PR #218, opened 2020-06-24
Move weight_decay handling from opt to layer

### LINKED ISSUE #213: Weight_decay handling should be done in `backwarding`
When weight decay is added to a layer, `Optimizer::calculate` takes `weight_decay` term and use that to update `djdw`

I think it should be done in the `layer::backwarding` phase and `optimizer` shouldn't care about `weight_decay` term as it is not about optimizing but updating derivatives.

for example `FullyConnectedLayer::backwarding` should be like

```cpp
FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
   if(should_apply_weight_decay) {
      djdw.add_i(weight, weight_decay.lambda);
   }

   /** rest of the calculation */
}
```

zhoonit:
- remove weight_decay from `Optimizer::calculate` signature
- fix bugs that weight_decay was applying twice
- apply weight decay to `fc_layer`

please note that `conv2d_layer::backwarding` also need to handle weight
decay after this PR is merged.

Resolves #213

**Self evaluation:**

### DIFF
--- a/nntrainer/src/fc_layer.cpp
+++ b/nntrainer/src/fc_layer.cpp
@@ -161,10 +161,11 @@ void FullyConnectedLayer::copy(std::shared_ptr<Layer> l) {
 
 Tensor FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
   Tensor ret = derivative.dot(weight.transpose("0:2:1"));
-  Tensor djdw = input.transpose("0:2:1").dot(derivative);
-
-  opt.calculate(djdw, derivative, weight, bias, iteration, this->init_zero,
-                weight_decay);
+  Tensor djdw = input.chain()
+                  .transpose("0:2:1")
+                  .dot(derivative)
+                  .applyIf(this->isWeightDecayL2Norm(), _LIFT(add_i), weight, weight_decay.lambda)
+                  .run();
 
   return ret;
 }
--- a/nntrainer/src/optimizer.cpp
+++ b/nntrainer/src/optimizer.cpp
@@ -77,26 +77,16 @@ int Optimizer::initialize(TensorDim d, bool set_tensor) {
   return status;
 }
 
-void Optimizer::calculate(Tensor &djdw, Tensor &djdb, Tensor &weight,
-                          Tensor &bias, int iteration, bool init_zero,
-                          WeightDecayParam weight_decay) {
+void Optimizer::calculate(const Tensor &djdw, const Tensor &djdb,
+                          Tensor &weight, Tensor &bias, int iteration,
+                          bool init_zero) {
   Tensor djdwAvg, djdbAvg;
-  if (weight_decay.type == WeightDecayType::l2norm) {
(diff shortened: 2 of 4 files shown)


nntrainer/src/fc_layer.cpp
```diff
@@ -200,11 +200,16 @@ Tensor FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
...
 
+  bool isL2norm = weight_decay.type == WeightDecayType::l2norm;
+
   Tensor ret = djdb.dot(weight.transpose("0:2:1"));
-  Tensor djdw = input.transpose("0:2:1").dot(djdb);
+  Tensor djdw = input.chain()
+                  .transpose("0:2:1")
+                  .dot(djdb)
+                  .applyIf(isL2norm, _LIFT(add_i), weight, weight_decay.lambda)
+                  .run();
```
jijoongmoon, line 211: Then, conv2d layer also have to handle weight decay by itself.

nntrainer/src/optimizer.cpp
```diff
@@ -77,26 +77,16 @@ int Optimizer::initialize(TensorDim d, bool set_tensor) {
...
-                          WeightDecayParam weight_decay) {
+void Optimizer::calculate(const Tensor &djdw, const Tensor &djdb,
+                          Tensor &weight, Tensor &bias, int iteration,
+                          bool init_zero) {
   Tensor djdwAvg, djdbAvg;
-  if (weight_decay.type == WeightDecayType::l2norm) {
-    djdw = djdw.add(weight.multiply(weight_decay.lambda));
-  }
-
   float ll = popt.learning_rate;
```
jijoongmoon, line 84: Good to remove this. this was duplication of weight decay.

nntrainer/src/fc_layer.cpp
```diff
@@ -200,11 +200,16 @@ Tensor FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
     }
   }
 
+  bool isL2norm = weight_decay.type == WeightDecayType::l2norm;
```
kparichay, line 203: how about add this as a common generic "protected" function in `layer.h`, and any function which has weight can use this?

zhoonit: Updated, thanks for the suggestion!

kparichay: I am sorry about unclear comment earlier. I meant applying weight decay - operation and weights.
Independent on the type of weights, the weight decay will be applied in the same way.
IMO, it would be better to write out the operation `.applyIf(this->isWeightDecayL2Norm(), _LIFT(add_i), weight, weight_decay.lambda)` once dependent on the `weight` as an argument.

zhoonit: Making a new operation something like `Tensor::applyWeightDecay(weightdecay)` comes to my mind.
Do you have any better idea?

PR #219, opened 2020-06-24
[bias] Bias updation missing for sgd

kparichay:
Bias updatation fixed for sgd where it only happened when bias was initialized with 0
For adam, bias updatation was happening twice

cc. @jijoongmoon Did I misinterpret a feature as a bug?


nntrainer/src/optimizer.cpp
```diff
@@ -102,6 +101,7 @@ void Optimizer::calculate(Tensor &djdw, Tensor &djdb, Tensor &weight,
   switch (type) {
   case OptType::sgd:
     weight.add_i(djdwAvg, -ll);
+    bias.add_i(djdbAvg, -ll);
     break;
```
jijoongmoon, line 105: Nice!

Issue #217, opened 2020-06-24
Add broadcasting support to element-wise operation to Tensor


zhoonit:
It would be better to support element wise operation with broadcasting support

zhoonit:
- Add stride / is_contiguous
- Make data vector shared_ptr
- Change setValue / getValue / print using stride
- Allow view of a current tensor

PR #220, opened 2020-06-24
Add stride and contiguous flag to tensor

zhoonit:
- Add stride
- Add checking is countigous flag to tensor

This patch is part of #217

**Self evaluation:**

### DIFF
--- a/nntrainer/include/tensor.h
+++ b/nntrainer/include/tensor.h
@@ -51,7 +51,7 @@ public:
   /**
    * @brief     Constructor of Tensor
    */
-  Tensor() : dim(){};
+  Tensor() : _is_contiguous(true), dim(){};
 
   /**
    * @brief     Constructor of Tensor with batch size one
@@ -331,6 +331,8 @@ public:
    */
   Tensor apply(std::function<Tensor(Tensor)> f) const;
 
+  Tensor apply_i(std::function<int(const Tensor &)> f) const;
+
   /**
    * @brief     Print element
    * @param[in] out out stream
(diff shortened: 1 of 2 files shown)


nntrainer/include/tensor.h
```diff
@@ -466,15 +468,30 @@ class Tensor {
    */
   int setDim(TensorDim d);
 
+  /**
+   * @brief     return if current tensor is contiguous, if not, you can't write
+   *            on this tensor
+   * @retval    bool is contigous
+   */
+  const bool isContiguous() const noexcept { return _is_contiguous; }
```
kparichay, line 476: can you elaborate on the use of this variable? 
Also, when is this false as all the constructors set them `true`.

nntrainer/include/tensor.h
```diff
@@ -51,7 +51,7 @@ class Tensor {
   /**
    * @brief     Constructor of Tensor
    */
-  Tensor() : dim(){};
+  Tensor() : _is_contiguous(true), dim(){};
```
dongju-chae, line 54: when `_is_contiguous` goes false?

zhoonit: It will become false when non-contiguous tensor is generated (tensor that refers to the non-sequential memory).

This PR is more of a preparation for #217. For now, every tensor is contiguous. Broadcasted tensor will be non-contiguous.

dongju-chae: Got it

nntrainer/include/tensor.h
zhoonit: It will be `false` when non-contiguous tensor is available. (such as broadcasted Tensor, or read-only transposed tensor)

I am planning to make `data` a shared pointer and reference it in diffre

While, I could've added more commit to show it's usage, I chose to split PR to prevent `n commits/1 PR`

kparichay: Got it.

PR #223, opened 2020-06-24
[activation] Derivative for activation

kparichay:
The derivative of softmax has been handcrafted to be different from others
Refer to https://github.com/nnstreamer/nntrainer/blob/2a650512813db6ce3bba828b5790066fbc655f14/nntrainer/src/fc_layer.cpp#L264-L267 for the original implementation
Softmax requires softmax(x) as input for derivative while other activations require x as input for derivative


kparichay:
@zhoonit Please verify once.

PR #224, opened 2020-06-24
[optimizer] Optimizer to manage list of weights @open sesame 06/25 13:00

kparichay:
Update optimizer to handle a list of weights than weights and bias individually
as layers like RNN, LSTM will have more than just (w,b)

### DIFF
--- a/nntrainer/include/optimizer.h
+++ b/nntrainer/include/optimizer.h
@@ -154,15 +154,13 @@ public:
   int initialize(TensorDim d, bool setTensor);
 
   /**
-   * @brief     calculate optimizer and Update Weight & Bais
-   * @param[in] dJdW Weight derivative
-   * @param[in] dJdB Bias derivative
-   * @param[in/out] Weight Weight Tensor
-   * @param[in/out] Bias Bias Tensor
+   * @brief     apply gradient to weights
+   * @param[in] weights vector of weights
+   * @param[in] gradients vector of corresponding gradients
    * @param[in] iteration nth epoch number
    */
-  void calculate(const Tensor &djdw, const Tensor &djdb, Tensor &weight,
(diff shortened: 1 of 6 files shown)


nntrainer/include/optimizer.h
```diff
@@ -161,8 +161,8 @@ class Optimizer {
    * @param[in/out] Bias Bias Tensor
    * @param[in] iteration nth epoch number
    */
-  void calculate(const Tensor &djdw, const Tensor &djdb, Tensor &weight,
-                 Tensor &bias, int iteration);
+  void calculate_gradients(std::vector<std::pair<Tensor, Tensor>> val_grad,
```
zhoonit, line 164: Would this be applicable to avoid memcopy and update the given weight?
```suggestion
  void calculate_gradients(std::vector<std::pair<Tensor & , const Tensor &>> val_grad,
```

kparichay: updated to `std::vector<Tensor> &weights, std::vector<Tensor> &gradients` as separate list of gradients will be available in each layer anyways in future.
Note: `std::vector<Tensor &>` cant be done as the type `Tensor &` wont be assignable.

kparichay: Updated to `std::vector<std::reference_wrapper<Tensor>> &weights, std::vector<std::reference_wrapper<Tensor>> &gradients` as per offline discussion.

jijoongmoon:
I think conv2d backwarding must be fixed. 
``` 
 opt.calculate(djdw, delBias[i], filters[i], bias[i], iteration);
````

kparichay:
> I think conv2d backwarding must be fixed.
> 
> ```
>  opt.calculate(djdw, delBias[i], filters[i], bias[i], iteration);
> ```

Applied.

PR #226, opened 2020-06-24
[wait for #224] [gradients] Get gradient of each layer

kparichay:
Added function to get the gradient of each layer
Each layer is responsible to fill in the list containing the gradients


nntrainer/src/optimizer.cpp
```diff
@@ -65,64 +65,71 @@ int Optimizer::initialize(TensorDim d, bool set_tensor) {
...
-    break;
-  }
-  case OptType::unknown:
-  default:
-    break;
+      wv.multiply_i(popt.beta2);
+      wv.add_i(x_grad.multiply(x_grad), 1 - popt.beta2);
+
+      x.add_i(wm.divide(wv.apply(sqrtEps)), -ll);
+      break;
```
jijoongmoon, line 124: bias update?

kparichay: This operates over the list of weights. So, the caller function will add all the weights and bias as a list
https://github.com/nnstreamer/nntrainer/blob/a112a6064633e7495fd5cdd882a59a1483d213d7/nntrainer/src/fc_layer.cpp#L175-L176

jijoongmoon: I see. I missed it.

nntrainer/src/fc_layer.cpp
```diff
@@ -168,6 +168,10 @@ Tensor FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
                   .applyIf(this->isWeightDecayL2Norm(), _LIFT(add_i), weight, weight_decay.lambda)
                   .run();
 
+  gradients.clear();
+  gradients.push_back(djdw);
+  gradients.push_back(djdb);
+
```
jijoongmoon, line 174: how do you manage the order of this gradients if there are multiple djdw and djdb. are you going to set the order like djdw[0], djdb[0], djdw[1], djdb[1]? or djdw[0], djdw[1], djdw[2], djdb[0], djdb[1]...? the former order is more proper I think. cause apply_gradient function asks this order.

kparichay: > how do you manage the order of this gradients if there are multiple djdw and djdb. are you going to set the order like djdw[0], djdb[0], djdw[1], djdb[1]? or djdw[0], djdw[1], djdw[2], djdb[0], djdb[1]...? 

Yes this will be the order.

nntrainer/include/layer.h
```diff
@@ -290,6 +290,14 @@ class Layer {
    */
   int setCost(CostType c);
 
+  const std::shared_ptr<std::vector<Tensor>> getGradients() {
+    std::vector<Tensor> grad;
+    for (int i =2; i< 0; i++)
+      grad.push_back(gradients[i]);
+
+    return std::make_shared<std::vector<Tensor>>(grad);
```
zhoonit, line 298: IMO we need to use `rvalue` to avoid copy.
```suggestion
    return std::make_shared<std::vector<Tensor>>(std::move(grad));
```

nntrainer/include/layer.h
```diff
@@ -402,6 +409,11 @@ class Layer {
   WeightDecayParam weight_decay;
 
   WeightIniType weight_ini_type;
+
+  /**
+   * @brief     Gradient for the weights in this layer
+   */
+  std::vector<std::reference_wrapper<Tensor>> gradients;
```
zhoonit, line 416: Not related to this PR though, how about moving `std::vector<Tensor> weights` out here so we can access to the current weights from out side as well?

nntrainer/include/layer.h
kparichay: may be copy is needed so that caller does not modify the internal copy?

nntrainer/include/layer.h
kparichay: good suggestion, will do in a separate PR, and add corresponding getter.

nntrainer/include/layer.h
kparichay: updated.

PR #228, opened 2020-06-24
[unittest] generate fc unittests @open sesame 06/30 15:31

kparichay:
Generate forward unittests for fully connected layer
Added minor bug fixes as well


test/input_gen/genInput.py
```diff
@@ -1,24 +1,38 @@
...
 import struct
+import random
+
+DEBUG = True
+
+# Fix the seeds across frameworks
+SEED = 1234
+tf.compat.v1.reset_default_graph()
+os.environ['TF_CUDNN_DETERMINISTIC'] = '1'
+os.environ['PYTHONHASSEED'] = str(SEED)
```
zhoonit, line 27: I guess this does not have any effect as [environment variable is processed in launch process](https://stackoverflow.com/questions/25684349/can-os-environpythonhashseed-be-set-dynamically-from-within-an-application).

test/input_gen/genInput.py
```diff
@@ -81,15 +93,137 @@ def conv2d_tensorflow(x, kernel, batch, width, height, channel, k_width, k_heigh
...
+                    raise 'unknown loss'
+
+                if opt == 'sgd':
+                    optimizer = tf.keras.optimizers.SGD()
+                elif opt == 'adam':
+                    optimizer = tf.keras.optimizers.Adam()
+                else:
+                    raise 'unknown optimizer'
+
+                def make_loss_callable(loss):
```
zhoonit, line 146: Q. is this function used somewhere?

test/input_gen/genInput.py
```diff
@@ -58,9 +70,9 @@ def gen_input(outfile_name, batch, channel, height, width):
 # @param[in] pad padding : SAME VALID FULL
 # @param[in] bias bias data
 # @return tf_o calculated result
-def conv2d_tensorflow(x, kernel, batch, width, height, channel, k_width, k_height, k_num, stride, pad, bias):
-    x_trans = np.transpose(x,[0,3,2,1])
-    kernel = np.transpose(kernel, [3,2,1,0])
+def conv2d_tf(x, kernel, batch, width, height, channel, k_width, k_height, k_num, stride, pad, bias):
+    x_trans = np.transpose(x,[0,2,3,1])
```
jijoongmoon, line 74: is this right? 2,3,1?

test/input_gen/genInput.py
kparichay: Yes, it's true. 
Removed setting `PYTHONHASHSEED` environment variable setting.
Moved `TF_CUDNN_DETERMINISTIC` environment variable before the TensorFlow is imported.

test/input_gen/genInput.py
jijoongmoon: Tensor order of tensorflow is NWHC, then. Got it. Thanks.

Issue #185, opened 2020-06-17
Adding copy and move constructor for tensor


kparichay:
In my understanding, `class tensor` does not has an explicitly written copy constructor. This leads to the use of the default copy constructor for itself and its elements. The default copy constructor for a `std::vector<>` copies all the elements inside the vector. As `std::vector<float> data` in `tensor` exists, a new vector is created and copied with each copy constructor use.

The repository code uses the copy constructor with `tensor` at many places (with direct uses such as `Tensor x = y`). This leads to unintentional data copy.

The same applies to move constructor.

jijoongmoon:
Right. I agreed.

zhoonit:
I am gonna be making copy assignment operator and move assignment operator

kparichay:
> I am gonna be making copy assignment operator and move assignment operator

If we have shared_ptr, IMO we do not need to write the copy and move assignment operator again.
@zhoonit close this once shared_ptr PR merges.

kparichay:
> I am gonna be making copy assignment operator and move assignment operator

If we have shared_ptr, IMO we do not need to write the copy and move assignment operator again.
@zhoonit close this once shared_ptr PR merges.

zhoonit:
I think we need a consensus on this topic.

You know in python family tensor library, `shared_ptr` kind of implementation is used. for example in numpy:

```python
>> import numpy as np
>> 
>> a = np.arange(5)
>> b = a
>> b[2] = -1
>> print(a)
array([ 0,  1, -1,  3,  4])
```

If you want to copy, you need to explicitly call copy like...

```python
>> import numpy as np
>> 
>> a = np.arange(5)
>> b = a.copy()
>> b[2] = -1
>> print(a)
array([ 0,  1, 2,  3,  4])
```

However, in c++ we have `reference operator` and `move assignment operator` in c++ which can do such job 

for example

```cpp
>> Tensor a = {1, 2, 3, 4, 5} // just an illustration.
>> Tensor &b = a;
>> b.getData()[2] = -1

>> std::cout << a;
1, 2, -1, 4, 5
```
or if you need to change the scope of a certain Tensor....
```cpp
>> Tensor a = {1, 2, 3, 4, 5}
>> Tensor b = std::move(a);
>> b.getData()[2] = -1;
```
Which python can't do.


So I think it is more natural to use copy assignment operator to copy the inner data structure as well (or use flag sth like copy-on-write?)




Another concern is that changing the semantics of copy operator as noted

[...]

PR #229, opened 2020-06-25
Change Tensor data to shared_ptr

zhoonit:
This PR proposes to change Tensor data to shared_ptr.

Part of #217

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -63,7 +63,7 @@ static auto rng = [] {
 
 Tensor::Tensor(const TensorDim d) {
   dim = d;
-  this->data = std::vector<float>(dim.getDataLen());
+  this->_data = std::make_shared<std::vector<float>>(dim.getDataLen());
   _is_contiguous = true;
   setZero();
 }
@@ -71,7 +71,7 @@ Tensor::Tensor(const TensorDim d) {
 Tensor::Tensor(int height, int width) {
   dim.height(height);
   dim.width(width);
-  this->data = std::vector<float>(dim.getDataLen());
+  this->_data = std::make_shared<std::vector<float>>(dim.getDataLen());
   _is_contiguous = true;
   setZero();
 }
@@ -80,7 +80,7 @@ Tensor::Tensor(int channel, int height, int width) {
   dim.height(height);
   dim.width(width);
   dim.batch(channel);
-  this->data = std::vector<float>(dim.getDataLen());
+  this->_data = std::make_shared<std::vector<float>>(dim.getDataLen());
   _is_contiguous = true;
   setZero();
 }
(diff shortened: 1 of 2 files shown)


kparichay:
In my understanding, making tensor data to shared_ptr changes the semantics of the copy constructor of the tensor. 
Without shared_ptr, a new copy of the vector data was being generated (using default copy constructor of vector). However, with shared_ptr, that copy is replaced with just a reference count increment (using default copy constructor of shared_ptr). It reduces memory copies :)
Please check that the usages of tensor in other files can work as per the updated copy constructor.

zhoonit:
> In my understanding, making tensor data to shared_ptr changes the semantics of the copy constructor of the tensor.
> Without shared_ptr, a new copy of the vector data was being generated (using default copy constructor of vector). However, with shared_ptr, that copy is replaced with just a reference count increment (using default copy constructor of shared_ptr). It reduces memory copies :)
> Please check that the usages of tensor in other files can work as per the updated copy constructor.

Ur right. I gonna right a test and check. Thanks :)

nntrainer/src/tensor.cpp
```diff
@@ -63,15 +63,15 @@ static auto rng = [] {
 
 Tensor::Tensor(const TensorDim d) {
   dim = d;
-  this->data = std::vector<float>(dim.getDataLen());
+  this->_data = std::make_shared<std::vector<float>>(dim.getDataLen());
```
dongju-chae, line 66: This is not related to this PR but I think you can declare and use a base constructor of `Tensor` initializing common variables.

zhoonit: That is actually a good idea. Would do in another PR. Thanks for the suggestion!

zhoonit:
> In my understanding, making tensor data to shared_ptr changes the semantics of the copy constructor of the tensor.
> Without shared_ptr, a new copy of the vector data was being generated (using default copy constructor of vector). However, with shared_ptr, that copy is replaced with just a reference count increment (using default copy constructor of shared_ptr). It reduces memory copies :)
> Please check that the usages of tensor in other files can work as per the updated copy constructor.

As this PR is not about changing the interface of Tensor. I am to make this PR keep the copying behavior as before(using copy / move constructor) and deal with the interface later after things are settle down in #185.  @kparichay would that be fine with you?

kparichay:
@zhoonit Please once confirm about this PR with @wooksong before moving forward.

zhoonit:
> @zhoonit Please once confirm about this PR with @wooksong before moving forward.

As far as i understood, his concern was: 
1. copying `vector<T>` is expensive compare to `vector <T *>` 
2. `vector<T *>`, `T *` could be a dangling pointer
3. elements in vector <T> should be destroyed correctly. (I checked and destructor of elements are called when vector is destoryed except the case of `2.` )

I think this should be considered when we expand tensor to another type. 

So far, I think his concern is not related to the PR. This PR aims to make `_data` accessible to other `tensor`.

zhoonit:
There were few concerns that this PR is potentially dangerous, I decided to implement this feature(sharing inner data) in other way.

Issue #233, opened 2020-06-25
Using standard exceptions internally


zhoonit:
Current exceptions are mostly done with int value.

With this strategy, stack unwinding is done at programmer's hand.

However, c++ has ability to throw exception and unwind call stack automatically until a certain function catches.
 
I think it is more safe to exploit this feature to prevent potential error.

For example. `TensorDim::setTensorDim`  throws `INVALID_PARAM` and this is hardly caught in the test (I am making up a PR for that) 

capi  binding can wrap a error handler function to catch std::exceptions and map `int`

Please review @jijoongmoon, @kparichay 

cf) https://en.cppreference.com/w/cpp/error/exception

kparichay:
I agree to using exceptions internally till handling the end layer of C where errno is needed.

kparichay:
Lets minimze internal c based error code use and possibly remove dependency from nntrainer_error.h and ml-api-common.h of nntrainer. C-API can depend on it.
This will be crucial for CC-API.

PR #237, opened 2020-06-25
Add exception boundary to capi @open sesame 07/03 11:09

zhoonit:
- Add a functor that returns `errno` to corresponding `exception` 

See also: #233 

**Self evaluation:**

### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -28,9 +28,41 @@
 #include <stdarg.h>
 #include <string.h>
 
+/**
+ * @brief     function to wrap an exception to predefined error value
+ * @param[in] func must be wrapped inside lambda []() -> int otherwise compile
+ * error will be raised
+ * @retval    errorno
+ */
+template <typename F> static int nntrainer_exception_boundary(F &&func) {
+  try {
+    return func();
+  } catch (std::invalid_argument &e) {
+    ml_loge("%s %s", typeid(e).name(), e.what());
+    return ML_ERROR_INVALID_PARAMETER;
+  } catch (std::range_error &e) {
+    ml_loge("%s %s", typeid(e).name(), e.what());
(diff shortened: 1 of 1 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -28,10 +28,42 @@
...
+ */
+template <typename F> static int nntrainer_exception_boundary(F &&func) {
+  try {
+    return func();
+  } catch (std::invalid_argument &e) {
+    ml_loge("%s %s", typeid(e).name(), e.what());
+    return ML_ERROR_INVALID_PARAMETER;
+  } catch (std::range_error &e) {
+    ml_loge("%s %s", typeid(e).name(), e.what());
+    return ML_ERROR_INVALID_PARAMETER;
```
kparichay, line 45: ML_ERROR_RESULT_OUT_OF_RANGE??

zhoonit: https://github.com/nnstreamer/nntrainer/blob/bb2dd7e383eda96d9502683b76c55ce72f9659f2/api/capi/include/nntrainer.h#L62-L72

I actually reverted that change because it was not defined and I was not sure if it is okay to change `ml_error_e`.
@jijoongmoon  could we add another ML_ERROR to `capi`?

kparichay: uhh, it seems like nntrainer_error.h and nntrainer.h are not in sync.

jijoongmoon: This is for the tizen api so have to sync accrodingly.

myungjoo: You are breaking the compatibility with Tizen ML API sets here.

It's time to consider using nnstreamer-capi-devel.rpm to get error-code header files (ml-api-common.h)

zhoonit: Added #243 to resolve this issue.

Issue #239, opened 2020-06-26
Refactor ini file configuration


kparichay:
ini file has some redundancies as well as ambiguities.
Refactor ini file configuration.

kparichay:
- [ ] Update ini parsing - #240
- [ ] Make ini variables more intuitive
- [ ] Make flatten as a parameter than layer
- [ ] Simplify `neuralnet::init`

kparichay:
Supporting skip connection would look something like this:
```diff
 # Layer Section : Name
 [inputlayer]
 Type = input
 HiddenSize = 128	# Input Layer Dimension
 Activation = sigmoid
 
 [fc1layer]
 Type = fully_connected
 HiddenSize = 20
 Activation = sigmoid
 
 [fc2layer]
+InputLayers = fc1layer                 # Regular connection - redundant
 Type = fully_connected
 HiddenSize = 20
 Activation = sigmoid
 
 [fc3layer]
+InputLayers = fc1layer , fc2layer       # Skip connection with concat operation
 Type = fully_connected
 HiddenSize = 20
 Activation = sigmoid
 
 [outputlayer]
+InputLayers = fc1layer + fc3_layer     # Skip connection with addition operation
 Type = output
 HiddenSize = 3
 Activation = sigmoid
```

Original Network
Input -- fc1 -- fc2 -- fc3 -- Output

New Network with skip connection
```
Input -- fc1 -- fc2 --- =fc3 -- + --Output
             |____ | ___|            |
                       |__________|
```

The intention is to support skip connection with element-wise addition or contact operation intuitively without needing the user to add another layer for this purpose. Internally, this would add `elementwise-add` or `concat` layer depending on the type of skip connection used. 
@jijoongmoon @zhoonit @myungjoo Please make suggestions.

myungjoo:
```
Input -- fc1 -- fc2 --+- fc3 --+- Output
          |___________|        |
          |____________________|
```
?

Looks good to me.

When we allow to use .circle format, we will still need this .ini file to supply training info for "deployed (inference-only) model" to retrain (adjust weights, add layers, ...). We may need a small workshop to share the vision of nntrainer 2021-2022 next week or later (w/ all related members)

kparichay:
Right, my diagram got a bit buggy.
The below will be supported.
```
Input -- fc1 -- fc2 --+- fc3 --+- Output
          |___________|        |
          |____________________|
```

However the example is actually for below:
```
[...]

PR #240, opened 2020-06-26
[ini] Update ini parsing and format

kparichay:
Update the parsing and format of ini input file

Related issue : #239 


test/unittest/unittest_nntrainer_internal.cpp
```diff
@@ -153,7 +152,7 @@ TEST(nntrainer_NeuralNetwork, init_07_p) {
   status = NN.setConfig("./test.ini");
   EXPECT_EQ(status, ML_ERROR_NONE);
   status = NN.init();
-  EXPECT_EQ(status, ML_ERROR_NONE);
+  EXPECT_EQ(status, ML_ERROR_INVALID_PARAMETER);
```
kparichay, line 155: @jijoongmoon please verify is this should fail or pass. Passing no `TrainData` is acceptable?

kparichay: Can you also clarify on `LabelData`, it's used by all train/valid/test it seems, That's a bit odd.

zhoonit:
Could you elaborate As-is / To-be of this PR if possible?
Thanks!

nntrainer/src/neuralnet.cpp
```diff
@@ -124,118 +124,137 @@ int NeuralNetwork::setConfig(std::string config) {
...
-    NN_INI_RETURN_STATUS();
-
-  } else if (iniparser_find_entry(ini, "DataSet:Tflite")) {
-    ml_loge("Error: Not yet implemented!");
-    return ML_ERROR_INVALID_PARAMETER;
-  } else {
-    data_buffer = std::make_shared<DataBufferFromCallback>();
+  /** Parse the DataSet section */
+  data_buffer = std::make_shared<DataBufferFromCallback>();
+  section_names_iter =
```
jijoongmoon, line 204: This 203 line has conflict with line 212

nntrainer/src/neuralnet.cpp
```diff
@@ -124,118 +124,137 @@ int NeuralNetwork::setConfig(std::string config) {
...
+      NN_INI_RETURN_STATUS();
+    }
   }
 
+  /** Parse all the layers defined as sections in order */
   TensorDim previous_dim;
-  for (unsigned int i = 0; i < layers_name.size(); i++) {
+  for (section_names_iter = section_names.begin();
+      section_names_iter != section_names.end(); ++section_names_iter) {
     bool last = false;
```
jijoongmoon, line 235: So, first layer should go first and last layer should go last, then. Strictly follow in order.

nntrainer/src/neuralnet.cpp
```diff
@@ -948,44 +1007,43 @@ int NeuralNetwork::addLayer(std::shared_ptr<Layer> layer) {
...
-  ActiType act = layer->getActivationType();
-  if (act != ACT_UNKNOWN) {
-    if (layers.back()->getType() == LAYER_ACTIVATION) {
-      /* possible throw here? */
-      ml_logw("Activation defined right after activation");
-    }
+std::shared_ptr<Layer> NeuralNetwork::_make_act_layer(ActiType act) {
+  if (layers.back()->getType() == LAYER_ACTIVATION) {
+    /** User defined activation layer. Do not add another activation layer after this */
+    ml_loge("Error: double activation layers.");
```
jijoongmoon, line 1013: Could you talk to @zhoonit about it? He has intention to have warning instead of error.

test/unittest/unittest_nntrainer_internal.cpp
jijoongmoon: 1. If you mean the TrainData in ini file,  it's acceptable. We have two options to get training data. one is from file. and the other is from callback which defined by user.
2. Label Data keyword is used to calculate loss. Is there any other way to get the label data?

jijoongmoon: Discussed off-line and make agreement.

nntrainer/src/neuralnet.cpp
zhoonit: NBD, I think this can easily be fixed it it becomes a requirement :)

nntrainer/include/layer.h
```diff
@@ -123,7 +123,7 @@ class Layer {
       type(LAYER_UNKNOWN),
       loss(0.0),
       cost(COST_UNKNOWN),
-      activation_type(ACT_UNKNOWN),
+      activation_type(ACT_NONE),
```
zhoonit, line 126: Please note that, ACT_NONE will set an layer with identity activation function w.r.t. current implementation of `activation layer` in `neuralnet::_make_act_layer`

nntrainer/src/neuralnet.cpp
```diff
@@ -550,30 +585,51 @@ int NeuralNetwork::init(std::shared_ptr<Optimizer> optimizer,
   status = setProperty(arg_list);
   NN_RETURN_STATUS();
 
-  loss = 10000.0;
+  /** Note: number of entries in layers will change. */
```
zhoonit, line 588: This is good

nntrainer/include/layer.h
kparichay: Right, I updated to result in an error if the activation type is unknown.
IMO itis better to make it a no-op (no activation) than keep it unknown. Regarding the overhead, unit layers can be optimized later.
Can update it to `unknown` if you think its better.

nntrainer/src/neuralnet.cpp
kparichay: In my understanding, this situation only happens due to internal bug as user cannot request for a separate LAYER_ACTIVATION as of now. So, I changed this to error.

nntrainer/src/neuralnet.cpp
kparichay: Yes, just like the older ini format required `Layers` to give the order of layers.
```
Layers = inputlayer \
         fc1layer \
         batchnormalization \
        outputlayer    #Layers of Neuralnetwork
```

nntrainer/src/neuralnet.cpp
kparichay: Yes, this has been updated.

nntrainer/include/layer.h
zhoonit: I think `ACT_NONE` is better :)

nntrainer/src/neuralnet.cpp
kparichay: Yes, this is the location where the order of layers is defined.
Please refer https://github.com/nnstreamer/nntrainer/pull/240#issuecomment-651039288 for detailed example.

PR #242, opened 2020-06-26
[optimizer] Bug fix

### LINKED ISSUE #241: Optimizer doesn't update weights
The updated optimizer creates list of weight references that are sent to the optimizer.
However, the copy constructor/copy assignment operator for vector in tensor copies the whole data if not handled carefully.
Hence, the weights processed by the optimizer were just a copy than the original weights, and the updates never reach the original weights.

kparichay:
Tensor copy constructor and copy assigment operator creates a copy of the vector
This led to bug in optimizer which updated the copy of the weight than the weight itself
Fixed by refernce of weight in optimizer

Resolves #241


zhoonit:
BTW,  `*.in` is going to be ignored in #228. Is addition of `*.in` in this PR intended?

kparichay:
> BTW, `*.in` is going to be ignored in #228. Is addition of `*.in` in this PR intended?

ohh yes, updated.

PR #245, opened 2020-07-01
Restructure inner data structure inside Tensor

zhoonit:
This PR restructures Tensor to enable sharing between Tensor. as well as refactoring the inner data structure to not use vector also providing significant speed increases(20s -> 10s in example application)

- Change inner data structure
- Refactor Tensor ctors
- Add copy/move ctors & assignment operators

**Self evaluation:**

### DIFF
--- a/nntrainer/include/tensor.h
+++ b/nntrainer/include/tensor.h
@@ -31,13 +31,13 @@ extern "C" {
 }
 #endif
 
+#include <array>
 #include <cmath>
 #include <fstream>
 #include <iostream>
 #include <memory>
 #include <regex>
 #include <tensor_dim.h>
-#include <vector>
 
 namespace nntrainer {
 
@@ -48,31 +48,33 @@ class LazyTensor;
  */
 class Tensor {
 public:
-  /**
-   * @brief     Constructor of Tensor
-   */
-  Tensor() : _is_contiguous(true), dim(){};
+  Tensor(const TensorDim &d, float *buf = nullptr) :
+    dim(d),
+    strides{{1, 2, 3}},
+    is_contiguous(true),
+    data(new float[d.getDataLen()], std::default_delete<float[]>())
+    {
+    // todo: initialize appropriate strides
+    if (buf == nullptr) {
+      setZero();
+    } else {
+      float *data = getData();
+      unsigned int len = length();
+
+#ifdef USE_BLAS
+      cblas_scopy(len, buf, 1, data, 1);
+#else
+      for (unsigned int i = 0; i < len; ++i) {
+        data[i] = buf[i];
+      }
+#endif
+    }
+  }
 
   /**
-   * @brief     Constructor of Tensor with batch size one
-   * @param[in] dim TensorDim
+   * @brief     Basic Constructor of Tensor
    */
-  Tensor(const TensorDim dim);
-
-  /**
-   * @brief     Constructor of Tensor with batch size one
-   * @param[in] heihgt Height of Tensor
-   * @param[in] width Width of Tensor
-   */
-  Tensor(int height, int width);
-
-  /**
-   * @brief     Constructor of Tensor
-   * @param[in] channel Channel of Tensor
-   * @param[in] heihgt Height of Tensor
-   * @param[in] width Width of Tensor
-   */
-  Tensor(int channel, int height, int width);
+  Tensor() : Tensor(TensorDim()){};
 
   /**
    * @brief     Constructor of Tensor
@@ -81,19 +83,24 @@ public:
    * @param[in] heihgt Height of Tensor
    * @param[in] width Width of Tensor
    */
-  Tensor(int batch, int channel, int height, int width);
-
-  /**
-   * @brief   Constructor of Tensor
-   * @param[in] d data for the Tensor with batch size one
-   */
-  Tensor(std::vector<std::vector<float>> const &d);
+  Tensor(int batch, int channel, int height, int width) :
+    Tensor(TensorDim(batch, channel, height, width)){};
 
   /**
    * @brief     Constructor of Tensor
-   * @param[in] d data for the Tensor
+   * @param[in] channel Channel of Tensor
+   * @param[in] heihgt Height of Tensor
+   * @param[in] width Width of Tensor
    */
-  Tensor(std::vector<std::vector<std::vector<float>>> const &d);
+  Tensor(int channel, int height, int width) :
+    Tensor(1, channel, height, width){};
+
+  /**
+   * @brief     Constructor of Tensor with batch size one
+   * @param[in] heihgt Height of Tensor
+   * @param[in] width Width of Tensor
+   */
+  Tensor(int height, int width) : Tensor(1, 1, height, width){};
 
   /**
    * @brief     Constructor of Tensor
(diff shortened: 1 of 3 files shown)


nntrainer/include/tensor.h
```diff
@@ -48,70 +48,120 @@ class LazyTensor;
...
 class Tensor {
 public:
-  /**
-   * @brief     Constructor of Tensor
-   */
-  Tensor() : _is_contiguous(true), dim(){};
+  Tensor(const TensorDim &d, float *buf = nullptr) :
+    dim(d),
+    _is_contiguous(true),
+    _data(new float[d.getDataLen()], [](float *p) { delete[] p; }) {
```
kparichay, line 54: how about using std::default_delete<float[]>() ?

nntrainer/src/tensor.cpp
```diff
@@ -143,38 +131,15 @@ void Tensor::setRandUniform(float min, float max) {
...
-    for (unsigned int k = 0; k < dim.height(); ++k)
-      for (unsigned int l = 0; l < dim.width(); ++l)
-        this->setValue(0, j, k, l, d[j][k][l]);
-}
-
 Tensor::Tensor(
   std::vector<std::vector<std::vector<std::vector<float>>>> const &d) {
 
   dim.batch(d.size());
   dim.channel(d[0].size());
```
kparichay, line 138: before accessing d[0] and so forth, lets ensure that the vector isn't  empty.

nntrainer/include/tensor.h
```diff
@@ -491,20 +539,26 @@ class Tensor {
...
-  bool _is_contiguous;
+  /**
+   * @brief     return Data pointer of Tensor
+   * @retval    float pointer
+   */
+  float *getData() { return _data.get(); }
+
+  /**< handle the data as a std::shared_ptr<float> type */
   TensorDim dim;
+  std::array<int, MAXDIM> _strides;
```
kparichay, line 553: any reason to use `_` as a prefix to all these variables?

nntrainer/include/tensor.h
```diff
@@ -48,70 +48,120 @@ class LazyTensor;
...
+   *  @param[in] Tensor &&
+   */
+  Tensor(Tensor &&rhs) noexcept = default;
+
+  /**
+   * @brief  Copy assignment operator.
+   * @param[in] rhs Tensor to be copied.
+   */
+  // todo: refactor operator= to consider allocated size for the data
+  Tensor &operator=(Tensor &rhs);
```
kparichay, line 146: `const Tensor &rhs`??

nntrainer/src/tensor.cpp
```diff
@@ -61,50 +61,36 @@ static auto rng = [] {
...
-}
-
-Tensor::Tensor(int height, int width) {
-  dim.height(height);
-  dim.width(width);
-  this->data = std::vector<float>(dim.getDataLen());
-  _is_contiguous = true;
-  setZero();
+Tensor &Tensor::operator=(Tensor &rhs) {
+  Tensor tmp(rhs);
```
kparichay, line 65: is this extra?

nntrainer/include/tensor.h
zhoonit: Fixed. Thanks :)

nntrainer/src/tensor.cpp
zhoonit: Sure. Thank you!

nntrainer/include/tensor.h
zhoonit: It is easier to tell private member variable from public member variables as well as adding some kind of awareness that you are using private variable.

AFAIK, variable name policy hasn't been discussed w.r.t. this matter and open to opinion about this :)

nntrainer/src/tensor.cpp
zhoonit: Implemented incorrectly. Now it is fixed. Thanks.

nntrainer/src/tensor.cpp
```diff
@@ -61,50 +61,36 @@ static auto rng = [] {
...
-Tensor::Tensor(int channel, int height, int width) {
-  dim.height(height);
-  dim.width(width);
-  dim.batch(channel);
-  this->data = std::vector<float>(dim.getDataLen());
-  _is_contiguous = true;
-  setZero();
-}
+Tensor &Tensor::operator=(Tensor &&rhs) noexcept {
+  std::swap(dim, rhs.dim);
```
kparichay, line 71: this works too but its recommended to create a internal swap function for this.

zhoonit: AFAIK, from c++11 it is find to use `std::swap` because it does not use copy constructors if possible which wasn't the case before c++11. 
However, I think I should add move ctor in `TensorDim` later on. 
Thanks for pointing out and please tell me if I got something wrong!

cf) http://www.cplusplus.com/reference/algorithm/swap/

kparichay: yes, you are right.
what I meant what create a function `void Tensor::swap(Tensor &lhs, Tensor &rhs)`.

nntrainer/include/tensor.h
jijoongmoon: let's do not use _. looks more complicated.

nntrainer/src/tensor.cpp
```diff
@@ -61,50 +61,36 @@ static auto rng = [] {
...
+
+  if (len != rhs.length())
     return false;
 
-  for (size_t i = 0; i < data.size(); ++i) {
-    if (std::isnan(data[i]) || std::isnan(rhs.data[i]) ||
-        std::fabs(data[i] - rhs.data[i]) > epsilon)
+  const float *data = getData();
+  const float *rdata = getData();
+
```
jijoongmoon, line 90: ``` *rdata = rhs.getData() ``` ??

nntrainer/src/tensor.cpp
```diff
@@ -113,22 +99,24 @@ bool Tensor::operator== (const Tensor &rhs) const {
 
 float Tensor::getValue(unsigned int batch, unsigned int c, unsigned int h,
                        unsigned int w) const {
-  return this->data[batch * dim.channel() * dim.height() * dim.width() +
-                    c * dim.height() * dim.width() + h * dim.width() + w];
+  return getData()[batch * dim.channel() * dim.height() * dim.width() +
+                   c * dim.height() * dim.width() + h * dim.width() + w];
```
jijoongmoon, line 103: how about ``` getData()[batch*dim.getFeatureLen() + c*dim.height()*dim.width() + h* dim.width() + w] ``` ??

nntrainer/src/tensor.cpp
zhoonit: Added :) Thanks.

nntrainer/src/tensor.cpp
zhoonit: Right. Thanks!

nntrainer/src/tensor.cpp
zhoonit: Yes it is better. Thank you!

nntrainer/include/tensor.h
zhoonit: Updated.

PR #247, opened 2020-07-02
[activation] Add missing config for activation layer

kparichay:
Add missing initialization and setting input/output dimension for activation layer
Also updated setting previous dimension to be from the last computation layer

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -431,8 +431,6 @@ int NeuralNetwork::init() {
       NN_INI_RETURN_STATUS();
       break;
     }
-    previous_dim = layers.back()->getOutputDimension();
-
     const char *acti_str = iniparser_getstring(
       ini, (layer_name + ":Activation").c_str(), unknown);
     ActiType act = (ActiType)parseType(acti_str, TOKEN_ACTI);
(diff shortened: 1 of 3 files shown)


nntrainer/src/neuralnet.cpp
```diff
@@ -431,15 +431,15 @@ int NeuralNetwork::init() {
...
-    previous_dim = layers.back()->getOutputDimension();
-
     const char *acti_str = iniparser_getstring(
       ini, (layer_name + ":Activation").c_str(), unknown);
     ActiType act = (ActiType)parseType(acti_str, TOKEN_ACTI);
 
     layers.back()->setActivation(act);
     status = initActivationLayer(act);
     NN_INI_RETURN_STATUS();
+
```
jijoongmoon, line 441: Some layers do not need the activation such as pooling, flatten. No need to check every layer.

nntrainer/src/neuralnet.cpp
```diff
@@ -1021,24 +1022,21 @@ std::shared_ptr<Layer> NeuralNetwork::_make_act_layer(ActiType act) {
       std::make_shared<ActivationLayer>();
 
     act_layer->setActivation(act);
+    act_layer->setInputDimension(prev->getOutputDimension());
```
jijoongmoon, line 1025: In Layer, if ActiType == ACT_UNKNOW, then there will be error. But there is an layer which does not need activation such as flatten, pooling. I think we have to consider it.

``` int Layer::setActivation(ActiType acti) {
  int status = ML_ERROR_NONE;
  if (acti == ACT_UNKNOWN) {
    ml_loge("Error:have to specify activation function");
  }
  activation_type = acti;

  return status;
}```

PR #248, opened 2020-07-02
[ini] update ini bias init and flatten as feature

kparichay:
bias init name is changes to bias_init_zero to make it more readable
flatten is now a layer feature externally rather than as a new layer itself
    

nntrainer/src/fc_layer.cpp
```diff
@@ -84,14 +84,18 @@ int FullyConnectedLayer::setProperty(std::vector<std::string> values) {
...
+    case PropertyType::bias_init_zero: {
+      status = setBoolean(this->bias_init_zero, value);
       NN_RETURN_STATUS();
     } break;
     case PropertyType::activation:
       status = setActivation((ActiType)parseType(value, TOKEN_ACTI));
       NN_RETURN_STATUS();
       break;
+    case PropertyType::flatten:
+      status = setBoolean(flatten, value);
```
jijoongmoon, line 96: I wonder that how it works in fc layer if this flatten is true? maybe do nothing?
And I think we also have to add this flatten property into pooling2D Layer as well.

kparichay: Yes, in the cases where the flattening is not needed, then flattening will not do anything.
Will add to pooling2D layer too.
However, flattening can be needed for FC layer too - FC layer input can be of sort (3,2,1,4) with kernel (4,5) and output (3,2,1,5). So, in this case, flattening will produce a result of (3,1,1,10).

PR #249, opened 2020-07-02
[ NETWORK ] initialize network for conv2d/pooling2d/flatten @open sesame 07/02 19:47

jijoongmoon:
This PR provides initialization of conv2d, pooling2d, flatten layer
for training configuraion file.

Also modified to update mutiple weights in the optimizer.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/optimizer.h
+++ b/nntrainer/include/optimizer.h
@@ -158,6 +158,17 @@ public:
   int initialize(TensorDim d, bool setTensor);
 
   /**
+   * @brief     initialize optimizer. Initialize Weight if it is adam
+   * @param[in] list_d TensorDim list (in an order of W0, B0, W1, B1, etc.)
+   * @param[in] setTensor true if the layer need weight update.
+   *            Input Layer and Batch Normalization layer won't need it.
+   *            Therefore, it sets false.
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   */
+  int initialize(std::vector<Tensor> list_d, bool setTensor);
+
+  /**
    * @brief     apply gradient to weights
    * @param[in] weights vector of weights
    * @param[in] gradients vector of corresponding gradients
--- a/nntrainer/src/conv2d_layer.cpp
+++ b/nntrainer/src/conv2d_layer.cpp
@@ -117,7 +117,17 @@ int Conv2DLayer::setOptimizer(Optimizer &opt) {
   if (status != ML_ERROR_NONE)
     return status;
 
-  return this->opt.initialize(dim, true);
+  std::vector<Tensor> list_d;
+  for (unsigned int i = 0; i < filter_size; ++i) {
+    for (unsigned int j = 0; j < 2; ++j) {
+      list_d.push_back(Tensor(dim));
+    }
+    for (unsigned int j = 0; j < 2; ++j) {
+      list_d.push_back(Tensor(1, 1, 1));
+    }
+  }
+
+  return this->opt.initialize(list_d, true);
 }
 
 Tensor Conv2DLayer::backwarding(Tensor derivative, int iteration) {
--- a/nntrainer/src/flatten_layer.cpp
+++ b/nntrainer/src/flatten_layer.cpp
@@ -34,12 +34,11 @@ int FlattenLayer::initialize(bool last) {
   output_dim.height(1);
   output_dim.width(input_dim.getFeatureLen());
 
-  hidden = Tensor(output_dim);
-
   return status;
 }
 
 Tensor FlattenLayer::forwarding(Tensor in, int &status) {
+  hidden = Tensor(output_dim);
   memcpy(hidden.getData(), in.getData(),
          in.getDim().getDataLen() * sizeof(float));
 
@@ -56,6 +55,10 @@ Tensor FlattenLayer::backwarding(Tensor in, int iteration) {
   return ret;
 }
 
+int FlattenLayer::setProperty(std::vector<std::string> values) {
+  return ML_ERROR_NOT_SUPPORTED;
+};
+
 void FlattenLayer::copy(std::shared_ptr<Layer> l) {
   std::shared_ptr<FlattenLayer> from =
     std::static_pointer_cast<FlattenLayer>(l);
@@ -65,10 +68,6 @@ void FlattenLayer::copy(std::shared_ptr<Layer> l) {
   this->input_dim = from->input_dim;
   this->output_dim = from->output_dim;
   this->last_layer = from->last_layer;
-}
-
-int FlattenLayer::setProperty(std::vector<std::string> values) {
-  return ML_ERROR_NOT_SUPPORTED;
 }
 
 } /* namespace nntrainer */
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -58,7 +58,7 @@ static bool is_file_exist(std::string file_name) {
 }
 
 static int parseWeightDecay(dictionary *ini, std::string layer_name,
-                          WeightDecayParam &weight_decay) {
+                            WeightDecayParam &weight_decay) {
   char unknown[] = "Unknown";
   int status = ML_ERROR_NONE;
   weight_decay.type = (WeightDecayType)parseType(
@@ -113,7 +113,6 @@ int NeuralNetwork::setConfig(std::string config) {
     ml_loge("Error: Cannot open model configuration file");
     return ML_ERROR_INVALID_PARAMETER;
   }
-
   this->config = config;
 
   return status;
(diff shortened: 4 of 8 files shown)


nntrainer/src/conv2d_layer.cpp
```diff
@@ -111,7 +114,17 @@ int Conv2DLayer::setOptimizer(Optimizer &opt) {
...
+  for (unsigned int i = 0; i < filter_size; ++i) {
+    for (unsigned int j = 0; j < 2; ++j) {
+      list_d.push_back(Tensor(dim));
+    }
+    for (unsigned int j = 0; j < 2; ++j) {
+      list_d.push_back(Tensor(1, 1, 1));
+    }
+  }
+
+  return this->opt.initialize(list_d, true);
```
kparichay, line 127: How about directly passing the `weights` from `Layer`. This will be available once you rebase.

nntrainer/src/neuralnet.cpp
```diff
@@ -431,15 +486,18 @@ int NeuralNetwork::init() {
...
     }
-    const char *acti_str = iniparser_getstring(
-      ini, (layer_name + ":Activation").c_str(), unknown);
-    ActiType act = (ActiType)parseType(acti_str, TOKEN_ACTI);
+    previous_dim = layers.back()->getOutputDimension();
 
-    layers.back()->setActivation(act);
-    status = initActivationLayer(act);
-    NN_INI_RETURN_STATUS();
+    if (layer_type != LAYER_IN && layer_type != LAYER_POOLING2D &&
```
kparichay, line 491: Do we need this condition here? 
IMO, the activation layer should be able to add at the end of any layer as our implementation allows it.
This might add `ACT_NONE` in some of the cases which we can optimize out later.

nntrainer/src/neuralnet.cpp
```diff
@@ -918,7 +1016,7 @@ int NeuralNetwork::train_run() {
           return status;
         }
         std::cout << "#" << i + 1 << "/" << epoch;
-        data_buffer->displayProgress(iter, nntrainer::BUF_TRAIN, getLoss());
+        data_buffer->displayProgress(count++, nntrainer::BUF_TRAIN, getLoss());
```
kparichay, line 1019: Can you elaborate on the usage of `count` as I think `iter` does the same job.
`iter` tracks class state and can be stored to continue the training.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -111,7 +114,17 @@ int Conv2DLayer::setOptimizer(Optimizer &opt) {
   if (status != ML_ERROR_NONE)
     return status;
 
-  return this->opt.initialize(dim, true);
+  std::vector<Tensor> list_d;
+  for (unsigned int i = 0; i < filter_size; ++i) {
+    for (unsigned int j = 0; j < 2; ++j) {
```
kparichay, line 119: Does this loop of 2 creates `wm, wv` pair? If yes, lets do this in the optimizer than here.

nntrainer/include/optimizer.h
```diff
@@ -153,6 +153,17 @@ class Optimizer {
    */
   int initialize(TensorDim d, bool setTensor);
 
+  /**
+   * @brief     initialize optimizer. Initialize Weight if it is adam
+   * @param[in] list_d TensorDim list
```
kparichay, line 158: Lets add some comment on the order of elements this list should maintain (like W0, B0, W1, B1 if applicable etc)

nntrainer/src/neuralnet.cpp
jijoongmoon: This is show how many data is calculated in every epoch.

jijoongmoon: this counter is used for counting how many data is calculated in every epoch. However, iter is used to update decay.

kparichay: Got it.

nntrainer/src/neuralnet.cpp
jijoongmoon: After rebase, this can be removed.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: We need to fix some issues related with weights. There will be garbage when we set the weights. This will be done next PR.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: Keep it for now. when we fix the weights issue, then we can replace it.

nntrainer/include/optimizer.h
jijoongmoon: done.

nntrainer/src/flatten_layer.cpp
```diff
@@ -34,12 +34,12 @@ int FlattenLayer::initialize(bool last) {
   output_dim.height(1);
   output_dim.width(input_dim.getFeatureLen());
 
-  hidden = Tensor(output_dim);
-
   return status;
 }
 
 Tensor FlattenLayer::forwarding(Tensor in, int &status) {
+  hidden = Tensor(in.batch(), output_dim.channel(), output_dim.height(),
```
kparichay, line 41: hidden = Tensor(output_dim) ??

nntrainer/src/neuralnet.cpp
```diff
@@ -346,13 +348,69 @@ int NeuralNetwork::init() {
...
+                  (int *)size);
+
+      NN_INI_RETURN_STATUS();
+      status =
+        pooling2d_layer->setSize(size, Layer::PropertyType::pooling_size);
+
+      NN_INI_RETURN_STATUS();
+      status = getValues(
+        POOLING2D_DIM,
+        iniparser_getstring(ini, (layer_name + ":stride").c_str(), unknown),
```
kparichay, line 380: lets not use unknown, but sensible defaults - getValues({1, 1}).

nntrainer/src/neuralnet.cpp
```diff
@@ -346,13 +348,69 @@ int NeuralNetwork::init() {
...
+          iniparser_getstring(ini, (layer_name + ":pooling").c_str(), unknown),
+          TOKEN_POOLING));
+
+      status = pooling2d_layer->initialize(last);
+      NN_INI_RETURN_STATUS();
+      layers.push_back(pooling2d_layer);
+    } break;
+
+    case LAYER_FLATTEN: {
+      std::shared_ptr<FlattenLayer> flatten_layer =
```
kparichay, line 404: replace with 
```
status = initFlattenLayer();
NN_INI_RETURN_STATUS();
```

nntrainer/src/pooling2d_layer.cpp
```diff
@@ -52,6 +50,8 @@ int Pooling2DLayer::initialize(bool last) {
 }
 
 Tensor Pooling2DLayer::forwarding(Tensor in, int &status) {
+  hidden = Tensor(in.batch(), output_dim.channel(), output_dim.height(),
```
kparichay, line 53: same here - `hidden = Tensor(output_dim)`

nntrainer/src/neuralnet.cpp
jijoongmoon: But at this time, we do not know the layers[position]->getOuptutDimension() because this is the independent layer.

nntrainer/src/flatten_layer.cpp
jijoongmoon: right. fixed.

nntrainer/src/neuralnet.cpp
jijoongmoon: fixed. thanks.

nntrainer/src/pooling2d_layer.cpp
jijoongmoon: fixed. thanks.

nntrainer/src/neuralnet.cpp
kparichay: We know that this will not be the first layer. Hence, `layers[position]->getOutputDimension()` will exist as with `initFlattenLayer()`, position is set to be the last layer in layers.

kparichay: Got it.

nntrainer/src/optimizer.cpp
```diff
@@ -58,6 +58,30 @@ int Optimizer::setOptParam(OptParam p) {
...
+        tensors[i].channel() == 0) {
+      ml_loge("Error: Tensor Dimension must be greater than 0");
+      return ML_ERROR_INVALID_PARAMETER;
+    }
+
+    if (tensors[i + 1].height() == 0 || tensors[i + 1].width() == 0 ||
+        tensors[i + 1].channel() == 0) {
+      ml_loge("Error: Tensor Dimension must be greater than 0");
+      return ML_ERROR_INVALID_PARAMETER;
+    }
```
zhoonit, line 77: IMO. this is not necessary as tensorDim is using c++ style throw to check tensordim patched at https://github.com/nnstreamer/nntrainer/pull/234/commits/7a70a1fcc85b1f676171fe667852470c203feeac

jijoongmoon: Right. fixed. Thanks.

PR #250, opened 2020-07-02
[ MNIST ] mnist application

jijoongmoon:
This PR provides mnist application which includes:
. 1 Input Layer
. 2 Conv2d
. 2 Pooling2d
. 1 Flatten
. 1 Fully Connected Layer

**Self evaluation:**


Applications/mnist/jni/main.cpp
```diff
@@ -0,0 +1,293 @@
...
+ * @retval true/false false : end of data
+ */
+bool getMiniBatch_val(float *outVec, float *outLabel, int *status) {
+
+  std::vector<int> memI;
+  std::vector<int> memJ;
+  unsigned int count = 0;
+  int data_size = total_val_data_size;
+
+  std::string filename = "mnist_trainingSet.dat";
```
zhoonit, line 203: Could you check if this is intended?

Applications/mnist/jni/main.cpp
```diff
@@ -0,0 +1,293 @@
...
+bool valduplicate[total_label_size * total_val_data_size];
+
+
+/**
+ * @brief     step function
+ * @param[in] x value to be distinguished
+ * @retval 0.0 or 1.0
+ */
+float stepFunction(float x) {
+  if (x > 0.9) {
```
kparichay, line 76: How about add a tolerance variable and then compare `x + tolerance > 1.0` and `x - tolerance < 0.0`?

Applications/mnist/jni/main.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+
+/**
+ * @brief     Generate Random integer value between min to max
+ * @param[in] min : minimum value
+ * @param[in] max : maximum value
+ * @retval    min < random value < max
+ */
+static int rangeRandom(int min, int max) {
+  int n = max - min + 1;
```
kparichay, line 95: `std::uniform_int_distribution<int> (min, max)`??

Applications/mnist/jni/main.cpp
```diff
@@ -0,0 +1,293 @@
...
+   * @brief     Neural Network Create & Initialization
+   */
+  nntrainer::NeuralNetwork NN(config);
+  NN.init();
+  NN.readModel();
+
+  /**
+   * @brief     Neural Network Train & validation
+   */
+  NN.train(getMiniBatch_train, getMiniBatch_val, nullptr);
```
kparichay, line 286: These functions `getMiniBatch_train/val` with same semantics are used everywhere. Do we want to provide these with our headers/C-API? If yes, will add.

Applications/mnist/jni/main.cpp
jijoongmoon: yes. it is intended.

Applications/mnist/jni/main.cpp
jijoongmoon: You may aware that this is not used for comparison. we can do this with tolerance though.

Applications/mnist/jni/main.cpp
jijoongmoon: Well, this does not use c-api. We can rewrite this with C-API. As you can see, this uses cpp. if you are talking about C++API, then, yes. we definitely needs it.

kparichay: Fair enough. These can be removed with C++ API.

PR #252, opened 2020-07-02
Rework bn forward & backward @open sesame 07/08 16:10

zhoonit:
Rework bn layer forward & backward pass.

~Please review but not merge before verification process is prepared~ 

**Self evaluation:**


nntrainer/src/bn_layer.cpp
```diff
@@ -92,58 +93,63 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
 
 Tensor BatchNormalizationLayer::forwarding(Tensor in, int &status) {
```
kparichay, line 94: If I am not wrong, the forward pass of batch normalization is different for training and evaluation.
So, I think you will need to update `forwarding` interface and differentiate them for train and eval;

zhoonit: Yes, I need to work for that one.

kparichay: #255 has been merged. It should be helpful here.

zhoonit: Thanks a lot ;)

zhoonit: To keep commit size small, I will work on this issue in different PR. Thanks :)

zhoonit:
PR Ready:

- Add verification with tf

nntrainer/src/bn_layer.cpp
```diff
@@ -37,10 +38,10 @@ int BatchNormalizationLayer::initialize(bool last) {
   dim = input_dim;
   output_dim = dim;
 
-  this->gamma = Tensor(dim.channel(), dim.batch(), dim.width());
-  this->beta = Tensor(dim.channel(), dim.batch(), dim.width());
-  beta.setZero();
-  gamma.setZero();
+  this->mu = Tensor(dim.channel(), dim.height(), dim.width());
```
kparichay, line 41: create a temp dim with batch_size = 1 and use it here?

kparichay: or infact does dim need to keep the batch dimension?

nntrainer/src/bn_layer.cpp
```diff
@@ -87,69 +87,71 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
...
 Tensor BatchNormalizationLayer::forwarding(Tensor in, int &status) {
   Tensor temp;
-  assert(dim.batch() > 0);
-  hidden = in;
+  this->input = in;
 
-  mu = in.sum(0).multiply(1.0 / dim.batch());
+  this->mu = in.average(0);
 
   temp = in.subtract(mu);
```
kparichay, line 94: let's rename `temp` as something like `mean_normalized_in`.

nntrainer/src/bn_layer.cpp
```diff
@@ -87,69 +87,71 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
 
 Tensor BatchNormalizationLayer::forwarding(Tensor in, int &status) {
   Tensor temp;
-  assert(dim.batch() > 0);
-  hidden = in;
+  this->input = in;
```
kparichay, line 90: Lets add a comment here at the top that this is for `TRAIN` mode, and `EVAL` mode is a TODO

nntrainer/include/bn_layer.h
```diff
@@ -125,6 +125,7 @@ class BatchNormalizationLayer : public Layer {
   Tensor var;
   Tensor gamma;
   Tensor beta;
+  Tensor x_hat;
```
kparichay, line 128: Add description on this variable or rename `x_hat` to `x_normalized`

nntrainer/src/bn_layer.cpp
```diff
@@ -87,69 +87,71 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
...
+  dx_hat = dy.multiply(gamma);
+
+  dx = dx_hat.chain()
+         .multiply_i(batch)
+         .subtract_i(dx_hat.sum(0))
+         .subtract_i(x_hat.multiply(dx_hat.multiply(x_hat).sum(0)))
+         .divide_i(var.multiply(batch))
+         .run();
+
+  weights.clear();
```
kparichay, line 132: Let's do this at initialization only.

nntrainer/src/bn_layer.cpp
```diff
@@ -37,10 +38,10 @@ int BatchNormalizationLayer::initialize(bool last) {
   dim = input_dim;
   output_dim = dim;
 
-  this->gamma = Tensor(dim.channel(), dim.batch(), dim.width());
-  this->beta = Tensor(dim.channel(), dim.batch(), dim.width());
-  beta.setZero();
-  gamma.setZero();
+  this->mu = Tensor(dim.channel(), dim.height(), dim.width());
+  this->var = Tensor(dim.channel(), dim.height(), dim.width());
```
kparichay, line 42: Do we need to store these variables as class variables? They are only used locally for forwarding.
The same goes for saving them as well.

IMO, we store the moving average of these variables which are used in the `EVAL` mode, not these variables directly.

test/input_gen/genInput.py
```diff
@@ -198,41 +202,52 @@ def fc_tf(x, kernel, label, bias, activation, train=False, loss='mse', opt='sgd'
...
-        bn_result = sess.run(bnlayer, feed_dict={tf_input: x})
-        grad_result = sess.run(grad, feed_dict={tf_input: x})
+
+        old_var = sess.run(input_variables, feed_dict=f_dict)
+        bn_result = sess.run(bnlayer, feed_dict=f_dict)
+        grad_result = sess.run(grad, feed_dict=f_dict)
+
         updated_gamma = sess.run(input_variables[1] - grad_result[1])
-        updated_beta = sess.run(input_variables[1] - grad_result[2])
+        updated_beta = sess.run(input_variables[2] - grad_result[2])
```
kparichay, line 237: How about SGD optimizer here? nntrainer now has the same defaults as tensorflow for SGD. So, IMO, you don't have to worry about learning rate or subtraction, and might be a more robust test.

nntrainer/src/bn_layer.cpp
zhoonit: Tensor constructor delegates dimension to 1 when it is omitted. So IMO it is okay.

nntrainer/src/bn_layer.cpp
zhoonit: mu and var needs to be updated and saved through training and this mu and var is used when inference.
More patches needed for that. (eg. adding momentum properties, save & read momentum to the file, using saved mu / var when infer... )

Changing `mu` to `moving_mu`  could be done though, IMO it is unnecessary since only `moving_mu` is used class wide(and for `var`)

nntrainer/src/bn_layer.cpp
zhoonit: > does dim need to keep the batch dimension?

I think it is good idea to use `dim` to represent those dimension though. I'll make a patch :)

nntrainer/src/bn_layer.cpp
kparichay: aah, you intend `mu` to store the moving average. 

> Changing mu to moving_mu could be done though, IMO it is unnecessary since only moving_mu is used class wide(and for var)

Yeah, how about just add some comment at its declaration about this?

jijoongmoon:
Could you check the initialization of epsilon? when you create bn layer, you set epsilon 0.001. but it will be set at setOptimizer() as zero. What is the default value?

nntrainer/src/bn_layer.cpp
```diff
@@ -63,10 +64,9 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
     NN_RETURN_STATUS();
 
     unsigned int type = parseLayerProperty(key);
-
     switch (static_cast<PropertyType>(type)) {
     case PropertyType::input_shape:
-      status = dim.setTensorDim(values[0].c_str());
+      status = input_dim.setTensorDim(value.c_str());
       break;
```
jijoongmoon, line 70: Is it necessary to give option about input shape? I wonder that is it possible to exist bn layer alone? I mean the input shape is totally depends on previous layer?

nntrainer/src/bn_layer.cpp
zhoonit: Can do. Thanks for the suggestion :)

nntrainer/src/bn_layer.cpp
```diff
@@ -87,69 +87,71 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
 
 Tensor BatchNormalizationLayer::forwarding(Tensor in, int &status) {
   Tensor temp;
-  assert(dim.batch() > 0);
-  hidden = in;
+  this->input = in;
 
-  mu = in.sum(0).multiply(1.0 / dim.batch());
+  this->mu = in.average(0);
```
jijoongmoon, line 92: It is not relevant this PR but I think the implementation of this average (int axis) is something wrong. It looks like only works for axis 0 (batch). cause it divides with batch not axis. Could you check it as well?

test/input_gen/genInput.py
zhoonit: I found out that `tf.keras.optimizer` does not get along with `tf.gradients` as it returns `tf.Tensor` for some and spits out `NotImplementedError` to use `optimizer.get_gradients`, it seems that I need to define some kind of loss. Making a more of an integrated test with loss layer.

While It would be worth testing with loss defined, I think checking the basic local backward pass is more accurate for the purpose. I will just leave the way it is for now and make a patch if I (or others) found a good solution for that.

zhoonit:
> Could you check the initialization of epsilon? when you create bn layer, you set epsilon 0.001. but it will be set at setOptimizer() as zero. What is the default value?

Yes. I changed it to use `BN layer epsilon`. Thanks :)

nntrainer/src/bn_layer.cpp
zhoonit: #262 will enable input_shape option to bn layer anyway. So, we need to explicitly prohibit setting `input_shape` after all. I can't find any reason to prohibit setting input_shape for bnlayer. You are right that this is not necessary for bn layer though.

nntrainer/src/bn_layer.cpp
zhoonit: Yes. Thanks for pointing out. I made an issue for that 👍

nntrainer/src/bn_layer.cpp
zhoonit: done.

nntrainer/src/bn_layer.cpp
zhoonit: done.

nntrainer/include/bn_layer.h
zhoonit: done.

nntrainer/src/bn_layer.cpp
zhoonit: done.

test/input_gen/genInput.py
kparichay: Alright, then add a testcase later with loss.

nntrainer/include/bn_layer.h
```diff
@@ -121,10 +121,15 @@ class BatchNormalizationLayer : public Layer {
 private:
   Tensor weight;
   Tensor bias;
+  //moving mu used for inferencing.
   Tensor mu;
+  //moving var used for inferencing.
   Tensor var;
+  //current variance used in training
```
myungjoo, line 128: You may want to start using doxygen-format comments. These comments may make you confused a few months later.

zhoonit: Updated. Thanks :)

nntrainer/include/bn_layer.h
```diff
@@ -121,10 +121,25 @@ class BatchNormalizationLayer : public Layer {
 private:
   Tensor weight;
   Tensor bias;
+  /**
+   * @brief moving mu used for inferencing.
+   *        momentum * mu + (1 - momenutm) * mu of current batch is used
+   */
```
myungjoo, line 127: No. you usually don't do this for class/struct properties. 

https://www.doxygen.nl/manual/docblocks.html
> int var; /**< Detailed description after the member */

or
> int var; ///< Detailed description after the member */

nntrainer/src/bn_layer.cpp
```diff
@@ -24,32 +24,44 @@
...
 #include <layer.h>
+#include <lazy_tensor.h>
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <parse_util.h>
 #include <util_func.h>
 
 namespace nntrainer {
 
+// todo: add channel wise bn for convolutional layer.
```
myungjoo, line 35: ```/// @todo ... ```

nntrainer/include/bn_layer.h
zhoonit: Thank you for pointing this out! One question though, is this rule applied also for comments inside a function?

nntrainer/src/bn_layer.cpp
```diff
@@ -79,70 +90,86 @@ int BatchNormalizationLayer::setProperty(std::vector<std::string> values) {
...
+    this->x_normalized = deviation.divide(cvar.apply(sqrtFloat));
 
-  status = ML_ERROR_NONE;
-  return ret;
+    this->hidden = x_normalized.chain().multiply_i(gamma).add_i(beta).run();
+
+    status = ML_ERROR_NONE;
+  } else {
+    /// NYI
+    throw std::runtime_error("hi");
```
kparichay, line 126: hi -> not yet implemented??

zhoonit: Ugh.. right lol

nntrainer/include/bn_layer.h
```diff
@@ -121,10 +121,19 @@ class BatchNormalizationLayer : public Layer {
...
   Tensor bias;
-  Tensor mu;
-  Tensor var;
+
+  Tensor mu;  /**< moving mu used for inferencing.
+                   momentum * mu + (1 - momenutm) * mu
+                   of current batch is used */
+  Tensor var; /**< moving var used for inferencing.
+                   momentum * var + (1 - momenutm) * var
+                   of current batch is used */
```
myungjoo, line 130: Yes, that's what I'm talking about. LGTM.

PR #253, opened 2020-07-03
[network] Add sensible defaults

### LINKED ISSUE #236: Make sensible defaults
Lets make defaults sensible for optimizers, loss and layer hyperparameters.

kparichay:
Added sensible defaults for the network initialization
This covers layers loss and optimizer properties
However defaults are not added for some of the core properties because they should be input by the user

Resolves #236

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -95,9 +95,6 @@ NeuralNetwork::NeuralNetwork() : NeuralNetwork("") {}
 
 NeuralNetwork::NeuralNetwork(std::string config) :
   batch_size(1),
-  learning_rate(0.01),
-  decay_rate(0.0),
-  decay_steps(0.0),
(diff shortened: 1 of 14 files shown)


nntrainer/src/neuralnet.cpp
```diff
@@ -316,15 +316,15 @@ int NeuralNetwork::init() {
...
         (int *)size);
       NN_INI_RETURN_STATUS();
       status = conv2d_layer->setSize(size, Layer::PropertyType::stride);
       NN_INI_RETURN_STATUS();
 
       status = getValues(CONV2D_DIM,
                          iniparser_getstring(
-                           ini, (layer_name + ":padding").c_str(), unknown),
+                           ini, (layer_name + ":padding").c_str(), getValues({0,0})),
                          (int *)size);
```
jijoongmoon, line 328: Default value of stride should be {1,1}.

PR #254, opened 2020-07-03
[git] ignore vscode configs and multiple vim backup files

kparichay:
ignore vscode configs and multiple vim backup files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,10 @@
 *.h~
 *.c~
 *.cc~
-*.swp
+*.sw?
+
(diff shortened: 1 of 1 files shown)


kparichay:
> LGTM. btw do you use vscode?

Yes, both vim and vscode :)
Do you too?

zhoonit:
> > LGTM. btw do you use vscode?
> 
> Yes, both vim and vscode :)
> Do you too?

Same here.

I was personally updating `./git/info/exclude`. Nice to have both here :)

PR #257, opened 2020-07-03
[unittest] Add forward unittest against tensorflow for loss, fc and activation

### LINKED ISSUE #267: Tensor::average(int axis)
Current implementation in `ca3e4fa` of Tensor::average(int axis) is only valid for axis 0

kparichay:
Add forward propagation unittests for fc layer with activation and loss.
This PR provides bug fixes for forward propagation of average tensor operator, softmax, MSE, sigmoid with cross entropy and softmax with cross entropy.
This also changes the semantics of average operator tensor. Average() performs average across all axis. However, average(axis) performs average for the given axis. Average(axis) has no default parameter anymore. Same for sum(axis).

Resolves #267


nntrainer/include/layer.h
```diff
@@ -425,6 +427,7 @@ class Layer {
    * @brief     Gradient for the weights in this layer
    */
   std::vector<std::reference_wrapper<Tensor>> gradients;
```
zhoonit, line 429: How about remove `reference_wrapper` here and use this variable as local variable inside `backwarding`  and lift this to a state saving vector? It seems that it is needed only for applying gradient.

```suggestion
  std::vector<Tensor> gradients;
```

kparichay: Yes, thats good.
But this would need the copy constructor of Tensor to be default and not create another copy of itself.
Lets do this after #258.

nntrainer/src/optimizer.cpp
```diff
@@ -85,7 +85,7 @@ void Optimizer::apply_gradients(
 
   float ll = popt.learning_rate;
   if (popt.decay_steps != -1) {
-    ll = ll * pow(popt.decay_rate, (iteration / popt.decay_steps));
+    // ll = ll * pow(popt.decay_rate, (iteration / popt.decay_steps));
   }
```
jijoongmoon, line 89: Do we have to comment out this learning rate decay?

kparichay: This is waiting for #253 as the earlier default value of `decay_steps` of `0` made a loss as `nan`

nntrainer/include/layer.h
zhoonit: Yes. I'll do that :)

kparichay:
This PR will be updated with just forwarding of various loss layers, and backwarding of FC layer.
Backwarding with loss layer has a major mismatch and will be done in a separate PR.

kparichay:
This PR has been redone. @zhoonit @jijoongmoon Please review.

nntrainer/src/tensor.cpp
```diff
@@ -887,7 +887,7 @@ Tensor Tensor::average(int axis) const {
 
   Tensor result;
   result = std::move(this->sum(axis));
-  result.divide_i(dim.batch());
+  result.divide_i(dim.getDim()[axis]);
```
zhoonit, line 890: Nice. please add 'resolves #267' to description

nntrainer/src/optimizer.cpp
zhoonit: guess this can be reverted now?

nntrainer/src/tensor.cpp
kparichay: Done.

zhoonit:
Looks good to go after checking https://github.com/nnstreamer/nntrainer/pull/257#discussion_r455522959

nntrainer/src/optimizer.cpp
kparichay: Removed unnecessary changes.

kparichay:
> Looks good to go after checking [#257 (comment)](https://github.com/nnstreamer/nntrainer/pull/257#discussion_r455522959)

Done.

Issue #260, opened 2020-07-06
C API support for editing layers with ini model


kparichay:
Add API to support models created with ini layers.

- [ ] Assign a name to all the layers in the model
- [ ] setProperty to support setting name for each layer
- [ ] Support get layer by name for C-API
- [ ] Get names of all the layers in the model for C-API
- [ ] Give default names to all the layers in the model 
- [ ] verify that each name is unique when adding a new layer

myungjoo:
Please keep in mind that we will be supporting different model formats (i.e., nnfw-runtime's .circle); the API should be not specific to the current .ini format. In such cases (nntrainer of 2021), we may have this .ini file and API calls to add info required for training not available in the given (exported/inference-only) .circle model.

kparichay:
@myungjoo Yes, the API is being tried to be kept generic.

kparichay:
This all done.

PR #261, opened 2020-07-06
[API] Search layer by name in C-API

kparichay:
Support searching of layer by name in C-API
This is part of the more changes to support this functionality

Related issue - #260

### DIFF
--- a/api/capi/include/nntrainer_internal.h
+++ b/api/capi/include/nntrainer_internal.h
@@ -24,8 +24,12 @@
 #ifndef __NNTRAINER_INTERNAL_H__
 #define __NNTRAINER_INTERNAL_H__
 
+#include <layer.h>
+#include <neuralnet.h>
 #include <nntrainer.h>
-#include <vector>
+#include <optimizer.h>
+#include <string>
+#include <unordered_map>
 
 #define ML_NNTRAINER_MAGIC 0x777F888F
 
@@ -48,7 +52,7 @@ typedef struct {
 typedef struct {
   uint magic;
   std::shared_ptr<nntrainer::NeuralNetwork> network;
-  std::vector<ml_nnlayer *> layers;
+  std::unordered_map<std::string, ml_nnlayer *> layers_map;
   ml_nnopt *optimizer;
 } ml_nnmodel;
 
@@ -90,6 +94,22 @@ typedef struct {
       return ML_ERROR_INVALID_PARAMETER;                             \
     }                                                                \
   } while (0)
+
+/**
+ * @brief Get neural network layer from the model with the given name.
+ * @details Use this function to get already created Neural Network Layer. The
+ * returned layer must not be deleted as it is owned by the model.
+ * @since_tizen 6.x
+ * @param[in] model The NNTrainer model handler from the given description.
+ * @param[in] layer_name Name of the already created layer.
+ * @param[out] layer The NNTrainer Layer handler from the given description.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Invalid parameter.
+ * @retval #ML_ERROR_CANNOT_ASSIGN_ADDRESS Cannot assign object.
+ */
+int ml_nnmodel_get_layer(ml_nnmodel_h model, const char *layer_name,
+                         ml_nnlayer_h *layer);
 
 #ifdef __cplusplus
 }
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -243,9 +243,9 @@ int ml_nnmodel_destruct(ml_nnmodel_h model) {
 
   if (nnmodel->optimizer)
     delete nnmodel->optimizer;
-  for (auto iter = nnmodel->layers.begin(); iter != nnmodel->layers.end();
-       ++iter)
-    delete (*iter);
+  for (auto &x : nnmodel->layers_map)
+    delete (x.second);
+  nnmodel->layers_map.clear();
   delete nnmodel;
 
   return status;
@@ -270,7 +270,7 @@ int ml_nnmodel_add_layer(ml_nnmodel_h model, ml_nnlayer_h layer) {
   status = nntrainer_exception_boundary(f);
   if (status == ML_ERROR_NONE) {
     nnlayer->in_use = true;
-    nnmodel->layers.push_back(nnlayer);
+    nnmodel->layers_map.insert({NL->getName(), nnlayer});
   }
 
   return status;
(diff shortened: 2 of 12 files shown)


nntrainer/src/neuralnet.cpp
```diff
@@ -1035,6 +1055,7 @@ std::shared_ptr<Layer> NeuralNetwork::_make_act_layer(ActiType act,
     std::shared_ptr<ActivationLayer> act_layer =
       std::make_shared<ActivationLayer>();
 
+    act_layer->setName(prev->getName() + "Activation");
     act_layer->setActivation(act);
```
jijoongmoon, line 1059: No related with this PR directly, I think when you have to make setLayer property after get the layer with this getLayer,  we have to extra concern about this Activation. if we change the activation, then we have to change  this activation layer as well.

kparichay: Yes, thats correct. May be create activation layer once the model is final and fixed.

nntrainer/src/neuralnet.cpp
```diff
@@ -1035,6 +1055,7 @@ std::shared_ptr<Layer> NeuralNetwork::_make_act_layer(ActiType act,
     std::shared_ptr<ActivationLayer> act_layer =
       std::make_shared<ActivationLayer>();
 
+    act_layer->setName(prev->getName() + "Activation");
```
zhoonit, line 1058: How about adding some delimiter like ":"?

kparichay: done.

kparichay:
@jijoongmoon @zhoonit Please review once more. This PR still aims for supporting getting layer by name. However, the implementation inside has been updated.

api/capi/include/nntrainer_internal.h
```diff
@@ -48,7 +52,7 @@ typedef struct {
 typedef struct {
   uint magic;
   std::shared_ptr<nntrainer::NeuralNetwork> network;
-  std::vector<ml_nnlayer *> layers;
+  std::unordered_map<std::string, ml_nnlayer *> layers_map;
```
zhoonit, line 55: How about...
  1. changing this to ordered map 
  2. putting this inside `nntrainer`? 

Considering we are going to change the saving format of model

nntrainer/src/neuralnet.cpp
```diff
@@ -1066,6 +1091,36 @@ int NeuralNetwork::setOptimizer(std::shared_ptr<Optimizer> optimizer) {
   return ML_ERROR_NONE;
 }
 
+void NeuralNetwork::ensureName(std::shared_ptr<Layer> &layer,
```
zhoonit, line 1094: I found this guide helpful https://www.modernescpp.com/index.php/c-core-guidelines-passing-smart-pointer

```suggestion
void NeuralNetwork::ensureName(std::shared_ptr<Layer> layer,
```

api/capi/include/nntrainer_internal.h
kparichay: 1. Is there any advantage to using ordered_map here over unordered_map? (given that unordered_map will give O(1) (amortized) search time compared to log(N) for map?)
2. Note that this is API specific as are storing API structure in this map to be returned to the user. As the API structure is not available and useful inside the nntrainer namespace, hence I did not add it to the namespace.

zhoonit: Sorry for lack of details. My suggestion was to change `NeuralNetwork::std::vector<std::shared_ptr<Layer>> layers;` to `ordered_map`. 🌵 
1.  No reason to change in this specific context.  if 🌵 is not done.
2.  If we do 🌵 I was thinking that we do not need this put `layers_map` inside api since the functionality would be part of the library.

kparichay: The reason to have `layers_map` in the API is to extract the associated `ml_nnlayer *` object.
When user create a layer, `nn_layer *` object is created and returned.
Now when user tries to get a layer which has already been created, its better not to create another `ml_nnlayer *` object, but to reuse the already created one.

> Sorry for lack of details. My suggestion was to change NeuralNetwork::std::vector<std::shared_ptr<Layer>> layers; to ordered_map. 🌵

I am not sure if there is a way to get reference to the `ml_nnlayer *` object for a layer whose `std::shared_ptr<Layer>` is obtained from 🌵

nntrainer/src/neuralnet.cpp
kparichay: Done.

api/capi/include/nntrainer_internal.h
zhoonit: Got it. Thanks for the explanation :)

api/capi/src/nntrainer.cpp
```diff
@@ -243,9 +243,9 @@ int ml_nnmodel_destruct(ml_nnmodel_h model) {
 
   if (nnmodel->optimizer)
     delete nnmodel->optimizer;
-  for (auto iter = nnmodel->layers.begin(); iter != nnmodel->layers.end();
-       ++iter)
-    delete (*iter);
+  for (auto &x : nnmodel->layers_map)
```
jijoongmoon, line 246: I think nnmodel->layers and nnmodel->layers_map are not same number of layer due to the activation and loss layer. Is it alright if we delete layer according to layers_map?

api/capi/src/nntrainer.cpp
```diff
@@ -303,12 +303,48 @@ int ml_nnmodel_set_optimizer(ml_nnmodel_h model, ml_nnopt_h optimizer) {
...
+int ml_nnmodel_get_layer(ml_nnmodel_h model, const char *layer_name,
+                         ml_nnlayer_h *layer) {
+  int status = ML_ERROR_NONE;
+  ml_nnmodel *nnmodel;
+  ML_NNTRAINER_CHECK_MODEL_VALIDATION(nnmodel, model);
+
+  std::shared_ptr<nntrainer::NeuralNetwork> NN;
+  std::shared_ptr<nntrainer::Layer> NL;
+
+  std::unordered_map<std::string, ml_nnlayer *>::iterator layer_iter =
```
jijoongmoon, line 315: I think this codes needs some description. It finds the layer:
1. layers_map
if not found : 2. find layers in NN
if not found : 3. make new layer and return ( just add new layer at the end of model?)
am i right?

api/capi/src/nntrainer.cpp
kparichay: Yes, nnmodel->layers_map will not contain all the layers.
This loop deletes memory allocated corresponding to `struct ml_nnlayer` object.
The internal layers are deleted with finalize (just like you said).

api/capi/src/nntrainer.cpp
kparichay: yes, using `ml_nnmodel_add_layer` is wrong. Entry should be only added to layers_map, and not to NeuralNetwork struct. Added #297 for bug fix and some description of this function.

PR #262, opened 2020-07-06
[property] Set common properties in layer.h

kparichay:
Many layers has common properties which are defined in layer.h
As they are in layer.h and are expected to be used in most layers, lets set their properties in layer.h itself
This reduces a lot of redundancies
If some property is not required to be handled by a specific layer in rare case, handle it as an error in that layer itself

Related Issue #260


jijoongmoon:
is their any way to give specific limits depends on layer?

kparichay:
> is their any way to give specific limits depends on layer?

If the property is not used many layers, then it should be removed from layer.h and added to its own layer.
If only a rare few should not be allowed to set that, then that layer should handle it with an error handler.

PR #263, opened 2020-07-06
[property] Add name property for layer

kparichay:
Add name property for layers

### DIFF
--- a/nntrainer/include/layer.h
+++ b/nntrainer/include/layer.h
@@ -26,6 +26,7 @@
 #include <fstream>
 #include <iostream>
 #include <optimizer.h>
+#include <set>
 #include <tensor.h>
 #include <tensor_dim.h>
 #include <vector>
@@ -327,6 +328,21 @@ public:
   bool getFlatten() { return flatten; }
 
   /**
+   * @brief     Set name of the layer
+   */
+  int setName(std::string name);
+
+  /**
+   * @brief     Set name of the layer
+   */
+  std::string getName();
+
+  /**
+   * @brief   Get base name of the layer
+   */
+  virtual std::string getBaseName() = 0;
+
+  /**
    * @brief     Property Enumeration
    *            0. input shape : string
    *            1. bias zero : bool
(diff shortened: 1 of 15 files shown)


nntrainer/include/layer.h
```diff
@@ -460,6 +492,8 @@ class Layer {
    */
   std::vector<std::reference_wrapper<Tensor>> weights;
 
+  int def_name_count;
```
zhoonit, line 495: How about making this variable static?

nntrainer/include/layer.h
```diff
@@ -322,6 +325,29 @@ class Layer {
...
+   * @brief   Get base name of the layer
+   */
+  virtual std::string getBaseName() = 0;
+
+  /**
+   * @brief     Ensure that layer has a name
+   */
+  void ensureName() {
+    if (name.empty())
+      name = getBaseName() + std::to_string(def_name_count++);
```
zhoonit, line 348: I think it is better to have prefix on generated name since name sth like "flatten1" could be taken by user.

nntrainer/include/layer.h
kparichay: aah yes, that is a bug. Updated.

nntrainer/include/layer.h
kparichay: This does not ensure that such a situation won't happen. Rather created a set of names of layer to ensure this.

nntrainer/src/layer.cpp
```diff
@@ -171,4 +178,42 @@ int Layer::setProperty(std::vector<std::string> values) {
...
+int Layer::setName(std::string name) {
+  int status = ML_ERROR_NONE;
+  std::set<std::string>::iterator iter;
+
+  if (name == this->name)
+    return status;
+
+  iter = layer_names.find(name);
+  if (name.empty()) {
+    status = ML_ERROR_INVALID_PARAMETER;
```
jijoongmoon, line 190: is this right? name.empty()? if it is right, then it should be checked earlier than this. I guess.

kparichay: Updated.

PR #273, opened 2020-07-07
[API] Update C-API @open sesame 06/08 13:02

kparichay:
Update C-API and corresponding code changes

Major Changes:

1. ml_nnmodel_compile_with_conf() replaced with ml_nnmodel_construct_with_conf().
This new function loads the model from the config file but does not initialize.
ml_nnmodel_compile() should be called after ml_nnmodel_construct_with_conf() before training.

2. ml_nnmodel_compile() does not take optimizer as an input.
Rather, use ml_nnmodel_set_optimizer() to set the optimizer for the model.

3. init() from neuralnet has been updated to loadFromConfig() and does not initialize the model anymore.
Rather call init() after loadFromConfig() to initialize.
This also allows updating the model config after loading model with loadFromConfig().

4. init(optimizer, args_list) has been replaced with init()
Rather call setOptimizer(optimizer) to set the optimizer and
setProperty(args_list) to set the properties before calling init().

5. Bug fixes in checkValidation()

### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -97,10 +97,11 @@ int ml_nnmodel_construct(ml_nnmodel_h *model) {
   return status;
 }
 
-int ml_nnmodel_compile_with_conf(const char *model_conf, ml_nnmodel_h model) {
+int ml_nnmodel_construct_with_conf(const char *model_conf, ml_nnmodel_h *model) {
   int status = ML_ERROR_NONE;
   ml_nnmodel *nnmodel;
-  std::shared_ptr<nntrainer::NeuralNetwork> nn;
+  std::shared_ptr<nntrainer::NeuralNetwork> NN;
+  returnable f;
 
   std::ifstream conf_file(model_conf);
   if (!conf_file.good()) {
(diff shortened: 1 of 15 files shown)


zhoonit:
FYI) #277 need merge to pass CI clang checker

api/capi/src/nntrainer.cpp
```diff
@@ -258,6 +265,27 @@ int ml_nnmodel_add_layer(ml_nnmodel_h model, ml_nnlayer_h layer) {
...
+  std::shared_ptr<nntrainer::NeuralNetwork> NN;
+  std::shared_ptr<nntrainer::Optimizer> opt;
+
+  NN = nnmodel->network;
+  opt = nnopt->optimizer;
+
+  returnable f = [&]() { return NN->setOptimizer(opt); };
+
+  status = nntrainer_exception_boundary(f);
+
```
zhoonit, line 285: Suggestion: if fail setting optimizer, destruct optimizer?

kparichay: IMO, this might be overkill.

kparichay:
V2:
Restrict the properties which can be set for the network after the compilation of the model has been done

api/capi/src/nntrainer.cpp
zhoonit: Yeah you're right. It's not creating optimizer so it shouldn't delete

kparichay: Yes, if this is to be added, this would extend to deletion of optimizer if set property fails and spiral from there.

nntrainer/src/neuralnet.cpp
```diff
@@ -499,29 +449,19 @@ int NeuralNetwork::init() {
...
     bool flatten =
       iniparser_getboolean(ini, (layer_name + ":Flatten").c_str(), false);
-    if (flatten) {
-      status = initFlattenLayer();
-      NN_INI_RETURN_STATUS();
-    }
+    layers.back()->setFlatten(flatten);
+
     previous_dim = layers.back()->getOutputDimension();
   }
```
jijoongmoon, line 460: I'm not sure that this has meaning. output dimension is set when the layer is initialized with input dimension. If you are going to initialize layer later, then this could be removed.

jijoongmoon:
code (line 631~640) is repeated. how about making some func or macro?
``` if (i != 0) {
        TensorDim def_init_dim;
        if (def_init_dim == layers[i]->getInputDimension()) {
          layers[i]->setInputDimension(previous_dim);
        } else if (previous_dim != layers[i]->getInputDimension()) {	
          ml_loge("Dimension mismatch between layers.");
          NN_RETURN_STATUS();
        }
      }	
```

nntrainer/src/neuralnet.cpp
kparichay: Yes, removed.

kparichay:
> code (line 631~640) is repeated. how about making some func or macro?

Done.

PR #274, opened 2020-07-07
[Proposal] Add UpdatableParam to manage weights / gradients

zhoonit:
_~Current phase of this PR is more of a proposal. I'd like to get some reviews and **approval** before proceed since this contains major change in `Layer`~_

This PR propose an `UpdateableParam` to manage trainable param altogether so that we can get the weight related parameters  easily.

- Add `UpdatableParam`
- Change `Optimizer::apply_gradient` signature
- Attach `UpdatableParam` to manage weights in layers ~(NYI except for `fc_layer` to provide a concept)~

**Self evaluation:**

### DIFF
--- a/nntrainer/include/layer.h
+++ b/nntrainer/include/layer.h
@@ -25,6 +25,7 @@
 
 #include <fstream>
 #include <iostream>
+#include <memory>
 #include <optimizer.h>
 #include <set>
 #include <tensor.h>
@@ -130,7 +131,8 @@ public:
     weight_decay(),
     weight_ini_type(WEIGHT_XAVIER_UNIFORM),
     flatten(false),
-    trainable(true) {}
+    trainable(true),
+    param_size(0) {}
 
   /**
    * @brief     Destructor of Layer Class
@@ -163,15 +165,17 @@ public:
 
   /**
    * @brief     read layer Weight & Bias data from file
+   * @note      derived class can call this to get/save updatableParams
    * @param[in] file input file stream
    */
-  virtual void read(std::ifstream &file) = 0;
+  virtual void read(std::ifstream &file);
 
   /**
    * @brief     save layer Weight & Bias data from file
+   * @note      derived class can call this to get/save updatableParams
    * @param[in] file output file stream
    */
-  virtual void save(std::ofstream &file) = 0;
+  virtual void save(std::ofstream &file);
 
   /**
    * @brief     set Property of layer
@@ -219,7 +223,7 @@ public:
    * @brief     Copy Layer
    * @param[in] l Layer to be copied
    */
-  virtual void copy(std::shared_ptr<Layer> l) = 0;
+  virtual void copy(std::shared_ptr<Layer> l);
 
   /**
    * @brief     set Batch Normalization Layer followed
@@ -307,20 +311,10 @@ public:
   void setTrainable(bool train) { trainable = train; }
 
   /**
-   * @brief     get gradients
-   * @retval    shared ptr of vector of all tensors
+   * @brief     get updatable params of all
+   * @retval    vector of all params
    */
-  std::shared_ptr<std::vector<Tensor>> getGradients() {
-    return getObjFromRef(gradients);
-  }
-
-  /**
-   * @brief     get weights
-   * @retval    shared ptr of vector of all tensors
-   */
-  std::shared_ptr<std::vector<Tensor>> getWeights() {
-    return getObjFromRef(weights);
-  }
+  std::shared_ptr<UpdatableParam> getParams() { return params; }
 
   /**
    * @brief     get if the output of this layer must be flatten
@@ -480,25 +474,61 @@ protected:
   bool trainable;
 
   /**
-   * @brief     Gradient for the weights in this layer
-   * @note      The order of gradients should match the order in weights
+   * @brief     reserve memory for @a params and set @a param_size
+   * @exception std::invalid_argument when param_size is already set and
+   * shouldn't be changed again.
    */
-  std::vector<std::reference_wrapper<Tensor>> gradients;
+  void setParamSize(unsigned int psize) {
+
+    // @note Need opinion about this
+    // if (param_size > 0) {
+    //   throw std::invalid_argument("param size can't be set once it is set");
+    // }
+
+    param_size = psize;
+    params = std::shared_ptr<UpdatableParam>(
+      new UpdatableParam[psize], std::default_delete<UpdatableParam[]>());
+  }
 
   /**
-   * @brief     weights in this layer
-   * @note      The weights are combined with their corresponding bias
-   *            For example- with W0, W1, B0 and B1, weights would be of format
-   *            {W0, B0, W1, B1}.
+   * @brief     get data alias at param position.
+   * @exception std::out_of_range for index out of range
    */
-  std::vector<std::reference_wrapper<Tensor>> weights;
+  UpdatableParam &paramsAt(const unsigned int position) {
+    if (position >= param_size) {
+      throw std::out_of_range("index out of range");
+    }
+
+    return params.get()[position];
+  }
+
+  /**
+   * @brief     updatable params in this layer. This contains params of layers.
+   * @note      UpdatableParam has weights and gradients paired.
+   */
+  std::shared_ptr<UpdatableParam> params;
+
+  unsigned int param_size; /**< length of UpdatableParam * params.
+                                This shouldn't be changed
+                                after initiation
+                                use setParamSize() to avoid
+                                setting parameters twice */
 
 private:
   /**
-   * @brief     Convert vector of reference to vector of objects
+   * @brief     Set containing all the names of layers
    */
-  std::shared_ptr<std::vector<Tensor>>
-  getObjFromRef(std::vector<std::reference_wrapper<Tensor>> &elements);
+  static std::set<std::string> layer_names;
+
+  /**
+   * @brief     Count assigned to layer names declared by default
+   */
+  static int def_name_count;
+
+  /**
+   * @brief     Ensure that layer has a name
+   */
+  void ensureName();
 };
 } // namespace nntrainer
 
--- a/nntrainer/src/bn_layer.cpp
+++ b/nntrainer/src/bn_layer.cpp
@@ -32,28 +32,31 @@
 
 namespace nntrainer {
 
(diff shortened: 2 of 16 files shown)


nntrainer/include/layer.h
```diff
@@ -111,24 +111,37 @@ typedef enum {
...
 } WeightIniType;
 
+/**
+ * @brief
+ */
+struct UpdatableParam {
+  Tensor weight;           /**<  weight to be updated and used */
+  Tensor grad;             /**<  gradient for the weight */
+  std::string name;        /**<  identifier of the param */
+  bool update_through_opt; /**<  determines the param should be updated through
```
kparichay, line 121: how about renaming this to trainable?

nntrainer/src/fc_layer.cpp
```diff
@@ -53,9 +55,9 @@ int FullyConnectedLayer::initialize(bool last) {
     bias.setRandUniform(-0.5, 0.5);
   }
 
-  weights.clear();
-  weights.push_back(weight);
-  weights.push_back(bias);
+  params.clear();
+  params.push_back({std::move(weight), Tensor(dim), "FC:weight", true});
```
kparichay, line 59: setting the last parameter as `true & trainable`?

zhoonit:
Had a discussion with @kparichay had some aggrement on 

- remove vectors to keep track of params instead use array
- remove boolean in `UpdatableParams`

zhoonit:
v2:

 - Change params to arrays (and remove unused variables)
 - Use `Layers::copy, save, read` to process `params`
 - Update derived layers to use params
 - Fix conv2d logic error

zhoonit:
Add a separate commit to: 
 - enable exceptions
 - bump jni build to c++14

nntrainer/include/layer.h
```diff
@@ -479,20 +473,34 @@ class Layer {
...
-   * @brief     Gradient for the weights in this layer
-   * @note      The order of gradients should match the order in weights
+   * @brief     reserve memory for @a params and set @a param_size
+   * @exception std::invalid_argument when param_size is already set and
+   * shouldn't be changed again.
    */
-  std::vector<std::reference_wrapper<Tensor>> gradients;
+  void setParamSize(unsigned int psize) {
+
+    // @note Need opinion about this
```
kparichay, line 482: Lets enable for this now. Each layer functionality is fixed for now, later this can be changed.

nntrainer/include/layer.h
```diff
@@ -398,8 +392,8 @@ class Layer {
   std::string name;
 
   /**
-   * @brief     check if current layer's weight decay type is l2norm
-   * @return    bool is weightdecay type is L2 Norm
+   * @brief        check if current layer's weight decay type is l2norm
```
kparichay, line 395: Extra indentation?

nntrainer/include/layer.h
```diff
@@ -509,10 +517,16 @@ class Layer {
   void ensureName();
 
   /**
-   * @brief     Convert vector of reference to vector of objects
+   * @brief     updatable params in this layer. This contains layers of params .
```
kparichay, line 520: This contains params of the layers?

nntrainer/include/layer.h
zhoonit: Ugh,, I found out that layer unittest need change. I'll just hold this for now and make a patch after this is merged

nntrainer/include/layer.h
```diff
@@ -479,20 +473,34 @@ class Layer {
...
-  std::vector<std::reference_wrapper<Tensor>> weights;
+  UpdatableParam &paramsAt(const unsigned int position) {
+    if (position >= param_size) {
+      throw std::out_of_range("index out of range");
+    }
+
+    return params.get()[position];
+  }
 
-private:
```
kparichay, line 495: why move all of these to protected?

nntrainer/src/conv2d_layer.cpp
```diff
@@ -30,30 +31,36 @@ int Conv2DLayer::initialize(bool last) {
...
-      Tensor(input_dim.batch(), Kdim.channel(), Kdim.height(), Kdim.width()));
-    delBias.push_back(Tensor(input_dim.batch(), 1, 1, 1));
-    filters.push_back(Knl);
-    weights.push_back(Knl);
+    Tensor bias = Tensor(1, 1);
 
-    Tensor B(1, 1, 1, 1);
     if (!bias_init_zero) {
-      B.apply([&](float x) { return random(); });
+      bias.apply([&](float x) { return random(); });
```
kparichay, line 52: `setZero()` in `else` case?

nntrainer/src/bn_layer.cpp
```diff
@@ -151,30 +159,20 @@ Tensor BatchNormalizationLayer::backwarding(Tensor dy, int iteration) {
...
   return dx;
 }
 
-void BatchNormalizationLayer::read(std::ifstream &file) {
-  mu.read(file);
-  var.read(file);
-  gamma.read(file);
-  beta.read(file);
-}
+void BatchNormalizationLayer::read(std::ifstream &file) { Layer::read(file); }
```
kparichay, line 169: i think read and save need not be pure virtual now. In another PR possibly.

nntrainer/include/layer.h
zhoonit: Oh, I just put it back. Don't know why yesterday me deleted that

nntrainer/include/layer.h
zhoonit: Fixed :)

nntrainer/include/layer.h
zhoonit: Fixed :)

nntrainer/src/bn_layer.cpp
```diff
@@ -32,28 +32,31 @@
...
   mu.setZero();
   var.setValue(1);
   gamma.setZero();
   beta.setZero();
 
-  weights.clear();
-  weights.push_back(gamma);
-  weights.push_back(beta);
+  setParamSize(4);
+  paramsAt(0) = {std::move(mu), Tensor(dim), "BN:moving_average"};
```
kparichay, line 56: As not all params will use `apply_gradients` like in this case, how about later allowing not allocating memory for their grads?

nntrainer/src/bn_layer.cpp
zhoonit: Yes. it is not pure virtual in this PR.

nntrainer/src/bn_layer.cpp
zhoonit: Yeah, do I 'have to' use pointer to make null kind of thing in c++?

nntrainer/src/bn_layer.cpp
kparichay: So, this definition can be just removed from here?

nntrainer/src/conv2d_layer.cpp
zhoonit: Right. fixed :)

nntrainer/src/bn_layer.cpp
zhoonit: Right. Thanks :)

nntrainer/src/bn_layer.cpp
kparichay: One way might be is to allow making empty tensors?

zhoonit: Yup.

kparichay: I guess this can be done later.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -169,7 +171,7 @@ Tensor Conv2DLayer::backwarding(Tensor derivative, int iteration) {
           sum += derivative.getValue(b, i, j, k);
         }
       }
-      delBias[i].setValue(0, 0, 0, 0, sum);
+      delBias.setValue(0, 0, 0, 0, sum);
     }
```
jijoongmoon, line 175: this should be delBias.setValue(b,0,0,0,sum). this is fixed at #292 which includes others. I strongly recommend you to merge this after #292 and #288 and #250.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -95,20 +96,21 @@ Tensor Conv2DLayer::forwarding(Tensor in, int &status) {
...
+
     for (unsigned int i = 0; i < filter_size; ++i) {
-      status = conv2d(in_padded.getData(), in_padded.getDim(),
-                      filters[i].getData(), filters[i].getDim(), output.data(),
-                      stride, bias[i].getValue(b, 0, 0, 0));
+      Tensor &filter = paramsAt(i).weight;
+      Tensor &bias = paramsAt(i + filter_size).weight;
+      status = conv2d(in_padded.getData(), in_padded.getDim(), filter.getData(),
+                      filter.getDim(), output.data(), stride,
+                      bias.getValue(b, 0, 0, 0));
```
jijoongmoon, line 109: this should be bias.getValue(0,0,0,0)

nntrainer/src/conv2d_layer.cpp
zhoonit: If there is a lot of change in `conv2dlayer`, how about I revert the change of conv2d_layer.cpp in this PR to reduce dependency between PRs?

nntrainer/include/layer.h
```diff
@@ -306,20 +310,10 @@ class Layer {
...
-  }
-
-  /**
-   * @brief     get weights
-   * @retval    shared ptr of vector of all tensors
-   */
-  std::shared_ptr<std::vector<Tensor>> getWeights() {
-    return getObjFromRef(weights);
-  }
+  std::shared_ptr<UpdatableParam> getParams() { return params; }
```
jijoongmoon, line 316: how about implement getGradients() and getWeights(). Cause user have to handle if he want to get gradient only.

zhoonit: Yes, I'll work on that once this is merged. Thanks

zhoonit:
I'll wait for #292, #288, #250

zhoonit:
Work done and it is ready to be merged :)

jijoongmoon:
Did you check the implementation with mnist example in Application/mnist ? It it is working without error, then it is good to go.

Issue #270, opened 2020-07-07
Add summary to capi


zhoonit:
User of nntrainer should have ways to review current model construct. so there should be a way to print summary of a model.

zhoonit:
- [ ] Add print methods to tensors
- [ ] Add print methods to layers
- [ ] Add print methods to neural network

zhoonit:
Need polishing a bit though, it is done for now.

PR #276, opened 2020-07-08
Enhance Tensor print

zhoonit:
- Add tensor address to the print
- Handle the case when tensor is too large
- Add test

See also: #270

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -783,17 +783,25 @@ Tensor Tensor::apply(std::function<float(float)> f) const {
 Tensor Tensor::apply(std::function<Tensor(Tensor)> f) const { return f(*this); }
 
 void Tensor::print(std::ostream &out) const {
-  unsigned int i, j, k, l;
-  std::stringstream ss;
-
+  out << "<Tensor Object at " << this << ">" << std::endl;
   const float *data = getData();
-  for (k = 0; k < dim.batch(); k++) {
-    for (l = 0; l < dim.channel(); l++) {
-      for (i = 0; i < dim.height(); i++) {
-        for (j = 0; j < dim.width(); j++) {
-          out << data[k * dim.getFeatureLen() + l * dim.width() * dim.height() +
-                      i * dim.width() + j]
-              << " ";
+
+  unsigned int len = length();
+
+  out << dim;
+
+  if (len > 100) {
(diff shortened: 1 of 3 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -783,17 +783,24 @@ Tensor Tensor::apply(std::function<float(float)> f) const {
...
+  unsigned int len = length();
+
+  if (len > 100) {
+    out << '[' << data[0] << ' ' << data[1] << ' ' << data[2] << " ... "
+        << data[len - 3] << ' ' << data[len - 2] << ' ' << data[len - 1] << ']'
+        << std::endl;
+    return;
+  }
+
+  unsigned int i, j, k, l;
```
wooksong, line 798: 1. If these iterator variables are used only for the following 'for' loops, you can move them into corresponding 'for' statements.

2. Comments for the following 'for' statements: 
dim.batch(), dim.channel(), dim.height(), dim.width(), and dim.getFeatureLen() would be invoked multiple times for no reasons. Please use variables for them outside the following 'for' loops.

zhoonit: Q. Regrading 2. those are inline functions. Does using a variable outside of the loop make a difference?

zhoonit: Updated for 1. Thanks :)

nntrainer/src/tensor.cpp
```diff
@@ -783,17 +783,25 @@ Tensor Tensor::apply(std::function<float(float)> f) const {
...
-          out << data[k * dim.getFeatureLen() + l * dim.width() * dim.height() +
-                      i * dim.width() + j]
-              << " ";
+
+  unsigned int len = length();
+
+  out << dim;
+
+  if (len > 100) {
+    out << '[' << data[0] << ' ' << data[1] << ' ' << data[2] << " ... "
```
kparichay, line 794: How about skip printing data in the model summary. If someone needs data, they can directly inquire tensor.

kparichay:
Other than @wooksong comments, LGTM.

nntrainer/src/tensor.cpp
wooksong: A for 2. https://isocpp.org/wiki/faq/inline-functions#inline-and-perf

zhoonit: Thanks you for the explanation :)

PR #279, opened 2020-07-08
[API] Update C-API semantics

kparichay:
Update C-API semantics about layer and optimizer
While the layer and optimizer are in use, do allow deletion of layer and optimizer
Further, once the layer and optimizer are set/added to a model, their ownership is transferred to the model
So, no need to delete them separately. Deleting the model will directly delete them as well.
    

jijoongmoon:
I wonder that if you check the Applicaiton/Tizen_CAPI as well??

kparichay:
> I wonder that if you check the Applicaiton/Tizen_CAPI is run properly as well??

Yes, those were missed. Updated them as well.

kparichay:
> LGTM. rebase plz.

Done

PR #282, opened 2020-07-08
Add print for layer

zhoonit:
- Add layer print and override '<<' for ofstream

See also: #270

**Self evaluation:**

### DIFF
--- a/nntrainer/src/fc_layer.cpp
+++ b/nntrainer/src/fc_layer.cpp
@@ -158,4 +158,13 @@ Tensor FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
 
   return ret;
 }
+
+void FullyConnectedLayer::print(std::ostream &out) const {
+  Layer::print(out);
+
+  out << "unit: " << unit << std::endl << "= weight info =" << std::endl;
+
+  weight.print(out);
+  bias.print(out);
+}
 } /* namespace nntrainer */
--- a/nntrainer/src/layer.cpp
+++ b/nntrainer/src/layer.cpp
@@ -25,6 +25,7 @@
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <parse_util.h>
+#include <typeinfo>
 #include <util_func.h>
 
 namespace nntrainer {
@@ -213,6 +214,22 @@ void Layer::ensureName() {
 
     layer_names.insert(name);
   }
+}
+
+void Layer::print(std::ostream &out) const {
+  /// @todo change type enum to actual string value
+  out << '<' << typeid(*this).name() << "Object at" << this << '>' << std::endl
+      << "Layer Type: " << type << std::endl
+      << "bias_initiate_to_zero: " << bias_init_zero << std::endl
+      << "Weight Decay: " << static_cast<int>(weight_decay.type) << " | "
+      << weight_decay.lambda << std::endl
+      << "Weight Initiation Type: " << weight_ini_type << std::endl
+      << "Loss: " << loss << std::endl
+      << "Cost: " << cost << std::endl
+      << "last layer: " << last_layer << std::endl
+      << "activation type after: " << activation_type << std::endl
+      << "flatten layer after: " << flatten << std::endl
+      << "Trainable: " << trainable << std::endl;
 }
 
 } /* namespace nntrainer */
--- a/test/unittest/unittest_nntrainer_layers.cpp
+++ b/test/unittest/unittest_nntrainer_layers.cpp
@@ -210,6 +210,14 @@ TEST_F(nntrainer_FullyConnectedLayer, initialize_01_p) {
 }
 
 /**
+ * @brief Fully Connected Layer Print test (Wait 4 #274)
+ */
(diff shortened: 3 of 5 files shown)


test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -209,6 +209,13 @@ TEST_F(nntrainer_FullyConnectedLayer, initialize_01_p) {
   EXPECT_EQ(status, ML_ERROR_NONE);
 }
 
+/**
+ * @brief Fully Connected Layer Print test (Waiting for #276)
+ */
+// TEST_F(nntrainer_FullyConnectedLayer, print_01_p) {
+//   std::cerr << layer;
+// }
+
```
jijoongmoon, line 218: #276 is merged. rebase plz.

zhoonit:
PR itself is fine though, I'll make a test and remove `DO NOT MERGE`

nntrainer/src/fc_layer.cpp
```diff
@@ -158,4 +158,13 @@ Tensor FullyConnectedLayer::backwarding(Tensor derivative, int iteration) {
 
   return ret;
 }
+
+void FullyConnectedLayer::print(std::ostream &out) const {
+  Layer::print(out);
+
+  out << "unit: " << unit << std::endl << "= weight info =" << std::endl;
+
+  weight.print(out);
```
kparichay, line 167: is it possible to add a mechanism to restrict the level of information?
This will be useful for the C-API as we don't expose tensor level information (abstracted by layer) in the API, and printing summary for tensors might not be a good choice.

nntrainer/src/layer.cpp
```diff
@@ -215,4 +216,20 @@ void Layer::ensureName() {
   }
 }
 
+void Layer::print(std::ostream &out) const {
+  /// @todo change type enum to actual string value
+  out << '<' << typeid(*this).name() << "Object at" << this << '>' << std::endl
+      << "Layer Type: " << type << std::endl
+      << "Activation type: " << activation_type << std::endl
+      << "bias_initiate_to_zero: " << bias_init_zero << std::endl
```
kparichay, line 224: not every layer uses `bias` and `weight decay`. Let each layer add this if they use it?

nntrainer/src/layer.cpp
```diff
@@ -215,4 +216,20 @@ void Layer::ensureName() {
...
+      << "Weight Decay: " << static_cast<int>(weight_decay.type) << " | "
+      << weight_decay.lambda << std::endl
+      << "Weight Initiation Type: " << weight_ini_type << std::endl
+      << "Loss: " << loss << std::endl
```
kparichay, line 228: IMO, loss and cost is redundant with each layer as it is same for the whole network.

nntrainer/src/layer.cpp
```diff
@@ -215,4 +216,20 @@ void Layer::ensureName() {
...
+      << "Cost: " << cost << std::endl
+      << "last layer: " << last_layer << std::endl
+      << "Need flatten: " << flatten << std::endl
```
kparichay, line 231: As flatten is always added as a new layer, lets not print it here. Flatten can be printed here if print in flatten layer is disabled. Same applies for activation as well.

nntrainer/src/fc_layer.cpp
zhoonit: I think this must be done at least internally though, I will make PRs as follows:

- [ ] Add neuralnet print
- [ ] Add summary control

since I am gonna heavily rely on delegation, it would be more reasonable to review control level's altogether

nntrainer/src/layer.cpp
zhoonit: I'll delete cost 
loss is a bit subtle. Current implementation save's weight decay loss inside each layer (not in loss layer).

zhoonit: may be change name to `Loss added`  and support when weight_decay is available?

nntrainer/src/layer.cpp
zhoonit: before init() is called, flatten layer isn't attached. so IMO it is needed. Instead I clarified name to "flatten layer after"

nntrainer/src/layer.cpp
kparichay: How about represent it as `weight regularization loss`?
For loss layer, it can just be `loss`.

nntrainer/src/layer.cpp
kparichay: Then lets give this only if the layer is not initialized. Same for activation.
It will be awkward to give it out twice, as layer property and a new layer after the layer is initialized.

zhoonit: yes

nntrainer/src/layer.cpp
zhoonit: Sounds good, I'll do that

nntrainer/src/layer.cpp
zhoonit: I am going to use #270 to filter out valid properties. I guess this should be done in another PR.

zhoonit:
I'll close this PR and rework. I need to some restructuring.

PR #294, opened 2020-07-10
Add setProperty for each property type

zhoonit:
Since Layer::setProperty is iterating through vectors, only part that
needs overriding is setting each particular property.

This PR add setProperty by type to relieve the issue.

w.r.t #270, this patch enables layers to check if given property type
 is valid.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/conv2d_layer.cpp
+++ b/nntrainer/src/conv2d_layer.cpp
@@ -287,60 +287,60 @@ int Conv2DLayer::setFilter(int f) {
   return status;
 }
 
-int Conv2DLayer::setProperty(std::vector<std::string> values) {
+void Conv2DLayer::setProperty(const PropertyType type,
+                              const std::string &value) {
   int status = ML_ERROR_NONE;
 
-  for (unsigned int i = 0; i < values.size(); ++i) {
-    std::string key;
-    std::string value;
-
-    status = getKeyValue(values[i], key, value);
-    NN_RETURN_STATUS();
-
-    unsigned int t = parseLayerProperty(key);
-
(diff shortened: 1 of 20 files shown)


nntrainer/src/conv2d_layer.cpp
```diff
@@ -284,60 +284,60 @@ int Conv2DLayer::setFilter(int f) {
...
+  case PropertyType::kernel_size:
+    if (!value.empty()) {
       status = getValues(CONV2D_DIM, value, (int *)(kernel_size));
-      NN_RETURN_STATUS();
+      throw_status(status);
       if (kernel_size[0] == 0 || kernel_size[1] == 0) {
-        ml_loge("Error: kernel_size must be greater than 0");
-        return ML_ERROR_INVALID_PARAMETER;
+        throw std::out_of_range(
+          "[Conv2DLayer] kernel_size must be greater than 0");
```
jijoongmoon, line 306: Looks like it does not write error log in log file.

zhoonit: Error handling through standard exception separates handling error and throwing.
Error is not handled here. It is handled at the point where it is caught.

In our code, it is guaranteed to be caught at the api layer and print log. Although it could be caught earlier and handled if needed.

https://github.com/nnstreamer/nntrainer/blob/d15c3d60048bceee9224e8680b892ac61ca48d72/api/capi/src/nntrainer.cpp#L37-L59

jijoongmoon: I see. Looks reasonable.

PR #302, opened 2020-07-14
[API] Update C-API dataset handling @open sesame 07/14 17:45

kparichay:
Update C-API to add dataset handling
Changes added to data buffer:
- Bug fix for data buffer using already freed memory
- Data buffer set properties moved out from neural network
- Data buffer generator now takes list of array for multiple inputs
    
### DIFF
--- a/Applications/Classification/jni/Android.mk
+++ b/Applications/Classification/jni/Android.mk
@@ -10,7 +10,7 @@ endif
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../libs/arm64-v8a
 NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include \
-	$(LOCAL_PATH)/../../../api/capi/include \
+	$(LOCAL_PATH)/../../../api \
 	$(LOCAL_PATH)/../../../api/capi/include/platform
 endif
 
--- a/Applications/Classification/jni/Android.mk
+++ b/Applications/Classification/jni/Android.mk
@@ -9,7 +9,9 @@ endif
 
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../libs/arm64-v8a
-NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include
+NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include \
+	$(LOCAL_PATH)/../../../api/capi/include \
+	$(LOCAL_PATH)/../../../api/capi/include/platform
 endif
 
 NNTRAINER_APPLICATION := $(LOCAL_PATH)/../..
@@ -45,7 +47,7 @@ include $(PREBUILT_SHARED_LIBRARY)
 
 include $(CLEAR_VARS)
 LOCAL_ARM_NEON := true
-LOCAL_CFLAGS += -std=c++14 -Ofast -mcpu=cortex-a53 -Ilz4-nougat/lib 
+LOCAL_CFLAGS += -std=c++14 -Ofast -mcpu=cortex-a53 -Ilz4-nougat/lib
 LOCAL_LDFLAGS += -Llz4-nougat/lib/obj/local/arm64-v8a/
 LOCAL_CXXFLAGS += -std=c++14
 LOCAL_CFLAGS += -pthread -fopenmp -fexceptions
--- a/Applications/mnist/jni/Android.mk
+++ b/Applications/mnist/jni/Android.mk
@@ -10,7 +10,7 @@ endif
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../libs/arm64-v8a
 NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include \
-	$(LOCAL_PATH)/../../../api/capi/include \
+	$(LOCAL_PATH)/../../../api \
 	$(LOCAL_PATH)/../../../api/capi/include/platform
 endif
 
--- a/Applications/mnist/jni/Android.mk
+++ b/Applications/mnist/jni/Android.mk
@@ -9,7 +9,9 @@ endif
 
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../libs/arm64-v8a
-NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include
+NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include \
+	$(LOCAL_PATH)/../../../api/capi/include \
+	$(LOCAL_PATH)/../../../api/capi/include/platform
 endif
 
 include $(CLEAR_VARS)
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -20,6 +20,9 @@
  * @author Jijoong Moon <jijoong.moon@samsung.com>
  * @bug No known bugs except for NYI items
  */
+#include <databuffer.h>
+#include <databuffer_file.h>
+#include <databuffer_func.h>
 #include <ml-api-common.h>
 #include <neuralnet.h>
 #include <nntrainer_internal.h>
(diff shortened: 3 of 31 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -483,6 +458,157 @@ int ml_train_optimizer_set_property(ml_train_optimizer_h optimizer, ...) {
...
 }
 
+int ml_train_model_set_dataset(ml_train_model_h model,
+                               ml_train_dataset_h dataset) {
+  int status = ML_ERROR_NONE;
+  ml_train_model *nnmodel;
+  ml_train_dataset *nndataset;
+
+  ML_NNTRAINER_CHECK_MODEL_VALIDATION(nnmodel, model);
+  ML_NNTRAINER_CHECK_DATASET_VALIDATION(nndataset, dataset);
```
zhoonit, line 468:
```cpp
if (dataset->in_use) {
   /** fail validation ? */
}
```

api/capi/src/nntrainer.cpp
```diff
@@ -483,6 +458,157 @@ int ml_train_optimizer_set_property(ml_train_optimizer_h optimizer, ...) {
...
+  data = nndataset->data_buffer;
+
+  returnable f = [&]() { return NN->setDataBuffer(data); };
+
+  status = nntrainer_exception_boundary(f);
+  if (status == ML_ERROR_NONE) {
+    nndataset->in_use = true;
+    if (nnmodel->dataset)
+      nnmodel->dataset->in_use = false;
+    nnmodel->dataset = nndataset;
```
zhoonit, line 483: It seems that if there is any previous dateset, it could dangle although it would be user's responsibility.

api/capi/src/nntrainer.cpp
```diff
@@ -483,6 +458,157 @@ int ml_train_optimizer_set_property(ml_train_optimizer_h optimizer, ...) {
...
+  nndataset->magic = ML_NNTRAINER_MAGIC;
+  nndataset->data_buffer =
+    std::make_shared<nntrainer::DataBufferFromDataFile>();
+  nndataset->in_use = false;
+
+  std::shared_ptr<nntrainer::DataBufferFromDataFile> callback_buffer =
+    std::static_pointer_cast<nntrainer::DataBufferFromDataFile>(
+      nndataset->data_buffer);
+
+  if (train_file) {
```
zhoonit, line 543: train file can't be null w.r.t documentation, so return error when !train_file or change docs?

nntrainer/src/databuffer.cpp
```diff
@@ -26,11 +26,14 @@
 #include <condition_variable>
 #include <cstring>
 #include <databuffer.h>
+#include <databuffer_file.h>
+#include <databuffer_func.h>
```
zhoonit, line 30: Q. is there a reason for this? I am concerning that having derived class in base class.

nntrainer/include/databuffer_func.h
```diff
@@ -31,11 +31,18 @@
...
+#include <nntrainer_internal.h>
 #include <thread>
 #include <vector>
 
 namespace nntrainer {
 
+/**
+ * @brief   Dataset generator callback type declaration
+ */
+typedef std::function<std::remove_pointer<ml_train_datagen_cb>::type>
```
zhoonit, line 43: IMO, it'd better to:
   1. declare `ml_train_datagen_cb` -> `datagen_cb_ptr` inside here
   2. alias this ptr from capi

otherwise it lib depends on api making circular dependency.

in this case `nntrainer_internal.h <- nntrainer.h <- lib <- nntrainer_internal.h`

nntrainer/include/neuralnet.h
```diff
@@ -286,16 +268,11 @@ class NeuralNetwork {
...
-    label_data = 5,
-    buffer_size = 6,
-    batch_size = 7,
-    epochs = 8,
-    model_file = 9,
-    continue_train = 10,
-    unknown = 11,
+    batch_size = 2,
+    epochs = 3,
+    model_file = 4,
```
zhoonit, line 273: I like this :+1:

api/capi/src/nntrainer.cpp
kparichay: If there is any previous dataset, then there are two cases - 
1. Already set dataset was created using ini file. In this case, it is a shared_ptr internally and will be automatically freed.
2. Already set dataset was set by user. In this case, user has access to ml_train_dataset_h which must be appropriately freed by the user using ml_train_dataset_destroy.

api/capi/src/nntrainer.cpp
kparichay: done.

api/capi/src/nntrainer.cpp
kparichay: added returning error.

nntrainer/src/databuffer.cpp
kparichay: That was erronous. Removed.

nntrainer/include/databuffer_func.h
kparichay: My intension is to not define the same function pointer twice. I updated the dependency of `databuffer_func.h` to `nntrainer.h` than `nntrainer_internal.h`. Does that satisfy the no circular dependency?
I do not understand what you mean by aliasing?

api/capi/src/nntrainer.cpp
zhoonit: Thanks :)

nntrainer/include/databuffer_func.h
zhoonit: I might have picked wrong term. I mean... header will end up referenced both ways. (capi -> lib) and (lib <- capi)

https://github.com/nnstreamer/nntrainer/blob/f45060b9d570bdb61063790dc5ac6c1aace26646/nntrainer/meson.build#L1-L4

I think `../api/capi/include` shouldn't be included to `nntrainer_inc` if possible.
If that's the case, nntrainer-devel will be super set of capi-nntrainer-devel,
If there is `api/python` or `api/cpp` binding, it has to include `api/capi` as well and so on..

Overall making it hard to manage the package.

What I meant was this would be better if is sth like:

##### in `nntrainer.h`
```cpp
#include "databuffer_func.h"
typedef datagen_cb_ptr ml_train_datagen_cb
```

##### in `databuffer_func.h`
```cpp
// remove #include "nntrainer.h" 
typedef int (*datagen_cb_ptr)(float **input, float **label, bool *last);
```

kparichay: > #include "databuffer_func.h"
> typedef datagen_cb_ptr ml_train_datagen_cb

Using this would expose databuffer_func.h to the outside.
I will get comments from other members and update correspondingly.

kparichay: As per discussion with @myungjoo, the proposed change by this PR is fine.

Applications/Classification/jni/Android.mk
```diff
@@ -9,7 +9,9 @@ endif
 
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../libs/arm64-v8a
-NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include
+NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include \
+	$(LOCAL_PATH)/../../../api/capi/include \
```
jijoongmoon, line 13: as far as I know, we are not using capi for android. why do we need capi includes here? do we need them?

Applications/mnist/jni/Android.mk
```diff
@@ -9,7 +9,9 @@ endif
 
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../libs/arm64-v8a
-NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include
+NNTRAINER_INCLUDES := $(LOCAL_PATH)/../../../nntrainer/include \
+	$(LOCAL_PATH)/../../../api/capi/include \
+	$(LOCAL_PATH)/../../../api/capi/include/platform
```
jijoongmoon, line 14: mnist example does not use capi. Why do we need them here?

Applications/Classification/jni/Android.mk
kparichay: api/capi/incluide/platform is to provide ml-api-common.h which is used by nntrainer_error.h
api/capi/include is to provide nntrainer.h which is used by databuffer_func.h for data generator callback declaration

Applications/mnist/jni/Android.mk
kparichay: same as above.

nntrainer/meson.build
```diff
@@ -1,5 +1,6 @@
 nntrainer_inc = [
-  include_directories('./include')
+  include_directories('./include'),
+  include_directories('../api/capi/include')
```
jijoongmoon, line 3: I think we have to remove this include.

kparichay:
@jijoongmoon As per offline conversation, created a common API header for the common declarations and used it in nntrainer.

kparichay:
Rebased.

Issue #303, opened 2020-07-14
Data Buffer refactor


kparichay:
Implementation of data buffer needs quite some refactoring.

zhoonit:
As discussed offline, I am taking this issue

- [ ] Make a class that manages one buf type & current databuffer wraps around the class.
- [ ] Fine grained buffer (validation) states 
- [ ] Change vector to a struct containing void *, size and return

kparichay:
@zhoonit It is desirable for databuffer to use tensor to allocate the memory. However, any other dependency onto tensor is not required as databuffer just treats memory as void * with bytes size.
Also, when refactor, please make some space for adding multiple parallel thread readers (in future, if not now).

zhoonit:
- related bugs #1322

kparichay:
Track https://github.com/pravega/gstreamer-pravega/issues/83

zhoonit:
Backlogging some undone works:

- [ ] Change property of the random data generator. 
- [ ] finalize return size as well.
- [ ] DataProducer::generator returns single element than batch (check if this is okay with tizen)
- [ ] change semantics of DataProducer::size()
- [ ] change signature of DataProducer::generator

zhoonit:
Add interrupt mechanism for the startFetchWorker

zhoonit:
There can be more PR's as per upcoming reviews, but #1510 is the planned last patch of the work.

zhoonit:
We have a mt support issue with less complicated locks left

PR #305, opened 2020-07-14
[data] Data buffer refactor

kparichay:
~Data buffer refactor~

~Related Issue: #303~

This patch adds data buffer v2
Data buffer v2 is designed to minimize memory copies and support many threads
The data buffer creates a buffer pool where memory is already allocated
The reader threads fill these allocated memory and reports back to the main thread
The allocated memory pool is allocated in batchsize units so that this memory can be directly
used in the tensor

The interface of the data buffer is just one where we set the callback function
The file interface will handled with the callback interface itself (create a callback to load the file data)

**Self evaluation:**

### DIFF
--- a/nntrainer/include/databuffer_v2.h
+++ b/nntrainer/include/databuffer_v2.h
@@ -1,0 +1,325 @@
+// SPDX-License-Identifier: Apache-2.0-only
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	databuffer_v2.h
+ * @date	3 September 2020
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is databuffer class for Neural Network
+ *
+ * @todo TODO: Support multi files for dataset with files
+ * @todo TODO: Support multi threads with more than 1 thread and use thread
+ * pooling
+ * @todo TODO: Support label size to be 0 for inference based scenarios
+ * @todo TODO: rename data buffer to dataset
+ * @todo TODO: manage with just 1 buffer
+ * @todo TODO: consider different data structure for buffer when using shuffle
+ */
+
+#ifndef __DATABUFFER_V2_H__
+#define __DATABUFFER_V2_H__
+#ifdef __cplusplus
+
+#include <condition_variable>
+#include <list>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <thread>
+#include <tuple>
+#include <vector>
+
+#include <nntrainer-api-common.h>
+#include <tensor.h>
+
+namespace nntrainer {
+
+/**
+ * @brief   States the collect thread running in the background
+ */
+enum class ThreadStates {
+  THREAD_NULL,            /**< not yet initialized */
+  THREAD_READY,           /**< initialized but not yet started */
+  THREAD_RUNNING,         /**< started and running */
+  THREAD_REQUEST_TO_STOP, /**< main thread has request to stop, background
+                             thread is in the process of being stopped */
+  THREAD_STOPPED, /**< background thread has stopped but not yet cleaned up */
+  THREAD_EPOCH_FINISHED, /**< background threads have finished the epoch */
+  THREAD_ERROR           /**< an error has occurred in background threads */
+};
+
+/**
+ * @brief   Dataset generator callback type declaration
+ */
+typedef std::function<std::remove_pointer<ml_train_datagen_cb>::type>
+  datagen_cb;
+
+/**
+ * @class   DataBuffer Data Buffers
+ * @brief   Data Buffer for read and manage data
+ */
+class DataBuffer_v2 {
+public:
+  /**
+   * @brief     Constructor
+   */
+  DataBuffer_v2() :
+    type(DataBufferType::DATA_BUFFER_UNKNOWN),
+    buffer_len(1),
+    avail_buffer_idx(0),
+    total_data_entries(0),
+    batch_size(1),
+    num_threads(1),
+    generator(nullptr),
+    gen_user_data(nullptr),
+    thread_state(ThreadStates::THREAD_NULL) {
+    buffer.clear();
+    batched_buffer.clear();
+    label_size.resize(1, 0);
+    input_size.resize(1, 0);
+  }
+
+  /**
+   * @brief     Destructor
+   */
+  ~DataBuffer_v2() { stop(); }
+
+  /**
+   * @brief     Initialize Buffer with set properties
+   * @throws std::invalid_argument
+   * @throws std::runtime_error
+   */
+  void init();
+
+  /**
+   * @brief     start the thread for collection the data
+   * @throws std::runtime_error
+   */
+  void start();
+
+  /**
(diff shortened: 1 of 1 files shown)


nntrainer/include/databuffer.h
```diff
@@ -187,17 +169,32 @@ class DataBuffer {
...
    * @param[in] BufferType training, validation, test
    * @param[in] outVec feature data ( minibatch size )
    * @param[in] outLabel label data ( minibatch size )
    * @retval    true/false
    */
-  virtual bool getDataFromBuffer(
-    BufferType type,
-    std::vector<std::vector<std::vector<std::vector<float>>>> &out_vec,
-    std::vector<std::vector<std::vector<std::vector<float>>>> &out_label);
+  virtual std::pair<Tensor, Tensor> getDataFromBuffer(BufferType type);
```
zhoonit, line 197: Need update document.

zhoonit:
Could you add what is being changed with this PR?

nntrainer/include/databuffer.h
kparichay: Updated.
As an extra note, this interface allows getting data from generator/file without needing to create multiple copies.

nntrainer/include/databuffer.h
```diff
@@ -215,40 +211,19 @@ class DataBuffer {
    */
   int setBufSize(unsigned int n);
 
-  /**
-   * @brief     set mini batch size
-   * @param[in] mini batch size
-   * @retval #ML_ERROR_NONE Successful.
-   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
-   */
-  int setMiniBatch(unsigned int n);
```
kparichay, line 224: setFeatureSize() already gives in batch size.

kparichay:
@zhoonit @jijoongmoon Please review this once. This updates the interface. The implementation follows in #514. 
It has been split to make review easy. Putting everything at once in #514 made review tough as noted by @jijoongmoon.

nntrainer/include/databuffer_v2.h
```diff
@@ -0,0 +1,325 @@
...
+namespace nntrainer {
+
+/**
+ * @brief   States the collect thread running in the background
+ */
+enum class ThreadStates {
+  THREAD_NULL,            /**< not yet initialized */
+  THREAD_READY,           /**< initialized but not yet started */
+  THREAD_RUNNING,         /**< started and running */
+  THREAD_REQUEST_TO_STOP, /**< main thread has request to stop, background
```
zhoonit, line 46: Just a suggestion. `THREAD_DONE` ? `THREAD_JOINABLE` ?

nntrainer/include/databuffer_v2.h
```diff
@@ -0,0 +1,325 @@
...
+                             thread is in the process of being stopped */
+  THREAD_STOPPED, /**< background thread has stopped but not yet cleaned up */
+  THREAD_EPOCH_FINISHED, /**< background threads have finished the epoch */
+  THREAD_ERROR           /**< an error has occurred in background threads */
```
zhoonit, line 50: I think it'd be better if we separate this to threadResult? because those state are not mutually exclusive.

eg. A thread can be  `thread_request_to_stop` and `thread_error`

nntrainer/include/databuffer_v2.h
```diff
@@ -0,0 +1,325 @@
...
+   */
+  void stop();
+
+  /**
+   * @brief     Get data from buffer to the passed input and label vectors
+   * @param[in] inputs list of input tensors
+   * @param[in] inputs list of label tensors
+   * @retval false if end of epoch, else true
+   */
+  int getdata(std::vector<sharedtensor> &inputs,
```
zhoonit, line 114: How about `getData` for consistency?

zhoonit: Can sharedTensor with size be used instead to match the arguments with Layer::forwarding?

nntrainer/include/databuffer_v2.h
```diff
@@ -0,0 +1,325 @@
...
+
+    if (num_labels != label_size.size()) {
+      label_size.resize(num_labels, 0);
+    }
+  }
+
+  /**
+   * @brief     set the size of the label data
+   * @param[in] bytes size in bytes
+   * @param[in] idx index of the label
```
zhoonit, line 148: Q. What does index do in this function?

nntrainer/include/databuffer_v2.h
```diff
@@ -0,0 +1,325 @@
...
+  void setLabelSize(const size_t bytes, const unsigned int idx = 0) {
+    if (bytes == 0)
+      throw std::invalid_argument("Label size should be more than 0");
+
+    if (idx >= label_size.size()) {
+      if (label_size.size() > 1)
+        throw std::invalid_argument(
+          "Index exceeds the total size set for the label");
+      label_size.resize(idx + 1);
+    }
```
zhoonit, line 160: Is this intended that resize only `if labe_size.size() == 1`?

nntrainer/include/databuffer_v2.h
kparichay: how aobut `THREAD_REQUESTED_TO_STOP`. IMO these are more verbose and reader understands what it means.
THREAD_DONE/THREAD_JOINABLE means thread is finished. THREAD_WAITING is too generic.

nntrainer/include/databuffer_v2.h
kparichay: yes. The default label size is 1. If the caller does not edit the label size and tries to directly set a value outside this range, then it is allowed (determining size as caller sets the value). However, if the caller sets the label size to be higher, and then tries to set size outside the range, that's an error.

nntrainer/include/databuffer_v2.h
kparichay: right.

nntrainer/include/databuffer_v2.h
kparichay: Yes, they are not mutually exclusive. However, they have priorities in a way. 
Once thread state is THREAD_ERROR, no more state change.
It cant be in both states.

nntrainer/include/databuffer_v2.h
zhoonit: Got the meaning. let's not change.

nntrainer/include/databuffer_v2.h
zhoonit: Acked

nntrainer/include/databuffer_v2.h
```diff
@@ -0,0 +1,325 @@
...
+  /**
+   * @brief     function for thread to stop collection of the data
+   * @throws std::runtime_error
+   */
+  void stop();
+
+  /**
+   * @brief     Get data from buffer to the passed input and label vectors
+   * @param[in] inputs list of input tensors
+   * @param[in] inputs list of label tensors
```
dongju-chae, line 111: It's `getData` and where is the location of data stored? Perhaps, `@param[out]`?

kparichay: Thanks for the review.
Currently this patch has been abandoned as it is low priority.
This will be taken back in a couple of weeks. Then I will update as per your suggestion.

zhoonit:
As refactoring the databuffer is handled with different approach, this PR is closed

PR #306, opened 2020-07-14
[data buffer] Bug fix for data buffer

kparichay:
data buffer updateData() passes the reference of a on-stack variable
which is deleted after the function call exits. However the callee function
remains on a new thread causing problems.

Added bug fix for this.

### DIFF
--- a/nntrainer/src/databuffer_file.cpp
+++ b/nntrainer/src/databuffer_file.cpp
@@ -134,7 +134,7 @@ int DataBufferFromDataFile::init() {
   return status;
 }
 
-void DataBufferFromDataFile::updateData(BufferType type, int &status) {
+void DataBufferFromDataFile::updateData(BufferType type) {
   unsigned int max_size = 0;
   unsigned int buf_size = 0;
   unsigned int *rest_size = NULL;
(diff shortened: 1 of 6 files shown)


nntrainer/src/databuffer_file.cpp
```diff
@@ -237,7 +236,6 @@ void DataBufferFromDataFile::updateData(BufferType type, int &status) {
       try {
         if (position > file_length || position > ULLONG_MAX) {
           ml_loge("Error: Cannot exceed max file size");
```
zhoonit, line 238: Let's standardize not to use ml_log when throw since it will be handled :)
refering to: https://github.com/nnstreamer/nntrainer/pull/294#discussion_r453427550

nntrainer/src/databuffer_func.cpp
```diff
@@ -135,8 +135,8 @@ int DataBufferFromCallback::setFunc(
   return status;
 }
 
-void DataBufferFromCallback::updateData(BufferType type, int &status) {
-  status = ML_ERROR_NONE;
+void DataBufferFromCallback::updateData(BufferType type) {
+  int status = ML_ERROR_NONE;
```
zhoonit, line 139: should we throw when `status != ML_ERROR_NONE` at line 199? I can't comment at line 199

kparichay: The data buffer is going to refactored heavily with #305. It would be covered in it.

nntrainer/src/databuffer_file.cpp
kparichay: Done.

nntrainer/src/databuffer_func.cpp
zhoonit: Alright

PR #308, opened 2020-07-14
Refactor nntrainer load from config

zhoonit:
`loadFromConfig` has many duplicated logics with `layer::setProperty` and
others

This PR refactor `loadFromConfig` to reuse already present logic.

- Add `std::out_of_range` exception to setProperty to represent validity
- Change error to warning when input_dim is not present at the head of
the network(for ini)
- Change `weightIni` -> `weight_ini` in ini for consistency
- Change unittest accordingly
- Separate dataset, network parser

**Self evaluation:**


nntrainer/src/layer.cpp
```diff
@@ -211,7 +211,7 @@ void Layer::setProperty(const PropertyType type, const std::string &value) {
   default:
     std::string msg =
       "[Layer] Unknown Layer Property Key for value" + std::string(value);
-    throw std::invalid_argument(msg);
+    throw std::out_of_range(msg);
```
kparichay, line 214: out_of_range? IMO, out of range is for accessing out of bounds.

nntrainer/src/flatten_layer.cpp
```diff
@@ -59,7 +59,7 @@ Tensor FlattenLayer::backwarding(Tensor in, int iteration) {
 
 void FlattenLayer::setProperty(const PropertyType type,
                                const std::string &value) {
-  throw std::invalid_argument("[Flatten Layer] setProperty not supported");
+  throw std::out_of_range("[Flatten Layer] setProperty not supported");
```
kparichay, line 62: out_of_range? IMO, out of range is for accessing out of bounds.

nntrainer/src/neuralnet.cpp
```diff
@@ -132,71 +117,23 @@ int NeuralNetwork::setConfig(std::string config) {
...
   cost = (CostType)parseType(iniparser_getstring(ini, "Network:Cost", unknown),
                              TOKEN_COST);
-  model = iniparser_getstring(ini, "Network:Model", model_name);
+  model = iniparser_getstring(ini, "Network:Model", "model.bin");
   batch_size = iniparser_getint(ini, "Network:Minibatch", batch_size);
 
   /** Default to adam optimizer */
   status = opt.setType((OptType)parseType(
-    iniparser_getstring(ini, "Network:Optimizer", unknown), TOKEN_OPT));
+    iniparser_getstring(ini, "Network:Optimizer", "adam"), TOKEN_OPT));
```
kparichay, line 136: lets not set default value manually as this put default values in multiple places harder to update later. Rather create the OptParam object earlier, and take its getType() which will return the default type.

nntrainer/src/neuralnet.cpp
```diff
@@ -132,71 +117,23 @@ int NeuralNetwork::setConfig(std::string config) {
...
-    std::string sec_name_lower(sec_name);
-    std::transform(sec_name_lower.begin(), sec_name_lower.end(),
-                   sec_name_lower.begin(),
-                   [](unsigned char c) { return std::tolower(c); });
-    section_names.push_back(sec_name_lower);
-  }
-
-  /** Parse the Network section and its properties */
-  section_names_iter =
-    std::find(section_names.begin(), section_names.end(), "network");
```
kparichay, line 180: this is necessary to check that "network" section exists in the ini file.
otherwise we will have a default network created rather than giving an error.

nntrainer/src/neuralnet.cpp
```diff
@@ -213,252 +150,119 @@ int NeuralNetwork::loadFromConfig() {
...
-                            "xavier_uniform"),
-        TOKEN_WEIGHTINI));
+    if (!sec_name) {
+      ml_loge("Error: Unable to retrieve section names from ini.");
+      return ML_ERROR_INVALID_PARAMETER;
+    }
 
-      status = parseWeightDecay(ini, layer_name, weight_decay);
-      NN_INI_RETURN_STATUS();
+    if (strncasecmp("network", sec_name, 7) == 0) {
```
kparichay, line 224: create a temporary string obj, and then use length of it than 7?

nntrainer/src/neuralnet.cpp
```diff
@@ -213,252 +150,119 @@ int NeuralNetwork::loadFromConfig() {
...
+      status = loadNetworkConfig((void *)ini);
```
kparichay, line 225: with this implementation, we state that a valid ini file can be without `netwoork` section. Is that intended?

zhoonit: The code undergoes validation starting from `:298` it would fail if there is no network section.

It is moved to segregate validation phase for better readability (as this function is not critical w.r.t performance)

nntrainer/src/flatten_layer.cpp
zhoonit: There are two case for `std::invalid_argument`
1. when property type is not valid 🌵 
2. when given value is not valid 🍰 

I needed to differentiate 🌵 and 🍰 .
While it'd be better off to derive a custom exception to represent the case. I just made 🌵 `std::out_of_range` to save time considering that accessing invalid property is kinda similar to accessing out of bounds.

flatten layer case is 🌵 since every property type is invalid.

```cpp
  /**
   * ...
   * @exception std::out_of_range     when property type is not valid for the
   * particular layer
   * @exception std::invalid_argument invalid argument
   */
  virtual void setProperty(const PropertyType type,
                           const std::string &value = "");
```

nntrainer/src/neuralnet.cpp
```diff
@@ -467,21 +271,52 @@ int NeuralNetwork::loadFromConfig() {
...
-    /** Add flatten layer */
-    bool flatten =
-      iniparser_getboolean(ini, (layer_name + ":Flatten").c_str(), false);
-    layers.back()->setFlatten(flatten);
+    status = addLayer(layer);
+    NN_INI_RETURN_STATUS();
+  }
+
+  /**< Additional validation and handling for the neural network */
+  if (iniparser_find_entry(ini, "network") == 0) {
```
kparichay, line 298: how about doing in this the start, and calling `loadNetworkConfig((void *)ini);` in the else of this than in the loop above.
may be same of the data buffer - move outside the above loop.

nntrainer/src/neuralnet.cpp
```diff
@@ -467,21 +271,52 @@ int NeuralNetwork::loadFromConfig() {
...
   status = data_buffer->setMiniBatch(batch_size);
   NN_INI_RETURN_STATUS();
 
+  if (layers.empty()) {
+    ml_loge("there is no layer section in the ini file");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (layers.front()->getInputDimension() == TensorDim()) {
+    ml_logw("input dimension for the first layer seems not specified thus set "
```
kparichay, line 316: is this needed?? rather default tensor to be empty and then inputLayer will itself give error when dimension is not set.

nntrainer/src/flatten_layer.cpp
kparichay: As the handling of this case (`out_of_range`) is very simple (empty), I think it's not a big overhead to make a custom exception. It would make code more readable.

nntrainer/src/neuralnet.cpp
zhoonit: checked that this is verified in `NeuralNetwork::inti()` so removed.

nntrainer/src/neuralnet.cpp
zhoonit: moved this before loop

nntrainer/src/neuralnet.cpp
zhoonit: done

nntrainer/src/neuralnet.cpp
zhoonit: I'll leave this as an issue: setting sensible default for optimizer.

nntrainer/src/flatten_layer.cpp
zhoonit: Making it is not so hard though, If we are going to make a custom exception it should be managed globally because it need be caught by `nntrainer_exception_boundary`.  Considering this, do you have opinion to where to put custom exceptions? maybe `nntrainer_error.h`?

kparichay: nntrainer_error.h or util_func.h

zhoonit: Alright, I'll do that in other PR to control size

nntrainer/src/neuralnet.cpp
```diff
@@ -213,252 +150,128 @@ int NeuralNetwork::loadFromConfig() {
...
+      continue;
+    }
 
-      addLayer(fc_layer);
-    } break;
-    case LAYER_BN: {
-      std::shared_ptr<BatchNormalizationLayer> bn_layer =
-        std::make_shared<BatchNormalizationLayer>();
+    /** Parse all the layers defined as sections in order */
+    ml_logi("parsing layers properties... undefined properties will be ignored "
```
kparichay, line 246: `undefined properties will be filled with default values.`??
Also is this needed for every layer? Probably giving this out once should be enough.

test/nntrainer_test_util.cpp
```diff
@@ -62,6 +62,7 @@ void replaceString(const std::string &from, const std::string &to,
   }
 
   std::ofstream data_file(file.c_str());
+
```
kparichay, line 65: remove unnecessary additions?

test/unittest/unittest_nntrainer_internal.cpp
```diff
@@ -114,17 +114,18 @@ TEST(nntrainer_NeuralNetwork, load_config_03_n) {
...
+TEST(nntrainer_NeuralNetwork, load_config_04_p) {
   int status = ML_ERROR_NONE;
   RESET_CONFIG("./test.ini");
   replaceString("Input_Shape = 32:1:1:62720", "", "./test.ini", config_str);
   nntrainer::NeuralNetwork NN;
   status = NN.setConfig("./test.ini");
   EXPECT_EQ(status, ML_ERROR_NONE);
   status = NN.loadFromConfig();
-  EXPECT_EQ(status, ML_ERROR_INVALID_PARAMETER);
+  EXPECT_EQ(status, ML_ERROR_NONE);
```
kparichay, line 128: can you add a unittest here with `NN.init()` which fails when initializing a ini file based network with no input_shape?

zhoonit: Actually, I'm making a parameterized, automated test for load config. I'll include that in that PR. Thanks

kparichay: ok.

nntrainer/include/layer.h
```diff
@@ -537,11 +537,16 @@ class Layer {
    * @param[in] type property type to be passed
    * @param[in] value value to be passed, if empty string is passed, do nothing
    * but throws error when @a type is invalid
+   * @exception std::out_of_range     when property type is not valid for the
+   * particular layer
```
jijoongmoon, line 541: Is it out_of_range or invalid_parameter?

zhoonit: it is out_of_range for this PR. Later gonna make custom exception for this to keep PR size small(#315)

nntrainer/include/layer.h
```diff
@@ -537,11 +537,16 @@ class Layer {
...
    * @exception std::invalid_argument invalid argument
    */
   virtual void setProperty(const PropertyType type,
                            const std::string &value = "");
 
+  /// @todo move virtual void setProperty to public and remove this
+  friend class NeuralNetwork;
+
```
jijoongmoon, line 549: do you think it is necessary to make NeuralNetwork as friend?

zhoonit: `virtual void setProperty` have to move to public. But it's lengthy and trivial. so I made friend for now and going to patch this  in other PR.

test/unittest/unittest_nntrainer_internal.cpp
zhoonit: This should be dealt with #316

nntrainer/src/neuralnet.cpp
zhoonit: moved location and clarified the description as
```cpp
  ml_logi(
    "parsing ini... invalid properties does not cause error, rather be ignored "
    "only invalid value for valid property will abort the process");
```

test/nntrainer_test_util.cpp
zhoonit: done

nntrainer/include/layer.h
zhoonit: Patched at #320

nntrainer/include/layer.h
zhoonit: Patched at #320

PR #313, opened 2020-07-15
[CAPI] Add functions to be supported later

kparichay:
Add functions to be supported later in the internal capi header.

### DIFF
--- a/api/capi/include/nntrainer_internal.h
+++ b/api/capi/include/nntrainer_internal.h
@@ -120,6 +120,77 @@ typedef struct {
  */
 int ml_train_model_get_layer(ml_train_model_h model, const char *layer_name,
                              ml_train_layer_h *layer);
+/**
+ * @brief Get all neural network layer names from the model.
+ * @details Use this function to get already created Neural Network Layer names.
+ * This can be used to obtain layers when model is defined with ini file.
+ * @note The caller must free the list of the layer names.
+ * @since_tizen 6.x
+ * @param[in] model The NNTrainer model handler from the given description.
+ * @param[out] layers_name List of names of layers in the model ended with NULL.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Invalid parameter.
+ * @retval #ML_ERROR_CANNOT_ASSIGN_ADDRESS Cannot assign object.
+ */
+int ml_train_model_get_all_layer_names(ml_train_model_h model,
+                                       const char **layers_name[]);
+
+/**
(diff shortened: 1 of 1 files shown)


api/capi/include/nntrainer_internal.h
```diff
@@ -111,6 +111,75 @@ typedef struct {
...
+ * connection, then they should support multiple inputs. Otherwise
+ * #ML_ERROR_INVALID_PARAMETER is returned. If length of @a output_layer_names
+ * is 0, then this layer will be treated as one of the output layers, and a loss
+ * will be attached to this based on network configuration. If both @a
+ * input_layer_names and @a output_layer_names are empty, then this layer is
+ * attached at the end of the output layer of the network. In case of multiple
+ * output layers, this layer is attached next to the last created output layer.
+ */
+int ml_train_model_add_layer_v2(ml_train_model_h model, ml_train_layer_h layer,
+                                const char *input_layer_names[],
```
jijoongmoon, line 181: how about ml_train_model_add_layer_at?

kparichay: `_at` is a little incomplete. I actually wanted to keep same function name but different signature. Hence v2.
How about `ml_train_model_insert_layer`??

zhoonit: insert_layer looks good

kparichay: Done.

api/capi/include/nntrainer_internal.h
```diff
@@ -120,6 +120,76 @@ typedef struct {
...
+ * @since_tizen 6.x
+ * @param[in] model The NNTrainer model handler from the given description.
+ * @param[out] layers_name List of names of layers in the model ended with NULL.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Invalid parameter.
+ * @retval #ML_ERROR_CANNOT_ASSIGN_ADDRESS Cannot assign object.
+ */
+int ml_train_model_get_all_layer_names(ml_train_model_h model,
+                                       const char **layers_name[]);
```
zhoonit, line 136: should this be freed after use?

kparichay: Added in details.

zhoonit:
Are we gonna implement this API within this year? If not, I wonder if there is a way to tell this is not implemented yet?

kparichay:
> Are we gonna implement this API within this year? If not, I wonder if there is a way to tell this is not implemented yet?

Can return ML_ERROR_NOT_SUPPORTED, but given that they are not exposed outside, it should be fine.
On the side note though, probably some of them will have to be supported.

PR #320, opened 2020-07-16
Follow-up after #308

### LINKED ISSUE #315: Add custom exception for some errors
https://github.com/nnstreamer/nntrainer/pull/308#discussion_r454772057

1. Declare custom errors if needed (to `nntrainer_error.h`)
2. Handle it at least at capi exception boundary

zhoonit:
- Now friendship between layer and network is over
- Move `setProperty(propType)` to public
- Add custom exception in `_error.h`
- Change `setProperty` `std::out_of_range` ->
`exception::invalid_property`
- Add error code boundary to `nntrainer_exception_boundary`
- Change inherited docs to @copydoc

resolves #315 

**Self evaluation:**

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com

### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -23,8 +23,8 @@
 #include <databuffer.h>
 #include <databuffer_file.h>
 #include <databuffer_func.h>
-#include <ml-api-common.h>
 #include <neuralnet.h>
+#include <nntrainer_error.h>
 #include <nntrainer_internal.h>
 #include <nntrainer_log.h>
 #include <parse_util.h>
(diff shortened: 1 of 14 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -23,8 +23,8 @@
 #include <databuffer.h>
 #include <databuffer_file.h>
 #include <databuffer_func.h>
-#include <ml-api-common.h>
```
kparichay, line 26: huh.. why remove this? ml_train_dataset_create_with_generator() needs this.

zhoonit: I think '*api' can and should solely depend on the <nntrainer_error.h> since #307.
<nntrainer_error.h> is superset of <ml-api-common.h> we have to make a boundary for that.

(theoretically) declaring error code twice otherwise.

kparichay: aah, I confused with nntrainer-api-common.h

PR #323, opened 2020-07-16
[optimzer/layers] Gradient dimension should match weight dim and sum

### LINKED ISSUE #280: Gradient contains batch dimension
Gradients in the each layer being passed to the optimizers contain batch axis.
This should be averaged in each layer where gradients are stored than in the optimizers.

kparichay:
Gradient dimension should match weight dimension
Currently optimizer applies averaging of gradients, which is not correct
Apply averaging of gradients before calling applyGradients

Further, Gradients of the layer should perform sum over batch than average
This is consistent with tensorflow's implementation

Resolves #280

### DIFF
--- a/nntrainer/src/bn_layer.cpp
+++ b/nntrainer/src/bn_layer.cpp
@@ -53,10 +53,10 @@ int BatchNormalizationLayer::initialize(bool last) {
   beta.setZero();
 
   setParamSize(4);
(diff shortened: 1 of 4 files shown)


kparichay:
> LGTM. But needs following pr about initialize gradient tensor with same dimension of weight.

Yes, that's on way.

PR #324, opened 2020-07-17
[fc/activation] Unittest for fc, activation layers and sgd optimizer

kparichay:
Add combined unittest for fc and activation layers (sigmoid and softmax) backwarding
Note that this is done without the loss
As SGD is used for updates, this also acts as testing optimizer sgd

Fixes related to these tests were already merged earlier in #323


test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -347,6 +364,18 @@ class nntrainer_FullyConnectedLayer_TFmatch
     EXPECT_EQ(status, ML_ERROR_NONE);
   }
 
+  void loadUpdatedWeightsGradients(const char *file_uw, const char *file_g) {
+    for (int idx = 0; idx < 2; ++idx) {
+      new_w.push_back(
+        nntrainer::Tensor(layer.getParams().get()[idx].weight.getDim()));
```
zhoonit, line 370: There is safer version of accessing element. `layer::paramsAt(idx)` that checks range.

kparichay: This is much better.
However, that function is protected. Made it public.

zhoonit: Nice :+1:

Issue #326, opened 2020-07-17
Dont throw in case given property is not supported


kparichay:
check if the property is needed to be set or given by the user.
If the property given by user is not supported, throw error.

PR #330, opened 2020-07-17
Fix neuralnetwork property logic

zhoonit:
exception::invalid_property for not allowed property should not be
ignored for some cases. So, parsing process is patched accordingly

cc. @kparichay 

**Self evaluation:**


zhoonit:
Is this resolves #326 ?

kparichay:
> Is this resolves #326 ? @kparichay

Yes, it did. #326 is closed.

PR #331, opened 2020-07-17
[ Bug ] Fix initialize buffer for adam optimizer

jijoongmoon:
. Add initialize(std::shared_ptr<UpdatableParam> params, ...)
. Update initailize conv2d for optimizer
. Update to do addition for calcuate derivatives of filter

**Self evaluation:**

### DIFF
--- a/nntrainer/src/conv2d_layer.cpp
+++ b/nntrainer/src/conv2d_layer.cpp
@@ -32,31 +32,27 @@ int Conv2DLayer::initialize(bool last) {
 
   this->last_layer = last;
   dim = TensorDim(1, input_dim.channel(), kernel_size[0], kernel_size[1]);
+  TensorDim bias_dim = TensorDim(1, 1, 1, 1);
 
   std::string kernelPrefix = "Conv2d:filter";
   std::string biasPrefix = "Conv2d:bias";
   setParamSize(filter_size * 2);
 
-  // fixme: #280
-  TensorDim gradDim(input_dim.batch(), dim.channel(), dim.height(),
-                    dim.width());
-  TensorDim gradBiasDim(input_dim.batch(), 1, 1, 1);
-
   for (unsigned int i = 0; i < filter_size; ++i) {
     Tensor Knl = initializeWeight(dim, weight_ini_type, status);
     NN_RETURN_STATUS();
 
-    Tensor bias = Tensor(1, 1);
+    Tensor bias = Tensor(bias_dim);
 
     if (!bias_init_zero) {
       bias.apply([&](float x) { return random(); });
     } else {
       bias.setZero();
     }
-    Tensor delK(gradDim);
+    Tensor delK(dim);
     delK.setZero();
 
-    Tensor delBias(gradBiasDim);
+    Tensor delBias(bias_dim);
     delBias.setZero();
 
     /*< @note: order of weight and bias are:
@@ -125,20 +121,7 @@ int Conv2DLayer::setOptimizer(Optimizer &opt) {
   if (status != ML_ERROR_NONE)
     return status;
 
-  std::vector<Tensor> list_d;
-  for (unsigned int i = 0; i < filter_size; ++i) {
-    for (unsigned int j = 0; j < 2; ++j) {
-      list_d.push_back(Tensor(dim));
-    }
-    for (unsigned int j = 0; j < 2; ++j) {
-      list_d.push_back(Tensor(1, 1, 1));
-    }
-  }
-  std::vector<Tensor>::iterator iter;
-  for (iter = list_d.begin(); iter != list_d.end(); ++iter) {
-    (*iter).setZero();
-  }
-  return this->opt.initialize(list_d, true);
+  return this->opt.initialize(getParams(), param_size, true);
 }
 
 Tensor Conv2DLayer::backwarding(Tensor derivative, int iteration) {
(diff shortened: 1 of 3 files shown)


nntrainer/src/conv2d_layer.cpp
```diff
@@ -138,7 +140,8 @@ int Conv2DLayer::setOptimizer(Optimizer &opt) {
   for (iter = list_d.begin(); iter != list_d.end(); ++iter) {
     (*iter).setZero();
   }
-  return this->opt.initialize(list_d, true);
```
kparichay, line 141: do we need creation and assignment of list_d now?

nntrainer/src/conv2d_layer.cpp
```diff
@@ -53,10 +53,10 @@ int Conv2DLayer::initialize(bool last) {
     } else {
       bias.setZero();
     }
-    Tensor delK(gradDim);
+    Tensor delK(dim);
```
kparichay, line 56: let's fix the value of `gradDim` to be used than setting value here manually with `dim`. Or remove `gradDim` variable.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -53,10 +53,10 @@ int Conv2DLayer::initialize(bool last) {
...
     delK.setZero();
 
-    Tensor delBias(gradBiasDim);
+    Tensor delBias(1, 1);
     delBias.setZero();
```
kparichay, line 61: ditto.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -152,7 +155,7 @@ Tensor Conv2DLayer::backwarding(Tensor derivative, int iteration) {
   for (unsigned int b = 0; b < input_dim.batch(); ++b) {
     Tensor in_padded = zero_pad(b, input, padding);
     TensorDim p_dim(1, 1, in_padded.height(), in_padded.width());
-
+    float sum = 0.0;
     /// @fixme: #280
     for (unsigned int i = 0; i < filter_size; i++) {
       Tensor &delK = paramsAt(i).grad;
```
kparichay, line 161: delK should be initialized with `setZero` here to remove the effect of training from previous iteration.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: yes. better to remove. Thanks.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: yes. no need. forgot to remove. thanks.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: right. thanks.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -53,10 +48,10 @@ int Conv2DLayer::initialize(bool last) {
     } else {
       bias.setZero();
     }
-    Tensor delK(gradDim);
+    Tensor delK(dim);
     delK.setZero();
 
-    Tensor delBias(gradBiasDim);
+    Tensor delBias(1, 1);
```
kparichay, line 54: Lets not use `(1, 1)` but use the dimension of the bias. Its more readable.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -164,19 +154,20 @@ Tensor Conv2DLayer::backwarding(Tensor derivative, int iteration) {
...
       // Calculate delBias [ batch , 1, 1, filter_size]
 
-      float sum = 0.0;
       for (unsigned int j = 0; j < derivative.height(); ++j) {
         for (unsigned int k = 0; k < derivative.width(); ++k) {
           sum += derivative.getValue(b, i, j, k);
         }
       }
-      delBias.setValue(b, 0, 0, 0, sum);
+      delBias.setValue(0, 0, 0, 0, sum);
```
kparichay, line 170: Should this be doing addition than just setValue??

nntrainer/src/optimizer.cpp
```diff
@@ -58,15 +58,25 @@ int Optimizer::setOptParam(OptParam p) {
...
+    for (unsigned int i = 0; i < param_size; ++i) {
+      UpdatableParam &param = param_data[i];
+
+      Tensor &weight = param.weight;
+      Tensor &grad = param.grad;
+      Tensor w = Tensor(weight.getDim());
+      w.setZero();
+      Tensor g = Tensor(grad.getDim());
+      g.setZero();
+      weight_mv.push_back(std::pair<Tensor, Tensor>(w, g));
```
zhoonit, line 77: This would avoid unnecessary copy
```suggestion
      std::pair<Tensor, Tensor> p = std::make_pair(std::move(w), std::move(g));
      weight_mv.push_back(std::move(p));
```

nntrainer/src/conv2d_layer.cpp
```diff
@@ -164,19 +154,20 @@ Tensor Conv2DLayer::backwarding(Tensor derivative, int iteration) {
           derivative.getAddress(b * derivative.getDim().getFeatureLen() +
                                 i * derivative.height() * derivative.width()),
           in_dim, output.data(), stride, 0.0);
-        memcpy(delK.getAddress(b * delK.getDim().getFeatureLen() + j * o_size),
-               output.data(), o_size * sizeof(float));
+        float *del = delK.getAddress(j * o_size);
+        for (unsigned k = 0; k < o_size; ++k) {
+          del[i] += output[i];
```
zhoonit, line 159: Can `/// @fixme: #280` go away?

jijoongmoon: yes. I think so. we are not going to use gradient with batch size from now on.

nntrainer/src/optimizer.cpp
jijoongmoon: fixed. thanks.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: fixed. thanks.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: fixed. thanks.

PR #344, opened 2020-07-21
[activation] Update signature of activation prime

kparichay:
Update signature to accommodate derivative in activation_prime
This allows softmax_prime to be managed with current activation_layer class

Activation unit tests out of util
Set error tolerance level to e-6
This has exposed a bug with batch normalization unit test - their golden values are 0
So, disabled it for now
@zhoonit please check it


nntrainer/src/activation_layer.cpp
```diff
@@ -257,6 +200,52 @@ Tensor ActivationLayer::softmax(Tensor const &t) {
...
+  int height = x.getHeight();
+  bool is_derivative = true;
+
+  Tensor PI = Tensor(x.getDim());
+
+  const float *xp = x.getData();
+  const float *d = derivative.getData();
+  float *pp = PI.getData();
+
+  /** @todo update default tensorDim to be 0 and not 1 */
```
zhoonit, line 217: What'd be the semantics of 0 tensor dim?
for example `2:3:4:0`?
tensorDim is not allowed to be 0 it might cause bugs otherwise. Even though If we allow this, it should not be default.

zhoonit: If we need `NULL` for derivative, I think we could use 0 but should be only exposed by sth like `TensorDim::getNull()`.

kparichay: The default `tensor()` creates a tensor of dimension `1:1:1:1` and also allocates memory for it.
IMO the default() tensor must be an empty tensor with no memory allocated.

zhoonit: I think it's similar to std::string() creates `""`, the minimal possible value for the existence to avoid abnormal cases. otherwise we need extra validation before using tensor.

For example, user creates
1.  `Tensor()`
2. set Tensor height to  2
3. now tensor has shape of "0:0:2:0" so it is not usable.

too check emptiness, `std::optional` strategy seems better option(it is c++17 though, concept is not hard to implement, or we could bump to c++17)

cf) https://en.cppreference.com/w/cpp/utility/optional

kparichay: Yes, one of these must be done,  as a separate PR.

jijoongmoon: Let's make another issue about it. (#359)

PR #354, opened 2020-07-22
[API] Make C-API thread safe

### LINKED ISSUE #287: make nntrainer thread safe
Lets make nntrainer thread safe atleast from the API.

kparichay:
Added thread locking safety for ml_train C-API
There is a global lock to assist with destruction of object
Each object gets its own lock which maintains safety when accessing/modifying that object

Resolves #287

### DIFF
--- a/api/capi/include/nntrainer_internal.h
+++ b/api/capi/include/nntrainer_internal.h
@@ -28,6 +28,7 @@
 #define __NNTRAINER_INTERNAL_H__
 
 #include <layer.h>
+#include <mutex>
 #include <neuralnet.h>
 #include <nntrainer.h>
 #include <nntrainer_log.h>
@@ -43,29 +44,36 @@ extern "C" {
 
 /**
  * @brief Struct to wrap neural network layer for the API
+ * @note model mutex must be locked before layer lock, if model lock is needed
  */
 typedef struct {
   uint magic;
   std::shared_ptr<nntrainer::Layer> layer;
   bool in_use;
+  std::mutex m;
 } ml_train_layer;
 
 /**
  * @brief Struct to wrap neural network optimizer for the API
+ * @note model mutex must be locked before optimizer lock, if model lock is
+ * needed
  */
 typedef struct {
   uint magic;
   std::shared_ptr<nntrainer::Optimizer> optimizer;
   bool in_use;
+  std::mutex m;
 } ml_train_optimizer;
 
 /**
  * @brief Struct to wrap data buffer for the API
+ * @note model mutex must be locked before dataset lock, if model lock is needed
  */
 typedef struct {
   uint magic;
   std::shared_ptr<nntrainer::DataBuffer> data_buffer;
   bool in_use;
+  std::mutex m;
 } ml_train_dataset;
 
 /**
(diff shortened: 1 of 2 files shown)


api/capi/include/nntrainer_internal.h
```diff
@@ -77,12 +85,13 @@ typedef struct {
   std::unordered_map<std::string, ml_train_layer *> layers_map;
   ml_train_optimizer *optimizer;
   ml_train_dataset *dataset;
+  std::mutex m;
```
myungjoo, line 88: Are you sure you want to add a lock to every hierarchy here?

It is usually the root of deadlocks and excessive overheads.

api/capi/src/nntrainer.cpp
```diff
@@ -231,45 +253,62 @@ int ml_train_model_destroy(ml_train_model_h model) {
...
-  }
-
+static int ml_train_model_get_summary_util(ml_train_model_h model,
+                                           ml_train_summary_type_e verbosity,
+                                           std::stringstream &ss) {
   int status = ML_ERROR_NONE;
   ml_train_model *nnmodel;
 
-  ML_NNTRAINER_CHECK_MODEL_VALIDATION(nnmodel, model);
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
```
jijoongmoon, line 285: I don't think we need to lock when we just display summary of network.

api/capi/src/nntrainer.cpp
```diff
@@ -388,7 +433,9 @@ int ml_train_model_get_layer(ml_train_model_h model, const char *layer_name,
                              ml_train_layer_h *layer) {
   int status = ML_ERROR_NONE;
   ml_train_model *nnmodel;
-  ML_NNTRAINER_CHECK_MODEL_VALIDATION(nnmodel, model);
+
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
```
jijoongmoon, line 438: Do we need to set the lock when user just get the layer? May be lock when set the layer value or change value of layer?

api/capi/src/nntrainer.cpp
kparichay: We need lock for summary as well because a user can delete the network while getting the summary.

api/capi/src/nntrainer.cpp
kparichay: As getting the layer involves accessing the model elements, and user can delete the model, this lock is needed.
Later, we can move to reader/writer locks for better performance.

api/capi/include/nntrainer_internal.h
kparichay: In the current implementation, each object (mode, layer, dataset, and optimizer) have their own memory for the wrapper along with a shared_ptr to the actual object which is used internally. The lock in each object is to protect itself from being deleted while setting its own properties as each object has its own create(), set_property() and destroy() functions. The locks don't try to protect the shared_ptr but just the wrapper itself.

These locks are acquired together when an object is added into another (adding layer/optimizer/dataset to a model). The hierarchy of taking locks is also fixed - Model lock is always acquired first, and then layer/optimizer/dataset locks are taken.

Can you please provide some hints to reduce the usage of locks here.

PR #363, opened 2020-07-23
[C-API] C-API updated with tizen ACR

kparichay:
Updating C-API with tizen ACR

### DIFF
--- a/nntrainer/src/databuffer_func.cpp
+++ b/nntrainer/src/databuffer_func.cpp
@@ -204,7 +204,7 @@ void DataBufferFromCallback::updateData(BufferType type) {
     testReadyFlag = DATA_NOT_READY;
     if (buf_size - (*cur_size) > 0) {
       /** @todo Update to support multiple inputs later */
-      status = callback(vec_arr, veclabel_arr, &endflag);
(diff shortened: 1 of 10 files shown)


kparichay:
This PR is to update the Tizen C-API headers and documentation.
Corresponding changes in source and unit tests/applications will be reflected once the API review moves forward.

nntrainer/src/databuffer_func.cpp
```diff
@@ -204,7 +204,7 @@ void DataBufferFromCallback::updateData(BufferType type) {
     testReadyFlag = DATA_NOT_READY;
     if (buf_size - (*cur_size) > 0) {
       /** @todo Update to support multiple inputs later */
-      status = callback(vec_arr, veclabel_arr, &endflag);
+      status = callback(vec_arr, veclabel_arr, &endflag, NULL);
```
zhoonit, line 207: Is this okay not to pass userdata in this context?

kparichay: Currently, we don't have user data.
This will be added to the properties and then passed once data buffer is refactored.

PR #366, opened 2020-07-24
[ Coverity ] Fix Coverity Issues

jijoongmoon:
This PR incluide Fixs of Coverity Issues.

**Self evaluation:**

### DIFF
--- a/Applications/Classification/jni/main.cpp
+++ b/Applications/Classification/jni/main.cpp
@@ -119,10 +119,6 @@ static int rangeRandom(int min, int max) {
  * @param[out] feature_input save output of tflite
  */
 void getFeature(const string filename, vector<float> &feature_input) {
-  int input_size;
-  int output_size;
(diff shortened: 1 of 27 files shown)


nntrainer/src/bn_layer.cpp
```diff
@@ -34,6 +34,20 @@ namespace nntrainer {
 
 enum class BNParams { mu, var, gamma, beta };
 
+BatchNormalizationLayer &BatchNormalizationLayer::
+operator=(BatchNormalizationLayer &&rhs) noexcept {
```
kparichay, line 38: Do we need to create this move constructor? In my knowledge, it is recommended to not write a copy/move constructor if it is the same as the default generated one (which I think in this case and below cases is).

jijoongmoon: In order to pass the coverity issues, we may have to specify explicitly I guess.

Issue #374, opened 2020-07-29
Refactor plans for negative cases


zhoonit:
#### Add way to check applications 
- [x] Add automated test to check sanity of applications (#375)
  - [x] returns 0 for proper input (positive test)
  - [x] ~returns 0 for inproper input (negative test)~ (cancelled for now)
  - [x] ~ensure that output does not change (maybe?)~ (cancelled for now)

#### Add validation
- [x] resolve undefined behavior of network::init and enable negative ini cases (#316) (#384)
- [x] use delegates strategy for checkValidation(#387) -> postphoned and closed in this umbrella

#### Prune unnecessary functions / variable
- [x] Delete cost, bn_follow from layer.h (#384)
- [x] Unify optimizer initialize to have only params (#381)
- [x] ~add validation initLossLayer~ (Done in `neuralnet::init`)
- [x] change exception::invalid_property -> exception::not_supported (#380)

#### Minor refactors as well
- [x] Delete VERIFY_SET_DIMENSION (#384)
- [x] is_file_exist -> isFileExist (#384)

zhoonit:
Project done, waiting for review and merge

zhoonit:
will be closed when #393 is merged

PR #384, opened 2020-07-29
Refactor neuralnet::init

### LINKED ISSUE #316: Undefined behavior when loadFromCompile fail and init
There should be a validation bit (or method) to save status of `loadFromCompile` (simillar to `file.good()`)

Currently, even if loadFromCompile fail, init is callable without error and lead to undefined behavior.

zhoonit:
- Add determinancy and check to
- Delete cost / bn_follow from layer
- Enable negative test for modelfile
- fix IniSection::setEntry is wrongly using key for an erase behavior
- Delete `VERIFY_SET_DIMENSION()`

Resolves #316
See also #374

**Self evaluation:**

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -43,21 +43,6 @@
     }                              \
   } while (0)
 
-#define VERIFY_SET_DIMENSION()                                     \
-  do {                                                             \
-    if (i != 0) {                                                  \
-      if (def_init_dim == layers[i]->getInputDimension()) {        \
-        layers[i]->setInputDimension(previous_dim);                \
-      } else if (previous_dim != layers[i]->getInputDimension()) { \
-        status = ML_ERROR_INVALID_PARAMETER;                       \
-        ml_loge("Dimension mismatch between layers.");             \
-        NN_RETURN_STATUS();                                        \
-      }                                                            \
-    }                                                              \
-  } while (0)
-
-#define CONV2D_DIM 2
-
 namespace nntrainer {
 
 /**
@@ -65,7 +50,7 @@ namespace nntrainer {
  * @param[in] filename file path to check
  * @retval    boolean true if exists
  */
-static bool is_file_exist(std::string file_name) {
+static bool isFileExist(std::string file_name) {
   std::ifstream infile(file_name);
   return infile.good();
 }
@@ -88,7 +73,7 @@ NeuralNetwork::NeuralNetwork(std::string config) : NeuralNetwork() {
 }
 
 void NeuralNetwork::setConfig(std::string config) {
-  if (!is_file_exist(config)) {
+  if (!isFileExist(config)) {
     std::stringstream ss;
     ss << "Cannot open model configuration file, filename : " << config;
     throw std::invalid_argument(ss.str().c_str());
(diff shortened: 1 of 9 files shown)


nntrainer/src/neuralnet.cpp
```diff
@@ -438,92 +433,54 @@ int NeuralNetwork::setTrainConfig(std::vector<std::string> values) {
...
-        std::static_pointer_cast<Conv2DLayer>(layers[i]);
-
-      status = layers[i]->setCost(cost);
-      NN_RETURN_STATUS();
-
-      status = layers[i]->initialize(last);
-      NN_RETURN_STATUS();
-
-      status = conv2d_layer->setOptimizer(opt);
-      NN_RETURN_STATUS();
```
jijoongmoon, line 467: so, we cannot set different optimizer for each layer then. Am I right?

zhoonit: This PR does not change the behavior as per setting optimizer. As per now, yes.

nntrainer/src/neuralnet.cpp
```diff
@@ -320,18 +314,21 @@ int NeuralNetwork::initLossLayer() {
...
   if (updated_cost == COST_ENTROPY) {
     if (layers.back()->getType() != LAYER_ACTIVATION) {
       ml_loge("Error: Cross Entropy need last layer to have softmax or sigmoid "
               "activation.");
       return ML_ERROR_NOT_SUPPORTED;
     }
 
-    std::shared_ptr<ActivationLayer> act_layer =
-      std::static_pointer_cast<ActivationLayer>(layers.back());
+    std::shared_ptr<Layer> act_layer = layers.back();
```
kparichay, line 328: this variable is now unused?

nntrainer/src/neuralnet.cpp
```diff
@@ -320,18 +314,21 @@ int NeuralNetwork::initLossLayer() {
...
     layers.pop_back();
 
-    switch (act_layer->getActivationType()) {
+    switch (layers.back()->getActivationType()) {
```
kparichay, line 331: This might lead to nasty bugs. Now, you are getting the activation type from the previous layer which is not activation layer anymore.
Cache this value before popping out the activation layer, and use from it than from the last layer.

nntrainer/src/neuralnet.cpp
```diff
@@ -438,92 +433,54 @@ int NeuralNetwork::setTrainConfig(std::vector<std::string> values) {
...
-
-      if (i == 0) {
+      if (first) {
         ml_loge("Batch normalization layer cannot be first layer of network");
         status = ML_ERROR_INVALID_PARAMETER;
         NN_RETURN_STATUS();
       }
-      layers[i - 1]->setBNfollow(true);
+      /// no break intended
+    case LAYER_CONV2D:
```
kparichay, line 467: can you add something like fallthrough is intended?

nntrainer/src/neuralnet.cpp
zhoonit: Yes. Now me is spoiling past me's work lol. will fix that.

nntrainer/src/neuralnet.cpp
jijoongmoon: I think we have plan to try change opt for a layer sooner or later. If we are doing this soon, then we have to reverse current implementation?

zhoonit: We do not need to revert this patch. It will need less amount of change after this patch since we only have to change `layer::setOptimizer.` (currently, `conv2d::setOptimizer`, `layer::setOptimizer`, `fc::setOptimizer`...  and so on)

nntrainer/src/neuralnet.cpp
zhoonit: Yup. Fixed :)

nntrainer/src/neuralnet.cpp
zhoonit: It is fixed

PR #385, opened 2020-07-30
Update setDataset / loadDataset

### LINKED ISSUE #379: NeuralNetwork::loadDatasetConfig is overvalidated
```cpp
int NeuralNetwork::loadDatasetConfig(void *_ini) {
   /* ... */
    data_buffer = std::make_shared<DataBufferFromDataFile>();
    std::shared_ptr<DataBufferFromDataFile> dbuffer =
      std::static_pointer_cast<DataBufferFromDataFile>(data_buffer);

    status = dbuffer->setDataFile(
      iniparser_getstring(ini, "DataSet:TrainData", ""), DATA_TRAIN);
    NN_INI_RETURN_STATUS();
    status = dbuffer->setDataFile(
      iniparser_getstring(ini, "DataSet:ValidData", ""), DATA_VAL);
    NN_INI_RETURN_STATUS();
    status = dbuffer->setDataFile(
      iniparser_getstring(ini, "DataSet:TestData", ""), DATA_TEST);
    NN_INI_RETURN_STATUS();
    status = dbuffer->setDataFile(
      iniparser_getstring(ini, "DataSet:LabelData", ""), DATA_LABEL);
    NN_INI_RETURN_STATUS();
    /// fixme: #299
    status = data_buffer->setBufSize(
      iniparser_getint(ini, "DataSet:BufferSize", batch_size));
    NN_INI_RETURN_STATUS();

   /* ... */
}
```

This shouldn't fail when there is no valid, or test dataset in accordance with CAPI.

I am not quite understanding the bufsize.

Can this be omitted @kparichay ?

zhoonit:
When path is given and invalid for dataset, it should fail. This patch updates
accordingly.

resolves #379 

**Self evaluation:**
2. Run test: [X]Passed [ ]Failed []Skipped

### DIFF
--- a/nntrainer/src/databuffer_file.cpp
+++ b/nntrainer/src/databuffer_file.cpp
@@ -290,20 +290,19 @@ int DataBufferFromDataFile::setDataFile(std::string path, DataType type) {
   case DATA_VAL: {
     validation[type] = true;
     if (!data_file.good()) {
-      ml_logw("Warning: Cannot open validation data file. Cannot validate "
+      ml_loge("Error: Cannot open validation data file. Cannot validate "
               "training result");
       validation[type] = false;
-      break;
+      return ML_ERROR_INVALID_PARAMETER;
     }
     val_name = path;
   } break;
   case DATA_TEST: {
     validation[type] = true;
     if (!data_file.good()) {
-      ml_logw(
-        "Warning: Cannot open test data file. Cannot test training result");
+      ml_loge("Error: Cannot open test data file. Cannot test training result");
       validation[type] = false;
-      break;
+      return ML_ERROR_INVALID_PARAMETER;
     }
     test_name = path;
   } break;
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -132,6 +132,7 @@ int NeuralNetwork::loadNetworkConfig(void *_ini) {
   return status;
 }
 
+/// @fixme: 370
 int NeuralNetwork::loadDatasetConfig(void *_ini) {
   int status = ML_ERROR_NONE;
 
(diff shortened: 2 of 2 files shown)


nntrainer/src/databuffer_file.cpp
```diff
@@ -289,20 +289,19 @@ int DataBufferFromDataFile::setDataFile(std::string path, DataType type) {
   case DATA_VAL: {
     validation[type] = true;
     if (!data_file.good()) {
-      ml_logw("Warning: Cannot open validation data file. Cannot validate "
+      ml_loge("Error: Cannot open validation data file. Cannot validate "
               "training result");
       validation[type] = false;
-      break;
+      return ML_ERROR_INVALID_PARAMETER;
```
jijoongmoon, line 295: I think the reason we set the break is that we still can run the training without validation data and test data set.

zhoonit: Yes. `setSomething` should validates if given input is correct if not return error. It is the responsibility of caller how the error should treated.
e.g) `setValidationDataSet` should return error when it `validationDataSet` is not present. because it was not successful. 

This patch do as follows
1. if `dataset:validData` key is not present, it does not call `setDataFile` because it is not necessary
2. if `dataset:validData` key is present but path is invalid, `loadConfig` fails
3. if `dataset:trainData` key is not present, `loadConfig` fails

jijoongmoon: Then, I think we need some kind of mode. We have to check at least train data is provided properly when Training mode. And also have to add run validation Mode only if no train data and validation data is provided without train.

zhoonit: > Then, I think we need some kind of mode. We have to check at least train data is provided properly when Training mode. And also have to add run validation Mode only if no train data and validation data is provided without train. I mean we have to run the mode according to which data set is provided.

We can do this but that issue is not related to this PR. 

Setter still have to verify if current input is valid. and it is caller's responsibility whether to call this or not.  It is not related to whether is is required property or not.

In other words, It is expected behavior that if given input is not valid, setter return error.

For example, our capi expects to receive error when file is not valid and current implementation of `setDataFile` is actually causing bug to capi. ( eg) when `valid_file` is not found )

https://github.com/nnstreamer/nntrainer/blob/c16261fd34fe0cad582638fa62d4da5208d1b614/api/capi/src/nntrainer.cpp#L674-L714

PR #394, opened 2020-07-31
Print log file and line for better debuging

zhoonit:
- Add file and line to the logger
- Delete log tag for linux distro (as it is obvious and consumes space)

**Self evaluation:**

### DIFF
--- a/nntrainer/include/nntrainer_log.h
+++ b/nntrainer/include/nntrainer_log.h
@@ -55,13 +55,30 @@
 #else /* Linux distro */
 #include <nntrainer_logger.h>
 
-#define ml_logi(...) __nntrainer_log_print(NNTRAINER_LOG_INFO, __VA_ARGS__)
+#if !defined(ml_logi)
+#define ml_logi(format, args...)                                            \
+  __nntrainer_log_print(NNTRAINER_LOG_INFO, "(%s:%s:%d) " format, __FILE__, \
+                        __func__, __LINE__, ##args)
+#endif
 
-#define ml_logw(...) __nntrainer_log_print(NNTRAINER_LOG_WARN, __VA_ARGS__)
+#if !defined(ml_logw)
+#define ml_logw(format, args...)                                            \
+  __nntrainer_log_print(NNTRAINER_LOG_WARN, "(%s:%s:%d) " format, __FILE__, \
+                        __func__, __LINE__, ##args)
+#endif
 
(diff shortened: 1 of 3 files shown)


nntrainer/include/nntrainer_log.h
```diff
@@ -55,13 +55,30 @@
 #else /* Linux distro */
 #include <nntrainer_logger.h>
 
-#define ml_logi(...) __nntrainer_log_print(NNTRAINER_LOG_INFO, __VA_ARGS__)
+#if !defined(ml_logi)
+#define ml_logi(format, args...)                                         \
+  __nntrainer_log_print(NNTRAINER_LOG_INFO, "(%s:%d) " format, __FILE__, \
+                        __LINE__, ##args)
```
kparichay, line 61: can you also add __FUNC__??

nntrainer/src/nntrainer_logger.cpp
```diff
@@ -88,26 +88,26 @@ void Logger::log(const std::string &message,
   std::stringstream ss;
   switch (loglevel) {
   case NNTRAINER_LOG_INFO:
-    ss << "[NNTRAINER INFO  ";
```
kparichay, line 91: when nntrainer is used with different libraries like nnstreamer etc, this might help differentiate nntrainer logs with nnstreamer logs or in tizen etc.

zhoonit: Yes. How about NNTR ? it is just to long :(

nntrainer/include/nntrainer_log.h
zhoonit: Will do. Thanks

nntrainer/src/nntrainer_logger.cpp
kparichay: Yes, but for the chance that it will be used for other distros, lets keep it uniform.
NNTR or TRAIN or TRAINER etc.

jijoongmoon: I think NNTRAINER is ok.

nntrainer/include/nntrainer_log.h
zhoonit: done

nntrainer/src/nntrainer_logger.cpp
zhoonit: Add tag done.

Issue #395, opened 2020-07-31
write iniparser wrapper


zhoonit:
We need to make an parser wrapper.

from the ini parser:
  1. it does not guarantee given value for the key is able to parse or not. 
  2. some of the parse logic has undefined behavior. (like using `atof` in `iniparser_getdouble`)

zhoonit:
Handled with #404 now with different strategy

Issue #392, opened 2020-07-30
Activation Layer is not parsed correctly


zhoonit:
type activation is not parsed correctly

Issue #391, opened 2020-07-30
load network conf ignores invalid value


zhoonit:

zhoonit:
iniparser_getdouble only checks if there is a key but does not check if given value is okay to parse which cause undefined behavior when user passes wrong value to iniparser_getdouble. This need some amount of refactoring

zhoonit:
This will be done with #400

zhoonit:
#1549 resolves the problem

Issue #269, opened 2020-07-07
Layer properties as map


kparichay:
Update properties internally to be set with std::map<std::string, std::any> with upgrade to c++17.

zhoonit:
This will be handled with #404 different strategy that ensures compile time type

PR #400, opened 2020-08-03
Add property handler

zhoonit:
Add property handler. Current validation of `setProperty` is not done
centrally, hiding bugs when invalid props are given as well as resulting
in many duplicated codes.

This property handler take care of save, load, setting default,
validation of props. So there is no need to take care of properties.
plus, this would enable automated model file save.

See also: #269, #395, #392, #391

**Self evaluation:**

### DIFF
--- a/nntrainer/include/properties.h
+++ b/nntrainer/include/properties.h
@@ -33,18 +33,22 @@ template <typename T> struct prop_traits {
    * @brief validation logic for type T
    * @param[in] value Target to have registration
    */
-  static bool is_valid(value_type value) {  throw std::runtime_error("no trait type can be validated"); }
+  static bool is_valid(value_type value) {
+    throw exception::not_supported("no trait type can be validated");
+  }
 
   /**
    * @brief get string key of current property
    */
-  static std::string getKey() {  throw std::runtime_error("no trait type have a key"); }
+  static std::string getKey() {
+    throw exception::not_supported("no trait type have a key");
+  }
 
   /**
    * @brief get default value of current type
    */
   static value_type getDefault() {
-    throw std::runtime_error("no trait type does not have default");
+    throw exception::not_supported("no trait type does not have default");
   }
 
   /**
@@ -52,17 +56,17 @@ template <typename T> struct prop_traits {
    * @param[in] target Target to be converted to string
    */
   static std::string serialize(value_type target) {
-    throw std::runtime_error("Only property type can be serialized");
+    throw exception::not_supported("Only property type can be serialized");
   }
 
   /**
    * @brief deserialization for type T
    * @param[in] str string to be converted to actual type T
-   * @throw std::runtime_error if T is not property
+   * @throw exception::not_supported if T is not property
    * @throw std::invalid_argument if @a str is not deserializable.
    */
   static value_type deserialize(std::string &str) {
-    throw std::runtime_error("Only property type can be deserialized");
+    throw exception::not_supported("Only property type can be deserialized");
   }
 };
 
@@ -163,26 +167,38 @@ public:
   typedef T prop_type;
 
   /**
+   * @brief Construct a new propertyholder object
+   *
+   */
+  _PropertyHolder() : data(prop_traits<T>::getDefault()) {}
+
+  /**
    * @brief property holder destructor
    */
   ~_PropertyHolder() {}
 
-  _PropertyHolder() {}
-
+  /**
+   * @brief get the actual value it is holding.
+   *
+   * @return value_type data
+   */
   value_type get() { return data; }
 
-  void set(value_type _data) {
-    if (prop_traits<T>::is_valid(_data) == false) {
+  /**
+   * @brief set values according to value type
+   *
+   * @param data data to set
+   * @throw if prop_type is not defined, it throws exception::not_supported
+   * @throw if @a data is not valid, it throws exception::not_supported
+   */
+  void set(value_type data) {
+    if (prop_traits<T>::is_valid(data) == false) {
       std::stringstream ss;
-      ss << "current value is not valid: " << prop_traits<T>::serialize(_data);
-
+      ss << "current value is not valid: " << prop_traits<T>::serialize(data)
+         << "type" << prop_traits<T>::getKey();
       throw std::invalid_argument(ss.str().c_str());
     }
-    force_set(_data);
-  }
-
-  void force_set(value_type _data) {
-    data = _data;
+    this->data = data;
   }
 
   /**
@@ -218,9 +234,10 @@ template <typename... _PropTypes> struct Properties {
   /**
    * @brief Property getter that uses type directly. It is recommened to use
    * this instead of string verion of getter
-   * @code LayerType foo = std::get<LayerType>();
+   * @code LayerType foo = props.template get<LayerType>();
    */
-  template <typename _PropType> _PropType get() {
+  template <typename _PropType>
+  typename prop_traits<_PropType>::value_type get() {
     return std::get<_PropertyHolder<_PropType>>(items).get();
   }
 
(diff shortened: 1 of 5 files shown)


zhoonit:
This PR will be open after making some tests

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+   * @brief property holder destructor
+   */
+  ~_PropertyHolder() {}
+
+  _PropertyHolder() {}
+
+  value_type get() { return data; }
+
+  void set(value_type _data) {
+    if (prop_traits<T>::is_valid(_data) == false) {
```
kparichay, line 179: how about update the return value of `is_valid` to be true when valid or rename to `is_invalid`?

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+      std::stringstream ss;
+      ss << "current value is not valid: " << prop_traits<T>::serialize(_data);
+      throw std::invalid_argument(ss.str().c_str());
+    }
+    force_set(_data);
+  }
+
+  void force_set(value_type _data) { data = _data; }
```
kparichay, line 187: do we need to expose this?

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+  static_assert(prop_traits<T>::is_property, "Only property type can be used");
+
+public:
+  typedef typename prop_traits<T>::value_type value_type;
+  typedef T prop_type;
+
+  /**
+   * @brief property holder destructor
+   */
+  ~_PropertyHolder() {}
```
kparichay, line 172: lets set to default value here?

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+   * @retval string reprenstation of current value
+   */
+  virtual std::string getProp() = 0;
+};
+
+/**
+ * @brief actual holder of property. This class holds pointer of actual type.
+ * Which will be residing on Properties::items;
+ */
+template <typename T> class _PropertyHolder : public PropertyHolder {
```
kparichay, line 162: Just a minor thing but IMO, this class should be `PropertyHolder` and the parent class should be `_PropertyHolder`.

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+  typedef T prop_type;
+
+  /**
+   * @brief property holder destructor
+   */
+  ~_PropertyHolder() {}
+
+  _PropertyHolder() {}
+
+  value_type get() { return data; }
```
kparichay, line 176: add doxygen entry here and at other places.

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+  typedef Color value_type;
+
+  static std::string getKey() { return "color"; }
+
+  static bool is_valid(Color c) { return c != Color::unknown; }
+
+  static Color getDefault() { return Color::unknown; }
+
+  static std::string serialize(Color c) {
+    switch (c) {
```
kparichay, line 107: lets append append `key` in return of serialize and correspondingly parse it back in de-serialize

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+  ~_PropertyHolder() {}
+
+  _PropertyHolder() {}
+
+  value_type get() { return data; }
+
+  void set(value_type _data) {
+    if (prop_traits<T>::is_valid(_data) == false) {
+      std::stringstream ss;
+      ss << "current value is not valid: " << prop_traits<T>::serialize(_data);
```
kparichay, line 181: add key value here as well.

nntrainer/include/properties.h
```diff
@@ -0,0 +1,351 @@
...
+public:
+  typedef typename prop_traits<T>::value_type value_type;
+  typedef T prop_type;
+
+  /**
+   * @brief property holder destructor
+   */
+  ~_PropertyHolder() {}
+
+  _PropertyHolder() {}
```
kparichay, line 174: lets set to default value here?

nntrainer/include/properties.h
zhoonit: `is_valid` returns true already. only undefined types return `std::runtime_error`

nntrainer/include/properties.h
zhoonit: `force_se`t is needed to initialize data to a value that is invalid. I'll add doxygen for the function

nntrainer/include/properties.h
zhoonit: Could you elaborate on this? The parent class is the exposed interface and the child classes are implementations.

nntrainer/include/properties.h
zhoonit: This function is deleted.

nntrainer/include/properties.h
zhoonit: Done

nntrainer/include/properties.h
zhoonit: Done

nntrainer/include/properties.h
zhoonit: Done

nntrainer/include/properties.h
```diff
@@ -0,0 +1,363 @@
...
+ *  You only have to add property struct and prop_traits.                    *
+ *                                                                           *
+ *  It is recommened to use unique key for each types.                       *
+ *  However, it is allowed to use duplicated key                             *
+ *  When those properties are mutually exclusive                             *
+ *****************************************************************************/
+
+/**< color is here for demonstration purpose. */
+
+enum class Color {
```
jijoongmoon, line 85: I don't think we need to demonstrate in header. comment out this codes or have to add this in test codes.

nntrainer/include/properties.h
```diff
@@ -0,0 +1,363 @@
...
+   * @retval string reprenstation of current value
+   */
+  virtual std::string getProp() = 0;
+};
+
+/**
+ * @brief actual holder of property. This class holds pointer of actual type.
+ * Which will be residing on Properties::items;
+ */
+template <typename T> class _PropertyHolder : public PropertyHolder {
```
jijoongmoon, line 162: please not to use a name starting with ```_```

test/unittest/unittest_properties.cpp
```diff
@@ -0,0 +1,392 @@
...
+    case value_type::UNKNOWN:
+      return "unknown";
+    default:
+      /// should not reach here
+      throw std::invalid_argument("c is not serializable");
+    }
+  }
+
+  static value_type deserialize(const std::string &val) {
+    if (val == "white")
```
jijoongmoon, line 133: better to use switch and string::compare

test/unittest/unittest_properties.cpp
```diff
@@ -0,0 +1,392 @@
...
+      return "tropical";
+    case value_type::UNKNOWN:
+      return "unknown";
+    default:
+      throw std::invalid_argument("given parameter is not serializable");
+    }
+  }
+
+  static value_type deserialize(const std::string &val) {
+    if (val == "pome")
```
jijoongmoon, line 83: better to use switch and string compare?

nntrainer/include/properties.h
```diff
@@ -0,0 +1,363 @@
...
+#include <sstream>
+#include <string>
+#include <tuple>
+#include <vector>
+
+#include <nntrainer_error.h>
+
+namespace nntrainer {
+template <typename T> struct prop_traits {
+  static constexpr bool is_property = false; /**< check if this is property */
```
jijoongmoon, line 28: struct? class? What made you to use struct? Looks like it is ok to use class. Cause struct, class mixed and make codes more complicated

jijoongmoon:
Correct me, if I'm wrong. It looks like it is to get the properties and has feature to set the properties for developers and users.
First of all, we have to decide that we are give this feature to the users. If we are not, then it looks like overspec I guess.
And the codes looks quite complicate.
From the top to bottom, is it like
properties --> _PropertyHolder --> PropertyHolder --> proptypes --> prop_traits ?

zhoonit:
Closing this as it is postponed to the next year

nntrainer/include/properties.h
kparichay: my bad. I thought _PropertyHolder was going to be exposed.

PR #401, opened 2020-08-03
[Layer] Update forward and backward layer functions

### LINKED ISSUE #398: Forwarding and backwarding need more arguments
Forwarding and backwarding takes only 1 tensor as input and output.
However to support skip connection, multiple inputs and outputs are to be supported.

kparichay:
Update the forwarding and backwarding layer functions to support
multiple inputs and outputs.
This is crucial to support concat, element wise add, etc kind of layers
as well as skip connection.
Further, forward no longer returns status but rather throws.

Resolves #398

**Self evaluation:**

### DIFF
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -134,27 +134,27 @@ public:
   int init();
 
   /**
-   * @brief     forward propagation
-   * @param[in] input Input Tensor X
-   * @retval    Output Tensor Y
+   * @brief     Forward Propagation of the neural network
+   * @param[in] input List of Input Tensors taken by the neural network
+   * @retval    List of Output Tensors
    */
-  Tensor forwarding(Tensor input, int &status);
+  sharedTensor forwarding(sharedTensor input);
 
   /**
-   * @brief     forward propagation
-   * @param[in] input Input Tensor X
-   * @param[in] label Input Tensor Y2
-   * @retval    Output Tensor Y
+   * @brief     Forward Propagation of the neural network
+   * @param[in] input List of Input Tensors taken by the neural network
+   * @param[in] label List of Label Tensors for the model
+   * @retval    List of Output Tensors
    */
-  Tensor forwarding(Tensor input, Tensor output, int &status);
+  sharedTensor forwarding(sharedTensor input, sharedTensor label);
 
   /**
-   * @brief     back propagation to update W & B
-   * @param[in] input Input Tensor X
-   * @param[in] expectedOutput Lable Tensor Y
-   * @param[in] iteration Epoch Number for ADAM
+   * @brief     Forward Propagation of the neural network
+   * @param[in] input List of Input Tensors taken by the neural network
+   * @param[in] label List of Label Tensors for the model
+   * @param[in] iteration Iteration Number for the optimizer
    */
-  int backwarding(Tensor input, Tensor expected_output, int iteration);
+  void backwarding(sharedTensor input, sharedTensor label, int iteration);
 
   /**
    * @brief     save model and training parameters into file
--- a/nntrainer/src/activation_layer.cpp
+++ b/nntrainer/src/activation_layer.cpp
@@ -51,20 +51,23 @@ int ActivationLayer::initialize(bool last) {
   return ML_ERROR_NONE;
 }
 
-Tensor ActivationLayer::forwarding(Tensor in, int &status) {
-  status = ML_ERROR_NONE;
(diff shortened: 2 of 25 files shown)


nntrainer/src/flatten_layer.cpp
```diff
@@ -37,24 +37,21 @@ int FlattenLayer::initialize(bool last) {
   return status;
 }
 
-Tensor FlattenLayer::forwarding(Tensor in, int &status) {
-  hidden = Tensor(in.batch(), output_dim.channel(), output_dim.height(),
-                  output_dim.width());
+sharedTensor FlattenLayer::forwarding(sharedTensor in) {
+  input = in.get()[0];
+
+  hidden = Tensor(output_dim);
```
jijoongmoon, line 43: I think this should be as it is. This can be differ according to train and inference. Cause currently we are config base data with training. ( Of course this is one of the issues we have to resolve)

jijoongmoon:
LGTM! Great work. This will make nntrainer faster I think. But I think we have to check with Application test examples. If there is no issues then it is good to go.

nntrainer/src/flatten_layer.cpp
kparichay: Fixed for both flatten and pooling layers.

nntrainer/include/neuralnet.h
```diff
@@ -134,27 +134,27 @@ class NeuralNetwork {
   int init();
 
   /**
-   * @brief     forward propagation
-   * @param[in] input Input Tensor X
-   * @retval    Output Tensor Y
+   * @brief     Forward Propation of the neural network
```
zhoonit, line 137:
```suggestion
   * @brief     Forward Propagation of the neural network
```

nntrainer/include/neuralnet.h
```diff
@@ -134,27 +134,27 @@ class NeuralNetwork {
...
    */
-  Tensor forwarding(Tensor input, int &status);
+  sharedTensor forwarding(sharedTensor input);
 
   /**
-   * @brief     forward propagation
-   * @param[in] input Input Tensor X
-   * @param[in] label Input Tensor Y2
-   * @retval    Output Tensor Y
+   * @brief     Forward Propation of the neural network
```
zhoonit, line 144: ditto

nntrainer/src/activation_layer.cpp
```diff
@@ -51,20 +51,23 @@ int ActivationLayer::initialize(bool last) {
...
-  status = ML_ERROR_NONE;
+sharedTensor ActivationLayer::forwarding(sharedTensor in) {
+  input = in.get()[0];
+  hidden = _act_fn(input);
 
-  input = in;
-  hidden = _act_fn(in);
-
-  return hidden;
+  return MAKE_SHARED_TENSOR(hidden);
```
zhoonit, line 58:
```suggestion
  return MAKE_SHARED_TENSOR(std::move(hidden));
```

nntrainer/src/conv2d_layer.cpp
```diff
@@ -79,24 +79,25 @@ void Conv2DLayer::read(std::ifstream &file) { Layer::read(file); }
 
 void Conv2DLayer::save(std::ofstream &file) { Layer::save(file); }
 
-Tensor Conv2DLayer::forwarding(Tensor in, int &status) {
+sharedTensor Conv2DLayer::forwarding(sharedTensor in) {
+  int status = ML_ERROR_NONE;
+  input = in.get()[0];
```
zhoonit, line 84: Minor though, I think in this context should be star. (and for other occurrences of get()[0])
```suggestion
  input = *in;
```

nntrainer/include/neuralnet.h
kparichay: done

nntrainer/include/neuralnet.h
kparichay: done

nntrainer/src/conv2d_layer.cpp
kparichay: yes, `*in` is more suitable for layers who only deal with a single input than `in.get()[0]`.
Done.

nntrainer/src/activation_layer.cpp
kparichay: Can we do this? hidden is the class object variable and not local.

zhoonit: Oh I missed that. nvm

PR #393, opened 2020-07-30
Add more test cases

zhoonit:
Add test cases. some cases are not working so it is disabled.

After fixing bugs in #388 it tc can be opend again accordingly.

**Changes proposed in this PR:**
- Fix some typos
- Fix bugs that activation is not working
- Add some debug logs
- Fix activation layer is not parsed correctly

See also #388
resolved #392
---final pr that resolves #374 

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

PR #402, opened 2020-08-03
[Layer] Add support for element-wise add layer

kparichay:
Add support for element-wise add layer along with basic unittests
This will form the base layer for supporting skip connections
    
**Self evaluation:**
    

nntrainer/src/elem_add_layer.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <elem_add_layer.h>
+#include <layer.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+int ElemAddLayer::initialize(bool last) {
```
zhoonit, line 24: Add validation for `num_inputs`?

nntrainer/src/elem_add_layer.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+sharedTensor ElemAddLayer::forwarding(sharedTensor in) {
+  input = in;
+  hidden = Tensor(in->batch(), output_dim.channel(), output_dim.height(),
+                  output_dim.width());
+  hidden.setZero();
+
+  for (unsigned int idx = 0; idx < num_inputs; ++idx) {
+    if (input_dim != in.get()[idx].getDim())
```
zhoonit, line 45: I think we should either pass the input size or sharedTensor holds size sooner or later.

zhoonit: Or `Layer::forwarding(sharedTensor in, int size = 1)` might work

nntrainer/src/elem_add_layer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    hidden.add_i(in.get()[idx]);
+  }
+
+  return MAKE_SHARED_TENSOR(hidden);
+}
+
+sharedTensor ElemAddLayer::backwarding(sharedTensor derivative, int iteration) {
+  sharedTensor ret = std::shared_ptr<Tensor>(new Tensor[num_inputs],
+                                             std::default_delete<Tensor[]>());
+  Tensor deriv = *derivative.get();
```
zhoonit, line 57: IMO `get()` is not needed. 
```suggestion
  Tensor deriv = *derivative;
```

nntrainer/include/layer.h
```diff
@@ -91,6 +92,7 @@ typedef enum {
   LAYER_FLATTEN,
   LAYER_LOSS,
   LAYER_ACTIVATION,
+  LAYER_ELEM_ADD,
```
zhoonit, line 95: For parsing mechanism to work correctly, the order should be:

```
  LAYER_ACTIVATION,
  LAYER_ELEM_ADD,
  LAYER_LOSS
```

I made a patch for activation layer in #393

nntrainer/src/elem_add_layer.cpp
kparichay: can you elaborate on what kind of validation.
`num_inputs` is Initialized with valid value and setProperty() ensures that the value remains valid.

nntrainer/src/elem_add_layer.cpp
kparichay: Done.

nntrainer/include/layer.h
kparichay: Done.

nntrainer/src/elem_add_layer.cpp
kparichay: Then maybe have that info inside the sharedTensor itself?

zhoonit: > Then maybe have that info inside the sharedTensor itself?

I like this idea better.

nntrainer/src/elem_add_layer.cpp
zhoonit: if that's the case I think it is not needed.

nntrainer/src/elem_add_layer.cpp
kparichay: > Or `Layer::forwarding(sharedTensor in, int size = 1)` might work

This won't work as the caller (NeuralNetwork) has no way to know the size while calling this function.

> Then maybe have that info inside the sharedTensor itself?

Putting size inside sharedTensor means that it should be able to represent a simple std::shared_ptr<Tensor> and std::shared_ptr<Tensor[]> with the same interface.
I will do this in a separate PR.

nntrainer/src/elem_add_layer.cpp
```diff
@@ -0,0 +1,98 @@
...
+  return MAKE_SHARED_TENSOR(hidden);
+}
+
+sharedTensor ElemAddLayer::backwarding(sharedTensor derivative, int iteration) {
+  sharedTensor ret = std::shared_ptr<Tensor>(new Tensor[num_inputs],
+                                             std::default_delete<Tensor[]>());
+
+  for (unsigned int idx = 0; idx < num_inputs; ++idx) {
+    Tensor &t = ret.get()[idx];
+    t = derivative->divide(num_inputs);
```
kparichay, line 59: remove divide

nntrainer/include/elem_add_layer.h
```diff
@@ -0,0 +1,111 @@
...
+#include <tensor.h>
+#include <vector>
+
+namespace nntrainer {
+
+/**
+ * @class   ElemAdd Layer
+ * @brief   ElemAdd Layer
+ */
+class ElemAddLayer : public Layer {
```
kparichay, line 30: Update name to AdditionLayer

kparichay:
V2:
As requested by @jijoongmoon, updated elem-add layer to addition layer.

jijoongmoon (approved):
LGTM

nntrainer/src/addition_layer.cpp
```diff
@@ -0,0 +1,94 @@
...
+                                             int iteration) {
+  sharedTensor ret = std::shared_ptr<Tensor>(new Tensor[num_inputs],
+                                             std::default_delete<Tensor[]>());
+
+  for (unsigned int idx = 0; idx < num_inputs; ++idx) {
+    Tensor &t = ret.get()[idx];
+    t = *derivative;
+  }
+
+  return ret;
```
zhoonit, line 62: can we return `drivative` instead of making `ret`?

nntrainer/src/addition_layer.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <layer.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+int AdditionLayer::initialize(bool last) {
+  int status = ML_ERROR_NONE;
```
zhoonit, line 24: Fail when num_input is not set?

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -1424,6 +1425,71 @@ TEST(nntrainer_ActivationLayer, forward_backward_01_p) {
...
+  sharedTensor input = std::shared_ptr<nntrainer::Tensor>(
+    new nntrainer::Tensor[1], std::default_delete<nntrainer::Tensor[]>());
+  nntrainer::Tensor &in = input.get()[0];
+
+  in = nntrainer::Tensor(layer.getInputDimension());
+
+  EXPECT_THROW(layer.forwarding(input), std::runtime_error);
+}
+
+TEST_F(nntrainer_AdditionLayer, forwarding_03) {
```
zhoonit, line 1480: missing `_p`

zhoonit (changes requested):
Please review comment above. Thanks :)

nntrainer/src/addition_layer.cpp
kparichay: No, we need to return an array of the size of `num_inputs`. Each of these derivatives will be sent back to the corresponding previous layer. One can optimize this to be just 1 derivative and let the caller send each copy. But cannot guarantee the proper working of this layer.

nntrainer/src/addition_layer.cpp
kparichay: As you already know, num_inputs is initialized with value of 1.
So, there is no way to know if it is uninitialized. One might argue that if `num_inputs` is 1, this layer can fail, but rather, it should be optimized out later by the graph.

test/unittest/unittest_nntrainer_layers.cpp
kparichay: added

nntrainer/src/addition_layer.cpp
zhoonit: Right.

nntrainer/src/addition_layer.cpp
zhoonit: How about set default value of num_inputs to be 0? Most of the time, setting sensible default works for other properties. However, for this case, I don't see the use case of num_inputs with default value are used. If there is no use case for default value, num_input must be set before used properly and we should validate accordingly as well.

kparichay: done.

zhoonit (approved):
LGTM 🌝

PR #408, opened 2020-08-05
[TensorDim] Change behaviour of default tensorDim constructor

### LINKED ISSUE #359: Tensor default dimension & allocation
Default tensor construct should have empty buffer. However currently we have 1:1:1:1 tensor and allocate one buffer.

kparichay:
Default tensor dim constructor sets up tensorDim of all 0 which are invalid to use by nature
Setting any dimension resets all the 0 dimensions to 1
This removes allocation of memory by tensor on default constructor yet allows using default constructor of tensor

Resolves #359

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -66,7 +66,18 @@ Tensor::Tensor(const TensorDim &d, float *buf) :
   dim(d),
   strides{{1, 2, 3}},
   is_contiguous(true),
-  data(new float[d.getDataLen()], std::default_delete<float[]>()) {
+  data(d.getDataLen() == 0
+         ? nullptr
+         : std::shared_ptr<float>(new float[d.getDataLen()],
+                                  std::default_delete<float[]>())) {
+  if (d.getDataLen() == 0) {
+    if (buf != nullptr) {
+      throw std::runtime_error(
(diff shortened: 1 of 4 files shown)


zhoonit:
Are we not gonna delete default constructor?

kparichay:
> Are we not gonna delete default constructor?

As discussed offline, lets keep default constructor with zero size.

zhoonit:
Could you put `resolves #359`?

nntrainer/src/tensor.cpp
```diff
@@ -66,7 +66,17 @@ Tensor::Tensor(const TensorDim &d, float *buf) :
   dim(d),
   strides{{1, 2, 3}},
   is_contiguous(true),
-  data(new float[d.getDataLen()], std::default_delete<float[]>()) {
+  data(std::shared_ptr<float>(static_cast<float *>(nullptr))) {
```
zhoonit, line 69:
```suggestion
  data(nullptr) {
```

zhoonit: Or how about this to avoid reinitialization

```suggestion
  data(d.getDataLen() == 0 ? nullptr: std::shared_ptr<float> ...  )) {
```

kparichay: Done.

PR #413, opened 2020-08-06
Add safeguard to prevent loadFromConf twice

### LINKED ISSUE #382: reset neuralnet when it failes to load
neuralnet::loadFromConfig can fail anymoment, but once it failes, layers added remains in the neuralnet which can lead to buggy behavior

zhoonit:
- This patch disables loadFromConfig after loading

v2:
resolves #382 

**Self evaluation:**

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -66,7 +66,8 @@ NeuralNetwork::NeuralNetwork() :
   continue_train(false),
   iter(0),
   initialized(false),
-  def_name_count(0) {}
+  def_name_count(0),
+  loadedFromConfig(false) {}
 
 NeuralNetwork::NeuralNetwork(std::string config) : NeuralNetwork() {
   this->setConfig(config);
@@ -164,6 +165,11 @@ int NeuralNetwork::loadDatasetConfig(void *_ini) {
 }
 
 int NeuralNetwork::loadFromConfig() {
+  if (loadedFromConfig == true) {
+    ml_loge("cannnot do loadFromConfig twice");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
   int status = ML_ERROR_NONE;
   std::string ini_file = config;
   int num_ini_sec = 0;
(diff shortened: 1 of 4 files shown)


nntrainer/src/neuralnet.cpp
```diff
@@ -164,6 +165,13 @@ int NeuralNetwork::loadDatasetConfig(void *_ini) {
 }
 
 int NeuralNetwork::loadFromConfig() {
+  if (loadedFromConfig == true) {
+    ml_loge("cannnot do loadFromConfig twice");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  loadedFromConfig = true;
```
kparichay, line 173: this should be set at the end of function.
this will remain true even if loadFromConfig fails.

zhoonit: Done

zhoonit:
> LGTM
> However, there exists an issue that if loading fails in between, then `layers` is not emptied. Or I should say there must be a `layers.clear()` before loading layers to solve it.

Please notice edited https://github.com/nnstreamer/nntrainer/pull/413#discussion_r466227259

nntrainer/src/neuralnet.cpp
kparichay: This adds more issues to a function which already has more errors.
Keep loadFromConfig to the end. 
Reset can be done separately in a new PR, but setting it at start creates more issues where this value also has to be reset.
In general, such values as configured/loaded etc should be set at the end of the call.

zhoonit: > it at start creates more issues

Can you elaborate on this? Since the function is not atomic, it has to be called only once fail or succeed. If you allow it to call again when fail, it will be running without exposing bugs.

kparichay: If it fails once with some incorrect config, and then called again with correct config, it will fail as loadFromConfig is not reset correctly when failed.

zhoonit: > If it fails once with some incorrect config, and then called again with correct config, it will fail as loadFromConfig is not reset correctly when failed.

If it is, it ought to go bottom. The problem is it isn't as #382 states

zhoonit:
> Need changes based on [#413 (comment)](https://github.com/nnstreamer/nntrainer/pull/413#discussion_r466268138)

Resolved #382 with additional commit. @kparichay  PTAL.

nntrainer/src/neuralnet.cpp
kparichay: Yes. Now its good.

PR #417, opened 2020-08-07
[Delegate] Add delegate support header

kparichay:
Add delegate support header
This supports settings backend and device
Added some properties but they are experimental

These will made by the user if needed, else defaults will be used by the model during compile

Related #659

### DIFF
--- a/nntrainer/include/delegate.h
+++ b/nntrainer/include/delegate.h
@@ -1,0 +1,187 @@
+// SPDX-License-Identifier: Apache-2.0-only
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	delegate.h
+ * @date	7 Aug 2020
+ * @brief	This is Delegate Class for the Neural Network
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ * @note This class is experimental and subject to major modifications.
+ *
+ */
+
+#ifndef __DELEGATE_H__
+#define __DELEGATE_H__
+
+namespace nntrainer {
+
+/**
+ * @class Backend
+ * @brief Backend to be used for the operations to use.
+ * @note If some operation is not supported on set backend, that operation
+ * implemented with default backend will be used.
+ *
+ */
+class Backend {
+public:
+  /**
+   * @brief Backend type enumeration
+   */
+  enum class BackendType {
+    armcl, /**< Arm Compute Library for backend */
+    blas,  /**< Libopenblas for backend */
+    base   /**< Internally implemented operations for backend */
+  };
+
+  /**
+   * @brief Construct a new Backend object
+   *
+   * @param backend Backend of be used. Defaults to base backend
+   * @param num_t Number of threads. Defaults to 1 thread
+   */
+  Backend(BackendType backend = BackendType::base, int num_t = 1) :
+    backend(backend),
+    num_threads(num_t) {}
+
+  /**
+   * @brief Set the number of threads for the backend if supported
+   *
+   * @param num_threads Number of threads
+   */
+  void setNumThreads(unsigned int num_threads);
+
+private:
+  /**
+   * @brief backend type stored
+   */
+  BackendType backend;
+
+  /**
+   * @brief number of threads stored
+   */
+  unsigned int num_threads;
+};
+
+/**
+ * @class Device
+ * @brief Device to be used for the operations to run.
+ * @note The selected delegate device will restrict the supported backend.
+ *
+ */
+class Device {
+public:
+  /**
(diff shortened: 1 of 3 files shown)


nntrainer/include/delegate.h
```diff
@@ -0,0 +1,179 @@
...
+   * @param precision_loss True to allow loss in precision, else false
+   */
+  void allowPrecisionLoss(bool precision_loss) {
+    throw std::logic_error("Not yet implemented");
+    this->precision_loss = precision_loss;
+  }
+
+private:
+  DeviceType device;     /**< Device type for the object */
+  float memory_fraction; /**< Maximum memory fraction which can be used by the
```
zhoonit, line 148: Can this be double?

temp.patch
```diff
@@ -0,0 +1,197 @@
+commit d1267681ea0e3060807c865c7f520ed6b66e30bf
```
zhoonit, line 1: I guess this file is added unintentionally

nntrainer/include/delegate.h
```diff
@@ -0,0 +1,179 @@
...
+    npu  /**< NPU as the device */
+  };
+
+  /**
+   * @brief Construct a new Device object
+   *
+   * @param device Device delegate to be used for the operations to run. Default
+   * device is cpu.
+   */
+  Device(DeviceType device = DeviceType::cpu) :
```
zhoonit, line 88: IMO, it is better to have ctor with full argument with default so that we can instantiate object with full configuration.

```suggestion
  Device(DeviceType device = DeviceType::cpu, float memory_fraction = 1.0, soft_pacement = false, precision_loss = false) :
```

zhoonit (approved):
Overall LGTM

nntrainer/include/delegate.h
kparichay: This can be double.
However, this is still experimental and I don't expect users to specify more than 2 places decimal value (precision of lower than 0.01 will be rare).

temp.patch
kparichay: removed.

nntrainer/include/delegate.h
kparichay: yes, updated.
Also updated for backend class as well.

kparichay:
rebased.

nntrainer/include/delegate.h
```diff
@@ -0,0 +1,187 @@
...
+ */
+class Backend {
+public:
+  /**
+   * @brief Backend type enumeration
+   */
+  enum class BackendType {
+    armcl, /**< Arm Compute Library for backend */
+    blas,  /**< Libopenblas for backend */
+    base   /**< Internally implemented operations for backend */
```
jijoongmoon, line 36: This is related with #491

jijoongmoon:
Before we move forward, we have to study what kind of op these backend can privdes. blas just provides linear algebra calculation not the convolution level and I'm not sure but acl supports different level. after studying this and we could decide which member function could be in backend.

nntrainer/include/delegate.h
kparichay: This header just stores the configuration passed by the user and validates it.
#491 implements one such backend. The structure in this header will allow allow tensors or layers to check the user set configuration and appropriately choose backend at runtime.

kparichay:
> Before we move forward, we have to study what kind of op these backend can privdes. blas just provides linear algebra calculation not the convolution level and I'm not sure but acl supports different level. after studying this and we could decide which member function could be in backend.

As this comment https://github.com/nnstreamer/nntrainer/pull/417#discussion_r481858120 states, this header just stores the configuration.
Yes, for various backends, the interface will be same. Some will below tensor level (blas), at tensor level (like eigen) or at layer level (nnfw, acl). So, #491 blas_interface allows to use lower level libraries with common interface. 
With delegate struct, tensor or layer can appropriately delegate their task to appropriate header implementation or backend.

jijoongmoon (approved):
LGTM

nntrainer/include/delegate.h
zhoonit: I think backend type (and arguably device type... if there are two heterogenuous gpu???) should be string based as it can be added or removed from time to time.
I think we can keep this enum for now since we don't have the managing system(global context), so leaving this up to you for now :)

kparichay: custom layers are going to be really common to be done, and using string for them is really convenient.
For backends and device, string works but I will keep enum for now, and then change to string when we understand more about what various types of hardware are possible 
- different versions of devices are possible which would require a version variable (will be added in next PR)
- different types of the same group of devices are possible (npu1, npu2) - then it might be better to add enum in this case 
This all depends on how we optimize graphs with delegates and what limitations we get with layers for different devices. Till then, I will leave it as it is.

zhoonit: Sounds good

zhoonit (approved):
LGTM

PR #418, opened 2020-08-07
[model] Change the semantics for setting the batchsize

### LINKED ISSUE #299: Overlap with minibatch and inputshape
Input shape takes batch dimension as input. However, this can be in conflict with `minibatch` argument in ini file format.

kparichay:
Batchsize as a property was earlier set in input_shape for a layer
as well as for the model. However, this was conflicting and prone to errors.

This patch disables setting batch size while setting the input_shape.
If batch_size is more than 1 in the input_shape, a warning is issued.
And this value is later overriden by the batchsize set for the model.
This same batchsize is also used for the dataset as well.

Resolves #299

**Self evaluation:**


nntrainer/src/neuralnet.cpp
```diff
@@ -536,6 +529,9 @@ sharedTensor NeuralNetwork::forwarding(sharedTensor input) {
 sharedTensor NeuralNetwork::forwarding(sharedTensor input, sharedTensor label) {
   sharedTensor X;
 
+  if (input->getDim().batch() > batch_size)
```
zhoonit, line 532: Q. is it valid for `input->getDim().batch() < batch_size` ?

nntrainer/src/neuralnet.cpp
```diff
@@ -536,6 +529,9 @@ sharedTensor NeuralNetwork::forwarding(sharedTensor input) {
...
+    throw std::logic_error("Error: mismatch in batchsize for data and model.");
```
zhoonit, line 533: Could you add `std::logic_error` to `nntrainer_exception_boundary`?

kparichay: done

nntrainer/src/parse_util.cpp
```diff
@@ -380,10 +384,10 @@ unsigned int parseDataProperty(std::string property) {
   return (unsigned int)DataBuffer::PropertyType::unknown;
 }
 
-int setInt(int &val, std::string str) {
+int setUint(unsigned int &val, const std::string &str) {
   int status = ML_ERROR_NONE;
   try {
-    val = std::stoi(str.c_str());
+    val = (unsigned int)std::stoi(str.c_str());
```
zhoonit, line 390:
```suggestion
    val = std::stoul(str.c_str());
```

nntrainer/src/tensor_dim.cpp
```diff
@@ -44,11 +44,21 @@ void TensorDim::resetLen() {
...
+    throw std::invalid_argument(
+      "[TensorDim] Tensor Dimension index should be between 0 and 4");
+
+  return dim[idx];
+}
+
 void TensorDim::setTensorDim(unsigned int idx, unsigned int value) {
-  if (value == 0) {
+  if (idx < 0 || idx > MAXDIM)
+    throw std::invalid_argument(
```
zhoonit, line 57: how about `throw std::out_of_range` ? We might need to differentiate those two cases.

nntrainer/src/tensor_dim.cpp
```diff
@@ -61,7 +71,7 @@ int TensorDim::setTensorDim(std::string input_shape) {
     std::sregex_iterator(input_shape.begin(), input_shape.end(), words_regex);
   auto words_end = std::sregex_iterator();
   int cur_dim = std::distance(words_begin, words_end);
-  if (cur_dim > MAXDIM) {
+  if (cur_dim <= 0 || cur_dim > MAXDIM) {
     ml_loge("Tensor Dimension should be less than 4");
```
zhoonit, line 75:
```suggestion
    ml_loge("Tensor Dimension should be between 1 and 4");
```

zhoonit: Slightly out of topic though, I think line 80~83 / 85~86 can be safely deleted.

nntrainer/src/layer.cpp
```diff
@@ -158,8 +158,29 @@ void Layer::setProperty(const PropertyType type, const std::string &value) {
     break;
   case PropertyType::input_shape:
     if (!value.empty()) {
+      unsigned int cache_batch_size = 1;
+      /** cache original value of batch size */
+      if (input_dim.batch()) {
```
kparichay, line 163: @zhoonit Updated.

nntrainer/src/parse_util.cpp
kparichay: done.

nntrainer/src/tensor_dim.cpp
kparichay: done.

nntrainer/src/tensor_dim.cpp
kparichay: Done.
Removed both checks and update `stoi` to `stoul` on line 80.

Issue #388, opened 2020-07-30
Bugs from more negative cases


zhoonit:
## descriptive debug logs
- Add more log debug to get better information

## bugs
- [x] Activation layer itself is making activation again (#390) (#429)
- [x] reset neuralnet when it failes to load (#382) (#413)
- [x] init_loss_layer should fail when cost is unknown (with different log)
- [x] buf size is determined after minibatch size is set so buf size can be smaller than minibatch (#389)
~ - [ ] loadNetworkConf ignores invalid values from ini (#391) ~ (dropped from the umbrella)

zhoonit:
Closing this as all of the issues have been resolved except for dropped one.

PR #429, opened 2020-08-11
Additional validation and fixes for network::init

### LINKED ISSUE #390: Activation layer type itself is making activation again
because we always make layer even for the activation layer itself, it is messing up activation ini

zhoonit:
- Add term `realization`. This term denotes that enum type is turned
into an instance
- Change `initActivationLayer` -> `realizeActivationType`
- Change `initFlattenLayer` -> `relizeFlattenType`
- Delete `_make_act_layer`
- Add `TensorDim::isEmpty()`
- Clarify `checkValidation` -> `isInitializable` and move beforehand
validation logics
- Fix logic that prevents double activation
- Move layer name validaton phase `addLayer` -> `isInitializable`
- Prevent realization of `ACT_NONE`
- Sort unittest_modelfile order and update comments
- Enable few test cases

Resolves #390
See also #388

**Self evaluation:**

### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -216,7 +216,7 @@ int ml_train_model_compile(ml_train_model_h model, ...) {
   if (status != ML_ERROR_NONE)
     return status;
 
-  f = [&]() { return NN->checkValidation(); };
+  f = [&]() { return NN->isInitializable(); };
   status = nntrainer_exception_boundary(f);
 
   return status;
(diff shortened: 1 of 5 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -216,7 +216,7 @@ int ml_train_model_compile(ml_train_model_h model, ...) {
   if (status != ML_ERROR_NONE)
     return status;
 
-  f = [&]() { return NN->checkValidation(); };
+  f = [&]() { return NN->isInitializable(); };
```
jijoongmoon, line 219: I think this is for just checking the configuration before Initialize.

zhoonit: Yes. It checks if it is able to initialize. So I clarified the name

PR #435, opened 2020-08-13
[ Bug ] Fix calculating training loss

jijoongmoon:
In this PR, calculation of training loss is fixed.
Previously, network loss is calculated using last mini_batch size data set.
This should be summarized through out whole data set and averaged.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -580,7 +580,6 @@ float NeuralNetwork::getLoss() {
   for (unsigned int i = 0; i < layers.size(); i++) {
     loss += layers[i]->getLoss();
   }
-
   return loss;
 }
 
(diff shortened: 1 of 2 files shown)


myungjoo:
Wow.. you finally got that bug. Great!

myungjoo:
It'd be better if you'd describe what was the problem with the previous code, though.

nntrainer/src/neuralnet.cpp
```diff
@@ -701,17 +693,16 @@ int NeuralNetwork::train_run() {
         }
         std::cout << "#" << i + 1 << "/" << epoch;
         data_buffer->displayProgress(count++, nntrainer::BUF_TRAIN, getLoss());
+        training_loss += getLoss();
```
kparichay, line 696: This is good catch.

nntrainer/src/neuralnet.cpp
```diff
@@ -575,14 +575,7 @@ void NeuralNetwork::backwarding(sharedConstTensor input,
...
 
-float NeuralNetwork::getLoss() {
-  loss = 0.0f;
-  for (unsigned int i = 0; i < layers.size(); i++) {
-    loss += layers[i]->getLoss();
-  }
-
-  return loss;
-}
+float NeuralNetwork::getLoss() { return layers.back()->getLoss(); }
```
kparichay, line 578: This will ignore the loss by the weight decay loss. Is this intentional?

jijoongmoon: fixed.

PR #441, opened 2020-08-18
[TensorOp] Add Tensor Op class

kparichay:
Add tensor op class
Also update FC layer as per tensor op class
Also add graph class for tensor op

**Self evaluation:**

### DIFF
--- a/nntrainer/include/graph.h
+++ b/nntrainer/include/graph.h
@@ -1,0 +1,166 @@
+// SPDX-License-Identifier: Apache-2.0-only
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	graph.h
+ * @date	14 August 2020
+ * @brief	This is Graph Class of Neural Network
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ */
+
+#ifndef __GRAPH_H__
+#define __GRAPH_H__
+#ifdef __cplusplus
+
+#include <tensor_op.h>
+#include <vector>
+
+namespace nntrainer {
+
+/**
+ * @class   Graph
+ * @brief   Graph of Tensors and Tensor Operations
+ */
+class Graph {
+public:
+  Graph() {}
+
+  ~Graph() {}
+
+  /**
+   * @brief Register the tensor op in the graph along with its input/output
+   * tensors
+   *
+   * @param op Operation to be registered
+   */
+  void registerTensorOp(TensorOp op) {}
+
+  /**
+   * @brief Get the all the tensors object
+   *
+   * @return std::vector<Tensor>& List of all the tensor objects for the all
+   * registered ops
+   */
+  std::vector<Tensor> &getTensors() { return all_tensors; }
+
+  /**
+   * @brief Get the all the input tensors object
+   *
+   * @return std::vector<Tensor>& List of all the input tensor objects for the
+   * graphs
+   * @details These are the tensors to be updated every iteration when training
+   * the model
+   */
+  std::vector<Tensor> getInputs() { return {}; }
+
+  /**
+   * @brief Get the all the output tensors object
+   *
+   * @return std::vector<Tensor>& List of all the output tensor objects for the
+   * graphs
+   * @details These are the tensors where the output of the model for every
+   * iteration will be available when training the model
+   */
+  std::vector<Tensor> getOutputs() { return {}; }
+
+  /**
+   * @brief Add this tensor to the list of output tensors
+   *
+   * @param tensor Tensor to be added as output tensor
+   */
+  void addToOutputs(Tensor tensor) {}
+
+  /**
+   * @brief Set the optimizer to be used for this graph optimization
+   *
+   * @param opt Optimizer to be set
+   */
+  void setOptimizer(Optimizer opt) {}
+
+  /**
+   * @brief Run the given tensorOp with the current inputs and its dependencies
+   *
+   * @param op Operation to be run
+   */
+  void compute(TensorOp op) {}
+
+  /**
+   * @brief Run the given tensor with the current inputs
+   *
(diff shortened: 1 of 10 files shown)


nntrainer/include/graph.h
```diff
@@ -0,0 +1,166 @@
...
+   */
+  void registerTensorOp(TensorOp op) {}
+
+  /**
+   * @brief Get the all the tensors object
+   *
+   * @return std::vector<Tensor>& List of all the tensor objects for the all
+   * registered ops
+   */
+  std::vector<Tensor> &getTensors() { return all_tensors; }
```
zhoonit, line 47: Q. So, is `std::vector<Tensor>` :cactus: gonna be changed to a map to save inputs and outputs?
I am confused if :cactus: gonna be a storage(for inputs/outputs) or graph or both?

kparichay: In a graph, we have a list of vectors and a list of edges connecting them.
Here, std::vector<Tensor> is the information on each edge.

nntrainer/src/fc_layer_op.cpp
```diff
@@ -0,0 +1,113 @@
...
+  dim.batch(1);
+  Tensor weight = initializeWeight(dim, weight_ini_type, status);
+  NN_RETURN_STATUS();
+
+  setParamSize(2);
+  paramsAt(0) = {std::move(weight), Tensor(weight.getDim()), "FC:weight"};
+  paramsAt(1) = {std::move(bias), Tensor(bias.getDim()), "FC:bias"};
+
+  TensorParam w =
+    TensorParam(paramsAt(0).weight, Tensor(weight.getDim()), "FC:weight", true);
```
jijoongmoon, line 63: may be ```TensorParam(paramsAt(0).weight, paramsAt(0).grad, "FC:weight"}``` and TensorParam(paramsAt(1).weight, paramsAt(1).grad,"FC:bias",true)```

nntrainer/src/fc_layer_op.cpp
```diff
@@ -0,0 +1,113 @@
...
+           .run()
+           .sum(0);
+}
+
+void FullyConnectedLayer::applyGrad(int iteration) {
+  int idx = 0;
+  for (auto &tp : inputs_op) {
+    if (tp.trainable && trainable) {
+      opt.apply_gradients(
+        std::make_shared<UpdatableParam>(params.get()[idx - 1]), 1, iteration);
```
jijoongmoon, line 109: Do we need updatableParam cause we have tensorparam.

nntrainer/src/neuralnet.cpp
```diff
@@ -824,6 +824,24 @@ int NeuralNetwork::addLayer(std::shared_ptr<Layer> layer) {
...
+  if (type == LAYER_UNKNOWN)
+    return ML_ERROR_INVALID_PARAMETER;
+
+  if (initialized) {
+    return ML_ERROR_NOT_SUPPORTED;
+  }
+
+  ensureName(layer);
+  std::shared_ptr<TensorOp> op = std::dynamic_pointer_cast<TensorOp>(layer);
+  graph.registerTensorOp(*op);
```
jijoongmoon, line 840: I'm not sure if tensorop could be tensor calculation level not a layer level. Cause there are calculation with tensor values like kernel calculation or make proper data type in convolution.

nntrainer/src/fc_layer_op.cpp
kparichay: Yes, they will be merged. This PR is just to demonstrate a sample example.

nntrainer/src/neuralnet.cpp
kparichay: As you said, custom op for tensor is not good.
As you might have noticed, `tensor_op`'s method `compute_op` and other methods do not take any arguments.
The reason is to allow multiple arguments.

When tensorOp does tensor level calculation later, lazy tensor would have to be deprecated as tensor would do all the lazily itself.

Please tell me if I misunderstood your question.

kparichay:
Note: this setup has an issue where layer configuration cannot be passed to the `tensor_op` while computing.
For example - in convolution `forwarding`, various configuration values such as stride, padding etc needs to be accessible to tensor_op so as to perform this function (this problem is only for layers. Tensor calculation does not need such configurations). Two possible solutions - 
1. TensorOp has properties than layers and then it can use it easily
2. These properties be passed to the computation functions (forwarding/backwarding or computeop/computegrad). However, in this case, graph needs to store these configurations for these tensorOp (essentially making it same as above).

Further, with operations such as gemm etc (which don't operate at tensor level), must think about how they will be covered with tensorOp.

nntrainer/src/fc_layer_op.cpp
kparichay: yes, that is better. Will be updated to this style.

zhoonit:
> Note: this setup has an issue where layer configuration cannot be passed to the `tensor_op` while computing.
> For example - in convolution `forwarding`, various configuration values such as stride, padding etc needs to be accessible to tensor_op so as to perform this function (this problem is only for layers. Tensor calculation does not need such configurations). Two possible solutions -
> 
> 1. TensorOp has properties than layers and then it can use it easily
> 2. These properties be passed to the computation functions (forwarding/backwarding or computeop/computegrad). However, in this case, graph needs to store these configurations for these tensorOp (essentially making it same as above).
> 
> Further, with operations such as gemm etc (which don't operate at tensor level), must think about how they will be covered with tensorOp.

I think..
1. Tensor Ops should have inner state by inheriting base class(TensorOps)
2. Gemm should be interfaced by tensor level.

kparichay:
> I think..
> 
> 1. Tensor Ops should have inner state by inheriting base class(TensorOps)
> 2. Gemm should be interfaced by tensor level wrapper operation

For 2. , yes that sounds good. If any bigger operation wants to use gemm, then that operation should becomes a tesnorOp itself.
However for 1, there comes a requirement for a structure to hold these properties of various data types and what not (maybe std::map(std::string, std::any) might satisfy).

kparichay:
Closing as this isn't needed now.

PR #442, opened 2020-08-19
[ Conv2D ] Add Thread to calculate backwarding

jijoongmoon:
Add Thread to accelerate backwarding calculation.
Seperate DelK Calculation.

### DIFF
--- a/nntrainer/src/conv2d_layer.cpp
+++ b/nntrainer/src/conv2d_layer.cpp
@@ -19,6 +19,7 @@
 #include <nntrainer_log.h>
 #include <parse_util.h>
 #include <string>
+#include <thread>
 #include <util_func.h>
 
 namespace nntrainer {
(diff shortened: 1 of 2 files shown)


myungjoo:
Threading for performance enhancement requires validation with experiments in target devices. You need to check if there are excessive overheads from it.

Although I'm not familiar with C++/std/thread (so I'm not sure if they are already addressed with std::thread or not), you need to check:
- The need for thread pooling: if (re)creating threads incur some overheads, you may consider thread pooling (persistent threading?)
- Task distribution (skewed? even?), and critical sections (safety).
- Overheads / inefficiency due to data locking.

jijoongmoon:
Certainly we have to test it on device. But I share the pc result about mnist example ( 1 epoch ).

without thread.
```
#0/1 - Training Loss: 2.30256 >> [ Accuracy: 11.5927% - Validation Loss : 2.29918 ]
real 0m5.183s
user 0m10.069s
sys 0m0.305s
```
with thread
```
#0/1 - Training Loss: 2.30256 >> [ Accuracy: 11.5927% - Validation Loss : 2.29918 ] 
real 0m4.433s
user 0m.9.592s
sys 0m0.160s ```

File: nntrainer/src/conv2d_layer.cpp [hunk 1]
```diff
@@ -176,9 +174,17 @@ sharedConstTensor Conv2DLayer::backwarding(sharedConstTensor derivative,
...
   }
+  return;
+}
 
-  // Calculate delS : returns ( Full pad )
+sharedConstTensor Conv2DLayer::backwarding(sharedConstTensor derivative,
+                                           int iteration) {
+
+  // Calculate delK : [batch, channel, height, width ] * filter_size
+  std::thread delk_thread(&Conv2DLayer::calculateDelK, this, derivative);
```
kparichay, line 184: Lets not create new thread everytime, but create 1 thread at runtime and reuse it existing thread.

kparichay:
Creating thread internally in backwarding for a layer is a CPU specific solution.
These multiple threads might not directly speedup for GPU/other devices.

jijoongmoon:
> Creating thread internally in backwarding for a layer is a CPU specific solution.
> These multiple threads might not directly speedup for GPU/other devices.

I guess these calculations can be done at the same time. They are not totally independent. And I think con2d function itself can be accelerated by the GPU/other devices. not the layer I guess. However reusing thread pool is better I guess. I will leave it at #453. Let's have this issues in mind when we implement delegate.

PR #450, opened 2020-08-19
[neuralnet] Train loop index variable reuse remove

kparichay:
Multiple nested loop with same variable names are being used. Fix that.
Also, reporting for train epoch should start from 1 and not from 0.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -668,7 +668,7 @@ int NeuralNetwork::train(std::vector<std::string> values) {
 int NeuralNetwork::train_run() {
   int status = ML_ERROR_NONE;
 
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM

nntrainer/src/neuralnet.cpp
```diff
@@ -707,7 +707,7 @@ int NeuralNetwork::train_run() {
     }
 
     saveModel();
-    std::cout << "#" << i << "/" << epoch
+    std::cout << "#" << epoch_idx + 1 << "/" << epoch
               << " - Training Loss: " << training_loss / count;
```
jijoongmoon, line 711: It's ok to start with 1. but tensorflow also start with 0. No big deal.

Issue #133, opened 2020-06-08
Add way to do cross check implementations


zhoonit:
Currently there is limited way to check if the current implementation is logically correct. (Aside from it is running and have pretty good result.)

Cross checking the implementation with other frameworks is needed.

kparichay:
Make tensorflow operations for each layer:
1. Forward output
2. Gradient for the weights in layer
3. Updated weight

Networks to use:
1. 1 FC layer with varied activations and loss training - tests FC layer, activations and loss
2. 2 FC layers with training - tests transfer of gradient from back from FC layer

Add more for other layers.

jijoongmoon:
We checked bullet items as below with tensorflow.
. conv2d : done
. fc : done
. pooling2d : done
. flatten : done
. optimizer : sgd done
. loss : mse, cross-sigmoid, cross-softmax : done

kparichay:
@jijoongmoon conv2d backward verification is still to be done as per https://github.com/nnstreamer/nntrainer/blob/master/test/unittest/unittest_nntrainer_layers.cpp#L949

jijoongmoon:
Adam validation is done with #376

zhoonit:
Can this be closed now?

kparichay:
Once MNIST comparison with tensorflow is done, add this to a test which is run in every PR.

jijoongmoon:
We are close this issues. We reopen or make new issue after implementing Transpose convolution.

PR #454, opened 2020-08-20
[mnist] Update tensorflow training for mnist application

kparichay:
Update tensorflow training example for mnist application
This with same initialization as nntrainer (shown with zero initialization)
matches the final accuracy as well as loss with nntrainer

See also #133

**Self evaluation:**

cc. @jijoongmoon

### DIFF
--- a/Applications/mnist/Tensorflow/Training_Keras.py
+++ b/Applications/mnist/Tensorflow/Training_Keras.py
@@ -41,7 +41,7 @@ np.random.seed(SEED)
 batch_size =32
 Learning = True
 Test = False
-num_epoch = 1500
+num_epoch = 50
 DEBUG = True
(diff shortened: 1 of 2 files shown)


Applications/mnist/Tensorflow/Training_Keras.py
```diff
@@ -44,6 +44,7 @@
 Test = False
 num_epoch = 1500
 DEBUG = True
+USE_FIT = False
```
jijoongmoon, line 47: how about USE_FIT = True as an default.

kparichay: done.

jijoongmoon (approved):
LGTM

PR #455, opened 2020-08-20
[tensor] restricted setDim is now reshape

### LINKED ISSUE #412: We should prohibit changing dataLen after it is copy constructed,
this will lead to segfault, if dataLen is stretched. Because inner data won't resize anyway.

if datalen stretches, tensor itself should reassign and copy all the data. (Would that be contiguous? if setTensorDim has diffrent dimension?)

More than welcome to suggest ideas to alleviate this problem.

It can be a future problem.

kparichay:
Tensor setDim() is renamed to reshape() as tensor should only be allowed to reshape
Resizing it invalidates the invariant for the other holders of the same data of tensor (see also #412)

This patch renames setDim() to reshape() which throws if there is an attempt to change size
Further getDim() now returns a copy of the dimension to disallow changing dimension of a tensor directly

Also some aliases in Tensor (like getWidth()) have been removed. Rather, width() like functions are kept
to maintain coherency with TensorDim() interface.

Resolves #412

**Self evaluation:**

### DIFF
--- a/nntrainer/src/activation_layer.cpp
+++ b/nntrainer/src/activation_layer.cpp
@@ -148,16 +148,16 @@ Tensor ActivationLayer::softmax(Tensor const &t) {
    * shiftx_logit = logit - max_batch(logit)
    * softmax = exp(shiftx_logit) / (sum(exp(shiftx_logit)))
    */
-  int batch = t.getBatch();
-  int channel = t.getChannel();
-  int height = t.getHeight();
-  int width = t.getWidth();
+  unsigned int batch = t.batch();
+  unsigned int channel = t.channel();
+  unsigned int height = t.height();
+  unsigned int width = t.width();
   float *dp;
   float *rp;
   const float *tp;
 
   Tensor result(t.getDim());
-  Tensor divisor(t.getWidth());
+  Tensor divisor(t.width());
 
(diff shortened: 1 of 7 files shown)


nntrainer/src/activation_layer.cpp
```diff
@@ -148,44 +148,44 @@ Tensor ActivationLayer::softmax(Tensor const &t) {
    * shiftx_logit = logit - max_batch(logit)
    * softmax = exp(shiftx_logit) / (sum(exp(shiftx_logit)))
    */
-  int batch = t.getBatch();
-  int channel = t.getChannel();
-  int height = t.getHeight();
-  int width = t.getWidth();
+  unsigned int batch = t.batch();
+  unsigned int channel = t.channel();
```
jijoongmoon, line 152: getTensorDim(0)?

jijoongmoon (approved):
LGTM. check format please.

kparichay:
Rebased and updated to clang-format.

zhoonit:
Q. If Tensor is constructed without dimension, can we initialize with dimension later on?

kparichay:
With this commit, no. It can be only reshaped.
I guess that can be supported easily as per need.

PR #461, opened 2020-08-21
[ini] Update Network section to Model

### LINKED ISSUE #318: Change [network] -> [model]
since [network] can actually be KNN or regression that aren't actually network,

how about we change [network] section naming to [model]?

kparichay:
Update section name from Network to Model
Also update internal naming in loadModel and correspondingly unittests

Resolves #318

**Self evaluation:**


kparichay:
@jijoongmoon Should NeuralNetwork class be also updated to model. Also documentation says "The aim of the NNtrainer is to develop Software Framework to train neural network model". How about update `neural network models` to `machine learning models` (because KNN don't come into the category of neural networks and Reinforcement learning are at AI/ML level, not NN).

Issue #457, opened 2020-08-21
Remove KNN Example until we support KNN.


jijoongmoon:
Current implementation of KNN does not use NNTrainer. So I think we need to remove it until we implement KNN (#458).

kparichay:
Lets just rename KNN/jni/main.cpp to main_tflite.cpp and create main.cpp which return NYI rather than removing the code (assuming that KNN is going to be implemented).

PR #465, opened 2020-08-21
[ Application ] change nntrainer_knn to knn_sample

jijoongmoon:
Current implementation of KNN does not use nntrainer. However it is
still good transfer learning example using KNN.
Therefore, we change the execution file name from nntrainer_knn to
knn_sample.
main.cpp with NYI is added instead until we implement it with
nntrainer.

Issue #457


zhoonit (approved):
LGTM

PR #468, opened 2020-08-24
[ Coverity ] Fix coverity issues.

jijoongmoon:
This PR inclues coverity issues fixes.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -819,6 +819,8 @@ void Tensor::print(std::ostream &out) const {
     return;
   }
 
+  std::ios init(NULL);
+  init.copyfmt(out);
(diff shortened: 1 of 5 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -832,6 +834,7 @@ void Tensor::print(std::ostream &out) const {
     }
     out << "-------" << std::endl;
   }
+  out.copyfmt(init);
```
kparichay, line 837: I am not sure about this.
Rest all LGTM

jijoongmoon: In order to use arrange the digit for output, we use setw. But we didn't restore the original. This is for that purpose.

zhoonit (approved):
LGTM

PR #481, opened 2020-08-26
[Tensor] out-of-place to in-place call uniformly

kparichay:
Out-of-place operations call in-place operations with different clones and checks everywhere
Make a common abstraction to do this and use it for cleaner and readable code.


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM.

PR #482, opened 2020-08-26
[tensor] Update tensor operations to use cblas

### LINKED ISSUE #480: Use cblas operations for tensor
Switch out self-written operations for tensors to use cblas operators for better performance and higher precision.

kparichay:
Update tensor operations to use cblas
Further, update various operations to use stl functions than implement locally
Bug fix on sum_by_batch which calculated sum of absolute values earlier

Resolves #480


jijoongmoon (approved):
LGTM.

PR #484, opened 2020-08-26
[tensor] Update tensor apply

kparichay:
Update tensor apply to use std::transform than a simple loop


jijoongmoon (approved):
LGTM

Issue #483, opened 2020-08-26
Optimize layers with cblas


kparichay:
Optimize layers with cblas:
- [x] fc layer #487
- [x] conv layer #477
- [x] softmax #485
- [ ] pooling ()

kparichay:
Pooling layer cannot be directly opitmized with blas.
Can use std::transform and other c++ optimizations which have been deferred to later.

PR #485, opened 2020-08-26
[softmax] Update softmax implementation

kparichay:
Update softmax implementation to use tensor operations than running manual loops

See also #483

### DIFF
--- a/nntrainer/src/activation_layer.cpp
+++ b/nntrainer/src/activation_layer.cpp
@@ -11,9 +11,12 @@
  *
  */
 
-#include <activation_layer.h>
 #include <fstream>
+#include <functional>
 #include <iostream>
+#include <vector>
+
+#include <activation_layer.h>
 #include <layer.h>
 #include <lazy_tensor.h>
 #include <nntrainer_error.h>
@@ -22,7 +25,6 @@
 #include <parse_util.h>
 #include <tensor.h>
 #include <util_func.h>
-#include <vector>
 
 namespace nntrainer {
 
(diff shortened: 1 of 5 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -239,6 +239,20 @@ Tensor Tensor::add(float const &value) {
   return result;
 }
 
+void Tensor::saxpy(const unsigned int N, const float alpha, const float *X,
+                   const int incX, float *Y, const int incY) {
```
jijoongmoon, line 243: I think we have to make this dummy func in different file for better managing. Also we have to add more blas func like gemm.

nntrainer/src/activation_layer.cpp
```diff
@@ -148,48 +150,37 @@ Tensor ActivationLayer::softmax(Tensor const &t) {
...
-    }
+  // take exp
+  Tensor result = divisor.apply(exp_util);
+  rp = result.getData();
+  // take sum over batch
+  Tensor sum = result.sum_by_batch();
+
+  for (unsigned int k = 0; k < batch; k++) {
+    int index = k * feat_len;
+    std::transform(rp + index, rp + index + feat_len, rp + index,
```
jijoongmoon, line 179: transform is very helpful.

jijoongmoon (approved):
LGTM

nntrainer/src/tensor.cpp
kparichay: yes.

PR #487, opened 2020-08-27
[dot] Update tensor dot implementation

kparichay:
Update the implementation of tensor dot product
This update applies dot on the last dimesnion of input and second last dimension of passed tensor `m`
This allows use of dot in fc layer directly eliminating calling transpose explicitly but rather use transpose in blas

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -593,142 +593,127 @@ Tensor Tensor::sum(int axis) const {
 }
 
 /**
- * If the dim.batch() size of m is one, the it is reused for
- * every calculation along with dim.batch().
- * TODO: support dim.channel() > 1.
+ * @note: This dot product flattens the fist 3 axis for the purpose of
+ * computation. So, while performing, these matrices are behaving as 2-D
+ * matrices. The dimensions are restored while returning back the tensor.
  */
-Tensor Tensor::dot(Tensor const &m) const {
-  if (dim.width() != m.dim.height()) {
-    throw std::runtime_error("Error dim.width() != m.dim.height()");
+Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
+  if (m.dim.rank() > 2) {
+    throw exception::not_supported("Error: support only for rank of dot "
+                                   "matrix <= 2");
+  }
+  if (trans && dim.rank() > 2) {
+    throw exception::not_supported("Error: support only for rank of dot "
+                                   "matrix <= 2 with trans");
   }
 
-  if (dim.channel() != 1 || m.dim.channel() != 1) {
-    throw std::runtime_error("Error channel() != 1");
-  }
+  unsigned int dim1 = batch() * channel() * height();
+  unsigned int dim2 = width();
+  unsigned int mdim1 = m.batch() * m.channel() * m.height();
+  unsigned int mdim2 = m.width();
+  Tensor result;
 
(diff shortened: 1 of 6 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -592,143 +592,118 @@ Tensor Tensor::sum(int axis) const {
...
+  Tensor result;
 
-  int mwidth = m.dim.width();
-  Tensor result(dim.batch(), 1, dim.height(), mwidth);
+  unsigned int M, N, K, lda, ldb, ldc;
+
+  if (!trans && !trans_m) {
+    if (dim2 != mdim1)
+      throw std::runtime_error(
+        "Error: imcompatible dimensions for dot product");
```
zhoonit, line 611:
```suggestion
        "Error: incompatible dimensions for dot product");
```

And for others below

nntrainer/src/tensor.cpp
```diff
@@ -592,143 +592,118 @@ Tensor Tensor::sum(int axis) const {
...
- * If the dim.batch() size of m is one, the it is reused for
- * every calculation along with dim.batch().
- * TODO: support dim.channel() > 1.
- */
-Tensor Tensor::dot(Tensor const &m) const {
-  if (dim.width() != m.dim.height()) {
-    throw std::runtime_error("Error dim.width() != m.dim.height()");
+Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
+  if (m.dim.rank() > 2) {
+    throw std::runtime_error("Error: support only for rank of dot matrix <= 2");
```
zhoonit, line 597:
```suggestion
    throw exception::not_supported("Error: support only for rank of dot matrix <= 2");
```

nntrainer/src/tensor.cpp
```diff
@@ -592,143 +592,118 @@ Tensor Tensor::sum(int axis) const {
...
-/**
- * If the dim.batch() size of m is one, the it is reused for
- * every calculation along with dim.batch().
- * TODO: support dim.channel() > 1.
- */
-Tensor Tensor::dot(Tensor const &m) const {
-  if (dim.width() != m.dim.height()) {
-    throw std::runtime_error("Error dim.width() != m.dim.height()");
+Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
+  if (m.dim.rank() > 2) {
```
zhoonit, line 596: We discussed offline to add safeguard to `when (this->dim.rank() > 2)` in case of `trans` is true

jijoongmoon:
I think once we merge this PR, then we need to modify little bit something like #491

nntrainer/src/tensor.cpp
kparichay: done.

nntrainer/src/tensor.cpp
kparichay: done.

nntrainer/src/tensor.cpp
kparichay: done.

kparichay:
> I think once we merge this PR, then we need to modify little bit something like #491

Yes, once #491 is merged, can directly use gemm from there.

kparichay:
more unit tests for dot will soon follow.

jijoongmoon (approved):
LGTM

zhoonit (approved):
LGTM :)

PR #492, opened 2020-08-28
[unittest] Make unittest data close to int

kparichay:
Make unittest data close to int
This allows matching outputs with lower tolerance and higher precision allowing error in logic to found easily

Also adds a bug fix for pooling global max
    
**Self evaluation:**
    

nntrainer/include/tensor.h
```diff
@@ -512,7 +512,7 @@ class Tensor {
    * @param[in] d TensorDim
    * @note      Throws std::invalid_argument if size mismatch
    */
-  void reshape(TensorDim d);
+  void reshape(const TensorDim d);
```
zhoonit, line 515: If applicable... 
```suggestion
  void reshape(const TensorDim &d);
```

nntrainer/src/pooling2d_layer.cpp
```diff
@@ -111,7 +114,10 @@ sharedConstTensor Pooling2DLayer::backwarding(sharedConstTensor derivative,
   } break;
   case PoolingType::global_max: {
     for (unsigned int i = 0; i < derivative->getDim().getDataLen(); ++i) {
-      out[max_idx[i]] += derivative->getData()[i];
+      float der = derivative->getData()[i] / max_idx_global[i].size();
+      for (unsigned int m = 0; m < max_idx_global[i].size(); m++) {
+        out[max_idx_global[i][m]] += der;
```
jijoongmoon, line 119: Looks like it is averaging. is it right implementation?

kparichay: It averages in case there are multiple locations which have the same location and are the max values.
This is the implementation which matches tensorflow.
For global_max, tensorflow requires avg of these max values.
However, for max, tensorflow takes the first max value.
This difference in implementation is shown with int unittests as range (1, 10) produces similar integer numbers for the pooling to operate on.

jijoongmoon: Ohh. I got it. I didn't check there are multiple same max value.

nntrainer/include/tensor.h
kparichay: done.

zhoonit (approved):
LGTM

PR #493, opened 2020-08-28
[loadconfig] Load model and dataset in config at start

### LINKED ISSUE #389: buf size is determined after minibatch size is set at load config
+ batch seems not updated when loadConfig. please check this as well

kparichay:
Load model and dataset in config at start so that any dependencies
which model, dataset and layer have among them can be easily solved

Resolves #389

**Self evaluation:**

cc. @zhoonit

### DIFF
--- a/nntrainer/src/model_loader.cpp
+++ b/nntrainer/src/model_loader.cpp
@@ -38,6 +38,11 @@ namespace nntrainer {
 int ModelLoader::loadModelConfigIni(dictionary *ini, NeuralNetwork &model) {
   int status = ML_ERROR_NONE;
 
+  if (iniparser_find_entry(ini, "Model") == 0) {
+    ml_loge("there is no [Model] section in given ini file");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
   /** Default to neural network model type */
   model.net_type = (nntrainer::NetType)parseType(
     iniparser_getstring(ini, "Model:Type", unknown), TOKEN_MODEL);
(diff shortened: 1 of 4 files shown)


zhoonit (commented):
So this fixes the order of [model] and [dataset], am i correct?

nntrainer/src/model_loader.cpp
```diff
@@ -107,11 +117,12 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
   status = parse_and_set("Dataset:LabelData", DATA_LABEL, true);
   NN_RETURN_STATUS();
 
-  /// fixme: #299, #389
   int bufsize = iniparser_getint(ini, "DataSet:BufferSize", model.batch_size);
   ml_logd("buf size: %d", bufsize);
   status = model.data_buffer->setBufSize(bufsize);
   NN_RETURN_STATUS();
+  status = model.data_buffer->setBatchSize(model.batch_size);
```
zhoonit, line 124: This should be done before :122

zhoonit (commented):
Could  you uncomment `unittest_nntrainer_modelfile.cpp:186` and `unittest_nntrainer_modelfile.cpp:194` to check it is working fine?

nntrainer/src/model_loader.cpp
zhoonit: More Idealy, IMO, setBufSize shouldn't rely one batchSize rather initialize checks if bufsize > batchsize though. I think it is slightly out of topic.

kparichay:
> So this fixes the order of [model] and [dataset], am i correct?

Yes, this is fixed. Also fixes the issue where batch size is used before it is set.

nntrainer/src/model_loader.cpp
kparichay: > This should be done before :122

It is not necessary. bufsize is already set at line 120 and initialized with model.batch_size. This line sets model.data_buffer.batch_size.

kparichay:
> Could you uncomment `unittest_nntrainer_modelfile.cpp:186` and `unittest_nntrainer_modelfile.cpp:194` to check it is working fine?

Enabled the test cases, they pass now.

nntrainer/src/model_loader.cpp
kparichay: Updated.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #496, opened 2020-09-01
[dataset] check bufsize at init

kparichay:
This patch changes the initialization of buffer size.
The buffer size is now initialized with value 1 if user does not set it.
Then the proper checks on buffer size are performed at init
and buffer size is appropriately resized then.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/databuffer.cpp
+++ b/nntrainer/src/databuffer.cpp
@@ -324,12 +324,6 @@ int DataBuffer::setClassNum(unsigned int num) {
 
 int DataBuffer::setBufSize(unsigned int size) {
   int status = ML_ERROR_NONE;
-  /// TODO: move this to initialization than here.
-  if (size < batch_size) {
-    ml_loge("Error: buffer size must be greater than batch size");
-    SET_VALIDATION(false);
-    return ML_ERROR_INVALID_PARAMETER;
-  }
   train_bufsize = size;
   val_bufsize = size;
   test_bufsize = size;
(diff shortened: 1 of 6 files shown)


jijoongmoon (approved):
LGTM

nntrainer/src/databuffer.cpp
```diff
@@ -347,6 +341,46 @@ int DataBuffer::setBatchSize(unsigned int size) {
...
+  }
+
+  if (!class_num) {
+    ml_loge("Error: number of class must be set");
+    SET_VALIDATION(false);
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (!this->input_dim.getFeatureLen()) {
+    ml_loge("Error: featuer size must be set");
```
jijoongmoon, line 368: %s/featuer/feature

jijoongmoon:
Correct me if I'm wrong.
buffer size is starting with 1. and if buffer size is smaller than batch_size then buffer size will be batch_size. and if max_train is smaller than batch_size, then max_train will be batch_size. and when it the buffer is filled, it is stop to fill the buffer and died if there is no data to fill. then main thread will not get the data as much as the batch_size from buffer ( cause there is not enough data ) but the buffer thread broad cast "end of data" when it is died and train process just will stop without train.

nntrainer/src/databuffer.cpp
kparichay: done.

kparichay:
In my understanding, this patch does not change the behavior of the buffer size. The initialization of buffer size does happen with 1. However, at the start of the init of the databuffer, the buffer size is set back to batch size as it would have happened earlier. And then, the whole initialization of the data buffer happens as regular.

Independent of the PR, regarding the case you described above, I assume you are talking about databuffer_file as databuffer_func has max_train as 0. In databuffer_file, if max_train is smaller than batch size, then the total data in the file is smaller than batch size. This is an error case. There are 3 outcomes:
1. exit with error
2. reduce batch size and let the training complete
3. read more data multiple times from the file and somehow fill the batch

I think option 1 is better. Please advice how do you think this should be managed.
Note these changes will be reflected in a different PR.

zhoonit (approved):
LGTM

PR #505, opened 2020-09-03
Fix app test fail

### LINKED ISSUE #501: app_classification_capi_func test fails on my machine
It fails with the following log. 
 9/21 app_classification_capi_func   FAIL           0.24s (exit status 234 or signal 106 SIGinvalid)

I think it is my machine that has a problem though, I am reporting for a record.

zhoonit:
Fix app classification capi func test fails with negative number

- Add destroy routine
- s/model_path/save_path/

Resolves #501

**Self evaluation:**

### DIFF
--- a/Applications/Tizen_CAPI/capi_func.c
+++ b/Applications/Tizen_CAPI/capi_func.c
@@ -44,11 +44,11 @@ bool file_exists(const char *filename) {
   return (stat(filename, &buffer) == 0);
 }
 
-#define NN_RETURN_STATUS()         \
-  do {                             \
-    if (status != ML_ERROR_NONE) { \
-      return status;               \
-    }                              \
+#define NN_DESTORY_MODEL_RETURN_STATUS() \
+  do {                                   \
+    printf("run unsuccessful \n");       \
+    ml_train_model_destroy(model);       \
+    return 1;                            \
   } while (0)
 
 /**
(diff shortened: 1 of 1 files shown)


Applications/Tizen_CAPI/capi_func.c
```diff
@@ -299,75 +292,118 @@ int main(int argc, char *argv[]) {
...
+  if (status != ML_ERROR_NONE) {
+    goto ERROR;
+  }
 
   return 0;
+
+ERROR:
+  printf("run unsuccessful \n");
+
+  ml_train_layer_destroy(layers[0]);
```
kparichay, line 403: This isnt right. 
Once layers, optimizer and dataset have been added to the model, these destroy will fail.
lets not call them for all the error cases.

zhoonit: I am aware of that. My intention is that if layers/opt/dataset are added to the model, it is essentially noop while before they are set to the model, it will destroy. Correct me, if I am wrong.

kparichay: @zhoonit They are noop from the point of c functions, however it returns error. These applications are going to be references for people to create their own applications. Lets call destroy in proper order.

zhoonit: > Instead of using goto, I recommend to add these into macro. NN_RETURN_STATUS()

Thanks for the suggestion. I am afraid using macro is restricted for this case to properly delete objects.

zhoonit: Updated

jijoongmoon (approved):
LGTM

Applications/Tizen_CAPI/capi_func.c
```diff
@@ -299,75 +299,118 @@ int main(int argc, char *argv[]) {
...
+  if (status != ML_ERROR_NONE) {
+    ml_train_layer_destroy(layers[0]);
+    NN_RETURN_STATUS();
+  }
 
   /* create fully connected layer */
   status = ml_train_layer_create(&layers[1], ML_TRAIN_LAYER_TYPE_FC);
-  NN_RETURN_STATUS();
+  if (status != ML_ERROR_NONE) {
+    ml_train_layer_destroy(layers[1]);
```
kparichay, line 331: destroy layer[0] as well.

Applications/Tizen_CAPI/capi_func.c
```diff
@@ -299,75 +299,118 @@ int main(int argc, char *argv[]) {
...
+  }
 
   /* set property for fc layer */
   status = ml_train_layer_set_property(
     layers[1], "unit= 10", "activation=softmax", "bias_initializer=zeros",
     "weight_regularizer=l2norm", "weight_regularizer_constant=0.005",
     "weight_initializer=xavier_uniform", NULL);
-  NN_RETURN_STATUS();
+  if (status != ML_ERROR_NONE) {
+    ml_train_layer_destroy(layers[1]);
```
kparichay, line 341: ditto

Applications/Tizen_CAPI/capi_func.c
```diff
@@ -299,75 +299,118 @@ int main(int argc, char *argv[]) {
...
+    NN_RETURN_STATUS();
+  }
 
   /* create optimizer */
   status = ml_train_optimizer_create(&optimizer, ML_TRAIN_OPTIMIZER_TYPE_ADAM);
-  NN_RETURN_STATUS();
+  if (status != ML_ERROR_NONE) {
+    ml_train_optimizer_destroy(optimizer);
```
kparichay, line 355: destroy layers

kparichay: The same applies to all the error based exits below

Applications/Tizen_CAPI/capi_func.c
zhoonit: It is already added to the model so do not need to destroy.

Applications/Tizen_CAPI/capi_func.c
zhoonit: ditto

Applications/Tizen_CAPI/capi_func.c
zhoonit: Done

Applications/Tizen_CAPI/capi_func.c
zhoonit: `    ml_train_layer_destroy(layers[1]);` gone

Applications/Tizen_CAPI/capi_func.c
```diff
@@ -299,75 +299,116 @@ int main(int argc, char *argv[]) {
...
+    NN_RETURN_STATUS();
+  }
 
   /* train model with data files : epochs = 10 and store model file named
    * "model.bin" */
-  status = ml_train_model_run(model, "epochs=10", "model_file=model.bin", NULL);
-  NN_RETURN_STATUS();
+  status = ml_train_model_run(model, "epochs=10", "save_path=model.bin", NULL);
+  if (status != ML_ERROR_NONE) {
+    NN_RETURN_STATUS();
```
kparichay, line 404: ml_train_model_destroy?

kparichay (approved):
LGTM

Applications/Tizen_CAPI/capi_func.c
kparichay: yes

Applications/Tizen_CAPI/capi_func.c
```diff
@@ -46,9 +46,9 @@ bool file_exists(const char *filename) {
 
 #define NN_RETURN_STATUS()         \
```
kparichay, line 47: can rename to NN_DESTROY_MODEL_RETURN_STATUS.

zhoonit: Done

PR #513, opened 2020-09-04
[Tensor] Add multiple axes support

zhoonit:
This patch adds multiple axes support for sum and average.
which is esentially needed to build bn_layer

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -340,13 +340,13 @@ Tensor Tensor::sum_by_batch() {
 /**
  * @brief Calculate sum according to the axis.
  */
-Tensor Tensor::sum(int axis, float alpha) const {
+Tensor Tensor::sum(unsigned int axis, float alpha) const {
   Tensor ret;
 
   const float *data = getData();
 
   if (axis >= 4)
-    throw std::out_of_range("Error: Dimension cannot exceed 3");
+    throw std::out_of_range("Error: axis is invalid");
 
   if (dim.getDim()[axis] == 1 and alpha == 1.0)
     return this->clone();
@@ -403,6 +403,18 @@ Tensor Tensor::sum(int axis, float alpha) const {
   default:
     throw std::out_of_range("Error: Dimension cannot exceed 3");
   }
+  return ret;
+}
+
+Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
+  if (axes.empty())
+    throw std::invalid_argument("empty axes given");
+
+  Tensor ret = this->sum(axes[0], alpha);
+
+  for (unsigned int i = 1; i < axes.size(); ++i)
+    ret = ret.sum(axes[i]);
+
   return ret;
 }
 
@@ -673,12 +685,31 @@ void Tensor::read(std::ifstream &file) {
 /**
  * @brief Calculate average value according to the axis.
  */
-Tensor Tensor::average(int axis) const {
+Tensor Tensor::average(unsigned int axis) const {
+  if (axis >= MAXDIM)
+    throw std::out_of_range(
+      "negative axis or axis more then MAXDIM is invalid");
+
   unsigned int axis_size = dim.getDim()[axis];
   if (axis_size == 1)
     return this->clone();
 
   return this->sum(axis, 1.0 / ((float)axis_size));
+}
+
+Tensor Tensor::average(const std::vector<unsigned int> &axes) const {
+  if (axes.empty())
+    return this->average();
+
+  TensorDim ret_shape;
+  for (const auto &idx : axes) {
+    if (idx >= MAXDIM) {
+      throw std::out_of_range("axis more then MAXDIM is invalid");
+    }
+    ret_shape.setTensorDim(idx, dim.getTensorDim(idx));
+  }
+
+  return this->sum(axes, 1.0 / (float)ret_shape.getDataLen());
 }
 
 /**
--- a/nntrainer/include/tensor.h
+++ b/nntrainer/include/tensor.h
@@ -306,7 +306,16 @@ public:
    * @param[in] alpha Scale the sum by this value
    * @retval    Calculated Tensor
    */
-  Tensor sum(int axis, float alpha = 1.0) const;
+  Tensor sum(unsigned int axis, float alpha = 1.0) const;
+
+  /**
+   * @brief sum all the Tensor by multiple axes
+   *
+   * @param axes axes to sum along
+   * @param alpha Scale the sum by this value
+   * @return Tensor
+   */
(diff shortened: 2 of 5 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -411,8 +411,8 @@ Tensor Tensor::sum(int axis, float alpha) const {
 
   const float *data = getData();
 
-  if (axis >= 4)
-    throw std::out_of_range("Error: Dimension cannot exceed 3");
+  if (axis < 0 || axis >= 4)
```
kparichay, line 414: change axis to unsigned int than adding the check of `axis < 0`.

kparichay: Also, can axis be equal to 0 and 4 both valid? out

nntrainer/src/tensor.cpp
```diff
@@ -545,6 +545,18 @@ Tensor Tensor::sum(int axis, float alpha) const {
   return ret;
 }
 
+Tensor Tensor::sum(const std::vector<int> &axes, float alpha) const {
+  if (axes.empty() || axes.size() > MAXDIM)
+    throw std::invalid_argument("empty or too many axes given");
+
+  Tensor ret = this->sum(axes[0], alpha);
+
```
jijoongmoon, line 553: It looks alpha scales multiple times. Could you check this?

nntrainer/src/tensor.cpp
```diff
@@ -829,13 +841,29 @@ void Tensor::read(std::ifstream &file) {
  * @brief Calculate average value according to the axis.
  */
 Tensor Tensor::average(int axis) const {
+  if (axis < 0 || axis >= MAXDIM)
```
jijoongmoon, line 844: better to use unsigned int? or do we support -1 like python?

nntrainer/src/tensor.cpp
zhoonit: > change axis to unsigned int than adding the check of axis < 0.
Changed to unsigned int. 

> Also, can axis be equal to 0 and 4 both valid? out
Do you mean axis 0 and 4 are both 0? If that's the case semantics for the axis has to be changed. I think it'd better be done in a different patch.

nntrainer/src/tensor.cpp
```diff
@@ -829,13 +841,29 @@ void Tensor::read(std::ifstream &file) {
...
 }
 
+Tensor Tensor::average(const std::vector<int> &axes) const {
+  if (axes.size() > MAXDIM)
+    throw std::invalid_argument("too many axes given");
+
+  if (axes.empty())
+    return this->average();
+
+  int axes_size = std::accumulate(axes.begin(), axes.end(), 0);
```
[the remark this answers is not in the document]
zhoonit: Fixed

nntrainer/src/tensor.cpp
zhoonit: Right. It is fixed and I added a regression test.

nntrainer/src/tensor.cpp
zhoonit: If we support a negative axis, I think it'd better be done in a different patch. I changed to unsigned int for now. Thanks.

nntrainer/src/tensor.cpp
kparichay: > Also, can axis be equal to 0 and 4 both valid? out

I meant that there are 4 dimensions. However,  current check allows for 5 values on axis [0, 4].
I think it should be [0, 4)

nntrainer/src/tensor.cpp
```diff
@@ -673,14 +685,37 @@ void Tensor::read(std::ifstream &file) {
...
+Tensor Tensor::average(const std::vector<unsigned int> &axes) const {
+  if (axes.size() > MAXDIM)
+    throw std::invalid_argument("too many axes given");
+
+  if (axes.empty())
+    return this->average();
+
+  unsigned int axes_size = 1;
+  for (const auto idx : axes) {
+    if (idx >= MAXDIM) {
```
kparichay, line 709: Create a new dimension where the given axes will be set to 1.
Then use `axes_size = dim.getDataLen() / newDim.getDataLen()`
This ensures that even though an axes is passed twice, it will produce correct answer.

nntrainer/src/tensor.cpp
```diff
@@ -673,14 +685,37 @@ void Tensor::read(std::ifstream &file) {
...
+
   unsigned int axis_size = dim.getDim()[axis];
   if (axis_size == 1)
     return this->clone();
 
   return this->sum(axis, 1.0 / ((float)axis_size));
 }
 
+Tensor Tensor::average(const std::vector<unsigned int> &axes) const {
+  if (axes.size() > MAXDIM)
```
kparichay, line 701: Maybe this is redundant when line 709 check exists.

zhoonit: FYI, line 701 checks if given axes should be less than 4. for example {0, 1, 2, 3, 0} is invalid. 
I'll allow the same axes passing twice with deleting these bits.

line 709 checks if each element is over maxdim for example {0, 3, 5} is invalid.

kparichay: Right, however {0,1,2,0} is still allowed. 
https://github.com/nnstreamer/nntrainer/pull/513#discussion_r485320976 can address this.
But should {0,1,2,0} be treated any different from {0,1,2,2,0}, I am not so sure.

So, even if axis.sizes() is more than MAXDIM, it will be covered in the checks later.
But when the input is correct, this is just a redundant check.

zhoonit: Rearranged. PTAL

nntrainer/src/tensor.cpp
```diff
@@ -673,14 +685,33 @@ void Tensor::read(std::ifstream &file) {
...
 
   return this->sum(axis, 1.0 / ((float)axis_size));
 }
 
+Tensor Tensor::average(const std::vector<unsigned int> &axes) const {
+  if (axes.empty())
+    return this->average();
+
+  TensorDim ret_shape;
+  for (const auto idx : axes) {
```
kparichay, line 705: just a minor thing - `const auto &idx`.

kparichay (approved):
LGTM

nntrainer/src/tensor.cpp
kparichay: Looks good now.

jijoongmoon (approved):
LGTM. I will merge this after fix of minor thing.

nntrainer/src/tensor.cpp
zhoonit: Done

PR #514, opened 2020-09-04
[wait for long] [DataBuffer_v2] Add support for data buffer with file

kparichay:
This patch adds data buffer v2
Data buffer v2 is designed to minimize memory copies and support many threads
The data buffer creates a buffer pool where memory is already allocated
The reader threads fill these allocated memory and reports back to the main thread
The allocated memory pool is allocated in batchsize units so that this memory can be directly
used in the tensor

Implementation of notification mechanism, and some data management is remaining

**Self evaluation:**


nntrainer/include/databuffer_file_cb.h
```diff
@@ -0,0 +1,80 @@
...
+ * @brief Data buffer with file helpers class
+ */
+class DataBufferFileUserData {
+public:
+  DataBufferFileUserData(const std::string file) :
+    size(0),
+    cur_loc(0),
+    filename(file) {
+    data_stream = std::ifstream(filename, std::ios::in | std::ios::binary);
+    if (!data_stream.good())
```
jijoongmoon, line 39: move to cpp?

kparichay: I will move.

nntrainer/include/databuffer_file_cb.h
```diff
@@ -0,0 +1,80 @@
...
+
+  size_t size, cur_loc;
+  std::ifstream data_stream;
+  std::string filename;
+  std::vector<size_t> inputs_size, labels_size;
+};
+
+/**
+ * @brief   Dataset generator callback for handling data from dat format files
+ */
```
jijoongmoon, line 75: are you going to write details about variables?

kparichay: yes, they will be added.

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,231 @@
...
+  if (buffer_len < batch_size)
+    buffer_len = batch_size;
+
+  /**
+   * batch size data is allocated at once in the buffer for data to avoid
+   * copying to data back into another holder which has contiguous data.
+   * This limits the buffer length to be a multiple of batch size.
+   */
+  batched_buffer_len = (buffer_len / batch_size);
+  buffer_len = batched_buffer_len * batch_size;
```
jijoongmoon, line 123: I think this is duplication. buffer_len is used to set batched_buffer_len and set the buffer_len using this.

jijoongmoon:
Need to fix CI errors.

PR #515, opened 2020-09-04
[ BLAS ] Modify to use CUBLAS for gemm

jijoongmoon:
Move CUBLAS gemm routine into blas_interface

**Self evaluation:**

### DIFF
--- a/nntrainer/src/blas_interface.cpp
+++ b/nntrainer/src/blas_interface.cpp
@@ -123,7 +123,33 @@ void sgemm(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, CBLAS_TRANSPOSE TransB,
            const float *B, const unsigned int ldb, const float beta, float *C,
            const unsigned int ldc) {
 
-#ifdef USE_BLAS
+#ifdef USE_CUBLAS
+  int devID = 0;
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, devID);
+  float *d_A, *d_B, *d_C;
+
+  unsigned int size_A = M * K * sizeof(float);
+  unsigned int size_B = K * N * sizeof(float);
+  unsigned int size_C = M * N * sizeof(float);
+
+  cudaMalloc((void **)&d_A, size_A);
+  cudaMalloc((void **)&d_B, size_B);
+  cudaMemcpy(d_A, A, size_A, cudaMemcpyHostToDevice);
+  cudaMemcpy(d_B, B, size_B, cudaMemcpyHostToDevice);
+  cudaMalloc((void **)&d_C, size_C);
+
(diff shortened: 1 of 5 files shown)


nntrainer/src/blas_interface.cpp
```diff
@@ -123,13 +123,41 @@ void sgemm(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, CBLAS_TRANSPOSE TransB,
...
+  cudaMalloc((void **)&d_A, size_A);
+  cudaMalloc((void **)&d_B, size_B);
+  cudaMemcpy(d_A, A, size_A, cudaMemcpyHostToDevice);
+  cudaMemcpy(d_B, B, size_B, cudaMemcpyHostToDevice);
+  cudaMalloc((void **)&d_C, size_C);
+
+  cublasHandle_t handle;
+  cublasCreate(&handle);
+
+  cublasOperation_t transA = trans ? CUBLAS_OP_T : CUBLAS_OP_N;
```
kparichay, line 145: trans variables here should be compared with CblasTrans.

nntrainer/src/blas_interface.cpp
```diff
@@ -123,13 +123,41 @@ void sgemm(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, CBLAS_TRANSPOSE TransB,
...
+  cublasOperation_t transB = trans_m ? CUBLAS_OP_T : CUBLAS_OP_N;
+  cublasSgemm(handle, CUBLAS_OP_N, CUBLAS_OP_N, N, M, K, &alpha, d_B, N, d_A, K,
+              &beta, d_C, N);
+
+  cudaMemcpy(C, d_C, size_C, cudaMemcpyDeviceToHost);
+  cublasDestroy(handle);
+#else
```
kparichay, line 152: use #elseif defined(USE_BLAS)

jijoongmoon: That's because when there are both Definition. if DUSE_BLAS & DUSE_CUBLAS is defined, then gemm use CUBLAS and others use BLAS.

nntrainer/src/blas_interface.cpp
jijoongmoon: fixed. thanks.

nntrainer/src/blas_interface.cpp
kparichay: Exactly. For this the following should be used in gemm
```c
#if defined USE_CUBLAS
...
#elif defined USE_BLAS
...
#endif
```

jijoongmoon: fixed.

zhoonit (approved):
LGTM

kparichay (approved):
LGTM

PR #517, opened 2020-09-07
[wait for #529] [tensor] Update default constructor of tensor to not throw

kparichay:
The default constructor does no heap memory allocation (size is 0)
However it can still throw which isn't right
This patch updates the tensor default constructor to not throw because of heap allocation
to create less issues with static checkers which requires to add try catch around tensor declarations

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -71,24 +71,19 @@ static auto rng = [] {
   return rng;
 }();
 
-Tensor::Tensor(const TensorDim &d, const float *buf) :
-  dim(d),
-  strides{{1, 2, 3}},
-  is_contiguous(true),
-  data(d.getDataLen() == 0
-         ? nullptr
-         : std::shared_ptr<float>(new float[d.getDataLen()],
-                                  std::default_delete<float[]>())) {
-  if (d.getDataLen() == 0) {
-    return;
-  }
+Tensor::Tensor(const TensorDim &d, const float *buf) : Tensor() {
+  dim = d;
+  if (d.getDataLen() != 0) {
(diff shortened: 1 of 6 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -76,24 +76,19 @@ static auto rng = [] {
...
-    return;
-  }
-
-  // todo: initialize appropriate strides
-  if (buf != nullptr) {
-    float *data = getData();
-    unsigned int len = length();
-
-    scopy(len, buf, 1, data, 1);
+Tensor::Tensor(const TensorDim &d, const float *buf) : Tensor() {
```
zhoonit, line 79: IMO, `dim = d` should be done even if it is empty. It might cause a problem when tensordim starts to have any meta information.

zhoonit (approved):
LGTM with minor comment

nntrainer/src/tensor.cpp
kparichay: Good catch. Updated.

kparichay: This creates an issue with the bn_layer.
bn creates weights without gradient but passes it to the optimizer.
The optimizer creates its own variables with that empty gradient which throws error of creating tensor with 0 dimensions.
How about I disable the bn unit test and you can check this while you implementing it?

zhoonit: AFAIK, this should not affect as bn_layer does not pass weights that does not have grads. I'll check it through anyway. Thanks.

kparichay: @zhoonit #529 resolves this issue.

jijoongmoon (approved):
LGTM

PR #520, opened 2020-09-07
[wait for #305] [databuffer_v2] Add databuffer v2 implementation

kparichay:
Add databuffer v2 implementation
This implements the buffer and thread management along
    
**Self evaluation:**
    

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+        throw std::logic_error("Background thread is not running anymore");
+    }
+
+    bool data_available;
+    /** check if data is available */
+    {
+      std::lock_guard<std::mutex> batched_buffer_data_avail_check_lock(
+        batched_buffer_m);
+      auto batched_buffer_elem = batched_buffer.front();
+      data_available = std::get<2>(batched_buffer_elem) == batch_size;
```
zhoonit, line 194: It seems like we don't need to use batched_buffer if we check `data_available` by available_buffer_idx > batch_size making code and locks simpler?

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    /**
+     * THREAD_REQUEST_TO_STOP : requested by main thread to stop
+     * THREAD_ERROR : thread is in error state, exit
+     * THREAD_EPOCH_FINISHED : thread finished work, exit
+     */
+    {
+      std::lock_guard<std::mutex> thread_stop(thread_m);
+      if (thread_state != ThreadStates::THREAD_RUNNING) {
+        thread_state = ThreadStates::THREAD_STOPPED;
```
zhoonit, line 267: THREAD_ERROR will be overridden here. Is this possible scenario and is that okay?

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+    /** if the buffer space is received, then call generator cb and get the data
+     */
+    if (buffer_space_available) {
+      void **input_holder = std::get<0>(buffer_elem);
+      void **label_holder = std::get<1>(buffer_elem);
+
+      int status = generator((float **)input_holder, (float **)label_holder,
+                             &last_element, gen_user_data);
+
+      if (status != ML_ERROR_NONE && !last_element) {
```
zhoonit, line 292:
```suggestion
      if (status == ML_ERROR_NONE && !last_element) {
```

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+        /** Update the batched buffer element loaded count */
+        std::lock_guard<std::mutex> batched_buffer_update_lock(
+          batched_buffer_m);
+        auto batched_buffer_elem = getNthBatchedBufferElement();
+        std::get<2>(batched_buffer_elem) += 1;
+        /** notify on buffer produce */
+        buffer_cond_filled.notify_all();
+      } else if (status == ML_ERROR_NONE) {
```
zhoonit, line 300: if `status != ML_ERROR_NONE` it should go to error handling section not here.

```suggestion
      } else if (status == ML_ERROR_NONE && last_element == true) {
```

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+        thread_state = ThreadStates::THREAD_STOPPED;
+        buffer_cond_filled.notify_all();
+        return;
+      }
+    }
+
+    /** check if buffer has space available to load the data */
+    {
+      std::lock_guard<std::mutex> buffer_data_fill_lock(buffer_m);
+      buffer_space_available = buffer_len - avail_buffer_idx - 1 > 0;
```
zhoonit, line 276: Check buffer overflow somewhere?

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+        thread_state != ThreadStates::THREAD_EPOCH_FINISHED)
+      throw std::runtime_error("Calling get data on a thread not running");
+  }
+
+  /** Keep on looping till a batch of data is ready */
+  while (true) {
+    {
+      std::lock_guard<std::mutex> thread_start(thread_m);
+      if (thread_state == ThreadStates::THREAD_EPOCH_FINISHED)
+        return false;
```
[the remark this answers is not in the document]
kparichay: Switched to -EINVAL. However, I am not sure what is the best return value without introducing a new return value.

nntrainer/src/databuffer_v2.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+        if (avail_buffer_idx < 0 || avail_buffer_idx > buffer_len)
+          throw std::logic_error("Buffer overflow error");
+      }
+
+      /** Add new space to the buffer */
+      pushBatchedData();
+      /** Wake up background threads waiting for more space to fill data */
+      buffer_cond_hungry.notify_all();
+      return true;
```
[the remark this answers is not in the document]
kparichay: changed to return 0.

nntrainer/src/databuffer_v2.cpp
kparichay: Yes, in case of multiple background threads, it will be overridden. 
Updated.

kparichay: Actually there is no much use for THREAD_REQUEST_TO_STOP and THREAD_STOPPED state.
Merged into THREAD_JOINABLE as per https://github.com/nnstreamer/nntrainer/pull/305/files#r484586941.

Issue #464, opened 2020-08-21
MNIST application is not deterministic


kparichay:
MNIST application with just fc layer and loss run over 10 times does not start from the same location.
This might be issue of data loading or network or some initialization and needs quick fix.

I think @zhoonit saw this kind of similar bug in his sample application too.

kparichay:
This un-deterministic behavior is also there on `nntrainer_capi_nnmodel.train_with_file_01_p` unittest.

kparichay:
@jijoongmoon If you have verified this, please close.

jijoongmoon:
fixed by #522

Issue #427, opened 2020-08-10
Nondeterministic behavior on Databuffer


zhoonit:
When it runs to the full data(to the end), it sometimes returns -22 sometimes runs fine.

kparichay:
@zhoonit Can you please once if this resolved on your side. Also please check if the un-deterministic beahvior of the example app is also fixed.

zhoonit:
> @zhoonit Can you please once if this resolved on your side. Also please check if the un-deterministic beahvior of the example app is also fixed.

Yes I'll check with this PR

zhoonit:
It is still Nondeterministic. reported here: https://github.com/nnstreamer/nntrainer/pull/428#issuecomment-689253629

jijoongmoon:
closed by #525

PR #522, opened 2020-09-07
[ Coverity ] Fix Race Condition issue about databuffer

jijoongmoon:
. Add Lock to set data ready flag.
. change order to set the condition

. Fix undeterministic behavior of databuffer
. related issues : #427, #464

**Self evaluation:**

### DIFF
--- a/nntrainer/src/databuffer.cpp
+++ b/nntrainer/src/databuffer.cpp
@@ -121,6 +121,7 @@ int DataBuffer::run(BufferType type) {
 
 int DataBuffer::clear(BufferType type) {
   int status = ML_ERROR_NONE;
+  NN_EXCEPTION_NOTI(DATA_NOT_READY);
   switch (type) {
   case BUF_TRAIN: {
     train_running = false;
@@ -130,7 +131,6 @@ int DataBuffer::clear(BufferType type) {
     this->train_data_label.clear();
     this->cur_train_bufsize = 0;
     this->rest_train = max_train;
-    trainReadyFlag = DATA_NOT_READY;
   } break;
   case BUF_VAL: {
     val_running = false;
@@ -140,7 +140,6 @@ int DataBuffer::clear(BufferType type) {
     this->val_data_label.clear();
     this->cur_val_bufsize = 0;
     this->rest_val = max_val;
-    valReadyFlag = DATA_NOT_READY;
   } break;
   case BUF_TEST: {
     test_running = false;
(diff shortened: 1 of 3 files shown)


nntrainer/src/databuffer.cpp
```diff
@@ -187,12 +185,15 @@ bool DataBuffer::getDataFromBuffer(BufferType type, vec_4d &outVec,
...
     std::vector<int> list;
-    std::unique_lock<std::mutex> ultrain(readyTrainData);
-    cv_train.wait(ultrain, [this]() -> int { return trainReadyFlag; });
-
-    if (train_data.size() < batch_size || trainReadyFlag == DATA_ERROR ||
-        trainReadyFlag == DATA_END) {
-      return false;
+    while (true) {
+      std::unique_lock<std::mutex> ultrain(readyTrainData);
+      cv_train.wait(ultrain, [this]() -> int { return trainReadyFlag; });
```
zhoonit, line 190: Could you check if this is valid?
https://en.cppreference.com/w/cpp/thread/condition_variable/wait

```suggestion
      cv_train.wait(ultrain, [this]() -> bool{ return trainReadyFlag; });
```

nntrainer/src/databuffer.cpp
```diff
@@ -187,12 +185,15 @@ bool DataBuffer::getDataFromBuffer(BufferType type, vec_4d &outVec,
...
+      if (trainReadyFlag == DATA_ERROR || trainReadyFlag == DATA_END) {
```
zhoonit, line 191:
```suggestion
      if (trainReadyFlag == DATA_ERROR || trainReadyFlag == DATA_END || train_data.empy()) {
```

to eliminate edge cases? (and for other cases)

nntrainer/src/databuffer.cpp
```diff
@@ -374,10 +383,15 @@ int DataBuffer::init() {
   this->cur_val_bufsize = 0;
   this->cur_test_bufsize = 0;
 
+  readyTrainData.lock();
```
kparichay, line 386: do we need locking here? this is init and no thread is active yet.

jijoongmoon: coverity says we need.

nntrainer/src/databuffer.cpp
jijoongmoon: it will be checked at line 194

nntrainer/src/databuffer.cpp
jijoongmoon: fixed. thanks.

nntrainer/src/databuffer.cpp
kparichay: ok

kparichay (approved):
LGTM

zhoonit (approved):
LGTM

PR #524, opened 2020-09-09
[ Example ] Add Inference Mode & Example

jijoongmoon:
This PR includes:
. Add Inferenece Mode
  Add 'is_train' to identify mode.
  : Default is training mode(true). If this is set false, then it is
    inference mode. If it is inference mode, then data buffer is not
    initialized.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -73,7 +73,20 @@ public:
   /**
    * @brief     Constructor of NeuralNetwork Class
    */
-  NeuralNetwork();
+  NeuralNetwork(bool istrain = true) :
+    is_train(istrain),
+    batch_size(1),
+    epochs(1),
+    loss(0.0f),
+    loss_type(LossType::LOSS_UNKNOWN),
+    weight_initializer(WEIGHT_UNKNOWN),
+    net_type(NET_UNKNOWN),
+    data_buffer(NULL),
+    continue_train(false),
+    iter(0),
+    initialized(false),
+    def_name_count(0),
+    loadedFromConfig(false) {}
 
   /**
    * @brief     Destructor of NeuralNetwork Class
@@ -99,6 +112,7 @@ public:
     swap(lhs.layer_names, rhs.layer_names);
     swap(lhs.def_name_count, rhs.def_name_count);
     swap(lhs.loadedFromConfig, rhs.loadedFromConfig);
+    swap(lhs.is_train, rhs.is_train);
   }
 
   /**
@@ -127,6 +141,7 @@ public:
 
   /**
    * @brief     Create and load the Network with ini configuration file.
+   * @param[in] config config file path
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
@@ -233,6 +248,13 @@ public:
   int train(std::vector<std::string> values);
 
   /**
+   * @brief     Run NeuralNetwork inference
+   * @param[in] X input tensor
+   * @retval shared_ptr<const Tensor>
+   */
+  sharedConstTensor inference(const Tensor X);
+
+  /**
    * @brief     Run NeuralNetwork train with callback function by user
    * @param[in] train_func callback function to get train data. This provides
    * batch size data per every call.
@@ -252,6 +274,18 @@ public:
    * @retval #ML_ERROR_INVALID_PARAMETER not ready to init.
    */
   int isInitializable();
+
+  /**
+   * @brief     check train or inference. Default is True (train)
+   * @retval bool True : training Mode. False : Inference Mode
+   */
+  bool isTrain() { return is_train; };
+
+  /**
+   * @param[in] train true for train mode. false for inference mode
+   * @brief     set Running Mode : Train or Inference
+   */
+  void setMode(bool train) { is_train = train; };
 
   /**
    * @brief     add layer into neural network model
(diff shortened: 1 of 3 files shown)


kparichay:
Is an application really needed for just inference mode (this is just a feature)?
IMO, unit test is more suitable for this than application as with unittests, we can outputs and everything but not with applications.
If you think application is really needed, then IMO adding it to an existing application like MNIST is better, otherwise, there is just too much code copy and maintenance become difficult with this.

jijoongmoon:
Application is just for the test. Nothing more. The reason I write the Application test, we do not have any inference only case. 
According to this Application test code, I check which and where should be changed slightly. Well, It could be done in unittest as well. I will change this later pr. ( I think bitmap helper should be kind of util so and use for unit test or application which needed. ) 
I will delete the application in this pr and make another include this.

nntrainer/src/model_loader.cpp
```diff
@@ -50,8 +50,10 @@ int ModelLoader::loadModelConfigIni(dictionary *ini, NeuralNetwork &model) {
   model.loss_type = (LossType)parseType(
     iniparser_getstring(ini, "Model:Loss", unknown), TOKEN_LOSS);
   model.save_path = iniparser_getstring(ini, "Model:Save_path", "./model.bin");
-  model.batch_size =
-    iniparser_getint(ini, "Model:Batch_Size", model.batch_size);
+  if (model.isTrain()) {
```
kparichay, line 53: !model.isTrain()?

nntrainer/src/model_loader.cpp
```diff
@@ -240,8 +242,10 @@ int ModelLoader::loadFromIni(std::string ini_file, NeuralNetwork &model) {
   status = loadModelConfigIni(ini, model);
   NN_INI_RETURN_STATUS();
 
-  status = loadDatasetConfigIni(ini, model);
-  NN_INI_RETURN_STATUS();
+  if (model.isTrain()) {
```
kparichay, line 245: ditto

nntrainer/include/neuralnet.h
```diff
@@ -75,6 +75,12 @@ class NeuralNetwork {
    */
   NeuralNetwork();
 
+  /**
+   * @param[in] istrain train mode if true. else inference model
+   * @brief     Constructor of NeuralNetwork Class
+   */
+  NeuralNetwork(bool istrain);
```
kparichay, line 82: add a default value to this `istrain` and then have only 1 constructor.

nntrainer/src/neuralnet.cpp
```diff
@@ -56,7 +56,23 @@ NeuralNetwork::NeuralNetwork() :
   iter(0),
   initialized(false),
   def_name_count(0),
-  loadedFromConfig(false) {}
+  loadedFromConfig(false),
+  is_train(true) {}
+
+NeuralNetwork::NeuralNetwork(bool istrain) :
```
kparichay, line 62: this won't be needed when the constructors are merged

nntrainer/src/model_loader.cpp
```diff
@@ -240,8 +242,10 @@ int ModelLoader::loadFromIni(std::string ini_file, NeuralNetwork &model) {
...
+    status = loadDatasetConfigIni(ini, model);
```
kparichay, line 246: Given there is no data set for the model in inference mode, are you planning to manually feed in the data bypassing the dataset? or the dataset must be set externally in this case?

nntrainer/src/model_loader.cpp
jijoongmoon: no.. I guess 1 is enough for inference.

kparichay:
> Application is just for the test. Nothing more. The reason I write the Application test, we do not have any inference only case.
> According to this Application test code, I check which and where should be changed slightly. Well, It could be done in unittest as well. I will change this later pr. ( I think bitmap helper should be kind of util so and use for unit test or application which needed. )
> I will delete the application in this pr and make another include this.

We should try writing unittest first to check correctness of the features than applications. Application tests are good, but in that case, its better to merge with existing application, if possible.

nntrainer/src/model_loader.cpp
jijoongmoon: this databuffer is needed when train mode.

nntrainer/src/model_loader.cpp
jijoongmoon: dataset is not needed I guess for inference.

nntrainer/src/neuralnet.cpp
jijoongmoon: but this will make easier. I guess. otherewise we need some api for set 'is_train'

nntrainer/src/model_loader.cpp
zhoonit: batch_size will not be parsed if is not given anyway. As this code is gonna be frozen for the release, IMO we can let it parse unless there is a strict reason not to. And for dataset as well.

nntrainer/src/neuralnet.cpp
zhoonit: I guess @kparichay means merging this to the default constructor with default value set. 
```suggestion
NeuralNetwork::NeuralNetwork(bool istrain = true) :
 batch_size(1),
 epochs(1),
 ...
 is_train(istrain) {}
```

zhoonit (approved):
LGTM

nntrainer/src/model_loader.cpp
kparichay: as discussed offline, the data will be passed directly.

nntrainer/src/model_loader.cpp
jijoongmoon: for inference, we do not need batch_size.

jijoongmoon:
> > Application is just for the test. Nothing more. The reason I write the Application test, we do not have any inference only case.
> > According to this Application test code, I check which and where should be changed slightly. Well, It could be done in unittest as well. I will change this later pr. ( I think bitmap helper should be kind of util so and use for unit test or application which needed. )
> > I will delete the application in this pr and make another include this.
> 
> We should try writing unittest first to check correctness of the features than applications. Application tests are good, but in that case, its better to merge with existing application, if possible.

I guess this is case by case.

PR #527, opened 2020-09-10
[Tensor] Add broadcasting support for add_i ops @open sesame 09/14 19:00

zhoonit:
Batchnormalization Layer needs extensive broadcasting support other than
batchwise add.
Add broadcasting to `add` family while keeping vectorization from cblas
as much as possible.

- Initialize actual strides
- Implement loop mechanism to `add_i`
- Add extensive tests for broadcasted add_i

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -73,11 +73,11 @@ static auto rng = [] {
 
 Tensor::Tensor(const TensorDim &d, const float *buf) : Tensor() {
   dim = d;
+  strides = d.computeStrides();
   if (d.getDataLen() != 0) {
     data = std::shared_ptr<float>(new float[d.getDataLen()],
                                   std::default_delete<float[]>());
 
-    // todo: initialize appropriate strides
     if (buf != nullptr) {
       float *data = getData();
       unsigned int len = length();
@@ -148,8 +148,7 @@ void Tensor::setRandUniform(float min, float max) {
 }
 
 Tensor::Tensor(
-  std::vector<std::vector<std::vector<std::vector<float>>>> const &d) :
-  strides{{1, 2, 3}} {
+  std::vector<std::vector<std::vector<std::vector<float>>>> const &d) {
 
   if (d.empty() || d[0].empty() || d[0][0].empty() || d[0][0][0].empty()) {
     throw std::out_of_range(
@@ -160,6 +159,7 @@ Tensor::Tensor(
   dim.channel(d[0].size());
   dim.height(d[0][0].size());
   dim.width(d[0][0][0].size());
+  strides = dim.computeStrides();
   data = std::shared_ptr<float>(new float[dim.getDataLen()],
                                 std::default_delete<float[]>());
   is_contiguous = true;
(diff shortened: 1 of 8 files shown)


test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -448,6 +448,204 @@ TEST(nntrainer_Tensor, add_i_01_n) {
...
+      114, 115, 116, 117, 118, 119, 120, 121};
+    nntrainer::Tensor answer(ref_dim, answer_data);
+    int status = t.add_i(m);
+    EXPECT_EQ(status, ML_ERROR_NONE);
+    EXPECT_EQ(t, answer);
+  }
+  {
+    nntrainer::Tensor t = ranged(3, 2, 4, 5);
+    nntrainer::Tensor m = ranged(1, 1, 1, 1);
+    float answer_data[] = {
```
jijoongmoon, line 618: It might be clear when m is set by other number like 1 or 2.

nntrainer/src/tensor.cpp
```diff
@@ -775,4 +811,58 @@ Tensor Tensor::standardization() const {
...
+      continue;
+    }
+
+    if (dim.getTensorDim(i) == m_dim.getTensorDim(i)) {
+      e.strides[i] = m.strides[i];
+      continue;
+    }
+
+    std::stringstream ss;
+    ss << "[computeExternalLoop] given shape is not broadcastable,\n"
```
kparichay, line 835: Update to `broadcasting only allowed for dimension value of only 1`. This is gives the reason why this wont work.

nntrainer/src/tensor.cpp
```diff
@@ -775,4 +811,58 @@ Tensor Tensor::standardization() const {
...
+  if (m.length() > this->length())
+    throw exception::not_supported("broadcasting *this is not supported");
+
+  const TensorDim m_dim = m.getDim();
+
+  ExternalLoopInfo e;
+
+  for (unsigned int i = 0; i < MAXDIM; ++i) {
+    if (m_dim.getTensorDim(i) == 1) {
+      e.strides[i] = 0;
```
kparichay, line 825: Add a contructor for `ExternalLoopInfo` inside that struct where these values are initialized (may be to 0).
Then remove this.

nntrainer/src/tensor.cpp
```diff
@@ -775,4 +811,58 @@ Tensor Tensor::standardization() const {
...
+  for (int axis = 2; axis >= 0; --axis) {
+    unsigned int cur_dim = m_dim.getTensorDim(axis);
+
+    if (first_dim_none) {
+      if (cur_dim != 1) {
+        e.buffer_axis = axis;
+        break;
+      }
+    } else {
+      if (cur_dim == 1) {
```
kparichay, line 854: there should also be a check here to match `dim.getTensorDim(axis) == m_dim.getTensorDim(axis)`??
Consider the case of dim = (3,5,1,4) and m_dim = (3,1,1,4)
e.buffer_axis = 2 and I think that's what not is intended.

nntrainer/src/tensor.cpp
```diff
@@ -224,24 +224,59 @@ Tensor Tensor::add(float const &value) { CLONE_OP_I(add_i, value); }
...
-    }
+  unsigned int offset = 0;
+  unsigned int m_offset = 0;
+
+  if (e.buffer_axis == -1) {
+    saxpy(e.buffer_size, alpha, mbuf, e.strides[3], buf, strides[3]);
   } else {
-    if (dim.batch() != m.dim.batch()) {
-      return ML_ERROR_INVALID_PARAMETER;
+    for (unsigned int b = 0; b < batch(); ++b) {
```
kparichay, line 245: find the first n axis till which stride matches,
then loop over n axis in 1-D fashion
and then call add_i over the new data itself.
The current method is not scalable to multiple operations IMO.

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -1163,7 +1361,7 @@ TEST(nntrainer_Tensor, constructor_from_shared_const_ptr_shares_variable_n) {
   nntrainer::Tensor B = *A;
   nntrainer::Tensor C = A->clone();
 
-  B.setValue(3, 4, 5, 6, 2.0f);
+  B.setValue(2, 3, 4, 5, 2.0f);
```
kparichay, line 1364: any reason to update this here?

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -1210,6 +1408,87 @@ TEST(nntrainer_Tensor, print_large_size) {
   EXPECT_EQ(ss.str(), expected.str());
 }
 
+TEST(nntrainer_Tensor, DISABLED_private_external_loop_n) {
```
kparichay, line 1411: any plans to enable this test?

test/unittest/unittest_nntrainer_tensor.cpp
zhoonit: It was wrong to update this index out of bound #532. Wonder how it managed to pass all the tests so far.

test/unittest/unittest_nntrainer_tensor.cpp
zhoonit: I debugged with this test and left it to demonstrate the usage. So, no plans to enable this test. Do you think we should delete this?

nntrainer/src/tensor.cpp
zhoonit: This `if` statement checks if the given shape is broadcastable. I could set `0` as the default stride. but IMO, setting strides here shows the mechanism better in that if TensorDim is equal to 1, it sets stride to 0.

nntrainer/src/tensor.cpp
zhoonit: Done

test/unittest/unittest_nntrainer_tensor.cpp
zhoonit: Changed. Thanks

nntrainer/src/tensor.cpp
zhoonit: Nice catch. 
I added test case for dim = (3, 5, 1, 4) and m_dim = (1, 5, 1, 4) on which buffer size should be 20.

nntrainer/src/tensor.cpp
zhoonit: I agree that the implementaion could be improved. Could you explain a bit more in person?

test/unittest/unittest_nntrainer_tensor.cpp
zhoonit: Enabled.

nntrainer/src/tensor.cpp
kparichay: `If TensorDim == 1, stride is used as default initialized value of 0`.
In c++, using struct/class, its best to have constructor initializing its members

nntrainer/src/tensor.cpp
kparichay: As discussed offline, lets update to a more manageable implementation.

zhoonit:
`Add_i` will be reimplemented in #534 . I'll squash this PR to #534.

PR #529, opened 2020-09-11
[bn] bug fix for setting optimizer for bn layer

### LINKED ISSUE #528: bn layer sets optimizer for non-trainable vairables
This patch provides bugfix for setting the optimizer for bn layer
although the weight updates are called only for the trainable params,
the optimizer is initialized with all the weights which are not trainable
and have no gradients
This results in unnecessary memory allocation for them as well as
creates issues for #517

kparichay:
This patch provides bugfix for setting the optiimzer for bn layer
although the weight updates are called only for the trainable params,
the optimizer is initialized with all the weights which are not trainable
and have no gradients
This results in un-necessary memory allocation for them as well as
creates issues for #517

This patch adds a count for trainable_param_size and a getTrainableParams() for the layer
and instead of all params, only trainable params of the layer are passed to the optimizer

An alternate was for the optimizer to check gradient size and ignore the ones with 0 size
but thats over-engineering at this point.

Resolves #528 

**Self evaluation:**

cc. @zhoonit

### DIFF
--- a/nntrainer/include/optimizer.h
+++ b/nntrainer/include/optimizer.h
@@ -33,9 +33,10 @@ namespace nntrainer {
  * @brief UpdatableParam that could be updated thorugh optimizer
  */
 struct UpdatableParam {
-  Tensor weight;    /**<  weight to be updated and used */
-  Tensor grad;      /**<  gradient for the weight */
-  std::string name; /**< name of the parameter */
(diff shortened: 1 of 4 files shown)


zhoonit (approved):
LGTM :)

jijoongmoon (approved):

PR #533, opened 2020-09-11
[dataset] Support setting user_data for generator dataset

kparichay:
Support setting user_data for the dataset using generators
For now, there is only 1 user_data for all 3 train/valid/test dataset
The format for setting user_data is
"ml_train_dataset_set_property(dataset, "user_data", (void *) data, NULL)"

It is to be passed two consecutive arguments, and with list being terminated by NULL

**Self evaluation:**


kparichay:
Note that user_data is initialized as NULL, so no need to set user_data to null.
Calling "ml_train_dataset_set_property(dataset, "user_data", NULL, NULL)" will result in error.

zhoonit (approved):
LGTM

Issue #539, opened 2020-09-14
app_DeepQ fail with current upstream/master


zhoonit:
ninja test
[95/95] Running all tests.
 1/18 app_knn                        OK             3.45s
 2/18 app_logistic                   OK             1.15s
 3/18 app_DeepQ                      FAIL           0.19s (killed by signal 6 SIGABRT)
 4/18 app_classification             OK             1.65s
 5/18 app_classification_func        OK             21.13s
 6/18 app_training                   OK             6.60s
 7/18 unittest_tizen_capi            OK             9.58s
 8/18 unittest_tizen_capi_layer      OK             0.16s
 9/18 unittest_tizen_capi_optimizer  OK             0.10s
10/18 unittest_tizen_capi_dataset    OK             0.15s
11/18 unittest_nntrainer_activations OK             0.08s
12/18 unittest_nntrainer_internal    OK             0.39s
13/18 unittest_nntrainer_layers      OK             0.30s
14/18 unittest_nntrainer_lazy_tensor OK             0.07s
15/18 unittest_nntrainer_tensor      OK             0.05s
16/18 unittest_util_func             OK             0.07s
17/18 unittest_databuffer_file       OK             0.09s
18/18 unittest_nntrainer_modelfile   OK             4.05s

Ok:                 17  
Expected Fail:      0   
Fail:               1   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

The output from the failed tests:

 3/18 app_DeepQ                      FAIL           0.19s (killed by signal 6 SIGABRT)

--- command ---
08:12:51 /data/nntrainer/build/Applications/ReinforcementLearning/DeepQ/jni/nntrainer_deepq /data/nntrainer/Applications/ReinforcementLearning/DeepQ/jni/DeepQ.ini
--- stderr ---
terminate called after throwing an instance of 'std::invalid_argument'
  what():  [TensorDim] Trying to assign value <=0 to tensor dim
-------

jijoongmoon:
closed by #543

PR #543, opened 2020-09-15
[ Bug ] Fix Tensor Copy

jijoongmoon:
we only copy if tensor len == 0

related issues : #539

**Self evaluation:**

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -72,8 +72,9 @@ static auto rng = [] {
 }();
 
 Tensor::Tensor(const TensorDim &d, const float *buf) : Tensor() {
-  dim = d;
+
   if (d.getDataLen() != 0) {
+    dim = d;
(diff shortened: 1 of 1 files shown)


nntrainer/src/tensor_dim.cpp
```diff
@@ -57,9 +57,9 @@ void TensorDim::setTensorDim(unsigned int idx, unsigned int value) {
   if (idx > MAXDIM)
     throw std::out_of_range(
       "[TensorDim] Tensor Dimension index should be between 0 and 4");
-  if (value <= 0)
```
kparichay, line 60: where do we need the case to remove this.
Current semantics is "allow initialization of TensorDim to 0" but "setting TensorDim to 0 or copying a Tensor/TensorDim with Dim 0 is not allowed"

kparichay (approved):
LGTM now

zhoonit (approved):
LGTM

PR #547, opened 2020-09-15
[ Bug ] Free content of g_file_get_contents

jijoongmoon:
Add free when return

**Self evaluation:**

### DIFF
--- a/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
+++ b/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
@@ -186,6 +186,7 @@ int NNTrainer::loadModel() {
   } catch (...) {
     ml_loge("Cannot load model from config\n");
     model->finalize();
+    g_free(content);
     return -1;
   }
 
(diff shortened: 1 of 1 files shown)


nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
```diff
@@ -176,6 +176,7 @@ int NNTrainer::loadModel() {
 
   if (!g_file_get_contents(model_config, &content, &file_size, &file_error)) {
     ml_loge("Error reading model config!! - %s", file_error->message);
+    g_free(content);
```
kparichay, line 179: this is not needed.
As the [documentation](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-file-get-contents) says, if error, content will be set to NULL.

kparichay (approved):
LGTM

zhoonit (approved):
LGTM

nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
jijoongmoon: > this is not needed but will work with it too as `g_free()` accepts NULL.
> As the [documentation](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-file-get-contents) says, if error, content will be set to NULL.

Good. then, I will remove this g_free.

Issue #559, opened 2020-09-17
Distributive law breaks with fairly large error


zhoonit:
The example illustrated in https://github.com/nnstreamer/nntrainer/commit/cb0ffe082bf9f4ee72c0b53e61cc713c078704ea fail with fairly large error. For example, 

```
../test/unittest/unittest_nntrainer_tensor.cpp:2039: Failure
      Expected: data1[i]
      Which is: -54.629936
To be equal to: data2[i]
      Which is: -54.630127
```

PR #560, opened 2020-09-17
[Tensor] Fix bug in the broadcast support @open sesame 09/18 11:48

zhoonit:
Fix bug that strides and buffer axis are miscalculated

- Clarified consecutive-one strategy and same-stride(dimension matching) strategy
- Change the last stride to 0 only if it is using a consecutive-one strategy
- Add regression test

Resolves: #559

**Self evaluation:**

Cc: Jijoong Moon <jijoong.moon@samsung.com>

### DIFF
--- a/nntrainer/src/tensor.cpp
+++ b/nntrainer/src/tensor.cpp
@@ -243,6 +243,17 @@ int Tensor::operator_i(
   std::function<void(const BroadcastInfo &e, float *, const float *)> v_func) {
 
   BroadcastInfo e;
+
+  /// shortcut to cover when dimension matches
+  /// note that buffer_size, the last stride is only used in v_func but it might
+  /// be changed
+  if (dim == m.dim) {
+    e.buffer_size = length();
+    e.strides[3] = 1;
+    v_func(e, getData(), m.getData());
+    return ML_ERROR_NONE;
+  }
+
   try {
     e = this->computeBroadcastInfo(m);
   } catch (std::exception &err) {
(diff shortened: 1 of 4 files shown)


nntrainer/src/tensor.cpp
```diff
@@ -785,12 +785,12 @@ BroadcastInfo Tensor::computeBroadcastInfo(const Tensor &m) {
   BroadcastInfo e;
 
   for (unsigned int i = 0; i < MAXDIM; ++i) {
-    if (m_dim.getTensorDim(i) == 1) {
+    if (dim.getTensorDim(i) == m_dim.getTensorDim(i)) {
```
jijoongmoon, line 788: I think it would be better to leave comment if the order is matter.

jijoongmoon: I think we don't need to calculate the BroadcastInfo if both length is same to save calculation.

nntrainer/src/tensor.cpp
```diff
@@ -802,30 +802,41 @@ BroadcastInfo Tensor::computeBroadcastInfo(const Tensor &m) {
...
+    }
+
+    e.buffer_size *= dim.getTensorDim(axis);
+  }
+
+  /// check strategy that uses consecutive ones
+  if (m_dim.getTensorDim(3) == 1) {
+    unsigned int inner_loop_size = 1;
+    int axis;
+    for (axis = 3; axis >= 0; --axis) {
```
jijoongmoon, line 823: It can start from 2 I guess.

nntrainer/src/tensor.cpp
zhoonit: Both done 👍

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #561, opened 2020-09-17
Fix bn layer backwarding @open sesame 09/18 14:15

zhoonit:
- Change formula to calculate bn layer backward propagation
- Enable conv2d test case
- Add test case when batch is 1
- Renew test case to use non constant grad_ys

**Self evaluation:**


nntrainer/src/bn_layer.cpp
```diff
@@ -161,27 +160,30 @@ sharedConstTensor
...
 
-  int batch = input_dim.batch();
+  int N = 1;
+
+  for (auto &axis : axes_to_reduce) {
+    N *= input_dim.getTensorDim(axis);
+  }
 
-  dgamma = x_normalized.multiply(deriv).sum(axes_to_reduce);
+  ///@note faster implementation. Should work but it does not match with tf
```
zhoonit, line 175: remove this line. it matches now

nntrainer/include/tensor.h
```diff
@@ -302,6 +302,13 @@ class Tensor {
    */
   Tensor divide(Tensor const &m) const;
 
+  /**
+   * @brief    Tensor power Element by Element
+   * @param[in] float Divisor Tensor
+   * @retval Calculated Tensor
+   */
+  Tensor pow(float m) const;
```
jijoongmoon, line 310: does it give us better accuracy?

zhoonit: The validation result says it is equivalent to the very small tolerance. I made this for checking purposes but decided leave here as it is a basic operation.

nntrainer/src/tensor.cpp
```diff
@@ -238,6 +239,15 @@ int Tensor::subtract_i(float const &value) { return this->add_i(-value); }
 
 Tensor Tensor::subtract(float const &value) { return this->add(-value); }
 
+Tensor Tensor::pow(float exponent) const {
+  Tensor result(dim);
+
+  for (unsigned int i = 0; i < length(); ++i)
+    *result.getAddress(i) = powf(*getAddress(i), exponent);
```
kparichay, line 246: use std::transform than a loop here.
Also, i wonder about the accuracy of pow for the exponent of 2 vs using cblas multiply? Is using `powf` more accurate?

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -853,21 +853,57 @@ TEST_F(nntrainer_BatchNormalizationLayer, forward_backward_training_01_p) {
...
-  EXPECT_NO_THROW(backward_result = *layer.backwarding(
-                    MAKE_SHARED_TENSOR(constant(1.0, 3, 1, 1, 12)), 1));
+  nntrainer::Tensor backward_in(layer.getOutputDimension());
+  loadFile("tc_bn_fc_1_goldenBNLayerBackwardDxIn.out", backward_in);
+
+  nntrainer::Tensor backward_result =
+    *layer.backwarding(MAKE_SHARED_TENSOR(backward_in), 1);
 
-  matchOutput(backward_result, "tc_bn_fc_1_goldenBNLayerBackwardDx.out");
+  // matchOutput(backward_result, "tc_bn_fc_1_goldenBNLayerBackwardDx.out");
```
kparichay, line 864: intentional?

zhoonit: Oh no.. seems like it hasn't verified yet for conv

nntrainer/src/tensor.cpp
zhoonit: Accuracy is verified with (0.5f and `sqrtFloat`)
This is for negative and non-integer power.

Updated to use apply instead

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -853,12 +853,13 @@ TEST_F(nntrainer_BatchNormalizationLayer, forward_backward_training_01_p) {
...
 
   EXPECT_NO_THROW(forward_result = layer.forwarding(MAKE_SHARED_TENSOR(in)));
-
   matchOutput(*forward_result, "tc_bn_fc_1_goldenBNResultForward.out");
 
-  nntrainer::Tensor backward_result;
-  EXPECT_NO_THROW(backward_result = *layer.backwarding(
-                    MAKE_SHARED_TENSOR(constant(1.0, 3, 1, 1, 12)), 1));
+  nntrainer::Tensor backward_in(layer.getOutputDimension());
+  loadFile("tc_bn_fc_1_goldenBNLayerBackwardDxIn.out", backward_in);
```
kparichay, line 859: this test the backwardDx. `AfterUpdate` will be in next unittest?

zhoonit: Yes. Do you prefer it to be deleted now?

kparichay: No its fine. I just wanted to know.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #581, opened 2020-09-21
[layers] Move constructors must be noexcept

kparichay:
Make all move constructors to be noexcept
Without noexcept, compiler many times does not use move constructor but rather uses copy constructor

**Self evaluation:**


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

Issue #199, opened 2020-06-18
Provides C++ API


jijoongmoon:
We have to provides C++ API

myungjoo:
For C++ APIs, you need to communicate with SNAP people.

kparichay:
- [ ] Cleanup existing headers (to allow to expose these directly)
- [ ] Align the API with SNAP people
- [ ] Move various functions to protected and declare classes as friends - then the classes can be directly exposed

kparichay:
@jijoongmoon Close this issue?

PR #584, opened 2020-09-22
[refactor]  Cleanup headers

kparichay:
This patch cleans up the headers of the internal codes so that the headers can directly be exposed for the c++ API
Also removed applyIf from lazy_tensor (after checking with @zhoonit) as it exposed #define in header and was not used
Update enums to enum class before headers are exposed

Related Issue: #199 

**Self evaluation:**

Signed-off-by: Parichay Kapoor


nntrainer/include/layer.h
```diff
@@ -340,26 +297,16 @@ class Layer {
    * @brief     set weight decay parameters
    * @param[in] w struct for weight decay
    */
-  void setWeightRegularizer(WeightRegularizerParam w) {
-    weight_regularizer = w;
-  }
+  // void setWeightRegularizer(WeightRegularizerParam w) {
```
zhoonit, line 300: Any reason for this?

nntrainer/include/tensor.h
```diff
@@ -580,6 +579,7 @@ class Tensor {
     return strides;
   }
 
+  // TODO: move to private
```
zhoonit, line 582: There is a test with this function. I guess the test can be removed because it is tested through `opeator_i`

nntrainer/src/blas_interface.cpp
```diff
@@ -13,6 +13,9 @@
 
 #include <blas_interface.h>
 
+#include <cmath>
+#include <stdexcept>
```
zhoonit, line 17: include `nntrainer_error.h` instead?

nntrainer/src/bn_layer.cpp
```diff
@@ -53,17 +53,17 @@ int BatchNormalizationLayer::initialize() {
   }
 
   Tensor mu =
-    initializeWeight(dim, initializers[static_cast<int>(BNParams::mu)], status);
-  Tensor var = initializeWeight(
-    dim, initializers[static_cast<int>(BNParams::var)], status);
+    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::mu)]);
+  Tensor var =
+    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::var)]);
   NN_RETURN_STATUS();
```
zhoonit, line 59: NN_RETURN_STATUS could be deleted (same for 61~66)

nntrainer/src/fc_layer.cpp
```diff
@@ -45,10 +45,10 @@ int FullyConnectedLayer::initialize() {
   TensorDim dim = output_dim;
   dim.height(input_dim.width());
   dim.batch(1);
-  Tensor weight = initializeWeight(dim, weight_initializer, status);
+  Tensor weight = getInitializedTensor(dim, weight_initializer);
   NN_RETURN_STATUS();
```
zhoonit, line 49: ditto (and for 52)

nntrainer/include/layer.h
kparichay: my bad, un-commented.

nntrainer/include/tensor.h
kparichay: I am currently disabling the test than removing it. We can think about removing the test or making the test class a friend class for testing purposes wrapped with `#if #endif`.

kparichay: Ohh that test is already DISABLED and its not an-automated test. Removing it.

nntrainer/src/blas_interface.cpp
kparichay: done.

nntrainer/src/bn_layer.cpp
kparichay: done.

nntrainer/src/fc_layer.cpp
kparichay: done.

zhoonit (approved):
Grete work 👍

jijoongmoon (approved):
LGTM. rebase please.

PR #585, opened 2020-09-22
[weight] weight class to simplify headers and enums

kparichay:
Added a weight class to simplify headers All weight-related enums and properties go to the weight header rather being dumped in layer or optimizer.

**Self evaluation:**

Signed-off-by: Parichay Kapoor

### DIFF
--- a/nntrainer/src/bn_layer.cpp
+++ b/nntrainer/src/bn_layer.cpp
@@ -52,22 +52,18 @@ int BatchNormalizationLayer::initialize() {
       axes_to_reduce.push_back(i);
   }
 
-  Tensor mu =
-    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::mu)]);
-  Tensor var =
-    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::var)]);
-
-  Tensor gamma =
-    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::gamma)]);
-  Tensor beta =
-    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::beta)]);
-
-  setParamSize(4);
-  paramsAt(0) = {std::move(mu), Tensor(), "BN:moving_mean", false};
+  setNumWeights(4);
+  weightAt(0) =
+    std::move(Weight(dim, initializers[static_cast<int>(BNParams::mu)], false,
+                     "BN:moving_mean"));
   ///@todo shift var to std to save computation
-  paramsAt(1) = {std::move(var), Tensor(), "BN:moving_variance", false};
-  paramsAt(2) = {std::move(gamma), Tensor(gamma.getDim()), "BN:gamma"};
-  paramsAt(3) = {std::move(beta), Tensor(beta.getDim()), "BN:beta"};
+  weightAt(1) =
+    std::move(Weight(dim, initializers[static_cast<int>(BNParams::var)], false,
+                     "BN:moving_variance"));
+  weightAt(2) = std::move(Weight(
+    dim, initializers[static_cast<int>(BNParams::gamma)], true, "BN:gamma"));
+  weightAt(3) = std::move(Weight(
+    dim, initializers[static_cast<int>(BNParams::beta)], true, "BN:beta"));
 
   return status;
 }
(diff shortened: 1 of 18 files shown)


nntrainer/src/bn_layer.cpp
```diff
@@ -52,22 +52,18 @@ int BatchNormalizationLayer::initialize() {
...
-    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::gamma)]);
-  Tensor beta =
-    getInitializedTensor(dim, initializers[static_cast<int>(BNParams::beta)]);
-
-  setParamSize(4);
-  paramsAt(0) = {std::move(mu), Tensor(), "BN:moving_mean", false};
+  setNumWeights(4);
+  weightAt(0) =
+    std::move(Weight(dim, initializers[static_cast<int>(BNParams::mu)], false,
+                     "BC:moving_mean"));
```
zhoonit, line 58: s/BC/BN ?

nntrainer/src/weight.cpp
```diff
@@ -0,0 +1,90 @@
...
+    break;
+  }
+}
+
+void Weight::swap(Weight &lhs, Weight &rhs) noexcept {
+  std::swap(lhs.var, rhs.var);
+  std::swap(lhs.initializer, rhs.initializer);
+  std::swap(lhs.trainable, rhs.trainable);
+  if (lhs.trainable)
+    std::swap(lhs.grad, rhs.grad);
```
zhoonit, line 75: I think this should be swapped to get predictable result.

Whatif rhs.trainable == true?

jijoongmoon (approved):
LGTM, then we do not need UpdatableParam any more.

nntrainer/src/weight.cpp
```diff
@@ -0,0 +1,90 @@
...
+    var.setRandUniform(-1.0f * sqrtFloat(6.0f / (dim.height())),
+                       sqrtFloat(6.0 / (dim.height())));
+    break;
+  default:
+    break;
+  }
+}
+
+void Weight::swap(Weight &lhs, Weight &rhs) noexcept {
+  std::swap(lhs.var, rhs.var);
```
zhoonit, line 71: Tensor has it's on swap.

I think it's better to depend on ADL or specialize std::swap

https://stackoverflow.com/questions/11562/how-to-overload-stdswap/2684544#2684544

nntrainer/src/weight.cpp
```diff
@@ -0,0 +1,90 @@
...
+  std::swap(lhs.initializer, rhs.initializer);
+  std::swap(lhs.trainable, rhs.trainable);
+  if (lhs.trainable)
+    std::swap(lhs.grad, rhs.grad);
+  std::swap(lhs.name, rhs.name);
+}
+
+Weight &Weight::operator=(const Weight &rhs) {
+  Weight tmp(rhs.var.getDim(), rhs.initializer, rhs.trainable, rhs.name);
```
zhoonit, line 80: 1. I guess we should have copy/move ctor to keep [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three)

2. is var and grad copied here?

nntrainer/src/weight.cpp
```diff
@@ -0,0 +1,90 @@
...
+  this->swap(*this, tmp);
+  return *this;
+}
+
+Weight &Weight::operator=(Weight &&rhs) noexcept {
```
zhoonit, line 85: ditto

nntrainer/src/weight.cpp
kparichay: so, if `rhs.trainable == false`, `rhs.grad` is a empty tensor. And swapping empty tensor gives error (copy a tensor with zero dimension). Hence the check.

nntrainer/src/weight.cpp
kparichay: I will add copy/move ctor.
As swap is done,  just ptr are exchanged, no deep copy.

nntrainer/src/weight.cpp
kparichay: ditto :)

nntrainer/src/weight.cpp
zhoonit: We discussed offline to fix Tensor::swap to friend void

nntrainer/src/weight.cpp
zhoonit: Yes. that makes `if (lhs.trainable and rhs.traninable)`. But this breaks the semantics of swap.
If copying an empty tensor gives an error, I think we need a fix inside tensor.

kparichay: If a tensor is not valid (empty == dimension of 0), then we shouldn't be copying it in the first place, as copy requires setting the dimension value to 0. Allowing this would allow copying an uninitialized tensor to an initialized tensor, which results in using an uninitialized tensor. Copying an empty tensor is made to give an error to stop allow unintended use of uninitialized tensor.

nntrainer/src/bn_layer.cpp
kparichay: done.

nntrainer/src/weight.cpp
kparichay: Updated this swap to ADL, and updated tensor, tensor_dim to also be a friend.

kparichay:
V2:
Update swap to swap gradients even if not trainable
Update swap of Tensor, TensorDim to be friend functions
Copy and move assignements are now default

zhoonit (approved):
LGTM

PR #588, opened 2020-09-23
[inference] Inference bug fix @open sesame 09/23 15:03

### LINKED ISSUE #587: Inference does not act on last layer
If last layer is activaiton, then it is sometimes merged with loss layer.
In this case, this layer is not run with inference of the model.

kparichay:
Added inference bug fix to run the last activation layer when it has been merged in the loss layer This is done by supporting forwarding(in) method for loss which was earlier not supported We just have to be careful for this not be called in any other case other than inference

Later this will be changed to handled with train/eval mode (which I think @zhoonit will add for bn layer)

Resolves #587

**Self evaluation:**


jijoongmoon (approved):
LGTM!

PR #591, opened 2020-09-23
[dist/tizen] Package tensor filter properly

kparichay:
NNtrainer tensor filter was packaged with nntrainer which isnt right
Now nntrainer tensor filter is packaged separately and depends on nnstreamer and nntrainer both

**Self evaluation:**


jijoongmoon (approved):
LGTM

PR #600, opened 2020-09-24
[batchsize] Change the semantics of batch size @open sesame 09/24 16:27

kparichay:
This patch changes the semantics of the way batchsize is used in the library
1. batch_size is no longer property of the layer. It can still be set externally by the model. The method to set is still public but will soon be changed to private.
2. batch_size of the input/label/derivative tensor provided to the forwarding/backwarding function can no longer be arbitrary. It must match the batch_size set to the model and the layer. This change in semantics to follow for a long-term design where memory for input/output is pre-allocated.
3. batch_size can now be set at train time than the earlier design where the batch_size had to be set at init time. This comes from the design change that the memory for the model weights can be allocated init time, which is not dependent on the batch size. However, the memory for input/output should be allocated at train/inference as the batch size can be different for these. In the current design, memory is allocated every iteration. Later, when memory is allocated at once and reused, change in batch size will change the memory at once at the first iteration (rather doing this at init times). This change is necessary to allow running inference without the need to initialize the model again.

**Self evaluation:**


nntrainer/src/addition_layer.cpp
```diff
@@ -38,8 +38,7 @@ int AdditionLayer::initialize() {
 }
 
 sharedConstTensor AdditionLayer::forwarding(sharedConstTensor in) {
-  hidden = Tensor(in->batch(), output_dim.channel(), output_dim.height(),
-                  output_dim.width());
+  hidden = Tensor(input_dim);
```
jijoongmoon, line 41: is it right? for the forwarding during inference? Cause inference input batch_size is not batch_size as in configuration file. And is it same dim  input and hidden dimension for addition layer?

zhoonit:
So, batch_size should be moved to run(...) officially.
I am good with the semantic change. We should keep in mind that this kinda affects (exposed) API.

Inside tizen docs
```
### Compile the model

Compiling a model finalizes the model with loss.
Once compiled, any modification to the properties of the model is restricted.
Adding layers or changing the optimizer or dataset of the model is not permitted either:

```c
int status = ML_ERROR_NONE;
status = ml_train_model_compile(model, "loss=cross", "batch_size=9", NULL);
```

## Train the model

Now, the model is ready to train. Train model as follows:

```c
int status = ML_ERROR_NONE;
status = ml_train_model_run(model, "epochs=20", "save_path=model.bin", NULL);
```
```

File: nntrainer/src/neuralnet.cpp [hunk 2]
```diff
@@ -172,6 +164,14 @@ int NeuralNetwork::setTrainConfig(std::vector<std::string> values) {
       continue_train = cont_train;
       opt.setProperty({values[i]});
     } break;
+    case PropertyType::batch_size: {
```
zhoonit, line 167: How about neuralnet->setBatchSize() setting batch of layers and databuffer(if any) as well?

zhoonit: Bigger suggestion:
How about s/`setTrainConfig`/`setRunConfig?Option?`  (or equivalent) and integrate inference option as well?

So that we can align with ml_train_model_run(...) api and add some run configs as well.
e.g.) 
"run mode=train/eval"
"metrics=accuracy"...?

nntrainer/src/addition_layer.cpp
kparichay: For forwarding inference, batch_size is reset with the inference method before calling forwarding on the layers (line 380 in neuralnetwork.cpp).
The updation of batch_size in the inference method updated the input dimension and hidden dimension (output dimension) both.

kparichay:
> So, batch_size should be moved to run(...) officially.
> I am good with the semantic change. We should keep in mind that this kinda affects (exposed) API.

The API still is still valid. The user can set the batch_size at `compile`. This patch just allows setting batch_size even at `run` method. The description of the API doesn't talk about the specific properties which can be set at compile or run. So, I think we need to add them to the documentation at once?

nntrainer/src/neuralnet.cpp
kparichay: Updated to add neuralnetwork->setBatchSize()
For the bigger suggestion, opened #606.

nntrainer/src/addition_layer.cpp
kparichay: > And is it same dim input and hidden dimension for addition layer?

As far as I know, most of the layers don't change batch size. If they do in future , then their `setBatchSize()` will have be overloaded to handle that.

jijoongmoon: I recommend to run Applications and check accuracy as well.

nntrainer/src/neuralnet.cpp
```diff
@@ -395,10 +411,13 @@ int NeuralNetwork::train(std::vector<std::string> values) {
     return ML_ERROR_INVALID_PARAMETER;
   }
 
-  /** Setup data buffer properties */
-  status = data_buffer->setBatchSize(batch_size);
+  status = setTrainConfig(values);
   NN_RETURN_STATUS();
 
+  /** set batch size just before training */
+  setBatchSize(batch_size);
```
jijoongmoon, line 418: Is it ok setBatchSize after the initialization?

zhoonit (approved):
LGTM after checking with the applications :)

nntrainer/src/neuralnet.cpp
kparichay: Yes, initialization only sets the layer-specific information, and this not dependent on the batch size (weights, bias, other parameters). This just adds a constraint that the memory for the inputs/outputs must be allocated at the `train/inference/run` methods than at `init` which I think is even better.

jijoongmoon:
> > So, batch_size should be moved to run(...) officially.
> > I am good with the semantic change. We should keep in mind that this kinda affects (exposed) API.
> 
> The API still is still valid. The user can set the batch_size at `compile`. This patch just allows setting batch_size even at `run` method. The description of the API doesn't talk about the specific properties which can be set at compile or run. So, I think we need to add them to the documentation at once?

I think not this release. We can do it next release. we are managing separate branch, so, it's ok.

kparichay:
@jijoongmoon Verified that all the applications run, and verified accuracy remains same with and without this patch.

kparichay:
V2:
Rebased
Updated validation to run on a whole batch of data at once
Also updated Tensor.argmax() to perform on batch of data than on the whole data

zhoonit:
LGTM for v2

jijoongmoon (approved):
LGTM

Issue #597, opened 2020-09-24
batch normalization not ready for inference


kparichay:
for batch normalization layer, in the inference/validation/test should set the trainable for the layer or the dependent parameter layer uses for checking if in train/eval mode should be flipped and flipped back in training.

zhoonit:
JJ and I have discussed about this matter. We are gonna have a member is_training for each layer. How's this working for you?

kparichay:
let's create a `ModelAttr` which contains certain attributes about the model, and the reference to this can be shared with all the layers, than all the layers having new members for such things.

zhoonit:
> let's create a `ModelAttr` which contains certain attributes about the model, and the reference to this can be shared with all the layers, than all the layers having new members for such things.

Good idea

PR #601, opened 2020-09-24
[inference] Remove deciding for inference at model construction

kparichay:
Currently, with is_train, user has to decide at model construction if the model is to be used for inference or train but we still allow changing it. We dont need that difference, rather using inference/train should be allowed at anytime by just changing the batch-size (this has been allowed with the previous commit)

Now, once the train is done, with model loaded in memory, inference can be called directly. Inference can be run by creating and loading the model from disk.

**Self evaluation:**

### DIFF
--- a/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
+++ b/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
@@ -151,7 +151,7 @@ void NNTrainer::loadModel() {
   if (model_config == nullptr)
     throw std::invalid_argument("model config is null!");
 
-  model = new nntrainer::NeuralNetwork(false);
+  model = new nntrainer::NeuralNetwork();
   model->loadFromConfig(model_config);
 #if (DBG)
   gint64 stop_time = g_get_real_time();
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -74,8 +74,7 @@ public:
   /**
    * @brief     Constructor of NeuralNetwork Class
    */
-  NeuralNetwork(bool istrain = true) :
-    is_train(istrain),
+  NeuralNetwork() :
     batch_size(1),
     epochs(1),
     loss(0.0f),
@@ -220,18 +219,6 @@ public:
   int setDataBuffer(std::shared_ptr<DataBuffer> data_buffer);
 
   /**
-   * @brief     check train or inference. Default is True (train)
-   * @retval bool True : training Mode. False : Inference Mode
-   */
-  bool isTrain() { return is_train; };
-
-  /**
-   * @param[in] train true for train mode. false for inference mode
-   * @brief     set Running Mode : Train or Inference
-   */
-  void setMode(bool train) { is_train = train; };
-
-  /**
    * @brief     add layer into neural network model
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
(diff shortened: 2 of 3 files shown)


zhoonit (commented):
I agree we do not need to set `is_train` at the construction phase.
I think It is good to have a state that represents the current mode of NeuralNetwork though (whether it is set to be training or eval). related to the suggestion. https://github.com/nnstreamer/nntrainer/pull/600/files#r494142824

kparichay:
> I agree we do not need to set `is_train` at the construction phase.
> I think It is good to have a state that represents the current mode of NeuralNetwork though (whether it is set to be training or eval). related to the suggestion. https://github.com/nnstreamer/nntrainer/pull/600/files#r494142824

Yes the state is good, but it is only a temporary state, which should not be associated with the model object. At any given time, the model object can be in running phase, but be it inference or training should;t matter. The mode/state which decide which internal function to use but can you please give why that state should be associated the model object.

jijoongmoon:
I recommend to run Applications and check if there is no issues.

zhoonit:
> Yes the state is good, but it is only a temporary state, which should not be associated with the model object. At any given time, the model object can be in running phase, but be it inference or training should;t matter. The mode/state which decide which internal function to use but can you please give why that state should be associated the model object.

As discussed in #597, Neuralnet needs to tell the layers if it's inference or training.

One can be done by using the state of the model and share the state with each node. 🍰 
or it can be done by differentiating forward function 🌵.
eg)
ops.forward(bool is_train = true)
or 
ops.forward_train()
ops.forward_infer()

🌵 can be done though, JJ and I have discussed to go for the 🍰 to not change the ops interface.

Keras internally works like :cake:. keras requires to pass {tf.keras.backend.learning_phase(): true} in the feeddict to control the run mode.

kparichay:
@zhoonit Yes, share the running mode with the layers is a good way and must be done. However, it still does not need to be associated with object of the model, as once the train/inference method completes, that state is invalid.

Lets share this state with #597.

zhoonit (approved):
LGTM

jijoongmoon:
I think this is ok. But I have one thing to mention is that we might have to inference with different input like tfmodel or NNFW input. In that case, we have to configure differently and this is_train might be required.

jijoongmoon (approved):
LGTM after rebase. But I have one thing to mention is that we might have to inference with different input like tfmodel or NNFW input. In that case, we have to configure differently and this is_train might be required.

kparichay:
Rebased.
@jijoongmoon Even if we have NNFW or tflite input to perform inference with, the inference mode is valid only while that inference call is being performed, so IMO, `is_train` like variables won't be needed to be exposed outside. It might be possible to have such variables internally.

jijoongmoon:
> Rebased.
> @jijoongmoon Even if we have NNFW or tflite input to perform inference with, the inference mode is valid only while that inference call is being performed, so IMO, `is_train` like variables won't be needed to be exposed outside. It might be possible to have such variables internally.

Well, I don't agree with it, but let's keep this for now. We will discuss when it comes to implement.

kparichay:
> Well, I don't agree with you. IMO, the reason we don't need to have is_train is that we configure model same way in both train and inference cases. But if we are going to use NNFW or tflite the inference can be performed by NNFW and tflite but train must be conduct by nntrainer and that means we need to distinguish model configuration. but let's keep this for now. We will discuss when it comes to implement. rebase please.

I am not sure if this will still be needed.
Yes, we can determine later when time comes.

PR #603, opened 2020-09-24
[model] Merge model finalize to destructor

kparichay:
Merge model finalize to destructor itself
User need not call finalize explicitly

Other minor updates:
- Merge all train methods into 1
- isInitializable() is made private. Removed some of the name checking from this function
to when the layer itself was created.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -256,7 +256,7 @@ int NeuralNetwork::init() {
 /**
  * @brief     free layers
  */
-void NeuralNetwork::finalize() {
+NeuralNetwork::~NeuralNetwork() {
   layers.erase(layers.begin(), layers.end());
 
   if (data_buffer) {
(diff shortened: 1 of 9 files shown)


nntrainer/src/neuralnet.cpp
```diff
@@ -542,39 +536,34 @@ int NeuralNetwork::isInitializable() {
...
-    return ML_ERROR_INVALID_PARAMETER;
-
   if (initialized) {
     return ML_ERROR_NOT_SUPPORTED;
   }
 
   ensureName(layer);
+
+  /** Validate the layer to be added */
+  status = layer->checkValidation();
```
zhoonit, line 552: If we validate and add layer_names only inside `addLayer`, it should be the single source of writing the layer. I think it'd be best if we could restrict the write access for the `layers`.

For example, `layers.insert` is being used for some methods.

kparichay: Yes, insertion of layers would be inserted once we have graph object. That will have a single interface to add layer via the graph object. 
In the current code, as a hot fix, each `layer.insert()` is attached with `ensureName(layer)` for this purpose. Graph will solve all these issues. (I can update `addLayer` to include a position and then `addLayer` would insert at that position than at the end, but anyway graph will be coming next month, so skipping it.)

zhoonit: Sounds fare enough

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #605, opened 2020-09-24
Add print preset

zhoonit:
- Add print preset for model and layer
- Add model print flags
- move printFlagEnums and print(out, flag) to private

**Self evaluation:**

Cc: Parichay Kapoor <pk.kapoor@samsung.com>

### DIFF
--- a/nntrainer/include/layer.h
+++ b/nntrainer/include/layer.h
@@ -120,26 +120,19 @@ typedef enum {
 } WeightInitializer;
 
 /**
- * @brief   Print Options when printing layer info
- */
-typedef enum {
-  // clang-format off
-  PRINT_INST_INFO  = (1 << 0), /**< Option to print type & instance address info */
-  PRINT_SHAPE_INFO = (1 << 1), /**< Option to print shape information, invalid before initiation*/
-  PRINT_PROP       = (1 << 2), /**< Option to print properties */
-  PRINT_PROP_META  = (1 << 3), /**< Option to print properties that describe meta info
-                                    e.g) layer activation type for non-activation layer. */
-  PRINT_WEIGHTS    = (1 << 4), /**< Option to print weights */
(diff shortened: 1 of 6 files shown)


nntrainer/include/layer.h
```diff
@@ -119,27 +119,20 @@ typedef enum {
...
-  // clang-format on
-} LayerPrintOption;
-
 /**
  * @class   Layer Base class for layers
  * @brief   Base class for all layers
  */
 class Layer {
 public:
+  enum class PrintPreset {
```
kparichay, line 128: I think this enum might only need PRINT_SUMMARY*? Differentiation between PRINT_SUMMARY* and PRINT_ALL could be better done with `ML_TRAIN_SUMMARY_LAYER` and `ML_TRAIN_SUMMARY_TENSOR`.

zhoonit: >Differentiation between PRINT_SUMMARY* and PRINT_ALL could be better done with ML_TRAIN_SUMMARY_LAYER and ML_TRAIN_SUMMARY_TENSOR.
 
This can be done. However, I also need differentiation between PRINT_SUMMARY, PRINT_SUMMARY_META which is depending on the `initialized` member.

Similar things can happen for more delicate presets. So I think it is better that the model chooses what to print depending on its state since the layer does not know about the model's state or more, it can be printed without a model owns the layer.

Please let me know in detail if you have a better design. I might have missed some point

kparichay: Yes, model should help decide the summary level for layer (differentiate between SUMMARY and SUMMARY_META). But the decision of TENSOR should not be taken by model, but rather than trickle down to layer.

jijoongmoon (approved):
LGTM. I think we can go with for now and modify later.

kparichay (approved):
Its really minor things
its good for now

zhoonit:
> Its really minor things
> its good for now

I think several decisions should be made regarding the topic. I'll leave an issue to deal with it later.

PR #609, opened 2020-09-25
[application] Add output verification with gtest

kparichay:
Commit- 1: 
Add output verification with gtest for transfer learning application

Commit- 2: 
Add output loss verification with gtest for MNIST application for 5 epochs

**Self evaluation:**

### DIFF
--- a/Applications/TransferLearning/Draw_Classification/jni/meson.build
+++ b/Applications/TransferLearning/Draw_Classification/jni/meson.build
@@ -7,6 +7,9 @@ gst_dep = dependency('gstreamer-'+gst_api_version)
 nnstreamer_capi_dep = dependency('', required: false)
 if build_platform == 'tizen'
   nnstreamer_capi_dep = dependency('capi-nnstreamer')
+  if not gtest_dep.found()
+    error('Gtest dependency not found for tizen')
+  endif
 endif
 
 training_sources = [
@@ -15,7 +18,7 @@ training_sources = [
 
 e = executable('nntrainer_training',
   training_sources,
-  dependencies: [app_utils_dep, iniparser_dep, nntrainer_capi_dep, nnstreamer_capi_dep, tflite_dep],
+  dependencies: [app_utils_dep, iniparser_dep, nntrainer_capi_dep, nnstreamer_capi_dep, tflite_dep, gtest_dep],
   include_directories: include_directories('.'),
   install: get_option('install-app'),
   install_dir: application_install_dir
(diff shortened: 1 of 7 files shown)


Applications/TransferLearning/Draw_Classification/jni/meson.build
```diff
@@ -15,7 +18,7 @@ training_sources = [
 
 e = executable('nntrainer_training',
   training_sources,
-  dependencies: [app_utils_dep, iniparser_dep, nntrainer_capi_dep, nnstreamer_capi_dep, tflite_dep],
+  dependencies: [app_utils_dep, iniparser_dep, nntrainer_capi_dep, nnstreamer_capi_dep, tflite_dep, gtest_dep],
```
jijoongmoon, line 21: I think this should be separated according to get_option("enable-test")

kparichay: then how about we add -DENABLE_TEST as well. So that applications will only run test with `enable-test`?

zhoonit (commented):
any reason to use gtest though? Can we put the checks at the end of the main and return non-negative instead?

So that the application can validate itself?

kparichay:
> any reason to use gtest though? Can we put the checks at the end of the main and return non-negative instead?
> 
> So that the application can validate itself?

Yes, for application to validate itself.
Regular check would fail the application with just giving out specific information. Gtest gives out mismatch information and some good test name too. Gtest is already a dependency with enable-test, so added.

Applications/TransferLearning/Draw_Classification/jni/meson.build
jijoongmoon: > then how about we add -DENABLE_TEST as well. So that applications will only run test with `enable-test`?

Sounds good to me.!

packaging/nntrainer.spec
```diff
@@ -213,8 +213,15 @@ bash %{test_script} ./test
...
 export NNSTREAMER_FILTERS=$(pwd)/build/nnstreamer/tensor_filter
 pushd build
+rm -rf model.bin
 TF_APP=Applications/TransferLearning/Draw_Classification
 ./${TF_APP}/jni/nntrainer_training ../${TF_APP}/res/Training.ini ../${TF_APP}/res
+
+rm -rf model.bin
+cp ../Applications/MNIST/jni/mnist_trainingSet.dat .
+MNIST_APP=Applications/MNIST
+./${MNIST_APP}/jni/nntrainer_mnist ../${MNIST_APP}/res/mnist.ini
```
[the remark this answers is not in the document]
kparichay: this doesnt directly work with run_test. I will have to check inside it and maybe make some updates to it.
I guess this will come with allowing app-test with ubuntu.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #613, opened 2020-09-28
[Filter] Replace g_assert

zhoonit:
g_assert ends the program rather than give chance for calls to handle
the error code. This patch replaces g_assert with return value

v2. Remove assertion codes and unify codes to throw

- Remove assertion codes instead throw to the top level and catch
altogether
- Merge `NNTrainer::init()` logic to constructor

**Self evaluation:**

### DIFF
--- a/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
+++ b/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
@@ -17,8 +17,9 @@
  */
 
 #include <algorithm>
-#include <map>
 #include <limits>
+#include <map>
+#include <sstream>
 #include <unistd.h>
 
 #include <nnstreamer_plugin_api.h>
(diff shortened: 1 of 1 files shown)


nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
```diff
@@ -166,7 +168,7 @@ int NNTrainer::loadModel() {
   gchar *content = nullptr;
   GError *file_error = nullptr;
 
-  g_assert(model_config != nullptr);
+  g_return_val_if_fail(model_config != nullptr, -1);
```
kparichay, line 171: How about removing all the asserts like one in line 83 and possibly others.
also, I think this check must be moved to line 86 and then this will become a throw.

myungjoo:
One thing that you should mind:

In shared library, it is not recommended to use assert. Try to write a full error messages and error handling routines.

nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
zhoonit: Yes, I'll make a V2 and upload

zhoonit:
v2. Remove assertion codes and unify codes to throw

- Remove assertion codes instead throw to the top level and catch
altogether
- Merge `NNTrainer::init()` logic to constructor

jijoongmoon (approved):
LGTM

nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
```diff
@@ -258,36 +230,32 @@ static void nntrainer_close(const GstTensorFilterProperties *prop,
...
 
     nntrainer_close(prop, private_data);
   }
 
-  nntrainer = new NNTrainer(model_file);
-  if (nntrainer == NULL) {
-    g_printerr("Failed to allocate memory for filter subplugin: nntrainer\n");
+  try {
+    nntrainer = new NNTrainer(model_file, prop);
+  } catch (std::bad_alloc &e) {
```
kparichay, line 247: I think this is redundant? `std::exception` can catch this exception too and do the same action.
Return value would be different but -1 is used in other places in this function as well.

kparichay (approved):

kparichay:
Please rebase once to pass the CI.

nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
zhoonit: Done

PR #616, opened 2020-10-05
[ Packaging ] Meta Package for NNTrainer

jijoongmoon:
NNTrainer Meta :
 - nntrainer-core
 - nnstreamer-nntrainer if nnstreamer_filter is on
 - capi-nntrainer if with tizen

**Self evaluation:**

### DIFF
--- a/packaging/nntrainer.spec
+++ b/packaging/nntrainer.spec
@@ -59,10 +59,24 @@ BuildRequires:	nnstreamer-devel
 %endif #nnstreamer_filter
 %endif  # tizen
 
+Requires:	nntrainer-core = %{version}-%{release}
+
+%if  0%{?nnstreamer_filter}
+Requires:	nnstreamer-nntrainer = %{version}-%{release}
+%endif #nnstreamer_filter
+%if %{with tizen}
+Requires:	capi-nntrainer = %{version}-%{release}
+%endif #tizen
+
+%description
+NNtrainer Meta package for tizen
+
+%package core
+Summary:	Software framework for traning neural networks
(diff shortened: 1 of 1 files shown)


zhoonit (approved):
LGTM

packaging/nntrainer.spec
```diff
@@ -254,6 +268,8 @@ cp -r result %{buildroot}%{_datadir}/nntrainer/unittest/
 %postun -p /sbin/ldconfig
 
 %files
```
kparichay, line 270: remove this? as nntrainer package depends on nntrainer-core, these files will be installed by core.

jijoongmoon: This should be empty package. It's only for the resolve dependency.

kparichay: I am not so sure about this.
Please once check - https://wiki.tizen.org/Security/Application_installation_and_Manifest. It says that "Each RPM package in MSM must have a manifest file".

kparichay (approved):

packaging/nntrainer.spec
jijoongmoon: I think this is not actually installed. The packages which has dependency with it and has manifest file are installed.

myungjoo: If a package does NOT have a binary or executable (.so, .out, .sh, and so on), it has no execution incurred by the package, so MSM has nothing to register for the package. So, if the package has no "execution files", you don't need to care MSM. That's why we do not add manifest file to packages without .so, .out, .sh files.

PR #618, opened 2020-10-05
[neuralnet] Add alternatives to getLoss

### LINKED ISSUE #617: MNIST application fails
MNIST application loss value fails matching in main branch

kparichay:
getLoss used to get the current loss of the model which was based
on the previous batch of data which the network ran on.
This does not allow getting training/validation loss.
Added getTrainingLoss and getValidationLoss for this purpose.
And update getLoss description to include this information.

As MNIST application was using getLoss() which returns the
loss of the last ran element, this value was changed with #600
as with #600 last element is a batch of data than just 1 data element.
The application is updated to now compare all three loss with
updated values.
So, this patch fixes that bug in main branch as well.

Resolves #617

**Self evaluation:**

### DIFF
--- a/Applications/MNIST/jni/main.cpp
+++ b/Applications/MNIST/jni/main.cpp
@@ -90,6 +90,8 @@ const float tolerance = 0.1;
 std::string data_path;
 
 float training_loss = 0.0;
+float validation_loss = 0.0;
+float last_batch_loss = 0.0;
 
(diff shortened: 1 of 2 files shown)


zhoonit (approved):
Nice catch!

jijoongmoon (approved):
LGTM

Issue #621, opened 2020-10-06
[buffer] Queue is not related properly


zhoonit:

zhoonit:
Resolved via #623

PR #624, opened 2020-10-06
[Apply Fix to Main][Databuffer] Fix queue is not emptied properly

zhoonit:
There was an issue that databuffer is not emptying queue properly.
This patch fixes the issue.o

Reoslves #621

**Self evaluation:**


jijoongmoon:
@zhoonit You don't need to do pr twice. I will handle this.

zhoonit:
> @zhoonit You don't need to do pr twice. I will handle this.

Thanks

PR #626, opened 2020-10-06
Refactor getDataFromBuffer

### LINKED ISSUE #611: Adding flatten layer to cs slowing down significantly
To make inference happening, I added flatten layer to the customshortcut.
1280:7:7:1 -> 1:62720:1:1

One epoch takes around 0.1s to 2sec (didn't exactly measure the time)

I guess there is memcopy happening around flatten layer??

zhoonit:
This patch rearranges getDataFromBuffer to not rely on nested buffers to
avoid unnecessary allocation & deallocation.

This patch provides notable speedup for some cases.

resolves #611 

**Self evaluation:**

### DIFF
--- a/nntrainer/src/databuffer.cpp
+++ b/nntrainer/src/databuffer.cpp
@@ -220,147 +220,80 @@ int DataBuffer::clear() {
   return status;
 }
 
-bool DataBuffer::getDataFromBuffer(BufferType type, vec_4d &outVec,
-                                   vec_4d &outLabel) {
-  unsigned int J, i, j, k, L, l;
-  unsigned int width = input_dim.width();
-  unsigned int height = input_dim.height();
-  unsigned int channel = input_dim.channel();
+bool DataBuffer::getDataFromBuffer(BufferType type, float *out, float *label) {
+
+  using QueueType = std::vector<std::vector<float>>;
+
+  auto wait_for_data_fill = [](std::mutex &ready_mutex,
+                               std::condition_variable &cv, DataStatus &flag,
+                               const unsigned int batch_size,
+                               QueueType &queue) {
+    while (true) {
+      std::unique_lock<std::mutex> ul(ready_mutex);
+      cv.wait(ul, [&]() -> bool { return flag; });
+      if (flag == DATA_ERROR || flag == DATA_END)
+        return queue.size() < batch_size ? false : true;
+
+      if (flag == DATA_READY && queue.size() >= batch_size)
+        return true;
+    }
+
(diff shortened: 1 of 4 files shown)


jijoongmoon (approved):
LGTM! Much Better!

zhoonit:
MNIST result matches. No problem :)

nntrainer/src/neuralnet.cpp
```diff
@@ -448,12 +448,16 @@ int NeuralNetwork::train_run() {
 
     int count = 0;
 
+    TensorDim inputDim = getInputDimension();
+    TensorDim labelDim{batch_size, 1, 1,
```
kparichay, line 452: why use just width of layers.back() than its full output dimension?

nntrainer/src/neuralnet.cpp
```diff
@@ -448,12 +448,16 @@ int NeuralNetwork::train_run() {
...
+                       layers.back()->getOutputDimension().width()};
+    std::shared_ptr<float> in, label;
```
kparichay, line 454: how about creating a tensor here itself with the given shape. Then passing float shared_ptr from `tensor.getData()` to `getDataFromBuffer()`?

nntrainer/src/databuffer.cpp
```diff
@@ -220,147 +220,87 @@ int DataBuffer::clear() {
...
-          return false;
-        else
-          break;
-      }
-      if (valReadyFlag == DATA_READY && val_data.size() >= batch_size) {
-        break;
-      }
-    }
+      std::unique_lock<std::mutex> ul(ready_mutex);
+      cv.wait(ul, [&]() -> bool { return flag; });
```
kparichay, line 235: condition_variables have spurios wakes - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-wait
So, there can be spurious wakeups, and this will just throw for it - it must  NOT.

nntrainer/src/neuralnet.cpp
zhoonit: Right. I'll change

nntrainer/src/neuralnet.cpp
zhoonit: good idea. It'll be done :)

zhoonit: Changed to use `float * out, float * label` that is already allocated from a tensor.

kparichay (approved):
LGTM

PR #632, opened 2020-10-08
[ Max Pooling ] Fix bug when max_id created

jijoongmoon:
This PR includs:
  - make Layer::setBatch virtual function to set batch size
  - max_id for max poooling should be created at initialization time
  but the batch size is not set properly at this time. So when the
  setBatch is called for each layer, it is set properly.

TODO: depnding on inference and training, max_id should be handled
properly. It is not needed for inference.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/pooling2d_layer.cpp
+++ b/nntrainer/src/pooling2d_layer.cpp
@@ -168,6 +168,16 @@ int Pooling2DLayer::setSize(int *size, PropertyType type) {
   return status;
 }
 
+void Pooling2DLayer::setBatch(unsigned int batch) {
+  Layer::setBatch(batch);
+
+  if (pooling_type == PoolingType::max) {
+    max_idx.resize(output_dim.getDataLen());
+  } else if (pooling_type == PoolingType::global_max) {
+    max_idx_global.resize(output_dim.getDataLen());
+  }
+}
+
 void Pooling2DLayer::copy(std::shared_ptr<Layer> l) {
   std::shared_ptr<Pooling2DLayer> from =
     std::static_pointer_cast<Pooling2DLayer>(l);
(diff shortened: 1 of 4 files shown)


nntrainer/src/pooling2d_layer.cpp
```diff
@@ -168,6 +168,19 @@ int Pooling2DLayer::setSize(int *size, PropertyType type) {
   return status;
 }
 
+void Pooling2DLayer::setBatch(unsigned int batch) {
+  input_dim.setTensorDim(0, batch);
```
kparichay, line 172: use `Layer::setBatch(batch)` than having the same implementation here.

nntrainer/src/pooling2d_layer.cpp
```diff
@@ -168,6 +168,19 @@ int Pooling2DLayer::setSize(int *size, PropertyType type) {
...
+  output_dim.setTensorDim(0, batch);
+
+  if (pooling_type == PoolingType::max) {
+    max_idx.resize(output_dim.getDataLen());
+  }
+
+  if (pooling_type == PoolingType::global_max) {
```
kparichay, line 179: I think should be `else if`?

zhoonit (approved):
LGTM after comments above has been resolved

nntrainer/src/pooling2d_layer.cpp
jijoongmoon: done.

nntrainer/src/pooling2d_layer.cpp
jijoongmoon: done.

kparichay (approved):
LGTM

PR #633, opened 2020-10-08
[optimizer] Refactor optimizer

kparichay:
This patch refactors optimizer in the following fashion:
- split optimizer implementations of different types to derived classes of adam and sgd
- create a optimizer_factory to create the optimizer class objs based on its type
  - this can be directly used with ccapi
- OptParam struct has been removed
- applyGradients has been broken down into different methods
- updated associated unittests

In a future PR, I will create subfolders in src/include to manage files better.

See also #199

**Self evaluation:**

### DIFF
--- a/nntrainer/include/optimizer.h
+++ b/nntrainer/include/optimizer.h
@@ -37,45 +37,29 @@ namespace nntrainer {
  */
 enum class OptType { sgd = 0, adam = 1, unknown = 2 };
 
-/**
- * @brief     type for the Optimizor to save hyper-parameter
- */
-typedef struct _OptParam {
-  float learning_rate;
-  double beta1;
-  double beta2;
-  double epsilon;
-  float decay_rate;
-  float decay_steps;
-  bool continue_train; /** Continue training with previous tensors for adam */
-
-  _OptParam(OptType type = OptType::adam) :
-    learning_rate(0.001f),
-    beta1(0.9f),
-    beta2(0.999f),
-    epsilon(1.0e-7f),
-    decay_rate(1.0f),
-    decay_steps(-1.0f),
-    continue_train(false) {
-    if (type == OptType::sgd) {
-      learning_rate = 0.01f;
-    }
-  }
-} OptParam;
-
 class Optimizer {
+
+  /** Allow layer to initialize optimizer with itself */
+  friend class Layer;
+
 public:
   /**
-   * @brief     Constructor of Optimizer Class
+   * @brief     Default Constructor of Optimizer Class
    */
-  Optimizer() : type(OptType::unknown), popt() {}
-
-  Optimizer(const OptType type, OptParam popt);
+  Optimizer(const OptType t, float lr, float decay_rate = 1.0f,
+            float decay_steps = -1.0f, float continue_train = false) :
+    type(t),
+    learning_rate(lr),
+    decay_rate(decay_rate),
+    decay_steps(decay_steps),
+    continue_train(continue_train) {
+    checkValidation();
+  }
 
   /**
    * @brief     Destructor of Optimizer Class
    */
-  ~Optimizer() {}
+  virtual ~Optimizer() {}
 
   /**
    * @brief  copy constructor
@@ -102,14 +86,6 @@ public:
   Optimizer &operator=(Optimizer &&rhs) = default;
 
   /**
-   * @brief     set Optimizer Type
-   * @param[in] t Optimizer type
-   * @retval #ML_ERROR_NONE Successful.
-   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
-   */
-  int setType(OptType t);
-
-  /**
    * @brief     get Optimizer Type
    * @retval    Optimizer type
    */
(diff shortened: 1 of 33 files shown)


nntrainer/src/optimizer.cpp
```diff
@@ -34,73 +34,25 @@
...
+  double ll = learning_rate;
 
-int Optimizer::setOptParam(OptParam p) {
-  int status = ML_ERROR_NONE;
-  if (p.learning_rate <= 0) {
-    ml_loge("Error: learning_rate should be grater than 0 (%f)",
-            p.learning_rate);
-    return ML_ERROR_INVALID_PARAMETER;
+  if (decay_steps != -1) {
+    ll = ll * pow(decay_rate, (iteration / decay_steps));
```
zhoonit, line 46: `powf` for the precision?

nntrainer/src/adam.cpp
```diff
@@ -0,0 +1,155 @@
...
+int Adam::initialize(std::shared_ptr<Weight> weight_list,
+                     unsigned int num_weights, bool set_tensor) {
+  int status = ML_ERROR_NONE;
+  weight_mv.clear();
+
+  if (set_tensor) {
+    for (unsigned int i = 0; i < num_weights; ++i) {
+      Weight &w = weight_list.get()[i];
+
+      // TODO: only trainable weights must be sent to optimizer
```
zhoonit, line 34: Since trainable weights might not be in the contiguous memory space, this might be the only option to filter weights effectively.

E.g)
weights: `[a, b, c]`, and `b.trainable = false`

nntrainer/include/optimizer.h
```diff
@@ -119,27 +95,19 @@ class Optimizer {
    * @brief     get Learning Rate
    * @retval    Learning rate
    */
-  float getLearningRate() { return popt.learning_rate; };
+  float getLearningRate() { return learning_rate; };
```
zhoonit, line 98: `virtual float getLearningRate()` and change `adam::getLearningRate` to float?

nntrainer/src/adam.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+      file.seekg(total_size, std::ifstream::cur);
+    }
+  } else {
+    ml_logw("Not loading saved optimizer parameters due to mismatched type");
+  }
+}
+
+void Adam::save(std::ofstream &file) {
+  file.write((char *)&type, sizeof(OptType));
```
zhoonit, line 147:
```suggestion
    Optimizer::save(file)
```

nntrainer/src/adam.cpp
```diff
@@ -0,0 +1,155 @@
...
+    status = ML_ERROR_NONE;
+    break;
+  }
+
+  throw_status(status);
+}
+
+void Adam::read(std::ifstream &file) {
+  OptType loaded_type;
+  file.read((char *)&loaded_type, sizeof(OptType));
```
zhoonit, line 126: Optimizer::read(file) will throw so `if(loaded_type == type)` can be removed

nntrainer/include/optimizer.h
```diff
@@ -201,36 +150,82 @@ class Optimizer {
...
+
+  /**
+   * @brief     apply gradient to the given weight
+   * @param[in] weight Weight and gradient set to be updated
+   * @param[in] tensor_idx Idx of this tensor in the tensors list
+   * @param[in] num_weights size of the array
+   * @param[in] iteration nth epoch number
+   * @note weight which is called upon can be assumed to be trainable
+   */
+  virtual void apply_gradient(Weight &weight, int tensor_idx, double updated_lr,
```
zhoonit, line 225: I think double and float are mixed for the learning rate. Could you check once up?

nntrainer/src/optimizer.cpp
kparichay: pow operates on double, powf operates on float.

nntrainer/src/adam.cpp
kparichay: I think the weights passed here to be trainable or not is fixed in logic of the layer.
So, layers can possibly two set of containers for weights - trianable and non-trainable
or just index them better.

Reason is that this check is performed in critical path of training - applyGradients()

nntrainer/include/optimizer.h
kparichay: I think you are mixing two things.
There are two functionsin optimizer.h:
`float getLearningRate()` which is a public function
`virtual double getLearningRate(int iteration)` which is a private function and overloaded in adam.h

nntrainer/src/adam.cpp
kparichay: that's my bad, updated.

nntrainer/src/adam.cpp
kparichay: `if(loaded_type == type)` is still needed. 
Updated `optimizer.read()` to  only throw for unknown types.
it will just load the optimizer type for checking, and then the overload optimizer, if the type matches can load corresponding data.

nntrainer/include/optimizer.h
kparichay: `updated_lr` is intentionally double. Learning rate is handled as a double internally. However, for public view and the initial value are stored as a float but any operation or intermediate results of learning rate are performed as double.

nntrainer/src/adam.cpp
zhoonit: Got it

nntrainer/src/optimizer.cpp
zhoonit: No, https://github.com/nnstreamer/nntrainer/pull/633#discussion_r503057084 this explained Thanks.

nntrainer/include/optimizer.h
zhoonit: Got it

nntrainer/include/optimizer.h
zhoonit: Alright

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #642, opened 2020-10-12
[IntegratedTest] Add test generator

zhoonit:
Add test generator and model recorder for integrated test cases.

This runs multiple epochs and save related informations.

**Self evaluation:**

### DIFF
--- a/test/input_gen/genModelTests.py
+++ b/test/input_gen/genModelTests.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+##
+# Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+#
+# @file getModelTests.py
+# @date 13 October 2020
+# @brief Generate tc using KerasRecorder
+# @author Jihoon lee <jhoon.it.lee@samsung.com>
+
+import warnings
+
+from .recorder import KerasRecorder
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=FutureWarning)
+    import numpy as np
+    import tensorflow as tf
+    from tensorflow.python import keras as K
+
+if __name__ == "__main__":
+    inp = K.Input(shape=(3, 3))
+    a = K.layers.Dense(5)(inp)
+    b = K.layers.Dense(5)(a)
+    c = K.layers.Dense(10)(b)
+    d = K.layers.Activation("softmax")(c)
+
+    KerasRecorder(
+        file_name="a.info",
+        inputs=inp,
+        outputs=[inp, a, b, c, d],
+        input_shape=(3, 3),
+        label_shape=(3, 10),
+        loss_fn=tf.keras.losses.CategoricalCrossentropy(),
+    ).run(2)
--- a/test/input_gen/recorder.py
+++ b/test/input_gen/recorder.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+##
+# Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+#
+# @file recorder.py
+# @date 13 October 2020
+# @brief Generate tc from given keras model
+# @author Jihoon lee <jhoon.it.lee@samsung.com>
+
+from functools import wraps
+import sys
+import os
+import warnings
+import random
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=FutureWarning)
+    import numpy as np
+    import tensorflow as tf
+    from tensorflow.python import keras as K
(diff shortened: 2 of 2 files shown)


test/input_gen/genModelTests.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+##
+# Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+#
+# @file recorder.py
+# @brief Generate tc using KerasRecorder
```
kparichay, line 7: missing date?

test/input_gen/recorder.py
```diff
@@ -0,0 +1,127 @@
...
+class KerasRecorder:
+    def __init__(
+        self,
+        file_name,
+        inputs,
+        outputs,
+        input_shape,
+        label_shape,
+        loss_fn=None,
+        optimizer=tf.keras.optimizers.SGD(lr=1.0),
```
kparichay, line 46: Maybe in a later PR, but you would want another argument here `from_logits`.
In this case, the activation for the last layer (for eg softmax) is amalgamated in the loss itself.
So, the output of the last layer must add that activation by itself (just for returning and not for training) to match the output with nntrainer.
This is needed because amalgamated losses are mathematically more stable than the separated versions.

test/input_gen/recorder.py
```diff
@@ -0,0 +1,127 @@
...
+# @brief Record Keras model with some watchers attached
+# @note  The class might need to go through some rework for non-sequential model
+# in case of the order of graph traversal is diffrent from NNTrainer
+class KerasRecorder:
+    def __init__(
+        self,
+        file_name,
+        inputs,
+        outputs,
+        input_shape,
```
kparichay, line 43: I think you will need a batch size separately in this

test/input_gen/recorder.py
```diff
@@ -0,0 +1,127 @@
...
+        label_shape,
+        loss_fn=None,
+        optimizer=tf.keras.optimizers.SGD(lr=1.0),
+    ):
+        self.inputs = inputs
+        self.outputs = outputs
+        self.model = K.Model(inputs=inputs, outputs=outputs)
+        self.loss_fn = loss_fn
+        self.optimizer = optimizer
+        self.file = open(file_name, "wb")
```
kparichay, line 53: how about raise if file already exists, or warn?

test/input_gen/recorder.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+            self.optimizer.apply_gradients(zip(gradients, layer.trainable_weights))
+
+        self._write_items(results[-1])
+
+    ##
+    # @brief run function
+    # @param epoch number of iteration to run
+    def run(self, epoch):
```
kparichay, line 125: I think this might be better named as iteration

test/input_gen/recorder.py
zhoonit: Isn't this done by adding 

`loss_fn=tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)`
and omit softmax activation on he last layer?

test/input_gen/recorder.py
zhoonit: Could you explain more about this? I am not getting the benefit from separating the batchsize

test/input_gen/genModelTests.py
zhoonit: updated

test/input_gen/recorder.py
zhoonit: Done

test/input_gen/recorder.py
zhoonit: Done

test/input_gen/recorder.py
kparichay: Yes, this is done for training, but inference, softmax is still needed (which requires extra conditions than just setting `from_logits=True`).

test/input_gen/recorder.py
kparichay: In tf.keras, input layer takes input_shape and batch_size separately.
As some tests will want to run over different batches, batch_size might be needed.

zhoonit: Sounds like it is better to restructure the class to pass input_shape, label_shape and filename to `self.run` phase. To make the behavior consistent, I'll add a test using this python script first and make another PR to restructure this file to make sure things don't break down.

test/input_gen/recorder.py
zhoonit: Ditto. I will do this in diffrent PR. For now, I'll leave a note in my issue

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #647, opened 2020-10-19
[Bug/Act] Fix setActivation call properly

zhoonit:
`Act::setActivation` should be called for activationLayer.
However, because `virtual Layer::setActivation` had diffrent signature
`Act::setActivation`, There was no way this function can be called.

This patch fixes the issue.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/layer.h
+++ b/nntrainer/include/layer.h
@@ -240,6 +240,13 @@ public:
   int setOptimizer(std::shared_ptr<Optimizer> opt);
 
   /**
+   * @brief Get the Optimizer object
+   *
+   * @return std::shared_ptr<Optimizer> optimizer
+   */
+  std::shared_ptr<Optimizer> getOptimizer() { return opt; }
+
+  /**
    * @brief     Activation Type Getter
    * @retval    Activation Type.
    */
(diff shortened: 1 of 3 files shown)


nntrainer/include/activation_layer.h
```diff
@@ -76,7 +76,7 @@ class ActivationLayer : public Layer {
    *
    * @param[in] ActivationTypeeActivationTypeeActivationTypeet
    */
-  void setActivation(ActivationType acti_type);
+  int setActivation(ActivationType acti_type);
```
kparichay, line 79: instead of changing the return type of this to int, change the signature of `setActivaiton` in `layer.h` to void?

zhoonit: Sure I'll do that

jijoongmoon:
> There was a bug that set activation is not called in proper order.
> This fixes the issue.
> 
> **Self evaluation:**
> 
> 1. Build test: [X]Passed [ ]Failed [ ]Skipped
> 2. Run test: [X]Passed [ ]Failed [ ]Skipped

It might be better if you have signed-off here as well.

zhoonit:
> It might be better if you have signed-off here as well. and It might be even better if you give more specific bug case

I'll update Thanks!

jijoongmoon (approved):
LGTM

zhoonit:
v2. updated `setActivation` to void, update commit message :)

nntrainer/include/activation_layer.h
zhoonit: Done

kparichay (approved):
LGTM

PR #648, opened 2020-10-19
[Weight] Add ctors

### LINKED ISSUE #652: copy assigment of Weight must be disabled
Copy assignment of weight must be disabled.
It is only used in `Layer::copy` with wrong usage where weights aren't actually copied.
Refer to discussion https://github.com/nnstreamer/nntrainer/pull/648#discussion_r508174968

zhoonit:
Weight did not have copy/move ctor which they should have.
This patch adds them.

Resolves #652 

**Self evaluation:**

### DIFF
--- a/nntrainer/include/weight.h
+++ b/nntrainer/include/weight.h
@@ -102,12 +102,26 @@ public:
   }
 
   /**
+   * @brief Copy constructor for weight
+   *
+   * @param rhs weight to construct from
+   */
+  Weight(const Weight &rhs);
+
+  /**
+   * @brief Move constructor for weight
+   *
+   * @param rhs weight to construct from
+   */
+  Weight(Weight &&rhs) = default;
+
+  /**
    * @brief copy assigment
    *
    * @param rhs copy from
    * @return Weight& Updated weight
    */
-  Weight &operator=(const Weight &rhs) = default;
+  Weight &operator=(const Weight &rhs);
 
   /**
    * @brief move assignment
--- a/nntrainer/src/weight.cpp
+++ b/nntrainer/src/weight.cpp
@@ -16,6 +16,20 @@
 
 namespace nntrainer {
 
+Weight::Weight(const Weight &rhs) :
+  initializer(rhs.initializer),
+  trainable(rhs.trainable),
+  name(rhs.name) {
+  var = rhs.var.clone();
+  grad = rhs.grad.clone();
+}
+
+Weight &Weight::operator=(const Weight &rhs) {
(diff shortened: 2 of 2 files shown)


nntrainer/include/weight.h
```diff
@@ -168,6 +181,7 @@ class Weight {
    */
   Tensor &getGradientRef() { return grad; }
 
+private:
```
kparichay, line 184: unnecessary change

nntrainer/include/weight.h
```diff
@@ -101,6 +101,20 @@ class Weight {
     swap(lhs.name, rhs.name);
   }
 
+  /**
+   * @brief Copy constructor for weight
+   *
+   * @param rhs weight to construct from
+   */
+  Weight(const Weight &rhs) = default;
```
kparichay, line 109: is this essentially needed? If some layer writer uses these, then the weights of both the layers would be bonded together
and create nasty bugs.

zhoonit: If we have a copy assignment, we need a copy constructor.  If we are gonna delete a copy assignment, I think it's fine for me.  I am gonna be using this for the integrated tests, btw, but I can work around I think we decide to not allow copy ctor/assignment.

I got what you mean though, it seems that there should be a broader patch that changes copy assignment as well.

nntrainer/include/weight.h
zhoonit: This should have gone to a separate patch. I'll move this after the discussion. 

I am thinking to open this up for the test. Is there a reason Ref 's are private?

kparichay: Yes, so that any other class other the friend class cannot directly update this as an operation. 
I know `float * data` is accessible, but if someone is changing that, its intentional.
Getting weight by reference is only allowed for object owners which is this case is the layer and its optimizer.

nntrainer/include/weight.h
kparichay: Copy assignment is wrong. 
It is currently only used in `Layer::copy` with wrong usage. It must create a deep copy of the weights but it rather binds the weights of the layer together as copy constructor of Tensor binds the memory together.

nntrainer/include/weight.h
zhoonit: The change has been reverted. I'll use data for now.

kparichay (commented):
LGTM
good bug fix

kparichay (approved):

jijoongmoon (approved):
LGTM

PR #649, opened 2020-10-19
[NN] call setBatchSize at init

zhoonit:
setBatchSize is not called at init which cause a problem when calling
layers from outside.
This fixes the issue

**Self evaluation:**

### DIFF
--- a/nntrainer/src/neuralnet.cpp
+++ b/nntrainer/src/neuralnet.cpp
@@ -249,6 +249,7 @@ int NeuralNetwork::init() {
     ml_logd("layer name: %s", l->getName().c_str());
 
   initialized = true;
+  setBatchSize(batch_size);
   return status;
 }
 


nntrainer/src/neuralnet.cpp
```diff
@@ -249,6 +249,7 @@ int NeuralNetwork::init() {
     ml_logd("layer name: %s", l->getName().c_str());
 
   initialized = true;
+  setBatchSize(batch_size);
```
kparichay, line 252: batch size is to be decided at running time (training/testing)

zhoonit: There is a scenario that a user is calling an equivalent consecutive function calls to alternate `training()/testing()` after `NN.init()` which Keras or other frameworks support to. 

In this scenario, because batch has not set (`NN.batchsize` has set but `setBatchSize` has not been called and it is not visible from API's point of view), it does not properly even it has been initiated. This patch is to alleviate the problem.

kparichay: Calling `setBatchSize()` at `init()` does not solve the problem but actually just hides it.
`setBatchSize()` must be called with each call to `training()/testing()` as different batch size can be used and is supported.

In the `nerualnet.cpp` I can see, `inference` and `train` both call `setBatchSize()` before running the model.
Validation and testing do not do this in the current implementation as their batchsize is not part of the interface yet.

zhoonit: Do you have a suggestion? as long as we are opening other plumbing api's  (like NN->forward, getFlatgraph...) we should resolve this problem anyway.

 I had an alternative which didn't seem appealing:
1. create a `getBatchSize()`
2. call `setBatchSize(getBatchSize()` at user side.

kparichay: resolved offline.

kparichay (approved):
LGTM

jijoongmoon (approved):
We discussed about this quite ago, it is set batch size at init then. LGTM

PR #656, opened 2020-10-21
Layer optimizer factory bugfix

### LINKED ISSUE #655: [layer/optimizer] Bug in variadic templates
layer/optimizer factory constructor used variadic templates.
However, variadic templates require same signature for all the objects constructed by factory.
These factory methods needs updation.

kparichay:
layer/optimizer factory constructor used variadic templates
However, that wasn't used and could not be used with the way it was made
The way to fix it was using c-style variable number of arguments in between
to allow different function signatures work with templates at compile time
as c-style variable number of arguments are not interpreted at compile time.

This creates an interface like below:
```
createLayer(LayerType::LAYER_FC, 4, ActivationType::ACT_RELU);
createLayer(LayerType::LAYER_CONV, 4, std::array<int, 2>{1,1}, std::array<int, 2>{2,2});
```
vs the current interface
```
createLayer(LayerType::LAYER_FC, {"unit=4", "activation=relu"})
createLayer(LayerType::LAYER_CONV, 4, {"strides=1:1", "filter=2:2"});
```

The first is more compact and performs some checks on types (not all) etc.
However, the second is more expressive and readable.
I have update the internal and external factory methods which makes the final
style to the latter version.

Simplified c-API implementation to also use these factory methods.

Resolves #655
More functionality for cc-api will be added as per need basis. This includes exposing Weight/Tensor with the API.

**Self evaluation:**


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #662, opened 2020-10-21
[ TENSOR ] Change to get Tensor Vector

jijoongmoon:
Currently we only take and push one tensor per layer. For the skip
connection or other layer which takes and make multiple tensors, we
need to handle tensors as an input and output of layer.

In this PR, tensor input chaneged to tensor vector.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -149,7 +149,7 @@ public:
    * @param[in] input List of Input Tensors taken by the neural network
    * @retval    List of Output Tensors
    */
-  sharedConstTensor forwarding(sharedConstTensor input);
+  sharedConstTensors forwarding(sharedConstTensors input);
 
   /**
    * @brief     Forward Propagation of the neural network
@@ -157,8 +157,8 @@ public:
    * @param[in] label List of Label Tensors for the model
    * @retval    List of Output Tensors
    */
-  sharedConstTensor forwarding(sharedConstTensor input,
-                               sharedConstTensor label);
+  sharedConstTensors forwarding(sharedConstTensors input,
+                                sharedConstTensors label);
 
   /**
    * @brief     Backward Propagation of the neural network
@@ -166,7 +166,7 @@ public:
    * @param[in] label List of Label Tensors for the model
    * @param[in] iteration Iteration Number for the optimizer
    */
-  void backwarding(sharedConstTensor input, sharedConstTensor label,
+  void backwarding(sharedConstTensors input, sharedConstTensors label,
                    int iteration);
 
   /**
@@ -205,7 +205,7 @@ public:
    * @param[in] X input tensor
    * @retval shared_ptr<const Tensor>
    */
-  sharedConstTensor inference(const Tensor X);
+  sharedConstTensors inference(sharedConstTensors X);
 
   /**
    * @brief     Run NeuralNetwork train with callback function by user
--- a/nntrainer/src/conv2d_layer.cpp
+++ b/nntrainer/src/conv2d_layer.cpp
@@ -64,9 +64,9 @@ void Conv2DLayer::read(std::ifstream &file) { Layer::read(file); }
 
 void Conv2DLayer::save(std::ofstream &file) { Layer::save(file); }
 
-sharedConstTensor Conv2DLayer::forwarding(sharedConstTensor in) {
+sharedConstTensors Conv2DLayer::forwarding(sharedConstTensors in) {
   int status = ML_ERROR_NONE;
-  input = *in;
+  input = *in[0];
 
   if (normalization) {
     input = input.normalization();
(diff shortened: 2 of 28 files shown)


nntrainer/include/neuralnet.h
```diff
@@ -207,7 +207,7 @@ class NeuralNetwork {
    * @param[in] X input tensor
    * @retval shared_ptr<const Tensor>
    */
-  sharedConstTensor inference(const Tensor X);
```
kparichay, line 210: input for inference must also be `sharedTensors`?
This is to keep the semantics of multiple inputs/multiple outputs.

nntrainer/src/conv2d_layer.cpp
```diff
@@ -229,9 +229,9 @@ sharedConstTensor Conv2DLayer::backwarding(sharedConstTensor derivative,
       input.getAddress(b * input.getDim().getFeatureLen()));
 
     status = conv2d_gemm(
-      derivative->getAddress(b * derivative->getDim().getFeatureLen()),
-      TensorDim(1, derivative->channel(), derivative->height(),
-                derivative->width()),
+      derivative[0]->getAddress(b * derivative[0]->getDim().getFeatureLen()),
```
kparichay, line 232: I think it might be more readable to get `derivative[0]` out separately.

nntrainer/src/neuralnet.cpp
```diff
@@ -280,11 +280,11 @@ sharedConstTensor NeuralNetwork::forwarding(sharedConstTensor input) {
...
  */
-sharedConstTensor NeuralNetwork::forwarding(sharedConstTensor input,
-                                            sharedConstTensor label) {
-  sharedConstTensor X;
+sharedConstTensors NeuralNetwork::forwarding(sharedConstTensors input,
+                                             sharedConstTensors label) {
+  sharedConstTensors X;
 
-  if (input->getDim().batch() > batch_size)
+  if (input[0]->getDim().batch() > batch_size)
```
kparichay, line 287: as the input is coming from outside, lets also check that the size of the vector matches with number of inputs for the first layer. It is sufficient to be done only for the first layer, as the input from user can be erroneous but intermediate layers can be trusted and have been verified.
Same for the label going to the backwarding.

nntrainer/include/layer.h
```diff
@@ -128,16 +128,16 @@ class Layer {
    * @param[in] in List of Input Tensors taken by this layer
    * @retval    List of Output Tensors
    */
-  virtual sharedConstTensor forwarding(sharedConstTensor in) = 0;
+  virtual sharedConstTensors forwarding(sharedConstTensors in) = 0;
```
kparichay, line 131: please also update weight_list from shared_ptr<Weight> to vector<Weight>.
It is supposed to follow the same semantics as tensor list.

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -1789,13 +1808,13 @@ TEST(nntrainer_ActivationLayer, forward_backward_01_p) {
                  nntrainer::ActivationLayer::relu((l - 4) * 0.1 * (i + 1)));
   nntrainer::Tensor result;
   EXPECT_NO_THROW(result =
-                    layer.forwarding(MAKE_SHARED_TENSOR(input)).get()[0]);
+                    layer.forwarding({MAKE_SHARED_TENSOR(input)})[0].get()[0]);
```
zhoonit, line 1811: Because shared_ptr shouldn't be interpreted as an array from this PR, I guess it is worth changing `sharedTensor.get()[0]` to `*sharedTensor` as well :)

and for others parts! 
```suggestion
                    *layer.forwarding({MAKE_SHARED_TENSOR(input)})[0]);
```

zhoonit (approved):
LGTM except for a minor comment

nntrainer/src/neuralnet.cpp
jijoongmoon: I don't think multiple input for layers allowed even if it is input layer. IMO, if someone want to use multiple tensor, then he have to use concatenate or add layer first after set individual input layer.

I think We need to kind of tensors class for better handling but it can be separate pr though. Let's assume that this pr is to enable skip connection.

nntrainer/include/layer.h
jijoongmoon: Well, I'm not sure. Basic rule here is that layer only has one input except several layer.

test/unittest/unittest_nntrainer_layers.cpp
jijoongmoon: We are using both in other codes. We have to find and change it separate pr.

nntrainer/src/neuralnet.cpp
kparichay: Then in the later PR, we can have a check on the count of the number of input tensors and the number of input layers.

nntrainer/include/layer.h
kparichay: shared_ptr<Weight> is representing a fixed size array of weights with the number elements in array stored separately.
So shared_ptr<Weight> can be replaced with vector<Weight>.

jijoongmoon: In order to do this, there are a lot more changes. I will do it later pr.

nntrainer/src/conv2d_layer.cpp
jijoongmoon: done.

nntrainer/include/neuralnet.h
jijoongmoon: fixed. Thanks.

nntrainer/include/layer.h
kparichay: Sure.

kparichay (approved):
LGTM

PR #663, opened 2020-10-22
[optimizer] private ownership of optimizer in layer

kparichay:
This patch updates the ownership semantics of optimizer in the layer.
Before this patch, although, each layer creates its own optimizer, yet it was allowed to
be shared with a shared_ptr. With this patch, optimizer is a unique_ptr meaning 
that it is owned by the layer and cannot be shared directly with other objects.
    
    **Self evaluation:**
    
### DIFF
--- a/nntrainer/include/layer_internal.h
+++ b/nntrainer/include/layer_internal.h
@@ -321,8 +321,7 @@ protected:
   /**
    * @brief     Optimizer for this layer
    */
-  // TODO: fix with #630
-  std::shared_ptr<Optimizer> opt;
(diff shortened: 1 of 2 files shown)


zhoonit (approved):
As discussed offline, it would be better `this->weights` has optimizer dependent parameter like `weight_mv` and weight could be shared while `weight_mv` remains a unique pointer, we've made a consensus that it is good to go for now :)

jijoongmoon (approved):
LGTM

kparichay:
Closing this PR, making optimizer unique does not allow user/unittest to get it. This can be done later when merged with weights.

Issue #660, opened 2020-10-21
Simplify transfer learning application


kparichay:
The most basic transfer learning application is very complicated because nntrainer does not support specifying backbone from the API. Rather, the user is expected to handle all of it themselves.

Basic support:
- [ ] Support backbone to be specified as a filename in ini file
  - [ ] If this is a tensorflow-lite file, run tflite on the input data first before giving the input to nntrainer
    - [x] support with nnstreamer #683 
    - [x] support directly with tflite to allow easier support on Android #693 
    - [ ] Update transfer learning/other applications to use these backbones #693 #729 
  - [x] If this is nntrainer file, load nntrainer layers and preprend them to the current nntrainer graph - #665

Advanced support: 
- [x] Nested backbones -  #665
- [x] Multiple backbone -  #665
- [x] Allow trainable/non-trainable - #667
- [x] Other frameworks backbone - use nnstreamer single API to run these #683
- [ ] Support backbone specific features from https://github.com/nnstreamer/nntrainer/issues/660#issuecomment-716313854 - #726
- [ ] Support caching the outputs of backbone as that part of the model is fixed. - postponed.
- [ ] backbone conversion - if tflite model file is provided, but that backbone needs training, load weights from tflite backbone to nntrainer and treat it as nntrainer backbone. - postponed.

More advanced support to be added.

kparichay:
Sharing design here which @zhoonit described in an offline meeting earlier:

```ini
[Model]
backbone : filename.${ext}
```
Extensions - ini (handled with nntrainer), tflite/etc (handled with external frameworks)

Note: 
In the existing ini file design, as `Model` does not have a type, for now, defining multiple models to modularize the model declaration was not possible. However, with the backbone, we can make a modular model block and then use it again and again as shown below.

```ini
# resnetblock.ini
[Model]
# Define the resnet block here

....
[InputLayer]
...
[Conv1Layer]
...
[Conv2Layer]
...
[Conv3Layer]
...
[Conv4Layer]
...

```

And then in the model ini, use the block backbone to describe the model modularly.
```ini
[Model]
# Define the resnet model here

....
[InputLayer]
...
[Block1]
backbone=resnetblock.ini
...
[Block2]
backbone=resnetblock.ini

[Pooling2dLayer]
...
[Block3]
backbone=resnetblock.ini
...
[Block4]
backbone=resnetblock.ini
...

[FCLayer]
...
```
[...]

PR #661, opened 2020-10-21
[Test] Add ParamTest scaffolding for model tests 

zhoonit:
Add Parameterized test scaffolding for integrated tests with minor
changes

**Additional Changes proposed in this PR:**
- Add `TensorDim::TensorDim(const std::string &shape)
- Move `Tensor::epsilon` to public

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>
cc: Parichay Kapoor <pk.kapoor@samsung.com>

PR #665, opened 2020-10-22
[model] Add support of ini based backbone to the model

kparichay:
This patch adds support of ini based backbone to the model neural network
From the point of view of ini file, backbone is treated as a layer itself.
This allows a graph of layers to be represented as a layer itself in the
ini file.

With this design, backbone must be specified as a layer with property backbone
as shown below with a sample pseudo-ini:
```ini
[Model]
...

[Block1]
backbone: base_block.ini

[PoolLayer]
type: pooling2d

[Block2]
backbone: base_block.ini
```

ModelLoader loads the layer configuration from the backbone independently
and then extends the existing graph in the main model with this newly created
graph from the backbone ini.

The names of all layers which are inserted from the backbone to a model are
prefixed with the name of the backbone for easier management and for the user
to identify/manage the layers from a backbone.

The patch allows nested backbones and multiple backbones in a model description.
~Unittests for this backbone support will follow in the next patch.~
Extensive unittests have also been added.

See also #660

**Self evaluation:**

### DIFF
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -243,6 +243,15 @@ public:
   int addLayer(NodeType layer);
 
   /**
+   * @brief     join passed graph into the existing graph model
+   * @param[in] graph graph to be added/to extend
+   * @note It is assumed that this model is valid by itself
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   */
+  int extendGraph(GraphType graph, std::string prefix = "");
+
+  /**
    * @brief     set optimizer for the neural network model
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
@@ -286,6 +295,14 @@ public:
    * @retval flatGraph of the current graph
    */
   FlatGraphType getFlatGraph() { return layers; }
+
+  /**
+   * @brief get current graph from the model
+   * @note graph contains pointer to the actual nodes, which is not deeply
+   * copied.
+   * @retval current graph
+   */
+  GraphType getGraph() { return layers; }
 
   /**
    * @brief     Set loss type for the neural network.
@@ -424,7 +441,8 @@ private:
   /**
    * @brief     Ensure that layer has a name
    */
-  void ensureName(NodeType layer, const std::string &prefix = "");
+  void ensureName(NodeType layer, const std::string &prefix = "",
+                  bool force_rename = false);
 
   /**
    * @brief     Swap function for the class
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -245,6 +245,7 @@ public:
   /**
    * @brief     join passed graph into the existing graph model
    * @param[in] graph graph to be added/to extend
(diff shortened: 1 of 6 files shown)


jijoongmoon:
I think we need to way to define last layer of backbone ini like ```exit_layer = [ layer name ] ```. With this, user can define and use freely.

kparichay:
> I think we need to way to define last layer of backbone ini like `exit_layer = [ layer name ] `. With this, user can define and use freely.

That is a good suggestion. That will allow using a subsection of backbone ini file.
The proposal to support this has been added in https://github.com/nnstreamer/nntrainer/issues/660#issuecomment-716313854.
This will be implemented as a later PR.

nntrainer/include/neuralnet.h
```diff
@@ -242,6 +242,15 @@ class NeuralNetwork : public ml::train::Model {
    */
   int addLayer(NodeType layer);
 
+  /**
+   * @brief     join passed graph into the existing graph model
+   * @param[in] graph graph to be added/to extend
```
dongju-chae, line 247: Missing argument?

kparichay: added.

dongju-chae (approved):
LGTM

test/include/nntrainer_test_util.h
```diff
@@ -122,6 +127,21 @@ typedef enum {
 class nntrainerIniTest
   : public ::testing::TestWithParam<
       std::tuple<const char *, const std::vector<IniSection>, int>> {
+public:
+  static void save_ini(const char *filename, std::vector<IniSection> sections,
```
zhoonit, line 131: FYI, There was a major refactoring on ini helper classes.

https://github.com/nnstreamer/nntrainer/pull/661/commits/4f17acd6e406421eb69e2ee4ccd9fa42e6b44c7c

1. `nntrainerIniTest` has moved to `nntrainer_modelfile.cpp` since it is a tailored class for ini test
2. Instead, there is `IniTestWrapper` for general usage of `Ini` handling. So, you can create an instance for each ini file. As it is merged, IMHO, it removes the need of this function and `getIni` function. (filemode argument can be added to IniTestWrapper for the `getIni`)

I should've reviewed this PR earlier and told you beforehand. Sorry about that :(

test/include/nntrainer_test_util.h
```diff
@@ -91,6 +92,10 @@ class IniSection {
 
   IniSection operator+(const std::string &s) { return IniSection(*this) += s; }
 
+  void rename(const std::string &name) { section_name = name; }
```
zhoonit, line 95: FYI, in 4f17acd, I have made SectionName only ctor, so it can be done like
`IniSection("block2") + backbone_valid` to provide isolation between each section.

zhoonit (approved):
LGTM 👍

test/include/nntrainer_test_util.h
kparichay: its alright, I will update with rebase.

test/include/nntrainer_test_util.h
kparichay: I tried this but it failed, didn't investigate, will update in a later PR.

kparichay:
rebased.

jijoongmoon (approved):
LGTM

PR #671, opened 2020-10-27
[ Layer ] Multiple Input Dimensions @open sesame 10/28 14:06

jijoongmoon:
Current implementaion only take one input. In order to take multiple
input, input_dim / output_dim should be vector type.
This PR includes this fixes except about addition layer which requires
actual multiple input. This will be done consequtive PR.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/layer_internal.h
+++ b/nntrainer/include/layer_internal.h
@@ -83,7 +83,10 @@ public:
     trainable(trainable_),
     num_weights(0),
     num_inputs(1),
-    num_outputs(1) {}
+    num_outputs(1) {
+    input_dim.resize(1);
+    output_dim.resize(1);
+  }
 
   /**
    *  @brief  Move constructor of Layer.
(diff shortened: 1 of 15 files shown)


nntrainer/include/layer_internal.h
```diff
@@ -83,7 +83,10 @@ class Layer : public ml::train::Layer {
     trainable(trainable_),
     num_weights(0),
     num_inputs(1),
-    num_outputs(1) {}
+    num_outputs(1) {
```
zhoonit, line 86: I think we should get rid of those two, as vector has it's own size, this violates Single source of truth principle

zhoonit: I'll make an issue to deal separately

zhoonit (approved):
LGTM

kparichay (approved):
LGTM

PR #672, opened 2020-10-28
[Wait4PR#671][ Layer ] Add Concat Layer @open sesame 10/28 14:16

jijoongmoon:
This PR includes Concatenate layer.

**Self evaluation:**

### DIFF
--- a/nntrainer/src/concat_layer.cpp
+++ b/nntrainer/src/concat_layer.cpp
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
+ *
+ * @file	concat_layer.cpp
+ * @date	27 Oct 2020
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is Concat Layer Class for Neural Network
+ *
+ */
+
+#include <concat_layer.h>
+#include <cstring>
+#include <layer_internal.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+int ConcatLayer::initialize() {
+  int status = ML_ERROR_NONE;
+  unsigned int channel = 0;
+  if (num_inputs == 0) {
+    ml_loge("Error: number of inputs are not initialized");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  for (unsigned int idx = 0; idx < num_inputs; ++idx) {
+    channel += input_dim[idx].channel();
+  }
+
+  output_dim[0] = input_dim[0];
+  output_dim[0].channel(channel);
+
+  return status;
+}
+
+sharedConstTensors ConcatLayer::forwarding(sharedConstTensors in) {
+
+  TensorDim d = in[0]->getDim();
+  unsigned int concat_channel = d.channel();
+
+  for (unsigned int idx = 1; idx < num_inputs; ++idx) {
+    TensorDim dim = in[idx]->getDim();
+
+#ifdef DEBUG
+    for (unsigned int i = 2; i < d.rank(); ++i) {
+      if (d[i] != dim[i])
+        throw std::runtime_error("Error: concat layer requires same "
+                                 "shape from  all input layers");
+    }
+#endif
+    concat_channel += dim.channel();
+  }
+
+#ifdef DEBUG
+  if (concat_channel != output_dim[0].channel()) {
+    throw std::runtime_error(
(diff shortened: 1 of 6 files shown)


nntrainer/src/concat_layer.cpp
```diff
@@ -0,0 +1,124 @@
...
+    concat_channel += dim.channel();
+  }
+
+  if (concat_channel != output_dim[0].channel()) {
+    throw std::runtime_error(
+      "Error: Sum of channel of input layers is not same with output channel");
+  }
+
+  hidden = Tensor(output_dim[0]);
+  hidden.setZero();
```
kparichay, line 64: this is redundant

nntrainer/src/concat_layer.cpp
```diff
@@ -0,0 +1,124 @@
...
+sharedConstTensors ConcatLayer::forwarding(sharedConstTensors in) {
+
+  TensorDim d = in[0]->getDim();
+  unsigned int concat_channel = d.channel();
+
+  for (unsigned int idx = 1; idx < num_inputs; ++idx) {
+    TensorDim dim = in[idx]->getDim();
+
+    for (unsigned int i = 2; i < d.rank(); ++i) {
+      if (d[i] != dim[i])
```
kparichay, line 51: how about moving these rank dependent checks to initialise with input_dim and rather just compare in[i].getDim() with input_dim[i]? lets keep forwarding (critical path) least checks.

kparichay: Further, I think that these checks of the dimension of `in` and `input_dim` can be all skipped from `forwarding` as nntrainer layers passes the data to this layer. If input to the model is correct dimension, this function will automatically receive the correct dimensions. This is the reason no other layer checks the dimension of the input in forwarding/backwarding.

zhoonit: Maybe it could live inside `#ifdef DEBUG`

nntrainer/src/concat_layer.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  return ret;
+}
+
+void ConcatLayer::setProperty(const PropertyType type,
+                              const std::string &value) {
+  int status = ML_ERROR_NONE;
+
+  switch (type) {
+  case PropertyType::num_inputs: {
```
zhoonit, line 110: IMO, it is more natural to set num_inputs from parsing input_layers in and disallow setting for `num_inputs`

```ini
[concat1]
Type = concat
input_layers ="input_layer1, input_layer2"
```

jijoongmoon: This num_inputs may not required according to following implementation. I will change this as it implemented.

nntrainer/src/concat_layer.cpp
jijoongmoon: fixed. Thanks

nntrainer/src/concat_layer.cpp
jijoongmoon: fixed. Thanks

nntrainer/src/concat_layer.cpp
```diff
@@ -0,0 +1,126 @@
...
+    for (unsigned int i = 2; i < d.rank(); ++i) {
+      if (d[i] != dim[i])
+        throw std::runtime_error("Error: concat layer requires same "
+                                 "shape from  all input layers");
+    }
+#endif
+    concat_channel += dim.channel();
+  }
+
+  if (concat_channel != output_dim[0].channel()) {
```
kparichay, line 60: this check and computation should also move to DEBUG.

nntrainer/src/concat_layer.cpp
```diff
@@ -0,0 +1,126 @@
...
+sharedConstTensors ConcatLayer::backwarding(sharedConstTensors derivative,
+                                            int iteration) {
+  sharedConstTensors ret;
+  TensorDim d = derivative[0]->getDim();
+
+  unsigned int position = 0;
+  for (unsigned int idx = 0; idx < num_inputs; ++idx) {
+    TensorDim in_dim = input_dim[idx];
+    sharedTensor t = std::shared_ptr<Tensor>(new Tensor(in_dim),
+                                             std::default_delete<Tensor[]>());
```
kparichay, line 92: I am not if this should call `delete Tensor[]` than `delete Tensor`.

nntrainer/src/concat_layer.cpp
kparichay: @jijoongmoon I think this check is still needed at the initialize phase to ensure that the connected layers set by the model are compatible.

nntrainer/src/concat_layer.cpp
jijoongmoon: fixed. Thanks

nntrainer/src/concat_layer.cpp
jijoongmoon: fixed. Thanks

kparichay (approved):
LGTM

zhoonit (approved):
LGTM :)

PR #674, opened 2020-10-28
[Fix/Optimizer] Fix decay_rate application

zhoonit:
There was a bug that decay_rate were always applied when it is set to
default value. Becuase `decay_steps != -1` was evaluated to true.

This patch fixes the issue.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/optimizer_internal.h
+++ b/nntrainer/include/optimizer_internal.h
@@ -52,7 +52,7 @@ public:
    * @brief     Default Constructor of Optimizer Class
    */
   Optimizer(const OptType t, float lr, float decay_rate = 1.0f,
-            float decay_steps = -1.0f, float continue_train = false) :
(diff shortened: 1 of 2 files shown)


jijoongmoon (approved):
LGTM

kparichay (changes requested):
decay_steps should be made integer as it counts the number of steps which can't be float.
Please update `decay_steps` to unsigned int type to resolve this issue than this hot fix.
Also, add a check that decay_step must be positive, (decay_step of 0 is also an error)

zhoonit:
> decay_steps should be made integer as it counts the number of steps which can't be float.
> Please update `decay_steps` to unsigned int type to resolve this issue than this hot fix.
> Also, add a check that decay_step must be positive, (decay_step of 0 is also an error)

Fixed. Thank you for the suggestion :)

kparichay (approved):
LGTM

PR #675, opened 2020-10-28
Fix adam iteration start to 1 at model_test

zhoonit:
~From the paper, adam's time t start from 1. So other framework do like
`iteration + 1` because they start iteration from 0. However, nntrainer
start iteration from 1 so, iteration + 1 is not correct.~

~Also, there was a bug that iteration is not counted correctly for the
train data, this patch fixes the issue~

- Fix bug that epoch_idx is not saved for continue_train
- Fix iteration to start from 0 in model test

**Self evaluation:**

### DIFF
--- a/nntrainer/include/neuralnet.h
+++ b/nntrainer/include/neuralnet.h
@@ -80,13 +80,14 @@ public:
   NeuralNetwork() :
     batch_size(1),
     epochs(1),
+    epoch_idx(0),
+    iter(0),
     loss(0.0f),
     loss_type(LossType::LOSS_UNKNOWN),
     weight_initializer(WeightInitializer::WEIGHT_UNKNOWN),
     net_type(NetType::UNKNOWN),
     data_buffer(nullptr),
     continue_train(false),
-    iter(0),
     initialized(false),
     def_name_count(0),
     loadedFromConfig(false) {}
@@ -343,6 +344,10 @@ private:
 
   unsigned int epochs; /**< Maximum Epochs */
 
+  unsigned int epoch_idx; /**< Number of epoch_idx  */
+
+  unsigned int iter; /**< iterations trained */
+
   float loss; /**< loss */
 
   LossType loss_type; /**< Loss Function type */
@@ -362,8 +367,6 @@ private:
 
   bool continue_train; /**< Continue train from the previous state of optimizer
    and iterations */
-
-  uint64_t iter; /**< Number of iterations trained */
 
   bool initialized; /**< Network is initialized */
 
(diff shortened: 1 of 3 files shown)


jijoongmoon (approved):
LGTM

nntrainer/src/neuralnet.cpp
```diff
@@ -455,7 +455,7 @@ int NeuralNetwork::train_run() {
       if (data_buffer->getDataFromBuffer(nntrainer::BufferType::BUF_TRAIN,
                                          in->getData(), label->getData())) {
         try {
-          backwarding({in}, {label}, iter++);
+          backwarding({in}, {label}, epoch_idx);
```
kparichay, line 458: This is wrong. This loop runs for multiple iterations and backwarding needs that iteration count to apply the gradient.
This update passes epoch count which remains the same for all the iterations in the epoch.

nntrainer/include/neuralnet.h
```diff
@@ -321,6 +321,8 @@ class NeuralNetwork : public ml::train::Model {
 
   unsigned int epochs; /**< Maximum Epochs */
 
+  unsigned int epoch_idx; /**< Number of iterations trained */
```
kparichay, line 324: You are mixing up epochs and iterations here in naming and its description.
Please keep them separate. 
Epochs count the total number of times the whole dataset is looped over.
Iterations count the total number of batches given to the model (this must NOT reset at the end of each epoch).

kparichay (changes requested):
I think you are mixing epochs and iterations which are very different.

nntrainer/include/neuralnet.h
zhoonit: Got it. Thanks for the clarification. I'll fix it as per the comment.

Don't we need to save the epoch_idx though? I think it is natural that when the user stops training, we should save the `epoch_idx` and `iteration` and start from the epoch.

kparichay: Yes, that's why `iter` member variable in `NeuralNetwork` is being saved.

zhoonit:
> I think you are mixing epochs and iterations which are very different.

v2. 
- Fix bug that epoch_idx is not saved for continue_train
- Fix iteration to start from 0 in model test

Fixed. Thanks :)

nntrainer/src/neuralnet.cpp
```diff
@@ -344,6 +344,7 @@ void NeuralNetwork::saveModel() {
   std::ofstream model_file(save_path, std::ios::out | std::ios::binary);
   for (unsigned int i = 0; i < layers.size(); i++)
     layers[i]->save(model_file);
+  model_file.write((char *)&epoch_idx, sizeof(epoch_idx));
```
kparichay, line 347: do we really need to store/load both epoch_idx and iter?
epoch_idx can be derived from iter right?

kparichay: I think epoch_idx can be local to `train_run` function and be locally derived there itself.

zhoonit: Can we calculate epoch_idx when we are using `dataset_with_generator`?
I think we can calculate epoch_idx iff `iteration per epoch` is known to the `neuralnet`.

kparichay: May be cache iteration per epoch after the epoch then (initialized to 0)?
Then epoch_idx can be inferred, and it can be further used later to display progress too.

kparichay (approved):
LGTM

PR #677, opened 2020-10-28
[Fix/bn] Fix batchnormalize layer @open sesame 10/29 10:43

zhoonit:
Fix batchnormalizationLayer that epsilon is included toward saving
moving variacne that could lead to long term inaccuracy

**Self evaluation:**

### DIFF
--- a/nntrainer/src/bn_layer.cpp
+++ b/nntrainer/src/bn_layer.cpp
@@ -132,12 +132,13 @@ sharedConstTensors BatchNormalizationLayer::forwarding(sharedConstTensors in) {
     deviation = input.subtract(cmu);
 
     cvar = deviation.pow(2.0f).average(axes_to_reduce);
-    cvar.add_i(epsilon);
 
     mu.multiply_i(momentum);
     mu.add_i(cmu, 1 - momentum);
     var.multiply_i(momentum);
     var.add_i(cvar, 1 - momentum);
+
+    cvar.add_i(epsilon);
 
     invstd = cvar.pow(-0.5f);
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM

nntrainer/src/bn_layer.cpp
```diff
@@ -127,13 +127,14 @@ sharedConstTensors BatchNormalizationLayer::forwarding(sharedConstTensors in) {
...
 
     cvar = deviation.pow(2.0f).average(axes_to_reduce);
-    cvar.add_i(epsilon);
 
     mu.multiply_i(momentum);
     mu.add_i(cmu, 1 - momentum);
     var.multiply_i(momentum);
     var.add_i(cvar, 1 - momentum);
 
+    cvar.add_i(epsilon);
```
kparichay, line 136: This can avoid long-term inaccuracy but does cause short term inaccuracy.
If trained for only 1 iteration where cvar is 0, this change would make var 0 causing UB in evaluation mode as var remained zero.
If you are planning on doing this, you must add epsilon to var (line 144).
I urge to look at pytorch/tensorflow implementation style for this in order to avoid any differences or error cases as they have been tried and tested by a lot of people.

kparichay (changes requested):
This can lead to UB.

nntrainer/src/bn_layer.cpp
zhoonit: Good point. Checked bn implmentation in tf and added epsilon to the inference mode 👍

kparichay (approved):
LGTM now

PR #683, opened 2020-10-29
[backbone] Support for external backbone

kparichay:
Added support for external backbone with nnstreamer
    This is enabled for both ubuntu and Tizen
    This patch creates a nnstreamer layer to support running different kind of model files
    with nnstreamer single c-api
    
Commit - 2
    Added support for external backbone
    Also added a small add.tflite file for unittests
    trainable is not supported with external backbones for now

~Commit - 3
    Update transfer learning application to use backbone
    Now feature extractor has been removed from the application
    and dependency on tflite is also removed
    However, the caching of features from feature extractor is not yet supported.
    This results in the application to be slow to run full 1000 epochs on gbs
    Till caching is supported, application test from gbs build is removed~
 
See also #660 

Signed-off-by: Parichay Kapoor <pk.kapoor@samsung.com


Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -36,20 +34,9 @@
 #include <time.h>
 #include <unistd.h>
 
-#if defined(__TIZEN__)
 #include <nnstreamer-single.h>
 #include <nnstreamer.h>
 #include <nntrainer_internal.h>
-#else
-#include "tensorflow/contrib/lite/interpreter.h"
```
jijoongmoon, line 44: Don't need to delete this example. You could make another application for this I guess.

kparichay: This application change will have to revert as the nnstreamer_layer is not supported on android.
I will make a tflite_layer for backend and update example with that.

nntrainer/include/tensor_dim.h
```diff
@@ -110,6 +110,11 @@ class TensorDim {
     return {dim[1] * dim[2] * dim[3], dim[2] * dim[3], dim[3], 1};
   }
 
+  /**
+   * @breif squeeze non-zero dimensions to inner dimensions in-place
```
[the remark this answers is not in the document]
zhoonit: We had an offline discussion to change this method to rotate

nntrainer/src/nnstreamer_layer.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+int NNStreamerLayer::finalizeError(int status) {
+  if (in_res) {
+    ml_tensors_info_destroy(in_res);
+    in_res = nullptr;
+  }
+
+  if (out_res) {
+    ml_tensors_info_destroy(out_res);
+    in_res = nullptr;
```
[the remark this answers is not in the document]
kparichay: done.

nntrainer/include/tensor_dim.h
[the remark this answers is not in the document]
kparichay: So, this has been changed to `reverse()` for now.
NNStreamer reverses the dimension it reads from frameworks and stores it. 
This layer would just counter it.
Rotation of tensor dim is not sufficient but must be accompanied with tensor transpose (has been added as a todo).

nntrainer/src/model_loader.cpp
```diff
@@ -307,23 +331,28 @@ int ModelLoader::loadFromIni(std::string ini_file, NeuralNetwork &model,
...
+    if (backbone == unknown) {
+      status = loadLayerConfigIni(ini, layer, sec_name);
+      NN_INI_RETURN_STATUS();
+    } else if (fileIni(backbone)) {
+      status = loadBackboneConfigIni(ini, backbone, model, sec_name);
+      NN_INI_RETURN_STATUS();
       continue;
+    } else {
+      status = loadBackboneConfigExternal(ini, backbone, layer, sec_name);
+      NN_INI_RETURN_STATUS();
```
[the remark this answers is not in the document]
kparichay: moved out for two cases, except for the middle case which continues.

api/ccapi/include/layer.h
```diff
@@ -39,6 +40,7 @@ enum class LayerType {
   LAYER_ADDITION,                       /** Addition Layer type */
   LAYER_CONCAT,                         /** Concat Layer type */
   LAYER_LOSS,                           /** Loss Layer type */
+  LAYER_BACKBONE_NNSTREAMER,            /** Backbone using NNStreamer */
```
jijoongmoon, line 43: Are you going to add nntrainer backbone here as well?

nntrainer/include/layer_internal.h
```diff
@@ -213,7 +213,7 @@ class Layer : public ml::train::Layer {
    * @brief     set trainable for this layer
    * @param[in] train to enable/disable train
    */
-  void setTrainable(bool train) noexcept { trainable = train; }
+  virtual void setTrainable(bool train) { trainable = train; }
```
jijoongmoon, line 216: Could you elaborate the trainable?? Like conv, fc are trainable layers and input layer is not. Or does it means we are able to set conv with not-trainable?? For the nnstreamer layer, it is not trainable I guess. but if we are using nntrainer ini as a backbone, we can train. In this case, is it depends on user to set trainable or untrainable?? I think we haven't decided how to use this variable.

nntrainer/include/model_loader.h
```diff
@@ -80,6 +80,10 @@ class ModelLoader {
    * @param[in/out] layer layer to be loaded
    * @param[in] layer_name name of the layer to be loaded
    */
+  int loadLayerConfigIniCommon(dictionary *ini, std::shared_ptr<Layer> &layer,
```
jijoongmoon, line 83: Need to fix the doxygen and could you let me know what is the diff common and without common?

nntrainer/include/model_loader.h
```diff
@@ -94,6 +98,20 @@ class ModelLoader {
                             NeuralNetwork &model,
                             const std::string &backbone_name);
 
+  /**
+   * @brief     load backbone config from ini
+   * @param[in] ini dictionary containing the config
```
jijoongmoon, line 103: please update the brief. And I wonder that what external means.

nntrainer/include/nnstreamer_layer.h
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	nnstreamer_layer.h
+ * @date	26 October 2020
+ * @brief	This is class to encapsulate nnstreamer as a layer of Neural Network
+ * @see		https://github.com/nnstreamer/nntrainer
```
jijoongmoon, line 8: Could you check this tab?

nntrainer/include/nnstreamer_layer.h
```diff
@@ -0,0 +1,115 @@
...
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
```
jijoongmoon, line 10: ditto.

nntrainer/include/nnstreamer_layer.h
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * @copydoc Layer::forwarding(sharedConstTensors in)
+   */
+  sharedConstTensors forwarding(sharedConstTensors in);
+
+  /**
+   * @copydoc Layer::backwarding(sharedConstTensors in, int iteration)
+   */
+  sharedConstTensors backwarding(sharedConstTensors in, int iteration);
+
```
jijoongmoon, line 59: I wonder.. can we do backwarding with nnstreamer?

nntrainer/src/model_loader.cpp
```diff
@@ -307,21 +331,25 @@ int ModelLoader::loadFromIni(std::string ini_file, NeuralNetwork &model,
...
      * If this section is a backbone, load backbone section from this
      * @note The order of backbones in the ini file defines the order on the
      * backbones in the model graph
      */
     const char *backbone =
       iniparser_getstring(ini, (sec_name + ":Backbone").c_str(), unknown);
-    if (backbone != unknown) {
-      loadBackboneConfigIni(ini, backbone, model, sec_name);
+    if (backbone == unknown) {
+      status = loadLayerConfigIni(ini, layer, sec_name);
```
jijoongmoon, line 345: IMO, backbone is also layer type so that, we can handle like one of the layer. I mean instead of add loadBAckbonConfigIni separately, we could do this inside loadLayerConfigIni with LayerType keyword.

api/ccapi/include/layer.h
kparichay: nntrainer backbone does not need a type.
it directly modifies the graph (layer vector for now).

nntrainer/include/nnstreamer_layer.h
kparichay: As NNStreamer_layer class extends Layer, where backwarding is pure virtual, this function has to be defined.
So, it defined here and throws not supported if called.

nntrainer/src/model_loader.cpp
kparichay: With loadLayerConfigIni, the layer is always added to the model. This is not desired for backbone with ini. Further with ini backbone, with more support to come for setting scaling factors, etc, lets keep it separate for now, and will merge once all features are supported.
For backbone with external frameworks, `loadBackboneConfigExternal` is actually just a wrapper for `loadLayerIni` which sets the modelfile property.

nntrainer/include/layer_internal.h
kparichay: > For the nnstreamer layer, it is not trainable I guess. but if we are using nntrainer ini as a backbone, we can train. In this case, is it depends on user to set trainable or untrainable?? 

trainable is by default set to true for a layer created with nntrainer ini file.
the exception is backbone ini, where the default trainable is false (must be set true with `trainable` property of the backbone in ini or API).
For external framework layers (tflite/nnstreamer), setting trainable to true will throw exception, setting to false is only allowed.

> Could you elaborate the trainable?? Like conv, fc are trainable layers and input layer is not. Or does it means we are able to set conv with not-trainable?? 

Yes, this property can be set individually for the layer from ini. There is no limitation to the layers which can be set to trainable/non-trainable. Setting a layer to trainable does not essentially mean that that layer has some variables which are going to be trained (flatten layer allows setting trainable to true/false but does not effect the behavior of the model training/inference).

nntrainer/include/model_loader.h
kparichay: done.

nntrainer/include/model_loader.h
kparichay: done.

nntrainer/src/model_loader.cpp
kparichay: Both the loadConfig for backboneExternal and Layer can be merged if we update the ini style.
Current style:
```ini
[block1]
backbone: tmp.ini/tflite
```

New style
```ini
[block1]
type: backbone
modelfile: tmp.ini/tflite
```

However, IMO, first is more direct and intuitive.

api/ccapi/include/layer.h
jijoongmoon: > nntrainer backbone does not need a type.
> it directly modifies the graph (layer vector for now).

That means if user add sub .ini file in backbone, then it parse and initialize?

nntrainer/include/layer_internal.h
jijoongmoon: > Setting a layer to trainable does not essentially mean that that layer has some variables which are going to be trained (flatten layer allows setting trainable to true/false but does not effect the behavior of the model training/inference).
So, if user set trainable as false for conv layer, it just skip update weight gradient. I think we haven't implement this so far. But we are going to. Am I right?

jijoongmoon (approved):
Nice discussion about how we support backbone network. LGTM

nntrainer/include/layer_internal.h
zhoonit: >So, if user set trainable as false for conv layer, it just skip update weight gradient. I think we haven't implement this so far. But we are going to. Am I right?

We have implemented this. We are skipping updating the gradient (in Optimizer::apply_gradients)

zhoonit (approved):
LGTM :)

api/ccapi/include/layer.h
kparichay: yes, the backbone ini file is parsed and a backbone graph is made.
Then the backbone graphs is merged into the original graph.

nntrainer/include/nnstreamer_layer.h
kparichay: done.

nntrainer/include/nnstreamer_layer.h
kparichay: done.

nntrainer/include/layer_internal.h
kparichay: IMO also this is already supported. In some cases where its not supported, I will add it :)

kparichay: @jijoongmoon Currently we only conv2d, fc and bn layer which updates weights.
Check line 353 for conv2d_layer.cpp, 129 for fc_layer.cpp. It is missing for bn_layer, which will be added soon.

zhoonit: I'll add if you are okay

kparichay: @zhoonit Thanks but I already added.

Issue #469, opened 2020-08-24
Support skip connections with user API


kparichay:
- [x] support skip connections with ini-file (https://github.com/nnstreamer/nntrainer/issues/239#issuecomment-659126498 and https://github.com/nnstreamer/nntrainer/issues/239#issuecomment-653440135)
- [x] update C-API to be able to add such models

jijoongmoon:
done with #712

kparichay:
Let's not close this till we have a graph example with API and able to create a skip connection with API.

PR #684, opened 2020-10-29
[ Layer ] Add input_layers keyword

jijoongmoon:
This PR includes enabling 'input_layers' keyword.
With this keyword, we could specify layer's input tensor.

. Added skip in parse_util to remove space and '[',']' in string and
split with ',' delimiter.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/parse_util.h
+++ b/nntrainer/include/parse_util.h
@@ -25,6 +25,7 @@
 #ifdef __cplusplus
 
 #include <iostream>
+#include <regex>
 #include <string>
 #include <vector>
 
@@ -178,6 +179,14 @@ int getKeyValue(std::string input_str, std::string &key, std::string &value);
 const char *getValues(std::vector<int> values, const char *delimiter = ",");
 
 int getValues(int n_str, std::string str, int *value);
+
+/**
+ * @brief     split string into vector with delimiter regex
+ * @param[in] str string
+ * @param[in] reg regular expression to use as delimiter
+ * @retval    output string vector
+ */
+std::vector<std::string> split(const std::string &s, std::regex &reg);
 
 /**
  * @brief     print instance info. as <Type at (address)>
--- a/nntrainer/src/layer.cpp
+++ b/nntrainer/src/layer.cpp
@@ -198,6 +198,16 @@ void Layer::setProperty(const PropertyType type, const std::string &value) {
       bias_initializer = (WeightInitializer)parseType(value, TOKEN_WEIGHT_INIT);
     }
     break;
+  case PropertyType::input_layers:
+    if (!value.empty()) {
+      std::regex reg("\\,+");
+      std::vector<std::string> concat_layers = split(value, reg);
+      // TODO set num_inputs properly
+      num_inputs = 1;
+      if (concat_layers.size() > 1)
+        num_inputs = concat_layers.size();
+    }
+    break;
   default:
     std::string msg =
       "[Layer] Unknown Layer Property Key for value " + std::string(value);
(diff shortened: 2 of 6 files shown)


zhoonit:
Related to #469

test/unittest/unittest_nntrainer_graph.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "nntrainer_test_util.h"
+
+namespace initest {
+typedef enum {
+  LOAD = 1 << 0, /**< should fail at load */
+  INIT = 1 << 1, /**< should fail at init */
+} IniFailAt;
+};
+
+class nntrainerGraphTest
```
zhoonit, line 25: Any reason for the duplicated test? I think this can go into `modelfile` test

jijoongmoon: More test cases are coming which related with graph like initialize, compiling, forwarding and so on.

zhoonit: If the purpose is to use Ini, I think it should use `IniTestWrapper`. This test checks if model file load & compiles and initiate without any problem which is not related to graph

jijoongmoon: This is just skeleton at his stage. I will add more test cases accordingly.

zhoonit (approved):
LGTM

nntrainer/include/parse_util.h
```diff
@@ -37,6 +38,8 @@ namespace nntrainer {
     }                              \
   } while (0)
 
+#define NUM_SKIP_CHAR 3
```
kparichay, line 41: lets move this to cpp file

nntrainer/src/layer.cpp
```diff
@@ -198,6 +198,15 @@ void Layer::setProperty(const PropertyType type, const std::string &value) {
       bias_initializer = (WeightInitializer)parseType(value, TOKEN_WEIGHT_INIT);
     }
     break;
+  case PropertyType::input_layers:
+    if (!value.empty()) {
+      std::regex reg("\\,+");
```
kparichay, line 203: make "\\,+" as a const string?

nntrainer/src/layer.cpp
```diff
@@ -198,6 +198,15 @@ void Layer::setProperty(const PropertyType type, const std::string &value) {
...
+      std::vector<std::string> concat_layers = split(value, reg);
+      num_inputs = 1;
+      if (concat_layers.size() > 1)
```
kparichay, line 206: what does check means? If the parse string reveals size is 0, then we should set num_inputs to that and init would give it out as a error. But here, it seems that in case concat_layers.size() is 0, then num_inputs=1 but then what does it point to?

nntrainer/src/layer.cpp
jijoongmoon: Want to use ',' as a delimiter and others like '+','-','*','%' will be added.

nntrainer/src/layer.cpp
jijoongmoon: yes. you are right. But there is layer which not need input like input_layers. It is different story if we are using databuffer as an layer, then input_layer need input_layers = 'databuffer_name'. That means every layer needs layer_inputs. That's what I plan to implement. set num_input=1 is just temporary and once every thing is implemented properly, I will fix it.

nntrainer/include/parse_util.h
jijoongmoon: done. Thanks.

nntrainer/src/layer.cpp
kparichay: Yes, thats why declare it as a string as a separate variable.
Also, might consider moving all of this to parse_util later.

nntrainer/src/layer.cpp
kparichay: Then lets add a `TODO/FIXME` here.

kparichay (approved):
LGTM

PR #690, opened 2020-11-03
[concat] Move validity checks to init

kparichay:
Move the dimension checks for validity of the various inputs
for concat layer are moved to initialize().
They are kept back in forwarding() under DEBUG conditional.

**Self evaluation:**

### DIFF
--- a/nntrainer/include/tensor_dim.h
+++ b/nntrainer/include/tensor_dim.h
@@ -99,7 +99,8 @@ public:
   bool isEmpty() const { return len == 0; }
   unsigned int rank() const;
 
-  unsigned int &operator[](unsigned int index);
+  unsigned int &operator[](const unsigned int index);
+  const unsigned int &operator[](const unsigned int index) const;
 
   /**
    * @brief Calculate standard strides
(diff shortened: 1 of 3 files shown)


0001-Application-Update-application-to-use-backbone.patch
```diff
@@ -0,0 +1,377 @@
+From c455b5ca2c394f0be271afe54e7a53ceaa5f7f54 Mon Sep 17 00:00:00 2001
+From: Parichay Kapoor <pk.kapoor@samsung.com>
```
jijoongmoon, line 2: Could you check this patch file??

kparichay: my bad, removed.

nntrainer/include/tensor_dim.h
```diff
@@ -100,6 +100,7 @@ class TensorDim {
   unsigned int rank() const;
 
   unsigned int &operator[](unsigned int index);
+  unsigned int operator[](unsigned int index) const;
```
zhoonit, line 103:
```suggestion
  const unsigned int &operator[](const unsigned int index) const;
```

jijoongmoon (approved):
LGTM

nntrainer/include/tensor_dim.h
kparichay: good suggestion, done.

zhoonit (approved):
LGTM :)

Issue #152, opened 2020-06-12
[ Proposal ] Plug-able & Modularize Layer   


jijoongmoon:
Make Layer Plugable & Modularize to be able to re-configure at compile time

kparichay:
Extend this to compile a layer individually as a shared object, and nntrainer can take a path to load all such layers.
Direct use case is external layers - so that external layers don't have to shipped directly with nntrainer but can still be supported if needed.
Check https://github.com/nnstreamer/nntrainer/pull/693#discussion_r517718088 thread for more.

PR #693, opened 2020-11-03
[backbone] Added native support for tflite backbone

kparichay:
Added native support for tflite backbone
    The unittests are verified for tflite backbone as tflite backbone
    is preferred over nnstreamer backbone.
    Interfacing directly with tflite API allows directly using tensor memory
    for forwarding, avoid memcpy incurred with nnstreamer backbone.
    
TfLite takes input shape of format NHWC.
    However, nntrainer takes the shape NCHW.
    This will be resolved later with a transpose operator later.

See also #660
    
**Self evaluation:**
    

jni/Android.mk
```diff
@@ -15,11 +15,38 @@ $(warning INIPARSER SRC is going to be downloaded!)
...
+include $(CLEAR_VARS)
+
+NNTRAINER_JNI_ROOT := $(NNTRAINER_ROOT)/jni
+
+TENSORFLOW_VERSION := 1.13.1
+
+ifndef TENSORFLOW_ROOT
+ifneq ($(MAKECMDGOALS),clean)
+$(warning TENSORFLOW_ROOT is not defined!)
+$(warning TENSORFLOW SRC is going to be downloaded!)
```
jijoongmoon, line 32: I think this should go to the Application or install this depending an option to use tflite backbone or not. I think we should not have tight dependency with tflite.

nntrainer/include/tflite_layer.h
```diff
@@ -0,0 +1,102 @@
...
+#define __TENSORFLOW_LITE_H__
+#ifdef __cplusplus
+
+#include <layer_internal.h>
+#include <tensor.h>
+
+#include <tensorflow/contrib/lite/interpreter.h>
+#include <tensorflow/contrib/lite/kernels/register.h>
+#include <tensorflow/contrib/lite/model.h>
+// #include <tensorflow/contrib/lite/string_util.h>
```
zhoonit, line 24: intended?

Applications/KNN/jni/Android.mk
```diff
@@ -9,10 +9,13 @@ endif
 
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/../../../
+endif
```
zhoonit, line 12: Good catch.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -103,189 +93,12 @@ void array_set_zero(float *data, size_t num_elem) {
...
- * @brief     Get Feature vector from tensorflow lite
- *            This creates interpreter & inference with ssd tflite
+ * @brief     Load input image data from bmp files and normalize to float
  * @param[in] filename input file path
- * @param[out] feature_input save output of tflite
+ * @param[out] feature_input normalized images
  */
-void getInputFeature(const TFLiteData &tflite_data, const std::string filename,
-                     float *feature_input) {
+void getInputFeature(const std::string filename, float *feature_input) {
```
zhoonit, line 101: Since some people are referring to the code...
```
void getInputFeature(const std::string &filename, float *feature_input) {
```

jni/Android.mk
kparichay: if tflite backbone is enabled to be supported, then in order to compile tflite backbone layer, nntrainer will have to compile with tflite for now.
Later, we can make pluginable layers which would allow compiling layers independently and plugigng them into the nntrainer. However, this is not directly possible with current design.

nntrainer/include/tflite_layer.h
kparichay: will remove.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
kparichay: Will be done.

meson.build
```diff
@@ -132,6 +132,11 @@ if get_option('enable-nnstreamer-backbone')
   add_project_arguments('-DENABLE_NNSTREAMER_BACKBONE=1', language:['c','cpp'])
 endif
 
+tflite_dep = dependency('tensorflow-lite', required: false)
```
zhoonit, line 135: I wonder if we have to limit the tflite version?

nntrainer/src/model_loader.cpp
```diff
@@ -249,17 +249,27 @@ int ModelLoader::loadBackboneConfigExternal(dictionary *ini,
                                             const std::string &backbone_config,
                                             std::shared_ptr<Layer> &layer,
                                             const std::string &backbone_name) {
-#ifdef ENABLE_NNSTREAMER_BACKBONE
+  LayerType type = LayerType::LAYER_UNKNOWN;
+
+#if defined(ENABLE_NNSTREAMER_BACKBONE)
```
zhoonit, line 254: Maybe later... how about we delegate && detect inside backbone layer and use only LayerType::LAYER_BACKBONE. I think we can say LAYER_BACKBONE has two implementations.

nntrainer/include/tflite_layer.h
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+#ifndef __TENSORFLOW_LITE_H__
+#define __TENSORFLOW_LITE_H__
+#ifdef __cplusplus
+
+#include <layer_internal.h>
+#include <tensor.h>
+
+#include <tensorflow/contrib/lite/interpreter.h>
```
zhoonit, line 21: Maybe later... How about using pimple idiom for this layer to isloate tflite dependency?

nntrainer/src/tflite_layer.cpp
```diff
@@ -0,0 +1,147 @@
...
+  Layer::copy(l);
+
+  std::shared_ptr<TfLiteLayer> from = std::static_pointer_cast<TfLiteLayer>(l);
+  this->modelfile = from->modelfile;
+}
+
+sharedConstTensors TfLiteLayer::backwarding(sharedConstTensors derivative,
+                                            int iteration) {
+  throw exception::not_supported(
+    "Backwarding is not supported for nnstreamer layer");
```
zhoonit, line 145:
```suggestion
    "Backwarding is not supported for nnstreamer layer");
```

nntrainer/include/tflite_layer.h
kparichay: I was thinking more along the lines of #152  because with pimpl, we still need to compile together.

zhoonit (approved):
LGTM. Please merge after reviewing the minor comments :)

nntrainer/src/tflite_layer.cpp
kparichay: done.

nntrainer/include/tflite_layer.h
kparichay: done.

jni/Android.mk
kparichay: Added an option to disable tflite dependency in Android.mk for android build.
I am not able to read this option from meson_options.txt
Please let me know if you know how to.

meson.build
kparichay: I am following nnstreamer here. They support version checks for tf1/2.
I will add them as a separate PR as it will not be that direct for android.

jijoongmoon (approved):
LGTM

PR #712, opened 2020-11-05
[ Graph ] Neural Network Graph

jijoongmoon:
In this PR, NetworkGraph Class is introduced and compile method is
implemented.
During the compile process, it takes layers vector which is created by
model_loader and create required layers like Activation, Flatten,
Concat or Addition Layers. Also it modify inputs relatively and
Calculate the order of calculation. So far, only serial processing is
supported.

**Self evaluation:**


nntrainer/include/network_graph.h
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
+ *
+ * @file	network_graph.h
+ * @date	19 Oct 2020
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is Convolution Layer Class for Neural Network
```
anyj0527, line 10: Might require renaming?

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
+ *
+ * @file	network_graph.h
+ * @date	19 Oct 2020
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is Convolution Layer Class for Neural Network
```
anyj0527, line 10: Also in here

nntrainer/include/network_graph.h
jijoongmoon: fixed. Thanks

nntrainer/src/network_graph.cpp
jijoongmoon: fixed. Thanks

nntrainer/include/neuralnet.h
```diff
@@ -129,6 +130,8 @@ class NeuralNetwork : public ml::train::Model {
    */
   int loadFromConfig(std::string config);
 
+  int compile();
```
kparichay, line 133: I think it might be good to expose this with c++ API?

nntrainer/include/network_graph.h
```diff
@@ -0,0 +1,81 @@
...
+  int addLossLayer(const LossType loss_type);
+
+  int setEdge();
+
+private:
+  void topologicalSortUtil(unsigned int ith, bool visited[],
+                           std::stack<LayerNode> &Stack);
+
+  unsigned int num_node;
+  std::vector<std::list<LayerNode>> adj;
```
kparichay, line 72: lets add some comments for there variables usage and also add doxygen for the member functions of this class.

kparichay, line 72: @jijoongmoon I think this is important to consider at start as this would require big changes, but how about using `std::unordered_map<LayerNode, std::list<LayerNode>>` for adj? This makes searching and traversing the graph easier. You can also consider `std::unordered_map<std::string, std::list<LayerNode>>` where string in the layer name.

nntrainer/include/network_graph.h
```diff
@@ -0,0 +1,81 @@
...
+  std::vector<LayerNode> Sorted;
+  std::set<std::string> layer_names;
```
kparichay, line 74: so these things will move out of the neuralnetwork class right?

nntrainer/include/network_graph.h
```diff
@@ -0,0 +1,81 @@
...
+
+  void ensureName(std::shared_ptr<Layer> layer, const std::string &prefix,
+                  bool force_rename);
+
+  int setGraphNode(std::vector<std::shared_ptr<Layer>> layers,
+                   const LossType loss_type);
+
+  int realizeMultiInputType(Layer &current);
+
+  int realizeActivationType(Layer &current,
```
kparichay, line 57: Also this is a suggestion to modularize this - separate this into two classes -  let says `GraphCore` which only does graph related stuff like `addEdge`, `addNode`, `topologicalSort` and make this class templated.
Then `NetworkGraph` class will have an instance of `GraphCore<LayerNode>` and will directly use it, making it easier to manage graph and these realization functions.
This will also make commits smaller and easier to review.

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
...
+    visited[i] = false;
+
+  for (unsigned int i = 0; i < num_node; ++i) {
+    if (visited[i] == false) {
+      topologicalSortUtil(i, visited, Stack);
+    }
+  }
+
+  while (Stack.empty() == false) {
+    Sorted.push_back(Stack.top());
```
kparichay, line 82: Does sorted need to be a class member? I think it just needs to be a local member for this function.

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+int NetworkGraph::realizeMultiInputType(Layer &current) {
+  int status = ML_ERROR_NONE;
+  if (current.num_inputs == 1)
+    return ML_ERROR_NONE;
+
+  if (current.num_inputs > 1) {
+    std::shared_ptr<Layer> layer =
+      nntrainer::createLayer(LayerType::LAYER_CONCAT);
+    ensureName(layer, current.getName());
```
kparichay, line 132: Instead of having `ensureName()` in various realization functions, lets move it to a common location in `addLayerNode()`?

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
...
+  adj.push_back(l);
+  num_node++;
+}
+
+void NetworkGraph::topologicalSortUtil(unsigned int ith, bool visited[],
+                                       std::stack<LayerNode> &Stack) {
+  visited[ith] = true;
+
+  std::list<LayerNode>::iterator i;
+  for (i = adj[ith].begin(); i != adj[ith].end(); ++i) {
```
kparichay, line 47: Lets add a note here to make this function iterative here later.
Recursive functions (with big depth like 100 which is easily possible here) arent good for production code.

nntrainer/include/network_graph.h
```diff
@@ -0,0 +1,81 @@
...
+
+  void topologicalSort();
+
+  void addLayerNode(std::shared_ptr<Layer>);
+
+  unsigned int getNumNode() { return num_node; }
+
+  LayerNode &getLayerNode(unsigned int ith);
+
+  LayerNode &getLayerNode(const std::string layer_name);
```
zhoonit, line 47: Maybe reference operator?
```suggestion
  LayerNode &getLayerNode(const std::string &layer_name);
```

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
...
+      topologicalSortUtil(i, visited, Stack);
+    }
+  }
+
+  while (Stack.empty() == false) {
+    Sorted.push_back(Stack.top());
+    Stack.pop();
+  }
+
+  for (unsigned int i = 0; i < Sorted.size(); ++i) {
```
zhoonit, line 86: is this left intended?

+edit) There are other couts I guess it is for demo purpose for now :)

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
...
+    }
+  }
+
+  throw std::invalid_argument("Cannot find Layer");
+}
+
+void NetworkGraph::topologicalSort() {
+  std::stack<LayerNode> Stack;
+
+  bool *visited = new bool[num_node];
```
zhoonit, line 71: should this be deleted after line 79?

nntrainer/src/network_graph.cpp
```diff
@@ -0,0 +1,364 @@
...
+      if (layers[i]->input_layers[j] == current.getName()) {
+        layers[i]->input_layers[j] = layer->getName();
+      }
+    }
+  }
+
+  return ML_ERROR_NONE;
+}
+
+int NetworkGraph::addLossLayer(const LossType loss_type) {
```
zhoonit, line 229: Maybe, we have to deal with multiple outputs and heterogeneous loss type.

nntrainer/include/neuralnet.h
jijoongmoon: Yes. I will. Thanks.

nntrainer/include/network_graph.h
jijoongmoon: Yes. that's good idea. but I think we can use it after we complete the whole processes.

nntrainer/include/network_graph.h
jijoongmoon: These are not exposed by neural network I guess. after compile, neural network just call forwarding or backwarding of graph I guess.

nntrainer/include/network_graph.h
jijoongmoon: I will do it later. Cause the model create layer in model_loader, and there is layers vector. at this time, each layers are not initialized. In stead of create layers as an vector, we could create GraphCore directly.

nntrainer/src/network_graph.cpp
jijoongmoon: It is needed to store the order of layers as an 1D vector.

nntrainer/src/network_graph.cpp
jijoongmoon: yes. good. Thanks.

nntrainer/src/network_graph.cpp
jijoongmoon: I haven't add any node or documentation for this pr. I will add.  ( It's just basic algorithm for tree structure. )

nntrainer/src/network_graph.cpp
jijoongmoon: This is just to check the result after ordering the network graph. We will not have this at the end.

nntrainer/src/network_graph.cpp
jijoongmoon: yes. multiple output will come in PR #714

nntrainer/src/network_graph.cpp
jijoongmoon: yes. right. I will.

nntrainer/include/network_graph.h
kparichay: Changing from vector to unordered_map would require big changes, so I think its much easier to switch as early as possible.

nntrainer/include/network_graph.h
kparichay: Graph must expose functions which provides an iterator view of the graph. NeuralNetwork will choose which functions to call or directly make the call.

nntrainer/include/network_graph.h
kparichay: Yes, later works too but this decoupling is needed.

nntrainer/include/network_graph.h
jijoongmoon: > Graph must expose functions which provides an iterator view of the graph. NeuralNetwork will choose which functions to call or directly make the call.

I think some of functions only to display the graph structure. There are many layers included by network graph. Like out layer, addition layers etc. So, eventually yes, we have to expose graph. But not all I say. And Also I'm not sure what you mean choose which functions?? I mean what do you expect which functions to choose?

nntrainer/include/network_graph.h
jijoongmoon: > Yes, later works too but this decoupling is needed.

Agreed. Which is oop does.

nntrainer/include/network_graph.h
kparichay: By choose I meant if forwarding to be run or backwarding to be run will be chosen by NeuralNetwork. Depending on train/inference, NeuralNetwork class can choose/decide.

jijoongmoon:
TODO
  
- [ ] After make node list of graph, we have to find root. (That means it should be the only one input for now.).
      Need to support multiple input and support search.
- [ ] unordered_map ( I'm not sure though )
- [ ] More Unittest 
- [ ] Proper Example ( for now mini resnet is tested with one iteration )
- [ ] Check Making Graph with API (C and C++) for non-ini initialization
- [ ] etc.

nntrainer/graph/network_graph.cpp
```diff
@@ -212,6 +212,15 @@ int NetworkGraph::realizeActivationType(
   layer->input_layers.clear();
   layer->input_layers.push_back(current.getName());
 
+  layer->num_outputs = current.num_outputs;
+  layer->output_layers.clear();
```
kparichay, line 222: add a check that the number of output layers here are 1?

nntrainer/graph/network_graph.cpp
```diff
@@ -212,6 +212,15 @@ int NetworkGraph::realizeActivationType(
   layer->input_layers.clear();
   layer->input_layers.push_back(current.getName());
 
+  layer->num_outputs = current.num_outputs;
```
kparichay, line 220: add a check that the number of output layers here are 1?

nntrainer/graph/network_graph.cpp
```diff
@@ -276,17 +290,20 @@ void NetworkGraph::setOutputLayers(std::vector<std::shared_ptr<Layer>> layers) {
 
   for (unsigned int idx = 0; idx < layers.size(); ++idx) {
     unsigned int count = 0;
+    std::cout << layers[idx]->getName() << " : ";
```
kparichay: update to ml_logd?

nntrainer/graph/network_graph.cpp
```diff
@@ -276,17 +290,20 @@ void NetworkGraph::setOutputLayers(std::vector<std::shared_ptr<Layer>> layers) {
...
     for (unsigned int i = 0; i < layers.size(); ++i) {
       if (layers[i]->getName() == layers[idx]->getName())
         continue;
       for (unsigned int j = 0; j < layers[i]->input_layers.size(); ++j) {
         if (layers[i]->input_layers[j] == layers[idx]->getName()) {
-          layers[idx]->output_layers.push_back(layers[i]->input_layers[j]);
+          layers[idx]->output_layers.push_back(layers[i]->getName());
+          std::cout << layers[idx]->output_layers[count] << ", ";
```
kparichay: same

nntrainer/graph/network_graph.cpp
```diff
@@ -409,8 +426,18 @@ int NetworkGraph::setEdge() {
   for (unsigned int i = 0; i < adj.size(); ++i) {
 
     std::list<LayerNode>::iterator iter;
-    std::cout << i << " : " << getLayerNode(i).layer->getName() << std::endl;
-    ;
+    std::cout << i << " : " << getLayerNode(i).layer->getName() << " ("
```
kparichay: more debug code?

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+      topologicalSortUtil((*i).index, visited, Stack);
+  }
+
+  Stack.push(getLayerNode(ith));
+}
+
+LayerNode &NetworkGraph::getLayerNode(unsigned int ith) {
+
+  std::list<LayerNode>::iterator iter;
+  for (unsigned int i = 0; i < num_node; ++i) {
```
kparichay, line 57: node->index is made using num_node and its location is at the ith index itself in the `adj` vector.
Lets make this `getLayerNode` as a constant time operation than O(N) operation.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+    }
+  }
+
+  throw std::invalid_argument("Cannot find Layer");
+}
+
+void NetworkGraph::topologicalSort() {
+  std::stack<LayerNode> Stack;
+
+  bool *visited = new bool[num_node];
```
kparichay, line 81: lets use `vector<bool>` or add it as a TODO here.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+  for (unsigned int i = 0; i < num_node; ++i)
+    visited[i] = false;
```
kparichay, line 83: lets use memset.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+  delete[] visited;
+
+  while (Stack.empty() == false) {
+    Sorted.push_back(Stack.top());
+    Stack.pop();
+  }
+
+  for (unsigned int i = 0; i < Sorted.size(); ++i) {
+    std::cout << Sorted[i].layer->getName() << " ";
```
kparichay: debug code? use ml_logd

nntrainer/graph/network_graph.cpp
jijoongmoon: We might need to change the order ( not yet supported ). Let's assume that user write loss for the first element in ini file but set properly as an last layer. In that case object might be in first but the index must be the last.

nntrainer/graph/network_graph.cpp
jijoongmoon: I think it is better the checking in setOutputLayer(). I will.

nntrainer/graph/network_graph.cpp
jijoongmoon: done.

nntrainer/graph/network_graph.cpp
jijoongmoon: done.

nntrainer/graph/network_graph.cpp
kparichay: as per line 35, index of the node is assigned at the time of node creation and it inserted in the adj list at that moment.
adj list order also remains same as it represent the graph data. So, I dont understand how a node in adj list and order of layer in ini file are related.

nntrainer/graph/network_graph.cpp
kparichay: in setOutputLayer also possible, but for activation, only 1 output and 1 input is supported. So, it must be ensured that when setting the output here, the number of output is 1.

nntrainer/graph/network_graph.cpp
```diff
@@ -77,33 +78,24 @@ LayerNode &NetworkGraph::getSortedLayerNode(unsigned int ith) {
 
 void NetworkGraph::topologicalSort() {
   std::stack<LayerNode> Stack;
+  std::vector<bool> visited(num_node);
```
kparichay, line 81: std::vector<bool> visited(num_node) = {0};

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+  layer->setName(name);
+}
+
+int NetworkGraph::realizeMultiInputType(Layer &current) {
+  int status = ML_ERROR_NONE;
+  if (current.num_inputs == 1)
+    return ML_ERROR_NONE;
+
+  if (current.num_inputs > 1) {
+    std::shared_ptr<Layer> layer = nntrainer::createLayer("concat");
```
kparichay: how does the graph choose between concat and addition layer in case of multi input?
if its NYI, please add a TODO.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+  adj[ith].push_back(node);
+}
+
+void NetworkGraph::addLayerNode(std::shared_ptr<Layer> layer) {
+  std::list<LayerNode> l;
+  LayerNode *node = new LayerNode();
+
+  node->layer = layer;
+  node->index = num_node;
+
```
kparichay, line 35: please add ensureName() here.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+  if (current.getType() == "flatten") {
+    ml_loge(
+      "It is not allowed to realize flatten layer, possibly flatten layer is "
+      "added right after flatten");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  std::shared_ptr<Layer> layer = nntrainer::createLayer("flatten");
+
+  ensureName(layer, current.getName());
```
kparichay, line 166: remove this. addLayerNode() should has it.

kparichay (commented):
The changes you make based on the comment, please make it in the corresponding commit than in the latest commit. Otherwise, it is very difficult to track the new change as this PR is already huge.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+  layer->num_inputs = 1;
+  layer->input_layers.clear();
+  layer->input_layers.push_back(current.getName());
+
+  addLayerNode(layer);
+
+  for (unsigned int i = 0; i < layers.size(); ++i) {
```
kparichay, line 174: please create a function like `updateNameInLayers(std::vector<std::shared_ptr<Layer>> layers, const std::string &name)` which does this. This is currently repeatedly written in many places.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+    current.num_inputs = 1;
+    current.input_layers.clear();
+    current.input_layers.push_back(layer->getName());
+    addLayerNode(layer);
+  }
+  return status;
+}
+
+int NetworkGraph::realizeFlattenType(
+  Layer &current, std::vector<std::shared_ptr<Layer>> layers) {
```
kparichay, line 151: std::vector<std::shared_ptr<Layer>> **&**layers

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+        layers[i]->input_layers[j] = layer->getName();
+      }
+    }
+  }
+
+  return ML_ERROR_NONE;
+}
+
+int NetworkGraph::realizeActivationType(
+  Layer &current, std::vector<std::shared_ptr<Layer>> layers) {
```
kparichay, line 186: std::vector<std::shared_ptr<Layer>> **&** layers

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+    /// ActivationType::ACT_NONE does not need realization
+    return ML_ERROR_NONE;
+  }
+
+  if (num_node == 0) {
+    ml_loge("layer is empty");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (current.getType() == "activation") {
```
kparichay, line 199: if (current.getType() == ActivationLayer::type) {

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+void NetworkGraph::setOutputLayers(std::vector<std::shared_ptr<Layer>> layers) {
+
+  for (unsigned int idx = 0; idx < layers.size(); ++idx) {
+    unsigned int count = 0;
+    for (unsigned int i = 0; i < layers.size(); ++i) {
+      if (layers[i]->getName() == layers[idx]->getName())
+        continue;
+      for (unsigned int j = 0; j < layers[i]->input_layers.size(); ++j) {
+        if (layers[i]->input_layers[j] == layers[idx]->getName()) {
+          layers[idx]->output_layers.push_back(layers[i]->input_layers[j]);
```
kparichay: This sets the `layers[idx]->getName()` as the output layers of `layers[idx]` which I think is a bug.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+  layer->num_inputs = 1;
+  layer->input_layers.clear();
+  layer->input_layers.push_back(input_str);
+
+  addLayerNode(layer);
+
+  return ML_ERROR_NONE;
+}
+
+void NetworkGraph::setOutputLayers(std::vector<std::shared_ptr<Layer>> layers) {
```
kparichay, line 303: std::vector<std::shared_ptr<Layer>> **&** layers

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+          count++;
+        }
+      }
+    }
+    layers[idx]->num_outputs = count;
+  }
+}
+
+int NetworkGraph::realizeMultiOutputType(
+  Layer &current, std::vector<std::shared_ptr<Layer>> layers) {
```
kparichay, line 336: std::vector<std::shared_ptr<Layer>> **&** layers

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+  int status = ML_ERROR_NONE;
+  if (current.num_outputs == 1)
+    return ML_ERROR_NONE;
+
+  if (current.num_outputs > 1) {
```
kparichay, line 341: this check is redundant.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+  layer->setName(name);
+}
+
+int NetworkGraph::realizeMultiInputType(Layer &current) {
+  int status = ML_ERROR_NONE;
+  if (current.num_inputs == 1)
+    return ML_ERROR_NONE;
+
+  if (current.num_inputs > 1) {
```
kparichay, line 133: redundant check.

nntrainer/graph/network_graph.cpp
kparichay: I think it should be `nntrainer::createLayer(layer->getType())`

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+void NetworkGraph::topologicalSort() {
+  std::stack<LayerNode> Stack;
+
+  bool *visited = new bool[num_node];
+  for (unsigned int i = 0; i < num_node; ++i)
+    visited[i] = false;
+
+  for (unsigned int i = 0; i < num_node; ++i) {
+    if (visited[i] == false) {
+      topologicalSortUtil(i, visited, Stack);
```
kparichay, line 87: topologicalSortUtil() call here MUST only be made for nodes which represent input nodes. Otherwise, this implementation would result in wrong solution if input layer is not the node at 0.

nntrainer/graph/network_graph.cpp
```diff
@@ -281,6 +294,9 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
   layer->input_layers.clear();
   layer->input_layers.push_back(input_str);
 
+  std::shared_ptr<LossLayer> temp = std::dynamic_pointer_cast<LossLayer>(layer);
```
kparichay, line 294: use createLoss directly than using createLayer at line 284.

nntrainer/graph/network_graph.cpp
```diff
@@ -302,9 +318,18 @@ void NetworkGraph::setOutputLayers(std::vector<std::shared_ptr<Layer>> layers) {
...
+      layers[idx]->num_outputs = count;
+      layers[idx]->output_dim.resize(count);
+    }
     std::cout << std::endl;
   }
+
+  if (layers.back()->num_outputs == 0) {
+    layers.back()->num_outputs = 1;
+    layers.back()->output_dim.resize(1);
+    layers.back()->output_layers.push_back("exit");
```
kparichay: what is the significance of layers.back()? Also what is `exit` here?

nntrainer/graph/network_graph.cpp
```diff
@@ -392,6 +417,13 @@ int NetworkGraph::setGraphNode(std::vector<std::shared_ptr<Layer>> layers,
   return status;
 }
 
+void NetworkGraph::setNumNetBufferSize() {
+  for (unsigned int i = 0; i < Sorted.size(); ++i) {
+    Sorted[i].input.resize(Sorted[i].layer->input_layers.size());
```
kparichay: Sorted[i].input is a Tensor. I wonder how this is working?

kparichay (commented):
It seems like layer->input_layers and layer->output_layers should store ptr than names, as this is a requirement at a lot of places, or maybe a function to solve this as this is making a lot of simple code convoluted in this PR.

nntrainer/graph/network_graph.cpp
```diff
@@ -490,4 +496,43 @@ int NetworkGraph::setEdge() {
...
+  for (auto const &layer_node : Sorted) {
+    layer_node.layer->setBatch(batch_size);
+  }
+}
+
+sharedConstTensors NetworkGraph::forwarding(sharedConstTensors input) {
+  for (unsigned int i = 0; i < Sorted.size() - 1; ++i) {
+    LayerNode &layer_node = Sorted[i];
+    if (layer_node.layer->getType() == LayerType::LAYER_IN) {
+      layer_node.layer->forwarding(input);
```
kparichay, line 498: why LAYER_IN not following the same semantics as other layers?

nntrainer/graph/network_graph.cpp
```diff
@@ -490,4 +496,43 @@ int NetworkGraph::setEdge() {
...
+
+  return out;
+}
+
+void NetworkGraph::backwarding(sharedConstTensors output, int iteration) {
+
+  for (unsigned int i = Sorted.size() - 1; i > 0; i--) {
+    LayerNode &layer_node = Sorted[i];
+    if (layer_node.layer->getType() == LayerType::LAYER_LOSS) {
+      layer_node.layer->backwarding(iteration, output);
```
kparichay, line 520: same, why LAYER_LOSS not following the same semantics as other LAYERS?

nntrainer/layers/bn_layer.cpp
```diff
@@ -177,16 +175,15 @@ BatchNormalizationLayer::backwarding(sharedConstTensors derivative,
   dx_2.subtract_i(deviation.divide(cvar).multiply(
     deviation.multiply(deriv).sum(axes_to_reduce)));
 
-  Tensor dx = dx_2.multiply(dx_1);
+  Tensor &dx = net_input[0]->grad;
+  dx = dx_2.multiply(dx_1);
   dx.divide_i(N);
```
kparichay, line 180: is it ok to modify this in-place?
also where this grad is initialized?

nntrainer/layers/flatten_layer.cpp
```diff
@@ -41,21 +41,14 @@ int FlattenLayer::initialize() {
...
 
-sharedConstTensors FlattenLayer::forwarding(sharedConstTensors in) {
-  input = *in[0];
-  hidden = input;
-
-  hidden.reshape(output_dim[0]);
-
-  return {MAKE_SHARED_TENSOR(hidden)};
+void FlattenLayer::forwarding(sharedConstTensors in) {
+  memcpy(net_hidden[0]->var.getData(), net_input[0]->var.getData(),
```
kparichay, line 45: is it possible to set net_hidden[0] = net_input[0] here?

nntrainer/layers/layer_internal.h
```diff
@@ -103,16 +108,15 @@ class Layer : public ml::train::Layer {
    * @param[in] in List of Input Tensors taken by this layer
    * @retval    List of Output Tensors
    */
-  virtual sharedConstTensors forwarding(sharedConstTensors in) = 0;
+  virtual void forwarding(sharedConstTensors in = {}) = 0;
```
kparichay, line 114: This is really good. Later, input argument can also be removed.

nntrainer/layers/layer.cpp
```diff
@@ -61,6 +61,22 @@ void Layer::setBatch(unsigned int batch) {
     output_dim[idx].setTensorDim(0, batch);
 }
 
+std::vector<Tensor> Layer::getHidden() {
```
kparichay, line 64: getHidden -> getOutput?

nntrainer/layers/layer.cpp
```diff
@@ -61,6 +61,22 @@ void Layer::setBatch(unsigned int batch) {
...
+  std::vector<Tensor> ret;
+  for (unsigned int i = 0; i < num_outputs; ++i) {
+    ret.push_back(net_hidden[i]->var);
+  }
+  return ret;
+}
+
+std::vector<Tensor> Layer::getGradient() {
```
kparichay, line 72: getGradient -> getDerivatives?

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
...
 }
 
+sharedConstTensors Layer::forwarding_with_val(sharedConstTensors input,
+                                   sharedConstTensors in) {
+
+  for(unsigned int i=0;i<num_inputs;++i){
+    net_input[i]->var = *input[i];
+  }
+
+  if (num_outputs == 0)
```
kparichay, line 110: is this possible? (I assume this function can only run once layer is initialized).

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
...
+    throw("invalid number of outputs");
+
+  if (num_outputs != net_hidden.size())
```
kparichay, line 113: same, this would be verified in initialize()

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
   this->num_outputs = l->num_outputs;
 }
 
+sharedConstTensors Layer::forwarding_with_val(sharedConstTensors input,
```
kparichay, line 103: what is the difference between input and in?

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
...
+
+  for(unsigned int i =0; i<num_outputs;++i){
+    out.push_back(MAKE_SHARED_TENSOR(net_hidden[i]->var));
+  }
+
+  return out;
+}
+
+sharedConstTensors Layer::backwarding_with_val(int iteration,
+                                      sharedConstTensors deriv,
```
kparichay: same, whats the difference between deriv and in?

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
...
+                                      sharedConstTensors in) {
+
+  for (unsigned int i = 0; i < num_outputs; ++i) {
+    net_hidden[i]->grad = *deriv[i];
+  }
+
+  if (num_inputs == 0)
```
kparichay, line 135: ditto

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
...
+    throw("invalid number of inputs");
+
+  if (num_inputs != net_input.size())
```
kparichay, line 138: ditto

nntrainer/layers/layer_internal.h
```diff
@@ -301,6 +311,29 @@ class Layer : public ml::train::Layer {
     output_dim.resize(num_outputs);
   }
 
+  std::vector<Tensor> getHidden();
+
+  std::vector<Tensor> getGradient();
+
+  void resizeNetInput(unsigned int size){net_input.resize(size);}
```
kparichay: do all these netInput/netOutput related functions need to be public? I think this information is available at via properties and be set at initialize(). For example, resizeNetInput() is same as setProperty("num_inputs=..");

nntrainer/models/neuralnet.cpp
```diff
@@ -199,6 +199,7 @@ int NeuralNetwork::compile() {
   NN_RETURN_STATUS();
 
   model_graph.topologicalSort();
+  setBatchSize(batch_size);
```
kparichay: any reason to set batch size here than at train directly? This changes a lot of semantics which we earlier set.

nntrainer/models/model_loader.cpp
```diff
@@ -279,6 +279,12 @@ int ModelLoader::loadBackboneConfigIni(dictionary *ini,
     graph[0]->setProperty(Layer::PropertyType::input_shape, input_shape);
```
kparichay, line 279: I am not sure here but is graph[0] still valid here?

nntrainer/graph/network_graph.cpp
jijoongmoon: > as per line 35, index of the node is assigned at the time of node creation and it inserted in the adj vector at that moment.
> adj vector order also remains same as it represent the graph data. So, I dont understand how a node in adj vector and order of layer in ini file are related.
> In my understanding, the order of the node->index is just the order in which nodes are created and represents no other information (same for order of nodes in adj vector)

Yes. you can see it is now. What I'm saying that will be changed soon when we do TODO After make node list of graph, we have to find root. --> When we do find the root then, I think index will be re-ordered. That means idx is not same any more.

nntrainer/graph/network_graph.cpp
jijoongmoon: Yes. I will. 
> Also move this call out of here to the original function. This function should just realize flatten layer. The handling of side-effects should take place in the caller function.

I don't think it it be out of here. This is only for adding flatten layer and If it is not possible, then it is the error of adding flatten layer. ( If it has error during this addLayerNode, then it shouldn't go further I guess )

nntrainer/graph/network_graph.cpp
kparichay: The index of the nodes remains same as they dont represent any ordering. Root of the graph can be any node and can be multiple (will have to be supported some day). So, the order of nodes in adj list should not have a meaning. If needed, there can be a different list which has meaning to order of nodes like Sorted.

nntrainer/graph/network_graph.cpp
kparichay: As any error in creating graph should not change the graph state, or leave graph in un-defined state, giving error before or after does not change. The behavior of updating name is associated with the caller - where a subgraph is to be added - but has nothing to do with realization of the layer.

jijoongmoon: > As any error in creating graph should not change the graph state, or leave graph in un-defined state, giving error before or after does not change. The behavior of updating name is associated with the caller - where a subgraph is to be added - but has nothing to do with realization of the layer.

For this PR, realization means adding Layer which associated with input and output layers setting. we make split the function with realization, adding layer with setting input and output layer then. But this can be done later PR.

nntrainer/graph/network_graph.cpp
jijoongmoon: > The index of the nodes remains same as they dont represent any ordering. Root of the graph can be any node and can be multiple (will have to be supported some day). So, the order of nodes in adj list should not have a meaning. If needed, there can be a different list which has meaning to order of nodes like Sorted.

Let's leave as it is. we need it now. We will see and if we don't need it , then we remove later.

nntrainer/graph/network_graph.cpp
jijoongmoon: That's what I'm saying. input layer must be root and node 0. If we have to support multiple inputs then, we need update graph sort.

nntrainer/graph/network_graph.cpp
jijoongmoon: I think ensureName could be outside. This is only place ensureName effective if it is in addLayerNode. Others like Activation, MultiOutput and Input, the name is required before addLayerNode. Otherwise, it is redundant check.

nntrainer/graph/network_graph.cpp
jijoongmoon: let's keep it outside.

nntrainer/graph/network_graph.cpp
jijoongmoon: not work.

nntrainer/graph/network_graph.cpp
jijoongmoon: done.

nntrainer/graph/network_graph.cpp
jijoongmoon: done.

nntrainer/graph/network_graph.cpp
jijoongmoon: done.

nntrainer/graph/network_graph.cpp
jijoongmoon: done

nntrainer/graph/network_graph.cpp
jijoongmoon: done

nntrainer/graph/network_graph.cpp
jijoongmoon: Actually I intended not to include *_layer.h file here. But once I includes them to support like ActivationLayer::type, I can create loss layer directly. done.

nntrainer/graph/network_graph.cpp
jijoongmoon: done

nntrainer/graph/network_graph.cpp
jijoongmoon: done

nntrainer/graph/network_graph.cpp
jijoongmoon: done

nntrainer/graph/network_graph.cpp
jijoongmoon: Input needs input data. That is the layer which set the first data. If you are look at the assignMem(), it does not assign input memory buffer as well. It just take memory assigned from outside.

nntrainer/graph/network_graph.cpp
jijoongmoon: There is two kind of forwarding. one is to calculate the loss and without calculation loss. In order to calculate the loss, it just take input (which is return from previous layer) and calculate old fashion way. I though it's not a big deal. We can change this later pr though.

nntrainer/layers/bn_layer.cpp
jijoongmoon: net_input[0]->grad is the return derivative and will be overwritten. No need to initialized. I don't see any problems here. Let me know if it is.

nntrainer/layers/flatten_layer.cpp
jijoongmoon: If we use like this, the dimension is also copyed. So I don't think it is possible.

nntrainer/layers/layer.cpp
jijoongmoon: I followed old verables. But I guess we could change. done.

nntrainer/layers/layer.cpp
jijoongmoon: done.

nntrainer/layers/layer.cpp
jijoongmoon: my bad, I think in must be deleted. done

nntrainer/layers/layer.cpp
jijoongmoon: I was thinking set everything based 1 is not good at that time. Currently we just set as 1 with no other questions. I guess it is not good. but we still doing with 1 so, it is not possible. I will delete it. done.

nntrainer/layers/layer.cpp
jijoongmoon: done.

nntrainer/layers/layer.cpp
jijoongmoon: done.

nntrainer/layers/layer.cpp
jijoongmoon: we need it.

nntrainer/layers/layer_internal.h
jijoongmoon: but for the loss calculation, we need to get the label. that why in is still there.

nntrainer/models/model_loader.cpp
jijoongmoon: getGraph is just get the layers which generated by model loader. It is not the model_grap.Sorted. So graph[0] is still valid. and also we need to consider that the first layer of ini might be the first of layers. ( I guess we need to change )

nntrainer/graph/network_graph.cpp
kparichay: Then please add this a note that not only just 1 input layer is supported but also that the input layer must be the first layer to be specified.

Also, if only 1 input layer is supported, then do we need the loop at line 75. All the nodes must be accessible starting from the input node. If not, we have a problem than starting topologicalSortUitl from there.

nntrainer/layers/flatten_layer.cpp
kparichay: yes, i meant like below:
```
Tensor temp = net_input[0];
temp.setDim(net_hidden[0].getDim());
net_hidden[0] = temp;
```

nntrainer/models/model_loader.cpp
kparichay: yeah, lets add a TODO here.

nntrainer/layers/flatten_layer.cpp
jijoongmoon: done.

nntrainer/models/model_loader.cpp
jijoongmoon: done

nntrainer/graph/network_graph.cpp
```diff
@@ -404,6 +436,16 @@ LayerNode &NetworkGraph::getLayerNode(const std::string &layer_name) {
   throw std::invalid_argument("Cannot find Layer");
 }
 
+LayerNode &NetworkGraph::getSortedLayerNode(const std::string &layer_name) {
```
zhoonit, line 462: Q. is this different from `getLayerNode(const std::string &layer_name)` ?

nntrainer/graph/network_graph.cpp
```diff
@@ -294,6 +294,12 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
   layer->input_layers.clear();
   layer->input_layers.push_back(input_str);
 
+  if (layer->output_layers.size() == 0) {
+    layer->num_outputs = 1;
+    layer->output_dim.resize(1);
+    layer->output_layers.push_back("exit");
```
zhoonit: Q. What is this exit for?

nntrainer/graph/network_graph.cpp
zhoonit: I think it is better for graph has the reference to inputs and set first before first forwarding happens to unify the api, reduce (dynamic) type checking. Maybe later patch

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -212,63 +210,61 @@ void NodeWatcher::read(std::ifstream &in) {
...
-  ss << "backward failed at " << node->getName() << " at iteration "
+  ss << "backward failed at " << node.layer->getName() << " at iteration "
      << iteration;
   std::string err_msg = ss.str();
 
-  nntrainer::sharedConstTensors out = node->backwarding(deriv, iteration);
+  std::vector<nntrainer::Tensor> out = node.layer->getGradient();
+
   if (should_verify) {
+    verify(out[0], expected_dx, err_msg);
```
zhoonit, line 264: Any reason for the order change?
IMO gradient should be verified beforehand,

nntrainer/graph/network_graph.cpp
jijoongmoon: That's why I left the note like
```After make node list of graph, we have to find root. (That means it should be the only one input for now.).
Need to support multiple input and support search.```

File: nntrainer/layers/loss_layer.cpp [hunk 59]
```diff
@@ -42,8 +42,33 @@ int LossLayer::initialize() {
...
 
+sharedConstTensors LossLayer::forwarding_with_val(sharedConstTensors in,
+                                         sharedConstTensors label) {
+
+  for(unsigned int i=0;i<num_inputs;++i){
+    net_input[i]->var = *in[i];
+  }
+
+  if (num_outputs == 0)
+    throw("invalid number of outputs");
```
zhoonit: I guess throwing const char * is unintended
throw std::invalid_argument("");

nntrainer/layers/layer.cpp
```diff
@@ -84,6 +100,74 @@ void Layer::copy(std::shared_ptr<Layer> l) {
...
+sharedConstTensors Layer::backwarding_with_val(int iteration,
+                                      sharedConstTensors deriv,
+                                      sharedConstTensors in) {
+
+  for (unsigned int i = 0; i < num_outputs; ++i) {
+    net_hidden[i]->grad = *deriv[i];
+  }
+
+  if (num_inputs == 0)
+    throw("invalid number of inputs");
```
zhoonit: I guess throwing const char * is unintended, I guess this is deleted though

nntrainer/graph/network_graph.cpp
```diff
@@ -140,7 +139,7 @@ int NetworkGraph::realizeMultiInputType(Layer &current) {
     return ML_ERROR_NONE;
 
   if (current.num_inputs > 1) {
-    std::shared_ptr<Layer> layer = nntrainer::createLayer("concat");
+    std::shared_ptr<Layer> layer = nntrainer::createLayer("addition");
```
zhoonit: Could you check if this has to be addition only? I think this should be either concat or addition.

nntrainer/graph/network_graph.cpp
```diff
@@ -386,7 +385,13 @@ int NetworkGraph::setGraphNode(std::vector<std::shared_ptr<Layer>> layers,
     Layer &l = *layers[i];
     ml_logd("layer name: %s", l.getName().c_str());
 
-    if (l.getType() != "addition" || l.getType() != "concat") {
+    if (l.getType() == "input") {
+      l.num_inputs = 1;
+      l.input_layers.clear();
+      l.input_layers.push_back("data");
```
zhoonit: Q. what is data?

nntrainer/graph/network_graph.cpp
jijoongmoon: I don't think it is bug. Let's say ```a --> b --> c```. then the output of a is must be  b. and input of b must be a. So.. ```layers[0]->getName(): a == layer[1]->input_layers[0] : b --> true``` and set ```layers[0]->output_layers.push_back(layer[1]->input_layers[0] : b)```. then the output of a is b and input of b is a. I think everything set properly. I guess. Let me know if I'm wrong.

also ``` layers[idx]->output_layers.push_back(layers[i]<-- must be the string. )```

test/unittest/unittest_nntrainer_modelfile.cpp
```diff
@@ -366,7 +387,7 @@ TEST(nntrainerIniTest, backbone_p_04) {
 /**
  * @brief Ini file unittest matching model with and without backbone
  */
-TEST(nntrainerIniTest, backbone_p_05) {
+TEST(nntrainerIniTest, DISABLED_backbone_p_05) {
```
zhoonit: Any reason this remains DISABLED?

nntrainer/models/neuralnet.cpp
```diff
@@ -212,6 +214,11 @@ int NeuralNetwork::initialize() {
     return ML_ERROR_NOT_SUPPORTED;
   }
 
+  if (!compiled) {
```
zhoonit, line 168: How about change `isInitializable()` -> `isCompilable()`.

nntrainer/graph/network_graph.cpp
jijoongmoon: different. but we could merge. getLayerNode get the layer from layers, but getSortedLayerNode get the layer form Sorted which includes layers which does not exist in layers.

jijoongmoon: I think it is the same: just for the getLayerNode with layer_name. if it is idx, then it differ. I will fix.

nntrainer/graph/network_graph.cpp
jijoongmoon: I think this is the place need to fix. Input Layer is not the only one takes input. Conv can takes and others too. Need to fix later PR.

nntrainer/models/neuralnet.cpp
jijoongmoon: done.

nntrainer/models/neuralnet.cpp
```diff
@@ -747,17 +748,37 @@ int NeuralNetwork::extendGraph(GraphType graph, std::string prefix) {
...
      * and ensure it is unique in this new graph
      */
+    std::string org_name = prefix + layer->getName();
     ensureName(layer, prefix, true);
+    sub_in_out.insert(std::make_pair(org_name, layer->getName()));
+
+    for (unsigned int i = 0; i < layer->input_layers.size(); ++i) {
+      if (sub_in_out.find(prefix + layer->input_layers[i]) !=
+          sub_in_out.end()) {
+        layer->input_layers[i] = sub_in_out[prefix + layer->input_layers[i]];
```
zhoonit, line 654: IMO, there seems few points to improve.

1. Renewing input_layers should be done after the loop of creating `sub_in_out` is done. `sub_in_out` might not contain the name it should have (will lead to an error for some cases)
2. For the same reason, not finding the key should throw an error.
3. `sub_in_out` should be a variable inside this function. so that it is cleared when the function starts. And the variable is only used inside this function
4. keys don't need to have the prefix added. It is redundant.

test/unittest/unittest_nntrainer_models.cpp
jijoongmoon: fixed.

nntrainer/utils/parse_util.cpp
```diff
@@ -485,6 +485,9 @@ std::vector<std::string> split(const std::string &s, std::regex &reg) {
 }
 
 bool istrequal(const std::string &a, const std::string &b) {
+  if (a.size() != b.size())
```
zhoonit, line 488: :+1:

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+  int status = ML_ERROR_NONE;
+
+  setOutputLayers(layers);
+
+  for (unsigned int i = 0; i < layers.size(); ++i) {
+    Layer &l = *layers[i];
+    ml_logd("layer name: %s", l.getName().c_str());
+
+    if (l.getType() != "addition" || l.getType() != "concat") {
```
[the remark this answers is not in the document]
zhoonit: Done in later commit

test/unittest/unittest_nntrainer_modelfile.cpp
zhoonit: This is reverted after adding backbone support

nntrainer/graph/network_graph.cpp
zhoonit: This was fixed at 266bb3e. Resolving this issue.

nntrainer/graph/network_graph.cpp
zhoonit: Found the usage, resolving the comment

nntrainer/layers/layer_internal.h
jijoongmoon: This is only for the testing. We made unittest cases layer by layer. not based on graph. If we are using graph, then we don't need. graph will set the net_input at compile time. However, we need to set somehow if we are using just for the layer. 
these are same for the
``` resizeNetInput, resizeNetOutput, getNumInputs, getNumOutputs, setInputBuffer, setOutputBuffer ```
we need to enable these func for the test only.

File: nntrainer/models/neuralnet.cpp [hunk 68]
```diff
@@ -560,6 +464,34 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X) {
   return out;
 }
 
+int NeuralNetwork::assignMem() {
```
zhoonit, line 412: I think the strategy is good though, IMO, `assignMem` should be private and used only internally because this is not part of the API. Meaning that if someone is creating a new type of model, they shouldn't care about this method.

nntrainer/models/neuralnet.cpp
jijoongmoon: changed.

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+void NetworkGraph::addEdge(unsigned int ith, LayerNode node) {
+  if (ith > num_node - 1)
+    throw std::invalid_argument("Exceed total number of layer");
+
+  adj[ith].push_back(node);
+}
+
+void NetworkGraph::addLayerNode(std::shared_ptr<Layer> layer) {
+  std::list<LayerNode> l;
+  LayerNode *node = new LayerNode();
```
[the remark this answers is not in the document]
jijoongmoon: done.

nntrainer/graph/network_graph.cpp
zhoonit: Kinda self answered.

nntrainer/graph/network_graph.cpp
jijoongmoon: just dummy for the later use.

nntrainer/layers/loss_layer.cpp
jijoongmoon: fixed.

nntrainer/layers/layer.cpp
zhoonit: Fixed

nntrainer/graph/network_graph.cpp
zhoonit: Will recomment on the latest patch

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+    current.num_outputs = 1;
+    current.output_layers.clear();
+    current.output_layers.push_back(layer->getName());
+    addLayerNode(layer);
+  }
+  return status;
+}
+
+int NetworkGraph::setGraphNode(std::vector<std::shared_ptr<Layer>> layers,
```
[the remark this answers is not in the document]
zhoonit: Fixed

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,536 @@
...
+
+  for (unsigned int idx = 0; idx < layers.size(); ++idx) {
+    unsigned int count = 0;
+    for (unsigned int i = 0; i < layers.size(); ++i) {
+      if (istrequal(layers[i]->getName(), layers[idx]->getName()))
+        continue;
+      for (unsigned int j = 0; j < layers[i]->input_layers.size(); ++j) {
+        if (istrequal(layers[i]->input_layers[j], layers[idx]->getName())) {
+          layers[idx]->output_layers.push_back(layers[i]->getName());
+          count++;
```
zhoonit, line 329: How about using `layers[idx]->output_layers.size()` and remove count to reduce complexity?

nntrainer/graph/network_graph.cpp
```diff
@@ -0,0 +1,536 @@
...
+
+  layer->setName(name);
+}
+
+int NetworkGraph::realizeMultiInputType(Layer &current) {
+  int status = ML_ERROR_NONE;
+  if (current.num_inputs == 1)
+    return ML_ERROR_NONE;
+
+  std::shared_ptr<Layer> layer = nntrainer::createLayer("addition");
```
zhoonit, line 159: I wonder if this should remain "addition" as there are two ways merge multiple inputs (concat and addition).

nntrainer/models/neuralnet.cpp
jijoongmoon: 1. sub_in_out is not renewed.
for 2, if it is not found then definitely, it is not the error.
about 3, we need it global. using inside this function only is not what I indented. It must be the unique throughout the network.
about 4. due to the reason 3, prefix is required.

nntrainer/models/neuralnet.cpp
jijoongmoon: yes. but like always, unittest is matter. on second thought, they might have to care about this method to set memory differently.

nntrainer/graph/network_graph.cpp
jijoongmoon: fixed.

nntrainer/graph/network_graph.cpp
jijoongmoon: fixed. concat is not supported in this PR. We need to define token and parse and set concat. Maybe another PR.

nntrainer/graph/network_graph.cpp
kparichay: ok.

nntrainer/graph/network_graph.cpp
kparichay: @zhoonit In this case, as we use getType() which is layers internal function, and compare it against its own type, is istrequal required? When comparing externally provided string, its understandable, but when working with internal strings, given its allocated by the framework itself, I think direct should also work.

nntrainer/graph/network_graph.cpp
kparichay: thanks.

nntrainer/graph/network_graph.cpp
kparichay: yes.

nntrainer/layers/bn_layer.cpp
kparichay: Ignore my comment, this is correct.
I was confused.

nntrainer/layers/layer.cpp
kparichay: I thought net_hidden is the structure managed by the graph.

nntrainer/models/neuralnet.cpp
zhoonit: I wonder if `sub_in_out` has some use outside of this function. Currently, there is no occurrence of `sub_in_out` and there is no other way to access the entry outside of this function (since key is processed inside this function and not saved elsewhere).

Plus, there are some issues when this stays global.

1. we need to ensure `org_name` is unique which seems not guaranteed, line 767 gaurantee's it is unique, so we cannot be sure if it is unique before line 767.

and for 🌵, there will be a hidden error if input layer is processed after the current layer, which is hard to find the cause of the problem.
e.g)
```ini

[a]
input_layers=b

[b]
...
```
ini sections does not have guaranteed order so, even if [b] is before [a], it might be processed before [a]. So I think not finding new names for the input layers should throw

So updating `sub_in_out` should be done like

```cpp
for (auto& layer : graph) {
 update sub_in_out
}

for(auto& layer: graph) {
 update layer information
}
```

nntrainer/graph/network_graph.cpp
zhoonit: We know this will work for internal usage, however semantic states type is case insensitive, if the part we assumed it would stay internal happens to have external usage, it will be very hard to catch the error. 
Just my opinion though, If we start to tell internal and external types, we might be ending up into a situation hard to deal with. 

If there is no visible performance improvement, I think it is better to stick to semantics.
If we have a performance problem, I think we'd better change semantics (this is not easy though) or at least strictly restrict the usage.

nntrainer/models/neuralnet.cpp
zhoonit: ok

nntrainer/layers/layer.cpp
jijoongmoon: > I thought net_hidden is the structure managed by the graph.

Yes. you are right. but this function is only the unittest which is not use graph structure. Directly call with input.

nntrainer/models/neuralnet.cpp
jijoongmoon: I think ini does not guarantee problem is not just for this. we have to handle this sepeartely. sub_in_out need be global. we have to consider that used same sub_graph repeatly with different section name.

zhoonit: Understood the reason behind the code. Thanks

zhoonit (approved):
Great work! LGTM :)

kparichay (approved):
L Great TM

PR #714, opened 2020-11-05
[ Layer ] Add Output Layer to support multiple output

jijoongmoon:
Currently there is no way to support multiple output.
However it is not neccessary to be explicit support.
Instead if user use "output_layers = layername0, layername1", then nntrainer
automatically add output layer. ( comparing ouput dimension and input
layer's dimension, we could decide it is addition or split operation. )

Also, if certain layer is specify as an input layer from mutiple
layers, then it should be added implicitly.

TODO: support split operation

**Self evaluation:**

### DIFF
--- a/nntrainer/utils/parse_util.cpp
+++ b/nntrainer/utils/parse_util.cpp
@@ -277,7 +277,9 @@ unsigned int parseType(std::string ll, InputType t) {
  * moving_variance_initializer = 22
  * gamma_initializer = 23
  * beta_initializer = 24
- * input_layers = 25
+ * modelfile = 25
+ * input_layers = 26
+ * output_layers = 27
  *
  * InputLayer has 0, 1, 2, 3 properties.
  * FullyConnectedLayer has 1, 4, 6, 7, 8, 9 properties.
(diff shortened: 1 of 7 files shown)


kparichay:
> Instead if user use "output_layers = layername0, layername1", then nntrainer
automatically add output layer. ( comparing ouput dimension and input
layer's dimension, we could decide it is addition or split operation. )

Just a clarification - `output_layers` essentially creates a copy of the current output which are then passed to different layers? Like in the example here, the output of this layer is given as input to `layername0` and `layername1`?

jijoongmoon:
> > Instead if user use "output_layers = layername0, layername1", then nntrainer
> > automatically add output layer. ( comparing ouput dimension and input
> > layer's dimension, we could decide it is addition or split operation. )
> 
> Just a clarification - `output_layers` essentially creates a copy of the current output which are then passed to different layers? Like in the example here, the output of this layer is given as input to `layername0` and `layername1`. Is this right?

Yes. or you can use directly like in resnet.

zhoonit (approved):
LGTM :)

nntrainer/utils/parse_util.cpp
```diff
@@ -278,14 +278,15 @@ unsigned int parseType(std::string ll, InputType t) {
  * gamma_initializer = 23
  * beta_initializer = 24
  * input_layers = 25
+ * output_layers = 26
```
kparichay, line 281: something strange about this count - its 27 in Layer::PropertyType
please check
I think modelfile is missing here in docs.

kparichay (approved):

PR #720, opened 2020-11-06
[Fix] Throw when read/save fails

### LINKED ISSUE #698: Object::read should check eof
Currently, object::read  (layer::read, or others...) is not checking eof, it acts like it read successfully even though it failed to read

zhoonit:
Currently, read save didn't throw when failing on those operation.
From this patch, if an object fails to read, it throws an exception
to prevent false positive on test result

**Self evaluation:**

resolves #698


nntrainer/layers/layer.cpp
```diff
@@ -38,6 +38,11 @@ void Layer::setActivation(ActivationType acti) {
 }
 
 int Layer::setOptimizer(std::shared_ptr<Optimizer> opt) {
+  if (!opt) {
+    this->opt = opt;
```
kparichay, line 42: this is wrong for adam. If two layers share the optimizer, then the corresponding values inside the optimizer are also shared which creates issue. Thats why, a new optimizer is always created.

nntrainer/models/neuralnet.cpp
```diff
@@ -358,15 +363,23 @@ void NeuralNetwork::saveModel() {
...
+  NeuralNetwork tmp(*this);
+
   std::ifstream model_file(save_path, std::ios::in | std::ios::binary);
   for (unsigned int i = 0; i < layers.size(); i++)
-    layers[i]->read(model_file);
+    tmp.layers[i]->read(model_file);
   if (continue_train) {
-    model_file.read((char *)&epoch_idx, sizeof(epoch_idx));
-    model_file.read((char *)&iter, sizeof(iter));
+    checkedRead(model_file, (char *)&tmp.epoch_idx, sizeof(epoch_idx),
```
kparichay, line 373: Even without continue_train, this value must be read. Otherwise the values read after this will be wrong.

packaging/nntrainer.spec
```diff
@@ -284,8 +284,7 @@ pushd build
 
 rm -rf model.bin
 TF_APP=Applications/TransferLearning/Draw_Classification
-cp ../${TF_APP}/res/ssd_mobilenet_v2_coco_feature.tflite .
-./${TF_APP}/jni/nntrainer_training ../${TF_APP}/res/Training.ini ../${TF_APP}/res
+$(pwd)/${TF_APP}/jni/nntrainer_training $(pwd)/../${TF_APP}/res/Training.ini $(pwd)/../${TF_APP}/res
```
kparichay, line 287: add $(pwd) to TF_APP itself

nntrainer/models/neuralnet.cpp
zhoonit: Good point. I'll fix it

nntrainer/layers/layer.cpp
zhoonit: Reverted.

nntrainer/models/neuralnet.cpp
zhoonit: Done

packaging/nntrainer.spec
zhoonit: Done

nntrainer/models/neuralnet.cpp
```diff
@@ -433,6 +444,11 @@ int NeuralNetwork::train(std::vector<std::string> values) {
 int NeuralNetwork::train_run() {
   int status = ML_ERROR_NONE;
 
+  if (!continue_train) {
```
kparichay, line 447: IMO this should move to `readModel`.
If the user calls the train function twice manually, then it should start from the previously loaded epoch in the model.

zhoonit: I think we have to clarify what `continue_train` means.

There seems several cases can be expected when continue_train is **false**

1. Strong reset: at the start of the `train_run()` we reset everything (weights and others...)
2. Weak reset: calling train_run() twice will continue from where the user has stopped. (but the epoch and arguably iteration(?) is changed)
3. No reset(for the current session): callling train_run() trains from stopped iteration and epoch. 

IMO `2` is what I think turning off continue_train. For the case of `3`, as long as `train_run()` finishes. second `train_run()` is essentially noop. even though continue_train is off. 

Our code states, 
```cpp
  bool continue_train; /**< Continue train from the previous state of optimizer
   and iterations */
```

@jijoongmoon  Could you give some opinion as well?

kparichay: The total training weights are divided into two parts - 
1. Model weights - this represents the weights which are needed when model is being inferred. This will just include weights of the layers and other layer configuration related hyper parameters like stride, padding etc.
2. Training weights - this represents weights to facilitate the training and continue from previously saved training instance. This will include optimizer related weights, epochs, iterations etc.

So, with `continue_train`, training weights must be loaded. Without, `continue_train` training weights must not be loaded.
However, there is another meaning to where the continue_train must be applied to - read or train.
If `continue_train` is applied to `train`, then the training_weights must all be reset at every call to train when Training_weights is called.
If `continue_train` is applied to `read/load`, then training weights must not be reset with every call to train.

That said, these both can be merged with `train` itself. `train` function saves the model and training weights at its end, however, it does not load it at its starting. `train` function must be modified to `read/load` at its start, where `read/load` can utilize the `continue_train` and appropriately do the job.

zhoonit: I guess this should be handled separately than just a quick fix as some application explicitly calls `readModel()`. I'll leave this as it is for now and deal from #736.

kparichay (approved):
LGTM

zhoonit:
Merged with #721

PR #721, opened 2020-11-06
[#720] [Optimizer] Change enum type to string

zhoonit:
- **#720** [Fix] Throw when read/save fails
- [Optimizer] Change enum type to string
```
    [Optimizer] Change enum type to string
    
    Change optimizer type to string
    
    **Self evaluation:**
    
```


nntrainer/optimizers/adam.h
```diff
@@ -92,6 +91,8 @@ class Adam : public Optimizer {
   double getEpsilon() { return epsilon; }
 
 private:
+  static const std::string type;
```
kparichay, line 94: this type must belong to optimizer base class and extended class of adam.

nntrainer/optimizers/sgd.cpp
```diff
@@ -15,6 +15,8 @@
 
 namespace nntrainer {
 
+const std::string SGD::type = "sgd";
```
kparichay, line 18: ditto

nntrainer/optimizers/optimizer_internal.h
```diff
@@ -151,23 +136,12 @@ class Optimizer : public ml::train::Optimizer {
   virtual void setProperty(const PropertyType type,
                            const std::string &value = "");
 
-  /**
-   * @brief     get the base name for the optimizer
-   * @retval    base name of the optimizer
-   */
-  virtual std::string getBaseName() = 0;
```
kparichay, line 158: this is still needed, mutliple optimizers with same name can exist and might be added to the model later.
Each optimizer must be able to provide a basename in order to build it a unique name.

nntrainer/optimizers/sgd.h
```diff
@@ -39,10 +38,12 @@ class SGD : public Optimizer {
                       int iteration);
 
   /**
-   * @brief     get the base name for the optimizer
-   * @retval    base name of the optimizer
+   * @copydoc Optimizer::getType()
    */
-  std::string getBaseName() { return "SGD"; };
+  const std::string getType() const { return SGD::type; }
```
kparichay, line 43: ditto - move to optimizer_internal.h

kparichay (changes requested):
Please keep a type/string object in optimizer object. It must be possible to infer type from optimizer object once created.

nntrainer/optimizers/adam.h
zhoonit: It is inferred through `getType()` not through object since virtual static member is not possible by nature.

nntrainer/optimizers/optimizer_internal.h
zhoonit: I think `getBaseName` is exactly same as `getType`. Would there be a reason to distinguish those two?

nntrainer/optimizers/sgd.h
zhoonit: Optimizer.h has getType(). Could I get the reason for optimizer_internal to have this?

nntrainer/optimizers/adam.h
kparichay: I meant from the purpose of baseclass being able to access it. I didn't realize base class (optimizer.h) has `getType()`.

nntrainer/optimizers/sgd.h
kparichay: I missed that optimizer.h has getType().

nntrainer/optimizers/optimizer_internal.h
kparichay: I thought this would fail as getBaseName was being used to assign names. But its not being used for optimizer.
getType() works as long as it is unique.

kparichay (approved):
LGTM

zhoonit:
We had an offline discussion to use integer mapping as well.

zhoonit:
I think there is not much to be done in this patch and to apply suggested change. I have changed enum class to enum in the api. I'll apply the idea when implementing `register_optimizer`.

api/ccapi/include/optimizer.h
```diff
@@ -50,7 +50,7 @@ class Optimizer {
    * @brief     get Optimizer Type
    * @retval    Optimizer type
    */
-  virtual OptimizerType getType() = 0;
```
kparichay, line 53: I guess now both are valid??

api/ccapi/include/optimizer.h
```diff
@@ -121,7 +121,7 @@ class Optimizer {
  * @brief Factory creator with constructor for optimizer
  */
 std::unique_ptr<Optimizer>
-createOptimizer(OptimizerType type,
```
kparichay, line 124: and I guess as enum remains, this will also remain??

api/ccapi/include/optimizer.h
zhoonit: Integer value will be dynamic for custom layer so I think we could only retrieve the integer from the context.

api/ccapi/include/optimizer.h
zhoonit: Yes it is added again in #740(To make PR merge faster) There are some untidy dependencies between PRs. I'll look on to the merge order.

nntrainer/optimizers/optimizer_factory.cpp
```diff
@@ -9,43 +9,56 @@
...
+#include <sstream>
 
 #include <adam.h>
 #include <optimizer_factory.h>
 #include <sgd.h>
 
 namespace nntrainer {
 
+static bool istrequal(const std::string &a, const std::string &b) {
+  return std::equal(a.begin(), a.end(), b.begin(),
```
jijoongmoon, line 22: isstrequal??

jijoongmoon (approved):
LGTM.

api/ccapi/include/optimizer.h
kparichay: Yes, given enums exist, there must be a method to get that enum value given the optimizer object.
The return type can be integer or OptimizerType, it's about readability as it does not matter with enum.

PR #725, opened 2020-11-10
[Layer] Change layer type to string

zhoonit:
- ~**#720** [Fix] Throw when read/save fails~
- ~**#721** [Optimizer] Change enum type to string~
- [Layer] Change layer type to string
```
This patch changes layer type to string

- Add Layer::getType() and Layer::type
- Add `istrequal` to the `parse_util` for case insensitive compare
- Test changes accordingly

**Self evaluation:**

```

### DIFF
--- a/api/capi/src/nntrainer_util.cpp
+++ b/api/capi/src/nntrainer_util.cpp
@@ -37,13 +37,19 @@ ml_optimizer_to_nntrainer_type(ml_train_optimizer_type_e type) {
 /**
  * @brief Convert nntrainer API layer type to neural network layer type
  */
-nntrainer::LayerType ml_layer_to_nntrainer_type(ml_train_layer_type_e type) {
+const std::string ml_layer_to_nntrainer_type(ml_train_layer_type_e type) {
   switch (type) {
   case ML_TRAIN_LAYER_TYPE_FC:
-    return nntrainer::LayerType::LAYER_FC;
+    return nntrainer::FullyConnectedLayer::type;
   case ML_TRAIN_LAYER_TYPE_INPUT:
-    return nntrainer::LayerType::LAYER_IN;
+    return nntrainer::InputLayer::type;
+  case ML_TRAIN_LAYER_TYPE_UNKNOWN:
+  /// fall through intended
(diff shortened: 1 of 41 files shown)


jijoongmoon (approved):
LGTM

api/ccapi/include/layer.h
```diff
@@ -26,26 +26,6 @@
 namespace ml {
 namespace train {
 
-/**
- * @brief     Enumeration of layer type
- */
-enum class LayerType {
```
kparichay, line 32: lets not delete this as we intend to support both enum and string.

api/ccapi/include/layer.h
```diff
@@ -198,7 +184,8 @@ class Layer {
  * @brief Factory creator with constructor for layer
  */
 std::unique_ptr<Layer>
-createLayer(LayerType type, const std::vector<std::string> &properties = {});
+createLayer(const std::string &type,
```
kparichay, line 187: ditto.

api/capi/src/nntrainer_util.cpp
```diff
@@ -37,13 +37,19 @@ ml_optimizer_to_nntrainer_type(ml_train_optimizer_type_e type) {
 /**
  * @brief Convert nntrainer API layer type to neural network layer type
  */
-nntrainer::LayerType ml_layer_to_nntrainer_type(ml_train_layer_type_e type) {
+const std::string ml_layer_to_nntrainer_type(ml_train_layer_type_e type) {
```
kparichay, line 40: as ml_train_layer_type_e and ml::train::LayerType must be kept consistent, it would be good to create a common function which converts int to string and vice versa for both capi and ccapi.

api/ccapi/include/layer.h
zhoonit: Done

api/capi/src/nntrainer_util.cpp
zhoonit: Made a function

kparichay (approved):
LGTM

Issue #361, opened 2020-07-23
Save model file format update


kparichay:
Save model struct with save model
Model structure should be saved (may be as ini) along with saving the weights.

kparichay:
From #272 : Update save file format to be a dictionary based on layer `name : type, data`
From #290 : Save, read last saved time, loss for the model

kparichay:
Updated saved format must also save layer configs like for conv layer, strides/padding etc must be saved which are currently not stored and is a BUG.

kparichay:
This will be handled with #898 and #899.

PR #726, opened 2020-11-10
[backbone] Add ini backbone properties

kparichay:
Added ini backbone properties :
    - scaleSize - scale the size of the model
    - preload - load the weights of this backbone model before adding to the model
    preload has issues which requires the format of the model file to update
    wait for #361 

Commit2:
    Added support for subgraph of a given ini backbone
    Added corresponding unittests as well
    
**Self evaluation:**
    
### DIFF
--- a/docs/configuration-ini.md
+++ b/docs/configuration-ini.md
@@ -281,7 +281,27 @@ Start with a "[ ${layer name} ]" which must be unique throughtout the model. In 
 
 2. ```trainable = <bool>```
 
-   If this backbone must be trained. Only supported for ini backbones (nntrainer models).
+   If this backbone must be trained (defaults to false). Only supported for ini backbones (nntrainer models).
+
+3. ```Preload = <bool>```
+
+   Load pretrained weights from the saved modelfile of backbone (defaults to false). Only supported for ini backbone (nntrainer models).
+
+4. ```ScaleSize = <float>```
+
+   Scale the size of the layers from backbone (defaults to 1.0). This applies for fully connected and convolution layer for now, where the units and the output channels are scaled respectively. Only supported for ini backbone (nntrainer models). If the model is being scaled, it cannot be preloaded from the saved modelfile. Only of the two options, ScaleSize and Preload, must be set at once.
+
+5. ```InputShape = <string>```
+
+   Set the shape of the input layer for the backbone model. Only supported for ini backbones (nntrainer models).
+
+6. ```InputLayer = <string>```
+
+   Choose the start layer for the backbone. This allows taking a subgraph starting with the specified layer name as a backbone. Only supported for ini backbones (nntrainer models).
+
+7. ```OutputLayer = <string>```
+
+   Choose the end layer for the backbone. This allows taking a subgraph ending with the specified layer name as a backbone. Only supported for ini backbones (nntrainer models).
 ``
 Below is sample backbone section.
 
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -618,6 +618,43 @@ int NeuralNetwork::extendGraph(GraphType graph, std::string prefix) {
   return ML_ERROR_NONE;
 }
 
+std::vector<std::shared_ptr<Layer>>
+NeuralNetwork::getGraph(const std::string &input_layer,
+                        const std::string &output_layer) {
+  /** count layers after output layer */
+  unsigned int num_layers_remove_end = 0;
+  if (!output_layer.empty()) {
+    for (auto iter = layers.rbegin(); iter != layers.rend(); iter++) {
+      if ((*iter)->getName() != output_layer)
+        num_layers_remove_end++;
+      else
+        break;
+    }
+  }
+
+  if (num_layers_remove_end == layers.size())
+    return {};
+
+  /** count layers before input layer */
+  unsigned int num_layers_remove_start = 0;
+  if (!input_layer.empty()) {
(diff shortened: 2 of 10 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -603,6 +603,44 @@ int NeuralNetwork::extendGraph(GraphType graph, std::string prefix) {
   return ML_ERROR_NONE;
 }
 
+std::vector<std::shared_ptr<Layer>>
+NeuralNetwork::getGraph(std::string input_layer, std::string output_layer) {
```
zhoonit, line 607: How about `const std::string &` for the arguments?

nntrainer/models/neuralnet.cpp
```diff
@@ -603,6 +603,44 @@ int NeuralNetwork::extendGraph(GraphType graph, std::string prefix) {
...
+  /** count layers after output layer */
+  unsigned int num_layers_remove_end = 0;
+  if (!output_layer.empty()) {
+    for (auto iter = layers.rbegin(); iter != layers.rend(); iter++) {
+      std::cout << (*iter)->getName() << std::endl;
```
zhoonit, line 612: Maybe change this to `ml_log?`

nntrainer/models/neuralnet.cpp
```diff
@@ -603,6 +603,44 @@ int NeuralNetwork::extendGraph(GraphType graph, std::string prefix) {
...
+
+  if (num_layers_remove_end == layers.size())
+    return {};
+
+  /** count layers before input layer */
+  unsigned int num_layers_remove_start = 0;
+  if (!input_layer.empty()) {
+    for (auto iter = layers.begin();
+         iter != layers.end() - num_layers_remove_end; iter++) {
+      std::cout << (*iter)->getName() << std::endl;
```
zhoonit, line 628: ditto :)

nntrainer/models/neuralnet.cpp
```diff
@@ -603,6 +603,44 @@ int NeuralNetwork::extendGraph(GraphType graph, std::string prefix) {
...
+      if ((*iter)->getName() != input_layer)
+        num_layers_remove_start++;
+      else
+        break;
+    }
+  }
+
+  /** copy the graph and return */
+  GraphType ret;
```
zhoonit, line 637: Maybe later, we might have to use flatgraph type to iterate through the graph or graph itself supports iterators

zhoonit (approved):
Minor comments, LGTM

nntrainer/models/neuralnet.cpp
kparichay: removed.

nntrainer/models/neuralnet.cpp
kparichay: removed.

nntrainer/models/neuralnet.cpp
kparichay: Yes, this will directly depend on graph_core API exposed by the graph.
For now, this is tightly dependent on the graph being implemented as a vector.

nntrainer/models/neuralnet.cpp
kparichay: good suggestion - done.

docs/configuration-ini.md
```diff
@@ -281,7 +281,27 @@ Start with a "[ ${layer name} ]" which must be unique throughtout the model. In
 
 2. ```trainable = <bool>```
 
-   If this backbone must be trained. Only supported for ini backbones (nntrainer models).
+   If this backbone must be trained (defaults to false). Only supported for ini backbones (nntrainer models).
+
+3. ```Preload = <bool>```
+
+   Load pretrained weights from the saved modelfile of backbonE (defaults to false). Only supported for ini backbone (nntrainer models).
```
jijoongmoon, line 288: backbone?

docs/configuration-ini.md
```diff
@@ -281,7 +281,27 @@ Start with a "[ ${layer name} ]" which must be unique throughtout the model. In
...
+
+4. ```ScaleSize = <float>```
+
+   Scale the size of the layers from backbone (defaults to 1.0). This applies for fully connected and convolution layer for now, where the units and the output channels are scaled respectively. Only supported for ini backbone (nntrainer models).
```
jijoongmoon, line 292: I wonder that if developer want to reduce the weight which is already trained, then how it works? I mean, let's say the filter size of conv layer is 10 and pre-loaded. If he set the scale 0.5, then how it works?

docs/configuration-ini.md
kparichay: done.

docs/configuration-ini.md
kparichay: Yes, I think this is a very important issue which should have been made clear in the documentation. 
Only one of preload or scaling can be done. Both the options activated together cannot be done for now.
Updated this to documentation.

jijoongmoon: > Yes, I think this is a very important issue which should have been made clear in the documentation.
> Only one of preload or scaling can be done. Both the options activated together cannot be done for now.
> Updated this to documentation.

Then it means scaling is only validate for the backbone network which is ready for training, then. Cause even if it is trained already, we can use scale, if we are going to train the backbone. Or If we are not going to train, then it is not allowed.

kparichay: Exactly.

jijoongmoon (approved):
LGTM

PR #728, opened 2020-11-10
[AppContext] Add AppContext

zhoonit:
This patch add basic app context with setting current working directory

**Self evaluation:**

Implemented without `std::experimental::filesystem`
~Current implementation (as of `f0d969f`) uses std::experimental::filesystem.~

~**ndk does not support std::filesystem until r22**~
~So either 1. we have to support from r22 or 2. reimplement related method in this patch.~

~IMO, I think we should stick to r21 since other people might not be using r22. and reimplement `app_context.cpp`~

### DIFF
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file	 app_context.cpp
+ * @date	 10 November 2020
+ * @brief	 This file contains app context related functions and classes that
+ * manages the global configuration of the current environment
+ * @see		 https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug	   No known bugs except for NYI items
+ *
+ */
+#include <dirent.h>
+#include <iostream>
+#include <sstream>
+
+#include <app_context.h>
+#include <nntrainer_log.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+AppContext AppContext::instance;
+
+/**
+ * @brief initiate global context
+ *
+ */
+static void init_global_context_nntrainer(void) __attribute__((constructor));
+
+/**
+ * @brief finialize global context
+ *
+ */
+static void fini_global_context_nntrainer(void) __attribute__((destructor));
+
+static void init_global_context_nntrainer(void) {}
+
+static void fini_global_context_nntrainer(void) {}
+
+AppContext &AppContext::Global() { return AppContext::instance; }
+
+static const std::string func_tag = "[AppContext::setWorkingDirectory] ";
+
+void AppContext::setWorkingDirectory(const std::string &base) {
+  DIR *dir = opendir(base.c_str());
+
+  if (!dir) {
+    std::stringstream ss;
+    ss << func_tag << "path is not directory or has no permission: " << base;
+    throw std::invalid_argument(ss.str().c_str());
+  }
+  closedir(dir);
+
+  char *ret = realpath(base.c_str(), nullptr);
+
+  if (ret == nullptr) {
+    std::stringstream ss;
+    ss << func_tag << "failed to get canonical path for the path: ";
+    throw std::invalid_argument(ss.str().c_str());
+  }
+
+  working_path_base = std::string(ret);
+  ml_logd("working path base has set: %s", working_path_base.c_str());
+  free(ret);
+}
+
+const std::string AppContext::getWorkingPath(const std::string &path) {
+
+  /// if path is absolute, return path
+  if (path[0] == '/') {
+    return path;
+  }
+
+  if (working_path_base == std::string()) {
+    return path == std::string() ? "." : path;
+  }
+
+  return path == std::string() ? working_path_base
+                               : working_path_base + "/" + path;
+}
+
+} // namespace nntrainer
--- a/nntrainer/app_context.h
+++ b/nntrainer/app_context.h
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file	 app_context.h
+ * @date	 10 November 2020
+ * @brief	 This file contains app context related functions and classes that
+ * manages the global configuration of the current environment
+ * @see		 https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#ifndef __APP_CONTEXT_H__
+#define __APP_CONTEXT_H__
+
+#include <string>
+
+namespace nntrainer {
+
+/**
+ * @class AppContext contains user-dependent configuration
+ * @brief App
+ */
+class AppContext {
+public:
+  /**
+   * @brief Get Global app context.
+   *
+   * @return AppContext&
+   */
+  static AppContext &Global();
+
+  /**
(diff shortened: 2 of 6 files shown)


nntrainer/meson.build
```diff
@@ -43,6 +43,7 @@ endforeach
 
 nntrainer_common_sources = [
   'nntrainer_logger.cpp',
+  'app_context.cpp'
```
kparichay, line 46: app_context.h must be added to nntrainer_headers if you it want to be able to import in applications.

nntrainer/app_context.h
```diff
@@ -0,0 +1,61 @@
...
+   * @brief Get Working Path from a relative or representation of a path
+   * strating from @a working_path_base.
+   * @param[in] path to make full path
+   * @return If absolute path is given, returns @a path
+   * If relative path is given and working_path_base is not set, return
+   * canonical path based on current working directory If relative path is given
+   * and working_path_base has set, return canonical path from current working
+   * directory
+   */
+  const std::string getPath(const std::string &path = "");
```
kparichay, line 53: `getAbsolutePath` or `getWorkingPath`?

nntrainer/app_context.cpp
```diff
@@ -0,0 +1,85 @@
...
+ *
+ */
+static void fini_global_context_nntrainer(void) __attribute__((destructor));
+
+static void init_global_context_nntrainer(void) {}
+
+static void fini_global_context_nntrainer(void) {}
+
+AppContext &AppContext::Global() {
+  static AppContext instance;
```
kparichay, line 41: Any reason to not create a private member `static AppContext global` inside `AppContext` class itself?

nntrainer/app_context.cpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+static void init_global_context_nntrainer(void) __attribute__((constructor));
+
+/**
+ * @brief finialize global context
+ *
+ */
+static void fini_global_context_nntrainer(void) __attribute__((destructor));
+
+static void init_global_context_nntrainer(void) {}
```
kparichay, line 36: may this function should configure the global appcontext working path as cwd?

nntrainer/app_context.cpp
zhoonit: static local are guaranteed to be thread safe by standard. I am not sure global variable(static data member) is thread safe or not. For now, I couldn't think of the cases that I have to access in otherway. So I have gone to the static local variable

nntrainer/meson.build
zhoonit: Done. Thanks

nntrainer/app_context.cpp
zhoonit: `getWorkingPath` will return relative path if is not set explicitly to give user flexibility. (e.g call `chdir()` later on user's code)

nntrainer/app_context.h
zhoonit: That sounds more clear. Changed

nntrainer/app_context.h
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file	app_context.h
+ * @date	10 November 2020
+ * @brief	This file contains app context related functions and classes that
+ * manages the global configuration of the current environment
+ * @see		https://github.com/nnstreamer/nntrainer
```
jijoongmoon, line 9: check the tab?

nntrainer/app_context.cpp
kparichay: static local members of a function of a class and static member of a class are both thread safe as long as there is 1 caller.
Two callers to the member function of a class modifying that thread variable will not be thread safe and will need synchronization primitives (not needed for now).
But this limits the static variable to that function. Later when other functionalities would also be provided, class static variable would be more versatile IMO.

jijoongmoon: So this is the only one then?

nntrainer/app_context.cpp
```diff
@@ -0,0 +1,85 @@
...
+  /// if path is absolute, return path
+  if (path[0] == '/') {
+    return path;
+  }
+
+  if (working_path_base == std::string()) {
+    return path == std::string() ? "." : path;
+  }
+
+  return path == std::string() ? working_path_base
```
kparichay, line 81: is the path here assumed here to be canonical by itself?
As per documentation, output is always canonical path but input might not be. And there is no code to convert it to canonical path.

jijoongmoon:
Could you check the ci?

test/unittest/unittest_nntrainer_appcontext.cpp
```diff
@@ -0,0 +1,106 @@
...
+TEST_F(nntrainerAppContextDirectory, readFromGetPath_p) {
+  nntrainer::AppContext ac = nntrainer::AppContext::Global();
+
+  std::string path = ac.getWorkingPath("testfile.txt");
+  EXPECT_EQ(path, "testfile.txt");
+
+  ac.setWorkingDirectory("testdir");
+
+  path = ac.getWorkingPath("testfile.txt");
+  EXPECT_EQ(path, current_directory + "/testdir/testfile.txt");
```
kparichay, line 63: I wonder about this. 
As per line 60, working directory is `testdir` aka `./testdir`, but it matches with `/testdir`.
Am I understanding something wrong here?

kparichay, line 63: as per this example, `setWorkingDirectory("testdir")` sets `${CWD}\testdir`.
Then, in that case, the default working directory must be `${CWD}`? but this is opposite to unittest on line 58.

When no working directory is set, CWD must be evaluated always with `getWorkingPath()` in case user changed it.

nntrainer/app_context.cpp
kparichay: Lock will not protect against changing while network is running. If user changes while running, and its owned by the user, its on them IMO.

zhoonit: I'll change this to static member

It's a bit different story but model will copy own the context to avoid temporal coupling. Changing context after setting model will not take effect. it won't be that expensive because context is a small class

test/unittest/unittest_nntrainer_appcontext.cpp
zhoonit: I think user can expect it to be relative path if working path is not set. If we fixate to current `${CWD}`. changing `chdir()` would take no effect.(and also our itc, utc will break) It is intended behavior so it is tested. However, doc was wrong. My apologies.

nntrainer/app_context.cpp
zhoonit: Yeah you are right, it should say absolute, not canonical. Fixed documentation

nntrainer/app_context.h
zhoonit: Done. Thank you!

nntrainer/app_context.cpp
zhoonit: > So this is the only one then? I wonder that is it ok to change this in the middle of processing by other network? Maybe add some lock?

Local static member is guaranteed to be thread-safe from c++11

kparichay (approved):
LGTM

nntrainer/app_context.cpp
jijoongmoon: > > So this is the only one then? I wonder that is it ok to change this in the middle of processing by other network? Maybe add some lock?
> 
> Local static member is guaranteed to be thread-safe from c++11

What I mean is that we shouldn't allow changing this in the middle of processing. That's what I mean. Stop and changing it and run is Ok. but changing it in the middle of processing, may cause problem I guess.

zhoonit: The issue is handled in #733. Model copies the app_context which fixate the app context used for the model. So even if we change in the middle of the process, it remains safe.

jijoongmoon (approved):
LGTM

PR #735, opened 2020-11-11
[ccapi] Syntactic sugar constructors

### LINKED ISSUE #734: ccapi add syntactic sugar
Add syntactic sugar methods for ccapi to make API easier to write and use for optimizer, layer and dataset, for their factor constructor methods.

kparichay:
Added syntactic sugar constructors for optimizers
which allow making optimizer closer to existing API
and more readable.

Resolves #734

**Self evaluation:**

### DIFF
--- a/api/ccapi/src/factory.cpp
+++ b/api/ccapi/src/factory.cpp
@@ -55,6 +55,47 @@ createOptimizer(const OptimizerType &type,
 }
 
 /**
+ * @brief Factory creator with constructor for loss
+ */
+static std::unique_ptr<Layer>
+createLoss(nntrainer::LossType type,
+           const std::vector<std::string> &properties) {
+  std::unique_ptr<Layer> layer = nntrainer::createLoss(type);
+
+  if (layer->setProperty(properties) != ML_ERROR_NONE)
+    throw std::invalid_argument("Set properties failed for layer");
+
+  return layer;
+}
+
+namespace loss {
+
+/**
+ * @brief Helper function to create loss layer
+ */
+static std::unique_ptr<Layer> Loss(nntrainer::LossType type,
+                                   const std::vector<std::string> &properties) {
+  return createLoss(type, properties);
+}
+
+/**
+ * @brief Helper function to create mse layer
+ */
+std::unique_ptr<Layer> MSE(const std::vector<std::string> &properties) {
+  return Loss(nntrainer::LossType::LOSS_MSE, properties);
+}
+
+/**
+ * @brief Helper function to create cross entropy layer
+ */
+std::unique_ptr<Layer>
+CrossEntropy(const std::vector<std::string> &properties) {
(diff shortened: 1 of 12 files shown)


zhoonit (approved):
LGTM

kparichay:
@zhoonit My apologies, but can you please review once more. Direct methods are added for layers and loss as well.

api/ccapi/src/factory.cpp
```diff
@@ -40,6 +40,161 @@ std::unique_ptr<Layer> createLayer(LayerType type,
...
+           const std::vector<std::string> &properties) {
+  std::unique_ptr<Layer> layer = nntrainer::createLoss(type);
+
+  if (layer->setProperty(properties) != ML_ERROR_NONE)
+    throw std::invalid_argument("Set properties failed for layer");
+
+  return layer;
+}
+
+namespace loss {
```
zhoonit, line 171: Are we gonna open up loss to be generated? I think our code is not ready for this.

zhoonit: How about we have an alias sth like using Loss = Layer?

kparichay: The loss_layer was already opened. However, there was no way to choose the loss type for the created loss layer.
This patch allows that.

kparichay: > How about we have an alias sth like using Loss = Layer?

creating two namespace purpose is to keep loss and layer decoupled. If later, loss is removed a layer, it would require updates here.

zhoonit:
As #721 has been merged, please wait for #740

test/ccapi/unittest_ccapi.cpp
```diff
@@ -50,17 +50,30 @@ TEST(ccapi_layer, construct_01_n) {
...
-  EXPECT_NO_THROW(ml::train::createLayer(ml::train::LayerType::LAYER_CONV2D));
-  EXPECT_NO_THROW(
-    ml::train::createLayer(ml::train::LayerType::LAYER_POOLING2D));
-  EXPECT_NO_THROW(ml::train::createLayer(ml::train::LayerType::LAYER_FLATTEN));
-  EXPECT_NO_THROW(
-    ml::train::createLayer(ml::train::LayerType::LAYER_ACTIVATION));
-  EXPECT_NO_THROW(ml::train::createLayer(ml::train::LayerType::LAYER_ADDITION));
-  EXPECT_NO_THROW(ml::train::createLayer(ml::train::LayerType::LAYER_LOSS));
+  EXPECT_NO_THROW(ml::train::layer::Input());
+  EXPECT_NO_THROW(ml::train::layer::FullyConnected());
```
jijoongmoon, line 54: Looks good.

jijoongmoon (approved):
Looks good to me.

api/ccapi/src/factory.cpp
zhoonit: Okay

PR #740, opened 2020-11-11
[Optimizer] Add enum type factory

zhoonit:
- ~**#720** [Fix] Throw when read/save fails~
- ~**#721** [Optimizer] Change enum type to string~
- [Optimizer] Add enum type factory
```
This patch adds integer factory to align with capi

**Self evaluation:**

```

### DIFF
--- a/nntrainer/optimizers/optimizer_factory.cpp
+++ b/nntrainer/optimizers/optimizer_factory.cpp
@@ -13,12 +13,31 @@
 #include <sstream>
 
 #include <adam.h>
+#include <nntrainer_error.h>
 #include <optimizer_factory.h>
 #include <parse_util.h>
 #include <sgd.h>
 
 namespace nntrainer {
 
+/// helper function to convert enum to string
+/// @todo this should be integrated into appcontext
+const std::string optimizerIntToStrType(const OptType &type) {
+  switch (type) {
+  case OptType::ADAM:
+    return "adam";
+  case OptType::SGD:
+    return "sgd";
+  case ML_TRAIN_OPTIMIZER_TYPE_UNKNOWN:
+  /// fall through intended
+  default:
+    throw exception::not_supported(
+      "[opt_integer_to_string_type] Not supported type given");
+  }
+
+  throw exception::not_supported(
+    "[opt_integer_to_string_type] Not supported type given");
+}
 /**
  * @brief Factory creator with copy constructor
  */
@@ -39,6 +58,12 @@ std::unique_ptr<Optimizer> createOptimizer(const std::string &type,
   throw std::invalid_argument(ss.str().c_str());
 }
 
+std::unique_ptr<Optimizer> createOptimizer(const OptType &type,
+                                           const Optimizer &opt) {
+  const std::string &s = optimizerIntToStrType(type);
+  return createOptimizer(s, opt);
+}
+
 /**
  * @brief Factory creator with constructor
  */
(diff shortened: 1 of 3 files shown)


nntrainer/optimizers/optimizer_factory.cpp
```diff
@@ -22,6 +23,25 @@ static bool istrequal(const std::string &a, const std::string &b) {
   return std::equal(a.begin(), a.end(), b.begin(),
                     [](char a, char b) { return tolower(a) == tolower(b); });
 }
+
+/// helper function to convert enum to string
+/// @todo this should be integrated into appcontext
+static const std::string integer_to_string_type(const OptType &type) {
+  switch (type) {
```
jijoongmoon, line 30: now we support enum and string both.

jijoongmoon (approved):
LGTM

nntrainer/optimizers/optimizer_factory.cpp
```diff
@@ -22,6 +23,25 @@ static bool istrequal(const std::string &a, const std::string &b) {
   return std::equal(a.begin(), a.end(), b.begin(),
                     [](char a, char b) { return tolower(a) == tolower(b); });
 }
+
+/// helper function to convert enum to string
+/// @todo this should be integrated into appcontext
+static const std::string integer_to_string_type(const OptType &type) {
```
kparichay, line 29: I wonder why not use unordered_map?

nntrainer/optimizers/optimizer_factory.h
```diff
@@ -14,21 +14,35 @@
...
+
 /**
  * @brief Factory creator with constructor
  */
 std::unique_ptr<Optimizer> createOptimizer(const std::string &type);
 
+/**
+ * @brief Factory creator with constructor using enum(integer)
+ */
+std::unique_ptr<Optimizer> createOptimizer(const OptType &type);
```
kparichay, line 44: Can you also bring back the corresponding function of ccapi which got deleted.

nntrainer/optimizers/optimizer_factory.cpp
zhoonit: This is code bit is temporary before moving into appcontext. #745 will use unordered map.

nntrainer/optimizers/optimizer_factory.h
zhoonit: Yup, will do in 30 min

nntrainer/optimizers/optimizer_factory.cpp
kparichay: Then maybe this PR can be merged with 745? as it seems like most of this patch will be replaced with 745.

zhoonit: Since #735 has to wait a lot and #745 need a lot more work...  I think it's better to merge this for #735 :)

nntrainer/optimizers/optimizer_factory.h
zhoonit: Done

nntrainer/optimizers/optimizer_factory.cpp
zhoonit: Yes

nntrainer/optimizers/optimizer_factory.cpp
kparichay: cool.

kparichay (approved):
LGTM

PR #743, opened 2020-11-11
[model] Handle loss layer to be added from user

### LINKED ISSUE #744: Allow loss layer to be created and added by user
Allow loss layer to be created and added by the user.

kparichay:
Handle loss layer to be added from the user which is created with c++ API

Resolves #744

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -67,20 +67,30 @@ int NeuralNetwork::initLossLayer() {
   int status = ML_ERROR_NONE;
   LossType updated_loss_type = loss_type;
 
-  if (layers.empty()) {
-    status = ML_ERROR_INVALID_PARAMETER;
-    NN_RETURN_STATUS();
+  if (layers.back()->getType() != LossLayer::type) {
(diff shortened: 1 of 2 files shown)


jijoongmoon (approved):
LGTM

nntrainer/models/neuralnet.cpp
```diff
@@ -186,9 +188,22 @@ int NeuralNetwork::init() {
...
+  if (initialized) {
+    ml_loge("Error: Initializing the model again");
+    return ML_ERROR_NOT_SUPPORTED;
+  }
+
   status = isInitializable();
   NN_RETURN_STATUS();
 
+  /** Add loss layer if not already added */
+  if (layers.back()->getType() != LayerType::LAYER_LOSS) {
```
zhoonit, line 200: How about add a warning if the user is overriding the loss set to the model?

zhoonit (approved):
LGTM :)

nntrainer/models/neuralnet.cpp
kparichay: done.

PR #745, opened 2020-11-11
[AppContext] Add register, invoke factory @open sesame 11/26 18:15

zhoonit:
- ~#728 [AppContext] Add AppContext~
- ~#720 [Fix] Throw when read/save fails~
- ~#721  [Optimizer] Change enum type to string~
- [AppContext] Add register, invoke factory
```
  [AppContext] Add registerer,invoke factory methods
    
    **Changes proposed in this PR:**
    - Add factory registerer
    - Add factory invoker
    - Register built-in objects to each layers(postponed)
    - Add tests
    
    **Self evaluation:**
    1. Build test: [X]Passed [ ]Failed [ ]Skipped
    2. Run test: [X]Passed [ ]Failed [ ]Skipped
    
    Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>
```

PR #746, opened 2020-11-13
[AppContext][Optimizer] Register Default ops at the beginning @open sesame 11/27 15:35

zhoonit:
- ~**#728** [AppContext] Add AppContext~
- ~**#745** [AppContext] Add registerer,invoke factory methods~
- [AppContext] Register Default ops at the begining
```
- Register default optimizer at the beginning of the load

**Self evaluation:**

```

### DIFF
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -15,15 +15,15 @@
 #include <iostream>
 #include <sstream>
 
+#include <adam.h>
 #include <app_context.h>
 #include <nntrainer_log.h>
+#include <sgd.h>
 #include <util_func.h>
 
 namespace nntrainer {
 
 std::mutex factory_mutex;
-
-AppContext AppContext::instance;
 
 /**
  * @brief initiate global context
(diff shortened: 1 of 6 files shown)


nntrainer/app_context.cpp
```diff
@@ -35,11 +35,22 @@ static void init_global_context_nntrainer(void) __attribute__((constructor));
...
+  ac.registerFactory(ml::train::createOptimizer<Adam>);
+  ac.registerFactory(AppContext::unknownFactory<ml::train::Optimizer>,
+                     "unknown", OptType::UNKNOWN);
+}
 
 static void fini_global_context_nntrainer(void) {}
 
-AppContext &AppContext::Global() { return AppContext::instance; }
+AppContext &AppContext::Global() {
+  static AppContext instance;
```
zhoonit, line 51: There seems a static init order problem (or might be compiler error...) when I made this to global variable. (http://www.cs.technion.ac.il/users/yechiel/c++-faq/static-init-order.html) which led to a floating point error. (referring to https://stackoverflow.com/questions/13580823/sigfpe-when-accessing-unordered-map)

Changing this to a function local static variable. Since it does not make much of a difference for us.

jni/Android.mk
```diff
@@ -83,7 +83,7 @@ NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/optimizers/optimizer_factory.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/util_func.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/parse_util.cpp \
-                  $(NNTRAINER_ROOT)/nntrainer/app_context.cpp
+									$(NNTRAINER_ROOT)/nntrainer/app_context.cpp
```
lhs8928, line 86: I think this one should be a space but tab

zhoonit: Fixed

jijoongmoon (approved):
LGTM

kparichay (approved):
LGTM

Issue #732, opened 2020-11-11
Features (Exceptions&Overheads) + Survey (Memory) Requests from Clients


myungjoo:
Survey: this week.
Features: consider to include for 2021 development items

---

## Survey
- Peak Memory Consumption
    - With a Tizen GUI App
    - With a console App without UI

## Features
- Overhead (CPU/memory consumption) control based on the app status (Foreground/Background) or system status (user actively using it. user is leaving the device alone)
- Exception handling or checkpointing for a sudden power outage during trainig.

myungjoo:
@jijoongmoon Please assign someone to measure peak memory consuption.

lhs8928:
About Peak Memory Consumption survey:
With Cifar10 transfer learning without tflite the maximum resident set size is about 67Mb.
The reason why this much memory is required will be analyzed.

myungjoo:
> About Peak Memory Consumption survey:
> With Cifar10 transfer learning without tflite the maximum resident set size is about 67Mb.
> The reason why this much memory is required will be analyzed.

Please describe the size of model and its structure as well.

zhoonit:
For GUI application [here](https://github.com/nnstreamer/nntrainer/tree/main/Applications/Tizen_native/CustomShortcut)

## A. The application flow 

### 1. idle stage
App is launched but nothing has been done

### 2. drawing and feature extract
draw 10 samples to personalize and the feature is extracted with nnstreamer API, model is used [mobilenetv2.tflite](https://github.com/nnstreamer/nntrainer/tree/main/Applications/Tizen_native/CustomShortcut/res/mobilenetv2.tflite)

### 3. training
 Do training with nntrainer API

### 4. drawing + inferencing
 User draw one shot and do inference with nnstreamer API (pipeline tflite model and feed to nntrainer model)
### MISC.
Demonstration of drawing + feature extraction + training (stage 2 + stage 3)
![img](https://raw.githubusercontent.com/nnstreamer/nntrainer/main/docs/images/customshortcut_training.webp)

Demonstration of drawing + inferencing (stage 4)
![img](https://raw.githubusercontent.com/nnstreamer/nntrainer/main/docs/images/customshortcut_inference.webp)


## B. Test Setup

### Model / test input

- input: width=224, height=224, format = rgb, normalized to `[-1, 1]` -> 1:3:224:224
- feature extractor: mobilenetv2 model structure is lengthy so it is attached  (size of modelfile is 8.5M)
[...]

Issue #670, opened 2020-10-26
Update the signature of out-of-place tensor operations


kparichay:
out-of-place tensor operations return output tensor without taking in an output tensor.
This makes it impossible to perform an operation given a pre-allocated output tensor without a memcopy.

This is also needed for optimized graph operations - 
Once the first iteration is run, the peak memory consumption of the application must not increase. This means that the output memory once allocated by a layer must be reused. Current design not only leads to de-allocation/re-allocation of memory, again and again, causing time overhead, but in the worst case, can make the peak memory consumption twice of what is actually should be.

Already discussed signature for this (check confluence) -
`Tensor op(Tensor &in, Tensor &out);`

kparichay:
This will help reduce the signature of peak memory as the total memory required will get fixed after first iteration with this.
#732 peak memory will be affected by this.

kparichay:
Update return types to be `Tensor &`

PR #747, opened 2020-11-13
[tensor] Tensor operations change style

kparichay:
**Commit1:**
Update tensor op signature for dot
    Further, fc layer has been updated based on this signature
    
**Commit2:**
This patch updates tensor signature to include output for apply
    Correspondingly, activation layer and loss layer have been updated.
    
More changes in the patch
    - Convolution layer has been updated to reuse its output derivative.
    - added a getBatchSlice() which provides a slice of the original tensor by batch
    and avoids creating a copy everytime

Note : Using a class like Weight for inputs/outputs has been inteniontionally skipped to limit the size of this patch. Further, the `ret_derivative` is not created as a vector, will be done when extended to a new class. 

See also #670 #732

**Self evaluation:**
    
### DIFF
--- a/nntrainer/layers/activation_layer.cpp
+++ b/nntrainer/layers/activation_layer.cpp
@@ -49,7 +49,7 @@ int ActivationLayer::initialize() {
 sharedConstTensors ActivationLayer::forwarding(sharedConstTensors in) {
   input = *in[0];
   /// @note @a _act_fn is expected to work out of place and not modify @a input
-  hidden = _act_fn(input);
+  _act_fn(input, hidden);
 
   return {MAKE_SHARED_TENSOR(hidden)};
 }
@@ -57,18 +57,17 @@ sharedConstTensors ActivationLayer::forwarding(sharedConstTensors in) {
 sharedConstTensors ActivationLayer::backwarding(sharedConstTensors derivative,
                                                 int iteration) {
   Tensor deriv = *derivative[0];
-  Tensor ret;
   if (activation_type == ActivationType::ACT_SOFTMAX)
-    ret = _act_prime_fn(hidden, deriv);
+    ret_derivative = _act_prime_fn(hidden, ret_derivative, deriv);
   else
-    ret = _act_prime_fn(input, deriv);
+    ret_derivative = _act_prime_fn(input, ret_derivative, deriv);
 
-  return {MAKE_SHARED_TENSOR(std::move(ret))};
+  return {MAKE_SHARED_TENSOR(ret_derivative)};
 }
 
 int ActivationLayer::setActivation(
-  std::function<Tensor(Tensor const &)> const &activation_fn,
-  std::function<Tensor(Tensor const &, Tensor const &)> const
+  std::function<Tensor(Tensor const &, Tensor &)> const &activation_fn,
+  std::function<Tensor(Tensor const &, Tensor &, Tensor const &)> const
     &activation_prime_fn) {
   _act_fn = activation_fn;
   _act_prime_fn = activation_prime_fn;
@@ -77,12 +76,12 @@ int ActivationLayer::setActivation(
 }
 
 int ActivationLayer::setActivation(
-  std::function<Tensor(Tensor const &)> const &activation_fn,
-  std::function<Tensor(Tensor const &)> const &activation_prime_fn) {
+  std::function<Tensor(Tensor const &, Tensor &)> const &activation_fn,
+  std::function<Tensor(Tensor const &, Tensor &)> const &activation_prime_fn) {
   _act_fn = activation_fn;
-  _act_prime_fn = [activation_prime_fn](Tensor const &x,
+  _act_prime_fn = [activation_prime_fn](Tensor const &x, Tensor &ret_derivative,
                                         Tensor const &derivative) {
-    return derivative.multiply(activation_prime_fn(x));
+    return derivative.multiply(activation_prime_fn(x, ret_derivative));
   };
 
   return ML_ERROR_NONE;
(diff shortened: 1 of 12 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -260,6 +266,16 @@ Tensor Tensor::pow(float exponent) const {
   return apply([=](float in) { return powf(in, exponent); });
 }
 
+Tensor Tensor::getBatchSlice(unsigned int offset, unsigned int size) const {
+  Tensor ret = *this;
```
jijoongmoon, line 270: This will help a lot.

jijoongmoon (approved):
LGTM. This will save memory consumption.

nntrainer/tensor/tensor.cpp
```diff
@@ -445,12 +451,20 @@ Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
...
+}
+
 /**
  * @note: This dot product flattens the fist 3 axis for the purpose of
  * computation. So, while performing, these matrices are behaving as 2-D
  * matrices. The dimensions are restored while returning back the tensor.
  */
-Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
+Tensor Tensor::dot(Tensor const &m, Tensor &result, bool trans,
+                   bool trans_m) const {
```
zhoonit, line 477: How about we make sure  `&m.getData() = &result.getData()`? Eventhough m is const, underlying data can change in this case.

nntrainer/tensor/tensor.cpp
```diff
@@ -266,6 +266,16 @@ Tensor Tensor::pow(float exponent) const {
   return apply([=](float in) { return powf(in, exponent); });
 }
 
+Tensor Tensor::getBatchSlice(unsigned int offset, unsigned int size) const {
```
zhoonit, line 269: Size seems not used. Could you check this out?

nntrainer/tensor/tensor.cpp
zhoonit: I guess this will prevail, so instead putting all checks in every operation, I wish there is a better way to do

zhoonit:
IMO, the return signature should be void or int, as `Tensor &out` is already outparam. I think it would be easy to misuse the operation if we return the outparam as well.

for example,
if we do `A = B.op(C, out=D)`.
We have to memorize that A and D are the same.

In the case of `ret = (A + B - C) x D`, `A.add(B, ret).subtract(C, ret).dot(D, ret)` seems same but I think it is UB.

Sorry for the late design discussion.

nntrainer/tensor/tensor.cpp
zhoonit: `&result != this` is necessary in this case as well

nntrainer/layers/activation_layer.cpp
```diff
@@ -49,26 +49,25 @@ int ActivationLayer::initialize() {
 sharedConstTensors ActivationLayer::forwarding(sharedConstTensors in) {
   input = *in[0];
   /// @note @a _act_fn is expected to work out of place and not modify @a input
-  hidden = _act_fn(input);
+  hidden = _act_fn(input, hidden);
```
zhoonit, line 52: Just confirming I guess no need for the assignment?

nntrainer/tensor/tensor.cpp
kparichay: you mean inverse of `&m.getData() = &result.getData()`?? (that they should not be same?)
but this wont be sufficient, it will have to be `&m.getData() + m.length()` and `&result.getData() + result.length()` (if result is preallocated) are not intersecting.

nntrainer/tensor/tensor.cpp
kparichay: good finding, updated.

kparichay:
> IMO, the return signature should be void or int, as `Tensor &out` is already outparam. I think it would be easy to misuse the operation if we return the outparam as well.

The purpose of this design is to allow chaining of operations.
However, this does not enforce to use of this style. The caller can ignore the return data.

> for example,
> if we do `A = B.op(C, out=D)`.
> We have to memorize that A and D are the same.

In `A = B.op(C, out=D)`, yes, the writer has to notice that A and D are the same. The writer can assign different names (like A, D) but has to know that A, D not only points to the same memory but in fact is the same tensor itself.

> In the case of `ret = (A + B - C) x D`, `A.add(B, ret).subtract(C, ret).dot(D, ret)` seems same but I think it is UB.

I dont understand why this will be UB though (as long as dimension constraints are satisfied). Below is something that should work and I cant identify the difference from the example above.
```c
Tensor ret;
ret = A.add(B, ret);
ret = ret.subtract(C, ret);
ret = ret.dot(D, ret);
```

nntrainer/layers/activation_layer.cpp
kparichay: done.

nntrainer/tensor/tensor.cpp
zhoonit: sorry, I meant sth like,

```cpp
if(m.getData() == result.getData() || this->getData() == result.getData()) {
 throw
} 
```

As you commented out, I guess checking if intersecting should be done

zhoonit:
> ```c
> Tensor ret;
> ret = A.add(B, ret);
> ret = ret.subtract(C, ret);
> ret = ret.dot(D, ret);
> ```

My concern was that Tensor ret.dot(D, ret) can't be done. As dot can't have inplace operator. For now, let's proceed and have an in-depth improvement later :)

zhoonit (approved):
LGTM :)

nntrainer/tensor/tensor.cpp
kparichay: These overlapping checks and return type updates will do in next PR if that's ok, as they will have to be done for a class of operations.

zhoonit: No worries

PR #748, opened 2020-11-13
[conv] Remove standardization and normalization

kparichay:
Remove support for standardization and normalization from convolution operations
These operations do not support backwarding for now, and cannot be forwarded with conv2d layer for this reason
For now, this is only supported with input layer which is never backwarded

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -81,14 +81,6 @@ sharedConstTensors Conv2DLayer::forwarding(sharedConstTensors in) {
   TensorDim &in_dim = input_dim[0];
   TensorDim &out_dim = output_dim[0];
 
-  if (normalization) {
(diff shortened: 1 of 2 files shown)


jijoongmoon:
Just make sure, then if user want to use conv2d layer as an first layer and need normalization or standardization, then we give proper notice that he need input layer. or we can allow it if it is first layer?

zhoonit (approved):
LGTM

kparichay:
> Just make sure, then if user want to use conv2d layer as an first layer and need normalization or standardization, then we give proper notice that he need input layer. or we can allow it if it is first layer?

When user will try to set standardization and normalization for convolution layer, it will result in an error as it is not supported by convolution layer. For now, user will have to use input layer for this.

jijoongmoon (approved):
LGTM

PR #749, opened 2020-11-13
[tensor] Reduce overall memory overhead

kparichay:
Reduce overall tensor memory overhead
    - standardization and normalization now take in-place
    - input and label tensor in train are now allocated only once externally and reused for each epoch
    that being used in each epoch
    - add_i does not allocate new memory now
    - removed support for normalization and standardization for conv layer
    
See also #670 #732

**Self evaluation:**
    

jijoongmoon (approved):
LGTM.

zhoonit (approved):
LGTM

PR #752, opened 2020-11-16
[var_grad] Add var_grad for input/output lists

kparichay:
Added var_grad for input/output lists which also combines derivatives
    This is the baseclass for the weights
    This update will help with graph class
    
**Self evaluation:**
    
### DIFF
--- a/nntrainer/tensor/var_grad.h
+++ b/nntrainer/tensor/var_grad.h
@@ -1,0 +1,154 @@
+// SPDX-License-Identifier: Apache-2.0-only
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	var_grad.h
+ * @date	13 November 2020
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is Var_Grad Class for Neural Network
+ *
+ */
+
+#ifndef __VAR_GRAD_H__
+#define __VAR_GRAD_H__
+
+#include <tensor.h>
+
+namespace nntrainer {
+
+/**
+ * @class   Var_Grad
+ * @brief   Variable with Gradient, and its corresponding trainable property
+ */
+class Var_Grad {
+
+public:
+  /**
+   * @brief Var_Grad default constructor
+   */
+  Var_Grad() : trainable(false) {}
+
+  /**
+   * @brief Construct a new Var_Grad object
+   *
+   * @param dim Variable and gradient tensor dimension
+   * @param train If the variable is trainable
+   * @param name Name for this Var_Grad
+   */
+  Var_Grad(const TensorDim &dim, bool train = true,
+           const std::string &name = "");
+
+  /**
+   * @brief Swap for Var_Grad
+   *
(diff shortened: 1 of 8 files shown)


meson.build
```diff
@@ -37,8 +37,8 @@ warning_flags = [
   '-Wvla',
   '-Wpointer-arith',
   '-Wno-error=varargs',
-  '-O2',
-  '-ftree-vectorize'
+  #  '-O2',
```
zhoonit, line 40: Maybe unintended comment?

nntrainer/tensor/var_grad.h
```diff
@@ -0,0 +1,154 @@
...
+  Var_Grad() : trainable(false) {}
+
+  /**
+   * @brief Construct a new Var_Grad object
+   *
+   * @param dim Variable and gradient tensor dimension
+   * @param train If the variable is trainable
+   * @param name Name for this Var_Grad
+   */
+  Var_Grad(const TensorDim &dim, bool train = true, std::string name = "");
```
zhoonit, line 40: Maybe `const std::string & name`?

nntrainer/tensor/var_grad.h
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * @brief Swap for Var_Grad
+   *
+   * @param lhs Swap to
+   * @param rhs Swap from
+   * @note Only swap gradient if trainable
```
zhoonit, line 47: I guess this note should go away

zhoonit (commented):
Could I get an explanation for the `var_grad`? I don't see the usage in the code and can't think of a reason to factor out those

meson.build
kparichay: done.

nntrainer/tensor/var_grad.h
kparichay: done.

nntrainer/tensor/var_grad.h
kparichay: done.

kparichay:
> Could I get an explanation for the `var_grad`? I don't see the usage in the code and can't think of a reason to factor out those

`var_grad` will replace inputs, outputs, derivative, ret_derivative for layers.

zhoonit (approved):
LGTM

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -1034,16 +1034,13 @@ TEST_F(nntrainer_Conv2DLayer, save_read_01_p) {
...
-
-  read_file.seekg(0, std::ios::beg);
 
   std::ifstream read_file2("save1.bin");
   ASSERT_TRUE(read_file2.good());
 
   float d1, d2;
 
-  for (int i = 0; i < (5 * 5 * 3 * 6) + 6; ++i) {
+  for (int i = 0; i < (5 * 5 * 3 * 12) + 12; ++i) {
```
lhs8928, line 1043: Does this change is related to var_grad?

kparichay: It does not directly. The test in this case was not being performed on the full weight which was revealed when using this patch.

lhs8928:
LGTM

lhs8928 (approved):

jijoongmoon (approved):
LGTM

PR #753, opened 2020-11-16
[Layer] Add built-in ops to the context @open sesame 11/30 11:08

zhoonit:
- ~**#745** [AppContext] Add registerer,invoke factory methods~
- ~**#746** [AppContext] Register Default ops at the begining~
- [AppContext] Fix key is case sensitive
```
In current semantics, type key should be case insensitive however case
was sensitive, this patch fixes the issue.

**Self evaluation:**

```
- [Layer] Add built-in ops to the context
```
- Add default layers to the global context

**Self evaluation:**

```

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -21,6 +21,7 @@
 #include <string>
 #include <vector>
 
+#include <ml-api-common.h>
 #include <nntrainer-api-common.h>
 
 namespace ml {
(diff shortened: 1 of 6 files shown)


jijoongmoon (approved):
LGTM

kparichay (approved):
LGTM

api/ccapi/include/layer.h
```diff
@@ -363,6 +364,23 @@ CrossEntropy(const std::vector<std::string> &properties = {});
 
 } // namespace loss
 
+/**
+ * @brief General Layer Factory function to register Layer
+ *
+ * @param props property representation
+ * @return std::unique_ptr<ml::train::Layer> created object
+ */
+template <typename T,
```
kparichay, line 373: Let's add a comment in `layer_internal` and `optimizer_internal` to ensure to make a unit test with `createLayer/createOptimizer`, otherwise, their templated function won't be made.

zhoonit: Isn't the reference made because `createLayer/createOptimizer` is used in `app_context.cpp`?

Please correct me if I have made a misunderstanding

kparichay: Then a comment that all new layers to be used must be included in `app_context.cpp`.

zhoonit: Done

PR #755, opened 2020-11-16
[graph] Skip backwarding for non-trainable layers

kparichay:
This patch skips the backwarding for the non-trainable layers.
    Further, the last trainable layer skips calcDerivative as well.
    This results in much fewer calculations as well as more importantly, reduced memory.
    
See also #732
    
**Self evaluation:**
    
### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -98,6 +98,16 @@ LayerNode &NetworkGraph::getSortedLayerNode(unsigned int ith) {
   throw std::invalid_argument("Cannot find Layer");
 }
 
+void NetworkGraph::countNonTrainableLayersAtBegin() {
+  /** TODO: update for multiple inputs when it is supported */
+  for (auto iter = Sorted.cbegin(); iter != Sorted.cend(); iter++) {
+    if ((*iter).layer->getTrainable()) {
+      skip_non_trainable_layers = iter - Sorted.cbegin();
+      break;
+    }
+  }
+}
+
 void NetworkGraph::topologicalSort() {
   std::stack<LayerNode> Stack;
   std::vector<bool> visited(num_node);
@@ -118,6 +128,9 @@ void NetworkGraph::topologicalSort() {
     Sorted.push_back(Stack.top());
     Stack.pop();
   }
+
+  /** TODO: this will be replaced with a corresponding graph function */
+  countNonTrainableLayersAtBegin();
 }
 
 void NetworkGraph::ensureName(std::shared_ptr<Layer> layer,
@@ -517,7 +530,9 @@ sharedConstTensors NetworkGraph::forwarding(sharedConstTensors input) {
 
 void NetworkGraph::backwarding(sharedConstTensors output, int iteration) {
 
-  for (unsigned int i = Sorted.size() - 1; i > 0; i--) {
+  /** First layer in Sorted must a INPUT layer */
+  for (unsigned int i = Sorted.size() - 1; i > skip_non_trainable_layers + 1;
+       i--) {
     LayerNode &layer_node = Sorted[i];
     if (istrequal(layer_node.layer->getType(), nntrainer::LossLayer::type)) {
       layer_node.layer->backwarding(iteration, output);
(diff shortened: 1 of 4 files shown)


kparichay:
With this patch, the peak memory requirement for CIFAR10 transfer learning application (mentioned in #732) is reduced to 38252KB. Below is the analysis of this memory requirement :
- weight of the fc layer (there is only one layer in the model) - 62720 (input size) x 10 (number of outputs) x 4 (size of float) = `2450 KB`
- gradient of the FC layer (same size as weight) = `2450 KB`
- adam optimizer weights (two tensors with same shape as weight) - 2 x 62720 x 10 x 4 = `4900 KB`
- input for the layer - 32 (batch size) x 62720 (input size) x 4 (float size) = `7840 KB`

Total bare minimum memory requirement (sum of the above) = `17640 KB`
Peak heap memory required by nntrainer (evaluated using [heaptrack](https://github.com/KDE/heaptrack) = `26.52 MB`.This includes the next input buffer of size 7840KB which has been pre-fetched to be executed next for the best training performance. Excluding this pre-fetched buffer memory requirement brings the Peak heap memory to `19316 KB`. This difference of `1676 KB` is under investigation.

The total peak memory requirement is high because that value includes nntrainer shared object and its dependencies.

nntrainer/models/neuralnet.cpp
```diff
@@ -309,8 +322,13 @@ void NeuralNetwork::backwarding(sharedConstTensors input,
   forwarding(input, label);
 
   sharedConstTensors output = label;
-  for (unsigned int i = layers.size() - 1; i > 0; i--)
+  for (unsigned int i = layers.size() - 1; i > skip_non_trainable_layers + 1;
+       i--)
     output = layers[i]->backwarding(output, iteration);
+
+  /** The last trainable layer need not calculate the derivatives */
+  layers[skip_non_trainable_layers + 1]->calcGradient(output);
```
zhoonit, line 330: Have you consider the api point of view? If neuralnet starts to know about this, we now have `layer::forward`, `layer::calcDeriv`, `layer::calcGradient` and `layer::applyGradient`. Also we have  `layer::backward` function but most likely it should be inherited from `nntrainer::internal`

kparichay: From the API point of view, `calDerivative`, `calcGradient` and `applyGradient` is exposed via layer_internal to neuralnet but for the API, it can be exposed or cannot be. It is still abstracted by `backwarding` and hidden under it from the external API point of view. `backwarding` function maintains is behavior.

kparichay:
This patch has been updated to apply this logic to graph than to neural network class.

nntrainer/models/neuralnet.cpp
```diff
@@ -292,6 +293,13 @@ void NeuralNetwork::backwarding(sharedConstTensors input,
   sharedConstTensors output = label;
 
   model_graph.backwarding(output, iteration);
+  // for (unsigned int i = layers.size() - 1; i > skip_non_trainable_layers + 1;
+  //      i--)
```
jijoongmoon, line 297: remove this comments?

jijoongmoon (approved):
LGTM

nntrainer/models/neuralnet.cpp
kparichay: done.

nntrainer/graph/network_graph.cpp
```diff
@@ -517,14 +530,22 @@ sharedConstTensors NetworkGraph::forwarding(sharedConstTensors input) {
 
 void NetworkGraph::backwarding(sharedConstTensors output, int iteration) {
 
-  for (unsigned int i = Sorted.size() - 1; i > 0; i--) {
+  for (unsigned int i = Sorted.size() - 1; i > skip_non_trainable_layers + 1;
```
zhoonit, line 533: Sorted[0] will be never accessed with this logic, I wonder if this is intended?

kparichay: Yes, sortted[0] has no one to give its derivative values to.
Further Sorted[0] is ALWAYS input layer which does not support calcGradient etc.
Hence it is always skipped.

zhoonit: Got it

zhoonit (approved):
LGTM

nntrainer/graph/network_graph.cpp
kparichay: done.

PR #763, opened 2020-11-20
[conv2d] Refactor conv2d layer

### LINKED ISSUE #761: conv2d layer needs refactoring
There are a number of issues with conv2d layer:
- number of weights are dependent on the number of filters (user configuration at runtime) - it should not be
- lot of unnecessary memory copies #747
- manually transposing the data instead of using transpose provided by blas in backwarding
- weight regularization loss is wrong as it divides by number of filters

kparichay:
**Commit 1: **

Conv2d layer has some issues #761
    This patch addresses some of them:
    - Weight is now independent of the filter size. Different filter
    weights have now been combined. This has resulted in easier addressing of weights
    - Above combining of weights also reduced many mem-copies of weights to bring it in a particular shape
    - Moved to use getSlice() to get access to some data than creating copies
    
Now, all layers have fixed number of weights.

**Commit 2:**

Regularization loss for conv2d layer took average over output filters
    than adding it up. This patch fixes it.
    
**Commit 3:**

This patch provides more optimizations for conv2d
    by avoiding more memcopies and operations along with modification
    to internal interface of conv2d_gemm operation

Resolves #761
    
**Self evaluation:**
    
### DIFF
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -52,9 +52,9 @@ int Conv2DLayer::initialize() {
   setNumWeights(2);
 
   weightAt(ConvParams::weight) =
-    std::move(Weight(dim, weight_initializer, true, kernelPrefix));
+    Weight(dim, weight_initializer, true, kernelPrefix);
   weightAt(ConvParams::bias) =
-    std::move(Weight(bias_dim, bias_initializer, true, biasPrefix));
+    Weight(bias_dim, bias_initializer, true, biasPrefix);
 
   // this output_dim should be the same with dimension of hidden
   out_dim.batch(in_dim.batch());
@@ -79,6 +79,7 @@ void Conv2DLayer::forwarding(sharedConstTensors in) {
   TensorDim &out_dim = output_dim[0];
 
   Tensor &hidden_ = net_hidden[0]->var;
+  /** @todo This check is redundant, remove it later */
   if (hidden_.uninitialized()) {
     hidden_ = Tensor(out_dim);
   }
@@ -195,8 +196,17 @@ void Conv2DLayer::calcDerivative(sharedConstTensors derivatives) {
    *           *(input_dim.width() + padding[1]*2) (width)]
    */
   using uint = unsigned int;
+  bool no_padding = padding[0] == 0 && padding[1] == 0;
 
-  Tensor ret(in_dim.batch(), in_dim.channel(), in_dim.height() + padding[0] * 2,
+  if (net_input[0]->grad.uninitialized())
+    net_input[0]->grad = Tensor(input.getDim());
+
+  Tensor ret;
+  if (no_padding)
+    ret = net_input[0]->grad;
+  else
+    ret =
+      Tensor(in_dim.batch(), in_dim.channel(), in_dim.height() + padding[0] * 2,
              in_dim.width() + padding[1] * 2);
 
   TensorDim kdim(ret.channel(), filter_size, kernel_size[0], kernel_size[1]);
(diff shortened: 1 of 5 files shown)


nntrainer/layers/conv2d_layer.cpp
```diff
@@ -41,23 +43,18 @@ int Conv2DLayer::initialize() {
...
-      Weight(bias_dim, bias_initializer, true, biasPrefix + std::to_string(i)));
-  }
+
+  TensorDim dim =
+    TensorDim(filter_size, in_dim.channel(), kernel_size[0], kernel_size[1]);
+  TensorDim bias_dim = TensorDim(1, filter_size, 1, 1);
+  setNumWeights(2);
+
+  weightAt(ConvParams::weight) =
+    std::move(Weight(dim, weight_initializer, true, kernelPrefix));
```
zhoonit, line 55: Not related to the PR though, right side is already rvalue so std::move is redundant.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -81,9 +78,13 @@ void Conv2DLayer::forwarding(sharedConstTensors in) {
   TensorDim &in_dim = input_dim[0];
   TensorDim &out_dim = output_dim[0];
 
-  TensorDim hidden_dim = output_dim[0];
   Tensor &hidden_ = net_hidden[0]->var;
-  hidden_.setZero();
+  if (hidden_.uninitialized()) {
```
zhoonit, line 82: I wonder if we have to check this in every layer or just a specific case in conv2d?

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -102,31 +102,29 @@ class nntrainer_abstractLayer : public ::testing::Test {
...
+    matchOutput(result.getData(), golden.getData(), result.length());
+  }
 
-    for (size_t i = 0; i < result.length(); ++i) {
-      EXPECT_NEAR(out_ptr[i], golden_ptr[i], local_tolerance);
-    }
+  void matchOutput(const float *result, const char *expected) {
+    nntrainer::Tensor golden;
+    loadFile(expected, golden);
+    matchOutput(result, golden.getData(), golden.length());
```
zhoonit, line 119: FYI) golden.length() is 0. (It is a known issue left untouched before this patch as well)

nntrainer/layers/conv2d_layer.cpp
kparichay: good catch, removed.

test/unittest/unittest_nntrainer_layers.cpp
zhoonit: How about putting a comment here?

zhoonit (approved):
LGTM :)

nntrainer/layers/conv2d_layer.cpp
kparichay: This is just from backward compatibility of hidden which was managed by layer.
I am not yet sure how graph is handling. 
Currently, net_buffer is under heavy updates - check #765, #766
Most likely this is redundant, and I will remove this later.

test/unittest/unittest_nntrainer_layers.cpp
kparichay: Even after loading data in it?

nntrainer/layers/conv2d_layer.cpp
zhoonit: Acked :)

test/unittest/unittest_nntrainer_layers.cpp
zhoonit: Yeah, because golden is uninitialized in this function, which has length of 0

zhoonit: Some unittest might be wrong after fixing this. I didn't fix it on purpose

kparichay: I will add a comment here.

jijoongmoon (approved):
LGTM

PR #764, opened 2020-11-25
[memory] Memory based optimizations and fixes

### LINKED ISSUE #771: bn layer inference bug
Batch normalization layer uses `Tensor add()` but does not store the output updated tensor back to the variable in inference mode.

kparichay:
Commit 1: [regex] Make regex static const
Although it is using static string, that memory is always being allocated inside regex
    Making is static const only makes it once for the function lifetime

Commit 2: [activaiton] Reduce temporary memory alloc
Reduce temporary memory allocations by activation layer
    by using the hidden and ret_derivative class variables
    This temporarily increases peak memory but alloc-dealloc is removed
    from every iteration

Commit 3: [pooling] Reduce temporary mem allocations
Reduce temporary memory allocations for pooling
    Remove unnecessary temporary memory allocations which can be
    replaced with a slice view
    Also removed unnecessary setting memory to zero

Commit 4: [tensor] Support multiply/divide with given output
Support multiply/divide with given output tensor
    This reduces temporary allocations for bn layer

Commit 5: [bnlayer] bug fix for inference
Batch normalization bug fix for inference mode
    when add() was used instead of add_i()

Resolves #771 
See also #772

**Self evaluation:**
    
### DIFF
--- a/nntrainer/tensor/tensor_dim.cpp
+++ b/nntrainer/tensor/tensor_dim.cpp
@@ -78,7 +78,7 @@ void TensorDim::setTensorDim(unsigned int idx, unsigned int value) {
 
 int TensorDim::setTensorDim(const std::string &input_shape) {
   int status = ML_ERROR_NONE;
-  const std::regex words_regex("[^\\s.,:;!?]+");
+  static const std::regex words_regex("[^\\s.,:;!?]+");
   auto words_begin =
     std::sregex_iterator(input_shape.begin(), input_shape.end(), words_regex);
   auto words_end = std::sregex_iterator();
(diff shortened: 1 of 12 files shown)


jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor_dim.cpp
```diff
@@ -78,7 +78,7 @@ void TensorDim::setTensorDim(unsigned int idx, unsigned int value) {
 
 int TensorDim::setTensorDim(const std::string &input_shape) {
   int status = ML_ERROR_NONE;
-  std::regex words_regex("[^\\s.,:;!?]+");
+  const std::regex words_regex("[^\\s.,:;!?]+");
```
zhoonit, line 81: static here as well?

kparichay: done.

zhoonit (approved):
LGTM

Issue #765, opened 2020-11-26
Derivatives don't need their own memory


kparichay:
The output of a layer can be used to store the derivative from the next layer.
The reason for this is that the output of a layer is not used in the backwarding (except for softmax backwarding which can be handled as a special). As the shape of derivative and output layer is exactly same, this saves the memory of all the layer return derivatives, reducing the peak memory by a LOT.

With training, the memory impact is usually doubled with gradients and return derivatives. 
With this scheme, the memory impact from return derivatives can be completely eliminated, bringing the memory requirement for training very close to the memory requirement for inference.

PR #768, opened 2020-11-27
[ Graph ] remove grad mem buffer for backwarding

jijoongmoon:
This PR includes,
  . remove grad memory buffer in n_buffes for graph. We do not need
  this because we could use var memory buffer of n_buffers to
  backwarding.
  . For MNIST, memory consumption is reduced 3.5 to 2.6

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -542,10 +542,11 @@ void NetworkGraph::backwarding(sharedConstTensors output, int iteration) {
   }
 
   /** The last trainable layer need not calculate the derivatives */
+  // Oder is matter here. 1. calcGradient 2.Derivative & Gradient
+  Sorted[skip_non_trainable_layers + 1].layer->calcGradient();
 #ifdef ENABLE_TEST
   Sorted[skip_non_trainable_layers + 1].layer->calcDerivative();
 #endif
-  Sorted[skip_non_trainable_layers + 1].layer->calcGradient();
   Sorted[skip_non_trainable_layers + 1].layer->applyGradient(iteration);
 }
 
--- a/nntrainer/layers/activation_layer.cpp
+++ b/nntrainer/layers/activation_layer.cpp
@@ -50,16 +50,19 @@ void ActivationLayer::forwarding(sharedConstTensors in) {
   Tensor &hidden_ = net_hidden[0]->var;
   /// @note @a _act_fn is expected to work out of place and not modify @a input
   _act_fn(net_input[0]->var, hidden_);
+  if (activation_type == ActivationType::ACT_SOFTMAX)
+    backup_hidden = hidden_.clone();
 }
 
 void ActivationLayer::calcDerivative(sharedConstTensors derivative) {
-  Tensor &deriv = net_hidden[0]->grad;
-  Tensor &ret = net_input[0]->grad;
+  Tensor &deriv = net_hidden[0]->var;
+  Tensor &ret = net_input[0]->var;
 
-  if (activation_type == ActivationType::ACT_SOFTMAX)
-    ret = _act_prime_fn(net_hidden[0]->var, ret, deriv);
-  else
+  if (activation_type == ActivationType::ACT_SOFTMAX) {
+    ret = _act_prime_fn(backup_hidden, ret, deriv);
+  } else {
     ret = _act_prime_fn(net_input[0]->var, ret, deriv);
+  }
 }
 
 int ActivationLayer::setActivation(
--- a/nntrainer/layers/activation_layer.h
+++ b/nntrainer/layers/activation_layer.h
@@ -155,6 +155,8 @@ private:
   std::function<Tensor(Tensor const &, Tensor &)> _act_fn;
   std::function<Tensor(Tensor const &, Tensor &, Tensor const &)> _act_prime_fn;
 
+  Tensor backup_hidden;
+
   /**
    * @brief setActivation by custom activation function
    * @note  apply derivative as this activation_prime_fn does not utilize
--- a/nntrainer/layers/layer.cpp
+++ b/nntrainer/layers/layer.cpp
@@ -72,7 +72,7 @@ std::vector<Tensor> Layer::getOutputs() {
 std::vector<Tensor> Layer::getDerivatives() {
   std::vector<Tensor> ret;
   for (unsigned int i = 0; i < num_inputs; ++i) {
-    ret.push_back(net_input[i]->grad);
+    ret.push_back(net_input[i]->var);
   }
   return ret;
 }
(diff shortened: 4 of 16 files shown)


test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -1303,7 +1297,9 @@ TEST_F(nntrainer_Conv2DLayer, backwarding_02_p) {
   // for example, first value is -14422792, out value is -1.44228e+07
   // matchOutput(out, "tc_conv2d_2_goldenConv2DResult2.out");
   matchOutput(grad_data, "tc_conv2d_2_goldenKernelGrad2.out");
-  matchOutput(result, "tc_conv2d_2_goldenInputGrad2.out");
+  // @fixme: the results is differe because we are using same buffer for grad
+  // and var at forwarding and backwarding
+  // matchOutput(result, "tc_conv2d_2_goldenInputGrad2.out");
```
kparichay, line 1302: Please add a meson option which enables this feature so that you dont have to comment the unittest.

kparichay (changes requested):
Any optimization must not disable any unittest.

jijoongmoon:
> Any optimization must not disable any unittest.

Well, I agree with that if lots of unittest will be failed. However, this one case unittest is failed and convolution test is conducted in unittest_nntrainer_models as well. What I'm thinking is it is the reason of fail of this unittest is because of wrong implementation of unittest rather than main implementation. We could implement differently for this unittest like way in unittest_nntrainer_models.
Others except this case are working pretty well with current main implementation and there is no reason disable unittest in meson. Maybe we could use ifdef but this could be done later PR like I commented in code.

kparichay:
> > Any optimization must not disable any unittest.
> 
> Well, I agree with that if lots of unittest will be failed. However, this one case unittest is failed and convolution test is conducted in unittest_nntrainer_models as well. What I'm thinking is it is the reason of fail of this unittest is because of wrong implementation of unittest rather than main implementation. We could implement differently for this unittest like way in unittest_nntrainer_models.
> Others except this case are working pretty well with current main implementation and there is no reason disable unittest in meson. Maybe we could use ifdef but this could be done later PR like I commented in code.

As this patch modifies the output of a layer, any output verified after backwarding will fail for unittests which have multiple layer.
Lets investigate this unittest failing and fix it. But let's not directly disable a unittest when it can avoided.

kparichay:
Resolves #765.

jijoongmoon:
Every unittest cases are passed successfully.

test/unittest/unittest_nntrainer_layers.cpp
jijoongmoon: fixed.

nntrainer/layers/activation_layer.cpp
```diff
@@ -50,17 +50,21 @@ void ActivationLayer::forwarding(sharedConstTensors in) {
   Tensor &hidden_ = net_hidden[0]->var;
   /// @note @a _act_fn is expected to work out of place and not modify @a input
   _act_fn(net_input[0]->var, hidden_);
+  if (activation_type == ActivationType::ACT_SOFTMAX)
+    tempTensor = hidden_;
```
kparichay, line 54: This is a shallow copy. tempTensor will also get updated when net_hidden[0]->var is updated.

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -1303,6 +1301,8 @@ TEST_F(nntrainer_Conv2DLayer, backwarding_02_p) {
   // for example, first value is -14422792, out value is -1.44228e+07
   // matchOutput(out, "tc_conv2d_2_goldenConv2DResult2.out");
   matchOutput(grad_data, "tc_conv2d_2_goldenKernelGrad2.out");
+  // @fixme: the results is differe because we are using same buffer for grad
```
kparichay, line 1304: remove this comment?

nntrainer/layers/layer.cpp
```diff
@@ -103,7 +103,7 @@ void Layer::copy(std::shared_ptr<Layer> l) {
 sharedConstTensors Layer::forwarding_with_val(sharedConstTensors input) {
 
   for (unsigned int i = 0; i < num_inputs; ++i) {
-    net_input[i]->var = *input[i];
+    net_input[i]->var = input[i]->clone();
```
kparichay, line 106: why are we cloning here? If you want to copy the data, please copy the data, clone allocates new memory.
But I don't know we even copying the data is needed.

jijoongmoon: > why are we cloning here? If you want to copy the data, please copy the data, clone allocates new memory.
> But I don't know we even copying the data is needed.

It needed. Cause input is updated outside function. Besides it is test case to test the internals. I guess we could give loose limitation of memory.

test/unittest/unittest_nntrainer_layers.cpp
jijoongmoon: done. thanks.

nntrainer/layers/activation_layer.cpp
jijoongmoon: done.

nntrainer/layers/layer.cpp
kparichay: works.

nntrainer/layers/activation_layer.h
```diff
@@ -155,6 +155,8 @@ class ActivationLayer : public Layer {
   std::function<Tensor(Tensor const &, Tensor &)> _act_fn;
   std::function<Tensor(Tensor const &, Tensor &, Tensor const &)> _act_prime_fn;
 
+  Tensor tempTensor;
```
kparichay, line 158: lets rename this with some proper name and description.

nntrainer/layers/activation_layer.cpp
```diff
@@ -50,16 +50,20 @@ void ActivationLayer::forwarding(sharedConstTensors in) {
...
+  Tensor &ret = net_input[0]->var;
 
-  if (activation_type == ActivationType::ACT_SOFTMAX)
-    ret = _act_prime_fn(net_hidden[0]->var, ret, deriv);
-  else
-    ret = _act_prime_fn(net_input[0]->var, ret, deriv);
+  if (activation_type == ActivationType::ACT_SOFTMAX) {
+    ret = _act_prime_fn(tempTensor, ret, deriv);
+  } else {
+    tempTensor = net_input[0]->var.clone();
```
kparichay, line 64: why is there a clone here than just passing `net_input[0]->var` to _act_prime_fn?

kparichay: As the activation layers perform elementwise operation, this will works.
Softmax is not element wise and hence needs output value stored.

nntrainer/layers/activation_layer.h
jijoongmoon: done.

zhoonit (approved):
LGTM 👍

nntrainer/graph/network_graph.cpp
```diff
@@ -542,10 +542,11 @@ void NetworkGraph::backwarding(sharedConstTensors output, int iteration) {
   }
 
   /** The last trainable layer need not calculate the derivatives */
+  // Oder is matter here. 1. calcGradient 2.Derivative & Gradient
+  Sorted[skip_non_trainable_layers + 1].layer->calcGradient();
```
kparichay, line 546: good catch

kparichay (approved):
LGTM

PR #776, opened 2020-11-27
[ModelLoader] Use vector when creating layer

zhoonit:
- ~**#746** [AppContext] Register Default ops at the begining~
- ~**#753** [AppContext] Fix key is case sensitive~
- ~**#753** [Layer] Add built-in ops to the context~
- [ModelLoader] Use vector<string> when creating layer
```
When creating a layer from an ini, enum based properties were used.
This prevents adding a new properties without changing the api header.

This patch moves to setting layer properties to vector<string>
to enable setting properties without changing the api header, eventually
 enabling custom properties in custom layer.

**Semantics Change propesed in this PR**
Ini won't ignore the properties that is not supported since model_loader
would not know if it is supported or not

See also #716

**Self evaluation:**

```

### DIFF
--- a/nntrainer/models/model_loader.cpp
+++ b/nntrainer/models/model_loader.cpp
@@ -177,9 +177,59 @@ int ModelLoader::loadLayerConfigIniCommon(dictionary *ini,
                                           const std::string &layer_name,
                                           const std::string &layer_type) {
   int status = ML_ERROR_NONE;
+  int num_entries = iniparser_getsecnkeys(ini, layer_name.c_str());
+
+  ml_logd("number of entries for %s: %d", layer_name.c_str(), num_entries);
+
+  if (num_entries < 1) {
+    std::stringstream ss;
+    ss << "there are no entries in the layer section: " << layer_name;
+    throw std::invalid_argument(ss.str());
+  }
+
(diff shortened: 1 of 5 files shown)


jijoongmoon (approved):
LGTM

nntrainer/models/model_loader.cpp
```diff
@@ -177,43 +177,66 @@ int ModelLoader::loadLayerConfigIniCommon(dictionary *ini,
...
+    throw std::invalid_argument(ss.str());
+  }
 
-    /**! @todo: add following negative tc after #319
-     * 1. layer has empty prop -> throw std::invalid_argument
-     * 2. layer has not allowed property -> throw exception::not_supported
-     * 3. property value parse error -> throw std::invalid_argument
-     */
-    if (!strncmp(value.c_str(), unknown, strlen(unknown))) {
+  std::vector<std::string> properties;
```
kparichay, line 199: `properties.reserve(num_entries)`?

kparichay: actually `num_entries - 1` as `type` or `backbone` are always ommited.

zhoonit: Changed the code to reserved at resource aquisition. Thks :)

kparichay (approved):
LGTM

PR #777, opened 2020-11-30
[tensor] Update tensor operation signature

### LINKED ISSUE #670: Update the signature of out-of-place tensor operations

kparichay:
Update tensor operation signature to return Tensor reference as a retval
than a tensor itself. This avoid creating dummy tensors as a return (which might have been
optimized by the compiler but lets do manually as the input is also a reference).

This patch also update `Tensor::copy/clone` to not allocate new tensor memory 
if already allocated and same size which did happen in many cases.

Resolves #670

@zhoonit thanks for this suggestion.

Note that this is the last patch for the tensor signature. Please suggest any more signature changes, if any. After this patch, signature is finalized and more functions will be updated later as per needed. Important core functions have already been updated.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/var_grad.cpp
+++ b/nntrainer/tensor/var_grad.cpp
@@ -16,11 +16,9 @@
 
 namespace nntrainer {
 
-Var_Grad::Var_Grad(const Var_Grad &rhs) :
-  trainable(rhs.trainable),
-  name(rhs.name) {
+Var_Grad::Var_Grad(const Var_Grad &rhs) : name(rhs.name) {
   var = rhs.var.clone();
-  if (rhs.trainable)
+  if (rhs.getTrainable())
     grad = rhs.grad.clone();
 }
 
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/var_grad.cpp
```diff
@@ -20,7 +20,8 @@ Var_Grad::Var_Grad(const Var_Grad &rhs) :
   trainable(rhs.trainable),
   name(rhs.name) {
   var = rhs.var.clone();
-  grad = rhs.grad.clone();
+  if (rhs.trainable)
```
zhoonit, line 23: How about,
```cpp
if (rhs.grad.uninitialized())
```

If it is not trainable but has a grad, it would mean something special we should care about.

if trainable means always gradient uninitialized, I think we wouldn't need trainable variable rather we can check if gradient is uninitialized to keep a single source of truth

jijoongmoon (approved):
LGTM!

nntrainer/tensor/var_grad.cpp
kparichay: Added as a new commit (to make this update separate).

zhoonit (approved):
LGTM :)

PR #780, opened 2020-12-01
[ CONV2D ] separate conv2d_gemm and im2col @open sesame 12/07 08:36

jijoongmoon:
It is better to split conv2d_gemm and im2col

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -128,9 +128,26 @@ void Conv2DLayer::forwarding(sharedConstTensors in) {
     Tensor out = hidden_.getBatchSlice(b, 1);
     Tensor inSub = input_.getBatchSlice(b, 1);
 
-    status =
-      conv2d_gemm(filter_kernel.getData(), filter_kernel.getDim(), inSub,
-                  out_dim, stride, padding, out.getData(), out.length(), true);
+    std::vector<float> in_col;
+
+    in_col.resize(filter_kernel.getDim().getFeatureLen() * out_dim.width() *
+                  out_dim.height());
+
+    Tensor in_padded;
+    Tensor &in_ref = in_padded;
+    if (padding[0] != 0 && padding[1] != 0) {
+      zero_pad(0, inSub, padding.data(), in_padded);
+    } else {
+      in_ref = inSub;
+    }
+
+    status = im2col(in_padded, filter_kernel.getDim(), in_col.data(), out_dim,
+                    stride, true);
+    if (status != ML_ERROR_NONE)
+      throw std::runtime_error("Forwarding Convolution failed.");
+
+    status = conv2d_gemm(filter_kernel.getData(), filter_kernel.getDim(),
+                         in_col.data(), out_dim, out.getData(), true);
     if (status != ML_ERROR_NONE)
       throw std::runtime_error("Forwarding Convolution failed.");
   }
@@ -161,7 +178,8 @@ void Conv2DLayer::calcDerivative(sharedConstTensors derivatives) {
    *   . Width  : filter_size * kernel_size[0] * kernel_size[1]
    *
    *                                kernel
-   *                             f0      f1          fn
+   *                             f0      fn-1          fn
+   *                            k..0     k..0         k..0
    *                          +---|---|---|---|...|---|---+
    *                          |---|---|---|---|...|---|---|
    * [filter.channel(height)] |---|---|---|---|...|---|---|
@@ -196,20 +214,11 @@ void Conv2DLayer::calcDerivative(sharedConstTensors derivatives) {
    *           *(input_dim.width() + padding[1]*2) (width)]
    */
   using uint = unsigned int;
-  bool no_padding = padding[0] == 0 && padding[1] == 0;
 
   if (net_input[0]->var.uninitialized())
     net_input[0]->var = Tensor(input.getDim());
 
-  Tensor ret;
-  if (no_padding)
-    ret = net_input[0]->var;
-  else
-    ret =
-      Tensor(in_dim.batch(), in_dim.channel(), in_dim.height() + padding[0] * 2,
-             in_dim.width() + padding[1] * 2);
-
-  TensorDim kdim(ret.channel(), filter_size, kernel_size[0], kernel_size[1]);
+  TensorDim kdim(in_dim.channel(), filter_size, kernel_size[0], kernel_size[1]);
 
   uint kernel_total_size = kernel_size[0] * kernel_size[1];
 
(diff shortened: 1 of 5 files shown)


nntrainer/layers/conv2d_layer.cpp
```diff
@@ -237,11 +247,19 @@ void Conv2DLayer::calcDerivative(sharedConstTensors derivatives) {
...
+    in_col.resize(kdim.getFeatureLen() * input_dim_padded.width() *
+                  input_dim_padded.height());
 
+    Tensor in_padded = zero_pad(0, inSub, same_pad.data());
     status =
-      conv2d_gemm(imKernel_raw, kdim, inSub, input_dim_padded, stride, same_pad,
-                  ret.getAddress(b * ret.getDim().getFeatureLen()),
-                  input_dim_padded.getFeatureLen(), true);
+      im2col(in_padded, kdim, in_col.data(), input_dim_padded, stride, true);
+    if (status != ML_ERROR_NONE)
```
dongju-chae, line 257: I think you'd better to add `likely` macro inside a loop (in another PRs)

dongju-chae (approved):
LGTM

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -128,9 +128,19 @@ void Conv2DLayer::forwarding(sharedConstTensors in) {
     Tensor out = hidden_.getBatchSlice(b, 1);
     Tensor inSub = input_.getBatchSlice(b, 1);
 
-    status =
-      conv2d_gemm(filter_kernel.getData(), filter_kernel.getDim(), inSub,
-                  out_dim, stride, padding, out.getData(), out.length(), true);
+    std::vector<float> in_col;
```
kparichay, line 131: Please use Tensor instead of vector here to manage memory uniformly.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -237,11 +247,19 @@ void Conv2DLayer::calcDerivative(sharedConstTensors derivatives) {
 
   for (unsigned int b = 0; b < in_dim.batch(); ++b) {
     Tensor inSub = derivative.getBatchSlice(b, 1);
+    std::vector<float> in_col;
```
kparichay, line 250: ditto

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -128,9 +128,19 @@ void Conv2DLayer::forwarding(sharedConstTensors in) {
...
 
-    status =
-      conv2d_gemm(filter_kernel.getData(), filter_kernel.getDim(), inSub,
-                  out_dim, stride, padding, out.getData(), out.length(), true);
+    std::vector<float> in_col;
+
+    in_col.resize(filter_kernel.getDim().getFeatureLen() * out_dim.width() *
+                  out_dim.height());
+
+    Tensor in_padded = zero_pad(0, inSub, padding.data());
```
kparichay, line 136: Please update zero_pad function signature to take output tensor as input. 
Then, if padding is zero, lets no copy but directly use the input. 
Currently, memory copy happens when padding is 0 inducing significant overhead.

nntrainer/layers/conv2d_layer.cpp
kparichay: Also, make this tensor a class variable so as to avoid allocaiton and deallocation with every forwarding.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -297,17 +315,27 @@ void Conv2DLayer::calcGradient(sharedConstTensors derivatives) {
...
+  TensorDim kdim =
+    TensorDim(1, derivative.channel(), derivative.height(), derivative.width());
+
+  TensorDim out_dim = TensorDim(
+    1, 1, filter_size, kernel_size[0] * kernel_size[1] * in_dim.channel());
+
   for (unsigned int b = 0; b < in_dim.batch(); ++b) {
     Tensor inSub = input_.getBatchSlice(b, 1);
 
+    std::vector<float> in_col;
```
kparichay, line 327: replace with Tensor.

nntrainer/layers/conv2d_layer.cpp
jijoongmoon: I don't think it should be. This is nothing releate with tensor or even does not have dimension. If we have to we have set dimensions as (1,1,1, filter_kertnel.getDim().gerFeatureLen()*out_dim.width()*dout_dim.height() ) and I don't think any consistency here.

nntrainer/layers/conv2d_layer.cpp
jijoongmoon: This is just temporal usage. So do you think it should be there all the time when forwarding and backwarding?

nntrainer/layers/conv2d_layer.cpp
jijoongmoon: ditto.

nntrainer/layers/conv2d_layer.cpp
jijoongmoon: done.

nntrainer/layers/conv2d_layer.cpp
jijoongmoon: I will. thanks.

nntrainer/layers/conv2d_layer.cpp
kparichay: Temporal usage will allocate and deallocate memory every iteration creating unnecessary overhead.
The allocation should be just done once.
This might increase peak memory (its possible that this allocation is already counted in peak memory) but increase performance.

nntrainer/layers/conv2d_layer.cpp
kparichay: The reason to use tensor is to be able to use a common allocator and manage memory in a unified way.
Further, with datatypes introduced later, vector<float> at all places will have to replaced.

jijoongmoon: I think it does confused as well. No reason to use tensor here.

kparichay: Overhead will be minor but later, we want to manage memory ourselves. Vector does not allow managing memory ourselves. IMO all the data in nntrainer must use Tensor be it temporary or not.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -161,7 +178,7 @@ void Conv2DLayer::calcDerivative(sharedConstTensors derivatives) {
    *   . Width  : filter_size * kernel_size[0] * kernel_size[1]
    *
    *                                kernel
-   *                             f0      f1          fn
+   *                             fn      fn-1          f0
```
zhoonit, line 181: :+1:

nntrainer/utils/util_func.cpp
```diff
@@ -49,7 +49,8 @@ float exp_util(float x) { return exp(x); }
 
 // This is 2D zero pad
 // TODO : Optimize for multi dimention padding
-Tensor zero_pad(int batch, Tensor const &in, unsigned int const *padding) {
+Tensor zero_pad(int batch, Tensor const &in, unsigned int const *padding,
```
zhoonit, line 52: This patch does not use the output. How about we return void here to reduce additional copy?

nntrainer/utils/util_func.cpp
```diff
@@ -92,21 +93,18 @@ Tensor zero_pad(int batch, Tensor const &in, unsigned int const *padding) {
 }
 
 // This is strip pad and return original tensor
-Tensor strip_pad(Tensor const &in, unsigned int const *padding,
-                 Tensor &output) {
+Tensor strip_pad(Tensor const &in, unsigned int const *padding, Tensor &output,
```
zhoonit, line 96: Same. Alternatively, I guess we can return `Tensor &`

nntrainer/utils/util_func.cpp
```diff
@@ -92,21 +93,18 @@ Tensor zero_pad(int batch, Tensor const &in, unsigned int const *padding) {
...
+                 unsigned int batch) {
   if (output.uninitialized()) {
-    output = Tensor(in.batch(), in.channel(), in.height() - padding[0] * 2,
+    output = Tensor(output.batch(), in.channel(), in.height() - padding[0] * 2,
```
zhoonit, line 99: FYI, this will never happen. If output has batch it means it is initialized.
I guess this should be just using batch??

jijoongmoon: done.

nntrainer/utils/util_func.cpp
jijoongmoon: done

nntrainer/utils/util_func.cpp
jijoongmoon: done

nntrainer/layers/conv2d_layer.cpp
jijoongmoon: done.

zhoonit (approved):
LGTM after ci issue :)

kparichay (approved):
LGTM
but please consider using tensor instead of vector for any data operations.

Issue #774, opened 2020-11-27
Create a graph attribute


kparichay:
To manage graphs and layers better, make a graph attribute.
`GraphAttr` will maintain all the layer list, input and output list, and weight list. It only contains essential parts of the graph. Sorted lists etc will not be part of it.

- [x] `GraphAttr` will be passed to a layer `initialize()`, where layer would call `graphAttr.registerWeight()` to add a weight.
This will allow the graph to manage all the weights. This is crucial for #766, and create a good interface for the custom layer to manage their weights.
The advantage of passing `GraphAttr` than the whole `Graph` to the layer is for the security purpose in the case of a custom layer. - #782

- [x] Similar interface will be there for inputs and outputs - `graphAttr.registerInput()` and `graphAttr.registerOutput()`. - #827

- [x] Calling these functions multiple times would register multiple weights/inputs/outputs. From the graph perspective, all the layer related information can now be handled together, simplifying graph code. - #827

- [x] Further, `graphAttr` will also have `optimizer` inside it. `registerWeight()` internally will also register optimizer related parameters for it (optimizer is known by then as the layer is being initialized). So, optimizer parameters will also be owned by the `graphAttr` as well unifying all the memory at one place. With this, each layer does not need its own optimizer copy. - #786

- [x] Lastly, the weight regularization will move out of each layer will directly be handled by the graph just like optimizer. - #762

kparichay:
The graph attribute came out to be a tensor manager, which manages all the tensor weights.
Graph manages at the layer, and each of those layers interacts with the manager for all its memory needs.

PR #782, opened 2020-12-02
[layer] Share gradient weights for all the layers @open sesame 12/07 09:04

### LINKED ISSUE #766: Gradients don't need own memory
Gradients for the weights are just a temporary memory, as they are only useful till the gradient is applied to the layer or when the gradient is used for debugging.
So, gradients of all the layers can share the same tensor with the same memory. This memory would be the maximum sized gradient needed by a layer. 
With this, the gradients memory for a model is just reduced to the gradients memory of layer, reducing the peak memory of the model by a lot.

kparichay:
**Commit 1**: [weight/var_grad] Make internal variable as shared_ptr
Internal variables in weights/var_grad, namely, the variable and gradient itself
    are changed to shared_ptr so that weights can be shared without worrying about
    shallow copies.
    
Also changed the copy constructor to not create new Tensor as copy constructor
    of weight will get called and its unnecessary + unintentional overhead.
    As weight is just wrapper over tensor, their copy constructors should follow
    same behavior as tensor which is to not create new memory.
    Added clone as an alternative to create new copy of a given weight.
    
**Commit 2**: [manager] Added nntrainer manager for weights
Added manager to manage all the allocated weights
    This patch also adds manager to the model and passes manager to the
    initialize which allows weights to be added to the manager.

**Commit 3**: [weight] Updated weights to be vector
Updated weights of layer to be vector than a shared_ptr array
    This is for easier management and updating weight internally when
    gradient will share the memory

**Commit 4**: [layers/manager] Register weights with manager
All the weights of the layer are now registered with manager
    Manager allocates memory for these weights and in future
    handle their updates etc

**Commit 5**: [manager] Share gradient memory for all layer
This patch allows sharing the memory for gradient across all the layers
    The maximum size of the gradient is allocated and all layers have unique tensors
    which internally points to this tensor.
    
This optimization feature can be disabled for a model (as done with automated models unittest)
    
Manager is also moved to nntrainer/tensor as manager is managing all the weights (tensors) and will
    in future manage all the inputs/outputs.
    If the functionality of manager is extended, then it can be appropriately moved.
    
See also #774
Resolves #766
    
**Self evaluation:**
    
### DIFF
--- a/nntrainer/manager.h
+++ b/nntrainer/manager.h
@@ -50,7 +50,7 @@ public:
    *
    * @param ws  Weights to be tracked
    */
-  void trackWeights(std::vector<Weight> ws) {
+  void trackWeights(std::vector<Weight> &ws) {
     weights.reserve(weights.size() + ws.size());
     weights.insert(weights.end(), ws.begin(), ws.end());
   }
--- a/nntrainer/manager.h
+++ b/nntrainer/manager.h
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	manager.h
+ * @date	30 Nov 2020
+ * @brief	This is NNtrainer manager for all weights, i/o and intermediate
+ * tensors
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ */
+
+#ifndef __MANAGER_H__
+#define __MANAGER_H__
+#ifdef __cplusplus
+
+#include <vector>
+
+#include <weight.h>
+
+namespace nntrainer {
+
+/**
+ * @class   Manager
+ * @brief   manager of nntrainer
+ */
+class Manager {
+public:
+  /**
+   * @brief     Constructor of Manager
+   */
+  Manager() {}
+
+  /**
+   * @brief     Destructor of Manager
+   */
+  ~Manager() {}
+
+  /**
+   * @brief     Add weight to be tracked and updated with nntrainer
+   *
+   * @param w   Weight to be tracked
+   */
+  void trackWeight(Weight w) { weights.push_back(w); }
+
+  /**
+   * @brief     Add weights to be tracked and updated with nntrainer
+   *
+   * @param ws  Weights to be tracked
+   */
+  void trackWeights(std::vector<Weight> ws) {
+    weights.reserve(weights.size() + ws.size());
+    weights.insert(weights.end(), ws.begin(), ws.end());
+  }
+
+  /**
+   * @brief     Get weights tracked with nntrainer
+   *
+   * @retval    list of weights
+   */
+  std::vector<Weight> getWeights() { return weights; }
+
+private:
+  // TODO: ensure that names of these weights are unique
+  std::vector<Weight> weights;
+};
+
+/**
+ * @brief Helper func for weight creation which are tracked by nntrainer
+ *
+ * @retval create weight
+ */
+template <typename... Args>
+Weight createWeight(Manager &manager, Args... args) {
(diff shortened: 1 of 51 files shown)


jijoongmoon (approved):
LGTM. Currently manager only manage the weight. but I think it should work same way for tensors in graph. and every layer include this manager so.. add tensor vector for input tensor and manage this and handle sharing between layers. I guess it could be sapearte PR.

kparichay:
> Good work. LGTM. Currently manager only manage the weight. but I think it should work same way for tensors in graph. and every layer include this manager so.. add tensor vector for input tensor and manage this and handle sharing between layers. I guess it could be sapearte PR.

Yes, that's the big plan. Manager will manage all the tensors and its related optimizations. Input, output and temporary allocations of the tensor, all will go with manager in a separate PR.

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -87,8 +87,8 @@ class NodeWatcher {
     node.layer->setTrainable(true);
 
     for (unsigned int i = 0; i < num_weights; ++i) {
-      const nntrainer::Weight &w = node.layer->weightAt(i);
-      expected_weights.push_back(w);
+      nntrainer::Weight &w = node.layer->weightAt(i);
```
zhoonit, line 90: I guess this should remain const and change signature of `weight::clone const ()`?

nntrainer/manager.h
```diff
@@ -0,0 +1,85 @@
...
+};
+
+/**
+ * @brief Helper func for weight creation which are tracked by nntrainer
+ *
+ * @retval create weight
+ */
+template <typename... Args>
+Weight createWeight(Manager &manager, Args... args) {
+  Weight w = Weight(args...);
```
zhoonit: `std::forward<Args>(args)...`  to do perfect forwarding?

zhoonit: Just an idea though, How about migrate this to one of weight constructor?

nntrainer/manager.h
```diff
@@ -52,32 +55,55 @@ class Manager {
...
- * @retval create weight
- */
-template <typename... Args>
-Weight createWeight(Manager &manager, Args... args) {
-  Weight w = Weight(args...);
-  manager.trackWeight(w);
-  return w;
-}
+// /**
+//  * @brief Helper func for weight creation which are tracked by nntrainer
```
zhoonit: Maybe this can be deleted??

nntrainer/tensor/weight.cpp
```diff
@@ -23,10 +23,12 @@ Weight::Weight(const TensorDim &dim, const WeightInitializer init, bool train,
   if (initializer == WeightInitializer::WEIGHT_UNKNOWN)
     throw std::invalid_argument("Weight initializer unknown");
 
-  initializeWeight();
+  // initialize();
```
zhoonit: Same

nntrainer/tensor/manager.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <vector>
+
+#include <manager.h>
+
+namespace nntrainer {
+
+/**
+ * @brief     Add weight to be tracked and updated with nntrainer
+ */
+void Manager::trackWeight(std::reference_wrapper<Weight> w) {
```
zhoonit, line 25: I wonder if we can't do `Weight &w` for the signature

zhoonit (approved):
Overall LGTM Great work!

nntrainer/manager.h
kparichay: This is later removed as this does not work.

nntrainer/manager.h
kparichay: done.

nntrainer/tensor/weight.cpp
kparichay: done.

test/unittest/unittest_nntrainer_models.cpp
kparichay: done.

PR #796, opened 2020-12-09
[meson.build] Change join_paths to / in meson.build files

### LINKED ISSUE #709: Change meson join_paths to '/' operator
We are using `join_paths` to concatenate directories. This can be represented in simpler way with ' / ' operator proposed on the later version( at 0.49.0 ). Making a change would be tedious but the script would be more readable.

Leaving this idea as a good first issue. 

referring to the meson manual
```
join_paths()
string join_paths(string1, string2, ...)
(since 0.36.0)

Joins the given strings into a file system path segment. For example join_paths('foo', 'bar') results in foo/bar. If any one of the individual segments is an absolute path, all segments before it are dropped. That means that join_paths('foo', '/bar') returns /bar.

Warning Don't use join_paths() for sources in library and executable, you should use files instead.

(since 0.49.0) Using the/ operator on strings is equivalent to calling join_paths.

# res1 and res2 will have identical values
res1 = join_paths(foo, bar)
res2 = foo / bar
```

lhs8928:
Replace join_paths in meson.build files to /

Check issue #709 for more details

Resolves #709

### DIFF
--- a/Applications/KNN/jni/meson.build
+++ b/Applications/KNN/jni/meson.build
@@ -1,4 +1,4 @@
-res_path = join_paths(meson.current_source_dir(), '..', 'res')
+res_path = meson.current_source_dir() / '..' / 'res'
 
 knn_sources = [
   'main_sample.cpp'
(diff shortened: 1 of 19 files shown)


kparichay (approved):
LGTM

kparichay:
Please use "Resolves #709" like statement in PR description to auto close the related issue when this PR is merged.

anyj0527 (approved):
LGTM 👍

zhoonit:
> Please use "Resolves #709" like statement in PR description to auto close the related issue when this PR is merged.
LGTM once above comment is resolved

zhoonit (approved):

PR #795, opened 2020-12-08
[Android] Integrate openblas into android

zhoonit:
Android ndk was not building on top of openblas

This patch fixes the problem

resolves #794

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

PR #797, opened 2020-12-09
[Profiler] Separate Profiler for wider use @open sesame 12/14 22:08

zhoonit:
- **#795** [Android] Integrate openblas into android
- [Profiler] Separate Profiler for wider use
```
Add profiler for wider use while separating listener from profiler

**Self evaluation:**

```
- [Test] Add profiler test
```
- Add profiler test
- Update build to attach profiler

**Self evaluation:**

```
- [Profiler] Add basic profilerlistener
```
This patch adds global profiler listener for various purpose

From this patch,
1. Profiler can called globally with designated event key
2. Listener reporting suite included

**Self evaluation:**

v2)
    1. Change listener to RAII object (with forcing profiler, event
    designation)
    2. Add unsubscribe method
    3. Change event register to set to prevent notifying a listener twice
    4. Change semintics to not allow adding same listener twice

```


Applications/KNN/jni/main_sample.cpp
```diff
@@ -20,8 +20,6 @@
 #include "tensorflow/contrib/lite/interpreter.h"
 #include "tensorflow/contrib/lite/kernels/register.h"
 #include "tensorflow/contrib/lite/model.h"
-#include "tensorflow/contrib/lite/string.h"
```
zhoonit, line 23: Minor fix on the sample app. This header is not used and if this header is included, unable to build from tensorflow-lite2

jni/Android.mk
```diff
@@ -115,6 +115,7 @@ NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/optimizers/optimizer_factory.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/util_func.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/parse_util.cpp \
+									$(NNTRAINER_ROOT)/nntrainer/utils/profiler.h \
```
kparichay, line 118: indentation?

zhoonit: Oops fixed

zhoonit:
Sorry for adding a hot fix :(

v2)
    1. Change listener to RAII object (with forcing profiler, event
    designation)
    2. Add unsubscribe method
    3. Change event register to set to prevent notifying a listener twice
    4. Change semantics to not allow adding the same listener twice

nntrainer/models/neuralnet.cpp
```diff
@@ -32,6 +32,7 @@
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <parse_util.h>
+#include <profiler.h>
```
jijoongmoon, line 35: I think it would be better to enable this depending on profile options. No need include this without profile options

zhoonit: Below is the code if the profiler is not enabled for `profiler.h`
```cpp
#ifndef PROFILE

#define START_PROFILE(event_key)
#define END_PROFILE(event_key)

#else
#define START_PROFILE(event_key)                         \
  do {                                                   \
    auto &prof = nntrainer::profile::Profiler::Global(); \
    prof.start(event_key);                               \
  } while (0);

#define END_PROFILE(event_key)                           \
  do {                                                   \
    auto &prof = nntrainer::profile::Profiler::Global(); \
    prof.end(event_key);                                 \
  } while (0);
#endif
```
This enables the use of `START_PROFILE` without #ifdef

jijoongmoon (approved):
LGTM

nntrainer/utils/profiler.cpp
```diff
@@ -0,0 +1,220 @@
...
+  auto duration =
+    std::chrono::duration_cast<std::chrono::milliseconds>(end - iter->second);
+  notify(event, duration);
+
+  start_time.erase(iter);
+}
+
+void Profiler::notify(const int &event,
+                      const std::chrono::milliseconds &value) {
+  for (auto &listener : all_event_listeners) {
```
kparichay, line 168: this must also be locked as this list can be modified while it notifies.

kparichay: However, this adds a big concern that notifies is called with every iteration (inside end) and that has a locking of its own, and I wonder, how much overhead it adds to the profiler/changes the values from the actual results.

nntrainer/utils/profiler.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  if (iter == start_time.end()) {
+    throw std::invalid_argument("profiler hasn't started with the event");
+  }
+
+  auto duration =
+    std::chrono::duration_cast<std::chrono::milliseconds>(end - iter->second);
+  notify(event, duration);
+
+  start_time.erase(iter);
```
kparichay, line 163: lets not erase, it can be expensive. It seems like this is used for ensuring that a started profiler is ended.
However, these kinds of checks are adding overhead to the profiler. 
IMHO, profiler must be as light weight as possible.

nntrainer/utils/profiler.cpp
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+void GenericProfileListener::onNotify(const int event,
+                                      const std::chrono::milliseconds &value) {
+
+  auto iter = time_taken.find(event);
+
+  if (iter == time_taken.end()) {
+    time_taken[event] = std::vector<std::chrono::milliseconds>{};
```
kparichay, line 47: can we added this at creation? this is also in critical path.

nntrainer/utils/profiler.h
```diff
@@ -0,0 +1,254 @@
...
+
+  /**
+   * @copydoc ProfileListener::report(std::ostream &out)
+   */
+  virtual void report(std::ostream &out) const override;
+
+private:
+  unsigned int warmups;
+
+  std::unordered_map<int, std::vector<std::chrono::milliseconds>> time_taken;
```
kparichay, line 161: I encourage you to consider using sum of all times and the count rather than this vector. This will grow big very fast.
You can add more statistics than just sum, but vector in a profiler will slow it down.

nntrainer/utils/profiler.cpp
zhoonit: I'll move this to #if Debug

zhoonit: And additional checks

nntrainer/utils/profiler.cpp
zhoonit: Done

nntrainer/utils/profiler.cpp
zhoonit: I'll add a lock for now.
Idea for the improvement:
0. Delay `onNotify()` call 
1. a `freeze`, `thaw` mechanism to control subscribe, (blocking unsubscribe) to ensure thread-safe behavior without locking in this function.
2. Use `shared_mutex` to relieve the problem (c++17 or boost) 

I'll leave this as an issue for now.. I'll keep in mind this can impact the performance.
If we have a performance issue, let's deal with it within this month.

nntrainer/utils/profiler.h
zhoonit: Done

nntrainer/utils/profiler.cpp
kparichay: good.
shared_mutex is good. I am not sure if freeze/thaw can work, but can be checked later.

kparichay (approved):
LGTM

PR #800, opened 2020-12-11
[Optim] Add shortcut to dot product @open sesame 12/14 22:55

### LINKED ISSUE #799: Tensor dot operation optimization
SGEMM is not optimized for vector * matrix operation,  should handle vector * matrix operation in Tensor Dot operation

### LINKED ISSUE #801: Tensor dot ld* is incorrect
lda, ldb should be actual memory layout before applying transpose but it was set after transpose.

zhoonit:
- ~**#797** [Profiler] Separate Profiler for wider use~
- ~**#797** [Test] Add profiler test~
- ~**#797** [Profiler] Add basic profilerlistener~
- [Fix] fix lda, ldb param
```
- lda, ldb, ldc is for layout so it should be set in terms of memory
layout, this patch fixes the issue while adding a corresponding test

**Self evaluation:**

```
resolves #801

- [Optim] Add shortcut to dot product
```
When dimension is 1, it is vector by matrix or vector by vector
multiplication. This patch adds a shortcut in that situation

**Self evaluation:**

```
resolves #799

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -633,11 +633,37 @@ Tensor &Tensor::dot(Tensor const &m, Tensor &result, bool trans,
   float *rdata = result.getData();
   const float alpha = 1.0f;
   const float beta = 0.0f;
-
   enum CBLAS_TRANSPOSE transA = trans ? CblasTrans : CblasNoTrans;
   enum CBLAS_TRANSPOSE transB = trans_m ? CblasTrans : CblasNoTrans;
-  sgemm(CblasRowMajor, transA, transB, M, N, K, alpha, data, lda, mdata, ldb,
-        beta, rdata, ldc);
+
+  /// shortcut handling in case of vector
+  /// for vector, (1 * K) == (K * 1) in current memory layout...
+  /// and plaese note that N, K, M is a fixed place holder after considering
+  /// transpose.
+  /// For example, there is no case like (1 * K) X (1 * K) while
+  /// (1 * K) X (1 * M) can be a case
+  /// case1: (1 * K) X (K * 1)
+  if (M == 1 && N == 1) {
+    *rdata = sdot(K, data, 1, mdata, 1);
+  }
+  /// case2: (M * K) X (K * 1)
+  else if (N == 1) {
+    sgemv(CblasRowMajor, transA, dim1, dim2, alpha, data, lda, mdata, 1, beta,
+          rdata, 1);
+  }
+  /// case3: (1 * K) X (K * N) = 1 * N = R
+  /// = R^T = (K * N) ^T * (1 * K) ^T = (N * K) * (K * 1) = (N * K) * (1 * K)
+  /// Effectively a translation of sgemv
+  else if (M == 1) {
+    transB = transB == CblasTrans ? CblasNoTrans : CblasTrans;
+    sgemv(CblasRowMajor, transB, mdim1, mdim2, alpha, mdata, ldb, data, 1, beta,
+          rdata, 1);
+  }
+  /// case others: use gemm
+  else {
+    sgemm(CblasRowMajor, transA, transB, M, N, K, alpha, data, lda, mdata, ldb,
+          beta, rdata, ldc);
+  }
 
   return result;
 }
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -564,7 +564,8 @@ Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
 /**
  * @note: This dot product flattens the fist 3 axis for the purpose of
  * computation. So, while performing, these matrices are behaving as 2-D
- * matrices. The dimensions are restored while returning back the tensor.
+ * matrices. The dimensions are restored while returning back the tensor
+ * in case of trans is false.
  */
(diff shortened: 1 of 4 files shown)


jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -633,9 +633,33 @@ Tensor &Tensor::dot(Tensor const &m, Tensor &result, bool trans,
   float *rdata = result.getData();
   const float alpha = 1.0f;
   const float beta = 0.0f;
-
   enum CBLAS_TRANSPOSE transA = trans ? CblasTrans : CblasNoTrans;
   enum CBLAS_TRANSPOSE transB = trans_m ? CblasTrans : CblasNoTrans;
+
+  /// shortcut calculation in case of vector
+  /// case1: (1 * K) X (K * 1)
+  if (M == 1 && N == 1) {
```
kparichay, line 641: in this scenario, trans are directly ignored? If yes, then please add that to the documentation of the function.

nntrainer/tensor/tensor.cpp
```diff
@@ -633,9 +633,33 @@ Tensor &Tensor::dot(Tensor const &m, Tensor &result, bool trans,
...
+    *rdata = sdot(K, data, 1, mdata, 1);
+    return result;
+  }
+
+  /// case2: (M * K) X (K * 1)
+  if (N == 1) {
+    sgemv(CblasRowMajor, transA, dim1, dim2, alpha, data, lda, mdata, 1, beta,
```
kparichay, line 648: similarly add that transB is ignored. If not ignored, it would generate `(M * K) X (1 * K)` which should ideally give an error. As this is just optimization and not changing functional behavior, lets add it as a documentation.

kparichay: or add a check that transB must be false (which would be the case if gemm was being used here.)

nntrainer/tensor/tensor.cpp
```diff
@@ -633,9 +633,33 @@ Tensor &Tensor::dot(Tensor const &m, Tensor &result, bool trans,
...
+
+  /// shortcut calculation in case of vector
+  /// case1: (1 * K) X (K * 1)
+  if (M == 1 && N == 1) {
+    *rdata = sdot(K, data, 1, mdata, 1);
+    return result;
+  }
+
+  /// case2: (M * K) X (K * 1)
+  if (N == 1) {
```
kparichay, line 647: how about making these all else if and lastly else with just 1 return statement?

nntrainer/tensor/tensor.cpp
```diff
@@ -633,9 +633,33 @@ Tensor &Tensor::dot(Tensor const &m, Tensor &result, bool trans,
...
+    sgemv(CblasRowMajor, transA, dim1, dim2, alpha, data, lda, mdata, 1, beta,
+          rdata, 1);
+    return result;
+  }
+
+  /// case3: (1 * K) X (K * N) = 1 * N = R
+  /// = R^T = (K * N) ^T * (1 * K)
```
kparichay, line 654: this will become `(K * N) ^T * (1 * K) ^ T` -> `N * K X K * 1` right? S
Just like above, please add check for transA or add comment that its not needed.

kparichay:
I have left some minor comments so that the behavior of dot remains same.
Overall good job in finding the underlying issue and solving it.

nntrainer/tensor/tensor.cpp
zhoonit: for a vector A (1 * K), A^T == A in terms of memory layout. The case is considered in the test :)

nntrainer/tensor/tensor.cpp
zhoonit: `(1 * K) == (K * 1)` in terms of memory layout. This is considered and tested in the unittest. Thanks :)

nntrainer/tensor/tensor.cpp
zhoonit: I'll add additional comment that (1 * K) == (K * 1) 👍

nntrainer/tensor/tensor.cpp
zhoonit: Done

nntrainer/tensor/tensor.cpp
zhoonit: Oh, If you mean `(M * K) X (1 * K)` giving an error, it is handled before getting in to this line.

nntrainer/tensor/tensor.cpp
kparichay: Right, `1*k` and `k*1` are same from the perspective from the memory, but in tensor_dim, they are not. Hence, it must be cleared.

zhoonit: Ahh, sorry for the misunderstanding, trans are not ignored as trans are resolved to get appropriate (N, M, K) before entering this line. 
Please note that N, M, K denotes a placeholder for (N * K) x (K * M). So case of (1 * K) X (1 * K) is not possible as it should be (1 * K) X (1 * M).

I added some more notes here to clarify.

kparichay: thanks

kparichay (approved):
LGTM

PR #804, opened 2020-12-14
[Test] Add macro to check if backbone is enabled @open sesame 12/16 14:30

### LINKED ISSUE #803: CCAPI unittest fail when backbone is not enabled

zhoonit:
When backbone is not enabled, test fails because backbone is not enabled
This patch adds a define in the test so that test can pass when backbone
is not enabled

resolves #803


kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #806, opened 2020-12-16
[Manager] Add shared memory allocation strategy @open sesame 12/16 17:54

zhoonit:
- ~[Graph] Move assignMem to graph~ Removed by request
```
neuralNetworks::assignMem is graph actually graph related.
This patch moves assignMem to graph.

**Self evaluation:**

```
- [TensorDim] Add initializer list ctor
```
This patch adds a tensordim
initializer list ctor to easily pass as a functional argument

**Self evaluation:**

```
- [Tensor] Add Tensor Wrap method
```
Add Tensor some factory methods
1. burrows external memory and use from
2. create from shared pointer without copy

To restrict unwanted use, those methods are static methods
called `Tensor::Wrap`

**Self evaluation:**

```
- [Manager] Add memory allocator
```
This patch adds memory allocator(mmap and corresponding shared memory
object) to manager.h

**Self evaluation:**

[Manager] Add shared memory allocation strategy

Currently, memory allocation is done in tensor but memories are
scattered away and not allocated in the shared memory region while some
hardware accel requires us to pass a shared memory object.

This patch add a strategy that manager owns a memory chunk and lend
memory to a tensor. In this manner, tensor does not own the memory so it
shouldn't be deallocated when Tensor destructs. This will enable to pass
a memory block to hardware accel (eg.NNAPI)

**Self evaluation:**

- Verified that this patch has no or little impact on PC / Android
with single FC inference pass

```
- [Android] Manage ndk to deal with changes
```
1. Upgrade ndk version to 29
2. Add dependent library
3. Fix syntax for Application.mk

```

- [Manager/Fix] Disallow copy ctor of manager
```
Since manager is holding a memory, it shouldn't be copied as ownership
becoms not clear. This patch delets copy ctor / assignment ops. While
chainging signature for members and functions that uses manager

**Self evaluation:**

```
**Added Commit**
- [Manager] Add MMaped memory
```   
    There was a requirement to separate weight memory region and grad memory
    region.
    To easily separate those two, this patch introduces new abstraction:
    `MMapedMemory` while separating weight and grad mmap
    
```

### DIFF
--- a/nntrainer/models/neuralnet.h
+++ b/nntrainer/models/neuralnet.h
@@ -95,6 +95,7 @@ public:
     loss_type(LossType::LOSS_UNKNOWN),
     weight_initializer(WeightInitializer::WEIGHT_UNKNOWN),
     net_type(NetType::UNKNOWN),
+    manager(std::make_shared<Manager>()),
     data_buffer(nullptr),
     continue_train(false),
     initialized(false),
@@ -369,7 +370,7 @@ public:
    * @note This optimization has no performance overhead.
    */
   void setGradientMemoryOptimization(bool opt) {
-    manager.setGradientMemoryOptimization(opt);
+    manager->setGradientMemoryOptimization(opt);
   }
 
   /**
@@ -483,7 +484,7 @@ private:
 
   GraphType layers; /**< vector for store layer pointers */
 
-  Manager manager; /**< nntrainer manager */
+  std::shared_ptr<Manager> manager; /**< nntrainer manager */
 
   std::shared_ptr<DataBuffer> data_buffer; /**< Data Buffer to get Input */
 
@@ -572,6 +573,7 @@ private:
     swap(lhs.layer_names, rhs.layer_names);
     swap(lhs.def_name_count, rhs.def_name_count);
     swap(lhs.loadedFromConfig, rhs.loadedFromConfig);
+    swap(lhs.manager, rhs.manager);
   }
 
   /**
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -33,18 +33,100 @@
 #include <nntrainer_log.h>
 
 namespace nntrainer {
+MMapedMemory::MMapedMemory(size_t size, bool allocate_fd) :
+  fd(-1),
+  buf(nullptr),
+  buf_size(0) {
+
+#ifndef __ANDROID__
+  if (allocate_fd) {
+    /// @todo create a file in tmpfs and bind to memfs
+    /// memfd_create is not available for number of platforms so this is
+    /// commented
+    // auto fd_ = memfd_create("", 0);
+    // if (fd_ < 0) {
+    //   throw std::runtime_error("[Manager] creating mem fd failed");
+    // }
+    // if (ftruncate(fd_, size) < 0) {
+    //   throw std::runtime_error("[Manager] truncating fd failed");
+    // }
+    ml_logi("[MMapedMemory] fd creation is not supported in this platform");
+    allocate_fd = false;
+  }
+#endif
+  int fd_ = -1;
+  void *buf_ = nullptr;
+
+  if (allocate_fd) {
+#ifdef __ANDROID__
+    /// unfortunately, memfd_create is not supported before android level 30
+    fd_ = ASharedMemory_create("", size);
+    if (fd_ < 0) {
+      throw std::runtime_error("[MMapedMemory] creating mem fd failed");
+    }
+
+    if (ASharedMemory_setProt(fd_, PROT_READ | PROT_WRITE) < 0) {
+      // unlink / close the given fd here
+      close(fd_);
+      throw std::runtime_error("[MMapedMemory] Setting prot failed");
+    }
+
+    buf_ = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd_, 0);
+#endif
+  } else {
+    buf_ = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS,
+                fd_, 0);
+  }
+
+  if (buf_ == MAP_FAILED) {
+    if (fd_ != -1) {
+      // unlink / close the given fd here
+      close(fd_);
+    }
+
+    throw std::runtime_error("[MMapedMemory] mmap failed");
+  }
+
+  fd = fd_;
+  buf = buf_;
+  buf_size = size;
+
+  ml_logd("[MMapedMemory] memory acquired size: %zu, fd: %d, addr: %p",
+          buf_size, fd, buf);
+}
+
+MMapedMemory::~MMapedMemory() {
+#ifdef DEBUG
+  assert(buf_size > 0 && fd > 0);
+#endif
+
+  if (fd != -1) {
+    if (close(fd) < 0) {
+      ml_logw("[MMapedMemory] closing fd failed on destruction please check");
+    }
+  }
+
+  if (buf != nullptr) {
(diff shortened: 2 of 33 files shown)


nntrainer/tensor/manager.cpp
```diff
@@ -35,37 +61,148 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
...
+    return Tensor();
+  };
+
+  AllocFunc allocate_weight = allocate_none;
+  AllocFunc allocate_grad = allocate_none;
+
+  if (use_shared_memory) {
+    size_t grad_size =
+      enable_gradient_memory_opt ? max_grad_size : total_grad_size;
+    size_t total_size = total_weight_size + grad_size;
```
kparichay, line 101: how do you handle the scenario when enable_gradient_memory_opt is false and a layer's weights are set to trainable which earlier were not trainable and vice versa.

nntrainer/tensor/manager.cpp
```diff
@@ -35,37 +61,148 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
...
+
+    size_t weight_bytes_size =
+      (total_weight_size + total_grad_size) * sizeof(float);
+    initializeSharedMemory(weight_bytes_size);
+
+    allocate_grad = allocate_weight = [&](const TensorDim &dim, size_t offset) {
+      return Tensor::Wrap(buf, dim, offset);
+    };
+
+  } else {
```
kparichay, line 116: the if case allocates weight and grad both
however, the else case only allocates grad?

nntrainer/tensor/manager.cpp
```diff
@@ -2,27 +2,53 @@
...
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
  *
  */
 
+#ifdef __ANDROID__
+#include <android/sharedmem.h>
+#endif
+
+#include <cassert>
```
kparichay, line 20: add DEBUG around it.

Applications/Custom/LayerClient/jni/pow.cpp
```diff
@@ -89,7 +89,7 @@ int PowLayer::setProperty(std::vector<std::string> values) {
   return nntrainer::Layer::setProperty(unhandled_values);
 }
 
-int PowLayer::initialize(nntrainer::Manager &manager) {
+int PowLayer::initialize(std::shared_ptr<nntrainer::Manager> manager) {
```
kparichay, line 92: I wonder where the copy constructor is called. 
In my understanding, passing the reference does not invoke copy constructor.
Also, ownership is with the neuralnetwork with `Manager manager`, and passed reference will be invalidated after model class is done with. But with `shared_ptr<Manager>`, layer can create a copy and keep it alive.
I dont understand why have a shared_ptr here (and it overhead), can you please elaborate on this.

nntrainer/tensor/manager.cpp
zhoonit: We made a offline discussion to manage two mmap (one for gradient, and another one for weights in case of reallocation happens)

nntrainer/tensor/manager.cpp
zhoonit: This is defaulted to allocate_none

nntrainer/tensor/manager.cpp
zhoonit: Will do

Applications/Custom/LayerClient/jni/pow.cpp
zhoonit: We made an offline discussion to check if this can be reverted.

nntrainer/tensor/manager.cpp
zhoonit: Done

nntrainer/tensor/manager.cpp
zhoonit: Done

Applications/Custom/LayerClient/jni/pow.cpp
zhoonit: Reverted

nntrainer/tensor/manager.cpp
zhoonit: Done with a new commit. Please take another look

zhoonit:
Added Commit

[Manager] Add MMaped memory
```
There was a requirement to separate weight memory region and grad memory
region.
To easily separate those two, this patch introduces no abstraction:
`MMapedMemory` while separating weight and grad mmap
    
```

nntrainer/models/neuralnet.h
```diff
@@ -479,6 +480,7 @@ class NeuralNetwork : public ml::train::Model {
     swap(lhs.layer_names, rhs.layer_names);
     swap(lhs.def_name_count, rhs.def_name_count);
     swap(lhs.loadedFromConfig, rhs.loadedFromConfig);
+    swap(lhs.manager, rhs.manager);
```
kparichay, line 483: is this intended? this does not copy but rather binds the memory of the two models together.

zhoonit: Yes, this will allow sharing the memory between model. I think manager should be unique_ptr, naturally model(or more specifically, graph) shouldn't be copy constructible but movable.

This will require bigger change so I decided to go with shared_ptr for now.

jijoongmoon: I think, in order to share the memory manager between models,  there needs to be more careful changes. As far as I understand, it is not considered for now. currently, one model has one manager. and I'm not sure some reason to share the memory manager though.

nntrainer/tensor/manager.cpp
```diff
@@ -2,27 +2,138 @@
...
+    // }
+    // if (ftruncate(fd_, size) < 0) {
+    //   throw std::runtime_error("[Manager] truncating fd failed");
+    // }
+    ml_logi("[MMapedMemory] fd creation is not supported in this platform");
+    allocate_fd = false;
+  }
+#endif
+  int fd_ = -1;
+  void *buf_ = nullptr;
```
jijoongmoon, line 58: looks duplicate with line 37

nntrainer/tensor/tensor.cpp
```diff
@@ -92,6 +92,35 @@ Tensor::Tensor(const TensorDim &d, const float *buf) : Tensor() {
   }
 }
 
+Tensor Tensor::Wrap(float *buf, const TensorDim &d, int offset) {
+  if (d.getDataLen() == 0 || buf == nullptr) {
```
jijoongmoon, line 96: It more like map rather than wrap. I guess.

nntrainer/models/neuralnet.h
zhoonit: Yes. This will be a unique ptr and the Model won't be copy constructible (but movable), this requires some bigger change so I'll leave it as an issue and do this in later PR.

nntrainer/tensor/manager.cpp
zhoonit: This is intended for the exception safety guarantee.
`fd` and `buf` aren't assigned if the construction fails.

nntrainer/tensor/tensor.cpp
zhoonit: I feel like map implies one to one mapping. wrap implies it wraps a memory with a tensor to use tensor methods. 

Thanks for the suggestion :) @kparichay Could you give an opinion as well?

zhoonit: Changed to Tensor::Map :)

jijoongmoon (approved):
LGTM

kparichay (approved):
LGTM

PR #810, opened 2020-12-17
[layer] Support in-place batch normalization

kparichay:
Support in-place batch normalization where the batch normalization
input/output is not stored and is over-written by the next layer.

This patch removes the input/output memory requirement when using
batch normalization layer.

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -165,6 +165,35 @@ int NeuralNetwork::compile() {
   return status;
 }
 
+void NeuralNetwork::inPlaceBatchNormOptimization() {
+  auto &sorted = model_graph.getSorted();
+
+  for (unsigned int idx = 1; idx < sorted.size() - 1; ++idx) {
+    auto &l = sorted[idx].layer;
+    if (l->getType() == BatchNormalizationLayer::type) {
+      /** @note assumes BatchNormalizationLayer is only for single in/out tensor
+       */
+      if (l->input_layers.size() != 1)
+        throw std::runtime_error("Internal error in the formed graph");
+
+      auto &prev_layer = model_graph.getLayerNode(l->input_layers[0]).layer;
+
+      unsigned int loc;
+      auto layer_name = l->getName();
+      for (loc = 0; loc < prev_layer->output_layers.size(); ++loc)
+        if (prev_layer->output_layers[loc] == layer_name)
+          break;
+
+      if (loc == prev_layer->output_layers.size())
+        throw std::runtime_error("Internal error in the formed graph.");
+
+      /** Share tensor with next layer */
+      prev_layer->net_hidden[loc] = l->net_hidden[0];
+      l->net_input[0] = l->net_hidden[0];
+    }
+  }
+}
+
 int NeuralNetwork::initialize() {
   int status = ML_ERROR_NONE;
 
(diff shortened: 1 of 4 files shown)


kparichay:
The reason this works is that batch-normalization stores a processed version of input and does not need storing the input.

nntrainer/models/neuralnet.cpp
```diff
@@ -164,6 +164,29 @@ int NeuralNetwork::compile() {
...
+      /** @note assumes BatchNormalizationLayer is only for single in/out tensor
+       */
+      auto &prev_layer = model_graph.getLayerNode(l->input_layers[0]).layer;
+
+      unsigned int loc;
+      auto layer_name = l->getName();
+      for (loc = 0; loc < prev_layer->output_layers.size(); ++loc)
+        if (prev_layer->output_layers[loc] == layer_name)
+          break;
+
```
zhoonit, line 182: How about adding a check `if loc == prev_layer.output_layers.size()`
Otherwise net_hidden[loc] can overflow?

nntrainer/models/neuralnet.cpp
```diff
@@ -164,6 +164,29 @@ int NeuralNetwork::compile() {
...
 
+void NeuralNetwork::inPlaceBatchNormOptimization() {
+  auto &sorted = model_graph.getSorted();
+
+  for (unsigned int idx = 1; idx < sorted.size() - 1; ++idx) {
+    auto &l = sorted[idx].layer;
+    if (l->getType() == BatchNormalizationLayer::type) {
+      /** @note assumes BatchNormalizationLayer is only for single in/out tensor
+       */
+      auto &prev_layer = model_graph.getLayerNode(l->input_layers[0]).layer;
```
zhoonit, line 175: Assure `l->input_layers.size() == 1`?

nntrainer/models/neuralnet.h
```diff
@@ -100,7 +100,8 @@ class NeuralNetwork : public ml::train::Model {
     compiled(false),
     def_name_count(0),
     loadedFromConfig(false),
-    app_context(app_context_) {}
+    app_context(app_context_),
+    in_place_bn_layer_optimization(true) {}
```
zhoonit, line 104: How about we add a default argument for the option?

zhoonit (commented):
as far as I understood, this optimization is implementation specific. If it is, could you add a note about how this works? My concern is that this might not work for other accelerated batch norm.

zhoonit (approved):
LGTM with minor comments

nntrainer/models/neuralnet.cpp
kparichay: added a throw.

nntrainer/models/neuralnet.cpp
kparichay: added a throw.

nntrainer/models/neuralnet.h
kparichay: added.

kparichay:
> as far as I understood, this optimization is implementation specific. If it is, could you add a note about how this works? My concern is that this might not work for other accelerated batch norm.

Added a note in the optimization function declaration.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #827, opened 2020-12-23
[optimization] In-place memory optimization for activation layer

### LINKED ISSUE #826: in-place memory optimization layers
Provide in-place memory optimization for layers.

kparichay:
Commit 1: [inputlayer] Input layer must be non-trainable 
Input layer must always be non-trainable as it does not support backwarding operation

Commit 2: [manager] Manager tracks input/output memory
Manager tracks input/output memory and allocates it
based on if the execution is training or inference

Commit 3: [layer] Use gradient instead of variable for derivative
Use gradient instead of variable for derivative
Manager internally sets gradient memory same as variable for the optimization
but hides this kind of optimizations from the layer

Commit 4: [activation] Making activation in-place
Added activation layer to be in-place.
Each layer now allocates memory for its output than for its input.

For activation layer, if its memory is optimized, then the memory
for the layer behind activation layer is not allocated.
And the memory for the derivative of the activation layer is shared
among all such layers.

Commit 5: [model/test] Duplicate models test for optimization
Run models test twice, once with all the optimizations enabled
and then once with all the optimizations disabled.

This ensures that both modes work properly.

See Also #774
Resolves #826

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/activation_layer.cpp
+++ b/nntrainer/layers/activation_layer.cpp
@@ -50,24 +50,19 @@ void ActivationLayer::forwarding(sharedConstTensors in) {
   Tensor &hidden_ = net_hidden[0]->getVariableRef();
   /// @note @a _act_fn is expected to work out of place and not modify @a input
   _act_fn(net_input[0]->getVariableRef(), hidden_);
-  if (activation_type == ActivationType::ACT_SOFTMAX)
-    backup_hidden = hidden_.clone();
 }
 
 void ActivationLayer::calcDerivative(sharedConstTensors derivative) {
   Tensor &deriv = net_hidden[0]->getGradientRef();
   Tensor &ret = net_input[0]->getGradientRef();
+  Tensor &in = net_hidden[0]->getVariableRef();
 
-  if (activation_type == ActivationType::ACT_SOFTMAX) {
-    ret = _act_prime_fn(backup_hidden, ret, deriv);
-  } else {
-    ret = _act_prime_fn(net_input[0]->getVariableRef(), ret, deriv);
-  }
+  ret = _act_prime_fn(in, ret, deriv);
 }
 
 int ActivationLayer::setActivation(
   std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
-  std::function<Tensor &(Tensor const &, Tensor &, Tensor const &)> const
+  std::function<Tensor &(Tensor &, Tensor &, Tensor const &)> const
     &activation_prime_fn) {
   _act_fn = activation_fn;
   _act_prime_fn = activation_prime_fn;
@@ -77,13 +72,12 @@ int ActivationLayer::setActivation(
 
 int ActivationLayer::setActivation(
   std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
-  std::function<Tensor &(Tensor const &, Tensor &)> const
-    &activation_prime_fn) {
+  std::function<Tensor &(Tensor &, Tensor &)> const &activation_prime_fn) {
   _act_fn = activation_fn;
-  _act_prime_fn = [activation_prime_fn](Tensor const &x, Tensor &ret_derivative,
+  _act_prime_fn = [activation_prime_fn](Tensor &x, Tensor &ret_derivative,
                                         Tensor const &derivative) -> Tensor & {
-    ret_derivative = activation_prime_fn(x, ret_derivative);
-    ret_derivative.multiply_i(derivative);
+    x = activation_prime_fn(x, x);
+    ret_derivative = derivative.multiply(x, ret_derivative);
 
     return ret_derivative;
   };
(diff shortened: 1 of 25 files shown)


kparichay:
With this patch, the peak memory overhead for VGG is reduced down from 512MB to roughly 395MB.

nntrainer/layers/activation_layer.cpp
```diff
@@ -50,24 +50,19 @@ void ActivationLayer::forwarding(sharedConstTensors in) {
   Tensor &hidden_ = net_hidden[0]->getVariableRef();
   /// @note @a _act_fn is expected to work out of place and not modify @a input
   _act_fn(net_input[0]->getVariableRef(), hidden_);
```
zhoonit, line 52: If activation function is not inplacable, this might run into a problem.

I think we should have:
1. type traits for inplacable operations
2. Deal each activation layer saperately.

For now, it looks goot to go for me.

nntrainer/layers/activation_layer.cpp
```diff
@@ -228,15 +222,15 @@ Tensor &ActivationLayer::softmaxPrime(Tensor const &x, Tensor &output,
 float ActivationLayer::sigmoid(float x) { return 1.0f / (1.0f + exp_util(-x)); }
 
 float ActivationLayer::sigmoidPrime(float x) {
-  float sprime = sigmoid(x);
-  return sprime * (1.0f - sprime);
+  // float sprime = sigmoid(x);
```
zhoonit, line 225: Unintended comment?

nntrainer/layers/activation_layer.cpp
```diff
@@ -228,15 +222,15 @@ Tensor &ActivationLayer::softmaxPrime(Tensor const &x, Tensor &output,
...
+  return x * (1.0f - x);
 }
 
 float ActivationLayer::tanhFloat(float x) { return (float)tanh(x); }
 
 float ActivationLayer::tanhPrime(float x) {
-  float th = (float)tanh(x);
-  return 1.0f - th * th;
+  // float th = (float)tanh(x);
```
zhoonit, line 232: Same

nntrainer/layers/activation_layer.cpp
```diff
@@ -77,13 +72,12 @@ int ActivationLayer::setActivation(
...
-  std::function<Tensor &(Tensor const &, Tensor &)> const
-    &activation_prime_fn) {
+  std::function<Tensor &(Tensor &, Tensor &)> const &activation_prime_fn) {
   _act_fn = activation_fn;
-  _act_prime_fn = [activation_prime_fn](Tensor const &x, Tensor &ret_derivative,
+  _act_prime_fn = [activation_prime_fn](Tensor &x, Tensor &ret_derivative,
                                         Tensor const &derivative) -> Tensor & {
-    ret_derivative = activation_prime_fn(x, ret_derivative);
-    ret_derivative.multiply_i(derivative);
+    x = activation_prime_fn(x, x);
```
zhoonit, line 79: Q. is this one okay when the optimization is off?

nntrainer/layers/activation_layer.cpp
kparichay: As of now, softmax activation layer does not work in-place.
For these kind of layers, the manager which is performing the in-place optimization is supposed to (and is currently also) ensure that those activation layers do not operate in-place as that will produce wrong result.

nntrainer/layers/activation_layer.cpp
kparichay: Check the last commit which ensures that everything is tested with optimization on as well as off.
The reason this works is that all the layers work with their inputs in the backward operation.
This `x` is actually the input of the next layer which has already performed its backward operation (note that backward start from last layer to the first layer). So, overwriting this `x` does not affect other layers.

nntrainer/layers/activation_layer.cpp
kparichay: will remove

nntrainer/layers/activation_layer.cpp
kparichay: same :)

nntrainer/layers/activation_layer.cpp
kparichay: check https://github.com/nnstreamer/nntrainer/pull/827/files#diff-8136da3afc181f3163a8f8a15d86d132abee8c07de2c9b49bc2cf0759e380a69R544
in-place optimizer function ensures that in-place optimization is only performed on layers which support in-place operations.

Later I will add a function for `layer` that would return if the `layer` can operate in-place. If yes, the in-place optimization function will make it in-place, else skip it.

zhoonit: Sounds good

nntrainer/layers/activation_layer.cpp
zhoonit: Thanks for the confirmation :)

zhoonit: When optimization is off, this shouldn't replace the backward result isn't it?

kparichay: When the optimization is off, still behaves the same as of now.
Layer's behavior must remain the same independent of the optimization.
The layer shouldn't know if the optimization is being done or not.

The activation layer now always works in-place with this. It is up to the model to use that information for optimization.

zhoonit: Oh I C. Thanks for the explanation :)

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #840, opened 2020-12-29
[manager] Optimize input/output memory for inference

### LINKED ISSUE #790: Inference memory reuse
In the inference mode, the inputs/outputs are not required to be stored, and that memory can be reused.
This is the same optimization that #766 applies for the gradients.

kparichay:
Optimize input/output memory for inference by using a shared buffer
where the max([sum(input_l, output_l)) for l from all layers]) memory
is allocated for inference.

Baseline working unittest added with models unittest which ensures
that inference works with and without optimizations without any
failures. Value verification tests is done by nnstreamer subplugin of
nntrainer.

Resolves #790

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -217,10 +217,6 @@ int NeuralNetwork::initialize() {
 
         l.setInputDimension(in_layer.getOutputDimension()[location], i);
       }
-    } else {
-      auto in_out = manager->TrackLayerInOuts(l.getType(), l.getName(),
-                                              l.getInputDimension());
-      l.setInputBuffers(in_out);
     }
 
     /**
@@ -233,8 +229,8 @@ int NeuralNetwork::initialize() {
     REGISTER_EVENT(l.getName(), lnode.event_key)
     opt->addOptimizerVariable(l.getWeightsRef());
 
-    auto in_out = manager->TrackLayerInOuts(l.getType(), l.getName(),
-                                            l.getOutputDimension());
+    auto &in_out = manager->trackLayerOutputs(
+      l.getType(), l.getName(), l.getOutputDimension(), l.getInputDimension());
     l.setOutputBuffers(in_out);
 
     /** Connect the output of the previous layers with the input of the current
@@ -254,34 +250,13 @@ int NeuralNetwork::initialize() {
         l.net_input[i] = model_graph.getLayerNode(l.input_layers[i])
                            .layer->net_hidden[location];
       }
-      // <<<<<<< d5b08b6c082d5ab907768f9293434db7f82a8cd2
-      //     } else {
-      //       manager->TrackLayerInOuts(l.getName(), l.getInputDimension(),
-      //                                 l.getTrainable());
-      //       l.setInputBuffers(manager->getInputsLayer(-1));
-      //     }
-      //
-      //     status = l.initialize(*manager);
-      //     NN_RETURN_STATUS();
-      //
-      //     REGISTER_EVENT(l.getName(), lnode.event_key)
-      //     opt->addOptimizerVariable(l.getWeightsRef());
-      //   }
-      //
-      //   auto &last_layer = model_graph.Sorted.back().layer;
-      //   manager->TrackLayerInOuts(last_layer->getName(),
-      //                             last_layer->getOutputDimension(),
-      //                             last_layer->getTrainable());
-      //   auto in_out = manager->getInputsLayer(-1);
-      //
-      //   for (unsigned int i = 0; i < last_layer->num_outputs; ++i) {
-      //     last_layer->net_hidden[i] = in_out[i];
-      //   }
-      //
-      // =======
+    } else {
+      auto &in_out = manager->trackLayerInputs(l.getType(), l.getName(),
+                                               l.getInputDimension(),
+                                               l.getOutputDimension());
+      l.setInputBuffers(in_out);
     }
   }
-  // >>>>>>> [activation] Making activation in-place
   setBatchSize(batch_size);
 
   if (in_place_optimization) {
@@ -310,7 +285,6 @@ NeuralNetwork::~NeuralNetwork() {
  */
 sharedConstTensors NeuralNetwork::forwarding() {
   return model_graph.forwarding();
-  // model_graph.Sorted.back().layer->forwarding();
 }
 
 /**
@@ -326,7 +300,7 @@ sharedConstTensors NeuralNetwork::forwarding(sharedConstTensors input,
     model_graph.getSortedLayerNode(model_graph.getSorted().size() - 1).layer;
 
   if (label.empty())
-    last_layer->net_hidden.clear();
+    last_layer->net_hidden[0]->getGradientRef() = Tensor();
   else
     last_layer->net_hidden[0]->getGradientRef() = *label[0].get();
 
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -30,6 +30,7 @@
 #include <vector>
 
 #include <activation_layer.h>
+#include <flatten_layer.h>
 #include <manager.h>
 #include <nntrainer_log.h>
 
@@ -122,14 +123,18 @@ MMapedMemory::~MMapedMemory() {
 
 Manager::Manager(bool enable_gradient_memory_opt_,
                  bool enable_derivative_memory_opt_,
-                 bool enable_activation_memory_opt_, bool use_shared_memory_) :
+                 bool enable_activation_memory_opt_,
+                 bool enable_inference_inout_memory_opt_,
+                 bool use_shared_memory_) :
   total_weight_size(0),
   total_grad_size(0),
   max_grad_size(0),
   max_derivative_size(0),
+  max_shared_inout(0),
   enable_gradient_memory_opt(enable_gradient_memory_opt_),
   enable_derivative_memory_opt(enable_derivative_memory_opt_),
   enable_activation_memory_opt(enable_activation_memory_opt_),
+  enable_inference_inout_memory_opt(enable_inference_inout_memory_opt_),
   use_shared_memory(use_shared_memory_) {}
 
 Manager::~Manager() {}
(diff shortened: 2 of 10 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -254,34 +250,13 @@ int NeuralNetwork::initialize() {
...
-      //                             last_layer->getTrainable());
-      //   auto in_out = manager->getInputsLayer(-1);
-      //
-      //   for (unsigned int i = 0; i < last_layer->num_outputs; ++i) {
-      //     last_layer->net_hidden[i] = in_out[i];
-      //   }
-      //
-      // =======
+    } else {
+      auto in_out = manager->trackLayerInputs(l.getType(), l.getName(),
```
zhoonit, line 254: auto &in_out ?

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -360,6 +361,12 @@ void GraphWatcher::compareFor(const std::string &reference,
         it->backward(iteration, true, !optimize);
     }
   }
+
+  /**
+   * This inference is to ensure that inference runs with/without optimizations
+   * for various kinds of models
+   */
+  nn.inference(input);
```
zhoonit, line 369: Q. is this EXPECT_NO_THROW?

nntrainer/tensor/manager.cpp
```diff
@@ -245,40 +250,87 @@ void Manager::initialize() {
...
 
+std::vector<std::shared_ptr<Var_Grad>> &
+Manager::trackLayerOutputs(const std::string &layer_type,
+                           const std::string &layer_name,
+                           const std::vector<TensorDim> &output_dim,
+                           const std::vector<TensorDim> &input_dim) {
+  if (enable_inference_inout_memory_opt && input_dim.empty()) {
+    throw std::invalid_argument(
+      "Input dimensions are required with inference memory opt.");
+  } else {
```
zhoonit, line 288: Could you double check this if else statement?

what if `enable_inference_inout_memory_opt` is true && `!input_dim.empty()` vs `enable_inference_inout_memory_opt` is false

nntrainer/tensor/manager.cpp
```diff
@@ -245,40 +250,87 @@ void Manager::initialize() {
...
+
+std::vector<std::shared_ptr<Var_Grad>> &
+Manager::trackLayerInputs(const std::string &layer_type,
+                          const std::string &layer_name,
+                          const std::vector<TensorDim> &input_dim,
+                          const std::vector<TensorDim> &output_dim) {
+  if (enable_inference_inout_memory_opt && output_dim.empty()) {
+    throw std::invalid_argument(
+      "Output dimensions are required with inference memory opt.");
+  } else {
```
zhoonit, line 311: same

nntrainer/tensor/manager.cpp
```diff
@@ -245,40 +250,87 @@ void Manager::initialize() {
...
+    for (auto const &dim : input_dim)
+      shared_inout += dim.getDataLen();
+
+    max_shared_inout = std::max(max_shared_inout, shared_inout);
+  }
+
+  return trackLayerInOuts(layer_type, layer_name + ":InOut", input_dim);
+}
+
 void Manager::untrackLayerInOuts(const std::string layer_name) {
```
zhoonit, line 326: out of topic though, const std::string & ?

nntrainer/tensor/manager.cpp
```diff
@@ -245,40 +250,87 @@ void Manager::initialize() {
...
   auto var_name = layer_name + ":InOut" + std::to_string(0);
 
   for (unsigned int cnt = 0; cnt < in_outs.size(); cnt++) {
     if (!in_outs[cnt].empty() && in_outs[cnt][0]->getName() == var_name) {
       in_outs.erase(in_outs.begin() + cnt);
       is_act_type.erase(is_act_type.begin() + cnt);
-      break;
+      is_flat_type.erase(is_flat_type.begin() + cnt);
```
zhoonit, line 333: Is removing break intended here?

nntrainer/tensor/manager.cpp
```diff
@@ -287,24 +339,57 @@ void Manager::untrackLayerInOuts(const std::string layer_name) {
...
   for (unsigned int idx = 0; idx < in_outs.size(); idx++) {
     auto &l_io = in_outs[idx];
     size_t offset = 0;
     bool is_last_layer = idx == in_outs.size() - 1;
     for (auto &io : l_io) {
-      if (enable_derivative_memory_opt && !is_last_layer) {
+      if (!trainable) {
+        Tensor shared_inout_cur = Tensor();
+        // For flatten layer, do not assign new memory
+        if (count > 0 && is_flat_type[count])
```
zhoonit, line 368: code clean: can `count` be replaced to idx??

nntrainer/tensor/manager.cpp
```diff
@@ -287,24 +339,57 @@ void Manager::untrackLayerInOuts(const std::string layer_name) {
...
+          use_first_last = 1 - use_first_last;
```
zhoonit, line 369: I think this should go outside of this loop, because the inner loop is for multiple inputs, outputs

nntrainer/tensor/manager.cpp
```diff
@@ -287,24 +339,57 @@ void Manager::untrackLayerInOuts(const std::string layer_name) {
...
+  // Allocate shared input/output memory for inference
+  // @note Memory for label is not allocated here as inference doesnt has label
+  Tensor shared_inout;
+  if (!trainable && enable_inference_inout_memory_opt)
+    shared_inout = Tensor(max_shared_inout);
+
+  // TODO: do not allocate memory for flatten layer
+
   size_t count = 0;
+  bool use_first_last = 0;
```
zhoonit, line 359: Could you add comment here? I understood from the offline explanation but we might forget later :p

nntrainer/tensor/manager.cpp
```diff
@@ -287,24 +339,57 @@ void Manager::untrackLayerInOuts(const std::string layer_name) {
...
+      if (!trainable) {
+        Tensor shared_inout_cur = Tensor();
+        // For flatten layer, do not assign new memory
+        if (count > 0 && is_flat_type[count])
+          use_first_last = 1 - use_first_last;
+        if (enable_inference_inout_memory_opt) {
+          if (use_first_last) {
+            // Create tensor with from the front of shared tensor
+            shared_inout_cur =
+              shared_inout.getSharedDataTensor(io->getDim(), 0);
```
zhoonit, line 374: Also offset need to be start from the previous input

nntrainer/tensor/manager.cpp
```diff
@@ -287,24 +339,57 @@ void Manager::untrackLayerInOuts(const std::string layer_name) {
...
+          } else {
+            // Create tensor with from the back of shared tensor
+            shared_inout_cur = shared_inout.getSharedDataTensor(
+              io->getDim(), max_shared_inout - io->getDim().getDataLen());
```
zhoonit, line 378: same

nntrainer/models/neuralnet.cpp
kparichay: done.

test/unittest/unittest_nntrainer_models.cpp
kparichay: added.

nntrainer/tensor/manager.cpp
kparichay: `if enable_inference_inout_memory_opt is true && !input_dim.empty()` -> then use input_dim to estimate the required memory, and then track layer inouts.
`if enable_inference_inout_memory_opt is false` -> added an extra check to estimate required memory when enable_inference_inout_memory_opt is true.

nntrainer/tensor/manager.cpp
kparichay: same

nntrainer/tensor/manager.cpp
kparichay: Yes, as both input and output currently use the same name when added tracked input/output.
So, it is possible to have two memory allocations with same name, and hence break was removed.

Updated names to be different and added break.

nntrainer/tensor/manager.cpp
kparichay: done.

nntrainer/tensor/manager.cpp
kparichay: done.

nntrainer/tensor/manager.cpp
kparichay: good catch - done.

nntrainer/tensor/manager.cpp
kparichay: done.

nntrainer/tensor/manager.cpp
kparichay: done.

nntrainer/tensor/manager.cpp
kparichay: same

jijoongmoon (approved):
LGTM

nntrainer/tensor/manager.cpp
```diff
@@ -122,14 +123,18 @@ MMapedMemory::~MMapedMemory() {
...
+                 bool use_shared_memory_) :
   total_weight_size(0),
   total_grad_size(0),
   max_grad_size(0),
   max_derivative_size(0),
+  max_shared_inout(0),
   enable_gradient_memory_opt(enable_gradient_memory_opt_),
   enable_derivative_memory_opt(enable_derivative_memory_opt_),
   enable_activation_memory_opt(enable_activation_memory_opt_),
+  enable_inference_inout_memory_opt(true),
```
zhoonit, line 137:
```suggestion
  enable_inference_inout_memory_opt(enable_inference_inout_memory_opt_),
```

nntrainer/tensor/manager.cpp
zhoonit: awesome

zhoonit (approved):
LGTM except for one small issue :)

nntrainer/tensor/manager.cpp
kparichay: done

PR #841, opened 2020-12-31
[transfer learning] Enable test on ubuntu

kparichay:
Enable testing of the trained model on ubuntu

### DIFF
--- a/Applications/TransferLearning/Draw_Classification/jni/main.cpp
+++ b/Applications/TransferLearning/Draw_Classification/jni/main.cpp
@@ -25,7 +25,7 @@
  *
  */
 
-#if defined(__TIZEN__) && defined(ENABLE_TEST)
+#if defined(NNSTREAMER_AVAILABLE) && defined(ENABLE_TEST)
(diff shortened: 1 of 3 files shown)


zhoonit (commented):
this will need nnstreamer as a dependency.
Could you check the meson dependency as well? 

It seems like that only when `enable-nnstreamer-tensor-filter` is on we add nnstreamer as a dependency.

kparichay:
`capi-nnstreamer` should itself add `nnstreamer` as dependency. So, we shouldn't add `nnstreamer` as dependency as `nnstreamer` might change their dependency which makes nntrainer code dependent on nnstreamer code.

jijoongmoon:
I think we need to support this example if there is no nnstreamer support.

jijoongmoon (approved):
LGTM

kparichay:
> I think we need to support this example if there is no nnstreamer support. Except that LGTM

I will add a check on nnstreamer support and only then use nnstreamer related code.

zhoonit (approved):
LGTM

PR #842, opened 2021-01-04
[mem-opt] Inference memory optimizations

kparichay:
Commit 1 : [manager] Donot allocate adam for inference 
Donot allocate adam and gradient memory for weights
when the model is being executed for inference

Commit 2 : [neuralnet] Donot set batchsize in initialize 
Donot set batch size in initialize as that allocates memory for pooling
layer. batch size is not directly set in inference/training itself.

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -257,7 +257,8 @@ int NeuralNetwork::initialize() {
       l.setInputBuffers(in_out);
     }
   }
-  setBatchSize(batch_size);
+  setBatchSize();
+
   // Allocate and initialize weights
   manager->initializeWeights();
 
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -258,12 +258,12 @@ int NeuralNetwork::initialize() {
     }
   }
   setBatchSize(batch_size);
+  // Allocate and initialize weights
+  manager->initializeWeights();
 
   if (in_place_optimization) {
     model_graph.inPlaceOptimize(*manager);
   }
-
-  manager->initialize();
 
   initialized = true;
   return status;
@@ -474,7 +474,7 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X) {
 
 int NeuralNetwork::assignMem(bool trainable) {
   // TODO: directly replace this
-  manager->initializeInOuts(trainable);
+  manager->initializeTensors(trainable);
   return ML_ERROR_NONE;
 }
 
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -174,22 +174,12 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
   max_grad_size = std::max(max_grad_size, grad_size);
 }
 
-/**
- * @brief Allocate and initialize the weight variable
- */
-void Manager::initialize() {
-  if (total_weight_size == 0) {
-    ml_logw("Nothing done on initialize because there is no weight registered");
-    return;
-  }
-  using AllocFunc = std::function<Tensor(const TensorDim &, size_t)>;
-
+Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   AllocFunc allocate_none = [](const TensorDim &dim, size_t) {
     return Tensor();
   };
 
-  AllocFunc allocate_weight = allocate_none;
-  AllocFunc allocate_grad = allocate_none;
+  AllocFunc allocate_func = allocate_none;
 
   if (use_shared_memory) {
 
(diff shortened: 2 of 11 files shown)


nntrainer/tensor/manager.cpp
```diff
@@ -287,24 +341,58 @@ void Manager::untrackLayerInOuts(const std::string layer_name) {
  * @brief Initialize the inputs/outputs for the layer
  */
 void Manager::initializeInOuts(bool trainable) {
+  initialize(trainable);
```
kparichay, line 344: this is wrong
this will reset the weights which have been loaded from the saved file.

kparichay: resolved.

nntrainer/models/neuralnet.cpp
```diff
@@ -258,13 +258,13 @@ int NeuralNetwork::initialize() {
     }
   }
   setBatchSize(batch_size);
+  // Allocate and initialize weights
+  manager->initializeWeights();
```
zhoonit, line 261: I wonder about a reason for the rearrangement.
Is it because inPlaceOptimize should have to happen after initializeWeights?

nntrainer/tensor/manager.cpp
```diff
@@ -174,22 +174,12 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
...
- * @brief Allocate and initialize the weight variable
- */
-void Manager::initialize() {
-  if (total_weight_size == 0) {
-    ml_logw("Nothing done on initialize because there is no weight registered");
-    return;
-  }
-  using AllocFunc = std::function<Tensor(const TensorDim &, size_t)>;
-
+Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
```
zhoonit, line 177: I like this :)
Could you reflect https://github.com/nnstreamer/nntrainer/pull/859/commits/e4f66430754e4b5e5366a714512ccf43fe26a398 here?
I think the commit suits here.

nntrainer/tensor/manager.cpp
```diff
@@ -207,24 +197,67 @@ void Manager::initialize() {
...
+
+  for (auto &l_w : weights) {
+    for (auto &w : l_w) {
+      Weight &weight = w.get();
+      auto dim = weight.getDim();
+      Tensor weight_prealloc = allocate_weight(dim, weight_offset);
+      Tensor grad_prealloc = Tensor();
+
+      weight_offset += dim.getDataLen();
+      weight.initialize(weight_prealloc, Tensor(), false);
```
zhoonit, line 245: how about separate initializing gradient as well (thus making `weight::initialize` to `weight::initializeWeight`), instead caller having the flexibility to call weights and gradients if needed?

test/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -40,23 +40,23 @@ static void nntrainer_capi_model_comp_metrics(ml_train_model_h model,
     &summary1);
   EXPECT_EQ(status, ML_ERROR_NONE);
 
-  EXPECT_FLOAT_EQ(std::strtof(summary1, nullptr), train_loss);
+  // EXPECT_FLOAT_EQ(std::strtof(summary1, nullptr), train_loss);
```
zhoonit, line 43: I wonder if this is intended

nntrainer/models/neuralnet.cpp
kparichay: This is because `readModels` should happen after `initializeWeights`, but before `train/inference`.
However, gradient is not requried in inference. Hence the rearrangement.

nntrainer/tensor/manager.cpp
kparichay: Added separate methods.

test/tizen_capi/unittest_tizen_capi.cpp
kparichay: removed.

nntrainer/models/neuralnet.cpp
zhoonit: Noted Thanks

nntrainer/models/neuralnet.cpp
```diff
@@ -257,7 +257,6 @@ int NeuralNetwork::initialize() {
       l.setInputBuffers(in_out);
     }
   }
-  setBatchSize(batch_size);
```
zhoonit, line 285: as far as I remember, this was added 
from https://github.com/nnstreamer/nntrainer/pull/649#discussion_r508173637

If a user is prohibited to call `forward` or `backward`.
I think this changes is fine.

If she is allowed to, either she has to explicitly call `setBatchSize` or `setBatchSize` should be done at `initailize` phase.
I think some of our application doesn't follow this and might break.

kparichay: `setBatchSize` currently creates an overhead of memory being allocated with pooling layer, with not the final batch size.
I will resolve this with an additional interface in the tensor manager for allocating extra memory for a layer.

zhoonit: Cool

zhoonit (approved):
LGTM :)

nntrainer/tensor/manager.cpp
```diff
@@ -172,24 +174,15 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
...
-  using AllocFunc = std::function<Tensor(const TensorDim &, size_t)>;
-
+Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   AllocFunc allocate_none = [](const TensorDim &dim, size_t) {
     return Tensor();
   };
 
-  AllocFunc allocate_weight = allocate_none;
-  AllocFunc allocate_grad = allocate_none;
+  AllocFunc allocate_func = allocate_none;
```
jijoongmoon, line 185: I wonder that if the user_shared_memory is false, then, the current implementation is working?

nntrainer/tensor/var_grad.cpp
```diff
@@ -24,20 +24,21 @@ Var_Grad::Var_Grad(const TensorDim &dim, bool train, const std::string &name) :
   grad = std::make_shared<Tensor>();
 }
 
-void Var_Grad::initialize(const Tensor &weights_preallocated,
-                          const Tensor &grad_preallocated, bool gtrain) {
-  if (!weights_preallocated.uninitialized()) {
-    var = std::make_shared<Tensor>(weights_preallocated);
+void Var_Grad::initializeWeight(const Tensor &preallocated, bool gtrain) {
+  if (!preallocated.uninitialized()) {
```
jijoongmoon, line 28: we do not need this gtrain here.

nntrainer/tensor/weight.cpp
```diff
@@ -24,9 +24,8 @@ Weight::Weight(const TensorDim &dim, const WeightInitializer init, bool train,
     throw std::invalid_argument("Weight initializer unknown");
 }
 
-void Weight::initialize(const Tensor &weights_preallocated,
-                        const Tensor &grad_preallocated) {
-  Var_Grad::initialize(weights_preallocated, grad_preallocated);
+void Weight::initializeWeight(const Tensor &weights_preallocated, bool gtrain) {
+  Var_Grad::initializeWeight(weights_preallocated, gtrain);
```
jijoongmoon, line 28: do we need gtrain?

nntrainer/tensor/var_grad.cpp
kparichay: `gtrain` is kept here for completeness and to maintain the same interface as `initializeGradient`

nntrainer/tensor/weight.cpp
kparichay: same as above - to maintain same interface as `initializeGradient`

nntrainer/tensor/manager.cpp
kparichay: `unittest_nntrainer_models` runs all its unit tests twice - once with all the optimizations on and once again with all the optimizations off. So, I think it works.
Currently, we don't check different combinations of optimizations which IMO would be too much.

nntrainer/tensor/var_grad.cpp
jijoongmoon: Is there any reason to maintain same interface with Gradient ?

kparichay: No advantage, other than ease of reading.

jijoongmoon: Well, I don't think it makes easy. It might be more like miss leading that it is going to use somewhere. It would be better to delete.

kparichay: done.

jijoongmoon (approved):
LGTM

PR #843, opened 2021-01-04
[dynamic finetuning] Adding dynamic-finetuning code

kparichay:
Commit 1 : [dyanmic training] Adding dynamic-training code …

Added dynamic-training code with both max and l2norm mode
Verified working with existing examples given the threshold

Commit 2: [dynamic-training] Add dynamic training using derivatives …

Added dynamic training using derivatives where the decision to
apply the gradient is calculated using the derivative received
without calculating the gradient itself.

**Self evaluation:**

### DIFF
--- a/nntrainer/models/dynamic_training_optimization.cpp
+++ b/nntrainer/models/dynamic_training_optimization.cpp
@@ -1,0 +1,144 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file   dynamic_training_optimization.cpp
+ * @date   5 January 2021
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is Dynamic Training Optimization for Neural Network
+ *
+ */
+
+#include <random>
+#include <vector>
+
+#include <dynamic_training_optimization.h>
+#include <layer_internal.h>
+#include <tensor.h>
+#include <util_func.h>
+
+namespace nntrainer {
+DynamicTrainingOptimization::DynamicTrainingOptimization(int threshold_,
+                                                         int skip_n_iter) :
+  threshold(threshold_),
+  enabled(false),
+  epsilon(1e-7),
+  skip_n_iterations(skip_n_iter) {
+  reduce_op = reduceByNorm;
+  calc_ratio_op = ratioUsingDerivative;
+  rng.seed(getSeed());
+  dist = std::uniform_real_distribution<float>(0.0, 1.0);
+}
+
+/**
+ * @brief     Check if the given weights can skip updating
+ * @note true if should be applied, else false
+ */
+bool DynamicTrainingOptimization::checkIfApply(
+  const std::vector<Weight> &weights, const std::shared_ptr<Var_Grad> &input,
+  const std::shared_ptr<Var_Grad> &output,
+  const std::shared_ptr<Optimizer> &opt, int iteration) {
+  if (!enabled || iteration < skip_n_iterations)
+    return true;
+
+  std::vector<bool> apply;
+  apply.reserve(weights.size());
+
+  for (auto const &weight : weights)
+    apply.push_back(checkIfApply(weight, input, output, opt, iteration));
+
+  return std::accumulate(apply.begin(), apply.end(), true,
+                         std::logical_and<bool>());
+}
+
+/**
+ * @brief     Check if the given weight can skip updating
+ * @note true if should be applied, else false
+ */
+bool DynamicTrainingOptimization::checkIfApply(
+  const Weight &weight, const std::shared_ptr<Var_Grad> &input,
+  const std::shared_ptr<Var_Grad> &output,
+  const std::shared_ptr<Optimizer> &opt, int iteration) {
+  if (iteration < skip_n_iterations)
+    return true;
+
+  if (!weight.getTrainable() || weight.getGradientRef().uninitialized())
+    return true;
+
+  float reduced_ratio = calc_ratio_op(weight, input, output, reduce_op);
+
+  return checkIfApply(reduced_ratio, (float)opt->getLearningRate(iteration));
+}
+
+/**
+ * @brief   Calculate the ratio of update to the weight using derivative
+ */
+float DynamicTrainingOptimization::ratioUsingDerivative(
+  const Weight &weight, const std::shared_ptr<Var_Grad> &input,
+  const std::shared_ptr<Var_Grad> &output,
+  std::function<float(Tensor const &)> reduce_op) {
+  float reduced_derivative = reduce_op(output->getGradientRef());
+  float reduced_input = reduce_op(input->getVariableRef());
+  float reduced_weight = reduce_op(weight.getVariableRef());
+  float reduced_grad = reduced_derivative * reduced_input;
+
+  return reduced_grad / reduced_weight;
+}
+
+/**
+ * @brief   Calculate the ratio of update to the weight using gradient
+ */
+float DynamicTrainingOptimization::ratioUsingGradient(
+  const Weight &weight, const std::shared_ptr<Var_Grad> &input,
+  const std::shared_ptr<Var_Grad> &output,
+  std::function<float(Tensor const &)> reduce_op) {
+  Tensor ratio = weight.getGradientRef().divide(weight.getVariableRef());
+  return reduce_op(ratio);
+}
+
+/**
+ * @brief   Check if the update should be applied or skipped
+ * @note true if should be applied, else false
+ */
+bool DynamicTrainingOptimization::checkIfApply(float reduced_ratio,
+                                               float learning_rate) {
+  /**
+   * If the reduced update ratio is higher than 1, then always apply update.
+   * If the reduced update raito is less than 1, then apply it with
+   * probability = update ratio
+   */
+  if (dist(rng) < reduced_ratio * learning_rate / threshold)
+    return true;
+
+  return false;
+}
+
+/**
+ * @brief     Operation to decide if update should be skipped
+ * @note      Calculate l0 norm of the tensor
+ */
+float DynamicTrainingOptimization::reduceByMax(Tensor const &ratio) {
+  return ratio.max_abs();
+}
+
+/**
+ * @brief     Operation to decide if update should be skipped
+ * @note      Calcalate l2 norm of the tensor averaged by its size
+ */
+float DynamicTrainingOptimization::reduceByNorm(Tensor const &ratio) {
+  float l2norm = ratio.l2norm();
+  return l2norm / std::sqrt(ratio.length());
+}
+
+/**< Different types of reduce operations */
+const std::string DynamicTrainingOptimization::dft_opt_max = "max";
+const std::string DynamicTrainingOptimization::dft_opt_norm = "norm";
+
+const std::string DynamicTrainingOptimization::dft_opt_mode_gradient =
+  "gradient";
+const std::string DynamicTrainingOptimization::dft_opt_mode_derivative =
+  "derivative";
+
+} /* namespace nntrainer */
--- a/nntrainer/models/dynamic_training_optimization.h
+++ b/nntrainer/models/dynamic_training_optimization.h
@@ -2,12 +2,33 @@
 /**
  * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
  *
- * @file   activation_layer.cpp
+ * @file   dynamic_training_optimization.h
  * @date   4 January 2021
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Parichay Kapoor <pk.kapoor@samsung.com>
  * @bug    No known bugs except for NYI items
  * @brief  This is Dynamic Training Optimization for Neural Network
+ *
+ * Dynamic training aims to optimize the cost of applying the gradient.
+ * The cost of applying the gradient includes the cost of the optimizer (adam,
+ * etc) where the optimizer variables are updated, and the cost of actually
+ * updating the weights (which can be non-trivial with bigger weights and
+ * distributed training).
+ *
+ * There are two supported modes:
+ * 1. Gradient Mode: The already calculated gradient is used to estimate if this
+ * gradient must be used to update the weight, or if this update must be
+ * skipped.
+ *
+ * 2. Derivative Mode: This mode tries to estimate an approximate gradient with
+ * low cost in order to save the cost of calculating gradient. This cost of
+ * calculating gradient is wasted if the gradient is not going to be applied.
+ *
+ * There are two supported reduction operations which reduce the gradient and
+ * the weight to a single value in order to compare it with a threshold.
+ * If the reduced value is less than threshold, the update is performed with
+ * some probabilty proportional to the value. If the reduced value is higher
+ * than threshold, then the update is always performed.
  *
  */
 
@@ -20,7 +41,6 @@
 
 #include <layer_internal.h>
 #include <tensor.h>
-#include <util_func.h>
 
 namespace nntrainer {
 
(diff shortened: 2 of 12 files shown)


nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * @brief     Set threshold for optimization
+   */
+  void setThreshold(float threshold_) { threshold = threshold_; };
+
+  /**
+   * @brief     Set the reduce operation for dynamic optimization
+   */
+  void setOp(std::string op) {
```
zhoonit, line 45:
```suggestion
  void setOp(const std::string &op) {
```
How about using const &?

nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * @brief     Disable the optimization
+   */
+  void disable() { enabled = false; }
+
+  /**
+   * @brief     Set the mode for optimization
+   */
+  void setMode(std::string mode_) {
```
zhoonit, line 67: same

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+  dist = std::uniform_real_distribution<float>(0.0, 1.0);
+}
+
+/**
+ * @brief     Check if the given weights can skip updating
+ * @note true if should be applied, else false
+ */
+bool DynamicTrainingOptimization::checkIfApply(
+  const std::vector<Weight> &weights, const std::shared_ptr<Var_Grad> input,
+  const std::shared_ptr<Var_Grad> output, const std::shared_ptr<Optimizer> opt,
```
zhoonit, line 41: I guess those can be `const std::shared_ptr<T> &` to save locking?

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * @brief     Check if the given weight can skip updating
+ * @note true if should be applied, else false
+ */
+bool DynamicTrainingOptimization::checkIfApply(
+  const Weight &weight, const std::shared_ptr<Var_Grad> &input,
+  const std::shared_ptr<Var_Grad> &output,
+  const std::shared_ptr<Optimizer> &opt, int iteration) {
+  if (iteration < skip_n_iterations)
```
lhs8928, line 64: Already check at line 43

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return true;
+
+  if (!weight.getTrainable() || weight.getGradientRef().uninitialized())
```
lhs8928, line 67: Isn't it right return false(skip) when trainable is set to false?

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+  const std::shared_ptr<Var_Grad> output, const std::shared_ptr<Optimizer> opt,
+  int iteration) {
+  if (!enabled || iteration < skip_n_iterations)
+    return true;
+
+  std::vector<bool> apply;
+  apply.reserve(weights.size());
+
+  for (auto const &weight : weights)
+    apply.push_back(checkIfApply(weight, input, output, opt, iteration));
```
zhoonit, line 50: instead of apply vector, how about return here when false?

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ * @brief   Check if the update should be applied or skipped
+ * @note true if should be applied, else false
+ */
+bool DynamicTrainingOptimization::checkIfApply(float reduced_ratio,
+                                               float learning_rate) {
+  /**
+   * If the reduced update ratio is higher than 1, then always apply update.
```
lhs8928, line 108: According to the code below it seems that reduced update ratio is higher than 1 would it skip the update

nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,183 @@
...
+   * @brief     Check if the gradient mode is used for optimization
+   */
+  bool isGradientMode() {
+    if (enabled && calc_ratio_mode == dft_opt_mode_gradient)
+      return true;
+    return false;
+  }
+
+  /**
+   * @brief     Set initial iterations to skip from optimization
```
lhs8928, line 96: One simple question.
Does this variable means that at first n_iteration it want to skip the gradient or apply the gradient?

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+  return ratio.max_abs();
+}
+
+/**
+ * @brief     Operation to decide if update should be skipped
+ * @note      Calcalate l2 norm of the tensor averaged by its size
+ */
+float DynamicTrainingOptimization::reduceByNorm(Tensor const &ratio) {
+  float l2norm = ratio.l2norm();
+  return (l2norm * l2norm) / ratio.length();
```
lhs8928, line 132: Is there any reason to square l2norm?

nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,183 @@
...
+    else if (mode_ == dft_opt_mode_gradient)
+      calc_ratio_op = ratioUsingGradient;
+    else
+      enabled = false;
+  }
+
+  /**
+   * @brief     Check if the derivative mode is used for optimization
+   */
+  bool isDerivativeMode() {
```
lhs8928, line 80: What is the derivative mode and when it is used?

nntrainer/models/dynamic_training_optimization.h
kparichay: done

nntrainer/models/dynamic_training_optimization.h
kparichay: done

nntrainer/models/dynamic_training_optimization.cpp
kparichay: currently, all the values are accumulated with `logical_and`, but this is not always true.
In the final version (which will be next PR), weights and biases are separated.
So, this will be completely refactored there.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: This function is also public and can be called from outside.
Hence the second check

nntrainer/models/neuralnet.cpp
```diff
@@ -346,25 +378,20 @@ void NeuralNetwork::backwarding(int iteration) {
...
    * The last trainable layer need not calculate the derivatives
-   * Do not change this order:
-   * 1. calcGradient
-   * 2. calcDerivative
-   * 3. applyGradient
    */
-  last_layer->calcGradient();
 #ifdef ENABLE_TEST
-  last_layer->calcDerivative();
+  backwarding((iter_end - 1)->layer, iteration, true);
```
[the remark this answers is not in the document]
kparichay: udpated.

nntrainer/models/dynamic_training_optimization.h
kparichay: added in the description.

nntrainer/models/dynamic_training_optimization.cpp
kparichay:
```python3
l2norm =  sqrt(sum([(x^2) for x in data]))   # python based pseudo code
l2norm / length = sqrt(... / (length ^ 2)) 
```

As you can see, simply l2norm by length actually divides by `length^2`.

nntrainer/models/dynamic_training_optimization.h
kparichay: for `first_n_iteration`, dynamic training optimization will be skipped, and gradients will always be applied.
Updated description to be clearer.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: This was inverted. Updated the return values.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: If trainable is set to false, dynamic training optimization cannot make the decision to skip the update or not as the gradient/derivative might not be available.
So, the decision making here is to go with default behavior as if there was no dynamic optimization, which is to return true in this case.

Please let me know if this is counter-intuitive.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: I updated to sqrt of the whole thing.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: ohh, i missed in this function. Updated.

nntrainer/models/dynamic_training_optimization.cpp
zhoonit: Noted.

nntrainer/models/dynamic_training_optimization.h
lhs8928: Got it.
But it is not used. Will use this in next PR?

nntrainer/models/dynamic_training_optimization.cpp
lhs8928: Understood.

nntrainer/models/dynamic_training_optimization.h
kparichay: aah yes.

nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,210 @@
...
+public:
+  /**
+   * @brief     Constructor of DynamicFineTuning Optimization
+   */
+  DynamicTrainingOptimization(int threshold_ = 1, int skip_n_iter = 1);
+
+  /**
+   * @brief     Set threshold for optimization
+   */
+  void setThreshold(float threshold_) { threshold = threshold_; };
```
jijoongmoon, line 61: need to add check if the threshold is greater than 1?

nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,210 @@
...
+
+  /**
+   * @brief     Set the reduce operation for dynamic optimization
+   */
+  void setOp(const std::string &op) {
+    if (op == dft_opt_max)
```
jijoongmoon, line 67: may be, provide some notification? if the op is not proper?

nntrainer/models/dynamic_training_optimization.h
```diff
@@ -0,0 +1,210 @@
...
+   * @brief     Set the mode for optimization
+   */
+  void setMode(const std::string &mode_) {
+    calc_ratio_mode = mode_;
+    if (mode_ == dft_opt_mode_derivative)
+      calc_ratio_op = ratioUsingDerivative;
+    else if (mode_ == dft_opt_mode_gradient)
+      calc_ratio_op = ratioUsingGradient;
+    else
+      enabled = false;
```
jijoongmoon, line 95: give proper notification?

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return true;
+
+  float reduced_ratio = calc_ratio_op(weight, input, output, reduce_op);
+
+  return checkIfApply(reduced_ratio, (float)opt->getLearningRate(iteration));
+}
+
+/**
+ * @brief   Calculate the ratio of update to the weight using derivative
+ */
```
jijoongmoon, line 77: better to add comments about input is actual input and output is output about forwarding direction.

nntrainer/models/dynamic_training_optimization.h
kparichay: Threshold is used a divider against the reduced value. 
So threshold can be anything but zero.

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+float DynamicTrainingOptimization::ratioUsingDerivative(
+  const Weight &weight, const std::shared_ptr<Var_Grad> &input,
+  const std::shared_ptr<Var_Grad> &output,
+  std::function<float(Tensor const &)> reduce_op) {
+  float reduced_derivative = reduce_op(output->getGradientRef());
```
jijoongmoon, line 82: is it reduced already in batch direction?

nntrainer/models/dynamic_training_optimization.cpp
```diff
@@ -0,0 +1,144 @@
...
+ */
+bool DynamicTrainingOptimization::checkIfApply(float reduced_ratio,
+                                               float learning_rate) {
+  /**
+   * If the reduced update ratio is higher than 1, then always apply update.
+   * If the reduced update raito is less than 1, then apply it with
+   * probability = update ratio
+   */
+  if (dist(rng) < reduced_ratio * learning_rate / threshold)
+    return true;
```
jijoongmoon, line 113: need to check if threshold is greater than 0.

nntrainer/models/dynamic_training_optimization.h
kparichay: Will add throw.

nntrainer/models/dynamic_training_optimization.h
kparichay: will add throw.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: This is the old method. The final method using approximate gradient over batch will be added in next PR.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: Added while setting the threshold.

nntrainer/models/dynamic_training_optimization.cpp
kparichay: added.

nntrainer/models/dynamic_training_optimization.h
jijoongmoon: I think at least we have to give them a notification what threshold which greater than 1.0 means. if the threshold is greater than 1, then it means only layer which approximation of gradient is greater than weight is updated. IMO, most of layer does not updated in normal cases.

kparichay: Then that notification should be give for even smaller thresholds.
Threshold of 1e-3 for resnet50 led to 75% skip of gradient while 1e-3 for VGG16 led to over 90% skip.
So, I am not sure what notification can be given for threshold over 1, as there is no direct meaning to the threshold values by itself.

jijoongmoon: does it means most layers are going to update if threshold over 1 is used.  Am I right? If it is, then it is kind of the opposite of user's intention.

kparichay: So, the value of threshold cannot be directly compared to 1. Its range is a lot dependent on model, data and training configuration.

nntrainer/tensor/var_grad.h
```diff
@@ -211,6 +211,20 @@ class Var_Grad {
    */
   Tensor &getGradientRef() { return *grad.get(); }
 
+  /**
+   * @brief Get the variable tensor (by reference)
+   *
+   * @return Tensor Variable tensor
+   */
+  Tensor &getVariableRef() const { return *var.get(); }
```
zhoonit, line 219: return const reference? (and also for :226)

kparichay: In many cases, we need to modify the tensor or call in_place operations on this.
That won't be possible with const.

zhoonit: If modifying tensor take places, shouldn't it be `var_grad` rather than `const var_grad` in the first place?

kparichay: done.

zhoonit: Glad that it didn't.

zhoonit (approved):
Great work man 👍

nntrainer/models/dynamic_training_optimization.h
jijoongmoon: > Its range is a lot dependent on model, data and training configuration.

Well, if you say so.. let's see. leave as it is.

jijoongmoon (approved):
LGTM

PR #846, opened 2021-01-06
[data augmentation] Added data augmentation

kparichay:
Added data augmentation for vgg model
Data augmentation support random flip and random translate
both applied on each image in a batch independently.

### DIFF
--- a/nntrainer/layers/preprocess_flip_layer.cpp
+++ b/nntrainer/layers/preprocess_flip_layer.cpp
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	preprocess_flip_layer.cpp
+ * @date	20 January 2020
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is Preprocess Random Flip Layer Class for Neural Network
+ *
+ */
+
+#include <random>
+
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <preprocess_flip_layer.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+const std::string PreprocessFlipLayer::type = "preprocess_flip";
+
+const std::string PreprocessFlipLayer::flip_horizontal = "horizontal";
+const std::string PreprocessFlipLayer::flip_vertical = "vertical";
+const std::string PreprocessFlipLayer::flip_horizontal_vertical =
+  "horizontal_and_vertical";
+
+int PreprocessFlipLayer::initialize(Manager &manager) {
+  output_dim = input_dim;
+
+  rng.seed(getSeed());
+  flip_dist = std::uniform_real_distribution<float>(0.0, 1.0);
+
+  return ML_ERROR_NONE;
+}
+
+void PreprocessFlipLayer::setProperty(const PropertyType type,
+                                      const std::string &value) {
+  switch (type) {
+  case PropertyType::flip_direction:
+    if (!value.empty()) {
+      if (istrequal(value, flip_horizontal)) {
+        flipdirection = FlipDirection::horizontal;
+      } else if (istrequal(value, flip_vertical)) {
+        flipdirection = FlipDirection::vertical;
+      } else if (istrequal(value, flip_horizontal_vertical)) {
+        flipdirection = FlipDirection::horizontal_and_vertical;
+      } else {
+        throw std::invalid_argument("Argument flip direction is invalid");
+      }
+    }
+    break;
+  default:
+    Layer::setProperty(type, value);
+    break;
+  }
+}
+
+void PreprocessFlipLayer::forwarding() {
+  using std::swap;
+  bool fliph, flipw;
+
+  for (unsigned int idx = 0; idx < input_dim.size(); idx++) {
+    Tensor &hidden_ = net_hidden[idx]->getVariableRef();
+    Tensor &input_ = net_input[idx]->getVariableRef();
+    unsigned int width = input_dim[idx].width();
+    unsigned int height = input_dim[idx].height();
+
+    for (unsigned int b = 0; b < input_dim[idx].batch(); b++) {
+      fliph = flipw = false;
+      if (flip_dist(rng) < 0.5 && flipdirection != FlipDirection::vertical)
+        flipw = true;
+
+      if (flip_dist(rng) < 0.5 && flipdirection != FlipDirection::horizontal)
+        fliph = true;
+
+      if (!flipw && !fliph)
+        continue;
+
+      if (flipw) {
+        for (unsigned int c = 0; c < input_dim[idx].channel(); c++)
+          for (unsigned int h = 0; h < input_dim[idx].height(); h++)
+            for (unsigned int w = 0; w < input_dim[idx].width() / 2; w++)
+              swap(*input_.getAddress(b, c, h, w),
+                   *input_.getAddress(b, c, h, width - w - 1));
+      }
+      if (fliph) {
+        for (unsigned int c = 0; c < input_dim[idx].channel(); c++)
+          for (unsigned int h = 0; h < input_dim[idx].height() / 2; h++)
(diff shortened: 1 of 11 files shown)


kparichay:
@jijoongmoon Is the dependency on opencv for data augmentation acceptable or we must devise these layers ourselves?

nntrainer/layers/input_layer.cpp
```diff
@@ -55,26 +84,48 @@ void InputLayer::setProperty(const PropertyType type,
...
+  Tensor &input_ = net_input[0]->getVariableRef();
 
   if (normalization)
-    hidden_.normalization_i();
+    input_.normalization_i();
   if (standardization)
-    hidden_.standardization_i();
+    input_.standardization_i();
+
+  unsigned int batch = input_dim[0].batch();
```
kparichay, line 94: if (augmentation), then do augmentation, else hidden_ = input_

kparichay: resolved

jijoongmoon:
I'm not sure about opencv dependency. I think it might be better to use Eigen or use openBLAS.

kparichay:
> I'm not sure about opencv dependency. I think it might be better to use Eigen or use openBLAS. Let's discuss further.

WIth this PR, opencv dependency is optional. The code works without opencv, just the `random_translate` based data augmentation is not supported. We can change the dependency to some other library if we want.

nntrainer/layers/preprocess_flip_layer.cpp
```diff
@@ -0,0 +1,115 @@
...
+      }
+      if (fliph) {
+        for (unsigned int c = 0; c < input_dim[idx].channel(); c++)
+          for (unsigned int h = 0; h < input_dim[idx].height() / 2; h++)
+            for (unsigned int w = 0; w < input_dim[idx].width(); w++)
+              swap(*input_.getAddress(b, c, h, w),
+                   *input_.getAddress(b, c, height - h - 1, w));
+      }
+    }
+    hidden_ = input_;
```
jijoongmoon, line 98: Can we use in-place ?? input and hidden?

nntrainer/layers/preprocess_flip_layer.h
```diff
@@ -0,0 +1,114 @@
...
+ * @brief   Preprocess FLip Layer
+ */
+class PreprocessFlipLayer : public Layer {
+public:
+  /**
+   * @brief     Constructor of Preprocess FLip Layer
+   */
+  template <typename... Args>
+  PreprocessFlipLayer(Args... args) :
+    Layer(args...),
```
jijoongmoon, line 36: How about.. nntrainer::augmetation::FilpLayer?

nntrainer/layers/preprocess_translate_layer.cpp
```diff
@@ -0,0 +1,134 @@
...
+            input_mat.at<cv::Vec3f>(h, w)[c] = input_.getValue(b, c, h, w);
+
+      cv::warpAffine(input_mat, output_mat, affine_transform_mat,
+                     output_mat.size(), cv::WARP_INVERSE_MAP,
+                     cv::BORDER_REFLECT);
+
+      for (unsigned int c = 0; c < input_dim[idx].channel(); c++)
+        for (unsigned int h = 0; h < input_dim[idx].height(); h++)
+          for (unsigned int w = 0; w < input_dim[idx].width(); w++)
+            hidden_.setValue(b, c, h, w, output_mat.at<cv::Vec3f>(h, w)[c]);
```
jijoongmoon, line 112: Once we finish line # 103, we can use the buffer for line112. I guess.

nntrainer/layers/preprocess_flip_layer.cpp
kparichay: This implementation will work in-place, as we directly modify the input.
The manager has to assign the memory with in-place setting though, which will be done outside this PR.

It is fine to use in-place here as no layer behind this layer can train.

nntrainer/layers/preprocess_flip_layer.h
kparichay: Then how about nntrainer::dataset::FlipLayer?
As input normalization, resizing etc will also be added, and they are although for dataset but not augmentation.

nntrainer/layers/preprocess_translate_layer.cpp
kparichay: The bottleneck is line 105, where `input_mat` and `output_mat` have to be different.

nntrainer/layers/preprocess_flip_layer.h
jijoongmoon: We could make sub-namespace for others too. I guess, let's start from here.

jijoongmoon: > Then how about nntrainer::dataset::FlipLayer?
> As input normalization, resizing etc will also be added, and they are although for dataset but not augmentation.
currently we support input normalization as an property. I think then it should be the same with this. So.. That could be a layer?

nntrainer/layers/preprocess_translate_layer.cpp
jijoongmoon: > The bottleneck is line 105, where `input_mat` and `output_mat` have to be different.
> We can use `input_` and `hidden_` to be the same for now, but it might lead to optimizations, as this layer cannot work in-place. It just seems to work in-place because of the extra memory allocated by OpenCV.

I'm saying for the opencv case.

nntrainer/layers/preprocess_flip_layer.cpp
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+    break;
+  default:
+    Layer::setProperty(type, value);
+    break;
+  }
+}
+
+void PreprocessFlipLayer::forwarding() {
```
zhoonit, line 62: Could you check if flipdirection is used in the function?

nntrainer/layers/preprocess_translate_layer.h
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+#ifndef __PREPROCESS_TRANSLATE_LAYER_H__
+#define __PREPROCESS_TRANSLATE_LAYER_H__
+#ifdef __cplusplus
+
+#include <random>
+
+#if defined(ENABLE_DATA_AUGMENTATION_OPENCV)
+#include <opencv2/highgui/highgui.hpp>
```
zhoonit, line 21: instead of including header, how about declare `class cv::Mat;` here?

nntrainer/layers/preprocess_flip_layer.h
kparichay: With property, the user has to set two configurations - layertype and property. But with property, it can be done with just 1 configuration.
Additionally, it's being coherent with other popular frameworks, but we can make it as a property like activation layer.

nntrainer/layers/preprocess_translate_layer.h
kparichay: That would allow compilation but would break while running. Wouldn't that be opposite of declaring dependency at compile-time?

zhoonit: Correct me If I am wrong, those `cv::Mat` is private and used only inside `.cpp`. So there seems declaring here shouldn't cause any undefined reference error. if someone would access those variables, would have a runtime error but it seems the case would never happen.

nntrainer/layers/preprocess_translate_layer.cpp
kparichay: done.

nntrainer/layers/preprocess_flip_layer.cpp
kparichay: oops.
Added.

zhoonit (approved):
LGTM :)

nntrainer/layers/preprocess_translate_layer.h
kparichay: The `cv::Mat` object in private section creates objects in header, so forward declare wont work. 
There is a way to convert them to unique_ptr, but its too much for now as opencv dependency wont be kept for long anyway.

zhoonit: Yes. That's too much. Thanks for the confirmation!

nntrainer/layers/preprocess_flip_layer.h
jijoongmoon: If we are supporting both ways, then I think it's ok.

jijoongmoon (approved):
LGTM

PR #852, opened 2021-01-14
[ Fix ] Fix Logistic Regression Example Error

jijoongmoon:
This PR includes fixes about logistic regression application

Change forwarding function

**Self evaluation:**

### DIFF
--- a/Applications/LogisticRegression/jni/main.cpp
+++ b/Applications/LogisticRegression/jni/main.cpp
@@ -217,7 +217,8 @@ int main(int argc, char *argv[]) {
 
       try {
         float answer =
-          NN.forwarding({MAKE_SHARED_TENSOR(nntrainer::Tensor({o}))})[0]
+          NN.forwarding({MAKE_SHARED_TENSOR(nntrainer::Tensor({o}))},
+                        {MAKE_SHARED_TENSOR(nntrainer::Tensor({l}))})[0]
             ->apply(stepFunction)
             .getValue(0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


Applications/LogisticRegression/jni/main.cpp
```diff
@@ -217,7 +217,8 @@ int main(int argc, char *argv[]) {
 
       try {
         float answer =
-          NN.forwarding({MAKE_SHARED_TENSOR(nntrainer::Tensor({o}))})[0]
+          NN.forwarding({MAKE_SHARED_TENSOR(nntrainer::Tensor({o}))},
+                        {MAKE_SHARED_TENSOR(nntrainer::Tensor({l}))})[0]
```
kparichay, line 221: I wonder what effect does this change has. 
IMO the only difference this will make is calculate loss, which in this case should not affect the running or the result.

lhs8928 (approved):
LGTM

Applications/LogisticRegression/jni/main.cpp
jijoongmoon: I guess it is temporal. we need to provide inference API. not call forward directly in the application. If we call directly.. there is error because hidden is not set properly.

kparichay: Got it.

kparichay (approved):
LGTM

zhoonit (approved):
LGTM. I wonder if we could temporarily use `NeuralNetwork::inference` for this case.

Issue #856, opened 2021-01-18
in / out / weights shouldn't be allocated inside layer


zhoonit:
Since in / out / weights are managed outside, there shouldn't be any case of reallocating in / out / weights (const std::shared_ptr would do?).

kparichay:
These objects can change, if model is allowed to change dynamically (and then initialize again before usage).
One of the examples is dynamically removing and freeing the memory associated with data augmentation layers.

zhoonit:
> These objects can change, if model is allowed to change dynamically (and then initialize again before usage).
> One of the examples is dynamically removing and freeing the memory associated with data augmentation layers.

I guess weights are okay to change dynamically. I think in & out shouldn't since it interacts with each other.
I'll check data augmentation layer to get a better understanding

kparichay:
Data augmentation layer PR is not finalized yet. I will finalize it today.
Data augmentation layer will follow the same semantics as any other layer.
However, it is supposed to be not present in the graph in inference mode (a quick workaround will be to have a no-op with data augmentation layers in inference stage but a user getting the graph might be confused looking at data augmentation layers being executed in inference stage).

kparichay:
I think this is resolved with layer_v2?

zhoonit:
This is a bit different story but I guess this can be closed as it is stale now.

PR #857, opened 2021-01-18
[Tensor] Add outplace method for arithmetic ops

zhoonit:
Add outplace ops with already allocated tensor.

**Self evaluation:**

See also #856

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -183,16 +183,27 @@ Tensor::Tensor(
           this->setValue(i, j, k, l, d[i][j][k][l]);
 }
 
-int Tensor::multiply_i(float const &value) {
+Tensor Tensor::multiply(float const &value, Tensor &out) const {
+  /// @note this is not depending on multiply_i as there is an optimized version
+  /// of multiply_i
+  CREATE_IF_EMPTY_DIMS(out, getDim());
+  const float *data = getData();
 
+  std::transform(data, data + length(), out.getData(),
+                 std::bind2nd(std::multiplies<float>(), value));
+
+  return out;
+}
+
+Tensor Tensor::multiply(float const &value) { CLONE_OP_I(multiply_i, value); }
+
+int Tensor::multiply_i(float const &value) {
   float *data = getData();
   unsigned int len = length();
 
   sscal(len, value, data, 1);
   return ML_ERROR_NONE;
 }
-
-Tensor Tensor::multiply(float const &value) { CLONE_OP_I(multiply_i, value); }
 
 int Tensor::divide_i(float const &value) {
   if (value == 0.0f) {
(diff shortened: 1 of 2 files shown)


jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -183,17 +183,26 @@ Tensor::Tensor(
           this->setValue(i, j, k, l, d[i][j][k][l]);
 }
 
-int Tensor::multiply_i(float const &value) {
+Tensor Tensor::multiply(float const &value, Tensor &out) const {
+  CREATE_IF_EMPTY_DIMS(out, getDim());
```
kparichay, line 187: Let's add a note here that the implementation of `multiply` does not depend on `multiply_i`.
Instead, uses a different methodology as `sscal` only works inplace.

nntrainer/tensor/tensor.cpp
```diff
@@ -219,8 +228,27 @@ int Tensor::add_i(float const &value) {
   return ML_ERROR_NONE;
 }
 
+Tensor Tensor::add(Tensor const &m, Tensor &out, float const alpha) const {
+  CREATE_IF_EMPTY_DIMS(out, getDim());
+  scopy(length(), getData(), 1, out.getData(), 1);
```
kparichay, line 233: do we really need the copy? we can do it in a single call to transform.

nntrainer/tensor/tensor.h
```diff
@@ -250,6 +250,14 @@ class Tensor {
    */
   Tensor multiply(float const &value);
 
+  /**
+   * @brief     multiply value element by element
+   * @param[in] value Divisor
```
kparichay, line 255: divisor -> multiplier

zhoonit: Done

nntrainer/tensor/tensor.cpp
zhoonit: This will be changed in upcoming PR. Thanks :)

nntrainer/tensor/tensor.cpp
zhoonit: Done

kparichay (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -219,8 +230,27 @@ int Tensor::add_i(float const &value) {
...
 
+Tensor Tensor::add(float const &value, Tensor &out) const {
+  const float *data = getData();
+
+  CREATE_IF_EMPTY_DIMS(out, getDim());
+
+  std::transform(data, data + length(), out.getData(),
+                 std::bind2nd(std::plus<float>(), value));
+
+  return out;
```
jijoongmoon, line 251: so.. if we are using this... the dimensions is going to change. and if it is the same.. then.. we don't need to another buffer?

zhoonit: you mean out dimension? if uninitialized, it is initialized with `this->getDim()`, if dimension is not correct it is UB 🍫 .

 🍫  is intended as the function is in critical path.

jijoongmoon: On second thought... It generates a quite of bug which is really hard to find.

zhoonit: I think it's good idea. I'll update on #870 since this patch is being depended upon other PR :)

zhoonit: Update done :)

PR #858, opened 2021-01-18
[Fix/TFlite] Fix tflite allocation

zhoonit:
Now, memory alllocation is handled outside of each layer.
Accordingly, allocating out tensor shouldn't be done inside a layer.

For the same reason, loss layer backwarding needs some fix, for now
it is just commented and will be handled soon

This patch handles the issue for tflite layer

**Self evaluation:**

See also #856

### DIFF
--- a/nntrainer/layers/tflite_layer.cpp
+++ b/nntrainer/layers/tflite_layer.cpp
@@ -98,36 +98,44 @@ void TfLiteLayer::setProperty(const PropertyType type,
 
 void TfLiteLayer::forwarding(bool training) {
 #ifdef DEBUG
-  std::vector<TensorDim> dims;
   if (net_input.size() != input_dim.size())
     throw std::invalid_argument("Provided number of input dimensions mismatch");
 
-  for (int idx = 0; idx < dims.size(); idx++) {
+  for (unsigned int idx = 0; idx < input_dim.size(); idx++) {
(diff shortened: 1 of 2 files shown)


kparichay (approved):
LGTM

nntrainer/layers/tflite_layer.cpp
```diff
@@ -102,32 +102,25 @@ void TfLiteLayer::forwarding() {
...
     interpreter->tensor(in_indices[idx])->data.raw =
       (char *)net_input[idx]->getVariableRef().getData();
 
   auto out_indices = interpreter->outputs();
-  out.resize(out_indices.size());
   for (size_t idx = 0; idx < out_indices.size(); idx++) {
-    out[idx] = MAKE_SHARED_TENSOR(output_dim[idx]);
     interpreter->tensor(out_indices[idx])->data.raw =
-      (char *)out[idx]->getData();
+      reinterpret_cast<char *>(net_hidden[idx]->getVariableRef().getData());
```
jijoongmoon, line 118: I think we need to check the dimension.

zhoonit: It will match as line:60 sets dimension from tflite, I added a failsafe checker for DEBUG mode :)

jijoongmoon (approved):
LGTM

PR #863, opened 2021-01-19
[meson] Update meson for ubuntu 20.04

kparichay:
Update meson to work with ubuntu 20.04
Also add some missing checks

Signed-off-by: Parichay Kapoor <pk.kapoor@samsung.com>

Issue #860, opened 2021-01-18
MNIST: inference example to the implementation.


lekogabi:
Dear NNTrainer Team!

I am trying to add an inference to the MNIST example. The training is working well. My modifications can be found here:
https://pastebin.com/3Ebb6MX9

One way to run after the build:
./nntrainer_mnist inference ../res/mnist.ini mnist_trainingSet.dat

I followed the inference example from the LogisticRegression implementation, but unfortunately I cannot make it to work.

Could you please take a look at it and give me some hints how to make some corrections, or could you provide a simple inference example to the MNIST implementation?

Thanks in advance!

Regards,
Gabor Leko

kparichay:
@lekogabi Thanks for the issue, I will take a look at it and report back today.

kparichay:
I tried the code you shared. The code ran successfully but there was a warning by the `cblas_sgemm`

```bash
** On entry to SGEMM  parameter number  3 had an illegal value
```

This is because the input data you are using is not of the appropriate shape.  In the MNIST model, the convolution layer expects a 2D input, however, the input provided is a 1-D input. To solve this, please apply the below change to your application code:

```diff
-auto answer = NN->inference({MAKE_SHARED_TENSOR(nntrainer::Tensor({o}))});
+nntrainer::Tensor T =nntrainer::Tensor({o});
+T.reshape({1,1,28,28});
+auto answer = NN->inference({MAKE_SHARED_TENSOR(T)});
```

`Convolution layer` does not perform shape checks only at compile time and not at runtime for optimization purposes.
In the current implementation, the `input layer` does not reshape the layer directly as in certain cases, reshape is not enough and transpose is required.

Please try this and let me know if your problem is resolved.

kparichay:
#864 will log the error and return empty output in case of input dimension mismatch in inference.

lekogabi:
Hi,
Thank you. It solved my problem.
Gabor

PR #864, opened 2021-01-19
[#863] [inference] Add input validation for inference

kparichay:
Add input validation for inference of the neural network

See also #860

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -463,6 +463,33 @@ void NeuralNetwork::setBatchSize(unsigned int batch) {
     throw std::invalid_argument("Error setting batchsize for the dataset");
 }
 
+bool NeuralNetwork::validateInput(sharedConstTensors X) {
+
+  auto &first_layer = model_graph.getSortedLayerNode(0).layer;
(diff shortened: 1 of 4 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -472,9 +499,12 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X) {
     setBatchSize(X[0]->batch());
   }
 
+  sharedConstTensors out;
+  if (!validateInput(X))
+    return out;
```
zhoonit, line 504: how about throw instead of return out?

zhoonit (approved):
LGTM

nntrainer/models/neuralnet.cpp
kparichay: Line 516 follows the same behavior of return out with a log.
So does train().
I think let's leave it with the same behavior for now and later change all to throw exceptions on error.

zhoonit: Sounds decent

jijoongmoon (approved):
LGTM

PR #870, opened 2021-01-21
[Tensor] rearrange methods

zhoonit:
- ~**#857** [Tensor] Add outplace method for arithmetic ops~
- [Tensor/Clean] Relocate tensor methods
```
This patch rearranges arithmetic methods while adding some missing
outplace operation signature.

Order is

1. inplace -> outplace -> outplace (with allocated memory)
2. multiply -> divide -> add -> subtract
3. scalar -> tensor

**Self evaluation:**

```
- [Tensor] Rearrange methods 
```
- Add missing out param methods
- Change way it is delegated for some methods
- Rename s/operator_/apply_broadcast
- Assure dimension checks
- remove `operator_i` and `operator_i_util`

**Self evaluation:**

```

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -59,19 +59,6 @@
     if (tensor.uninitialized())           \
       tensor = Tensor(__VA_ARGS__);       \
   } while (0);
-
-/** do clone of this, perform the operation and return the output */
-#define CLONE_OP_I(op, ...)                        \
-  do {                                             \
-    Tensor clone = this->clone();                  \
-    if (clone.op(__VA_ARGS__) != ML_ERROR_NONE) {  \
-      std::stringstream ss;                        \
-      ss << "Error: op " << __func__ << " failed"; \
-      throw std::invalid_argument(ss.str());       \
-    }                                              \
-    return clone;                                  \
-  } while (0);
-
 namespace nntrainer {
 
 /**
@@ -238,77 +225,6 @@ Tensor &Tensor::multiply(float const &value, Tensor &out) const {
   return apply(f, out);
 }
 
-int Tensor::divide_i(float const &value) {
-  if (value == 0.0f) {
-    return ML_ERROR_INVALID_PARAMETER;
-  }
-  this->divide(value, *this);
-  return ML_ERROR_NONE;
-}
-
-Tensor Tensor::divide(float const &value) const {
-  Tensor t;
-  return divide(value, t);
-}
-
-Tensor &Tensor::divide(float const &value, Tensor &out) const {
-  auto f = std::bind(std::divides<float>(), std::placeholders::_1, value);
-  /// @todo add unittest
-  if (value == 0.0f) {
-    std::stringstream ss;
-    ss << "[Tensor] divide by value failed, value: " << value;
-    throw std::invalid_argument(ss.str().c_str());
-  }
-  return apply(f, out);
-}
-
-int Tensor::add_i(float const &value) {
-  this->add(value, *this);
-  return ML_ERROR_NONE;
-}
-
-Tensor Tensor::add(float const &value) const {
-  Tensor t;
-  return add(value, t);
-}
-
-Tensor &Tensor::add(float const &value, Tensor &out) const {
-  /// @todo add unittest
-  auto f = std::bind(std::plus<float>(), std::placeholders::_1, value);
-  return apply(f, out);
-}
-
-int Tensor::subtract_i(float const &value) {
-  this->subtract(value, *this);
-  return ML_ERROR_NONE;
-}
-
-Tensor Tensor::subtract(float const &value) const {
-  Tensor t;
-  return subtract(value, t);
-}
-
-Tensor &Tensor::subtract(float const &value, Tensor &out) const {
-  /// @todo add unittest
-  auto f = std::bind(std::minus<float>(), std::placeholders::_1, value);
-  return apply(f, out);
-}
-
-int Tensor::pow_i(float exponent) {
-  pow(exponent, *this);
-  return ML_ERROR_NONE;
-}
-
-Tensor Tensor::pow(float exponent) const {
-  Tensor t;
-  return pow(exponent, t);
-}
-
-Tensor &Tensor::pow(float exponent, Tensor &out) const {
-  auto f = [exponent](float in) { return powf(in, exponent); };
-  return apply(f, out);
-}
-
 int Tensor::multiply_i(Tensor const &m) {
   try {
     this->multiply(m, *this);
@@ -340,6 +256,30 @@ Tensor &Tensor::multiply(Tensor const &m, Tensor &output) const {
   return output;
 }
 
+int Tensor::divide_i(float const &value) {
+  if (value == 0.0f) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+  this->divide(value, *this);
+  return ML_ERROR_NONE;
+}
+
+Tensor Tensor::divide(float const &value) const {
+  Tensor t;
+  return divide(value, t);
+}
+
+Tensor &Tensor::divide(float const &value, Tensor &out) const {
+  auto f = std::bind(std::divides<float>(), std::placeholders::_1, value);
+  /// @todo add unittest
+  if (value == 0.0f) {
+    std::stringstream ss;
+    ss << "[Tensor] divide by value failed, value: " << value;
+    throw std::invalid_argument(ss.str().c_str());
+  }
+  return apply(f, out);
+}
+
 int Tensor::divide_i(Tensor const &m) {
   try {
     this->divide(m, *this);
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -259,23 +370,48 @@ int Tensor::add_i(Tensor const &m, float const alpha) {
...
+Tensor &Tensor::add(Tensor const &m, Tensor &out, float const alpha) const {
+  /// v1 (avg, min, max)
+  /// for 10k + 10k this took 2, 2, 3
+  /// for 4k x 4k + 4k x 4k this took 16740, 15638, 32347
+  /// for 40k x k + 40k x k this took 41427, 39533, 57240
+  /// fyi) out already allocated so no additional initialization happening
+  out.fill(*this, true);
+  // CREATE_IF_EMPTY_DIMS(out, getDim());
+  // scopy(length(), getData(), 1, out.getData(), 1);
+  if (out.add_i(m, alpha) != ML_ERROR_NONE) {
```
zhoonit, line 386: Need opinion which version we should choose based on the benchmark.

jijoongmoon:
Any reason for rearrangement of method or just for readability?

zhoonit:
> Any reason for rearrangement of method or just for readability?

There were some missing outparam methods and it was hard to notify because methods were unorganized.
The first commit handles the issue :)

jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -213,14 +241,18 @@ int Tensor::divide_i(float const &value) {
...
+Tensor Tensor::divide(float const &value) const {
   if (value == 0.0f) {
     throw std::runtime_error("Error: Divide by zero");
   }
 
   CLONE_OP_I(divide_i, value);
 }
 
+Tensor Tensor::divide(float const &value, Tensor &out) const {
+  throw nntrainer::exception::not_supported("NYI");
```
kparichay: can you add this same as `multiply` with the same interface?

nntrainer/tensor/tensor.cpp
```diff
@@ -251,36 +275,86 @@ Tensor Tensor::add(float const &value, Tensor &out) const {
...
+
+Tensor Tensor::pow(float exponent) const {
+  return apply([=](float in) { return powf(in, exponent); });
+}
+
+int Tensor::pow_i(float exponent) {
+  return apply_i([=](float in) { return powf(in, exponent); });
+}
+
+int Tensor::multiply_i(Tensor const &m) {
```
kparichay, line 300: as this PR is doing re-arranging, let club all the forms of multiply together, and same for other operations too.

nntrainer/tensor/tensor.cpp
```diff
@@ -251,36 +275,86 @@ Tensor Tensor::add(float const &value, Tensor &out) const {
...
+Tensor Tensor::subtract(float const &value) const { return this->add(-value); }
 
-  unsigned int buffer_size; /**< virtual size of the buffer */
-  int buffer_axis;          /**< the smallest axis that should be looped.
-                                 -1 means no loop needed*/
-  std::array<unsigned int, MAXDIM>
-    strides; /**< modified strides for the loop */
-};
+Tensor Tensor::subtract(float const &value, Tensor &out) const {
+  throw nntrainer::exception::not_supported("NYI");
```
kparichay: lets add this as well, based on add.

nntrainer/tensor/tensor.cpp
```diff
@@ -292,20 +366,20 @@ Tensor Tensor::add(Tensor const &m, float const alpha) const {
...
 
-Tensor Tensor::pow(float exponent) const {
-  return apply([=](float in) { return powf(in, exponent); });
-}
+Tensor Tensor::subtract(Tensor const &m) const { return add(m, -1); }
 
-int Tensor::pow_i(float exponent) {
-  return apply_i([=](float in) { return powf(in, exponent); });
+Tensor &Tensor::subtract(Tensor const &m, Tensor &out) const {
+  throw nntrainer::exception::not_supported("NYI");
```
kparichay: same

nntrainer/tensor/tensor.cpp
zhoonit: I found that multiply with tensor has more common characteristic with divide with tensor then multiply with scalar hence this arrangement. If we are to add another arithmetic operator (e.g sparse tensor???), I think those could be easily cluster as well.

kparichay: It will be best if the arrangement in the header and source files are synced.
From a code reviewer/reader perspective, who wants to find a function, say multiply, it's difficult to find a required function at different sections of the files compared to the arrangement where all the different interfaces of multiply listed together.

zhoonit: Sounds reasonable, I'll change :)

nntrainer/tensor/tensor.cpp
zhoonit: you mean throw?  it is implemented at following commit :)

nntrainer/tensor/tensor.cpp
zhoonit: Done

nntrainer/tensor/tensor.h
```diff
@@ -271,31 +293,37 @@ class Tensor {
...
-   * @param[out] alpha Values to be scaled
-   * @retval #ML_ERROR_NONE  Successful
-   * @retval #ML_ERROR_INVALID_PARAMETER Invalid Parameter
+   * @brief     Divide value element by element
+   * @param[in] value Divisor
+   * @param[out] out out parameter to store the result
+   * @retval    Calculated Tensor
    */
-  int add_i(Tensor const &m, float const alpha = 1);
+  Tensor divide(float const &value, Tensor &out) const;
```
kparichay: Tensor &

nntrainer/tensor/tensor.cpp
```diff
@@ -214,162 +216,184 @@ int Tensor::multiply_i(float const &value) {
...
 
 int Tensor::multiply_i(Tensor const &m) {
-  auto f = [&](const BroadcastInfo &e, float *buf, const float *m_buf) {
-    for (unsigned int i = 0; i < e.buffer_size; ++i) {
-      *buf *= *m_buf;
-      buf += strides[3];
-      m_buf += e.strides[3];
-    }
-  };
+  try {
```
kparichay, line 217: do we need to have `try catch` here?

kparichay: If this is because of the signature, update the signature to be void, but lets not add `try catch` in critical path.

nntrainer/tensor/tensor.cpp
```diff
@@ -214,162 +216,184 @@ int Tensor::multiply_i(float const &value) {
...
 
 int Tensor::divide_i(Tensor const &m) {
-  auto f = [&](const BroadcastInfo &e, float *buf, const float *m_buf) {
-    for (unsigned int i = 0; i < e.buffer_size; ++i) {
-      *buf /= *m_buf;
-      buf += strides[3];
-      m_buf += e.strides[3];
-    }
-  };
+  try {
```
kparichay, line 272: same

nntrainer/tensor/tensor.cpp
kparichay: Also check this at other places as well.

nntrainer/tensor/tensor.cpp
```diff
@@ -889,10 +872,35 @@ void Tensor::reshape(const TensorDim &d) {
...
     throw std::invalid_argument("Error: reshape cannot change the tensor size");
   }
+
   dim = d;
   strides = d.computeStrides();
 }
 
+void Tensor::fill(const Tensor &from, bool initialize) {
+  if (initialize && this->uninitialized()) {
+    return this->copy(from);
```
kparichay, line 869: return in a function which returns void?

nntrainer/tensor/tensor.cpp
zhoonit: This is not newly added, so all the benchmarks was based on this data path.
As far as I know, overhead is significant only when the error occurs.

I think it is fine because
1. Throwing error is not at the critical path
1. Operation related error is supposed to return integer code in this semantic.

We could check if there is a noticeable difference with `noexcept` or not catching but I think this could be done later.

nntrainer/tensor/tensor.cpp
```diff
@@ -406,6 +378,21 @@ Tensor &Tensor::subtract(Tensor const &m, Tensor &out) const {
   return add(m, out, -1);
 }
 
+int Tensor::pow_i(float exponent) {
```
kparichay, line 381: Please rearrange in headers as well.

zhoonit: This is synced with header. you mean moving to else where?

nntrainer/tensor/tensor.h
zhoonit: Fixed in the following PR :)

nntrainer/tensor/tensor.cpp
zhoonit: Updated

nntrainer/tensor/tensor.cpp
zhoonit: Added issue #887

nntrainer/tensor/tensor.cpp
kparichay: I looked at previous commit.

kparichay (approved):
LGTM

nntrainer/tensor/tensor.cpp
jijoongmoon: don't we need catch the return value for c API?

zhoonit: There is a throw boundary in capi so in capi side it's okay.

Issue #871, opened 2021-01-21
Certain order of layers causes Tensor exception


klmntoth:
Using con2d layer followed by batch_normalization and activation layers cause std::invalid_argument("Creating shared tensor of size bigger than tensor memory.") exception during train.

To reproduce here is a modified ini file for the mnist example: https://gist.github.com/klmntoth/f222c770776b8627494c58979d196548
The same thing happens when you add "activation = sigmoid" to the conv2d_c1_bn layer and delete the conv2d_c1_act one.
But if you swap the batch_normalization and activation layers then everything works.

I traced back the error and it has to do something with this optimization part: https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/graph/network_graph.cpp#L568
when i swap line 569 with 570 all above layer combinations work (no idea if correctly though).

kparichay:
Thanks for reporting this issue. I will check this.
Meanwhile, I recommend you to disable all the optimizations with NNTrainer.
```c
// Create a model
auto nn = nntrainer::NeuralNetwork();                                                                                                                                                                                                                                

// Disable all the optimizations before model initialization
nn.setGradientMemoryOptimization(optimize);                                                                                                                                                                                                                                      
nn.setDerivativeMemoryOptimization(optimize);                                                                                                                                                                                                                                    
nn.setInPlaceLayerOptimization(optimize);                                                                                                                                                                                                                                        
nn.setInferenceInOutMemoryOptimization(optimize);
```

kparichay:
@klmntoth #872 has been added to solve the issue you reported. 
There was a bug in the in-place optimization of layers. In-place optimization of consecutive layers when they are in a trainable state cannot be done and the check to stop that from happening had a bug. #872 adds a quick fix for this. I verified with the new MNIST model file you shared and with #872, the training reaches 81.85% with 100 epochs and 
I tested with a few other model configurations as well and they seem to work too.

Please check on your end and let us know if this works for you.

#872 do not perform memory optimization to the maximum (this only occurs in rare scenarios). A more memory-optimized version will be added soon.

klmntoth:
after #872 our model works without disabling the optimizations. thanks for the quick fix!

PR #872, opened 2021-01-22
[optimization] Bug fix for in-place layer optimization

kparichay:
Inplace layer optimization is performed for multiple layers - activation and batch normalization layers
and this list will increase with data augmentation etc.
However, the in-place layers cannot work correctly consecutively if these layers are trainable.
They can work perfectly is they dont need to pass the derivative back.

For now, this patch limits two consecutive layers to be in-place.
This will be made generic later dependent on the trainable and inPlace property of the layer.

See also #871

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -28,6 +28,13 @@
 
 namespace nntrainer {
 
+/**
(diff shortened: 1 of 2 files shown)


jijoongmoon (approved):
LGTM

zhoonit (approved):
LGTM

Issue #879, opened 2021-01-25
Decouple memory allocation with tensor


kparichay:
Decouple the memory allocation from the tensor allocation itself. This will have to do at multiple levels:
1. Allow tensors to be created without memory allocation, and allocate memory later. - required by #878 
2. Pass allocators to tensors which will allocate memory, handle all the tensors in the manager and ensure all memory is allocated before the final run.
3. Lazy allocate - do not allocate unless that tensor is going to be used - just like lazy page loading in kernel memory management.

kparichay:
1 and 3 are done by #880.
2 will be done as a separate topic later.

PR #876, opened 2021-01-22
[Loss] Fix loss layer were allocating a new mem @open sesame 02/19 18:25

zhoonit:
- [Tensor/Clean] Relocate tensor methods
- [Tensor] rearrange methods
- [Loss] Fix loss layer were allocating a new mem
```
As allocation is managed in `manager`, layer shouldn't be allocate a new
memory that are managed by memory.
However, loss layer was allocating a new memory. This patch fixes the issue.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>
Cc: Parichay Kapoor <pk.kapoor@samsung.com>
```

Issue #867, opened 2021-01-20
Inplace to be a layer property


kparichay:
Each layer must support a property - `inPlace` - which must be set by the layer creator.
This tells the tensor manager if a certain layer **can** operate in place for the backwarding as well as forwarding.
Then there must be a `operatingMode` which when set to `inPlace` will make the layer operate in place (layer must operate out-of-place in debug mode) and memory optimizations can be done.

These are critical for data augmentation layers being added.

kparichay:
Also support `skipOpitmize` property for a layer, which must be checked by the manager before performing various optimizations for it.

zhoonit:
Also support `requireLabel` property for a layer, if this is on, label is somehow fed to the layer.

kparichay:
Also support `isWorkingInPlace` which will be set if the layer is optimized for in-place operation.

kparichay:
- leaving `skipOptimize` for later. For now, manager only performs in-place optimizations for the layer which can be controlled by the layer, and gradient/derivative optimizations can be controlled for the model
- `inWorkingInPlace` is also left to be added later when needed

PR #880, opened 2021-01-25
[tensor/manager] Fix in-place layer optimization bug

### LINKED ISSUE #878: InPlace layer optimization bug
In-place layer optimization has yet another bug.

In general, with derivative optimization, the output tensor and derivative tensor of a layer share the same memory.
With in-place layer optimization, in-place layers do not allocate their own output tensors, but rather use their neighbor's tensors. To handle the derivative memory requirement, in-place layers use a shared memory of derivatives among all the in-place layers.

Current in-place layer optimization deletes the tensors from the previous layers and re-assigns tensors from the current layer (which will behave in-place) to the previous layer. However, this creates an issue that the derivative memory management is different for the current layer and the previous layer. 
For example, with FC layer and activation layer, with the activation layer in-place, they will share the same input/output tensors. But the derivative memory management for the two layers has to be handled separetely.

kparichay:
**Commit 1**: [tensor] Support late memory allocation for tensors
Support late memory allocation for tensors.
This helps create tensor wrapper elements and allocate memory
later when needed.
This allows finalizing the graph, creating all the tensors and graph
connections for all the layers which can be stored in offline setup
and then loaded directly from file.

This decoupling allows reusing tensors with every call to train
where we can allocate and deallocate memory without having to create
and do tensor management again and again.

In short term, this is necessary for in-place layer optimization.
See Also #879

The implementation requires caching the source tensor when creating
shared tensors as shared tensors cannot be created unless the source
tensor memory has been allocated.

**Commit 2**: In-place bug fix 
This patch applied bug fix for in-place layer optimization.
As in-place layers and other layers manages derivative memory
differently, in-place layers cannot directly re-use the tensors
of their neighboring layers. But rather have to use tensors differently.
This patch applies this bug-fix.
See also #878

**Commit 3:** [Manager] Support lazy memory allocation with manager
Support lazy memory allocation with manager

**Commit 4:** [bug fixes] Update shared mem and in-place opt
Update shared memory and in-place opt to share memory with previous
layers with different strategies based on the type of optimization the
layer is doing for reducing its memory usage.

Added weight and var_grad to create weight/var_grad with lazy
memory allocation.
Update SrcSharedTensor to remove the check is source is allocated
as it cannot be guaranteed.
Bug fix for tensor is allocated

Update nntrainer extension for nnstreamer to use a different
ini file for inference which does not require changing batch size.
Added minor bug fixes to the patch

**Commit 5:** [tensor] Split tensor constructor into 2    
Create a new tensor constructor for lazy allocation
    than adding the lazy allocation feature to the existing one

Resolves #878
See also #876

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -11,6 +11,8 @@
  *
  */
 
+#include <sstream>
+
 #include <activation_layer.h>
 #include <addition_layer.h>
 #include <bn_layer.h>
@@ -576,22 +578,49 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
         continue;
 
       /** Share tensor with next layer */
-      l->net_input[0] = l->net_hidden[0]; // I2 = O2
+      /**
+       * Assume two layers, L1 and L2, with I and O corresponding to the layer
+       * outputs. Assume L2 to be the layer, needing in-place optimization.
+       */
       if (l->getType() == BatchNormalizationLayer::type) {
-        prev_layer->net_hidden[loc] = l->net_hidden[0]; // O1 = O2
+        /**
+         * With batch normalization, neither input nor output of the layer are
+         * requried for calculatin gradient and derivative. Just input
+         * derivative is required. In this scenraio, L2 is assumed to be batch
+         * normaliztion layer, and L1 is assumed to be a non-in-place layer.
+         * Hence, L1 layer's output and L2 layer's input var_grad are modified.
+         */
+        auto &inplace_shared_vg_ptr = l->net_hidden[0];
+        l->net_input[0] = inplace_shared_vg_ptr;             /// I2 = O2
+        prev_layer->net_hidden[loc] = inplace_shared_vg_ptr; /// O1 = O2
       } else if (l->getType() == ActivationLayer::type) {
-        // THIS IS NOT WORKING BECAUSE INITIALIZES MAKES NEW SHARED_PTR OF TENSOR
-        // THAN KEEPING THE SAME SHARED_PTR AND JUST CHANGING THE TENSOR IN IT
-        prev_layer->net_hidden[loc]->getGradientRef() =
-          l->net_hidden[0]->getVariableRef(); // O1.G = O2.V
-        prev_layer->net_hidden[loc]->getVariableRef() =
-          l->net_hidden[0]->getVariableRef(); // O1.V = O2.V
+        /**
+         * For activation layer, output of the layer and input derivative, both
+         * , are requried for calculating the gradient and derivative. In this
+         * scenraio, L2 is assumed to be activation layer, and L1 is assumed to
+         * be a non-in-place layer.
+         * Hence, L1 layer's output and L2 layer's input var_grad are
+         * differently. L1 layer operates out of place and share the memory for
+         * its output (O1.V) and input derivative (O1.G). L2 layer operates
+         * in-place and use a common shared derivative memory for their
+         * derivatives (handled in manager.cpp).
+         */
+        /**
+         * @note As this updates the tensors used in the prev_layer->net_hidden
+         * (O1), the tensors inside l->net_input (I2) are automatically updated
+         * as they share the same var_grad object.
+         */
+        auto &inplace_shared_vg = *l->net_hidden[0].get();
+        prev_layer->net_hidden[loc]->updateVariableByVariable(
+          inplace_shared_vg); /// O1.G = O2.V
+        prev_layer->net_hidden[loc]->updateGradientByVariable(
+          inplace_shared_vg); /// O1.V = O2.V
+      } else {
+        std::stringstream ss;
+        ss << l->getType();
+        ss << " layer is not supported for in-place optimization";
+        throw std::runtime_error(ss.str());
       }
-
-      /** prev layer should use this shared output for derivative though */
-      // prev_layer->net_hidden[loc] =
-      //   std::shared_ptr<Var_Grad>(*prev_layer->net_hidden[loc].get());
-      // prev_layer->net_hidden[loc].get
 
       /** Untrack the memory for this layer */
       manager.untrackLayerInOuts(prev_layer->getName());
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -576,8 +576,22 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
         continue;
 
       /** Share tensor with next layer */
-      l->net_input[0] = l->net_hidden[0];
-      prev_layer->net_hidden[loc] = l->net_hidden[0];
+      l->net_input[0] = l->net_hidden[0]; // I2 = O2
+      if (l->getType() == BatchNormalizationLayer::type) {
+        prev_layer->net_hidden[loc] = l->net_hidden[0]; // O1 = O2
+      } else if (l->getType() == ActivationLayer::type) {
+        // THIS IS NOT WORKING BECAUSE INITIALIZES MAKES NEW SHARED_PTR OF TENSOR
+        // THAN KEEPING THE SAME SHARED_PTR AND JUST CHANGING THE TENSOR IN IT
+        prev_layer->net_hidden[loc]->getGradientRef() =
+          l->net_hidden[0]->getVariableRef(); // O1.G = O2.V
+        prev_layer->net_hidden[loc]->getVariableRef() =
+          l->net_hidden[0]->getVariableRef(); // O1.V = O2.V
+      }
+
+      /** prev layer should use this shared output for derivative though */
+      // prev_layer->net_hidden[loc] =
+      //   std::shared_ptr<Var_Grad>(*prev_layer->net_hidden[loc].get());
+      // prev_layer->net_hidden[loc].get
 
       /** Untrack the memory for this layer */
       manager.untrackLayerInOuts(prev_layer->getName());
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -68,17 +68,14 @@ int Conv2DLayer::initialize(Manager &manager) {
     ml_logw("Warning: the length of previous layer dimension is one");
   }
 
-  std::string kernelPrefix = "Conv2d:filter";
-  std::string biasPrefix = "Conv2d:bias";
-
   TensorDim dim =
     TensorDim(filter_size, in_dim.channel(), kernel_size[0], kernel_size[1]);
   TensorDim bias_dim = TensorDim(1, filter_size, 1, 1);
 
   if (weights.empty()) {
     weights.reserve(2);
-    weights.emplace_back(dim, weight_initializer, true, false, kernelPrefix);
-    weights.emplace_back(bias_dim, bias_initializer, true, false, biasPrefix);
+    weights.emplace_back(dim, weight_initializer, true, "Conv2d:filter");
+    weights.emplace_back(bias_dim, bias_initializer, true, "Conv2d:bias");
     manager.trackWeights(weights);
   } else {
     weights[ConvParams::weight].reset(dim, weight_initializer, true);
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -77,8 +77,8 @@ int Conv2DLayer::initialize(Manager &manager) {
 
   if (weights.empty()) {
     weights.reserve(2);
-    weights.emplace_back(dim, weight_initializer, true, kernelPrefix);
-    weights.emplace_back(bias_dim, bias_initializer, true, biasPrefix);
+    weights.emplace_back(dim, weight_initializer, true, false, kernelPrefix);
+    weights.emplace_back(bias_dim, bias_initializer, true, false, biasPrefix);
     manager.trackWeights(weights);
   } else {
     weights[ConvParams::weight].reset(dim, weight_initializer, true);
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -189,7 +189,8 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
     /// this creates memory and sets to @a memory and returns AllocFunc
     auto get_allocfunc = [](const unsigned int weight_size,
                             std::unique_ptr<MMapedMemory> &memory) {
-      if (weight_size >= std::numeric_limits<unsigned int>::max() / sizeof(float)) {
+      if (weight_size >=
+          std::numeric_limits<unsigned int>::max() / sizeof(float)) {
         throw std::invalid_argument(
           "weights exceed maximum size supported for shared memory");
       }
@@ -213,7 +214,7 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   } else if (!is_weight) {
     /** only for gradients */
     if (max_grad_size > 0 && enable_gradient_memory_opt) {
-      shared_grad = Tensor(TensorDim({(unsigned int)max_grad_size}), nullptr, false);
+      shared_grad = Tensor({(unsigned int)max_grad_size}, nullptr, false);
 
       allocate_func = [this](const TensorDim &dim, unsigned int offset) {
         return shared_grad.getSharedDataTensor(dim, offset);
@@ -436,7 +437,7 @@ void Manager::initializeTensors(bool trainable) {
 
   // Allocate shared derivative memory
   if (max_derivative_size > 0 && enable_activation_memory_opt && trainable)
-    shared_deriv = Tensor(TensorDim({max_derivative_size}), nullptr, false);
+    shared_deriv = Tensor({max_derivative_size}, nullptr, false);
 
   // @todo Do not count memory of the input tensor of the input layer in the
   // estimate of max_shared_inout as it is not used
@@ -444,7 +445,7 @@ void Manager::initializeTensors(bool trainable) {
   // Allocate shared input/output memory for inference
   // @note Memory for label is not allocated here as inference doesnt has label
   if (!trainable && enable_inference_inout_memory_opt)
-    shared_inout = Tensor(TensorDim({max_shared_inout}), nullptr, false);
+    shared_inout = Tensor({max_shared_inout}, nullptr, false);
 
   /**
    * A single buffer (shared_inout) provides memory for inputs and outputs of a
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -246,7 +246,7 @@ void Manager::initializeWeights() {
       Tensor grad_prealloc = Tensor();
 
       weight_offset += dim.getDataLen();
-      weight.initializeWeight(weight_prealloc);
+      weight.initializeVariable(weight_prealloc);
     }
   }
 
@@ -300,7 +300,7 @@ void Manager::initializeGradients() {
         grad_prealloc = allocate_grad(dim, grad_offset);
         grad_offset += dim.getDataLen();
       }
-      weight.initializeGrad(grad_prealloc, true);
+      weight.initializeGradient(grad_prealloc, true);
     }
   }
 }
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -214,7 +214,7 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   } else if (!is_weight) {
     /** only for gradients */
     if (max_grad_size > 0 && enable_gradient_memory_opt) {
-      shared_grad = Tensor({(unsigned int)max_grad_size}, nullptr, false);
+      shared_grad = Tensor(TensorDim({max_grad_size}), false);
 
       allocate_func = [this](const TensorDim &dim, unsigned int offset) {
         return shared_grad.getSharedDataTensor(dim, offset);
@@ -437,7 +437,7 @@ void Manager::initializeTensors(bool trainable) {
 
   // Allocate shared derivative memory
   if (max_derivative_size > 0 && enable_activation_memory_opt && trainable)
-    shared_deriv = Tensor({max_derivative_size}, nullptr, false);
+    shared_deriv = Tensor(TensorDim({max_derivative_size}), false);
 
   // @todo Do not count memory of the input tensor of the input layer in the
   // estimate of max_shared_inout as it is not used
@@ -445,7 +445,7 @@ void Manager::initializeTensors(bool trainable) {
   // Allocate shared input/output memory for inference
   // @note Memory for label is not allocated here as inference doesnt has label
   if (!trainable && enable_inference_inout_memory_opt)
-    shared_inout = Tensor({max_shared_inout}, nullptr, false);
+    shared_inout = Tensor(TensorDim({max_shared_inout}), false);
 
   /**
    * A single buffer (shared_inout) provides memory for inputs and outputs of a
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -157,12 +157,12 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
   std::vector<std::reference_wrapper<Weight>> layer_weights;
   layer_weights.reserve(ws.size());
 
-  size_t weight_size = 0;
-  size_t grad_size = 0;
+  unsigned int weight_size = 0;
+  unsigned int grad_size = 0;
 
   for (auto &w : ws) {
     layer_weights.emplace_back(std::ref(w));
-    size_t len = w.getDim().getDataLen();
+    unsigned int len = w.getDim().getDataLen();
     weight_size += len;
     if (w.getTrainable())
       grad_size += len;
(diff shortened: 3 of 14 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -80,18 +80,80 @@ static auto rng = [] {
...
+ */
+class SrcSharedTensor {
+public:
+  /**
+   * @brief   Constructor for the class
+   */
+  SrcSharedTensor(): src(nullptr), off(0) {}
+
+  SrcSharedTensor(const Tensor *tensor, unsigned int offset): src(tensor), off(offset) {}
+
```
jijoongmoon, line 116: If we get the tensor obj from SrcSharedTenor with offset.. I think we need to compare the last buffer from offset to prevent the overflow.

nntrainer/tensor/tensor.cpp
```diff
@@ -80,18 +80,80 @@ static auto rng = [] {
...
+};
+
+void Tensor::allocate() {
+  if (data)
+    /// already allocated
+    return;
+
+  if (src_tensor) {
+    /// allocate data based on the source tensor
+    data = std::shared_ptr<float>(src_tensor->tensor()->data, src_tensor->tensor()->data.get() + src_tensor->offset());
```
jijoongmoon: If we get the tensor obj from SrcSharedTenor with offset.. I think we need to compare the last buffer from offset to prevent the overflow.

nntrainer/tensor/tensor.cpp
```diff
@@ -336,9 +398,17 @@ Tensor Tensor::getSharedDataTensor(const TensorDim dim_,
...
-  ret.data = std::shared_ptr<float>(this->data, this->data.get() + offset);
   ret.strides = ret.dim.computeStrides();
 
+  /**
+   * In this case, its the caller's responsibility to ensure that allocate() is
+   * called for this function before operating on this tensor.
+   */
+  if (this->data)
+    ret.data = std::shared_ptr<float>(this->data, this->data.get() + offset);
+  else
```
jijoongmoon: Is it possible to get the data from this->data.get()+offset1 to offset2?

nntrainer/tensor/tensor.cpp
```diff
@@ -80,18 +80,80 @@ static auto rng = [] {
...
       copy(buf);
   }
 }
 
+/**
+ * @class SrcSharedTensor
+ * @brief Source of the shared tensor
+ */
+class SrcSharedTensor {
+public:
```
jijoongmoon, line 107: I wonder.. does the tensor has this SrcSharedTensor always? How about just set the data with data pointer from shared tensor ?

nntrainer/tensor/tensor.cpp
kparichay: This is updated in the last commit of this PR, where the src_tensor is always the original tensor which will allocate the memory, and offset is set based on the original source tensor.

nntrainer/tensor/tensor.cpp
kparichay: I don't understand the reference to offset1 and offset2 here. Can you please elaborate?

nntrainer/tensor/tensor.cpp
kparichay: > How about just set the data with data pointer from shared tensor ?

That is one approach. However, with this, there need to be two extra variables - offset from srcSharedTensor, and a variable to indicate if this actual data or srcSharedTensor, and creates Tensor object more complicated and heavy.
With SrcSharedTensor, the Tensor class size remains cleaner with smaller increase in object size.

nntrainer/tensor/tensor.cpp
jijoongmoon: What I means.. making shared tensor from in the middle of memory chunck.

nntrainer/graph/network_graph.cpp
```diff
@@ -576,8 +576,22 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
...
+      if (l->getType() == BatchNormalizationLayer::type) {
+        prev_layer->net_hidden[loc] = l->net_hidden[0]; // O1 = O2
+      } else if (l->getType() == ActivationLayer::type) {
+        // THIS IS NOT WORKING BECAUSE INITIALIZES MAKES NEW SHARED_PTR OF TENSOR
+        // THAN KEEPING THE SAME SHARED_PTR AND JUST CHANGING THE TENSOR IN IT
+        prev_layer->net_hidden[loc]->getGradientRef() =
+          l->net_hidden[0]->getVariableRef(); // O1.G = O2.V
+        prev_layer->net_hidden[loc]->getVariableRef() =
+          l->net_hidden[0]->getVariableRef(); // O1.V = O2.V
+      }
```
jijoongmoon, line 608: is it right to point grad and var to 02.v?

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -77,8 +77,8 @@ int Conv2DLayer::initialize(Manager &manager) {
 
   if (weights.empty()) {
     weights.reserve(2);
-    weights.emplace_back(dim, weight_initializer, true, kernelPrefix);
-    weights.emplace_back(bias_dim, bias_initializer, true, biasPrefix);
+    weights.emplace_back(dim, weight_initializer, true, false, kernelPrefix);
+    weights.emplace_back(bias_dim, bias_initializer, true, false, biasPrefix);
```
jijoongmoon: So.. when we initialize layer, there is no memory allocation?

nntrainer/tensor/tensor.cpp
```diff
@@ -327,6 +387,17 @@ Tensor Tensor::getBatchSlice(unsigned int offset, unsigned int size) const {
...
 
+void Tensor::createSharedDataTensor(const Tensor &src, Tensor &dest,
+                                    unsigned int offset) {
+  if (src.data)
+    dest.data = std::shared_ptr<float>(src.data, src.data.get() + offset);
+  else if (!src.src_tensor)
+    dest.src_tensor = std::make_shared<SrcSharedTensor>(&src, offset);
+  else
+    dest.src_tensor = std::make_shared<SrcSharedTensor>(
+      src.src_tensor->tensor(), offset + src.src_tensor->offset());
```
jijoongmoon, line 398: So.. if there is src.src_tensor, then we are going to use memory from src.offset() to offset of dest? If it is, then it is not possible to share from src.data?

nntrainer/tensor/tensor.cpp
```diff
@@ -80,18 +80,80 @@ static auto rng = [] {
   return rng;
 }();
 
-Tensor::Tensor(const TensorDim &d, const float *buf) : Tensor() {
+Tensor::Tensor(const TensorDim &d, const float *buf, bool alloc_now) : Tensor() {
```
zhoonit: Q. giving `buf` means `alloc_now` always true?

nntrainer/tensor/tensor.cpp
```diff
@@ -80,18 +80,80 @@ static auto rng = [] {
...
+      throw std::runtime_error("Accessing empty src tensor");
+
+    /**
+     * Allocate source tensor if not already allocated.
+     */
+    if (!src->isAllocated())
+      throw std::runtime_error("Source tensor must be allocated before allocating dependent tensors");
+      // src->allocate();
+
+    return src;
```
zhoonit, line 124: return src + offset?

nntrainer/tensor/tensor.cpp
kparichay: To remedy this, in the last commit, if the source tensor (say `t_s1`) provided depends on another source tensor (say `t_s2`), then the tensor to be made in this situation is made to directly depend on `t_s2`.

nntrainer/graph/network_graph.cpp
kparichay: Yes, this is intended. The derivative optimization of different layers happens differently.
For layers working in-place, like activation, these layers share derivative from common shared memory.
However, for layers that are not in-place, these layers use their output memory as the derivative memory.

nntrainer/layers/conv2d_layer.cpp
kparichay: Right.

nntrainer/tensor/tensor.h
```diff
@@ -841,8 +863,17 @@ class Tensor {
...
+
   std::shared_ptr<float> data;
 
+  /**<
+   * When using shared_data with tensor, this stores the ptr of the source
+   * tensor which handles the full memory. If tensor data is already allocated,
+   * this does not affect the tensor. If the tensor data is not allocated, and
+   * src_ptr is valid, this tensor will use the memory allocated by the src_ptr
+   */
+  std::shared_ptr<SrcSharedTensor> src_tensor;
```
zhoonit, line 888: I think, in the long term, we should have storage object or equivalent to manage the memory, tensor often used as a wrapper to a memory. Thus making our intentions more clear by using storage if we are to manage memory.

nntrainer/tensor/tensor.cpp
zhoonit: how about hiding this logic into `src_tensor->allocate()`

nntrainer/graph/network_graph.cpp
```diff
@@ -576,8 +576,22 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
         continue;
 
       /** Share tensor with next layer */
-      l->net_input[0] = l->net_hidden[0];
-      prev_layer->net_hidden[loc] = l->net_hidden[0];
+      l->net_input[0] = l->net_hidden[0]; // I2 = O2
```
zhoonit: Q. what is I2 O2 O1.G O2.V??

nntrainer/tensor/tensor.cpp
kparichay: The function first check If `src.data` exists (meaning tensor already has memory allocated). If it is, then we directly make this tensor based on the src tensor.
If `src.data` does not exist (meaning tensor does not memory allocated), and `src.src_tensor` does not exist (meaning the src tensor does not exist on other tensor), then create a SrcTensor around the src.
If `src.src_tensor` exists, then use the `src.src_tensor` to create the required `src_tensor` to avoid recursive dependency.
Some more explanation in https://github.com/nnstreamer/nntrainer/pull/880#discussion_r567550639.

> So.. if there is src.src_tensor, then we are going to use memory from src.offset() to offset of dest? If it is, then it is not possible to share from src.data?

The function first tries to share from `src.data`. If src.data is not yet allocated, then it goes to src_tensor.

nntrainer/tensor/tensor.h
```diff
@@ -57,12 +58,17 @@ class Tensor {
...
 
   /**
    * @brief     Constructor of Tensor with dimension/buf
+   * @param d Tensor dim for this tensor
+   * @param bug buffer
+   * @alloc_now If the memory of the tensor must be allocated
+   * @throws if buf is not null and allow_now is false
    */
-  Tensor(const TensorDim &d, const float *buf = nullptr);
+  Tensor(const TensorDim &d, const float *buf = nullptr, bool alloc_now = true);
```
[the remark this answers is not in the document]
kparichay: Tensor(TensorDim, buf = nullptr, alloc_now) would also be needed, where we can store `buf` in the `SrcSharedTensor`?

nntrainer/graph/network_graph.cpp
kparichay: I will add details in comments.

nntrainer/graph/network_graph.cpp
```diff
@@ -576,8 +576,22 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
...
+      if (l->getType() == BatchNormalizationLayer::type) {
```
zhoonit, line 581: caution for this if,
we will going to need `is_batchnorm_type`, `is_activation_type` kind of property if this cannot be unified.

nntrainer/graph/network_graph.cpp
zhoonit: Input to current (I2)
Output to current (O2)

nntrainer/graph/network_graph.cpp
zhoonit: We could have strategy instead I guess?

kparichay: The strategy is to support `supportInPlace()` in the layer.

kparichay: #867

nntrainer/graph/network_graph.cpp
```diff
@@ -576,8 +576,22 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
...
+        prev_layer->net_hidden[loc] = l->net_hidden[0]; // O1 = O2
+      } else if (l->getType() == ActivationLayer::type) {
+        // THIS IS NOT WORKING BECAUSE INITIALIZES MAKES NEW SHARED_PTR OF TENSOR
+        // THAN KEEPING THE SAME SHARED_PTR AND JUST CHANGING THE TENSOR IN IT
+        prev_layer->net_hidden[loc]->getGradientRef() =
+          l->net_hidden[0]->getVariableRef(); // O1.G = O2.V
```
zhoonit: How about having a designated memory(for safety) or at least alias for readability?

nntrainer/graph/network_graph.cpp
```diff
@@ -576,8 +576,22 @@ void NetworkGraph::inPlaceOptimize(const std::string &layer_type,
...
+        prev_layer->net_hidden[loc]->getVariableRef() =
+          l->net_hidden[0]->getVariableRef(); // O1.V = O2.V
+      }
+
+      /** prev layer should use this shared output for derivative though */
+      // prev_layer->net_hidden[loc] =
```
zhoonit: comment intended?

nntrainer/tensor/tensor.h
kparichay: Yes, the storage object would wrap both `data` and `src_tensor`.
It will be done. For now, this PR has been rushed to solved the bug ASAP.

nntrainer/layers/conv2d_layer.cpp
zhoonit: Not for the PR, but there is an idea. 

manager calls layer::get_weight() before layer::initialize() and it tracks all the weights. we can extract this to a manager (idea for api change)

nntrainer/tensor/manager.cpp
```diff
@@ -177,23 +177,23 @@ void Manager::trackWeights(std::vector<Weight> &ws) {
 }
 
 Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
-  AllocFunc allocate_none = [](const TensorDim &dim, size_t) {
+  AllocFunc allocate_none = [](const TensorDim &dim, unsigned int) {
```
zhoonit, line 180: any reason for the change?

nntrainer/tensor/manager.cpp
```diff
@@ -206,16 +206,16 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
...
-      size_t grad_size =
+      unsigned int grad_size =
         enable_gradient_memory_opt ? max_grad_size : total_grad_size;
       allocate_func = get_allocfunc(grad_size, grad_mmaped_memory);
     }
   } else if (!is_weight) {
     /** only for gradients */
     if (max_grad_size > 0 && enable_gradient_memory_opt) {
-      Tensor shared_grad = Tensor(TensorDim({(unsigned int)max_grad_size}));
+      shared_grad = Tensor(TensorDim({(unsigned int)max_grad_size}), nullptr, false);
```
zhoonit:
```suggestion
      shared_grad = Tensor({(unsigned int)max_grad_size}, nullptr, false);
```

nntrainer/tensor/var_grad.cpp
```diff
@@ -16,19 +16,18 @@
 
 namespace nntrainer {
 
-Var_Grad::Var_Grad(const TensorDim &dim, bool train, const std::string &name) :
+Var_Grad::Var_Grad(const TensorDim &dim, bool train, bool alloc_now_, const std::string &name) :
```
zhoonit: What's the difference from just having tensor dim and not allocated tensor?

nntrainer/tensor/tensor.cpp
kparichay: `allocate` is the only function which allocates/sets `data` in tensor.
I didn't want to split it.

nntrainer/graph/network_graph.cpp
kparichay: Added comments.

nntrainer/graph/network_graph.cpp
kparichay: Removed in a later commit.

nntrainer/tensor/manager.cpp
kparichay: Tensor uses unsigned int, and it causes `unsigned int` vs `long unsigned int` comparison warnings at a lot of places.

nntrainer/tensor/var_grad.cpp
kparichay: With alloc_now_ == true, when Var_Grad::initialize() is called, then the tensors memory is allocated.
With alloc_now_ == true, Var_Grad::initialize() will not allocated memory.

`NetworkGraph::inPlaceOptimization` takes place after Tensor in Var_Grad objects have been allocated and initialization has been done. But if the memory is already allocated with initialization, these later optimizations do not reduce the peak memory requirements.

Plus, by separating initialize() and allocate(), all the optimizations to decide how tensors across the layers interface (input and output connectivity between layers) is not required to done every time.

nntrainer/tensor/tensor.h
zhoonit: Yup

nntrainer/tensor/tensor.cpp
zhoonit: Sure

nntrainer/tensor/manager.cpp
zhoonit: As far as I remember, size_t can effectively check max overflow but unsigned int is not guaranteed as max supported size is different. I guess not a big deal for now

nntrainer/layers/conv2d_layer.cpp
kparichay: Sounds good, but this would have to be an added interface and not a replacement for the current method.
With just that interface, a layer, not part of the model would not be able to register weights with manager.

nntrainer/graph/network_graph.cpp
kparichay: designated memory(for safety) -> you mean a designated variable?

nntrainer/tensor/manager.cpp
kparichay: done.

kparichay:
Please note that there will be a wiki page soon on NNTrainer explaining all the memory optimizations.

nntrainer/layers/conv2d_layer.cpp
zhoonit: If we make sure neuralnetwork calls sth like `manager.trackWeights(layer.getWeights)`. For API point of view, layer does not need to know about the manager.

nntrainer/tensor/tensor.h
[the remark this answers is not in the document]
zhoonit: Current (before this patch) semantics, `Tensor(TensorDim)` does the work.
So `Tensor(TensorDim, bool allocate_now)` covers all the new semantics

else, I think we could consider tag dispatching.

nntrainer/graph/network_graph.cpp
zhoonit: yup. sth like `Tensor inPlaceSharedSlot`

or `Tensor & in_place_shared_slot = l->net_hidden[0]->getVariableRef()` for readability

nntrainer/layers/conv2d_layer.cpp
kparichay: As the API exposes layer functionality, a user cannot then create just a layer (without wrapping it in model), without exposing the manager to the API (which IMO should not be exposed).

nntrainer/tensor/manager.cpp
kparichay: Yes, size_t is better, but it requires repo wide updates, which is beyond the scope of this PR.

nntrainer/tensor/var_grad.cpp
zhoonit: Yes. I mean `Tensor::Tensor(Tensor&dim, nullptr, false)` vs `TensorDim` instead for non-allocated tensor. Because `allocation` happens always later, In this case `initialize` does not need to deal with `tensor(dim, null, false)` but just `tensordim` or `weights` that wraps around the `tensordim`.

nntrainer/tensor/tensor.h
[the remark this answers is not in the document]
kparichay: With just `Tensor(TensorDim, bool allocate_now)`, how do you lazy allocate a tensor but with a provided `buf`?

nntrainer/tensor/var_grad.cpp
kparichay: `Tensor::Tensor(Tensor&dim, nullptr, false)` vs `Tensor::Tensor(TensorDim)` are different as `alloc_now_` defaults to true.

nntrainer/tensor/tensor.h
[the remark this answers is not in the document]
zhoonit: if buf is provided, allocate_now should be always true. So there seems no such case.

```
@throws if buf is not null and allow_now is false
```

nntrainer/graph/network_graph.cpp
kparichay: updated.

nntrainer/tensor/var_grad.cpp
zhoonit: Yes, I mean between `Tensor::Tensor(Tensor&dim, nullptr, false)` vs `TensorDim`. And they could share weights (a wrapper of TensorDim in this stage) if needed.
Let's discuss this later as this would be slightly out of boundary of the PR

nntrainer/tensor/tensor.h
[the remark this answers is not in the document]
kparichay: For this PR. Later, buf will be cached in SrcSharedTensor.
I can do it in this PR itself.

nntrainer/tensor/var_grad.cpp
kparichay: I get what you mean. Semantically, they contain the same information. `Tensor::Tensor(Tensor&dim, nullptr, false)` is just a wrapper over `TensorDim` which will later contain memory.

nntrainer/tensor/tensor.cpp
jijoongmoon: Could you add some comments when they are used? and if there is the case which src.data and src.src_tensor is true?

nntrainer/tensor/var_grad.cpp
```diff
@@ -16,19 +16,22 @@
...
+  else
+    grad = std::make_shared<Tensor>();
 }
 
 void Var_Grad::initializeWeight(const Tensor &preallocated) {
   if (!preallocated.uninitialized()) {
-    var = std::make_shared<Tensor>(preallocated);
-  } else {
-    var = std::make_shared<Tensor>(dim);
+    var->makeSharedDataTensor(preallocated);
```
jijoongmoon, line 34: s/initializeWeight/initializeVariables?

nntrainer/tensor/tensor.h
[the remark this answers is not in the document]
kparichay: Updated.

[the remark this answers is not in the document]
zhoonit: Awesome

test/test_models/models/mnist_inf.ini
```diff
@@ -0,0 +1,11 @@
...
+[Model]
+Type = NeuralNetwork	# Network Type : Regression, KNN, NeuralNetwork
+Optimizer = adam 	# adam (Adamtive Moment Estimation)
+Loss = cross  		# Loss function : mse (mean squared error)
+Save_Path = "../test_models/models/model.bin"  	# model path to save / read
+batch_size = 1		# batch size
+
+[mnist]
+backbone = "../test_models/models/mnist.ini"
+Input_Shape = 1:28:28
```
jijoongmoon, line 11: It's how we use backbone!

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -1911,6 +1915,7 @@ TEST(nntrainer_LossLayer, forward_loss_unknown_n) {
     layer.getType(), layer.getName(), layer.getOutputDimension()));
 
   manager.initializeTensors(true);
+  manager.allocateTensors();
```
jijoongmoon, line 1918: We need this from now on.

test/test_models/models/mnist_inf.ini
kparichay: yes :)

test/unittest/unittest_nntrainer_layers.cpp
kparichay: yes, when using without model.
model.inference() and model.train() call this internally.

nntrainer/tensor/var_grad.cpp
zhoonit: Yes and I think var_grad can do this. 
This could be discussed later when we need to reduce the complexity of tensor I guess.

zhoonit (approved):
LGTM :)

nntrainer/tensor/var_grad.cpp
kparichay: done as a new commit in this PR.

nntrainer/tensor/tensor.cpp
kparichay: done

nntrainer/tensor/var_grad.cpp
kparichay: Yes, to reduce the overhead of tensor object.

jijoongmoon (approved):
LGTM

PR #883, opened 2021-01-26
[CAPI/acr] Update tizen capi header

zhoonit:
- Setting void *user_data
- Update layer type enum

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -33,17 +33,18 @@ namespace train {
 enum LayerType {
   LAYER_IN = ML_TRAIN_LAYER_TYPE_INPUT, /** Input Layer type */
   LAYER_FC = ML_TRAIN_LAYER_TYPE_FC,    /** Fully Connected Layer type */
-  LAYER_BN,                             /** Batch Normalization Layer type */
-  LAYER_CONV2D,                         /** Convolution 2D Layer type */
-  LAYER_POOLING2D,                      /** Pooling 2D Layer type */
-  LAYER_FLATTEN,                        /** Flatten Layer type */
-  LAYER_ACTIVATION,                     /** Activation Layer type */
-  LAYER_ADDITION,                       /** Addition Layer type */
-  LAYER_CONCAT,                         /** Concat Layer type */
-  LAYER_OUT,                            /** Multi Output Layer type */
-  LAYER_LOSS,                           /** Loss Layer type */
-  LAYER_BACKBONE_NNSTREAMER,            /** Backbone using NNStreamer */
-  LAYER_BACKBONE_TFLITE,                /** Backbone using TFLite */
+  LAYER_BN = ML_TRAIN_LAYER_TYPE_BN,    /** Batch Normalization Layer type */
+  LAYER_CONV2D = ML_TRAIN_LAYER_TYPE_CONV2D, /** Convolution 2D Layer type */
+  LAYER_POOLING2D = ML_TRAIN_LAYER_TYPE_POOLING2D, /** Pooling 2D Layer type */
+  LAYER_FLATTEN = ML_TRAIN_LAYER_TYPE_FLATTEN,     /** Flatten Layer type */
+  LAYER_ACTIVATION =
+    ML_TRAIN_LAYER_TYPE_ACTIVATION,              /** Activation Layer type */
+  LAYER_ADDITION = ML_TRAIN_LAYER_TYPE_ADDITION, /** Addition Layer type */
+  LAYER_CONCAT = ML_TRAIN_LAYER_TYPE_CONCAT,     /** Concat Layer type */
+  LAYER_MULTIOUT = ML_TRAIN_LAYER_TYPE_MULTIOUT, /** Multi Output Layer type */
+  LAYER_LOSS,                                    /** Loss Layer type */
+  LAYER_BACKBONE_NNSTREAMER,                  /** Backbone using NNStreamer */
+  LAYER_BACKBONE_TFLITE,                      /** Backbone using TFLite */
   LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN /** Unknown */
 };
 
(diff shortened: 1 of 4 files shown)


api/nntrainer-api-common.h
```diff
@@ -27,8 +27,15 @@ extern "C" {
...
-  ML_TRAIN_LAYER_TYPE_FC,           /**< Fully Connected Layer */
+  ML_TRAIN_LAYER_TYPE_INPUT = 0,  /**< Input Layer */
+  ML_TRAIN_LAYER_TYPE_FC,         /**< Fully Connected Layer */
+  ML_TRAIN_LAYER_TYPE_BN,         /**< Batch Normalization Layer (since 6.5) */
+  ML_TRAIN_LAYER_TYPE_CONV2D,     /**< Convolution 2D Layer (since 6.5) */
+  ML_TRAIN_LAYER_TYPE_POOLING2D,  /**< Pooling 2D Layer (since 6.5) */
+  ML_TRAIN_LAYER_TYPE_FLATTEN,    /**< Flatten Layer (since 6.5) */
+  ML_TRAIN_LAYER_TYPE_ACTIVATION, /**< Activation Layer (since 6.5) */
+  ML_TRAIN_LAYER_TYPE_ADDITION,   /**< Addition Layer (since 6.5) */
+  ML_TRAIN_LAYER_TYPE_CONCAT,     /**< Concat Layer (since 6.5) */
```
kparichay, line 38: Please add nnstreamer and tflite backbone layer as well.

api/capi/include/nntrainer.h
```diff
@@ -357,6 +357,34 @@ int ml_train_dataset_create_with_generator(ml_train_dataset_h *dataset,
...
+ * @param[in] valid_user_data user data to be passed when @a valid_cb is being
+ * called. Can be null.
+ * @param[in] test_user_data user data to be passed when @a test_cb is being
+ * called. Can be null.
+ * @return @c 0 on successs. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successsful
+ * @retval #ML_ERROR_INVALID_PARAMETER if @a dataset is not created from @a
+ * ml_train_dataset_create_with_generator
+ */
+int ml_train_dataset_generator_set_user_data(ml_train_dataset_h *dataset,
```
kparichay, line 384: If you are adding this, please revert the changes from https://github.com/nnstreamer/nntrainer/pull/533

zhoonit: Agh.. I forgot we had that functionality already.

I think #533 style is better, in this case, we just have to change the document.

I wonder if we can split those into train_user_data, valid_user_data, ... as far as I remember there was a problem with this.
"ml_train_dataset_set_property(dataset, "train_user_data", (void *) data, " NULL)"

kparichay: Yes, we have to support 3 user data corresponding to each train, valid and test.
That will have to added w.r.t. #533.

api/nntrainer-api-common.h
zhoonit: We discussed offline to skip this change

api/capi/include/nntrainer.h
zhoonit: We discussed offline to add something like "train_user_data", "validation_user_data", "test_user_data" w.r.t #533

api/nntrainer-api-common.h
```diff
@@ -27,8 +27,15 @@ extern "C" {
  * @since_tizen 6.0
  */
 typedef enum {
-  ML_TRAIN_LAYER_TYPE_INPUT = 0,    /**< Input Layer */
-  ML_TRAIN_LAYER_TYPE_FC,           /**< Fully Connected Layer */
+  ML_TRAIN_LAYER_TYPE_INPUT = 0,  /**< Input Layer */
+  ML_TRAIN_LAYER_TYPE_FC,         /**< Fully Connected Layer */
```
kparichay, line 31: you can consider making it alphabetical.

kparichay (approved):
LGTM

api/nntrainer-api-common.h
zhoonit: Input and FC is released so I guess it is not possible?

kparichay: We do not any applications with nntrainer, so IMO it can be done.
Please take advice from @again4you on this.

zhoonit: If we keep making it sorted, this will break someday let alone additional ACR we have to go thorugh.
As we are going to keep adding layers. Let's keep it as it is if there's no good reason.

kparichay: works.

api/ccapi/include/layer.h
```diff
@@ -33,17 +33,18 @@ namespace train {
...
-  LAYER_BACKBONE_TFLITE,                /** Backbone using TFLite */
+  LAYER_BN = ML_TRAIN_LAYER_TYPE_BN,    /** Batch Normalization Layer type */
+  LAYER_CONV2D = ML_TRAIN_LAYER_TYPE_CONV2D, /** Convolution 2D Layer type */
+  LAYER_POOLING2D = ML_TRAIN_LAYER_TYPE_POOLING2D, /** Pooling 2D Layer type */
+  LAYER_FLATTEN = ML_TRAIN_LAYER_TYPE_FLATTEN,     /** Flatten Layer type */
+  LAYER_ACTIVATION =
+    ML_TRAIN_LAYER_TYPE_ACTIVATION,              /** Activation Layer type */
+  LAYER_ADDITION = ML_TRAIN_LAYER_TYPE_ADDITION, /** Addition Layer type */
+  LAYER_MULTIOUT = ML_TRAIN_LAYER_TYPE_MULTIOUT, /** Multi Output Layer type */
+  LAYER_CONCAT,                                  /** Concat Layer type */
```
jijoongmoon, line 44: We need this concat. resnet needs it.

zhoonit: Updated

jijoongmoon (approved):
LGTM.

PR #884, opened 2021-01-26
[Fix] Reflect changes to upstream/main

zhoonit:
From merging some big prs there happend some inconsistency which casued
a build break. This patch solves the issue

- Use manager.initializeTensor() in the unittest
- Add training signature to forwarding

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/preprocess_translate_layer.h
+++ b/nntrainer/layers/preprocess_translate_layer.h
@@ -64,27 +64,29 @@ public:
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
-  int initialize(Manager &manager);
+  int initialize(Manager &manager) override;
 
   /**
    * @copydoc Layer::forwarding()
    */
-  void forwarding();
+  void forwarding(bool training = true) override;
 
   /**
    * @copydoc Layer::calcDerivative()
    */
-  void calcDerivative();
+  void calcDerivative() override;
 
   /**
    * @copydoc Layer::setTrainable(bool train)
    */
-  void setTrainable(bool train);
+  void setTrainable(bool train) override;
 
   /**
    * @copydoc Layer::getType()
    */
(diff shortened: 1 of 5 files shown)


nntrainer/layers/preprocess_translate_layer.h
```diff
@@ -64,27 +64,29 @@ class PreprocessTranslateLayer : public Layer {
...
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
-  int initialize(Manager &manager);
+  int initialize(Manager &manager) override;
 
   /**
    * @copydoc Layer::forwarding()
    */
-  void forwarding();
+  void forwarding(bool training = true) override;
```
jijoongmoon, line 72: is it right?? training = true?

zhoonit: Yes, semantic is defaulted to be training = true when just do `forwarding()` for every layer.

For augmentation layer, there is no difference.

jijoongmoon: What I mean.. what it means if the training = true for translate layer. does it update weight?? or others?

kparichay (changes requested):
For augmentation layer, their forwarding must be no_op (`hidden_ = input_`) when `training = false`.
Please update

zhoonit:
> For augmentation layer, their forwarding must be no_op (`hidden_ = input_`) when `training = false`.
> Please update

Updated.

nntrainer/layers/preprocess_translate_layer.h
zhoonit: As per below comment, when training is false, there is no augmentation happening. when training is true, augmentation take places.

> For augmentation layer, their forwarding must be no_op (`hidden_ = input_`) when `training = false`.
> Please update

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #889, opened 2021-01-26
[manager] Bug fix using tensor map

### LINKED ISSUE #892: Manager bug when allocating gradient
Manager has a bug when allocating memory for gradient for batch normalization layer, as Tensor::map does not perform memory checks. Replacing Tensor::map() with Tensor::getSharedTensor() uncovers that bug.

kparichay:
Commit 1: [manager] Disable user_shared_memory 
Disable user_shared_memory to as NNAPI is not required to be supported.
It is further needed to decouple tensor structure allocation and its
internal memory allocation.

Commit 2: [manager] bug fix for Tensor::map
This patch exposes a bug from Tensor::Map
where the offset is not checked when assigning the data.

Commit 3: [tensor] Update interface for tensor::map …

Update interface for tensor::map to include the size of the original
buffer to ensure that the buffer contains enough memory required
by the tensor shape wrapping around the memory.
Added another negative unittest for it.

Resolves #892

This is because of the bug in manager for batch normalization layer.

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -124,8 +124,7 @@ MMapedMemory::~MMapedMemory() {
 Manager::Manager(bool enable_gradient_memory_opt_,
                  bool enable_derivative_memory_opt_,
                  bool enable_activation_memory_opt_,
-                 bool enable_inference_inout_memory_opt_,
-                 bool use_shared_memory_) :
+                 bool enable_inference_inout_memory_opt_) :
   total_weight_size(0),
   total_grad_size(0),
   max_grad_size(0),
@@ -136,7 +135,7 @@ Manager::Manager(bool enable_gradient_memory_opt_,
   enable_derivative_memory_opt(enable_derivative_memory_opt_),
   enable_activation_memory_opt(enable_activation_memory_opt_),
   enable_inference_inout_memory_opt(enable_inference_inout_memory_opt_),
-  use_shared_memory(use_shared_memory_) {}
+  use_shared_memory(false) {}
 
 Manager::~Manager() {}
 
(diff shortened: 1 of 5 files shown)


nntrainer/tensor/tensor.h
```diff
@@ -73,6 +73,8 @@ class Tensor {
    * @param offset offset to be used from current
    * @return Tensor object
    * @throws std::invalid_argument if buf is null
+   *
+   * @note This method is unsafe as it does not check buf size.
```
zhoonit, line 77: How about adding todo to add bufsize?

nntrainer/tensor/manager.cpp
```diff
@@ -212,11 +212,13 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   } else if (!is_weight) {
     /** only for gradients */
     if (max_grad_size > 0 && enable_gradient_memory_opt) {
-      std::shared_ptr<float> window(new float[max_grad_size],
-                                    std::default_delete<float[]>());
+      std::shared_ptr<Tensor> shared_grad =
+        std::make_shared<Tensor>(TensorDim({(unsigned int)max_grad_size}));
```
zhoonit, line 216: How about having a memory block instead of a tensor to add extra overhead?
We can add bufsize argument to Tensor::Map.

If you have an ongoing updates regarding this, I think it's fine.

kparichay: In an upcoming PR, this memory is going to be allocated later, when needed.
To allow lazy allocation, wrapping it with tensor will be required.

zhoonit: I'll update tensor map for the future use anyway.
Could you update to `nntrainer::Tensor shared_grad` ? as it is captures it does not need to be a shared pointer or we can use unique_ptr

kparichay: Done. 
Updated tensor map as part of this PR itself.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #891, opened 2021-01-26
[Tensor] Handle copy edge case

zhoonit:
When copying uninitialized tensor to another unintialized tensor,
tensor::copy tried to reshape uninitialized dimension.

This patch fixes the issue.

**Self evaluation:**

### DIFF
--- a/test/unittest/unittest_nntrainer_layers.cpp
+++ b/test/unittest/unittest_nntrainer_layers.cpp
@@ -2118,27 +2118,6 @@ TEST_F(nntrainer_AdditionLayer, setProperty_02_n) {
   EXPECT_EQ(status, ML_ERROR_INVALID_PARAMETER);
 }
 
-TEST_F(nntrainer_AdditionLayer, forwarding_01_n) {
-  setProperty("num_inputs=1");
-
-  sharedTensor input = std::shared_ptr<nntrainer::Tensor>(
-    new nntrainer::Tensor[1], std::default_delete<nntrainer::Tensor[]>());
-  nntrainer::Tensor &in = *input;
-
-  in = nntrainer::Tensor();
-
-  nntrainer::Manager manager;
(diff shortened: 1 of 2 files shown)


test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -2118,7 +2118,7 @@ TEST_F(nntrainer_AdditionLayer, setProperty_02_n) {
   EXPECT_EQ(status, ML_ERROR_INVALID_PARAMETER);
 }
 
-TEST_F(nntrainer_AdditionLayer, forwarding_01_n) {
+TEST_F(nntrainer_AdditionLayer, DISABLED_forwarding_01_n) {
```
zhoonit, line 2121: This was throwing because of the bug revealed in this patch.

setting `num_inputs=1` does not throw in addition layer forwarding.

kparichay: can you apply patch for this bug in this PR as well?

zhoonit: I'd like to ask if it is a negative case or not, because

1. we don't usually verify input when forwarding
2. even in intialize() function, num_input = 1 is allowed.

kparichay: As far as I can see, this unit test throws due to input dimension mismatch and not because of num_input (there is no check on number of inputs in addition layer).
This can be removed addition layer as long as tensor class throws in this case when dimension mismatch.

zhoonit: Clarified. I'll delete this test :)

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM!

PR #894, opened 2021-01-27
[Fix] Resolve static init order fiasco

### LINKED ISSUE #893: std::invalid_argument with examples after building with gbs
Hello, I've just built nntrainer with gbs and after installing it on the device when running the examples I get a C++ exceptions.
```
/lib/nntrainer/bin/applications# ./nntrainer_mnist
terminate called after throwing an instance of 'std::invalid_argument'
  what():  cannot register factory with already taken key:
 ```
This seems to happen before the main even started, as you can see I am not passing the right number of argument so the example should just quit.
Here is the backtrace from gdb:
```
#0  __GI_raise (sig=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x42b6bc20 in __GI_abort () at abort.c:79
#2  0x42d89100 in __gnu_cxx::__verbose_terminate_handler() () from /lib/libstdc++.so.6
#3  0x42d87a64 in ?? () from /lib/libstdc++.so.6
#4  0x42d87ab2 in std::terminate() () from /lib/libstdc++.so.6
#5  0x42d87cd2 in __cxa_throw () from /lib/libstdc++.so.6
#6  0xb605a0ae in nntrainer::AppContext::registerFactory<ml::train::Optimizer>(std::function<std::unique_ptr<ml::train::Optimizer, std::default_delete<ml::train::Optimizer> > (std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int) (this=this@entry=0xb61d293c <nntrainer::AppContext::Global()::instance>, factory=..., key=...,
    int_key=<optimized out>, int_key@entry=0) at /usr/lib/gcc/armv7l-tizen-linux-gnueabi/9.2.0/include/c++/bits/allocator.h:153
#7  0xb605a382 in nntrainer::AppContext::registerFactory<ml::train::Optimizer> (this=0xb61d293c <nntrainer::AppContext::Global()::instance>, factory=<optimized out>, key=..., int_key=0)
    at /usr/lib/gcc/armv7l-tizen-linux-gnueabi/9.2.0/include/c++/new:174
#8  0xb6058cd2 in global constructors keyed to 65535_0_databuffer.cpp.o.91324 () at /usr/lib/gcc/armv7l-tizen-linux-gnueabi/9.2.0/include/c++/bits/hashtable_policy.h:433
#9  0x41010268 in call_init (l=<optimized out>, argc=argc@entry=1, argv=argv@entry=0xbe1ff794, env=env@entry=0xbe1ff79c) at dl-init.c:72
#10 0x4101036c in call_init (env=<optimized out>, argv=<optimized out>, argc=<optimized out>, l=<optimized out>) at dl-init.c:30
#11 _dl_init (main_map=0x41031910, argc=1, argv=0xbe1ff794, env=0xbe1ff79c) at dl-init.c:136
#12 0x41000a84 in ?? () from /lib/ld-linux.so.3
```
Any help would be appreciated.
Thank you,

### PR DESCRIPTION
**Pull request by zhoonit**
As static initialization order out of a translation unit is undefined,
initializing global caused some undefined behavior.

Initializing global app context is delayed until it is first called.

See https://gcc.gnu.org/legacy-ml/gcc-patches/2017-03/msg00863.html

See also #893 (verified)

**Self evaluation:**

### DIFF
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -49,22 +49,18 @@ namespace nntrainer {
 std::mutex factory_mutex;
 
 /**
- * @brief initiate global context
- *
- */
-static void init_global_context_nntrainer(void) __attribute__((constructor));
-
-/**
  * @brief finialize global context
  *
  */
 static void fini_global_context_nntrainer(void) __attribute__((destructor));
 
-static void init_global_context_nntrainer(void) {
+static void fini_global_context_nntrainer(void) {}
+
+std::once_flag global_app_context_init_flag;
+
+static void add_default_object(AppContext &ac) {
   /// @note all layers should be added to the app_context to gaurantee that
   /// createLayer/createOptimizer class is created
-  auto &ac = AppContext::Global();
-
   using OptType = ml::train::OptimizerType;
   ac.registerFactory(ml::train::createOptimizer<SGD>, SGD::type, OptType::SGD);
   ac.registerFactory(ml::train::createOptimizer<Adam>, Adam::type,
@@ -108,10 +104,9 @@ static void init_global_context_nntrainer(void) {
                      LayerType::LAYER_UNKNOWN);
 }
 
-static void fini_global_context_nntrainer(void) {}
-
 AppContext &AppContext::Global() {
   static AppContext instance;
+  std::call_once(global_app_context_init_flag, add_default_object, instance);
   return instance;
 }
 

### CODE REVIEW

File: nntrainer/app_context.cpp [hunk 1]
```diff
@@ -48,23 +48,11 @@ namespace nntrainer {
...
- *
- */
-static void init_global_context_nntrainer(void) __attribute__((constructor));
+std::once_flag global_app_context_init_flag;
 
-/**
- * @brief finialize global context
- *
- */
-static void fini_global_context_nntrainer(void) __attribute__((destructor));
```
kparichay, line 61: how about we keep this?

zhoonit: updated

nntrainer/app_context.cpp
```diff
@@ -108,10 +104,9 @@ static void init_global_context_nntrainer(void) {
                      LayerType::LAYER_UNKNOWN);
 }
 
-static void fini_global_context_nntrainer(void) {}
```
kparichay, line 111: then keep this too?

zhoonit: Yup. It has moved to :57.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #896, opened 2021-01-27
[Util] Add gtest style throw macro @open sesame 02/16 13:03

### LINKED ISSUE #895: Add util to throw with stringstream
it is bothersome to check and throw with formatted item, make a simple wrapper to do this in one line.

zhoonit:
Add gtest style throw macro for productivity.
With this patch, if you have to throw, do in one line

`NNTR_THROW_IF(true, std::invalid_argument) << log << what << ever`

**Self evaluation:**

resolves #895 

v2: Added `NNTR_THROW_IF_CLEANUP(pred, std::invalid_argument, cleanup_func)` to take a cleanup function before throw.

### DIFF
--- a/nntrainer/nntrainer_error.h
+++ b/nntrainer/nntrainer_error.h
@@ -24,12 +24,65 @@
 #define ML_ERROR_RESULT_OUT_OF_RANGE (-ERANGE)
 #endif
 
+#include <functional>
+#include <sstream>
 #include <stdexcept>
+
+#define NNTR_THROW_IF(pred, err) \
+  if ((pred))                    \
+  nntrainer::exception::ErrorNotification<err>()
+
+#define NNTR_THROW_IF_CLEANUP(pred, err, cleanup_func) \
+  if ((pred))                                          \
+  nntrainer::exception::ErrorNotification<err>(cleanup_func)
+
 namespace nntrainer {
 
 /// @note underscore_case is used for ::exception to keep in accordance with
 /// std::exception
 namespace exception {
+
+/**
+ * @brief Error Notification class, error is thrown when the class is destroyed.
+ * DO NOT use this outside as this contains throwing destructor.
+ *
+ * @tparam Err Error type that except cstring as an argument.
+ */
+template <typename Err,
+          typename std::enable_if_t<std::is_base_of<std::exception, Err>::value,
+                                    Err> * = nullptr>
+class ErrorNotification {
+public:
+  /**
+   * @brief Construct a new Error Notification object
+   *
+   */
+  explicit ErrorNotification() : cleanup_func([] {}) {}
+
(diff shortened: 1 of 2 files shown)


nntrainer/nntrainer_error.h
```diff
@@ -24,13 +24,72 @@
...
 /// std::exception
 namespace exception {
 
+namespace internal {
+
+/**
+ * @brief Error Notification class, error is thrown when the class is destroyed.
+ * DO NOT use this outside as this contains throwing destructor.
+ *
+ * @tparam Err Error type that except cstring as an argument.
```
kparichay, line 53: tparam -> param?

nntrainer/nntrainer_error.h
```diff
@@ -24,13 +24,72 @@
 #define ML_ERROR_RESULT_OUT_OF_RANGE (-ERANGE)
 #endif
 
+#include <functional>
+#include <sstream>
 #include <stdexcept>
+
+#define NNTR_THROW_IF(pred, err) \
+  if (!(pred)) {                 \
```
kparichay, line 32: i wonder why `if(!pred) ... else ...` than just `if(pred) ...`?

kparichay (approved):

nntrainer/nntrainer_error.h
zhoonit: I think branch prediction would be the same or minimal as 99% or the time it would go to non-throw.
It is just a pure readability. I think it would be interesting to benchmark this though.

nntrainer/nntrainer_error.h
zhoonit: It's template parameter, so tparam is used 
https://www.doxygen.nl/manual/commands.html#cmdtparam

nntrainer/nntrainer_error.h
zhoonit: Changed to `if((pred))`

nntrainer/nntrainer_error.h
```diff
@@ -24,13 +24,70 @@
...
+  nntrainer::exception::internal::ErrorNotification<err>(cleanup_func)
+
 namespace nntrainer {
 
 /// @note underscore_case is used for ::exception to keep in accordance with
 /// std::exception
 namespace exception {
 
+namespace internal {
+
```
jijoongmoon, line 46: do we need internal namespace?

zhoonit: This is to ask other's never use this class. Should I change name or remove the namespace?

jijoongmoon: I think we don't need.

zhoonit: Removed

zhoonit:
This commit is merged with #911, thus fix will be opened with new PR

PR #907, opened 2021-01-27
[ ResNet ] Add Resnet18 Application

jijoongmoon:
This PR includes ini configuration file for resnet18

**Self evaluation:**

### DIFF
--- a/Applications/ResNet/res/resnet18.ini
+++ b/Applications/ResNet/res/resnet18.ini
@@ -1,0 +1,505 @@
+#
+# SPDX-License_Identifier: Apache-2.0
+# @file resnet18.ini
+# @author Jijoong Moon <jijoong.moon@samsung.com>
+# @date Jan 27 2021
+# @brief input configuration of Residual Net 18
+#
+
+[Model]
+Type = NeuralNetwork
+Learning_rate = 1e-4
+Epochs = 2
+Optimizer = adam
+Loss = cross
+Save_Path = "model.bin"
+batch_size = 128
+beta1 = 0.9
+beta2 = 0.999
+epsilon = 1e-7
+
+[inputlayer]
+Type = input
+Input_Shape = 3:32:32
+
+[first_layer_112_1]
+Type=conv2d
+input_layers = inputlayer
+kernel_size = 3,3
+bias_initializer = zeros
+weight_initializer = xavier_uniform
+filters=64
+stride =1,1
+padding=1,1
+
+[preparation_bn]
+Type=batch_normalization
+input_layers = first_layer_112_1
+epsilon = 1.0e-6
+momentum = 0.9
+Activation = relu
+beta_initializer = zeros
+gamma_initializer = ones
+moving_mean_initializer = zeros
+moving_variance_initializer = ones
+
+[out0]
+Type=output
+input_layers=preparation_bn
+
+[B1_residue_1_conv_3]
+Type=conv2d
+input_layers=out0
+kernel_size = 3,3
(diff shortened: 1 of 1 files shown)


Applications/ResNet/res/resnet18.ini
```diff
@@ -0,0 +1,505 @@
...
+stride =1,1
+padding=1,1
+
+[B1_residue_2_bn]
+Type=batch_normalization
+input_layers = B1_residue_1_conv_3
+epsilon = 1.0e-6
+momentum = 0.9
+Activation = relu
+beta_initializer = zeros
```
kparichay, line 66: if these are default, can we omit their definition?
this makes ini file bigger and difficult to read.

Applications/ResNet/res/resnet18.ini
```diff
@@ -0,0 +1,505 @@
...
+gamma_initializer = ones
+moving_mean_initializer = zeros
+moving_variance_initializer = ones
+
+[out0]
+Type=output
+input_layers=preparation_bn
+
+[B1_residue_1_conv_3]
```
kparichay, line 50: Can we define a block in a separate file and use it as a backbone here.
It is almost impossible to verify if this is the correct/desired architecture.

Applications/ResNet/res/resnet18.ini
```diff
@@ -0,0 +1,505 @@
...
+Type=pooling2d
+input_layers = out8
+pool_size = 4,4
+stride =4,4
+padding = 0,0
+pooling = average
+
+[concatenate_1]
+Type=concat
+input_layers=max_pooling2d_1, average_pooling2d_1
```
kparichay, line 487: Original ResNet18 has no concatenation.
Plus concatenation is not exposed with the new C-API.
Let's not use `concat` layer.

Applications/ResNet/res/resnet18.ini
```diff
@@ -0,0 +1,505 @@
...
+
+[linear]
```
kparichay, line 489: lets either rename to `conv**` or change type to `fully_connected`. It's the same essentially, but just for readability.

Applications/ResNet/res/resnet18.ini
jijoongmoon: Yes. That's what I thought too. but let's do it after finishing the evaluation.

Applications/ResNet/res/resnet18.ini
jijoongmoon: We will expose concat. I ask. I think it is good chance to test concat though.

Applications/ResNet/res/resnet18.ini
jijoongmoon: yes. it is. but I think it can be as it is to compare the reference.

Applications/ResNet/res/resnet18.ini
jijoongmoon: I'm not sure. I will decide if it is ok to use default when the evaluation is done.

zhoonit (approved):
LGTM

Applications/ResNet/res/resnet18.ini
kparichay: Alright.

Applications/ResNet/res/resnet18.ini
kparichay: Yes, of course, this naming is just for our convenience.

kparichay (approved):
LGTM

PR #908, opened 2021-01-27
[AppContext] Add register plugin function @open sesame 03/02 13:20

zhoonit:
- ~**#896** [Util] Add gtest style throw macro~
- [API] Add layer pluggable definition
```
Add layer pluggable definition to be used when creating a custom layer
to be plugged into nntrainer

**Self evaluation:**

```
- [Pluggable] Add layer wrapper for plugged layer
```
As layer loaded from a `dlsym` cannot destroyed by `delete`, we must
need a way to delete it from destroy funcion. This patch add a simple
wrapper to deal with the problem.

Additionally, This patch is making functions in layer_internal
virtual which paves way to clean up api. Please think
`plugged_layer.h` as a layer api draft

**Self evaluation:**

```
- [AppContext] Add register plugin function
```
This patch adds register plugin function to be used by loading library

**Self evaluation:**

```
- [Test] Add test to create layer from an example
```
This patch adds libpow_layer.so as an example.
Also this patch adds a simple test to load and register the so file.

**Self evaluation:**

```


Applications/Custom/pow.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#ifdef DEBUG
+  std::cout << "input: " << net_input[0]->getVariable();
+  std::cout << "output: " << net_hidden[0]->getVariable();
+  PowUtil::pause();
+#endif
+}
+
+void PowLayer::calcDerivative() {
+/// intended here to demonstrate that PowLayer::backwarding is being called
```
jijoongmoon, line 120: name is calcDerivative?? backward??

jijoongmoon:
Basically, layer_internal.h includes layer.h and this pluggable layer needs to include layer.h only tool. So.. I don't think it is good to change every function in layer_intenal.h as a virtual.

nntrainer/app_context.cpp
```diff
@@ -136,18 +156,43 @@ void AppContext::setWorkingDirectory(const std::string &base) {
...
+
+  auto layer = pluggable->createfunc();
+  auto type = layer->getType();
+  NNTR_THROW_IF_CLEANUP(type == "", std::invalid_argument, close_dl)
+    << func_tag << "custom layer must specify type name, but it is empty";
+  pluggable->destroyfunc(layer);
+
+  FactoryType<ml::train::Layer> factory_func =
+    [pluggable](const PropsType &prop) {
+      std::unique_ptr<ml::train::Layer> layer =
```
kparichay, line 189: I just wonder about the purpose of `PluggedLayer`, as `PluggedLayer` interface must be always same as ml::train::Layer (nntrainer::Layer for now).
How about just `std::unique_ptr<ml::train::Layer> layer = std::unique_ptr<nntrainer::Layer, decltype(&pluggable->destroyfunc)>(pluggable->createfunc(), &pluggable->destroyfunc)` or did I understood something wrong?

zhoonit: That was my first try. The problem is `std::unique_ptr<nntrainer::Layer, decltype(&pluggable->destroyfunc)>` is different type from `std::unique_ptr<ml::train::Layer>`, so it does not play well with other unique_ptr. One option is we move every deleter to lambda but it is like x4 memory for a pointer, let alone change in a lot of spot.

kparichay: `std::unique_ptr<nntrainer::Layer, decltype(&pluggable->destroyfunc)>` should be able to reinterpret as `std::unique_ptr<ml::train::Layer>`??

zhoonit: > std::unique_ptr<nntrainer::Layer, decltype(&pluggable->destroyfunc)> should be able to reinterpret as std::unique_ptr<ml::train::Layer>??

This won't call `destroyfunc()` as it is reinterpreted. thus not viable choice

> Additionally, std::unique_ptr<nntrainer::Layer, decltype(&pluggable->destroyfunc)> will anyway be replaced with std::unique_ptr<ml::train::Layer, decltype(&pluggable->destroyfunc)> once ml::train::Layer interface is finalized.

It doesn't have to be in this setup in the cost of pimpl delegation. Downside of changing `std::unique_ptr<Layer>` to `std::unique_ptr<Layer, lambda>`, This affects the performance let alone usability (it affects usability because you always need `std::unique_ptr<Layer, std::function...>` in every 
signature and variables.)

you might wanna read https://dev.krzaq.cc/post/you-dont-need-a-stateful-deleter-in-your-unique_ptr-usually/

kparichay: If it can't, IMO, it would be better to have an alias for `std::unique_ptr<ml::train::layer, lambda>` and use it, than maintaining another Layer extension `PluggedLayer` and then `PluggedOptimizer`, `PluggedModel` and so on.

zhoonit: > If it can't, IMO, it would be better to have an alias for std::unique_ptr<ml::train::layer, lambda> and use it, than maintaining another Layer extension PluggedLayer and then PluggedOptimizer, PluggedModel and so on.

plus, the unique_ptr won't be default constructible 
```cpp
using PtrType<Layer> = std::unique_ptr<ml::train::layer, lambda>
std::vector<PtrType<Layer>> // this will be compile error as PtrType<Layer> is not default constructible.
```

kparichay (approved):
LGTM

nntrainer/layers/plugged_layer.h
```diff
@@ -0,0 +1,258 @@
...
+   * @copydoc Layer::calcDerivative()
+   */
+  void calcDerivative() override { layerImpl->calcDerivative(); }
+
+  /**
+   * @copydoc Layer::calcGradient()
+   */
+  void calcGradient() override { layerImpl->calcGradient(); }
+
+  /**
```
jijoongmoon, line 89: don't we need layerImpl->applyGradient ??

Applications/Custom/pow.cpp
zhoonit: Currently, `calcDerivative()` and `calcGradient` are called in `neuralnet` so overriding `backwarding` doesn't work and not needed.

nntrainer/layers/plugged_layer.h
zhoonit: Yes, thank you for noticing hardly noticeable issue :)

zhoonit: Updated

Applications/Custom/LayerPlugin/layer_plugin_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+const char *RES_PATH = std::getenv("RES_PATH");
+
+TEST(AppContext, DlRegisterOpen_p) {
+  ASSERT_NE(RES_PATH, nullptr) << "RES_PATH environment value must be set";
+  auto &ac = nntrainer::AppContext::Global();
+
+  ac.registerLayerPlugin("libpow_layer.so", RES_PATH);
+
+  auto layer = ml::train::createLayer("pow");
```
myungjoo, line 30: If this is the standard method to register a custom op, it LGTM!

jijoongmoon (approved):
LGTM

PR #911, opened 2021-01-28
[Build] Decouple gtest from nntrainer_test_util @open sesame 02/15 17:40

### LINKED ISSUE #910: clean up gtest for the tests
gtest seems not available in the tizen or possibly other normal platform, 
Also, we don't have need every main, we can replace it to gtest 
let's make gtest self-contained for portability

zhoonit:
- **#896**  [Util] Add gtest style throw macro
- [Build] Decouple gtest from nntrainer_test_util
```
As nntrainer_test_util had gtest, it was preventing from compiler checks
some trivial bugs (like sign compare) and had some weird bug.

This patch decouples gtest from nntrainer_test_util while changing gtest
to static build.

Resolves #910

**Self evaluation:**

```

### DIFF
--- a/nntrainer/nntrainer_error.h
+++ b/nntrainer/nntrainer_error.h
@@ -24,12 +24,69 @@
 #define ML_ERROR_RESULT_OUT_OF_RANGE (-ERANGE)
 #endif
 
+#include <functional>
+#include <sstream>
 #include <stdexcept>
+
+#define NNTR_THROW_IF(pred, err) \
+  if ((pred))                    \
+  nntrainer::exception::internal::ErrorNotification<err>()
+
+#define NNTR_THROW_IF_CLEANUP(pred, err, cleanup_func) \
+  if ((pred))                                          \
+  nntrainer::exception::internal::ErrorNotification<err>(cleanup_func)
+
 namespace nntrainer {
 
 /// @note underscore_case is used for ::exception to keep in accordance with
 /// std::exception
 namespace exception {
+
+namespace internal {
+
+/**
+ * @brief Error Notification class, error is thrown when the class is destroyed.
+ * DO NOT use this outside as this contains throwing destructor.
+ *
+ * @tparam Err Error type that except cstring as an argument.
+ */
+template <typename Err,
+          typename std::enable_if_t<std::is_base_of<std::exception, Err>::value,
+                                    Err> * = nullptr>
+class ErrorNotification {
+public:
+  /**
+   * @brief Construct a new Error Notification object
+   *
+   */
+  explicit ErrorNotification() : cleanup_func([] {}) {}
+
+  explicit ErrorNotification(std::function<void()> cleanup_func_) :
+    cleanup_func(cleanup_func_) {}
+
+  /**
+   * @brief Destroy the Error Notification object, Error is thrown when
+   * destroying this
+   *
+   */
+  ~ErrorNotification() noexcept(false) {
+    cleanup_func();
+    throw Err(ss.str().c_str());
+  }
+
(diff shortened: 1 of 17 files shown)


nntrainer/nntrainer_error.h
```diff
@@ -24,13 +24,72 @@
...
 #endif
 
+#include <functional>
+#include <sstream>
 #include <stdexcept>
+
+#define NNTR_THROW_IF(pred, err) \
+  if (!(pred)) {                 \
+  } else                         \
+    nntrainer::exception::internal::ErrorNotification<err>()
```
myungjoo, line 34: This looks extremely dangerious. Can you make the macro closed?

E.g.,
```
#define NNTR_THROW_IF(pred, err, message) \
   do { \
      ....
   } while (0)
```

myungjoo: Note that the following code works. So don't worry:
```
#include <iostream>
#define A(a)\
        do { \
                std::cout << a << std::endl; \
        } while (0)

int main()
{
        A(123123 << std::endl << "33" << std::endl);
}
```

myungjoo: Question:
  Why do you need the empty "then" statement here? Are you trying to hack the branch-prediction mechanism?

zhoonit: Not closing the stream intentional. This enables logging directly to a stream, as [gtest](https://stackoverflow.com/a/16491892) or [aosp](https://cs.android.com/android/platform/superproject/+/master:packages/modules/NeuralNetworks/runtime/CompilationBuilder.cpp;l=103) does.

They also do not close the stream at the end.

I tried to minimize the code that needs to be inside the macro. Similar to implementation in [aosp](https://cs.android.com/android/platform/superproject/+/master:system/libbase/include/android-base/logging.h;drc=master;l=219) is doing.
If it doesn't sound right, I will change to the suggested style.

zhoonit: > Question:
> Why do you need the empty "then" statement here? Are you trying to hack the branch-prediction mechanism?

It is not necessary, I'll change.

myungjoo: > > Question:
> > Why do you need the empty "then" statement here? Are you trying to hack the branch-prediction mechanism?
> 
> It is not necessary, I'll change.

No. keep as it is. If you are trying their style, Having an empty "then statement" is safer. Assume you have :
```
if (some condition) NNTR_THROW_IF (X, Y) << "message";
else NNTR_THROW_IF (Z, W) << "message2";
```

myungjoo: Keep the original code if your intention is to follow their style. The whole statement is a single C statement; thus, do-while is not really required.

zhoonit: Awesome :)

jijoongmoon (approved):
LGTM

test/meson.build
```diff
@@ -1,23 +1,23 @@
...
+  include_directories: nntrainer_test_inc
+)
+
+nntrainer_testutil_dep = declare_dependency(
+  link_with: nntrainer_testutil_lib,
+  include_directories: nntrainer_test_inc
+)
+
+nntrainer_test_deps += nntrainer_testutil_dep
+nntrainer_test_deps += gtest_dep
```
kparichay, line 20: while you are editing around gtest, can you disable unit tests build if gtest_dep not found.

zhoonit: Added a check if gtest not found, make error when test is enabled

myungjoo (approved):

PR #912, opened 2021-01-28
[#960] [Fix] Make mnist application runnable from install

### LINKED ISSUE #731: Save path has a default value
Currently save_path has a default value when set from ini but we don't have default save_path when setting from a code.

```cpp
  model.save_path = iniparser_getstring(ini, "Model:Save_path", "./model.bin");
```

We have two options

1. moving default save_path to `Neuralnet::Constructor()`
2. remove default save_path

zhoonit:
commit 1 - [App] install application 
```
mnist_trainingSet.dat was missing in the installation,
which prevented the app to run.

This patch fixes the issue.

**Self evaluation:**

See #893
```

commit 2 - [Fix/Bug] NeuralNet check default save path
```
Add a check to see if save_path is valid if not, issue a warning

**Semantic changes:**
From this patch, default save path is gone

**Self evaluation:**

```

resolves #731


zhoonit:
Please note that this patch requires save path modification to pass the test, it will be done and this PR will be rebased.

nntrainer/models/model_loader.cpp
```diff
@@ -62,8 +62,8 @@ int ModelLoader::loadModelConfigIni(dictionary *ini, NeuralNetwork &model) {
   model.loss_type = (LossType)parseType(
     iniparser_getstring(ini, "Model:Loss", unknown), TOKEN_LOSS);
   const std::string &save_path =
-    iniparser_getstring(ini, "Model:Save_path", "./model.bin");
-  model.setSavePath(model.app_context.getWorkingPath(save_path));
+    iniparser_getstring(ini, "Model:Save_path", "tmp_model.bin");
+  model.setSavePath(save_path);
```
kparichay, line 66: I wonder why the prefix of `getWorkingPath` has been removed?

zhoonit: `model.setSavePath` internally calls `save_path` so this is not needed here.

zhoonit:
This pr needs a patch regarding #731 before merge

nntrainer/models/model_loader.cpp
kparichay: So, if model has a `getSavePath` later, it will also have to add `getWorkingPath`?

zhoonit: `setSavePath` adds it so it won't need `getWorkingPath`

zhoonit:
v2. amended last commit to remove default path to resolve #731 

Add a check to see if save_path is valid if not, issue a warning

**Semantic changes:**
From this patch, default save path is gone

**Self evaluation:**

jijoongmoon (approved):
LGTM

kparichay (approved):
LGTM

PR #915, opened 2021-01-29
Load layer plugin from directories @open sesame 03/02 16:11

zhoonit:
This patch mainly adds the functionality to add `.so` from a directory (by api, environment variable, and configuration ini)

- ~**#894** [Fix] Resolve static init order fiasco~
- ~**#896** [Util] Add gtest style throw macro~
- ~**#908** [API] Add layer pluggable definition~
- ~**#908** [Pluggable] Add layer wrapper for plugged layer~
- ~**#908** [AppContext] Add register plugin function~
- ~**#908** [Test] Add test to create layer from an example~
- [AppContext] Default Path loader
```
Add appcontext default path loader.

- add `AppContext::registerLayerFromDirectory`
- call `add_extension_object` routine for `AppContext::Global()`. The
routine searches the directory of `NNTRAINER_PATH` and from
`nntrainer.ini`(installing ini in upcoming PR)

**Self evaluation:**

```
- [Conf] Add default conf nntrainer.ini
```
This patch adds nntrainer.ini (defaulted to be `/etc/nntrainer.ini`)
The path is defaulted to be '/etc/nntrainer.ini' but it is subjected to
change by changing `--sysconfdir=''` and `--prefix`

**Self evaluation:**

```

### DIFF
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -55,14 +55,13 @@
 static std::string layerlib_suffix = "layer.so";
 static const std::string func_tag = "[AppContext] ";
 
-constexpr const char *DEFAULT_CONF_PATH = "/etc/nntrainer.ini";
-
-constexpr const char *getConfPath() {
 #ifdef NNTRAINER_CONF_PATH
-  return NNTRAINER_CONF_PATH;
+constexpr const char *DEFAULT_CONF_PATH = NNTRAINER_CONF_PATH;
+#else
+constexpr const char *DEFAULT_CONF_PATH = "/etc/nntrainer.ini";
 #endif
-  return DEFAULT_CONF_PATH;
-}
+
+constexpr const char *getConfPath() { return DEFAULT_CONF_PATH; }
 
 namespace nntrainer {
 
@@ -80,7 +79,7 @@ bool endswith(const std::string &target, const std::string &suffix) {
  *
  * @return std::string plugin path
  */
-std::string getPluginPathConf() {
+std::string getPluginPathConf(const std::string &suffix) {
   std::string conf_path{getConfPath()};
 
   NNTR_THROW_IF(!isFileExist(conf_path), std::invalid_argument)
@@ -93,7 +92,11 @@ std::string getPluginPathConf() {
 
   auto freedict = [ini] { iniparser_freedict(ini); };
 
-  const char *path = iniparser_getstring(ini, "plugins:layer", NULL);
+  std::string s{"plugins:"};
+
+  s += suffix;
+
+  const char *path = iniparser_getstring(ini, s.c_str(), NULL);
   NNTR_THROW_IF_CLEANUP(path == nullptr, std::invalid_argument, freedict)
     << func_tag << "plugins layer failed";
 
@@ -110,6 +113,14 @@ std::string getPluginPathConf() {
 std::vector<std::string> getPluginPaths() {
   std::vector<std::string> ret;
 
+  /*** @note NNTRAINER_PATH is an environment variable stating a @a directory
+   * where you would like to look for the layers, while NNTRAINER_CONF_PATH is a
+   * (buildtime hardcoded @a file path) to locate configuration file *.ini file
+   */
+  /*** @note for now, NNTRAINER_PATH is a SINGLE PATH rather than serise of path
+   * like PATH environment variable. this could be improved but for now, it is
+   * enough
+   */
   const char *env_path = std::getenv("NNTRAINER_PATH");
   if (env_path != nullptr) {
     if (isFileExist(env_path)) {
@@ -122,7 +133,7 @@ std::vector<std::string> getPluginPaths() {
   }
 
   try {
-    std::string conf_path = getPluginPathConf();
+    std::string conf_path = getPluginPathConf("layer");
     ret.emplace_back(conf_path);
     ml_logd("DEFAULT CONF PATH, path: %s", conf_path.c_str());
   } catch (std::exception &e) {
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -15,6 +15,10 @@
 #include <dlfcn.h>
 #include <iostream>
 #include <sstream>
+#include <string>
+#include <vector>
+
+#include <iniparser.h>
 
 #include <app_context.h>
 #include <nntrainer_error.h>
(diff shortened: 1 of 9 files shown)


nntrainer/app_context.cpp
```diff
@@ -196,4 +303,31 @@ int AppContext::registerLayerPlugin(const std::string &path,
...
+
+  NNTR_THROW_IF(dir == nullptr, std::invalid_argument)
+    << func_tag << "failed to open the directory: " << base_path;
+
+  struct dirent *entry;
+
+  std::vector<int> keys;
+  while ((entry = readdir(dir)) != NULL) {
+    if (endswith(entry->d_name, layerlib_suffix)) {
+      try {
```
jijoongmoon, line 318: so.. to register custome layer, it must ends with ".......layer.so"??

zhoonit: Yes. This is tentative. Any suggestions are welcome :)

myungjoo:
Q1. are you going to let nntrainer load ALL plugins at launch-time if the plugins are available in a specified directory?
Q2. (or) are you going to make a default plugin directory so that users do not need to specify the directory if it's in the default directory?

zhoonit:
> Q1. are you going to let nntrainer load ALL plugins at launch-time if the plugins are available in a specified directory?
> Q2. (or) are you going to make a default plugin directory so that users do not need to specify the directory if it's in the default directory?
> 
> Please explain your intention more concretely.

There are two ways to load plugins

1. Runtime api call
  - `AppContext::registerLayerFromDirectory(const std::string & directory_path)` to add a directory 
  - `AppContext::registerLayer(const std::string & so_path)`
2. Launch-time loading from default directory. Directory is either defined at
  - `NNTRAINER_PATH` environment variable (analogous to PKG_CONFIG_PATH or PYTHON_PATH)
  - Default path defined in `nntrainer.ini`.

jijoongmoon (approved):
LGTM

nntrainer/app_context.cpp
```diff
@@ -47,8 +51,104 @@
...
+constexpr const char *getConfPath() {
+#ifdef NNTRAINER_CONF_PATH
+  return NNTRAINER_CONF_PATH;
+#endif
+  return DEFAULT_CONF_PATH;
+}
+
 namespace nntrainer {
 
+namespace {
```
kparichay, line 69: name of the namespace?

nntrainer/app_context.cpp
```diff
@@ -47,8 +51,104 @@
...
 #endif
 
+/// add #ifdef across platform
+static std::string layerlib_suffix = "layer.so";
+static const std::string func_tag = "[AppContext] ";
+
+constexpr const char *DEFAULT_CONF_PATH = "/etc/nntrainer.ini";
+
+constexpr const char *getConfPath() {
+#ifdef NNTRAINER_CONF_PATH
```
kparichay, line 61: How about set NNTRAINER_CONF_PATH to DEFAULT_CONF_PATH with `#define`. 
and let this function can simply return NNTRAINER_CONF_PATH.

nntrainer/app_context.cpp
```diff
@@ -47,8 +51,104 @@
...
+
+/**
+ * @brief Get the plugin paths
+ *
+ * @return std::vector<std::string> list of paths to search for
+ */
+std::vector<std::string> getPluginPaths() {
+  std::vector<std::string> ret;
+
+  const char *env_path = std::getenv("NNTRAINER_PATH");
```
kparichay, line 113: can you add a comment here differentiating NNTRAINER_PATH and NNTRAINER_CONF_PATH.

nntrainer/app_context.cpp
```diff
@@ -47,8 +51,104 @@
...
+    << func_tag << "There is no existing config file";
+  ml_logd("%s conf path: %s", func_tag.c_str(), conf_path.c_str());
+
+  dictionary *ini = iniparser_load(conf_path.c_str());
+  NNTR_THROW_IF(ini == nullptr, std::runtime_error)
+    << func_tag << "loading ini failed";
+
+  auto freedict = [ini] { iniparser_freedict(ini); };
+
+  const char *path = iniparser_getstring(ini, "plugins:layer", NULL);
```
kparichay, line 96: it would be better to have `layer` as an argument.

nntrainer/app_context.cpp
```diff
@@ -107,32 +207,39 @@ static void add_default_object(AppContext &ac) {
...
-
-namespace {
-const std::string getFullPath(const std::string &path,
-                              const std::string &base) {
-  /// if path is absolute, return path
-  if (path[0] == '/') {
-    return path;
+  for (auto &path : dir_list) {
+    try {
+      ac.registerLayerFromDirectory(path);
```
kparichay, line 215: from the point of view of this function, shouldn't this just call a hypothetical `ac.registerFromDirectory(path)`. `registerFromDirectory` can figure out if the file inside is a layer or other type based on prefix?

nntrainer/app_context.cpp
kparichay: Is a suffix needed? IMO different folders for different types - layers, optimizers, etc. - might be better.

nntrainer/app_context.cpp
```diff
@@ -199,4 +306,31 @@ int AppContext::registerLayerPlugin(const std::string &path,
...
+
+  std::vector<int> keys;
+  while ((entry = readdir(dir)) != NULL) {
+    if (endswith(entry->d_name, layerlib_suffix)) {
+      try {
+        int key = registerLayer(entry->d_name, base_path);
+        keys.emplace_back(key);
+      } catch (std::exception &e) {
+        closedir(dir);
+        throw;
```
kparichay, line 326: throw e?

nntrainer/app_context.h
```diff
@@ -94,8 +94,16 @@ class AppContext {
...
-  int registerLayerPlugin(const std::string &library_path,
-                          const std::string &base_path = "");
+  int registerLayer(const std::string &library_path,
+                    const std::string &base_path = "");
+
+  /**
+   * @brief register Layer from a directory.
+   *
+   * @param base_path a directory path to search layer's
+   * @return std::vector<int> list of integer key to create a layer
```
kparichay, line 104: add a note here for the expected behavior when the same layer is registered twice.

kparichay: you might also consider adding a test case for the same.

nntrainer/app_context.cpp
zhoonit: This is intended, https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces

nntrainer/app_context.cpp
zhoonit: What you've said sounds about right.

We could have

```
void registerFromDirectory(path) {
ac.registerLayerFromDirectory(path);
ac.registerOptimizerFromDirectory(path);
ac.registerModelsFromDirectory(path);
}
```

or if we fix the policy as having only one folder and put everything inside. below could work too.
```
void registerFromDirectory(path, const std::string prefix = "_layer.so") {
}
```

It seems more straight forward to have `ac.registerLayerFromDirectory` for now, later extend in a way you recommended when we start extending the feature to other modules after fixing some policies related to this.

nntrainer/app_context.cpp
zhoonit: `throw` is used to rethrow original throw object without loss of information.

refering to https://stackoverflow.com/questions/2360597/c-exceptions-questions-on-rethrow-of-original-exception

nntrainer/app_context.h
zhoonit: It is responsibility of `registerFactory` and there is a testcase for that: `unittest_nntrainer_appcontext.cpp:RegisterFactoryWithClashingKey_n`

I'll add a note comment.

nntrainer/app_context.cpp
zhoonit: Done.

nntrainer/app_context.cpp
zhoonit: Done

nntrainer/app_context.cpp
zhoonit: Done

nntrainer/app_context.h
kparichay: Then please add a comment to ensure that NNTRAINER_PATH cannot contain duplicate paths.
Also, if we don't support multiple paths in NNTRAINER_PATH, add as a comment as well.

kparichay (approved):
LGTM

nntrainer/app_context.h
zhoonit: Good point, I'll add

zhoonit: Added to the place where NNTRAINER_PATH is explained

PR #918, opened 2021-01-29
[adam] Reduce temporary memory allocation

### LINKED ISSUE #917: Temporary allocation of memory in adam
Temporary allocation of memory is being done in Adam optimizer for each weight.
This is being done for each weight with every iteration, leading to a lot of temporary allocations as well as higher peak memory.

kparichay:
Reduce temporary memory allocation for adam
This is done by reusing gradient memory to calculate the final
update which is to be applied for the weight.

This reduces temporary memory allocations, which were being every epoch
for each weight, but also reduces peak memory usage.

Resolves #917

This has a side effect that the after `applyGradient()`, `getGradient()` will return
the `update` to be applied to the weight than the gradient itself.


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #919, opened 2021-01-29
Move weight regularization @open sesame 02/02 20:01

### LINKED ISSUE #762: Weight regularization must move out of layers
Weight regularization must move out of each layer to a common location.

### LINKED ISSUE #774: Create a graph attribute

kparichay:
**Commit 1**: [weight/layer] Move weight regularization out of layers
Move weight regularization out of layers to weights
and remove same code from the all the layers.
Loss and grads from weight regularization is done by the
weight itself.

**Commit 2**: [weight] Move apply gradient to weight
Move apply gradient which will just add the gradient
to the variable multiplied with the gradient.

**Commit 3**: [optimizer] Update to camelcase
Update apply_gradient(s) to applyGradient(s)

Resolves #762 #774

### DIFF
--- a/nntrainer/optimizers/optimizer.cpp
+++ b/nntrainer/optimizers/optimizer.cpp
@@ -46,8 +46,8 @@ double Optimizer::getLearningRate(int iteration) {
   return ll;
 }
 
-void Optimizer::apply_gradients(std::vector<Weight> &weight_list,
-                                int iteration) {
+void Optimizer::applyGradients(std::vector<Weight> &weight_list,
+                               int iteration) {
 
   if (weight_list.empty())
     return;
@@ -61,7 +61,7 @@ void Optimizer::apply_gradients(std::vector<Weight> &weight_list,
     /** calculate regularization gradient before applying the gradient */
     weight.calcRegularizationGradient();
 
-    apply_gradient(weight, ll, iteration);
+    applyGradient(weight, ll, iteration);
   }
 }
 
--- a/nntrainer/optimizers/optimizer.cpp
+++ b/nntrainer/optimizers/optimizer.cpp
@@ -58,6 +58,9 @@ void Optimizer::apply_gradients(std::vector<Weight> &weight_list,
     if (!weight.getTrainable())
       continue;
 
+    /** calculate regularization gradient before applying the gradient */
+    weight.calcRegularizationGradient();
+
     apply_gradient(weight, ll, iteration);
   }
 }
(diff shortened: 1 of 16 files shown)


jijoongmoon (approved):
LGTM

nntrainer/optimizers/optimizer.cpp
```diff
@@ -58,6 +58,9 @@ void Optimizer::apply_gradients(std::vector<Weight> &weight_list,
     if (!weight.getTrainable())
       continue;
 
+    /** calculate regularization gradient before applying the gradient */
+    weight.calcRegularizationGradient();
```
zhoonit, line 62: I wonder if this logic can go inside `weight::apply_gradient`.

zhoonit (approved):
Nice work :)

nntrainer/optimizers/optimizer.cpp
kparichay: Adam takes in the gradient and updates the gradient which is to be applied to the weight.
So, the regularization gradient must be added before adam operation updates the weight.

Moving it inside the `weight::apply_gradient` would make regularization skip adam. This might be useful for `weight_decay` (weight decay skips adam operations on those gradient) when we support it later.

Issue #927, opened 2021-02-02
Use nntrainer model for feature extraction


klmntoth:
Hello!
I would like to use nntrainer as a feature extractor, like how tflite is used in the KNN example. But calling forwaring/inference uses the automatically added loss layer. Right now after calling inference i'm using `model.getFlatGraph().back()->getOutputs() ...` to get the values from the last layer added by me. Is there a better way to do this?

kparichay:
Hello
There are multiple ways to extract the features from a pre-trained model - 

- As you described, running the `inference()` method and then `model.getFlatGraph().back()->getOutputs()`. However, this becomes complicated when extracting features from other layers.

Unfortunately, the current implementation of NNTrainer always attached a loss to the end of the model. With the loss attached, the inference includes the affect of the loss. In order to extract the features manually, there is no better way for now. 

Soon we will be supporting extracting nntrainer trained model with TfLite compatible format #899 and then, you can directly use your trained model for feature extraction.

If the extracted features are going to be used later for more processing, you can directly use the model till the last layer as a backbone for your final task.
If the goal is just to extract the features, there is another hack you can try meanwhile is to use a loss (MSE in this case) for the inference model which does not affect the output of the layer. The inference model can use a submodel of the full training model as a backbone, and then use the features correspondingly.

kparichay:
We will add support for allowing no loss in the model, in order to facilitate your requirement.
The PR for this will be up soon.

kparichay:
Please check #934 which allows setting loss as `none` or skipping adding loss to a model.
You can create two model definition files (ini), one for training and another for feature extraction (consider using backbone while specifying desired output layer), load the trained model binary file in the model created for feature extraction.

Please let me know if you have more questions on this.

klmntoth:
Can you please give me an example of an ini file for feature extraction? I tried to make one with `loss=none` and using my original ini as backbone, but i'm getting way different values then using the method i posted before. Also i tried to simply use `setProperty({"loss=none"})` from the code using my originial ini, but i'm getting the same different values as with the second ini.

kparichay:
As you noticed, using the method you posted before vs using `loss=none` can give different results in two scenarios:
1. Last layer is softmax and loss is cross-entropy
2. Last layer is sigmoid and loss is cross-entropy
For numerical stability, these two layers are merged together as a single operation. So, when you use your method, `model.getFlatGraph().back()->getOutputs()`, this would return the output of the model without softmax/sigmoid.
[...]

PR #934, opened 2021-02-04
[loss] Allow no loss in the model @open sesame 02/19 17:43

kparichay:
Allow setting no loss in the model
This allows inferencing a model, and creating submodels
with backbones to infer some particular output features

Updated existing unittest with no loss to succeed
and added more unittests to validate that models run successfully
without loss

See Also #927

**Self evaluation:**

### DIFF
--- a/nntrainer/utils/parse_util.cpp
+++ b/nntrainer/utils/parse_util.cpp
@@ -71,8 +71,9 @@ unsigned int parseType(std::string ll, InputType t) {
    * @brief     Loss Function String from configure file
    *            "mse"  : Mean Squared Error
    *            "caterogical" : Categorical Cross Entropy
+   *            "none" : No Loss attached to the model
    */
-  std::array<std::string, 2> loss_string = {"mse", "cross"};
+  std::array<std::string, 3> loss_string = {"mse", "cross", "none"};
 
   /**
    * @brief     Model Type String from configure file
--- a/test/unittest/unittest_nntrainer_models.cpp
+++ b/test/unittest/unittest_nntrainer_models.cpp
@@ -20,6 +20,7 @@
 #include <bn_layer.h>
 #include <input_layer.h>
 #include <layer.h>
+#include <loss_layer.h>
 #include <neuralnet.h>
 #include <weight.h>
 
@@ -195,6 +196,9 @@ public:
                   const nntrainer::TensorDim &label_shape,
                   unsigned int iterations);
 
+  void validateFor(const std::string &reference,
+                   const nntrainer::TensorDim &label_shape);
+
 private:
   std::array<nntrainer::Tensor, 2>
   prepareData(std::ifstream &f, const nntrainer::TensorDim &label_dim);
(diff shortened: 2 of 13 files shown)


jijoongmoon (approved):
LGTM

nntrainer/utils/parse_util.cpp
```diff
@@ -136,6 +137,8 @@ unsigned int parseType(std::string ll, InputType t) {
 
   switch (t) {
   case TOKEN_LOSS:
+    if (ll.empty())
```
lhs8928, line 140: 1. I think ll never goes to empty even the loss is omitted. 
Instead it would be a "none".
Check line 62 at model_loader.cpp

2. Also would it be better to declare a none at the beginning in loss_string, LossType?
Then these 2 line could be deleted :)

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -446,20 +475,36 @@ class nntrainerModelTest
...
+    EXPECT_EQ(std::get<0>(GetParam()), std::get<0>(GetParam()));
+  }
 }
 
 /**
  * @brief check given ini is failing/suceeding at load
  */
 TEST_P(nntrainerModelTest, model_test_optimized) {
+  /** Check model with all optimizations on */
+  if (getIniName().find("validate") == std::string::npos) {
```
zhoonit, line 492: IMO, `*_vallidate` should have different test suite than `model_test` to keep consistency and reduce complexity.

How about we have this validation test added to `unittest_nntrainer_modelfile`? I think this is good spot because other modelfile should have a simplerun test as well.

This isn't directly related to PR. I am willing to work on it after this is merged.

zhoonit (approved):
LGTM

zhoonit:
Please rebase this, there was a change regarding api dependency.

test/unittest/unittest_nntrainer_models.cpp
kparichay: A different test suite can be created. However, then each test suite will have to create its own models. If you planning to update test-suite, it would be good if unittests are able to merge and use each other defined models.

The reason for adding the tests here was to validate for now and then later run the same model.
Given this is a valid model, this test wants to check the training/inferencing this model as a user would do it (later add checking exact values).

> How about we have this validation test added to unittest_nntrainer_modelfile? I think this is good spot because other modelfile should have a simplerun test as well.

`unittest_nntrainer_modelfile` provides initialization check only.

IMO these tests can be done later.

nntrainer/utils/parse_util.cpp
kparichay: will do.

test/unittest/unittest_nntrainer_models.cpp
zhoonit: Understood, if that's the intention I think it explains good reason to leave `*_validate` here

zhoonit (approved):
LGTM after CI pass

nntrainer/utils/parse_util.cpp
kparichay: removed the empty check. `none` is already part of `loss_string`.

PR #937, opened 2021-02-04
[model] Allow update batch size

### LINKED ISSUE #914: Fix changing batchsize after allocation
With #880, changing batch size after memory is allocated creates issue as #880 does not changes the tensor object but tried to reuse the same tensor object, which requires changing size of an already allocated tensor which is not allowed.

The solution is to allow changing only the batch size of a tensor, which will be supported soon.
It has been kept outside #880 to control the size of that patch.

kparichay:
Commit 1: [tensor] Allow updating batch size after allocation 
Allow updating batch size after memory allocation for tensor.
If previously allocated memory is first set to null and then allocated again
with the new batch size. As tensors share memory, the previously allocated
might not be freed when set to null and can lead to high peak memory usage.
It is recommended to free the memory for all tensors whose batch size is changing
first, update batch size and then allocate again to keep the peak memory requirement
at check.

As the memory is allowed to be re-allocated, the src_tensor for a tensor is not
reset anymore after memory allocation. When reallocating, the src_tensor is used
again in order to maintain the memory management done once duing initialization.

This patch also sets appropriate interface in Var_Grad and Weight classes as well.

Commit 2: [model] Allow updating batch size after allocation
Allow updating the batch size after memory allocation has been done
This allows changing the batch size for training anytime.

Further, it allows changing the batch size from training to inference and back
which is demonstrated with MNIST example in nnstreamer element of nntrainer.
The model file starts with batch size of 32, which is changed to 1 for inference.

Also added another unittest in cpp-api which sets different batch size in different
stages of the model, after initialization, after compile, after train and trains multiple times
with different batch sizes to validate that the training works.

Resolve #914

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/manager.h
+++ b/nntrainer/tensor/manager.h
@@ -240,9 +240,21 @@ public:
       max_derivative_size *= batch;
       max_shared_inout *= batch;
     }
+
+    /**
+     * All the tensors must be deallocated first and then allocated.
+     * Deallocating and allocating tensors one by one can potentially lead to
+     * high requirement of the peak memory requirement.
+     */
+    deallocateInOuts();
+    deallocateDerivatives();
+
     for (auto &in_out : in_outs)
       for (auto &vg : in_out)
         vg->setBatchSize(batch);
+
+    allocateInOuts();
+    allocateDerivatives();
   }
 
   /**
@@ -255,6 +267,18 @@ public:
     allocateGradients();
     allocateInOuts();
     allocateDerivatives();
+  }
+
+  /**
+   * @brief Deallocate memory for all the managed tensors
+   */
+  void deallocateTensors(bool dealloc_weights = false) {
+    if (dealloc_weights)
+      deallocateWeights();
+
+    deallocateGradients();
+    deallocateInOuts();
+    deallocateDerivatives();
   }
 
   /**
@@ -276,6 +300,29 @@ public:
    * @brief Allocate memory for all the managed layer derivatives
    */
   void allocateDerivatives();
+
+  /**
+   * @brief Deallocate memory for all the weights
+   */
+  void deallocateWeights();
+
+  /**
+   * @brief Deallocate memory for all the gradients of the weights
+   *
+   */
+  void deallocateGradients();
+
+  /**
+   * @brief Deallocate memory for all the input and output tensors
+   *
+   */
+  void deallocateInOuts();
+
+  /**
+   * @brief Deallocate memory for all the inputs and outputs derivative tensors
+   *
+   */
+  void deallocateDerivatives();
 
 private:
   // TODO: ensure that names of these weights are unique
--- a/nntrainer/tensor/tensor.h
+++ b/nntrainer/tensor/tensor.h
@@ -823,6 +823,32 @@ public:
   unsigned int width() const { return dim.width(); }
 
   /**
+   * @brief     update batch size for this tensor
+   * @param     batch size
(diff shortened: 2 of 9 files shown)


nntrainer/tensor/var_grad.h
```diff
@@ -253,6 +252,32 @@ class Var_Grad {
     resetGradient();
   }
 
+  /*
+   * @brief Allocate memory for the variable and gradient
+   */
+  void alllocate() {
+    allocateVariable();
```
jijoongmoon, line 259: alllocate --> allocate?

jijoongmoon (approved):
LGTM

nntrainer/tensor/var_grad.h
kparichay: will do.

nntrainer/tensor/tensor.h
```diff
@@ -822,6 +822,32 @@ class Tensor {
...
+   *
+   * @note      If this tensor is re-allocated, then the memory might not be
+   * immediately freed as the tensor already dependeing on this tensor also
+   * share the same memory. So, the peak memory consumption in worst case can
+   * reach the total memory requirements of a model with old batchsize and the
+   * new batch size. It is recommended to first deallocate all the tensors,
+   * updateBatch and then allocate again to avoid such issues.
+   */
+  void updateBatch(unsigned int batch) {
+    dim.batch(batch);
```
zhoonit, line 846: update stride here?

test/ccapi/unittest_ccapi.cpp
```diff
@@ -272,6 +272,75 @@ TEST(nntrainer_ccapi, train_dataset_with_generator_01_p) {
...
+
+  /** Update batch size after train */
+  EXPECT_EQ(model->setProperty({"batch_size=32"}), ML_ERROR_NONE);
+  EXPECT_NO_THROW(model->train());
+
+  /** Update batch size after train */
+  EXPECT_EQ(model->setProperty({"batch_size=4"}), ML_ERROR_NONE);
+  EXPECT_NO_THROW(model->train());
+
+  // EXPECT_FLOAT_EQ(model->getTrainingLoss(), 2.2109976);
```
zhoonit, line 339: I wonder if this comment is intended?

nntrainer/tensor/manager.h
```diff
@@ -349,6 +396,8 @@ class Manager {
    * @param[in] is_weight true if weight, else false meaning its gradient
    */
   AllocFunc getAllocFunc(bool is_weight);
+
+  void deallocateTensorsButWeights();
```
zhoonit, line 400: Could you check if this function is used / implemented? I might have missed :(

nntrainer/tensor/tensor.h
kparichay: changing batch does not change the stride. Please check `computeStrides` in `tensor_dim.h`.

test/ccapi/unittest_ccapi.cpp
kparichay: wasn't intended, updated.

nntrainer/tensor/var_grad.h
kparichay: done.

nntrainer/tensor/manager.h
kparichay: removed.

zhoonit (approved):
LGTM

PR #942, opened 2021-02-05
[meson] Set meson buildtype to release

kparichay:
Setting -O optimization manually gives meson warnings.
```
meson.build:59: WARNING: Consider using the built-in optimization level instead of using "-O2".
```

Instead update meson buildtype to release
This sets debug to 0 and optimization to level 3.

Note that this patch increases the optimization level of the library from 2 to 3.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,7 +6,8 @@ project('nntrainer', 'c', 'cpp',
     'werror=true',
     'warning_level=1',
     'c_std=gnu89',
-    'cpp_std=c++14'
+    'cpp_std=c++14',
+    'buildtype=release'
   ]
 )
 add_project_arguments('-DMIN_CPP_VERSION=201402', language:['c','cpp'])
@@ -37,7 +38,6 @@ warning_flags = [
   '-Wvla',
   '-Wpointer-arith',
   '-Wno-error=varargs',
-  '-O2',
   '-ftree-vectorize'
 ]
 


meson.build
```diff
@@ -6,7 +6,8 @@ project('nntrainer', 'c', 'cpp',
     'werror=true',
     'warning_level=1',
     'c_std=gnu89',
-    'cpp_std=c++14'
+    'cpp_std=c++14',
+    'buildtype=release'
   ]
```
jijoongmoon, line 11: any reason to change?? much faster??

jijoongmoon (approved):
LGTM

zhoonit:
Meson has buildtypes as a [builtin options](https://mesonbuild.com/Builtin-options.html) which is defaulted to be debug.
Deleting -O2 seems good, how about we relying on the meson option?

lhs8928 (approved):
LGTM

zhoonit (changes requested):
Meson has buildtype as a builtin options which is defaulted to be debug.
Deleting -O2 seems good, and for the flag, how about we relying on the meson option?

kparichay:
> Meson has buildtype as a builtin options which is defaulted to be debug.
> Deleting -O2 seems good, and for the flag, how about we relying on the meson option?

Default option of `debug` for `buildtype` keeps optimization level to be 0 and `debug` mode enabled - https://mesonbuild.com/Builtin-options.html, which IMO isn't desirable for release.
If you want more fine-grained control, `buildtype` can be `plain` and specify optimization `-o2` with meson variables.

zhoonit:
Aside from it, could you check `nntrainer.spec:279` which is building `plain`. I think it would override `buildtype=release` ?

zhoonit (approved):
>  Default option of debug for buildtype keeps optimization level to be 0 and debug mode enabled - https://mesonbuild.com/Builtin-options.html, which IMO isn't desirable for release.
If you want more fine-grained control, buildtype can be plain and specify optimization -o2 with meson variables.

I missed that it was on the default option section. Sounds reasonable

PR #949, opened 2021-02-15
[Fix] Update lazy alloc ctor to lazy alloc

zhoonit:
This patch fixes lazy alloc ctor to lazy allocate which were delegated
to eager ctor thus causing allocating tensor twice for the lazy
allocated tensor

**Self evaluation:**


jijoongmoon (approved):
LGTM

PR #955, opened 2021-02-17
[Waiting for #971] [Conv2D] Rework conv2d layer for various scenario

### LINKED ISSUE #921: Missing conv2d multistride test
There is no conv2d multistride test, we need a multistrided test to test forwarding & backwarding

zhoonit:
- [Conv2D] Rework conv2d layer
```
This patch contains major rework of conv2d layer to cope with
padded, strided configuration

**Major changes proposed in this PR**
- Rework conv2d::calcDerivative by
  - Add and exploit col2im method
- Rework conv2d::calcGradient by
  - Add dilation arugment to im2col

**Minor changes proposed in this PR:**
- Delete obsolete profiler

**Effect on performance**
- strip_pad has been removed, thus less memcpy, allocation

**Major restriction**
This patch mainly refactors the flow of how conv2d backpropagates.
There were some delegate issues still left to be handled. Those issue
can be handled on top of this patch.

- if calculated output is not an integer either on forward / backward
this will be hard error for now.
- same padding can only be applicable when kernel size is odd.

**Self evaluation:**

```
- [Test] Add conv various situation test
```
Convolutional layer when variable stride is given case was missing from
the test, the patch adds it

- basic conv for reference
- conv with same padding
- conv with multi strides
- conv with multi strides + same padding

**Self evaluation:**

```

resolves #921


nntrainer/layers/conv2d_layer.cpp
```diff
@@ -164,135 +209,47 @@ void Conv2DLayer::forwarding(bool training) {
...
-
   Tensor &derivative = net_hidden[0]->getGradientRef();
+  Tensor &input_derivative = net_input[0]->getGradientRef();
   Tensor &filter_kernel = weightAt(ConvParams::weight).getVariableRef();
 
-  std::array<unsigned int, CONV2D_DIM> same_pad;
-  same_pad[0] = kernel_size[0] - 1;
-  same_pad[1] = kernel_size[1] - 1;
-
-  /** Calculate return derivative
```
kparichay, line 201: is it intended to remove this description? or if the procedure has changed, update the description than deleting it.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -474,21 +431,41 @@ void Conv2DLayer::im2col(const Tensor &in, const TensorDim &kdim,
...
+
+  /// effective kernel height considering dilation
+  unsigned int eff_k_height = (k_height - 1) * dilation[0] + 1;
+  /// effective kernel width considering dilation
+  unsigned int eff_k_width = (k_width - 1) * dilation[1] + 1;
+
+  unsigned int out_height = (height - eff_k_height) / mstride[0] + 1;
+  unsigned int out_width = (width - eff_k_width) / mstride[1] + 1;
+
+  NNTR_THROW_IF(height < eff_k_height, std::invalid_argument)
```
kparichay, line 443: let's move these checks to `initialize()`, there is no need to perform the same check with every iteration.

nntrainer/layers/conv2d_layer.h
```diff
@@ -121,6 +121,29 @@ class Conv2DLayer : public Layer {
...
+   * @param[in] kdim kernel dimesion for define number of row
+   * @param[in] padding padding information
+   * @param[in] mstride stride value : x, y direction
+   * @param[in] dilation kernel dilation factor : x, y each
+   * @param[in] channel_mode loop with channel first
+   * @param[out] out out tensor to put, if uninitialized, allocate a new tensor
+   * and set padding
+   * @note if out is initialized tensor, setting padding is skipped.
+   */
+  static void im2col(const Tensor &in, const TensorDim &kdim,
```
kparichay, line 141: I wonder why this is public?

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -26,30 +26,71 @@
...
-void register_event() {
-  pad_profile_key = profile::Profiler::Global().registerEvent("zero_pad");
-  im2col_key = profile::Profiler::Global().registerEvent("im2col");
-  conv_gemm_profile_key =
-    profile::Profiler::Global().registerEvent("conv_gemm");
-
-  add_bias_key = profile::Profiler::Global().registerEvent("add_bias_key");
-  clean_up = profile::Profiler::Global().registerEvent("clean_up");
-  temp_key = profile::Profiler::Global().registerEvent("temp_key");
+void col2im(const Tensor &col_matrix, const TensorDim &kdim,
```
kparichay, line 31: this should be static.
Also, this is conv2d class static member.

nntrainer/layers/conv2d_layer.cpp
zhoonit: The procedure is deleted, so the description is removed

before) we calculate derivative out of conv operation
now) we calculated
col_matrix = w_trans (dot) derivative
return_derivative = col2im(col_matrix)

Now, the code itself became pretty much self-descriptive, I'll add a small comment here.

nntrainer/layers/conv2d_layer.cpp
zhoonit: This will have to move to line 126 eventually, it's more than moving this logic but needs few more checks as the function is used in two place. 
I'll do it in a separate PR instead will add a todo comment here.

nntrainer/layers/conv2d_layer.h
zhoonit: I intentionally aligned this to https://github.com/nnstreamer/nntrainer/pull/922/files#diff-96ba836ffd47c0f12c47c69ec6a55741d43c093ca1c8b35239b3039f857ab080R140-R143

Is the PR okay when I hide this function? 
If that's the case, the function doesn't need to be exposed in the header.

nntrainer/layers/conv2d_layer.cpp
zhoonit: I think there is no reason for im2col and col2im be exposed outside of this cpp except for the reason stated above.

So it is hidden inside `conv2d_layer.cpp` with anonymous namespace

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -26,30 +26,71 @@
...
+  unsigned hdilation = dilation[0];
+  unsigned wdilation = dilation[1];
+
+  /// image considering padding
+  unsigned im_eff_height = im_height + 2 * ph;
+  unsigned im_eff_width = im_width + 2 * pw;
+  image.setZero();
+
+  int h_stride_end = im_eff_height - eff_k_height - ph;
+  int w_stride_end = im_eff_width - eff_k_width - pw;
```
dongju-chae, line 62: Is it okay just converting an unsigned integer to signed one?
I think those values are supposed to be small enough but this style seems to have potential problems.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -26,30 +26,71 @@
...
+
+  unsigned col_w = 0;
+  for (int hs = -ph; hs <= h_stride_end; hs += hstride) {
```
dongju-chae, line 65: same here

nntrainer/layers/conv2d_layer.cpp
zhoonit: Thank you for the review!

Those numbers should be at maximum 1000 otherwise the algorithm won't run.
This function is in the critical path, so the indexes are optimized in a way to minimize calculation and would hard to add checks that won't likely to happen.

Although it should be better to be checked beforehand, I added an additional todo in `initialize()` to precalculate if it would overflow.

dongju-chae: Got it. Thanks for the explanation :)

nntrainer/layers/conv2d_layer.h
kparichay: That PR will need more changes than just im2col being public. For now, let's keep it private or outside the header.

nntrainer/layers/conv2d_layer.cpp
kparichay: then, for now, let's just make it a static function in this cpp.

kparichay (approved):
LGTM

nntrainer/layers/conv2d_layer.h
zhoonit: Done :)

nntrainer/layers/conv2d_layer.cpp
zhoonit: Done

kparichay:
Retry DeepCode

jijoongmoon (approved):
LGTM

Issue #974, opened 2021-03-03
Memory allocation issues


kparichay:
1. Memory allocation for a model is done again and again per run, without allocating memory from the previous run. - #976
2. Memory deallocating at the destruction of model is not done properly. - #976
3. Memory allocation must be done again at setting the batch size, and previous memory must be deallocated before next memory is allocated. - #976
4. Memory must be freed at the end of the training run.
5. Memory must not be freed at the end of inference run, but this memory must be freed before starting another training run. Also create interface to free inference run memory.
6. None of these will free weight memory for the model. This will be freed at the destruction of the model only.

kparichay:
Also, add unittests for the manager, initialize, and allocation.

kparichay:
Create another class in Manager to manage states of weights/grads/inouts/deriv as suggested by @zhoonit. This will be done in a new PR soon.

zhoonit:
Could you consider below as discussed offline a few days ago?
1. restricting the ordering of `initialize - allocate - deallocate - deinitialize` (else hard error).

kparichay:
Yes, this will be done soon. As we discussed offline earlier, 
1. the manager code will be more modularized (if possible), and check the ordering of `initialize - (allocate - deallocate) - deinitialize`
2. rename functions in weights and var_grads from `initialize` to what they actually do.

jijoongmoon:
You can close this if it is done.

kparichay:
Once #1001 is merged, https://github.com/nnstreamer/nntrainer/issues/974#issuecomment-797207164 needs to be addressed, which is about refactoring manager and weights.
As this is more towards refactoring and not for addressing any allocation issues, I have moved its tracking to #986.
This will close once #1001 is merged.

Issue #965, opened 2021-02-26
memory usage increasing with each inference


klmntoth:
After #880 each call to inference on the same model increases the memory usage. The memory used during inference is not freed after the inference is done. The used memory is not freed even after the model is deleted.

Example to reproduce: https://gist.github.com/klmntoth/cd47b6ce6830b89fe2c31b0284c13b71 replace the main.cpp in the MNIST example with this. This ends up using 318MB ram.

Running Valgrind's Massif tool shows that all of that is used up by [this](https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/tensor/tensor.cpp#L162) allocation.
Massif output: https://gist.github.com/klmntoth/77e0bac1289534a554133f302569ec45

kparichay:
Thanks for reporting the issue. I am looking into this and will make a patch for this soon.

klmntoth:
Sorry I said it wrong in my original comment, the memory is not freed when the model is deleted.
Also, Thanks for the fast reply!

kparichay:
I added #976 which resolves the memory allocation issue. The patch resolves memory being added again and again, with each inference or train run.

> Sorry I said it wrong in my original comment, the memory is not freed when the model is deleted.

I could reproduce this with the main branch latest commit. I verified with #976 that no such issue exist. The memory consumption with MNIST does not exceed 2MB on my local system for the code you shared.
I ran with a bigger model - number of filters in the MNIST model increased from 6 and 12 to 600 and 1200 to test the memory. Maximum memory after 1 run vs 1000 runs remained same (around 1.5GB).

Please let me know if #976 resolves the issues on your side. There are a couple more issues with memory allocation which are being tracked at #974. They will be resolved this week.

klmntoth:
I tried #976 and it resolved the memory problem, Thank you!

PR #975, opened 2021-03-03
[manager] Add check in manager for multiple init and allocate

kparichay:
Add a check in the manager to avoid multiple initializations of the same
variables and their allocations. This happens when inference/train is called
multiple times on the same model.

This patch adds a check inside the manager which does not re-init or allocate
if it is already initialized/allocated.

See Also #965 #974

**Self evaluation:**


kparichay:
Merged into a common PR in #976.

PR #976, opened 2021-03-03
[manager] Manage memory allocation and de-allocation

kparichay:
**Commit 1**: [manager] Add check in manager for multiple init and allocate 
Add a check in the manager to avoid multiple initializations of the same
variables and their allocations. This happens when inference/train is called
multiple times on the same model.

This patch adds a check inside the manager which does not re-init or allocate
if it is already initialized/allocated.

**Commit 2:** [manager] Support deallocation of memory
Create a proper support for deallocation of memory by the manager which
for now will be done with destruction of manager or the model itself.

This frees all the memory involved in the model for inference or training.

**Commit 3:** [manager] Check on re-initialize
Add a check on re-initialize for the ccapi unittest
As re-initialize is removed, the memory intialization is also updated
which changes the final loss values

See Also #974 #975
This limits the excessive memory consumption by inference mode as reported in #974

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -277,6 +277,7 @@ int NeuralNetwork::initialize() {
  * @brief     free layers
  */
 NeuralNetwork::~NeuralNetwork() {
+  manager.reset();
   layers.erase(layers.begin(), layers.end());
 
   if (data_buffer) {
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -134,13 +134,18 @@ Manager::Manager(bool enable_gradient_memory_opt_,
   max_derivative_size(0),
   max_shared_inout(0),
   weights_initialized(false),
+  tensors_initialized(false),
+  weights_allocated(false),
+  tensors_allocated(false),
   enable_gradient_memory_opt(enable_gradient_memory_opt_),
   enable_derivative_memory_opt(enable_derivative_memory_opt_),
   enable_activation_memory_opt(enable_activation_memory_opt_),
   enable_inference_inout_memory_opt(enable_inference_inout_memory_opt_),
   use_shared_memory(false) {}
 
-Manager::~Manager() {}
+Manager::~Manager() {
+  // TODO: deallocate everything associated
+}
 
 /**
  * @brief     Add weight to be tracked and updated with nntrainer
@@ -250,6 +255,7 @@ void Manager::initializeWeights() {
     for (auto &w : l_w) {
       Weight &weight = w.get();
       auto dim = weight.getDim();
+      /** This will allocate memory for weights right now */
       Tensor weight_prealloc = allocate_weight(dim, weight_offset);
       Tensor grad_prealloc = Tensor();
 
@@ -259,6 +265,7 @@ void Manager::initializeWeights() {
   }
 
   weights_initialized = true;
+  weights_allocated = true;
 }
 
 void Manager::allocateWeights() {
@@ -268,6 +275,8 @@ void Manager::allocateWeights() {
       weight.allocateVariable();
     }
   }
+
+  weights_allocated = true;
 }
 
 void Manager::deallocateWeights() {
@@ -547,6 +556,8 @@ void Manager::initializeTensors(bool trainable) {
     }
     use_first_last = 1 - use_first_last;
   }
+
+  tensors_initialized = true;
 }
 
 } // namespace nntrainer
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -248,6 +248,9 @@ void Manager::initializeWeights() {
     return;
   }
 
+  if (weights_initialized)
+    return;
+
   AllocFunc allocate_weight = getAllocFunc(true);
 
   unsigned int weight_offset = 0;
@@ -266,10 +269,14 @@ void Manager::initializeWeights() {
   }
 
   weights_initialized = true;
+  /** weights are allocated without delay */
   weights_allocated = true;
 }
 
 void Manager::allocateWeights() {
+  if (weights_allocated)
+    return;
+
   for (auto &l_w : weights) {
     for (auto &w : l_w) {
       Weight &weight = w.get();
@@ -491,18 +498,21 @@ void Manager::initializeTensors(bool trainable) {
   if (!weights_initialized)
     initializeWeights();
 
-  // Allocate gradients
+  if (tensors_initialized)
+    return;
+
+  // Initialize gradients
   if (trainable)
     initializeGradients();
 
-  // Allocate shared derivative memory
+  // Initialize shared derivative memory
   if (max_derivative_size > 0 && enable_activation_memory_opt && trainable)
     shared_deriv = Tensor(TensorDim({max_derivative_size}), false);
 
   // @todo Do not count memory of the input tensor of the input layer in the
   // estimate of max_shared_inout as it is not used
 
-  // Allocate shared input/output memory for inference
+  // Initialize shared input/output memory for inference
   // @note Memory for label is not allocated here as inference doesnt has label
   if (!trainable && enable_inference_inout_memory_opt)
     shared_inout = Tensor(TensorDim({max_shared_inout}), false);
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -143,9 +143,10 @@ Manager::Manager(bool enable_gradient_memory_opt_,
   enable_inference_inout_memory_opt(enable_inference_inout_memory_opt_),
   use_shared_memory(false) {}
 
-Manager::~Manager() {
-  // TODO: deallocate everything associated
-}
+/**
+ * @brief Destructor
+ */
+Manager::~Manager() { reset(); }
 
 /**
  * @brief     Add weight to be tracked and updated with nntrainer
@@ -286,6 +287,8 @@ void Manager::deallocateWeights() {
       weight.deallocateVariable();
     }
   }
+
+  weights_allocated = false;
 }
 
 void Manager::allocateGradients() {
--- a/nntrainer/tensor/manager.h
+++ b/nntrainer/tensor/manager.h
@@ -168,16 +168,30 @@ public:
 
   /**
    * @brief Reset the manager state
+   * @note The tensors assigned to the layers are not reset. They will be
+   * automatically reset once the model is initialized again.
    */
   void reset() {
-    weights.clear();
-    max_grad_size = 0;
+    deallocateTensors(true);
+    weights_allocated = false;
+    tensors_allocated = false;
+
     total_weight_size = 0;
     total_grad_size = 0;
+    max_grad_size = 0;
+    max_derivative_size = 0;
     max_shared_inout = 0;
+
     weight_mmaped_memory.reset();
     grad_mmaped_memory.reset();
+
     in_outs.clear();
+    weights.clear();
+    is_act_type.clear();
+    is_flat_type.clear();
+
+    weights_initialized = false;
+    tensors_initialized = false;
   }
 
   /**
@@ -246,15 +260,20 @@ public:
      * Deallocating and allocating tensors one by one can potentially lead to
      * high requirement of the peak memory requirement.
      */
-    deallocateInOuts();
-    deallocateDerivatives();
+
+    if (tensors_allocated) {
+      deallocateTensors();
+      deallocateDerivatives();
+    }
 
     for (auto &in_out : in_outs)
       for (auto &vg : in_out)
         vg->setBatchSize(batch);
 
-    allocateInOuts();
-    allocateDerivatives();
+    if (tensors_allocated) {
+      allocateInOuts();
+      allocateDerivatives();
+    }
   }
 
   /**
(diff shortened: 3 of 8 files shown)


nntrainer/tensor/manager.cpp
```diff
@@ -250,6 +255,7 @@ void Manager::initializeWeights() {
     for (auto &w : l_w) {
       Weight &weight = w.get();
       auto dim = weight.getDim();
+      /** This will allocate memory for weights right now */
```
jijoongmoon, line 259: we do allocation and initialization weight here? or do we call allocateWeights first before initialization? If it is second, then we need to check the weights_allocated before initialization.

nntrainer/tensor/manager.h
```diff
@@ -261,12 +261,15 @@ class Manager {
...
       allocateWeights();
-    allocateGradients();
-    allocateInOuts();
-    allocateDerivatives();
+
+    if (!tensors_allocated) {
+      allocateGradients();
+      allocateInOuts();
+      allocateDerivatives();
+      tensors_allocated = true;
```
jijoongmoon, line 290: So.. tensor means tensors for gradient and input / output and derivatives?

nntrainer/tensor/manager.cpp
```diff
@@ -286,6 +287,8 @@ void Manager::deallocateWeights() {
       weight.deallocateVariable();
     }
   }
+
+  weights_allocated = false;
```
jijoongmoon, line 291: weights_initialized = false; ??

nntrainer/tensor/manager.h
```diff
@@ -168,16 +168,30 @@ class Manager {
...
     weight_mmaped_memory.reset();
     grad_mmaped_memory.reset();
+
     in_outs.clear();
+    weights.clear();
+    is_act_type.clear();
+    is_flat_type.clear();
+
+    weights_initialized = false;
+    tensors_initialized = false;
```
jijoongmoon, line 194: I think this tensors_initialized value setting should be inside of deallocateTensors.

nntrainer/tensor/manager.h
```diff
@@ -168,16 +168,30 @@ class Manager {
...
    * @brief Reset the manager state
+   * @note The tensors assigned to the layers are not reset. They will be
+   * automatically reset once the model is initialized again.
    */
   void reset() {
-    weights.clear();
-    max_grad_size = 0;
+    deallocateTensors(true);
+    weights_allocated = false;
+    tensors_allocated = false;
```
jijoongmoon, line 177: This is duplication.

nntrainer/tensor/manager.cpp
kparichay: > we do allocation and initialization weight here? 

Yes, weights initialization and allocation are done together here.

The check is missing - added a check of weights_initialized here.

nntrainer/tensor/manager.h
kparichay: yes. Tensors required to infer/train the model.

nntrainer/tensor/manager.cpp
kparichay: This is a bit tricky. 
weight_allocated and weight_initialized variables are not synced. It is just that `initializeWeights()` does allocation as well - this can be separated as well, but there is not advantage to it and delays loading the saved model file.
Weights are still initialized (with correct configuration) after dealloc. And can be allocated again with initialization.

nntrainer/tensor/manager.h
kparichay: done.

kparichay: deallocateTensors does not de-initialize. So, IMO, tensors_initialized should not be modified inside `deallocateTensors`. Only allocation related variables must be edited.

nntrainer/tensor/manager.h
kparichay: removed.

jijoongmoon:
Could you remove unnecessary files like logs and patches?

kparichay:
> Could you remove unnecessary files like logs and patches?

my bad, removed.

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -762,9 +762,13 @@ INSTANTIATE_TEST_CASE_P(
     mkModelTc(conv_same_padding_multi_stride, "3:1:1:10", 10),
     mkModelTc(conv_no_loss_validate, "3:1:1:10", 1),
     mkModelTc(conv_none_loss_validate, "3:1:1:10", 1)
-), [](const testing::TestParamInfo<nntrainerModelTest::ParamType>& info){
```
zhoonit, line 765: I wonder if this doesn't work for you?

kparichay: I am still on Ubuntu 16.04, and libgtest is old for xenial.
I will update to 20.04 once nnstreamer build is fixed for it, and then update this.

nntrainer/tensor/manager.h
```diff
@@ -334,7 +337,11 @@ class Manager {
   unsigned int max_grad_size; /**< max trainable weight required by a layer */
   unsigned int max_derivative_size; /**< max derivative required by a layer */
   unsigned int max_shared_inout;    /**< max memory for in/outs for inference */
+
   bool weights_initialized; /**< track if weights have been initialized */
+  bool tensors_initialized; /**< track if other tensors have been initialized */
```
zhoonit, line 330: Q. what does mean by others? Could you classify by derivative + in_outs + gradients?

zhoonit: Self answered: yes

nntrainer/tensor/manager.cpp
```diff
@@ -547,6 +556,8 @@ void Manager::initializeTensors(bool trainable) {
     }
     use_first_last = 1 - use_first_last;
   }
+
+  tensors_initialized = true;
```
zhoonit, line 573: tensors_initialized includes gradients so from the invariant defined in the header. If `trainable` is false, `tensor_initialized` is nor false or true. I think it'd better to have a bit flag or equivalent to tell which one is initialized or not.

nntrainer/models/neuralnet.cpp
```diff
@@ -277,6 +277,7 @@ int NeuralNetwork::initialize() {
  * @brief     free layers
  */
 NeuralNetwork::~NeuralNetwork() {
+  manager.reset();
```
zhoonit, line 280: Was this causing the bug? manager should reset when NeuralNetwork destroys without this line.

kparichay: The bug was the missing check if the memory is already allocated.
This one just ensures that memory is released before layers are removed.

nntrainer/tensor/manager.h
```diff
@@ -276,57 +295,28 @@ class Manager {
    * @brief Deallocate memory for all the managed tensors
    */
   void deallocateTensors(bool dealloc_weights = false) {
```
zhoonit, line 295: Could you specify the semantics of `tensors`?
it seems to state one of three from time to time.
1. gradient + weights + inouts + derivatives
2. gradient + inouts + derivatives
3. inouts + derivatives.

test/unittest/unittest_nntrainer_models.cpp
zhoonit: ah I see. Thanks

nntrainer/tensor/manager.cpp
```diff
@@ -491,18 +498,21 @@ void Manager::initializeTensors(bool trainable) {
...
-  // Allocate gradients
+  if (tensors_initialized)
+    return;
+
+  // Initialize gradients
   if (trainable)
     initializeGradients();
 
-  // Allocate shared derivative memory
+  // Initialize shared derivative memory
```
zhoonit, line 508: isn't this allocate?

test/ccapi/unittest_ccapi.cpp
```diff
@@ -371,9 +371,9 @@ TEST(nntrainer_ccapi, train_batch_size_update_after) {
   EXPECT_EQ(model->setProperty({"batch_size=4"}), ML_ERROR_NONE);
   EXPECT_NO_THROW(model->train());
 
-  EXPECT_FLOAT_EQ(model->getTrainingLoss(), 1.9582682);
-  EXPECT_FLOAT_EQ(model->getValidationLoss(), 2.1831701);
-  EXPECT_FLOAT_EQ(model->getLoss(), 2.1985414);
+  EXPECT_FLOAT_EQ(model->getTrainingLoss(), 1.9613363);
```
zhoonit, line 374: As far as I can tell, initialize doesn't have reinitialize, may be because the init order changed so affect the loss?

nntrainer/tensor/manager.h
```diff
@@ -173,8 +173,6 @@ class Manager {
    */
   void reset() {
     deallocateTensors(true);
-    weights_allocated = false;
```
zhoonit: I reckon deallocateTensors changes weights_allocated to false?

kparichay: deallocateTensors() calls deallocateWeights() which sets weights_allocated to false.

nntrainer/tensor/manager.cpp
kparichay: no, this is `initialize`. Line 510 only sets `tensorDim` but does not allocate.

nntrainer/tensor/manager.h
kparichay: I will rename the function to `deallocateAllTensors()`.
Tensors will consistently refer to `gradient + inouts + derivatives`.
And all tensors will include weights as well.

nntrainer/tensor/manager.cpp
zhoonit: Ah it's false. okay

test/ccapi/unittest_ccapi.cpp
kparichay: The issue is not from the initialize but the optimizer.
In the older version, the memory was allocated again and again, which initialized the adam parameters from the scratch for each run.
With this patch, the memory is not re-allocated. And the adam optimizer parameters start from the values of the previous iteration.

nntrainer/tensor/manager.cpp
kparichay: The reason the init variables are not separated is that all 3 tensors (inouts, gradients, and derivatives) are all clubbed together in the abstraction exposed by the manager.
For internal purposes, I will add a variable to store the state of the trainable used for initialization, which will also match with the exposed abstraction provided by the manager.

kparichay: I will do this in a separate PR as this introduces some big changes, when considering about inference/train being called again and again.

nntrainer/tensor/manager.h
kparichay: Actually, deallocateTensors() is supposed to give same abstraction as allocateTensors() and initializeTensors().
which is meant for `gradient + inouts + derivatives`.
deallocateTensors() provides an additional option to deallocate weights as well (but does not by default).
allocateTensors() ensures that weights are allocated internally (intended to allocate the trio only).

Please suggest alternative if this is confusing.

nntrainer/tensor/manager.cpp
zhoonit: Sound decent.

nntrainer/tensor/manager.h
zhoonit: My concern is about the invariants seems,

InitializeTensors, might not allocates gradients (if it is not trainable).
```cpp
void func1() {
  initializeTensors(false); // initailzes weight + inout + deriv
  initializeTensors(true); // this doesn't initialize grad although it should(?) or fail because already initialized?
}

void func2() {
  initilizeTensors(false); // initialize weight + inout + deriv
  allocateTensors(); // allocates weight + inout + deriv + grad (it shouldn't allocate grad) 
  deallocateTensors(); // deallocates grad (it shouldn't deallocate grad) 
}

void func3() {
  allocateTensors(); // if it is not initialized, should this be hard error? if initialized before, is grad initialized or not?
}

void func4() {
  deallocateTensors(); // if not allocated before, Should this be an hard error?
}
```

test/ccapi/unittest_ccapi.cpp
zhoonit: I see, thanks for the explanation :)

nntrainer/tensor/manager.cpp
kparichay: Added #981.

nntrainer/tensor/manager.h
kparichay: check #981 once, if it does what you need.

> InitializeTensors, might not allocates gradients (if it is not trainable).

`InitializeTensors` will never allocate gradient, only initialize. But, if `initializeTensors` didn't initialize gradients, then gradients wont be initialized. `funcTensors` provides an abstraction which applies processing the `func` on all 3 - `gradients, inouts and derivatives`, as if the same `func` was applied on all 3 individually. In case of not trainable, `initializeGradients` would have not done anything. 
But that is now handled by `initializeTensors` and `initializeGradients` just initialize gradients if asked to. Decision making has moved to `funcTensors`.

zhoonit: I checked #981, it's orthogonal to this.

I'll get to you in person to explain.

zhoonit: Discussed offline, this will be treated with a separate PR.

zhoonit (approved):
LGTM

nntrainer/tensor/manager.h
kparichay: This will be handled in or over #981.

jijoongmoon (approved):
LGTM

PR #979, opened 2021-03-03
[model] Support freeing memory with train and inference

### LINKED ISSUE #974: Memory allocation issues

kparichay:
**Commit 1:** [neuralnet] Support deallocate of tensors 
Support deallocate and allocation of tensors from the neuralnet.

Also perform the deallocation of tensors after each train run.
Once a training is performed, the memory associated for that training
except the weights variables will be freed.

The memory associated with inference will not be freed until freed manually.
This will require calling deallocate() with the model object.
Note that calling a train() after inference() will free the inference memory
and the train will allocate its own memory which will be freed at the end
of the training.

**Commit 2:** [model] Free memory for inference 
Support freeing memory for inference with free_mem argument in inference
This defaults to true.

See Also #965  #974

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.h
+++ b/nntrainer/models/neuralnet.h
@@ -162,13 +162,20 @@ public:
   int initialize();
 
   /**
-   * @brief     Assign Graph Memory. This should be called after initialize.
-   * @TODO      Consider to move Network Graph Class
+   * @brief     Allocate memory for the model. This should be called after initialize.
    * @param[in] trainable Assign memory for inference or train mode
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
-  int assignMem(bool trainable = true);
+  int allocate(bool trainable = true);
+
+  /**
+   * @brief     Deallocate memory for the model.
+   * @param[in] trainable Assign memory for inference or train mode
+   * @retval #ML_ERROR_NONE Successful.
+   * @note This does not free the model graph but only the weight tensors, and input/output/gradient/derivative tensors if any.
+   */
+  int deallocate();
 
   /**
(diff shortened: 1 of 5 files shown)


nntrainer/models/neuralnet.h
```diff
@@ -254,7 +256,7 @@ class NeuralNetwork : public ml::train::Model {
    * @param[in] X input tensor
    * @retval shared_ptr<const Tensor>
    */
-  sharedConstTensors inference(sharedConstTensors X);
+  sharedConstTensors inference(sharedConstTensors X, bool free_mem = true);
```
jijoongmoon, line 259: If we keep using inference which is initialized, then this free_mem should be false.. then it does make sure not to allocate the memory? ( may be using tensors_allocated and weight_allocated variable?)

kparichay: > then it does make sure not to allocate the memory?

Yes, if `free_mem` is passed as false, then no new memory is allocated (using exactly tensors_allocated and weight_allocated) resulting in reduction of used bandwidth as well as faster inference. 
`free_mem` can be set to true with last inference or that memory can be freed manually with `model.deallocate()`.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #982, opened 2021-03-04
[ Layer ] Add Skeleton for Embedding Layer

jijoongmoon:
In this PR,
Skeleton Code for Embedding Layer is included.
- header / source and some modification for layer factory.

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -47,6 +47,7 @@ enum LayerType {
   LAYER_LOSS,                                    /** Loss Layer type */
   LAYER_BACKBONE_NNSTREAMER,                  /** Backbone using NNStreamer */
   LAYER_BACKBONE_TFLITE,                      /** Backbone using TFLite */
+  LAYER_EMBEDDING,                            /** Embedding Layer type */
   LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN /** Unknown */
 };
 
@@ -97,8 +98,14 @@ public:
    *            23. gamma_initializer : string (type),
    *            24. beta_initializer" : string (type)
    *            25. modelfile : model file for loading config for backbone layer
-   *            26. input_layers" : string (type)
-   *            27. output_layers" : string (type)
+   *            26. input_layers : string (type)
+   *            27. output_layers : string (type)
+   *            28. trainable :
+   *            29. flip_direction
+   *            30. random_translate
+   *            31. in_dim : int ( input dimension for embedding layer )
+   *            32. out_dim : int ( output dimesion for embedding layer )
+   *            33. in_length : int ( input length for embedding layer )
    */
   enum class PropertyType {
     input_shape = 0,
@@ -132,6 +139,9 @@ public:
     trainable = 28,
     flip_direction = 29,
     random_translate = 30,
+    in_dim = 31,
+    out_dim = 32,
+    in_length = 33,
     unknown
   };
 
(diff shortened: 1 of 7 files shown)


api/ccapi/include/layer.h
```diff
@@ -311,6 +321,14 @@ BackboneTFLite(const std::vector<std::string> &properties = {}) {
   return createLayer(LayerType::LAYER_BACKBONE_TFLITE, properties);
 }
 
+/**
+ * @brief Helper function to create Embedding layer
+ */
+inline std::unique_ptr<Layer>
+Embedding(const std::vector<std::string> &properties = {}) {
```
zhoonit, line 328: Because this is a default layer, how about registering explicitly from `app_context.cpp@add_default_object`.

nntrainer/layers/embedding.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <layer_internal.h>
+#include <lazy_tensor.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+const std::string EmbeddingLayer::type = "embedding_layer";
```
zhoonit, line 24: how about setting type name as "embedding" to keep the convention consistent?

zhoonit (approved):
LGTM :)

api/ccapi/include/layer.h
```diff
@@ -97,8 +98,14 @@ class Layer {
...
    *            24. beta_initializer" : string (type)
    *            25. modelfile : model file for loading config for backbone layer
-   *            26. input_layers" : string (type)
-   *            27. output_layers" : string (type)
+   *            26. input_layers : string (type)
+   *            27. output_layers : string (type)
+   *            28. trainable :
+   *            29. flip_direction
+   *            30. random_translate
+   *            31. in_dim : int ( input dimension for embedding layer )
```
kparichay, line 106: i wonder how `in_dim` is different from `input_shape`?

nntrainer/layers/embedding.h
```diff
@@ -0,0 +1,109 @@
...
+#include <layer_internal.h>
+#include <tensor.h>
+
+namespace nntrainer {
+
+/**
+ * @class   EmbeddingLayer
+ * @brief   EmbeddingLayer
+ */
+class EmbeddingLayer : public Layer {
```
kparichay, line 27: This is not related to this PR, but if this layer's implementation is close to FC layer, you consider extending `FullyConnectedLayer` than `Layer` class in the future PR.

api/ccapi/include/layer.h
jijoongmoon: done.

nntrainer/layers/embedding.cpp
jijoongmoon: done

api/ccapi/include/layer.h
jijoongmoon: yes. it differs. in_dim is vocab size. and the input size is related with input length. and the out dim is the size of embedding space.

nntrainer/layers/embedding.h
jijoongmoon: I thonght that too.. but I guess It is easier starting from layer.

api/ccapi/include/layer.h
kparichay: Then how about rename in_dim to vocab_size or add in comments?

> the out dim is the size of embedding space.

Is out_dim supposed to match with output_dimension of the layer? Please add such invariants in the comments of this property.

nntrainer/layers/embedding.h
kparichay: Yes, extend to FCLayer is for later.

api/ccapi/include/layer.h
jijoongmoon: as you can see, those number is just unsigned integer. not the dim. output dimension must be composed with batch, in_length, out_dim.

kparichay: ok.

kparichay (approved):
LGTM

PR #984, opened 2021-03-08
[ LAYER ] Embedding Layer Inference

jijoongmoon:
This PR includes inference calculation of Embedding Layer.
- Implementation of Embedding Layer
- Few test cases for inference. ( More test will be added )

**Self evaluation:**


nntrainer/layers/embedding.cpp
```diff
@@ -23,10 +23,36 @@ namespace nntrainer {
...
+
+  output_dim[0].height(in_length);
+  output_dim[0].width(out_dim);
+  input_dim[0].width(in_length);
+
+  TensorDim dim = output_dim[0];
+
+  dim.height(in_dim);
+  dim.width(out_dim);
+  dim.batch(1);
```
zhoonit, line 44: I have a small code clean suggestion :)
 `TensorDim dim({in_dim, out_dim});` or `TensorDim dim({output_dim.channel(), in_dim, out_dim});`

zhoonit: I wonder if `channel() != 1` it should be error or okay?

nntrainer/layers/embedding.cpp
```diff
@@ -64,7 +90,31 @@ void EmbeddingLayer::setProperty(const PropertyType type,
 }
 
 void EmbeddingLayer::forwarding(bool training) {
-  // NYI
+  Tensor &weight =
+    weightAt(static_cast<int>(EmbeddingParams::weight)).getVariableRef();
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  for (unsigned int b = 0; b < input_.getDim().batch(); ++b) {
```
zhoonit, line 98: Minor code suggestion :)
```suggestion
  for (unsigned int b = 0; b < input_.batch(); ++b) {
```

nntrainer/layers/embedding.cpp
```diff
@@ -64,7 +90,31 @@ void EmbeddingLayer::setProperty(const PropertyType type,
...
+    Tensor in_sub = input_.getBatchSlice(b, 1);
+    Tensor out_sub = hidden_.getBatchSlice(b, 1);
+    float *in_data = in_sub.getData();
+
+    for (unsigned int i = 0; i < in_length; ++i) {
+      if (in_data[i] < 0)
```
zhoonit, line 104: I wonder if `in_data[i] < 0` is acceptable or error?

kparichay, line 104: add a check that `in_data[i]` is an unsigned integer? based on its usage in line 108.

nntrainer/layers/embedding.cpp
```diff
@@ -64,7 +90,31 @@ void EmbeddingLayer::setProperty(const PropertyType type,
...
+        continue;
+
+      float *weight_data =
+        &(weight.getData()[static_cast<uint>(in_data[i]) * out_dim]);
```
kparichay, line 108: I recommend to `weight.getAddress()` with corresponding `b,c,h,w`. 
This would remove the need for `getBatchSlice` and simplify the code.

nntrainer/layers/embedding.cpp
```diff
@@ -64,7 +90,31 @@ void EmbeddingLayer::setProperty(const PropertyType type,
...
+      float *out_data = &(out_sub.getData()[i * out_dim]);
+      for (unsigned int j = 0; j < out_dim; ++j) {
```
kparichay, line 110: use `memcpy` or `scopy` instead of this loop?

nntrainer/layers/embedding.cpp
kparichay: also `in_data[i]` must be less than `in_dim`?

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -2184,6 +2185,54 @@ TEST_F(nntrainer_AdditionLayer, DISABLED_forwarding_03_p) {
...
+
+  nntrainer::Manager manager;
+
+  manager.setInferenceInOutMemoryOptimization(false);
+  layer.setInputBuffers(manager.trackLayerInputs(
+    layer.getType(), layer.getName(), layer.getInputDimension()));
+  layer.setOutputBuffers(manager.trackLayerOutputs(
+    layer.getType(), layer.getName(), layer.getOutputDimension()));
+
+  manager.initializeTensors(true);
```
kparichay, line 2230: recommended using `initializeTensors(False)` is only using `forwarding`.

nntrainer/layers/embedding.cpp
jijoongmoon: already done in #988

nntrainer/layers/embedding.cpp
jijoongmoon: if it is less then 0.. then it just skipped I guess. It is just padding. In this version, we assume it starts from 0. 
it is modified #988

nntrainer/layers/embedding.cpp
jijoongmoon: for now, channel need to be 1. I will add check.

nntrainer/layers/embedding.cpp
jijoongmoon: done.

nntrainer/layers/embedding.cpp
jijoongmoon: I will add check.

nntrainer/layers/embedding.cpp
jijoongmoon: I think getAddress with some calculations of index might be messy and drop the readability.  I think either one of them is ok.

nntrainer/layers/embedding.cpp
zhoonit: I see thank you!

zhoonit (approved):
LGTM after issues are resolved :)

kparichay (approved):
LGTM

PR #985, opened 2021-03-08
Fix/svace issue 210308

lhs8928:
Handle NO_CATCH issues

Added missing try-catch statement
resolve 457555, 457556, 457557, 457558

### DIFF
--- a/Applications/SimpleShot/task_runner.cpp
+++ b/Applications/SimpleShot/task_runner.cpp
@@ -256,8 +256,13 @@ int main(int argc, char **argv) {
     return 1;
   };
 
-  std::shared_ptr<ml::train::Optimizer> optimizer =
-    ml::train::optimizer::SGD({"learning_rate=0.1"});
+  std::shared_ptr<ml::train::Optimizer> optimizer;
+  try {
+    optimizer = ml::train::optimizer::SGD({"learning_rate=0.1"});
+  } catch (...) {
+    std::cerr << "creating optimizer failed";
+    return 1;
+  }
 
   if (model->setOptimizer(optimizer) != 0) {
     std::cerr << "failed to set optimizer" << std::endl;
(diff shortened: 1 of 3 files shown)


nntrainer/layers/conv2d_layer.cpp
```diff
@@ -143,13 +143,13 @@ static void im2col(const Tensor &in, const TensorDim &kdim,
   unsigned int out_height = (height - eff_k_height) / mstride[0] + 1;
   unsigned int out_width = (width - eff_k_width) / mstride[1] + 1;
 
-  NNTR_THROW_IF(height < eff_k_height, std::invalid_argument)
+  NNTR_THROW_IF(height < eff_k_height + ph, std::invalid_argument)
```
zhoonit, line 146: Adding padding to eff_k_height make this statement meaningless. I wonder what this is trying to solve?

lhs8928: There is an UNDERFLOW svace issue at line 187 (457562, 457563)
`
    int h_stride_end = height - eff_k_height - ph;
    int w_stride_end = width - eff_k_width - pw;
`
which means height not just greater than eff_k_height 
but should be greater than eff_k_height + ph.

zhoonit: What you are trying to solve is actually a false positive.

please look at the 457562 case for example.

1. the static analyzer is assuming that 'height' is < 'eff_k_height' (which is wrong)
2. because `height - eff_k_height > 0`, `h_stride_end - height - eff_k_height - ph > -ph` 
3. so, it means that if `UINT_MAX > ph > INT_MAX`, it will underflow
5 when `UINT_MAX > ph > INT_MAX`, let alone this algorithm won't run at this scale, `unsigned int height = in_height + ph * 2` will overflow before reaching this line.
4. so the problem reduces to if we have to check every overflow when we add two unsigned int together when it is never likely to happen. 

plus, consider this case

in_height = 3
ph = 1
eff_k_height = 4,

this case has to run but `height < eff_k_height + ph` will effectively throw

lhs8928: Understood.

kparichay (approved):
LGTM

zhoonit (approved):
LGTM

again4you (approved):
Nice catch!

PR #988, opened 2021-03-09
[ LAYER ] Add backwarding of Embedding Layer

jijoongmoon:
This PR includes
 - Backwarding implementation of Embedding Layer
 - Add Test Cases ( More Test Cases will be added )

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/embedding.cpp
+++ b/nntrainer/layers/embedding.cpp
@@ -104,17 +104,21 @@ void EmbeddingLayer::forwarding(bool training) {
     float *in_data = input_.getAddress(b * input_.getDim().getFeatureLen());
 
     for (unsigned int i = 0; i < in_length; ++i) {
-      if (in_data[i] < 0)
+      if (in_data[i] > in_dim) {
+        throw std::invalid_argument("input word index is greater than in_dim");
+      }
+
+      // Assume padding is 0 and index always start from 1.
+      // If in_data[i] - 1 < 0, then it skips.
+      if (in_data[i] - 1 < 0)
         continue;
 
       float *weight_data =
-        weight.getAddress(static_cast<uint>(in_data[i]) * out_dim);
+        weight.getAddress(static_cast<uint>(in_data[i] - 1) * out_dim);
       float *out_data =
         hidden_.getAddress(b * hidden_.getDim().getFeatureLen() + i * out_dim);
 
-      for (unsigned int j = 0; j < out_dim; ++j) {
-        out_data[j] = weight_data[j];
-      }
+      std::copy(weight_data, weight_data + out_dim, out_data);
     }
   }
 
@@ -133,10 +137,43 @@ void EmbeddingLayer::copy(std::shared_ptr<Layer> l) {
 }
 
 void EmbeddingLayer::calcDerivative() {
-  // NYI
+  // Uncomment this after fixing issues backwarding of first layer. (Issues
+  // #1017)
+  // throw exception::not_supported(
+  //   "calcDerivative for Embedding layer is not supported");
+  return; // intended
 }
+
 void EmbeddingLayer::calcGradient() {
-  // NYI
+  Tensor &djdw =
+    weightAt(static_cast<int>(EmbeddingParams::weight)).getGradientRef();
+  Tensor &derivative_ = net_hidden[0]->getGradientRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  djdw.setZero();
+
+  // TODO:
+  // This is to calculate gradient with current implementation of optimizer.
+  // In order to accelerate, we need to better way like using index to weight.
+
+  for (unsigned int b = 0; b < input_.batch(); ++b) {
+    float *in_data = input_.getAddress(b * input_.getDim().getFeatureLen());
+
+    for (unsigned int i = 0; i < in_length; ++i) {
+      // Assume padding is 0 and index always start from 1.
+      // If in_data[i] - 1 < 0, then it skips.
+      if (in_data[i] - 1 < 0)
+        continue;
+
+      float *djdw_data =
+        djdw.getAddress(static_cast<uint>(in_data[i] - 1) * out_dim);
+      float *grad_data = derivative_.getAddress(
+        b * derivative_.getDim().getFeatureLen() + i * out_dim);
+
+      std::transform(djdw_data, djdw_data + out_dim, grad_data, djdw_data,
+                     std::plus<float>());
+    }
+  }
 }
 
 } // namespace nntrainer
--- a/test/unittest/unittest_nntrainer_layers.cpp
+++ b/test/unittest/unittest_nntrainer_layers.cpp
@@ -2252,6 +2252,46 @@ TEST_F(nntrainer_EmbeddingLayer, forwarding_01_p) {
   EXPECT_NO_THROW(layer.forwarding_with_val({input}));
 }
 
+TEST_F(nntrainer_EmbeddingLayer, forwarding_backwarding_01_p) {
+  float sentence[36] = {45, 16, 32, 27, 34, 33, 0,  0,  0, 0, 0,  0,
+                        24, 2,  27, 34, 33, 37, 32, 27, 3, 0, 0,  0,
+                        22, 27, 16, 28, 35, 33, 7,  2,  2, 3, 33, 35};
+
+  sharedTensor input = std::shared_ptr<nntrainer::Tensor>(
+    new nntrainer::Tensor[1], std::default_delete<nntrainer::Tensor[]>());
+
+  nntrainer::Tensor &in = *input;
+
+  in = nntrainer::Tensor(nntrainer::TensorDim(3, 1, 1, 12), sentence);
+
+  nntrainer::Manager manager;
+
+  manager.setInferenceInOutMemoryOptimization(false);
+  layer.setInputBuffers(manager.trackLayerInputs(
+    layer.getType(), layer.getName(), layer.getInputDimension()));
+  layer.setOutputBuffers(manager.trackLayerOutputs(
+    layer.getType(), layer.getName(), layer.getOutputDimension()));
+
+  manager.initializeTensors(true);
+  manager.allocateTensors();
+
+  EXPECT_NO_THROW(layer.forwarding_with_val({input}));
+
+  nntrainer::Tensor derivatives(3, 1, 12, 8);
+
+  for (unsigned int i = 0; i < derivatives.getDim().getDataLen(); ++i) {
+    derivatives.getData()[i] = 1.0;
+  }
+
+  setOptimizer(nntrainer::OptType::ADAM, "learning_rate=1.0");
+
+  allocateMemory();
+
+  nntrainer::Tensor result;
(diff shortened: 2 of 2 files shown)


nntrainer/layers/embedding.cpp
```diff
@@ -104,17 +104,19 @@ void EmbeddingLayer::forwarding(bool training) {
     float *in_data = input_.getAddress(b * input_.getDim().getFeatureLen());
 
     for (unsigned int i = 0; i < in_length; ++i) {
-      if (in_data[i] < 0)
+      if (in_data[i] > in_dim) {
```
zhoonit, line 107: if (`in_data[i] == in_dim`) is allowed, I wonder if `in_data[i] == 0` is defined in this case? for keras, only positive integer seems allowed.

nntrainer/layers/embedding.cpp
```diff
@@ -104,17 +104,19 @@ void EmbeddingLayer::forwarding(bool training) {
...
+        throw std::invalid_argument("input word index is greater then in_dim");
```
zhoonit, line 108:
```suggestion
        throw std::invalid_argument("input word index is greater than in_dim");
```

nntrainer/layers/embedding.cpp
```diff
@@ -104,17 +104,19 @@ void EmbeddingLayer::forwarding(bool training) {
...
+      }
+
+      if (in_data[i] - 1 < 0)
```
zhoonit, line 111: if this is not a hard error, I guess we have to fill the output to zero?

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -2260,6 +2260,46 @@ TEST_F(nntrainer_EmbeddingLayer, forwarding_01_p) {
...
+  for (unsigned int i = 0; i < derivatives.getDim().getDataLen(); ++i) {
+    derivatives.getData()[i] = 1.0;
+  }
+
+  setOptimizer(nntrainer::OptType::ADAM, "learning_rate=1.0");
+
+  allocateMemory();
+
+  nntrainer::Tensor result;
+  EXPECT_NO_THROW(result = *layer.backwarding_with_val(
```
zhoonit, line 2299: Is the test matching  numbers? if not could you please left a comment?

zhoonit (approved):
LGTM

nntrainer/layers/embedding.cpp
kparichay: Does the indexing of data start from 0 or 1? If 0, then `in_data[i] == 0` is valid.

kparichay, line 111: Does the indexing of data start from 0 or 1? If 0, then `in_data[i] == 0` is valid.

nntrainer/layers/embedding.cpp
```diff
@@ -133,10 +135,38 @@ void EmbeddingLayer::copy(std::shared_ptr<Layer> l) {
 }
 
 void EmbeddingLayer::calcDerivative() {
-  // NYI
+  return; // intended
```
kparichay, line 138: It seems like embedding layer does not pass the derivative back. If yes, then please throw here.

nntrainer/layers/embedding.cpp
```diff
@@ -133,10 +135,38 @@ void EmbeddingLayer::copy(std::shared_ptr<Layer> l) {
...
+
+  // TODO:
+  // This is to calculate gradient with current implementation of optimizer.
+  // In order to accelerate, we need to better way like using index to weight.
+
+  for (unsigned int b = 0; b < input_.batch(); ++b) {
+    float *in_data = input_.getAddress(b * input_.getDim().getFeatureLen());
+
+    for (unsigned int i = 0; i < in_length; ++i) {
+      if (in_data[i] - 1 < 0)
```
kparichay, line 157: This is not needed. This check was already verified in `forwarding`.

nntrainer/layers/embedding.cpp
```diff
@@ -133,10 +135,38 @@ void EmbeddingLayer::copy(std::shared_ptr<Layer> l) {
...
+        continue;
+
+      float *djdw_data =
+        djdw.getAddress(static_cast<uint>(in_data[i] - 1) * out_dim);
+      float *grad_data = derivative_.getAddress(
+        b * derivative_.getDim().getFeatureLen() + i * out_dim);
+
+      for (unsigned int j = 0; j < out_dim; ++j) {
```
kparichay, line 165: please use memset here instead of a loop.

nntrainer/layers/embedding.cpp
```diff
@@ -133,10 +135,38 @@ void EmbeddingLayer::copy(std::shared_ptr<Layer> l) {
...
+  Tensor &djdw =
+    weightAt(static_cast<int>(EmbeddingParams::weight)).getGradientRef();
+  Tensor &derivative_ = net_hidden[0]->getGradientRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  djdw.setZero();
+
+  // TODO:
+  // This is to calculate gradient with current implementation of optimizer.
+  // In order to accelerate, we need to better way like using index to weight.
```
kparichay, line 151: :100:

nntrainer/layers/embedding.cpp
jijoongmoon: done.

nntrainer/layers/embedding.cpp
jijoongmoon: It must be skipped.  It is just padding. No need to calculate.

jijoongmoon: It is padding and padding could be less or equal zero.

nntrainer/layers/embedding.cpp
jijoongmoon: No. It need to be update the gradient. so.. It is not like input layer. Input Layer does not need to do backwarding. But we do not have detail control about it. This need another change but.. for not it is ok as it is. ( You can see line 353 in neuralnet.cpp )

nntrainer/layers/embedding.cpp
jijoongmoon: It is needed. it is for the skipping of calculation.

nntrainer/layers/embedding.cpp
jijoongmoon: It is not assign. It addition.

test/unittest/unittest_nntrainer_layers.cpp
jijoongmoon: I will add another PR for test number after PR#1002 and #1004

nntrainer/layers/embedding.cpp
kparichay: `in_data` is the input data. I wonder how is it padding? can you please elaborate on the padding part?
I also wonder why the embedding layer needs padding.

nntrainer/layers/embedding.cpp
kparichay: This layer calculates the derivative for the next layer in backpropagation.
In order for the network to run correctly, either this layer must give back the derivative or throw to ensure that the layer behind it (if any) does not receive any garbage gradient.
Just a simple return invalidates the invariant of the network training.

nntrainer/layers/embedding.cpp
kparichay: Can you please add in the comments the scenario where in_data[i] is less than 1?
Also, can you write the semantics of in_data - that it must be an unsigned integer etc.

nntrainer/layers/embedding.cpp
kparichay: then please use something like `std::transform (djdw.begin(), djdw.end(), grad_data.begin(), djdw.begin(), std::plus<float>());`

nntrainer/layers/embedding.cpp
jijoongmoon: well.. the words is indexed as number. and the input is sentence. the maximum size of words is in_length.. and sometimes it is shorter than this. when it is short.. we need to fill out with padding. 
' Well done!' could be [ 1, 2, 0, 0] with in_length is 4.

nntrainer/layers/embedding.cpp
jijoongmoon: I will leave another issue for this. but not this pr.

nntrainer/layers/embedding.cpp
jijoongmoon: does it faster than this? if it is, let me know.

nntrainer/layers/embedding.cpp
jijoongmoon: Like I describe earlier, it is padding.

nntrainer/layers/embedding.cpp
kparichay: Then how making this check `in_data[i] == 0`?

nntrainer/layers/embedding.cpp
kparichay: yes, it is faster with `std::transform()` as it is easier for compile to optimize these, and from c++17, allows parallel computation with transform.

nntrainer/layers/embedding.cpp
kparichay: Please raise a issue for this :)

nntrainer/layers/embedding.cpp
jijoongmoon: It also can be -1.

kparichay: Can you please add the possible values and what they denote as comments? BTW, what does -1 denote?

nntrainer/layers/embedding.cpp
jijoongmoon: done

nntrainer/layers/embedding.cpp
jijoongmoon: I guess It depends how user set. As you know, index starts 0 or 1. right? If index starts 0, then it will -1. if it starts from 1, the padding will be 0. Nothing more.

nntrainer/layers/embedding.cpp
```diff
@@ -133,10 +135,41 @@ void EmbeddingLayer::copy(std::shared_ptr<Layer> l) {
...
+
+  for (unsigned int b = 0; b < input_.batch(); ++b) {
+    float *in_data = input_.getAddress(b * input_.getDim().getFeatureLen());
+
+    for (unsigned int i = 0; i < in_length; ++i) {
+      if (in_data[i] - 1 < 0)
+        continue;
```
lhs8928, line 162: Cause if padding is exist then it will be 0 until the end of in_data, so I think break will be better than continue.

nntrainer/layers/embedding.cpp
kparichay: Then this condition will come out wrong if the actual index starts from 0, as this condition discards them too.

lhs8928 (approved):
LGTM

nntrainer/layers/embedding.cpp
kparichay: when using break, you also need to think about masking the inputs for future.

kparichay (approved):
LGTM

Issue #888, opened 2021-01-26
`nntrainer::forwarding(input, label, training)` need attention


zhoonit:
1. `nntrainer::forwarding` is supposed **to be able to take less than batchsize** input, however, as the network graph is compiled beforehand, it does not have fixed value anymore (see validateInput(sharedConstTensors X))
   1. input, label reassigns the memory slot

This is one of major issue that deepq is failing

kparichay:
> nntrainer::forwarding is supposed to be able to take less than batchsize input, however, as the network graph is compiled beforehand, it does not have fixed value anymore (see validateInput(sharedConstTensors X))

BatchSize is a value which takes crucial role in the calculation. Providing partially filled in the batches will result in wrong results. For example, consider batch normalization layer or loss layer (where result is directly dependent on the batch size). To handle this, `inference/train` sets the appropriate batch size before calling the `forward/backward`. `validateInput()` method is called after the batch size is update based on the batchsize provided as input and memory allocated updated as per the batch size.

> input, label reassigns the memory slot

For training, input, label memory must not be allocated as well, and must be received from the databuffer. This will be done as a part of #303.
This replacement must happen inference. The bug is that the manager must not allocate memory for them for inference. I will take a look at this.

kparichay:
> nntrainer::forwarding is supposed to be able to take less than batchsize input

Inference gives the abstraction of being able to take variable size batch size input at the cost of memory allocations/deallocations with each run (batch size inside the layers must be updated as explained in the comment for the correct functioning). This can be optimized by hinting the optimizer to not keep memory allocated for bigger batch size and reuse it for smaller batch size but I am not sure how frequent is this scenario.

Issue #986, opened 2021-03-08
C++ API and architecture refactor


kparichay:
This issue is to track the re-factor of the exposed headers and the C++ API.
This will update the NNTrainer architecture to the class diagram finalized with @jijoongmoon.

- [x] Update internal implementation to use only the methods exposed by those headers and not an internal implementation
  - [x] Remove all friends of layers
  - [x] Manage input/output properties better for layer - track at #1454
  - [x] add unittests for ensureName related  to #1009
  - [x] refactor graph to move graph node related members from layers
    - [x] #1029 
    - [x] consider removing adj lists once compile is done to save some space. - #1049
    - [x] move layers connection related properties out to graph - #1142 #1182
  - [x] tensor constructor to support initialization - https://github.com/nnstreamer/nntrainer/pull/1031#discussion_r597551354 - pointed out by @myungjoo - #1452
  - [x] refactor model loader to allow custom properties #1032 
  - [x] prepare custom loss #1256
  - [x] many more
- [x] Update headers and their interfaces
  - [x] Make c++ api equivalent to c API - done with #1050
  - [x] C++ API must be superset of https://github.com/nnstreamer/nntrainer/pull/1010#discussion_r594866783 or the nntrainer filter must be updated to accommodate with the API. - #1461
  - [x] Make c api directly use c++ API - #1050
  - [x] divide `*_internal` to `*_devel` which will be just an interface and `*_impl` which will be the default implementation. For example, Layer <- Layer_devel <- Layer_impl. - #1426
- [ ] Update packaging and applications correspondingly #1450

PR #990, opened 2021-03-09
[Manager] Remove alloc for first/last layer during inference

kparichay:
Remove the allocation of input of the first layer and label of the last
layer during inference. This is because those tensors are overridden at
the call to inference by the tensors provided externally by the user.
Note that the output of the last layer will still be allocated by the manager.

Also refactor initializeTensors() in manager into two functions
separated for inference and training, along with a couple of bug fixes inside them.

See Also #974 #888

**Self evaluation:**

cc. @zhoonit Please check if this resolves the second issue in #888.

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -491,21 +491,14 @@ void Manager::deallocateDerivatives() {
   }
 }
 
-void Manager::initializeTensors() {
-  // Initialize gradients
-  if (model_training)
-    initializeGradients();
-
-  // Initialize shared derivative memory
-  if (max_derivative_size > 0 && enable_activation_memory_opt && model_training)
-    shared_deriv = Tensor(TensorDim({max_derivative_size}), false);
-
-  // @todo Do not count memory of the input tensor of the input layer in the
-  // estimate of max_shared_inout as it is not used
+void Manager::initializeTensorsInference() {
+  // @todo Do not count memory of the input tensor of the input layer and
+  // output tensor of the last layer in the estimate of max_shared_inout as it
+  // is not used
 
   // Initialize shared input/output memory for inference
   // @note Memory for label is not allocated here as inference doesnt has label
-  if (!model_training && enable_inference_inout_memory_opt)
+  if (enable_inference_inout_memory_opt)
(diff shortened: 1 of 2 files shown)


zhoonit:
This is not apparently solving the problem around deepq. I think we need some solid baseline how forwarding, backwarding works beforehand. Some of the logics are tangled and prone to fail.

kparichay:
> This is not apparently solving the problem around deepq.

DeepQ application has many other issues. From the first look, the mainNet and targetNet do not allocate memory for the internal layers (it directly calls forwarding), the batch size for the model is not set, etc. Solving the issues in DeepQ requires debugging the written application which is beyond the scope of this PR. 

This PR intends to remove the unnecessary allocation of inputs/labels inside the manager for a model for inference mode. This will not solve the DeepQ problems as there other issues in the application itself as mentioned above.

> I think we need some solid baseline how forwarding, backwarding works beforehand. Some of the logics are tangled and prone to fail.

`forwarding` and `backwarding` are not supposed to be exposed as the API. The application should call inference. Calling forwarding means handling setting batch size appropriately and taking care of memory management in the application itself (which will have to be done in DeepQ if `forwarding` is to be used directly.)

zhoonit:
> forwarding and backwarding are not supposed to be exposed as the API.

If we are to allow users to make, something complex training flow, don't we need to expose the forwarding, backwarding as an api?

kparichay:
> If we are to allow users to make, something complex training flow, don't we need to expose the forwarding, backwarding as an api?

IMO a better way would be to create a customLayer/customModel which does the complex training part. Opening forwarding/backwarding means that all the inits/allocations/batchsize verification/etc would have to be done by user, inducing a lot of bugs. 

Creating customModel is not yet possible but will be after #986 is done to some extent.

kparichay:
> I think we need some solid baseline how forwarding, backwarding works beforehand. Some of the logics are tangled and prone to fail.

Please make some PRs for this, as this PR and others related to #974 are focused towards manager managing the memory and not related for forwarding/backwarding related issues.

zhoonit:
True that, forwarding, backwarding issue is a bit orthogonal to this.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #996, opened 2021-03-09
[model] Update batch size check

kparichay:
Update batch size check to exactly match with the model
This is because batch size is used by certain layers for calculation and
running an input with wrong batch (with correct memory allocations) can result
in wrong output.

This is independent of the running a batch size 16 with memory allocated for
bigger batch size. That will still work with this patch as long as batch size
is set correctly for the model, and memory allocated is at least big enough
to handle this batch size.

See also #888

**Self evaluation:**

cc. @zhoonit thanks for making #888 clearer in person.


zhoonit (approved):
LGTM :)

jijoongmoon (approved):
LGTM

PR #1001, opened 2021-03-10
[weights] Bug fix + lazy allocation

### LINKED ISSUE #974: Memory allocation issues

kparichay:
Commit 1: [layers] Weight creation bug fix
The name of the weight being passed as string was being interpreted as a string
for the variable alloc_now. This patch fixes it by passing the arguments
appropriately.

Commit 2: [Weights] Split weight variable init and allocSplit the initialization and memory allocation for weights
3 exposed bugs with this has been resolved:
- manager does not allow tracking var_grads once initialized
- resetGradient confirms the allocation is done before accessing the memory
- allocateVariable() calls the correct initializer now

Further, the logic of reinitialize in unittest for layers has been
split into two parts - initialize and reinitialize where
reinitialize will reset layer and manager and then call initialize.

Resolves #974

### DIFF
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -63,15 +63,17 @@ int BatchNormalizationLayer::initialize(Manager &manager) {
   if (weights.empty()) {
     weights.reserve(4);
     weights.emplace_back(dim, initializers[BNParams::mu],
-                         WeightRegularizer::NONE, 1.0f, false, true,
+                         WeightRegularizer::NONE, 1.0f, false, false,
                          "BN::moving_mean");
     weights.emplace_back(dim, initializers[BNParams::var],
-                         WeightRegularizer::NONE, 1.0f, false, true,
+                         WeightRegularizer::NONE, 1.0f, false, false,
                          "BN::moving_variance");
     weights.emplace_back(dim, initializers[BNParams::gamma],
-                         WeightRegularizer::NONE, 1.0f, true, true, "BN::gamma");
+                         WeightRegularizer::NONE, 1.0f, true, false,
+                         "BN::gamma");
     weights.emplace_back(dim, initializers[BNParams::beta],
-                         WeightRegularizer::NONE, 1.0f, true, true, "BN::beta");
+                         WeightRegularizer::NONE, 1.0f, true, false,
+                         "BN::beta");
     manager.trackWeights(weights);
   } else {
     weights[BNParams::mu].reset(dim, initializers[BNParams::mu],
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -63,15 +63,15 @@ int BatchNormalizationLayer::initialize(Manager &manager) {
   if (weights.empty()) {
     weights.reserve(4);
     weights.emplace_back(dim, initializers[BNParams::mu],
-                         WeightRegularizer::NONE, 1.0f, false,
+                         WeightRegularizer::NONE, 1.0f, false, true,
                          "BN::moving_mean");
     weights.emplace_back(dim, initializers[BNParams::var],
-                         WeightRegularizer::NONE, 1.0f, false,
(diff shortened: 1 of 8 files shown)


nntrainer/layers/bn_layer.cpp
```diff
@@ -63,15 +63,15 @@ int BatchNormalizationLayer::initialize(Manager &manager) {
   if (weights.empty()) {
     weights.reserve(4);
     weights.emplace_back(dim, initializers[BNParams::mu],
-                         WeightRegularizer::NONE, 1.0f, false,
+                         WeightRegularizer::NONE, 1.0f, false, true,
```
zhoonit: Nice catch! how about making weight ctor explicit to prevent further mistake?
```diff
- Weight(
+ explicit Weight(f
    const TensorDim &dim,
    const WeightInitializer init = WeightInitializer::WEIGHT_XAVIER_UNIFORM,
    const WeightRegularizer reg = WeightRegularizer::NONE,
    const float reg_const = 1.0f, bool train = true, bool alloc_now = true,
    std::string name = "");
```

zhoonit (approved):
LGTM with minor comments :-)

meson.build
```diff
@@ -7,7 +7,7 @@ project('nntrainer', 'c', 'cpp',
     'warning_level=1',
     'c_std=gnu89',
     'cpp_std=c++14',
-    'buildtype=release'
+    'buildtype=debug'
```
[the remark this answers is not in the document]
kparichay: not intended. will remove.

nntrainer/layers/bn_layer.cpp
kparichay: added for weight and var_grad.

nntrainer/tensor/weight.h
```diff
@@ -79,11 +79,11 @@ class Weight : public Var_Grad {
    * @param train If the variable is trainable
    * @param name Name for this weight
    */
-  Weight(
+  explicit Weight(
     const TensorDim &dim,
     const WeightInitializer init = WeightInitializer::WEIGHT_XAVIER_UNIFORM,
     const WeightRegularizer reg = WeightRegularizer::NONE,
-    const float reg_const = 1.0f, bool train = true, bool alloc_now = true,
+    const float reg_const = 1.0f, bool train = true, bool alloc_now = false,
```
jijoongmoon, line 86: . add brief @param about alloc_now. 
. I wonder the reason of false default value is because to track the weight in manager??
  Could you elaborate why manager does not allow manager to track initialized var_grads?
  and so.. we cannot allocate buffer memory during the initialization of layer??

kparichay: > . add brief @param about alloc_now.

Added.

> . I wonder the reason of false default value is because to track the weight in manager??
Could you elaborate why manager does not allow manager to track initialized var_grads?
and so.. we cannot allocate buffer memory during the initialization of layer??

The default is false to ensure the same behavior of inouts+derivatives and weights+gradients. Earlier, weights were eagerly allocated, but others were not. Now, all 4 (weights/grads/inouts/derivs) wrappers are allocated without memory first, and they are allocated whenever required.
Manager allows tracking initialized var_grads. Further, as alloc_now is still a parameter that allows creating weights/var_grads which allocate memory at that moment.
Layers are allowed to allocate memory for weight in their init by passing `allow_now = true`, but making its default value to false means that nntrainer encourages to use false and let the manager handle allocation appropriately than allocating upon init.

jijoongmoon (approved):
LGTM

PR #1007, opened 2021-03-12
[loader] Change semantics of path in ini

zhoonit:
Ini description is expected to be somewhat determinisitic.

But paths inside ini is not deterministic. For example, if you have
save_path=model.bin,

if A application trains it and B application try to use it, reading the
same ini does not gaurantee B to read to the weights from A.

It is because A and B can have diffrent working path.

To make it consistent, this patch changes semantics of paths' in ini.

1. if model.appcontext.hasWorkingDirectory(), path inside ini are interpreted from
the workingDirectory
2. else, it is interpreted **relative to where ini is located at.**

**Self evaluation:**

### DIFF
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -145,6 +145,20 @@ std::vector<std::string> getPluginPaths() {
   return ret;
 }
 
+/**
+ * @brief Get the Full Path from given string
+ * @details path is resolved in the following order
+ * 1) if @a path is absolute, return path
+ * ----------------------------------------
+ * 2) if @a base == "" && @a path == "", return "."
+ * 3) if @a base == "" && @a path != "", return @a path
+ * 4) if @a base != "" && @a path == "", return @a base
+ * 5) if @a base != "" && @a path != "", return @a base + "/" + path
+ *
+ * @param path path to calculate from base
+ * @param base base path
+ * @return const std::string
+ */
 const std::string getFullPath(const std::string &path,
                               const std::string &base) {
   /// if path is absolute, return path
--- a/nntrainer/models/model_loader.cpp
+++ b/nntrainer/models/model_loader.cpp
@@ -64,7 +64,7 @@ int ModelLoader::loadModelConfigIni(dictionary *ini, NeuralNetwork &model) {
   const std::string &save_path =
     iniparser_getstring(ini, "Model:Save_path", unknown);
   if (save_path != unknown) {
-    model.setSavePath(save_path);
+    model.setSavePath(resolvePath(save_path));
   }
   model.batch_size =
     iniparser_getint(ini, "Model:Batch_Size", model.batch_size);
@@ -151,7 +151,7 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
       return required ? ML_ERROR_INVALID_PARAMETER : ML_ERROR_NONE;
     }
 
-    return dbuffer->setDataFile(dt, model.app_context.getWorkingPath(path));
+    return dbuffer->setDataFile(dt, resolvePath(path));
   };
 
   status = parse_and_set("DataSet:TrainData", DATA_TRAIN, true);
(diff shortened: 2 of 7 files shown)


jijoongmoon (approved):
LGTM

nntrainer/app_context.cpp
```diff
@@ -145,6 +145,19 @@ std::vector<std::string> getPluginPaths() {
   return ret;
 }
 
+/**
+ * @brief Get the Full Path from given string
+ * path is resolved in the following order
```
kparichay, line 150: move from here to the rest of the `@brief` description under `@details`?

nntrainer/models/model_loader.cpp
```diff
@@ -456,7 +455,38 @@ int ModelLoader::loadFromIni(std::string ini_file, NeuralNetwork &model,
  * @brief     load all of model and dataset from given config file
  */
 int ModelLoader::loadFromConfig(std::string config, NeuralNetwork &model) {
-  return loadFromConfig(config, model, false);
+
+  NNTR_THROW_IF(model_file_context != nullptr, std::runtime_error)
```
kparichay, line 459: can you change this throw to `return error` or change other return error to throw for this function to keep it consistent?

kparichay (commented):
> else, it is interpreted relative to where ini is located at.

Can you please describe the outcome of various scenarios for this? Some scenarios of the top of my head:
- two models with different ini location and shared appContext, is path different for each?
- new ini configuration is loaded, is path interpreted updated?
- is currentWorkingDirectory updated when interpreting these paths?
- if user sets relative savePath as a property, is it wrt to ini or currentWorkingDirectory?

zhoonit:
> Can you please describe the outcome of various scenarios for this? Some scenarios of the top of my head:
> 
1. two models with different ini location and shared appContext, is path different for each?

It will be essentially same for both but it depends on if the model app_context has working path set.
```cpp
   return appContext.hasWorkingPathSet() ? app_context.getWorkingPath() : model_file_context.getWorkingPath();
```

2.  new ini configuration is loaded, is path interpreted updated?

Path will be interpreted _before_ setting into the model. So, path won't interpreted again. if Path is newly given from the new ini conf, it will be interpreted relative to the ini. (if you have working directory set for the model, it will be used for old and new ini config though).

If you have a good reason to avoid this behavior, it will be more reasonable to use api or absolute path.

3. is currentWorkingDirectory updated when interpreting these paths?

No, more like, every model_file has it's own working directory and it is used if model's working directory is not set.

4. if user sets relative savePath as a property, is it wrt to ini or currentWorkingDirectory?

Extending from 3, wrt to ini if model's working directory is not set else model's working directory.

Current related doc is below. Please throw me some idea if the description is a bit shady.

```cpp
   /* from ModelLoader::resolvePath
   * @note  if path is absolute path, return path.
   *        if app_context has working directory set, resolve from app_context
   *        if not, resolve path assuming model_path is the current directory.
   *        The behavior relys on the semantics of getWorkingPath();
   */
  ```

nntrainer/app_context.cpp
zhoonit: Done

nntrainer/models/model_loader.cpp
zhoonit: Done to `return error`

lhs8928 (approved):
LGTM

kparichay:
> > two models with different ini location and shared appContext, is path different for each?
> It will be essentially same for both if app_context.getWorkingPath() is set, if not it will be different.

Please update the documentation (docs/*.md) for this case.

> > if user sets relative savePath as a property, is it wrt to ini or currentWorkingDirectory?
> Extending from 3, wrt to ini if model's working directory is not set else model's working directory.

So, save path will always be wrt to ini. I wonder if wrt to local directory vs wrt to ini directory - which is more intuitive. IMO, the behavior is not intuitive for a user who does not set currentWorkingDirectory() manually.

zhoonit:
> So, save path will always be wrt to ini

Not always, if model app_context has working directory, it is relative to the working directory.

> I wonder if wrt to local directory vs wrt to ini directory 

Although, local directory seems more intuitive, the reason for this design change is that, app can easily gets to some undefined zone if it is based on applications working directory as a default.

1. if same application running in different working directory, it will fail to load the save path
2. if same model is used across application, it will fail to load the save path (except for a single case that all applications are in the same directory)

> BTW, the ini file path, if the currentWorkingDirectory is not set, what is it wrt?

It is based on the ini file's absolute path so `model_file_context` (compared to `model_context`) is guaranteed to have working directory set.

zhoonit:
Let me update the description to make it clearer.

PR #1009, opened 2021-03-12
[model] Move layers out of model

kparichay:
Move layers from model to graph.
This places all the graph related elements in the graph
and easy to extend and reason about model.
This shows the redundancy in the graph which shall be handled in
upcoming PRs.

Remove friendship between classes as it makes extending the interface
difficult. Some friendships exist which will be removed in upcoming
PRs.

See Also #986

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -584,7 +584,7 @@ NetworkGraph::getGraph(const std::string &input_layer,
 }
 
 void NetworkGraph::extendGraph(std::vector<std::shared_ptr<Layer>> graph,
-                               std::string prefix) {
+                               std::string &prefix) {
 
   /**
    * The input_layers for graph[0] here is provided to the backbone by the ini
--- a/nntrainer/graph/network_graph.h
+++ b/nntrainer/graph/network_graph.h
@@ -178,7 +178,7 @@ public:
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
   void extendGraph(std::vector<std::shared_ptr<Layer>> graph,
-                   std::string prefix);
+                   std::string &prefix);
 
   /**
    * @brief     Ensure that layer has a name


jijoongmoon (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -522,6 +544,83 @@ std::vector<TensorDim> NetworkGraph::getOutputDimension() {
...
+  /** copy the graph and return */
+  std::vector<std::shared_ptr<Layer>> ret;
+  std::copy(layers.begin() + num_layers_remove_start,
+            layers.end() - num_layers_remove_end, std::back_inserter(ret));
+
+  return ret;
+}
+
+void NetworkGraph::extendGraph(std::vector<std::shared_ptr<Layer>> graph,
+                               std::string prefix) {
```
zhoonit, line 585: const std::string &prefix ?

nntrainer/graph/network_graph.h
```diff
@@ -226,6 +298,26 @@ class NetworkGraph {
...
 
+  /**
+   * @brief     Copy the graph
+   * @param[in] from Graph Object to copy
+   * @retval    Graph Object copyed
+   */
+  NetworkGraph &copy(NetworkGraph &from) {
+    if (this != &from) {
+      for (unsigned int i = 0; i < layers.size(); i++)
+        layers[i]->copy(from.layers[i]);
```
zhoonit, line 309: 1. How about copy constructor instead?
2. I wonder if copying layer is enough.

nntrainer/tensor/weight.h
```diff
@@ -49,17 +49,6 @@ enum class WeightInitializer {
...
-  friend class Layer;
-  friend class Conv2DLayer;
-  friend class FullyConnectedLayer;
-  friend class BatchNormalizationLayer;
-
-  /** Declare opitmizer as friend to get variable/gradient reference */
-  friend class Optimizer;
-  friend class SGD;
-  friend class Adam;
-
```
zhoonit, line 62: 🥇 :+1:

lhs8928 (approved):
Great

nntrainer/graph/network_graph.h
kparichay: This PR just intends to move `layers` and its related code out of model to graph.
More refining is being done in #1016. I will add it here.

> How about copy constructor instead?

Yes, #1016 will do it.

> I wonder if copying layer is enough.

This will depend on the semantics of the layer itself. Currently this is very wrong as this expects `this->layers.size()` is already same as `from.layers.size()`. All this being done in #1016. This PR just moves things out of model to graph without any improvements or major bug fixes.

nntrainer/graph/network_graph.cpp
kparichay: done.

nntrainer/graph/network_graph.h
zhoonit: Awesome

kparichay:
The changes proposed in this PR are already merged with #1013.
This PR just contains the changes requested in review which were not merged with #1013.

PR #1016, opened 2021-03-16
[graph] Refactor graph + many bug fixes

kparichay:
Update graph to remove layers and have a single data structure
for maintainence purposes. Now, adj represents the nodes list
which is unsorted. Sorted list of nodes is available after topological
sort in Sorted.

Additional Bug fixes:
- getSortedLayerNode was giving out wrong indices, its fixed now
- topological sort now empties sorted list before adding new elements
- other minor fixes to make adj replace layers

The compile has moved to graph and easier API exposed to model.
    Also many public exposed functions reduced.

Add bug fixes for layer realizations for making the graph
    Many scenarios were missed about creating connections for the graph
    which this patch fixes.
Further, this patch reduce the isCompilable() checks as the first
    layer cant be determined till the topologicalSort() of the graph is done.
    Added checkCompiledGraph() where first layer and some more checks like
    types are checked.

Add compiled status to graph
    Compiling multiple times or modifying the graph after compilation
    is erronous. Getting sorted list before completing the compilation
    is also an error.

Cleanup redundant member variables from graph
    Rename some of the functions. Some others to be modified
    have been added as TODO.
    getters have modified to work better.
Some redundant members have been removed from layer_internal as well.

I know this PR has become quite long, my apologies but the changes are all connected and necessary to resolve the bugs inside model_graph. I have tried to split the changes into smaller commits for easier review. 

**Self evaluation:**


kparichay:
@nnstreamer/nnstreamer This introduces some big changes to the model graph along with many fixes. Please review thoroughly.

nntrainer/graph/network_graph.cpp
```diff
@@ -37,35 +38,68 @@ namespace nntrainer {
 static const std::vector<std::string> in_place_layers = {
   ActivationLayer::type, BatchNormalizationLayer::type};
 
+int NetworkGraph::compile(const LossType loss_type) {
```
jijoongmoon, line 41: c and c++ api has compile. So.. API is still in neuralnet and it it wrapper of this?

kparichay: `NeuralNetwork::Compile` still exists, and it is just a wrapper for this.
In later stages, once `NeuralNetwork::Compile()` will decide if `NetworkGraph::Compile` must be done or it is already done offline etc.

nntrainer/graph/network_graph.cpp
```diff
@@ -181,13 +190,19 @@ int NetworkGraph::realizeMultiInputType(Layer &current) {
...
   ensureName(layer, current.getName());
+
   layer->setNumInputs(current.getNumInputs());
   layer->input_layers.clear();
   for (unsigned int i = 0; i < current.input_layers.size(); ++i)
     layer->input_layers.push_back(current.input_layers[i]);
 
+  layer->setNumOutputs(1);
+  layer->output_layers.clear();
+  layer->output_layers.push_back(current.getName());
```
jijoongmoon, line 204: setNumOutputs is called when the layer is created. You can remove this. 
Is it dummy? set the layer->output_layers with current.getName()?? it is not actually. it could be the output of the current layer.

nntrainer/graph/network_graph.cpp
```diff
@@ -217,6 +232,20 @@ int NetworkGraph::realizeFlattenType(Layer &current) {
   layer->input_layers.clear();
   layer->input_layers.push_back(current.getName());
 
+  if (current.getNumOutputs() != 1)
+    return ML_ERROR_INVALID_PARAMETER;
+
+  layer->setNumOutputs(current.getNumOutputs());
+  layer->output_layers.clear();
+  for (unsigned int i = 0; i < current.getNumOutputs(); ++i)
+    layer->output_layers.push_back(current.output_layers[i]);
```
jijoongmoon, line 241: I think flatten layer shouldn't have multi output.

nntrainer/graph/network_graph.cpp
```diff
@@ -268,33 +298,43 @@ int NetworkGraph::realizeActivationType(Layer &current) {
   current.output_layers.clear();
   current.output_layers.push_back(layer->getName());
 
-  addLayerNode(layer);
-
   updateNameInLayers(current.getName(), layer->getName());
+  addLayerNode(layer);
```
jijoongmoon, line 302: I wonder the order is a matter? layer is the pointer.. so.. It might not be an issue.

nntrainer/graph/network_graph.cpp
```diff
@@ -402,47 +449,61 @@ int NetworkGraph::realizeMultiOutputType(Layer &current) {
...
+        return ML_ERROR_INVALID_PARAMETER;
+      }
+    }
+  }
+
   return ML_ERROR_NONE;
 }
 
-int NetworkGraph::setGraphNode(const LossType loss_type) {
+int NetworkGraph::setGraphNode() {
```
jijoongmoon, line 485: is it ok if we remove the consideration of the loss layer?

nntrainer/graph/network_graph.cpp
```diff
@@ -577,15 +633,32 @@ NetworkGraph::getGraph(const std::string &input_layer,
 
   /** copy the graph and return */
   std::vector<std::shared_ptr<Layer>> ret;
-  std::copy(layers.begin() + num_layers_remove_start,
-            layers.end() - num_layers_remove_end, std::back_inserter(ret));
+  std::transform(adj.begin() + num_layers_remove_start,
+                 adj.end() - num_layers_remove_end, std::back_inserter(ret),
+                 [](auto const &elem) { return elem.front().layer; });
```
jijoongmoon, line 638: transform is better than std::copy?

nntrainer/graph/network_graph.cpp
```diff
@@ -577,15 +633,32 @@ NetworkGraph::getGraph(const std::string &input_layer,
...
+
+std::vector<std::shared_ptr<Layer>> NetworkGraph::getLayers() {
+  std::vector<std::shared_ptr<Layer>> ret;
+  if (compiled) {
+    std::transform(Sorted.begin(), Sorted.end(), std::back_inserter(ret),
+                   [](auto const &elem) { return elem.layer; });
+  } else {
+    std::transform(adj.begin(), adj.end(), std::back_inserter(ret),
+                   [](auto const &elem) { return elem.front().layer; });
+  }
```
jijoongmoon, line 651: I guess. once it is compiled.. then we do not need adj anymore. We could remove it.

nntrainer/graph/network_graph.cpp
kparichay: Loss layer can't be added till we know where to add it. Earlier code that the order of layers in the ini is sensible but that can be wrong. So, first, the graph is created without loss and sorted. After sorting, the graph knows the last layer, and the loss layer is inserted at the end.

nntrainer/graph/network_graph.cpp
kparichay: std::copy is better but we need a modified version of each element. Hence, std::transform is used.

nntrainer/graph/network_graph.cpp
kparichay: that's interesting, I will try and if it works, will add as a new commit.

nntrainer/graph/network_graph.cpp
kparichay: line 235 ensures that this loops runs only once.

nntrainer/graph/network_graph.cpp
kparichay: In this case, `multiInputLayer` is added as utility to the current layer so that it can merge (add or concat) all its inputs and provide it as a single input to the `current` layer. So, the `output_layer` of the new `layer` must be `current`. Please let me know if I got the understanding wrong.

`setNumOutputs(1)` is done for clarity and reading purposes (you are right that it is initialized to 1 anyway, so I think compiler will optimize it out as it is a compile time constant).

nntrainer/graph/network_graph.cpp
kparichay: yes, the order does not matter.
In fact, there is no need for `updateNameInLayers()` here. I will remove it.

nntrainer/graph/network_graph.cpp
```diff
@@ -96,14 +97,7 @@ LayerNode &NetworkGraph::getLayerNode(unsigned int ith) {
...
 LayerNode &NetworkGraph::getSortedLayerNode(unsigned int ith) {
-
-  for (unsigned int i = 0; i < Sorted.size(); ++i) {
-    if (Sorted[i].index == ith) {
-      return Sorted[i];
-    }
-  }
-
-  throw std::invalid_argument("Cannot find Layer");
+  return Sorted[ith];
```
zhoonit: If Sorted.empty() throw?

nntrainer/graph/network_graph.cpp
```diff
@@ -268,33 +267,44 @@ int NetworkGraph::realizeActivationType(Layer &current) {
...
   updateNameInLayers(current.getName(), layer->getName());
+  addLayerNode(layer);
 
   return status;
 }
 
 int NetworkGraph::addLossLayer(const LossType loss_type) {
+
   int status = ML_ERROR_NONE;
+
```
zhoonit, line 310: Check if Sorted.empty() ?

nntrainer/graph/network_graph.cpp
kparichay: Turns out there was a bug related to this. Fix added about renaming the connections.

nntrainer/graph/network_graph.cpp
```diff
@@ -331,44 +340,48 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
...
+  topologicalSort();
 
   return ML_ERROR_NONE;
 }
 
 void NetworkGraph::setOutputLayers() {
 
-  if (layers.back()->getNumOutputs() > 0 &&
-      layers.back()->output_layers.size() > 0) {
+  auto &last_layer = adj.back().front().layer;
```
zhoonit, line 385: check if adj is empty?

nntrainer/graph/network_graph.cpp
```diff
@@ -331,44 +340,48 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
...
+  if (last_layer->getNumOutputs() > 0 && last_layer->output_layers.size() > 0) {
     throw std::runtime_error("last layer already has a output layer");
   }
 
-  layers.back()->setNumOutputs(1);
-  layers.back()->output_layers.clear();
-  layers.back()->output_layers.push_back("__exit__");
+  last_layer->setNumOutputs(1);
```
zhoonit, line 390: I wonder if it is always 1?

zhoonit: Yes it is. __exit__

nntrainer/graph/network_graph.cpp
kparichay: I thought c++ used to throw in this situation, but it does not. Added.

nntrainer/graph/network_graph.cpp
```diff
@@ -402,12 +415,12 @@ int NetworkGraph::realizeMultiOutputType(Layer &current) {
...
 
 int NetworkGraph::isCompilable() {
-  if (layers.empty()) {
+  if (adj.empty()) {
     ml_loge("Layer is empty");
     return ML_ERROR_INVALID_PARAMETER;
   }
 
-  Layer &l = *layers[0];
+  Layer &l = *adj[0].front().layer;
```
zhoonit: Is the first layer guaranteed to be the first layer?

nntrainer/graph/network_graph.cpp
kparichay: this is already guaranteed with `isCompilable()` which ensures that `adj` is not empty. And `topologicalSort` by definition ensures that Sorted will not be empty. 
For completeness, added a check at the end of topologicalSort.

nntrainer/graph/network_graph.cpp
kparichay: ensured by `isCompilable()`.

nntrainer/graph/network_graph.cpp
kparichay: Yes, it is, the statement is just for easier reading.

nntrainer/graph/network_graph.cpp
```diff
@@ -496,26 +504,28 @@ LayerNode &NetworkGraph::getLayerNode(const std::string &layer_name) {
   throw std::invalid_argument("Cannot find Layer");
 }
 
-int NetworkGraph::setEdge() {
-  int status = ML_ERROR_NONE;
+void NetworkGraph::setEdge(unsigned int adj_idx) {
+  std::list<LayerNode>::iterator iter = adj[adj_idx].begin();
```
zhoonit: check if adj_idx exist in `adj`?

nntrainer/graph/network_graph.cpp
kparichay: You are right. This is handled in a later commit of this PR. You can check the final version of this function as a check.

nntrainer/graph/network_graph.cpp
kparichay: setEdge is a private function called by internal functions only. 
The way it is called ensures that `adj_idx < adj.size()`.

nntrainer/graph/network_graph.h
```diff
@@ -153,9 +153,9 @@ class NetworkGraph {
    * @retval Layer
    */
   std::shared_ptr<Layer> getLayer(const std::string &layer_name) {
-    for (auto iter = layers.begin(); iter != layers.end(); ++iter) {
-      if ((*iter)->getName() == layer_name) {
-        return *iter;
+    for (auto iter = adj.begin(); iter != adj.end(); ++iter) {
+      if ((*iter).front().layer->getName() == layer_name) {
```
zhoonit: I wonder about the semantics of layer_name, is it `istrequal` or case-sensitive?

nntrainer/graph/network_graph.cpp
```diff
@@ -268,33 +267,44 @@ int NetworkGraph::realizeActivationType(Layer &current) {
...
   LossType updated_loss_type = loss_type;
   if (num_node == 0) {
     status = ML_ERROR_INVALID_PARAMETER;
     NN_RETURN_STATUS();
   }
 
+  LayerNode last_node = Sorted.back();
   if (updated_loss_type == LossType::LOSS_ENTROPY) {
-    if (getLayerNode(num_node - 1).layer->getType() != "activation") {
+    if (last_node.layer->getType() != "activation") {
```
zhoonit, line 327: istrequal here?

nntrainer/graph/network_graph.cpp
```diff
@@ -552,23 +562,23 @@ NetworkGraph::getGraph(const std::string &input_layer,
...
 
   /** count layers before input layer */
   unsigned int num_layers_remove_start = 0;
   if (!input_layer.empty()) {
-    for (auto iter = layers.begin();
-         iter != layers.end() - num_layers_remove_end; iter++) {
-      if ((*iter)->getName() != input_layer)
+    for (auto iter = adj.begin(); iter != adj.end() - num_layers_remove_end;
+         iter++) {
+      if ((*iter).front().layer->getName() != input_layer)
```
zhoonit, line 627: istrequal here?

nntrainer/graph/network_graph.h
kparichay: that's a good question. Layer names are meant to a unique case-sensitive string. So, IMO using `istrequal` goes against that invariant.
This must be done all over code though.

nntrainer/graph/network_graph.cpp
kparichay: Followup on this.

nntrainer/graph/network_graph.cpp
kparichay: This is different from the layer name. what is our policy for this?

nntrainer/graph/network_graph.h
```diff
@@ -53,22 +53,10 @@ class NetworkGraph {
...
-  /**
-   * @brief Sorting and Define order to calculate : Depth First Search
-   */
-  void topologicalSort();
-
-  /**
-   * @brief Create new LayerNode and add into Graph
-   * @param[in] layer shared_ptr of Layer
+   * @brief     Compile the graph
+   * @param[in] loss_type loss for the graph
```
zhoonit, line 57: `@return int`?

nntrainer/graph/network_graph.h
```diff
@@ -352,6 +249,119 @@ class NetworkGraph {
...
+   */
+  int addLossLayer(const LossType loss_type);
+
+  /**
+   * @brief     set Multi Output Layer
+   */
+  void setOutputLayers();
+
+  /**
+   * @brief     Ensure that layer has a name
```
zhoonit, line 345: could you update the doc here?

nntrainer/graph/network_graph.cpp
```diff
@@ -402,12 +415,12 @@ int NetworkGraph::realizeMultiOutputType(Layer &current) {
 }
 
 int NetworkGraph::isCompilable() {
```
zhoonit, line 451: How about adding here, a bool or some logic to check if compiled already and check it here to prevent compiling twice?

nntrainer/graph/network_graph.cpp
zhoonit: I checked, type and layer names are case-insensitive.

nntrainer/graph/network_graph.cpp
zhoonit: Cool

nntrainer/graph/network_graph.cpp
zhoonit: Awesome

nntrainer/graph/network_graph.cpp
zhoonit: https://github.com/nnstreamer/nntrainer/pull/1016/commits/55dab436b8e7be5c59b7a68a8cf3601e4406fd25#r595745132

nntrainer/graph/network_graph.h
zhoonit: It's good, we can leave this as an issue, with good first issue labeled.

nntrainer/graph/network_graph.cpp
```diff
@@ -440,25 +459,31 @@ int NetworkGraph::isCompilable() {
...
-  // TODO: move this to the input layer and not first layer
-  if (l.getInputDimension().size() == 0) {
-    ml_loge("InputDimension of first layer is not set");
-    return ML_ERROR_INVALID_PARAMETER;
-  }
+  return ML_ERROR_NONE;
+}
 
+int NetworkGraph::checkCompiledGraph() {
+  auto &l = Sorted[0].layer;
```
zhoonit, line 461: Check if Sorted.empty()?

nntrainer/graph/network_graph.cpp
```diff
@@ -440,25 +459,31 @@ int NetworkGraph::isCompilable() {
...
       istrequal(type, LossLayer::type)) {
-    ml_loge("%s cannot be the first layer, type: %s", l.getName().c_str(),
+    ml_loge("%s cannot be the first layer, type: %s", l->getName().c_str(),
             type.c_str());
     return ML_ERROR_INVALID_PARAMETER;
   }
 
+  /** Dimension of input layers must be known */
+  for (auto const &lnode : Sorted) {
+    if (lnode.layer->getType() == InputLayer::type) {
```
zhoonit, line 474: istrequal?

test/unittest/unittest_nntrainer_modelfile.cpp
```diff
@@ -817,7 +817,7 @@ TEST(nntrainerIniTest, backbone_n_21) {
   EXPECT_EQ(NN.compile(), ML_ERROR_NONE);
   EXPECT_EQ(NN.initialize(), ML_ERROR_NONE);
   EXPECT_EQ(NN.getNetworkGraph().getSorted().size(), 3u);
-  // EXPECT_EQ(NN.getGraph().size(), 3);
+  // EXPECT_EQ(NN.getUnsortedLayers().size(), 3);
```
zhoonit, line 820: I wonder if this test can be deleted?

nntrainer/graph/network_graph.cpp
```diff
@@ -84,11 +84,22 @@ void NetworkGraph::addLayerNode(std::shared_ptr<Layer> layer) {
...
-  node->index = num_node;
+  node->index = adj.size();
 
   l.assign(1, *node);
   adj.push_back(l);
-  num_node++;
+}
+
+LayerNode &NetworkGraph::getLayerNode(unsigned int ith) {
+
```
zhoonit, line 95: check if ith exist?

nntrainer/graph/network_graph.cpp
```diff
@@ -84,11 +84,22 @@ void NetworkGraph::addLayerNode(std::shared_ptr<Layer> layer) {
...
+  if (adj[ith].front().index != ith)
+    throw std::runtime_error("Graph internal index mismatch");
+
+  return adj[ith].front();
+}
+
+LayerNode &NetworkGraph::getSortedLayerNode(unsigned int ith) {
```
zhoonit, line 102: check if ith exist?

nntrainer/graph/network_graph.cpp
zhoonit: But we can make layer names as case-sensitive, leaving as an issue I guess.

nntrainer/graph/network_graph.cpp
zhoonit: Turned out that there is a bool `compiled` I wonder if that can be used here as well?

nntrainer/graph/network_graph.cpp
kparichay: i will let #1024 handle it.

nntrainer/graph/network_graph.cpp
kparichay: this is called internally in `Compile`, and sorted is guaranteed to be non-empty by `topologicalSort`.

zhoonit: Cool

nntrainer/graph/network_graph.h
kparichay: done

nntrainer/graph/network_graph.h
kparichay: done.

nntrainer/graph/network_graph.cpp
kparichay: The check was being done in `compile()` itself, but it's better suited here. Moved here.

nntrainer/graph/network_graph.cpp
kparichay: done.

nntrainer/graph/network_graph.cpp
kparichay: done.

zhoonit (approved):
Nice work! LGTM :)

test/unittest/unittest_nntrainer_modelfile.cpp
kparichay: this test has been deleted.

kparichay:
rebased.

nntrainer/graph/network_graph.cpp
jijoongmoon: right. BTW, Isn't it set at NetworkGraph::setOutputLayers()? it is ok to set twice??

nntrainer/graph/network_graph.cpp
jijoongmoon: > line 235 ensures that this loop runs only once.

Do you mean current.getNumOutputs() is always 1? I think it is better to set the output_layer after all the input is set properly. It if is calculated layer by layer.. then there might be an error, I guess. Cause.. if the current number of output is multiple.. that means.. we needs multi-out layer.. so.. Could you check this?

nntrainer/graph/network_graph.cpp
kparichay: `setOutputLayers()` is added here just for readability. Yes, it is alright to set twice or reset it.

nntrainer/graph/network_graph.cpp
kparichay: Yes, the current.getNumOutputs() must always be 1. Flatten is used only in this way.  If flatten has to be used with multi-output, then the flatten must be set before the multi-output. If the flatten is supposed to be used for a single output of the multi-output, then it should be inserted manually there inside the ini model.
If you still think there is a bug in the situation you described, can you please describe a minimal model showing the bug.

jijoongmoon (approved):
We discuss off-line. LGTM!

PR #1028, opened 2021-03-19
Fix/coverity issue 210319

lhs8928:
Commit 1:  Handle Unused value issue

Store status value using originStatus to protect from overwriting
resolves 1216574, 1216576, 1216588, 1216594, 

Commit 2: Handle Structually dead code(UNREACHABLE) issues

Deleted the code that followed by switch case statement
resolves 1216577, 1216583, 1216591, 1216597

Commit 3: Handle Unchecked return value(CHECKED_RETURN) issues

Added check the return value
resolve 1222078, 1222441

Commit 4: Handle Not restoring ostream format(STREAM_FORMAT_STATE) issue

Restore outputstream adjustflag to standard stream
resolve: 1222083

Commit 5: Handle Large stack use(STACK_USE) issue

Coverity warns that in embedded application stack overflow may occur.
Because featureVector size(1080000) is bigger than maximum stack size(250000).
So define featureVector at the global area.
resolves: 1216560, 1216582

Commit 6: Handle uncaught exception issue

Added try catch statement
resolves: 1143556, 1144840

### DIFF
--- a/Applications/TransferLearning/Draw_Classification/jni/main.cpp
+++ b/Applications/TransferLearning/Draw_Classification/jni/main.cpp
@@ -277,6 +277,7 @@ int getInputFeature_c(const std::string filename, float *feature_input) {
 int testModel(const char *data_path, const char *model) {
 #if defined(NNSTREAMER_AVAILABLE)
   int status = ML_ERROR_NONE;
+  int new_status;
   ml_pipeline_h pipe;
   ml_pipeline_src_h src;
   ml_pipeline_sink_h sink;
@@ -363,17 +364,28 @@ int testModel(const char *data_path, const char *model) {
 
 fail_info_release:
   ml_tensors_info_destroy(in_info);
-
-  status = ml_pipeline_stop(pipe);
+  new_status = ml_pipeline_stop(pipe);
+  if (status == ML_ERROR_NONE && new_status != ML_ERROR_NONE) {
+    status = new_status;
+  }
 
 fail_sink_release:
-  status = ml_pipeline_sink_unregister(sink);
+  new_status = ml_pipeline_sink_unregister(sink);
+  if (status == ML_ERROR_NONE && new_status != ML_ERROR_NONE) {
+    status = new_status;
+  }
 
 fail_src_release:
-  status = ml_pipeline_src_release_handle(src);
+  new_status = ml_pipeline_src_release_handle(src);
+  if (status == ML_ERROR_NONE && new_status != ML_ERROR_NONE) {
+    status = new_status;
+  }
 
 fail_pipe_destroy:
-  status = ml_pipeline_destroy(pipe);
+  new_status = ml_pipeline_destroy(pipe);
+  if (status == ML_ERROR_NONE && new_status != ML_ERROR_NONE) {
+    status = new_status;
+  }
 
 fail_exit:
   return status;
--- a/Applications/TransferLearning/Draw_Classification/jni/main.cpp
+++ b/Applications/TransferLearning/Draw_Classification/jni/main.cpp
@@ -77,6 +77,10 @@ const char *label_names[LABEL_SIZE] = {"happy", "sad", "soso"};
 float inputVector[EPOCH_SIZE][INPUT_SIZE];
 float labelVector[EPOCH_SIZE][LABEL_SIZE];
 
+#if defined(NNSTREAMER_AVAILABLE)
+float featureVector[INPUT_SIZE];
+#endif
+
 #if defined(APP_VALIDATE)
 /** Benchmark output values */
 const float test_output_benchmark[TOTAL_TEST_SIZE] = {
@@ -334,7 +338,6 @@ int testModel(const char *data_path, const char *model) {
       goto fail_info_release;
     }
 
-    float featureVector[INPUT_SIZE];
     status = getInputFeature_c(test_file_path, featureVector);
     free(test_file_path);
     if (status != ML_ERROR_NONE)
--- a/Applications/TransferLearning/Draw_Classification/jni/main.cpp
+++ b/Applications/TransferLearning/Draw_Classification/jni/main.cpp
@@ -455,16 +455,35 @@ int main(int argc, char *argv[]) {
     loadAllData(data_path, inputVector, labelVector);
   } catch (...) {
     std::cout << "Failed loading input images." << std::endl;
+#if defined(__TIZEN__)
+    set_feature_state(NOT_CHECKED_YET);
+#endif
     return 1;
   }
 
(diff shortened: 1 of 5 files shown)


api/capi/src/nntrainer_util.cpp
```diff
@@ -41,9 +41,6 @@ ml_optimizer_to_nntrainer_type(ml_train_optimizer_type_e type) {
     throw nntrainer::exception::not_supported(
       "[ml_optimmizer_to_nntrainer_type] Not supported type given");
   }
-
-  throw nntrainer::exception::not_supported(
```
zhoonit, line 45: I wonder why this is deleted? We know this is unreachable section but this is to prevent no return from future code changes

zhoonit: And for others

lhs8928: Yes this is unreachable code and coverity point this out.
For that reason should I comment this line?

zhoonit: IMO, it's better to tell coverity it's a false positive.

lhs8928: I classified it as intentional.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -363,17 +365,28 @@ int testModel(const char *data_path, const char *model) {
 
 fail_info_release:
   ml_tensors_info_destroy(in_info);
-
+  originStatus = status;
```
zhoonit, line 374: I think this could coverity result could be ignore next time.

zhoonit (approved):
LGTM

Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -229,7 +234,10 @@ void sink_cb(const ml_tensors_data_h data, const ml_tensors_info_h info,
   int max_idx = -1;
   float max_val = 0; // last layer is softmax, so all values will be positive
 
-  ml_tensors_info_get_tensor_dimension(info, 0, dim);
+  status = ml_tensors_info_get_tensor_dimension(info, 0, dim);
+  if (status != ML_ERROR_NONE) {
+    return;
```
kparichay, line 239: add a log statement here.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -363,17 +371,28 @@ int testModel(const char *data_path, const char *model) {
 
 fail_info_release:
   ml_tensors_info_destroy(in_info);
-
+  originStatus = status;
   status = ml_pipeline_stop(pipe);
+  NN_STATUS_RETURN();
```
kparichay, line 376: This exits without `sink_unregister`, `src_release_handle` and `destroy_pipeline` if `pipeline_stop` fails. This will leak memory. Please correct this.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -440,12 +459,22 @@ int main(int argc, char *argv[]) {
   }
 
   /** Do the training */
-  status = trainModel(config.c_str());
+  try {
+    status = trainModel(config.c_str());
+  } catch (...) {
+    std::cerr << "Failed train model\n";
+    return 1;
```
kparichay, line 466: let's not exist here. Please reset the `feature_state` back to `NOT_CHECKED_YET` if the `__TIZEN__` is defined.
This is a bug in the existing code as well in line 458.

kparichay (changes requested):
Please fix the possible memory leak and reset tizen feature state upon exit.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
lhs8928: All right.
But let say something went wrong so exit to fail_info_release and ml_pipeline_stop also failed.
Then which status should I return?

Applications/TransferLearning/Draw_Classification/jni/main.cpp
lhs8928: You mean no exit and just reset the feature_state?

Applications/TransferLearning/Draw_Classification/jni/main.cpp
kparichay: The original status of failure, which led to the exit to fail_info_release.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
kparichay: Exit is fine here, but reset the feature_state before exit.

Applications/TransferLearning/Draw_Classification/jni/main.cpp
lhs8928: Done

Applications/TransferLearning/Draw_Classification/jni/main.cpp
lhs8928: Done

Applications/TransferLearning/Draw_Classification/jni/main.cpp
lhs8928: Done

Applications/TransferLearning/Draw_Classification/jni/main.cpp
```diff
@@ -363,17 +371,16 @@ int testModel(const char *data_path, const char *model) {
 
 fail_info_release:
   ml_tensors_info_destroy(in_info);
-
-  status = ml_pipeline_stop(pipe);
+  ml_pipeline_stop(pipe);
```
kparichay, line 374: For future you can consider the below style to avoid coverity issues in cases like these.
```c++
new_status = ml_pipeline_stop(pipe);
if (new_status != ML_ERROR_NONE && status == ML_ERROR_NONE)
  status = new_status;
```

Let's see if the current code makes an issue (of ignoring the return value).

kparichay (approved):
LGTM

Applications/TransferLearning/Draw_Classification/jni/main.cpp
lhs8928: I think coverity will make a return value issues.
I applied your suggestion

kparichay: This style returns the first error that occurred in the code (as other errors can be caused by the first error).
If Coverity gives the issues of `return value ignored`, then it can be marked false positive.

PR #1031, opened 2021-03-19
[models/test] Valgrind models test fix

kparichay:
Valgrind gives the error of uninitialized memory usage for this
unittest which is fixed with this patch.

**Self evaluation:**

### DIFF
--- a/test/unittest/unittest_nntrainer_models.cpp
+++ b/test/unittest/unittest_nntrainer_models.cpp
@@ -390,9 +390,13 @@ void GraphWatcher::compareFor(const std::string &reference,
 
 void GraphWatcher::validateFor(const std::string &reference,
                                const nntrainer::TensorDim &label_shape) {
-  nntrainer::sharedConstTensors input = {
-    MAKE_SHARED_TENSOR(nn.getInputDimension()[0])};
-  nntrainer::sharedConstTensors label = {MAKE_SHARED_TENSOR(label_shape)};
+  auto in_tensor = MAKE_SHARED_TENSOR(nn.getInputDimension()[0]);
+  in_tensor->setRandNormal();
+  nntrainer::sharedConstTensors input = {in_tensor};
+
+  auto label_tensor = MAKE_SHARED_TENSOR(label_shape);
+  label_tensor->setRandNormal();
+  nntrainer::sharedConstTensors label = {label_tensor};
 
   EXPECT_NO_THROW(nn.forwarding(input, label));
 


zhoonit (approved):
nice catch!

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -390,9 +390,13 @@ void GraphWatcher::compareFor(const std::string &reference,
...
-  nntrainer::sharedConstTensors input = {
-    MAKE_SHARED_TENSOR(nn.getInputDimension()[0])};
-  nntrainer::sharedConstTensors label = {MAKE_SHARED_TENSOR(label_shape)};
+  auto in_tensor = MAKE_SHARED_TENSOR(nn.getInputDimension()[0]);
+  in_tensor->setRandNormal();
+  nntrainer::sharedConstTensors input = {in_tensor};
+
+  auto label_tensor = MAKE_SHARED_TENSOR(label_shape);
+  label_tensor->setRandNormal();
+  nntrainer::sharedConstTensors label = {label_tensor};
```
myungjoo, line 399: If you need additional method call after initialization to avoid un-init-error, it appears that the constructor is incomplete. Are you sure that the current constructor that requires additional method call to complete initialization is complete enough?

kparichay: The current constructor allocates the memory block but does not initialize the memory (tensor is initialized but the memory it points to is not), to optimize for the scenario where the memory is to be loaded from the external file (example - loading weights from the saved file) or an operation will write over this memory (example - convolution or matmul. In case an operation requires its memory to be initialized/zeroed, it is performed by the operation).

@myungjoo Is it recommended to initialize the memory in the constructor? The constructor can be updated to initialize the memory by default, and optionally leave it uninitialized when required.

myungjoo: It's just me that is feeling weird to see two initialization functions are required for a single object, which is prone to human errors. Unless that is really necessary (e.g., the timing of "second" init is required much later than the "first" init),

```
a = constructor1(option1);
a.construct2(option2);
```
may become:
```
a = constructor_unified(option1, option2);
```

For the case option2 is determined much later, we may allow
```
a = constructor_unified(option1, DO_NOT_INIT_2ND_PART);
```
, which will keep some part uninitialized, which may incur ASSERT if some methods requiring this "2ND_PART" are called without calling ```construct2()```.


If these may be related with users' behaviors, not just our internal developers, this is going to be critical.

kparichay: From the point of internal usage of tensors, their memory is usually not initialized as mostly they are made for holding output of some operations and their memory will be overwritten by the result of the operation (except for weights which has its own wrapper to handle initialization, gradients, etc).

However, from the user's point of view, such an interface might be more intuitive. (note: in general, the user's interaction with tensor would come with input/output/weight of a model/layer. Either these tensors are already filled or the user has to fill them, and both of these mostly would not require initialization of tensors memory.) So, this will become like below:
```c++
// allocate tensor memory now with initializer (initializer defaults to none??)
T = Tensor(shape, alloc_now=true, initializer=none/zero/rand_norm/rand_uniform/void *mem)

// allocate tensor memory lazily (initializer not needed as underlying memory not allocated - give an error if the initializer is not none as a "void *buf" passed for initialization at construction might not be valid at the time of allocate.)
T = Tensor(shape, alloc_now=false)
// give initializer at the time of allocate
T.allocate(initializer=none/zero/rand_norm/rand_uniform)
```

myungjoo (approved):
LGTM. Because the author now understands the potential issue and this is a test case issue, we can move on.

test/unittest/unittest_nntrainer_models.cpp
kparichay: Updating the interface of the tensor will be tracked under #986 as the tensor is also supposed to be opened to the devel package.

jijoongmoon (approved):
LGTM

PR #1032, opened 2021-03-19
[optimizer] Add optimizer section to INI @open sesame 04/01 17:19

kparichay:
Add a new optimizer section to INI
This is essential to allow custom optimizers as the
properties of optimizers needs to be iterables.
Currently, optimizer needs to specified in the model, which
along being not the best design, limits the properties to
be set for the optimizer as the model loader has to know
which properties are for optimizer and which for model.

This patch separates the optimizer as a new section and updates
the unittests.

V2:
Added backward compatibility for optimizers.
Update ini loader for model to support more properties
than the ones which are hardcoded.

See also #986

**Self evaluation:**

### DIFF
--- a/nntrainer/models/model_loader.cpp
+++ b/nntrainer/models/model_loader.cpp
@@ -49,9 +49,17 @@ int ModelLoader::loadOptimizerConfigIni(dictionary *ini, NeuralNetwork &model) {
   int status = ML_ERROR_NONE;
 
   if (iniparser_find_entry(ini, "Optimizer") == 0) {
-    ml_logw("there is no [Optimizer] section in given ini file."
-            "This model can only be used for inference.");
+    if (!model.opt) {
+      ml_logw("there is no [Optimizer] section in given ini file."
+              "This model can only be used for inference.");
+    }
     return ML_ERROR_NONE;
+  }
+
+  /** Optimizer already set with deprecated method */
+  if (model.opt) {
+    ml_loge("Error: optimizers specified twice.");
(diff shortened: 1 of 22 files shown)


nntrainer/models/model_loader.cpp
```diff
@@ -70,61 +98,6 @@ int ModelLoader::loadModelConfigIni(dictionary *ini, NeuralNetwork &model) {
   model.batch_size =
     iniparser_getint(ini, "Model:Batch_Size", model.batch_size);
 
-  /** Default to adam optimizer */
-  const char *opt_type = iniparser_getstring(ini, "Model:Optimizer", "adam");
```
zhoonit, line 74: I think we should provide backward compatibility. There are number of people trying our library.

zhoonit: Would this be included to tizen 6.5m1? How about we change to tizen docs as well?

kparichay: will add backward compatibility.

zhoonit:
removing release critical for now

zhoonit:
LGTM. Could you add one test that contains old model format?

kparichay:
> LGTM. Could you add one test that contains old model format?

In `unittest_nntrainer_models`, only some of the ini models are update to match the updated model format. Others are still using the deprecated version.

zhoonit (approved):

jijoongmoon (approved):
LGTM

PR #1034, opened 2021-03-22
[rpm] Fix nnstreamer dependency @opensesame 03/30 01:49

### LINKED ISSUE #997: [tizen] nnstreamer.ini conf makes unable to test sub-plugin.
Currently, we have `nnstreamer-extra` is build-required for rpm.
bug gbs fail because tizen repo does not have nnstreamer-extra
`gbs build --define "unit_test 1"`

I wonder about the purpose of nnstreamer-extra. It just deletes `nnstreamer.ini` afaict

```spec
%if 0%{?unit_test}
BuildRequires:	nnstreamer-protobuf
BuildRequires:	nnstreamer-extra
BuildRequires:	gst-plugins-good-extra
BuildRequires:	python
%endif #unit_test
```

zhoonit:
This patch resolves nnstreamer dependency problem. Thus making rpm to
enable plugin related test.

This patch is wating for nnstreamer/nnstreamer#3147

resolves #997
**Self evaluation:**

### DIFF
--- a/packaging/nntrainer.spec
+++ b/packaging/nntrainer.spec
@@ -110,8 +110,7 @@ BuildRequires:	nnstreamer-devel
 %define enable_nnstreamer_tensor_filter -Denable-nnstreamer-tensor-filter=true
 
 %if 0%{?unit_test}
-BuildRequires:	nnstreamer-protobuf
-# BuildRequires:	nnstreamer-test-devel
+BuildRequires:	nnstreamer-test-devel
 BuildRequires:	gst-plugins-good-extra
 BuildRequires:	python
 %endif #unit_test
@@ -120,7 +119,7 @@ BuildRequires:	python
 
 Requires:	nntrainer-core = %{version}-%{release}
 
-%if  0%{?nnstreamer_filter}
+%if 0%{?nnstreamer_filter}
 Requires:	nnstreamer-nntrainer = %{version}-%{release}
 %endif #nnstreamer_filter
 %if %{with tizen}
@@ -161,6 +160,7 @@ Requires:	nntrainer = %{version}-%{release}
 Requires:	iniparser
 Requires:	%{capi_machine_learning_inference}
 Requires:	nnstreamer-tensorflow-lite
+BuildRequires:  nnstreamer-test-devel
 BuildRequires:	nnstreamer-tensorflow-lite
 BuildRequires:	tensorflow-lite-devel
 BuildRequires:	pkgconfig(jsoncpp)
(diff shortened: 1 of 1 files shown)


packaging/nntrainer.spec
```diff
@@ -321,44 +321,17 @@ meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} \
...
-# export NNSTREAMER_CONF=$(pwd)/test/nnstreamer/nnstreamer-test.ini
-# export NNSTREAMER_FILTERS=$(pwd)/build/nnstreamer/tensor_filter
-# pushd test/nnstreamer
-# bash runTest.sh
-# popd
-# %endif #nnstreamer_filter
+# todo: migrate this to meson test soon
+%if 0%{?nnstreamer_filter}
+pushd test/nnstreamer
+bash runTest.sh
```
myungjoo, line 332: Why are you running runTest.sh instead of ssat?

zhoonit: I wonder as well, replaced this to `ssat`

packaging/nntrainer.spec
```diff
@@ -321,44 +321,18 @@ meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} \
...
 
 # unittest for nntrainer plugin for nnstreamer
-# %if 0%{?nnstreamer_filter}
-# export NNSTREAMER_CONF=$(pwd)/test/nnstreamer/nnstreamer-test.ini
-# export NNSTREAMER_FILTERS=$(pwd)/build/nnstreamer/tensor_filter
-# pushd test/nnstreamer
+# todo: migrate this to meson test soon
+%if 0%{?nnstreamer_filter}
+pushd test/nnstreamer
+ssat
```
myungjoo, line 332: You may need to add a few options.
If you add postfixes (e.g., ```_n```) for negative cases, add ```-cn _n```
If you want to generate report after testing, add ```--summary <filename>```
If you want to avoid broken texts in OBS log, add ```-n```
and so on...

Anyway, this is not that critical, thus, I'll mark LGTM

myungjoo (approved):
LGTM

packaging/nntrainer.spec
```diff
@@ -321,44 +321,18 @@ meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} \
...
-%{meson_test} app_knn
-%{meson_test} app_logistic
-%{meson_test} app_mnist
-%{meson_test} app_cifar_classification
-%{meson_test} app_cifar_classification_func
-# %{meson_test} app_draw_classification
-%{meson_test} app_plugin_test
-%{meson_test} simpleshot_tests
-
+export NNSTREAMER_CONF=$(pwd)/test/nnstreamer/nnstreamer-test.ini
```
kparichay, line 324: I wonder if this is only required for ssat test, then this can be moved to inside of `%if 0%{?nnstreamer_filter}` as meson test sets their environment variables when defining the tests.

kparichay (approved):

packaging/nntrainer.spec
zhoonit: DrawClassification also requires this for now. 

There is a minor issue with including conf (which is only needed for the tizen build) so it is left for now. 
Later, `nnstremaer-test.ini` will be deprecated in favor of `nnstreamer-test.ini.in`, at the same time, this will migrate into meson test.

kparichay: sounds good.

zhoonit:
#3147 merged, so this can go group sr today.

jijoongmoon (approved):
LGTM

PR #1035, opened 2021-03-22
[ ActFunc ] Separate Activation Function from Activation Layer

jijoongmoon:
In order to use Activation function method in other Layer, it is
better to make sepearte the activation function definition and Layer.
In this PR, new ActiFunc Class is introduced and midified the
activation layer to use it.
RNN Layer also use this as an private variable.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -2,88 +2,77 @@
 /**
  * Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
  *
- * @file   activation_layer.h
- * @date   17 June 2020
+ * @file   acti_func.cpp
+ * @date   22 March 2021
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
  * @bug    No known bugs except for NYI items
- * @brief  This is Activation Layer Class for Neural Network
+ * @brief  This is Activation Function Class for Neural Network
  *
  */
 
-#ifndef __ACTIVATION_LAYER_H__
-#define __ACTIVATION_LAYER_H__
+#ifndef __ACTI_FUNC_H__
+#define __ACTI_FUNC_H__
 #ifdef __cplusplus
 
-#include <acti_func.h>
-#include <layer_internal.h>
 #include <tensor.h>
 
 namespace nntrainer {
 
 /**
- * @class   Activation Layer
- * @brief   Activation Layer
+ * @brief     Enumeration of activation function type
  */
-class ActivationLayer : public Layer {
+enum class ActivationType {
+  ACT_TANH,    /** tanh */
+  ACT_SIGMOID, /** sigmoid */
+  ACT_RELU,    /** ReLU */
+  ACT_SOFTMAX, /** softmax */
+  ACT_NONE,    /** no op */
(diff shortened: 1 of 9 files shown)


zhoonit (approved):
LGTM :)

lhs8928 (approved):
LGTM

nntrainer/layers/acti_func.h
```diff
@@ -0,0 +1,196 @@
...
+public:
+  /**
+   * @brief     Constructor of ActiFunc
+   */
+  ActiFunc(ActivationType at = ActivationType::ACT_NONE) { setActiFunc(at); }
+
+  /**
+   * @brief     Destructor of ActiFunc
+   */
+  ~ActiFunc(){};
```
kparichay, line 50: = default

nntrainer/layers/acti_func.h
```diff
@@ -0,0 +1,196 @@
...
+   */
+  Tensor &run_prime_fn(Tensor &in, Tensor &ret, Tensor const &deriv);
+
+  /**
+   * @brief       Calculate softmax for Tensor Type
+   * @param[in] x Tensor
+   * @param[out] output output Tensor
+   * @retval      Tensor
+   */
+  static Tensor &softmax(Tensor const &x, Tensor &output);
```
kparichay, line 83: If the purpose of this class is to expose an interface to be able to create activation functions and run the forward/backward given input/output, then let's make these static activation functions private, and if any layer wants to use these functions, they must create an object of this class.

jijoongmoon: > If the purpose of this class is to expose an interface to be able to create activation functions and run the forward/backward given input/output, then let's make these static activation functions private, and if any layer wants to use these functions, they must create an object of this class.

I'm not sure about this. The loss layer also use this Activation function.. In order to change to use actifunc object, then we have to make the wrapper of each function. and I don't think it is no different to use these function as public.

nntrainer/layers/acti_func.h
jijoongmoon: done.

nntrainer/layers/acti_func.h
kparichay: If we intend to directly use the functions given in this class, then can you please elaborate on the purpose of this class. 
As per my understanding of the PR description, it seems like this class facilitates the use of activation functions in other layers.

jijoongmoon: > , it seems like this class facilitates the use of activation functions in other layers.
It is the purpose of this class. These functions was in ActivationLayer Class and it was public. That's because loss layer use them directly. What I'm saying is it Is too much wrapper code if we make them private and make public code to use directly. I think it is ok as it is now.

kparichay: I agree that using ActiFunc obj with acti_func.h header is a good proposal and makes structure cleaner.
However, if the usage of activation functions is still directly using the ActiFunc obj (like in loss layer), then this class makes little difference.
I would prefer if loss layer can also be updated to have a ActiFunc obj just like RNNLayer will have. And now, if all users of ActiLayer will create its object, then there is no need to open individual activation functions as public.

jijoongmoon: This can be done in another PR then. I will make issue for it.( #1040 )

kparichay: alright.

kparichay (approved):
LGTM

PR #1038, opened 2021-03-23
[ RNN ] Implement initialization and forwarding of RNN Layer

jijoongmoon:
- [ RNN ] Add initialization of RNN

```
Add Initialization of RNN Layer
- Set dimension and Weight
- Add Unittest for initialization

**Self evaluation:**

```

- [ RNN ] Implementaion of forwarding
```
- Implementation of RNN forwarding.

**Self evaluation:**

```
- [ RNN ] support various activation function

```
In this PR, various activation functions are supported using ActiFunc
Class.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -28,6 +28,7 @@
 #include <output_layer.h>
 #include <parse_util.h>
 #include <profiler.h>
+#include <rnn.h>
 
 namespace nntrainer {
 
@@ -260,6 +261,12 @@ int NetworkGraph::realizeActivationType(Layer &current) {
   int status = ML_ERROR_NONE;
 
   ActivationType act = current.getActivationType();
+
+  if (current.getType() == RNNLayer::type) {
+    // No need to add activation layer for RNN Layer
+    current.setActivation(act);
+    return status;
+  }
 
   if (act == ActivationType::ACT_NONE) {
     /// ActivationType::ACT_NONE does not need realization
--- a/nntrainer/layers/rnn.cpp
+++ b/nntrainer/layers/rnn.cpp
@@ -65,6 +65,7 @@ int RNNLayer::initialize(Manager &manager) {
                          weight_regularizer_constant, true, "RNN:weight_hh");
     weights.emplace_back(bias_dim, bias_initializer, WeightRegularizer::NONE,
                          1.0f, true, "RNN:bias_h");
+    manager.trackWeights(weights);
   } else {
     weights[RNNParams::weight_xh].reset(dim_xh, weight_initializer,
                                         weight_regularizer,
@@ -75,6 +76,10 @@ int RNNLayer::initialize(Manager &manager) {
     weights[RNNParams::bias_h].reset(bias_dim, bias_initializer,
                                      WeightRegularizer::NONE, 1.0f, true);
   }
+
+  bias_dim.batch(input_dim[0].batch());
+  h_prev = Tensor(bias_dim);
+  h_prev.setZero();
 
   return status;
 }
@@ -87,6 +92,13 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
       status = setUint(unit, value);
       throw_status(status);
       output_dim[0].width(unit);
+    }
+    break;
+  case PropertyType::activation:
+    if (!value.empty()) {
+      ActivationType acti_type = (ActivationType)parseType(value, TOKEN_ACTI);
+      Layer::activation_type = acti_type;
+      acti_func.setActiFunc(acti_type);
     }
     break;
   default:
@@ -107,25 +119,39 @@ void RNNLayer::forwarding(bool training) {
   Tensor &hidden_ = net_hidden[0]->getVariableRef();
   Tensor &input_ = net_input[0]->getVariableRef();
 
+  Tensor temp;
+  Tensor hs_prev;
+  Tensor hs;
+
+  // TODO : check merge b and t index
   for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
     Tensor islice = input_.getBatchSlice(b, 1);
     Tensor oslice = hidden_.getBatchSlice(b, 1);
 
     for (unsigned int t = 0; t < islice.height(); ++t) {
-      Tensor xs = Tensor(TensorDim(1, 1, 1, islice.width()),
-                         islice.getAddress(t * islice.width()));
-      unsigned int id = 0;
+      Tensor xs = input_.getSharedDataTensor(TensorDim(1, 1, 1, islice.width()),
+                                             t * islice.width());
+      // Calculate Hidden
+      // activation(xs.dot(weight_xh).add(hs_prev.dot(weight_hh).add(bias_h)));
+      hs = oslice.getSharedDataTensor(TensorDim(1, 1, 1, oslice.width()),
+                                      t * oslice.width());
       if (t > 0) {
-        id = t - 1;
+        hs_prev = oslice.getSharedDataTensor(TensorDim(1, 1, 1, oslice.width()),
+                                             (t - 1) * oslice.width());
+      } else {
+        hs_prev = h_prev.getBatchSlice(b, 1);
       }
-      Tensor hs = Tensor(TensorDim(1, 1, 1, oslice.width()),
-                         oslice.getAddress(t * oslice.width()));
-      Tensor hs_prev = Tensor(TensorDim(1, 1, 1, oslice.width()),
-                              oslice.getAddress(id * oslice.width()));
-      // Calculate hs_t = tanh(Whh*h_(t-1) + Wxh*X_t))
-      hs = xs.dot(weight_xh).add(hs_prev.dot(weight_hh).add(bias_h));
+
+      hs_prev.dot(weight_hh, temp);
+
+      xs.dot(weight_xh, hs);
+      temp.add_i(bias_h);
+
+      hs.add_i(temp);
+      // TODO : In-place calculation for activation
       acti_func.run_fn(hs, hs);
     }
+    h_prev.getBatchSlice(b, 1).copy(hs);
   }
 }
 
@@ -142,11 +168,6 @@ void RNNLayer::calcDerivative() {
 
 void RNNLayer::calcGradient() {
   // NYI
-}
-
-void RNNLayer::setActivation(ActivationType acti_type) {
-  Layer::setActivation(acti_type);
-  acti_func.setActiFunc(acti_type);
 }
 
 } // namespace nntrainer
--- a/nntrainer/layers/rnn.cpp
+++ b/nntrainer/layers/rnn.cpp
@@ -11,6 +11,7 @@
  *
  */
 
+#include <cmath>
 #include <layer_internal.h>
 #include <lazy_tensor.h>
 #include <nntrainer_error.h>
@@ -95,7 +96,38 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
   }
 }
 
-void RNNLayer::forwarding(bool training) {}
+void RNNLayer::forwarding(bool training) {
+  Tensor &weight_xh =
+    weightAt(static_cast<int>(RNNParams::weight_xh)).getVariableRef();
+  Tensor &weight_hh =
+    weightAt(static_cast<int>(RNNParams::weight_hh)).getVariableRef();
+  Tensor &bias_h =
+    weightAt(static_cast<int>(RNNParams::bias_h)).getVariableRef();
+
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
+    Tensor islice = input_.getBatchSlice(b, 1);
+    Tensor oslice = hidden_.getBatchSlice(b, 1);
+
+    for (unsigned int t = 0; t < islice.height(); ++t) {
+      Tensor xs = Tensor(TensorDim(1, 1, 1, islice.width()),
+                         islice.getAddress(t * islice.width()));
+      unsigned int id = 0;
+      if (t > 0) {
+        id = t - 1;
+      }
+      Tensor hs = Tensor(TensorDim(1, 1, 1, oslice.width()),
+                         oslice.getAddress(t * oslice.width()));
+      Tensor hs_prev = Tensor(TensorDim(1, 1, 1, oslice.width()),
+                              oslice.getAddress(id * oslice.width()));
+      // Calculate hs_t = tanh(Whh*h_(t-1) + Wxh*X_t))
+      hs =
+        xs.dot(weight_xh).add(hs_prev.dot(weight_hh).add(bias_h)).apply(tanh);
+    }
+  }
+}
 
 void RNNLayer::copy(std::shared_ptr<Layer> l) {
   Layer::copy(l);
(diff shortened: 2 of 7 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -643,11 +643,17 @@ int NeuralNetwork::train_run() {
...
-      knn->net_hidden[0]->getGradientRef() = label;
+    layer_ = layer_node.layer;
+    if (layer_->getType() == "centroid_knn") {
+      layer_->net_hidden[0]->getGradientRef() = label;
+    }
+
+    if (layer_->getType() == "RNN"){
+      Tensor &var_ref= layer_->net_hidden[0]->getVariableRef();
+      // for now, we are using Xavir_normal for hidden initialization
+      var_ref.setRandNormal(0.0f, sqrtFloat(2.0f / (var_ref.width() + var_ref.height())));
```
kparichay: 1. `HiddenLayer value` must be a weight for RNN, and not the output. If it is an output, it can be overwritten by other layers/derivatives, but that value is needed for the next iteration. Further, during inference, its value would be overwritten due to optimization but hidden value is needed for the next iteration.
2. Initialization of hidden value must be done inside the RNN layer itself, and not by the model. `centroid_knn` does it in the model which is wrong. It requires to pass the label for the training/validation and should be done by forwarding/backwarding like done for loss layer.

nntrainer/layers/rnn.cpp
```diff
@@ -95,7 +96,38 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
...
+    weightAt(static_cast<int>(RNNParams::weight_xh)).getVariableRef();
+  Tensor &weight_hh =
+    weightAt(static_cast<int>(RNNParams::weight_hh)).getVariableRef();
+  Tensor &bias_h =
+    weightAt(static_cast<int>(RNNParams::bias_h)).getVariableRef();
+
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
```
kparichay, line 110: why use loops? Simply reshape to `batch * seq_length, 1, 1, input_dim[0].width()` and dot product with the weight.
Current code creates a lot of temporary memory allocations and is complicated making it prone to errors.

nntrainer/layers/rnn.cpp
```diff
@@ -144,4 +144,9 @@ void RNNLayer::calcGradient() {
   // NYI
 }
 
+void RNNLayer::setActivation(ActivationType acti_type) {
+  Layer::setActivation(acti_type);
```
kparichay, line 148: please override `setProperty` and add this in `setProperty`.
`setActivation` will soon be removed.

nntrainer/layers/rnn.cpp
```diff
@@ -95,7 +96,38 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
...
+
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
+    Tensor islice = input_.getBatchSlice(b, 1);
+    Tensor oslice = hidden_.getBatchSlice(b, 1);
+
+    for (unsigned int t = 0; t < islice.height(); ++t) {
+      Tensor xs = Tensor(TensorDim(1, 1, 1, islice.width()),
```
zhoonit, line 115: Code clean: how about `TensorDim(islice.width())`?

nntrainer/layers/rnn.cpp
```diff
@@ -95,7 +96,38 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
   }
 }
 
-void RNNLayer::forwarding(bool training) {}
+void RNNLayer::forwarding(bool training) {
```
zhoonit, line 99: As far as I understand, RNN retains the (last) previous output and use it when inference mode. I wonder if this behavior is considered

nntrainer/layers/rnn.h
```diff
@@ -93,10 +93,13 @@ class RNNLayer : public Layer {
   void setProperty(const PropertyType type,
                    const std::string &value = "") override;
 
+  void setActivation(ActivationType acti_type) override;
+
   static const std::string type;
 
 private:
   unsigned int unit;
+  ActiFunc acti_func;
```
kparichay, line 102: Please set the default activation function to `TanH`. RNN defaults to TanH activation and a user specifying RNN without activation must still remain a valid RNN.

nntrainer/layers/rnn.cpp
```diff
@@ -95,7 +96,38 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
...
+    Tensor oslice = hidden_.getBatchSlice(b, 1);
+
+    for (unsigned int t = 0; t < islice.height(); ++t) {
+      Tensor xs = Tensor(TensorDim(1, 1, 1, islice.width()),
+                         islice.getAddress(t * islice.width()));
+      unsigned int id = 0;
+      if (t > 0) {
+        id = t - 1;
+      }
+      Tensor hs = Tensor(TensorDim(1, 1, 1, oslice.width()),
```
zhoonit, line 121: Tensor(Tensordim, float * data) will create a new copy so I think this will not update hidden.

Could you make a small `model` test to verify the output?

nntrainer/layers/rnn.cpp
```diff
@@ -123,8 +123,8 @@ void RNNLayer::forwarding(bool training) {
       Tensor hs_prev = Tensor(TensorDim(1, 1, 1, oslice.width()),
                               oslice.getAddress(id * oslice.width()));
       // Calculate hs_t = tanh(Whh*h_(t-1) + Wxh*X_t))
-      hs =
-        xs.dot(weight_xh).add(hs_prev.dot(weight_hh).add(bias_h)).apply(tanh);
+      hs = xs.dot(weight_xh).add(hs_prev.dot(weight_hh).add(bias_h));
+      acti_func.run_fn(hs, hs);
```
kparichay, line 127: Please add a check in the graph that the next layer of the RNN must not be an activaiton layer.

nntrainer/graph/network_graph.cpp
```diff
@@ -227,6 +228,12 @@ int NetworkGraph::realizeActivationType(Layer &current) {
 
   ActivationType act = current.getActivationType();
 
+  if (current.getType() == RNNLayer::type) {
+    // No need to add activation layer for RNN Layer
+    current.setActivation(act);
```
kparichay, line 233: `none` activation layer is also supported?

jijoongmoon: I'm thinking of. but later PR, default activation is tanh.

nntrainer/layers/rnn.cpp
jijoongmoon: could you point out more specific? previous output??

nntrainer/graph/network_graph.cpp
kparichay: I think `none` must not be supported. Can be opened later if requested.

nntrainer/layers/rnn.cpp
jijoongmoon: I'm not thinking about any optimization here. at least this PR. We do not have any feature about the RNN. It will update one by one.

nntrainer/layers/rnn.cpp
jijoongmoon: done

nntrainer/layers/rnn.cpp
jijoongmoon: We discuss offline. I will update.

nntrainer/layers/rnn.h
jijoongmoon: come up this next PR

nntrainer/layers/rnn.cpp
kparichay: The current code is quite difficult to understand. This can be updated to be written functionally within 3-4 lines which are quite easy to reason about and review.

jijoongmoon: Ok. I will try to make it easier then. Besides, I think loop in batch is necessary because of update the hidden vector between batch.

nntrainer/layers/rnn.cpp
jijoongmoon: Right. I will update this.

nntrainer/layers/rnn.cpp
jijoongmoon: In network graph, there is realizeActivationLayer to create additional activation layer. but for RNN Layer, there is no need to do it.. so.. it just set default tanh if type is ACT_NONE. Talking about the user intended addition activation layer.. I think we don't need to care about. That's user intension. Also to calculate softmax.. user also have responsibility to add Fully Connected Layer. I think the restriction of activation layer is totally depending on user.

nntrainer/layers/rnn.cpp
jijoongmoon: I will. Thanks.

nntrainer/layers/rnn.cpp
kparichay: You can refer to the FC layer implementation or use FC layer inside RNN as well :)
The operation performed here is exactly the same as FC layer with an extra dimension (which can be merged with batchsize for simplicity).

nntrainer/layers/rnn.h
kparichay: Great.

nntrainer/layers/rnn.cpp
kparichay: The issue is that RNN uses activation function internally which is being operated in-place in the current code.
So, if an activation layer is added after it, which is also optimized to work in-place, it will create issues.
If there is no such restriction of activation layer being added to the graph, please add a `TODO:` in `inPlaceOptimize` in the graph to ensure that if the input layer of activation is RNN or RNN has activation internally (to be more precise), then that activation must not be inplace. This will later be updated to each layer provided `isOperatingInPlace` with #867.

nntrainer/layers/rnn.cpp
jijoongmoon: > You can refer to the FC layer implementation or use FC layer inside RNN as well :)
> The operation performed here is exactly the same as FC layer with an extra dimension (which can be merged with batchsize for simplicity).

I think there is more than that. This is updated.

nntrainer/layers/rnn.cpp
jijoongmoon: Let's see. how it works. RNNLayer still in the middle of development. so this kind of optimization can be considered later.

nntrainer/models/neuralnet.cpp
jijoongmoon: right. it removed. Thanks.

nntrainer/layers/rnn.cpp
kparichay: Alright. Please add a TODO as comments to keep it as a reminder.

nntrainer/layers/rnn.cpp
```diff
@@ -54,7 +105,66 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
...
+  Tensor temp;
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
+    Tensor islice = input_.getBatchSlice(b, 1);
+    Tensor oslice = hidden_.getBatchSlice(b, 1);
+    float *hidden_next = oslice.getData();
+
+    if (!training) {
+      if (b > 0) {
+        float *hidden_pre =
```
kparichay, line 127: I am a little confused. I think the hidden layer is not shared across the batch. Each batch has its own hidden information in a generic RNN.

kparichay:
```python
from tensorflow.keras.layers import Input, SimpleRNN
in = Input(shape=(3, 1), batch_size=10)
print(in.shape) # TensorShape([10, 3, 1])
rnn = SimpleRNN(5)(in) # can also enable return_state=True to get the state
print(rnn.shape) # TensorShape([10, 5]) - hidden cell value
```
The hidden state between batch size should not share their values.

jijoongmoon: for the inference, I guess it should. but backwarding, it's not.

kparichay: IMHO, even for inference, it should be not.

zhoonit: There are number of points that needs clarification.

I think there are two ways.

1. Keep states for the batch only 🌴 
2. Presist states 🐟 

The one I am referencing suggested to train with 🌴 and inference 🐟, Jijoong told me the implementation is done in 🐟  but later add `stateful=false` as a property.

keras seems to have two options.
https://keras.io/guides/working_with_rnns/#crossbatch-statefulness

nntrainer/models/neuralnet.cpp
```diff
@@ -632,10 +632,11 @@ int NeuralNetwork::train_run() {
       layer_->net_hidden[0]->getGradientRef() = label;
     }
 
-    if (layer_->getType() == "RNN"){
-      Tensor &var_ref= layer_->net_hidden[0]->getVariableRef();
+    if (layer_->getType() == "RNN") {
+      Tensor &var_ref = layer_->net_hidden[0]->getVariableRef();
```
zhoonit, line 636: I wonder if we can handle this on `RNN::initialize()`.

zhoonit (approved):
LGTM

nntrainer/layers/rnn.cpp
kparichay: Stateful and batch are very different concepts.
Stateful talks about keeping the hidden state across different runs.
Batch is about different data points running in parallel.

The behavior of RNN remains the same in inference and train for the batch and stateful. NNTrainer's RNN's default behavior must match other frameworks default behavior to avoid confusion.

> https://keras.io/guides/working_with_rnns/#crossbatch-statefulness

Although the name is `crossbatch`, the state is independent across the batch. This is just a way to selectively reset the state when needed. This does not shares the state across the batch (multiple data points given at once). The `batch` terminology used in the article refers to `iteration`, and it is equivalent to a simple behavior, where we can relinquish the power to reset the hidden state to the user.

nntrainer/models/neuralnet.cpp
```diff
@@ -625,11 +625,18 @@ int NeuralNetwork::train_run() {
...
-      knn->net_hidden[0]->getGradientRef() = label;
+    layer_ = layer_node.layer;
+    if (layer_->getType() == "centroid_knn") {
+      layer_->net_hidden[0]->getGradientRef() = label;
+    }
+
+    if (layer_->getType() == "RNN") {
+      Tensor &var_ref = layer_->net_hidden[0]->getVariableRef();
+      // for now, we are using Xavir_normal for hidden initialization
+      var_ref.setRandNormal(
```
kparichay, line 638: is this even required anymore?

nntrainer/layers/rnn.cpp
```diff
@@ -54,7 +109,49 @@ void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
...
+    weightAt(static_cast<int>(RNNParams::bias_h)).getVariableRef();
+
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  Tensor temp;
+  Tensor hs_prev;
+  Tensor hs;
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
```
kparichay, line 126: Please add a TODO here to try and transpose by the batch dimension and see if its faster that this implementation.

nntrainer/models/neuralnet.cpp
jijoongmoon: removed

nntrainer/models/neuralnet.cpp
jijoongmoon: removed, thanks.

nntrainer/layers/rnn.cpp
jijoongmoon: done.

kparichay (approved):
LGTM
Please remember that this version does not reset the hidden state per iteration, and behaves like a stateful RNN.

PR #1049, opened 2021-03-26
[graph] More refactors for graph + bug fix @open sesame 04/01 11:12

kparichay:
**Commit 1:** [graph] Update setting output layers in graph + bug fix 
Update setting output layers in graph to a unified location.
The graph is first made with just input layers.
Output layers are set at the end for the completeness at
one go than adding at each realization to avoid more errors.

Also, added a bug fix where the last layer in  the model
is not the last layer in the ini file. Now the last layer
is not the last layer in the adj but the layer without any
output_layers.

**Commit 2**: [graph] Remove graph edges after compile
This patch removes edges from the graph after compiling to
save the memory. This graph can be reconstructed by adding those edges
again from the information in the layers themselves.

See also #986

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -216,13 +216,13 @@ int NetworkGraph::realizeMultiInputType(Layer &current) {
   for (unsigned int i = 0; i < current.input_layers.size(); ++i)
     layer->input_layers.push_back(current.input_layers[i]);
 
-  layer->setNumOutputs(1);
-  layer->output_layers.clear();
-  layer->output_layers.push_back(current.getName());
+  layer->setNumOutputs(current.getNumOutputs());
 
-  current.setNumInputs(1);
+  current.setNumInputs(layer->getNumOutputs());
   current.input_layers.clear();
   current.input_layers.push_back(layer->getName());
+  /** output layers for layer obj will be set in setOutputLayers() */
+
   addLayerNode(layer);
 
   return status;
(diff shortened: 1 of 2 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -414,45 +412,34 @@ void NetworkGraph::setOutputLayers() {
...
     if (layer_idx->getNumOutputs() != layer_idx->output_layers.size()) {
-      layer_idx->setNumOutputs(layer_idx->output_layers.size());
+      if (layer_idx->output_layers.size() == 0) {
+        /** No output layer inplies its the last layer */
+        layer_idx->setNumOutputs(1);
+        layer_idx->output_layers.clear();
+        layer_idx->output_layers.push_back("__exit__");
+        last_layer_count += 1;
+      } else if (layer_idx->getNumOutputs() < layer_idx->output_layers.size()) {
+        /** this if the multi-output layer */
```
zhoonit, line 430: is `layer_idx->getNumOutputs() < `layer_idx->output_layers.size()` holds for multi-output layer? I wonder as realizeMultiOutput is setting the output.

nntrainer/graph/network_graph.cpp
```diff
@@ -287,23 +277,38 @@ int NetworkGraph::realizeActivationType(Layer &current) {
...
+  if (current.getNumOutputs() == 1)
+    return ML_ERROR_NONE;
+
+  std::shared_ptr<Layer> layer = nntrainer::createLayer(OutputLayer::type);
+  ensureName(layer, current.getName());
+
+  layer->setNumInputs(current.getNumInputs());
+  layer->input_layers.clear();
+  layer->input_layers.push_back(current.getName());
+  layer->setNumOutputs(current.output_layers.size());
```
zhoonit, line 310: I wonder if setting output here is intended as the commit msg suggests differently.

nntrainer/graph/network_graph.cpp
kparichay: Yes. `RealizeMultiOutput` has a different job. It is to insert an `outputlayer` and set this correctly. However, if an `outputlayer` is already set, then its `numOutputs` and `output_layers.size()` don't match.

nntrainer/graph/network_graph.cpp
kparichay: Yes, only the number of output layers is set. The output layers (connections in the graph) is set later.

zhoonit: I see. Thank you for the explanation :)

nntrainer/graph/network_graph.cpp
zhoonit: Understood 👍

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1050, opened 2021-03-26
[api] Sync cc-API with c-API @open sesame 04/01 11:12

kparichay:
1. Update layer cpp API to match the c-api interface. Move pluggableLayer related stuff to layer_internal.
2. Sync the cc-API of model with the c-API. Some unnecessary functions have been removed. Some others have been renamed.
3. Update dataset cc-api to match with the c-api. Also remove extra redundant functionality
4. Update c-api implementation to only depend on cc-api. Also add some minor additional functions in the cc-api to support the c-api functionality.

See Also #986

**Self evaluation:**


api/ccapi/include/layer.h
```diff
@@ -63,159 +63,56 @@ class Layer {
...
-   * @brief print using PrintPreset
-   *
-   * @param out oustream
-   * @param preset preset to be used
+   * @brief     set Property of layer
+   * @param[in] values values of property
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
-  virtual void printPreset(std::ostream &out,
```
zhoonit, line 217: shouldn't print remain here?

api/ccapi/include/model.h
```diff
@@ -91,19 +77,19 @@ class Model {
...
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
-  virtual int initialize() = 0;
+  virtual int compile() = 0;
 
   /**
-   * @brief     Compile Network. This should be called before initialize
+   * @brief     Initialize Network. This should be called after setting the
+   * property and compiling. hyperparameters.
```
zhoonit, line 90: I wonder about the last word `hyperparameters.`

Applications/MNIST/jni/main.cpp
```diff
@@ -334,7 +334,6 @@ int main(int argc, char *argv[]) {
     model->train();
     training_loss = model->getTrainingLoss();
     validation_loss = model->getValidationLoss();
-    last_batch_loss = model->getLoss();
```
zhoonit, line 337: Q. is getLoss gone?

api/ccapi/include/dataset.h
```diff
@@ -66,30 +67,26 @@ class Dataset {
...
-   * @param[in] path file path
+   * @brief     set property
+   * @param[in] values values of property
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   * @details   Properties (values) is in the format -
+   *  { std::string property_name, std::string property_val, ...}
    */
-  virtual int setDataFile(DatasetDataType type, std::string path) = 0;
+  virtual int setProperty(std::vector<std::string> values) = 0;
```
zhoonit, line 77: `const std::vector<> &` ? or `std::vector<>& `?

api/ccapi/include/dataset.h
```diff
@@ -66,30 +67,26 @@ class Dataset {
...
+   * user_data for callbacks
    * @param[in] values values of property
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   * @note      this is a superset of the setProperty(std::vector<std::string>)
+   * @details   Properties (values) is in the format -
+   *  { std::string property_name, void * property_val, ...}
    */
-  virtual int setProperty(std::vector<std::string> values) = 0;
+  virtual int setProperty(std::vector<void *> values) = 0;
```
zhoonit, line 89: Is this because of the callback?

api/capi/src/nntrainer.cpp
```diff
@@ -638,7 +641,7 @@ int ml_train_optimizer_create(ml_train_optimizer_h *optimizer,
 
   returnable f = [&]() {
     nnopt->optimizer =
-      nntrainer::createOptimizer(ml_optimizer_to_nntrainer_type(type));
+      ml::train::createOptimizer((ml::train::OptimizerType)type);
```
zhoonit, line 644: How about unifying to `ml_train_*_type_e` in the near future? 🐳 

It seems that using `ml::train::OptimizerType` is to exploit enum class. But having one typedef enum seems better than having same `typedef enum` and `enum class`.

api/ccapi/include/layer.h
kparichay: The ccAPI for now must be equivalent to cAPI.

Applications/MNIST/jni/main.cpp
kparichay: `getLoss` returns the loss for the last batch, which by definition is strange.
If certain training characteristics are required, the `metrics` struct should be returned with all the characteristics.

api/ccapi/include/dataset.h
kparichay: yes.

Applications/MNIST/jni/main.cpp
zhoonit: Yes. Sounds reasonable.

api/capi/src/nntrainer.cpp
kparichay: ml_train_optimizer_type_e is supposed to be a subset of ml::train::OptimizerType. Any new introduction of new optimizer will be directly added to ccAPI (keeping ccAPI always up to date). However, it will be added to cAPI separately as per the release requirement. If they are equal, they can be an alias to each other.

api/ccapi/include/layer.h
zhoonit: Acked

api/capi/src/nntrainer.cpp
kparichay: Just an FYI: `ml::train::OptimizerType` is also `typedef enum` and not  an `enum class`

api/ccapi/include/model.h
kparichay: done.

api/ccapi/include/dataset.h
kparichay: done.

zhoonit (approved):
LGTM after rebase :)

api/ccapi/include/dataset.h
zhoonit: Okay

jijoongmoon:
So.. capi is only use ccapi?

jijoongmoon (approved):
LGTM

PR #1051, opened 2021-03-29
[Pooling2D] Rework forward/backward

### LINKED ISSUE #825: Make padding zero cost
Lets make padding zero cost for pooling , convolution or any future layers.
Lets say convolution needs padding, then the input tensor `In_c` allocated for this must  be tensor of padded size (input shape  + padded). The previous layer which treats `In_c` as its output tensor will have the same memory but a tensor with different shape and stride (which does not include padding). In this way, once the previous layer fills in the output, convolution layer automatically receives padded input without any extra cost.

The same optimization will be done for pooling and for backwarding as well.
This not only reduces the cost of the memcopy of the padding but also the temporary memory allocation overhead and its corresponding rise in peak memory is also removed.

### LINKED ISSUE #1045: Add test Pooling with padding
AFAICT, pooling2d with padding test is missing, add the test.

zhoonit:
With this series of commit, number of bugs from pooling layer are fixed and tests are added for that.

- [Pooling2D/Test] Add model tests
```
Add model test for various types of pooling2d

**Self evaluation:**

```
- [Pooling2d] Rework pooling to avoid padding
```
This patch
1. set up scaffolding for the pooling to reuse some loop.
2. initialize index vector when setting batch (fixes a hidden
bug)
3. implements max pooling without additional memory allocation

**Self evaluation:**

```
- [Pooling2d] Rework Average Pooling
```
This patch refactors average pooling layer to not rely
on the pooling loop.

From this patch, max_idx is used to remember number of
effective elements used for the average calculation

Also add some checks on the `initailize`,
Also global average pooling is merged to average pooling

**Self evaluation:**

```
- [Pooling2D] Rework global max pooling
```
- Rework global max pooling to work with lambda function
- Remove switch cases for the main loop in pooling

**Self evaluation:**

```

resolves #825 resolves #1045


test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -751,6 +751,90 @@ INI conv_none_loss_validate(
   }
 );
 
+INI pooling_max_same_padding(
+  "pooling_max_same_padding",
+  {
+    nn_base + "learning_rate=0.1 | optimizer=sgd | loss=cross | batch_size=3",
+        I("input") + input_base + "input_shape=2:5:3",
```
kparichay, line 758: It would be good to have these tests for 2:3:5 to ensure the working along both height and width dimensions. These extra test cases can be skipped for global pooling versions.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -55,9 +55,24 @@ int Pooling2DLayer::initialize(Manager &manager) {
...
+    max_idx_global.resize(out_dim.batch() * output_dim[0].getFeatureLen());
+  }
+
   return status;
 }
 
+void Pooling2DLayer::setBatch(unsigned int batch) {
+  Layer::setBatch(batch);
+  if (pooling_type == PoolingType::max) {
+    max_idx.reserve(batch * output_dim[0].getFeatureLen());
```
kparichay, line 101: unless you really want this to be here, please don't allocate memory in setBatch.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -55,9 +55,24 @@ int Pooling2DLayer::initialize(Manager &manager) {
     out_dim.width(in_dim.channel());
   }
 
+  if (pooling_type == PoolingType::max) {
+    max_idx.reserve(in_dim.batch() * output_dim[0].getFeatureLen());
```
kparichay: unless you really want this to be here, please don't allocate memory in setBatch. Rather move it to the first forward/backward.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -140,8 +173,8 @@ void Pooling2DLayer::calcDerivative() {
     for (unsigned int b = 0; b < batch; ++b) {
       for (unsigned int i = 0; i < channel; ++i) {
         float del = deriv.getValue(b, 0, 0, i) / (p_size);
-        for (unsigned int j = 0; j < height; ++j) {
-          for (unsigned int k = 0; k < width; ++k) {
+        for (int j = 0; j < height; ++j) {
```
kparichay: is this intended?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -98,12 +115,28 @@ void Pooling2DLayer::calcDerivative() {
...
-    for (unsigned int i = 0; i < deriv.getDim().getDataLen(); ++i) {
-      out[max_idx[i]] += deriv.getData()[i];
+    auto iter = max_idx.begin();
+    const float *deriv_data = deriv.getData();
+    for (unsigned int b = 0; b < batch; ++b) {
+      for (unsigned int c = 0; c < channel; ++c) {
+        for (unsigned int i = 0; i < out_map_size; ++i) {
+          /// max_idx = -1 means the max idx was at the padding, so no need to
+          /// update
+          if (*iter != -1) {
```
kparichay, line 156: as `iter == -1` is ignored, is there any reason to store `-1` values in `max_idx` than rather not storing these indices in it at all?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -244,57 +277,104 @@ void Pooling2DLayer::setProperty(const PropertyType type,
...
+      int end_w = start_w + patch_width;
+
+      float max_val = std::numeric_limits<float>::lowest();
+
+      int cur_max_idx = -1;
+      for (int h = start_h; h < end_h; ++h) {
+        if (h < 0 || in_height <= h) {
+          continue;
+        }
+        for (int w = start_w; w < end_w; ++w) {
```
kparichay, line 350: `for (int w = std::max(0, start_w); w < std::min(end_w, in_width); ++w)` and remove the check below?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -244,57 +277,104 @@ void Pooling2DLayer::setProperty(const PropertyType type,
...
+  switch (pooling_type) {
+  case PoolingType::max: {
+    pool_fn = [&, this](const float *in_data, int start_h, int start_w) {
+      int end_h = start_h + patch_height;
+      int end_w = start_w + patch_width;
+
+      float max_val = std::numeric_limits<float>::lowest();
+
+      int cur_max_idx = -1;
+      for (int h = start_h; h < end_h; ++h) {
```
kparichay, line 346: same as the comment below

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -244,57 +277,104 @@ void Pooling2DLayer::setProperty(const PropertyType type,
...
+            max_val = val;
+            if (training) {
+              cur_max_idx = cur_idx;
+            }
+          }
+        }
+      }
+
+      if (training) {
+        max_idx.push_back(cur_max_idx);
```
kparichay, line 366: reserve `max_idx` at the start of this function.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -69,25 +84,27 @@ void Pooling2DLayer::forwarding(bool training) {
     hidden_ = Tensor(hidden_dim);
   }
 
-  if (pooling_type == PoolingType::max) {
-    max_idx.resize(output_dim[0].getDataLen());
-  } else if (pooling_type == PoolingType::global_max) {
-    max_idx_global.resize(output_dim[0].getDataLen());
+  if (training) {
+    if (pooling_type == PoolingType::global_max) {
+      max_idx_global.clear();
```
kparichay, line 115: move this inside `pooling2d` where this vector will be reserved again.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -69,25 +84,27 @@ void Pooling2DLayer::forwarding(bool training) {
...
+    pooling2d(in_sub, training, result);
   }
 }
 
 void Pooling2DLayer::calcDerivative() {
   unsigned int batch = input_dim[0].batch();
   unsigned int channel = input_dim[0].channel();
-  unsigned int height = input_dim[0].height();
-  unsigned int width = input_dim[0].width();
+  int height = input_dim[0].height();
```
kparichay, line 131: intended?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -30,7 +30,8 @@ int Pooling2DLayer::initialize(Manager &manager) {
   int status = ML_ERROR_NONE;
 
   if (input_dim.size() != 1 || output_dim.size() != 1) {
-    throw std::invalid_argument("Convolution layer only takes one input");
+    ml_loge("pooling layer only takes one input");
+    return ML_ERROR_NONE;
```
kparichay, line 34: return ML_ERROR_INVALID_PARAMETER?

jijoongmoon (approved):
LGTM

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -40,25 +41,55 @@ int Pooling2DLayer::initialize(Manager &manager) {
     ml_logw("Warning: the length of previous layer dimension is one");
   }
 
+  if (pooling_type == PoolingType::global_max ||
+      pooling_type == PoolingType::global_average) {
+    if (pool_size[0] != in_dim.height() || pool_size[1] != in_dim.width()) {
+      ml_logw(
+        "[Pooling2D] global_max, global_average does not except pool size");
```
kparichay, line 48: except??

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -40,25 +41,55 @@ int Pooling2DLayer::initialize(Manager &manager) {
...
+      pool_size[0] = in_dim.height();
+      pool_size[1] = in_dim.width();
+    }
+
+    if (padding[0] != 0 || padding[1] != 0) {
+      ml_logw("[Pooling2D] global_max, global_average does not except padding");
```
kparichay, line 54: same

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -40,25 +41,55 @@ int Pooling2DLayer::initialize(Manager &manager) {
...
+  if (pooling_type == PoolingType::global_max ||
+      pooling_type == PoolingType::global_average) {
+    if (pool_size[0] != in_dim.height() || pool_size[1] != in_dim.width()) {
+      ml_logw(
+        "[Pooling2D] global_max, global_average does not except pool size");
+      pool_size[0] = in_dim.height();
+      pool_size[1] = in_dim.width();
+    }
+
+    if (padding[0] != 0 || padding[1] != 0) {
```
kparichay, line 53: Is this something that our implementation does not support or it is something that is just not possible?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -40,25 +41,55 @@ int Pooling2DLayer::initialize(Manager &manager) {
...
+      ml_logw("[Pooling2D] global_max, global_average does not except padding");
+      padding[0] = 0;
+      padding[1] = 0;
+    }
+
+    if (stride[0] != 1 || stride[1] != 1) {
```
kparichay, line 59: if a user sets stride more than 1, it is better to give an error on initializing than eating that error and forcing the model to be different than intended by the user.

kparichay: Same applies for above checks as well. But you must distinguish between initialized values and user set values.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -40,25 +41,55 @@ int Pooling2DLayer::initialize(Manager &manager) {
...
+  out_dim.width((eff_in_width - pool_size[1]) / stride[1] + 1);
 
-  if (pooling_type == PoolingType::max ||
-      pooling_type == PoolingType::average) {
-    out_dim.height(
-      (in_dim.height() - pool_size[0] + 2 * padding[0]) / stride[0] + 1);
-    out_dim.width((in_dim.width() - pool_size[1] + 2 * padding[1]) / stride[1] +
-                  1);
+  if (pooling_type == PoolingType::global_max) {
+    max_idx_global.resize(out_dim.batch() * out_dim.getFeatureLen());
```
kparichay: Please move memory allocation to forward/backward unless necessary.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -40,25 +41,55 @@ int Pooling2DLayer::initialize(Manager &manager) {
...
-    out_dim.height(1);
-    out_dim.channel(1);
-    out_dim.width(in_dim.channel());
-  }
-
-  if (pooling_type == PoolingType::max) {
-    max_idx.reserve(in_dim.batch() * output_dim[0].getFeatureLen());
-  } else if (pooling_type == PoolingType::global_max) {
-    max_idx_global.resize(out_dim.batch() * output_dim[0].getFeatureLen());
+    max_idx.reserve(in_dim.batch() * out_dim.getFeatureLen());
```
kparichay, line 92: i wonder why the behavior of max_idx and max_idx_global is different (one resizes while the other reserves)?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -139,18 +163,30 @@ void Pooling2DLayer::calcDerivative() {
...
-            for (unsigned int pi = 0; pi < p_height; ++pi) {
-              for (unsigned int pj = 0; pj < p_width; ++pj) {
-                result.setValue(b, i, j + pi, k + pj,
-                                result.getValue(b, i, j + pi, k + pj) + del);
+          for (int k = -padding[1]; k <= width_stride_end; k += stride[1]) {
+            float del = deriv.getValue(b, i, J, K) / *iter;
+            iter++;
+            int patch_height_end = j + p_height;
+            int patch_width_end = k + p_width;
+            for (int h = j; h < patch_height_end; ++h) {
```
kparichay, line 181: please merge the conditions below in the loop itself.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -139,18 +163,30 @@ void Pooling2DLayer::calcDerivative() {
...
+              if (h < 0 || height <= h) {
+                continue;
+              }
+              for (int w = k; w < patch_width_end; ++w) {
```
kparichay, line 185: please merge the conditions below in the loop itself.

kparichay: This can be done at a lot of places. Please update them all.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -337,17 +359,47 @@ Tensor Pooling2DLayer::pooling2d(Tensor &in, bool training, Tensor &output) {
...
+            continue;
+          }
+          float val = in_data[h * in_width + w];
+          cnt++;
+          total += val;
+        }
+      }
+
+      if (training) {
+        max_idx.push_back(cnt);
```
kparichay, line 431: max_idx being pushed in average pooling?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -337,17 +359,47 @@ Tensor Pooling2DLayer::pooling2d(Tensor &in, bool training, Tensor &output) {
...
+      for (int h = start_h; h < end_h; ++h) {
+        if (h < 0 || in_height <= h) {
+          continue;
+        }
+        for (int w = start_w; w < end_w; ++w) {
+          if (w < 0 || in_width <= w) {
+            continue;
+          }
+          float val = in_data[h * in_width + w];
+          cnt++;
```
kparichay, line 425: please calculate `cnt` statically than adding in every loop.

nntrainer/layers/pooling2d_layer.cpp
kparichay: looks like max_idx stores the effective average counter. Can you please explain what is effective average counter?

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -197,10 +197,21 @@ void Pooling2DLayer::calcDerivative() {
...
   } break;
   case PoolingType::global_max: {
-    for (unsigned int i = 0; i < deriv.getDim().getDataLen(); ++i) {
-      float der = deriv.getData()[i] / max_idx_global[i].size();
-      for (unsigned int m = 0; m < max_idx_global[i].size(); m++) {
-        result_data[max_idx_global[i][m]] += der;
+    auto iter = max_idx_global.begin();
+    float *deriv_data = deriv.getData();
+    for (unsigned int b = 0; b < batch; ++b) {
+      for (unsigned int c = 0; c < channel; ++c) {
```
kparichay, line 203: if some indices are not being used, it would be good to combine them. I am not so sure if the compiler will.

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -360,6 +371,38 @@ Tensor Pooling2DLayer::pooling2d(Tensor &in, bool training, Tensor &output) {
...
+      if (training) {
+        max_idx_global.emplace_back();
+        indexor = max_idx_global.end() - 1;
+      }
+
+      for (int h = start_h; h < end_h; ++h) {
+        for (int w = start_w; w < end_w; ++w) {
+          int cur_idx = h * in_width + w;
+          float val = in_data[cur_idx];
+          if (max_val < val) {
```
kparichay, line 390: how about just calculate max_val and its idx, store them in local variables, and then later add to max_idx_global at once. Current checks are heavily error-prone and tough to reason.

test/unittest/unittest_nntrainer_models.cpp
zhoonit: Added, also testing for the multistride

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Yes, this reduces calculation when for-looping

nntrainer/layers/pooling2d_layer.cpp
zhoonit: max_idx == -1 has information that for this patch, calcDerivative has to be skipped. if we don't keep this, the memory layout breaks

nntrainer/layers/pooling2d_layer.cpp
zhoonit: For calcDerivative, I guess this can stay unsigned.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: We can only need to reserve once at initialize, I wonder the reasoning behind moving this to the start of this function.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Reverted.

nntrainer/layers/pooling2d_layer.cpp
kparichay: to remove memory allocation related to training/inference out of initialize.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Found that this is need for the next commit. For this commit, it is unsigned but for the next commit it becomes `int`

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Global pooling has no notion of padding, stride, pool_size

nntrainer/layers/pooling2d_layer.cpp
zhoonit: lol. updated

nntrainer/layers/pooling2d_layer.cpp
zhoonit: It's just a matter of policy but returning an error as you mentioned.

> But you must distinguish between initialized values and user set values.

we cannot distinguish if this is defaulted value or user set value. eg) pool_size must be same as in_dim.height(), in_dim.width(). but it defaults to `0, 0`

If we were to separate classes between pooling type, this could become an hard error.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Updated

nntrainer/layers/pooling2d_layer.cpp
zhoonit: max_idx_global will move to reserve in the upcoming PR. global_max_pooling is left untouched in this version.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: .clear() is guaranteed to keep it's capacity so it does not need to reserve. let's discuss this offline.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: this is to follow the semantics with tensorflow, I found that, if padding is part of the patch, it shouldn't be part of average calculation.

eg) pooling of below
x x x
x 3 3
x 3 3 
= 12 / 4 = 3

This counter has to be remembered or recalculated when backwarding as well.

nntrainer/layers/pooling2d_layer.cpp
kparichay: As user is allowed to set these, then lets compare it with its default values, and give error if not default.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Yes, I think it's good idea

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
kparichay: Sounds reasonable. Please add some explanations in the comments too then.

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
zhoonit: I agree this is heavily error-prone. Best way I can think of is using lambda but this makes the loop 2x slower (tested when building conv2d). Could you suggest me better way? I'll ask offline

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done within new commit

nntrainer/layers/pooling2d_layer.cpp
zhoonit: Done

nntrainer/layers/pooling2d_layer.cpp
zhoonit: redirecting duplicated discussion to
https://github.com/nnstreamer/nntrainer/pull/1051#discussion_r603766020

nntrainer/layers/pooling2d_layer.cpp
zhoonit: redirecting duplicated discussion to https://github.com/nnstreamer/nntrainer/pull/1051#discussion_r603766020

nntrainer/layers/pooling2d_layer.cpp
zhoonit: redirecting duplicated discussion to https://github.com/nnstreamer/nntrainer/pull/1051#discussion_r603766020

nntrainer/layers/pooling2d_layer.cpp
zhoonit: redirecting duplicated discussion to https://github.com/nnstreamer/nntrainer/pull/1051#discussion_r603766020

nntrainer/layers/pooling2d_layer.cpp
zhoonit: This is not critical so let's have a discussion after merge.

lhs8928 (approved):

PR #1053, opened 2021-03-30
[graph] Bug fix for no loss

### LINKED ISSUE #1055: loss layer cannot be set none
setting loss layer as none sets it to some other type.

kparichay:
Added bug fix for handling the scenario when there is no loss
specified. The issue was the difference in ordering of the loss
types in the parser and loss type enums.
This patch fixes it. Further, graph now simply returns if no loss found.

Resolves #1055

**Self evaluation:**


jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

PR #1056, opened 2021-03-30
[graph/ini] Support for default input_layers

### LINKED ISSUE #1046: Delete input_layers requirement
As application behavior for tizen 6.0M2 has no requirement for the input_layers, we should also support this before tizen 6.5m1 release.

So, if input_layers are not specified, we suppose that input_layers the previously defined layer in ini (same for `add_layer`)

kparichay:
This patch adds support for default input_layers
If no input_layers is specified, the layer above
the current layer in the ini file or the previously
adding layer in the model becomes the input layer for the current layer.
However, this connection is only made if the current layer
has no input layers.

Resolves #1046

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -78,6 +78,16 @@ void NetworkGraph::updateConnectionName(const std::string &from,
   }
 }
 
+void NetworkGraph::addDefaultInputLayers() {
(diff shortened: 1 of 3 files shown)


zhoonit (approved):
Nice quick patch :)

jijoongmoon (approved):
LGTM

PR #1057, opened 2021-03-30
Fix/coverity issue 210329

lhs8928:
Handle Unchecked return value issue

Added return value of remove function
resolves: 1222078, 1223055


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1058, opened 2021-03-30
fix/analysishub_issue_210330

lhs8928:
Handle BAD_OVERRIDE issue

Match type signatures with nntrainer::Optimizer by adding const
resolves: 458424, 458425

Self evaluation:

### DIFF
--- a/test/unittest/unittest_nntrainer_appcontext.cpp
+++ b/test/unittest/unittest_nntrainer_appcontext.cpp
@@ -92,39 +92,39 @@ TEST_F(nntrainerAppContextDirectory, notExisitingSetDirectory_n) {
 class CustomOptimizer : public nntrainer::Optimizer {
 public:
   /** Full custom optimizer example which overrides all functions */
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM

test/unittest/unittest_nntrainer_appcontext.cpp
```diff
@@ -94,7 +94,7 @@ class CustomOptimizer : public nntrainer::Optimizer {
   /** Full custom optimizer example which overrides all functions */
   const std::string getType() const { return "identity_optimizer"; }
 
-  float getLearningRate() { return 1.0f; }
+  float getLearningRate() const { return 1.0f; }
```
zhoonit, line 97: Could you put `override` here?
```suggestion
  float getLearningRate() const override { return 1.0f; }
```

zhoonit: and for other members as well?

zhoonit (approved):
LGTM

PR #1060, opened 2021-03-30
[ RNN ] Backwarding Implementation @open sesame 04/01 11:12

jijoongmoon:
This commit includes calculate gradient and derivatives of RNN Layer.

Resolves:

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/rnn.cpp
+++ b/nntrainer/layers/rnn.cpp
@@ -86,6 +86,7 @@ int RNNLayer::initialize(Manager &manager) {
 
 void RNNLayer::setProperty(const PropertyType type, const std::string &value) {
   int status = ML_ERROR_NONE;
+  // TODO : Add return_state property & api to get the hidden input
   switch (type) {
   case PropertyType::unit: {
     if (!value.empty()) {
@@ -123,6 +124,9 @@ void RNNLayer::forwarding(bool training) {
   Tensor hs_prev;
   Tensor hs;
 
+  if (training)
+    h_prev.setZero();
+
   // TODO : check merge b and t index
   for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
     Tensor islice = input_.getBatchSlice(b, 1);
@@ -151,7 +155,8 @@ void RNNLayer::forwarding(bool training) {
       // TODO : In-place calculation for activation
       acti_func.run_fn(hs, hs);
     }
-    h_prev.getBatchSlice(b, 1).copy(hs);
+    if (!training)
+      h_prev.getBatchSlice(b, 1).copy(hs);
   }
 }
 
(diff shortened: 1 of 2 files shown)


nntrainer/layers/rnn.cpp
```diff
@@ -163,11 +164,68 @@ void RNNLayer::copy(std::shared_ptr<Layer> l) {
...
+    Tensor xs_t = input_.getBatchSlice(b, 1);
+    Tensor hs_t = hidden_.getBatchSlice(b, 1);
+    dh_nx.setZero();
+
+    Tensor dh;
+    Tensor xs;
+    Tensor hs_prev;
+    Tensor hs;
+
+    for (unsigned int t = deriv_t.height(); t-- > 0;) {
```
zhoonit, line 201:
```suggestion
    for (int t = deriv_t.height(); t-- > 0;) {
```
It seems like t might go underflow

nntrainer/layers/rnn.cpp
```diff
@@ -163,11 +164,68 @@ void RNNLayer::copy(std::shared_ptr<Layer> l) {
 }
 
 void RNNLayer::calcDerivative() {
-  //  NYI
+  Tensor &derivative_ = net_hidden[0]->getGradientRef();
+  Tensor &weight =
+    weightAt(static_cast<int>(RNNParams::weight_xh)).getVariableRef();
+  Tensor &ret_ = net_input[0]->getGradientRef();
+
+  ret_ = derivative_.dot(weight, ret_, false, true);
```
zhoonit, line 172: Minor but this removes extra copy.

```suggestion
  derivative_.dot(weight, ret_, false, true);
```

nntrainer/layers/rnn.cpp
```diff
@@ -163,11 +164,68 @@ void RNNLayer::copy(std::shared_ptr<Layer> l) {
...
+      } else {
+        hs_prev = hs_t.getSharedDataTensor(TensorDim(1, 1, 1, hs_t.width()),
+                                           (t - 1) * hs_t.width());
+      }
+
+      if (t < deriv_t.height() - 1) {
+        dh.add_i(dh_nx);
+      }
+
+      hs = acti_func.run_prime_fn(hs, dh, dh);
```
zhoonit, line 220: Same.
```suggestion
      acti_func.run_prime_fn(hs, dh, dh);
```

nntrainer/layers/rnn.cpp
jijoongmoon: I don't think so. t never be negative.

nntrainer/layers/rnn.cpp
jijoongmoon: done.

nntrainer/layers/rnn.cpp
jijoongmoon: done.

nntrainer/layers/rnn.cpp
zhoonit: Right, I mistook.

zhoonit (approved):
LGTM

nntrainer/layers/rnn.cpp
```diff
@@ -163,11 +164,68 @@ void RNNLayer::copy(std::shared_ptr<Layer> l) {
...
+
+      if (t < deriv_t.height() - 1) {
+        dh.add_i(dh_nx);
+      }
+
+      acti_func.run_prime_fn(hs, dh, dh);
+      dh.multiply_i(hs);
+      djdb_h.add_i(dh);
+
+      djdw_x.add_i(xs.dot(dh, true, false));
```
kparichay, line 224: how about set replace add_i with providing the last parameter `beta` as 1.0 and initialize the djdw_x and djdw_h as zero?

nntrainer/layers/rnn.cpp
```diff
@@ -163,11 +164,68 @@ void RNNLayer::copy(std::shared_ptr<Layer> l) {
...
+      if (t == 0) {
+        hs_prev = Tensor(TensorDim(1, 1, 1, hs_t.width()));
+        hs_prev.setZero();
+      } else {
+        hs_prev = hs_t.getSharedDataTensor(TensorDim(1, 1, 1, hs_t.width()),
+                                           (t - 1) * hs_t.width());
+      }
+
+      if (t < deriv_t.height() - 1) {
+        dh.add_i(dh_nx);
```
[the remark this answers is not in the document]
jijoongmoon: yes. It is.

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -2351,24 +2351,39 @@ TEST_F(nntrainer_RNNLayer, initialize_01_p) {
...
+  allocateMemory();
   EXPECT_NO_THROW(layer.forwarding_with_val({input}, {}, false));
+
+  nntrainer::Tensor derivatives(2, 1, 3, 3);
+  derivatives.setValue(1.0);
+
+  setOptimizer(nntrainer::OptType::SGD, "learning_rate=1.0");
+
+  nntrainer::Tensor result;
+  EXPECT_NO_THROW(result = *layer.backwarding_with_val(
```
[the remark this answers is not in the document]
jijoongmoon: I will do later pr.

nntrainer/layers/rnn.cpp
jijoongmoon: right. I forget to set the gradients zero. I will add. 
>> add_i with providing the last parameter beta as 1.0
so.. you mean A= beta*A + B and if beta is 0.0 the A is zero??

kparichay: > so.. you mean A= beta*A + B and if beta is 0.0 the A is zero??

right. If beta is 0.0, then the original value of A will not affect the result.

test/unittest/unittest_nntrainer_layers.cpp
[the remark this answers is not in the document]
kparichay: ok.

nntrainer/layers/rnn.cpp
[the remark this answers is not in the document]
kparichay: There are two scenarios:
1. Hidden state is reset to zero every iteration - in this case, hidden state can be update in backwarding as it does not affect the next iteration.
2. Hidden state is NOT reset to zero every iteration - in this case, hidden state must retain its value for the next iteration.

In the forwarding, as per my understanding, it looks like scenario 2. Please let know, if I missed some part.

nntrainer/layers/rnn.cpp
jijoongmoon: It's good. I will update.

nntrainer/layers/rnn.cpp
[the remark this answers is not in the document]
jijoongmoon: Thanks. We made offiline discussion.

kparichay (approved):
LGTM

PR #1065, opened 2021-04-01
[ LSTM ] Layer forwarding implementation @open sesame 04/02 19:38

jijoongmoon:
This commit includes:
 . implementation of lstm forwarding implementation

Resolves:

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -49,6 +49,7 @@ enum LayerType {
   LAYER_BACKBONE_TFLITE,                      /** Backbone using TFLite */
   LAYER_EMBEDDING,                            /** Embedding Layer type */
   LAYER_RNN,                                  /** RNN Layer type */
+  LAYER_LSTM,                                 /** LSTM Layer type */
   LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN /** Unknown */
 };
 
@@ -106,6 +107,7 @@ public:
    * - out_dim : int ( output dimesion for embedding layer )
    * - in_length : int ( input length for embedding layer )
    * - recurrent_activation : string (type) - used only in lstm
+   * - dist_layer : string (type) - layer name to be distributed
    */
   /**
    * @brief     set Property of layer
@@ -265,6 +267,14 @@ Embedding(const std::vector<std::string> &properties = {}) {
 inline std::unique_ptr<Layer>
 RNN(const std::vector<std::string> &properties = {}) {
   return createLayer(LayerType::LAYER_RNN, properties);
+}
+
+/**
+ * @brief Helper function to create LSTM layer
+ */
+inline std::unique_ptr<Layer>
+LSTM(const std::vector<std::string> &properties = {}) {
+  return createLayer(LayerType::LAYER_LSTM, properties);
 }
 
 /**
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -107,7 +107,6 @@ public:
    * - out_dim : int ( output dimesion for embedding layer )
    * - in_length : int ( input length for embedding layer )
    * - recurrent_activation : string (type) - used only in lstm
-   * - dist_layer : string (type) - layer name to be distributed
    */
   /**
    * @brief     set Property of layer
--- a/nntrainer/layers/lstm.cpp
+++ b/nntrainer/layers/lstm.cpp
@@ -40,7 +40,7 @@ int LSTMLayer::initialize(Manager &manager) {
     throw std::invalid_argument("LSTM layer takes only one input");
   }
 
-  // input_dim = [ batch, 1, time_iterantion, feature_size ]
+  // input_dim = [ batch, 1, time_iteration, feature_size ]
   // outut_dim = [ batch, 1, time_iteration, hidden_size ( unit ) ]
   output_dim[0] = input_dim[0];
   output_dim[0].width(unit);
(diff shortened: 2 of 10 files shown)


api/ccapi/include/layer.h
```diff
@@ -106,6 +107,7 @@ class Layer {
    * - out_dim : int ( output dimesion for embedding layer )
    * - in_length : int ( input length for embedding layer )
    * - recurrent_activation : string (type) - used only in lstm
+   * - dist_layer : string (type) - layer name to be distributed
```
kparichay, line 110: intended?

nntrainer/layers/lstm.cpp
```diff
@@ -80,6 +81,11 @@ int LSTMLayer::initialize(Manager &manager) {
                                       WeightRegularizer::NONE, 1.0f, true);
   }
 
+  mem_cell =
+    std::make_shared<Var_Grad>(output_dim[0], true, true, "LSTM:mem_cell");
```
kparichay, line 85: Please try and use the `mem_cell` as a weight so that is memory is managed by the manager than managing it manually in the layer.

nntrainer/layers/lstm.cpp
```diff
@@ -74,7 +138,72 @@ void LSTMLayer::setRecurrentActivation(ActivationType activation) {
...
+  Tensor o;
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
+    Tensor islice = input_.getBatchSlice(b, 1);
+    Tensor oslice = hidden_.getBatchSlice(b, 1);
+    Tensor cell = m_cell_.getBatchSlice(b, 1);
+
+    for (unsigned int t = 0; t < islice.height(); ++t) {
+      Tensor xs = input_.getSharedDataTensor(TensorDim(1, 1, 1, islice.width()),
+                                             t * islice.width());
```
zhoonit, line 169: Very minor code clean idea :)
```suggestion
      Tensor xs = input_.getSharedDataTensor({ islice.width() }), t * islice.width());
```

nntrainer/layers/lstm.cpp
```diff
@@ -74,7 +138,72 @@ void LSTMLayer::setRecurrentActivation(ActivationType activation) {
...
+
+      f = temp.getSharedDataTensor(TensorDim(1, 1, 1, unit), 0);
+      g = temp.getSharedDataTensor(TensorDim(1, 1, 1, unit), temp.width());
+      i = temp.getSharedDataTensor(TensorDim(1, 1, 1, unit), 2 * temp.width());
+      o = temp.getSharedDataTensor(TensorDim(1, 1, 1, unit), 3 * temp.width());
+
+      recurrent_acti_func.run_fn(f, f);
+      recurrent_acti_func.run_fn(i, i);
+      recurrent_acti_func.run_fn(o, o);
+      acti_func.run_fn(g, g);
```
zhoonit, line 196: if an activation function is not element-wise, this would break. But I wonder if there is any case for this.

nntrainer/layers/lstm.cpp
```diff
@@ -26,12 +26,76 @@ const std::string LSTMLayer::type = "lstm";
...
+//  : [1, 1, unit (hidden_size) , unit (hidden_size) x NUM_GATE] -> f, g, i, o
+// - bias_h ( hidden bias )
+//  : [1, 1, 1, unit (hidden_size) x NUM_GATE] -> f, g, i, o
 int LSTMLayer::initialize(Manager &manager) {
   int status = ML_ERROR_NONE;
   if (getNumInputs() != 1) {
     throw std::invalid_argument("LSTM layer takes only one input");
   }
 
+  // input_dim = [ batch, 1, time_iterantion, feature_size ]
```
zhoonit, line 43:
```suggestion
  // input_dim = [ batch, 1, time_iteration, feature_size ]
```

nntrainer/layers/lstm.cpp
```diff
@@ -74,7 +138,72 @@ void LSTMLayer::setRecurrentActivation(ActivationType activation) {
...
+      recurrent_acti_func.run_fn(o, o);
+      acti_func.run_fn(g, g);
+
+      f.multiply(cs_prev, cs);
+      cs.add_i(g.multiply_i(i));
+
+      acti_func.run_fn(cs, hs);
+      hs.multiply_i(o);
+    }
+    h_prev.getBatchSlice(b, 1).copy(hs);
```
zhoonit, line 204: I think it's guaranteed that dimension is same for this case...

If they are different, copy will allocate a new tensor thus breaking memory layout.
I think later, we should have a different memory ready for this purpose.

zhoonit (approved):
LGTM with some minor codes :)

nntrainer/layers/lstm.cpp
kparichay: softmax is not element-wise, but I doubt someone would use it here though. Add it as a note though or restrict to be only ReLU, TanH, Sigmoid.

nntrainer/layers/lstm.cpp
kparichay: `tensor::copy(tensor)` performs the check to ensure that the size of src and destination tensor are the same.

zhoonit: Yes, I mean the case we intend to dump the data without allocation but if we can't it, it would be safer to throw.

nntrainer/layers/lstm.cpp
jijoongmoon: yes. there will be the limitation about the activation in later PR. only element wise activation is working.

nntrainer/layers/lstm.cpp
jijoongmoon: this is not the memory which managed by the manager. It is layer variables. the manager managed variable is h, weight_hh, weight_xh and bias which includes all variables, f, g, i and o.

nntrainer/layers/lstm.cpp
jijoongmoon: the dimension is same. I will add checking in later PR.

nntrainer/layers/lstm.cpp
jijoongmoon: I think it should be removed. Thanks.

api/ccapi/include/layer.h
jijoongmoon: I will remove. it accidentally there. my bad.

nntrainer/layers/lstm.cpp
kparichay: I guess there should be an interface with the manager to request for memory, as required in this case. Added #1086 for this.

kparichay (approved):
LGTM

Issue #1063, opened 2021-04-01
devel package and headers need some clean up


zhoonit:
1. devel package is not working due to missing header (or unnecessary inclusion at the header file)
2. capi-devel and nntrainer-devel is mixed.

kparichay:
This will be done after or with #986 .

zhoonit:
Cool. I worked a bit with @lhs8928 just to make it work for now. please refer to the PR that hs is about to make

zhoonit:
- loss type definition is included in `networkGraph` but type is defined at `loss_layer.h`

jijoongmoon:
We need to test writing the custom layer and custom optimizer with devel package.

kparichay:
Yes, that will be done as soon as layer refactoring is done (it is delayed in my schedule though)

zhoonit:
- remove parse_util.h from header that would be included.

PR #1066, opened 2021-04-01
[Meson] Adjust header files @open sesame 04/27 16:42

lhs8928:
1. Include some header files to nntrainer_headers in meson to provide to users when install it.
2. Adust header files to handle following issues caused by 1.

Self evaluation:

### DIFF
--- a/nntrainer/meson.build
+++ b/nntrainer/meson.build
@@ -4,7 +4,7 @@ nntrainer_inc = [
 ]
 
 nntrainer_sources = []
-nntrainer_headers = ['app_context.h']
+nntrainer_headers = ['app_context.h', 'nntrainer_error.h', 'nntrainer_log.h', 'nntrainer_logger.h']
 
 # Dependencies
 nntrainer_base_deps=[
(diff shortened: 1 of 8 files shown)


zhoonit:
Could you please update & check rpm / debian to include those header files as well?

zhoonit:
See also #1063

nntrainer/meson.build
```diff
@@ -4,7 +4,7 @@ nntrainer_inc = [
 ]
 
 nntrainer_sources = []
-nntrainer_headers = ['app_context.h']
+nntrainer_headers = ['app_context.h', 'nntrainer_error.h', 'nntrainer_log.h', 'nntrainer_logger.h']
```
kparichay, line 7: nntrainer_headers contains the headers which will be installed with nntrainer-devel.
so, if you want to include them, please install them with tizen and ubuntu packaging as well.

btw, why do you need these headers outside nntrainer?

zhoonit: Because some of our header file includes those headers making unable to build an application without those header.
We need some cleanup. But for now this patch makes devel working.

kparichay: please update those applications to not use headers. It will be just a small work while making this patch.

PR #1068, opened 2021-04-01
[Wait for #1065] [ TIME DIST ] Add skeleton Code for Time Distribution Layer

jijoongmoon:
This includes
 . Skeleton Code for Time Distribution Layer

Resolves:

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -50,6 +50,7 @@ enum LayerType {
   LAYER_EMBEDDING,                            /** Embedding Layer type */
   LAYER_RNN,                                  /** RNN Layer type */
   LAYER_LSTM,                                 /** LSTM Layer type */
+  LAYER_TIME_DIST,                            /** Time Distributed Layer type */
   LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN /** Unknown */
 };
 
@@ -107,6 +108,7 @@ public:
    * - out_dim : int ( output dimesion for embedding layer )
    * - in_length : int ( input length for embedding layer )
    * - recurrent_activation : string (type) - used only in lstm
+   * - distribute : bool
    */
   /**
    * @brief     set Property of layer
@@ -274,6 +276,14 @@ RNN(const std::vector<std::string> &properties = {}) {
 inline std::unique_ptr<Layer>
 LSTM(const std::vector<std::string> &properties = {}) {
   return createLayer(LayerType::LAYER_LSTM, properties);
+}
+
+/**
+ * @brief Helper function to create Time Distributed layer
+ */
+inline std::unique_ptr<Layer>
+TimeDistLayer(const std::vector<std::string> &properties = {}) {
+  return createLayer(LayerType::LAYER_TIME_DIST, properties);
 }
 
 /**
--- a/nntrainer/layers/time_dist.h
+++ b/nntrainer/layers/time_dist.h
@@ -29,10 +29,7 @@ public:
   /**
    * @brief     Constructor of Time Distribution Layer
    */
-  template <typename... Args>
-  TimeDistLayer(std::shared_ptr<Layer> iter_, Args... args) :
-    Layer(args...),
-    iter(iter_) {}
+  template <typename... Args> TimeDistLayer(Args... args) : Layer(args...) {}
 
   /**
    * @brief     Destructor of Time Distributed Layer
@@ -84,23 +81,12 @@ public:
    */
   const std::string getType() const override { return TimeDistLayer::type; };
 
-  using Layer::setProperty;
-
-  /**
-   * @copydoc Layer::setProperty(const PropertyType type, const std::string
-   * &value)
-   */
-  void setProperty(const PropertyType type,
-                   const std::string &value = "") override;
-
   static const std::string type;
 
 private:
   /* @brief Layer to be distributed through time */
-  std::shared_ptr<Layer> iter;
 
-  /* @brief Name of Layer be distributed through time */
-  std::string iter_name;
+  std::shared_ptr<Layer> dist_layer;
 };
 } // namespace nntrainer
 
--- a/nntrainer/layers/time_dist.h
+++ b/nntrainer/layers/time_dist.h
@@ -16,7 +16,6 @@
 #ifdef __cplusplus
 
 #include <tensor.h>
-#include <time_dist.h>
 
 namespace nntrainer {
 
@@ -34,7 +33,7 @@ public:
   /**
    * @brief     Destructor of Time Distributed Layer
    */
-  ~TimeDistLayer(){};
+  ~TimeDistLayer() = default;
 
   /**
    *  @brief  Move constructor.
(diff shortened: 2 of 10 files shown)


api/ccapi/include/layer.h
```diff
@@ -107,6 +107,9 @@ class Layer {
    *            31. in_dim : int ( input dimension for embedding layer )
    *            32. out_dim : int ( output dimesion for embedding layer )
    *            33. in_length : int ( input length for embedding layer )
+   *            34. recurrent_activation :  string (type) - lstm
+   *            35. dist_layer : string (type) - layer name to be distributed
```
zhoonit, line 111:
```suggestion
   *            35. dist_layer : string (type) - layer type name to be distributed
```

zhoonit (approved):
LGTM :)

api/ccapi/include/layer.h
jijoongmoon: No.. I think it must be layer name. not type name. This could be defined by user.

kparichay: You might need to think a little more about this. Taking `layer name` for which time distribution is to be done would require the graph to add a layer, then give that layer to `time distributed` layer and replace it inside the graph.
From the API perspective, adding `time distributed` layer to the graph removes the original layer from the graph which is also a bit weird. 

As a workaround, you might want to consider a design, where time_dist is not a layer but rather a wrapper for the layer. When time distributed is set (like flatten), then it adds an additional loop for each kind of layer. This might come out as a cleaner way and have less obstruction on the graph and less confusion for the user.

kparichay:
You might need to think a little more about this. Taking layer name for which time distribution is to be done would require the graph to add the original layer, then give that layer to time distributed layer and replace it inside the graph.
From the API perspective, adding time distributed layer to the graph removes the original layer from the graph which is also a bit weird. This design in a way creates multiple instances of a layer with shared configuration and parameters.

As a workaround, you might want to consider a design, where time_dist is not a layer but rather a feature for each layer. When time distributed is set (like flatten), then it adds an additional loop for each kind of layer. This might come out as a cleaner way and have less obstruction on the graph and less confusion for the user. This design allows each layer to directly work on higher dimensional inputs. Another advantage of this design is that majority of layers already support this (like activation layers, in/out layers, etc).

api/ccapi/include/layer.h
zhoonit: I see, understood :)

zhoonit: I wonder how it would be used in the ini. If input_layers is not specified, it is automatically connects to a previous layer.

```ini
[fclayer]
type=fc

[convlayer]
type=conv
// this will implicitly have fclayer as an input in the graph as well.

[dist_layer]
dist_layer=fclayer
input_layers=fc
```

jijoongmoon: yes. it is automatically connects the previous layer and dist_layer inside.

jijoongmoon:
> You might need to think a little more about this. Taking layer name for which time distribution is to be done would require the graph to add the original layer, then give that layer to time distributed layer and replace it inside the graph.

It is more like handle sub-graph inside time distributed layer. Input output of dist_layer are feeaded by distributed layer.. and the connection to outside.. totally done by time distributed layer. 

> From the API perspective, adding time distributed layer to the graph removes the original layer from the graph which is also a bit weird. This design in a way creates multiple instances of a layer with shared configuration and parameters.

Not remove original layer. There is no connection with dist_layer to out-side. It's only working inside of time distributed layer.

> As a workaround, you might want to consider a design, where time_dist is not a layer but rather a feature for each layer. When time distributed is set (like flatten), then it adds an additional loop for each kind of layer. This might come out as a cleaner way and have less obstruction on the graph and less confusion for the user. This design allows each layer to directly work on higher dimensional inputs. Another advantage of this design is that majority of layers already support this (like activation layers, in/out layers, etc).

Well. if we try to implement time distribution for every layer, then.. we change all the implementation of layer. input_dim and output dim also have to changed.. and I guess.. more. It could be messy. Let me think about it. and if we have to consider about custom layer .. then it make more difficult. we have to ask a developer to implement.

api/ccapi/include/layer.h
```diff
@@ -50,6 +50,7 @@ enum LayerType {
   LAYER_EMBEDDING,                            /** Embedding Layer type */
   LAYER_RNN,                                  /** RNN Layer type */
   LAYER_LSTM,                                 /** LSTM Layer type */
+  LAYER_TIME_DIST,                            /** Time Distributed Layer type */
```
kparichay, line 53: if this layer can be constructed and add to the model manually by the user, just keep in mind that this will need special handling in the graph, which can be done in a later PR.

kparichay:
> Well. if we try to implement time distribution for every layer, then.. we change all the implementation of layer. input_dim and output dim also have to changed.. and I guess.. more. It could be messy. Let me think about it. and if we have to consider about custom layer .. then it make more difficult. we have to ask a developer to implement.

We discussed offline to provide `distribution` as a property for the layer as the default interface.

api/ccapi/include/layer.h
jijoongmoon: This will come in next pr. #1074

nntrainer/layers/time_dist.h
```diff
@@ -0,0 +1,93 @@
...
+ * @bug		No known bugs except for NYI items
+ *
+ */
+
+#ifndef __TIME_DIST_H__
+#define __TIME_DIST_H__
+#ifdef __cplusplus
+
+#include <tensor.h>
+#include <time_dist.h>
```
kparichay, line 19: intended?

nntrainer/layers/time_dist.h
```diff
@@ -0,0 +1,93 @@
...
+  static const std::string type;
+
+private:
+  /* @brief Layer to be distributed through time */
+  std::shared_ptr<Layer> dist_layer;
+};
+} // namespace nntrainer
+
+#endif /* __cplusplus */
+#endif /* __FC_LAYER_H__ */
```
kparichay, line 93: FC_LAYER?

nntrainer/layers/time_dist.h
```diff
@@ -0,0 +1,93 @@
...
+public:
+  /**
+   * @brief     Constructor of Time Distribution Layer
+   */
+  template <typename... Args> TimeDistLayer(Args... args) : Layer(args...) {}
+
+  /**
+   * @brief     Destructor of Time Distributed Layer
+   */
+  ~TimeDistLayer(){};
```
kparichay, line 37: use `default` instead of `{}`

kparichay (approved):
LGTM other than the new minor comments

nntrainer/layers/time_dist.h
jijoongmoon: fixed

nntrainer/layers/time_dist.h
jijoongmoon: fixed

nntrainer/layers/time_dist.h
jijoongmoon: fixed

PR #1071, opened 2021-04-05
[API/Graph] Update graph node and layer objects

kparichay:
**Commit 1** :  [Graph] Add interface for graph node
Add interface for the graph node.
Graph will use this interface.

**Commit 2:** [graph] Add node the graph of layers
Create class for the graph of layers which will represent nodes
This class also represents the object which will be created with APIs from now.
The actual class object resides inside this layer node object.

Properties for this layer node will be intercepted by the layer node object,
and properties of the graph node will be set in the layer node.
Others will be passed down to the layer node.

See Aso #986

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/graph_node.h
+++ b/nntrainer/graph/graph_node.h
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	graph_node.h
+ * @date	1 April 2021
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is the graph node interface for c++ API
+ */
+
+#ifndef __GRAPH_NODE_H__
+#define __GRAPH_NODE_H__
+
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace nntrainer {
+
+/**
+ * @class   Layer Base class for the graph node
+ * @brief   Base class for all layers
+ */
+class GraphNode {
+public:
+  /**
+   * @brief     Destructor of Layer Class
+   */
+  virtual ~GraphNode() = default;
+
+  /**
+   * @brief     Get index of the node
+   *
+   */
+  virtual size_t getIndex() = 0;
+
+  /**
+   * @brief     Set index of the node
+   *
+   */
+  // virtual void setIndex(size_t) = 0;
+
+  /**
+   * @brief     Get the Name of the underlying object
+   *
+   * @return std::string Name of the underlying object
+   * @note name of each node in the graph must be unique
+   */
+  virtual std::string getName() noexcept = 0;
+
+  /**
+   * @brief     Set the Name of the underlying object
+   *
+   * @param[in] std::string Name for the underlying object
+   * @note name of each node in the graph must be unique
+   *
+   * @todo make it work with setProperty
+   */
+  // virtual int setName(std::string name) = 0;
+
+  /**
+   * @brief     Get the trainable property of the underlying object
+   *
+   * @return boolean true if trainable, else false
+   */
+  virtual bool getTrainable() noexcept = 0;
+
+  /**
+   * @brief     Set the properties for the node
+   *
+   * @param[in] properties properties of the node
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   *
+   * @note this shouldn't be virtual, this became virtual to support custom
+   * layer. should be reverted after layer.h can fully support custom layer
+   */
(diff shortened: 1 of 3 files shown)


nntrainer/graph/graph_node.h
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * @brief     Set the Name of the underlying object
+   *
+   * @param[in] std::string Name for the underlying object
+   * @note name of each node in the graph must be unique
+   *
+   * @todo make it work with setProperty
+   */
+  // virtual int setName(std::string name) = 0;
```
zhoonit, line 61: I guess this is intended to unify interface to set property

nntrainer/graph/graph_node.h
```diff
@@ -0,0 +1,91 @@
...
+   * @brief     Set the properties for the node
+   *
+   * @param[in] properties properties of the node
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   *
+   * @note this shouldn't be virtual, this became virtual to support custom
+   * layer. should be reverted after layer.h can fully support custom layer
+   */
+  virtual int setProperty(std::vector<std::string> properties) = 0;
```
zhoonit, line 80: I feel like I asked before, but is it restricted to do `const vector &` ?

nntrainer/layers/layer_node.h
```diff
@@ -0,0 +1,122 @@
...
+#include <layer.h>
+#include <layer_internal.h>
+
+namespace nntrainer {
+
+/**
+ * @class   LayerNode class
+ * @brief   layer node class for the graph
+ */
+class LayerNode : public ml::train::Layer, public GraphNode {
```
zhoonit, line 26: I like current status, is GraphNode gonna be type T?

nntrainer/layers/layer_node.h
```diff
@@ -0,0 +1,122 @@
...
+#ifdef PROFILE
+  int event_key;
+#endif
+
+private:
+  std::shared_ptr<nntrainer::Layer>
+    layer;      /**< The actual object in the graph node */
+  size_t index; /**< index of each node */
+
+  std::vector<std::string> input_layers;  /**< input layer names */
```
zhoonit, line 114: I wonder if this can be actually pointing to the node later?

`std::shared_ptr<LayerNode>`

zhoonit (approved):
LGTM

nntrainer/graph/graph_node.h
kparichay: Yes, but I am not sure if this can be surely eliminated. It will be removed if future PRs work without it.

nntrainer/graph/graph_node.h
kparichay: current implementation of `setProperty` modifies the properties itself. This will have to updated #1098

nntrainer/layers/layer_node.h
kparichay: It was `type T` but not, it is not needed anymore. It was redundant and has been removed.

nntrainer/layers/layer_node.h
kparichay: This is just taken from `Layer` class to mark what all functionalities will be moving to layer node. This can be changed to replaced or appended with `std::shared_ptr<LayerNode>`.

jijoongmoon:
Could you specify which has which ?? GraphNode, LayerNode and Layer?

nntrainer/layers/layer_node.h
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file	graph_node.h
+ * @date	1 April 2021
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug		No known bugs except for NYI items
+ * @brief	This is the graph node interface for c++ API
```
dongju-chae, line 10: Plz check the overall doxygen comments.

nntrainer/graph/graph_node.h
```diff
@@ -0,0 +1,91 @@
...
+   */
+  // virtual void setIndex(size_t) = 0;
+
+  /**
+   * @brief     Get the Name of the underlying object
+   *
+   * @return std::string Name of the underlying object
+   * @note name of each node in the graph must be unique
+   */
+  virtual std::string getName() noexcept = 0;
```
jijoongmoon, line 51: so.. I guess is it the same with Layer Name?

jijoongmoon (approved):
LGTM

nntrainer/graph/graph_node.h
kparichay: yes.

nntrainer/layers/layer_node.h
kparichay: i will update this in the next PR.

PR #1074, opened 2021-04-05
[ TIME DIST ] Modify graph to recognize time distribute layer @open sesame 04/26 21:11

jijoongmoon:
if time distributed layer has activation, we have to add time
distributed layer for activation, and if it need to realize loss
layer, we also add the time distribute layer for loss.

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -29,6 +29,7 @@
 #include <parse_util.h>
 #include <profiler.h>
 #include <rnn.h>
+#include <time_dist.h>
 
 namespace nntrainer {
 
@@ -38,6 +39,13 @@ namespace nntrainer {
  */
 static const std::vector<std::string> in_place_layers = {
   ActivationLayer::type, BatchNormalizationLayer::type};
+
+static std::shared_ptr<Layer> distributeLayer(std::shared_ptr<Layer> l) {
+  std::shared_ptr<Layer> layer = nntrainer::createLayer(TimeDistLayer::type);
+  std::dynamic_pointer_cast<TimeDistLayer>(layer)->setDistLayer(l);
+
+  return layer;
+}
 
 int NetworkGraph::compile(const LossType loss_type) {
   int status = ML_ERROR_NONE;
@@ -196,8 +204,14 @@ void NetworkGraph::ensureName(std::shared_ptr<Layer> layer,
 
   std::set<std::string>::iterator iter;
   std::string name;
-  if (orig_name_empty)
-    orig_name = layer->getType();
+  if (orig_name_empty) {
+    if (layer->getType() == TimeDistLayer::type) {
+      orig_name =
+        std::dynamic_pointer_cast<TimeDistLayer>(layer)->getDistLayerType();
+    } else {
+      orig_name = layer->getType();
+    }
+  }
   std::string direct_name = prefix + orig_name;
 
   do {
@@ -299,11 +313,15 @@ int NetworkGraph::realizeActivationType(Layer &current) {
     return ML_ERROR_INVALID_PARAMETER;
   }
 
-  std::shared_ptr<Layer> layer = nntrainer::createLayer("activation");
+  std::shared_ptr<Layer> layer = nntrainer::createLayer(ActivationLayer::type);
+  layer->setActivation(act);
+
+  if (current.getType() == TimeDistLayer::type) {
+    layer = distributeLayer(layer);
+  }
 
   ensureName(layer, current.getName());
 
-  layer->setActivation(act);
   layer->setNumInputs(current.getNumInputs());
   layer->input_layers.clear();
   layer->input_layers.push_back(current.getName());
@@ -348,6 +366,12 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
   if (Sorted.back().layer->getType() == LossLayer::type)
     return status;
 
+  if (Sorted.back().layer->getType() == TimeDistLayer::type) {
+    if (std::static_pointer_cast<TimeDistLayer>(Sorted.back().layer)
+          ->getDistLayerType() == LossLayer::type)
+      return status;
+  }
+
   if (loss_type == LossType::LOSS_NONE) {
     return ML_ERROR_NONE;
   }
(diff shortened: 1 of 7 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -360,10 +384,21 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
 
   LayerNode last_node = Sorted.back();
   if (updated_loss_type == LossType::LOSS_ENTROPY) {
-    if (last_node.layer->getType() != "activation") {
-      ml_loge("Error: Cross Entropy need last layer to have softmax or sigmoid "
-              "activation.");
-      return ML_ERROR_NOT_SUPPORTED;
+
+    if (last_node.layer->getType() != ActivationLayer::type) {
+      if (last_node.layer->getType() == TimeDistLayer::type) {
```
dongju-chae, line 389: How about this?
```
auto type = last_node.layer->getType();
if (type != ActivationLayer::type && type == TimeDistLayer::type) {
  ...
} else {
  ...
}
```

nntrainer/layers/time_dist.cpp
```diff
@@ -45,6 +49,11 @@ void TimeDistLayer::copy(std::shared_ptr<Layer> l) {
   this->dist_layer = from->dist_layer;
 }
 
+void TimeDistLayer::setDistLayer(std::shared_ptr<Layer> l) {
+  dist_layer = l;
+  Layer::setActivation(l->getActivationType());
```
kparichay, line 54: setFlatten as well?

test/unittest/unittest_nntrainer_modelfile.cpp
```diff
@@ -849,6 +849,22 @@ TEST(nntrainerIniTest, backbone_based_on_working_directory_p) {
...
+  ScopedIni s{
+    "distribute_p1",
+    {nw_base, adam,
+     input + "-Activation" + "-Input_Shape" + "Input_Shape = 3:1:10:10",
+     out + "distribute=true"}};
+  nntrainer::NeuralNetwork NN;
+
+  EXPECT_EQ(NN.loadFromConfig(s.getIniName()), ML_ERROR_NONE);
+  EXPECT_EQ(NN.compile(), ML_ERROR_NONE);
+  // EXPECT_EQ(NN.initialize(), ML_ERROR_NONE);
```
kparichay, line 865: is this intended?

nntrainer/models/model_loader.cpp
```diff
@@ -306,18 +306,20 @@ int ModelLoader::loadLayerConfigIniCommon(dictionary *ini,
...
-      drived->setDistLayer(std::static_pointer_cast<Layer>(layer_));
+    std::shared_ptr<Layer> l = std::static_pointer_cast<Layer>(layer_);
+
+    if (l->getDistribute()) {
+      ml_logd("This %s layer is going to distributed", layer_name.c_str());
+
+      std::shared_ptr<Layer> dist_layer =
+        nntrainer::createLayer(nntrainer::TimeDistLayer::type);
+
+      std::dynamic_pointer_cast<TimeDistLayer>(dist_layer)->setDistLayer(l);
```
zhoonit, line 317: dynamic_cast breaks our api.

I think, later...., 1. we should put this into our api, 2. implement this with exisiting set of api, 3. introduce new object.

nntrainer/graph/network_graph.cpp
```diff
@@ -386,7 +421,16 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
   last_node = Sorted.back();
   std::string input_str = last_node.layer->getName();
 
-  std::shared_ptr<LossLayer> layer = std::make_unique<LossLayer>();
+  std::shared_ptr<Layer> layer = nntrainer::createLayer(LossLayer::type);
+
+  status =
+    std::dynamic_pointer_cast<LossLayer>(layer)->setLoss(updated_loss_type);
```
zhoonit, line 427: Later, I think we should merge this to setProperty as well. Could you put todo here?

zhoonit (approved):
LGTM

test/unittest/unittest_nntrainer_modelfile.cpp
jijoongmoon: This will enable at next PR.

nntrainer/layers/time_dist.cpp
jijoongmoon: Yes. we support flatten like activation for some cases. I will update next pr.

nntrainer/models/model_loader.cpp
jijoongmoon: > dynamic_cast breaks our api.

Could you give more specific?

nntrainer/graph/network_graph.cpp
jijoongmoon: Why do you think we need to merge this into setProperty?? It is Propery from internal loss layer. Not dist layer.

nntrainer/models/model_loader.cpp
zhoonit: API does not have the method `setDistLayer` while we are using it as a part of layer.
We shouldn't be relying on the concrete type, but only the interfaces. If it's not possible, it suggests that there'd be some problem with our interface.

nntrainer/graph/network_graph.cpp
zhoonit: loss layer has to be sooner or later open as an interface, so it would be better to use the existing interface if it's possible.

nntrainer/models/model_loader.cpp
jijoongmoon: Well we are not going to open this dist layer to the public. All the users have to do is add just dist_layer=true/false.

nntrainer/graph/network_graph.cpp
jijoongmoon: yes. loss layer might be open. but time dist layer is not. and I'm not sure if we have to setLoss should be opened.

nntrainer/models/model_loader.cpp
zhoonit: If we are to define this is not part of an api, I think it's better to directly construct distlayer instead of creating `<Layer>` and dynamic_cast, also remove from the `appcontext`.

nntrainer/graph/network_graph.cpp
zhoonit: I think this can be done as sth like `layer->setProperty("loss=loss_cross_entropy")` later :)

nntrainer/graph/network_graph.cpp
jijoongmoon: done

kparichay (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -360,8 +384,16 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
...
+    auto type = last_node.layer->getType();
+    if (type != ActivationLayer::type && type == TimeDistLayer::type) {
+      if (std::dynamic_pointer_cast<TimeDistLayer>(last_node.layer)
+            ->getDistLayerType() != ActivationLayer::type) {
+        ml_loge("Error: Cross Entropy need last layer to have softmax or "
+                "sigmoid activation.");
+        return ML_ERROR_NOT_SUPPORTED;
+      }
+    } else {
+      ml_loge("Error: Cross Entropy need last layer to have softmax or sigmoid"
```
zhoonit, line 396: Found issue. This should be somewhat like below.

```cpp
auto type = last_node.layer->getType();
if (type == "time_dist") {
 type = layer->getDistLayerType();
}

if (type != ActivationLayer::type) {
     /** handle error */
}
```

PR #1078, opened 2021-04-06
[Ini] Implement ini deserializer

zhoonit:
1. [Ini] Implement ini deserializer 

```This patch implements ini deserializer as well as changing the interface
signature to get filepath.

Many NYI parts will be implemented soon.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/ini_interpreter.cpp
+++ b/nntrainer/compiler/ini_interpreter.cpp
@@ -1,0 +1,148 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   ini_interpreter.cpp
+ * @date   02 April 2021
+ * @brief  NNTrainer Ini Interpreter (partly moved from model_loader.c)
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+#include <ini_interpreter.h>
+
+#include <vector>
+
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <util_func.h>
+
+static constexpr const char *FUNC_TAG = "[IniInterpreter] ";
+
+static constexpr const char *UNKNOWN_STR = "UNKNOWN";
+static constexpr const char *NONE_STR = "NONE";
+static constexpr const char *MODEL_STR = "model";
+static constexpr const char *DATASET_STR = "dataset";
+static constexpr const char *OPTIMIZER_STR = "optimizer";
+
+namespace nntrainer {
+
+namespace {
+
+/** @todo:
+ *  1. deprecate tag dispatching along with #1072
+ *  2. deprecate getMergeableGraph (extendGraph should accept graph itself)
+ */
+class PlainLayer {};    /**< Plain Layer tag */
+class BackboneLayer {}; /**< Backbone Layer tag */
+
+template <typename T>
+static std::shared_ptr<Layer> section2layer(dictionary *ini,
+                                            const std::string &sec_name) {
+  /// NYI!
+  return nullptr;
+}
+
+static std::shared_ptr<GraphRepresentation>
+section2graph(dictionary *ini, const std::string &sec_name) {
+  /// NYI!
+  return nullptr;
+}
+
+bool graphSupported(const std::string &backbone_name) {
+  /// NYI!
+  return true;
+}
+
+std::vector<std::shared_ptr<Layer>>
+getMergeableGraph(std::shared_ptr<const GraphRepresentation>) {
+  return {};
+};
+
+} // namespace
+
+void IniGraphInterpreter::serialize(
+  std::shared_ptr<const GraphRepresentation> representation,
+  const std::string &out) {}
+
+std::shared_ptr<GraphRepresentation>
+IniGraphInterpreter::deserialize(const std::string &in) {
+
+  NNTR_THROW_IF(in.empty(), std::invalid_argument)
+    << FUNC_TAG << "given in file is empty";
+
+  NNTR_THROW_IF(!isFileExist(in), std::invalid_argument)
+    << FUNC_TAG << "given ini file does not exist";
+
+  dictionary *ini = iniparser_load(in.c_str());
+  NNTR_THROW_IF(!ini, std::runtime_error) << "loading ini failed";
+
+  auto freedict = [ini] { iniparser_freedict(ini); };
+
+  /** Get number of sections in the file */
+  int num_ini_sec = iniparser_getnsec(ini);
+  NNTR_THROW_IF_CLEANUP(num_ini_sec < 0, std::invalid_argument, freedict)
+    << FUNC_TAG << "invalid number of sections.";
+
+  std::shared_ptr<GraphRepresentation> graph =
+    std::make_shared<GraphRepresentation>();
+
+  try {
+    for (int idx = 0; idx < num_ini_sec; ++idx) {
+      auto sec_name_ = iniparser_getsecname(ini, idx);
+      NNTR_THROW_IF_CLEANUP(!sec_name_, std::runtime_error, freedict)
+        << "parsing a section name returned error, filename: " << in
+        << "idx: " << idx;
+
+      ml_logd("probing section_name: %s", sec_name_);
+      std::string sec_name(sec_name_);
+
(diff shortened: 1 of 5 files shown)


nntrainer/compiler/compiler.h
```diff
@@ -7,7 +7,21 @@
...
+ *
+ * GraphRepresentaion g;
+ * g.setInterpreter(tfliteInterpreter);
+ * g.setCompiler(nntrainerCPUCompiler);
+ * auto tflite_file = std::open('a.tflite');
+ * tflite_file >> g;
+ *
+ * ExecutableGraph eg = g.compile();
+ *
+ * g.setInterpreter(iniInterpreter);
```
jijoongmoon: So.. we does not consider tflite backbone case? ( This is tflite + nntrainer case )

nntrainer/compiler/compiler.h
```diff
@@ -97,12 +113,124 @@ class GraphInterpreter {
...
+class GraphRepresentation {
+public:
+  /**
+   * @brief Construct a new Graph Representation object
+   *
+   * @param interpreter_ interpreter to use
+   * @param compiler_ compiler to use
+   */
+  GraphRepresentation(std::shared_ptr<GraphInterpreter> interpreter_ = nullptr,
+                      std::shared_ptr<GraphCompiler> compiler_ = nullptr) :
```
jijoongmoon: I thought GraphCompiler use graphRepresentation. This looks the opposite now.

nntrainer/compiler/compiler.h
```diff
@@ -97,12 +113,124 @@ class GraphInterpreter {
...
+   * @brief Set the Graph Compiler object
+   *
+   * @param compiler_ compiler
+   */
+  void setGraphCompiler(std::shared_ptr<GraphCompiler> compiler_) {
+    compiler = compiler_;
+  }
+
+private:
+  std::shared_ptr<GraphInterpreter> interpreter;
```
jijoongmoon: I think GraphInterpreter and Compiler has graphrepresentation. Could you share your thought?

nntrainer/compiler/compiler.h
zhoonit: Yes, this is initial design commit, only the last commit is valid for this PR. I'll rebase now :)

nntrainer/compiler/compiler.h
zhoonit: This is rebased, this used to be the initial design, I just rebased :)

nntrainer/compiler/compiler.h
zhoonit: Sorry for the messy commit from waiting PR, now this commit is rebased :)

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file ini_interpreter.cpp
+ * @date 02 April 2021
+ * @brief NNTrainer Ini Interpreter (partly moved from model_loader.c)
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author	Jijoong Moon <jijoong.moon@samsung.com>
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
```
dongju-chae, line 11: Let's fix indent

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -0,0 +1,148 @@
...
+namespace {
+
+/** @todo:
+ *  1. deprecate tag dispatching along with #1072
+ *  2. deprecate getMergeableGraph (extendGraph should accept graph itself)
+ */
+class PlainLayer {};    /**< Plain Layer tag */
+class BackboneLayer {}; /**< Backbone Layer tag */
+
+template <typename T>
```
dongju-chae, line 42: What's its purpose?

zhoonit: This is specializes the function for parsing PlainLayer and parsing BackboneLayer (eventually this specialization won't be needed...)
Related commit:
https://github.com/nnstreamer/nntrainer/pull/1079/commits/15b4e4e375f232448bae21d841bb2ff3850b20dd

nntrainer/compiler/ini_interpreter.cpp
zhoonit: done

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -0,0 +1,148 @@
...
+      NNTR_THROW_IF_CLEANUP(!sec_name_, std::runtime_error, freedict)
+        << "parsing a section name returned error, filename: " << in
+        << "idx: " << idx;
+
+      ml_logd("probing section_name: %s", sec_name_);
+      std::string sec_name(sec_name_);
+
+      if (istrequal(sec_name, MODEL_STR) || istrequal(sec_name, DATASET_STR) ||
+          istrequal(sec_name, OPTIMIZER_STR)) {
+        /// dedicated sections so skip
```
kparichay, line 105: where are these dedicated sections called?

zhoonit: serializers are meant to work only on graph. Those are not related to graph. So they are not commonly treated inside an interpreter. unless we start to put those objects inside the graph.

nntrainer/compiler/ini_interpreter.h
```diff
@@ -0,0 +1,81 @@
...
+ * @date 02 April 2021
+ * @brief NNTrainer Ini Interpreter
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug No known bugs except for NYI items
+ */
+#include <iostream>
+#include <memory>
+
+#include <iniparser.h>
```
jijoongmoon, line 15: I wonder that model loader also use iniparser to parse. Is it to save graph to ini ?? or read ini? If it is to read ini then.. I guess better to use model_loader.

nntrainer/compiler/ini_interpreter.h
```diff
@@ -0,0 +1,81 @@
...
+                                         const std::string &section);
+
+  /**
+   * @brief Create a Layer From Backbone Config
+   *
+   * @param ini ini if throw, ini will be freed.
+   * @param section section name
+   * @return std::shared_ptr<Layer> layer
+   */
+  std::shared_ptr<Layer> loadBackboneConfigIni(dictionary *ini,
```
jijoongmoon, line 73: I think this ini and section info should get from model_loader?

nntrainer/compiler/ini_interpreter.h
zhoonit: Yes this is duplication so reading graph part will move here and related function call will be removed. https://github.com/nnstreamer/nntrainer/pull/1085

nntrainer/compiler/ini_interpreter.cpp
kparichay: so, deserialize will be called in combination with `model loader` or `deserialize` will use model loader internally.
also, section2layer and section2graph will also call `model loader` internally?

nntrainer/compiler/ini_interpreter.h
zhoonit: Yes, this function is a duplication from the `model_loader` this will be removed from `model_loader`

nntrainer/compiler/ini_interpreter.cpp
kparichay: from #1079, it seems like `section2layer` does not call model loader, but parses the ini itself. So, I am little confused how is the work distributed - layers and graph are parsed by ini_interpreter and these sections by the `model loader`?
If my understanding is correct, it would be best to move all to ini_interpreter or let all be model loader and call it from here.

zhoonit: 1. `model loader`  delegates parsing graph to the interpreter by calling `interpreter->deserialize`
2. so graph parsing related logics are moving to `ini_interpreter`
3. `model loader` uses `interpreter` but `interpreter` won't use `model loader`
+. Training related objects will be still parsed inside the `model loader`

zhoonit: The boundary of `interpreter` interface is parsing graph only. In this way, we can parse other type of backbones into nntrainer graph as well.

kparichay: Understood. Please add the dsecription above to `model loader` or `ini_interpreter`?

kparichay (approved):
LGTM

nntrainer/compiler/ini_interpreter.cpp
zhoonit: Added to the `interpreter.h`

kimjh12 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #1079, opened 2021-04-06
[IniInterpreter] Add section2layer

zhoonit:
- [IniInterpreter] Add section2layer 

```
This patch add plain `section2layer`.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/ini_interpreter.cpp
+++ b/nntrainer/compiler/ini_interpreter.cpp
@@ -39,10 +39,89 @@ namespace {
 class PlainLayer {};    /**< Plain Layer tag */
 class BackboneLayer {}; /**< Backbone Layer tag */
 
-template <typename T>
-static std::shared_ptr<Layer> section2layer(dictionary *ini,
+/**
+ * @brief convert section to list of string based properties
+ *
+ * @param ini ini handler
+ * @param sec_name section name
+ * @return std::vector<std::string> list of properties
+ */
+std::vector<std::string> section2properties(dictionary *ini,
                                             const std::string &sec_name) {
-  /// NYI!
+  int num_entries = iniparser_getsecnkeys(ini, sec_name.c_str());
+  NNTR_THROW_IF(num_entries < 1, std::invalid_argument)
+    << "there are no entries in the layer section: " << sec_name;
+  ml_logd("number of entries for %s: %d", sec_name.c_str(), num_entries);
+
+  std::unique_ptr<const char *[]> key_refs(new const char *[num_entries]);
+  NNTR_THROW_IF(iniparser_getseckeys(ini, sec_name.c_str(), key_refs.get()) ==
+                  nullptr,
+                std::invalid_argument)
+    << "failed to fetch keys for the section: " << sec_name;
+
+  std::vector<std::string> properties;
+  properties.reserve(num_entries - 1);
+
+  for (int i = 0; i < num_entries; ++i) {
+    std::string key(key_refs[i]);
+    std::string prop_key = key.substr(key.find(":") + 1);
+
+    if (istrequal(prop_key, "type") || istrequal(prop_key, "backbone")) {
+      continue;
+    }
+
+    std::string value = iniparser_getstring(ini, key_refs[i], UNKNOWN_STR);
+    NNTR_THROW_IF(value == UNKNOWN_STR || value.empty(), std::invalid_argument)
+      << "parsing property failed key: " << key << " value: " << value;
+
+    ml_logd("parsed properties: %s=%s", prop_key.c_str(), value.c_str());
+    properties.push_back(prop_key + "=" + value);
+  }
+
+  properties.push_back("name=" + sec_name);
+
+  return properties;
+}
+
+/**
+ * @brief convert section to a layer object (later it might be layer node)
+ *
+ * @tparam T tag
+ * @param ini dictionary * ini
+ * @param sec_name section name
+ * @param ac app context to search for the layer
+ * @return std::shared_ptr<Layer> return layer object
+ */
+template <typename T>
+std::shared_ptr<Layer> section2layer(dictionary *ini,
+                                     const std::string &sec_name,
+                                     const AppContext &ac) {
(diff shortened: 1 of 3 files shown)


nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -39,10 +39,88 @@ namespace {
...
+    iniparser_getstring(ini, (sec_name + ":Type").c_str(), UNKNOWN_STR);
+  NNTR_THROW_IF(layer_type == UNKNOWN_STR, std::invalid_argument)
+    << "section name is invalid, section name: " << sec_name;
+
+  auto properties = section2properties(ini, sec_name);
+  std::shared_ptr<ml::train::Layer> layer_ =
+    ac.createObject<ml::train::Layer>(layer_type, properties);
+
+  auto layer = std::static_pointer_cast<Layer>(layer_);
+  layer->setName(sec_name);
```
kparichay, line 115: can this be added to the properties as well?

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -39,10 +39,88 @@ namespace {
...
+    << "failed to fetch keys for the section: " << sec_name;
+
+  std::vector<std::string> properties;
+  properties.reserve(num_entries - 1);
+
+  for (int i = 0; i < num_entries; ++i) {
+    std::string key(key_refs[i]);
+    std::string prop_key = key.substr(key.find(":") + 1);
+
+    if (istrequal(prop_key, "type") || istrequal(prop_key, "backbone")) {
```
kparichay, line 69: please follow the latest `model loader` code style where the list of properties to be skipped is given by each section - layer, backbone, optimizer, etc have different properties to be skipped.

nntrainer/compiler/ini_interpreter.cpp
zhoonit: Good idea, updated

nntrainer/compiler/ini_interpreter.cpp
zhoonit: I checked this restriction and it works or fine with model_loader. please get me right if I am missing something.

kparichay: My earlier concern was that `type` and `backbone` can be passed externally as parameters, but as this is being used only for parsing layers, its fine.

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -39,10 +39,89 @@ namespace {
...
-template <typename T>
-static std::shared_ptr<Layer> section2layer(dictionary *ini,
+/**
+ * @brief convert section to list of string based properties
+ *
+ * @param ini ini handler
+ * @param sec_name section name
+ * @return std::vector<std::string> list of properties
+ */
+std::vector<std::string> section2properties(dictionary *ini,
```
kparichay, line 49: as `section2properties()` strongly overlaps with `parseProperties()` in ModelLoader, I recommend trying to keep only 1 implementation, either here or in ModelLoader, for easier maintenance. (you can make `ModelLoader::parseProperties()` as static public and use in `section2properties()` or vice versa)

zhoonit: Since model_loader is including this header, I'll make model_loader to use this with some tweak in other PR.

kparichay: Then you might want to make this function more generic (taking the properties keys to be skipped).

kparichay: sure, that works too.

jijoongmoon: Even, do we need model loader?

zhoonit: Yes, model loader will call those functions altogether.

jijoongmoon: I wonder that .. model loader ( which will be removed ) = section2properties + compiler?

zhoonit: model loader will parses and loads graph(from serializers) + optimizers + dataset + model hyper parameters.

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -39,10 +39,89 @@ namespace {
...
+    std::string value = iniparser_getstring(ini, key_refs[i], UNKNOWN_STR);
+    NNTR_THROW_IF(value == UNKNOWN_STR || value.empty(), std::invalid_argument)
+      << "parsing property failed key: " << key << " value: " << value;
+
+    ml_logd("parsed properties: %s=%s", prop_key.c_str(), value.c_str());
+    properties.push_back(prop_key + "=" + value);
+  }
+
+  properties.push_back("name=" + sec_name);
+
```
jijoongmoon, line 82: are we going to use this "a=b" format for properties?? If it is.. then we also need to parse it somehow.. isn't it overhead?

nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -39,10 +39,89 @@ namespace {
...
+ * @brief convert section to a layer object (later it might be layer node)
+ *
+ * @tparam T tag
+ * @param ini dictionary * ini
+ * @param sec_name section name
+ * @param ac app context to search for the layer
+ * @return std::shared_ptr<Layer> return layer object
+ */
+template <typename T>
+std::shared_ptr<Layer> section2layer(dictionary *ini,
```
jijoongmoon, line 96: Which one is better?? I mean.. section2layer?? or propertiest2layer? There is some overlap between section2layer and section2properties.

nntrainer/compiler/ini_interpreter.cpp
zhoonit: This is to conform to the API. Later, we can have better way to directly import. I have an idea but I guess it's low priority for now :).

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

PR #1087, opened 2021-04-08
[Ini] Propose ini wrapper

zhoonit:
- [Ini] Propose ini wrapper
This patch adds `IniSection` which was often used inside unittest. 

```
Most codes are taken from the test directly.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/utils/ini_wrapper.cpp
+++ b/nntrainer/utils/ini_wrapper.cpp
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file ini_wrapper.cpp
+ * @date 08 April 2021
+ * @brief NNTrainer Ini Wrapper helps to save ini
+ * @note this is to be used with ini_interpreter
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug No known bugs except for NYI items
+ */
+#include <ini_wrapper.h>
+
+#include <nntrainer_error.h>
+#include <parse_util.h>
+#include <regex>
+
+namespace nntrainer {
+
+IniSection::IniSection(const std::string &name) : section_name(name) {}
+
+IniSection::IniSection(const std::string &section_name,
+                       const std::string &entry_str) :
+  IniSection(section_name) {
+  setEntry(entry_str);
+}
+
+IniSection::IniSection(IniSection &from, const std::string &section_name,
+                       const std::string &entry_str) :
+  IniSection(from) {
+  if (!section_name.empty()) {
+    this->section_name = section_name;
+  }
+  if (!entry_str.empty()) {
+    setEntry(entry_str);
+  }
+}
+
+void IniSection::print(std::ostream &out) {
+  out << '[' << section_name << ']' << std::endl;
+  for (auto &it : entry)
+    out << it.first << " = " << it.second << std::endl;
+}
+
+void IniSection::setEntry(const std::map<std::string, std::string> &entry) {
+  for (auto &it : entry) {
+    this->entry[it.first] = it.second;
+  }
+}
+
+void IniSection::setEntry(const std::string &entry_str) {
+  // setting property separated by "|"
+  std::regex words_regex("[^|]+");
+
+  auto words_begin =
+    std::sregex_iterator(entry_str.begin(), entry_str.end(), words_regex);
+  auto words_end = std::sregex_iterator();
+
+  std::string key, value;
+  for (std::sregex_iterator i = words_begin; i != words_end; ++i) {
+    std::string cur = (*i).str();
+
+    if (cur[0] == '-') {
+      entry.erase(cur.substr(1));
+      continue;
+    }
+
+    int status = getKeyValue(cur, key, value);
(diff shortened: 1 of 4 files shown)


nntrainer/utils/ini_wrapper.h
```diff
@@ -0,0 +1,177 @@
...
+ * @file ini_wrapper.h
+ * @date 08 April 2021
+ * @brief NNTrainer Ini Wrapper helps to save ini
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug No known bugs except for NYI items
+ */
+
+#include <iostream>
+#include <map>
```
zhoonit, line 14: The code is taken from the test code. So marking it as a trivial.

nntrainer/utils/ini_wrapper.cpp
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file ini_wrapper.cpp
+ * @date 08 April 2021
+ * @brief NNTrainer Ini Wrapper helps to save ini
```
kparichay, line 7: is this going to be used to serialize the `GraphRepresentation` to ini format?

zhoonit: Yes.

kparichay: Please add to description.

zhoonit: done

jijoongmoon (approved):
LGTM

nntrainer/utils/ini_wrapper.cpp
```diff
@@ -0,0 +1,76 @@
...
+    out << it.first << " = " << it.second << std::endl;
+}
+
+void IniSection::setEntry(const std::map<std::string, std::string> &entry) {
+  for (auto &it : entry) {
+    this->entry[it.first] = it.second;
+  }
+}
+
+void IniSection::setEntry(const std::string &entry_str) {
```
kparichay, line 52: it's ok for now, but later, you could consider to use std::vector<std::string> than using a separator `|`.

nntrainer/utils/ini_wrapper.h
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * @brief Construct a new Ini Section object
+   *
+   * @param from Ini Section to copy from
+   * @param entry_str entry string to override the given section name
+   */
+  IniSection(IniSection &from, const std::string &entry_str) :
+    IniSection(from, "", entry_str) {}
+
+  IniSection() = default;
```
kparichay, line 67: is this needed?

nntrainer/utils/ini_wrapper.h
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * @brief set entry from the string representation
+   *
+   * @param entry_str setEntry as "Type = neuralnetwork | decayrate = 0.96 |
+   * -epochs = 1" will delete epochs, and overwrite type and decayrate
+   */
+  void setEntry(const std::string &entry_str);
+
+  std::string section_name; /**< section name of the ini section */
+  std::map<std::string, std::string>
```
kparichay, line 161: consider using `unordered_map` to make finding a section faster.

nntrainer/utils/ini_wrapper.cpp
zhoonit: yes, this could be optimized later, fyi) this method should kept for the test codes though.
I am putting a todo.

nntrainer/utils/ini_wrapper.h
zhoonit: Yes, trivial default ctor is needed for the containers (eg) `std::vector<IniSection>`)

nntrainer/utils/ini_wrapper.h
zhoonit: I changed to ordered map to keep the order predictable, I checked with current tests and overhead is minimal.

nntrainer/utils/ini_wrapper.cpp
kparichay: When this is updated, the test codes can be updated as well (in a later PR).

nntrainer/utils/ini_wrapper.h
kparichay: ok.

nntrainer/utils/ini_wrapper.h
kparichay: > I changed to ordered map to keep the order predictable,

Any reason for this? For the ini input file, we can't assume any order.

zhoonit: This is for testing purpose. If this is critical to the performance, I could change but I think it's fine as one section wouldn't have more than 10 or 20 entries.

nntrainer/utils/ini_wrapper.cpp
zhoonit: For sure :)

kparichay (approved):
LGTM

nntrainer/utils/ini_wrapper.h
zhoonit: Done

PR #1091, opened 2021-04-08
[Ini/Serialize] Implement logic to serialize

zhoonit:
- [Ini/Serialize] Implement logic to serialize 

```
This patch implements general concept of serializing a node.

Below is the plan to implement this logic

0. Implement logic to serialize [*]
1. Add export api
2. Add exporter
3. Add export test
4. Implement properties
5. combine with export
6. Add extra test

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/ini_interpreter.cpp
+++ b/nntrainer/compiler/ini_interpreter.cpp
@@ -13,10 +13,12 @@
  */
 #include <ini_interpreter.h>
 
+#include <sstream>
 #include <vector>
 
 #include <layer.h>
 #include <layer_factory.h>
+#include <ini_wrapper.h>
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <parse_util.h>
@@ -46,9 +48,7 @@ namespace {
 /** @todo:
  *  1. deprecate tag dispatching along with #1072
  *  2. deprecate getMergeableGraph (extendGraph should accept graph itself)
- */
-
-/**
+ *
  * @brief Plain Layer tag
  */
 class PlainLayer {};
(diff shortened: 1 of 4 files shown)


nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -247,7 +249,22 @@ getMergeableGraph(std::shared_ptr<const GraphRepresentation> graph,
 
 void IniGraphInterpreter::serialize(
   std::shared_ptr<const GraphRepresentation> representation,
-  const std::string &out) {}
+  const std::string &out) {
+
+  std::vector<IniSection> sections;
+  for (const auto &ln : representation->getSorted()) {
```
kparichay, line 255: any reason to use `getSorted()`?

kparichay: The ini representation is independent of the order of the sections.

zhoonit: AFAIK, there are some bugs if first layer is not input layer or output layer is before input layer and so on..
Even though above is sorted out, it is neater and more human readable to make it sorted.

kparichay: > AFAIK, there are some bugs if first layer is not input layer or output layer is before input layer and so on..

I think this was resolved when the network_graph was refactored (not yet with backbone though).

kparichay: > Even though above is sorted out, it is neater and more human readable to make it sorted.

Then lets use unsorted representation in debug mode to make sure that the all styles work.

zhoonit: Let me have a todo instead, I am also testing backbone.

kparichay: ok.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

kparichay (approved):
LGTM

PR #1092, opened 2021-04-08
[Serializer] Add exporter to export node

zhoonit:
This patch adds exporter to export node. This exploits visitor pattern
because, exporting need templated function to be executed, but virtual
functions does not allow templated function.

**Self evaluation:**

### DIFF
--- a/nntrainer/utils/node_exporter.h
+++ b/nntrainer/utils/node_exporter.h
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file node_exporter.h
+ * @date 08 April 2021
+ * @brief NNTrainer Node exporter
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug No known bugs except for NYI items
+ */
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+#ifndef __NODE_EXPORTER_H__
+#define __NODE_EXPORTER_H__
+
+namespace nntrainer {
+enum class ExportMethods {
+  METHOD_STRINGVECTOR = 0, /**< export to a string vector */
+  METHOD_TFLITE = 1,       /**< epxort to tflite */
+  METHOD_UNDEFINED = 999,  /**< undefined */
+};
+
+namespace {
+
+template <ExportMethods method> struct return_type { using type = void; };
+
+template <> struct return_type<ExportMethods::METHOD_STRINGVECTOR> {
+  using type = std::vector<std::pair<std::string, std::string>>;
+};
+} // namespace
+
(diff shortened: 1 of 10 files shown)


jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

nntrainer/utils/node_exporter.h
```diff
@@ -0,0 +1,88 @@
...
+   * @brief this function iterates over the property and process the property in
+   * a designated way.
+   *
+   * @tparam Ts type of elements
+   * @param props tuple that contains properties
+   * @param method method to export
+   */
+  template <typename... Ts>
+  void save_result(std::tuple<Ts...> &props, ExportMethods method) {
+    if (is_exported) {
```
kparichay, line 60: I wonder about this check. It is not allowed to export a set of properties twice, or essentially save the model twice?

zhoonit: I made a strong assumption that this would be only for one time use. But this was changed in the upcoming PR.
You can think `is_exported` is gone lol.

nntrainer/utils/node_exporter.h
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * @brief this function iterates over the property and process the property in
+   * a designated way.
+   *
+   * @tparam Ts type of elements
+   * @param props tuple that contains properties
+   * @param method method to export
+   */
+  template <typename... Ts>
+  void save_result(std::tuple<Ts...> &props, ExportMethods method) {
```
kparichay, line 59: just remember, this can lead to lots of classes version of this functions (i think worst case should be number of classes using exporter, but you can give a check).

Also, IMO this header will have to available with `nntrainer-devel` so that custom layer can save their properties with this.

nntrainer/utils/node_exporter.h
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * @brief Get the result object
+   *
+   * @tparam methods method to get
+   * @tparam T appropriate return type regarding the export method
+   * @return T T
+   */
+  template <ExportMethods methods, typename T = return_type<methods>>
+  T get_result() {
+    if (!is_exported) {
```
kparichay, line 77: is this the inverse of `save_result` which imports?

nntrainer/utils/node_exporter.h
kparichay: cool.

nntrainer/utils/node_exporter.h
zhoonit: Yes

nntrainer/utils/node_exporter.h
zhoonit: Yeah, I think we can have a dedicated api wrapping around this I guess.

nntrainer/utils/node_exporter.h
kparichay: This is fine for now, but i think this should be eliminated.

kparichay (approved):
LGTM

Issue #1094, opened 2021-04-08
Inference_inout_memory_opt in inference mode


lhs8928:
Problem: Enable inference_inout_memory_opt in inference could make a problem

Cause: If the networks are not flatten which means have a cycle, some of output of layer should be preserved.
For example, in resnet there is a residual output and if the opt is enabled this data could be corrupted.

Solution(?): Check if the layer have a multiple input or output.

zhoonit:
elaborating the yesterday's debugging,
for the resnet, in/out order is not even/odd because the input memory has to be kept to feed the addition layer.

we will need a test for the addition layer inside a model as well.

kparichay:
Memory optimizations have not been tested well with mutli-input/output layers as we didn't have any models using them properly.
This will have to be done soon.

PR #1095, opened 2021-04-08
[manager] disable inference_memory_opt

lhs8928:
Enable inference_inout_memory_opt in inference could make a problem

Cause: If the networks are not flatten which means have a cycle, some of output of layer should be preserved.
For example, in resnet there is a residual output and if the opt is enabled this data could be corrupted.

See Also #1094

Self evaluation:

### DIFF
--- a/nntrainer/tensor/manager.h
+++ b/nntrainer/tensor/manager.h
@@ -93,7 +93,7 @@ public:
   Manager(bool enable_gradient_memory_opt_ = true,
           bool enable_derivative_memory_opt_ = true,
           bool enable_activation_memory_opt_ = true,
-          bool enable_inference_inout_memory_opt_ = true);
+          bool enable_inference_inout_memory_opt_ = false);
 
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/manager.h
```diff
@@ -93,7 +93,7 @@ class Manager {
   Manager(bool enable_gradient_memory_opt_ = true,
           bool enable_derivative_memory_opt_ = true,
           bool enable_activation_memory_opt_ = true,
-          bool enable_inference_inout_memory_opt_ = true);
+          bool enable_inference_inout_memory_opt_ = false);
```
dongju-chae, line 96: How about adding some comments to describe this workaround?

dongju-chae (approved):
LGTM

kparichay (approved):
LGTM
Just to make the terminology clearer, the issue when using layers with multiple inputs/outputs (not a cycle, there can't be a cycle in this graph).

PR #1102, opened 2021-04-09
[Exporter] Add exporter test to prop test

zhoonit:
- [Exporter] Add exporter test to prop test 

```
This patch adds exporter test to prop test to ensure that it is working
fine.

**Self evaluation:**

```

- [Exporter] Add string vector exporter 

```
This patch adds string vector exporter implementation

Exporter will iterate through props (`iterate_props`) and will pass a
generic lambda to be called.

Later, We can retrieve the result by `get_result` if needed.

**Self evaluation:**

```

### DIFF
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -91,6 +91,7 @@ NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_factory.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_func.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_file.cpp \
+                  $(NNTRAINER_ROOT)/nntrainer/compiler/ini_interpreter.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/tensor/tensor.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/tensor/lazy_tensor.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/tensor/manager.cpp \
@@ -128,8 +129,7 @@ NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/ini_wrapper.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/parse_util.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/profiler.cpp \
-                  $(NNTRAINER_ROOT)/nntrainer/utils/profiler.cpp \
-                  $(NNTRAINER_ROOT)/nntrainer/compiler/node_exporter.cpp \
+                  $(NNTRAINER_ROOT)/nntrainer/utils/node_exporter.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/app_context.cpp
 
 # Add tflite backbone building
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -128,7 +128,8 @@ NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/ini_wrapper.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/parse_util.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/utils/profiler.cpp \
-                  $(NNTRAINER_ROOT)/nntrainer/compiler/ini_interpreter.cpp \
+                  $(NNTRAINER_ROOT)/nntrainer/utils/profiler.cpp \
+                  $(NNTRAINER_ROOT)/nntrainer/compiler/node_exporter.cpp \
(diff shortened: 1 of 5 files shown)


nntrainer/utils/node_exporter.h
```diff
@@ -56,11 +58,22 @@ class Exporter {
    * @param method method to export
    */
   template <typename... Ts>
-  void save_result(std::tuple<Ts...> &props, ExportMethods method) {
-    if (is_exported) {
-      throw std::invalid_argument("This exporter is already used");
+  void save_result(const std::tuple<Ts...> &props, ExportMethods method) {
+    switch (method) {
+    case ExportMethods::METHOD_STRINGVECTOR: {
+      auto callable = [this](auto &&prop, size_t index) {
```
lhs8928, line 64: It seems that variable index doesn't used anywhere

lhs8928 (approved):
LGTM

jijoongmoon:
I wonder the STRINGVECTOR exporter method is a kind of preprocessing before export for tflite or nntrainer??
If it is.. then.. I'm not sure they are same enum value.

jijoongmoon (approved):
I wonder the STRINGVECTOR exporter method is a kind of preprocessing before export for tflite or nntrainer??
If it is.. then.. I'm not sure they are same enum value.

For now, LGTM

jni/Android.mk
```diff
@@ -91,6 +91,7 @@ NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_factory.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_func.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_file.cpp \
+									$(NNTRAINER_ROOT)/nntrainer/compiler/ini_interpreter.cpp \
```
kparichay, line 94: check indent please.

nntrainer/utils/node_exporter.h
```diff
@@ -3,7 +3,7 @@
  * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
  *
  * @file node_exporter.h
- * @date 08 April 2021
+ * @date 09 April 2021
```
kparichay, line 6: is this intended?

nntrainer/utils/node_exporter.h
```diff
@@ -56,11 +73,22 @@ class Exporter {
    * @param method method to export
    */
   template <typename... Ts>
-  void save_result(std::tuple<Ts...> &props, ExportMethods method) {
-    if (is_exported) {
-      throw std::invalid_argument("This exporter is already used");
+  void save_result(const std::tuple<Ts...> &props, ExportMethods method) {
+    switch (method) {
+    case ExportMethods::METHOD_STRINGVECTOR: {
+      auto callable = [this](auto &&prop, size_t index) {
```
kparichay, line 79: please add some brief about this lambda function here.
also, as for iterate_prop, callable is a template parameter, you can consider removing `index` for readability, and include it in other methods when needed.

nntrainer/utils/node_exporter.h
```diff
@@ -72,16 +100,51 @@ class Exporter {
...
+  template <size_t I = 0, typename Callable, typename... Ts>
+  typename std::enable_if<(I < sizeof...(Ts)), void>::type
+  iterate_prop(Callable &&c, const std::tuple<Ts...> &tup) {
+    c(std::get<I>(tup), I);
+
+    iterate_prop<I + 1>(c, tup);
+  }
+
+  std::vector<std::pair<std::string, std::string>>
+    stored_result; /**< stored result */
```
kparichay, line 143: This is fine for now, but later you can consider if this is needed. This is dangerous as there is no way to know if this stored_result is valid now, as some properties might have changed since then.

nntrainer/utils/node_exporter.h
zhoonit: yes.

nntrainer/utils/node_exporter.h
zhoonit: I think it is sensible that returned value of get_property can always be invalidated after set_property is called.

I think it is also sensible not to call get_property in the middle of set_property.
But If we need to make it sure, we can have a mutex I guess.

nntrainer/utils/node_exporter.h
zhoonit: I'll put documentation here.

nntrainer/utils/node_exporter.h
zhoonit: It was 09 April but I just copy and pasted the doc and fixed it later lol

jni/Android.mk
zhoonit: Done

nntrainer/utils/node_exporter.h
kparichay: :)

nntrainer/utils/node_exporter.h
kparichay: The scenario you described is a legitimate one. However, there can be another scenario too.
```
model.train('batch_size=32');
model.saveResult();
model.train('batch_size=16');
model.getResult(); // <- batch size in this result will be wrong and not sure, if mutex can resolve this
```

kparichay (approved):
LGTM

PR #1104, opened 2021-04-09
[Props] Implement load properties

zhoonit:
- [Props] Implement load properties 
```
This patch implement `load_properties` which helps loading properties
for a tuple of properties.

**Self evaluation:**

```
- v2. [NDK] Add frtti to application build
```
As real time type information is used in nntrainer, frtti flag is turned
on to fix build issue

**Self evaluation:**

```

### DIFF
--- a/nntrainer/utils/node_exporter.h
+++ b/nntrainer/utils/node_exporter.h
@@ -15,6 +15,7 @@
 #include <vector>
 
 #include <nntrainer_error.h>
+#include <parse_util.h>
 
 #ifndef __NODE_EXPORTER_H__
 #define __NODE_EXPORTER_H__
@@ -114,40 +115,6 @@ public:
   const T &get_result();
 
 private:
-  /**
-   * @brief base case of iterate_prop, iterate_prop iterates the given tuple
-   *
-   * @tparam I size of tuple(automated)
-   * @tparam Callable generic lambda to be called during iteration
-   * @tparam Ts types from tuple
-   * @param c callable gerneric labmda
-   * @param tup tuple to be iterated
-   * @return void
-   */
-  template <size_t I = 0, typename Callable, typename... Ts>
-  typename std::enable_if<I == sizeof...(Ts), void>::type
-  iterate_prop(Callable &&c, const std::tuple<Ts...> &tup) {
-    // end of recursion;
-  }
-
-  /**
-   * @brief base case of iterate_prop, iterate_prop iterates the given tuple
-   *
-   * @tparam I size of tuple(automated)
-   * @tparam Callable generic lambda to be called during iteration
-   * @tparam Ts types from tuple
-   * @param c callable gerneric labmda
-   * @param tup tuple to be iterated
-   * @return not used
-   */
-  template <size_t I = 0, typename Callable, typename... Ts>
-  typename std::enable_if<(I < sizeof...(Ts)), void>::type
-  iterate_prop(Callable &&c, const std::tuple<Ts...> &tup) {
-    c(std::get<I>(tup), I);
-
-    iterate_prop<I + 1>(c, tup);
-  }
-
   std::vector<std::pair<std::string, std::string>>
     stored_result; /**< stored result */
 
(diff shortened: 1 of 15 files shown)


jijoongmoon (approved):
Is there any overhead to use type checking for every operation?

nntrainer/utils/parse_util.cpp
```diff
@@ -38,19 +38,23 @@
 
 namespace nntrainer {
 
-int getKeyValue(std::string &input_str, std::string &key, std::string &value) {
+int getKeyValue(const std::string &input_str, std::string &key,
```
kparichay, line 41: Awesome that you make this parameter const.
Now, every setProperty can use `const std::string &`.

nntrainer/utils/node_exporter.h
```diff
@@ -147,5 +114,90 @@ class Exporter {
...
+                std::tuple<Ts...> &props) {
+
+  std::vector<std::string> left = string_vector;
+
+  auto callable = [&left](auto &&prop, size_t index) {
+    std::string prop_key = std::remove_reference_t<decltype(prop)>::key;
+
+    for (auto iter = left.begin(); iter < left.end(); ++iter) {
+      std::string key, value;
+      int status = getKeyValue(*iter, key, value);
```
kparichay, line 186: in this logic, for every prop_key, the string is being parse into key and value in each loop.
Inverse the order of loops to parse the string just once, and re-use key-value pair.

nntrainer/utils/node_exporter.h
```diff
@@ -15,6 +15,7 @@
 #include <vector>
 
 #include <nntrainer_error.h>
+#include <parse_util.h>
```
kparichay, line 18: it seems like parse_util will now also be needed to export in devel header as node_exporter is needed by layer_internal.h

zhoonit: Actually, we are only using `getKeyValue` and most of them can be deprecated, at the end of the day, we can make a decision to put `getKeyValue` as static here instead of exposing `parse_util`

nntrainer/utils/node_exporter.h
zhoonit: Done

zhoonit:
> Is there any overhead to use type checking for every operation?

No, checking type happens at the compile time. So the whole property managing system is lightweight. (getting item from tuple happens at compile time as well)

nntrainer/utils/node_exporter.h
kparichay: alright.

nntrainer/utils/node_exporter.h
```diff
@@ -156,5 +123,104 @@ class Exporter {
...
+                   return std::make_pair(key, value);
+                 });
+
+  auto callable = [&left](auto &&prop, size_t index) {
+    std::string prop_key = std::remove_reference_t<decltype(prop)>::key;
+
+    for (auto iter = left.begin(); iter < left.end(); ++iter) {
+      if (istrequal(prop_key, iter->first) == true) {
+        from_string(iter->second, prop);
+        iter = left.erase(iter);
```
kparichay, line 207: huh, the properties which match the `props` are erased. So, left contains the properties which do not match, and are put into remainder and returned. I thought this function returns the properties which matches between props and string_vector.

nntrainer/utils/node_exporter.h
```diff
@@ -156,5 +123,104 @@ class Exporter {
...
+                     << property;
+                   return std::make_pair(key, value);
+                 });
+
+  auto callable = [&left](auto &&prop, size_t index) {
+    std::string prop_key = std::remove_reference_t<decltype(prop)>::key;
+
+    for (auto iter = left.begin(); iter < left.end(); ++iter) {
+      if (istrequal(prop_key, iter->first) == true) {
+        from_string(iter->second, prop);
```
kparichay, line 206: I wonder when are these properties set. The return value of from_string is ignored, and the iterator is erased as well.

zhoonit: The signature of `from_string` is `void from_string(const std::string &str, T &property)`. This sets the value.

nntrainer/utils/node_exporter.h
zhoonit: Updated to do less operation.

kparichay (approved):
LGTM

nntrainer/utils/node_exporter.h
kparichay: aah, I checked the wrong from_string.

PR #1105, opened 2021-04-12
[ TIME DIST ] Implement Initialization of Time Dist Layer @open sesame 04/27 19:44

jijoongmoon:
This commit includes,
  - Time Dist Layer initialization
  - Some modification of network graph

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -208,13 +208,9 @@ void NetworkGraph::ensureName(std::shared_ptr<Layer> layer,
   std::set<std::string>::iterator iter;
   std::string name;
   if (orig_name_empty) {
-    if (layer->getType() == TimeDistLayer::type) {
-      orig_name =
-        std::dynamic_pointer_cast<TimeDistLayer>(layer)->getDistLayerType();
-    } else {
-      orig_name = layer->getType();
-    }
+    orig_name = layer->getType();
   }
+
   std::string direct_name = prefix + orig_name;
 
   do {
@@ -319,11 +315,14 @@ int NetworkGraph::realizeActivationType(Layer &current) {
   std::shared_ptr<Layer> layer = nntrainer::createLayer(ActivationLayer::type);
   layer->setActivation(act);
 
-  if (current.getType() == TimeDistLayer::type) {
-    layer = distributeLayer(layer);
-  }
-
   ensureName(layer, current.getName());
+
+  if (current.getType() == TimeDistLayer::type) {
+    std::string unit_str = layer->getName();
+    layer->setName(layer->getName() + "_unit");
+    layer = distributeLayer(layer);
+    layer->setName(unit_str);
+  }
 
   layer->setNumInputs(current.getNumInputs());
   layer->input_layers.clear();
@@ -425,11 +424,14 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
     std::dynamic_pointer_cast<LossLayer>(layer)->setLoss(updated_loss_type);
   NN_RETURN_STATUS();
 
-  if (last_node.getObject()->getType() == TimeDistLayer::type) {
-    layer = distributeLayer(layer);
-  }
-
   ensureName(layer);
+
+  if (last_node.getObject()->getType() == TimeDistLayer::type) {
+    std::string unit_str = layer->getName();
+    layer->setName(unit_str + "_unit");
+    layer = distributeLayer(layer);
+    layer->setName(unit_str);
+  }
 
   last_node.getObject()->setNumOutputs(1);
   last_node.getObject()->output_layers.clear();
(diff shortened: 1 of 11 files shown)


nntrainer/layers/layer_internal.h
```diff
@@ -472,6 +472,14 @@ class Layer : public ml::train::Layer {
     net_hidden.resize(size);
   }
 
+  /**
+   * @brief Set the input dimension
+   * @param[in] d dimension to be set
+   */
+  void setInputDimension(std::vector<TensorDim> d) { input_dim = d; }
+
+  void setInputDimension(TensorDim d, unsigned int i) { input_dim[i] = d; }
```
zhoonit, line 481: could you check for the range before assign? svace will complain :p

zhoonit, line 481:
```suggestion
 void setInputDimension(const TensorDim &d, unsigned int i) { input_dim[i] = d; }
```

nntrainer/layers/layer_internal.h
```diff
@@ -472,6 +472,14 @@ class Layer : public ml::train::Layer {
     net_hidden.resize(size);
   }
 
+  /**
+   * @brief Set the input dimension
+   * @param[in] d dimension to be set
+   */
+  void setInputDimension(std::vector<TensorDim> d) { input_dim = d; }
```
zhoonit, line 479:
```suggestion
 void setInputDimension(const std::vector<TensorDim>& d) { input_dim = d; }
```

zhoonit (approved):
LGTM :)

nntrainer/graph/network_graph.cpp
```diff
@@ -316,12 +312,15 @@ int NetworkGraph::realizeActivationType(Layer &current) {
   std::shared_ptr<Layer> layer = nntrainer::createLayer(ActivationLayer::type);
   layer->setActivation(act);
 
+  ensureName(layer, current.getName());
+
   if (current.getType() == TimeDistLayer::type) {
+    std::string unit_str = layer->getName();
+    layer->setName(layer->getName() + "_unit");
```
kparichay, line 319: please ensure that this name does not already exist.

nntrainer/graph/network_graph.cpp
```diff
@@ -422,12 +421,14 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
     std::dynamic_pointer_cast<LossLayer>(layer)->setLoss(updated_loss_type);
   NN_RETURN_STATUS();
 
+  ensureName(layer);
   if (last_node.layer->getType() == TimeDistLayer::type) {
+    std::string unit_str = layer->getName();
+    layer->setName(unit_str + "_unit");
```
kparichay, line 427: same

nntrainer/layers/layer_internal.h
```diff
@@ -454,24 +466,54 @@ class Layer : public ml::train::Layer {
...
+   * @param[in] d dimension to be set
+   */
+  void setInputDimension(const std::vector<TensorDim> &d) { input_dim = d; }
+
+  /**
+   * @brief Set the input dimension
+   * @param[in] d dimension to be set
+   * @param[in] i axis
+   */
+  void setInputDimension(const TensorDim &d, unsigned int i) {
```
kparichay, line 510: is it necessary to be make these public? can you please use `setProperty` as these will have to removed eventually?

jijoongmoon: > is it necessary to be make these public? can you please use `setProperty` as these will have to removed eventually?

as you know, there are no interface to get the input dimension except input layer. The input dimension and output dimension is calculated inside of layer just before initialize() is called. So.. setProperty state is quite earlier than this and also it is not proper to ask more information from user.  Is there any reason not to be public ??

nntrainer/graph/network_graph.cpp
jijoongmoon: well.. I guess, layer->getName() is unique.. then this is unique. And the uniqueness is checked already.

nntrainer/graph/network_graph.cpp
jijoongmoon: > please ensure that this name does not already exist.

well.. I guess, layer->getName() is unique.. then this is unique. And the uniqueness is checked already.

kparichay: `layer->getName()` is unqiue, but layer->getName() + 'xyz' can exist. There is no check to verify this.

jijoongmoon: This will be updated in #1144

nntrainer/layers/layer_internal.h
kparichay: input dimension for a layer can be updated using `setProperty` just like any property.
`input_shape` is already supported as a basic property for all layers. 

> So.. setProperty state is quite earlier than this and also it is not proper to ask more information from user

`setProperty` still works for this, allowing to update the `input_shape`. As internal code will be doing it, it can be assumed that it will be done properly. Please consider updating this in a later PR.

> Is there any reason not to be public ??

Opening this up means that all custom layers will have to provide support for another function call.

kparichay (approved):
LGTM
Please check my other comment, and consider applying it in a later PR.

lhs8928 (approved):
LGTM

PR #1107, opened 2021-04-12
[IniSerializer] Add `serialize` to ini

zhoonit:
- [IniSerializer] Add `serialize` to ini 

```
- implements ini interpreter::serialize
- add corresponding test (reference shoud be equal to
serialize->deserialized graph)
- Fix some const correctness

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/ini_interpreter.cpp
+++ b/nntrainer/compiler/ini_interpreter.cpp
@@ -21,6 +21,7 @@
 #include <layer_factory.h>
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
+#include <node_exporter.h>
 #include <parse_util.h>
 #include <time_dist.h>
 #include <util_func.h>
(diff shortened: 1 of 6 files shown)


nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -275,12 +276,21 @@ void IniGraphInterpreter::serialize(
...
+    layer->export_to(e);
+
+    const auto key_val_pairs =
+      e.get_result<ExportMethods::METHOD_STRINGVECTOR>();
+
+    for (const auto &pair : key_val_pairs) {
+      s.setEntry(pair.first, pair.second);
+    }
+
+    sections.push_back(s);
```
jijoongmoon, line 289: Just for a reminder, the order of sections is matters.

test/unittest/compiler/unittest_interpreter.cpp
```diff
@@ -45,6 +45,41 @@ const std::string pathResolver(const std::string &path) {
...
+    nntrainer::Exporter rhs_export;
+
+    l.export_to(lhs_export);
+    r.export_to(rhs_export);
+
+    /*** fixme, there is one caveat that order matters in this form */
+    EXPECT_EQ(
+      lhs_export.get_result<nntrainer::ExportMethods::METHOD_STRINGVECTOR>(),
+      rhs_export.get_result<nntrainer::ExportMethods::METHOD_STRINGVECTOR>());
+  };
```
jijoongmoon, line 74: I wonder how we define "equal two network graphs"?

nntrainer/compiler/ini_interpreter.cpp
zhoonit: I think activation should exist as a layer when we serialize (we have an assumption that it's been already trained). 
And I agree that time disributed layer should be considered later.

test/unittest/compiler/unittest_interpreter.cpp
zhoonit: I agree, making a graph equal have some different way to define.
For now, I don't have any idea to cover all the requirements, so didn't actually make an equal operator.

My thinking is that we could make a discussion if equal operator is needed in the future.

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

PR #1112, opened 2021-04-13
[Tflite] Implement TfOpNode

zhoonit:
- [Tflite] Implement TfOpNode 

```
This patch implements TfOpNode which will be used to deserialize layer
node to a mere chunk of informations. Please note that those methods are
tentative and subject to change wile implementing other methods

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -11,15 +11,19 @@
  */
 #include <tflite_interpreter.h>
 
+#include <algorithm>
 #include <fstream>
 #include <memory>
 #include <string>
 
 #include <tf_schema_generated.h>
 
+#include <fc_layer.h>
 #include <nntrainer_error.h>
 #include <node_exporter.h>
 #include <tensor.h>
+#include <var_grad.h>
+#include <weight.h>
 
 #define UNUSED(x) x __attribute__((unused))
 #define WILL(x) ;
@@ -27,6 +31,7 @@
 static constexpr const char *FUNC_TAG = "[TFLITE INTERPRETER] ";
 
 namespace nntrainer {
+
 namespace {
 /**
  * @brief after finishing building, call this to safe to a file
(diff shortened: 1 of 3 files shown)


nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -56,12 +61,135 @@ void builder2file(const flatbuffers::FlatBufferBuilder &builder,
...
+  TfOpNode() = default;
+
+  /**
+   * @brief Construct a new Tf Op Node object from layer
+   * @note this is a shortcut to skip if layer does not need to be devided or
+   * fused
+   * @param layer layer that is converted to TfOpNode
+   */
+  TfOpNode(const Layer &layer) {
+    setInputs(layer.getInputRef());
```
kparichay, line 76: IMO for cleaner code, setInputs, setOutputs, setWeights can use a common implementation or the same static function with parameters to set.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -56,12 +61,135 @@ void builder2file(const flatbuffers::FlatBufferBuilder &builder,
...
+
+  /**
+   * @brief Set the Inputs object from layer
+   *
+   * @param inputs_ input to be inserted
+   */
+  void setInputs(const std::vector<std::shared_ptr<Var_Grad>> &inputs_) {
+    inputs.reserve(inputs_.size());
+    std::transform(inputs_.begin(), inputs_.end(), std::back_inserter(inputs),
+                   [](const auto &data) { return data.get(); });
```
kparichay, line 90: This code will create a new Var_Grad object for every element in the vector. Is this intended?

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -56,12 +61,135 @@ void builder2file(const flatbuffers::FlatBufferBuilder &builder,
  *
  */
 class TfOpNode {
```
kparichay, line 63: do you plan to use this node loading the tflite file as well?

zhoonit: I was but I might not as we decided to use tflite directly. Things are tentative regarding loading yet.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Could you give me an example? You mean templated code?

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: This is weak reference to the function. else I can't say they are equivalent.  The data structure does not own the pointer. Also, I think this case is fairly safe as it should be guaranteed that the given graph doesn't get destroyed while saving.

If you urge not to use, can you give me some alternatives for this purpose?

nntrainer/compiler/tflite_interpreter.cpp
kparichay: ok.

nntrainer/compiler/tflite_interpreter.cpp
kparichay: I meant a templated common function which these functions can call. Feel free to do in a later PR.

nntrainer/compiler/tflite_interpreter.cpp
kparichay: Yes, a templated common function which these functions can call. Feel free to do in a later PR.

zhoonit: Okay

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: > but if someone re-uses this later, this will be a hard bug to find.

I think putting this to shared_ptr hide bugs that model got destroyed but the opNode still works.

Another problem is that weights are not shared_ptr<Var_Grad>, I can't get the original reference of the weights other than this. If this is resolved, I can change to shared_ptr but for now shared_ptr can't be used.

Yet another problem is each `Var_Grad *` is used as a hash key to the index map. So index map will anyway going to have the raw pointer managing inside. (Defining a hash function for a shared_ptr is possible, but IMO this is an overkill for this stage, but later polished.)

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Alright. I'll try to migrate to tf data structure if possible.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: We discussed offline that this is (kind of) resolved using explicit ownership through `unique_ptr` in the upcoming PR.

kparichay (approved):
LGTM

PR #1113, opened 2021-04-13
[TfLite] Implement TfOpIdxMap

zhoonit:
- [TfLite] Implement TfOpIdxMap 

```
This patch implments TfOpIdxMap
Please note that graph::initialize has been extracted for modularity and
testability within this patch.

**Self evaluation:**

```


nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -199,30 +200,114 @@ class TfOpNode {
...
-private:
   /**
-   * @brief Bidirectional Index map
+   * @brief Get the Index of the data
    *
-   * @tparam T type of a underyling value
+   * @param key data that will be the key
+   * @return unsigned int index
    */
-  template <typename T> class BidirectionalIndexMap {
```
zhoonit: `BidirectionalIndexMap` extracted out of `TfOpIdxMap`

nntrainer/graph/network_graph.cpp
```diff
@@ -857,4 +857,79 @@ std::vector<LayerNode> &NetworkGraph::getSorted() {
   return Sorted;
 }
 
+int NetworkGraph::initialize(std::shared_ptr<Manager> manager) {
```
zhoonit, line 860: `NetworkGraph::iniaitialize` has been extracted from model for better testing & interface.

zhoonit: This is just copy & paste.

test/unittest/compiler/unittest_interpreter.cpp
```diff
@@ -157,16 +157,28 @@ TEST_P(nntrainerInterpreterTest, graphSerializeAfterDeserialize) {
...
+  auto g = makeGraph({fc0, fc1});
+  EXPECT_EQ(g->compile(nntrainer::LossType::LOSS_NONE), ML_ERROR_NONE);
+  EXPECT_EQ(g->initialize(manager), ML_ERROR_NONE);
+
+  manager->initializeWeights();
+  manager->allocateWeights();
+
   interpreter.serialize(g, "test.tflite");
+
+  manager->deallocateWeights();
```
zhoonit, line 181: fyi) if i don't deallocateWeights here, there is a segfault. But I think it's gcc's inherent bug.
related link: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89303

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -199,30 +200,114 @@ class TfOpNode {
...
+        for (auto &variable : variables) {
+          variable_map.addDataWhenNotFound(variable);
+        }
+      };
+
+    for (auto &op_node : nodes) {
+      update_opcode(op_node.getOpType());
+      update_variables(op_node.getInputs());
+      update_variables(op_node.getOutputs());
+      update_variables(op_node.getWeights());
```
kparichay, line 295: would it better to create a different map for weights? if you want to later use this map for saving the model to tflite, it saves the trouble of extracting the weights. Also, I am not sure at this point, but weights might need different treatment like datatypes etc compared to inputs/outputs in later time.

kparichay: Maybe buffer map can serve that purpose, but then full var_grad might be needed than just buffer.

nntrainer/models/neuralnet.cpp
```diff
@@ -168,90 +168,8 @@ int NeuralNetwork::initialize() {
...
-                           .getObject()
-                           ->net_hidden[location];
-      }
-    } else {
-      auto &in_out = manager->trackLayerInputs(
-        cur_type, l.getName(), l.getInputDimension(), l.getOutputDimension());
-      l.setInputBuffers(in_out);
-    }
-  }
+  status = model_graph.initialize(manager);
```
kparichay, line 171: :100:

kparichay: with this, does manager fit better inside graph than in model class?

test/unittest/compiler/unittest_interpreter.cpp
kparichay: Yes, that is because destructor of manager does not call `deallocateWeights()`, `reset()` does which is called by `NeuralNetwork` destructor. So, if you don't call `deallocateWeights` or `reset` here, the freeing of memory happens strangely.

Why does manager does not call deallocate upon destruction? -> because manager must reset before graph resets as manager just holds the references to weight objects. If manager is updated to hold shared_ptr of weights than references, then constraint can be removed.

kparichay (approved):
LGTM

test/unittest/compiler/unittest_interpreter.cpp
zhoonit: Understood. I guess we should untangle a bit

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Actually, as far as I remember, this is merged in few pr's ahead.

nntrainer/models/neuralnet.cpp
zhoonit: I think so

nntrainer/compiler/tflite_interpreter.cpp
kparichay: got it.

nntrainer/models/neuralnet.cpp
kparichay: then let's do so in a later PR.

test/unittest/compiler/unittest_interpreter.cpp
```diff
@@ -161,16 +161,28 @@ TEST_P(nntrainerInterpreterTest, graphSerializeAfterDeserialize) {
...
-  g->compile(nntrainer::LossType::LOSS_NONE);
+  auto g = makeGraph({fc0, fc1});
+  EXPECT_EQ(g->compile(nntrainer::LossType::LOSS_NONE), ML_ERROR_NONE);
+  EXPECT_EQ(g->initialize(manager), ML_ERROR_NONE);
+
+  manager->initializeWeights();
+  manager->allocateWeights();
+
   interpreter.serialize(g, "test.tflite");
+
```
jijoongmoon, line 184: you will add validation of test.tflite later??

jijoongmoon (approved):
LGTM

test/unittest/compiler/unittest_interpreter.cpp
zhoonit: yes :)

PR #1114, opened 2021-04-14
Handle Uninitialized scalar variable issue

lhs8928:
Set default recurrent_activation_type in parameter


zhoonit (approved):
LGTM

PR #1116, opened 2021-04-14
[Tflite] Implement serializing opcode

zhoonit:
- [Tflite] Implement serializing opcode 

```
This patch implements implementing serialize opcode by
`buildOperatorCodes`

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -250,6 +250,13 @@ public:
     return index2data[index];
   }
 
+  /**
+   * @brief Get the Data object
+   *
+   * @return const std::vector<T>& underlying data
+   */
+  const std::vector<T> &getData() const { return index2data; }
+
 private:
   std::unordered_map<T, unsigned int> data2index; /**< data -> index map */
   std::vector<T> index2data;                      /**< index -> data map */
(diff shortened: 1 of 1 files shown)


nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -331,10 +342,27 @@ buildBuffers(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+  const auto &op_codes = map.getIndexMap<tflite::BuiltinOperator>().getData();
+
+  std::vector<flatbuffers::Offset<tflite::OperatorCode>> fb_op_codes;
+  fb_op_codes.reserve(op_codes.size());
+
+  auto create_op_offset = [&fbb](const tflite::BuiltinOperator &op,
+                                 int32_t version = 1) {
+    tflite::OperatorCodeBuilder builder(fbb);
+    builder.add_deprecated_builtin_code(static_cast<int8_t>(op));
+    /// @todo find reason why version field is not shown
```
kparichay, line 354: what are the possible values of the version? is it possible to get this version from tflite::interpreter?

kparichay (approved):
LGTM

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: It's been a while after I checked the source code but I remember it was hardcoded to 1 in tf sourcefile as well.

kparichay: then lets keep track to their code if they update it, and let it be 1 for now :)

jijoongmoon (approved):
LGTM

PR #1118, opened 2021-04-14
[Tflite] Implement `BuildTensors`

zhoonit:
- [Tflite] Implement `BuildTensors` 

```
This patch add `BuildTensors` which build tensors into *.tflite file.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -252,7 +252,7 @@ public:
     NNTR_THROW_IF(search == key2index.end(), std::invalid_argument)
       << FUNC_TAG << "Cannot find index for key: " << key;
 
-    return *search;
+    return search->second;
   }
 
   /**
@@ -281,14 +281,6 @@ private:
 };
 
 /**
- * @brief Tfbuffer struct (not yet fully specified)
- *
- */
-struct TfBuffer {
-  size_t size;
-};
-
-/**
  * @brief tensorflow operation index map, this class manages operation index
  * mapping
  *
@@ -305,7 +297,9 @@ public:
 
     auto &buffer_map = getIndexMap<const float *, Buffer>();
     buffer_map.addDataWhenNotFound(
-      empty_buffer, {0, empty_buffer}); /// put empty buffer to first
+      nullptr, {0, empty_buffer}); // this represents undefined buffer
+    buffer_map.addDataWhenNotFound(
+      empty_buffer, {0, empty_buffer}); /// this represents empty buffer
 
     auto update_buffers = [&buffer_map](const TfOpNode::Variables &variables) {
       for (auto &variable : variables) {
@@ -363,6 +357,7 @@ TfOpNodes
 buildOpNodes(std::shared_ptr<const GraphRepresentation> representation) {
   TfOpNodes nodes;
   /// @todo, look ahead of layers to get nodes that can be fused
+  /// we will need to have a dedicated builder
   for (const auto &ln : representation->getSorted()) {
     nodes.emplace_back(*ln->getObject());
   }
@@ -376,9 +371,7 @@ buildBuffers(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
     map.getIndexMap<const float *, TfOpIdxMap::Buffer>().getData();
 
   std::vector<flatbuffers::Offset<tflite::Buffer>> fb_buffers;
-  fb_buffers.reserve(buffers.size() + 1);
-  /// add reserved buffer to denote undefined
-  fb_buffers.push_back(tflite::CreateBuffer(fbb));
+  fb_buffers.reserve(buffers.size());
 
   auto create_buffer_offset = [&fbb](const TfOpIdxMap::Buffer &buffer) {
     if (buffer.first == 0) {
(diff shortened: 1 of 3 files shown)


lhs8928 (approved):
LGTM

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,91 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+  }
+
+  return fbb.CreateVector(fb_dimension);
+}
+
+flatbuffers::Offset<flatbuffers::Vector<flatbuffers::Offset<tflite::Tensor>>>
+buildTensors(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
+  /// @todo: the actual (suqeezed) tensor dimension must be known before coming
+  /// here. For now, it is directly guessed for the fc layer
+  const auto &variables = map.getIndexMap<const Var_Grad *>().getData();
```
jijoongmoon, line 449: I wonder that we have buildBuffers with this getData(). and Do we need it for tensor as well?

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,91 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+ * @param type tflite Tensor type
+ * @return flatbuffers::Offset<flatbuffers::Vector<int32_t>>
+ */
+flatbuffers::Offset<flatbuffers::Vector<int32_t>>
+buildTesorShape(const TensorDim &dim, flatbuffers::FlatBufferBuilder &fbb,
+                int effective_dimension, bool is_signature = false) {
+  std::vector<int32_t> fb_dimension;
+
+  /// batch dimension is always considered with signature
+  if ((effective_dimension >> 3) & 1) {
```
jijoongmoon, line 431: Could you add more explanation about effective_dimension?

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: The return of `getData()` here is indexmap of variables. This is not related to buffer :)

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: We have fixed number of dimension, to support flexible dimension. we need to store the squeezed version of dimension somewhere. 

for example(3, 1, 1, 5) can be (3, 5), (3, 1, 5), (3, 1, 1, 5). `effective_dimension` distinguish between those.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -306,7 +298,9 @@ class TfOpIdxMap {
 
     auto &buffer_map = getIndexMap<const float *, Buffer>();
     buffer_map.addDataWhenNotFound(
-      empty_buffer, {0, empty_buffer}); /// put empty buffer to first
+      nullptr, {0, empty_buffer}); // this represents undefined buffer
```
kparichay, line 301: when would this be needed?

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,91 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
   return fbb.CreateVector(fb_op_codes);
 }
 
+/**
+ * @brief get Serialized Tensor shape
+ *
+ * @param dim Tensor dimension to convert
+ * @param type tflite Tensor type
```
kparichay, line 422: update the params?

nntrainer/compiler/tflite_interpreter.cpp
kparichay: can you add as comments the possible values of effective_dimension? 
also what does >> 3 signify - just checking the batch value?

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,91 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+  /// batch dimension is always considered with signature
+  if ((effective_dimension >> 3) & 1) {
+    fb_dimension.push_back(is_signature ? -1 : dim.batch());
+  }
+
+  for (unsigned int i = 1; i < MAXDIM; ++i) {
+    if (((effective_dimension >> (MAXDIM - i - 1)) & 1) == 0) {
+      continue;
+    }
+    fb_dimension.push_back(dim.getTensorDim(i));
```
kparichay, line 439: does the transpose needs to happen here for NCHW/NHWC? if yes, add comment for TODO. If handled else where, add comment that handled else where.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Yes, 3 signifies checking the batch value that will be always added in our case.
Added as a doxygen, this is changed to std::bitset<N> to make it explicit in few pr's ahead.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: This is part of tflite file spec. the first buffer index denotes undefined buffer.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Updated

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: transpose is handled elsewhere, this is where that weird `varGradTranspose` comes into.
This commit doesn't contain that part yet.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,91 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+    /// this is not true, this should be supported inherently from TensorDim
+    /// class.
+    int FIX_ME_FLAG = 0b1001;
+    auto shape = buildTesorShape(var->getDim(), fbb, FIX_ME_FLAG, false);
+    auto shape_sig = buildTesorShape(var->getDim(), fbb, FIX_ME_FLAG, true);
+    auto name = fbb.CreateString(var->getName());
+    auto tensor = var->getVariableRef();
+
+    unsigned int buffer_idx = 1;
+    if (!tensor.uninitialized() && tensor.isAllocated()) {
```
kparichay, line 466: in case, the tensor is not initialized, what difference would it make to the generated output file? this question is not for the weights but for the input/output tensors perspective.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,95 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+
+  std::vector<flatbuffers::Offset<tflite::Tensor>> fb_tensors;
+  fb_tensors.reserve(variables.size());
+
+  auto create_tensor = [&fbb, &buffer_map](const Var_Grad *var) {
+    /// @fixme: There is a fixed assumption that effective_dimension is 0b1001,
+    /// this is not true, this should be supported inherently from TensorDim
+    /// class.
+    int FIX_ME_FLAG = 0b1001;
+    auto shape = buildTesorShape(var->getDim(), fbb, FIX_ME_FLAG, false);
```
kparichay, line 464: buildTesorShape -> buildTensorShape?

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,95 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+    auto shape_sig = buildTesorShape(var->getDim(), fbb, FIX_ME_FLAG, true);
```
kparichay, line 465: can you add some comment as how shape is different from shape_sig here?

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Nice catch 😸

nntrainer/compiler/tflite_interpreter.cpp
kparichay: got it.

nntrainer/compiler/tflite_interpreter.cpp
kparichay: this is fine for now, but later, add a method like `getBatchIndex` or `getBatchLocation` in `TensorDim`, and right shift by that value?

nntrainer/compiler/tflite_interpreter.cpp
kparichay: alright. Please just add a small comment that transpose is already done or will be handled later at ...

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: it will point to a empty buffer (in our case, index of 1). `tflite::interpreter` will regard this as a tensor that needs input from user.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -422,12 +415,95 @@ buildOperatorCodes(const TfOpIdxMap &map, flatbuffers::FlatBufferBuilder &fbb) {
...
+    auto name = fbb.CreateString(var->getName());
+    auto tensor = var->getVariableRef();
+
+    unsigned int buffer_idx = 1;
+    if (!tensor.uninitialized() && tensor.isAllocated()) {
+      buffer_idx = buffer_map.getIndex(var->getVariableRef().getData());
+    }
+
+    tflite::TensorBuilder builder(fbb);
+    builder.add_name(name);
```
kparichay, line 475: this is not related to this PR, but can you check the requirements of the name by tflite and update #1185 correspondingly that we our name satisfy those requirements?

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Yes, Regarding this, I have made some (arguable) changes regarding this in the upcoming pr :)

kparichay: cool.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Added to the doxygen note.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Updated

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Updated

nntrainer/compiler/tflite_interpreter.cpp
kparichay: so, if a model is saved to tflite format before allocation input/output buffers, tflite assumes that all the input/output buffers must be given by the user?

zhoonit: For that reason, I made it not to save buffer(as far as I remember) for in/out, so that it can be inputted from user.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Sure

zhoonit: I'll update the name validator with regard to https://github.com/tensorflow/tensorflow/blob/8ed04bf988ff53e1e8eb40d78876328149fb5deb/tensorflow/python/framework/ops.py#L1922

nntrainer/compiler/tflite_interpreter.cpp
kparichay: > For that reason, I made it not to save buffer(as far as I remember) for in/out, so that it can be inputted from user.

got it.

> so, if a model is saved to tflite format before allocation input/output buffers, tflite assumes that all the input/output buffers must be given by the user?

I am talking the input/output buffers of the intermediate layers.

zhoonit: Ah, right. this will is not required to be filled from user. the operations fill it.

kparichay: So, empty buffers for model inputs/outputs are treated differently than empty buffers per operation inputs/outputs, right?

zhoonit: In terms of list of tensors it is treated the same. `SubGraph` table defines which index is model in and out.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -304,18 +332,23 @@ class TfOpIdxMap {
...
 
+  const float *get_empty_buffer() const { return empty_buffer; }
+
 private:
-  float empty_buffer[0]; /**< unintialized tensor points to this buffer */
+  float
+    empty_buffer[0]; /**< reserved unintialized tensor points to this buffer */
 
-  std::tuple<BidirectionalIndexMap<const float *>, /**< underlying buffer map */
+  std::tuple<BidirectionalIndexMap<const float *, Buffer>,   /**< buffer map */
```
jijoongmoon, line 351: I wonder, Would it be more as we support more layer?

nntrainer/compiler/tflite_interpreter.cpp
jijoongmoon: if (3,5) and effective dimension is 1, 0, 0, 1 .. then it is same with (3, 1,  1, 5)?? or.. it the dimension is (3, 1,1,5), and effective dimension is 0, 1,1, 0?? is it wrong? need to check?

nntrainer/compiler/tflite_interpreter.cpp
jijoongmoon: right.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: From spec, I think there wouldn't be no new map.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: If our dimension is 3, 1, 1, 5 and 1, 0, 0, 1. It means (3, 5) in flatbuffer schema

If our dimension is (3, 1, 1, 5) and (0, 1, 1, 0). This would be an invalid combination but interpreted as (1, 1) in flatbuffer schema

nntrainer/compiler/tflite_interpreter.cpp
kparichay: great, thanks for the explanation.

kparichay (approved):
LGTM

nntrainer/compiler/tflite_interpreter.cpp
jijoongmoon: So.. simply it ignore the dimension which is zero?

kparichay:
@zhoonit Please rebase.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: Yes. I guess we can decide to keep dimension if it is not one in the future.

PR #1119, opened 2021-04-14
[Tflite] Fix tf shape building @open sesame 05/24 16:44

zhoonit:
```
This patch fixes tf shape building, by adding `eff_dim_flag` and
`dyn_dim_flag`.

`eff_dim_flag` checks which dimension are actually used so that to be
squeezed.
`dyn_dim_flag` checks which dimension should be set to -1 (meaning it is
not fixed)

With those flag set, our tensor dim is able to convert to the dimension
representation that supports `-1` and not fixed size

Those fixes are apply to fc layer for now

**Self evaluation:**

```

- ~**[Pending MERGES - #1155]** Merge branch '"'"'@dox/tensordim'"'"' into @1067/tf_interpreter/10_base~

### DIFF
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -43,25 +43,28 @@ int FullyConnectedLayer::initialize(Manager &manager) {
     throw std::invalid_argument("Fully connected layer takes only one input");
   }
 
-  output_dim[0] = input_dim[0];
+  auto &in_dim = input_dim[0];
+  /// @todo fc actaully supports multidimensions. EffDimFlag shouldn't be fixed
+  /// like this.
+  in_dim.setEffDimFlag(0b1001);
+  in_dim.setDynDimFlag(0b1000);
+
+  output_dim[0] = in_dim;
   output_dim[0].width(unit);
 
-  TensorDim bias_dim = TensorDim();
-  bias_dim.setTensorDim(3, unit);
-
-  TensorDim dim = output_dim[0];
-  dim.height(input_dim[0].width());
-  dim.batch(1);
+  TensorDim bias_dim(1, 1, 1, unit, 0b0001);
+  TensorDim weight_dim(1, 1, in_dim.width(), unit, 0b0011);
 
   if (weights.empty()) {
     weights.reserve(2);
-    weights.emplace_back(dim, weight_initializer, weight_regularizer,
+    weights.emplace_back(weight_dim, weight_initializer, weight_regularizer,
                          weight_regularizer_constant, true, false, "FC:weight");
     weights.emplace_back(bias_dim, bias_initializer, WeightRegularizer::NONE,
                          1.0f, true, false, "FC:bias");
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/tensor_dim.h
```diff
@@ -323,13 +360,36 @@ class TensorDim {
...
   /**
    * @brief reset length
    *
    */
   void resetLen();
 
+  std::bitset<MAXDIM> eff_dim_flag; /**< dimension bit flag to define effective
+          dimension size */
+
+  std::bitset<MAXDIM> dyn_dim_flag; /**< dimension bit flag to define
```
kparichay, line 390: Can dimensions other than batch be dynamic? 
Maybe all dimensions can be dynamic but are there are any other possibilities?

zhoonit: Yes, all dimension can be dynamic.

nntrainer/layers/fc_layer.cpp
```diff
@@ -43,25 +43,26 @@ int FullyConnectedLayer::initialize(Manager &manager) {
     throw std::invalid_argument("Fully connected layer takes only one input");
   }
 
-  output_dim[0] = input_dim[0];
-  output_dim[0].width(unit);
+  auto &in_dim = input_dim[0];
+  in_dim.setEffDimFlag(0b1001);
```
jijoongmoon, line 47: It is not always true for rnn and lstm. Effective Dimension could be 1011.

zhoonit: Yes, I'll add a todo. There are some subtle problem regarding this I'd like to discuss later.

nntrainer/tensor/tensor_dim.h
kparichay: If the dynamic dimensions are other than the batch dimension, we must add a way to ensure that the calculation of output must work independent of it, else fail. Please add a TODO related to this.

kparichay (approved):
LGTM

nntrainer/tensor/tensor_dim.h
kparichay: food for thought - set `-1` to the values which are dynamic directly in the dimension. Just like a value of 0 inside dimension means it's not properly set yet, `-1` would mean it's not yet decided (this dimension's actual value would be 1 or more). Using this dimension for calculation purposes will work, but this dimension can't be used to create a Tensor or allocate any memory related to it.

zhoonit: Yeah, I am thinking that the reason they have a dynamic dimension is similar. Having both (effective, dynamic) dimension has advantage that when dimension is not set, there is a explicit default dimension. -1 could be used for our purpose as well.

> If the dynamic dimensions are other than the batch dimension, we must add a way to ensure that the calculation of output must work independent of it, else fail. Please add a TODO related to this.

Added a todo regarding your concern.

jijoongmoon (approved):
LGTM

PR #1121, opened 2021-04-16
[Tflite] Add model input, output

zhoonit:
- [Tflite] Add model input, output

```
This patch add model to track input and output

**Self evaluation:**

```

### DIFF
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -75,11 +75,24 @@ public:
    * fused
    * @param layer layer that is converted to TfOpNode
    */
-  TfOpNode(const Layer &layer) {
+  TfOpNode(const Layer &layer) : is_input(false), is_output(false) {
+    setInOut(layer);
     setInputs(layer.getInputRef());
     setOutputs(layer.getOutputRef());
     setWeights(layer.getWeightsRef());
     setOpType(layer.getType());
+  }
+
+  /**
+   * @brief Set the In Out object
+   *
+   */
+  void setInOut(const Layer &layer) {
+    auto &in = layer.getInputLayers();
+    is_input = std::find(in.begin(), in.end(), "__data__") != in.end();
+
+    auto &out = layer.getOutputLayers();
+    is_output = std::find(out.begin(), out.end(), "__exit__") != out.end();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -329,6 +361,24 @@ class TfOpIdxMap {
...
+      [&variable_map](const TfOpNode::Variables &variables,
+                      std::vector<int> &v) {
+        for (auto &variable : variables) {
+          auto index = variable_map.getIndex(variable);
+          v.push_back(index);
+        }
+      };
+
+    for (auto &op_node : nodes) {
+      if (op_node.isInput()) {
```
jijoongmoon, line 375: isInput is checking if this layer is input layer of network model and isOutput too. so.. It confused little bit. How about change the name such as isInputLayer or so.

jijoongmoon (approved):
LGTM

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -500,9 +526,13 @@ buildSubGraphs(const TfOpNodes &nodes, const TfOpIdxMap &map,
 
   /// @todo extract this to buildSubgraph if there is one or more subgraph
   auto name = fbb.CreateString("main");
+  auto inputs = fbb.CreateVector(map.getInputs());
+  auto outputs = fbb.CreateVector(map.getOutputs());
 
   auto builder = tflite::SubGraphBuilder(fbb);
   builder.add_tensors(tensors);
+  builder.add_inputs(inputs);
+  builder.add_outputs(outputs);
```
jijoongmoon, line 535: does Tflite need this inputs and outputs of model?

zhoonit: Yes, they are identified specially.

nntrainer/compiler/tflite_interpreter.cpp
zhoonit: I think the reason of confusion comes from we are not explicitly identifying tensor but only operation
But tf identifies tensors and operation.
I'll change to `isInputNode()` :)

zhoonit: Updated

lhs8928 (approved):
LGTM

PR #1143, opened 2021-04-28
[ TIME DIST ] Forwarding Implementation

jijoongmoon:
This commit includes
 . Implementation of Forwarding Implementation of Time Dist Layer

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/loss_layer.cpp
+++ b/nntrainer/layers/loss_layer.cpp
@@ -55,7 +55,7 @@ void LossLayer::forwarding(bool training) {
   switch (loss_type) {
   case LossType::LOSS_MSE: {
     // y2 <- y2 - y;
-    hidden_ = y;
+    hidden_.fill(y);
     if (label_exist) {
       Tensor &y2 = net_hidden[0]->getGradientRef();
       Tensor residual = y2.subtract(y);
--- a/nntrainer/layers/time_dist.cpp
+++ b/nntrainer/layers/time_dist.cpp
@@ -60,7 +60,79 @@ int TimeDistLayer::initialize(Manager &manager) {
 }
 
 void TimeDistLayer::forwarding(bool training) {
-  // NYI
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  Tensor hidden_g, h_g;
+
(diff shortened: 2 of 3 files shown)


nntrainer/layers/loss_layer.cpp
```diff
@@ -55,7 +55,7 @@ void LossLayer::forwarding(bool training) {
   switch (loss_type) {
   case LossType::LOSS_MSE: {
     // y2 <- y2 - y;
-    hidden_ = y;
+    hidden_.fill(y);
```
zhoonit, line 58: nice catch :)

nntrainer/layers/time_dist.cpp
```diff
@@ -60,7 +60,79 @@ int TimeDistLayer::initialize(Manager &manager) {
...
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
+  Tensor &input_ = net_input[0]->getVariableRef();
+
+  Tensor hidden_g, h_g;
+
+  TensorDim ho_dim = hidden_.getDim();
+  TensorDim in_dim = input_.getDim();
+
+  // input_.dim = [ 1, b, h, w ] : nntrainer only support 1,2,3 transpose
+  input_.reshape({1, in_dim[0], in_dim[2], in_dim[3]});
```
kparichay, line 72: i think this will have to `getBatchSlixe()` or `getSharedDataTensor()` as reshape requires the new shape and old shape to be of equal size, which is not the case here.

kparichay: I don't understand, how this is reducing the size and line 75 goes back to the original size?

jijoongmoon: the channel is 1. so.. the order of date is same between [ b, 1, h, w ] and [ 1, b, h, w ].  Therefore we just do the reshape for that. and Tensor only support C, H, W transpose. ( Now, It B, H, W ). Now Transpose ( [ 1, H, B, W] ) and reshape it to [ H, 1, B, W ].

lhs8928 (approved):
LGTM

nntrainer/layers/time_dist.cpp
kparichay: thanks for the explanation.

kparichay (approved):
LGTM

PR #1149, opened 2021-04-29
[Android] Add tflite interpreter option

zhoonit:
- [Android] Add tflite interpreter option

```
This patch add tflite interpreter option to ndk build

**Self evaluation:**

```

- [Refactor] Delegate tfNode creation to exporter

```
This patch enables delegating exporting logic to exporter.
Note that this is supposed to be 1:1 mapping, if it has to be m:n
mapping, it should be directly handled by the interpreter, which
actually can see the node.

Also s/`save_result`/`saveResult` and `get_result`/`getResult`

**Self evaluation:**

```

### DIFF
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -3,6 +3,15 @@ LOCAL_PATH := $(call my-dir)
 include $(CLEAR_VARS)
 
 ENABLE_TFLITE_BACKBONE := 1
+ENABLE_TFLITE_INTERPRETER := 1
+
+NEED_TF_LITE := 0
+
+ifeq ($(ENABLE_TFLITE_BACKBONE), 1)
+NEED_TF_LITE := 1
+else ifeq ($(ENABLE_TFLITE_INTERPRETER), 1)
+NEED_TF_LITE := 1
+endif
 
 ifndef NNTRAINER_ROOT
 NNTRAINER_ROOT := $(LOCAL_PATH)/..
@@ -31,7 +40,7 @@ include $(CLEAR_VARS)
 NNTRAINER_JNI_ROOT := $(NNTRAINER_ROOT)/jni
 
 # Build tflite if its backbone is enabled
-ifeq ($(ENABLE_TFLITE_BACKBONE),1)
+ifeq ($(NEED_TF_LITE),1)
 $(warning BUILDING TFLITE BACKBONE !)
 TENSORFLOW_VERSION := 2.3.0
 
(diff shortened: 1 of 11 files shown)


lhs8928 (approved):
LGTM

jni/Android.mk
```diff
@@ -3,6 +3,15 @@ LOCAL_PATH := $(call my-dir)
 include $(CLEAR_VARS)
 
 ENABLE_TFLITE_BACKBONE := 1
+ENABLE_TFLITE_INTERPRETER := 1
+
+NEED_TF_LITE := 0
+
+ifeq ($(ENABLE_TFLITE_BACKBONE), 1)
+NEED_TF_LITE := 1
```
jijoongmoon, line 11: how about..
``` 
ifeq ($(ENABLE_TFLITE_BACKBONE) $(ENABLE_TFLITE_INTERPRETER), 1)
NEED_TF_LITE :=1
endif
```

zhoonit: Tested on my local machine, `$(ENABLE_TFLITE_BACKBONE) $(ENABLE_TFLITE_INTERPRETER)` evaluates to `1 1` so `ifeq` does not work :(

kparichay (approved):
LGTM

jni/Android.mk
jijoongmoon: is it? they say it is working.

zhoonit: I think I need to understand the code. let me improve this later, thank you for the suggestion!

nntrainer/utils/node_exporter.h
```diff
@@ -106,24 +145,83 @@ class Exporter {
...
+
+class FullyConnectedLayer;
+/**
+ * @copydoc template <typename PropsType, typename NodeType> void
+ * Exporter::saveTflResult(const PropsType &props, const NodeType *self);
+ */
+template <>
+void Exporter::saveTflResult(const std::tuple<props::Unit> &props,
+                             const FullyConnectedLayer *self);
+#endif
```
jijoongmoon, line 223: It will be bigger than this. I think it would be better to manage these functions in a separate file. then we could remove ifdef in the code as well.

jijoongmoon (approved):
LGTM except for small suggestion.

nntrainer/utils/node_exporter.h
zhoonit: Yes, I'll make sure this be a seperate file :)

PR #1150, opened 2021-04-29
[Tensor] Add `transpose` with out

zhoonit:
- [Tensor] Add `transpose_i` 

```
This patch creates transpose_i, transposes(out) to manipulate given tensor
With exhaustive list of test

**Self evaluation:**

```


nntrainer/tensor/tensor.h
```diff
@@ -504,7 +504,22 @@ class Tensor {
...
+   * @param direction to transpose ex) 0:2:1
+   * @param[out] Tensor to saved to
+   * @retval Tensor& reference to the out
+   */
+  Tensor &transpose(const std::string &direction, Tensor &out) const;
+
+  /**
+   * @brief Transpose a tensor and put it back to @a this
+   *
+   * @param direction to transpose ex) 0:2:1
```
kparichay, line 520: i highly recommend you to add a note here, that this is NOT an in-place operation. It just returns the result in-place.

nntrainer/tensor/tensor.h
```diff
@@ -504,7 +504,22 @@ class Tensor {
...
+   */
+  void transpose_i(const std::string &direction);
```
kparichay, line 522: i highly recommend you to add a note here, that this is NOT an in-place operation. It just returns the result in-place.

kparichay: infact, unless really needed, lets make this private. This gives a false notion of in-place transpose operation while is it not.

zhoonit: Transpose cannot be non-copyable, from behavior point of view, it is what `_i` does. Can you suggest different name?

kparichay: How about allow create `Tensor &Tensor::transpose(const std::string &direction, Tensor &out)`, where out can be same as `this`, and this function can call `transpose_i` internally.

kparichay:
Please consider rebasing this PR directly to main, as IMO, it does not depend on other PRs and can be merged faster.

nntrainer/tensor/tensor.h
zhoonit: Sure, I'll change that way.

zhoonit: Done in a new commit. removed `transpose_i`

zhoonit:
Closing this to try automated merge pr.

PR #1157, opened 2021-04-30
[VarGrad] Add clone Transpose

zhoonit:
- [VarGrad] Add clone Transpose

```
This patch adds clone transpose to `var_grad`, this function clones the
current `var_grad` but transpose and save

**Self evaluation:**

```

### DIFF
--- a/nntrainer/tensor/var_grad.cpp
+++ b/nntrainer/tensor/var_grad.cpp
@@ -14,6 +14,8 @@
 #include <util_func.h>
 #include <var_grad.h>
 
+#include <nntrainer_error.h>
+
 namespace nntrainer {
 
 Var_Grad::Var_Grad(const TensorDim &dim, bool train, bool alloc_now_,
@@ -55,6 +57,21 @@ void Var_Grad::setTrainable(bool train) {
     bool alloc_now_ = var->isAllocated();
     grad = std::make_shared<Tensor>(var->getDim(), alloc_now_);
   }
+}
+
+Var_Grad
+Var_Grad::cloneTransposeVariableOnly(const std::string &direction) const {
+  Var_Grad vg(*this);
+  /// @todo: make this clonable even when not allocated
+  NNTR_THROW_IF(var->isAllocated() == false, std::invalid_argument)
+    << "transpose clone is only allowed when variable is allocated, var name: "
+    << getName();
+
+  vg.var = std::make_shared<Tensor>(var->transpose(direction));
+  vg.dim = vg.var->getDim();
+  vg.grad.reset();
+
+  return vg;
 }
 
 } // namespace nntrainer
--- a/nntrainer/tensor/var_grad.h
+++ b/nntrainer/tensor/var_grad.h
@@ -167,11 +167,24 @@ public:
    */
   Var_Grad clone() const {
     Var_Grad vg(*this);
+
+    /// @fixme even if var is not allocated, cloned var will have allocated
+    /// memory
     vg.var = std::make_shared<Tensor>(this->var->clone());
     vg.grad = std::make_shared<Tensor>(this->grad->clone());
 
     return vg;
   };
+
+  /**
+   * @brief transpose and clone variable, gradient is set to null here
+   * @warning returned var_grad has broken invariant, so the gradient should
+   * never be used
+   *
+   * @param direction direction to transpose
+   * @return Var_Grad new var_grad
+   */
+  Var_Grad cloneTransposeVariableOnly(const std::string &direction) const;
 
   /**
    * @brief Reset the variable and gradient


nntrainer/tensor/var_grad.cpp
```diff
@@ -57,4 +59,20 @@ void Var_Grad::setTrainable(bool train) {
   }
 }
 
+Var_Grad Var_Grad::cloneTranspose(const std::string &direction) const {
+  Var_Grad vg(*this);
+  NNTR_THROW_IF(var->isAllocated() == false, std::invalid_argument)
+    << "transpose clone is only allowed when variable is allocated, var name: "
+    << getName();
+
+  vg.var = std::make_shared<Tensor>(var->transpose(direction));
```
kparichay, line 68: var_grad will later have to store the transpose direction and do the transpose later when memory is allocated (once the clone is fixed).

kparichay: can you also please elaborate a bit where this will be used.

zhoonit: > var_grad will later have to store the transpose direction and do the transpose later when memory is allocated (once the clone is fixed).

I'll add todo

> can you also please elaborate a bit where this will be used.

This function is used for tflite convert. When converting to tflite, some layers need to be transposed. (I found that even fclayer needs transpose)

zhoonit: Adding todo done.

kparichay: > This function is used for tflite convert. When converting to tflite, some layers need to be transposed. (I found that even fclayer needs transpose)

I hope you mean the weights of the layers need transpose. In that case, IMO, transposing the gradients for storing the weights of the layer might be unnecessary.

zhoonit: Yes, you are right. This is to maintain `clone` semantics. In my usage, if grad is allocated, it will be cloned but destroyed right away.

I could add a boolean to pass grad copy but I'd like to minimize human error caused from too many states:whale: as of now, living with a negligible performance overhead.

:whale: would include how to copy initialized but not allocated variable, or uninitialized var copy, or skip grad but clone again with grad, things like that.

If we need to optimize this bit, this can be done easily anyway.
```
if (grad->isAllocated()) {
    vg.grad = std::make_shared<Tensor>(grad->transpose(direction));
  }
```

kparichay: When a model is training, grad will always be allocated. So, I am not sure how this can be optimized.
As the model file will not have to store the gradients, you might want to consider transposing the `weight` tensor (var_grad->var) than transposing the whole `weight` var_grad.

zhoonit: We can pass `cloneTranspose(bool no_grad = true)` if we need to optimize this function.

I am tracking `var_grad *` to do variable check. There is no good way to tell if weights  (or in_outs) are shared without tracking var_grad * as of now.

jijoongmoon: for now, every calculation in layer, we deal with var and grad separately. We use var_gad as an container, not the variables to calculate. If we are going to use it in the calculation, we need support every combination of them. Moreover, if we consider memory optimization of gradient, then.. there will be more complicated cases. Also .. it's getting complicated already.. for example, in the case of cloneTranspose(false), we do not know what the false means if we didn't look at the definition of function. We have similar cases in tensor already (like xs.dot(a, true, false)). So.. I'm not sure about this.

zhoonit: I agree. That's the reason I didn't implement that way. I think this can be done later after we sort things out more clearly.

kparichay: IMO, as saving this related to saving the weights, it is best to support `cloneTranspose()` for `Weight` class than the base class `Var_Grad`. In case of `Weight` class, we can use default `bool no_grad = true` to just transpose weights, and not its gradients. As weights are never shared as of now, we don't need bout any optimization. This also explicitly does not allow transposing `Var_Grad` objects (which can be inputs and outputs).

zhoonit: Another issue is that if we transpose only a variable, the invariant weight or var_grad is broken.

I think we can take those kind of risk if we have to. But I don't think there is a reason to do that as of now adding more complexity. And the behavior is related to VarGrad so wonder the reason why suggesting to move this to weight?

kparichay: > And the behavior is related to VarGrad so wonder the reason why suggesting to move this to weight?

Because the saving is to be done only for the weights.

zhoonit: I'll create `getTransposedVariable` instead.

zhoonit: Done

kparichay (approved):
LGTM

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1159, opened 2021-05-03
[Props] Separate base_propeties to a source

zhoonit:
- [Props] Separate base_propeties to a source

```
This patch separate base_properties into some specialization to source
file

The purpose is to save compile time and redundant doxygen, and clean up
length header.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

```

PR #1162, opened 2021-05-03
[Wait for #1159] [Props] Add vector props

zhoonit:
- [Props] Add vector props

```
This patch add vector property implementation.
Note that there is only unsigned int for now, this can be generalized
later.

**Self evaluation:**

```


zhoonit:
Changed the design, let me close this and reopen to prevent unnecessary review

PR #1163, opened 2021-05-04
Fix/svace_issue_210504

lhs8928:
Handle NO_EFFECT issue

Delete not effective condition
The type of variable i is unsigned int so i < 0 always false

resolves: 461267, 461268


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1165, opened 2021-05-04
[Props] Add vector/array handler

zhoonit:
- [Props] Add vector/array handler

```
This patch add vector property handler.

vectored property have sementics that number of size is not defined
arrayed property have sementics that number of size must match

**Self evaluation:**

```

### DIFF
--- a/nntrainer/utils/base_properties.h
+++ b/nntrainer/utils/base_properties.h
@@ -9,9 +9,13 @@
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
  * @bug No known bugs except for NYI items
  */
-#include <nntrainer_error.h>
+#include <array>
+#include <sstream>
 #include <string>
-#include <tensor_dim.h>
+#include <vector>
+
+#include <nntrainer_error.h>
+#include <parse_util.h>
 
 #ifndef __BASE_PROPERTIES_H__
 #define __BASE_PROPERTIES_H__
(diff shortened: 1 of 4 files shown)


jijoongmoon (approved):
LGTM

nntrainer/utils/base_properties.h
```diff
@@ -9,8 +9,13 @@
...
  * @bug No known bugs except for NYI items
  */
-#include <nntrainer_error.h>
+#include <array>
+#include <sstream>
 #include <string>
+#include <vector>
+
+#include <nntrainer_error.h>
+#include <parse_util.h>
```
kparichay, line 18: is it possible to do without including it here in the header?

test/unittest/unittest_properties.cpp
```diff
@@ -20,7 +22,7 @@
 #include <util_func.h>
 
 namespace { /**< define a property for testing */
-
+//
```
kparichay, line 25: intended?

nntrainer/utils/base_properties.h
```diff
@@ -252,13 +310,13 @@ template <> class Property<std::string> {
  *
  * struct custom_tag: int_prop_tag {};
  *
- * using tag_type = tag_cast<custom_tag, vector_prop_tag>::type
- * static_assert(<std::is_save_v<tag_type, custom_tag> == false);
+ * using tag_type = tag_cast<custom_tag, double_prop_tag>::type
```
kparichay, line 313: this is unrelated to this PR, but later, let's update double to float.

nntrainer/utils/base_properties.h
zhoonit: split function is used from the util. This can be avoided by copy-pasting this to base_properties.cpp, https://github.com/nnstreamer/nntrainer/issues/1063 updated.

test/unittest/unittest_properties.cpp
zhoonit: deleted

nntrainer/utils/base_properties.h
zhoonit: updated

kparichay (approved):
LGTM

PR #1173, opened 2021-05-10
[Permute] Add permute layer forward/backward

zhoonit:
- [Permute] Add permute layer forward/backward

```
This patch implements forward/backward with a following test

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/permute_layer.cpp
+++ b/nntrainer/layers/permute_layer.cpp
@@ -22,8 +22,8 @@
 
 namespace nntrainer {
 
-bool props::Direction::isValid(const unsigned int &value) const {
-  return value <= 2;
+bool props::PermuteDims::isValid(const unsigned int &value) const {
+  return 0 < value && value <= 3;
 }
 
 const std::string PermuteLayer::type = "permute";
@@ -36,9 +36,9 @@ const std::string PermuteLayer::type = "permute";
  * @return const std::string string to return
  */
 static std::string
-buildTrasposeString(const std::array<props::Direction, 3> &arr) {
+buildTrasposeString(const std::array<props::PermuteDims, 3> &arr) {
   std::stringstream ss;
-  ss << arr[0].get() << ':' << arr[1].get() << ':' << arr[2].get();
+  ss << arr[0].get() - 1 << ':' << arr[1].get() - 1 << ':' << arr[2].get() - 1;
   return ss.str();
 }
 
@@ -47,8 +47,8 @@ int PermuteLayer::initialize(Manager &manager) {
     std::bitset<3> check_transpose; /**< check if transpose contains all axis */
 
     for (int i = 0; i < 3; ++i) {
-      check_transpose.set(direction[i], true);
-      this->reverse_direction[direction[i]].set(i);
+      check_transpose.set(direction[i] - 1, true);
+      this->reverse_direction[direction[i] - 1].set(i + 1);
     }
 
     NNTR_THROW_IF(check_transpose.all() == false, std::invalid_argument)
@@ -76,10 +76,18 @@ int PermuteLayer::initialize(Manager &manager) {
   return ML_ERROR_NONE;
 }
 
-void PermuteLayer::forwarding(bool training) { /** NYI */
+void PermuteLayer::forwarding(bool training) {
+  auto &input_ = net_input[0]->getVariableRef();
+  auto &hidden_ = net_hidden[0]->getVariableRef();
+
+  input_.transpose(direction_str, hidden_);
 }
 
-void PermuteLayer::calcDerivative() { /** NYI */
+void PermuteLayer::calcDerivative() {
+  auto &input_grad = net_input[0]->getGradientRef();
+  auto &hidden_grad = net_hidden[0]->getGradientRef();
+
+  hidden_grad.transpose(rdirection_str, input_grad);
 }
 
 void PermuteLayer::copy(std::shared_ptr<Layer> l) {
--- a/nntrainer/layers/permute_layer.h
+++ b/nntrainer/layers/permute_layer.h
@@ -23,11 +23,11 @@ namespace nntrainer {
 
 namespace props {
 /**
- * @brief Direction property, direction property describes the axis to be
+ * @brief PermuteDims property, direction property describes the axis to be
  * transposed. to be used with array
  *
  */
-class Direction : public nntrainer::Property<unsigned int> {
+class PermuteDims : public nntrainer::Property<unsigned int> {
 public:
   static constexpr const char *key = "direction"; /**< unique key to access */
   using prop_tag = uint_prop_tag;                 /**< property type */
@@ -125,8 +125,8 @@ private:
   std::string rdirection_str; /**< transpose representation, @todo
                                         deprecate this */
 
-  std::array<props::Direction, 3> direction;
-  std::array<props::Direction, 3> reverse_direction;
+  std::array<props::PermuteDims, 3> direction;
+  std::array<props::PermuteDims, 3> reverse_direction;
 };
 
 } // namespace nntrainer
--- a/test/unittest/unittest_nntrainer_layers.cpp
+++ b/test/unittest/unittest_nntrainer_layers.cpp
@@ -32,6 +32,7 @@
 #include <nntrainer_test_util.h>
 #include <optimizer_factory.h>
 #include <parse_util.h>
+#include <permute_layer.h>
 #include <pooling2d_layer.h>
 #include <preprocess_flip_layer.h>
 #include <preprocess_translate_layer.h>
@@ -2433,7 +2434,7 @@ TEST_F(nntrainer_RNNLayer, backwarding_01_p) {
 }
 
 /**
- * @brief nntainer RNN Layer for test
+ * @brief nntainer LSTM Layer for test
  */
 class nntrainer_LSTMLayer
   : public nntrainer_abstractLayer<nntrainer::LSTMLayer> {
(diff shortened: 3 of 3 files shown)


nntrainer/utils/base_properties.h
```diff
@@ -20,30 +24,63 @@
...
+using remove_cv_ref_t = std::remove_cv_t<std::remove_reference_t<T>>;
 /**
- * @brief property tag to specialize functions based on this
+ * @brief property info to specialize functions based on this
  *
  * @tparam T property type
  */
-template <typename T> struct prop_tag { using type = typename T::prop_tag; };
+template <typename T> struct prop_info {
+  using prop_type = remove_cv_ref_t<T>;
```
jijoongmoon, line 35: Add more comments here? what is prop_type, tag_type, data_type with example?

nntrainer/layers/permute_layer.h
```diff
@@ -0,0 +1,97 @@
...
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   */
+  int initialize(Manager &manager) override;
+
+  /**
+   * @copydoc Layer::export_to(Exporter &exporter, ExportMethods method)
+   */
+  void export_to(
+    Exporter &exporter,
```
jijoongmoon, line 105: is it for the generate ini?

nntrainer/layers/permute_layer.cpp
```diff
@@ -20,9 +22,57 @@
 
 namespace nntrainer {
 
+bool props::Direction::isValid(const unsigned int &value) const {
```
jijoongmoon, line 25: is it direction or axis?? which one is better?

nntrainer/layers/permute_layer.h
```diff
@@ -14,11 +14,34 @@
...
 
 #include <array>
+#include <string>
 
 #include <base_properties.h>
 #include <layer_internal.h>
 #include <node_exporter.h>
 namespace nntrainer {
+
+namespace props {
```
jijoongmoon, line 24: Are we going to have this props namespace? if it is we need to think more. More code is coming.. so.. maybe we need more namespace like layer.. and optimizer.

nntrainer/layers/permute_layer.h
zhoonit: This is to generate ini, tflite, get properties.

nntrainer/layers/permute_layer.h
zhoonit: props namespace are used to explicitly tell it is derived from a property class and can be dealt as other properties are used.
This namespace is used inside `common_properties.h` `base_properties.h` and some custom properties.

```cpp
   props::Name name; // we can explicitly know from the code it is property, and we can expect some common functions like to_string, from_string with this class.
```

vs 
```cpp
  Name name; // we don't know it's a class or a property-ish class.
```
I think it's a policy we can decide. I wonder if we should not use namespace for the props?

nntrainer/utils/base_properties.h
zhoonit: Updated docs and added test code for clarity :)

nntrainer/layers/permute_layer.cpp
zhoonit: Thanks for the suggestion. Changed to `PermuteDims` referring from keras properties :)

nntrainer/layers/permute_layer.h
jijoongmoon: No. I'm saying, If we need, we have to use it other places to maintain consistency.

test/unittest/unittest_base_properties.cpp
```diff
@@ -96,6 +96,19 @@ TEST(BasicProperty, tagCast) {
   }
 }
 
+TEST(BasicProperty, propInfo) {
+  { /**< prop_info test */
+    using prop_type = nntrainer::prop_info<QualityOfBanana>::prop_type;
```
jijoongmoon, line 101: Let's say differently. for the property input_shape.
then
1. prop_type is input_shape.
2. tag_type is dimension
3. data_type is string 

??

jijoongmoon (approved):
LGTM

test/unittest/unittest_base_properties.cpp
zhoonit: Yes, that is current behavior 👍. For `#2`, this is left to incorporate with different property parser. (eg, dimension type for ini but maybe `string` type for tflite.)

nntrainer/layers/permute_layer.h
zhoonit: `common_properties.h` has props shared across layers.
and each layer can have its own properties. Should we not use namespace for that?

nntrainer/layers/permute_layer.cpp
```diff
@@ -76,10 +76,18 @@ int PermuteLayer::initialize(Manager &manager) {
...
+  auto &input_ = net_input[0]->getVariableRef();
+  auto &hidden_ = net_hidden[0]->getVariableRef();
+
+  input_.transpose(direction_str, hidden_);
 }
 
-void PermuteLayer::calcDerivative() { /** NYI */
+void PermuteLayer::calcDerivative() {
+  auto &input_grad = net_input[0]->getGradientRef();
+  auto &hidden_grad = net_hidden[0]->getVariableRef();
```
lhs8928, line 88: In the backwarding I think the hidden_grad should be net_hidden[0]->getGradientRef() rather than getVariableRef(), isn't it?

zhoonit: Nice catch! this should be getGradientRef(), thanks :)

lhs8928 (approved):
LGTM

nntrainer/layers/permute_layer.cpp
kparichay: @zhoonit Please update this before merge.

nntrainer/layers/permute_layer.cpp
```diff
@@ -36,7 +36,7 @@ const std::string PermuteLayer::type = "permute";
  * @return const std::string string to return
  */
 static std::string
-buildTrasposeString(const std::array<props::Direction, 3> &arr) {
+buildTrasposeString(const std::array<props::PermuteDims, 3> &arr) {
   std::stringstream ss;
   ss << arr[0].get() << ':' << arr[1].get() << ':' << arr[2].get();
```
kparichay, line 41: not directly related to this PR, but consider first getting unique values from this array and then applying the transpose operation. Also, the current implementation expects exactly 3 dimensions to be set for permute, but even two dimensions should be enough to use the permute layer.

zhoonit: You mean applying array directly? Yes, this should better be supported.

Since we have fixed number of dimension(4), I am concerning that the exact semantics of having two dimension, If this is solid, I think we can apply for two dimension.

kparichay: Specifying 3 dimensions for transposing 2 dimensions is an inconvenience. I will approve for this, this can be considered as TODO. Also, there needs to a way to ensure that all the values are unique.

zhoonit: > But, there needs to a way to ensure that all the values are unique

Could you check out if this is what you've stated? https://github.com/nnstreamer/nntrainer/blob/6cf6dd2844b4fd6bb49b2728ab06af2432ed60e4/nntrainer/layers/permute_layer.cpp#L47

> Specifying 3 dimensions for transposing 2 dimensions is an inconvenience.

afaik, for other frameworks it is forced that `rank(shape)` `len(direction)` must match, as we have semantics that `rank(shape) == 3`, I think we need stronger assumption or semantics to make it clear.

For example, given ​shape `2:1:4
if 2:1 is given, possible interpretations are
  1. `1 <-> 4`, because h is 1 when starting from channel
  1. invalid, because height 1 means there is height does not count.
  1. invalid, because `len(direction) == 2` must start from zero

if 0:1 is given, possible interpretations are
  1. `2 <-> 1` starting from channel
  1. `1 <-> 4` starting from height (but invalid because channel is > 2)
  1.  `2 <-> 4` starting from channel but height is 1 so it is not counted.

If we have solid interpretation considering we have fixed number of dimensions I think we could support num 2.

kparichay: > Could you check out if this is what you've stated?

Thanks for the link. I missed this. For later, you may consider moving such constraints to setProperty().

Agreed on the part that 2 dimensions might lead to confusion. 
However, this raises question (outside the scope of this PR), dimension `0:1:2` maps to `c:h:w` and in some cases, `0:1:2:3` maps to `b:c:h:w`. Let's make it consistent.

kparichay (approved):
LGTM

nntrainer/layers/permute_layer.cpp
zhoonit: >  this raises question

Yes, for now, we have firm semantics it is always `c:h:w` and current behavior (ignoring batch dimension) coherent with keras `PermuteLayer`. but `tf.transpose` will start from batch. 


Please throw me ideas if any.

My current thinking is transpose in the near future, so we might be needing two semantics (with different name I guess :p)

Or we can consider it as one semantic, i guess, if we never allow putting batch dimension inside *.ini or property.

Thank you for raising the issue :)

kparichay: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Permute
tf.keras.layers.permute starts dimensions with 1 (I think they left 0 for batch). 
Let's follow these semantics to avoid confusion.

zhoonit: Thanks, made a change to accept 1~3

PR #1178, opened 2021-05-10
fix/coverity_issue_210510

lhs8928:
Handle PARAMETER_HIDDEN, NULL_RETURNS issue

1. Check return value of malloc function
2. Rename function parameter of lambda function

resolve: 1216568, 12126575, 1222077, 1222080, 1227326, 1227327


jijoongmoon (approved):
LGTM

PR #1179, opened 2021-05-11
[ LSTM ] Backwarding Implementation

jijoongmoon:
This commit includes
 . calDerivative & calGradient of LSTM Layer
 . Modification of Forwarding of LSTM Layer

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/lstm.cpp
+++ b/nntrainer/layers/lstm.cpp
@@ -170,7 +170,7 @@ void LSTMLayer::forwarding(bool training) {
 
     for (unsigned int t = 0; t < islice.height(); ++t) {
       Tensor xs =
-        input_.getSharedDataTensor({islice.width()}, t * islice.width());
+        islice.getSharedDataTensor({islice.width()}, t * islice.width());
       hs = oslice.getSharedDataTensor({oslice.width()}, t * oslice.width());
       cs = cell.getSharedDataTensor({cell.width()}, t * cell.width());
       Tensor fgio_t =
@@ -194,15 +194,15 @@ void LSTMLayer::forwarding(bool training) {
       Tensor hi = fgio_t.getSharedDataTensor({unit}, unit * 2);
       Tensor ho = fgio_t.getSharedDataTensor({unit}, unit * 3);
 
-      recurrent_acti_func.run_fn(hf, hf);
-      recurrent_acti_func.run_fn(hi, hi);
-      recurrent_acti_func.run_fn(ho, ho);
-      acti_func.run_fn(hg, hg);
+      acti_func.run_fn(hf, hf);
+      acti_func.run_fn(hi, hi);
+      acti_func.run_fn(ho, ho);
+      recurrent_acti_func.run_fn(hg, hg);
 
       hf.multiply(cs_prev, cs);
-      cs.add_i(hg.multiply_i(hi));
+      cs.add_i(hg.multiply(hi));
 
-      acti_func.run_fn(cs, hs);
+      recurrent_acti_func.run_fn(cs, hs);
       hs.multiply_i(ho);
     }
     // size of h_prev and hs size is same : unit.
@@ -244,18 +244,20 @@ void LSTMLayer::calcGradient() {
   Tensor &hidden_ = net_hidden[0]->getVariableRef();
   Tensor &input_ = net_input[0]->getVariableRef();
   Tensor &m_cell_ = mem_cell->getVariableRef();
+  Tensor &dm_cell_ = mem_cell->getGradientRef();
 
   Tensor dh_nx = Tensor({derivative_.width()});
   Tensor dc_nx = Tensor({derivative_.width()});
 
   for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
     Tensor deriv_t = derivative_.getBatchSlice(b, 1);
+    Tensor derivc_t = dm_cell_.getBatchSlice(b, 1);
     Tensor xs_t = input_.getBatchSlice(b, 1);
     Tensor hs_t = hidden_.getBatchSlice(b, 1);
     Tensor cs_t = m_cell_.getBatchSlice(b, 1);
 
-    dh_nx.setZero();
     dc_nx.setZero();
+    dh_nx.setZero();
 
     Tensor dh;
     Tensor xs;
@@ -269,6 +271,8 @@ void LSTMLayer::calcGradient() {
 
     for (unsigned int t = deriv_t.height(); t-- > 0;) {
       dh = deriv_t.getSharedDataTensor({deriv_t.width()}, t * deriv_t.width());
+      dc =
+        derivc_t.getSharedDataTensor({derivc_t.width()}, t * derivc_t.width());
       xs = xs_t.getSharedDataTensor({xs_t.width()}, t * xs_t.width());
       hs = hs_t.getSharedDataTensor({hs_t.width()}, t * hs_t.width());
       cs = cs_t.getSharedDataTensor({cs_t.width()}, t * cs_t.width());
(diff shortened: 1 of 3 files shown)


nntrainer/layers/lstm.h
```diff
@@ -150,6 +150,7 @@ class LSTMLayer : public Layer {
   Tensor c_prev;
 
   std::shared_ptr<Var_Grad> mem_cell;
+  std::shared_ptr<Var_Grad> fgio;
```
kparichay, line 153: add some description? also for my understanding, this stores the output of the intermediate gates in the LSTM layer, right?

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
...
+    weightAt(static_cast<int>(LSTMParams::weight_xh)).getGradientRef();
+  Tensor &djdw_h =
+    weightAt(static_cast<int>(LSTMParams::weight_hh)).getGradientRef();
+  Tensor &djdb_h =
+    weightAt(static_cast<int>(LSTMParams::bias_h)).getGradientRef();
+  Tensor &weight_hh =
+    weightAt(static_cast<int>(LSTMParams::weight_hh)).getVariableRef();
+
+  Tensor &derivative_ = net_hidden[0]->getGradientRef();
+  Tensor &hidden_ = net_hidden[0]->getVariableRef();
```
kparichay, line 244: consider updating net_hidden to net_output someday as now it confuses with the hidden state of RNN and LSTM :).

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
 }
 
 void LSTMLayer::calcDerivative() {
-  // NYI
+  Tensor &derivative_ = fgio->getGradientRef();
+  Tensor &weight =
+    weightAt(static_cast<int>(LSTMParams::weight_xh)).getVariableRef();
+  Tensor &ret_ = net_input[0]->getGradientRef();
+  derivative_.dot(weight, ret_, false, true);
```
kparichay, line 230: is this complete?

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
 }
 
 void LSTMLayer::calcDerivative() {
-  // NYI
+  Tensor &derivative_ = fgio->getGradientRef();
```
kparichay, line 226: can you add a check here that this calculation was done in calcGradient()?
If an LSTM layer is set to be non-trainable, then this will break.

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
...
+
+      dc.multiply(cs_prev, dhf);
+      dc.multiply(hg, dhi);
+      dc.multiply(hi, dhg);
+      dc.multiply(hf, dc_nx);
+
+      acti_func.run_prime_fn(ho, dho, dho);
+      acti_func.run_prime_fn(hf, dhf, dhf);
+      acti_func.run_prime_fn(hi, dhi, dhi);
+      recurrent_acti_func.run_prime_fn(hg, dhg, dhg);
```
zhoonit, line 325: Could you check if this is okay?

`acti_func.run_prime_fn(a, b, b)`
```cpp
prim_fn(x, y, z) {
1: run_prim_fn(x, y)
2: return y * z; 
}
```
y is changed at line 1, in this case z is same as y so the data is changed.

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
...
+
+      float alpha = 1.0;
+      if (b != 0) {
+        alpha = 0.0;
+      }
+
+      djdb_h.add_i(dfgio_t, alpha);
+      djdw_x.add_i(xs.dot(dfgio_t, true, false), alpha);
```
zhoonit, line 333: Could you check if below is equivalent? 

```suggestion
      xs.dot(dfgio_t, djdw_x, true, false, alpha);
```

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
...
+      djdw_h.add_i(hs_prev.dot(dfgio_t, true, false), alpha);
+      dfgio_t.dot(weight_hh, dh_nx, false, true, 1.0);
```
zhoonit, line 335: I wonder the `beta` value is always 1.0? or maybe below?
```suggestion
      dfgio_t.dot(weight_hh, dh_nx, false, true, alpha);
```

nntrainer/layers/lstm.cpp
zhoonit: I checked this is done in `calcGradient()` I think if there is a dependency between those two, everything should be inside `calcDerivative`?

nntrainer/layers/lstm.cpp
jijoongmoon: yes. just like fully connected layer.

nntrainer/layers/lstm.cpp
jijoongmoon: yes. it is. but actually they are the same.

nntrainer/layers/lstm.cpp
zhoonit: From the offline discussion, I think this is beyond the scope of this PR. 

Since z is passed as `const` I think we shouldn't allow changing z. Let's deal this problem later :)

nntrainer/layers/lstm.cpp
```diff
@@ -218,11 +223,118 @@ void LSTMLayer::copy(std::shared_ptr<Layer> l) {
...
+  Tensor dc_nx = Tensor({derivative_.width()});
+
+  for (unsigned int b = 0; b < input_dim[0].batch(); ++b) {
+    Tensor deriv_t = derivative_.getBatchSlice(b, 1);
+    Tensor derivc_t = dm_cell_.getBatchSlice(b, 1);
+    Tensor xs_t = input_.getBatchSlice(b, 1);
+    Tensor hs_t = hidden_.getBatchSlice(b, 1);
+    Tensor cs_t = m_cell_.getBatchSlice(b, 1);
+
+    dh_nx.setZero();
```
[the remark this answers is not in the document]
jijoongmoon: done.

nntrainer/layers/lstm.cpp
jijoongmoon: Checked. It's Ok.

nntrainer/layers/lstm.cpp
jijoongmoon: I fixed.

nntrainer/layers/lstm.cpp
jijoongmoon: they are equivalent. but I will change after more unittest cases.

nntrainer/layers/lstm.h
jijoongmoon: You are right. Added. Thanks.

nntrainer/layers/lstm.cpp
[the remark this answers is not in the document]
zhoonit: Okay

zhoonit (approved):
LGTM

nntrainer/layers/lstm.cpp
kparichay: Please add a TODO here to add a check later to support  calcDerivative when this layer is non-trainable,

kparichay: Added as a new issue #1226 as discussed offline.

kparichay (approved):
LGTM

Issue #1176, opened 2021-05-10
VOC from AIC : requirements


kparichay:
1. [x] Concat layer support for INI and test the concat layer in a model @zhoonit 
   1. input_layers = A + B (current support only for A, B) (delayed)
   2. update realizeMultiInputType() (done)
   3. Test in a model (done)
   4. Maybe add unittest as well (done)

2. [x]  Model to simulate multiple inputs
    1. Combine the inputs  in the dataset and split  the inputs in the models
    2. corresponding unittests and app tests.

Moved generic multi inputs/outputs support to #1253 as the current requirements are completed with simulating multiple inputs for the model with a split layer.

kparichay:
@jijoongmoon Please check this once.

jijoongmoon:
We need to check if they need multiple datasets.

zhoonit:
Related issue https://github.com/nnstreamer/nntrainer/issues/239

kparichay:
@zhoonit Is concat layer part done, including support of  `input_layers=A,B` will insert a concat layer with A,B?

zhoonit:
> @zhoonit Is concat layer part done, including support of `input_layers=A,B` will insert a concat layer with A,B?

Concat is verified and supported for their requirement.

There are number of things to be done to complete this, if we are not duck tapping it. I think, though, AIC requirement is satisfied, so can be dealt separately.

kparichay:
Please open a new issue listing out the things to be done then.

kparichay:
Closing issue as the current requirements are met. Can re-opened if some bugs arise or create new issue for further requirements.

PR #1182, opened 2021-05-12
[graph] Add graph core class 

kparichay:
Add graph core class
This is mainly separating the existing implementation of core part
of the graph from NetworkGraph.
This patch does not contain new implementation but mostly contain
borrowed implementation from existing graph.

See Also #1176

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/graph_core.h
+++ b/nntrainer/graph/graph_core.h
@@ -1,0 +1,270 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file    network_graph.h
+ * @date    12 May 2020
+ * @see     https://github.com/nnstreamer/nntrainer
+ * @author  Jijoong Moon <jijoong.moon@samsung.com>
+ * @author  Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug     No known bugs except for NYI items
+ * @brief   This is Graph Core Class for Neural Network
+ *
+ */
+
+#ifndef __GRAPH_CORE_H__
+#define __GRAPH_CORE_H__
+#ifdef __cplusplus
+
+#include <iostream>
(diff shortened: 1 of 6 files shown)


jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.h
```diff
@@ -69,12 +69,21 @@ class LayerNode : public ml::train::Layer, public GraphNode {
...
   }
 
+  /**
+   * @brief     set name of layer
+   *
+   * @param[in] name Name of the layer
+   * @retval #ML_ERROR_NONE Successful.
+   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   */
+  int setName(const std::string &name) { return setProperty({"name=", name}); }
```
zhoonit, line 85: maybe this?
```suggestion
  int setName(const std::string &name) { return setProperty({"name=" + name}); }
```

nntrainer/graph/graph_core.h
```diff
@@ -0,0 +1,270 @@
...
+    typename std::vector<std::shared_ptr<GraphNode>>::const_reverse_iterator
+      const_reverse_iterator;
+
+  /**
+   * @brief     Constructor of Graph Core Class
+   */
+  GraphCore() : def_name_count(0) {}
+
+  /**
+   * @brief Create new GraphNode and add into Graph
```
zhoonit, line 53: is this also creating a new graphnode?

nntrainer/layers/layer_node.h
kparichay: yes, this was done in the next PR.

kparichay: done here as well.

nntrainer/graph/graph_core.h
kparichay: no, it is not. Updated documentation.

zhoonit (approved):
LGTM

PR #1193, opened 2021-05-18
[model] Model inference should not catch throw

### LINKED ISSUE #1191: model inference should not catch its throw
model inference should not catch its throw. `inference` does not return error but output. So, it is expected to throw.
Further, all the calls to inference expect it to throw to catch the error.

kparichay:
Model inference should not catch throw, and must throw errors
so that the callers can catch it and detect errors.
This has also revealed bug in time distribution layer test case which was hidden till now.

Resolves #1191.

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -480,18 +480,13 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
 
   sharedConstTensors out;
   if (!validateInput(X))
-    return out;
+    throw std::invalid_argument("Input validation failed.");
 
   allocate(false);
 
-  try {
-    START_PROFILE(profile::NN_FORWARD);
-    forwarding(X, {}, false);
-    END_PROFILE(profile::NN_FORWARD);
-  } catch (...) {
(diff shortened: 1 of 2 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -484,14 +484,9 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
...
-    START_PROFILE(profile::NN_FORWARD);
-    forwarding(X, {}, false);
-    END_PROFILE(profile::NN_FORWARD);
-  } catch (...) {
-    ml_loge("Failed to inference Model");
-    return out;
-  }
+  START_PROFILE(profile::NN_FORWARD);
+  forwarding(X, {}, false);
+  END_PROFILE(profile::NN_FORWARD);
```
zhoonit, line 489: Could you checkout line 482 as well? should we throw if `validateTensor` fails?

kparichay: done.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1204, opened 2021-05-21
[fc/tensor] Update fc layer bias derivative to in-place

### LINKED ISSUE #1202: FC calcGradient is not regarding multidimension
sum should be run axis through height and channel as well not only width.
Also, allocate shouldn't happen here

kparichay:
Update tensor sum to support in-place sum when done over
multiple dimensions.
Also update arguments order to have output tensor as a latter
argument.

Make bias derivative in-place for fc layer

**Self evaluation:**

resolves #1202


zhoonit (approved):
LGTM :)

zhoonit:
resolves #1202

jijoongmoon (approved):
LGTM

PR #1206, opened 2021-05-21
[Props] Add concept of empty to property

zhoonit:
- [Props] Add concept of empty to property

This patch adds empty concept to property instead of having insensible
value inside a property as a default

before this patch,
```cpp
auto a = Property<int>();
EXPECT_EQ(a.get(), -1); // let's assume -1 is default value which might not be sensible.
```
after this patch

```cpp
auto a = Property<int>();
EXPECT_THROW(a.get(), std::invalid_argument);
EXPECT_TRUE(a.empty());
```

so underlying value remains always null or valid.

Now, it became natural to distinguish mandatory props and non-mandatory
props

**Self evaluation:**

### DIFF
--- a/nntrainer/utils/base_properties.h
+++ b/nntrainer/utils/base_properties.h
@@ -10,6 +10,7 @@
  * @bug No known bugs except for NYI items
  */
 #include <array>
+#include <memory>
 #include <sstream>
 #include <string>
 #include <vector>
@@ -112,15 +113,41 @@ public:
    * @brief Construct a new Property object
    *
    */
-  Property() = default;
+  Property() : value(nullptr){};
 
   /**
-   * @brief Construct a new Property object, setting default skip validation on
-   * purpose
+   * @brief Construct a new Property object
    *
    * @param value default value
    */
-  Property(const T &value) : value(value){};
+  Property(const T &value_) { set(value_); }
+
+  /**
+   * @brief Copy Construct a new Property object
+   *
+   * @param rhs right side to copy from
+   */
+  Property(const Property &rhs) {
+    if (this != &rhs) {
+      value = std::make_unique<T>(*rhs.value);
+    }
+  }
+
+  /**
+   * @brief Copy assignment operator of a new property
+   *
+   * @param rhs right side to copy from
+   * @return Property& this
+   */
+  Property &operator=(const Property &rhs) {
+    if (this != &rhs) {
+      value = std::make_unique<T>(*rhs.value);
+    }
+    return *this;
+  };
+
+  Property(Property &&rhs) noexcept = default;
+  Property &operator=(Property &&rhs) noexcept = default;
 
   /**
    * @brief Destroy the Property object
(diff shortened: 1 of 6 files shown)


lhs8928 (approved):
LGTM

nntrainer/utils/base_properties.h
```diff
@@ -133,28 +160,44 @@ template <typename T> class Property {
...
+    return *value;
+  }
+
+  /**
+   * @brief check if property is empty
+   *
+   * @return true empty
+   * @return false not empty
+   */
+  bool empty() const { return value == nullptr; }
```
myungjoo, line 200: ```isEmpty()```?

Users may think this "empties data", not "checks if it's empty"

nntrainer/utils/base_properties.h
```diff
@@ -133,28 +160,44 @@ template <typename T> class Property {
...
   /**
    * @brief get the underlying data
    *
    * @return T& data
    */
-  T &get() { return value; }
+  T &get() {
+    NNTR_THROW_IF(value == nullptr, std::invalid_argument)
+      << "Cannot get property, property is empty";
+    return *value;
```
myungjoo, line 191: Your commit message, ```This patch adds empty concept to property instead of having invalid
value as a default.``` suggests that ```value == nullptr``` is allowed. Is it a code-bug? or an insufficient commit message?

nntrainer/utils/base_properties.h
```diff
@@ -133,28 +160,44 @@ template <typename T> class Property {
...
-  operator T &() { return value; }
+  operator T &() { return get(); }
 
   /**
    * @brief cast operator for property
    *
    * @return T value
    */
-  operator const T &() const { return value; }
+  operator const T &() const { return get(); }
```
myungjoo, line 170: T operator? operator T?

nntrainer/utils/base_properties.h
zhoonit: before this patch,
```cpp
auto a = Property<int>();
EXPECT_EQ(a.get(), -1); // let's assume -1 is default value which might not be sensible.
```
after this patch

```cpp
auto a = Property<int>();
EXPECT_THROW(a.get(), std::invalid_argument); 
EXPECT_TRUE(a.empty());
```

so value remains always null or valid.

Property wraps around a value. So, I'll change the commit msg to
```diff
- invalid value
+ invalid property
```

and add more explanation :)

nntrainer/utils/base_properties.h
```diff
@@ -133,28 +160,44 @@ template <typename T> class Property {
...
+    NNTR_THROW_IF(value == nullptr, std::invalid_argument)
+      << "Cannot get property, property is empty";
+    return *value;
+  }
+
+  /**
+   * @brief check if property is empty
+   *
+   * @return true empty
+   * @return false not empty
```
zhoonit, line 198: change this to retval

nntrainer/utils/base_properties.h
zhoonit: yes, I'll change this :)

nntrainer/utils/base_properties.h
zhoonit: Operator T & is a casting operator.
The intention here is to make using property as if it's underlying value type :)

```cpp
const auto name = property<string>();
std::string& name_ = name; ///this function is called
```

nntrainer/utils/base_properties.h
zhoonit: Sorry, I found an issue prevent to change to `isEmpty()`

there is a template function insided nntrainer like below

```cpp
template<typename T>
void some_func(T& data){

  if(data.empty()) {
   /// do something
  }
}

TEST(can_compile_this) {
  Property<int> prop;
  std::vector<Property<int>> vector_of_prop

  some_func(prop); // prop.isEmpty() will fail
  some_func(vector_of_prop);
}
```

as stl has fixed semantics that *.empty() for testing emptiness. (arrays, vectors or others...)

kparichay (approved):
LGTM

nntrainer/utils/base_properties.h
kparichay: is this already done?

zhoonit: Yes

zhoonit: chronologically,

1. Create this PR
2. Patch return -> retval (merged)
3. this PR patched (today)

nntrainer/utils/base_properties.h
myungjoo: Ok.. it seems that STL is the culprit. keep using empty() for this.

myungjoo (approved):
LGTM

jijoongmoon (approved):
LGTM

Issue #1208, opened 2021-05-21
Support split layer


kparichay:
This is an alternative to split the output of the dataset to provide multiple inputs.

PR #1210, opened 2021-05-21
[Split layer] Added support for split layer

kparichay:
Added split layer scaffolding with its properties and initialization
Add forwarding and backwarding for the split layer

See Also #1208

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/layer_internal.h
+++ b/nntrainer/layers/layer_internal.h
@@ -235,6 +235,7 @@ public:
    *            33. in_length : int ( input length for embedding layer )
    *            34. recurrent_activation :  string (type) - lstm
    *            35. distribute : bool
+   *            36.
    */
   enum class PropertyType {
     input_shape = 0,
@@ -273,6 +274,7 @@ public:
     in_length = 33,
     recurrent_activation = 34,
     distribute = 35,
+    split_dimension = 36,
     unknown
   };
 
--- a/nntrainer/layers/split_layer.cpp
+++ b/nntrainer/layers/split_layer.cpp
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
+ *
+ * @file   split_layer.cpp
+ * @date   21 May 2021
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is Split Layer Class for Neural Network
+ *
+ */
+
+#include <cstring>
+#include <layer_internal.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <split_layer.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+const std::string SplitLayer::type = "split";
+
+int SplitLayer::initialize(Manager &manager) {
+  int status = ML_ERROR_NONE;
+
+  if (split_dimension < 1) {
+    ml_loge("Error: cannot split along the batch dimension");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (split_dimension >= MAXDIM) {
+    ml_loge("Error: split dimension exceeding the total number of dimensions");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (getNumInputs() == 0) {
+    ml_loge("Error: number of inputs are not initialized");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (getNumInputs() > 1) {
+    ml_loge("Error: only a single input is supported with split layer");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  const TensorDim &in_dim = input_dim[0];
+  setNumOutputs(in_dim.getTensorDim(split_dimension));
+
+  TensorDim d = in_dim;
+  d.setTensorDim(split_dimension, 1);
+
+  for (unsigned int idx = 0; idx < getNumOutputs(); ++idx) {
+    output_dim[idx] = d;
+  }
+
+  /**
+   * Setup input_reshape_helper to which input will be reshaped in forwarding
+   * to facilitate easier processing.
+   */
+  input_reshape_helper = {1, 1, 1, 1};
+  for (unsigned int idx = 0; idx < split_dimension; ++idx) {
+    input_reshape_helper.batch(input_reshape_helper.batch() *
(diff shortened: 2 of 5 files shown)


nntrainer/layers/split_layer.cpp
```diff
@@ -0,0 +1,153 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
```
jijoongmoon, line 3: Change the copyrigit?

nntrainer/layers/split_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (getNumInputs() > 1) {
+    ml_loge("Error: only a single input is supported with split layer");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  const TensorDim &in_dim = input_dim[0];
+  setNumOutputs(in_dim.getTensorDim(split_dimension));
```
jijoongmoon, line 50: Could you add comments here? how exactly we expect the split_dimension from user? batch size must be same, and height and width??

nntrainer/layers/split_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * Setup input_reshape_helper to which input will be reshaped in forwarding
+   * to facilitate easier processing.
+   */
+  input_reshape_helper = {1, 1, 1, 1};
+  for (unsigned int idx = 0; idx < split_dimension; ++idx) {
+    input_reshape_helper.batch(input_reshape_helper.batch() *
+                               in_dim.getTensorDim(idx));
```
jijoongmoon, line 66: does it flatten the input ?? [ batch, channel, height, width ] --> [batch , 1, 1, channel, height, widht]?

kparichay: the dimension is changed to [dimensions_before_split_dimension, 1, split_dimension, dimensions_after_split_dimension].
for example, if [b,c,h,w] is being split at c, then this will be [b,1,c,h*w]. If being split at h, this will be [b*c,1,h,w]. If being split at b, then this will be [1,1,b,c*h*w].

nntrainer/layers/split_layer.cpp
kparichay: I will add.

nntrainer/layers/split_layer.cpp
kparichay: will do.

nntrainer/layers/layer_internal.h
```diff
@@ -273,6 +274,7 @@ class Layer {
     in_length = 33,
     recurrent_activation = 34,
     distribute = 35,
+    split_dimension = 36,
```
zhoonit, line 277: Wonder if you'd like to try properties. I am willing to reach out if you are interested :)

nntrainer/layers/split_layer.h
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
```
zhoonit, line 3: Copyright here as well?

nntrainer/layers/split_layer.cpp
zhoonit: Could you elaborate with an example? 
eg.) if [3, 4, 2, 5], splited on c, is it becoming {[3, 1, 4, 10]} ?  I wonder where does split occur?

nntrainer/layers/layer_internal.h
kparichay: how about it as the first thing on top of this PR?

nntrainer/layers/split_layer.cpp
kparichay: the reason for this reshape is the place of splitting. In the given example, as the split if happening at `c`, the output shapes must be `[3,1,2,5]x4`.  In order to achieve this, we need to keep the dimensions before and after the split dimension the same.
So, we merge all the dimensions before and after the dimensions, loop over the dimension before the split, split the split dimension, and copy the data with the size of the dimension after the split. So, in this case, we copy of size `10` at once.
And create two loops of outer dimensions `3` and `4`. This methodology expands to any number of dimensions and using any dimension to split. Unittests are missing but I will add them soon.

There is one limitation of this method - we always split into `n` pieces, if the split dimension is `n`. Later, we can expand to take number of splits as an argument if needed.

nntrainer/layers/split_layer.h
kparichay: will do.

nntrainer/layers/split_layer.cpp
jijoongmoon: > Later, we can expand to take number of splits as an argument if needed.
Yes. we need to support for not even dimensions cases.

nntrainer/layers/layer_internal.h
zhoonit: Sure, let me do it.

nntrainer/layers/split_layer.cpp
kparichay: > > Later, we can expand to take number of splits as an argument if needed.
> 
>  And It looks like the channel is always 1 and we will split always according to height. Therefore the number of splited output is the same as the number of height. Am I right?

The number of split output is the value of the dimension at the split_dimension. Channel is always 1 for the reshape dimension, but the actual dimension comes out differently. For example, for `[3,4,2,5]` with split_dimension set as `c`, then the output will be `[3,1,2,5] x 4`. If the split dimension is set as `h`, then the output will be `[3,4,1,5] x 2`.

nntrainer/layers/layer_internal.h
kparichay: let me do this one, and get adjusted to the updated format :).

nntrainer/layers/split_layer.cpp
kparichay: done.

nntrainer/layers/split_layer.h
kparichay: done.

nntrainer/layers/split_layer.cpp
kparichay: done.

jijoongmoon (approved):
LGTM. Need Unit Test Cases Later PR.

nntrainer/layers/layer_internal.h
zhoonit: Awesome :)

zhoonit (approved):
LGTM :)

PR #1219, opened 2021-05-26
[Recorder] Add multiout equivalent in transLayer @open sesame 06/09 15:12

zhoonit:
- [Recorder] Add multiout equivalent in transLayer

```
When getting gradient from `gradientTape` it adds all derivatives of the
dependent layer. While this behavior is true, we need to measure to what
extent each layer contributed to the layer.

This patch adds a multiout layer which consists of some stublayer that
seperates derivatives one by one.

**Self evaluation:**

```

### DIFF
--- a/test/input_gen/transLayer.py
+++ b/test/input_gen/transLayer.py
@@ -144,6 +144,36 @@ CHANNEL_LAST_LAYERS = (
     K.layers.MaxPool2D,
 )
 
+##
+# @brief Multiout wrapper layer, this class separate gradient
+# to calculate derivative properly
+# when calling, this returns [x] * @a num_output just like output Layer does in NNTrainer
+class MultiOutLayer(IdentityTransLayer):
+
+    ##
+    # @brief init function
+    # @param tf_layer tf_layer to get number of outputs
+    # @param num_output explicit number to generate number of output
+    def __init__(self, tf_layer = None, *args, num_output, **kwargs):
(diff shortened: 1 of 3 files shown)


jijoongmoon (approved):
LGTM

test/input_gen/transLayer.py
```diff
@@ -144,6 +144,36 @@ def to_nntr_weights(self, weights):
...
+    # @param tf_layer tf_layer to get number of outputs
+    # @param num_output explicit number to generate number of output
+    def __init__(self, tf_layer = None, *args, num_output, **kwargs):
+        if not tf_layer:
+            tf_layer = K.layers.Layer()
+
+        super().__init__(tf_layer, *args, **kwargs)
+
+        # this enables seperating gradient one by one
+        self.stub_layers = [K.layers.Lambda(lambda x: x + 0) for i in range(num_output)]
```
kparichay, line 164: How about simply
```suggestion
        self.stub_layers = [K.layers.Lambda(lambda x: x) for i in range(num_output)]
```

zhoonit: Checked this, and it didn't work. I suppose if I don't do x: x + 0, I think tensorflow optimizer optimizes this to a `noop`.

kparichay: ok.

kparichay (approved):
LGTM

PR #1240, opened 2021-06-02
[Product Ratings] Fix inconsistent batch

zhoonit:
- [Product Ratings] Fix inconsistent batch

```
This patch fixes inconcistent batch size between source file and model
configuration.

Also found a bug #1238, which will be dealt seperately.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

```

PR #1241, opened 2021-06-02
[SplitLayer] Support flexible dimension @open sesame 06/03 12:41

### LINKED ISSUE #1238: Split_layer requires specialized batch handling
Current implementation of split layer is assuming fixed set of batch size, this needs to be handled in a flexible manner.

zhoonit:
~**[Pending commit: #1240]**~

- [SplitLayer] Support flexible dimension

```
This patch adds support for the flexible dimension in split layer.

**Self evaluation:**

Cc: Parichay Kapoor <pk.kapoor@samsung.com>

```

resolves #1238

### DIFF
--- a/nntrainer/layers/split_layer.cpp
+++ b/nntrainer/layers/split_layer.cpp
@@ -72,10 +72,12 @@ int SplitLayer::initialize(Manager &manager) {
    * together and all the dimensions after the split_dimension to faciliate
    * easier splitting of the data.
    */
-  input_reshape_helper = {1, 1, 1, 1};
-  for (unsigned int idx = 0; idx < split_dimension; ++idx) {
-    input_reshape_helper.batch(input_reshape_helper.batch() *
-                               in_dim.getTensorDim(idx));
+  leading_helper_dim = 1;
+  input_reshape_helper.channel(1);
+  input_reshape_helper.height(1);
+  input_reshape_helper.width(1);
+  for (unsigned int idx = 1; idx < split_dimension; ++idx) {
+    leading_helper_dim *= in_dim.getTensorDim(idx);
   }
 
   input_reshape_helper.height(in_dim.getTensorDim(split_dimension));
@@ -92,7 +94,15 @@ int SplitLayer::initialize(Manager &manager) {
   output_reshape_helper = input_reshape_helper;
   output_reshape_helper.height(1);
 
+  setBatch(in_dim.batch());
+
   return status;
+}
+
+void SplitLayer::setBatch(unsigned int batch) {
+  Layer::setBatch(batch);
+  input_reshape_helper.batch(batch * leading_helper_dim);
+  output_reshape_helper.batch(batch * leading_helper_dim);
 }
 
 void SplitLayer::forwarding(bool training) {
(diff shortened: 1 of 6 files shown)


nntrainer/layers/split_layer.cpp
```diff
@@ -92,17 +92,38 @@ int SplitLayer::initialize(Manager &manager) {
   output_reshape_helper = input_reshape_helper;
   output_reshape_helper.height(1);
 
+  std::cout << output_reshape_helper;
```
kparichay, line 95: remove

nntrainer/layers/split_layer.cpp
```diff
@@ -92,17 +92,38 @@ int SplitLayer::initialize(Manager &manager) {
...
 }
 
+/**
+ * @brief create a helper dimension with a new batch set
+ *
+ * @param batch
+ * @param base
+ * @return TensorDim
+ */
+static TensorDim setHelperBatch(unsigned int batch, const TensorDim &base) {
```
kparichay, line 107: how about keep L76 the original version, and rather override setBatch() function here to update the input_reshape_helper/output_reshape_helper (will be a simple scale).
this will also keep rest of the code same and no overhead in fowarding/backwarding.

nntrainer/layers/split_layer.cpp
zhoonit: done

nntrainer/layers/split_layer.cpp
zhoonit: That was my first idea. That cannot be done because, in that way, setBatch has side effect.

consecutive call of 
setBatch(20);
setBatch(20);

will accumulate and there seems no way to know if we have to reset the batch or the batch has information of split dimensions.

Actually we can have another member to save this state. I'll make a tweak

zhoonit: Updated.

kparichay: looks good now.

kparichay (approved):
LGTM after CI pass and rebase.

zhoonit:
> LGTM after CI pass and rebase.

There is a problem, that in this design. `setBatch` has temporal dependency of `initialize`.
So `layer::initialize` must be called before `setBatch()`. That's why split_layer test fails.

Should we have embrace this dependency?

kparichay:
Let's cache the value of the current batch size.
Then set `input_reshape_helper.batch(cached_batch * leading_helper_dim);` where `leading_helper_dim` must be initialized to 1 inside `setBatch`.

inside initialize, use `input_reshape_helper.batch(cached_batch * leading_helper_dim);` as well.

zhoonit:
> Let's cache the value of the current batch size inside the split layer.
> Then set `input_reshape_helper.batch(cached_batch * leading_helper_dim);` where `leading_helper_dim` must be initialized to 1 inside `setBatch`.
> 
> inside initialize, use `input_reshape_helper.batch(cached_batch * leading_helper_dim);` as well.

This should work, will do

kparichay (approved):
LGTM

kparichay:
resolves #1238

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1248, opened 2021-06-03
[deb/test] Minor update to debian packaging and unittests

kparichay:
Add nnstreamer filter in the packaging
Also update the unittests to use the meson tests

Update unittest tolerance to be 1e-5 for debian build to match
across platforms (bionic, focal) and architectures.
Similar issues have occured for i586 tizen as well.
So, this patch lowers the tolerance for all ccapi and capi test.
This issue occurs to os and library versioning issues.

Update debian packaging:
- update the unittests to use meson tests
- update dh_missing to fail than give warnings

Remove run_unittests

**Self evaluation:**


jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

debian/nnstreamer-nntrainer-dev.install
```diff
@@ -0,0 +1 @@
+/usr/lib/*/libnnstreamer_filter_nntrainer.a
```
zhoonit, line 1: Are we using static?

kparichay: its released in tizen, so debian matches it for now.
if it is to be removed, then then need to check for tizen first.

zhoonit (approved):
LGTM

debian/nnstreamer-nntrainer-dev.install
zhoonit: I see, thanks

Issue #1251, opened 2021-06-04
Enable launchpad build


kparichay:
Enable launchpad build  for both bionic and focal:
- [x] update debian packaging #1236 
- [x] update install filenames for the API #1246 
- [x] set common tolerance for the unittests to pass on launchpad #1248
- [x] lower tolerance further just for debian builds

Do not close this issue till the launchpad build succeeds.

kparichay:
All builds pass except armhf bionic.

zhoonit:
Wonder if we can close this?

kparichay:
armhf bionic still fails.

zhoonit:
Sad

PR #1252, opened 2021-06-04
[pkg] Reduce error tolerance for all builds @open sesame 06/08 10:52

kparichay:
Reduce error tolerance for release builds to succeed build on launchpad and tizen builds.
The default error tolerance is only reduced for all release builds for all tests that use defined tolerance.
However, the CI will run at original tolerance for all the builds and will ensure that
values are matched with original precision.

The tolerance has to be reduced because different os
have different versions and implementations of the blas
which results in minor difference in the results of the float
operations. However, when the difference is accumulated over
a range of operations, the unittests fail to match the
exact values.

See Also #1251

**Self evaluation:**

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -21,3 +21,6 @@ option('capi-ml-inference-actual', type: 'string', value: 'capi-ml-inference',
         description: 'backward compatible dependency name of capi-ml-inference')
 option('capi-ml-common-actual', type: 'string', value: 'capi-ml-common',
         description: 'backward compatible dependency name of capi-ml-common')
+
+# detect build os
+option('reduce-tolerance', type: 'boolean', value: true)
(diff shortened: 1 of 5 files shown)


meson_options.txt
```diff
@@ -21,3 +21,6 @@ option('capi-ml-inference-actual', type: 'string', value: 'capi-ml-inference',
         description: 'backward compatible dependency name of capi-ml-inference')
 option('capi-ml-common-actual', type: 'string', value: 'capi-ml-common',
         description: 'backward compatible dependency name of capi-ml-common')
+
+# detect build os
+option('debian-build', type: 'boolean', value: false)
```
zhoonit, line 26: How about we have a reduce test toloerance option and instead of debian build option?

zhoonit: Tizen i586 has same problem, so...

kparichay: that will reduce the tolerance everywhere. for now, the problem seems limited to capi and ccapi unittests.

kparichay: Updated to reduce the tolerance for packaging, but CI runs at max tolerance (reduce tolerance enabled by default but CI will disable it)?

kparichay:
Updated to reduce the tolerance for packaging, but CI runs at max tolerance (reduce tolerance enabled by default but CI will disable it).

zhoonit (approved):
LGTM :)

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1260, opened 2021-06-08
[layer] add supportInPlace() as an interface for the layer

### LINKED ISSUE #867: Inplace to be a layer property

kparichay:
Support inPlace() as an interface for the layer.
If returns true, manager can set its inputs/outputs to be
the same memory location, but not necessarily.
Corresponding graph code is also updated.

Resolves #867

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/activation_layer.h
+++ b/nntrainer/layers/activation_layer.h
@@ -55,13 +55,13 @@ public:
    * @brief     Read Activation layer params. This is essentially noops for now.
    * @param[in] file input stream file
    */
-  void read(std::ifstream &file) override{/* noop */};
+  void read(std::ifstream &file) override{/** noop */};
 
   /**
    * @brief     Save Activation layer params. This is essentially noops for now.
    * @param[in] file output stream file
    */
-  void save(std::ofstream &file) override{/* noop */};
+  void save(std::ofstream &file) override{/** noop */};
 
   /**
    * @copydoc Layer::forwarding(bool training)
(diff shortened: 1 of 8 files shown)


nntrainer/layers/activation_layer.h
```diff
@@ -85,6 +86,11 @@ class ActivationLayer : public Layer {
    */
   const std::string getType() const override { return ActivationLayer::type; };
 
+  /**
+   * @copydoc Layer::supportInPlace()
+   */
+  bool supportInPlace() const override { return support_in_place; }
```
zhoonit, line 92: how about `acti_type != ActivationType::ACT_SOFTMAX` instead of member variable?

zhoonit (approved):
LGTM with a minor suggestion

nntrainer/layers/activation_layer.h
kparichay: it works, for now, might create problems when more activation types are added.

zhoonit: Yes, I was thinking it'd better have to calculate right on the way because this guarantees invariant at the cost of calculation, which in this case looks OK deal.

It's just a suggestion leaving decision up to you :)

kparichay: updated. Also, moved decision making to ActiFunc.

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1263, opened 2021-06-10
[Loss] Apply requireLabel to feed label

zhoonit:
- [Loss] Apply requireLabel to feed label

```
Currently, loss layer is the only one that can accept label which was
prohibiting building a custom loss layer.

**Major Changes**
- Apply requireLabel to feed label
- Last layer is not assumed loss layer when feeding label(still it is in
other places, need to be dealt)

**Minor Changes**
- Feeding multiple label is now possible (spliting label is future work)
- [Fix] Count Trainable includes loss layer
- [Fix] Handle cases When all layer is not trainable for
neuralnet::backwarding

**Self evaluation:**

```

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -50,10 +50,10 @@ int NetworkGraph::compile(const LossType loss_type) {
 
   graph.topologicalSort();
 
-  countNonTrainableLayersAtBegin();
-
   status = addLossLayer(loss_type);
   NN_RETURN_STATUS();
+
+  countNonTrainableLayersAtBegin();
 
   status = checkCompiledGraph();
   NN_RETURN_STATUS();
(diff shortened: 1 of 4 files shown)


lhs8928 (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -50,11 +50,11 @@ int NetworkGraph::compile(const LossType loss_type) {
 
   graph.topologicalSort();
 
-  countNonTrainableLayersAtBegin();
-
   status = addLossLayer(loss_type);
   NN_RETURN_STATUS();
 
+  countNonTrainableLayersAtBegin();
```
kparichay, line 56: good catch.

nntrainer/graph/network_graph.cpp
```diff
@@ -249,19 +249,32 @@ int NetworkGraph::realizeMultiOutputType(
   return status;
 }
 
-/** TODO: this needs special attention */
+/**
+ * @fixme: the implementation assumes loss layer should always be at the last
```
kparichay, line 253: yes, I think the loss layer setup has to be rewritten.

nntrainer/layers/time_dist.h
```diff
@@ -97,6 +97,11 @@ class TimeDistLayer : public Layer {
    */
   std::shared_ptr<Layer> &getDistLayer() { return dist_layer; };
 
+  /**
+   * @copydoc Layer::requireLabe()
+   */
+  bool requireLabel() const override { return dist_layer->requireLabel(); }
```
kparichay, line 103: can you consider not adding this here, but rather move this to LayerNode and use getLayer()->requireLabel().
This will cover the dist layer case automatically.
The purpose is to keep dist layer as light as possible, otherwise management becomes tough.

zhoonit: You mean sthlike below?

```cpp
LayerNode::requireLabe() {
 if (type==timedist_layer) {
     return layer->getLayer()->requireLabel()
 }
}
```

If it is, or not, Could you ping me if you've got time to discuss offline?

kparichay: no, the below code precisely:
```cpp
const bool LayerNode::requireLabe() const {
     return getLayer()->requireLabel();
}
```

zhoonit: Ah, yeah this has to be done, let me do it in follow up PR

kparichay: then make sure to remove the corresponding function from time dist layer.

kparichay (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -249,19 +249,32 @@ int NetworkGraph::realizeMultiOutputType(
...
+ * layer and the there is only one loss, this assumption is not true
+ */
```
jijoongmoon, line 255: Is it true? we only support only on loss?? I thought we can run multi-output.

nntrainer/layers/time_dist.h
zhoonit: Yup

nntrainer/graph/network_graph.cpp
zhoonit: Yes, for now, multi-output is supported only in the middle.

jijoongmoon (approved):
LGTM

PR #1271, opened 2021-06-11
[graph] Graph core cleanup

kparichay:
1. [graph] Make adjacency list lazily
Update graph to make adjacenct list lazily when required
during the topological sort.
Until then, keep the node_list which contains simply all the nodes.

2. [graph] Update node_names to unordered_set
Update node_names storing all the names of the nodes
in the graph for uniqueness to use unordered_set.

3. [graph] Graph iterator for sorted and unsorted 
Updated graph iteartor to run over unsorted list of nodes
if the graph is not yet sorted. If the graph is sorted,
the iterator iterates over sorted nodes.

4. [graph] Remove getNodes for the graph
Remove getNodes() which provided direct access to the data structure
holding the graph. Now, all access is done using the iterators.

5. [graph] Remove non-const iterators
Remove non-const iterators for the graph.
Update the corresponding usages for the graph.

Note: the pattern for (auto const &node : graph) has been
removed as this can begin()/end() in the background.
Workaround for this is to use std::as_const(graph) but this
is only available from c++17, so this patch uses
the basic const iterator for loop pattern.

See Also #986

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/graph_core.cpp
+++ b/nntrainer/graph/graph_core.cpp
@@ -22,33 +22,19 @@
 
 namespace nntrainer {
 
-void GraphCore::removeEdges() {
-  /**
-   * remove all edges to save memory.
-   * NodeList is kept for now for O(1) access of nodes by idx.
-   */
-  for (unsigned int i = 0; i < adj.size(); ++i) {
-    /**
-     * As this resize is guaranteed to not insert new elements, create a
-     * default element needed by resize.
-     */
-    adj[i].resize(1);
-  }
-}
-
 void GraphCore::addGraphNode(std::shared_ptr<GraphNode> node) {
-  node->setIndex(adj.size());
-  adj.push_back(std::list<std::shared_ptr<GraphNode>>({node}));
+  node->setIndex(node_list.size());
+  node_list.push_back(node);
 }
 
 const std::shared_ptr<GraphNode> &GraphCore::getNode(unsigned int ith) const {
   if (ith >= size())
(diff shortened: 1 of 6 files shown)


nntrainer/graph/graph_core.cpp
```diff
@@ -59,27 +45,47 @@ GraphCore::getSortedNode(unsigned int ith) const {
...
+    for (auto const &in_conn : node->getInputConnections()) {
+      if (istrequal(in_conn, "__data__"))
+        continue;
+      unsigned int to_node_id = getNode(in_conn)->getIndex();
+      adj[to_node_id].push_back(node);
+    }
+  }
+}
+
 void GraphCore::topologicalSortUtil(
```
zhoonit, line 66: How about making this as a static free function inside this source file?

kparichay: will do.

kparichay: actually, this cant be done. topologicalSortUtil uses getNode() which requires it to be member function.
There is work around to use directly get it from adj, but it uses the knowledge of how index is assigned, and I dont think its a good idea.

zhoonit: Okay

zhoonit (approved):
LGTM

kparichay:
rebasing for layer_v2 as it is needed for layer_v2 changes.

jijoongmoon (approved):
LGTM

PR #1285, opened 2021-06-16
[context] Layer context creation scaffolding

kparichay:
Added todo list for the layer context creation
Added minor new interfaces for the layer context
Disable in-place optimization and inference memory optimization

### DIFF
--- a/nntrainer/layers/layer_node.h
+++ b/nntrainer/layers/layer_node.h
@@ -273,7 +273,9 @@ public:
 private:
   // TODO: make this unique_ptr once getObject API is removed
   std::shared_ptr<nntrainer::Layer>
-    layer;      /**< The actual object in the graph node */
+    layer; /**< The actual object in the graph node */
+  // TODO: possibly remove, two identifiers for the same  node (name and index)
+  // can lead to issues later
   size_t index; /**< index of each node */
 
   /** TODO : move management of num_inputs to layer_node */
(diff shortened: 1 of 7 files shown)


nntrainer/layers/layer_node.h
```diff
@@ -286,7 +288,10 @@ class LayerNode : public ml::train::Layer, public GraphNode {
 
   RunLayerContext
     run_context; /**< context required for running/execution of the layer. This
-                    will also contain the properties of the layer. */
+                    will also contain the properties of the layer. The
+                    properties will be copied upon final creation. Editing
+                    properties of the layer after init will not the properties
+                    in the context/graph unless intended. */
```
zhoonit, line 294: I wonder about the purpose of this behavior. My concern is that we could save to the context with `std::any` but we can't save the type.

kparichay: to avoid updating properties once the layer is initialized.

kparichay: > My concern is that we could save to the context with std::any but we can't save the type.

you want to directly save to run_context?

zhoonit: How about layer node control this behavior or let it be free for now?

kparichay: let's keep it free for now. how to handle stopping the setProperties() after init() can be handled later.

zhoonit: Cool

zhoonit: > you want to directly save to run_context?

I mean properties have it's own type. So there is no way to have a member if we are not using std::any. Let's delay the discussion for now :)

kparichay: yes.

zhoonit (approved):
LGTM

lhs8928 (approved):
LGTM

kparichay:
> Android build failing, please check and label `PR/OK2GO`

done

PR #1287, opened 2021-06-16
[addition] Bug fix addition layer calcDerivative

kparichay:
Bug fix for addition layer calcDerivative().
addition layer assign the same tensor derivative memory back to its
other layers. If the other layers connecting with addition layers are
inplace, then this can lead to wrong results.

cc. @zhoonit


zhoonit (approved):
Nice, LGTM :)

jijoongmoon (approved):
LGTM

PR #1290, opened 2021-06-16
[manager] Support request Tensors and Weights 

kparichay:
Added support for request Tensors and Weights
Moved spec of Tensors and Weights to VarGrad and Weights correspondingly
Also created constructors for weights and var_grad with their specs

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -626,4 +626,50 @@ void Manager::deinitializeTensors() {
   tensors_initialized = false;
 }
 
+/**
+ * @brief     Create weights with the given spec
+ *
+ */
+std::vector<Weight *>
+Manager::requestWeights(const GraphNode &node,
+                        std::vector<Weight::Spec> &weights_spec) {
+  std::vector<Weight *> ret;
+  std::vector<std::unique_ptr<Weight>> weights_list;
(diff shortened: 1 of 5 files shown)


nntrainer/tensor/manager.cpp
```diff
@@ -626,4 +626,50 @@ void Manager::deinitializeTensors() {
   tensors_initialized = false;
 }
 
+/**
+ * @brief     Create weights with the given spec
+ *
+ */
+std::vector<Weight *>
+Manager::requestWeights(const GraphNode &node,
+                        std::vector<Weight::Spec> &weights_spec) {
```
zhoonit, line 635: how about `const std::vector<Weight::Spec> &` ?

kparichay: this creates issue `reference to local variable` issue.

zhoonit: Right

zhoonit (approved):
LGTM

nntrainer/tensor/manager.cpp
kparichay: i think the copy will be optimized by the compiler.

jijoongmoon (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1292, opened 2021-06-16
[Layer] Scaffold layer impl

zhoonit:
- [Layer] Scaffold layer impl

```
**Major changes**
- Create `LayerImpl`
- Change `LayerDevel` signature
  s/int initialize/void finalize
  s/int setProperty/void setProperty
- Move createLayer to layerImpl

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/layer_devel.h
+++ b/nntrainer/layers/layer_devel.h
@@ -61,23 +61,23 @@ public:
   virtual const std::string getType() const = 0;
 
   /**
-   * @brief     Initialize the layer
-   * @retval #ML_ERROR_NONE Successful.
-   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
+   * @brief     Finalize creating the layer,
    * @details   Input dimensions will be provided set in the context. This
    * function must set output dimensions in the given context. Further, context
    * can be used to request weights for the layer, and any extra tensor required
    * for the operation of the layer.
+   * @note      After calling this it is not allowed to
+   * change properties.
    * @note      No memory allocation must be performed in the initialization
    * step. Any tensor memory required must be requested to the context which
    * will be made available during execution of the layer with the context.
    */
-  virtual int initalize(InitContext &context) = 0;
+  virtual void finalize(InitContext &context) = 0;
 
   /**
    * @brief     Forward Propagation of a layer
-   * @param[in] context Context of the layer
-   * @param[in] training true if training, false if inference
+   * @param     context Context of the layer
+   * @param     training true if training, false if inference
    * @note      Output must be set in the output tensors.
    * @details   context provides access to the weights (if any), inputs,
    * outputs, and tensors (if any) for the layer. Input and output dimensions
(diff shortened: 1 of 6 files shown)


nntrainer/layers/layer_impl.h
```diff
@@ -0,0 +1,87 @@
...
+private:
+  bool finalized;                                 /**< check if finalized */
+  std::unique_ptr<std::tuple<>> layer_impl_props; /**< layer_impl_props */
+};
+
+/**
+ * @brief General Layer Factory function to register Layer
+ *
+ * @param props property representation
+ * @return std::unique_ptr<ml::train::Layer> created object
```
kparichay, line 74: nntrainer::Layer?

nntrainer/layers/layer_impl.h
```diff
@@ -0,0 +1,87 @@
...
+ */
+template <typename T,
```
kparichay, line 76: why move to LayerImpl? this will be called by model/graph/etc and they shouldnt have to include layer_impl.

zhoonit: I think the specification is the signature `std::unique_ptr<Layer>(const std::vector<std::string &props>)`. 

If we are going to make this implementation as a specification of an api, works for me.  But I guess we can delay this until we need it (maybe soon 😛 )

kparichay: isn't what the API calls? its left up to you where you keep this but no other header should import layer_impl.h except the ones which derive from LayerImpl (if they do, then a lot other headers will need importing as well). Others should import layer_devel.h

zhoonit: Yes. I'll keep in mind, if this is not possible this must go to layer_devel.h.

zhoonit: As discussed offline, this is reverted back

nntrainer/layers/layer_devel.h
```diff
@@ -174,6 +171,20 @@ class Layer {
 using CreateLayerFunc = nntrainer::Layer *(*)();
 using DestroyLayerFunc = void (*)(nntrainer::Layer *);
 
+/**
```
kparichay, line 174: this is fine for now, but later you can consider putting `LayerPluggable` stuff together, and moving `createLayer` out from the middle of it's related declarations and definitions.

kparichay (approved):
LGTM

lhs8928 (approved):
LGTM

nntrainer/layers/layer_impl.h
```diff
@@ -0,0 +1,73 @@
...
+class Exporter;
+
+enum class ExportMethods;
+/**
+ * @class   An abstract class to ease developing a layer
+ * @brief   An abstract class for all layers
+ *
+ */
+class LayerImpl : public Layer {
+
```
jijoongmoon, line 35: This ```public Layer``` is from layer_devel?

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_impl.h
zhoonit: Yes, it is from Layer :)

PR #1293, opened 2021-06-16
[Layer] Add const to the layer::getName()

zhoonit:
- [Layer] Add const to the layer::getName()

```
getName() is independent of const, so this patch adds const to
layer::getName();

**Self evaluation:**

```

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -133,7 +133,7 @@ public:
    * @note      This name might be changed once this layer is added to the model
    * to keep the name unique to the model
    */
-  virtual std::string getName() noexcept = 0;
+  virtual const std::string getName() const noexcept = 0;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


api/ccapi/include/layer.h
```diff
@@ -133,7 +133,7 @@ class Layer {
    * @note      This name might be changed once this layer is added to the model
    * to keep the name unique to the model
    */
-  virtual std::string getName() noexcept = 0;
+  virtual std::string getName() const noexcept = 0;
```
kparichay, line 136: as long we are updating, how about 
```suggestion
  virtual const std::string &getName() const noexcept = 0;
```

kparichay: just const is fine.

kparichay (approved):
LGTM now

api/ccapi/include/layer.h
zhoonit: Done

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1294, opened 2021-06-16
[Properties] Add float and boolean cases

zhoonit:
**[Pending commit: #1292]**

- [Properties] Add float and boolean cases

```
Add float and boolean case to be used later.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/utils/base_properties.cpp
+++ b/nntrainer/utils/base_properties.cpp
@@ -33,6 +33,23 @@ std::string str_converter<str_prop_tag, std::string>::from_string(
 }
 
 template <>
+std::string str_converter<bool_prop_tag, bool>::to_string(const bool &value) {
+  return value ? "true" : "false";
+}
+
+template <>
+bool str_converter<bool_prop_tag, bool>::from_string(const std::string &value) {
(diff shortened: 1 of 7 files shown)


nntrainer/utils/base_properties.cpp
```diff
@@ -32,6 +32,23 @@ std::string str_converter<str_prop_tag, std::string>::from_string(
   return value;
 }
 
+template <>
+std::string str_converter<bool_prop_tag, bool>::to_string(const bool &value) {
```
kparichay, line 36: shouldn't this also throw if value is empty? if a property is empty, we dont want to save it. or this will be handled somewhere else? My concern is if a prop is empty, then it should not be part of saved properties.

zhoonit: If empty we throw at the property level. You might wanna double check at `base_properties.h:182`.

kparichay: great, thanks for the check.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1295, opened 2021-06-16
[manager] Add support for request Inputs/outputs

kparichay:
Added support for requesting inputs and outputs
Also update implementation to use a common function for easier
management

Signed-off-by: Parichay Kapoor <pk.kapoor@samsung.com>

PR #1296, opened 2021-06-16
[wait for #1295] [graph] Support creating RunLayerContext 

kparichay:
Implement the function to create RunLayerContext given the
InitLayerContext and update it in the given LayerNode.

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -681,16 +681,25 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
   }
 }
 
-void NetworkGraph::init2runContext(InitLayerContext &init_context,
-                                   RunLayerContext &run_context) {
-  // NOTE: this just create the wrappers and does not actually memory inside
-  // these wrappers
-  // TODO: create wrappers for weights - initialize is already done, so the
-  // object creation can be done
-  // TODO: create wrapper for the temporary tensors
-  // TODO: create wrappers for outputs
-  // TODO: create a new context with these new wrappers and then copy assign
-  // run_context
+void NetworkGraph::updateRunContext(std::shared_ptr<Manager> &manager,
+                                    const std::shared_ptr<LayerNode> &lnode) {
+  /**
+   * using copy assignment allows setting run_context without adding more
+   * interfaces
+   */
+  const GraphNode &gnode = *lnode.get();
+  const InitLayerContext &init_context = lnode->getInitContext();
+  /**
+   * @todo must use existing properties like name/trainable of run_context to
+   * create the new run_context
+   */
+  // const RunLayerContext &run_context = lnode->getRunContext();
+
+  lnode->updateRunContext(RunLayerContext(
+    manager->requestWeights(gnode, init_context.getWeightsSpec()),
+    manager->requestInputs(gnode, init_context.getInputDimensions()),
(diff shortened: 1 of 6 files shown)


nntrainer/layers/layer_node.h
```diff
@@ -307,6 +307,30 @@ class LayerNode : public ml::train::Layer, public GraphNode {
...
+  const RunLayerContext &getRunContext() const { return run_context; }
+
+  /**
+   * @brief   Set run layer context
+   *
+   * @param  context Updated run layer context
+   */
+  void updateRunContext(const RunLayerContext &context) {
+    // TODO: ensure props/trainable must match
+    run_context = std::move(context);
```
zhoonit, line 331: Since move is a casting operation, we need signature of `RunLayerContext &&context` or `RunLayerContext &context` (non-const)
Otherwise copy will be invoked.

nntrainer/graph/network_graph.cpp
```diff
@@ -681,16 +681,25 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
...
+   */
+  const GraphNode &gnode = *lnode.get();
+  const InitLayerContext &init_context = lnode->getInitContext();
+  /**
+   * @todo must use existing properties like name/trainable of run_context to
+   * create the new run_context
+   */
+  // const RunLayerContext &run_context = lnode->getRunContext();
+
+  lnode->updateRunContext(RunLayerContext(
```
zhoonit, line 698: Update sounds like it's updating part of the context. how about `swapRunContext` ?

zhoonit: or how about sth like moving `creatingRunContext to initContext`?
```cpp
RunContext InitContext::CreateRunContext(Manager manager);
void LayerNode::initRunContext(Manager manager) {
   this->runContext = initContext.CreateRunContext(manager) 
}
```

nntrainer/layers/layer_node.h
kparichay: > Otherwise copy will be invoked.

can copy be invoked? as there is unique_ptr inside and that will give compile error.

nntrainer/graph/network_graph.cpp
kparichay: > Update sounds like it's updating part of the context. how about `swapRunContext` ?

Actually it is returning updated RunContext. RunContext does not has any interface to update it (to protect from users). So, only way to update is new RunContext.

kparichay: aah, i understand the confusion. This interface is temporary, and updated in #1301. Please check it once.

nntrainer/layers/layer_node.h
zhoonit: https://wandbox.org/permlink/JtK4a4npPR3bXtYc check this out.

It seems like unique_ptr does not exist inside context and nor move ctor/assign operator as well.

nntrainer/graph/network_graph.cpp
zhoonit: Thanks :)

nntrainer/layers/layer_node.h
kparichay: quite strange. Updated to `RunLayerContext &&context`.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

Issue #1297, opened 2021-06-16
[Proposal] std::any Layer::getProperty(string);


zhoonit:
For now, we have some (theoretical) properties frequently queried from other objects.
For example, flatten, trainable, supportInPlace, requireLabels are those kind of stuff.

# As is
To support this, we are making one api each to query this. But this can be problemetic, if we are considering the list is keep growing. and If we add a new API, we don't want to update the version. (or think about tizen capi)

# To Be

Move this to a layer property and make it available in LayerImpl level. When querying, we use std::any.

```cpp
class ActivationLayer: LayerImpl {
   ActivationLayer() : LayerImpl({"trainable=false", "require_label=false"}): {
     /** ... */
   }
}; 

TEST(query_trainable_option1) {
 auto layer = ActivationLayer();
 std::any is_trainable = layer.getProperties("trainable"); // if trainable is NA or empty, this will be empty
 EXPECT_NE(false, is_trainable.empty());
 bool &is_trainable_ = std::any_cast<std::reference_wrapper<bool>>(is_trainable).get();
 EXPECT_EQ(false, is_trainable_);
 is_trainable_ = true; // this will change original properties
}

TEST(query_trainable_option2) {
 auto layer = ActivationLayer();
 std::any is_trainable = layer.getProperties("trainable");
 EXPECT_EQ(false, std::any_cast<bool>(is_trainable).get()); // in this option, we cannot modify variable
 is_trainable = true; // this won't change the original properties
}
```

kparichay:
With this change, developers will have to use c++17 for their custom layers. Although our code is c++17, we did not force custom developers to also c++17 but with std::any it will.
How about starting with a simplified signature of returning bool/int (it currently satisfied minimal requirement), and add std::any with for developers who have c++17?

zhoonit:
We could consider void * (which c api will anyway)
I'll start with bool for now.

kparichay:
Yes, we can provide `void *`, which internally depends on `std::any`, and `std::any` version can be directly used if the developer uses c++17.

PR #1298, opened 2021-06-17
[Layer] Add trainable prop

zhoonit:
- [Layer] Add trainable prop

```
This patch add trainable property to layer

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/layer_impl.cpp
+++ b/nntrainer/layers/layer_impl.cpp
@@ -15,6 +15,7 @@
 #include <string>
 #include <vector>
 
+#include <common_properties.h>
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <node_exporter.h>
@@ -23,7 +24,7 @@ namespace nntrainer {
 
 LayerImpl::LayerImpl() :
   finalized(false),
-  layer_impl_props(std::make_unique<std::tuple<>>()) {}
+  layer_impl_props(std::make_unique<std::tuple<props::Trainable>>()) {}
 
 void LayerImpl::finalize(InitContext &context) {
   NNTR_THROW_IF(finalized, nntrainer::exception::not_supported)
@@ -32,7 +33,9 @@ void LayerImpl::finalize(InitContext &context) {
   finalized = true;
 }
 
-void LayerImpl::setProperty(const std::vector<std::string> &values) {}
+void LayerImpl::setProperty(const std::vector<std::string> &values) {
+  loadProperties(values, *layer_impl_props);
+}
 
 void LayerImpl::exportTo(Exporter &exporter,
                          const ExportMethods &method) const {}
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -67,6 +67,22 @@ public:
 };
 
 /**
+ * @brief trainable property, use this to set and check how if certain layer is
+ * trainable
+ *
+ */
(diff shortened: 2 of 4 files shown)


nntrainer/layers/layer_impl.cpp
```diff
@@ -23,7 +24,7 @@ namespace nntrainer {
 
 LayerImpl::LayerImpl() :
   finalized(false),
-  layer_impl_props(std::make_unique<std::tuple<>>()) {}
```
kparichay: props trainable will be in LayerNode and not in LayerImpl, and will be accessible to Layer via Init/Run context.

zhoonit: Thought about it, there are some issue

1. What if a layer is not trainable by default, or must be always non trainable?
2. When we request tensor, only layer knows which layer should be trainable or non-trainable
    eg) Even if we set trainable, weight1 and weight2 is trainable but weight3, weight4 must stay non-trainble
          Even if we set non-trainable, weight1 must be trainable.

zhoonit: My thinking is that layer has the boolean trainable but when request weight set true or false respectively.

kparichay: > Thought about it, there are some issue
> 
> 1. What if a layer is not trainable by default, or must be always non trainable?

yeah, there are pros and cons. To support `always non-trainable`, there will yet another function in layer devel like `supportTraining()` if trainable in LayerNode. If in LayerImpl, we need `getTrainable` or `getProperty` now. Both fine choices, left up to you.

> 2. When we request tensor, only layer knows which layer should be trainable or non-trainable
>    eg) Even if we set trainable, weight1 and weight2 is trainable but weight3, weight4 must stay non-trainble
>    Even if we set non-trainable, weight1 must be trainable.

yes, that must be controlled by the layer. When layer requestWeights() by context, layer gets full control of this. This true/false for each weight must be `and` with the trainable of the layer, and then correspondingly set to weight (not yet done, just my thought).

kparichay: If you choose to keep trainable in LayerImpl, please update LayerNode correspondingly as well (remove trainable from there).

zhoonit: Thanks for the suggestion, I'll update within this PR.

Could you also take a look over #1297 regarding this?

kparichay: done.

zhoonit: > I'll update within this PR.

trainable deleted from props

kparichay (approved):
LGTM

nntrainer/layers/layer_impl.cpp
jijoongmoon: >When we request tensor, only layer knows which layer should be trainable or non-trainable
eg) Even if we set trainable, weight1 and weight2 is trainable but weight3, weight4 must stay non-trainble
Even if we set non-trainable, weight1 must be trainable.

I wonder whether we are going to set the trainable in weight ??

kparichay:
Discussed offline, trainable will later move to LayerNode.
Lets merge this PR for now.

nntrainer/layers/layer_impl.cpp
zhoonit: Yes, we are setting trainable in weights and that won't be changing.

jijoongmoon (approved):
LGTM

PR #1301, opened 2021-06-17
[networkgraph] Network graph updated for Layer V2

kparichay:
Network graph updated to work with LayerV2
This involves settings input dimension in InitContext, and setting
up RunContext for each layer before their execution.

Further, some helper functions are also added in LayerNode.

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -379,7 +379,7 @@ int NetworkGraph::checkCompiledGraph() {
   /** Dimension of input layers must be known */
   for (auto iter = cbegin(); iter != cend(); iter++) {
     auto lnode = (*iter);
-    if (lnode->getObject()->getType() == InputLayer::type) {
+    if (lnode->getType() == InputLayer::type) {
       if (lnode->getInputDimensions().size() == 0) {
         ml_loge("InputDimension of first layer is not set");
         return ML_ERROR_INVALID_PARAMETER;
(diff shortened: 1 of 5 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -729,31 +750,67 @@ int NetworkGraph::initialize(std::shared_ptr<Manager> manager) {
...
     NN_RETURN_STATUS();
-
-    // TODO: call init2runContext
+#endif
+
+#if LAYER_V2
+    std::vector<Var_Grad *> inputs = {};
+    if (cur_type != InputLayer::type && idx != 0) {
+      if (input_map.find(lnode->getName()) == input_map.end())
+        throw std::runtime_error("Cannot find inpuy buffers for the node");
```
zhoonit, line 778: s/inpuy/input?

kparichay: done.

nntrainer/graph/network_graph.cpp
```diff
@@ -681,36 +681,57 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
...
+#if !LAYER_V2
+    auto &lptr = lnode->getObject();
+#endif
 
     /**
      * Set input dimension for all the layers.
      * For input layer, as input dimension is known, set input tensor.
      */
-    if (!first) {
+    if (cur_type != InputLayer::type && idx != 0) {
```
zhoonit, line 731: how about making this to a lambda? this logic is used in number of places `cur_type != InputLayer::type && idx != 0` but likely to change sometime

nntrainer/graph/network_graph.cpp
```diff
@@ -681,36 +681,57 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
...
       std::string l_pre_type = getSortedLayerNode(idx - 1)->getType();
```
zhoonit, line 732: how about making this to a lambda? this logic is used in number of places `cur_type != InputLayer::type && idx != 0` but likely to change sometime

nntrainer/graph/network_graph.cpp
```diff
@@ -681,36 +681,57 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
...
+    outputs, manager->requestTensors(gnode, init_context.getTensorsSpec())));
+
+  return outputs;
 }
 
 int NetworkGraph::initialize(std::shared_ptr<Manager> manager) {
   int status = ML_ERROR_NONE;
-  // TODO: don't delete adj list - use it here and make this more cleaner/faster
+  /** this contains the map from name to input tensors for each node */
+  std::unordered_map<std::string, std::vector<Var_Grad *>> input_map;
```
zhoonit, line 716: Nice, I guess this can directly replace tfliteNode sometime later

kparichay: yes, that will be good :)

nntrainer/graph/network_graph.cpp
```diff
@@ -681,45 +681,57 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
...
+    auto &lptr = lnode->getObject();
+#endif
 
     /**
      * Set input dimension for all the layers.
      * For input layer, as input dimension is known, set input tensor.
      */
-    if (!first) {
+    if (cur_type != InputLayer::type && idx != 0) {
       std::string l_pre_type = getSortedLayerNode(idx - 1)->getType();
```
zhoonit, line 732: I think there is some connection issue, my comment was gone.
--- From my brain memory:
Not related to the PR but pre layer is not the parent layer, so maybe adding todo here?

nntrainer/graph/network_graph.cpp
kparichay: will add.

nntrainer/graph/network_graph.cpp
kparichay: this variable is removed in the next PR #1302

zhoonit: Cool

nntrainer/graph/network_graph.cpp
```diff
@@ -696,45 +696,64 @@ void NetworkGraph::inPlaceOptimize(Manager &manager) {
...
   /**
    * using copy assignment allows setting run_context without adding more
    * interfaces
    */
   const GraphNode &gnode = *lnode.get();
   const InitLayerContext &init_context = lnode->getInitContext();
+  std::vector<Var_Grad *> inputs = prev_inputs;
+  if (inputs.empty())
+    inputs = manager->requestInputs(gnode, init_context.getInputDimensions());
+
```
jijoongmoon, line 712: It looks like using A output as an input of B in A-->B situation. is it enough just set like this? or are you going to optimize this?

kparichay: > It looks like using A output as an input of B in A-->B situation. 

Yes, this marks sharing the Var_Grad wrappers. This does not include the optimization of the shared tensors. That is done by the manager.

> is it enough just set like this? or are you going to optimize this?

Yes, this is enough to automatically redirect the output of layer A to the input of layer B.

jijoongmoon (approved):
LGTM

PR #1302, opened 2021-06-17
[network] Remove check for double activation

kparichay:
Remove the check for double activation, it is now allowed to have
two activation operations one after the other.


zhoonit (approved):
It is related to optimization right? LGTM

kparichay:
> It is related to optimization right? LGTM

No, this is related to remove an extra restriction we had added (not allowing two activation layers consecutive in the graph).

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

PR #1305, opened 2021-06-17
[manager] Support initialization/allocation of weights 

kparichay:
Support initialization and allocation of weights for LayerV2

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -719,6 +719,7 @@ NetworkGraph::updateRunContext(std::shared_ptr<Manager> &manager,
    */
   // const RunLayerContext &run_context = lnode->getRunContext();
   lnode->updateRunContext(RunLayerContext(
+    // TODO: update weights spec for trainable based on layer trainable prop
     manager->requestWeights(gnode, init_context.getWeightsSpec()), inputs,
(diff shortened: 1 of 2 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -704,6 +704,7 @@ NetworkGraph::updateRunContext(std::shared_ptr<Manager> &manager,
    */
   // const RunLayerContext &run_context = lnode->getRunContext();
   lnode->updateRunContext(RunLayerContext(
+    // TODO: update weights spec for trainable based on layer trainable prop
```
zhoonit, line 707: Shouldn't this must be done inside getWeightsSpec already before come to this spot.

zhoonit (approved):
LGTM

nntrainer/graph/network_graph.cpp
kparichay: It depends on when we allow to change trainable property. Maybe this should be done in the end of after finalize() in layer_node.h

kparichay: I will add its support in #1312 or its next PR.

zhoonit: Okay

jijoongmoon (approved):
LGTM

PR #1306, opened 2021-06-17
[AppContext] Integrate layer-devel @open sesame 06/22 16:38

zhoonit:
- [AppContext] Integrate layer-devel

```
This patch integrates layer devel to appcontext and plugin features.

From this patch, having it is able to include LayerV1 and Layer at the
same time.

Also adding a test

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/layer.cpp
+++ b/nntrainer/layers/layer.cpp
@@ -363,7 +363,7 @@ void LayerV1::print(std::ostream &out, unsigned int flags) {
   }
 };
 
-std::shared_ptr<LayerV1> getLayerDevel(std::shared_ptr<ml::train::Layer> l) {
+std::shared_ptr<LayerV1> getLayerV1Devel(std::shared_ptr<ml::train::Layer> l) {
   return std::static_pointer_cast<LayerNode>(l)->getObject();
 }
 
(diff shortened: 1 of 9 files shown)


kparichay (approved):
LGTM

nntrainer/layers/layer.cpp
```diff
@@ -363,7 +363,7 @@ void LayerV1::print(std::ostream &out, unsigned int flags) {
   }
 };
 
-std::shared_ptr<LayerV1> getLayerDevel(std::shared_ptr<ml::train::Layer> l) {
+std::shared_ptr<LayerV1> getLayerV1Devel(std::shared_ptr<ml::train::Layer> l) {
```
jijoongmoon, line 366: I quite dont understand why it needs getLayerDevel?? is there any reason?? If it is getLayerDevel.. then do we have getLayer as well?? and does it return different??

kparichay: This is for internal use only. This is a wrapper function for LayerNode::getObject() and this will be removed once getObject() is removed.

zhoonit: I think this is a temporary code when we were building `getLayerDevel`
@kparichay any opinions?

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1308, opened 2021-06-17
[Wait for #1306] [LayerNode] Add layer(devel) to the node

zhoonit:
**[Pending commit: #1306]**

- [LayerNode] Add layer(devel) to the node

```
- Add `Layer` to the node
- Remove `node_exporter.h` from `layer_node.h`
- Change compile time switch to runtime switch of checking layer v1 and
layer v2

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -16,18 +16,37 @@
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <time_dist.h>
+
+#include <common_properties.h>
 namespace nntrainer {
 
 LayerNode::LayerNode(std::shared_ptr<nntrainer::LayerV1> l, size_t idx) :
-  layer(l),
+  LayerNode(nullptr, l, idx) {}
+
+LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l, size_t idx) :
+  LayerNode(std::move(l), nullptr, idx) {}
+
+LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&layer_v2,
+                     std::shared_ptr<nntrainer::LayerV1> layer_v1, size_t idx) :
+  layerv1(layer_v1),
+  layer(std::move(layer_v2)),
   index(idx),
   flatten(false),
   distribute(false),
   activation_type(ActivationType::ACT_NONE),
-  props(props::Name()) {
-  if (layer->getType() == TimeDistLayer::type)
+  props(new std::tuple<props::Name>(props::Name())) {
+  if (layerv1 && layerv1->getType() == TimeDistLayer::type) {
     distribute = true;
+  } else if (layer && layer->getType() == TimeDistLayer::type) {
+    distribute = true;
+  }
 }
+
+/**
+ * @brief Destroy the Layer Node object
+ *
+ */
+LayerNode::~LayerNode() = default;
 
 /**
  * @brief Layer factory creator with constructor
@@ -92,7 +111,7 @@ void LayerNode::setProperty(const nntrainer::LayerV1::PropertyType type,
   switch (type) {
   case PropertyType::name:
     if (!value.empty()) {
-      std::get<props::Name>(props).set(value);
+      std::get<props::Name>(*props).set(value);
     }
     break;
   case PropertyType::flatten:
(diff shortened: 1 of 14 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -16,19 +16,38 @@
...
+  layer(std::move(layer_v2)),
   index(idx),
   flatten(false),
   distribute(false),
   activation_type(ActivationType::ACT_NONE),
-  props(props::Name()) {
-  if (layer->getType() == TimeDistLayer::type)
+  props(new std::tuple<props::Name>(props::Name())) {
+  if (layerv1 && layerv1->getType() == TimeDistLayer::type) {
+    distribute = true;
```
kparichay, line 39: please also add a check that layer inside TimeDistLayer is also valid.

nntrainer/layers/layer_node.h
```diff
@@ -337,7 +364,8 @@ class LayerNode : public ml::train::Layer, public GraphNode {
    * These properties are set for the layer by the user but are intercepted
    * and used in the node which forms the basic element of the graph.
    */
-  std::tuple<props::Name> props; /**< properties for the layer node */
+  std::unique_ptr<std::tuple<props::Name>>
```
kparichay, line 367: does this need unique_ptr?

nntrainer/layers/layer_node.h
```diff
@@ -309,11 +332,15 @@ class LayerNode : public ml::train::Layer, public GraphNode {
   friend std::ostream &operator<<(std::ostream &out, const LayerNode &l);
 
 private:
-  // TODO: make this unique_ptr once getObject API is removed
```
kparichay, line 311: add this TODO to getObject()?

nntrainer/layers/layer_node.h
```diff
@@ -309,11 +332,15 @@ class LayerNode : public ml::train::Layer, public GraphNode {
...
+  /// @todo: remove this
   std::shared_ptr<nntrainer::LayerV1>
+    layerv1; /**< The actual object in the graph node */
+
+  std::unique_ptr<nntrainer::Layer>
```
kparichay, line 339: how about a union of layerv1 and layer? both should not exist together.

nntrainer/layers/layer_node.cpp
```diff
@@ -212,8 +236,18 @@ void LayerNode::updateInputLayers(const unsigned int idx,
   input_layers[idx] = to;
 }
 
+void LayerNode::exportTo(Exporter &exporter,
+                         const ExportMethods &method) const {
+  exporter.saveResult(*props, method, this);
+  if (layerv1 == nullptr) {
+    /// have layer_v2 implementation
+  } else {
+    layerv1->export_to(exporter, method);
```
kparichay, line 245: `getLayer()->exportTo`

nntrainer/layers/layer_node.h
zhoonit: This is to pimple the props, unique_ptr allows incomplete type. Props are quite heavy to compile but overhead from indirection is negligible.

nntrainer/layers/layer_node.h
zhoonit: I was thinking `std::variant` but even in this case we need to check which one exist so,  we have to add another member, I though this is not a big advantage.

nntrainer/layers/layer_node.h
zhoonit: Also added to `getObject()`

nntrainer/layers/layer_node.cpp
zhoonit: Done

nntrainer/layers/layer_node.h
kparichay: sounds good.

nntrainer/layers/layer_node.h
kparichay: We will need some `finalized` kind of variable to track which context to pass to `setBatch()` (check #1312) which sits on top of this PR itself :).

zhoonit: Yeah, this will be deleted anyway so I guess it's fine, from the user's perspective they are only able to set only one regulated from the constructor.

kparichay: cool.

kparichay (approved):
LGTM

nntrainer/layers/layer_node.h
jijoongmoon: is it ok when we need to copy the properties?

zhoonit: Yes, copy is implemented for properties.

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1310, opened 2021-06-17
[LayerNode] Change flatten and distribute to prop

zhoonit:
- [LayerNode] Change flatten and distribute to prop

```
- Fix bug on `loadProperties`
- Add flatten and distribute to prop

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -17,8 +17,38 @@
 #include <nntrainer_log.h>
 #include <time_dist.h>
 
+#include <base_properties.h>
 #include <common_properties.h>
 namespace nntrainer {
+
+namespace props {
+class ActivationType;
+
+/**
+ * @brief Flatten property, true if needs flatten layer afterwards
+ */
+class Flatten : public Property<bool> {
+public:
+  Flatten() : Property<bool>() {}               /**< has default value of 0 */
+  static constexpr const char *key = "flatten"; /**< unique key to access */
+  using prop_tag = bool_prop_tag;               /**< property type */
+};
+
+/**
+ * @brief Distribute property, true if it distribute across layer
+ *
+ */
+class Distribute : public Property<bool> {
+public:
+  Distribute() : Property<bool>() {}
+  static constexpr const char *key = "distribute";
+  using prop_tag = bool_prop_tag;
+  bool isValid(const bool &v) const {
+    return empty() || !get();
+  } /**< distribute=true can be set strictly one time */
+};
+
+} // namespace props
 
 LayerNode::LayerNode(std::shared_ptr<nntrainer::LayerV1> l, size_t idx) :
   LayerNode(nullptr, l, idx) {}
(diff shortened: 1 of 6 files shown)


nntrainer/utils/node_exporter.h
```diff
@@ -309,10 +312,12 @@ loadProperties(const std::vector<std::string> &string_vector, Tuple &&props) {
   auto callable = [&left](auto &&prop, size_t index) {
     std::string prop_key = getPropKey(prop);
 
-    for (auto iter = left.begin(); iter < left.end(); ++iter) {
+    for (auto iter = left.begin(); iter < left.end();) {
       if (istrequal(prop_key, iter->first) == true) {
         from_string(iter->second, prop);
         iter = left.erase(iter);
+      } else {
+        iter++;
```
kparichay, line 320: good catch

nntrainer/layers/layer_node.h
```diff
@@ -162,13 +164,13 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    * @brief     get if the output of this layer must be flatten
    * @retval    flatten value
    */
-  bool getFlatten() const { return flatten; }
+  bool getFlatten() const;
```
kparichay, line 167: noexcept as well?

nntrainer/layers/layer_node.cpp
```diff
@@ -95,69 +121,53 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
...
+      auto &ac = nntrainer::AppContext::Global();
+      std::shared_ptr<nntrainer::LayerV1> dlayer =
+        ac.createObject<nntrainer::LayerV1>(TimeDistLayer::type);
+      std::static_pointer_cast<TimeDistLayer>(dlayer)->setDistLayer(layerv1);
+      layerv1 = dlayer;
+    }
+  }
+
+  if (layerv1 == nullptr) {
+    layer->setProperty(remainder);
```
kparichay, line 143: this should also be getLayer->setProperty()?

nntrainer/layers/layer_node.cpp
```diff
@@ -95,69 +121,53 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
...
+  } else {
+    auto &l = getLayer();
+    return l->setProperty(remainder);
```
kparichay, line 146: why is the behavior different than LayerV2?

zhoonit: V2 is void operation, V1 returns int.

zhoonit: So, if line 143 fails, it will throw

nntrainer/layers/layer_node.h
zhoonit: Done

nntrainer/layers/layer_node.cpp
kparichay: aah got it.

kparichay (approved):
LGTM

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.cpp
```diff
@@ -16,19 +16,67 @@
...
 namespace nntrainer {
 
+namespace props {
+class ActivationType;
+
+/**
+ * @brief Flatten property, true if needs flatten layer afterwards
+ */
+class Flatten : public Property<bool> {
+public:
```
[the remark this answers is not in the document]
zhoonit: Yes, this is here because the property should never be used outside of layer_node.

I wonder if we should not allow making properties for each object, but only inside `common_properties.h`?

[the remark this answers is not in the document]
zhoonit: > I wonder if we should not allow making properties for each object, but only inside common_properties.h?

If we decide to do this, I'll do it as a separate PR. There are couple of other properties need moving :)

PR #1311, opened 2021-06-18
[manager] Memory allocation for non-weight tensors 

kparichay:
Added memory allocation for non-weight tensors has been added
including inputs, outputs and tensors.
For now, this does include any optimization except the gradient
based optimization.

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -308,6 +308,7 @@ void Manager::allocateWeights() {
       }
     }
   }
+
   weights_allocated = true;
 }
 
@@ -326,6 +327,7 @@ void Manager::deallocateWeights() {
       }
     }
   }
+
   weights_allocated = false;
 }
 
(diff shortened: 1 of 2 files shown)


zhoonit (approved):
LGTM

nntrainer/tensor/manager.cpp
```diff
@@ -564,41 +565,68 @@ void Manager::initializeTensorsInference() {
...
+      for (auto &ts : layer_ts) {
+        ts->initialize(Tensor(), Tensor(), false);
+      }
+    }
 
     // In inference mode, do not allocate the memory for the input of the first
     // layer. These is the first entry in the in_outs. Inference() will override
     // input tensors of the first layer
-    if (is_first_layer)
+    for ([[maybe_unused]] auto &layer_ins : inputs_v2) {
```
lhs8928, line 625: Do we need this?

lhs8928 (approved):
LGTM

nntrainer/tensor/manager.cpp
kparichay: This is an empty loop to tell the reviewer that the developer has intentionally not allocated memory for inputs_v2. This code section can be commented, but the risk is that it can be deleted in some cleanup.

zhoonit: This silents the compiler warning as well.

jijoongmoon (approved):
LGTM

PR #1314, opened 2021-06-18
[Example] How to save and reuse model

zhoonit:
- [Example] How to save and reuse model

```
Please do not merge, this is demonstrative purpose only

```

### DIFF
--- a/Applications/SimpleShot/task_runner.cpp
+++ b/Applications/SimpleShot/task_runner.cpp
@@ -13,6 +13,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <neuralnet.h>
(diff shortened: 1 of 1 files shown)


Applications/SimpleShot/task_runner.cpp
```diff
@@ -286,6 +289,29 @@ int main(int argc, char **argv) {
     return 1;
   }
 
+  model->saveModel(); // this saves to save_path
+
+  ////////// when inferencing in different code ///////////////
+
```
zhoonit, line 295: I am assuming here that model is created and not reused from above code.
```cpp
std::unique_ptr<Model> model = created_for_inference();
```

PR #1317, opened 2021-06-21
[layernode/impl] Add properties to Layer Node and Impl 

kparichay:
1. Add input_dim property to layer node
2. Add weights/bias related properties such as initializer, regularizer,
etc to LayerImpl. This creates a differentiating factor of LayerImpl
from LayerDevel as LayerImpl provides a base class for layers with
weights/bias properties loaded.

### DIFF
--- a/nntrainer/layers/layer_impl.cpp
+++ b/nntrainer/layers/layer_impl.cpp
@@ -4,7 +4,7 @@
  *
  * @file   layer_impl.h
  * @date   21 June 2021
- * @brief  This is base Optimizer implementation class
+ * @brief  This is base layer implementation class
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
  * @bug    No known bugs except for NYI items
@@ -16,6 +16,7 @@
 #include <vector>
 
 #include <common_properties.h>
+#include <layer_internal.h>
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <node_exporter.h>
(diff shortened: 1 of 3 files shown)


zhoonit (approved):
LGTM

nntrainer/layers/layer_node.h
```diff
@@ -331,7 +331,9 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    */
   void addInputLayers(const std::string &in_layer) {
     input_layers.push_back(in_layer);
-    layerv1->setNumInputs(input_layers.size());
+    input_dim.resize(input_layers.size());
```
lhs8928, line 334: So the input_dim is managed in layer_node and also in layer_context?

lhs8928 (approved):
LGTM

nntrainer/layers/layer_node.h
kparichay: That's a good question.
`input_dim` in `LayerNode` stores the intermediate value of `input_dim` till `finalize` is called. 
The reason to do this way is to avoid an interface in `InitLayerContext` which allows modifying `input_dim` to layer developers.
I am open to ideas for this. Please propose a new design for this, and I will update this PR :).

nntrainer/layers/layer_impl.cpp
```diff
@@ -16,6 +16,7 @@
 #include <vector>
 
 #include <common_properties.h>
+#include <layer_internal.h>
```
jijoongmoon, line 19: Could you check line 7: base Optimizer Implementation Class??

nntrainer/layers/layer_impl.cpp
```diff
@@ -35,6 +36,65 @@ void LayerImpl::finalize(InitLayerContext &context) {
...
+  }
+}
+
+void LayerImpl::setProperty(const std::string &type_str,
+                            const std::string &value) {
+  int status = ML_ERROR_NONE;
+  LayerV1::PropertyType type =
+    static_cast<LayerV1::PropertyType>(parseLayerProperty(type_str));
+
+  switch (type) {
```
jijoongmoon, line 66: Still, it is not clear, which property should go layer_node and which should go here and layer. I'm sure you will make doc but please keep in mind that there are developers who think like me.

nntrainer/layers/layer_impl.h
```diff
@@ -52,7 +53,7 @@ class LayerImpl : public Layer {
   /**
    * @brief     finalize the layer
    * @throw     nntrainer::not_supported if try to initialize twice
-   * @copydoc   Layer::fianlize(InitLayerContext &context)
+   * @copydoc   Layer::finalize(InitLayerContext &context)
```
jijoongmoon, line 56: I wonder that LayerImpl is the abstract class for all layers?? (Line 37) You mean the abstraction level of this LayerImpl is higher than Layer, I guess. It looks like developer can not make their own layer with only including this LayerImpl if they don't know about the layer_devel.h

nntrainer/layers/layer_impl.cpp
kparichay: done.

nntrainer/layers/layer_impl.cpp
kparichay: Added description about the LayerImpl in details section of LayerImpl header, which covers the help it provides and the properties it does. Added the similar kind of description in LayerNode.h header as well.

nntrainer/layers/layer_impl.h
kparichay: I am thinking that all the layers which have weights/bias will extend LayerImpl, while others will extend LayerDevel.
LayerImpl for now only provides support for Weight/bias-related properties, so layers like activation/flatten/etc can extend LayerDevel directly IMO. What do you advise?

kparichay: > It looks like developer can not make their own layer with only including this LayerImpl if they don't know about the layer_devel.h

Yes, that is absolutely right. The developer will have to look at LayerDevel. LayerImpl is also an abstract class.

PR #1319, opened 2021-06-21
[CustomLoss] Implement mae loss layer

zhoonit:
- [CustomLoss] Implement mae loss layer

```
This patch implements mae loss layer forward and backward, also moves
other functions to the sourcefile instead of inlinining it.

**Self evaluation:**

```

### DIFF
--- a/Applications/Custom/mae_loss.cpp
+++ b/Applications/Custom/mae_loss.cpp
@@ -12,8 +12,57 @@
  */
 #include "mae_loss.h"
 
+#include <cmath>
+
+#include <tensor.h>
+
+constexpr const float EPSILON_ = 1e-7;
 namespace custom {
-const std::string MaeLossLayer::type = "mae_loss";
+
+int MaeLossLayer::initialize(nntrainer::Manager &manager) {
+  output_dim = input_dim;
+  return ML_ERROR_NONE;
+}
+
+int MaeLossLayer::setProperty(std::vector<std::string> values) {
+  /// this implementation makes to pass the test, this will change soon.
+  return values.size();
+}
+
+void MaeLossLayer::forwarding(bool training) {
+  nntrainer::Tensor &label = net_hidden[0]->getGradientRef();
+  nntrainer::Tensor &predicted = net_input[0]->getVariableRef();
+  nntrainer::Tensor &output = net_hidden[0]->getVariableRef();
+
+  bool with_label = !label.uninitialized();
+  if (with_label) {
+    predicted.subtract(label, output);
+    /// make Tensor::abs instead and use it here
+    output.apply_i(fabs);
+  } else {
+    output.fill(predicted);
+  }
+}
+
+void MaeLossLayer::calcDerivative() {
+  nntrainer::Tensor &predicted = net_input[0]->getVariableRef();
+  nntrainer::Tensor &label = net_hidden[0]->getGradientRef();
+
+  nntrainer::Tensor &deriv = net_input[0]->getGradientRef();
+
+  /// This can be saved at MaeLossLayer::forwarding, but this is done here on
+  /// purpose for demonstration purpose
+  predicted.subtract(label, deriv);
+
+  deriv.apply_i([](float x) {
+    if (fabs(x) < EPSILON_) {
+      return 0.0f;
+    }
+    return x > 0 ? 1.0f : -1.0f;
+  });
+}
+
+bool MaeLossLayer::requireLabel() const { return true; }
 
 #ifdef PLUGGABLE
 
--- a/Applications/Custom/mae_loss.h
+++ b/Applications/Custom/mae_loss.h
@@ -23,7 +23,7 @@ namespace custom {
 
 /**
  * @brief A sample loss layer which calculates mean absolute error from output
- * @todo update this to LayerV1
+ * @todo update this to LayerV2
  *
  */
 class MaeLossLayer final : public nntrainer::LayerV1 {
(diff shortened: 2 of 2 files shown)


Applications/Custom/mae_loss.h
```diff
@@ -43,44 +43,37 @@ class MaeLossLayer final : public nntrainer::LayerV1 {
   using nntrainer::LayerV1::setProperty;
 
   /**
-   * @brief     set Property of layer, currently only "exponent is accepted"
+   * @brief     set Property of layer
```
kparichay, line 46: update todo on line 26 to update to LayerV2.

Applications/Custom/mae_loss.cpp
```diff
@@ -12,8 +12,54 @@
...
+
+void MaeLossLayer::forwarding(bool training) {
+  nntrainer::Tensor &label = net_hidden[0]->getGradientRef();
+  nntrainer::Tensor &predicted = net_input[0]->getVariableRef();
+  nntrainer::Tensor &output = net_hidden[0]->getVariableRef();
+
+  bool with_label = !label.uninitialized();
+  if (with_label) {
+    predicted.subtract(label, output);
+    output.apply_i([](float x) { return x > 0 ? x : -x; });
```
kparichay, line 38: I recommend making a `Tensor::abs` and then use it here.

Applications/Custom/mae_loss.cpp
```diff
@@ -12,8 +12,54 @@
...
+  nntrainer::Tensor &label = net_hidden[0]->getGradientRef();
+
+  nntrainer::Tensor &deriv = net_input[0]->getGradientRef();
+
+  /// This can be saved at MaeLossLayer::forwarding, but this is done here on
+  /// purpose for demonstration purpose
+  predicted.subtract(label, deriv);
+
+  deriv.apply_i([](float x) {
+    if (fabs(x) < 1e-7) {
```
kparichay, line 55: please declare a layer const `epsilon` and use here?

Applications/Custom/mae_loss.cpp
```diff
@@ -12,8 +12,54 @@
...
+
+void MaeLossLayer::calcDerivative() {
+  nntrainer::Tensor &predicted = net_input[0]->getVariableRef();
+  nntrainer::Tensor &label = net_hidden[0]->getGradientRef();
+
+  nntrainer::Tensor &deriv = net_input[0]->getGradientRef();
+
+  /// This can be saved at MaeLossLayer::forwarding, but this is done here on
+  /// purpose for demonstration purpose
+  predicted.subtract(label, deriv);
```
kparichay, line 52: actually its already saved at `net_hidden[0]->getVariableRef()`:)

zhoonit: Ahh, hidden is absolute value. So unfortunately we can't use :(

Applications/Custom/mae_loss.cpp
zhoonit: I'll add todo instead.

Applications/Custom/mae_loss.cpp
zhoonit: Updated.

Applications/Custom/mae_loss.h
zhoonit: Done

Applications/Custom/mae_loss.cpp
kparichay: aah, right.

kparichay (approved):
LGTM

Applications/Custom/mae_loss.h
```diff
@@ -23,7 +23,7 @@ namespace custom {
 
 /**
  * @brief A sample loss layer which calculates mean absolute error from output
- * @todo update this to LayerV1
+ * @todo update this to LayerV2
```
jijoongmoon, line 26: do developers have to include layer_internal.h ?? or layer_devel.h? If this is the example for the developer, then I think it must be the layer_devel.h. Could you check this? And do we need to include manager.h? AFAK, we are not going to release manager.h.

Applications/Custom/mae_loss.cpp
```diff
@@ -12,8 +12,57 @@
...
 #include "mae_loss.h"
 
+#include <cmath>
+
+#include <tensor.h>
+
+constexpr const float EPSILON_ = 1e-7;
 namespace custom {
-const std::string MaeLossLayer::type = "mae_loss";
+
```
jijoongmoon, line 21: when the developer use their own layer in API ( not ini file ), How they call it? do they have to use custom namespace as defined here?

Applications/Custom/mae_loss.cpp
```diff
@@ -12,8 +12,57 @@
...
+int MaeLossLayer::initialize(nntrainer::Manager &manager) {
+  output_dim = input_dim;
```
jijoongmoon, line 23: The parameter of this function, manager might be temporary.?

Applications/Custom/mae_loss.h
zhoonit: It is layer_devel, after the migration, this will be moving to including layer_devel :)

Applications/Custom/mae_loss.cpp
zhoonit: namspace is up to the developer.

To create...
1. Register *.so and createLayer('custom_type') (similar to nnstreamer sub-plugin) 
2. Directly creates a instance out of the class.

Please refer to the details on the following PR :) https://github.com/nnstreamer/nntrainer/pull/1320

Applications/Custom/mae_loss.cpp
zhoonit: Yes, manager will go away and context will take place.

Applications/Custom/mae_loss.cpp
jijoongmoon: thanks.

Applications/Custom/mae_loss.cpp
jijoongmoon: got it.

jijoongmoon (approved):
LGTM

Issue #954, opened 2021-02-16
Is it possible to create a custom loss layer?


Str4t0n:
We would like to create our custom loss layer, but it is currently not possible to do so (or we did not find a solution for it yet), because nntrainer automatically adds its own loss layer to the model.

<details> 
  <summary>We are looking for something like this: </summary>

```ini
[Model]
Type = NeuralNetwork
...
[input]
Type = input
....
....
[lossLayer]
Type = mycustomloss
```

```cpp
auto &app_context = nntrainer::AppContext::Global();
app_context.registerFactory(ml::train::createLossLayer<custom::MyCustomLoss>);
```
```cpp
// or isntead of inherited from the base layer class
class MyCustomLoss : public nntrainer::Layer { ... }

// we could derive from the loss layer
class MyCustomLoss : public nntrainer::LossLayer { ... }
```
</details>

So that the nntrainer will know that its not necessary to add its own loss layer at the end .

zhoonit:
As of now, this is not directly supported although it will be with a abstractLossLayer to derive from.

Haven't tested but Below workaround should work.

<details>
<summary>implementing custom losslayer </summary>

1. apply #934 
1. implementing custom loss layer with type "loss"

```cpp
namespace custom  {
/// you could also derive from nntrainer::LossLayer
class MyCustomLoss : public nntrainer::Layer { 
  
  void forwarding(bool training) override {
     const Tensor &predicted = net_input[0]->getVariableRef(); // feature from the network can be accessed from here
     const Tensor &label = net_hidden[0]->getGradientRef(); // label can be accessed from here

    Tensor &hidden = net_hidden[0]->getVariableRef(); // loss calculated should go here

   /// do something
  }

  void calcDerivative() override {
     Tensor &ret_derivative = net_input[0]->getGradientRef(); // calculate derivative to this variable
     /// do something
  }
  const std::string getType() const override { return "loss"; }; // this hack the mechanism to detect loss layer.
}
}
```

</details>

<details>
<summary>using custom losslayer </summary>

1. do not add custom defined losslayer in the ini configuration

```ini
[Model]
Type = NeuralNetwork 
# don't type any loss here
...
[input]
Type = input
....
....
# [lossLayer] don't attach loss layer here
# Type = mycustomloss
```

2. do not register to the context

```cpp
// as type clashes, you won't be able to register the custom layer
// auto &app_context = nntrainer::AppContext::Global();
[...]

PR #1320, opened 2021-06-21
[CustomLoss] Update example

### LINKED ISSUE #1256: Custom Loss design and tests
- customLoss will be just like a regular custom layer
- For multiple losses, a constant value scalar for each loss needed
- requireLabel() in layer interface - then no  layer ahead of it in the graph

This is a subpart of #986

zhoonit:
- [CustomLoss] Update example

```
- Update example using LayerClient
- Add LayerClient example to AppTest

**Self evaluation:**

```

resolves #1256, #954

### DIFF
--- a/Applications/Custom/LayerClient/res/custom_layer_client.ini
+++ b/Applications/Custom/LayerClient/res/custom_layer_client.ini
@@ -3,8 +3,8 @@
 [Model]
 Type = NeuralNetwork
 Epochs = 100
-Loss = cross
 batch_size = 10
+# loss = cross  ###### Providing loss at model section should not be done
 
 [Optimizer]
 Type = sgd
@@ -26,3 +26,9 @@ Type = fully_connected
 unit = 10
 Bias_initializer = zeros
 Activation = softmax
+
+# Below is the custom loss layer,
+# This will directly replace loss in the Model section.
+# Thus, there should be no loss provided in the model section.
+[losslayer]
+Type = mae_loss
--- a/Applications/Custom/LayerClient/jni/Android.mk
+++ b/Applications/Custom/LayerClient/jni/Android.mk
@@ -62,7 +62,7 @@ LOCAL_ARM_MODE := arm
 LOCAL_MODULE := nntrainer_layer_client_example
 LOCAL_LDLIBS := -llog -landroid
 
-LOCAL_SRC_FILES := main.cpp $(LOCAL_PATH)/../../pow.cpp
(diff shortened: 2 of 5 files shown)


Applications/Custom/LayerClient/res/custom_layer_client.ini
```diff
@@ -26,3 +25,6 @@ Type = fully_connected
 unit = 10
 Bias_initializer = zeros
 Activation = softmax
+
+[losslayer]
+Type = mae_loss # custom loss layer
```
kparichay, line 30: i wonder if `loss = mae_loss` in `model` section also works?

kparichay: If yes, add as a comment here? if no, then it needs debugging.

zhoonit: The design here is that
1. Custom loss is not supported from model section(except the traditional basic ones)
2. We support loss as normal layers that is `requireLabel() == true`

To support multi loss, we aggregates loss from `requireLabel() == true`

So it is no bug but intended

kparichay: > Custom loss is not supported from model section(except the traditional basic ones)

What about a custom loss from .so?

zhoonit: Either from .so or directly from there source code which implements layer_devel.h

I'll update the comment inside ini file to cover the scenario you mentioned

kparichay (approved):
LGTM

Applications/Custom/LayerClient/res/custom_layer_client.ini
```diff
@@ -26,3 +26,9 @@ Type = fully_connected
 unit = 10
 Bias_initializer = zeros
 Activation = softmax
+
+# Below is the custom loss layer,
+# This will directly replace loss in the Model section.
+# Thus, there should be no loss provided in the model section.
+[losslayer]
+Type = mae_loss
```
jijoongmoon, line 34: I'm sure it it. But just check if there is type check between custom::mae_loss and mae_loss.

zhoonit: Yes there is, if mae_loss exist, custom::mae_loss will fail to register. so the developer of mae_loss must have to choose different type name 👍

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

PR #1321, opened 2021-06-22
[ Recurrent ] Implement Dropout for Recurrent Net

jijoongmoon:
In this commit, drop out for recurrent network is intrduced.
dropout property is introduced and if the element value of random
tensor is smaller than dropout rate, then it will be set zero.
The element which is not set zero, then it will scale with
1.0/(1.0-dropout).

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/gru.cpp
+++ b/nntrainer/layers/gru.cpp
@@ -161,6 +161,12 @@ void GRULayer::setProperty(const PropertyType type, const std::string &value) {
       throw_status(status);
     }
     break;
+  case PropertyType::dropout:
+    if (!value.empty()) {
+      status = setFloat(dropout_rate, value);
+      throw_status(status);
+    }
+    break;
   default:
     LayerV1::setProperty(type, value);
     break;
@@ -207,6 +213,10 @@ void GRULayer::forwarding(bool training) {
     for (unsigned int t = 0; t < islice.height(); ++t) {
       Tensor xs =
         islice.getSharedDataTensor({islice.width()}, t * islice.width());
+
+      if (dropout_rate > 0.0 && training) {
+        xs.multiply_i(xs.dropout_mask(dropout_rate));
+      }
       hs = oslice.getSharedDataTensor({oslice.width()}, t * oslice.width());
       Tensor zrg_t =
         zrg_.getSharedDataTensor({unit * NUM_GATE}, unit * t * NUM_GATE);
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -870,6 +870,24 @@ Tensor Tensor::transpose(const std::string &direction) const {
   return result;
 }
 
+Tensor Tensor::dropout_mask(float dropout) const {
+  Tensor result(dim);
+  result.setValue(1.0);
+  Tensor rand_temp(dim);
+  rand_temp.setRandUniform(0.0, 1.0);
+  float scale = 1.0 / (1 - dropout);
+
+  float *mask = result.getData();
+  float *random = rand_temp.getData();
+  for (unsigned int i = 0; i < length(); ++i) {
+    if (random[i] >= dropout)
+      mask[i] = mask[i] * scale;
+    else
+      mask[i] = 0.0;
+  }
+  return result;
+}
+
 int Tensor::apply_i(std::function<float(float)> f) {
   float *data = getData();
 
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -116,6 +116,7 @@ public:
    * - return_sequences : bool (type) - used only in lstm
    * - distribute : bool
    * - hidden_state_activation : string (type) - used only in lstm
+   * - drop_out : float (type) - drop out rate
    */
   /**
    * @brief     set Property of layer
--- a/nntrainer/layers/gru.h
+++ b/nntrainer/layers/gru.h
@@ -34,12 +34,13 @@ public:
     unsigned int unit_ = 0,
     ActivationType hidden_state_activation_type_ = ActivationType::ACT_NONE,
     ActivationType recurrent_activation_type_ = ActivationType::ACT_NONE,
-    bool sequence = false, Args... args) :
+    bool sequence = false, float dropout = 0.0, Args... args) :
     LayerV1(args...),
     unit(unit_),
     hidden_state_activation_type(hidden_state_activation_type_),
     recurrent_activation_type(recurrent_activation_type_),
-    return_sequences(sequence){};
+    return_sequences(sequence),
+    dropout_rate(dropout){};
 
   /**
    * @brief     Destructor of GRULayer
@@ -162,6 +163,11 @@ private:
    * @brief     variable to set return sequences
    */
   bool return_sequences;
+
+  /**
+   * @brief     drop out rate
+   */
+  float dropout_rate;
 };
 } // namespace nntrainer
 
--- a/nntrainer/layers/layer_internal.h
+++ b/nntrainer/layers/layer_internal.h
@@ -248,6 +248,7 @@ public:
    *            36. split_dimension : string (type)
    *            37. return_sequences :  bool (type) - lstm
    *            39. hidden_state_activation :  string (type) - lstm
+   *            40. dropout :  float (type) - drop out rate
    */
   enum class PropertyType {
     input_shape = 0,
@@ -289,6 +290,7 @@ public:
     split_dimension = 36,
     return_sequences = 37,
     hidden_state_activation = 38,
+    dropout = 39,
     unknown
   };
 
--- a/nntrainer/layers/lstm.cpp
+++ b/nntrainer/layers/lstm.cpp
@@ -146,6 +146,12 @@ void LSTMLayer::setProperty(const PropertyType type, const std::string &value) {
       throw_status(status);
     }
     break;
+  case PropertyType::dropout:
+    if (!value.empty()) {
+      status = setFloat(dropout_rate, value);
+      throw_status(status);
+    }
+    break;
   default:
     LayerV1::setProperty(type, value);
     break;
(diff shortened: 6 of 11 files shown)


kparichay (changes requested):
Please add dropout as a layer than adding it as a property of a layer. With the current design, each layer will have to inidividually implement dropout. But if dropout is a layer, then each layer can use dropout by attaching it in front of it. Please update the design.

jijoongmoon:
> Please add dropout as a layer than adding it as a property of a layer. With the current design, each layer will have to inidividually implement dropout. But if dropout is a layer, then each layer can use dropout by attaching it in front of it. Please update the design.

I think this should be differently. This pr only for the recurrent network. We can implement drop out layer as an formal layer later.

kparichay:
> > Please add dropout as a layer than adding it as a property of a layer. With the current design, each layer will have to inidividually implement dropout. But if dropout is a layer, then each layer can use dropout by attaching it in front of it. Please update the design.
> 
> I think this should deal differently. This pr only for the recurrent network. We can implement drop out layer as an formal layer later.

IMHO, it is a very small task to implement a dropout layer (less than half a day) but maintains extensibility and available for LSTM/GRU etc layers, as well as easier to debug and deal with. Given when dropout layer is added, this code will be removed, I think directly creating dropout layer will be better. I leave the final decision up to you.

nntrainer/layers/gru.cpp
```diff
@@ -207,6 +213,10 @@ void GRULayer::forwarding(bool training) {
     for (unsigned int t = 0; t < islice.height(); ++t) {
       Tensor xs =
         islice.getSharedDataTensor({islice.width()}, t * islice.width());
+
+      if (dropout_rate > 0.0 && training) {
```
kparichay, line 217: 0.0 -> epsilon (1e-6)

nntrainer/tensor/tensor.cpp
```diff
@@ -870,6 +870,24 @@ Tensor Tensor::transpose(const std::string &direction) const {
   return result;
 }
 
+Tensor Tensor::dropout_mask(float dropout) const {
+  Tensor result(dim);
+  result.setValue(1.0);
+  Tensor rand_temp(dim);
```
kparichay, line 876: can you cache this tensor from the caller, and give the mask to this function?

nntrainer/tensor/tensor.cpp
```diff
@@ -870,6 +870,24 @@ Tensor Tensor::transpose(const std::string &direction) const {
...
+  rand_temp.setRandUniform(0.0, 1.0);
+  float scale = 1.0 / (1 - dropout);
+
+  float *mask = result.getData();
+  float *random = rand_temp.getData();
+  for (unsigned int i = 0; i < length(); ++i) {
+    if (random[i] >= dropout)
```
kparichay, line 883: please create two loops - one which clips and value to 0/1 and the other which calls element-wise multiplication operation.

jijoongmoon: I cannot see why. could you elaborate?

nntrainer/tensor/tensor.cpp
jijoongmoon: could you elaborate why?

kparichay: dropout_mask is usually stored by the dropout layer. This is because it is needed for the backwarding operation. The derivatives also needs to be zero out.

nntrainer/tensor/tensor.cpp
kparichay: the proposed style reuses blas operations and will be faster, and easier to read as well.

nntrainer/tensor/tensor.cpp
jijoongmoon: > As far as I understand, it is not need in maks. mask is already applied in input. so.. for the training, it is already applied. But one thing I concern, the deterministic random number for the rand_temp. but nntrainer is already deteministic, so.. we don't need.

nntrainer/tensor/tensor.cpp
jijoongmoon: But I guess it is much better just use one loop, rather than two.

jijoongmoon:
> > > Please add dropout as a layer than adding it as a property of a layer. With the current design, each layer will have to inidividually implement dropout. But if dropout is a layer, then each layer can use dropout by attaching it in front of it. Please update the design.
> > 
> > 
> > I think this should deal differently. This pr only for the recurrent network. We can implement drop out layer as an formal layer later.
> 
> IMHO, it is a very small task to implement a dropout layer (**a few hours max**) but maintains extensibility and available for LSTM/GRU etc layers, as well as easier to debug and deal with (this PR adds repeated code for both RNN/LSTM which reduces maintainability). Given when dropout layer is added, this code will be removed, I think directly creating dropout layer will be better.

I don't think it is just small task. (If you think it is, - like you said few hours max task, then you can do it, please). Also in the API point of view, other frameworks supports this way as a property. If you are asking to use explicit layer declaration, then I will say no. We could support this drop out like flatten property and if we are going to, then needs realization and it takes more time including test. If you are saying both of them, then it is ok. So I guess this can be done later after the refactoring is finished. I'm fully aware of this, so don't worry too much. I'm sure it can be done after refactoring. ( again, if you like to do it while you are refactoring, then you can do it. Cause, it is just a few hours tasks. So please. ).

> Also, this breaks LayerV2 migration and refactoring.

I don't see why it breaks LayerV2 migration and refactoring? If it is because of just adding another property, then it should be supported.

nntrainer/tensor/tensor.cpp
kparichay: > But I guess it is much better just use one loop, rather than two.

If it is one loop with `if` statements inside it, then I am not sure how well the compiler can optimize.

nntrainer/tensor/tensor.cpp
kparichay: I am concerned about the derivatives in `calcDerivative`. I think it might also need some zeroing out with the same mask.
So something like (pseudocode) `derivative *= mask`.

kparichay:
> I don't think it is just small task. (If you think it is, - like you said few hours max task, then you can do it, please.). Also in the API point of view, other frameworks supports this way as a property. If you are asking to use explicit layer declaration for this kind of scenario, then I will say no. We could support this drop out like flatten property and if we are going to, then needs realization and it takes more time including test. If you are saying both of them, then it is ok. So I guess this can be done later after the refactoring is finished because it might needs changes. I'm fully aware of this, so don't worry too much. I'm sure it can be done after refactoring. ( again, if you like to do it while you are refactoring, then you can do it. Cause, it is just a few hours tasks. So please. ).

Yes, I meant it to be used just like flatten layer concept. I will add a `DropoutLayer` quickly based on your implementation in RNN/LSTM layers. And then realization will be added with/while refactoring.

> I don't see why it breaks LayerV2 migration and refactoring? If it is because of just adding another property, then it should be supported.

This property will ideally be suited best for LayerNode. I guess will have to be done later.

kparichay:
This PR provides an initial implementation for the dropout, and as it is being used with RNN/LSTM for now, lets keep it this way for verification and the implementation can be wrapped in a layer later. Let's go forward with this.

kparichay (approved):
LGTM other than the minor comment about comparison with 0.0

zhoonit (approved):
LGTM

jijoongmoon:
> > > Please add dropout as a layer than adding it as a property of a layer. With the current design, each layer will have to inidividually implement dropout. But if dropout is a layer, then each layer can use dropout by attaching it in front of it. Please update the design.
> > 
> > 
> > I think this should deal differently. This pr only for the recurrent network. We can implement drop out layer as an formal layer later.
> 
> IMHO, there might not be big overhead to implement a dropout layer (as the implementation has already been added in the PR, it just need wrapping as a separate layer) and maintains extensibility and available for LSTM/GRU etc layers, as well as easier to debug (this PR results in same logic of dropout for both RNN/LSTM).
> Also, this might break LayerV2 migration and refactoring.
> 
> This PR adds a new property that would be best suited to be inferred at the graph realization, and graph can add a new dropout layer (just like flatten) and extend support of dropout for all layers. That said, lets keep realization for later (after refactor).
> 
> If this PR is time sensitive, lets start with the design in this PR, and update to a new layer later.



> > > Please add dropout as a layer than adding it as a property of a layer. With the current design, each layer will have to inidividually implement dropout. But if dropout is a layer, then each layer can use dropout by attaching it in front of it. Please update the design.
> > 
> > 
> > I think this should deal differently. This pr only for the recurrent network. We can implement drop out layer as an formal layer later.
> 
> IMHO, there might not be big overhead to implement a dropout layer (as the implementation has already been added in the PR, it just need wrapping as a separate layer) and maintains extensibility and available for LSTM/GRU etc layers, as well as easier to debug (this PR results in same logic of dropout for both RNN/LSTM).
> Also, this might break LayerV2 migration and refactoring.
> 
> This PR adds a new property that would be best suited to be inferred at the graph realization, and graph can add a new dropout layer (just like flatten) and extend support of dropout for all layers. That said, lets keep realization for later (after refactor).
> 
> If this PR is time sensitive, lets start with the design in this PR, and update to a new layer later.

I think this design of droput for recurrent layer is also required as we discussed off-line. I will add new dropout layer in layer PR and also update this design with saving mask.

nntrainer/tensor/tensor.cpp
jijoongmoon: > If it is one loop with `if` statements inside it, then I am not sure how well the compiler can optimize. But this is for optimization part, so lets do it later.

I will fix in later pr.

nntrainer/tensor/tensor.cpp
jijoongmoon: > I am concerned about the derivatives in `calcDerivative`. I think it might also need some zeroing out with the same mask.
> So something like (pseudocode) `derivative *= mask`.

right. we need to save mask. I will update in later pr.

nntrainer/layers/gru.cpp
jijoongmoon: I will update it in later PR. Thanks.

PR #1323, opened 2021-06-22
[layernode] Move activation to LayerNode @open sesame 06/22 18:12

kparichay:
Move activation from Layer internal to LayerNode

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -143,10 +143,9 @@ int NetworkGraph::realizeFlattenType(
 
 int NetworkGraph::realizeActivationType(
   const std::shared_ptr<LayerNode> &in_node) {
-  LayerV1 &current = *in_node->getObject();
   int status = ML_ERROR_NONE;
 
-  ActivationType act = current.getActivationType();
+  ActivationType act = in_node->getActivationToBeRealized();
 
   if (act == ActivationType::ACT_NONE) {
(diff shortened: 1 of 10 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -172,7 +171,8 @@ int NetworkGraph::realizeActivationType(
   }
 
   std::shared_ptr<LayerV1> layer = lnode->getObject();
-  layer->setActivation(act);
+  lnode->setProperty({"activation=" + ActivationTypeStr[(unsigned int)act]});
+  in_node->setProperty({"activation=none"});
```
jijoongmoon, line 175: I wonder that if the activation property is set in layer_node, then does it set the layer property as well?? 
It looks like when we call setProperty of layer node, then setActivation() is called and seems like it does not set the layer property. Is it OK without set the layer property?

jijoongmoon (approved):
LGTM, except minor comments.

nntrainer/graph/network_graph.cpp
kparichay: > I wonder that if the activation property is set in layer_node, then does it set the layer property as well??

Yes, the layer property is also set. The throw in setProperty() after calling setActivation() leads to the setting the activation in layer as well (which is the activation layer). The throw method is replaced with a  better method in #1332.

zhoonit (approved):
lgtm

PR #1325, opened 2021-06-22
[layer] Update getTrainable with supportBackwarding

kparichay:
This patch introduces supportBackwarding.
Currently getTrainable does two jobs -
1. check if the layer is trainable
2. check if the layer can do backwarding

Now, the two is separated.
Further, the trainable property has been moved to LayerNode.

### DIFF
--- a/Applications/SimpleShot/layers/l2norm.h
+++ b/Applications/SimpleShot/layers/l2norm.h
@@ -82,12 +82,9 @@ public:
   const std::string getType() const override { return L2NormLayer::type; }
 
   /**
-   * @brief get boolean if the function is trainable
-   *
-   * @retval true trainable
-   * @retval false not trainable
+   * @copydoc bool supportBackwarding() const
    */
-  bool getTrainable() noexcept override { return false; }
+  bool supportBackwarding() const override { return false; };
 
   inline static const std::string type = "l2norm";
 };
--- a/nntrainer/compiler/ini_interpreter.cpp
+++ b/nntrainer/compiler/ini_interpreter.cpp
@@ -207,8 +207,8 @@ getMergeableGraph(std::shared_ptr<const GraphRepresentation> graph,
   // bool preload =
   //   iniparser_getboolean(ini, (sec_name + ":Preload").c_str(), true);
 
-  bool trainable =
-    iniparser_getboolean(ini, (sec_name + ":Trainable").c_str(), false);
+  const std::string &trainable =
+    iniparser_getstring(ini, (sec_name + ":Trainable").c_str(), "false");
   double scale_size =
     iniparser_getdouble(ini, (sec_name + ":ScaleSize").c_str(), 1.0);
 
@@ -217,11 +217,11 @@ getMergeableGraph(std::shared_ptr<const GraphRepresentation> graph,
     << "backbone cannot have non-positive scale_size. Current scale size: "
     << scale_size;
 
-  for (auto &layer : g) {
-    layer->getObject()->setTrainable(trainable);
-    layer->getObject()->resetDimension();
+  for (auto &lnode : g) {
+    lnode->setProperty({"trainable=" + trainable});
+    lnode->getObject()->resetDimension();
     if (scale_size != 1) {
-      layer->getObject()->scaleSize(scale_size);
+      lnode->getObject()->scaleSize(scale_size);
     }
     /** TODO #361: this needs update in model file to be of dictionary format */
     // if (preload) {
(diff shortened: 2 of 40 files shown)


nntrainer/compiler/ini_interpreter.cpp
```diff
@@ -207,8 +207,8 @@ getMergeableGraph(std::shared_ptr<const GraphRepresentation> graph,
   // bool preload =
   //   iniparser_getboolean(ini, (sec_name + ":Preload").c_str(), true);
 
-  bool trainable =
-    iniparser_getboolean(ini, (sec_name + ":Trainable").c_str(), false);
+  const std::string &trainable =
+    iniparser_getstring(ini, (sec_name + ":Trainable").c_str(), "false");
```
zhoonit, line 211: Not related to the PR but, is the default value false?

zhoonit (approved):
Nice one :)

nntrainer/compiler/ini_interpreter.cpp
kparichay: I am not sure about this either. This is our design choice.
False: this avoids user from creating extra trainable layers (i added some point, didn't want to leave it empty).
True: as this is training framework, this just sounds sensible. This also matches with each layer using `trainable=true` by default.

kparichay: I think true as default is better. what do you think?

zhoonit: I agree :) let's change this to true

kparichay: Changed to true as a new commit in this PR.

jijoongmoon: True. I guess.

Applications/SimpleShot/layers/l2norm.h
```diff
@@ -82,12 +82,9 @@ class L2NormLayer : public nntrainer::LayerV1 {
...
 
   /**
-   * @brief get boolean if the function is trainable
-   *
-   * @retval true trainable
-   * @retval false not trainable
+   * @copydoc bool supportBackwarding() const
    */
-  bool getTrainable() noexcept override { return false; }
+  bool supportBackwarding() const override { return false; };
```
jijoongmoon, line 87: The name of this is more appropriate.

nntrainer/graph/network_graph.cpp
```diff
@@ -86,7 +86,9 @@ void NetworkGraph::addLayerNode(std::shared_ptr<LayerV1> layer) {
 
 void NetworkGraph::countNonTrainableLayersAtBegin() {
   for (auto iter = cbegin(); iter != cend(); iter++) {
-    if ((*iter)->getObject()->getTrainable()) {
+    // TODO: check if getTrainable() was set and if trainable weights exist,
+    // then throw
+    if ((*iter)->getTrainable() && (*iter)->supportBackwarding()) {
```
jijoongmoon, line 91: Could you elaborate the difference between getTrainable and supportBackwarding?

nntrainer/layers/plugged_layer.h
```diff
@@ -143,14 +143,9 @@ class PluggedLayer : public nntrainer::LayerV1 {
...
-   * @copydoc Layer::setTrainable(bool train)
+   * @copydoc bool supportBackwarding() const
    */
-  void setTrainable(bool train) override { layerImpl->setTrainable(train); }
-
-  /**
-   * @copydoc Layer::getTrainable()
-   */
-  bool getTrainable() noexcept override { return layerImpl->getTrainable(); }
+  bool supportBackwarding() const override { return false; };
```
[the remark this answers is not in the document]
kparichay: you are right. Updated.

nntrainer/graph/network_graph.cpp
kparichay: getTrainable() means if the layer should be trained (user can set trainable = true/false for a layer).
supportBackwarding() means if the layer provides implementation to do backwarding (eg. input_layer does not support backwarding as calcDerivative is not implemented).

jijoongmoon: So.. for making clearer, for the Conv2D layer, supportBackwarding() is always true ( it looks more like property then ), and can be trainable true or false depending on option then. we also have training variable and weight also have trainable property. So if the layer trainable is set true, then all the trainable in weights for that layer is going to set trainable.. ?

kparichay:
I had a discussion offline with @jijoongmoon. Below is the summary:
- Trainable: Layer property which can be set by the user indicating if the layer should be trained or not
- Training (to be updated to Enum Class ExecutionMode: {Train, Inference} - maybe add eval later): the model is running for training or inference
- supportBackwarding: this is set by the layer writer - this is the guarantee given by the layer writer if the layer can backpropagate the gradients or not
- Weight::Trainable (to be updated to needs/hasGradient()): if the weight needs/has gradient or not

This will be updated to use the above nomenclature soon.

kparichay:
Updated as per described in the previous commit. Mode of executions are defined but not used. It will be used when refactoring the layers from #1328.

PR #1326, opened 2021-06-22
[CAPI] Propose save/load api

zhoonit:
- [CAPI] Propose save/load api

```
**Motivations**
1. Fine grained api required to save and load

- Add function to model.h  / `saveModel`, `loadModel`
- Mark `readModel()` deprecated (later remove...)
- Mark `saveModel()` to change `exportTo`
- Add `ModelSaveLoadFlags`

**Self evaluation:**

```

### DIFF
--- a/api/capi/include/nntrainer.h
+++ b/api/capi/include/nntrainer.h
@@ -216,6 +216,49 @@ int ml_train_model_set_dataset(ml_train_model_h model,
                                ml_train_dataset_h dataset);
 
 /**
+ * @brief Save the model after training
+ * @details Use this function to save the current model.
+ * When calling with @a ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS or
+ * ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS, the model must be compiled
+ * with @a ml_train_model_compile beforehand.
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to save
+ * @param[in] path_prefix  Path prefix to save the file. This function will save
+ * one or number of files using the given prefix. Note that this does not state
+ * the exact file name but prefix
+ * @param[in] option Option flag which part of model should be saved
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Path is a directory without a prefix, or
+ * the given path is invalid or model is not compiled.
+ * @retval #ML_ERROR_NOT_SUPPORTED Given flag is not valid for the model.
+ */
+int ml_train_model_save(ml_train_model_h model, const char *path_prefix,
+                        ml_train_model_save_load_flags_e option);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model, model must be compiled
+ * When calling with @a ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS or
+ * ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS, the model must be compiled
+ * with @a ml_train_model_compile beforehand.
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to load.
+ * @param[in] path_prefix  Path prefix to load the file. Note that this does not
+ * state the exact file name but prefix
+ * @param[in] option Option flag which part of model should be loaded.
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Path is a directory without a prefix, or
+ * the given path is invalid.
+ * @retval #ML_ERROR_NOT_SUPPORTED Given flag is not valid for the model
+ */
+int ml_train_model_load(ml_train_model_h model, const char *path_prefix,
+                        ml_train_model_save_load_flags_e option);
+
+/**
  * @brief Creates a neural network layer.
  * @details Use this function to create neural network layer.
  * @since_tizen 6.0
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -19,6 +19,7 @@
 #if __cplusplus >= MIN_CPP_VERSION
 
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include <nntrainer-api-common.h>
@@ -38,6 +39,30 @@ enum class ModelType {
   KNN,        /** k Nearest Neighbor */
   NEURAL_NET, /** Neural Network */
   UNKNOWN     /** Unknown */
+};
+
+/**
+ * @brief Model saving options
+ *
+ */
+enum class ModelSaveLoadFlags {
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS =
+    ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS, /**< params required for
+                                                        inferences, eg) model
+                                                        weights */
+  MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS =
+    ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS, /**< params required for
+                                                       training, eg) optimizer
+                                                       weights */
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_CONFIG =
+    1 << 2, /**< inference related model_definitions, eg) graph */
+  MODEL_SAVE_LOAD_FLAGS_TRAINING_CONFIG =
+    1 << 3, /**< training related model_definitions, eg) optimizer */
+
+  MODEL_SAVE_LOAD_FLAGS_PARAMS =
+    (1 << 0) | (1 << 1), /**< INFERENCE_PARAMS | TRAINING_PARAMS */
+  MODEL_SAVE_LOAD_FLAGS_CONFIG =
+    (1 << 2) | (1 << 3) /**< INFERENCE_CONFIG | TRAINING_CONFIG */
 };
 
 /**
@@ -95,13 +120,36 @@ public:
 
   /**
    * @brief     save model and training parameters into file
+   * @todo      deprecate this
    */
-  virtual void saveModel() = 0;
+  [[deprecated("use saveModel(const std::string &path_prefix)")]] virtual void
+  saveModel() = 0;
+
+  /**
+   * @brief  load model states and training parameters from a file
+   * @param path_prefix path_prefix to save the model, if full path is not
+   * given, it should be saved inside working directory
+   * @param option option to save parameters
+   */
+  virtual void save(const std::string &path_prefix,
+                    ModelSaveLoadFlags option =
+                      ModelSaveLoadFlags::MODEL_SAVE_LOAD_FLAGS_PARAMS){};
 
   /**
    * @brief     read model and training parameters from file
+   * @todo      deprecate this
    */
   virtual void readModel() = 0;
+
+  /**
+   * @brief  load model with regard to the option
+   * @param path_prefix path_prefix to save the model, if full path is not
+   * given, it should be saved inside working directory
+   * @param option option to save parameters
+   */
+  virtual void load(const std::string &path_prefix,
+                    ModelSaveLoadFlags option =
+                      ModelSaveLoadFlags::MODEL_SAVE_LOAD_FLAGS_PARAMS){};
 
   /**
    * @brief     Run Model training and validation
--- a/api/nntrainer-api-common.h
+++ b/api/nntrainer-api-common.h
@@ -164,6 +164,17 @@ typedef enum {
 } ml_train_summary_type_e;
 
 /**
+ * @brief Enumeration for the neural network
+ *
+ */
+typedef enum {
+  ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS =
+    1 << 0, /**< params required for inference, For example, model weights */
+  ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS =
+    1 << 1, /**< params required for training, For example, optimizer weights */
+} ml_train_model_save_load_flags_e;
+
+/**
  * @}
  */
 #ifdef __cplusplus


api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,51 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
+ * @param[in] path  Pth to save the file
+ * @param[in] file_version file format version
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER File path is invalid or model does not
+ * have any allocated states.
+ * @retval #ML_ERROR_NOT_SUPPORTED Not supported format.
+ */
+int ml_train_model_save_states(ml_train_model_h model, const char *path,
+                               int file_version);
```
zhoonit, line 239: There are three options regarding file_verison.

option 1. current PR (using `int file_version`)
option 2. using dedicated enum such as `ML_TRAIN_STATES_FILE_FORMAT`
option 3. we rely on the extension of the `path`

option 3. is simpler and intuitive but there is one big caveat that if we are having extension crash across versions, it might be impossible to deal with.

kparichay: `save_states` goes against the extension of the path.
`.tflite` saves the states as well as the model architecture while `.bin` + `.ini` has different behavior.

kparichay:
You must think about when the proposed APIs will be used (best to create a scenario where this will be used). If they are to be used after `ml_train_model_run`, this implies that `ml_train_model_save` will overwrite the previously saved model by `ml_train_model_run`? The abstraction of the APIs and their behavior needs to be stated.

api/capi/include/nntrainer.h
kparichay: The same issue exists with `ml_train_model_load`. When using with `tflite` format, does the API validates the model architecture or overwrites the model with the given architecture from `tflite` format.

zhoonit: We will need to save states only. This is also what tflite guys are trying to do.

See below pseudo code from their side.
```python
## This code is from https://github.com/tensorflow/community/pull/390
# Construct the interpreter and get the signature runners.
interpreter = tf.lite.Interpreter(TFLITE_FILE_PATH)
train = interpreter.get_signature_runner(method_name="train")
infer = interpreter.get_signature_runner(method_name="infer")
save = interpreter.get_signature_runner(method_name="save")
restore = interpreter.get_signature_runner(method_name="restore")
# Restore trainable weights from files
restore("/path/to/checkpoints")
# Run training for a few steps
train(train_input, train_labels)
# Save trainable weights back to checkpoint files files
save("/path/to/checkpoints")
# Developers can also use the trained model to run inference now
prediction = infer(input)
```

so states here maps to checkpoint, checkpoint will not have any model architecture saved but only weights.

It's a bit off the topic but, checking the tflite file, we can use tflite format directly to save the checkpoint (This might not be a good idea though).

zhoonit:
> You must think about when the proposed APIs will be used (best to create a scenario where this will be used). If they are to be used after `ml_train_model_run`, this implies that `ml_train_model_save` will overwrite the previously saved model by `ml_train_model_run`? The abstraction of the APIs and their behavior needs to be stated.

From the documentation, we allow setting state `save_path` in run. With `save_path`, we control restoring, updating the states.
With the proposed api, we handover the control to save and restore the model to the api user. Similar to proposed code in https://github.com/nnstreamer/nntrainer/pull/1326#discussion_r656710277

api/capi/include/nntrainer.h
zhoonit: > The same issue exists with ml_train_model_load. When using with tflite format, does the API validates the model architecture or overwrites the model with the given architecture from tflite format. 

I think it's better to use different extension than`*.tflite` although it's possible.

zhoonit: How about `ml_train_model_save_checkpoint`?

kparichay: > We will need to save states only. This is also what tflite guys are trying to do.

I think you mean sort of update the states (as the output tflite file should be able to run for inference standalone)?

> so states here maps to checkpoint, checkpoint will not have any model architecture saved but only weights.

This confuses me a bit, then inference would require two files as this checkpoint will not have model architecture?

> It's a bit off the topic but, checking the tflite file, we can use tflite format directly to save the checkpoint (This might not be a good idea though).

not so sure about this one, but thats a good point to raise.

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,43 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
+ * one or number of files using the given prefix. Note that this does not state
+ * the exact file name but prefix
+ * @param[in] option Option flag which part of model should be saved
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Path is a directory without a prefix, or
+ * the given path is invalid.
+ * @retval #ML_ERROR_NOT_SUPPORTED Given flag is not valid for the model
+ */
+int ml_train_model_save(ml_train_model_h model, const char *path_prefix,
```
kparichay, line 234: please add details for which state model should be in to support this - constructed, compiled, at least run once, etc.

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,43 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
+                        ml_train_model_save_load_flags_e option);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to save
```
kparichay, line 242: save -> load

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,43 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
                                ml_train_dataset_h dataset);
 
+/**
+ * @brief Save the model after training
+ * @details Use this function to save the current model
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to save
+ * @param[in] path_prefix  Path prefix to save the file. This function will save
+ * one or number of files using the given prefix. Note that this does not state
```
kparichay, line 225: maybe a small example might be good? or too much?

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,43 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
+ * the exact file name but prefix
+ * @param[in] option Option flag which part of model should be saved
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER Path is a directory without a prefix, or
```
kparichay, line 230: write behavior if the file already exists - overwrite or error (overwrite is better IMO).

kparichay: adding a number would also increase another argument for load - load latest/ or a specific number.

api/ccapi/include/model.h
```diff
@@ -95,14 +125,36 @@ class Model {
 
   /**
    * @brief     save model and training parameters into file
+   * @todo      deprecate this
```
kparichay, line 128: how about [[deprecate]]??

api/capi/include/nntrainer.h
zhoonit: > write behavior if the file already exists - overwrite or error

if we overwrite, do we clean the previously saved ones?

for example
```cpp
 save("here", config | params)
 save("here", params) <--- here do we leave config? or delete config?
```

api/ccapi/include/model.h
```diff
@@ -95,14 +125,36 @@ class Model {
...
    */
   virtual void saveModel() = 0;
 
+  /**
+   * @brief  load model states and training parameters from a file
+   * @param path_prefix path_prefix to save the model, if full path is not
+   * given, it should be saved inside working directory
+   * @param option option to save parameters
+   */
+  virtual void saveModel(const std::string &path_prefix,
```
kparichay, line 138: this is minor:  i wonder if `save()` is enough? or `Model` prefix is needed? given this is applied on `model` object anyway

api/ccapi/include/model.h
```diff
@@ -40,6 +41,35 @@ enum class ModelType {
...
+ * @brief Model saving options
+ *
+ */
+enum class ModelSaveLoadFlags {
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS =
+    1 << 0, /**< params required for inferences, eg) model weights */
+  MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS =
+    1 << 1, /**< params required for training, eg) optimizer weights */
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_CONFIG =
+    1 << 2, /**< inference related configurations, eg) graph */
```
kparichay, line 54: inference related configurations -> inference related model definition?

api/ccapi/include/model.h
```diff
@@ -40,6 +41,35 @@ enum class ModelType {
   UNKNOWN     /** Unknown */
 };
 
+/**
+ * @brief Model saving options
+ *
+ */
+enum class ModelSaveLoadFlags {
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS =
+    1 << 0, /**< params required for inferences, eg) model weights */
```
kparichay, line 50: 1 << 0 -> ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS

api/ccapi/include/model.h
```diff
@@ -40,6 +41,35 @@ enum class ModelType {
...
+  MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS =
+    1 << 1, /**< params required for training, eg) optimizer weights */
```
kparichay, line 52: 1 << 1 -> ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS

api/capi/include/nntrainer.h
kparichay: we can leave it to the user - or we can use numbered saves
@jijoongmoon what do you recommend?

api/capi/include/nntrainer.h
zhoonit: I'll think about it, I guess example can come when with some implementation and testing.

api/capi/include/nntrainer.h
zhoonit: Yeah, I guess we can leave this to the user. One vote for overwrite

api/ccapi/include/model.h
zhoonit: We are not allowing having deprecated warning, until we decide to change let's do in this fashion :p

api/ccapi/include/model.h
zhoonit: I think it's a good point. Changed as suggested

api/capi/include/nntrainer.h
zhoonit: Done

api/capi/include/nntrainer.h
zhoonit: Done

api/ccapi/include/model.h
zhoonit: Done

api/ccapi/include/model.h
zhoonit: Done

api/ccapi/include/model.h
zhoonit: Done

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,49 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
+ * When calling with @a ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS or
+ * ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS, the model must be compiled
+ * with @a ml_train_model_compile beforehand.
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to save
+ * @param[in] path_prefix  Path prefix to save the file. This function will save
+ * one or number of files using the given prefix. Note that this does not state
+ * the exact file name but prefix
+ * @param[in] option Option flag which part of model should be saved
```
jijoongmoon, line 230: Usually, when we are talking about saving the model, I think it might be better to make sure not to overwrite the current model or limit the number of saved models and kind of naming rule. I'm sure you have a plan for this.

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,49 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
 int ml_train_model_set_dataset(ml_train_model_h model,
                                ml_train_dataset_h dataset);
 
+/**
+ * @brief Save the model after training
+ * @details Use this function to save the current model.
+ * When calling with @a ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS or
+ * ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS, the model must be compiled
+ * with @a ml_train_model_compile beforehand.
```
jijoongmoon, line 223: if this function called with INFERENCE_PARAMS, then it will save the model just for inference ( Just weights )?
and if with TRAINING_PARAMS, then it save the model including hyper-params as well as inference?

api/capi/include/nntrainer.h
jijoongmoon: I guess numbered saves with limitations from uses.

api/capi/include/nntrainer.h
```diff
@@ -215,6 +215,49 @@ int ml_train_model_set_optimizer(ml_train_model_h model,
...
+ * @retval #ML_ERROR_NOT_SUPPORTED Given flag is not valid for the model.
+ */
+int ml_train_model_save(ml_train_model_h model, const char *path_prefix,
+                        ml_train_model_save_load_flags_e option);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model, model must be compiled
+ * When calling with @a ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS or
+ * ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS, the model must be compiled
```
jijoongmoon, line 244: I wonder if user set the INFERNECE_PARAMS, are we going to load the hyper-params or not. I think we don't need it. then we cannot call train if this option is given. If developer wants to train, they have to give TRAINING_PARAMS.

api/ccapi/include/model.h
```diff
@@ -40,6 +41,30 @@ enum class ModelType {
   UNKNOWN     /** Unknown */
 };
 
+/**
+ * @brief Model saving options
+ *
+ */
+enum class ModelSaveLoadFlags {
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_PARAMS =
```
jijoongmoon, line 49: I think we need check again how these options will work. Cause once we set these enum, It's going to be difficult to change.

api/capi/include/nntrainer.h
zhoonit: Current thinking is... 

- Some fixed configurations like learning rate, optimizers - `ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_CONFIG` training configuration (for now, only for ccapi)
- Training weights like iterations, adam weights - `ML_TRAIN_MODEL_SAVE_LOAD_FLAGS_TRAINING_PARAMS` (this is exposed)

api/capi/include/nntrainer.h
zhoonit: 100% agree :)

api/capi/include/nntrainer.h
zhoonit: Talked offline to clarify, that we will have training params including inference params :)

api/ccapi/include/model.h
zhoonit: As per offline discussion, unused enum for now will be commented

api/capi/include/nntrainer.h
kparichay: cool.

api/ccapi/include/model.h
kparichay: aah, got it. Then before M2 release, we will remove the deprecated APIs right?

zhoonit: Yes

api/capi/include/nntrainer.h
kparichay: So, what is the final conclusion on the overwrite/numbered saves?

api/ccapi/include/model.h
```diff
@@ -40,6 +41,30 @@ enum class ModelType {
...
+  MODEL_SAVE_LOAD_FLAGS_INFERENCE_CONFIG =
+    1 << 2, /**< inference related model_definitions, eg) graph */
+  MODEL_SAVE_LOAD_FLAGS_TRAINING_CONFIG =
+    1 << 3, /**< training related model_definitions, eg) optimizer */
+
+  MODEL_SAVE_LOAD_FLAGS_PARAMS =
+    (1 << 0) | (1 << 1), /**< INFERENCE_PARAMS | TRAINING_PARAMS */
+  MODEL_SAVE_LOAD_FLAGS_CONFIG =
+    (1 << 2) | (1 << 3) /**< INFERENCE_CONFIG | TRAINING_CONFIG */
+};
```
myungjoo, line 66: We will have multiple formats supported. How such info is expressed with this flag set?

zhoonit: Sorry for the late reply.

TL; DR. Current plan is we don't expose the file format directly, but leave it as a implementation detail of this API.
In this scheme, user can't and won't need to choose format.

The reason for this design choice is...

 1. There are use case of saving only some part of the graph
 some non-exhaustive and overlapping lists are:
    - saving weights to retrain later.
    - saving weights to inference later.
    - saving graph only for inference.
    - saving optimizer and hyper params to retrain later.
  2. Every format serves only their purpose, so we can't save everything in a single format. eg) tflite support only inference related weights + graph.
  3. If user designates a format as well as what to save, they will likely to call in a ill-formed combination.
  eg) Trying to saving hyperparams with *.tflite format.

 So, we decided user shouldn't know about the format directly but rely on the api, if they want to save now and restore the graph later.
If, we need to save a model in a specified, explicit format, we will have to support `ml_train_model_export`.

| Version           | call                                                                                             | Generated File                          |
|-------------------|--------------------------------------------------------------------------------------------------|-----------------------------------------|
| Tizen 6.5         | save(prefix, FLAGS_TRAIN_PARAM)                                                                  | ${prefix}.bin                           |
|                   | save(prefix, FLAGS_TRAIN_PARAM \| FLAGS_INFER_PARAM)                                             | ${prefix}.bin                           |
|                   | save(prefix, FLAGS_INFER_PARAM)                                                                  | ${prefix}.bin                           |
| Tizen near future | save(prefix, FLAGS_TRAIN_PARAM \| FLAGS_TRAIN_CONFIG)                                            | ${prefix}.bin, ${prefix}.ini             |
|                   | save(prefix, FLAGS_TRAIN_PARAM \| FLAGS_INFER_PARAM \| FLAGS_TRAIN_CONFIG \| FLAGS_INFER_CONFIG) | ${prefix}.bin, ${prefix}.ini             |
|       | save(prefix, FLAGS_INFER_PARAM \| FLAGS_INFER_CONFIG)                                            | ${prefix}.bin ${prefix}.ini  |
| Tizen future      | save(prefix, FLAGS_INFER_PARAM \| FLAGS_INFER_CONFIG)                                            | ${prefix}.tflite (something equivalent) |

api/capi/include/nntrainer.h
zhoonit: 1. If we number with given path,

- Over 100mb storage will be implicitly used everytime calling this api.
- It's hard to locate which to delete. Because the name mangling rule won't be exposed.
- So, user can only delete the whole model.

2. If we override,

- saving on the same path with different flag is grey area in terms of api behavior.
 eg) saving train params > saving inference params > saving train params (in *.bin file, this overrides inference params), (in otherfiles this might not override inference params)

3. If give error if the path is taken.

- Not user friendly if user wants to override.
- But predictable, and user can control over the behavior if they wants to.

IMO, the option 3 has the only way to guarantee the consistent semantics is to give error if the given path is taken.

side note) We do not support deletion of the saved model. With path_prefix strategy, it's not possible to easily delete a file, so I think it's better to get empty directory as an argument instead. 
Come to think of it, it seems we should have a directory as an argument instead of a path prefix.

api/ccapi/include/model.h
myungjoo: The enum for save/load is getting too complicated. I doubt if you can handle such complicated save/load protocols.
Try to minimize the possible save file formats (best = 1). And enumerate each of the format here. I understand that we can have a lot of different combinations because there are different types of data to be stored, but users would want only one format for their needs.

E.g., save (filepath, FLAG_TFLITECOMPAT_V1); save (filepath, FLAG_PYTORCHCOMPAT_V1); save (filepath, FLAG_RAWWEIGHT_V1); ...

zhoonit:
There are few concerns that needs some breaking changes, in order to reduce confusions, I'll close this and recreate a new PR :)

api/ccapi/include/model.h
zhoonit: Yes, I think this way is more configuration, I'll come up with suggested design 👍

PR #1328, opened 2021-06-23
[fclayer] Update fc layer for LayerV2

kparichay:
1. Update fc layer for LayerV2
2. Update node_exporter to export weights from run_context
Add corresponding const getters and setters in layer_context
Updated some fixes in var_grad
3. Update layer factory for LayerV2 for creation of layers of signature
LayerV2.

### DIFF
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -33,103 +33,94 @@ namespace nntrainer {
 
 enum FCParams { weight, bias };
 
-int FullyConnectedLayer::initialize(Manager &manager) {
+void FullyConnectedLayer::finalize(InitLayerContext &context) {
   auto unit = std::get<props::Unit>(fc_props).get();
-  int status = ML_ERROR_NONE;
 
-  if (getNumInputs() != 1) {
+  if (context.getNumInputs() != 1) {
     throw std::invalid_argument("Fully connected layer takes only one input");
   }
 
-  auto &in_dim = input_dim[0];
+  std::vector<TensorDim> output_dims(1);
+
   /// @todo fc actaully supports multidimensions. EffDimFlag shouldn't be fixed
   /// like this.
-  in_dim.setEffDimFlag(0b1001);
-  in_dim.setDynDimFlag(0b1000);
+  context.setEffDimFlagInputDimension(0, 0b1001);
+  context.setDynDimFlagInputDimension(0, 0b1000);
 
-  output_dim[0] = in_dim;
-  output_dim[0].width(unit);
+  /** set output dimensions */
+  auto const &in_dim = context.getInputDimensions()[0];
+  output_dims[0] = in_dim;
+  output_dims[0].width(unit);
+  context.setOutputDimensions(output_dims);
 
+  /** set weight specifications */
   TensorDim bias_dim(1, 1, 1, unit, 0b0001);
   TensorDim weight_dim(1, 1, in_dim.width(), unit, 0b0011);
 
-  if (weights.empty()) {
-    weights.reserve(2);
-    weights.emplace_back(weight_dim, weight_initializer, weight_regularizer,
-                         weight_regularizer_constant, true, false, "FC:weight");
-    weights.emplace_back(bias_dim, bias_initializer, WeightRegularizer::NONE,
-                         1.0f, true, false, "FC:bias");
-    manager.trackWeights(weights);
-  } else {
-    weights[FCParams::weight].reset(weight_dim, weight_initializer,
-                                    weight_regularizer,
-                                    weight_regularizer_constant, true);
-    weights[FCParams::bias].reset(bias_dim, bias_initializer,
-                                  WeightRegularizer::NONE, 1.0f, true);
-  }
+  weight_idx[FCParams::weight] =
+    context.requestWeight(weight_dim, weight_initializer, weight_regularizer,
+                          weight_regularizer_constant, "FC:weight", true);
 
-  return status;
+  weight_idx[FCParams::bias] = context.requestWeight(
+    bias_dim, bias_initializer, WeightRegularizer::NONE, 1.0f, "FC:bias", true);
+
+  // if (weights.empty()) {
+  //   weights.reserve(2);
+  //   weights.emplace_back(weight_dim, weight_initializer, weight_regularizer,
+  //                        weight_regularizer_constant, true, false,
+  //                        "FC:weight");
+  //   weights.emplace_back(bias_dim, bias_initializer, WeightRegularizer::NONE,
+  //                        1.0f, true, false, "FC:bias");
+  //   manager.trackWeights(weights);
+  // } else {
+  //   weights[FCParams::weight].reset(weight_dim, weight_initializer,
+  //                                   weight_regularizer,
+  //                                   weight_regularizer_constant, true);
+  //   weights[FCParams::bias].reset(bias_dim, bias_initializer,
+  //                                 WeightRegularizer::NONE, 1.0f, true);
+  // }
 }
 
-void FullyConnectedLayer::export_to(Exporter &exporter,
-                                    ExportMethods method) const {
-  LayerV1::export_to(exporter, method);
+void FullyConnectedLayer::exportTo(Exporter &exporter,
+                                   const ExportMethods &method) const {
+  Layer::exportTo(exporter, method);
   exporter.saveResult(fc_props, method, this);
 }
 
-void FullyConnectedLayer::setProperty(const PropertyType type,
-                                      const std::string &value) {
-  switch (type) {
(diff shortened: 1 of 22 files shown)


nntrainer/layers/fc_layer.h
```diff
@@ -95,19 +82,19 @@ class FullyConnectedLayer : public LayerV1 {
     return FullyConnectedLayer::type;
   };
 
-  using LayerV1::setProperty;
+  using Layer::setProperty;
```
zhoonit, line 90: I guess this can now go away.

nntrainer/layers/fc_layer.cpp
```diff
@@ -33,103 +33,94 @@ namespace nntrainer {
...
-  } else {
-    weights[FCParams::weight].reset(weight_dim, weight_initializer,
-                                    weight_regularizer,
-                                    weight_regularizer_constant, true);
-    weights[FCParams::bias].reset(bias_dim, bias_initializer,
-                                  WeightRegularizer::NONE, 1.0f, true);
-  }
-
-  return status;
+  weight_idx[FCParams::weight] =
```
zhoonit, line 60: is `runLayerContext.getWeight(0)` returns the first registered weight?

nntrainer/layers/fc_layer.cpp
```diff
@@ -33,103 +33,94 @@ namespace nntrainer {
...
-void FullyConnectedLayer::copy(std::shared_ptr<LayerV1> l) {
-  LayerV1::copy(l);
-
-  std::shared_ptr<FullyConnectedLayer> from =
-    std::static_pointer_cast<FullyConnectedLayer>(l);
-
-  std::get<props::Unit>(fc_props) = std::get<props::Unit>(from->fc_props);
-}
+void FullyConnectedLayer::calcDerivative(RunLayerContext &context) {
+  Tensor &weight = context.getWeight(weight_idx[FCParams::weight]);
```
zhoonit, line 107: Does `runLayerContext.getWeight(FCParams::weight)` returns the first registered weight?
or the weight index is always dynamic inside layer context?

nntrainer/layers/fc_layer.h
```diff
@@ -26,20 +26,17 @@ namespace nntrainer {
...
+class FullyConnectedLayer : public LayerImpl {
 public:
   /**
    * @brief     Constructor of Fully Connected Layer
    */
-  template <typename... Args>
-  FullyConnectedLayer(unsigned int unit_ = 0, Args... args) :
-    LayerV1(args...),
-    fc_props(props::Unit(unit_)) {}
+  FullyConnectedLayer() : LayerImpl(), fc_props(props::Unit(0)) {}
```
zhoonit, line 34: slightly off the topic, is Unit of zero valid prop? else I think we can leave it as empty.

nntrainer/utils/node_exporter.cpp
```diff
@@ -50,6 +50,25 @@ void Exporter::saveTflResult(const std::tuple<> &props, const LayerV1 *self) {
   createIfNull(tf_node);
 }
 
+static void saveTflWeights(TfOpNode *tf_node, const RunLayerContext &context,
```
zhoonit, line 53: Nice, could you remove `var_grad::cloneTransposeVariableOnly` within this PR?

nntrainer/utils/node_exporter.cpp
```diff
@@ -59,22 +78,15 @@ void Exporter::saveTflResult(
   tf_node->setInOut(*self);
   tf_node->setInputs(self->getObject()->getInputRef());
   tf_node->setOutputs(self->getObject()->getOutputRef());
+
+  saveTflWeights(tf_node.get(), self->getRunContext(), "0:2:1");
```
zhoonit, line 82: This behavior is specific to fc layer, could you check if this should move to line 89.

nntrainer/layers/fc_layer.cpp
kparichay: getWeight() will return the weight for the given index. This index is returned when requestWeight() is called.

nntrainer/layers/fc_layer.h
kparichay: will update to empty.

nntrainer/utils/node_exporter.cpp
kparichay: I will remove it as part of a new commit.

nntrainer/layers/fc_layer.h
kparichay: done.

nntrainer/utils/node_exporter.cpp
kparichay: right, this will be different for conv, bn, etc. Done.

nntrainer/layers/fc_layer.cpp
zhoonit: So, my concern is extra level of indirection is always required for every layer.

If returned index from `requiestWeight` is always the new index of an array, we can use `FCParams::weight`. instead of `weight_idx[FCParams::weight]`.

kparichay: This is to ensure that the change in declaration order does not affect the result.

nntrainer/utils/node_exporter.cpp
zhoonit: Let's delay the discussion later until we have some use case, I think we can easily change the behavior later, if we have other requirements.

kparichay: does "0:2:1" works for convolution layer?

nntrainer/layers/fc_layer.cpp
zhoonit: I was wondering about the semantic. If that's the case it might be better in this case we promote weight_idx to vector and move to layerImpl later :)

nntrainer/utils/node_exporter.cpp
zhoonit: I think it is likely to be specific to fc. They transpose and save weights for fc.
confer below
https://github.com/tensorflow/community/blob/master/rfcs/20190722-tflite-training/trans1.png

zhoonit (approved):
LGTM

nntrainer/layers/fc_layer.cpp
kparichay: Even moving to LayerImpl means that layer_context needs to provide a guarantee that it can always provide weight indices from 0 and in incremental fashion order. For this to work, weights and requested tensor will have to maintain two separate lists (which is done currently, but I am thinking the lists will have to updated based on LIFESPAN type).
So, I will update this when the memory optimizations are done.

nntrainer/utils/node_exporter.cpp
kparichay: then layer needs to specify while declaring the weights if its weights are to be transposed for save/load or we have to maintain a list of layers whose weights are to be transposed.

nntrainer/layers/fc_layer.cpp
```diff
@@ -33,103 +33,94 @@ namespace nntrainer {
 
 enum FCParams { weight, bias };
 
-int FullyConnectedLayer::initialize(Manager &manager) {
+void FullyConnectedLayer::finalize(InitLayerContext &context) {
   auto unit = std::get<props::Unit>(fc_props).get();
-  int status = ML_ERROR_NONE;
 
-  if (getNumInputs() != 1) {
+  if (context.getNumInputs() != 1) {
```
jijoongmoon, line 39: @zhoonit Is it ok for CTC of TIZEN if we remove the return value?

nntrainer/layers/layer_node.cpp
```diff
@@ -159,6 +172,32 @@ bool LayerNode::setProperty(const std::string &key, const std::string &value) {
...
+        in_dim.batch(1);
+      }
+      int status = in_dim.setTensorDim(value.c_str());
+      if (in_dim.batch() > 1) {
+        ml_logw("Batch size set with input dimension %d is ignored."
+                "Set batchsize property for the model to update batchsize.",
+                in_dim.batch());
+      }
+      /** set back to cache value of dimension */
+      in_dim.batch(cache_batch_size);
```
jijoongmoon, line 197: I guess it can be done later PR. If in_dim is a temporal variable then we could set it into context.

jijoongmoon (approved):
LGTM except for small recommendation.

nntrainer/layers/layer_node.cpp
kparichay: yes, I will do this in the next PR.

nntrainer/layers/fc_layer.cpp
zhoonit: Yes, our api captures throw and convert back as an integer.

PR #1330, opened 2021-06-23
[losslayer] Update loss layer with LayerV2 design @open sesame 07/05 16:40

kparichay:
Update loss layer with LayerV2 design.


nntrainer/layers/loss/loss_layer.cpp
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright (C) 2020 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
zhoonit, line 4: using spdx?

nntrainer/app_context.cpp
```diff
@@ -255,6 +256,19 @@ static void add_default_object(AppContext &ac) {
   ac.registerFactory(nntrainer::createLayer<PermuteLayer>, PermuteLayer::type,
                      LayerType::LAYER_PERMUTE);
 
+  /** register losses */
+  ac.registerFactory(nntrainer::createLayer<MSELossLayer>, MSELossLayer::type,
+                     LayerType::LAYER_LOSS_MSE);
+  ac.registerFactory(nntrainer::createLayer<CrossEntropyLossLayer>,
```
zhoonit, line 262: Wonder if we need to keep CrossEntropyLossLayer

nntrainer/graph/network_graph.cpp
```diff
@@ -265,8 +269,9 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
 
   auto const &updated_last_node = getSortedLayerNode(graph.size() - 1);
 
-  std::shared_ptr<LayerNode> lnode = createLayerNode(LossLayer::type);
-  lnode->setLossType(updated_loss_type);
+  // FIXME: fixme
```
zhoonit, line 271: Maybe comment missing?

nntrainer/graph/network_graph.cpp
```diff
@@ -214,16 +216,18 @@ int NetworkGraph::realizeMultiOutputType(
...
  * layer and the there is only one loss, this assumption is not true
  */
-int NetworkGraph::addLossLayer(const LossType loss_type) {
+int NetworkGraph::addLossLayer(const std::string &loss_type) {
   int status = ML_ERROR_NONE;
   auto const &last_node = LNODE(graph.getSortedNode(graph.size() - 1));
   auto last_layer_node = getSortedLayerNode(graph.size() - 1);
 
   if (last_layer_node->requireLabel()) {
+    if (last_layer_node->getType() != loss_type)
```
zhoonit, line 224: istrequal here?

nntrainer/app_context.cpp
zhoonit: Maybe we need as cross entropy can be the last layer

nntrainer/graph/network_graph.cpp
zhoonit: Case handling for cross entropy might be needed here.

nntrainer/graph/network_graph.cpp
```diff
@@ -236,9 +240,9 @@ int NetworkGraph::addLossLayer(const LossType loss_type) {
   /// @todo enable this
   /// if (num_layer_that_requires_label > 2) { return error; }
 
-  LossType updated_loss_type = loss_type;
+  std::string updated_loss_type = loss_type;
 
-  if (updated_loss_type == LossType::LOSS_ENTROPY) {
+  if (updated_loss_type == CrossEntropyLossLayer::type) {
```
zhoonit, line 244: istrequal here?

nntrainer/graph/network_graph.cpp
```diff
@@ -363,13 +368,14 @@ int NetworkGraph::checkCompiledGraph() {
...
   /** First layer cannot be activation, batch normalization or loss */
   const std::string &type = l->getType();
-  if (istrequal(type, ActivationLayer::type) ||
-      istrequal(type, BatchNormalizationLayer::type) ||
-      istrequal(type, LossLayer::type)) {
-    ml_loge("%s cannot be the first layer, type: %s", l->getName().c_str(),
-            type.c_str());
-    return ML_ERROR_INVALID_PARAMETER;
-  }
+  // FIXME: why such restriction exists
```
zhoonit, line 370: I think this was related to support Inplace

nntrainer/layers/entropy_layer.h
```diff
@@ -0,0 +1,110 @@
...
+enum class LossType {
+  LOSS_MSE,             /** Mean Squared Error */
+  LOSS_ENTROPY,         /** Cross Entropy */
+  LOSS_NONE,            /** No loss for this model */
+  LOSS_ENTROPY_SIGMOID, /** Cross Entropy amalgamated with sigmoid for stability
+                         */
+  LOSS_ENTROPY_SOFTMAX, /** Cross Entropy amalgamated with softmax for stability
+                         */
+  LOSS_UNKNOWN          /** Unknown */
+};
```
zhoonit: Can we delete this?

nntrainer/app_context.cpp
kparichay: Cross is opened with the APIs, so it kept. This layer can be created and added, but cant be finalized.
This must be merged with appropriate activations for it work.

nntrainer/layers/loss/loss_layer.cpp
kparichay: this was old code, but updated.

nntrainer/graph/network_graph.cpp
kparichay: removed.

nntrainer/graph/network_graph.cpp
kparichay: This is adding restriction on the first layer. If the user wants it, let them place it. 
Support inPlace should not be affected with this. 
Support inPlace must adjust itself than restricting the model architectures.

kparichay: Removed the commented code.

nntrainer/layers/entropy_layer.h
kparichay: done.

nntrainer/graph/network_graph.cpp
kparichay: > Case handling for cross entropy might be needed here.

You mean for cross vs cross_softmax/sigmoid? 
In the current code, the existing loss layer is a layer added by the user to the model in model description or the API. Below are some scenarios:
1. Already added = cross, loss_type = cross - PASS
2. Already added = cross_sigmoid, loss_type = cross - FAIL as the behavior of the two scenrios become different (in the former case, last layer before loss can be any layer, but for latter case, the previous layer must be sigmoid/softmax).

nntrainer/graph/network_graph.cpp
kparichay: done.

nntrainer/graph/network_graph.cpp
kparichay: > istrequal here?

done.

zhoonit: Thanks for the explanation.

Even for case 1. Isn't it invalid parameter?

since added cross but trying to add another cross after it, something like below?

```cpp
if(last_layer.requireLabel()) {
 return loss_string.empty() ? OK : ERROR
}
```

kparichay: I understand now. You prefer treating setting loss twice as an error.
This is being handled with @lhs8928 when he support multiple losses.
This PR just provides a temporary handling for it.

zhoonit: Cool :)

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1332, opened 2021-06-24
[layernode] Update throw to retval in setProperty 

kparichay:
Update throw to retval in setProperty when the property
is not set by LayerNode. Illegal properties vs non-captured properties
were being mixed in this case.
Now it has been separated by return false. If setProperty() throws, it
must propagate upwards as that error shows the error in the property
given.

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -132,18 +132,14 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
     status = getKeyValue(left_properties[i], key, value);
     NN_RETURN_STATUS();
 
-    unsigned int type = parseLayerProperty(key);
-
     if (value.empty()) {
       ml_logd("value is empty for layer: %s, key: %s, value: %s",
               getName().c_str(), key.c_str(), value.c_str());
       return ML_ERROR_INVALID_PARAMETER;
     }
 
-    try {
-      /// @note this calls derived setProperty if available
-      setProperty(static_cast<nntrainer::LayerV1::PropertyType>(type), value);
-    } catch (...) {
+    /// @note this calls derived setProperty if available
+    if (!setProperty(key, value)) {
       remainder.push_back(left_properties[i]);
     }
   }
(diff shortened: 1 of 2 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -158,29 +154,31 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
...
-  case PropertyType::activation:
-    if (!value.empty()) {
-      setActivation((ActivationType)parseType(value, TOKEN_ACTI));
-      if (getType() == ActivationLayer::type)
-        throw std::invalid_argument(
-          "Set property delegated to activation layer");
+  case PropertyType::activation: {
+    setActivation((ActivationType)parseType(value, TOKEN_ACTI));
+    if (getType() == ActivationLayer::type) {
+      ml_logi("Set property delegated to activation layer");
```
zhoonit, line 165: Is this temporary fix?

I think we should have hard assumption that key should map to the exact one property.

In this case, I think we can have type=relu, type=sigmoid those kind of way.

kparichay: The activation property can be interpreted in two ways :
- one by a non-activation layer - in this case, layernode  intercepts it and adds an activation layer
- one by activation layer - in this case, layernode let it through
Ideally, if all activation layers have been split, then the second use would go away.

zhoonit: How about we finalize if activation type is there and underlying layer type is activation, we set at their?

kparichay: It will lead to different behavior once activations are split into different classes.
Giving `activation` property to an `activation layer currently sets its type, but later, `activation` property will be fixed to adding a new activation layer.

> How about when we finalize() if activation type is there and underlying layer type is activation, we set at their, in this way we can keep the assumption

Further, LayerNode should not store the activation_type information once the node has been realized. With this style, the `activation_type` variable now has two meanings depending on the type of the layer.

zhoonit: With or without the suggested style, there is already two meanings on the activation type. So what to decide here is if we are delegating second meaning to the activation Layer or let layer_node handle it.

If something else should be processed _based on_ property, it should be strictly done at `finalize()`. because what we are doing is kind like a builder pattern without a builder.

For example, imagine `setProperty({distribute = true, distribute = false}); then there will be distribute layer inside distribute layer(current code) even if we override it to false. In order to prevent this behavior, I prohibited to set `distribute` twice, but it's just a temporary solution that shouldn't be done like that.

Another (not possible but plausible) scenario would be we don't even have a layer when setting a property.

So, the unified semantics that can cover every scenario is that we delay whatever processing depending on the property at `finalize()`. For now, activation is just a type so it wouldn't make a big difference, but if we are directly creating a new activation layer, that any has to be strictly done at `finalize()` phase anyway, in the same manner, I think it is sensible to handle `activation layer property in layer_node` at the `finalize()` phase.

kparichay: currently, all the properties intercepted by the layer_node are handled in the same way.
Many of these properties result in changes in the graph such as flatten, distribute and activation.
Delaying setting the activation to finalize would require doing it for all the properties.

Further, from the point of view of long term, getters would also have issue where some properties like activation are being returned by the layer_node but others are not. Activation issue can be mellowed by splitting all activations into each layer.

However, handling the bigger issues you added about setting properties like above are beyond the scope of this PR.
I will add it as a TODO.

zhoonit: true, thanks for the consideration :)

zhoonit (approved):
LGTM :)

jijoongmoon (approved):
LGTM

PR #1333, opened 2021-06-24
[layer/model] Reduce dependence on LayerV1

kparichay:
Reduce dependence on LayerNode()->getObject() which returns LayerV1
object.

Reduce the usage of getObject() for optimizer.
This patch changes optimizer interface to work on each weight
individually and adds a getWeightObject() for the LayerContext.
This allows removal of getObject() from the neural network class for
backwarding.

This patch also disable dynamic_training_optimization.

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -334,8 +334,8 @@ void LayerNode::finalize() {
 void LayerNode::forwarding(bool training) {
   if (layerv1)
     layerv1->forwarding(training);
-      else
-  layer->forwarding(run_context, training);
+  else
+    layer->forwarding(run_context, training);
 }
 
 /**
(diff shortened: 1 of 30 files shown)


zhoonit:
Could you clarify how many commit should be reviewed? If it's directly depending on #1332, it's looks like there is three commits.

kparichay:
> Could you clarify how many commit should be reviewed? If it's directly depending on #1332, it's looks like there is three commits.

Yes, there are 3 commits to be reviewed.

nntrainer/layers/layer_node.cpp
```diff
@@ -335,29 +332,46 @@ void LayerNode::finalize() {
  * @brief     Forward Propagation of a layer
  */
 void LayerNode::forwarding(bool training) {
+  if (layerv1)
+    layerv1->forwarding(training);
+      else
```
zhoonit: How clang-format formats is quite weird 🤣

zhoonit (approved):
LGTM

nntrainer/layers/layer_node.cpp
kparichay: yeah, i dont know why.

zhoonit:
Android build failing. I wonder this is intended?

jijoongmoon (approved):
LGTM

PR #1336, opened 2021-06-24
[Resnet] Create resnet model @open sesame 06/30 16:02

zhoonit:
- [Resnet] Create resnet model

```
This patch creates code to create resnet.

**Self evaluation:**

```

### DIFF
--- a/Applications/Resnet/jni/main.cpp
+++ b/Applications/Resnet/jni/main.cpp
@@ -4,6 +4,7 @@
  *
  * @file   main.cpp
  * @date   24 Jun 2021
+ * @todo   move resnet model creating to separate sourcefile
  * @brief  task runner for the resnet
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
@@ -16,6 +17,7 @@
 
 #include <layer.h>
 #include <model.h>
+#include <optimizer.h>
 
 using LayerHandle = std::shared_ptr<ml::train::Layer>;
 using ModelHandle = std::unique_ptr<ml::train::Model>;
@@ -31,23 +33,25 @@ using ModelHandle = std::unique_ptr<ml::train::Model>;
 template <typename T>
 static std::string withKey(const std::string &key, const T &value) {
   std::stringstream ss;
-  ss << "key=" << value;
+  ss << key << "=" << value;
   return ss.str();
 }
 
 template <typename T>
 static std::string withKey(const std::string &key,
-                           const std::vector<T> &value) {
-  if (value.empty()) {
-    throw std::invalid_argument("empty vector cannot be converted");
+                           std::initializer_list<T> value) {
+  if (std::empty(value)) {
+    throw std::invalid_argument("empty data cannot be converted");
   }
 
   std::stringstream ss;
-  ss << "key=";
-  for (unsigned int i = 0; i < value.size() - 1; ++i) {
-    ss << value.at(i) << ',';
+  ss << key << "=";
+
+  auto iter = value.begin();
+  for (; iter != value.end() - 1; ++iter) {
+    ss << *iter << ',';
   }
-  ss << value.back();
+  ss << *iter;
 
   return ss.str();
 }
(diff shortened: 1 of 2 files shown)


Applications/Resnet/jni/main.cpp
```diff
@@ -65,23 +68,140 @@ static std::string withKey(const std::string &key,
...
+                {withKey("unit", 100), withKey("activation", "softmax")}));
+
+  return layers;
+}
+
+ModelHandle createResnet18() {
+  ModelHandle model =
+    ml::train::createModel(ml::train::ModelType::NEURAL_NET,
+                           {withKey("loss", "cross"),
+                            withKey("batch_size", 128), withKey("epochs", 60)});
```
kparichay, line 180: You might want to reduce training times by a smaller number of epochs (possibly use pre-trained resnet18 for initialization) for a later PR.

Applications/Resnet/jni/main.cpp
```diff
@@ -65,23 +68,140 @@ static std::string withKey(const std::string &key,
...
+
+  for (auto layers : createResnet18Graph()) {
+    model->addLayer(layers);
```
kparichay, line 183: just for debug purposes, can you shuffle the layers vector and then add layers - it should result in the same model.

Applications/Resnet/jni/main.cpp
```diff
@@ -65,23 +68,140 @@ static std::string withKey(const std::string &key,
...
+  return layers;
+}
+
+ModelHandle createResnet18() {
+  ModelHandle model =
+    ml::train::createModel(ml::train::ModelType::NEURAL_NET,
+                           {withKey("loss", "cross"),
+                            withKey("batch_size", 128), withKey("epochs", 60)});
+
+  for (auto layers : createResnet18Graph()) {
```
kparichay, line 182: another suggestion, to be able to extend to other resnet models, use width, and number of block as a configuration variable. It does not has to be done in this PR, but keep it as a TODO.

Applications/Resnet/jni/main.cpp
```diff
@@ -65,23 +68,140 @@ static std::string withKey(const std::string &key,
...
+  }
+
+  layers.push_back(createLayer("pooling2d", {withKey("name", "last_p1"),
+                                             withKey("pooling", "average"),
+                                             withKey("pool_size", {4, 4})}));
+
+  layers.push_back(createLayer("flatten", {withKey("name", "last_f1")}));
+  layers.push_back(
+    createLayer("fully_connected",
+                {withKey("unit", 100), withKey("activation", "softmax")}));
```
kparichay, line 171: i recommend taking `100` as a label size, and move resnet model creation to `resnet_model.cpp` for modularity.

Applications/Resnet/jni/main.cpp
```diff
@@ -65,23 +68,140 @@ static std::string withKey(const std::string &key,
...
+ * @return vector of layers that contain full graph of resnet18
  */
-ModelHandle createResnet18() {
+std::vector<LayerHandle> createResnet18Graph() {
+  using ml::train::createLayer;
+
   std::vector<LayerHandle> layers;
-  return nullptr;
+
+  layers.push_back(
```
kparichay, line 135: i wonder where is the input layer? :)

Applications/Resnet/jni/main.cpp
zhoonit: yes, I'll tune epochs when training

Applications/Resnet/jni/main.cpp
zhoonit: I think there is already a bug, I'll do that after verification process is done :)

Applications/Resnet/jni/main.cpp
zhoonit: Yes, as we have resnet block it can be easily configurable. I"ll add it as a todo

Applications/Resnet/jni/main.cpp
zhoonit: Do we have label size property? could you explain a bit more?

Applications/Resnet/jni/main.cpp
zhoonit: when we have input_shape property, we don't need input layer. Could you clarify this?

Applications/Resnet/jni/main.cpp
zhoonit: I'll consider moving to resnet_model.cpp when tidying up. For now main is empty so, it's less complex

Applications/Resnet/jni/main.cpp
kparichay: alright.

Applications/Resnet/jni/main.cpp
kparichay: > Do we have label size property? could you explain a bit more?

Its just to modularize createResnetModel() function, not for layer property. As 100 is being unit size for last layer, keep it as a property of createResnetModel() will allow to use resnet for different size of labels. 

On a side note, when tidying up, add an option to load with/without last layer.

Applications/Resnet/jni/main.cpp
kparichay: Isn't input layer necessary for model creation. I maybe wrong here.

zhoonit: afaik, if input_shape is there it is treated input. I have tested it's running fine but if there is a problem, I'll look into it :)

Applications/Resnet/jni/main.cpp
zhoonit: Ah, I see, I'll take care of it on upcoming PRs :).

> On a side note, when tidying up, add an option to load with/without last layer.

Could you explain for this? is it for to skip loss layer? or more like `include_top = false` in keras?

kparichay: Yes.

Applications/Resnet/jni/main.cpp
kparichay: cool.

Applications/Resnet/jni/main.cpp
kparichay: > I'll consider moving to resnet_model.cpp when tidying up. For now main is empty so, it's less complex

Please also consider exposing two functions with c and c++ API model data types.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1339, opened 2021-06-25
[resnet] Implement test run to resnet

zhoonit:
- [resnet] Implement test run to resnet

```
**Changes proposed**
- Add dataloader with interface `next()`.
- Add RandomDataLoader as an example
- Minor bug fixes to the model architecture regarding name
- Add routine to train the model

**Self evaluation:**

```

### DIFF
--- a/Applications/Resnet/jni/cifar_dataloader.cpp
+++ b/Applications/Resnet/jni/cifar_dataloader.cpp
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   cifar_dataloader.h
+ * @date   24 Jun 2021s
+ * @brief  dataloader for cifar
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#include "cifar_dataloader.h"
+
+#include <nntrainer_error.h>
+#include <random>
+
+namespace nntrainer::resnet {
+
+RandomDataLoader::RandomDataLoader(const std::vector<TensorDim> &input_shapes,
+                                   const std::vector<TensorDim> &output_shapes,
+                                   int data_size) :
+  iteration(0),
+  iteration_for_one_epoch(data_size),
+  input_shapes(input_shapes),
+  output_shapes(output_shapes),
+  input_dist(0, 255),
+  label_dist(0, output_shapes.front().width() - 1) {
+  NNTR_THROW_IF(output_shapes.empty(), std::invalid_argument)
+    << "output_shape size empty not supported";
+  NNTR_THROW_IF(output_shapes.size() > 1, std::invalid_argument)
+    << "output_shape size > 1 is not supported";
+
+  iteration_for_one_epoch /= output_shapes.front().batch();
+}
+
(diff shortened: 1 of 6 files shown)


zhoonit:
v2. There is a patch added to prohibit testing, (it takes around two minutes in my machine so disabled with an option)

jijoongmoon (approved):
LGTM

zhoonit:
> Do you plan to implement resnet with ini format as well (may be using backbone)?

Yes, after saving api I am thinking to make one directly :)

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -0,0 +1,105 @@
...
+          *label = (generated_label == j);
+          label++;
+        }
+        break;
+      }
+      }
+    }
+  };
+
+  if (fill_last() == true) {
```
kparichay, line 77: so in this case, fill_last does not fill in any data?

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -0,0 +1,105 @@
...
+      case 1: { /// case of single integer value
+        *label = generated_label;
+        label++;
+        break;
+      }
+      default: { /// case of one hot
+        for (unsigned int j = 0; j < length; ++j) {
+          *label = (generated_label == j);
+          label++;
+        }
```
kparichay, line 70:
```suggestion
if (generated_label >= length)
  throw std::invalid_argument("Label out of range");
*(label + generated_label) = 1;
label += length;
```

Applications/Resnet/jni/main.cpp
```diff
@@ -136,6 +139,8 @@ std::vector<LayerHandle> createResnet18Graph() {
   layers.push_back(
     createLayer("conv2d", {
                             withKey("name", "conv0"),
+                            withKey("input_shape", "3:32:32"),
```
kparichay, line 142: consider using `input_layer` than setting input_shape for a given layer as a recommended practice, as this example will form benchmark for upcoming models.

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: is the label zero initialized? from nntrainer?

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: Yes

Applications/Resnet/jni/cifar_dataloader.cpp
kparichay: maybe not. Consider memset to set zero to the memory, and then set the one_hot value at once than a loop.

Applications/Resnet/jni/main.cpp
zhoonit: I'll do this in the upcoming PR, I think this will conflict consecutively.

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: Alright, I'll do this in the upcoming PR

zhoonit: Updated to #1340

Applications/Resnet/jni/main.cpp
kparichay: alright.

kparichay (approved):
LGTM

PR #1340, opened 2021-06-25
[Wait for #1339] [Resnet] Connect the model with cifar100  @open sesame 06/30 20:04

zhoonit:
**[Pending commit: #1339]**

- [Resnet] Connect the model with cifar100

```
- Implement cifar100dataloader
- Connect the data loader

**Self evaluation:**

```

### DIFF
--- a/Applications/Resnet/jni/cifar_dataloader.cpp
+++ b/Applications/Resnet/jni/cifar_dataloader.cpp
@@ -12,10 +12,49 @@
 
 #include "cifar_dataloader.h"
 
+#include <cstring>
 #include <nntrainer_error.h>
 #include <random>
 
 namespace nntrainer::resnet {
+
+namespace {
+
+/**
+ * @brief fill label to the given memory
+ *
+ * @param data data to fill
+ * @param length size of the data
+ * @param label label
+ */
+void fillLabel(float *data, unsigned int length, unsigned int label) {
+  if (length == 1) {
+    *data = label;
+    return;
+  }
+
+  memset(data, 0, length * sizeof(float));
+  *(data + label) = 1;
+}
+
+/**
+ * @brief fill last to the given memory
+ * @note this function increases iteration value, if last is set to true,
+ * iteration resets to 0
+ *
+ * @param[in/out] iteration current iteration
+ * @param iteration_per_epoch iteration per epoch
+ * @return bool true if iteration has finished
+ */
+bool setIteration(unsigned int &iteration, unsigned int iteration_per_epoch) {
+  if (iteration++ == iteration_per_epoch) {
+    iteration = 0;
+    return true;
+  }
+  return false;
+};
+
+} // namespace
 
 RandomDataLoader::RandomDataLoader(const std::vector<TensorDim> &input_shapes,
                                    const std::vector<TensorDim> &output_shapes,
@@ -35,16 +74,6 @@ RandomDataLoader::RandomDataLoader(const std::vector<TensorDim> &input_shapes,
 }
 
 void RandomDataLoader::next(float **input, float **label, bool *last) {
-  auto fill_last = [&last, this] {
-    if (iteration++ == iteration_for_one_epoch) {
-      iteration = 0;
-      *last = true;
-    } else {
-      *last = false;
-    }
-    return *last;
-  };
-
   auto fill_input = [this](float *input, unsigned int length) {
     for (unsigned int i = 0; i < length; ++i) {
       *input = input_dist(rng);
@@ -56,25 +85,13 @@ void RandomDataLoader::next(float **input, float **label, bool *last) {
                            unsigned int length) {
     for (unsigned int i = 0; i < batch; ++i) {
       unsigned int generated_label = label_dist(rng);
-
-      switch (length) {
-      case 1: { /// case of single integer value
-        *label = generated_label;
-        label++;
-        break;
-      }
-      default: { /// case of one hot
-        for (unsigned int j = 0; j < length; ++j) {
-          *label = (generated_label == j);
-          label++;
-        }
-        break;
-      }
-      }
+      fillLabel(label, length, generated_label);
+      label += length;
     }
   };
 
-  if (fill_last() == true) {
+  if (setIteration(iteration, iteration_for_one_epoch)) {
+    *last = true;
     return;
   }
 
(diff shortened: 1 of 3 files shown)


Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -12,11 +12,58 @@
...
+void fillLabel(float *data, unsigned int length, unsigned int label) {
+  if (length == 1) {
+    *data = label;
+    return;
+  }
+
+  int sum = 0;
+  /// if label is onehot
+  for (unsigned int j = 0; j < length; ++j) {
+    *data = (label == j);
```
kparichay, line 39: `*(data + label) = 1`;

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -93,13 +118,65 @@ void RandomDataLoader::next(float **input, float **label, bool *last) {
...
+    << error_msg << " Given file does not align with the format";
+
+  auto data_size = pos / (Cifar100DataLoader::SampleSize * splits);
+  idxes = std::vector<unsigned int>(data_size);
+  std::cout << "path: " << path << '\n';
+  std::cout << "data_size: " << data_size << '\n';
+  std::iota(idxes.begin(), idxes.end(), 0);
+  std::shuffle(idxes.begin(), idxes.end(), rng);
+
+  iteration_per_epoch = data_size / batch_size;
```
kparichay, line 142: This code ignores the last batch if data_size is not a multiple of batch_size. If this is intended, please add as a comment.

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -12,11 +12,58 @@
...
+/**
+ * @brief fill last to the given memory
+ * @note this function increases iteration value, if last is set to true,
+ * iteration resets to 0
+ *
+ * @param[in/out] iteration current iteration
+ * @param iteration_per_epoch iteration per epoch
+ * @return bool true if iteration has finished
+ */
+bool isLast(unsigned int &iteration, unsigned int iteration_per_epoch) {
```
kparichay, line 57: how about rename to `iterate` as it seems like the main purpose is to increase iteration?

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -93,13 +118,65 @@ void RandomDataLoader::next(float **input, float **label, bool *last) {
...
+  unsigned int idx_pos = current_iteration * batch;
+  if (isLast(current_iteration, iteration_per_epoch)) {
+    *last = true;
+    std::shuffle(idxes.begin(), idxes.end(), rng);
+    return;
+  }
+  unsigned int idx_pos_end =
+    current_iteration * batch; // iteration increased by 1
+
+  float *cur_input_tensor = *input;
```
kparichay, line 173: please add a comment that only a single input and label is being provided to the model here.

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: Done

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: Done

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: Updated to line 139

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: changed to `setIteration`

kparichay: minor - `updateIteration`? `setIteration` is fine too btw.

kparichay (approved):
LGTM

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: I'll change to updateIteration after validation :)

kparichay: cool.

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -93,13 +118,65 @@ void RandomDataLoader::next(float **input, float **label, bool *last) {
...
+
+  auto fill_one_sample = [this](float *input_, float *label_, int index) {
+    NNTR_THROW_IF(!file.good(), std::invalid_argument)
+      << "file is not good, reason: " << std::strerror(errno);
+    file.seekg(index * Cifar100DataLoader::SampleSize + 1, std::ios_base::beg);
+
+    uint8_t current_label;
+    file.read(reinterpret_cast<char *>(&current_label), sizeof(uint8_t));
+    fillLabel(label_, Cifar100DataLoader::NumClass, current_label);
+
```
jijoongmoon, line 149: do you assume that the dataset stores the data first or label first? it looks label data first. Could you check this? ( If you are using old dataset we used use, then the data is the first )

Applications/Resnet/jni/cifar_dataloader.h
```diff
@@ -107,6 +108,17 @@ class Cifar100DataLoader final : public DataLoader {
   void next(float **input, float **label, bool *last);
 
 private:
+  inline static constexpr int ImageSize = 3 * 32 * 32;
+  inline static constexpr int NumClass = 100;
+  inline static constexpr int SampleSize =
+    1 + 1 + 3 * 32 * 32; /**< 1 coarse label, 1 fine label, pixel size */
+
```
jijoongmoon, line 115: What do you mean coarse label and fine label? Could elaborate little bit?

Applications/Resnet/jni/cifar_dataloader.cpp
```diff
@@ -93,13 +118,65 @@ void RandomDataLoader::next(float **input, float **label, bool *last) {
...
+  NNTR_THROW_IF(!file.good(), std::invalid_argument)
+    << error_msg << " Cannot open file";
+
+  auto pos = file.tellg();
+  NNTR_THROW_IF((pos % Cifar100DataLoader::SampleSize != 0),
+                std::invalid_argument)
+    << error_msg << " Given file does not align with the format";
+
+  auto data_size = pos / (Cifar100DataLoader::SampleSize * splits);
```
jijoongmoon, line 127: It looks not handle when the splited data size is smaller than batch. It looks like just set the iteration_per_epoch is zero and not run. Is it intended?

jijoongmoon (approved):
LGTM with small recommandations.

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: This is directly using the raw data available from the official website. The format is slightly different, so this is using generator and instead of file based.

Applications/Resnet/jni/cifar_dataloader.h
zhoonit: Below raw format is used, cited from https://www.cs.toronto.edu/~kriz/cifar.html.

We are only using fine label instead of coarse label though :)
```
Binary version
The binary version of the CIFAR-100 is just like the binary version of the CIFAR-10, except that each image has two label bytes (coarse and fine) and 3072 pixel bytes, so the binary files look like this:
<1 x coarse label><1 x fine label><3072 x pixel>
...
<1 x coarse label><1 x fine label><3072 x pixel>
```

Applications/Resnet/jni/cifar_dataloader.cpp
zhoonit: Afaik, we do yet to have a good way to handle the iteration less than the batch size. 
I'll make this error when there is no batch size, in a new PR :)

kparichay: check drop_remainder property in https://www.tensorflow.org/api_docs/python/tf/data/Dataset.

zhoonit: The concern is commented at line 134.

PR #1345, opened 2021-06-30
[CAPI] Propose save/load api @open sesame 07/05 16:40

zhoonit:
- [CAPI] Propose save/load api

```
**Motivations**
1. Fine grained api required to save and load

- Add function to model.h  / `save`, `load`
- Mark `readModel()` deprecated (later remove...)
- Mark `saveModel()` to change `exportTo`
- Add `ModelSaveLoadFormat`
- Create capi-machine-learning-training-tizen-internal for internal api
support

**Self evaluation:**

```

### DIFF
--- a/api/capi/include/nntrainer-tizen-internal.h
+++ b/api/capi/include/nntrainer-tizen-internal.h
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   nntrainer-tizen-internal.h
+ * @date   30 June 2021
+ * @brief  NNTrainer CAPI header for tizen interanl api.
+ * @note   This header is designed to be used only in Tizen
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#ifndef __NNTRAINER_TIZEN_INTERNAL_H__
+#define __NNTRAINER_TIZEN_INTERNAL_H__
+
+#include <nntrainer.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Save the model
+ * @details Use this function to save the current model. @a format
+ * describes various formats in which various selections of the
+ * parameters of the models can be saved. Some formats may save
+ * parameters required for training. Some other formats may save model
+ * configurations. Unless stated otherwise, @a ml_train_model_compile() has to
+ * be called upon the @a model before calling this function.
+ * @see @a ml_train_model_format_e to check which part of the model is
+ * saved
+ * @note This function overrides the existing file without any notification.
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to save
+ * @param[in] file_path File path to save the file.
+ * @param[in] format Format flag to determine which format should be used to
+ * save
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER The given @a file_path is
+ * invalid or @a model is not compiled.
+ */
+int ml_train_model_save(ml_train_model_h model, const char *file_path,
+                        ml_train_model_format_e format);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model. @a format
+ * describes various formats in which various selections of the
+ * parameters of the models can be loaded. Some formats may load
+ * parameters required for training. Some other formats may load model
+ * configurations. Unless stated otherwise, @a ml_train_model_compile() has to
+ * be called upon the @a model before calling this function.
+ *
+ * @see @a ml_train_model_format_e to check which part of the model is
+ * loaded
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to load.
+ * @param[in] file_path File path to load the file.
+ * @param[in] format Format flag to determine which format should be used to
+ * loaded
+ * @return @c 0 on success, Otherwise a negative error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER The given @a file_path is
+ * invalid or @a model is not in valid state to load.
+ */
+int ml_train_model_load(ml_train_model_h model, const char *file_path,
+                        ml_train_model_format_e format);
+
+#ifdef __cplusplus
+}
+#endif /* __cplusplus */
+#endif // __NNTRAINER_TIZEN_INTERNAL_H__
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -19,6 +19,7 @@
 #if __cplusplus >= MIN_CPP_VERSION
 
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include <nntrainer-api-common.h>
@@ -38,6 +39,21 @@ enum class ModelType {
   KNN,        /** k Nearest Neighbor */
   NEURAL_NET, /** Neural Network */
   UNKNOWN     /** Unknown */
+};
+
+/**
+ * @brief Model saving options
+ *
+ */
+enum class ModelFormat {
+  MODEL_FORMAT_BIN =
+    ML_TRAIN_MODEL_FORMAT_BIN, /**< raw bin file saves parameters required
+for inference and training without any configurations*/
+
+  /** Comment intended */
+  // MODEL_FORMAT_INI = 1, /**< ini file with save_path defined */
+  // MODEL_FORMAT_INI_WITH_BIN = 2, /**< ini file with save_path
+  // defined */
 };
 
 /**
@@ -95,13 +111,35 @@ public:
 
   /**
    * @brief     save model and training parameters into file
+   * @todo      deprecate this
    */
-  virtual void saveModel() = 0;
+  [[deprecated("use saveModel(const std::string &path_prefix, "
+               "ModelFormat format)")]] virtual void
+  saveModel() = 0;
+
+  /**
+   * @brief  load model states and training parameters from a file
+   * @param file_path file_path to save the model, if full path is not
+   * given, it should be saved inside working directory
+   * @param format format to save parameters
+   */
+  virtual void save(const std::string &file_path,
+                    ModelFormat format = ModelFormat::MODEL_FORMAT_BIN){};
 
   /**
    * @brief     read model and training parameters from file
+   * @todo      deprecate this
    */
   virtual void readModel() = 0;
+
+  /**
+   * @brief  load model with regard to the format
+   * @param file_path file_path to save the model, if full path is not
+   * given, it should be saved inside working directory
+   * @param format format to save parameters
+   */
+  virtual void load(const std::string &file_path,
+                    ModelFormat format = ModelFormat::MODEL_FORMAT_BIN){};
 
   /**
    * @brief     Run Model training and validation
--- a/api/nntrainer-api-common.h
+++ b/api/nntrainer-api-common.h
@@ -164,6 +164,16 @@ typedef enum {
 } ml_train_summary_type_e;
 
 /**
+ * @brief Enumeration for the neural network
+ *
+ */
+typedef enum {
+  ML_TRAIN_MODEL_FORMAT_BIN =
+    0, /**< raw bin file saves parameters required for inference and training
+          without any configurations*/
+} ml_train_model_format_e;
+
+/**
  * @}
  */
 #ifdef __cplusplus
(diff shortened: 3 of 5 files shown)


zhoonit:
@myungjoo The comment from #1326 is reflected here. Please take another look :)

zhoonit:
@myungjoo, @jijoongmoon The comment from #1326 has been reflected here. PTAL :)

api/nntrainer-api-common.h
```diff
@@ -163,6 +163,16 @@ typedef enum {
                              information */
 } ml_train_summary_type_e;
 
+/**
+ * @brief Enumeration for the neural network
+ *
+ */
+typedef enum {
+  ML_TRAIN_MODEL_SAVE_LOAD_BIN =
```
myungjoo, line 171: ```ML_TRAIN_MODEL_FORMAT_BIN```?

api/ccapi/include/model.h
```diff
@@ -40,6 +41,21 @@ enum class ModelType {
   UNKNOWN     /** Unknown */
 };
 
+/**
+ * @brief Model saving options
+ *
+ */
+enum class ModelSaveLoadFormat {
+  MODEL_SAVE_LOAD_FORMAT_BIN =
```
myungjoo, line 49: ```MODEL_FORMAT_BIN```?
(SAVE-LOAD is unnecessary duplicated terms)

zhoonit: Yes, prefix save_load is deleted, across the patch.

api/nntrainer-api-common.h
zhoonit: Updated :)

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,74 @@
...
+ * parameters required for training. Some other formats may save model
+ * configurations. The model must be compiled and has allocated memory before
+ * calling this function.
+ * @see @a ml_train_model_format_e to check which part of the model is
+ * saved
+ * @note This function overrides the existing file without any notification.
+ *
+ * @since_tizen 6.5
+ * @param[in] model The NNTrainer model handle to save
+ * @param[in] file_path File path to save the file.
```
kparichay, line 37: is this going to be file path or file prefix?

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+int ml_train_model_save(ml_train_model_h model, const char *file_path,
+                        ml_train_model_format_e format);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model. @a format rules
+ * how and which part of the model should be saved. Some formats may save
+ * parameters required for training. Some other formats may save model
+ * configurations. The model must be compiled and has allocated memory before
```
kparichay, line 53: the state of the model depends on the `ml_train_model_format_e` value. Can you please specify it by case, and let other cases by TBD.

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,74 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Save the model after training
+ * @details Use this function to save the current model. @a format rules
+ * how and which part of the model should be saved. Some formats may save
+ * parameters required for training. Some other formats may save model
+ * configurations. The model must be compiled and has allocated memory before
```
kparichay, line 29: as there is no interface to allocate memory for the model, please change the requirement of the model state to be compiled/initialized depending on what the API states.

api/capi/include/nntrainer-tizen-internal.h
kparichay: as there is no interface to allocate memory for the model, please change the requirement of the model state to be compiled/initialized depending on what the API states.

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,74 @@
...
+ * invalid or @a model is not compiled.
+ */
+int ml_train_model_save(ml_train_model_h model, const char *file_path,
+                        ml_train_model_format_e format);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model. @a format rules
+ * how and which part of the model should be saved. Some formats may save
+ * parameters required for training. Some other formats may save model
```
kparichay, line 52: IMO I find these statements a little confusing. Maybe below statement can be better:
```suggestion
 * @details Use this function to load the current model. @a format 
 * describes various formats in which various selections of the 
 * parameters of the models can be saved.
```
Similar for load as well.

api/ccapi/include/model.h
```diff
@@ -40,6 +41,21 @@ enum class ModelType {
...
 
+/**
+ * @brief Model saving options
+ *
+ */
+enum class ModelFormat {
+  MODEL_FORMAT_BIN =
+    ML_TRAIN_MODEL_FORMAT_BIN, /**< raw bin file saves parameters required
+for inference and training without any configurations*/
+
```
jijoongmoon, line 52: at this time, we only save and load updated weights. ( there is only one format we support for now, FORMAT_BIN ).
That means, we are not update the ini. and if it is true, then the model save is always the same between training and infernece.

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,74 @@
...
+ * @param[in] file_path File path to save the file.
+ * @param[in] format Format flag to determine which format should be used to
+ * save
+ * @return @c 0 on success, Otherwise a negativse error value
+ * @retval #ML_ERROR_NONE Successful.
+ * @retval #ML_ERROR_INVALID_PARAMETER The given @a file_path is
+ * invalid or @a model is not compiled.
+ */
+int ml_train_model_load(ml_train_model_h model, const char *file_path,
+                        ml_train_model_format_e format);
```
jijoongmoon, line 69: We usually use different name of ini and bin. Even we have option to set the bin name in ini file. I wonder that it is also considered? ( file_path means ini file path?? )

api/capi/include/nntrainer-tizen-internal.h
zhoonit: It will be a single file. (if there are two file, one file will point to another one)

api/capi/include/nntrainer-tizen-internal.h
zhoonit: Done.

api/capi/include/nntrainer-tizen-internal.h
zhoonit: Done

api/capi/include/nntrainer-tizen-internal.h
zhoonit: Done

api/ccapi/include/model.h
zhoonit: Yes :) I think we will have `ML_TRAIN_MODEL_FORMAT_OPTIMIZED_FOR_INFERENCE`  this kind of enum.

api/capi/include/nntrainer-tizen-internal.h
zhoonit: Yes, if we state `ML_TRAIN_MODEL_FORMAT_INI_WITH_BIN`, bin name will be generated based on the ini file.

kparichay: So, in the two cases, with BIN/INI format, the user is expected to give the correct file format?

api/capi/include/nntrainer-tizen-internal.h
kparichay: so the file extension must be provided by the user then? how about we decide it based on the format?

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,77 @@
...
+#define __NNTRAINER_TIZEN_INTERNAL_H__
+
+#include <nntrainer.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Save the model after training
```
kparichay, line 25:
```suggestion
 * @brief Save the model
```
as calling save the model right after load is also eligible?

api/capi/include/nntrainer-tizen-internal.h
zhoonit: the saving file format will be ini as `INI_WITH_BIN` states. I think there has to be main format which points to another file, in the case of two files

api/capi/include/nntrainer-tizen-internal.h
zhoonit: I think it's commonly expected that user puts file path including the extension. Toco, pytorch, flatbuffers does not include extensions.

api/capi/include/nntrainer-tizen-internal.h
zhoonit: Done

api/capi/include/nntrainer-tizen-internal.h
kparichay: then our implementation will check if the extension matches the format (.bin for BIN and .ini for INI/INI+BIN)? if yes, add to documentation for the expected format.

api/capi/include/nntrainer-tizen-internal.h
kparichay: in case of ML_TRAIN_MODEL_FORMAT_BIN, it will be .bin right?

api/capi/include/nntrainer-tizen-internal.h
zhoonit: I think we don't need to, and should check or force the extension, as other's also support this, analogically, as we can edit something in notepad and can save as `.html`.

Adding this restriction is not to hard thing to do though :)

api/capi/include/nntrainer-tizen-internal.h
zhoonit: Yup

packaging/nntrainer.spec
```diff
@@ -215,6 +215,13 @@ Requires:        capi-machine-learning-training-devel = %{version}-%{release}
 %description -n capi-machine-learning-training-devel-static
 Static library of capi-machine-learning-training-devel package.
 
+%package -n capi-machine-learning-training-tizen-internal-devel
```
myungjoo, line 218: DO NOT ADD "capi" prefix for non public API packages.

zhoonit: Updated to nntrainer-tizen-internal-devel :)

api/capi/include/nntrainer-tizen-internal.h
kparichay: Yes, IMO the extension forcing just makes things easier to read.
@jijoongmoon what do you think?

api/capi/include/nntrainer-tizen-internal.h
kparichay: I think this discussion is leading to https://github.com/nnstreamer/nntrainer/pull/1345#discussion_r661396237. Let's continue it there.

api/capi/include/nntrainer-tizen-internal.h
zhoonit: After resolving this, i guess we can merge this?

kparichay (approved):
LGTM
The interface is good to go IMO, the enforcement of allowing only certain extensions can be discussed in the implementation.

myungjoo (commented):
LGTM

api/ccapi/include/model.h
```diff
@@ -95,14 +111,36 @@ class Model {
...
    * @brief     save model and training parameters into file
+   * @todo      deprecate this
    */
-  virtual void saveModel() = 0;
+  [[deprecated("use saveModel(const std::string &path_prefix, "
+               "ModelFormat format)")]] virtual void
+  saveModel() = 0;
+
+  /**
+   * @brief  load model states and training parameters from a file
```
lhs8928, line 121: please replace load with save

api/ccapi/include/model.h
```diff
@@ -95,14 +111,36 @@ class Model {
...
 
   /**
    * @brief     read model and training parameters from file
+   * @todo      deprecate this
    */
   virtual void readModel() = 0;
 
+  /**
+   * @brief  load model with regard to the format
+   * @param file_path file_path to save the model, if full path is not
```
lhs8928, line 137: please revise the @param. this document is for the save

lhs8928 (approved):
LGTM except the comment

api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -0,0 +1,77 @@
...
+ * invalid or @a model is not compiled.
+ */
+int ml_train_model_save(ml_train_model_h model, const char *file_path,
+                        ml_train_model_format_e format);
+
+/**
+ * @brief Load the model
+ * @details Use this function to load the current model. @a format
+ * describes various formats in which various selections of the
+ * parameters of the models can be saved. Some formats may save
```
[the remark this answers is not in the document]
zhoonit: Nice catch, updated :)

api/capi/include/nntrainer-tizen-internal.h
jijoongmoon: I guess we need to save with the name given by user. One thing we need to make sure is that this is for saving training result. not for the inference. We might have to have another api for inference later.. ( maybe after finishing tflite output generating )

jijoongmoon (approved):
LGTM.

PR #1348, opened 2021-07-01
[unittest] Enable models unittests @open sesame 07/05 19:36

kparichay:
Support input/output tensor allocation for layerv2 design

Enable models unittest for layerv2
Corresponding bugfixes are also added

Separate getNumInputs/Outputs semantics for inputs/outputs
and for connections as a node. Number of inputs/outputs will always be
more than 1, but number of input/output connections can be 0 for
input/output nodes of the graph.
This patch separates the two concepts, and its usage.

### DIFF
--- a/nntrainer/layers/activation_layer.cpp
+++ b/nntrainer/layers/activation_layer.cpp
@@ -47,7 +47,7 @@ void ActivationLayer::calcDerivative(RunLayerContext &context) {
   Tensor &ret = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
   Tensor &out = context.getOutput(SINGLE_INOUT_IDX);
 
-  ret = acti_func.run_prime_fn(out, ret, deriv);
+  acti_func.run_prime_fn(out, ret, deriv);
 }
 
 void ActivationLayer::setProperty(const std::vector<std::string> &values) {
--- a/nntrainer/layers/activation_layer.cpp
+++ b/nntrainer/layers/activation_layer.cpp
@@ -45,9 +45,9 @@ void ActivationLayer::forwarding(RunLayerContext &context, bool training) {
 void ActivationLayer::calcDerivative(RunLayerContext &context) {
   Tensor &deriv = context.getIncomingDerivative(SINGLE_INOUT_IDX);
   Tensor &ret = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
-  Tensor &in = context.getOutput(SINGLE_INOUT_IDX);
+  Tensor &out = context.getOutput(SINGLE_INOUT_IDX);
 
-  ret = acti_func.run_prime_fn(in, ret, deriv);
+  ret = acti_func.run_prime_fn(out, ret, deriv);
 }
 
 void ActivationLayer::setProperty(const std::vector<std::string> &values) {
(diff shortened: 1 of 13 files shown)


jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.cpp
```diff
@@ -427,6 +427,8 @@ void LayerNode::setBatch(unsigned int batch) {
       run_context.setBatch(batch);
       layer->setBatch(run_context, batch);
     } else {
+      // for (auto &dim : input_dim)
```
zhoonit, line 430: unintended comment?

nntrainer/layers/layer_node.h
```diff
@@ -441,6 +458,21 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
     }
   }
 
+  /**
+   * @brief Get the Weight object
+   *
+   * @param idx Identifier of the weight
+   * @return Tensor& Reference to the weight tensor
```
zhoonit, line 465: return is weight?

nntrainer/layers/activation_layer.cpp
```diff
@@ -45,9 +45,9 @@ void ActivationLayer::forwarding(RunLayerContext &context, bool training) {
 void ActivationLayer::calcDerivative(RunLayerContext &context) {
   Tensor &deriv = context.getIncomingDerivative(SINGLE_INOUT_IDX);
   Tensor &ret = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
-  Tensor &in = context.getOutput(SINGLE_INOUT_IDX);
+  Tensor &out = context.getOutput(SINGLE_INOUT_IDX);
 
-  ret = acti_func.run_prime_fn(in, ret, deriv);
+  ret = acti_func.run_prime_fn(out, ret, deriv);
```
zhoonit, line 50: `acti_fun.run_prim_fn(out, ret, deriv)` ?

nntrainer/layers/layer_node.h
zhoonit: Are we opening weight? or is it temporary?

nntrainer/layers/layer_node.cpp
kparichay: removed.

nntrainer/layers/layer_node.h
kparichay: This is just a wrapping interface, which takes already available information for weight and creates a wrapper for weight.
Further, this opens weights for internal use only. 
This is primarily added for unittest_models.

nntrainer/layers/activation_layer.cpp
kparichay: done

nntrainer/layers/layer_node.h
zhoonit: I see. I guess, in the long term it'd better refactor unittest_models later

zhoonit (approved):
LGTM :)

nntrainer/layers/layer_node.h
kparichay: If this function is not used anywhere else after layer_v2 refactor, the implementation here can move the unittest_models itself.

PR #1350, opened 2021-07-02
[test] Add common unittest for layers  @open sesame 07/05  19:36

kparichay:
Add common unittest for layers.
Also add validation for context and some more check in layer_node.
Further, some checks have been removed from LayerImpl which exists in
LayerNode.
The common tests have been split into two parts:
1. standalone: the environment must be manually created in these tests.
Further, these tests should not depend on any other
implementations/headers other than the ones on which LayerDevel also
depends on, like LayerNode, Manager, etc.
2. Dependent: the environment is created using the other
implementations.
Both these tests ensures that layer can work in correct environment and
that the environment creation in standalone version is also correct.


test/unittest/layers/layers_standalone_common_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+TEST_P(LayerSemantics, setPropertiesValidWithInvalid_n) {}
+
+TEST_P(LayerSemantics, setPropertiesValidInvalidOnly_n) {}
+
+TEST_P(LayerSemantics, finalizeOutputValidate_p) {
+  nntrainer::TensorDim in_dim({1, 1, 1, 1});
+  nntrainer::InitLayerContext init_context =
+    nntrainer::InitLayerContext({in_dim});
+  init_context.validate();
```
zhoonit, line 44: check if this is true?

test/unittest/layers/unitest_layer_node.cpp
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file        unittest_nntrainer_layers.cpp
```
zhoonit, line 5: Could you check the file name? s/unitest/unittest/

nntrainer/layers/loss/loss_layer.h
```diff
@@ -40,7 +40,13 @@ class LossLayer : public Layer {
   /**
    * @copydoc Layer::setProperty(const std::vector<std::string> &values)
    */
-  virtual void setProperty(const std::vector<std::string> &values) override {}
+  virtual void setProperty(const std::vector<std::string> &values) override {
+    if (!values.empty()) {
```
zhoonit, line 44: I think below macro can be applied for convenience.
```cpp
NNTR_THROW_IF(!values.empty(), std::invalid_argument) 
        << "[Layer] Unknown Layer Properties count = " << values.size()
```

test/unittest/layers/layers_common_tests.h
```diff
@@ -38,7 +38,7 @@ using LayerSemanticsParamType =
              std::string /** Type of Layer */,
              std::vector<std::string> /** Valid Properties */,
              std::vector<std::string> /** Invalid Properties */,
```
zhoonit, line 40: As per offline discussion yesterday, how about we delete invalid properties here?

test/unittest/layers/layers_standalone_common_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <layers_common_tests.h>
+
+#include <app_context.h>
+#include <layer_devel.h>
+
+constexpr unsigned SAMPLE_TRIES = 10;
+
+TEST_P(LayerSemantics, createFromAppContext_pn) {
+  auto &ac = nntrainer::AppContext::Global();
```
zhoonit, line 21: I guess App context is dependent test?

test/unittest/layers/layers_standalone_common_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+TEST_P(LayerSemantics, finalizeOutputValidate_p) {
+  nntrainer::TensorDim in_dim({1, 1, 1, 1});
+  nntrainer::InitLayerContext init_context =
+    nntrainer::InitLayerContext({in_dim});
+  init_context.validate();
+
+  // set necessary properties only
+  EXPECT_NO_THROW(layer->setProperty(valid_properties));
+
+  if (!must_fail) {
```
zhoonit, line 49: I understand that some rare layers are meant to fail, but in this case, I think it's better for the layer to have it's own `test` instead of using `test_p`, unless obeying the semantics.

test/unittest/layers/unittest_layers_loss.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <cross_entropy_softmax_loss_layer.h>
+#include <layers_common_tests.h>
+#include <mse_loss_layer.h>
+
+auto semantic_loss_cross_sigmoid = LayerSemanticsParamType(
+  nntrainer::createLayer<nntrainer::CrossEntropySigmoidLossLayer>,
+  nntrainer::CrossEntropySigmoidLossLayer::type, {}, {}, 0, false);
+
+INSTANTIATE_TEST_CASE_P(LossCrossSigmoid, LayerSemantics,
+                        ::testing::Values(semantic_loss_cross_sigmoid));
```
zhoonit, line 27:
```suggestion
INSTANTIATE_TEST_CASE_P(LossCrossSigmoid, LayerSemantics,
                        ::testing::Values(semantic_loss_cross_sigmoid, semantic_loss_cross_softmax, semantic_loss_mse, ... ));
```

zhoonit, line 27:
```suggestion
INSTANTIATE_TEST_CASE_P(LossCrossSigmoid, LayerSemantics,
                        ::testing::Values(semantic_loss_cross_sigmoid, semantic_loss_cross_softmax, semantic_loss_mse, ... ));
```

test/unittest/layers/layers_standalone_common_tests.cpp
kparichay: done.

test/unittest/layers/unitest_layer_node.cpp
kparichay: done.

nntrainer/layers/loss/loss_layer.h
kparichay: done.

test/unittest/layers/layers_standalone_common_tests.cpp
kparichay: moved.

kparichay: also moved layer context from LayerSemantics to dependent unittest.
Now LayerSemantics and standalone_unittest is not dependent on app_context.

test/unittest/layers/unittest_layers_loss.cpp
kparichay: done.

test/unittest/layers/layers_common_tests.h
kparichay: done.

test/unittest/layers/layers_standalone_common_tests.cpp
kparichay: I will leave this for now, as creating another set of tests for this layer will be cumbersome and reduce maintainability.
As this layer can be created and added by the user, the full testsuite has to be run.
I will find another way to handle this later.

test/unittest/layers/layers_standalone_common_tests.cpp
zhoonit: Cool

test/unittest/layers/layers_standalone_common_tests.cpp
zhoonit: Okay

zhoonit (approved):
LGTM after ci pass

jijoongmoon (approved):
LGTM

PR #1353, opened 2021-07-02
[Fix] Weight initializer stddev calculation

zhoonit:
- [Fix] Weight initializer stddev calculation

```
The weight initializer was fitted to only fully connected layer case,
this patch extends weight initializer to properly work with other
weights as well.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/tensor/weight.cpp
+++ b/nntrainer/tensor/weight.cpp
@@ -14,6 +14,8 @@
 #include <util_func.h>
 #include <weight.h>
 
+#include <nntrainer_error.h>
+
 namespace nntrainer {
 
 Weight::Weight(const TensorDim &dim, const WeightInitializer init,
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/weight.cpp
```diff
@@ -40,6 +42,29 @@ void Weight::runVariableInitializer() {
...
+  /// effective dimension then actual numbers here.
+  switch (dim.rank()) {
+  case 1:
+    fan_out = fan_in = dim.width();
+    break;
+  case 2: /// fully connected layers
+    fan_in = dim.height();
+    fan_out = dim.width();
+    break;
+  case 4: { /// convolution filters, @todo extend this to > 4
```
kparichay, line 57: if the dimension of some axis is 1, the rank will be 3 leading to fan_out/fan_in to be 1.

nntrainer/tensor/weight.cpp
```diff
@@ -40,6 +42,29 @@ void Weight::runVariableInitializer() {
...
 
+  unsigned int fan_in, fan_out;
+
+  /// @fixme: when unit is equal to one, this does not work we need to rely on
+  /// effective dimension then actual numbers here.
+  switch (dim.rank()) {
+  case 1:
+    fan_out = fan_in = dim.width();
+    break;
+  case 2: /// fully connected layers
```
kparichay, line 53: if rank == 2, but weight and height are 1 which is true for 1x1 conv filter, IMO this case doesn't handle it correctly.

zhoonit: Yes, there is `fixme` right above. Do you have a suggestion instead of relying on effective dimension?

kparichay: aah, I missed that fixme.

kparichay: How about below?
Case 1: B=1, C=1, and rank = 1 - current case 1
Case 2: B=1, C=1, and rank = 2 - current case 2
Default : current case 3

zhoonit: Updated :)

nntrainer/tensor/weight.cpp
zhoonit: Discussed together with below :)

kparichay (approved):
LGTM

nntrainer/tensor/weight.cpp
```diff
@@ -40,6 +42,24 @@ void Weight::runVariableInitializer() {
   Tensor &var_ref = getVariableRef();
   const TensorDim dim = var_ref.getDim();
 
+  unsigned int fan_in, fan_out;
+
+  /// @fixme: when unit is equal to one, this does not work, we need to rely on
+  /// effective dimension then actual numbers here. For now, some heuristics
+  /// added to infer what would be fan_in/fan_out
+  if (dim.batch() * dim.channel() * dim.height() == 1) {
+    fan_out = fan_in = dim.width();
```
jijoongmoon, line 51: do we need this?? if batch == channel == height ==1, then field_size = dim.width() . then fan_in and fan_out = dim.width().

jijoongmoon (approved):
LGTM with recommandations.

nntrainer/tensor/weight.cpp
zhoonit: Right :) I'll update.

PR #1354, opened 2021-07-05
[Padding] Add padding2d prop header @open sesame 07/05 16:40

zhoonit:
- [Padding] Add padding2d prop header

```
This patch adds padding2d property header.

Padding2D prop will be saved as a string and compute the integer when it
is needed

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -19,6 +19,9 @@
 #define __COMMON_PROPERTIES_H__
 
 namespace nntrainer {
+
+class TensorDim;
+
 namespace props {
 
 /**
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM!

nntrainer/layers/common_properties.h
```diff
@@ -189,6 +192,39 @@ class InputSpec : public nntrainer::Property<ConnectionSpec> {
...
 };
 
+/**
+ * @brief Padding2D property, this is used to calculate padding2D
+ * @details Padding2D is saved as a string. Upon calling Padding2D::compute,
+ * returns std::vector<unsigned int> which has computed padding2Ds, below
+ * formats are accepted valid
+ * 1. "same" (case insensitive literal string)
+ * 2. "valid" (case insensitive literal string)
+ * 3. "padding2D_all",
```
kparichay, line 202: Will these property options be open to the user? tf/pytorch does without these options, so I wonder how they do it and how compatible will be this property with their configuration.

zhoonit: option 4 is to conform with tizen 6.0 abi.
And numbered padding is also supported in pytorch.

FYI) maybe my comment is a bit misleading. option 3, 4, 5 is numbers

kparichay: > FYI) maybe my comment is a bit misleading. option 3, 4, 5 is numbers

I think this has resolved my confusion.

kparichay (approved):
LGTM

nntrainer/layers/common_properties.h
zhoonit: Updated the comment :)

PR #1356, opened 2021-07-05
[Padding] Add padding verification

zhoonit:
- [Padding] Add padding verification

```
This patch implements `Padding2D::isValid` and tests respectively.

Padding2D property is valid when
1. string is "valid" or "same"
2. comma seperated, non-integer value of size 1, 2, 4

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/common_properties.cpp
+++ b/nntrainer/layers/common_properties.cpp
@@ -14,6 +14,7 @@
 
 #include <nntrainer_error.h>
 #include <parse_util.h>
+#include <tensor_dim.h>
 
 #include <regex>
 #include <sstream>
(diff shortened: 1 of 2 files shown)


nntrainer/layers/common_properties.cpp
```diff
@@ -58,6 +59,45 @@ bool InputSpec::isValid(const ConnectionSpec &v) const {
...
 }
 
+/**
+ * @brief unsigned integer property, internally used to parse padding values
+ *
+ */
+class Padding_ : public nntrainer::Property<int> {
+public:
+  using prop_tag = int_prop_tag; /**< property type */
+};
```
jijoongmoon, line 69: is it integer?? If it is, then could you add more details in brief for each cases?

zhoonit: The `Padding2D` property is string. This class is used only internally inside `common_properties.cpp` to parse padding values.

jijoongmoon: so.. you mean, Padding2D is [ Padding_, Padding_ ] I guess. But If it is.. I guess. it's ok. but.. I prefer to use just < int, int >. It is better in the point of readability.

zhoonit: yes, it is std::vector<int, int>, this is like a helper function in a form of class :)

nntrainer/layers/common_properties.cpp
```diff
@@ -58,6 +59,45 @@ bool InputSpec::isValid(const ConnectionSpec &v) const {
...
+    return true;
+  }
+
+  /// case else: false
+  return false;
+}
+
+std::vector<unsigned int> Padding2D::compute(const TensorDim &input,
+                                             const TensorDim &kernel) {
+  return std::vector<unsigned int>();
```
kparichay, line 98: is this complete or implemeted in the next PR?

kparichay (approved):

nntrainer/layers/common_properties.cpp
kparichay: found in next PR :).

jijoongmoon (approved):
LGTM

PR #1357, opened 2021-07-05
[Padding] Add padding compute function

zhoonit:
- [Padding] Add padding compute function

```
This patch implements padding2d::compute with a test

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/common_properties.cpp
+++ b/nntrainer/layers/common_properties.cpp
@@ -95,7 +95,48 @@ bool Padding2D::isValid(const std::string &v) const {
 
 std::vector<unsigned int> Padding2D::compute(const TensorDim &input,
                                              const TensorDim &kernel) {
-  return std::vector<unsigned int>();
+  auto &padding_repr = get(); /// padding representation
+  if (istrequal(padding_repr, "valid")) {
+    return {0, 0, 0, 0};
+  }
+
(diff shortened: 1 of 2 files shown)


nntrainer/layers/common_properties.cpp
```diff
@@ -58,6 +59,86 @@ bool InputSpec::isValid(const ConnectionSpec &v) const {
...
+      return kernel_ - 1;
+    };
+
+    auto pad_horizontal = calculate_padding(input.width(), kernel.width());
+    auto pad_vertical = calculate_padding(input.height(), kernel.height());
+
+    auto pad_top = pad_vertical / 2;
+    auto pad_left = pad_horizontal / 2;
+
+    return {pad_top, pad_vertical - pad_top, pad_left,
```
kparichay, line 119: maybe these can be cached? it depends on if the compute function is to be called multiple times. If it is called once/twice, its fine.

kparichay (approved):

nntrainer/layers/common_properties.cpp
zhoonit: It is supposed to called at the finalize only, so I think it's fine :)

kparichay: then its alright.

jijoongmoon (approved):
LGTM

Issue #459, opened 2020-08-21
Support Dropout


kparichay:
Add dropout as a new layer.

zhoonit:
@kparichay  could you check if the issue has been resolved?

kparichay:
The implementation has been added by @jijoongmoon but the unittests are still to be done. They were delayed because layer unittests were broken till now.

PR #1359, opened 2021-07-05
[ Layer ] implementation of DropOut Layer

jijoongmoon:
In this commit,
  . implementation of DropOutLayer

Resolves:

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/dropout.cpp
+++ b/nntrainer/layers/dropout.cpp
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
+ *
+ * @file   dropout_layer.cpp
+ * @date   16 June 2020
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug	   No known bugs except for NYI items
+ * @brief  This is Dropout Layer Class for Neural Network
+ *
+ */
+
+#include <dropout.h>
+#include <layer_internal.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <parse_util.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+int DropOutLayer::initialize(Manager &manager) {
+  output_dim = input_dim;
+  // TODO : asking Tensor to manager
+  for (auto &t : input_dim) {
+    mask.push_back(std::make_shared<Tensor>(t, true));
+  }
+
+  return ML_ERROR_NONE;
+}
+
+void DropOutLayer::forwarding(bool training) {
+  auto &rate_ = std::get<props::DropOutSpec>(dropout_rate).get();
+  // Assume it is in-place calculation. It means input and output share mem
+  // buffer. So if the training is false, the output is the same with input. In
+  // orther words, there is nothing happen during inference.
+
+  if (training && rate_ > 0.0) {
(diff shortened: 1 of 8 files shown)


kparichay:
Resolves #459

zhoonit (approved):
LGTM :)

kparichay (approved):
LGTM
I am thinking that dropout layer can work inplace (for future), what do you think?

nntrainer/layers/dropout.cpp
```diff
@@ -0,0 +1,66 @@
...
+      Tensor &mask_ = *mask[i].get();
+
+      mask_ = input.dropout_mask(rate_);
+      input.multiply(mask_, hidden);
+    }
+  }
+}
+
+void DropOutLayer::calcDerivative() {
+  for (unsigned int i = 0; i < input_dim.size(); ++i) {
```
kparichay, line 49: add a check for `rate_ > 0.0` just like in forwarding?

nntrainer/layers/dropout.cpp
```diff
@@ -0,0 +1,66 @@
...
+    mask.push_back(std::make_shared<Tensor>(t, true));
+  }
+
+  return ML_ERROR_NONE;
+}
+
+void DropOutLayer::forwarding(bool training) {
+  auto &rate_ = std::get<props::DropOutSpec>(dropout_rate).get();
+
+  if (training && rate_ > 0.0) {
```
kparichay, line 36: if `rate_ == 0.0`, the output tensors go uninitialized?

nntrainer/layers/dropout.cpp
```diff
@@ -0,0 +1,66 @@
...
+  // TODO : asking Tensor to manager
+  for (auto &t : input_dim) {
+    mask.push_back(std::make_shared<Tensor>(t, true));
+  }
+
+  return ML_ERROR_NONE;
+}
+
+void DropOutLayer::forwarding(bool training) {
+  auto &rate_ = std::get<props::DropOutSpec>(dropout_rate).get();
```
kparichay, line 34: @zhoonit  can this throw if rate_ is not set by the user?

zhoonit: yes. if empty it throws

jijoongmoon:
> LGTM
> I am thinking that dropout layer can work inplace (for future), what do you think?

right. it works with in place.

nntrainer/layers/dropout.cpp
kparichay: Then perhaps this should be check at initialize phase for all the necessary properties like DropOutSpec here?

zhoonit: it'd better be but not necessarily I guess.

kparichay: i guess this can be verified with layerSemantics tests later, where not setting mandatory properties must fail initialize/finalize.

kparichay: I guess this discussion went a little off-topic for this PR, I will resolve this conversation.

lhs8928 (approved):
LGTM

nntrainer/layers/dropout.cpp
jijoongmoon: Current implementation assumes in-place operation. If it is, then it is same with input if rate_== 0.0 or inference mode as well. it means it is doing nothing.

nntrainer/layers/dropout.cpp
jijoongmoon: I will in later pr. Thanks.

PR #1360, opened 2021-07-06
[Wait for #1358] [Pooling] Apply padding property

zhoonit:
**[Pending commit: #1358]**

- [Pooling] Apply padding property

```
This patch applies padding property with tests

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

```

PR #1362, opened 2021-07-06
[pooling] Update to LayerV2 

kparichay:
**Commit 1:**
Update pooling layer to Layerv2 design.
Corresponding unittests for common and models are added and enabled
respectively.

setBatch() for layer node has been updated.
Further, some minor updates have been added to layer context to check if
they are ready to use.

**Commit 2:**
Update pooling layer to use helper tensors.
Instead of using vector memory, pooling layer will now use tensors
managed with the nntrainer.

As the memory is supposed to be integer, this patch currently interprets
float memory as integer memory. Note that no tensor operation is
performed on this memory as it would corrupt the data. Manual reading
and writing of data is done as was done with vector, but memory
management is moved out of the pooling layer with this patch.

This can be done formally when other data types are supported with
tensor class.

### DIFF
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -467,7 +467,7 @@ public:
    * @param idx index of the tensor (identifier)
    * @param batch Updated batch size
    */
-  void updateTensors(unsigned int idx, unsigned int batch) {
+  void updateTensor(unsigned int idx, unsigned int batch) {
     tensors[idx]->setBatchSize(batch);
   }
 
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -525,6 +525,22 @@ public:
    */
   const std::string &getName() const { return std::get<props::Name>(props); }
 
+  /**
+   * @brief   check if run context is set and is ready to use
+   *
+   * @return true if ready, else false
+   */
+  bool readyToUse() const {
+    /**
+     * assumption:
+     * 1. there must be atleast 1 input
+     * 2. the setter set everything at once
+     */
+    if (inputs.empty())
+      return false;
+    return !inputs[0]->getVariable().uninitialized();
+  }
+
 private:
   std::tuple<props::Name> props; /**< props of the layer */
   float loss;                    /**< loss of the layer */
--- a/nntrainer/layers/pooling2d_layer.cpp
+++ b/nntrainer/layers/pooling2d_layer.cpp
@@ -90,8 +90,14 @@ void Pooling2DLayer::finalize(InitLayerContext &context) {
    * = 12 / 4 = 3
    * // clang-format on
    */
-  pool_helper_idx = context.requestTensor(out_dim, "Pooling2d::helper_idx",
-                                          false, ITERATION_LIFESPAN);
+  if (pooling_type == PoolingType::global_max) {
+    pool_helper_idx = context.requestTensor(in_dim, "Pooling2d::helper_idx",
+                                            false, ITERATION_LIFESPAN);
+    pool_helper_size.resize(in_dim.batch() * in_dim.channel());
+  } else {
+    pool_helper_idx = context.requestTensor(out_dim, "Pooling2d::helper_idx",
+                                            false, ITERATION_LIFESPAN);
+  }
 }
 
 void Pooling2DLayer::forwarding(RunLayerContext &context, bool training) {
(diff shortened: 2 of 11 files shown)


test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -1272,7 +1272,7 @@ INSTANTIATE_TEST_CASE_P(
     mkModelTc(pooling_avg_same_padding_multi_stride, "3:1:1:10", 10),
     mkModelTc(pooling_avg_valid_padding, "3:1:1:10", 10),
     mkModelTc(pooling_global_avg, "3:1:1:10", 10),
-    mkModelTc(pooling_global_max, "3:1:1:10", 10),
+    // mkModelTc(pooling_global_max, "3:1:1:10", 10),
```
zhoonit, line 1275: Wonder why this is commented

kparichay: my bad for the incomplete description. GLOBAL_MAX is still WIP.

kparichay:
> unintended comment?

missing context?

zhoonit (commented):
> > unintended comment?
> 
> missing context?

I think something changed to github UI, please refer to https://github.com/nnstreamer/nntrainer/commit/2437c056be6730d9f0c729b501c17401e98eed81

zhoonit:
I have a little concern that this might clash with #1357

kparichay:
> I have a little concern that this might clash with #1357

It might, I guess we will sit together and merge this commit manually :).

nntrainer/layers/layer_context.h
```diff
@@ -525,6 +525,22 @@ class RunLayerContext {
...
+   */
+  bool readyToUse() const {
+    /**
+     * assumption:
+     * 1. there must be atleast 1 input
+     * 2. the setter set everything at once
+     */
+    if (inputs.empty())
+      return false;
+    return !inputs[0]->getVariable().uninitialized();
```
zhoonit, line 541: Is checking for the first input enough?

zhoonit: If assumption holds, it looks safe

nntrainer/layers/pooling2d_layer.h
```diff
@@ -37,17 +40,26 @@ class Pooling2DLayer : public LayerV1 {
     unknown = 4,
   };
 
+  /**
+   * @brief PaddingType Class
+   * @todo support keras type of padding
+   */
+  enum class PaddingType {
```
zhoonit, line 47: Could you delete this? I won't in #1360

kparichay: It is used in parse_util.cpp. 
I recommend you to delete it in #1360 as deleting also doesn't match the patch scope.

nntrainer/layers/layer_context.h
kparichay: yes, if the assumption stays valid. Even if we check all, and the assumption does not hold true, some required inputs might not be set, and checking all inputs wont be enough either.

nntrainer/layers/pooling2d_layer.h
zhoonit: I see, I'll do it after #1360, #1359 is merged.

test/unittest/unittest_nntrainer_models.cpp
kparichay: done.

nntrainer/layers/pooling2d_layer.h
kparichay: sure.

jijoongmoon (approved):
LGTM

nntrainer/layers/pooling2d_layer.cpp
```diff
@@ -181,21 +187,20 @@ void Pooling2DLayer::calcDerivative(RunLayerContext &context) {
...
-    unsigned int outer_loop = batch * channel;
-    for (unsigned int _ = 0; _ < outer_loop; _++) {
-      for (unsigned int i = 0; i < out_map_size; ++i) {
-        float der = *deriv_data / pool_helper.length();
-
-        for (unsigned int idx = 0; idx < pool_helper.length(); idx++) {
+    for (unsigned int b = 0; b < batch; b++) {
+      for (unsigned int c = 0; c < channel; c++) {
+        const int *iter =
+          pool_helper.getData<int>() + pool_helper.getIndex(b, c, 0, 0);
```
zhoonit, line 194: Note: This will be working for non-contiguous tensor but there is additional overhead of calculating the output index. It was the reason I tried to not put batch argument to this function.

I guess it is fine until we do some benchmarks.

zhoonit (approved):
LGTM

nntrainer/layers/pooling2d_layer.cpp
kparichay: When using std::vector, the batch and channel dimensions were merged. The same can be done with tensor, where the batch dimension would become batch * channel size. is that what you meant?

zhoonit: Yeah, sth like that.

kparichay: I will do this in a separte small PR.

PR #1363, opened 2021-07-06
[loss/unittest] Proper support for loss as layer

kparichay:
Support for loss as a layer with the API has been done.
However, the verification of this feature was missing.
This patch ensures that the added loss layer is verified and its
required conditions checked before finalizing the graph.

Add loss layer unittests for various loss configurations to be
supported. These tests do not ensure correct graph formation.
This will be done soon with models unittest.

Add golden tests for loss layer added as a layer.
    This patch modified the existing models test architecture to be
    described where layer is added externally.
    
More will be added ones more models tests are enabled.

cc. @zhoonit

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -53,6 +53,14 @@ int NetworkGraph::compile(const std::string &loss_type) {
   status = addLossLayer(loss_type);
   NN_RETURN_STATUS();
 
+  try {
+    finalizeLossLayer();
+  } catch (const std::exception &e) {
+    ml_loge("%s", e.what());
+    status = ML_ERROR_INVALID_PARAMETER;
+    NN_RETURN_STATUS();
+  }
+
   countNonTrainableLayersAtBegin();
 
   status = checkCompiledGraph();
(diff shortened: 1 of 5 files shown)


jijoongmoon (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -221,72 +229,76 @@ int NetworkGraph::realizeMultiOutputType(
...
-      status = ML_ERROR_INVALID_PARAMETER;
-    return status;
-  }
-
-  if (loss_type.empty()) {
-    return status;
+void NetworkGraph::finalizeLossLayer() {
+  auto loss_layer_node = getSortedLayerNode(graph.size() - 1);
+  std::string loss_type = loss_layer_node->getType();
+  if (graph.size() == 1) {
```
zhoonit, line 235: Could you explain what this line is checking?

is it checking that graph is only consists of a loss?

zhoonit (approved):
Thanks :)

nntrainer/graph/network_graph.cpp
kparichay: For the naive cross-entropy layer, the previous layer must be softmax or sigmoid. So, if graph size is 1, then its not possible. However, other losses/layers are possible.
This also provides added security that the check at L245 does not fail.

PR #1368, opened 2021-07-07
[layer] Enable addition and multiout layer for V2 design

kparichay:
Update addition and multioutput layer for V2 design.
Add corresponding unittests.

Enable graph unittests which creates resnet like model for testing.
Enable correspoding resnet like add in models unittest.

### DIFF
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -57,14 +57,16 @@ public:
    * @brief Construct a new Init Layer Context object
    *
    */
-  InitLayerContext() = default;
+  InitLayerContext() : InitLayerContext({}, 1) {}
 
   /**
    * @brief Construct a new Init Layer Context object
    *
    * @param dim Input dimensions for the layer
    */
-  InitLayerContext(const std::vector<TensorDim> &dim) : input_dim(dim) {}
+  InitLayerContext(const std::vector<TensorDim> &dim, unsigned int num_out) :
+    input_dim(dim),
+    num_outputs(num_out) {}
 
   /**
    * @brief Get the number of inputs for the layer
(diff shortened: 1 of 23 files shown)


nntrainer/layers/bn_layer.cpp
```diff
@@ -61,18 +61,21 @@ void BatchNormalizationLayer::finalize(InitLayerContext &context) {
       axes_to_reduce.push_back(i);
   }
 
-  weight_idx[BNParams::mu] = context.requestWeight(
-    dim, initializers[BNParams::mu], WeightRegularizer::NONE, 1.0f,
-    "BN::moving_mean", false);
-  weight_idx[BNParams::var] = context.requestWeight(
+  wt_idx[BNParams::mu] = context.requestWeight(dim, initializers[BNParams::mu],
```
kparichay, line 64: test comment

nntrainer/layers/layer_context.h
```diff
@@ -57,14 +57,16 @@ class InitLayerContext {
...
 
   /**
    * @brief Construct a new Init Layer Context object
    *
    * @param dim Input dimensions for the layer
    */
-  InitLayerContext(const std::vector<TensorDim> &dim) : input_dim(dim) {}
+  InitLayerContext(const std::vector<TensorDim> &dim, unsigned int num_out) :
+    input_dim(dim),
+    num_outputs(num_out) {}
```
jijoongmoon, line 69: I guess these are layer node properties. Does the initcontext need to have those? is it because multi out and multi input?

jijoongmoon:
I think it is great for now. But we need to start discussing about #1373 .

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_context.h
kparichay: Yes, it is layer node property.
The number of outputs are fixed by the connections in the graph. Some nodes need that information such as split node.
Further for other nodes, setting the output dimensions should not exceed num_outputs already set inside InitLayerContext.

PR #1369, opened 2021-07-07
[tests] Add more layer common unittests 

kparichay:
Add more layer common unittests without involving runContext for the
layers.


jijoongmoon (approved):
LGTM

PR #1372, opened 2021-07-08
[dataset] Remove label data

zhoonit:
- [dataset] Remove label data

```
Before this patch, label.dat was only used to get number of classes,
unfortunately, it was clashing with how databuffer_generator is getting
the number of classes and creating some inconsistency. This patch fixes
the issue by deprecating label data.

* Checked TCT and there is no critical issue that makes tests fail with
this PR.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/databuffer_file.cpp
+++ b/nntrainer/dataset/databuffer_file.cpp
@@ -293,24 +293,6 @@ int DataBufferFromDataFile::setDataFile(DataType type, std::string path) {
     }
     test_name = path;
   } break;
-  case DATA_LABEL: {
-    std::string data;
-    validation[type] = true;
-    if (!data_file.good()) {
-      ml_loge("Error: Cannot open label file");
-      SET_VALIDATION(false);
-      return ML_ERROR_INVALID_PARAMETER;
-    }
-    while (data_file >> data) {
-      labels.push_back(data);
-    }
-    if (class_num != 0 && class_num != labels.size()) {
-      ml_loge("Error: number of label should be same with number class number");
-      SET_VALIDATION(false);
-      return ML_ERROR_INVALID_PARAMETER;
-    }
-    class_num = labels.size();
-  } break;
   case DATA_UNKNOWN:
   default:
     ml_loge("Error: Not Supported Data Type");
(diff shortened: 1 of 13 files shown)


nntrainer/models/model_loader.cpp
```diff
@@ -224,8 +224,10 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
   NN_RETURN_STATUS();
   status = parse_and_set("DataSet:TestData", DATA_TEST, false);
   NN_RETURN_STATUS();
-  status = parse_and_set("Dataset:LabelData", DATA_LABEL, true);
-  NN_RETURN_STATUS();
+  const char *path = iniparser_getstring(ini, "Dataset:LabelData", NULL);
+  if (path != NULL) {
+    ml_logi("setting labeData is deprecated!, it is essentially noop now!");
```
kparichay, line 229: labeData -> labelData

nntrainer/dataset/databuffer_file.cpp
```diff
@@ -293,24 +293,6 @@ int DataBufferFromDataFile::setDataFile(DataType type, std::string path) {
...
-    }
-    while (data_file >> data) {
-      labels.push_back(data);
-    }
-    if (class_num != 0 && class_num != labels.size()) {
-      ml_loge("Error: number of label should be same with number class number");
-      SET_VALIDATION(false);
-      return ML_ERROR_INVALID_PARAMETER;
-    }
-    class_num = labels.size();
```
kparichay, line 312: so in case of file based databuffer, how is the number of class information provided?

zhoonit: It is unified to get from the width of the output shape. Later this is being change to directly take label shape so that, number of classes is no longer needed(around #1395)

nntrainer/models/model_loader.cpp
zhoonit: Nice catch, done :)

nntrainer/dataset/databuffer_file.cpp
kparichay: how about instead of taking the width of the output shape, we consume the exact size of the output shape, especially for cases when output is not for classification purposes?

zhoonit: Yes, similar thing is planned (we take label shape, if not given, taking output shape). For the knn we are planning to do, it takes single integer as a label and output is a vector.

jijoongmoon: So, we don't care about the label data anymore. It makes sense.

jijoongmoon (approved):
LGTM

nntrainer/dataset/databuffer_file.cpp
kparichay: cool.

kparichay (approved):
LGTM

PR #1376, opened 2021-07-08
[dataset] Clean up dataset enums

zhoonit:
~**[Pending commit: #1372]**~

- [dataset] Clean up dataset enums

```
`nntrainer::DataType` and `nntrainer::BufferType` is duplicated from
ccapi which was adding complication. This patch simply alternate those
types ccapi `DatasetType` / `DatasetDataUsageType`

**Self evaluation:**

```

### DIFF
--- a/Applications/LogisticRegression/jni/main.cpp
+++ b/Applications/LogisticRegression/jni/main.cpp
@@ -172,7 +172,8 @@ int main(int argc, char *argv[]) {
 
   std::shared_ptr<nntrainer::DataBufferFromCallback> DB =
     std::make_shared<nntrainer::DataBufferFromCallback>();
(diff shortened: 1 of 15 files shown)


kparichay:
This will be updated based on the offline discussion we had today?

zhoonit:
> This will be updated based on the offline discussion we had today?

Yes but not in this PR, so many enum is being used in like 16 PR so I'll make a new one to save time

kparichay:
> > This will be updated based on the offline discussion we had today?
> 
> Yes but not in this PR, so many enum is being used in like 16 PR so I'll make a new one to save time

alright.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM.

Issue #293, opened 2020-07-10
Restrict setParamSize to be called once only.


zhoonit:
This need minor change on unittest_nntrainer_layers. So it is postponed.

https://github.com/nnstreamer/nntrainer/pull/274#discussion_r452536997

PR #1377, opened 2021-07-08
[network_graph] multi output

### LINKED ISSUE #1367: Multi outputs
- The list of output layer will be maintained by graph_core.
- Switch order of topological sort and add loss layer function

lhs8928:
Commit 1:
    [graph_core] Implement input_list, output_list for multi input, output
    
    - Make input_list, output_list and its getter to support multi input, output
    
Commit 2:
    [network_graph] revise addlosslayer for multi output
    
     - reorder addlosslayer to run before topologicalsort
     - Replace node_list rather than delete when the loss is entropy
    
Commit 3:
    [unittest] multi output unittest
    
     - Modify unittest to test multi output case
     - Make setLabel function in neuralnet
       to set labels at all loss layers
    
    Resolve #1367
    
    **Self evaluation:**
    

nntrainer/graph/graph_core.h
```diff
@@ -205,6 +205,26 @@ class GraphCore {
    */
   void addLossToSorted();
 
+  void addInputNode(std::shared_ptr<GraphNode> &node) {
```
kparichay: I wonder if the graph_core needs this function? can graph_core extract this information by itself? I recommend to modify the modify graph core interface for graph core to interpret itself given the nodes, which are input nodes and which are output nodes.

kparichay: you might have to introduce a function (getOutputConnections() like getInputConnections()) in GraphNode.

kparichay:
Overall good job

nntrainer/graph/network_graph.cpp
```diff
@@ -459,6 +454,7 @@ int NetworkGraph::realizeGraph() {
     }
   }
   /// @todo add check that input_layers <-> output_layers does match.
+  /// @todo check whether graph has a cycle or bipartite
```
zhoonit, line 451: Could you explain why we need to check it is bipartite?

nntrainer/graph/graph_core.h
```diff
@@ -205,6 +205,26 @@ class GraphCore {
...
+    input_list.push_back(node);
+  }
+
+  const std::shared_ptr<GraphNode> &getInputNode(unsigned int idx) const {
```
zhoonit, line 205: Please add doxygen for the functions added.

nntrainer/models/neuralnet.cpp
```diff
@@ -478,8 +482,8 @@ bool NeuralNetwork::validateInput(sharedConstTensors X) {
   return true;
 }
 
-sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
-                                            bool free_mem) {
+std::vector<sharedConstTensors> NeuralNetwork::inference(sharedConstTensors X,
```
zhoonit, line 487: Currently, this is returning vector of vector.
The reasoning behind this is that each node can have multiple outputs.
eg)
out_layer1 -> out_tensor_a, out_tensor_b
out_layer2 -> out_tensor_c, out_tensor_d
out_layer3 -> out_tensor_e,

so returning [[out_tensor_a, out_tensor_b], [out_tensor_c, out_tensor_d], [out_tensor_e]]

We made a bit of offline discussion about this signature and rollback.

Reasoning,
1. NNStreamer or other frameworks expects/makes output of 1d list. (correct me if I am wrong)
2. If we start to allow nested output, not only 2d but nd list is possible. eg) multiple subgraph of above example
3. When receiving inputs, we are having 1D list (float **) not (float ***) so it is also predictable that output is also a type of 1D list.
3. Even though returning 1D result, we are not loosing much representational power because user will know about the output layout.

-------------- Off-the-topic issue
As our sort order is not determined, it will be requested to user to write the out layer names in order, to determine the output layout.

nntrainer/graph/network_graph.cpp
lhs8928: Not related to this pr, but when the multi input and multi output is supported the graph can be bipartite so we need to check it I guess.

nntrainer/graph/graph_core.h
lhs8928: So you means instread of addInputNode and addOutputNode which is called in realizeGraph by network_graph make a function in graph_core something like findinputOutputNode and call that function after running the realizeGraph. And getOutputConnection will be needed when making that function am I right?

kparichay: yes, you are right.

nntrainer/graph/network_graph.cpp
kparichay: bipartite graph can be cyclic (bipartite only ensures no cycles of odd length), so you might want to expand on this comment here.

nntrainer/graph/graph_core.h
kparichay: I wonder how doxygen CI passed for this PR.

nntrainer/models/neuralnet.cpp
kparichay: > As our sort order is not determined, it will be requested to user to write the out layer names in order, to determine the output layout.

If these names are to be taken, it better to take them during the model compile. In this way, the shape and size of every inference remains fixed.

kparichay: > We made a bit of offline discussion about this signature and rollback.

Keeping the current signature must be clarified - does it mean multiple output nodes are allowed when the output of each node has to be just 1 tensor or a single node with multiple outputs? IMO, the former case is better, as this can be expanded by making the tensor class more flexible (supporting more than 4 dimensions in the future).

> NNStreamer or other frameworks expects/makes output of 1d list. (correct me if I am wrong)

You are right. However IMO nntrainer's tensor filter element can define semantics for the conversion from nntrainer's inference to be returned to nnstreamer.

> If we start to allow nested output, not only 2d but nd list is possible. eg) multiple subgraph of above example

Can this be nd list (as long as a node can only handle list of tensors)?

zhoonit: >  IMO, the former case is better 

I agree :)

> Can this be nd list (as long as a node can only handle list of tensors)?

Yes,

for example.

subgraph 1. [[a], [b,c], [d,e]]
subgarph 2. [[f], [g, h]

output of sugraph 1 + subgraph 2 is 3D list.

This implies that we can't fully represent the list of list with nd graph where n is a fixed number. So the best we can do would be have semantics that is is always flattened for the model output.

nntrainer/graph/network_graph.cpp
lhs8928: Oh no sorry. I mean 2 separated graph not the bipartite. I will amend the comment

nntrainer/graph/graph_core.h
lhs8928: I've done what you said but check replaceOutputNode function that this also needs to removed. Actually we can also delete this function by realize input output node after addLossLayer. But in this case in the we didn't have a information which layer is a output node so we need to check all the layer to find output layer in addLossLayer

nntrainer/graph/graph_core.h
lhs8928: Added doxygen

nntrainer/models/neuralnet.cpp
lhs8928: reverted to return sharedtensors

nntrainer/graph/graph_core.cpp
```diff
@@ -167,20 +167,22 @@ void GraphCore::ensureName(GraphNode &node, const std::string &prefix,
...
+  unsigned int idx = from->getIndex();
+  to->setIndex(idx);
+  node_list[idx] = to;
+}
 
-  for (; after != node_list.end(); ++after) {
-    (*after)->setIndex(after_index++);
+void GraphCore::realizeInputOutputNode() {
+  for (auto iter = cbegin(); iter != cend(); ++iter) {
+    if (iter->getInputConnections()[0] == "__data__") {
```
zhoonit, line 179: Please note that this will have to change after merging v2. it will compile so it will be hard to find the problem.

nntrainer/graph/graph_core.h
```diff
@@ -195,15 +195,52 @@ class GraphCore {
   void ensureName(GraphNode &node, const std::string &prefix = "",
                   const std::string &postfix = "", bool force_rename = false);
 
+  void replaceNode(std::shared_ptr<GraphNode> from,
```
zhoonit, line 198: This has missing doxygen

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -1286,6 +1293,28 @@ INI multi_gru_return_sequence_with_batch(
   }
 );
 
+INI multi_output_resnet_like(
```
zhoonit, line 1296: is this considered resnet_like?

zhoonit (approved):
Overall good job! lgtm after ci pass + doxygen added

nntrainer/graph/graph_core.cpp
kparichay: yes, this has been added to the list of manual merging along with padding :)

kparichay, line 179: can you also add a check here ensuring that `iter->getInputConnections().size() == 1`?

nntrainer/graph/graph_core.cpp
```diff
@@ -167,20 +167,22 @@ void GraphCore::ensureName(GraphNode &node, const std::string &prefix,
...
+      input_list.push_back(*iter);
+    }
+    if (iter->getOutputConnections()[0] == "__exit__") {
```
kparichay, line 182: same here.

nntrainer/graph/graph_core.h
```diff
@@ -196,14 +196,56 @@ class GraphCore {
...
-  void addLossToSorted();
+  const std::shared_ptr<GraphNode> &getInputNode(unsigned int idx) const {
+    return input_list[idx];
+  }
+
+  /**
+   * @brief   getter of number of input nodes
+   * @return  number of input nodes
+   */
+  unsigned int getInputNodeSize() const { return input_list.size(); }
```
kparichay, line 219: getInputNodeSize() -> getNumInputNodes()?

nntrainer/graph/graph_core.h
```diff
@@ -196,14 +196,56 @@ class GraphCore {
...
+   */
+  const std::shared_ptr<GraphNode> &getOutputNode(unsigned int idx) const {
+    return output_list[idx];
+  }
+
+  /**
+   * @brief   getter of number of output nodes
+   * @return  number of output nodes
+   */
+  unsigned int getOutputNodeSize() const { return output_list.size(); }
```
kparichay, line 234: getNumOutputsNodes()

test/unittest/unittest_nntrainer_models.cpp
kparichay: maybe rename to multiple output model?

kparichay (approved):
LGTM other than some minor comments
Amazing Job

nntrainer/graph/graph_core.h
kparichay: Yes, that will be good. `addLossLayer()` can once iterate over all the nodes to find output layers - it's fine IMO.
It's an optimization that can be done in the next PR as this PR is already big enough.

nntrainer/graph/network_graph.cpp
kparichay: yes, thanks for the clarification.

nntrainer/models/neuralnet.cpp
kparichay: great.

nntrainer/graph/graph_core.h
lhs8928: Done

test/unittest/unittest_nntrainer_models.cpp
lhs8928: Rename to multiple_output_model

nntrainer/graph/graph_core.cpp
lhs8928: Done

nntrainer/graph/graph_core.cpp
lhs8928: Done

nntrainer/graph/graph_core.h
lhs8928: Done

nntrainer/graph/graph_core.h
lhs8928: Done

nntrainer/graph/network_graph.cpp
```diff
@@ -212,98 +214,90 @@ int NetworkGraph::realizeMultiOutputType(
...
-    return status;
-  }
+  for (unsigned int i = 0; i < graph.getNumOutputNodes(); ++i) {
+    auto output_layer_node = LNODE(graph.getOutputNode(i));
 
-  if (loss_type == LossType::LOSS_NONE) {
-    return status;
-  }
+    if (output_layer_node->getObject()->requireLabel()) {
+      continue;
```
kparichay, line 224: so, if some loss layers are added as layers in the model, then this continue will lead to `loss` being added after all the output nodes which do not contain loss nodes. Is this intended?

kparichay, line 224: also please add a todo here to realize the cross_entropy loss to softmax/sigmoid cross entropy loss.
This is already implemented in layer_v2, and will be added here upon merge.

lhs8928: I think you want to say that in that case it makes an error and that's what we've discussed. In this pr I didn't handle such cases. Maybe in later pr I will take care of.

lhs8928: I think from line 235 to 249 the realization of cross_entropy loss to softmax/sigmoid cross entropy loss exist. Am I misunderstand what you say?

kparichay: cool.

kparichay: If the output_layer_node is cross_entropy layer type, this loop will continue here. However, it also needs realization (not add new loss layer but udpate current loss layer type).

kparichay (approved):
LGTM

nntrainer/graph/network_graph.cpp
lhs8928: Oh now I understand. Thanks

nntrainer/graph/graph_core.h
```diff
@@ -235,6 +233,20 @@ class GraphCore {
...
   unsigned int getNumOutputNodes() const { return output_list.size(); }
 
+  /**
+   * @brief       replace output node
+   * @param idx   output node index to be replaced
+   * @param node  graph node shared pointer to replace
+   */
+  void replaceOutputNode(unsigned int idx, std::shared_ptr<GraphNode> node) {
+    output_list[idx] = node;
+  }
```
jijoongmoon, line 243: I wonder if the size of ouptut_list is the same with graph node size according to line #293

jijoongmoon (approved):
LGTM

PR #1379, opened 2021-07-09
[dataset] split train / val / test databuffer @open sesame 07/15 15:00

zhoonit:
- [dataset] Clean up dataset enums

```
`nntrainer::DataType` and `nntrainer::BufferType` is duplicated from
ccapi which was adding complication. This patch simply alternate those
types ccapi `DatasetType` / `DatasetDataUsageType`

**Self evaluation:**

```

- [dataset] split train / val / test databuffer

```
This patch splits train / val / test dataset.

It is also possible to set dataset separately from the model.

**Major Changes**
1. `auto dataset = createDataset(train_cb, val_cb, test_cb)` -> `auto
dataset_train = createDataset(train_cb)`
1. `NN.setDataset(dataset);` -> `NN.setDataset(DATA_TRAIN,
dataset_train)`

**Self evaluation:**

```

- [dataset] Clean up dataset enums

```
`nntrainer::DataType` and `nntrainer::BufferType` is duplicated from
ccapi which was adding complication. This patch simply alternate those
types ccapi `DatasetType` / `DatasetDataUsageType`

**Self evaluation:**

```

### DIFF
--- a/Applications/TransferLearning/CIFAR_Classification/jni/main_func.cpp
+++ b/Applications/TransferLearning/CIFAR_Classification/jni/main_func.cpp
@@ -273,10 +273,12 @@ int main(int argc, char *argv[]) {
   /**
    * @brief     Data buffer Create & Initialization
    */
-  std::shared_ptr<ml::train::Dataset> dataset;
+  std::shared_ptr<ml::train::Dataset> dataset_train, dataset_val;
   try {
-    dataset = createDataset(ml::train::DatasetType::GENERATOR, getBatch_train,
-                            getBatch_val);
+    dataset_train =
+      createDataset(ml::train::DatasetType::GENERATOR, getBatch_train);
+    dataset_val =
+      createDataset(ml::train::DatasetType::GENERATOR, getBatch_val);
   } catch (...) {
     std::cerr << "Error creating dataset" << std::endl;
     return 1;
@@ -306,7 +308,8 @@ int main(int argc, char *argv[]) {
     std::cerr << "Error during readModel, reason: " << e.what() << std::endl;
     return 1;
   }
-  model->setDataset(dataset);
+  model->setDataset(ml::train::DatasetDataUsageType::DATA_TRAIN, dataset_train);
+  model->setDataset(ml::train::DatasetDataUsageType::DATA_VAL, dataset_val);
 
   /**
    * @brief     Neural Network Train & validation
--- a/api/capi/include/nntrainer_internal.h
+++ b/api/capi/include/nntrainer_internal.h
@@ -25,6 +25,7 @@
 #ifndef __NNTRAINER_INTERNAL_H__
 #define __NNTRAINER_INTERNAL_H__
 
+#include <array>
 #include <mutex>
 #include <string>
 #include <unordered_map>
@@ -101,7 +102,7 @@ typedef struct {
  */
 typedef struct {
   uint magic;
-  std::shared_ptr<ml::train::Dataset> dataset;
+  std::array<std::shared_ptr<ml::train::Dataset>, 3> dataset;
   bool in_use;
   std::mutex m;
 } ml_train_dataset;
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -22,6 +22,7 @@
  * @bug No known bugs except for NYI items
  */
 
+#include <array>
 #include <cstdarg>
 #include <cstring>
 #include <sstream>
@@ -143,7 +144,16 @@ static int ml_train_dataset_create(ml_train_dataset_h *dataset,
   nndataset->in_use = false;
 
   returnable f = [&]() {
-    nndataset->dataset = ml::train::createDataset(type, train, valid, test);
+    nndataset->dataset[ML_TRAIN_DATASET_DATA_USAGE_TRAIN] =
+      ml::train::createDataset(type, train);
+    if (valid != nullptr) {
+      nndataset->dataset[ML_TRAIN_DATASET_DATA_USAGE_VALID] =
+        ml::train::createDataset(type, valid);
+    }
+    if (test != nullptr) {
+      nndataset->dataset[ML_TRAIN_DATASET_DATA_USAGE_TEST] =
+        ml::train::createDataset(type, test);
+    }
     return ML_ERROR_NONE;
   };
 
@@ -492,12 +502,35 @@ int ml_train_model_set_dataset(ml_train_model_h model,
   }
 
   std::shared_ptr<ml::train::Model> m;
-  std::shared_ptr<ml::train::Dataset> d;
 
   m = nnmodel->model;
-  d = nndataset->dataset;
 
-  returnable f = [&]() { return m->setDataset(d); };
+  returnable f = [&]() {
+    auto &[train_set, valid_set, test_set] = nndataset->dataset;
+    int status = ML_ERROR_NONE;
+    status =
+      m->setDataset(ml::train::DatasetDataUsageType::DATA_TRAIN, train_set);
+    if (status != ML_ERROR_NONE) {
+      return status;
+    }
+
+    if (valid_set != nullptr) {
+      status =
+        m->setDataset(ml::train::DatasetDataUsageType::DATA_VAL, valid_set);
+      if (status != ML_ERROR_NONE) {
+        return status;
+      }
+    }
+
+    if (test_set != nullptr) {
+      status =
+        m->setDataset(ml::train::DatasetDataUsageType::DATA_TEST, test_set);
+      if (status != ML_ERROR_NONE) {
+        return status;
+      }
+    }
+    return status;
+  };
 
   status = nntrainer_exception_boundary(f);
   if (status == ML_ERROR_NONE) {
@@ -735,7 +768,7 @@ int ml_train_dataset_set_property(ml_train_dataset_h dataset, ...) {
   int status = ML_ERROR_NONE;
   ml_train_dataset *nndataset;
   void *data;
-  std::shared_ptr<ml::train::Dataset> d;
+  std::array<std::shared_ptr<ml::train::Dataset>, 3> db;
 
   check_feature_state();
 
@@ -755,10 +788,21 @@ int ml_train_dataset_set_property(ml_train_dataset_h dataset, ...) {
     ML_TRAIN_GET_VALID_DATASET_LOCKED(nndataset, dataset);
     ML_TRAIN_ADOPT_LOCK(nndataset, dataset_lock);
 
-    d = nndataset->dataset;
+    db = nndataset->dataset;
   }
 
-  returnable f = [&]() { return d->setProperty(arg_list); };
+  returnable f = [&]() {
+    int status = ML_ERROR_NONE;
+    for (auto &d : db) {
+      if (d != nullptr) {
+        status = d->setProperty(arg_list);
+        if (status != ML_ERROR_NONE) {
+          return status;
+        }
+      }
+    }
+    return status;
+  };
   status = nntrainer_exception_boundary(f);
 
   return status;
--- a/nntrainer/dataset/databuffer_func.cpp
+++ b/nntrainer/dataset/databuffer_func.cpp
@@ -86,7 +86,7 @@ int DataBufferFromCallback::init() {
 }
 
 int DataBufferFromCallback::setGeneratorFunc(DatasetDataUsageType type,
-                                             datagen_cb func) {
+                                             datagen_cb func, void *user_data) {
 
   int status = ML_ERROR_NONE;
   switch (type) {
@@ -94,6 +94,7 @@ int DataBufferFromCallback::setGeneratorFunc(DatasetDataUsageType type,
     if (!func)
       return ML_ERROR_INVALID_PARAMETER;
     callback_train = func;
+    this->user_data = user_data;
     if (func)
       validation[0] = true;
     break;
--- a/nntrainer/models/model_loader.cpp
+++ b/nntrainer/models/model_loader.cpp
@@ -193,9 +193,7 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
   int status = ML_ERROR_NONE;
 
   if (iniparser_find_entry(ini, "Dataset") == 0) {
-    model.data_buffer = nntrainer::createDataBuffer(DatasetType::GENERATOR);
-    status = model.data_buffer->setBatchSize(model.batch_size);
-    return status;
+    return ML_ERROR_NONE;
   }
 
   if (iniparser_find_entry(ini, "DataSet:Tflite")) {
@@ -203,10 +201,14 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
     return ML_ERROR_INVALID_PARAMETER;
   }
 
-  model.data_buffer = nntrainer::createDataBuffer(DatasetType::FILE);
-  std::shared_ptr<DataBufferFromDataFile> dbuffer =
-    std::static_pointer_cast<DataBufferFromDataFile>(model.data_buffer);
+  model.data_buffers[static_cast<int>(DatasetDataUsageType::DATA_TRAIN)] =
+    nntrainer::createDataBuffer(DatasetType::FILE);
+  model.data_buffers[static_cast<int>(DatasetDataUsageType::DATA_VAL)] =
+    nntrainer::createDataBuffer(DatasetType::FILE);
+  model.data_buffers[static_cast<int>(DatasetDataUsageType::DATA_TEST)] =
+    nntrainer::createDataBuffer(DatasetType::FILE);
 
+  unsigned int bufsize = iniparser_getint(ini, "DataSet:BufferSize", 1);
   std::function<int(const char *, DatasetDataUsageType, bool)> parse_and_set =
     [&](const char *key, DatasetDataUsageType dt, bool required) -> int {
     const char *path = iniparser_getstring(ini, key, NULL);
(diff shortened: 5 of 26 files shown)


nntrainer/models/neuralnet.h
```diff
@@ -511,7 +522,8 @@ class NeuralNetwork : public ml::train::Model {
 
   std::shared_ptr<Manager> manager; /**< nntrainer manager */
 
-  std::shared_ptr<DataBuffer> data_buffer; /**< Data Buffer to get Input */
+  std::array<std::shared_ptr<DataBuffer>, 3>
+    data_buffers; /**< Data Buffers to get Input */
```
zhoonit, line 526: Now, neuralnetwork holds 3 data buffers (train, test, val each)

Applications/TransferLearning/CIFAR_Classification/jni/main_func.cpp
```diff
@@ -273,10 +273,12 @@ int main(int argc, char *argv[]) {
...
    */
-  std::shared_ptr<ml::train::Dataset> dataset;
+  std::shared_ptr<ml::train::Dataset> dataset_train, dataset_val;
   try {
-    dataset = createDataset(ml::train::DatasetType::GENERATOR, getBatch_train,
-                            getBatch_val);
+    dataset_train =
+      createDataset(ml::train::DatasetType::GENERATOR, getBatch_train);
+    dataset_val =
+      createDataset(ml::train::DatasetType::GENERATOR, getBatch_val);
```
zhoonit, line 281: Create each dataset for each on usage.

Applications/TransferLearning/CIFAR_Classification/jni/main_func.cpp
```diff
@@ -306,7 +308,8 @@ int main(int argc, char *argv[]) {
     std::cerr << "Error during readModel, reason: " << e.what() << std::endl;
     return 1;
   }
-  model->setDataset(dataset);
+  model->setDataset(ml::train::DatasetDataUsageType::DATA_TRAIN, dataset_train);
+  model->setDataset(ml::train::DatasetDataUsageType::DATA_VAL, dataset_val);
```
zhoonit, line 312: set by usage as well in the model side

api/capi/include/nntrainer_internal.h
```diff
@@ -101,7 +101,7 @@ typedef struct {
  */
 typedef struct {
   uint magic;
-  std::shared_ptr<ml::train::Dataset> dataset;
+  std::array<std::shared_ptr<ml::train::Dataset>, 3> dataset;
```
zhoonit, line 104: In the capi point of view, the change will be abstracted away by dataset having 3 databuffers.

api/capi/src/nntrainer.cpp
```diff
@@ -492,12 +501,35 @@ int ml_train_model_set_dataset(ml_train_model_h model,
...
+      }
+    }
+
+    if (test_set != nullptr) {
+      status =
+        m->setDataset(ml::train::DatasetDataUsageType::DATA_TEST, test_set);
+      if (status != ML_ERROR_NONE) {
+        return status;
+      }
+    }
```
zhoonit, line 530: Creating a dataset with (deprecated) capi will be split into three underlying calls.

api/capi/src/nntrainer.cpp
```diff
@@ -755,10 +787,21 @@ int ml_train_dataset_set_property(ml_train_dataset_h dataset, ...) {
...
 
-    d = nndataset->dataset;
+    db = nndataset->dataset;
   }
 
-  returnable f = [&]() { return d->setProperty(arg_list); };
+  returnable f = [&]() {
+    int status = ML_ERROR_NONE;
+    for (auto &d : db) {
+      if (d != nullptr) {
```
zhoonit, line 796: Upon calling  ml_train_set_dataset_set_property, it is calling setProperty for the underlying dataset which is owned by the handle.

kparichay (commented):
Let's remove DatasetType type while creating the dataset from the C++ API ( current signature `std::unique_ptr<Dataset> createDataset(DatasetType type, const char *file);`), and rather pass it while setting this dataset to the model? I do understand that this is needed for the C-API, so ml_train_dataset_h can store the dataset type inside its structure and pass it while setting the dataset appropriately.

There is just one downside, the C++ API and C API will differ at this stage. To avoid this, make create a new DatasetType for C++ `ANY` where the type of dataset is not yet known (which will be default value) and will be set when setting the dataset to the model. Now, the C++ API is a superset of C-API.

zhoonit:
> Let's remove DatasetType type while creating the dataset from the C++ API ( current signature std::unique_ptr<Dataset> createDataset(DatasetType type, const char *file);), and rather pass it while setting this dataset to the model? I do understand that this is needed for the C-API, so ml_train_dataset_h can store the dataset type inside its structure and pass it while setting the dataset appropriately.

> There is just one downside, the C++ API and C API will differ at this stage. To avoid this, make create a new DatasetType for C++ ANY where the type of dataset is not yet known (which will be default value) and will be set when setting the dataset to the model. Now, the C++ API is a superset of C-API.

Can't agree more. I will remove type on the upcoming PR's. Currently type is a member of databuffer and need some broader changes to directly apply here.

kparichay:
> > Let's remove DatasetType type while creating the dataset from the C++ API ( current signature std::unique_ptr createDataset(DatasetType type, const char *file);), and rather pass it while setting this dataset to the model? I do understand that this is needed for the C-API, so ml_train_dataset_h can store the dataset type inside its structure and pass it while setting the dataset appropriately.
> 
> > There is just one downside, the C++ API and C API will differ at this stage. To avoid this, make create a new DatasetType for C++ ANY where the type of dataset is not yet known (which will be default value) and will be set when setting the dataset to the model. Now, the C++ API is a superset of C-API.
> 
> Can't agree more. I will remove type on the upcoming PR's. Currently type is a member of databuffer and need some broader changes to directly apply here.

Then please a corresponding TODO / add to issue to keep track for it :).

nntrainer/dataset/databuffer_func.cpp
```diff
@@ -86,14 +86,15 @@ int DataBufferFromCallback::init() {
...
 int DataBufferFromCallback::setGeneratorFunc(DatasetDataUsageType type,
-                                             datagen_cb func) {
+                                             datagen_cb func, void *user_data) {
 
   int status = ML_ERROR_NONE;
   switch (type) {
   case DatasetDataUsageType::DATA_TRAIN:
     if (!func)
       return ML_ERROR_INVALID_PARAMETER;
     callback_train = func;
```
kparichay, line 96: I am assuming the dependency on the type of dataset this is (train/valid/test) will be removed in future PRs.

nntrainer/models/model_loader.cpp
```diff
@@ -232,13 +244,6 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
     ml_logi("setting labeData is deprecated!, it is essentially noop now!");
   }
 
-  status = model.data_buffer->setBatchSize(model.batch_size);
```
kparichay, line 230: is this intentionally removed without any replacement?

nntrainer/models/neuralnet.cpp
```diff
@@ -447,8 +448,13 @@ void NeuralNetwork::setBatchSize(unsigned int batch) {
   model_graph.setBatchSize(batch);
   manager->setBatchSize(batch);
 
-  if (data_buffer && data_buffer->setBatchSize(batch_size) != ML_ERROR_NONE)
-    throw std::invalid_argument("Error setting batchsize for the dataset");
+  for (auto &db : data_buffers) {
+    if (db != nullptr) {
+      int status = db->setBatchSize(batch_size);
+      ml_loge("[model] setting batchsize from data buffer failed");
```
kparichay, line 454: missing `status != ML_ERROR_NONE`?

nntrainer/models/neuralnet.cpp
```diff
@@ -551,15 +563,25 @@ int NeuralNetwork::train(std::vector<std::string> values) {
   status = allocate(true);
   NN_RETURN_STATUS();
 
-  /** Setup data buffer properties */
-  status = data_buffer->setClassNum(getOutputDimension()[0].width());
-  NN_RETURN_STATUS();
+  auto initiate_data_buffer = [this](std::shared_ptr<DataBuffer> &db) {
+    int status = db->setClassNum(getOutputDimension()[0].width());
```
kparichay, line 567: can we use `setLabelShape(std::vector<TensorDim>)`?

nntrainer/models/neuralnet.cpp
```diff
@@ -551,15 +563,25 @@ int NeuralNetwork::train(std::vector<std::string> values) {
...
+    NN_RETURN_STATUS();
 
-  status = data_buffer->setFeatureSize(getInputDimension()[0]);
-  NN_RETURN_STATUS();
+    status = db->setFeatureSize(getInputDimension()[0]);
```
kparichay, line 570: update setFeatureSize() to `setInputShape(std::vector<TensorDim>)`?

nntrainer/models/neuralnet.cpp
```diff
@@ -591,45 +613,57 @@ int NeuralNetwork::train_run() {
   auto &label = last_layer_node->getOutputGrad(0);
   auto &in = first_layer_node->getInput(0);
 
+  /// below constant is needed after changing
+  /// databuffer having train, valid, test -> train buffer, valid buffer, test
+  /// buffer After the cahgne, only data train is used inside a databuffer.
+  /// RUN_CONSTANT is a stub value to deal with the situation
+  auto RUN_CONSTANT = DatasetDataUsageType::DATA_TRAIN;
```
kparichay, line 620: RUN_CONSTANT is temporary right, and will be removed once databuffer internally removed its dependency on the type the buffer it is?

nntrainer/models/neuralnet.cpp
```diff
@@ -591,45 +613,57 @@ int NeuralNetwork::train_run() {
...
+
+  auto &[train_buffer, valid_buffer, test_buffer] = data_buffers;
+
+  if (train_buffer == nullptr) {
```
kparichay, line 624: This is outside the scope of this PR, but reduce to function to take a dataset and just run with it depending on its type.
This will simplify this function, and databuffer usage.

zhoonit:
> Then please a corresponding TODO / add to issue to keep track for it :).

Wait, we will need types. 

1. There are some other types of databuffer added in the upcoming PR that rely on the current semantics (eg) `random_categorical`)
2. for ini need to parse the same way
3. setting property after setting to the model won't be allowed.
4. we will need two set of set property in cc, capi. 

 I'll make a templated factory instead.
```cpp
typename <typename Args...>
std::unique_ptr<Dataset>
createDataset(DatasetType type,
              const std::vector<std::string> &properties = {}, Args &&args...);
```

nntrainer/dataset/databuffer_func.cpp
zhoonit: Correct 👍

nntrainer/models/model_loader.cpp
zhoonit: Yes, it's gotta be done at finalize or before call.

nntrainer/models/neuralnet.cpp
zhoonit: it's throwing status at line 455 which is being captured at exception boundary

nntrainer/models/neuralnet.cpp
zhoonit: Similar interface is being proposed. So for now it is matching with `databuffer_callback.`

nntrainer/models/neuralnet.cpp
zhoonit: same here :)

nntrainer/models/neuralnet.cpp
zhoonit: Correct ✔️

nntrainer/models/neuralnet.cpp
zhoonit: Yup, while refactoring, I'll do in this manner :)

nntrainer/models/model_loader.cpp
kparichay: cool.

nntrainer/models/neuralnet.cpp
kparichay: ml_loge is logging always even though the there was no error.

nntrainer/models/neuralnet.cpp
kparichay: alright. Maybe add a TODO for it here.

kparichay:
> > Then please a corresponding TODO / add to issue to keep track for it :).
> 
> Wait, we will need types.
> 
> 1. There are some other types of databuffer added in the upcoming PR that rely on the current semantics (eg) `random_categorical`)
> 2. for ini need to parse the same way
> 3. setting property after setting to the model won't be allowed.
> 4. we will need two set of set property in cc, capi.
> 
> I'll make a templated factory instead.
> 
> ```c++
> typename <typename Args...>
> std::unique_ptr<Dataset>
> createDataset(DatasetType type,
>               const std::vector<std::string> &properties = {}, Args &&args...);
> ```

Should these be other types of dataset or be a property of the existing datasets. Can you list out the types of dataset you are planning (maybe in an issue and we can continue the discussion there). 
Also, lets try to not attach type to dataset in internal code. It can done at the capi implementation side. For the ini also, it can be handled at the ini reading side.

nntrainer/models/neuralnet.cpp
zhoonit: Right, I'll change

zhoonit:
> Also, lets try to not attach type to dataset in internal code. It can done at the capi implementation side. For the ini also, it can be handled at the ini reading side.

I think attaching datatype is needed.

There are list of different dataset types which would likely to have same

1. Mock dataset which generates a random one-hot label (already implemented)
2. ImageFolder dataset (which will also receive a const char*)
3. Specified Format of dataset (perhaps in format of hdf5, receiving const char*)
4. Some custom datasets (just like pytorch support it)

Those use cases can be a property and we put switch cases like current loss layer or activation layer, but I think it's not very extensible. IMO, those would better be a polymorphic object like layer does.

kparichay:
> > Also, lets try to not attach type to dataset in internal code. It can done at the capi implementation side. For the ini also, it can be handled at the ini reading side.
> 
> I think attaching datatype is needed.
> 
> There are list of different dataset types which would likely to have same
> 
> 1. Mock dataset which generates a random one-hot label (already implemented)
> 2. ImageFolder dataset (which will also receive a const char*)
> 3. Specified Format of dataset (perhaps in format of hdf5, receiving const char*)
> 4. Some custom datasets (just like pytorch support it)
> 
> Those use cases can be a property and we put switch cases like current loss layer or activation layer, but I think it's not very extensible. IMO, those would better be a polymorphic object like layer does.
> 
> There soon will be a factory refactor PR, I think we can continue discussion there.

Dataset types like these are fine. Basically, the base dataset is databuffer using callback. All of these are just callbacks implemented for different file/format types. Thats what I meant by property. Depending on the property, this kind of dataset callback would be attached.

DatasetType is confusing keyword as of now, so this type will have to attach but I meant train/test and certain other types based on properties like shuffle etc.

zhoonit:
> DatasetType is confusing keyword as of now, so this type will have to attach but I meant train/test and certain other types based on properties like shuffle etc. When you have the base class ready for dataset, which all other types would extend, please share the PR explicitly for review.

I understood what you mean. I'll make a simple diagram in #303.

nntrainer/models/neuralnet.cpp
zhoonit: Done

nntrainer/models/neuralnet.cpp
zhoonit: Done

test/ccapi/unittest_ccapi.cpp
```diff
@@ -204,7 +204,7 @@ TEST(nntrainer_ccapi, train_with_config_01_p) {
   EXPECT_NO_THROW(model->train());
 
   EXPECT_NEAR(model->getTrainingLoss(), 4.434051, tolerance);
-  EXPECT_NEAR(model->getValidationLoss(), 2.9646113, tolerance);
+  EXPECT_NEAR(model->getValidationLoss(), 2.910938, tolerance);
```
kparichay, line 207: I wonder why this happens?

kparichay (approved):
overall LGTM
Great Job

test/ccapi/unittest_ccapi.cpp
zhoonit: This is because the rand engine has been changed.

before: one rand engine ran for train -> val -> test
now: three rand engine run for each.

nntrainer/models/neuralnet.cpp
jijoongmoon: when you do the refactoring, you may need to consider the thread conflict as well.

jijoongmoon (approved):
LGTM

jijoongmoon:
> > DatasetType is confusing keyword as of now, so this type will have to attach but I meant train/test and certain other types based on properties like shuffle etc. When you have the base class ready for dataset, which all other types would extend, please share the PR explicitly for review.
> 
> I understood what you mean. I'll make a simple diagram in #303.

I want to put another thought to remove file type support. I think we don't need to. Saving order or format of data is quite differ as well. and the data type is differ too. It could be really hard to support. And file format also can support with callback so.. we could deprecate the file support for dataset. then it could make some issues easier.

test/ccapi/unittest_ccapi.cpp
kparichay: right.

PR #1381, opened 2021-07-09
[dataset/cleanup] Remove usage from dataset impl

zhoonit:
- [dataset/cleanup] Remove usage from dataset impl

```
As each dataset now has one usage(train, valid or test).
This patch removes usage handles from dataset impl.

**Major Changes**
1. merge train/test/val members to one
2. clarify some variable names.
3. move global variables to members of an instance.
4. remove databuffer_utils.h as it is no longer used.

**Self evaluation:**

```


nntrainer/dataset/databuffer.h
```diff
@@ -201,101 +184,48 @@ class DataBuffer : public ml::train::Dataset {
...
-  std::thread test_thread;
+  /**< @todo below variable should be owned by databuffer that has fixed size */
+  unsigned int samples_per_epoch; /**< size of samples in the dataset */
+  unsigned int remaining_samples_per_epoch; /**< size of samples remaining in
+                                               current epoch */
 
-  std::vector<std::string> labels;
-  bool validation[NBUFTYPE];
+  bool is_running; /**< flag to check if running */
+  bool is_valid;   /**< check if current buffer is in valid state */
```
kparichay, line 228: how about is_valid -> initialized?

zhoonit: done. fyi) in the upcoming design this will not be used :)

kparichay: yes, that will be even better.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM with small suggestions.

zhoonit:
> Great. one small suggestion to have more unit test cases with this update. ( Cause we had issues related this dataset )

Thanks, as this is on the refactoring progress, I'll add test while refactoring :)

PR #1384, opened 2021-07-11
[layer] Update LSTM for layer_v2 

kparichay:
Update LSTM for layer_v2.
Also enable corresonding unittests and add common unittests.

Update LSTM implementation to reduce the use of temporary tensors
and request temporary tensors for the ones that are necessary.


zhoonit (approved):
LGTM

nntrainer/layers/lstm.cpp
```diff
@@ -228,91 +233,72 @@ void LSTMLayer::forwarding(bool training) {
       acti_func.run_fn(cs, hs);
       hs.multiply_i(ho);
     }
-    // size of h_prev and hs size is same : unit.
-    // size of c_prev and cs is same : unit.
-    h_prev.getBatchSlice(b, 1).copy(hs);
-    c_prev.getBatchSlice(b, 1).copy(cs);
```
jijoongmoon, line 234: I wonder if it pass the unittest?

jijoongmoon (approved):
LGTM

nntrainer/layers/lstm.cpp
kparichay: Yes, the golden unittest of RNN is in unittest_models which has been activated with this PR.
From what I understood, h_prev was used to cache the values of the current calculation for the next hidden state, and then the next hidden state was initialized with it. 
So, I updated to cache directly in the memory of the next hidden state directly.

jijoongmoon: great, thanks.

PR #1387, opened 2021-07-11
[applications] Enable working applications with v2

### LINKED ISSUE #1229: in_length in embedding is redundant
`in_length` in embedding is redundant.
@jijoongmoon Can you please check this once.

kparichay:
Enable working applications with v2. This is done by also enabling split, concat and embedding layer for the V2 design.
Adding some minor updates to the applications to match with the v2 implementation.

An extra property from embedding layer was also removed to work with the new LayerV2 semantics.

resolves #1229

### DIFF
--- a/Applications/TransferLearning/CIFAR_Classification/jni/main.cpp
+++ b/Applications/TransferLearning/CIFAR_Classification/jni/main.cpp
@@ -421,10 +421,19 @@ int main(int argc, char *argv[]) {
    * @brief     Neural Network Create & Initialization
    */
   nntrainer::NeuralNetwork NN;
+  int status = ML_ERROR_NONE;
   try {
-    NN.loadFromConfig(config);
-    NN.compile();
-    NN.initialize();
+    status = NN.loadFromConfig(config);
+    if (status != ML_ERROR_NONE)
+      return status;
+
+    status = NN.compile();
+    if (status != ML_ERROR_NONE)
+      return status;
+
+    status = NN.initialize();
+    if (status != ML_ERROR_NONE)
+      return status;
   } catch (...) {
     std::cerr << "Error during init" << std::endl;
     return 1;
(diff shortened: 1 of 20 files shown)


Applications/TransferLearning/CIFAR_Classification/jni/main.cpp
```diff
@@ -421,10 +421,19 @@ int main(int argc, char *argv[]) {
...
    */
   nntrainer::NeuralNetwork NN;
+  int status = ML_ERROR_NONE;
   try {
-    NN.loadFromConfig(config);
-    NN.compile();
-    NN.initialize();
+    status = NN.loadFromConfig(config);
+    if (status != ML_ERROR_NONE)
+      return status;
```
jijoongmoon, line 428: This looks strange. there is not going to return if there is error. they just throw the error I guess. Isn't it?

jijoongmoon (approved):
LGTM with minor comment

test/unittest/unittest_nntrainer_layers.cpp
```diff
@@ -2251,8 +2251,7 @@ class nntrainer_EmbeddingLayer
 
   virtual void prepareLayer() {
     int status = setProperty("in_dim=50 |"
-                             "out_dim=8 |"
-                             "in_length=12");
```
jijoongmoon, line 2271: So, do you expect to set this from input layer ?

Applications/TransferLearning/CIFAR_Classification/jni/main.cpp
kparichay: NN.loadFromConfig can throw as well as return error. So, added both.

zhoonit (approved):
LGTM

test/unittest/unittest_nntrainer_layers.cpp
kparichay: `in_length` can be interpreted directly based on the shape of the input. So, it has been removed.

Issue #904, opened 2021-01-27
Evaluate ResNet


jijoongmoon:

zhoonit:
Additional todo regarding this item

- [ ] speed benchmark of other frameworks with same data.
<details> <summary> benchmark results </summary>

| framework            | batch_size | real time   |   data_size |
|----------------------|------------|-------------|------------|
| nntrainer(fake data) | 32         | 11m37.97926 |     512    |
| nntrainer(fake data) | 128        | 10m45.761   |      512    |

</details> 

- [ ] Create readme.md 
readme.md must cite resnet and cifar100.

zhoonit:
Evaluating resnet done (checking speed hasn't been done but this can be done relatively easily later)

For the result please check out the [link](https://github.sec.samsung.net/jhoon-it-lee/cifar100_resnet/issues/5)

kparichay:
@zhoonit The links are internal. Please check the permissions and open for outside if possible.

PR #1388, opened 2021-07-12
[Resnet] Sync resnet app with validated model

zhoonit:
- [Resnet] Sync resnet app with validated model

```
This patch syncs resnet application with validated model.
Changes include:

1. s/setIteration/updateIteration
2. normalize input
3. fixed batchsize of 128 -> args received from the sysargs
4. output layer to copy
5. silent failed to read epoch idx and iteration

**Self evaluation:**

```

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -430,10 +430,17 @@ void NeuralNetwork::readModel() {
     (*iter)->read(model_file);
   }
 
-  checkedRead(model_file, (char *)&tmp.epoch_idx, sizeof(epoch_idx),
-              "[NeuralNetwork::readModel] failed to read epoch_idx");
-  checkedRead(model_file, (char *)&tmp.iter, sizeof(iter),
-              "[NeuralNetwork::readModel] failed to read iteration");
+  try {
+    /// this is assuming that the failure is allowed at the end of the file
+    /// read. so, after this line, additional read shouldn't be called
+    checkedRead(model_file, (char *)&tmp.epoch_idx, sizeof(epoch_idx),
+                "[NeuralNetwork::readModel] failed to read epoch_idx");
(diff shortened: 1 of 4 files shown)


myungjoo (approved):

nntrainer/models/neuralnet.cpp
```diff
@@ -430,10 +430,14 @@ void NeuralNetwork::readModel() {
...
-              "[NeuralNetwork::readModel] failed to read epoch_idx");
-  checkedRead(model_file, (char *)&tmp.iter, sizeof(iter),
-              "[NeuralNetwork::readModel] failed to read iteration");
+  try {
+    checkedRead(model_file, (char *)&tmp.epoch_idx, sizeof(epoch_idx),
+                "[NeuralNetwork::readModel] failed to read epoch_idx");
+    checkedRead(model_file, (char *)&tmp.iter, sizeof(iter),
+                "[NeuralNetwork::readModel] failed to read iteration");
+  } catch (...) {
+    std::cerr << "failed to read epoch idx, proceeding with default index\n";
```
kparichay, line 439: please add a comment here that this failure is allowed at the end of file read. If this has failed, no other read operation should be performed on this file.
Or maybe move model_file.close() inside try and catch both to emphasize on it.

kparichay (approved):

nntrainer/models/neuralnet.cpp
zhoonit: Good idea, added :)

jijoongmoon (approved):
LGTM with one miner. 
. Update the document with some reuslts. ( later PR, I guess )

jijoongmoon:
I wonder that validated model means keras model ? If user download the keras model and then, is this model configuration same with it?

zhoonit:
yes, they are in the internal repo. #904

I'll update those two in a separate pr.

PR #1389, opened 2021-07-12
[Wait for #1381] [Dataset] Introduce data producer

zhoonit:
**[Pending commit: #1381]**

- [Dataset] Introduce data producer

```
This patch introduces data producer which abstracts a single iteration
creator.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

```

PR #1391, opened 2021-07-12
[Dataset] Add random producer

zhoonit:
- [Dataset] Add random producer

```
Add onehot random producer which generatese a fake data

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/random_data_producers.cpp
+++ b/nntrainer/dataset/random_data_producers.cpp
@@ -1,0 +1,168 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   random_data_producers.cpp
+ * @date   09 July 2021
+ * @brief  This file contains various random data producers
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+#include <random_data_producers.h>
+
+#include <base_properties.h>
+#include <node_exporter.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+/**
+ * @brief Props containing min value
+ *
+ */
+class PropsMin : public Property<float> {
+public:
+  /**
+   * @brief Construct a new props min object with a default value
+   *
+   * @param value default value
+   */
+  PropsMin(float value = 0.0f) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key = "min"; /**< unique key to access */
+  using prop_tag = float_prop_tag;          /**< property type */
+};
+
+/**
+ * @brief Props containing max value
+ *
+ */
+class PropsMax : public Property<float> {
+public:
+  /**
+   * @brief Construct a new props max object with a default value
+   *
+   * @param value default value
+   */
+  PropsMax(float value = 1.0f) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key = "max"; /**< unique key to access */
+  using prop_tag = float_prop_tag;          /**< property type */
+};
+
+/**
+ * @brief Props containing number of samples
+ *
+ */
+class PropsDataSize : public Property<unsigned int> {
+public:
+  /**
+   * @brief Construct a new props data size object with a default value
+   *
+   * @param value default value
+   */
+  PropsDataSize(unsigned int value = 512) :
+    nntrainer::Property<unsigned int>(value) {}
+  static constexpr const char *key = "size"; /**< unique key to access */
+  using prop_tag = uint_prop_tag;            /**< property type */
+};
+
+RandomDataOneHotProducer::RandomDataOneHotProducer() :
+  rd_one_hot_props(new Props()) {}
+
+RandomDataOneHotProducer::~RandomDataOneHotProducer() {}
+
+const std::string RandomDataOneHotProducer::getType() const {
+  return RandomDataOneHotProducer::type;
+}
+
+unsigned long long
+RandomDataOneHotProducer::size(const std::vector<TensorDim> &input_dims,
+                               const std::vector<TensorDim> &label_dims) const {
+  return std::get<PropsDataSize>(*rd_one_hot_props).get();
+}
+
+void RandomDataOneHotProducer::setProperty(
+  const std::vector<std::string> &properties) {
+  auto left = loadProperties(properties, *rd_one_hot_props);
+  NNTR_THROW_IF(!left.empty(), std::invalid_argument)
+    << "There are unparsed properties, size: " << left.size();
+}
+
+DataProducer::Gernerator
+RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
+                                   const std::vector<TensorDim> &label_dims) {
+  /** check if the given producer is ready to finalize */
+  auto &[min_, max_, sz] = *rd_one_hot_props;
+
+  /// @todo expand this to non onehot case
+  NNTR_THROW_IF(std::any_of(label_dims.begin(), label_dims.end(),
+                            [](const TensorDim &dim) {
+                              return dim.channel() != 1 || dim.height() != 1;
+                            }),
+                std::invalid_argument)
+    << "Label dimension containing channel or height not allowed";
+
+  NNTR_THROW_IF(min_.get() > max_.get(), std::invalid_argument)
+    << "Min value is bigger then max value, min: " << min_.get()
+    << "max: " << max_.get();
+
+  NNTR_THROW_IF(size(input_dims, label_dims) == 0, std::invalid_argument)
+    << "size is zero, dataproducer does not provide anything";
+
+  /** prepare states for the generator */
+  std::vector<std::uniform_int_distribution<unsigned int>> label_chooser_;
+  label_chooser_.reserve(label_dims.size());
+  std::transform(label_dims.begin(), label_dims.end(),
+                 std::back_inserter(label_chooser_),
+                 [this](const TensorDim &label_dim) {
+                   return std::uniform_int_distribution<unsigned int>(
+                     0, label_dim.width() - 1);
+                 });
+
+  std::mt19937 rng;
+  rng.seed(getSeed());
+
+  /** DataProducer::Generator */
+  return [rng, sz = size(input_dims, input_dims), input_dims, label_dims,
+          min_ = min_.get(), max_ = max_.get(), current_iteration = 0ULL,
+          label_chooser = std::move(label_chooser_)]() mutable {
+    if (current_iteration++ == sz) {
+      current_iteration = 0;
+      return DataProducer::Iteration(true, {}, {});
+    }
+
+    auto populate_tensor = [&](const TensorDim &input_dim) {
+      Tensor t(input_dim);
+      t.setRandUniform(min_, max_);
+
+      return t;
+    };
+
+    auto populate_label =
+      [&](const TensorDim &label_dim,
+          std::uniform_int_distribution<unsigned int> &label_dist_) {
+        Tensor t(label_dim);
+        t.setZero();
+        for (unsigned int b = 0; b < t.batch(); ++b) {
+          t.setValue(b, 0, 0, label_dist_(rng), 1);
+        }
+        return t;
+      };
+
+    std::vector<Tensor> inputs;
+    inputs.reserve(input_dims.size());
+
+    std::vector<Tensor> labels;
+    labels.reserve(label_dims.size());
+
+    std::transform(input_dims.begin(), input_dims.end(),
+                   std::back_inserter(inputs), populate_tensor);
+
+    std::transform(label_dims.begin(), label_dims.end(), label_chooser.begin(),
+                   std::back_inserter(labels), populate_label);
+
+    return DataProducer::Iteration(false, inputs, labels);
+  };
+}
+} // namespace nntrainer
--- a/nntrainer/dataset/data_producers.h
+++ b/nntrainer/dataset/data_producers.h
@@ -85,7 +85,7 @@ public:
                               const std::vector<TensorDim> &label_dims) = 0;
 
   /**
-   * @brief get size of the iteration given input_dims, label_dims, if size
+   * @brief get size of total dataset given input_dims, label_dims, if size
    * cannot be determined, this function must return
    * DataProducer::SIZE_UNDEFINED;
    *
--- a/nntrainer/dataset/meson.build
+++ b/nntrainer/dataset/meson.build
@@ -2,7 +2,8 @@ dataset_sources = [
   'databuffer.cpp',
   'databuffer_factory.cpp',
   'databuffer_file.cpp',
-  'databuffer_func.cpp'
+  'databuffer_func.cpp',
+  'random_data_producers.cpp'
 ]
 
 dataset_headers = [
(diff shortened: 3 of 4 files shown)


nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <node_exporter.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+/**
+ * @brief Props containing min value
+ *
+ */
+class PropsMin : public Property<float> {
```
kparichay, line 25: this is minor but how about PropsRangeMin? corresponding key would become range_min? or you could also consider range_start.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+  PropsMin(float value = 0.0f) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key = "min"; /**< unique key to access */
+  using prop_tag = float_prop_tag;          /**< property type */
+};
+
+/**
+ * @brief Props containing max value
+ *
+ */
+class PropsMax : public Property<float> {
```
kparichay, line 41: correspondingly range max or range end.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+RandomDataOneHotProducer::size(const std::vector<TensorDim> &input,
+                               const std::vector<TensorDim> &output) const {
+  return std::get<PropsDataSize>(*rd_one_hot_props).get();
+}
+
+void RandomDataOneHotProducer::setProperty(
+  const std::vector<std::string> &properties) {
+  auto left = loadProperties(properties, *rd_one_hot_props);
+  NNTR_THROW_IF(!left.empty(), std::invalid_argument)
+    << "There is unparsed properties, size: " << left.size();
```
kparichay, line 89: There `are` unparsed properties

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+RandomDataOneHotProducer::~RandomDataOneHotProducer() {}
+
+const std::string RandomDataOneHotProducer::getType() const {
+  return RandomDataOneHotProducer::type;
+}
+
+unsigned long long
+RandomDataOneHotProducer::size(const std::vector<TensorDim> &input,
+                               const std::vector<TensorDim> &output) const {
+  return std::get<PropsDataSize>(*rd_one_hot_props).get();
```
kparichay, line 82: I wonder the purpose of the property PropsDataSize here? Shouldn't the size be size of input/output?

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+DataProducer::Gernerator
+RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
+                                   const std::vector<TensorDim> &label_dims) {
+  /** check if the given producer is ready to finalize */
+  auto &[min_, max_, sz] = *rd_one_hot_props;
+
+  NNTR_THROW_IF(std::any_of(label_dims.begin(), label_dims.end(),
+                            [](const TensorDim &dim) {
+                              return dim.channel() != 1 || dim.height() != 1;
```
kparichay, line 100: any reason to have this restriction? IMO consider this class to be a generic data producer which can work for super resolution or object detection samples as well than just classifier. Removing this check will do so.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+   *
+   * @param value default value
+   */
+  PropsMax(float value = 1.0f) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key = "max"; /**< unique key to access */
+  using prop_tag = float_prop_tag;          /**< property type */
+};
+
+/**
+ * @brief Props containing max value
```
kparichay, line 54: update brief?

nntrainer/dataset/random_data_producers.cpp
kparichay: In the implementation here, it seems like it returns "the total number of inputs + outputs pairs which the dataset can provide", but in documentation in #1389, the description seems to give the size of the data the producer will have to allocate/read for each iteration. 
Can you please either update the description or check if this return value is correct.

nntrainer/dataset/random_data_producers.cpp
zhoonit: This will affect upcoming PR which uses this property so I'll do it later :)

zhoonit: There are some tests related to the property here, I'll make it as a separate PR later :)

nntrainer/dataset/random_data_producers.cpp
zhoonit: Done

nntrainer/dataset/random_data_producers.cpp
zhoonit: Confirming that size here is number of iteration here, but as per discussion at the last Friday, the behavior of `size()` will likely to change to number of elements it has.

nntrainer/dataset/random_data_producers.cpp
zhoonit: My thinking is that there are number of different random data producer (categorical, sparse)... This is like instead 'loss' has one type + mode, each mode is one type. As the classname is `RandomDataOneHotProducer` this restriction is added.

Added in the class brief that
`* @brief RandomDataProducer which generates a onehot vector as a label`

nntrainer/dataset/random_data_producers.cpp
zhoonit: Done

nntrainer/dataset/random_data_producers.cpp
kparichay: alight.

nntrainer/dataset/random_data_producers.cpp
kparichay: Please update the documentation in this PR to match the intended behavior. In the header of data_producers, the description is a bit different or unclear.

nntrainer/dataset/random_data_producers.cpp
kparichay: Then it will be RandomDataProducer -> RandomSingleLabelDataProducer (as this cannot be used for multi-class classification problems). If you keep the current name, please add in description that only single label is supported.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+ */
+class PropsMax : public Property<float> {
+public:
+  /**
+   * @brief Construct a new props max object with a default value
+   *
+   * @param value default value
+   */
+  PropsMax(float value = 1.0f) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key = "max"; /**< unique key to access */
```
jijoongmoon, line 49: does it requre comparison with PropsMin when it set?? If it is to be property, then how we check when it is assigned as we planed... Just check type?

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+  using prop_tag = float_prop_tag;          /**< property type */
+};
+
+/**
+ * @brief Props containing number of iterations
```
jijoongmoon, line 54: is it data size or number of iteration?

nntrainer/dataset/random_data_producers.cpp
jijoongmoon: and also input and output are not relevant here. Is it because it is random?? But even if it is random.. I guess the required size of input is necessary to generate data.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+  NNTR_THROW_IF(std::any_of(label_dims.begin(), label_dims.end(),
+                            [](const TensorDim &dim) {
+                              return dim.channel() != 1 || dim.height() != 1;
+                            }),
+                std::invalid_argument)
+    << "Label dimension containing channel or height not allowed";
+
+  NNTR_THROW_IF(min_.get() > max_.get(), std::invalid_argument)
+    << "Min value is bigger then max value, min: " << min_.get()
+    << "max: " << max_.get();
```
jijoongmoon, line 107: So we just check the min, max values during finalize, not during set property.

nntrainer/dataset/random_data_producers.cpp
jijoongmoon: I recommend to keep RandomDataProducer with removing restriction.  If it is SingleLabel, then it is expected to have MultiLabel and also we need to define what is MultiLabel. ( Channel, Height?? )

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  /** DataProducer::Generator */
+  return [rng, sz = size(input_dims, input_dims), input_dims, label_dims,
+          min_ = min_.get(), max_ = max_.get(), current_iteration = 0ULL,
+          label_chooser = std::move(label_chooser_)]() mutable {
+    if (current_iteration++ == sz) {
+      current_iteration = 0;
+      return DataProducer::Iteration(true, {}, {});
+    }
+
```
jijoongmoon, line 133: I wonder if this random data is going to converge. I mean.. there is no relation between data and label.

nntrainer/dataset/random_data_producers.cpp
zhoonit: Could you explain why multilabel is not supported?

if you have more then one label, it's fine.
It's returning one-hot vector and it is the name of the class. (`RandomDataOneHotProducer`)

zhoonit: > I recommend to keep RandomDataProducer with removing restriction. If it is SingleLabel, then it is expected to have MultiLabel and also we need to define what is MultiLabel. ( Channel, Height?? )

It's already supporting multi label and not restricted to give single label

eg) giving label dims of {1:1:1:5, 1:1:1:7} will return two labels.

I think we can pass a vectors instead of vaguely using a single label.

The class we are dicussing is `RandomDataOneHotProducer` and it's serving it's purpose. If we need another data producer, I think we can add another class (the reason why `random_data_producers.h` not `random_data_producer.h`)

nntrainer/dataset/random_data_producers.cpp
zhoonit: Ah, I understood. This has changed to label_dims, also updated the documentation in he data_producers.h

nntrainer/dataset/random_data_producers.cpp
zhoonit: also s/`* @brief RandomDataProducer which generates a onehot vector as a label`/`* @brief A family of random data producer which generates a onehot vector as a label`

nntrainer/dataset/random_data_producers.cpp
zhoonit: For the record, Props data size has been changed to propsNumSamples in the future PR

nntrainer/dataset/random_data_producers.cpp
zhoonit: It's gotta be actually number_of_samples, I'll update the description :)

nntrainer/dataset/random_data_producers.cpp
zhoonit: Yes :), setProperty validates if given property is  valid. finalize validates if given combination is valid.

nntrainer/dataset/random_data_producers.cpp
zhoonit: Yes, this is just a sample whoever can use if they built a model but not data, but want to run to see if their model has a nyproblem.

jijoongmoon: > Yes, this is just a sample whoever can use if they built a model but not data, but want to run to see if their model has a nyproblem.

so.. I guess.. it is not guarantee to converge while training with this dataset. So.. it is not for training.. it is just for the test of generation of data?? On second thought.. then why do we need this?.. just functional test? But with this kind of dataset, we do not test except semantics.. I guess. Is it the reason to support this?

nntrainer/dataset/random_data_producers.cpp
jijoongmoon: > Could you explain why multilabel is not supported?
> 

What I mean is that we don't need to just adding another method if we support same feature with small change of others.

zhoonit: In that case, we will have a mode='something' rather than a type='something', this might be appropriate if we are to have another type of data producer. I think if we have a need for other data producer, we can decide at that time :)

nntrainer/dataset/random_data_producers.cpp
zhoonit: The intention here is that, as databuffer is going through a big change, this kind of simple implementation works as a reference. If I am to test on a batch queue or databuffer, we'd better test on a simplest possible implementation on the designed interface, so that the given data producer would have no problem but batch queue has problem in case of batch queue test fails.

The refactored databuffers and components like batch queues are tested on this referential implementation without having compilated, possibly problematic producer involved. Because it's simple, it didn't take more than two hours to implement this component.

nntrainer/dataset/random_data_producers.cpp
jijoongmoon: > In that case, we will have a mode='something' rather than a type='something', this might be appropriate if we are to have another type of data producer. I think if we have a need for other data producer, we can decide at that time :)

What I mean is, just removing the error checking of channel and height will work like @kparichay said.

nntrainer/dataset/random_data_producers.cpp
jijoongmoon: > The intention here is that, as databuffer is going through a big change, this kind of simple implementation works as a reference.

I guess this is matter when we develop the code but not the matter of actual user of nntrainer when the new databuffer is done.  And also if it is just for the test queue or databuffer, it need to be test rather than main feature.

> Because it's simple, it didn't take more than two hours to implement this component.

In my opinion, I fully understand the reason of this ( reducing debugging time and etc. ) but it is not a matter of implementation time issue. I mean, if we are going to support the custom data set ( I guess it is callback feature, correct me if it is not ), this random data set producer is more appropriate for that. Also, we do need example for it too.

zhoonit: I see :) I guess we can migrate this to custom example some time soon!
For now, we need a bit more infras to support this as a custom. (like ini parsing and app context registering.)

nntrainer/dataset/random_data_producers.cpp
zhoonit: Even if we delete this, the code is not supporting multi channel, multi height but only onehot for now.
I'll add a todo instead to support this.

jijoongmoon: > Even if we delete this, the code is not supporting multi channel, multi height but only onehot for now.
> I'll add a todo instead to support this.

Even if it is not able to support the multi channel or multi height, we can give more flexibility. Anyway, Let's leave it as in todo then.

nntrainer/dataset/random_data_producers.cpp
jijoongmoon: Thanks for your considering. I guess this is ok for now. But once we finished implementation the new dataset, then we need to migrate this to examples.

jijoongmoon (approved):
LGTM

nntrainer/dataset/random_data_producers.cpp
kparichay: > Could you explain why multilabel is not supported?

I guess in that case, there are multiple labels for a given input. So, the one hot output would be of dimension like {b,1,multi_labels, on_hot_vector_size} (where multi_labels value is not fixed and is different for each input), but leaving it as a TODO is fine for now :).

kparichay (approved):
LGTM

PR #1392, opened 2021-07-12
[Dataset] Add test for the random dataproducer

zhoonit:
- [Dataset] Add test for the random dataproducer

```
- Add random data producer tests
- Add abstract test for more data producers

**Self evaluation:**

```

### DIFF
--- a/test/unittest/datasets/data_producer_common_tests.cpp
+++ b/test/unittest/datasets/data_producer_common_tests.cpp
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file data_producer_common_tests.cpp
+ * @date 12 July 2021
+ * @brief Common test for nntrainer data producers (Param Tests)
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug No known bugs except for NYI items
+ */
+#include <gtest/gtest.h>
+
+#include <data_producer_common_tests.h>
+
+void DataProducerSemantics::SetUp() {
+  auto [producerFactory, properties, input_dims_, label_dims_, validator_,
+        result_] = GetParam();
+
+  producer = producerFactory(properties);
+  input_dims = std::move(input_dims_);
+  label_dims = std::move(label_dims_);
+  result = result_;
+  validator = std::move(validator_);
+
+  if (result != DataProducerSemanticsExpectedResult::SUCCESS) {
+    ASSERT_EQ(validator, nullptr)
+      << "Given expected result of not success, validator must be empty!";
+  }
+}
+
+void DataProducerSemantics::TearDown() {}
+
+TEST_P(DataProducerSemantics, finalize_pn) {
+  if (result == DataProducerSemanticsExpectedResult::FAIL_AT_FINALIZE) {
+    EXPECT_ANY_THROW(producer->finalize(input_dims, label_dims));
+  } else {
+    EXPECT_NO_THROW(producer->finalize(input_dims, label_dims));
+  }
+}
+
+TEST_P(DataProducerSemantics, fetch_one_epoch_or_10_iteration_pn) {
+  if (result != DataProducerSemanticsExpectedResult::SUCCESS) {
+    return; // skip this test
+  }
+
+  auto generator = producer->finalize(input_dims, label_dims);
+  auto sz = producer->size(input_dims, label_dims);
+  bool has_fixed_size = sz != nntrainer::DataProducer::SIZE_UNDEFINED;
+
+  if (!has_fixed_size) {
+    sz = 5;
+  }
+
+  for (unsigned i = 0; i < sz; ++i) {
+    auto [last, ins, labels] = generator();
+
+    ASSERT_FALSE(last) << " reached last at iteration: " << i << '\n';
+    if (validator) {
+      ASSERT_TRUE(validator(ins, labels))
+        << " failed validation for iteration: " << i << '\n';
+    }
+  }
+
+  if (has_fixed_size) {
+    {
+      auto [last, ins, labels] = generator();
+      EXPECT_TRUE(last);
+    }
+
+    {
+      auto [last, ins, labels] = generator();
+      EXPECT_TRUE(validator(ins, labels))
+        << "failed last validation after one epoch\n";
+      EXPECT_FALSE(last);
+    }
+  }
+}
--- a/test/unittest/datasets/unittest_random_data_producers.cpp
+++ b/test/unittest/datasets/unittest_random_data_producers.cpp
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file data_producer_common_tests.cpp
+ * @date 12 July 2021
+ * @brief Common test for nntrainer dataset producers (Param Tests)
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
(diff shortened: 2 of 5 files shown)


test/unittest/datasets/data_producer_common_tests.cpp
```diff
@@ -0,0 +1,78 @@
...
+  bool has_fixed_size = sz != nntrainer::DataProducer::SIZE_UNDEFINED;
+
+  if (!has_fixed_size) {
+    sz = 5;
+  }
+
+  for (unsigned i = 0; i < sz; ++i) {
+    auto [last, ins, labels] = generator();
+
+    ASSERT_FALSE(last) << " reached last at iteration: " << i << '\n';
```
kparichay, line 58: why use ASSERT than EXPECT?

test/unittest/datasets/unittest_random_data_producers.cpp
```diff
@@ -0,0 +1,66 @@
...
+        return false;
+      }
+      data++;
+    }
+    return true;
+  };
+
+  /// label validator: sum of all is equal to batch
+  auto label_valid = [](const nntrainer::Tensor &t) {
+    return fabs(t.batch() - t.sum({0, 1, 2, 3}).getValue(0, 0, 0, 0) < 1e-7);
```
kparichay, line 34: This would turn out to be valid if one of the batch values adds to 2 and some other value adds to 0.
Instead, please update the check for each batch, the sum to be 1.

test/unittest/datasets/unittest_random_data_producers.cpp
```diff
@@ -0,0 +1,66 @@
...
+  return f;
+}
+auto random_onehot_success = DataProducerSemanticsParamType(
+  createDataProducer<nntrainer::RandomDataOneHotProducer>,
+  {"min=0", "max=1", "size=10"}, {{3, 2, 4, 5}}, {{3, 1, 1, 10}},
+  random_onehot_validator(0, 1), DataProducerSemanticsExpectedResult::SUCCESS);
+
+auto random_onehot_min_over_max = DataProducerSemanticsParamType(
+  createDataProducer<nntrainer::RandomDataOneHotProducer>,
+  {"min=2", "max=1", "size=10"}, {{3, 2, 4, 5}}, {{3, 1, 1, 10}}, nullptr,
```
kparichay, line 56: is this testing the random_data_producer or testing the test setup itself :)?

test/unittest/datasets/unittest_random_data_producers.cpp
```diff
@@ -0,0 +1,66 @@
...
+  DataProducerSemanticsExpectedResult::FAIL_AT_FINALIZE);
+
+auto random_onehot_invalid_label_shape = DataProducerSemanticsParamType(
+  createDataProducer<nntrainer::RandomDataOneHotProducer>, {}, {{3, 2, 4, 5}},
+  {{3, 1, 2, 10}}, nullptr,
+  DataProducerSemanticsExpectedResult::FAIL_AT_FINALIZE);
```
jijoongmoon, line 62: For me, it looks like validator validates the random data itself. But I wonder if it is testing the random generator of C++. I'm not sure what it is trying to test?

test/unittest/datasets/data_producer_common_tests.cpp
zhoonit: If last is true, there is no meaning proceeding, while producing too many error logs making it hard to identify, where it was actually dead.

test/unittest/datasets/unittest_random_data_producers.cpp
zhoonit: Let me change this within the element-wise generator change.

test/unittest/datasets/unittest_random_data_producers.cpp
zhoonit: It's kinda both :p

test/unittest/datasets/unittest_random_data_producers.cpp
zhoonit: This test is to fixating semantics of data producer with some simple codes and simpler implementation.
The test itself is reused over other types of producers.

test/unittest/datasets/unittest_random_data_producers.cpp
zhoonit: added todo instead

test/unittest/datasets/data_producer_common_tests.cpp
kparichay: Then add a manual if. In my understanding, ASSERT is used for setting up the test case, but not the actual test itself.
In this case, it seems like you put this to avoid printing all the mismatch in the loops and just print the first, but IMO, the printing all can be helpful, as that information is helpful (it is possible for a random piece of data to mismatch but others do match and these observations hints what the error can be in general).

test/unittest/datasets/unittest_random_data_producers.cpp
kparichay: Please update this before this code/this type of check is used at other places :).

test/unittest/datasets/unittest_random_data_producers.cpp
kparichay: I think this test is a bit misleading, and this type of tests exists in future PR as well. The testing of the test setup can be done at once, and no need for the data producer types to do it again.

zhoonit: Could you explain why it's misleading? This test case tests random producer.

I mean I have tested the test setup while building this test, and if test code runs fine, it means test setup is correct and the test is correct so it's testing both.

test/unittest/datasets/unittest_random_data_producers.cpp
zhoonit: this type check is only used inside random producers, other producer will have it's own validator.

test/unittest/datasets/data_producer_common_tests.cpp
zhoonit: Continuing doesn't make sense after this line fails. Same things applied already to unittest_models. 
Adding a manual will do exactly same. 

Please refer to below.

http://google.github.io/googletest/primer.html#assertions

> Usually EXPECT_* are preferred, as they allow more than one failure to be reported in a test. However, you should use ASSERT_* if it doesn’t make sense to continue when the assertion in question fails.

test/unittest/datasets/unittest_random_data_producers.cpp
kparichay: cool.

test/unittest/datasets/unittest_random_data_producers.cpp
kparichay: > Could you explain why it's misleading? This test case tests random producer.

As per the test, this test will check if the validator if nullptr and then the other tests performed are exactly the same as the tests above it. So, in this way, I couldn't see which extra test does this unit test do.

zhoonit: the above test is positive test (DataProducerSemanticsExpectedResult::SUCCESS)
while this test is negative test (DataProducerSemanticsExpectedResult::FAIL_AT_FINALIZE), it is negative test because the case is that the property value "min" is over "max" which is meant to fail, it is not failing because the validator is nullptr but finalize fails.  I am not getting why it's exactly the same as the test above it. Wonder if I am missing something.

kparichay: aah, I thought it was because the validator is nullptr.
then its fine.

kparichay (approved):
LGTM

test/unittest/datasets/data_producer_common_tests.cpp
kparichay: I guess its our choice, either works :).

jijoongmoon (approved):
LGTM

test/unittest/datasets/unittest_random_data_producers.cpp
jijoongmoon: If it is just for the demo as an example, then it is ok for now. Thanks.

PR #1393, opened 2021-07-12
[application] Enable custom application

kparichay:
Update plugged layer for V2 design.
The corresponding updates in app_context have also been made.

Enable custom application for v2 layer design.
This also includes updating the power layer and mae loss layer implementation
with layerv2 and updating their corresponding unittests.

### DIFF
--- a/Applications/Custom/mae_loss.h
+++ b/Applications/Custom/mae_loss.h
@@ -14,82 +14,80 @@
 #define __MAE_LOSS_LAYER_H__
 #include <string>
 
-/// @todo migrate these to API and ensure those headers are exposed to devel
-#include <layer_internal.h>
-#include <manager.h>
-#include <tensor.h>
+#include <layer_context.h>
+#include <layer_devel.h>
+#include <node_exporter.h>
 
 namespace custom {
 
 /**
  * @brief A sample loss layer which calculates mean absolute error from output
- * @todo update this to LayerV1
  *
  */
-class MaeLossLayer final : public nntrainer::LayerV1 {
+class MaeLossLayer final : public nntrainer::Layer {
 public:
   /**
-   * @brief Construct a new Pow Layer object that does elementwise power
+   * @brief Construct a new MAE Layer object that does elementwise power
    *
    */
-  MaeLossLayer() : LayerV1() {}
+  MaeLossLayer() : Layer() {}
 
   /**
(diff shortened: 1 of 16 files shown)


Applications/Custom/mae_loss.h
```diff
@@ -14,82 +14,80 @@
...
-   * @brief initializing nntrainer
-   *
-   * @return int ML_ERROR_NONE if success
+   * @copydoc Layer::forwarding(RunLayerContext &context, bool training)
    */
-  int initialize(nntrainer::Manager &manager) override { /** NYI */
-    return 0;
-  }
+  void forwarding(nntrainer::RunLayerContext &context, bool training) override{
+    /** NYI */
```
zhoonit, line 53: FYI) Afaik, this was already implemented in main.

test/unittest/layers/meson.build
```diff
@@ -26,6 +50,8 @@ exe = executable(
   'unittest_layers', test_target,
   dependencies: [
     nntrainer_test_main_deps,
+    nntrainer_layer_common_standalone_tests_dep,
+    nntrainer_layer_common_dependent_tests_dep
```
zhoonit, line 54: How about we only have standalone test here, as this is an example for the layer developers?

zhoonit (approved):
Minor suggestion, LGTM  :)

Applications/Custom/mae_loss.h
kparichay: I will copy from main then, rather than relying on merging.

test/unittest/layers/meson.build
kparichay: you mean create two `unittest_layers` test targets?

zhoonit: Morelike not including dependent test, to think in a way this folder is in a separate repo, but it's super minor, so up to you :)

kparichay: My thought is to package nntrainer_layer_common_standalone_tests_lib with the devel_tests package, so the developer can test with it, like the application in this test. This is the unittest of the internal tests, which have extra unittests (dependent ones :))

zhoonit: Cool

jijoongmoon (approved):
LGTM

PR #1394, opened 2021-07-12
[application] SimpleShot application update for V2

kparichay:
SimpleShot application update for layer V2 design.
This includes updating the three implemented layers - centering, l2norm, and centroid_knn.
Further, task_runner has been updated to use only 1 loss kind of layer.

Unittests of layers have not been enabled, they will be enabled with unittest_layers.

### DIFF
--- a/Applications/SimpleShot/layers/centering.cpp
+++ b/Applications/SimpleShot/layers/centering.cpp
@@ -67,9 +67,6 @@ void CenteringLayer::forwarding(nntrainer::RunLayerContext &context,
   auto &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
   auto &input_ = context.getInput(SINGLE_INOUT_IDX);
 
-  std::cout << input_.getDim();
-  std::cout << hidden_.getDim();
-  std::cout << mean_feature_vector.getDim();
   input_.add(mean_feature_vector, hidden_, -1);
 }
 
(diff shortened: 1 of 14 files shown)


Applications/SimpleShot/layers/centering.cpp
```diff
@@ -69,15 +62,18 @@ void CenteringLayer::read(std::ifstream &file) {
...
-void CenteringLayer::forwarding(bool training) {
-  std::cout << net_input[0]->getVariableRef().getDim();
-  std::cout << net_hidden[0]->getVariableRef().getDim();
+void CenteringLayer::forwarding(nntrainer::RunLayerContext &context,
+                                bool training) {
+  auto &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+  auto &input_ = context.getInput(SINGLE_INOUT_IDX);
+
+  std::cout << input_.getDim();
+  std::cout << hidden_.getDim();
```
jijoongmoon, line 71: It is just for the debugging?

jijoongmoon:
I think ```Application/KNN/``` also need to update with KNN Layer as well. #1421

jijoongmoon (approved):
LGTM

Applications/SimpleShot/layers/centering.cpp
kparichay: this existed in the original code, but removed now.

kparichay:
> I think `Application/KNN/` also need to update with KNN Layer as well. #1421

Right, that should also be done. How about in main branch? Maybe this can be done after KNN layer is moved to nntrainer core?

zhoonit (approved):
LGTM

PR #1395, opened 2021-07-12
[Dataset] Add callback generator

zhoonit:
- [Dataset] Add callback generator

```
This patch add callback producer to abstract generator

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/func_data_producer.cpp
+++ b/nntrainer/dataset/func_data_producer.cpp
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   func_data_producer.cpp
+ * @date   12 July 2021
+ * @brief  This file contains various data producers from a callback
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#include <func_data_producer.h>
+
+#include <nntrainer_error.h>
+
+namespace nntrainer {
+
+FuncDataProducer::FuncDataProducer(datagen_cb datagen_cb, void *user_data_) :
+  cb(datagen_cb),
+  user_data(user_data_) {}
+
+FuncDataProducer::~FuncDataProducer() {}
+
+const std::string FuncDataProducer::getType() const {
+  return FuncDataProducer::type;
+}
+
+void FuncDataProducer::setProperty(const std::vector<std::string> &properties) {
+  NNTR_THROW_IF(!properties.empty(), std::invalid_argument)
+    << "properties is not empty, size: " << properties.size();
+}
+
+DataProducer::Gernerator
+FuncDataProducer::finalize(const std::vector<TensorDim> &input_dims,
+                           const std::vector<TensorDim> &label_dims) {
+  NNTR_THROW_IF(!this->cb, std::invalid_argument)
+    << "given callback is nullptr!";
+
+  auto input_data = std::shared_ptr<float *>(new float *[input_dims.size()],
+                                             std::default_delete<float *[]>());
+  auto label_data = std::shared_ptr<float *>(new float *[label_dims.size()],
+                                             std::default_delete<float *[]>());
+
+  return [cb = this->cb, ud = this->user_data, input_dims, label_dims,
+          input_data, label_data]() -> DataProducer::Iteration {
+    std::vector<Tensor> inputs;
+    inputs.reserve(input_dims.size());
+
+    float **input_data_raw = input_data.get();
+    float **label_data_raw = label_data.get();
+
+    for (unsigned int i = 0; i < input_dims.size(); ++i) {
+      inputs.emplace_back(input_dims[i]);
+      *(input_data_raw + i) = inputs.back().getData();
+    }
+
+    std::vector<Tensor> labels;
+    labels.reserve(label_dims.size());
+
+    for (unsigned int i = 0; i < label_dims.size(); ++i) {
+      labels.emplace_back(label_dims[i]);
+      *(label_data_raw + i) = labels.back().getData();
+    }
+
+    bool last = false;
+    int status = cb(input_data_raw, label_data_raw, &last, ud);
+    NNTR_THROW_IF(status != ML_ERROR_NONE, std::invalid_argument)
+      << "[DataProducer] Callback returned error: " << status << '\n';
+
+    if (last) {
+      return {true, {}, {}};
+    } else {
+      return {false, inputs, labels};
+    }
+  };
+}
+} // namespace nntrainer
--- a/nntrainer/dataset/func_data_producer.h
+++ b/nntrainer/dataset/func_data_producer.h
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   func_data_producer.h
+ * @date   12 July 2021
+ * @brief  This file contains various data producers from a callback
+ * @see    https://github.com/nnstreamer/nntrainer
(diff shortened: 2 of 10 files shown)


nntrainer/dataset/func_data_producer.cpp
```diff
@@ -0,0 +1,70 @@
...
+  cb(datagen_cb),
+  user_data(user_data_) {}
+
+FuncDataProducer::~FuncDataProducer() {}
+
+const std::string FuncDataProducer::getType() const {
+  return FuncDataProducer::type;
+}
+
+void FuncDataProducer::setProperty(const std::vector<std::string> &properties) {
```
kparichay, line 30: i think this is redundant.

nntrainer/dataset/func_data_producer.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+DataProducer::Gernerator
+FuncDataProducer::finalize(const std::vector<TensorDim> &input_dims,
+                           const std::vector<TensorDim> &label_dims) {
+  return [cb = this->cb, ud = this->user_data, input_dims,
+          label_dims]() -> DataProducer::Iteration {
+    std::vector<Tensor> inputs;
+    inputs.reserve(input_dims.size());
+    float **input_data = new float *[input_dims.size()];
```
kparichay, line 42: can you create this as a member of this class than being created at every call?

kparichay: also the corresponding free is missing.

nntrainer/dataset/func_data_producer.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    for (unsigned int i = 0; i < input_dims.size(); ++i) {
+      inputs.emplace_back(input_dims[i]);
+      *(input_data + i) = inputs.back().getData();
+    }
+
+    std::vector<Tensor> labels;
+    labels.reserve(label_dims.size());
+    float **label_data = new float *[labels.size()];
```
kparichay, line 51: same

nntrainer/dataset/func_data_producer.h
```diff
@@ -0,0 +1,76 @@
...
+   */
+  FuncDataProducer(datagen_cb datagen_cb, void *user_data_);
+
+  /**
+   * @brief Destroy the Func Data Producer object
+   *
+   */
+  ~FuncDataProducer();
+
+  inline static const std::string type = "batch_callback";
```
kparichay, line 48: batch_callback?

test/unittest/datasets/unittest_func_data_producers.cpp
```diff
@@ -0,0 +1,102 @@
...
+         labels[0] == expected_first_label &&
+         labels[1] == expected_second_label;
+};
+
+} // namespace
+
+std::unique_ptr<nntrainer::DataProducer>
+createConstantBatchProducer(const std::vector<std::string> &properties = {}) {
+  std::unique_ptr<nntrainer::DataProducer> ptr =
+    std::make_unique<nntrainer::FuncDataProducer>(getBatch, &user_data);
```
kparichay, line 82: can you make another unit test when you pass nullptr to this getBatch, and let it throw/fail and see our system handles it.

nntrainer/dataset/func_data_producer.cpp
zhoonit: Yes, but later this has a property so I'll leave it as it is

nntrainer/dataset/func_data_producer.cpp
zhoonit: Yes, I changed to unique_ptr and member of this lambda

nntrainer/dataset/func_data_producer.cpp
zhoonit: Done

nntrainer/dataset/func_data_producer.h
zhoonit: it is batched callback so batch callback, but as discussed earlier, this will changed to just callback

test/unittest/datasets/unittest_func_data_producers.cpp
zhoonit: Added

nntrainer/dataset/func_data_producer.h
kparichay: alright.

kparichay (approved):
LGTM

nntrainer/dataset/func_data_producer.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+FuncDataProducer::FuncDataProducer(datagen_cb datagen_cb, void *user_data_) :
+  cb(datagen_cb),
+  user_data(user_data_) {}
+
+FuncDataProducer::~FuncDataProducer() {}
+
+const std::string FuncDataProducer::getType() const {
+  return FuncDataProducer::type;
+}
```
jijoongmoon, line 28: Could you list up how many FuncDataProducer Types are?

nntrainer/dataset/func_data_producer.cpp
```diff
@@ -0,0 +1,79 @@
...
+FuncDataProducer::finalize(const std::vector<TensorDim> &input_dims,
+                           const std::vector<TensorDim> &label_dims) {
+  NNTR_THROW_IF(!this->cb, std::invalid_argument)
+    << "given callback is nullptr!";
+
+  auto input_data = std::shared_ptr<float *>(new float *[input_dims.size()],
+                                             std::default_delete<float *[]>());
+  auto label_data = std::shared_ptr<float *>(new float *[label_dims.size()],
+                                             std::default_delete<float *[]>());
+
```
jijoongmoon, line 45: It looks like we give mem buffer which we allocate. (correct me if I'm wrong) then.. It will be deleted when we destruct dataset. I guess how about the opposite? user gives buffer and we just are using it.

nntrainer/dataset/func_data_producer.cpp
```diff
@@ -0,0 +1,79 @@
...
+      *(input_data_raw + i) = inputs.back().getData();
+    }
+
+    std::vector<Tensor> labels;
+    labels.reserve(label_dims.size());
+
+    for (unsigned int i = 0; i < label_dims.size(); ++i) {
+      labels.emplace_back(label_dims[i]);
+      *(label_data_raw + i) = labels.back().getData();
+    }
```
jijoongmoon, line 65: Correct me if i'm wrong. It looks like we allocate tensor as required and user set the data in cb ( as in getBatch() in unit test case). How about opposite ? I mean user allocate the buffer and we just use it.. Cause if we destruct the dataset.. then the buffer will also removed. then there might be high chance to loose user data or there might be needed another copy in callback function.

zhoonit: It's absolutely right that we are allocating memory for the user. 
Thank you for the suggestion :)

I think there are some issues if we are to pass the ownership.

1. Semantic deviation/break
Currently(tizen 6.0+), it is allocated by us if we change the behavior, it breaks semantics.
Other producers will own it's buffer, then we have to discriminate callback producer to not delete the memory.

1. Memory ownership issue
if the user tries to reuse the memory it will likely to be invalidated (due to multiple workers on this callback at the same time or in-place call from model). if they allocate memory for each call, we have to ask the user to have a threadsafe vector to save the allocated pointer and free after `ml_train_model_destruct()`. I think this restrictions are hard to be kept by the users.

1.  Losing opportunity to optimize
It is allocated from the dataproducer. later it will be from databuffer. We will at least reuse the allocated memory later, we will let it be controlled by manager in the future)

nntrainer/dataset/func_data_producer.cpp
zhoonit: There is one type for now, later we can add :)

nntrainer/dataset/func_data_producer.cpp
jijoongmoon: > It is allocated from the dataproducer. later it will be from databuffer. We will at least reuse the allocated memory later, we will let it be controlled by the manager in the future)

so, the user copys the data to the buffer we allocate and we are not going to copy to queue. then.

Thanks for your explanation.

jijoongmoon (approved):
LGTM

PR #1396, opened 2021-07-12
[dataset] Add file producer

zhoonit:
- [dataset] Add file producer

```
This patch adds file producer which abstracts reading a raw file. Also,
this component makes sure that the file accepts various kind of input
shapes.

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/raw_file_data_producer.cpp
+++ b/nntrainer/dataset/raw_file_data_producer.cpp
@@ -1,0 +1,155 @@
+
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   raw_file_data_producer.cpp
+ * @date   12 July 2021
+ * @brief  This file contains raw file data producers, reading from a file
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#include <raw_file_data_producer.h>
+
+#include <memory>
+#include <numeric>
+#include <random>
+#include <vector>
+
+#include <common_properties.h>
+#include <nntrainer_error.h>
+#include <node_exporter.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+RawFileDataProducer::RawFileDataProducer() : raw_file_props(new PropTypes()) {}
+RawFileDataProducer::~RawFileDataProducer() {}
+
+const std::string RawFileDataProducer::getType() const {
+  return RawFileDataProducer::type;
+}
+
+unsigned long long
+RawFileDataProducer::size(const std::vector<TensorDim> &input_dims,
+                          const std::vector<TensorDim> &label_dims) const {
+  auto size_accumulator = [](const unsigned int &a, const TensorDim &b) {
+    return a + b.getFeatureLen();
+  };
+
+  auto sample_size =
+    std::accumulate(input_dims.begin(), input_dims.end(), 0u, size_accumulator);
+  sample_size = std::accumulate(label_dims.begin(), label_dims.end(),
+                                sample_size, size_accumulator);
+
+  auto path_prop = std::get<props::FilePath>(*raw_file_props);
+  auto file_size = path_prop.file_size();
+
+  /// checking alignment is a good way to make check if a file is valid,
+  /// unfortunately, our test dataset does not have this property
+  /// (trainingSet.dat, valSet.dat, testSet.dat) after checking, we can
+  /// uncomment below line.
+  // NNTR_THROW_IF((file_size % sample_size * RawFileDataProducer::pixel_size !=
+  // 0),
+  //               std::invalid_argument)
+  //   << " Given file does not align with the given sample size, sample size: "
+  //   << sample_size << " file_size: " << file_size;
+
+  return path_prop.file_size() /
+         (sample_size * RawFileDataProducer::pixel_size);
+}
+
+void RawFileDataProducer::setProperty(
+  const std::vector<std::string> &properties) {
+  auto left = loadProperties(properties, *raw_file_props);
+  NNTR_THROW_IF(!left.empty(), std::invalid_argument)
+    << "There is unparsed properties, size: " << left.size();
+}
+
+DataProducer::Generator
+RawFileDataProducer::finalize(const std::vector<TensorDim> &input_dims,
+                              const std::vector<TensorDim> &label_dims) {
+
+  /****************** Validation ****************/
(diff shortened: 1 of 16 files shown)


nntrainer/dataset/raw_file_data_producer.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <util_func.h>
+
+namespace nntrainer {
+
+/**
+ * @brief Props containing file path value
+ *
+ */
+class PropsPath : public Property<std::string> {
+public:
```
jijoongmoon, line 35: Isn't it better to collect all the property in one location? like in common_property?

nntrainer/dataset/raw_file_data_producer.cpp
```diff
@@ -0,0 +1,180 @@
...
+RawFileDataProducer::finalize(const std::vector<TensorDim> &input_dims,
+                              const std::vector<TensorDim> &label_dims) {
+
+  /****************** Validation ****************/
+  auto sz = size(input_dims, label_dims);
+  auto batch = input_dims[0].batch();
+
+  NNTR_THROW_IF(sz < batch, std::invalid_argument)
+    << "calculated sample size is less than a batch";
+
```
jijoongmoon, line 109: I'm not sure which one is better. but how about change the batch size with warrning, instead of giving error.

nntrainer/dataset/raw_file_data_producer.cpp
```diff
@@ -0,0 +1,180 @@
...
+ *
+ */
+class PropsPath : public Property<std::string> {
+public:
+  static constexpr const char *key = "path"; /**< unique key to access */
+  using prop_tag = str_prop_tag;             /**< property type */
+
+  bool isValid(const std::string &v) {
+    std::ifstream file(v, std::ios::binary | std::ios::ate);
+    return file.good();
```
jijoongmoon, line 41: AFIK, there will be a specific directory name coming up after this file dataset. In that case, what the type name might be if it is path? how about filepath, dirpath for making it clearer?

jijoongmoon (approved):
LGTM except for small suggestions.

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: Okay :) I'll make this to public.

zhoonit: > AFIK, there will be a specific directory name coming up after this file dataset. In that case, what the type name might be if it is path? how about filepath, dirpath for making it clearer?

One concern here is that, other's using this property need different validation logic. In this case it might be better to put it here, I think we can consider this later so I'll put this inside common properties :)

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: I'll change prop name to FilePath :)

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: As per the offline discussion two weeks ago, this will change to a element-wise generator so this validation will go away, I'll add todo here and apply this on the way to the migration :)

nntrainer/dataset/raw_file_data_producer.cpp
```diff
@@ -0,0 +1,180 @@
...
+RawFileDataProducer::~RawFileDataProducer() {}
+
+const std::string RawFileDataProducer::getType() const {
+  return RawFileDataProducer::type;
+}
+
+unsigned long long
+RawFileDataProducer::size(const std::vector<TensorDim> &input_dims,
+                          const std::vector<TensorDim> &label_dims) const {
+  auto size_accumulator = [](const unsigned int &a, const TensorDim &b) {
```
kparichay, line 66: const unsigned int &a -> const unsigned long long &a

nntrainer/dataset/raw_file_data_producer.cpp
```diff
@@ -0,0 +1,180 @@
...
+  /// unfortunately, our test dataset does not have this property
+  /// (trainingSet.dat, valSet.dat, testSet.dat) after checking, we can
+  /// uncomment below line.
+  // NNTR_THROW_IF((file_size % sample_size * RawFileDataProducer::pixel_size !=
+  // 0),
+  //               std::invalid_argument)
+  //   << " Given file does not align with the given sample size, sample size: "
+  //   << sample_size << " file_size: " << file_size;
+
+  return path_prop.file_size() /
```
kparichay, line 88: return file_size / ...

nntrainer/dataset/raw_file_data_producer.cpp
```diff
@@ -0,0 +1,180 @@
...
+  auto idxes_ = std::vector<unsigned int>();
+  idxes_.reserve(sz);
+  /// idxes point to the file position in bytes where a sample starts
+  std::generate_n(std::back_inserter(idxes_), sz,
+                  [sample_size, current = 0ULL]() mutable {
+                    auto c = current;
+                    current += sample_size * RawFileDataProducer::pixel_size;
+                    return c;
+                  });
+  std::shuffle(idxes_.begin(), idxes_.end(), rng_);
```
kparichay, line 137: I recommend to remove shuffle from here. Shuffling will be done from the external caller.
If you want to keep it for later remembrance, then consider adding as a comment.

kparichay (approved):

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: why, wonder how it worked.

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: Yes, I'll add a comment here

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: There is a file size right below. clang-format doesn't allow me to hoist the below line :p

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: Done

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: Done

nntrainer/dataset/raw_file_data_producer.cpp
zhoonit: Done

nntrainer/dataset/raw_file_data_producer.cpp
kparichay: ok.

PR #1403, opened 2021-07-13
[layer] Enable time dist layer for V2

kparichay:
Enable time dist layer for LayerV2 design.
This patch tries to simulate the InitContext and the RunContext inside
the time dist layer so that proper shapes of variables can be passed to
the internal layer. Further, context changing function calls are
replicated on the actual InitContext/RunContext by detecting changes and
making those function calls again.

LayerNode was updated to ensure that a layer is not getting distributed
again and again.
Some more getter APIs were added for TensorDim and LayerContext.

Unittests related to distribute have also been enabled with this patch.

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -104,16 +104,19 @@ LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l, size_t idx) :
 
 int LayerNode::setProperty(std::vector<std::string> properties) {
   int status = ML_ERROR_NONE;
+  bool already_distributed =
+    !std::get<props::Distribute>(*layer_node_props).empty() &&
+    std::get<props::Distribute>(*layer_node_props).get();
   auto left_properties = loadProperties(properties, *layer_node_props);
 
   /// note that setting distribute is only allowed for one time.
   /// until we have layerNode::finalize and must not except timedist layer
-  if (getDistribute()) {
-    // auto &ac = nntrainer::AppContext::Global();
-    // std::unique_ptr<nntrainer::Layer> dlayer =
-    //   ac.createObject<nntrainer::Layer>(TimeDistLayer::type);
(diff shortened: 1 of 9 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -132,19 +109,11 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
...
-      std::shared_ptr<nntrainer::LayerV1> dlayer =
-        ac.createObject<nntrainer::LayerV1>(TimeDistLayer::type);
-      std::static_pointer_cast<TimeDistLayer>(dlayer)->setDistLayer(layerv1);
-      layerv1 = dlayer;
-    }
+    // auto &ac = nntrainer::AppContext::Global();
+    // std::unique_ptr<nntrainer::Layer> dlayer =
+    //   ac.createObject<nntrainer::Layer>(TimeDistLayer::type);
+    // dynamic_cast<TimeDistLayer*>(dlayer.get())->setDistLayer(std::move(layer));
+    // layer = std::move(dlayer);
```
jijoongmoon: is it intended?

nntrainer/layers/layer_node.cpp
```diff
@@ -104,16 +104,19 @@ LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l, size_t idx) :
 
 int LayerNode::setProperty(std::vector<std::string> properties) {
   int status = ML_ERROR_NONE;
+  bool already_distributed =
+    !std::get<props::Distribute>(*layer_node_props).empty() &&
+    std::get<props::Distribute>(*layer_node_props).get();
```
jijoongmoon, line 109: I think those are same. if it is empty, get() return false..

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.cpp
kparichay: if the property is empty(), get() throws. Hence, the check of empty() before get().

zhoonit (approved):
LGTM

PR #1405, opened 2021-07-13
[Dataset] Add thread safe queue for the buffer

zhoonit:
- [Dataset] Add thread safe queue for the buffer

```
Currently, databuffer is not queuing any iteration but requesting buffer
on-demand basis.
This patch adds thread safe queue for the buffer to queue multiple
buffers

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/batch_queue.cpp
+++ b/nntrainer/dataset/batch_queue.cpp
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+ *
+ * @file   batch_queue.cpp
+ * @date   13 July 2021
+ * @brief  This file contains thread safe queue
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+#include <batch_queue.h>
+#include <chrono>
+
+#include <mutex>
+#include <nntrainer_error.h>
+#include <shared_mutex>
+
+using namespace std::literals::chrono_literals;
+
+namespace nntrainer {
+
+BatchQueue::BatchQueue(unsigned int queue_capacity_) :
+  queue_capacity(queue_capacity_) {
+  NNTR_THROW_IF(queue_capacity == 0, std::invalid_argument)
+    << "queue capacity of zero not supported!";
+}
+
+BatchQueue::BatchQueue(const BatchQueue &rhs) :
+  queue_capacity(rhs.queue_capacity) {}
+
+BatchQueue &BatchQueue::operator=(const BatchQueue &rhs) {
+  if (this == &rhs) {
+    return *this;
+  }
+  this->queue_capacity = rhs.queue_capacity;
+  return *this;
+}
+
+void BatchQueue::wait_and_push(T &&data) noexcept {
+  std::unique_lock<std::shared_mutex> lk(q_mutex);
+  q_writer_cv.wait(lk, [this] { return q.size() != queue_capacity; });
+  q.push(std::make_unique<T>(data));
+  q_reader_cv.notify_one();
+}
+
+std::unique_ptr<BatchQueue::T> BatchQueue::wait_and_pop() noexcept {
+  std::unique_lock<std::shared_mutex> lk(q_mutex);
+  q_reader_cv.wait(lk, [this] { return !q.empty(); });
+
+  /// @note this invalidates q.front(), but it is okay because it is locked and
+  /// popped right away
+  auto ptr = std::move(q.front());
+  q.pop();
+  q_writer_cv.notify_one();
+
+  return ptr;
+}
+
+bool BatchQueue::isFull() const {
+  std::shared_lock<std::shared_mutex> lk(q_mutex);
+  return queue_capacity == q.size();
+}
+
+bool BatchQueue::isEmpty() const {
+  std::shared_lock<std::shared_mutex> lk(q_mutex);
+  return q.empty();
+}
+
+} // namespace nntrainer
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -125,6 +125,7 @@ include $(CLEAR_VARS)
 NNTRAINER_SRCS := $(NNTRAINER_ROOT)/nntrainer/models/neuralnet.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/models/model_loader.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/models/dynamic_training_optimization.cpp \
+                  $(NNTRAINER_ROOT)/nntrainer/dataset/batch_queue.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer.cpp \
                   $(NNTRAINER_ROOT)/nntrainer/dataset/databuffer_factory.cpp \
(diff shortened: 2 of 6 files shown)


nntrainer/dataset/batch_queue.cpp
```diff
@@ -0,0 +1,68 @@
...
+ */
+#include <batch_queue.h>
+#include <chrono>
+
+#include <nntrainer_error.h>
+
+using namespace std::literals::chrono_literals;
+
+namespace nntrainer {
+BatchQueue::BatchQueue(unsigned int queue_capacity_) :
```
kparichay, line 21: Please ensure to take queue_capacity as the number of elements (not number of batches) as queue capacity controls the amount of memory consumed by the data prefetching. This should not change when batchsize is increased or decreased.

nntrainer/dataset/batch_queue.cpp
```diff
@@ -0,0 +1,68 @@
...
+  /// popped right away
+  auto ptr = std::move(q.front());
+  q.pop();
+  q_writer_cv.notify_one();
+
+  return ptr;
+}
+
+bool BatchQueue::isFull() const {
+  std::lock_guard<std::mutex> lk(q_mutex);
```
kparichay, line 59: I think this can be shared_lock (read lock)

nntrainer/dataset/batch_queue.cpp
```diff
@@ -0,0 +1,68 @@
...
+  return queue_capacity == q.size();
+}
+
+bool BatchQueue::isEmpty() const {
+  std::lock_guard<std::mutex> lk(q_mutex);
```
kparichay, line 64: same read lock.

nntrainer/dataset/batch_queue.cpp
```diff
@@ -0,0 +1,68 @@
...
+BatchQueue &BatchQueue::operator=(const BatchQueue &rhs) {
+  if (this == &rhs) {
+    return *this;
+  }
+  this->queue_capacity = rhs.queue_capacity;
+  return *this;
+}
+
+void BatchQueue::wait_and_push(T &&data) noexcept {
+  std::unique_lock<std::mutex> lk(q_mutex);
```
kparichay, line 39: as BatchQueue will be shared by the model and the dataset, I am guessing model will start dataset in a separate thread in this case?

nntrainer/dataset/batch_queue.cpp
zhoonit: It will be amount of iterations batch queue will have, if we decide to change element, it will be number of elements. data buffer will set number `buffersize / batchsize`. batch queue is constructed inside a function that knows the immutable batch size, so.. this won't change in the point of batch queue.

nntrainer/dataset/batch_queue.cpp
zhoonit: yes, this can be shared lock, I'll change :)

nntrainer/dataset/batch_queue.cpp
zhoonit: yes, I'll change!

nntrainer/dataset/batch_queue.cpp
zhoonit: actually it is abstracted inside the databuffer, so model won't know about the existence of a queue.

please refer to the below code from #1416
https://github.com/nnstreamer/nntrainer/blob/71c7d99f6c533ae28e87513d6e81593d73ae4a8b/nntrainer/models/neuralnet.cpp#L687-L707

kparichay: Great, thanks for the explanation.

nntrainer/dataset/batch_queue.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+std::unique_ptr<BatchQueue::T> BatchQueue::wait_and_pop() noexcept {
+  std::unique_lock<std::mutex> lk(q_mutex);
+  q_reader_cv.wait(lk, [this] { return !q.empty(); });
+
+  /// @note this invalidates q.front(), but it is okay because it is locked and
+  /// popped right away
+  auto ptr = std::move(q.front());
+  q.pop();
+  q_writer_cv.notify_one();
```
kparichay, line 53: It might be good to update this to notify_all() for later support for multiple writers? or add a note/todo.

kparichay (approved):
LGTM

nntrainer/dataset/batch_queue.cpp
zhoonit: sure :)

nntrainer/dataset/batch_queue.cpp
zhoonit: for the mutiple writer, it will be notify_one, because we don't want to wake up every writer when buffer has a room :p

kparichay: I recommend notify_all, as with multiple writers, notify_one might not do anything if all are awake, and lead to some workers sleeping.

zhoonit: if all are awake, i think it's intended to be noop? I feel like I am missing something, I'll ask offline :)

nntrainer/dataset/batch_queue.cpp
zhoonit: updated

nntrainer/dataset/batch_queue.cpp
zhoonit: updated :)

nntrainer/dataset/batch_queue.cpp
jijoongmoon: There will be the default value for queue_capacity and it can be defined by user. I wonder why do it needs to be <= buffersize? Of course, it will waste the memory but, I guess there is no need to constrain it. isn't it?

nntrainer/dataset/batch_queue.cpp
```diff
@@ -0,0 +1,71 @@
...
+  std::unique_lock<std::shared_mutex> lk(q_mutex);
+  q_writer_cv.wait(lk, [this] { return q.size() != queue_capacity; });
+  q.push(std::make_unique<T>(data));
+  q_reader_cv.notify_one();
+}
+
+std::unique_ptr<BatchQueue::T> BatchQueue::wait_and_pop() noexcept {
+  std::unique_lock<std::shared_mutex> lk(q_mutex);
+  q_reader_cv.wait(lk, [this] { return !q.empty(); });
+
```
jijoongmoon, line 51: I think just checking if the queue is empyt is not enought. We need to check the number of remaining queue size here. If it is less than batch size, we need to wait longer.

jijoongmoon (approved):
LGTM except small comments.

nntrainer/dataset/batch_queue.cpp
zhoonit: We discussed offline to add a todo here.

Additional requirements

1. maximize thread utilization
2. resource allocation should be done in parallel
3. minimize overhead

nntrainer/dataset/batch_queue.cpp
zhoonit: Yes, there is no constraint from the queue. It is from the who is using the queue. But the queue capacity is not equal to buffer size I think it's worth noting it here :)

nntrainer/dataset/batch_queue.cpp
zhoonit: one data point here is a single batch so that, if q is not empty, it means we have a batch fully ready :)

jijoongmoon: I see. Thanks.

PR #1407, opened 2021-07-13
[Dataset] Implement data getter

zhoonit:
- [Dataset] Implement data getter

```
This patch implements data getter.
There are two types of data getter.

1. queue multiple iterations and get one by one
(`DataBuffer::startFetchWorker()`, `DataBuffer::fetch()`)

2. generate one iteration on the fly (`DataBuffer::batcher()`)

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/databuffer.h
+++ b/nntrainer/dataset/databuffer.h
@@ -33,6 +33,7 @@
 #include <thread>
 #include <vector>
 
+#include <batch_queue.h>
 #include <data_producers.h>
 #include <dataset.h>
 #include <tensor_dim.h>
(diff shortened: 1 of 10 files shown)


nntrainer/dataset/databuffer.h
```diff
@@ -78,11 +79,49 @@ class DataBuffer : public ml::train::Dataset {
...
+  /**
+   * @brief Get the Generator object
+   * @remark the batch dimension of input_dims / label_dims must be same for
+   * all.
+   *
+   * @param input_dims dimension of input_dims
+   * @param label_dims dimension of label_dims
+   * @return DataProducer::Generator which generates an iteration
+   */
+  DataProducer::Generator batcher(const std::vector<TensorDim> &input_dims,
```
kparichay, line 117: can you please elaborate on this? the name doesn't match with the brief.
Also, startFetchWorker() wraps around this, so maybe this is intended for private?

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -79,7 +79,7 @@ const std::string RandomDataOneHotProducer::getType() const {
 unsigned long long
 RandomDataOneHotProducer::size(const std::vector<TensorDim> &input_dims,
```
kparichay, line 80: how about update this function as well?

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -79,7 +79,7 @@ const std::string RandomDataOneHotProducer::getType() const {
 unsigned long long
```
kparichay, line 79: update datatype to unsigned int or PropNumSamples data type to ull.

nntrainer/dataset/databuffer.h
zhoonit: This is sync method for the databuffer so intended to be a public method.

Later, if we ever going to support `model->forwarding` or equivalent, this will be needed.

Please refer to below, meanwhile I'll elaborate more :)

https://github.com/nnstreamer/nntrainer/blob/1da132aa82e08f5a3c49177ca42d6e4eb656747c/test/unittest/datasets/unittest_databuffer.cpp#L19-L34

nntrainer/dataset/random_data_producers.cpp
zhoonit: I'll change this to ull 👍

nntrainer/dataset/random_data_producers.cpp
zhoonit: I'll change this

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -54,17 +54,17 @@ class PropsMax : public Property<float> {
  * @brief Props containing number of samples
  *
  */
-class PropsDataSize : public Property<unsigned int> {
+class PropsNumSamples : public Property<unsigned int> {
```
jijoongmoon, line 57: Could you elaborate this property? What do you mean by the number of samples?

zhoonit: A random data producer has theoretical size. number of samples is used to set theoretical size of the random data producer's data size, so that it can end in indefinite amount of iteration.

Updated comment :)

jijoongmoon: Thanks.

jijoongmoon (approved):
LGTM

nntrainer/dataset/databuffer.h
kparichay: Then you might consider renaming to something like `getGenerator()` etc.

zhoonit: I'll rename it to somewhat similar to the suggestion on the future work :)

nntrainer/dataset/random_data_producers.cpp
zhoonit: Done

nntrainer/dataset/random_data_producers.cpp
zhoonit: Done

kparichay (approved):
LGTM

PR #1408, opened 2021-07-14
[layer] Update tflite included headers

kparichay:
Update tflite included headers for tensorflow2.
tensorflow/contrib signature headers work too but tensorflow2 signature
is better.


kparichay:
Closing this as tensorflow2 prefix is not done for android build.

PR #1409, opened 2021-07-14
[Dataset] Change factory signature

zhoonit:
- [Dataset] Change factory signature

```
Add properties to factory, given contructor argument, it is delegated to
the constructor

**Self evaluation:**

Cc: Parichay Kapoor<pk.kapoor@samsung.com>

```

### DIFF
--- a/api/ccapi/include/dataset.h
+++ b/api/ccapi/include/dataset.h
@@ -1,4 +1,3 @@
-
 // SPDX-License-Identifier: Apache-2.0
 /**
  * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
@@ -87,23 +86,40 @@ public:
 };
 
 /**
- * @brief Factory creator with constructor for dataset
+ * @brief Create a Dataset object with given arguements
+ *
+ * @param type dataset type
+ * @param properties property representations
+ * @return std::unique_ptr<Dataset> created dataset
  */
 std::unique_ptr<Dataset>
 createDataset(DatasetType type,
               const std::vector<std::string> &properties = {});
 
 /**
- * @brief Factory creator with constructor for dataset
+ * @brief Create a Dataset object
+ *
+ * @param type dataset type
+ * @param path path to a file or folder
+ * @param properties property representations
+ * @return std::unique_ptr<Dataset> created dataset
  */
-std::unique_ptr<Dataset> createDataset(DatasetType type, const char *file);
+std::unique_ptr<Dataset>
+createDataset(DatasetType type, const char *path,
+              const std::vector<std::string> &properties = {});
 
 /**
- * @brief Factory creator with constructor for dataset
+ * @brief Create a Dataset object
+ *
+ * @param type dataset type
+ * @param cb callback
+ * @param user_data user data
+ * @param properties property representations
+ * @return std::unique_ptr<Dataset> created dataset
  */
-std::unique_ptr<Dataset> createDataset(DatasetType type, datagen_cb cb,
-                                       void *user_data = nullptr);
-
+std::unique_ptr<Dataset>
+createDataset(DatasetType type, datagen_cb cb, void *user_data = nullptr,
+              const std::vector<std::string> &properties = {});
 } // namespace train
 } // namespace ml
 
--- a/nntrainer/dataset/databuffer_factory.cpp
+++ b/nntrainer/dataset/databuffer_factory.cpp
@@ -23,14 +23,12 @@ namespace nntrainer {
  */
 std::unique_ptr<DataBuffer> createDataBuffer(DatasetType type) {
   switch (type) {
-  case DatasetType::GENERATOR:
-    return std::make_unique<DataBufferFromCallback>();
   case DatasetType::FILE:
     return std::make_unique<DataBufferFromDataFile>();
   case DatasetType::UNKNOWN:
-    /** fallthrough intended */
+    [[fallthrough]];
   default:
-    throw std::invalid_argument("Unknown type for the dataset");
+    throw std::invalid_argument("Unsupported constructor type for the dataset");
   }
 }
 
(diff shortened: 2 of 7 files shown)


api/ccapi/include/dataset.h
```diff
@@ -87,23 +86,28 @@ class Dataset {
 };
 
 /**
- * @brief Factory creator with constructor for dataset
+ * @copydoc std::unique_ptr<Dataset>createDataset(DatasetType, const
```
kparichay, line 89: Lets have description in the API header, and you can use copydoc in the internal headers.

kparichay: Same for other functions in this file.

nntrainer/dataset/databuffer_factory.cpp
```diff
@@ -23,14 +23,12 @@ namespace nntrainer {
  */
 std::unique_ptr<DataBuffer> createDataBuffer(DatasetType type) {
   switch (type) {
-  case DatasetType::GENERATOR:
-    return std::make_unique<DataBufferFromCallback>();
   case DatasetType::FILE:
```
kparichay, line 26: so, the default type is file?

nntrainer/dataset/databuffer_file.h
```diff
@@ -46,6 +47,14 @@ class DataBufferFromDataFile final : public DataBuffer {
    */
   DataBufferFromDataFile() : DataBuffer(DatasetType::FILE) {}
 
+  /**
+   * @brief     Constructor
+   */
+  DataBufferFromDataFile(const std::string &path) : DataBufferFromDataFile() {
```
kparichay, line 53: the argument `path` is ignored?

nntrainer/models/model_loader.cpp
```diff
@@ -219,12 +212,17 @@ int ModelLoader::loadDatasetConfigIni(dictionary *ini, NeuralNetwork &model) {
...
+    try {
+      model.data_buffers[static_cast<int>(dt)] =
+        createDataBuffer(DatasetType::FILE, resolvePath(path).c_str());
+    } catch (...) {
+      ml_loge("path is not valid, path: %s", resolvePath(path).c_str());
+      return ML_ERROR_INVALID_PARAMETER;
+    }
 
-    dbuffer->setProperty({bufsizepros});
+    model.data_buffers[static_cast<int>(dt)]->setProperty({bufsizepros});
```
kparichay, line 223: move this inside try as well.

test/ccapi/unittest_ccapi.cpp
```diff
@@ -148,9 +148,9 @@ TEST(ccapi_dataset, construct_01_n) {
 /**
  * @brief Neural Network Dataset Contruct Test
  */
-TEST(ccapi_dataset, construct_02_p) {
-  EXPECT_NO_THROW(ml::train::createDataset(ml::train::DatasetType::GENERATOR));
-  EXPECT_NO_THROW(ml::train::createDataset(ml::train::DatasetType::FILE));
+TEST(ccapi_dataset, construct_02_n) {
+  EXPECT_THROW(ml::train::createDataset(ml::train::DatasetType::GENERATOR),
```
kparichay, line 152: Add the corresponding test for DatasetType::FILE?

api/ccapi/include/dataset.h
zhoonit: Updated :)

nntrainer/dataset/databuffer_factory.cpp
zhoonit: Generator is forced to have a callback argument, so it is not supported when only type is given

nntrainer/dataset/databuffer_file.h
zhoonit: This is intended. this will be removed so, no need to set path

nntrainer/models/model_loader.cpp
zhoonit: Right, updated

test/ccapi/unittest_ccapi.cpp
zhoonit: It is valid to create without argument for FILE so there is no such like this case.

nntrainer/dataset/databuffer_factory.cpp
kparichay: got it.

nntrainer/dataset/databuffer_file.h
kparichay: so the argument will be removed, or this struct?

test/ccapi/unittest_ccapi.cpp
kparichay: so the check of having file is added in the finalize of the Dataset?

nntrainer/dataset/databuffer_factory.cpp
kparichay: > Generator is forced to have a callback argument, so it is not supported when only type is given

It might be better to keep the same constraints for both cb/file datasets for consistency.

zhoonit: Yeah, I think we could decide this later, I'd like to keep the structure for reference for now :)

nntrainer/dataset/databuffer_file.h
zhoonit: This struct will be removed

test/ccapi/unittest_ccapi.cpp
zhoonit: Correct 👍

api/ccapi/include/dataset.h
```diff
@@ -87,23 +86,40 @@ class Dataset {
...
+ *
+ * @param type dataset type
+ * @param path path to a file or folder
+ * @param properties property representations
+ * @return std::unique_ptr<Dataset> created dataset
  */
-std::unique_ptr<Dataset> createDataset(DatasetType type, const char *file);
+std::unique_ptr<Dataset>
+createDataset(DatasetType type, const char *path,
+              const std::vector<std::string> &properties = {});
```
jijoongmoon, line 109: I wonder if the path can be properties.

zhoonit: Path is in supported as props. This is to support constructing with char * argument. 
While this can be fully supported via path properties, here it is added for completeness (because callback producer must have to have it's property and later others might too).

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1410, opened 2021-07-14
[layer] LayerV1 cleanup from LayerNode

kparichay:
Remove LayerNode dependence on LayerV1.
Further, NetworkGraph dependence on LayerV1 is also removed.

Some More bugfixes in the patch:
- Add initialization for the missing variables for LayerImpl class
- Add missing initialization in constructor for the weight class

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -81,19 +81,6 @@ createLayerNode(const std::string &type,
  * @brief Layer factory creator with constructor
  */
 std::unique_ptr<LayerNode>
-createLayerNode(std::shared_ptr<nntrainer::LayerV1> layer,
-                const std::vector<std::string> &properties) {
-  auto lnode = std::make_unique<LayerNode>(layer);
-  if (lnode->setProperty(properties) != ML_ERROR_NONE)
-    throw std::invalid_argument("Error setting layer properties.");
-
-  return lnode;
-}
-
-/**
- * @brief Layer factory creator with constructor
- */
-std::unique_ptr<LayerNode>
 createLayerNode(std::unique_ptr<nntrainer::Layer> &&layer,
                 const std::vector<std::string> &properties) {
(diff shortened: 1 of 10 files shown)


jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.h
```diff
@@ -739,15 +625,15 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    * @details this is layer inside the distribution layer if this layer node
    * is distributed.
    */
-  const std::shared_ptr<nntrainer::LayerV1> &getLayer() const;
+  const nntrainer::Layer *getLayer() const;
```
lhs8928, line 628: Is there any reason to use pointer instead of shared_pointer?

nntrainer/layers/layer_node.cpp
```diff
@@ -256,11 +220,12 @@ std::ostream &operator<<(std::ostream &out, const LayerNode &l) {
 }
 
 std::string LayerNode::getDistLayerType() const {
-  if (getDistribute())
-    return std::static_pointer_cast<TimeDistLayer>(layerv1)->getDistLayerType();
-  else
-    throw std::runtime_error(
-      "Get distribution layer type for non-distributed layer");
+  // if (getDistribute())
```
lhs8928, line 223: Intended comment?

lhs8928 (approved):
LGTM

nntrainer/layers/layer_node.h
kparichay: internal layer has not become a unique_ptr<>, however, unique_ptr<> cannot be returned.
for unique_ptr, in order to return the object to non-owners, the underlying ptr is returned.

nntrainer/layers/layer_node.cpp
kparichay: Yes, this is intended. This will be enabled + updated in the next patch where the distribution layer is enabled for V2 design.

PR #1418, opened 2021-07-15
[LayerV1] Delete LayerV1

kparichay:
This patch deletes LayerV1 headers and its implementations.
Some of the relevant codes from LayerV1 are moved to either LayerNode or
to LayerDevel.


myungjoo:
Great! Finally, you can merge the branches!

nntrainer/layers/layer_devel.h
```diff
@@ -50,6 +50,85 @@ enum class ExportMethods;
 class Layer {
 
 public:
+  /**
+   * @brief     Property Enumeration
+   *            0. input shape : string
+   *            1. bias zero : bool
```
lhs8928, line 56: from 1 to 9 documentation doesn't match. (bias zero should be move to 9)

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM. Now we upgrade Layer Implementation. We need to check once again the unit test and applications once again including accuracy, even if the unit test passed by ci but we need to check one by one for sure.

kparichay:
> LGTM. Now we upgrade Layer Implementation. We need to check once again the unit test and applications once again including accuracy. Even if the unit test passed by ci automatically, but we need to check one by one for sure.

Agreed, I will run a manual check for all the applications, before making the layer_v2 -> main branch merge.

nntrainer/layers/layer_devel.h
kparichay: updated.

kparichay: good catch.

kparichay:
> LGTM. Now we upgrade Layer Implementation. We need to check once again the unit test and applications once again including accuracy. Even if the unit test passed by ci automatically, but we need to check one by one for sure.

This will be done in patch #1426.

PR #1420, opened 2021-07-15
[Dataset] Add user_data set Property handler

zhoonit:
- [Dataset] Add user_data set Property handler

```
- PtrType properties
- Add user_data set property handler for backward competibility

**Self evaluation:**

```

### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -868,6 +868,7 @@ static int
 ml_train_dataset_set_property_for_mode_(ml_train_dataset_h dataset,
                                         ml_train_dataset_mode_e mode,
                                         const std::vector<void *> &args) {
+  static constexpr char USER_DATA[] = "user_data";
   int status = ML_ERROR_NONE;
   ml_train_dataset *nndataset;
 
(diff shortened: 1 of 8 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -888,7 +888,41 @@ ml_train_dataset_set_property_for_mode_(ml_train_dataset_h dataset,
...
-      status_ = db->setProperty(args);
+      std::vector<std::string> properties;
+      for (unsigned int i = 0; i < args.size(); ++i) {
+        char *key_ptr = (char *)args[i];
+        std::string key = key_ptr;
+        std::string value;
+
+        /** Handle the user_data as a special case, serialize the address and
+         * pass it to the databuffer */
+        if (key == "user_data") {
```
kparichay, line 899: PropsUserData::key?

kparichay: also istrequal()?

api/capi/src/nntrainer.cpp
```diff
@@ -888,7 +888,41 @@ ml_train_dataset_set_property_for_mode_(ml_train_dataset_h dataset,
...
+            return status_;
+          }
+          std::ostringstream ss;
+          ss << key << '=' << args[i + 1];
+          properties.push_back(ss.str());
+
+          /** As values of i+1 is consumed, increase i by 1 */
+          i++;
+
+        } else if (key.find("user_data") == 0) {
```
kparichay, line 914: this adds a limitation for the future properties which can include the term "user_data" in them.
recommend to rather use `key.startsWith("user_data=")` after removing all white spaces from the key.

nntrainer/utils/base_properties.h
```diff
@@ -316,6 +322,39 @@ template <typename Tag, typename DataType> struct str_converter {
...
+  }
+
+  /**
+   * @brief convert string to underlying value
+   *
+   * @param value value to convert to string
+   * @return DataType converted type
+   */
+  static DataType from_string(const std::string &value) {
+    std::stringstream ss(value);
```
jijoongmoon, line 351: I wonder if this is just for consistency with others or actual use cases? I think if the user wants to set the user_data from string, then they might set the string from outside of function and hand it over as a pointer.

jijoongmoon (approved):
LGTM except with minor comments.

nntrainer/utils/base_properties.h
zhoonit: I am thinking this is strictly internal use only and we should not allow user to set pointer.
The reasoning here is that, properties are designed to be saved to a file while saving an address will have no meaning because of address space layout randomization.

For the user_data, user_data can be set at runtime with below two functions and setting from ini won't be allowed.

1. `ml_train_set_property_for_mode(dataset, TRAIN, "user_data", &user_data, nullptr);` // this is more for the backward compatibility
2. `ml_train_dataset_add_generator(dataset, TRAIN, user_func, &user_data);`

api/capi/src/nntrainer.cpp
zhoonit: update to handle lower case and whitespace case, for the below comment :)
To decouple headers, I'll keep "user_data" here.

api/capi/src/nntrainer.cpp
zhoonit: Nice point, I have change to startsWith here 👍

api/capi/src/nntrainer.cpp
kparichay: capi is allowed to depend on nntrainer sources, so moving the key to a header and using here should be fine.
otherwise create `user_data` as a static const variable here and for line 915.

kparichay (approved):

api/capi/src/nntrainer.cpp
zhoonit: Sure, I'll change to static const.

PR #1423, opened 2021-07-19
[application] Disable MNIST unittest

kparichay:
MNIST unittest was not running as patch c87963c5433514bccdf0d13f733dd15de4356105
added some filecheck but returned 0 upon failure.
This patch removes returning on failure with just a warning.


jijoongmoon (approved):
LGTM

PR #1424, opened 2021-07-20
[activation] Update implementation for in-place 

### LINKED ISSUE #1422: MNIST training not as good
MNIST training on the main branch has some error. It is also possible that other applications have similar issues as well. 
Compared to the release branch 0.2.0, main branch training has two issues/symptoms:
1. Training accuracy does not exceed 60%
2. After saturating to 60%, training accuracy decays down to <40%.
3. Training accuracy increases quite slowly

cc. a.escudero@samsung.com

kparichay:
Update implementation to handle in-place and non-inplace scenarios.
With new memory scheme coming in, in-place activation optimization and
derivative optimization is disabled. This patch updates the activation
function implementations to work for in-place and non-in-place
scenarios.
RNN, LSTM and GRU use activation functions internally with always
in-place settings. So, both the modes are supported.
This patch specifically convers the activation function scenario.
The generic support of in-place will be done when manager is updated.

This resolves the MNIST training bug.

Update MNIST application benchmark values based on the updated weight
initialization.

Resolves #1422

### DIFF
--- a/nntrainer/layers/acti_func.cpp
+++ b/nntrainer/layers/acti_func.cpp
@@ -34,6 +34,9 @@ int ActiFunc::setActivation(
   std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
   std::function<Tensor &(Tensor &, Tensor &, Tensor const &)> const
     &activation_prime_fn) {
+  if (in_place)
+    return ML_ERROR_INVALID_PARAMETER;
+
   _act_fn = activation_fn;
   _act_prime_fn = activation_prime_fn;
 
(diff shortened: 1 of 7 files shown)


nntrainer/layers/acti_func.cpp
```diff
@@ -34,6 +34,9 @@ int ActiFunc::setActivation(
   std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
   std::function<Tensor &(Tensor &, Tensor &, Tensor const &)> const
     &activation_prime_fn) {
+  if (in_place)
```
jijoongmoon, line 37: This is for the softmax kind ? and it raise the error if it is in_place?

jijoongmoon (approved):
LGTM!

nntrainer/layers/acti_func.cpp
kparichay: Yes, this is for softmax kind. Yes, softmax does not support in-place. This is a quick fix.
This check will later move to finalize with layer_v2. I will do this once layer_v2 merges with main.

zhoonit (approved):
Nice catch :)

PR #1427, opened 2021-07-21
[ RNN ] Add mask for dropout 

jijoongmoon:
This commit includes :
  . implementation of mask for dropout

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/rnn.cpp
+++ b/nntrainer/layers/rnn.cpp
@@ -30,7 +30,7 @@ static constexpr size_t SINGLE_INOUT_IDX = 0;
 //  : [1, 1, unit (hidden_size) , unit (hidden_size)]
 // - bias_h ( hidden bias )
 //  : [1, 1, 1, unit (hidden_size)]
-enum RNNParams { weight_xh, weight_hh, bias_h, hidden_state };
+enum RNNParams { weight_xh, weight_hh, bias_h, hidden_state, dropout_mask };
 
 void RNNLayer::finalize(InitLayerContext &context) {
   auto unit = std::get<props::Unit>(props).get();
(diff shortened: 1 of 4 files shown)


nntrainer/layers/rnn.h
```diff
@@ -126,6 +127,8 @@ class RNNLayer : public LayerImpl {
    */
   float dropout_rate;
 
+  float epsilon;
```
kparichay, line 130: please add some description for this member variable.

kparichay (approved):
LGTM

nntrainer/layers/rnn.cpp
```diff
@@ -47,6 +47,11 @@ void RNNLayer::finalize(InitLayerContext &context) {
   output_dim = input_dim;
   output_dim.width(unit);
 
+  if (dropout_rate > epsilon) {
+    wt_idx[RNNParams::dropout_mask] = context.requestTensor(
+      output_dim, "RNN:dropout_mask", true, ITERATION_LIFESPAN);
```
kparichay, line 52: `true` -> false`? true is for trainable tensor (which has gradient as well). Do we need gradient for mask tensor itself?

jijoongmoon: is it? I thought it is "assign now" variable. I will fix it. Thanks.

nntrainer/layers/rnn.h
jijoongmoon: added. Thanks.

nntrainer/layers/rnn.cpp
kparichay: assign now is not part of layer_context interface. It is part of only Var_Grad and used only for internal purpose.

zhoonit (approved):
LGTM

PR #1430, opened 2021-07-21
[CAPI] Implement ml_train_model_get_input|output_dims

lhs8928:
Commit 1: 
    [tensor_dim] package tensor_dim.h with ccapi
    
     - Package tensor_dim with ccapi
     - MAXDIM belongs to tensorDim class
    
    Self evaluation:
    
Commit 2:
    [CAPI] Implement ml_train_model_get_input|output_dims
    
     - Implement ml_train_model_get_input_dimensions
     - Implement ml_train_model_get_output_dimensions
     - Make a unittest for implemented apis
    
    Self evaluation:
    
### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -968,6 +968,122 @@ int ml_train_dataset_destroy(ml_train_dataset_h dataset) {
   return status;
 }
 
+int ml_train_model_get_input_tensors_info(ml_train_model_h model,
+                                          ml_tensors_info_h *info) {
+  int status = ML_ERROR_NONE;
+  ml_train_model *nnmodel;
+  std::shared_ptr<ml::train::Model> m;
+  returnable f;
+
+  check_feature_state();
+
+  if (!info) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
+  m = nnmodel->model;
+  if (m == NULL) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  std::vector<ml::train::TensorDim> dims;
+  f = [&]() {
+    dims = m->getInputDimension();
+    return ML_ERROR_NONE;
+  };
+  status = nntrainer_exception_boundary(f);
+  if (status != ML_ERROR_NONE) {
+    return status;
+  }
+
+  status = ml_tensors_info_create(info);
+  if (status != ML_ERROR_NONE) {
+    return status;
+  }
+
+  status = ml_tensors_info_set_count(*info, dims.size());
+  if (status != ML_ERROR_NONE) {
+    ml_tensors_info_destroy(info);
+    return status;
+  }
+
+  for (unsigned int i = 0; i < dims.size(); ++i) {
+    status = ml_tensors_info_set_tensor_type(*info, i, ML_TENSOR_TYPE_FLOAT32);
+    if (status != ML_ERROR_NONE) {
+      ml_tensors_info_destroy(info);
+      return status;
+    }
+
+    status = ml_tensors_info_set_tensor_dimension(*info, i, dims[i].getDim());
+    if (status != ML_ERROR_NONE) {
+      ml_tensors_info_destroy(info);
+      return status;
+    }
+  }
+
+  return status;
+}
+
+int ml_train_model_get_output_tensors_info(ml_train_model_h model,
+                                           ml_tensors_info_h *info) {
+  int status = ML_ERROR_NONE;
+  ml_train_model *nnmodel;
+  std::shared_ptr<ml::train::Model> m;
+  returnable f;
+
+  check_feature_state();
+
+  if (!info) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
+  m = nnmodel->model;
+  if (m == NULL) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  std::vector<ml::train::TensorDim> dims;
+  f = [&]() {
+    dims = m->getOutputDimension();
+    return ML_ERROR_NONE;
+  };
+  status = nntrainer_exception_boundary(f);
+  if (status != ML_ERROR_NONE) {
+    return status;
+  }
+
+  status = ml_tensors_info_create(info);
+  if (status != ML_ERROR_NONE) {
+    return status;
+  }
+
+  status = ml_tensors_info_set_count(*info, dims.size());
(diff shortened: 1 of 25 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -968,6 +968,84 @@ int ml_train_dataset_destroy(ml_train_dataset_h dataset) {
...
+
+  check_feature_state();
+
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
+  m = nnmodel->model;
+
+  std::vector<nntrainer::TensorDim> dims = m->getInputDimension();
+
+  f = [&]() { return ml_tensors_info_set_count(*info, dims.size()); };
```
kparichay, line 986: where do you allocate `info`? Also please clarify in the header documentation if the info must be allocated by the caller or the callee (based on the function signature, I am assuming its callee).

kparichay: you don't need to wrap `ml_tensors_info_*` functions in `nntrainer_exception_boundary`? They are c-API functions and they will not throw, just return error code.

test/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -968,6 +968,147 @@ TEST(nntrainer_capi_summary, summary_02_n) {
   EXPECT_EQ(status, ML_ERROR_NONE);
 }
 
+TEST(nntrainer_capi_nnmodel, get_input_output_dimension_01_p) {
```
kparichay, line 971: please create a couple more test cases:
1. ml_tensors_info_h can only support till 16 inputs/outputs. Check if the model has than 16 inputs, then how does this API handle it.
2. trying to get the input/output dimension with an empty model.
3. trying to get input/output dimension before compiling the model
4. trying to get input/output dimension after destroying the model
5. verify that the dimension object stays valid/invalid when the model is destroyed

api/ccapi/include/model.h
```diff
@@ -28,6 +28,8 @@
 #include <layer.h>
 #include <optimizer.h>
 
+#include <tensor_dim.h>
```
kparichay, line 31: You will have to package tensor_dim.h with ccapi API, and move the header `tensor_dim.h` to ccapi location.
Then update the namespace to ml::train::TensorDim, and create an alias nntrainer::TensorDim.

kparichay: Update tizen, debian and android packaging.

jijoongmoon: Are we going to open tensor_dim to public?

kparichay: with this API design, TensorDim will have to be opened? or maybe the return type can be vector<int> but it will not completely represent TensorDim.

jijoongmoon: I mean this model.h is opened to user.. if tensor_dim.h is included this header.. then it means open to public. doesn't it?

kparichay: Yes, you are right.

api/capi/src/nntrainer.cpp
lhs8928: Done

test/tizen_capi/unittest_tizen_capi.cpp
lhs8928: First of all I made a test cases you mentioned above, but since the unittest_tizen_capi is turn off in #1328 we can't check the testcase in the upstream/main right now. So I tested on my local branch before #1328 is merged.
For the case 1, we are not implemented multi input so can't check it right now. 
For the case 5, dimension object stays valid even the model is destroyed.

kparichay: great.

api/capi/src/nntrainer.cpp
```diff
@@ -968,6 +968,82 @@ int ml_train_dataset_destroy(ml_train_dataset_h dataset) {
...
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
+  m = nnmodel->model;
+  if (m == NULL) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  std::vector<ml::train::TensorDim> dims = m->getInputDimension();
+
+  status = ml_tensors_info_set_count(*info, dims.size());
```
kparichay, line 989: The interface expects the allocation of info object by this function. 
The allocation of the ml_tensors_info_h must be done in this function than by the caller (like in the unittest).

api/ccapi/include/tensor_dim.h
```diff
@@ -22,16 +22,17 @@
...
-
-constexpr const size_t MAXDIM = 4;
+namespace ml {
+namespace train {
 
 /**
  * @brief Tensor Dimension. This class is used to save dimension information
  *
  */
 class TensorDim {
```
kparichay, line 32: this is minor but can you also move the implementations out of this header to its source, so that if some implementation is updated for bugfix, header remains same.

kparichay (approved):
Overall LGTM

api/capi/src/nntrainer.cpp
```diff
@@ -968,6 +968,82 @@ int ml_train_dataset_destroy(ml_train_dataset_h dataset) {
...
+  check_feature_state();
+
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
+  m = nnmodel->model;
+  if (m == NULL) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  std::vector<ml::train::TensorDim> dims = m->getOutputDimension();
```
zhoonit, line 1025: please wrap with exception boundary

api/capi/src/nntrainer.cpp
```diff
@@ -968,6 +968,82 @@ int ml_train_dataset_destroy(ml_train_dataset_h dataset) {
...
+  check_feature_state();
+
+  ML_TRAIN_GET_VALID_MODEL_LOCKED(nnmodel, model);
+  ML_TRAIN_ADOPT_LOCK(nnmodel, model_lock);
+  m = nnmodel->model;
+  if (m == NULL) {
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  std::vector<ml::train::TensorDim> dims = m->getInputDimension();
```
zhoonit, line 987: please wrap with exception boundary

lhs8928: Done

api/capi/src/nntrainer.cpp
lhs8928: Done

api/capi/src/nntrainer.cpp
lhs8928: Done

api/ccapi/include/tensor_dim.h
lhs8928: I made a another pr for this cause this will gonna too huge.
The following pr(#1456) is related with this.

jijoongmoon (approved):
LGTM

PR #1437, opened 2021-07-26
[weight/var_grad] Reduce weight/var_grad dependency for devel

kparichay:
The devel packaging has to be updated, where a lot of headers have to be removed.  This patch makes preparations for it to able to remove weight and var_grad related headers.
- optimizer factory has been deprecated and removed
- parse_utils have been divided into two as it needs exposure for node_exporter
- weight and var_grad are now no longer need for layer_devel requirements

Other minor refactoring is done to make removal of these headers possible.

See Also #986 


zhoonit (commented):
Good job 👍 it's getting cleaner and nicer, one concern here is that it will break this patch will break some codes in the nntrainer-internal that is using neuralnet headers. Do you have any thoughts for the smooth transition?

kparichay:
> Good job  it's getting cleaner and nicer, one concern here is that it will break this patch will break some codes in the nntrainer-internal that is using neuralnet headers. Do you have any thoughts for the smooth transition?

The update on the nntrainer-internal can be done first, and then merge this patch.

kparichay:
I will separate the last commit for packaging to separate PR. Then nntrainer-internal can be updated first before the actual packaging is updated.

nntrainer/layers/layer_context.cpp
```diff
@@ -0,0 +1,268 @@
...
+ * @brief Get regularization loss for the weight
+ *
+ * @param idx Identifier of the weight
+ * @return float Value of the loss
+ */
+float RunLayerContext::getWeightRegularizationLoss(unsigned int idx) const {
+  if (weights[idx]->hasGradient())
+    return weights[idx]->getRegularizationLoss();
+
+  return 0;
```
zhoonit, line 52: I wonder if it is valid to call for regularizationloss when they don't have gradient?
If it is, I guess we don't need check but return `weights[idx]->getRegularizationLoss()`?

nntrainer/layers/layer_context.cpp
```diff
@@ -0,0 +1,268 @@
...
+  return weights[idx]->getGradientRef();
+}
+
+/**
+ * @brief Get regularization loss for the weight
+ *
+ * @param idx Identifier of the weight
+ * @return float Value of the loss
+ */
+float RunLayerContext::getWeightRegularizationLoss(unsigned int idx) const {
```
zhoonit, line 48: maybe check idx and throw?

zhoonit (approved):
LGTM with minor comments :)

nntrainer/layers/layer_context.cpp
kparichay: A weight can be initialized to be non-trainable but with regularization for now. I guess if that is stopped, then what you said should work.

nntrainer/layers/layer_context.cpp
kparichay: i will replace `[]` with `.at()` which will do this for us :).

kparichay: I will update this in a separate PR.

nntrainer/layers/layer_context.cpp
zhoonit: Cool

jijoongmoon (approved):
LGTM

PR #1438, opened 2021-07-26
[CAPI] Add ml_train_model_get_layer to internal

zhoonit:
- [CAPI] Add ml_train_model_get_layer to internal

```
This patch adds ml_train_model_get_layer to internal api

**Self evaluation:**

```

### DIFF
--- a/api/capi/include/nntrainer-tizen-internal.h
+++ b/api/capi/include/nntrainer-tizen-internal.h
@@ -8,6 +8,7 @@
  * @note   This header is designed to be used only in Tizen
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
  * @bug    No known bugs except for NYI items
  *
  */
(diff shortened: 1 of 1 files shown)


api/capi/include/nntrainer-tizen-internal.h
```diff
@@ -21,6 +22,21 @@
 extern "C" {
 #endif /* __cplusplus */
 
+/**
+ * @brief Get neural network layer from the model with the given name.
+ * @details Use this function to get already created Neural Network Layer. The
+ * returned layer must not be deleted as it is owned by the model.
```
kparichay, line 28: Add a note that this works only for layers added by the API, and does not currently support layers supported for layers created by the ini.
I will try and add it, but if not, its better to match the documentation to the current status.

zhoonit: Okay :)

kparichay (approved):
LGTM

api/capi/include/nntrainer-tizen-internal.h
jijoongmoon: is it possible to modify the layer which given by this API? I mean, let's say user changes dimension, unit size. then whole network need to be compiled. If it is the case we are considering here, then we need to add this description.  Or we are not allow to change, allow just get the information of that layer, then we need to add description for this purpose too.

zhoonit: I think it's good point! I'll add description that modification after calling model_compile will have no effect.

jijoongmoon: Great. Thanks.

jijoongmoon (approved):
LGTM

PR #1448, opened 2021-07-29
[ Property ] Add output_shape property

jijoongmoon:
The output_shape should be defined to set the mutiple tensor output.
The semetic of it might be output_shape="1:1:2:1,1:2:3:3,3:2:1".
In order to set it, we need to define property first.

This commit includes.
  . implementation of TensorShape Property.
  . implementation of TensorsSpec Property.
  . implementation of OutputSpec Property.

### DIFF
--- a/nntrainer/layers/common_properties.cpp
+++ b/nntrainer/layers/common_properties.cpp
@@ -48,6 +48,38 @@ void FilePath::set(const std::string &v) {
 
 std::ifstream::pos_type FilePath::file_size() { return cached_pos_size; }
 
+bool TensorShape::isValid(const std::string &v) const {
+  // Tensor dimension does not allow the 0 & number of dimension shuld be less
+  // than MAXDIM 4. Also the white space will be ignored.
+
+  static std::regex allowed("[1-9:]:[1-9:]*");
+  auto is_valid = [](const std::string &value) {
+    auto values = split(value, std::regex("\\s*\\:\\s*"));
+    if (values.size() > TensorDim::MAXDIM)
+      return false;
+    // check if the values is nothing ( 1::1 )
+    for (auto item : values) {
+      if (item.empty())
+        return false;
+    }
+    return true;
+  };
+
+  return std::regex_match(v, allowed) && is_valid(v);
+}
+
+bool TensorsSpec::operator==(const TensorsSpec &rhs) const {
+  for (unsigned int i = 0; i < shapes.size(); ++i) {
+    if (!istrequal(to_string(shapes[i]), to_string(rhs.getTensorsShape()[i])))
+      return false;
+  }
+  return true;
+}
+
+bool OutputSpec::isValid(const TensorsSpec &v) const {
+  return v.getTensorsShape().size() > 0;
+}
+
 ConnectionSpec::ConnectionSpec(const std::vector<props::Name> &layer_ids_,
                                const std::string &op_type_) :
   op_type(op_type_),
@@ -240,6 +272,57 @@ str_converter<props::connection_prop_tag, props::ConnectionSpec>::from_string(
   props::Name n;
   n.set(value); // explicitly trigger validation using set method
   return props::ConnectionSpec({n});
+}
+
+template <>
+std::string
+str_converter<props::tensors_shape_prop_tag, props::TensorsSpec>::to_string(
+  const props::TensorsSpec &value) {
+
+  std::stringstream ss;
+  auto last_iter = value.getTensorsShape().end() - 1;
+  for (auto iter = value.getTensorsShape().begin(); iter != last_iter; ++iter) {
+    for (unsigned int i = 0; i < TensorDim::MAXDIM - 1; ++i)
+      ss << (*iter).get().getTensorDim(i) << ":";
+    ss << (*iter).get().getTensorDim(TensorDim::MAXDIM - 1) << ",";
+  }
+  for (unsigned int i = 0; i < TensorDim::MAXDIM - 1; ++i)
+    ss << (*last_iter).get().getTensorDim(i) << ":";
+  ss << (*last_iter).get().getTensorDim(TensorDim::MAXDIM - 1);
+
+  return ss.str();
+}
+
+template <>
+props::TensorsSpec
+str_converter<props::tensors_shape_prop_tag, props::TensorsSpec>::from_string(
+  const std::string &value) {
+
+  auto values = split(value, std::regex("\\s*\\,\\s*"));
+  auto generate_TensorShape_vector = [](const auto &values) {
+    props::TensorShape ts;
+    std::vector<props::TensorShape> shapes_;
+    shapes_.reserve(values.size());
+
+    for (auto &item : values) {
+      if (!ts.isValid(item)) {
+        break;
+      }
+      TensorDim dim(item);
+      props::TensorShape s(dim);
+      shapes_.emplace_back(s);
+    }
+
+    return shapes_;
+  };
+
+  auto shapes = generate_TensorShape_vector(values);
+
+  NNTR_THROW_IF((shapes.size() != values.size()), std::invalid_argument)
+    << "size of tensors spec should equal to values : " << shapes.size()
+    << " vs." << values.size();
+
+  return props::TensorsSpec(shapes);
 }
 
 } // namespace nntrainer
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -56,6 +56,41 @@ public:
 };
 
 /**
+ * @brief Tensor Shape property, TensorShape is an identifier of an object
+ *
+ */
+class TensorShape : public nntrainer::Property<nntrainer::TensorDim> {
+public:
+  /**
+   * @brief Construct a new TensorShape object
+   *
+   */
+  TensorShape() : nntrainer::Property<nntrainer::TensorDim>(){};
+
+  /**
+   * @brief Construct a new TensorShape object
+   * @param value default value of a TensorDim
+   *
+   */
+  TensorShape(const nntrainer::TensorDim &dim) :
+    nntrainer::Property<nntrainer::TensorDim>(dim){};
+
+  static constexpr const char *key =
+    "tensor_shape"; /**< unique key to access */
+
+  using prop_tag = dimension_prop_tag; /**< property type */
+
+  /**
+   * @brief TensorShape validator
+   *
+   * @param v string to validate
+   * @retval true if it contains unsigned int numbers and/or ':'
+   * @retval false if it is empty or contains non-valid character
+   */
+  bool isValid(const std::string &v) const;
+};
+
+/**
  * @brief unit property, unit is used to measure how many weights are there
  *
  */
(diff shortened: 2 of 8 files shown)


nntrainer/layers/common_properties.cpp
```diff
@@ -48,6 +48,33 @@ void FilePath::set(const std::string &v) {
...
 std::ifstream::pos_type FilePath::file_size() { return cached_pos_size; }
 
+bool TensorShape::isValid(const std::string &v) const {
+  // Tensor dimension does not allow the 0 & number of dimension shuld be less
+  // than MAXDIM 4. Also the white space will be ignored.
+
+  static std::regex allowed("[1-9:]*");
+  auto check_num_tensors = [](const std::string &value) {
+    auto values = split(value, std::regex("\\s*\\:\\s*"));
+    if (values.size() > TensorDim::MAXDIM)
```
kparichay, line 58: also, check that each split value inside the values array is a non-empty string. For example,  "1::1"?

nntrainer/layers/common_properties.cpp
```diff
@@ -48,6 +48,33 @@ void FilePath::set(const std::string &v) {
 
 std::ifstream::pos_type FilePath::file_size() { return cached_pos_size; }
 
+bool TensorShape::isValid(const std::string &v) const {
+  // Tensor dimension does not allow the 0 & number of dimension shuld be less
+  // than MAXDIM 4. Also the white space will be ignored.
+
+  static std::regex allowed("[1-9:]*");
```
kparichay, line 55: "[1-9:]*" -> "[1-9]:[1-9:]*"?
then `!v.empty()` can also be removed from L63.
I wonder if "::" will pass the current checks?

nntrainer/layers/common_properties.h
```diff
@@ -69,6 +104,84 @@ class Unit : public nntrainer::Property<unsigned int> {
   bool isValid(const unsigned int &v) const override { return v > 0; }
 };
 
+/**
+ * @brief TensorsSpec property, this defines tensors shape of output
```
kparichay, line 108: this defines tensors shape~ of output~? Then reuse this can be reused input_shape as well.

nntrainer/layers/common_properties.cpp
```diff
@@ -242,4 +269,46 @@ str_converter<props::connection_prop_tag, props::ConnectionSpec>::from_string(
...
+}
+
+template <>
+props::TensorsSpec
+str_converter<props::tensors_shape_prop_tag, props::TensorsSpec>::from_string(
+  const std::string &value) {
+  auto generate_TensorShape_vector = [](const std::string &value) {
+    auto values = split(value, std::regex("\\s*\\,\\s*"));
+    std::vector<props::TensorShape> shapes_;
+    shapes_.reserve(values.size());
```
kparichay, line 298: throw if `values` is empty?

nntrainer/layers/common_properties.cpp
```diff
@@ -242,4 +269,46 @@ str_converter<props::connection_prop_tag, props::ConnectionSpec>::from_string(
   return props::ConnectionSpec({n});
 }
 
+template <>
+std::string
+str_converter<props::tensors_shape_prop_tag, props::TensorsSpec>::to_string(
+  const props::TensorsSpec &value) {
```
kparichay, line 275: can this function be called with value where value.getTensorShape().size() == 0?

nntrainer/layers/layer_node.cpp
```diff
@@ -139,6 +140,14 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
     std::get<props::Distribute>(*layer_node_props).get();
   auto left_properties = loadProperties(properties, *layer_node_props);
 
+  size_t num_output_tensors = getNumOutputTensors();
+
+  if (num_output_tensors > 1) {
+    init_context =
+      InitLayerContext(init_context.getInputDimensions(),
+                       std::max((unsigned int)num_output_tensors, 1u));
```
kparichay, line 148: also resize output_layers to match?
also add check in layer_node.h L380 that when setOutputLayers is called from outside, this will also be resized (maybe check for mismatch)?
maybe create resizeOutputDimensions() like resizeInputDimensions() for help.

nntrainer/layers/layer_node.cpp
```diff
@@ -287,6 +306,15 @@ bool LayerNode::getFlatten() const {
   return flatten.get();
 }
 
+unsigned int LayerNode::getNumOutputTensors() const {
+  auto &output = std::get<props::OutputSpec>(*layer_node_props);
```
kparichay, line 310: can this mismatch with getNumOutputs()?

test/unittest/unittest_common_properties.cpp
```diff
@@ -250,6 +250,62 @@ TEST(Padding2D, given_padding_is_negative_02_n) {
...
+TEST(OutputSpecProperty, emptyString_n_01) {
+  using namespace nntrainer::props;
+  InputSpec actual;
+  EXPECT_THROW(nntrainer::from_string("", actual), std::invalid_argument);
+}
+
+TEST(OutputSpecProperty, noSeperator_n_01) {
+  using namespace nntrainer::props;
+  OutputSpec actual;
+  EXPECT_THROW(nntrainer::from_string("1:1:1 2:2:2", actual),
```
kparichay, line 305: add negative unittest with empty string "" and with ","?

test/unittest/unittest_common_properties.cpp
```diff
@@ -250,6 +250,62 @@ TEST(Padding2D, given_padding_is_negative_02_n) {
...
 }
 
+TEST(TensorShapeProperty, setPropertyValid_p) {
+  nntrainer::props::TensorShape n;
+  EXPECT_NO_THROW(from_string("1:1:1:1", n));
+
+  EXPECT_EQ(to_string(n), "1:1:1:1");
+
+  EXPECT_NO_THROW(from_string("1:1", n));
+  EXPECT_EQ(n.get(), nntrainer::TensorDim("1:1"));
```
kparichay, line 260: add negative unittest with just ":"?

nntrainer/layers/layer_node.cpp
jijoongmoon: I intentionally give this different name. Cause getNumOutputs is to get the connection number and this if get the number of output tensors. eventually total output size must be getNumOutputs x getNumOutputTensors.

nntrainer/layers/layer_node.cpp
jijoongmoon: right. we need to think the mismatch. I will fix and add some more code in next pr which is considering output layers. Meanwhile, we need to think the case.. Multi output connection layers with multi output tensors. In this pr, those are not considered yet.

nntrainer/layers/common_properties.h
jijoongmoon: right. it can be used for input_shape as well.

nntrainer/layers/layer_node.cpp
kparichay: yes.

nntrainer/layers/layer_node.cpp
kparichay: I think there is some confusion. `getNumOutputs` gives the number of outputs from the init context, which is the number of output dimensions. The number of connections is given by `getNumOutputConnections`.
So maybe getNumOutputs and getNumOutputsTensors should be in sync.

nntrainer/layers/common_properties.cpp
jijoongmoon: yes. I missed it. Thanks.

nntrainer/layers/common_properties.cpp
jijoongmoon: fixed.

nntrainer/layers/common_properties.cpp
jijoongmoon: if it is 0, then it will give empty string.

nntrainer/layers/layer_node.cpp
jijoongmoon: AFIK, both of them are the same for the current implementation. even if it is called differently. Cause the number of output of init_context is set by the number of the connection actually. What I'm thinking is we have to return the number of how many layers are connected if the getNumOutputConnections is called, and if you call the getNumOutputs() then you have to give the number of connection x number of output tensors. and if you call the getNumOutputTesnors(), it has to return the number of output tensors for each output.

test/unittest/unittest_common_properties.cpp
jijoongmoon: added.

test/unittest/unittest_common_properties.cpp
jijoongmoon: added. Thanks.

nntrainer/layers/common_properties.cpp
jijoongmoon: fixed.

nntrainer/layers/common_properties.cpp
kparichay: aah, then its fine, thanks.

nntrainer/layers/layer_node.cpp
kparichay: I think I am missing something. In my understanding, in a graph I thought we only had 1 tensor for each connection?
For if there are two nodes N1 and N2, both feeding their outputs to N3, then both N1 and N2 can only give 1 tensor to N3, making total of 2 tensors received by N3. Please correct me if I am wrong.

kparichay (approved):
LGTM

nntrainer/layers/layer_node.cpp
```diff
@@ -139,6 +140,14 @@ int LayerNode::setProperty(std::vector<std::string> properties) {
     std::get<props::Distribute>(*layer_node_props).get();
   auto left_properties = loadProperties(properties, *layer_node_props);
 
+  size_t num_output_tensors = getNumOutputTensors();
+
+  if (num_output_tensors > 1) {
```
zhoonit, line 145: Wonder if we can move this logic to `finalize()`?

nntrainer/layers/layer_node.cpp
```diff
@@ -186,6 +195,16 @@ bool LayerNode::setProperty(const std::string &key, const std::string &value) {
...
+  case PropertyType::output_shape: {
+    std::vector<TensorDim> output_dim = init_context.getOutputDimensions();
+    if (getNumOutputs() == 0)
+      output_dim.resize(1);
+
+    // for (auto iter = output_dim.cbegin();iter!= output_dim.cend();iter++){
+    // NYI
+    // }
+
+  } break;
```
zhoonit, line 207: FYI) This line won't reach as line 141 will already parse output_shape. But looks like this logic perhaps can be moved to `finalize()`

nntrainer/utils/base_properties.cpp
```diff
@@ -102,8 +102,18 @@ TensorDim str_converter<dimension_prop_tag, TensorDim>::from_string(
     tokens.push_back(token);
   }
 
-  NNTR_THROW_IF(tokens.size() > ml::train::TensorDim::MAXDIM,
-                std::invalid_argument)
+  static std::regex allowed("[1-9:]*");
```
zhoonit, line 105: is 10:10:100 allowed with this logic?

nntrainer/utils/base_properties.cpp
```diff
@@ -102,8 +102,18 @@ TensorDim str_converter<dimension_prop_tag, TensorDim>::from_string(
...
+  auto check_num_tensors = [](const std::string &value) {
+    auto values = split(value, std::regex("\\s*\\:\\s*"));
+    if (values.size() > TensorDim::MAXDIM || !std::regex_match(value, allowed))
```
zhoonit, line 108: if values.size() == 0 okay?

jijoongmoon:
close due to #1732

PR #1449, opened 2021-07-30
[optimizer] Make optimizer independent of weight

kparichay:
Cleanup optimizer interface where some extra interfaces have been
removed, and arguments interface has been updated.
Further, the dependency on weight header has been removed, and added an
optimizer context interface which provides an interface to weight
related tensors required by the optimizer.

Update the interface for the layer.h setProperty.

### DIFF
--- a/nntrainer/optimizers/optimizer_devel.h
+++ b/nntrainer/optimizers/optimizer_devel.h
@@ -86,9 +86,9 @@ public:
   };
 
   /**
-   * @brief     initialize optimizer.
+   * @brief     finalize optimizer.
    */
-  virtual void initialize(){};
+  virtual void finalize(){};
 
   /**
    * @brief     Read Training optimizer paramters from file
--- a/nntrainer/optimizers/optimizer_devel.h
+++ b/nntrainer/optimizers/optimizer_devel.h
@@ -19,8 +19,8 @@
 #include <memory>
 
 #include <optimizer.h>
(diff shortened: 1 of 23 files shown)


nntrainer/optimizers/optimizer_devel.h
```diff
@@ -88,25 +85,10 @@ class Optimizer : public ml::train::Optimizer {
...
-  virtual void setProperty(const std::string &key,
-                           const std::string &value) = 0;
-
   /**
    * @brief     initialize optimizer.
-   * @retval #ML_ERROR_NONE Successful.
-   * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    */
-  virtual int initialize() = 0;
+  virtual void initialize(){};
```
zhoonit, line 91: As we are making a api breaking change, may be finalize to keep modules consistent?

zhoonit (approved):
LGTM with a off-the-topic suggestion :)

nntrainer/optimizers/optimizer_devel.h
kparichay: will update.

kparichay: done.

jijoongmoon (commented):
LGTM

jijoongmoon (approved):

PR #1450, opened 2021-07-30
[devel/pkg] Update the devel packaging 

### LINKED ISSUE #986: C++ API and architecture refactor

kparichay:
Update the devel packaging to remove unnecessary headers.. Below is the list of packaging update:
- weight/var_grad headers are removed
- app_context and other graph, network related headers are removed. They
can be added once network/neural network headers are refactored for
custom models.
- manager and dynamic_training_optimization header is removed
- databuffer headers are commented. They will be updated once dataset
refactoring is done.

Resolves #986

### DIFF
--- a/debian/nntrainer-dev.install
+++ b/debian/nntrainer-dev.install
@@ -1,29 +1,23 @@
-/usr/include/nntrainer/app_context.h
+# node exporter and its dependencies
 /usr/include/nntrainer/nntrainer_error.h
-/usr/include/nntrainer/nntrainer_log.h
-/usr/include/nntrainer/nntrainer_logger.h
-/usr/include/nntrainer/acti_func.h
 /usr/include/nntrainer/common_properties.h
-/usr/include/nntrainer/weight.h
-/usr/include/nntrainer/var_grad.h
 /usr/include/nntrainer/base_properties.h
+/usr/include/nntrainer/props_util.h
 /usr/include/nntrainer/node_exporter.h
-/usr/include/nntrainer/parse_util.h
-/usr/include/nntrainer/graph_core.h
-/usr/include/nntrainer/graph_node.h
-/usr/include/nntrainer/databuffer.h
-/usr/include/nntrainer/databuffer_factory.h
-/usr/include/nntrainer/layer_context.h
-/usr/include/nntrainer/layer_devel.h
-/usr/include/nntrainer/neuralnet.h
+# tensor headers
 /usr/include/nntrainer/tensor.h
 /usr/include/nntrainer/tensor_wrap_specs.h
+# todo: update dataset headers
+/usr/include/nntrainer/databuffer.h
+/usr/include/nntrainer/databuffer_factory.h
+# layer headers
+/usr/include/nntrainer/layer_context.h
+/usr/include/nntrainer/layer_devel.h
+/usr/include/nntrainer/layer_impl.h
+# optimizer headers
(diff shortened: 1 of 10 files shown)


kparichay:
Please note that this patch will be breaking any independent applications code which uses internal headers directly.

debian/nntrainer-dev.install
```diff
@@ -17,6 +17,10 @@
 /usr/include/nntrainer/layer_devel.h
 /usr/include/nntrainer/neuralnet.h
 /usr/include/nntrainer/tensor.h
+<<<<<<< HEAD
+=======
+/usr/include/nntrainer/tensor_dim.h
+>>>>>>> [weight/var_grad] Remove exposure of weight/var_grad
```
zhoonit: Could you check this out?

zhoonit: I can see updated, maybe deleting this commit?

zhoonit (approved):
LGTM with minor issue :)

debian/nntrainer-dev.install
kparichay: done.

packaging/nntrainer.spec
```diff
@@ -431,33 +431,26 @@ cp -r result %{buildroot}%{_datadir}/nntrainer/unittest/
...
 %{_includedir}/nntrainer/databuffer_factory.h
+# layer headers
 %{_includedir}/nntrainer/layer_context.h
 %{_includedir}/nntrainer/layer_devel.h
-%{_includedir}/nntrainer/neuralnet.h
-%{_includedir}/nntrainer/tensor.h
-%{_includedir}/nntrainer/tensor_wrap_specs.h
+%{_includedir}/nntrainer/layer_impl.h
+# optimizer headers
+%{_includedir}/nntrainer/optimizer_context.h
```
jijoongmoon, line 451: Great!. Looks like more organized and simpler!

jijoongmoon (approved):
LGTM!

PR #1451, opened 2021-07-30
[ TEST ] remove layer_v2 test

jijoongmoon:
once we merged layer_v2, we remove layer_v2 unit test

### DIFF
--- a/test/unittest/layers/meson.build
+++ b/test/unittest/layers/meson.build
@@ -58,8 +58,3 @@ exe = executable(
   install: get_option('enable-test'),
   install_dir: application_install_dir
 )
-
-# @todo delete v2 once migration is done
-test('unittest_layers_v2', exe,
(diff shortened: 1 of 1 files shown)


test/unittest/layers/meson.build
```diff
@@ -58,8 +58,3 @@ exe = executable(
   install: get_option('enable-test'),
   install_dir: application_install_dir
 )
-
-# @todo delete v2 once migration is done
-test('unittest_layers_v2', exe,
```
zhoonit, line 63: Maybe s/unittest_layers_v2/unittest_layers/ ?

I think deleting this might remove test runs from gbs, pdebuild.

kparichay: The tests built in this are tested using this statement I think.
So, I think rename should do.

kparichay: Verified locally, deleting this disables this test runs.
Just rename is required.

PR #258, opened 2020-07-03
[bug fixes] [tensor] Tensor constructor should not set zero

kparichay:
Tensor constructors should not zero to the memory by default
This hides many bugs in the code and also leads to big overheads
Removed setting zero in constructor and corresponding many bug fixes

Signed-off-by: Parichay Kapoor <pk.kapoor@samsung.com>

PR #1452, opened 2021-07-30
[Tensor/weight] Support initialization of tensors with constructors

### LINKED ISSUE #1411: Provide initializer for requestTensor() in InitLayerContext
Provide support for initializer for the requested tensors in InitLayerContext.

kparichay:
- Update WeightInitializer to Tensor::Initiailzer
- Support initializer inside tensor with allocation
- Move explicit initialization inside var_grad to tensors
- Remove caching initialization inside weights and let it be handled by tensors
- Add support for initialization in requestTensors()

See Also #986 https://github.com/nnstreamer/nntrainer/pull/1031#discussion_r597551354
Resolves #1411

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -153,10 +153,12 @@ void Tensor::allocate() {
     data = std::shared_ptr<float>(src_tensor->tensor()->data,
                                   src_tensor->tensor()->data.get() +
                                     src_tensor->offset());
+    /** as this memory is shared, do NOT initialize */
   } else {
     /// allocate new memory for the tensor data
     data = std::shared_ptr<float>(new float[dim.getDataLen()],
                                   std::default_delete<float[]>());
+    initialize();
   }
 }
 
@@ -541,9 +543,11 @@ void Tensor::createSharedDataTensor(const Tensor &src, Tensor &dest,
    * @note src.data and src.src_tensor CAN co-exist. src.src_tensor is stored
    * if the batch size of src is updated and needs reallocation.
    */
-  if (src.data)
-    dest.data = std::shared_ptr<float>(src.data, src.data.get() + offset);
-  else if (!src.src_tensor)
+  dest.data = nullptr;
+  if (src.data) {
+    dest.src_tensor = std::make_shared<SrcSharedTensor>(&src, offset);
+    dest.allocate();
+  } else if (!src.src_tensor)
     dest.src_tensor = std::make_shared<SrcSharedTensor>(&src, offset);
   else
     dest.src_tensor = std::make_shared<SrcSharedTensor>(
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -239,6 +239,62 @@ void Tensor::setRandUniform(float min, float max) {
     std::uniform_real_distribution<float>(min, max));
 }
 
+void Tensor::initialize() {
+  if (empty() || !isAllocated())
+    return;
+
+  unsigned int fan_in, fan_out;
+
+  /// @fixme: when unit is equal to one, this does not work, we need to rely on
+  /// effective dimension then actual numbers here. For now, some heuristics
+  /// added to infer what would be fan_in/fan_out
+  if (dim.batch() * dim.channel() * dim.height() == 1) {
+    fan_out = fan_in = dim.width();
+  } else if (dim.batch() * dim.channel() == 1) { /// fc layer - 2-D tensor
+    fan_in = dim.height();
+    fan_out = dim.width();
+  } else { /// conv2d filters - 4d tensor, @todo extend this to > 4
+    auto field_size = dim.height() * dim.width();
+
+    // this also handles below cases.
+    // 1. fan_in = fan_out = 1 as well.
+    // 2. batch == 1, channel == 1 and height == 1, theoretical rank of 1
+    fan_in = dim.channel() * field_size;
+    fan_out = dim.batch() * field_size;
+  }
+
+  switch (initializer) {
+  case Tensor::Initializer::ZEROS:
+    setZero();
+    break;
+  case Tensor::Initializer::ONES:
+    setValue(1.0f);
+    break;
+  case Tensor::Initializer::LECUN_NORMAL:
+    setRandNormal(0.0f, sqrtFloat(1.0f / fan_in));
+    break;
+  case Tensor::Initializer::XAVIER_NORMAL:
+    setRandNormal(0.0f, sqrtFloat(2.0f / (fan_in + fan_out)));
+    break;
+  case Tensor::Initializer::HE_NORMAL:
+    setRandNormal(0.0f, sqrtFloat(2.0f / (fan_in)));
+    break;
+  case Tensor::Initializer::LECUN_UNIFORM:
+    setRandUniform(-1.0f * sqrtFloat(1.0f / fan_in), sqrtFloat(1.0f / fan_in));
+    break;
+  case Tensor::Initializer::XAVIER_UNIFORM:
+    setRandUniform(-1.0f * sqrtFloat(6.0f / (fan_in + fan_out)),
+                   sqrtFloat(6.0 / (fan_in + fan_out)));
+    break;
+  case Tensor::Initializer::HE_UNIFORM:
+    setRandUniform(-1.0f * sqrtFloat(6.0f / (fan_in)),
+                   sqrtFloat(6.0 / (fan_in)));
+    break;
+  default:
+    break;
+  }
+}
+
 Tensor::Tensor(
   std::vector<std::vector<std::vector<std::vector<float>>>> const &d) {
 
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -56,7 +56,7 @@
 
 #define CREATE_IF_EMPTY_DIMS(tensor, ...) \
   do {                                    \
-    if (tensor.uninitialized())           \
+    if (tensor.empty())                   \
       tensor = Tensor(__VA_ARGS__);       \
   } while (0);
 namespace nntrainer {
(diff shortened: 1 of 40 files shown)


nntrainer/tensor/var_grad.h
```diff
@@ -65,8 +67,9 @@ class Var_Grad {
   explicit Var_Grad(const Spec &spec) :
     Var_Grad(std::get<0>(spec), // TensorDim
              std::get<1>(spec), // need_gradient
+             std::get<2>(spec), // need_gradient
```
jijoongmoon, line 70: need_gradient --> initializer?

kparichay: done.

nntrainer/tensor/tensor.h
```diff
@@ -54,58 +54,48 @@ class SrcSharedTensor;
...
+  enum class Initializer {
+    ZEROS,          /** Zero initialization */
+    ONES,           /** One initialization */
+    LECUN_NORMAL,   /** LeCun normal initialization */
+    LECUN_UNIFORM,  /** uniform initialization */
+    XAVIER_NORMAL,  /** Xavier normal initialization */
+    XAVIER_UNIFORM, /** Xavier uniform initialization */
+    HE_NORMAL,      /** He normal initialization */
+    HE_UNIFORM,     /** He uniform initialization */
+    NONE            /** No initialization */
```
kparichay, line 70: FILE - to be added here later.

jijoongmoon (approved):
LGTM. please consider having an initializer class as we discussed offline later.

jijoongmoon (approved):
LGTM

kparichay:
> LGTM. please consider having an initializer class as we discussed offline later.

yes

nntrainer/tensor/tensor.cpp
```diff
@@ -239,6 +239,62 @@ void Tensor::setRandUniform(float min, float max) {
     std::uniform_real_distribution<float>(min, max));
 }
 
+void Tensor::initialize() {
+  if (empty() || !isAllocated())
```
zhoonit, line 245: early return here if initializer is none as well?

zhoonit: and if empty(), !isAllocated, shouldn't it be throwing?

nntrainer/tensor/var_grad.cpp
```diff
@@ -18,15 +18,15 @@
...
+                   bool ng, bool alloc_now_, const std::string &name) :
   dim(dim),
   need_gradient(ng),
   alloc_now(alloc_now_),
   name(name) {
-  var = std::make_shared<Tensor>(dim, alloc_now);
+  var = std::make_shared<Tensor>(dim, alloc_now, init);
   if (need_gradient)
-    grad = std::make_shared<Tensor>(dim, alloc_now);
+    grad = std::make_shared<Tensor>(dim, alloc_now, Tensor::Initializer::ZEROS);
```
zhoonit, line 33: if we allocate this to zero here, does this apply for the every iteration?

nntrainer/tensor/var_grad.cpp
```diff
@@ -26,6 +26,10 @@ Var_Grad::Var_Grad(const TensorDim &dim, const Tensor::Initializer init,
   name(name) {
   var = std::make_shared<Tensor>(dim, alloc_now, init);
   if (need_gradient)
+    /**
+     * @todo gradient initializer should be none, and then they should be set
```
zhoonit, line 30: I guess this comment is related to the question?

nntrainer/tensor/var_grad.cpp
kparichay: No, initializer (as the name suggests) initializes only upon allocation by default.
Tensor::initialize() can be called again to re-initialize the variable.

kparichay: It is currently set to zero to keep the original behavior. Updating the behavior of when the gradient is zeroed will be another PR.

nntrainer/tensor/var_grad.cpp
kparichay: yes, this is the corresponding to do for the answer above.

nntrainer/tensor/tensor.cpp
kparichay: > early return here if initializer is none as well?

will add

zhoonit (approved):
LGTM :)

nntrainer/tensor/tensor.cpp
kparichay: > and if empty(), !isAllocated, shouldn't it be throwing?

- in !isAllocated(), it will note the initialize value, and whenever access, the tensor will be initialized (allocate will initialize). This allows delayed initialized till the memory is allocated.
- if empty(), the above behavior will be followed, but here, we can throw as the tensor shape info is not available.

kparichay: call of initialize() is missing in allocate() (which I thought I did add). I will add it.

kparichay:
Note to self - add unit tests for tensor initialization taking place upon allocation.

nntrainer/tensor/tensor.cpp
zhoonit: So, is initialize must come after allocate? or vice versa? or have no order?

`empty()` previously was checking on `unintialized()` checks on the `size()`, which does not require `allocate()`

if allocate -> initialize
   allocate should not initialize but should be always separate call for the consistent semantics or initialize should be a private function? But it seems, allocate() need tensor dimension fixed so initialize has to be done first?

We might need different naming or firmer semantics. Delaying also sounds good though.

kparichay: initialize() ideally must come after allocate(), but calling before it is a no-op.
initialize(init=...) has an extra effect of setting which initializer to use.

When allocate() will be called, the memory will be allocated and tensor will be initialize() at that time (by calling intiailize() internally).

kparichay: > which follows different question, is uninitialized() == empty() ?

`uninitialized()` was confusing because of its name. So, it is removed.
`empty()` means size is 0.
`isAllocated()` means memory is allocated.

zhoonit: Looks like there is actually three things to be done.

1. setting tensor dimension (it's not initialize nor allocate)
2. initialize()
3. allocate()

If we will to delay the initialize(), this has to be done with a state and we shouldn't always initialize(). and could you check if  `empty()` have to throw at `allocate()`?

```cpp
void initialize() {
  if (unallocated() ) { 
    this->book_initialize = true; /// or throw
  }
}

void allocate() {
  if(this->book_initialize) {
    initialize();
  }
}
```

Other option might be fixating the order?

nntrainer/tensor/var_grad.cpp
zhoonit: Cool

nntrainer/tensor/tensor.cpp
kparichay: The delay of initialization is done with the allocation. Once allocation is done, initialization will always be done.
In default case, the initialization is NONE, which results in no-op.

zhoonit: I mean even if there is a initializer is not NONE, it shouldn't be done on allocation, it has to be done only when requested.

kparichay: In case of lazy allocation, requesting the initialization again later is a burden for the caller.

>  it shouldn't be done on allocation, it has to be done only when requested.

can you expand why it shouldn't be done on allocation and only when requested?

zhoonit:
```cpp
TEST(nntrainer_Tensor, initialize_03_p) {
  nntrainer::Tensor t({1, 2, 3, 4}, false,
                      nntrainer::Tensor::Initializer::ONES);
  t.allocate();

  nntrainer::Tensor golden(1, 2, 3, 4);
  golden.setValue(1);

  EXPECT_EQ(golden, t);
}
```

above didn't call `initialize()` but allocated.

case 1. if initialize() only takes places when allocating, we don't need to call `initialize()` but can be private.
case 2. if initialize() takes multiple times (say, every iteration), some line of code will contain `t.initialize()`, so we will be calling initialize() twice for the first iteration.

Moreover, if we don't initialize but allocate implicitly call initialize(I mean not delayed initialize() but we don't call initialize) this will have same effect as #258 points out

kparichay: > above didn't call initialize() but allocated + initialized.

The code requested a tensor which must be initialized with ONES when accessed. So, the above need not call initialize explicitly.

The initialize is called every time with allocating, and only once by design. This ensures that once the tensor is allocated, it is initialized if needed (refer to https://github.com/nnstreamer/nntrainer/pull/1031#discussion_r597551354 for more details).

> case 1. if initialize() only takes place when allocating, we don't need to call initialize() so it can be private.

initialize takes place when allocating doesn't mean it can't be done again. `initialize()` is public for users to re-initialize their tensors whenever they want.

> case 2. if initialize() takes multiple times (say, every iteration), some line of code will contain t.initialize(), so we will be calling initialize() twice for the first iteration.

This is not by design for tensors but can be done by the caller. If a tensor is to be initialized every iteration and the design is leading to double initialization, use the below sample code:
```
Tensor t = Tensor({SHAPE}, false, Tensor::Initialize::NONE);
.....
t.allocate();
.....
for (iter : iterations)
  t.inialize(Tensor::Iinitializer::ONES);
```

> Moreover, if we don't initialize but allocate implicitly call initialize(I mean not delayed initialize() but we don't call initialize) this will have same effect as #258 points out

#258 had an overhead of zeroing out every tensor. But the current default initialization is NONE. So, allocate implicitly call initialize will have no overhead. I dont understand how these two are related.

zhoonit: Seeing https://github.com/nnstreamer/nntrainer/pull/1031#discussion_r597551354 I got the intention :) So it's basically RAII to some extent.

I'd like to suggest hiding `initialize()` but leaving `initialize(Tensor::Initializer initializer)` to only allow below cases.

```cpp
Tensor t = Tensor({SHAPE}, false, Tensor::Initialize::NONE);
.....
t.allocate();
.....
for (iter : iterations)
  t.inialize(Tensor::Iinitializer::ONES);
```

kparichay: > I'd like to suggest hiding initialize() 

This is open for the purpose of gradient, to be able to zero it without having to explicitly use its initializer or storing it somewhere.

zhoonit: Cool

PR #1459, opened 2021-08-03
[test] capi unittest fix

kparichay:
C-API unittest was checking for throw but instead should be checking for
return value. This patch fixes the test.
Further, the unittests tries to free unallocated tensordim in error
cases. This patch resolves it too.

### DIFF
--- a/test/tizen_capi/unittest_tizen_capi.cpp
+++ b/test/tizen_capi/unittest_tizen_capi.cpp
@@ -1081,32 +1081,20 @@ TEST(nntrainer_capi_nnmodel, get_input_output_dimension_03_n) {
   status = ml_train_model_construct_with_conf(s.getIniName().c_str(), &handle);
   EXPECT_EQ(status, ML_ERROR_NONE);
 
-  EXPECT_THROW(ml_train_model_get_input_tensors_info(handle, &input_info),
-               std::invalid_argument);
-  EXPECT_THROW(ml_train_model_get_output_tensors_info(handle, &output_info),
-               std::invalid_argument);
-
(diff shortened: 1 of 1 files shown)


zhoonit (approved):
Nice catch!

test/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -1081,32 +1081,20 @@ TEST(nntrainer_capi_nnmodel, get_input_output_dimension_03_n) {
   status = ml_train_model_construct_with_conf(s.getIniName().c_str(), &handle);
   EXPECT_EQ(status, ML_ERROR_NONE);
 
-  EXPECT_THROW(ml_train_model_get_input_tensors_info(handle, &input_info),
-               std::invalid_argument);
-  EXPECT_THROW(ml_train_model_get_output_tensors_info(handle, &output_info),
-               std::invalid_argument);
-
-  status = ml_tensors_info_destroy(input_info);
```
jijoongmoon, line 1089: I guess we need to add comment which shows that destroy is only required when the getting info is success.

jijoongmoon (approved):
LGTM except minor comments

test/tizen_capi/unittest_tizen_capi.cpp
kparichay: @zhoonit  does it has to be a separete PR to be included for ACR? or it can be done here?

kparichay: checked with @zhoonit offline, the comments for the destroy info will be added as a separate PR.

Issue #1127, opened 2021-04-21
Memory optimization for Training and Inference


kparichay:
Memory optimization done by the manager works well for a sequential model. However, it does not work very well with model with multiple inputs and outputs (like resnet, etc). 
Memory optimization needs to
1. work with the graph than working with topologically sorted version of it
2. Memory optimization needs to use producer-consumer model, the producer is the GraphNode making the output, and consumer is the node which consumes that tensor as an input. With producer-consumer modelling, a tensor once produced must stay alive till its consumer has not consumed it.

kparichay:
More details are required to avoid fragmentation of memory in the common memory pool. Will add these details soon.

jijoongmoon:
In oder to do it, we need to check if we can get the multiple input and multiple output. That means multiple loss. So far,  we didn't considered it. I guess, we also need to support run the inference with multi-threaded ( not layer level, I mean graph level ). Then, the memory management will be more complicated.

kparichay:
For each layer operation, we create a buffer lifespan object containing start time, and end time.
- start time: the buffer must be available for use for this time
- end time: the buffer has been consumed at this point and is available for reuse
The time is set based on the topological sorting of the layers to be run. For example, consider a 3 layer model, L1 (fc layer), L2 (fc layer), and L3 (addition layer). So, in this case:
- L1 executes from time 0 to 1: In1 is the input buffer (start:0, end:1), out1 is the output buffer (start:1, end:3)
- L2 executes from time 1 to 2: In2 is the input buffer (start:1, end:2), out2 is the output buffer (start:2, end:3)
- L3 executes from time 2 to 3: out1 and out2 is the input buffers (start:1, end:3 for out1 and start:2, end:3 for out2), out3 is the output buffer (start:2, end:3)
In this case, we have 4 buffers with their lifespans set. Once we get lifespans in general for all input/output buffers, the memory optimization has been reduced to an interval overlap problem, which is easy to solve with a memory pool.

kparichay:
For expanding the above for training, assume a backwarding graph expanded outward, and backwarding can be assumed to be like forwarding with extra buffers (for gradients/derivatives), and same memory optimization will work for backwarding as well.

kparichay:
Preparation:
1. Given a GraphNode, get access to the corresponding input GraphNodes, output GraphNodes, its index in the sorted list of nodes (which is the time for execution)
[...]

PR #1462, opened 2021-08-04
[graph] Graph cleanup and prepare for memory optimization

kparichay:
This patch provides graph cleanup and preparation for memory optimizations:
- Remove layerv2 initialize code for network graph.
- Currently each graph node has two identifiers - name and idx - where
both were unique. This patch removes the idx as the indentifier.
The corresponding functions are also removed.
In order to support fast fetching of nodes, graphcore has a map nodes.
- This change revealed a bug with the names in the code. As ini does not
allow case sensitive names, the names from the ini are changed to lower
case. However, input_layers property names were allowed to be
case-sensitive. Further names set by the API were also case sensitive.
This patch forces all such names to be case-insensitive to lowercase.
- Update the check for checking if a given node is input node.
The check uses in-degree for the node to be 0.

See Also #1127

### DIFF
--- a/nntrainer/graph/graph_core.cpp
+++ b/nntrainer/graph/graph_core.cpp
@@ -114,7 +114,6 @@ void GraphCore::ensureName(GraphNode &node, const std::string &prefix_,
                            const std::string &postfix_, bool force_rename) {
   auto to_lower = [](const std::string &str) -> std::string {
     std::string ret = str;
-    ;
     std::transform(ret.begin(), ret.end(), ret.begin(),
                    [](unsigned char c) { return std::tolower(c); });
     return ret;
--- a/nntrainer/graph/graph_core.cpp
+++ b/nntrainer/graph/graph_core.cpp
@@ -23,26 +23,17 @@
 namespace nntrainer {
 
 void GraphCore::addGraphNode(std::shared_ptr<GraphNode> node) {
-  node->setIndex(node_list.size());
   node_list.push_back(node);
+  node_map[node->getName()] = node_list.size() - 1;
 }
 
 const std::shared_ptr<GraphNode> &GraphCore::getNode(unsigned int ith) const {
-  if (ith >= size())
-    throw std::invalid_argument("Exceed total number of nodes");
-
-  if (node_list[ith]->getIndex() != ith)
-    throw std::runtime_error("Graph internal index mismatch");
-
-  return node_list[ith];
+  return node_list.at(ith);
 }
 
 const std::shared_ptr<GraphNode> &
 GraphCore::getSortedNode(unsigned int ith) const {
-  if (ith >= Sorted.size())
-    throw std::invalid_argument("Exceed total number of nodes");
-
-  return Sorted[ith];
+  return Sorted.at(ith);
 }
 
 void GraphCore::makeAdjacencyList(
@@ -55,7 +46,7 @@ void GraphCore::makeAdjacencyList(
   /** make the connections */
   for (auto &node : node_list) {
     for (auto const &in_conn : node->getInputConnections()) {
-      unsigned int to_node_id = getNode(in_conn)->getIndex();
+      unsigned int to_node_id = getNodeIdx(in_conn);
       adj[to_node_id].push_back(node);
     }
   }
@@ -69,7 +60,7 @@ void GraphCore::topologicalSortUtil(
 
   std::list<std::shared_ptr<GraphNode>>::iterator i;
   for (i = adj[ith].begin(); i != adj[ith].end(); ++i) {
-    auto index = (*i)->getIndex();
+    auto index = getNodeIdx((*i)->getName());
     if (!visited[index])
       topologicalSortUtil(adj, index, visited, dfs_stack);
   }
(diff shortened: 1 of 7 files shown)


nntrainer/graph/graph_core.cpp
```diff
@@ -126,14 +110,26 @@ void GraphCore::addNode(std::shared_ptr<GraphNode> node, bool ensure_name) {
...
 }
 
-void GraphCore::ensureName(GraphNode &node, const std::string &prefix,
-                           const std::string &postfix, bool force_rename) {
-  std::string orig_name = node.getName();
+void GraphCore::ensureName(GraphNode &node, const std::string &prefix_,
+                           const std::string &postfix_, bool force_rename) {
+  auto to_lower = [](const std::string &str) -> std::string {
+    std::string ret = str;
+    ;
```
zhoonit, line 117: Additional semicolon?

nntrainer/graph/graph_core.h
```diff
@@ -290,6 +294,14 @@ class GraphCore {
...
   void
   makeAdjacencyList(std::vector<std::list<std::shared_ptr<GraphNode>>> &adj);
+
+  /**
+   * @brief     Get index of the node with given name
+   *
+   * @param     name Name of the node
+   * @return    internal index of the node
+   */
+  unsigned int getNodeIdx(const std::string &name);
```
zhoonit, line 304: We still have the index but it's accessible from the name, am I correct?

nntrainer/graph/network_graph.cpp
```diff
@@ -702,14 +701,11 @@ int NetworkGraph::initialize(std::shared_ptr<Manager> manager) {
...
-             ++j) {
-          if (in_layer_node->getOutputLayers()[j] == lnode->getName()) {
-            location = j;
-            break;
-          }
-        }
+        auto const &in_layer_out_connect = in_layer_node->getOutputLayers();
+        unsigned int location =
+          std::find(in_layer_out_connect.begin(), in_layer_out_connect.end(),
+                    lnode->getName()) -
```
zhoonit, line 705: how about adding logic to check if the layer is found?

nntrainer/layers/layer_node.h
```diff
@@ -369,7 +356,17 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    * @param layers Name of the layers
    */
   void setInputLayers(const std::vector<std::string> &layers) {
-    input_layers = layers;
+    auto to_lower = [](const std::string &str) -> std::string {
+      std::string ret = str;
+      ;
```
zhoonit, line 361: additonal semicolon?

zhoonit (approved):
LGTM with minor suggestion

nntrainer/graph/graph_core.cpp
kparichay: will remove.

nntrainer/graph/graph_core.h
kparichay: the index is only for internal usage. The index is an index given to a node, but index to the location where the node exists in the structure (adjacency list in this case).

nntrainer/layers/layer_node.h
kparichay: will remove.

nntrainer/graph/network_graph.cpp
kparichay: Will add this check in DEBUG mode.

nntrainer/graph/graph_core.h
zhoonit: I see

nntrainer/graph/graph_core.cpp
kparichay: done.

nntrainer/graph/network_graph.cpp
kparichay: done.

nntrainer/graph/graph_core.h
```diff
@@ -261,6 +264,7 @@ class GraphCore {
   std::vector<std::shared_ptr<GraphNode>> output_list;
   std::vector<std::shared_ptr<GraphNode>>
     node_list;                                    /**< Unordered Node List  */
+  std::unordered_map<std::string, int> node_map;  /**< Unordered Node map  */
```
jijoongmoon, line 267: what's the usage difference between node_names and node_map? I wonder if the node_names is removed.

jijoongmoon (approved):
LGTM except minor comment

nntrainer/graph/graph_core.h
kparichay: node_map provides the location of the node in the given list of nodes for faster access. node_names provides assurance that each names in node are unique.

You are right in thinking about their overlap. I think the functionality to ensure all names are unique can be done just with node_map. I will do this in a new PR soon.

kparichay: Now as you mention it, maybe node_map and node_list can be combined as well.
In a separate PR, I will combine as many as of these structures.

jijoongmoon: great. thanks for your consideration.

PR #1464, opened 2021-08-04
[layers/manager] Set tensor to usage mapping

kparichay:
This patch creates a var_grad/weight to usage index mapping for both forwarding and backwarding.
These indexes will be used by the manager and the memory allocator for understanding the overlapping use of tensors.
The changes to achieve are (summarized per commit):
- Move manager from NeuralNetwork class to NetworkGraph class.
Some corresponding minor adjustments are also added.
- Update the layers to make the requested weights and tensors names to be
unique. This done using the layer name with the postfix of the
weights/tensors name making the combination unqiue as layer name is
unique.
InitLayerContext is updated to provide access to the layer name while
finalizing the layer.
- Set default usage time index for all the tensors based on when the layer
is to be run in the sorted list.
The usage time indexes are stored in each GraphNode by the GraphCore.
When creating the sorted list, each layer is updated to store its
order location/index when executed in this sequence for forwarding as
well as backwarding.
- This patch adds the setup to catch lifespan set by the layer developer
when requesting for tensors.
Further, the execution order now consists of 3 values - fowarding,
calcGradient and calcDerivative. The memory optimization will now be
done using the 3 functions to be called in order.

NOTE: the time indexes for variable and gradient will be set individually by the manager. 
The rule followed for this will be that each variable will be made available for forward as well as backward, but gradient will be made available only during backward.

See Also #1127

### DIFF
--- a/nntrainer/graph/graph_core.cpp
+++ b/nntrainer/graph/graph_core.cpp
@@ -89,6 +89,8 @@ void GraphCore::topologicalSort() {
 
   while (dfs_stack.empty() == false) {
     Sorted.push_back(dfs_stack.top());
+    Sorted.back()->setExecLoc(
+      {Sorted.size(), (node_list.size() * 2) - Sorted.size() + 1});
     dfs_stack.pop();
   }
 
--- a/nntrainer/graph/graph_core.cpp
+++ b/nntrainer/graph/graph_core.cpp
@@ -89,8 +89,6 @@ void GraphCore::topologicalSort() {
 
   while (dfs_stack.empty() == false) {
     Sorted.push_back(dfs_stack.top());
-    Sorted.back()->setExecLoc(
-      {Sorted.size(), (node_list.size() * 2) - Sorted.size() + 1});
     dfs_stack.pop();
   }
 
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -60,6 +60,7 @@ int NetworkGraph::compile(const std::string &loss_type) {
   graph.topologicalSort();
 
   countNonTrainableLayersAtBegin();
+  setExecutionOrder();
 
   status = checkCompiledGraph();
   NN_RETURN_STATUS();
@@ -67,6 +68,21 @@ int NetworkGraph::compile(const std::string &loss_type) {
   compiled = true;
 
   return status;
+}
+
+void NetworkGraph::setExecutionOrder() {
+  auto node_count = graph.size();
+  /** @todo: remove backwarding count for non-trainble layers */
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    auto &node = *iter;
+    auto order_idx = iter - cbegin();
+    auto forward_order = order_idx;
+    auto calc_gradient_order = (node_count * 3) - (order_idx * 2) + 1;
+    /** calc derivative is called right after calc_gradient */
+    auto calc_derivative_order = calc_gradient_order + 1;
+    node->setExecutionOrder(
+      {forward_order, calc_gradient_order, calc_derivative_order});
+  }
 }
 
 void NetworkGraph::updateConnectionName(const std::string &from,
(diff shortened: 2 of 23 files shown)


zhoonit:
Could you clarify how many commits should be reviewed in this PR?

kparichay:
rebased.

test/unittest/layers/layers_standalone_common_tests.cpp
```diff
@@ -84,7 +84,7 @@ TEST_P(LayerSemantics, gettersValidate_p) {
 TEST_P(LayerSemantics, setBatchValidate_p) {
   nntrainer::TensorDim in_dim({1, 1, 1, 1});
   nntrainer::InitLayerContext init_context =
-    nntrainer::InitLayerContext({in_dim}, 1);
+    nntrainer::InitLayerContext({in_dim}, 1, "layer");
```
zhoonit, line 87: just an idea, model file might want to check for those kind of invariants: every tensor name has to be unique

zhoonit (approved):
LGTM

nntrainer/layers/layer_node.cpp
```diff
@@ -351,6 +351,8 @@ void LayerNode::finalize() {
   if (finalized)
     throw std::runtime_error("Finalizing a layer which is already finalized");
 
+  init_context = InitLayerContext(init_context.getInputDimensions(),
+                                  init_context.getNumOutputs(), getName());
```
jijoongmoon, line 356: I wonder layer_node has name and init_context has same name. So.. it might be possible to remove one of them?

nntrainer/graph/graph_core.cpp
```diff
@@ -89,6 +89,8 @@ void GraphCore::topologicalSort() {
 
   while (dfs_stack.empty() == false) {
     Sorted.push_back(dfs_stack.top());
+    Sorted.back()->setExecLoc(
+      {Sorted.size(), (node_list.size() * 2) - Sorted.size() + 1});
```
jijoongmoon: May be add the comment why Sorted.size() and (node_list.size() * 2) - Sorted.size() + 1? it looks like forwarding order and backwarding.

nntrainer/graph/network_graph.cpp
```diff
@@ -69,6 +70,21 @@ int NetworkGraph::compile(const std::string &loss_type) {
...
 
+void NetworkGraph::setExecutionOrder() {
+  auto node_count = graph.size();
+  /** @todo: remove backwarding count for non-trainble layers */
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    auto &node = *iter;
+    auto order_idx = iter - cbegin();
+    auto forward_order = order_idx;
+    auto calc_gradient_order = (node_count * 3) - order_idx + 1;
+    /** calc derivative is called right after calc_gradient */
```
jijoongmoon, line 81: I'm not sure but. calc_gradient_order might be node_count + 2*(order_idx)+1

test/unittest/layers/layers_standalone_common_tests.cpp
kparichay: When requesting weights, each weight name is ensured to be unique by the manager for all its tensor. Its added in #1487.

nntrainer/layers/layer_node.cpp
kparichay: `init_context` provides access to the name of the layer node to its underneath Layer::finalize method.

nntrainer/graph/network_graph.cpp
kparichay: Yes, you are right. I will update this.

nntrainer/layers/layer_node.cpp
jijoongmoon: if it is, then great. Thanks.

nntrainer/graph/network_graph.cpp
kparichay: done.

jijoongmoon (approved):
LGTM

PR #1465, opened 2021-08-04
[mesonbuild] update meson build

lhs8928:
- Make yocto build option
 - Check sysroot

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,11 +14,9 @@ add_project_arguments('-DMIN_CPP_VERSION=201703L', language:['c','cpp'])
 
 cc = meson.get_compiler('c')
 cxx = meson.get_compiler('cpp')
-build_platform = ''
 
-if get_option('enable-tizen')
+if get_option('platform') == 'tizen'
   # Pass __TIZEN__ to the compiler
-  build_platform = 'tizen'
   add_project_arguments('-D__TIZEN__=1', language:['c','cpp'])
 
   if get_option('enable-tizen-feature-check')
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -25,6 +26,13 @@ if get_option('enable-tizen')
     add_project_arguments('-D__FEATURE_CHECK_SUPPORT__', language: ['c', 'cpp'])
   endif
 endif
+if get_option('enable-yocto')
+  if build_platform == ''
+    build_platform = 'yocto'
+  else
+    error(multi_build_platform_err_msg_template.format(build_platform, 'yocto'))
```
jijoongmoon, line 33: This is little bit strange. is there no case in which the build_paltform='yocto' ?? If it is, then there will be error.

jijoongmoon (approved):
LGTM except for small comments.

meson.build
myungjoo: Logic of 29~35 is overly complex and is not general (extensible). Simplify it.

```
count = 0
if TIZEN
   ...
   count ++
endif
if YOCTO
   ...
   count ++;
endif

if (count > 1) error
```
should be enough.

meson_options.txt
```diff
@@ -1,4 +1,5 @@
 option('enable-tizen', type: 'boolean', value: false)
+option('enable-yocto', type: 'boolean', value: false)
```
myungjoo, line 2: Refer to https://mesonbuild.com/Build-options.html
```
option('platform', type: 'combo', choices: ['none', 'tizen', 'yocto'], value: 'none')
```
might be more pretty.

myungjoo (approved):
LGTM.

PR #1470, opened 2021-08-09
[Dataset] Implement samplewise producer for random

zhoonit:
- [Dataset] Implement samplewise producer for random

```
This patch implements samplewise producer for random and its
corresponding test abstraction

**Self evaluation:**

```

### DIFF
--- a/nntrainer/dataset/random_data_producers.cpp
+++ b/nntrainer/dataset/random_data_producers.cpp
@@ -78,6 +78,11 @@ const std::string RandomDataOneHotProducer::getType() const {
   return RandomDataOneHotProducer::type;
 }
 
+bool RandomDataOneHotProducer::isMultiThreadSafe() const {
+  /// @todo make this true, it is needed to test multiple worker scenario
+  return false;
+}
+
 unsigned int
 RandomDataOneHotProducer::size(const std::vector<TensorDim> &input_dims,
                                const std::vector<TensorDim> &label_dims) const {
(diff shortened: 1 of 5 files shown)


nntrainer/dataset/random_data_producers.cpp
```diff
@@ -169,4 +171,74 @@ RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
...
+
+    auto populate_label =
+      [&](Tensor *t, std::uniform_int_distribution<unsigned int> &label_dist_) {
+        t->setZero();
+        t->setValue(0, 0, 0, label_dist_(rng), 1);
+        return t;
+      };
+
+    auto &[inputs, labels] = sample;
+    std::transform(inputs.begin(), inputs.end(), inputs.begin(),
```
kparichay, line 230: I recommend to use `std::for_each` here as you intend to perform the function on each element in-place. WIth this you dont need to return in your lambdas and it also simplifies the usage.

nntrainer/dataset/data_producers.h
```diff
@@ -77,25 +97,63 @@ class DataProducer {
...
+   * @remark this function must assume that the batch dimension of each tensor
+   * dimension is one. If actual dimension is not one, this function must ignore
+   * the batch dimension and assume it to be one.
+   * @param input_dims input dimensions.
+   * @param label_dims label dimensions.
+   * @return Generator generator is a function taht generates a sample upon
+   * call.
+   */
+  virtual Generator_sample
+  finalize_sample(const std::vector<TensorDim> &input_dims,
```
kparichay, line 124: i recommend to name this `finalize`, and rename deprecated version to `finalize_batch`.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -169,4 +171,74 @@ RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
...
+  NNTR_THROW_IF(min_.get() > max_.get(), std::invalid_argument)
+    << "Min value is bigger then max value, min: " << min_.get()
+    << "max: " << max_.get();
+
+  /// @todo move this to higher order component
+  NNTR_THROW_IF(size(input_dims, label_dims) == 0, std::invalid_argument)
+    << "size is zero, dataproducer does not provide anything";
+
+  /** prepare states for the generator */
+  std::vector<std::uniform_int_distribution<unsigned int>> label_chooser_;
```
kparichay, line 200: if I am not wrong, this might not be MT safe.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -169,4 +171,74 @@ RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
...
+  label_chooser_.reserve(label_dims.size());
+  std::transform(label_dims.begin(), label_dims.end(),
+                 std::back_inserter(label_chooser_),
+                 [this](const TensorDim &label_dim) {
+                   return std::uniform_int_distribution<unsigned int>(
```
kparichay, line 205: if I am not wrong, this might not be MT safe.

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -169,4 +171,74 @@ RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
...
+      [&](Tensor *t, std::uniform_int_distribution<unsigned int> &label_dist_) {
+        t->setZero();
+        t->setValue(0, 0, 0, label_dist_(rng), 1);
+        return t;
+      };
+
+    auto &[inputs, labels] = sample;
+    std::transform(inputs.begin(), inputs.end(), inputs.begin(),
+                   populate_input);
+    std::transform(labels.begin(), labels.end(), label_chooser.begin(),
```
kparichay, line 232: verify that using label_chooser here is MT safe.

nntrainer/dataset/data_producers.h
```diff
@@ -43,12 +44,31 @@ class DataProducer {
...
+
+  /**
+   * @brief generator callable type which will fill a sample
+   * @todo rename this to Generator.
+   * @param[out] Sample @see DataProducer::Sample for detail, inside this Sample
+   * must contain allocated tensor before fed to this function.
+   * @return bool true if this is the last sample.
+   *
+   */
+  using Generator_sample = std::function<bool(Sample &)>;
```
kparichay: consider directly using `std::vector<Tensor *>, std::vector<Tensor *>` or update the `size()` to have a similar tuple based interface.

nntrainer/dataset/data_producers.h
```diff
@@ -77,25 +97,63 @@ class DataProducer {
...
+   * @remark this function must assume that the batch dimension of each tensor
+   * dimension is one. If actual dimension is not one, this function must ignore
+   * the batch dimension and assume it to be one
+   * @param input_dims input dimensions
+   * @param label_dims label dimensions
+   *
+   * @return size calculated size
+   */
+  virtual unsigned int
+  size_sample(const std::vector<TensorDim> &input_dims,
```
kparichay, line 158: update size_sample to more appropriate name - getDataSize() or getNumSamples()

nntrainer/dataset/data_producers.h
zhoonit: `finalize_batch` will be deleted within two days, I'll keep this to not confuse the reviewers (from previous commits that is not merged) but rename it at the end of the work :)

nntrainer/dataset/data_producers.h
zhoonit: This will be renamed to `DataProducer::size()` after rework is done :)

nntrainer/dataset/random_data_producers.cpp
zhoonit: Good point, it is not thread safe :)

nntrainer/dataset/data_producers.h
zhoonit: Sample will be a class onward.
I guess we can use ConnectionSpecs  (or SampleSpec if we need) later on for the size query.

nntrainer/dataset/random_data_producers.cpp
zhoonit: Done

nntrainer/dataset/data_producers.h
zhoonit: But I think it's better to pass std::vector<Tensor *>, std::vector<Tensor *> here. I'll change.

nntrainer/dataset/data_producers.h
kparichay: alright.

nntrainer/dataset/data_producers.h
kparichay: ok.

nntrainer/dataset/data_producers.h
kparichay: `DataProducer::size()` with or  without arguments?

zhoonit: it's with arguments.

nntrainer/dataset/data_producers.h
zhoonit: Done

nntrainer/dataset/random_data_producers.cpp
```diff
@@ -169,4 +174,70 @@ RandomDataOneHotProducer::finalize(const std::vector<TensorDim> &input_dims,
     return DataProducer::Iteration(false, inputs, labels);
   };
 }
+
+DataProducer::Generator_sample RandomDataOneHotProducer::finalize_sample(
+  const std::vector<TensorDim> &input_dims,
```
jijoongmoon, line 179: I thought it is also similar to layer implementation. we have this, ```finailze_sample()``` with ```finalize()``` is because we have both sample based and batch based implementation? and it will be the same for other producers as well?

jijoongmoon (approved):
LGTM other than minor comments

kparichay (approved):
LGTM

nntrainer/dataset/random_data_producers.cpp
zhoonit: `finalize()` will be deleted and `finalize_sample()` will be renamed to `finalize()`

The reason being we have two function is to make a gradual change :)

PR #1473, opened 2021-08-10
[Dataset] Rework func dataset to samplewise

zhoonit:
- [Dataset] Rework func dataset to samplewise

```
- Fix bugs in Tensor::Map
- Add shortcut for the Tensor::updateBatch
- Add func_data_producer with sample

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/split_layer.cpp
+++ b/nntrainer/layers/split_layer.cpp
@@ -106,12 +106,14 @@ void SplitLayer::forwarding(RunLayerContext &context, bool training) {
     output_.reshape(output_reshape_helper);
 
     for (unsigned int batch = 0; batch < input_.batch(); batch++) {
-      const Tensor source_tensor = Tensor::Map(
-        input_.getAddress(batch, 0, idx, 0), input_reshape_helper.width(),
-        {1, 1, 1, input_reshape_helper.width()});
-      Tensor dest_tensor = Tensor::Map(
-        output_.getAddress(batch, 0, 0, 0), output_reshape_helper.width(),
-        {1, 1, 1, output_reshape_helper.width()});
+      const Tensor source_tensor =
+        Tensor::Map(input_.getAddress(batch, 0, idx, 0),
+                    input_reshape_helper.width() * sizeof(float),
+                    {1, 1, 1, input_reshape_helper.width()});
+      Tensor dest_tensor =
+        Tensor::Map(output_.getAddress(batch, 0, 0, 0),
+                    output_reshape_helper.width() * sizeof(float),
+                    {1, 1, 1, output_reshape_helper.width()});
       dest_tensor.copy(source_tensor);
     }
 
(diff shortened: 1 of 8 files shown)


zhoonit:
resolves #1474

nntrainer/tensor/tensor.cpp
```diff
@@ -501,7 +501,8 @@ Tensor Tensor::getSharedDataTensor(const TensorDim dim_, unsigned int offset,
                                    bool reset_stride) const {
   Tensor ret = *this;
 
-  if (dim_.getDataLen() + offset > dim.getDataLen())
+  if (dim_.getDataLen() * sizeof(float) + offset >
```
kparichay, line 504: Good catch.
you can update to using `bytes()` after #1452 :)

nntrainer/tensor/tensor.cpp
```diff
@@ -170,7 +170,7 @@ Tensor Tensor::Map(float *buf, unsigned int size, const TensorDim &d,
       "[Tensor::Map] empty tensor dim is not allowed");
   }
 
-  if (d.getDataLen() + offset > size) {
+  if (d.getDataLen() * sizeof(float) + offset > size) {
```
kparichay, line 173: it would be better to rename size to bytes or size_bytes to avoid any confusion.

kparichay, line 173: It might be better to rename size to bytes to avoid confusion.

nntrainer/layers/split_layer.cpp
```diff
@@ -133,12 +135,14 @@ void SplitLayer::calcDerivative(RunLayerContext &context) {
...
-      const Tensor source_tensor = Tensor::Map(
-        output_.getAddress(batch, 0, 0, 0), output_reshape_helper.width(),
-        {1, 1, 1, output_reshape_helper.width()});
+      Tensor dest_tensor =
+        Tensor::Map(input_.getAddress(batch, 0, idx, 0),
+                    input_reshape_helper.width() * sizeof(float),
+                    {1, 1, 1, input_reshape_helper.width()});
+      const Tensor source_tensor =
+        Tensor::Map(output_.getAddress(batch, 0, 0, 0),
+                    output_reshape_helper.width() * sizeof(float),
```
kparichay, line 144: This works but as Tensor only deals with float, you can also consider updating Map to take offset as elements than bytes as it's a bit overhead to calculate bytes everywhere.

zhoonit: I think it's better to make the offset as elements, this is more like a bugfix rather then an enhancement, so let's deal this in different PR.

nntrainer/tensor/tensor.cpp
zhoonit: Done :)

nntrainer/tensor/tensor.cpp
zhoonit: Done

jijoongmoon (approved):
LGTM

nntrainer/layers/split_layer.cpp
```diff
@@ -133,12 +135,14 @@ void SplitLayer::calcDerivative(RunLayerContext &context) {
...
     for (unsigned int batch = 0; batch < input_.batch(); batch++) {
-      Tensor dest_tensor = Tensor::Map(input_.getAddress(batch, 0, idx, 0),
-                                       input_reshape_helper.width(),
-                                       {1, 1, 1, input_reshape_helper.width()});
-      const Tensor source_tensor = Tensor::Map(
-        output_.getAddress(batch, 0, 0, 0), output_reshape_helper.width(),
-        {1, 1, 1, output_reshape_helper.width()});
+      Tensor dest_tensor =
+        Tensor::Map(input_.getAddress(batch, 0, idx, 0),
+                    input_reshape_helper.width() * sizeof(float),
```
kparichay, line 140: replace `sizeof(float)` with `sizeof(input_.getValue(0))` or `create tensor.sizeElement()`.

kparichay (approved):
LGTM other than the minor comment.

lhs8928 (approved):
LGTM

nntrainer/layers/split_layer.cpp
zhoonit: we can have a tensor size type defined as a template meta function. using decltype. I'll do this as a separate PR.

nntrainer/layers/split_layer.cpp
zhoonit: we can have a tensor size type defined as a template meta function. using decltype. because sizeof(input_.getValue(0)) is runtime type inference. I'll do this as a separate PR.

zhoonit: https://github.com/nnstreamer/nntrainer/issues/1515

kparichay: cool.

PR #1475, opened 2021-08-10
[memory] Introduce memory pool and planner

kparichay:
- Simplify weight and gradient allocation out to the most basic strategy
so that the memory optimization can be applied on it.
With this patch, the memory for gradient is not shared anymore till the
memory optimization is implemented.
- This patch introduces memory planner interface which will provide a
layout plan given the various memories required by the model.
Different memory allocations plans/strategies will be implemented by
extending this class with the given interface.
- This patch introduces memory pool for tensors. All the tensors requiring
memory can ask memory pool for their memory. The memory pool collects
all the requirements, uses the provided memory planner to plan a memory
layout, allocates the required memory, and then finally returns the
allocated memory to the requesters.

There are certain limitations to this memory pool:
1. All the memory must be requested before layout planning and
allocation.
2. There is no support for freeing selectively allocated tensors
3. In order to request support for more tensors and their memories,
all previously allocated memory has to be freed.
4. MemoryPool for now allocates all its required memory in a single
block. Support for breaking the required memory into smaller blocks is
left a TODO for later (this can be useful for devices without IOMMU).

See Also #1127

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -252,30 +252,28 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
  * @brief Allocate and initialize the weight variable
  */
 void Manager::initializeWeights() {
-  if (total_weight_size == 0) {
-    ml_logw("Nothing done on initialize because there is no weight registered");
-    return;
-  }
 
   if (weights_initialized)
     return;
 
-  AllocFunc allocate_weight = getAllocFunc(true);
-
-  unsigned int weight_offset = 0;
-
   if (LAYER_V2) {
     for (auto &l_w : weights_v2) {
       for (auto &w : l_w) {
-        const auto &dim = w->getDim();
-
-        Tensor weight_prealloc = allocate_weight(dim, weight_offset);
-        weight_offset += dim.getDataLen();
-
-        w->initializeVariable(weight_prealloc);
+        w->initializeVariable();
+        // tensor_map(&w->getVariableRef(), requestMemory(w.getDim().size(), 0,
+        // MAX));
       }
     }
   } else {
+    if (total_weight_size == 0) {
+      ml_logw(
+        "Nothing done on initialize because there is no weight registered");
+      return;
+    }
+
+    AllocFunc allocate_weight = getAllocFunc(true);
+
+    unsigned int weight_offset = 0;
     for (auto &l_w : weights) {
       for (auto &w : l_w) {
         Weight &weight = w.get();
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/manager.cpp
```diff
@@ -377,28 +375,39 @@ void Manager::deallocateGradients() {
...
+    for (auto &l_w : weights_v2) {
+      for (auto &w : l_w) {
+        w->initializeGradient();
+        // auto exec_order = tensor_exec_order[w.getName()];
+        // tensor_map(&w->getGradientRef(), requestMemory(w.getDim().size(),
+        //       std::get<1>(exec_order), std::get<2>(exec_order) + 1));
+      }
+    }
+  } else {
+    if (total_weight_size == 0) {
```
zhoonit, line 388: Is this code for the layer v1 tests?

If it is, could you run a test with layer v1, if this is working fine before merging this PR?

nntrainer/tensor/manager.cpp
```diff
@@ -808,7 +817,7 @@ Manager::requestWeights(const GraphNode &node,
 }
 
 /**
- * @brief     Create weights with the given spec
+ * @brief     Create weight4 with the given spec
```
zhoonit, line 820: typo?

nntrainer/tensor/memory_pool.h
```diff
@@ -0,0 +1,123 @@
...
+  unsigned int requestMemory(size_t bytes, unsigned int start_time,
+                             unsigned int end_time);
+
+  /**
+   * @brief Planner the layout with memory planner
+   *
+   * @param planner The memory planner to be used for finalizing the layout
+   *
+   * @return The total memory requirement using this memory planner
+   * @return The efficiency of the memory layer with the given memory planner
```
zhoonit, line 54: returning pair?

nntrainer/tensor/memory_pool.h
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * @details The efficiency of the planner is calculated as the ratio of the
+   * theoretical minimum memory requirement divided by the memory requirement
+   * given by the memory planner.
+   *
+   * @details planLayout can be called multiple times as this does not perform
+   * any allocation but rather just plans the layout and stores the layout.
+   * Subsequent call to this function will overwrite any existing layout.
+   */
+  double planLayout(const MemoryPlanner &planner);
```
zhoonit, line 64: I guess this has to be pointer for polymorphism?

nntrainer/tensor/manager.h
```diff
@@ -444,6 +445,7 @@ class Manager {
     tensor_exec_order; /**< stores the order/location at which a given tensor is
                         going to be executed when the network is forwarded and
                         backwarded */
+  // std::unordered_map<Tensor &, unsigned int> tensor_map;
```
zhoonit, line 448: Intended comment?

nntrainer/tensor/memory_planner.h
```diff
@@ -0,0 +1,53 @@
...
+   * @param[out] memory_offset The offset of each memory from the base of the
+   * allocated memory
+   * @return The total memory required as per this strategy
+   *
+   * @details The minimum offset will be 0, and the maximum offset will be less
+   * than the sum of the memory_size vector.
+   */
+  virtual size_t planLayout(
+    const std::vector<size_t> &memory_size,
+    const std::vector<std::pair<unsigned int, unsigned int>> &memory_validity,
```
zhoonit, line 47: looks like size and validity is clubbed. Perhaps, it'd be better to make a struct for this. (or not). I totally fine with current.

zhoonit (approved):
LGTM with unimportant comments.

nntrainer/tensor/memory_pool.h
kparichay: MemoryPlanner is a stateless object. So, it can be const reference which can be created on fly.

zhoonit: Yes, I mean to utilize virtual function shouldn't it be a pointer?

kparichay: i tested with BinaryPlanner and it works (check the next PR).

nntrainer/tensor/manager.h
kparichay: yes, this is intended for now

nntrainer/tensor/memory_planner.h
kparichay: I will club these data structures together once all the manager-related memories use this.

nntrainer/tensor/memory_pool.h
kparichay: line 53 is typo, will remove.

nntrainer/tensor/manager.cpp
kparichay: will update.

nntrainer/tensor/manager.cpp
kparichay: total_weight_size is used by layer v1 (where requestWeights() interface was used).
As this variable was not needed in layerv2 style, it is just moved in the else case. So, there is no change in behavior for layer-v1 and layer-v2 by bringing this in.

There is no layerv1 tests which directly deals with this.

zhoonit: I wonder the reason for the maintaining this section rather then deleting it.

nntrainer/tensor/memory_pool.h
zhoonit: Cool. Thanks :)

nntrainer/tensor/memory_planner.h
zhoonit: Awesome

nntrainer/tensor/manager.cpp
kparichay: It is kept to able to use this strategy in a manual strategy if the automated strategy is not as good (although I am quite sure that automated will do better than manual but I cant guarantee).

nntrainer/tensor/manager.cpp
kparichay: done.

nntrainer/tensor/memory_pool.h
kparichay: done.

jijoongmoon (approved):
LGTM

PR #1477, opened 2021-08-11
[MemoryPlanner] Basic memory planner with unittests

kparichay:
Add support for a basic memory planner along with its unittests:
- This patch adds support for the basic memory planner.
The basic memory planner does not provide any memory sharing, and
results in all tensors being allocated in separate memories.
This forms the baseline for memory allocation.
Other memory planners will be enhancement over this.
- This patch adds unittests for the memory planner.
The unittests covers some of the extreme scenarios for memory planning
and ensures that the constraints are satisfied.
Note that these tests do not test the efficient of the planning but the
validity of the planned layout.
Corresponding bug fixes are also added.

See Also #1127

### DIFF
--- a/nntrainer/tensor/basic_planner.h
+++ b/nntrainer/tensor/basic_planner.h
@@ -26,24 +26,33 @@ namespace nntrainer {
  * @details Basic planner performs no memory optimization and provides no memory
  * sharing
  */
-class BasicPlanner {
+class BasicPlanner : public MemoryPlanner {
 public:
   /**
    * @brief BasicPlanner destructor
    *
    */
-  BasicPlanner() = default;
+  ~BasicPlanner() = default;
 
   /**
    * @copydoc MemoryPlanner::planLayout(
    * const std::vector<size_t> &memory_size,
    * const std::vector<std::pair<unsigned int, unsigned int>> &memory_validity,
    * std::vector<size_t> &memory_offset);
+   *
    */
   size_t planLayout(
     const std::vector<size_t> &memory_size,
     const std::vector<std::pair<unsigned int, unsigned int>> &memory_validity,
-    std::vector<size_t> &memory_offset);
+    std::vector<size_t> &memory_offset) const;
+
+  /**
+   * @copydoc MemoryPlanner::getType() const
+   *
+   */
+  const std::string &getType() const { return type; }
+
+  inline static const std::string type = "basic_planner";
 };
 
 } // namespace nntrainer
--- a/nntrainer/tensor/basic_planner.h
+++ b/nntrainer/tensor/basic_planner.h
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file   basic_planner.h
+ * @date   11 August 2021
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is Naive Memory Planner
+ *
+ */
+
+#ifndef __BASIC_PLANNER_H__
+#define __BASIC_PLANNER_H__
+
+#include <vector>
+
+#include <memory_planner.h>
+
+namespace nntrainer {
+
+/**
+ * @class   BasicPlanner
+ * @brief   Basic Memory Planner provides the basic plan for memory layout
+ * @details Basic planner performs no memory optimization and provides no memory
+ * sharing
+ */
+class BasicPlanner {
+public:
+  /**
+   * @brief BasicPlanner destructor
+   *
+   */
+  BasicPlanner() = default;
+
+  /**
+   * @copydoc MemoryPlanner::planLayout(
+   * const std::vector<size_t> &memory_size,
+   * const std::vector<std::pair<unsigned int, unsigned int>> &memory_validity,
+   * std::vector<size_t> &memory_offset);
+   */
+  size_t planLayout(
+    const std::vector<size_t> &memory_size,
+    const std::vector<std::pair<unsigned int, unsigned int>> &memory_validity,
+    std::vector<size_t> &memory_offset);
+};
+
+} // namespace nntrainer
+
+#endif /** __BASIC_PLANNER_H__ */
--- a/nntrainer/tensor/memory_pool.cpp
+++ b/nntrainer/tensor/memory_pool.cpp
@@ -1,0 +1,262 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2021 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file   memory_pool.cpp
+ * @date   11 August 2021
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is Memory Pool Class
+ */
+
+#include <numeric>
+#include <vector>
+
+#include <memory_pool.h>
+#include <nntrainer_error.h>
+
+namespace nntrainer {
+
+/**
+ * @brief Request Memory from memory pool
+ * @note start_time is inclusive, but end_time is exclusive
+ */
+unsigned int MemoryPool::requestMemory(size_t bytes, unsigned int start_time,
+                                       unsigned int end_time) {
+  if (mem_pool != nullptr)
+    throw std::invalid_argument(
+      "Deallocate memory pool before requesting more memory");
+
+  if (end_time <= start_time)
+    throw std::invalid_argument(
+      "Invalid validity range for the requested memory");
+
+  memory_size.push_back(bytes);
+  memory_validity.push_back({start_time, end_time});
+
+  /** invalidate min_pool_size if already there */
+  min_pool_size = 0;
+
+  return memory_size.size() - 1;
+}
+
+/**
+ * @brief Planner the layout with memory planner
+ *
+ * @details The efficiency of the planner is calculated as the ratio of the
+ * theoretical minimum memory requirement divided by the memory requirement
+ * given by the memory planner.
+ *
+ * @details planLayout can be called multiple times as this does not perform
+ * any allocation but rather just plans the layout and stores the layout.
+ * Subsequent call to this function will overwrite any existing layout.
+ */
+double MemoryPool::planLayout(const MemoryPlanner &planner) {
+  if (mem_pool != nullptr)
+    /** mem_pool must be deallocated when planLayout is being called */
+    throw std::runtime_error("Planning memory layout after allocation");
+
+  if (memory_size.empty())
+    throw std::runtime_error("Planning memory layout for empty pool");
+
+  /** calculate min_pool_size if not already calculated */
+  if (min_pool_size == 0)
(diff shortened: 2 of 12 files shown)


nntrainer/tensor/basic_planner.h
```diff
@@ -0,0 +1,51 @@
...
+ * @details Basic planner performs no memory optimization and provides no memory
+ * sharing
+ */
+class BasicPlanner {
+public:
+  /**
+   * @brief BasicPlanner destructor
+   *
+   */
+  BasicPlanner() = default;
```
zhoonit, line 35: I guess this is ~BasicPlanner() ?

nntrainer/tensor/memory_pool.cpp
```diff
@@ -0,0 +1,258 @@
...
+/**
+ * @brief Planner the layout with memory planner
+ *
+ * @details The efficiency of the planner is calculated as the ratio of the
+ * theoretical minimum memory requirement divided by the memory requirement
+ * given by the memory planner.
+ *
+ * @details planLayout can be called multiple times as this does not perform
+ * any allocation but rather just plans the layout and stores the layout.
+ * Subsequent call to this function will overwrite any existing layout.
```
zhoonit, line 53: I like this :)

nntrainer/tensor/memory_pool.cpp
```diff
@@ -0,0 +1,258 @@
...
+    throw std::runtime_error("Planning memory layout for empty pool");
+
+  /** calculate min_pool_size if not already calculated */
+  if (min_pool_size == 0)
+    min_pool_size = calcMinMemoryRequirement();
+
+  pool_size = planner.planLayout(memory_size, memory_validity, memory_offset);
+  if (pool_size < min_pool_size || !validateLayout())
+    throw std::runtime_error("Planned layout is not feasible");
+
```
zhoonit, line 70: throw if min_pool size is zero? (more for svace...)

nntrainer/tensor/memory_pool.cpp
```diff
@@ -0,0 +1,258 @@
...
+  return double(pool_size) / double(min_pool_size);
```
zhoonit, line 71: throw if min_pool size is zero? (more for svace...)

nntrainer/tensor/memory_pool.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * @brief Do the allocation of memory
+ *
+ */
+void MemoryPool::allocate() {
+  if (pool_size == 0)
+    throw std::runtime_error("Allocating memory pool with size 0");
+
+  mem_pool = malloc(pool_size);
```
zhoonit, line 82: throw if malloc fails?

nntrainer/tensor/memory_pool.cpp
```diff
@@ -0,0 +1,258 @@
...
+ * @param e1 end of interval 1
+ * @param s2 start of interval 2
+ * @param e2 end of interval 2
+ *
+ * @return true if overlap else false
+ *
+ * @note overlap check assumes are start is inclusive and end is exclusive
+ */
+template <typename T> static bool overlap(T s1, T e1, T s2, T e2) {
+  if (e1 <= s2 || e2 <= s1)
```
zhoonit, line 161: 1. Super minor, maybe return condition itself?
2. if this is not in critical path, how about adding a check to throw if s1 < e1 and s2 < e2?

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -0,0 +1,173 @@
...
+constexpr unsigned int MEM_BYTES = 50;
+constexpr unsigned int MEM_QUANT = 100;
+constexpr unsigned int INTERVAL_SIZE = 5;
+
+void MemoryPlannerValidate::SetUp() {
+  plan_type = GetParam();
+
+  planner = nullptr;
+  if (plan_type == nntrainer::BasicPlanner::type)
+    planner = std::move(std::make_unique<nntrainer::BasicPlanner>());
```
zhoonit, line 29: It looks like given value is already rvalue so no need to std::move.

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+  EXPECT_TRUE(validateOverflow(memory_size, memory_offset, pool_size));
+  if (planner->getType() == nntrainer::BasicPlanner::type) {
+    EXPECT_EQ(pool_size,
+              std::accumulate(memory_size.begin(), memory_size.end(), 0));
+    EXPECT_TRUE(validateNoOverlap(memory_size, memory_offset, pool_size));
+  } else {
+    EXPECT_EQ(pool_size,
+              *std::max_element(memory_size.begin(), memory_size.end()));
+    EXPECT_TRUE(validateAllOverlap(memory_size, memory_offset));
```
zhoonit, line 137: I wonder what this is for, is this checking planner->planLayerout is noop if nothing has been done?

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -0,0 +1,173 @@
...
+    if (memory_offset[idx] + memory_size[idx] > pool_size)
+      return false;
+
+  return true;
+}
+
+/**
+ * @brief Test memory planning with full overlap of requested memories
+ */
+TEST_P(MemoryPlannerValidate, full_overlap) {
```
zhoonit, line 87: could you clarify by renaming validityIsFullyOverlapping?

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -0,0 +1,173 @@
...
+  return true;
+}
+
+/**
+ * @brief Validate the provided layout does completely overlap
+ */
+static bool validateAllOverlap(const std::vector<size_t> &memory_size,
+                               const std::vector<size_t> &memory_offset) {
+  for (unsigned int idx = 0; idx < memory_size.size(); idx++) {
+    EXPECT_EQ(memory_offset[idx], memory_offset[0]);
```
zhoonit, line 64: do we have to check size as well? as the name states all overlap?

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -0,0 +1,173 @@
...
+  std::uniform_int_distribution<unsigned int> dist_interval(1, INTERVAL_SIZE);
+
+  std::vector<size_t> memory_size(MEM_QUANT);
+  for (unsigned int idx = 0; idx < memory_size.size(); idx++) {
+    memory_size[idx] = dist(rng);
+  }
+
+  std::vector<std::pair<unsigned int, unsigned int>> memory_validity(MEM_QUANT);
+  for (unsigned int idx = 0; idx < memory_validity.size(); idx++) {
+    memory_validity[idx] = {idx, idx + dist_interval(rng)};
```
zhoonit, line 156: Should we need to have memory_validity is not given in order as planner should not expect it is given in order?

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+  EXPECT_TRUE(validateOverflow(memory_size, memory_offset, pool_size));
+  if (planner->getType() == nntrainer::BasicPlanner::type) {
+    EXPECT_EQ(pool_size,
+              std::accumulate(memory_size.begin(), memory_size.end(), 0));
+    EXPECT_TRUE(validateNoOverlap(memory_size, memory_offset, pool_size));
+  } else {
+    EXPECT_EQ(pool_size,
+              *std::max_element(memory_size.begin(), memory_size.end()));
+    // EXPECT_TRUE(validateIntervalOverlap(memory_size, memory_offset));
```
zhoonit, line 171: Is this todo?

nntrainer/tensor/basic_planner.h
kparichay: aah yes, done.

nntrainer/tensor/memory_pool.cpp
kparichay: this cant be zero -
1. if it is zero, its calculated at line 65.
2. this calculation can't be zero as memory_size is non empty, and memory pool does not accept requests of zero size.

nntrainer/tensor/memory_pool.cpp
kparichay: done.

nntrainer/tensor/memory_pool.cpp
kparichay: 1. updated.
2. added under debug

test/unittest/memory/memory_planner_validate.cpp
kparichay: As there is no overlap, a good planner should give share a common memory to all the requests. 
The check is very rigid in the way that it expects all the memories to have same starting point. The check will be updated once other better planners are implemented.

test/unittest/memory/memory_planner_validate.cpp
kparichay: added as details in doxygen (kept name same for consistency with other unittests).

test/unittest/memory/memory_planner_validate.cpp
kparichay: If the starting points are same, then we are sure that they overlap.
There is no way to check the size, as the `playLayout` does not return size of each allocation but just the start point. To check size, the total size of the allocation is checked.
Also, in the next PR, the memories are filled with random values and check is the values are correct, which ensures further correctness on size and overlap.

test/unittest/memory/memory_planner_validate.cpp
kparichay: Planner sorts the memory validity on its own. I have added another line to shuffle the memory validity to cover the case your mentioned.

test/unittest/memory/memory_planner_validate.cpp
kparichay: This will be added once other memory planners will be implemented.

test/unittest/memory/memory_planner_validate.cpp
kparichay: done.

test/unittest/memory/memory_planner_validate.cpp
zhoonit: Cool

zhoonit (approved):
LGTM :)

test/unittest/memory/memory_planner_validate.cpp
zhoonit: Cool

test/unittest/memory/memory_planner_validate.cpp
zhoonit: Yup

nntrainer/tensor/memory_pool.cpp
zhoonit: Alright :)

jijoongmoon:
I wonder if there is also kind of property for memory planner in network model. If it is not, I recommend to provide such property so that user can set it his own responsibility and preferance.

jijoongmoon (approved):
LGTM.

kparichay:
> I wonder if there is also kind of property for memory planner in network model. If it is not, I recommend to provide such property so that user can set it his own responsibility and preferance.

The current plan is to have multiple planners available, and the manager will choose the planner with the max efficiency (note: memory pool returns the efficiency of the planner).
The user will provide option to enable memory optimization and the best optimization will be selected automatically.
As choosing the best planner can consume some time and cost, the best planner for a model (and maybe its layout) can be saved with the model itself.

Issue #1480, opened 2021-08-12
Issues found from address sanitizer


kparichay:
Below is the main branch compile message with address sanitizer (used `meson build --buildtype debug -Db_sanitize=address` for the build)
```bash
1/2] Running all tests.
 1/33 unittest_tizen_capi                     FAIL    29.21 s (exit status 1)
 2/33 unittest_tizen_capi_layer               OK       0.29 s 
 3/33 unittest_tizen_capi_optimizer           OK       0.25 s 
 4/33 unittest_tizen_capi_dataset             FAIL     0.43 s (exit status 1)
 5/33 unittest_nntrainer_activations          OK       0.28 s 
 6/33 unittest_nntrainer_internal             OK       0.24 s 
 7/33 unittest_nntrainer_lazy_tensor          OK       0.23 s 
 8/33 unittest_nntrainer_tensor               OK       0.29 s 
 9/33 unittest_util_func                      OK       0.27 s 
10/33 unittest_databuffer_file                OK       0.27 s 
11/33 unittest_nntrainer_modelfile            OK       0.38 s 
12/33 unittest_nntrainer_models               OK       3.23 s 
13/33 unittest_nntrainer_graph                OK      14.31 s 
14/33 unittest_nntrainer_appcontext           OK       0.27 s 
15/33 unittest_base_properties                OK       0.32 s 
16/33 unittest_common_properties              OK       0.32 s 
17/33 unittest_memory                         OK       0.32 s 
18/33 unittest_compiler                       OK       0.04 s 
19/33 unittest_layers_v2                      FAIL     0.57 s (exit status 1)
20/33 unittest_datasets                       OK       1.97 s 
21/33 unittest_ccapi                          OK      44.07 s 
22/33 app_knn                                 OK       0.22 s 
23/33 app_logistic                            OK       2.43 s 
24/33 app_mnist                               TIMEOUT 60.02 s 
25/33 app_cifar_classification                OK       4.12 s 
26/33 app_cifar_classification_func           OK       0.32 s 
27/33 app_draw_classification                 OK      37.80 s 
28/33 app_layer_client_api                    OK       1.97 s 
29/33 app_layer_client_ini                    OK       1.93 s 
30/33 app_plugin_test                         OK       0.47 s 
31/33 app_product_ratings                     OK       1.42 s 
32/33 simpleshot_tests                        OK       0.28 s 
33/33 app_simpleshot_sample                   OK       1.63 s 

Ok:                   29
[...]

PR #1481, opened 2021-08-12
[fixes] Add bug fixes found with address sanitizer

kparichay:
This patch adds bug fixes after adding fixes related to address
sanitzer:
- NNStreamerLayer wasnt freeing resources upon destruction
- accessing released model handle in capi unittest
- not releasing model handle before end of unittest

See Also #1480

### DIFF
--- a/nntrainer/layers/nnstreamer_layer.cpp
+++ b/nntrainer/layers/nnstreamer_layer.cpp
@@ -62,18 +62,9 @@ int NNStreamerLayer::nnst_info_to_tensor_dim(ml_tensors_info_h &out_res,
   return status;
 }
 
-NNStreamerLayer::~NNStreamerLayer() {
-  try {
-    finalizeError(ML_ERROR_NONE);
-  } catch (std::exception &e) {
-    std::cerr << "failed in destructor, reason: " << e.what();
-  }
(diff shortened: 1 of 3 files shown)


nntrainer/layers/nnstreamer_layer.cpp
```diff
@@ -62,18 +62,9 @@ int NNStreamerLayer::nnst_info_to_tensor_dim(ml_tensors_info_h &out_res,
...
-    std::cerr << "failed in destructor, reason: " << e.what();
-  }
-}
-
-void NNStreamerLayer::finalizeError(int status) {
-  if (status == ML_ERROR_NONE)
-    return;
+NNStreamerLayer::~NNStreamerLayer() { release(); }
 
+void NNStreamerLayer::release() {
```
zhoonit, line 67: I guess this should be marked as noexcept (and guarantee it is noexcept), else put try catch at line 65?

kparichay: added noexcept.

zhoonit (approved):
LGTM :)

jijoongmoon (approved):
LGTM

PR #1483, opened 2021-08-12
[fixes] Add bug fixes from asan

zhoonit:
- [fixes] Add bug fixes from asan

```
- Fix issues from address sanitizer
- Enable deprecated-declaration warning but disable for some files

**Self evaluation:**

```

### DIFF
--- a/test/tizen_capi/unittest_tizen_capi.cpp
+++ b/test/tizen_capi/unittest_tizen_capi.cpp
@@ -19,6 +19,7 @@
  * @author      Jijoong Moon <jijoong.moon@samsung.com>
  * @bug         No known bugs
  */
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include <gtest/gtest.h>
 
 #include <nntrainer.h>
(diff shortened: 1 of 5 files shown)


api/capi/src/nntrainer.cpp
```diff
@@ -959,6 +959,11 @@ int ml_train_dataset_destroy(ml_train_dataset_h dataset) {
 
   check_feature_state();
 
+  if (dataset == nullptr) {
```
kparichay, line 962: maybe I am missing something, but isn't this checked with nntrainer_internal.h L127? if not, then this check will be needed in all the destroy methods.

test/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -19,6 +19,7 @@
  * @author      Jijoong Moon <jijoong.moon@samsung.com>
  * @bug         No known bugs
  */
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
kparichay, line 22: :100:

api/capi/src/nntrainer.cpp
zhoonit: You are correct, this is removed :)

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1487, opened 2021-08-12
[vargrad/tensor] Tensor names + cleanup 

kparichay:
Note: number of commits for this PR = 3
This patch provides cleanup for VarGrad and Manager, and add name to Tensor:

1. Flatten the tensors (weights/inputs/outputs/tensors) list stored in the
manager. It was of the form vector<vector<>> to club the tensors
requested from a layer together. However, this is neither required now
nor is it sufficient. If required later, a mapping from node will be
created.

2. Move name as an identifier to the tensor.
This name has been mvoed from the Var_Grad object.
Var_Grad now does not contain name itself.
The semantics of name for Tensor currently follows from var-grad:
- empty name is allowed
- repititive names are allowed
Restrictions can and will be added over time.

3. This patch cleansup VarGrad implementation, and ends up removing all its
members except the variable and gradient tensors (fitting exactly to its
description).

Resolves nnstreamer#1485 
Resolves nnstreamer#1486
See Also  #1127

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -257,12 +257,10 @@ void Manager::initializeWeights() {
     return;
 
   if (LAYER_V2) {
-    for (auto &l_w : weights_v2) {
-      for (auto &w : l_w) {
-        w->initializeVariable();
-        // tensor_map(&w->getVariableRef(), requestMemory(w.getDim().size(), 0,
-        // MAX));
-      }
+    for (auto &w : weights_v2) {
+      w->initializeVariable();
+      // tensor_map(&w->getVariableRef(), requestMemory(w.getDim().size(), 0,
+      // MAX));
     }
   } else {
     if (total_weight_size == 0) {
@@ -295,10 +293,8 @@ void Manager::allocateWeights() {
     return;
 
   if (LAYER_V2) {
-    for (auto &l_w : weights_v2) {
-      for (auto &w : l_w) {
-        w->allocateVariable();
-      }
+    for (auto &w : weights_v2) {
+      w->allocateVariable();
     }
   } else {
     for (auto &l_w : weights) {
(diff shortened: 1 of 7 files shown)


zhoonit (approved):
LGTM :)

nntrainer/tensor/manager.cpp
```diff
@@ -548,20 +530,14 @@ void Manager::deallocateInOuts() {
   shared_inout.deallocate();
 
   if (LAYER_V2) {
-    for (auto &li : inputs_v2) {
-      for (auto &in : li) {
-        in->deallocateVariable();
-      }
+    for (auto &in : inputs_v2) {
```
jijoongmoon, line 533: Just checking, do we still have to have LAYER_V2 Conditional?

nntrainer/tensor/var_grad.cpp
```diff
@@ -19,10 +19,8 @@
 namespace nntrainer {
 
 Var_Grad::Var_Grad(const TensorDim &dim, const Tensor::Initializer init,
-                   bool ng, bool alloc_now_, const std::string &name) :
-  dim(dim),
-  need_gradient(ng),
-  alloc_now(alloc_now_) {
+                   bool need_gradient, bool alloc_now,
+                   const std::string &name) {
```
jijoongmoon, line 23: Isn't there any svace issues if we remove this initialization?

jijoongmoon (approved):
LGTM except for minor comment.

nntrainer/tensor/manager.cpp
kparichay: this is cleaned up after 2 more PRs.

nntrainer/tensor/var_grad.cpp
kparichay: the corresponding member variables have been removed from the class, so it's ok.

lhs8928 (approved):
LGTM

PR #1479, opened 2021-08-11
[MemoryPool] Add memory pool implementation with unittests

kparichay:
This patch adds memory pool implementation with its unittests which uses the basic planner as the planner
- This patch provides implementation for memory pool which uses a memory
planner to plan the layout of the memory and provides efficiency of the
given planner.
- This patch adds unittests for memory pool for its public APIs.
Further, memory validation is added for the requested memories and
those tests will be added for each planner. This ensures that the
memories returned by each planner do not overlap, and remain valid while
usage.
Fixes corresponding to the unittests are also added.

See ALso #1127

Signed-off-by: Parichay Kapoor <pk.kapoor@samsung.com>

PR #1492, opened 2021-08-17
[manager] Use memory pool for weights

kparichay:
This patch updates the use of memory pool for the weights of the model.
Correspondingly a pool object is added to the manager.
The pool is allocated in the weights allocation.

See also #1127

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -30,6 +30,7 @@
 #include <vector>
 
 #include <activation_layer.h>
+#include <basic_planner.h>
 #include <flatten_layer.h>
 #include <layer_node.h>
 #include <manager.h>
@@ -249,6 +250,12 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   return allocate_func;
 }
 
+std::pair<unsigned int, unsigned int>
+Manager::getValidity(const std::string &name) {
+  /** @todo calculate validity based on lifespan and usage */
+  return {0, std::numeric_limits<unsigned>::max()};
+}
+
 /**
  * @brief Allocate and initialize the weight variable
  */
@@ -260,9 +267,11 @@ void Manager::initializeWeights() {
   if (LAYER_V2) {
     for (auto &w : weights_v2) {
       w->initializeVariable();
-      // tensor_map(&w->getVariableRef(), requestMemory(w.getDim().size(), 0,
-      // MAX));
+      auto const &t_validity = getValidity(w->getName());
+      tensor_token_map[w->getName()] = pool.requestMemory(
+        w->getVariableRef().bytes(), t_validity.first, t_validity.second);
     }
+    pool.planLayout(BasicPlanner());
   } else {
     if (total_weight_size == 0) {
       ml_logw(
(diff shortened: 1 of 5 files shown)


nntrainer/tensor/manager.cpp
```diff
@@ -249,6 +250,12 @@ Manager::AllocFunc Manager::getAllocFunc(bool is_weight) {
   return allocate_func;
 }
 
+std::pair<unsigned int, unsigned int>
+Manager::getValitdity(const std::string &name) {
```
zhoonit, line 254: getValidity?

nntrainer/tensor/memory_pool.cpp
```diff
@@ -250,6 +250,22 @@ size_t MemoryPool::calcMinMemoryRequirement() {
                         return val1.second < val2.second;
                       });
   unsigned int last_interval = max_interval.second;
+  /** as weights stay valid for max duration, ignore this value and get the real
+   * max value */
+  if (last_interval == std::numeric_limits<unsigned int>::max()) {
+    max_interval = *std::max_element(
+      memory_validity.begin(), memory_validity.end(),
+      [last_interval](auto const &val1, auto const &val2) {
+        return ((val2.second != last_interval) && (val1.second < val2.second));
```
zhoonit, line 259: Is this calculation assumes that the given vector is sorted in some order?

nntrainer/tensor/memory_pool.cpp
```diff
@@ -250,6 +250,22 @@ size_t MemoryPool::calcMinMemoryRequirement() {
                         return val1.second < val2.second;
                       });
   unsigned int last_interval = max_interval.second;
+  /** as weights stay valid for max duration, ignore this value and get the real
```
zhoonit, line 253: Looks like this method needs some dedicated tests later if there isn't any?

zhoonit (approved):
LGTM

nntrainer/tensor/manager.cpp
kparichay: done.

nntrainer/tensor/memory_pool.cpp
kparichay: no, this doesn't assume any sorting order. This is trying to get the max value of the validity to determine the range of the validity. So, if the max value is UINT_MAX, it gets the 2nd highest value.

nntrainer/tensor/memory_pool.cpp
kparichay: https://github.com/nnstreamer/nntrainer/pull/1479 added them. Let me know if some cases are missed.

zhoonit: Checked, thanks :)

nntrainer/tensor/memory_pool.cpp
zhoonit: [7, 8]
[4, 6]
[4, MAX]

In this case, would [7, 8] be selected?

kparichay: yes.

kparichay: note that as we are only interested in the end_value, so effectively 8 is selected from the interval.

zhoonit: I see, thanks :)

jijoongmoon (approved):
LGTM

kparichay:
The commits in this patch have been made available to #1511.

PR #1494, opened 2021-08-17
[fix] ahub issues

lhs8928:
- Added try catch statement
 - Check value to avoid divide by zero
 - Added const keyword to match override function
 - Initialize member variable in constructor

### DIFF
--- a/nntrainer/dataset/databuffer.cpp
+++ b/nntrainer/dataset/databuffer.cpp
@@ -55,7 +55,7 @@ public:
    * @param value default value
    */
   PropsBufferSize(unsigned int value = 1) { set(value); }
-  bool isValid(const unsigned int &v) { return v > 0; }
+  bool isValid(const unsigned int &v) const override { return v > 0; }
   static constexpr const char *key = "buffer_size"; /**< unique key to access */
   using prop_tag = uint_prop_tag;                   /**< property type */
 };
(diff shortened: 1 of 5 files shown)


jijoongmoon (approved):
LGTM

nntrainer/dataset/databuffer.cpp
```diff
@@ -55,7 +55,7 @@ class PropsBufferSize : public Property<unsigned int> {
    * @param value default value
    */
   PropsBufferSize(unsigned int value = 1) { set(value); }
-  bool isValid(const unsigned int &v) { return v > 0; }
+  bool isValid(const unsigned int &v) const { return v > 0; }
```
zhoonit, line 58: could you put override here as well?

zhoonit (approved):
LGTM :)

nntrainer/layers/layer_node.cpp
```diff
@@ -124,7 +124,8 @@ LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l) :
   activation_type(ActivationType::ACT_NONE),
   layer_node_props(new PropsType(props::Name(), props::Flatten(),
                                  props::Distribute(), props::Trainable(),
-                                 props::Loss())) {
+                                 props::Loss())),
+  regularization_loss(0.0f) {
```
kparichay, line 128: check indentation here

kparichay (approved):
LGTM

nntrainer/dataset/databuffer.cpp
lhs8928: Done

PR #1506, opened 2021-08-19
[CAPI] Update description of ml_train_datagen_cb

zhoonit:
- [CAPI] Update description of ml_train_datagen_cb

```
This patch updates ml_train_datagen_cb with newly added api sets and
clarify the parameter @a last means.

**Self evaluation:**

```

### DIFF
--- a/api/nntrainer-api-common.h
+++ b/api/nntrainer-api-common.h
@@ -77,7 +77,7 @@ typedef enum {
  *
  * @details The user of the API must provide this callback function to supply
  * data to the model and register the callback with
- * ml_train_dataset_create_with_generator(). The model will call this callback
+ * ml_train_dataset_add_generator(). The model will call this callback
  * whenever it needs more data. This function should provide a single element of
  * input and label data in the passed containers. The containers passed by the
  * caller will already be allocated with sufficient space to contain the data.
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM

api/nntrainer-api-common.h
```diff
@@ -95,13 +95,16 @@ typedef enum {
  * number of samples are set as a property for this dataset. In this case, last
  * is only used for verification purposes. If total number of samples for the
  * dataset is unknown, this function will be called in sequence.
+ * @note If @a last false when not modified by the user. @a last has to be set
```
kparichay, line 98: If @a last false when not modified by the user. -> it sounds like `last` variable will be defaulted to false by the caller and user need not set it if not needed. Is this intentional?

zhoonit: Yes, it is current behavior and I think it's fine to behave like this.

kparichay: The described behavior works, but IMO, its best for the callee to always set this variable, and not assume about the default value set by the caller.

kparichay (approved):
LGTM

api/nntrainer-api-common.h
zhoonit: Let's delete this line although this behavior will exist to not break old codes.

PR #1509, opened 2021-08-20
[Wait for #1508] [Dataset] Change and apply callback handler

zhoonit:
**[Pending commit: #1508]**

- [Dataset] Change and apply callback handler

```
This patch change and apply callback handler from batchwise ->
samplewise

**Self evaluation:**

```


zhoonit:
From this patch, old callback will break, on merging this PR, nntrainer-internal has to be taken care of as well.

PR #1511, opened 2021-08-20
[TensorPool] Added TensorPool as a pool of tensors

kparichay:
- This patch introduces TensorPool which manages the list of all the
tensors with their execution orders and lifespan.
The conversion of lifespan and execution orders to the validity will be
done by the TensorPool.
TensorPool forms the middle ground between memory pool and manager for
easier maintainance.

- Added tensor pool implementation with unittests.
Tensorpool provides the interface where all the requested tensors are
managed at one places ensuring the uniqueness of name with their
lifespan and execution orders.
Corresponding unittests checking the working is added.

This patch aims to reduce the work and data structures from the manager and shift the workload to TensorPool.

Merged already approved #1492 in this PR.

See Also #1127

### DIFF
--- a/nntrainer/tensor/tensor_pool.cpp
+++ b/nntrainer/tensor/tensor_pool.cpp
@@ -1,0 +1,214 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Parichay Kapoor <pk.kapoor@samsung.com>
+ *
+ * @file   tensor_pool.cpp
+ * @date   19 Aug 2020
+ * @brief  This is TensorPool for all requested tensors
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug	   No known bugs except for NYI items
+ *
+ * @todo   add checks for request/updates that finalize is not done
+ * @todo   check before allocate that finalize is done
+ */
+
+#include <memory_pool.h>
+#include <tensor.h>
+#include <tensor_pool.h>
+#include <tensor_wrap_specs.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+/**
+ * @brief     Request tensor with the given spec
+ *
+ * @note returns empty tensor which will be filled when allocate is called.
+ * @note we assume that the caller checks if the exec_order and lifespan are
+ * compatible.
+ */
+Tensor *TensorPool::requestTensor(const TensorDim dim,
+                                  const std::vector<unsigned int> &exec_order,
+                                  TensorLifespan lifespan,
+                                  const std::string &name) {
+  if (pool.find(name) != pool.end())
+    throw std::invalid_argument("Cannot request tensor with same name");
+
+  if (dim.getDataLen() == 0)
+    throw std::invalid_argument("Cannot request tensor with size 0");
+
+  if (name.empty())
+    throw std::invalid_argument("Cannot request tensor with empty name");
+
+  pool[name] = {
+    std::make_unique<Tensor>(dim, false, Tensor::Initializer::NONE, name),
+    exec_order, lifespan, 0};
+
+  return pool[name].tensor.get();
+}
+
+/**
+ * @brief     Request tensor which has been already requested with the given
+ * spec
+ *
+ * @note returns empty tensor which will be filled when allocate is called.
+ * @note we assume that the caller checks if the exec_order and lifespan are
+ * compatible.
+ */
+Tensor *TensorPool::requestPrerequestedTensor(
+  const TensorDim dim, const std::vector<unsigned int> &exec_order,
+
+  TensorLifespan lifespan, const std::string &name) {
+  if (pool.find(name) == pool.end())
+    throw std::invalid_argument("Requested tensor not found");
+
+  auto &spec = pool[name];
+  if (spec.tensor->getDim() != dim)
+    throw std::invalid_argument("Request tensor dimension mismatch");
+
+  spec.exec_order.insert(spec.exec_order.end(), exec_order.begin(),
+                         exec_order.end());
+  spec.lifespan = enum_class_or<TensorLifespan>(spec.lifespan, lifespan);
+
+  return pool[name].tensor.get();
+}
+
+/**
+ * @brief finalize the requested tensors
+ *
+ * @details finalize the requested tensors, request memory for them and plan
+ * layout for their allocations.
+ */
+void TensorPool::finalize(const MemoryPlanner &planner,
+                          unsigned int start_order, unsigned int end_order) {
+  unsigned int bytes_requested = 0;
+  for (auto &entry : pool) {
+    auto &spec = entry.second;
+
+    if (spec.exec_order.empty())
+      continue;
+
+    /** 1. create the validity ranges for the all the requested tensors */
+    unsigned int validity_start =
+      *std::min_element(spec.exec_order.begin(), spec.exec_order.end());
+    unsigned int validity_end =
+      *std::max_element(spec.exec_order.begin(), spec.exec_order.end());
+
+    /**
+     * use lifespan to update the validity.
+     * if the validity is long term, the tensor must stay valid for the
+     * complete duration.
+     */
+    if (isTensorLongTerm(spec.lifespan)) {
+      validity_start = start_order;
+      validity_end = end_order;
+    }
+
+    /** 2. for each tensor request if it is in the provided range */
+    if (validity_end < start_order || validity_start > end_order)
+      continue;
+    validity_start = std::max(validity_start, start_order);
+    validity_end = std::max(validity_end, end_order);
+
+    /**
+     * 3. requestMemory for all the tensors and set their tokens
+     * @note +1 is to make the validity_end exlusive in the interval range
+     */
+    spec.token = mem_pool.requestMemory(spec.tensor->bytes(), validity_start,
+                                        validity_end + 1);
+    bytes_requested += spec.tensor->bytes();
+  }
+
+  /** 4. finalizeLayout for the memory pool. */
+  if (bytes_requested > 0)
+    mem_pool.planLayout(planner);
+}
+
+/**
+ * @brief Set the batch size for the inputs/outputs of the layers
+ */
+void TensorPool::setBatchSize(const std::string &name, unsigned int batch) {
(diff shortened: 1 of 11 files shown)


nntrainer/tensor/tensor_pool.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * @brief     Request tensor which has been already requested with the given
+ * spec
+ *
+ * @note returns empty tensor which will be filled when allocate is called.
+ * @note we assume that the caller checks if the exec_order and lifespan are
+ * compatible.
+ */
+Tensor *TensorPool::requestPreallocatedTensor(
```
zhoonit, line 59: How about sth like Prerequested?

nntrainer/tensor/tensor_pool.cpp
```diff
@@ -0,0 +1,214 @@
...
+     * if the validity is long term, the tensor must stay valid for the
+     * complete duration.
+     */
+    if (isTensorLongTerm(spec.lifespan)) {
+      validity_start = start_order;
+      validity_end = end_order;
+    }
+
+    /** 2. for each tensor request if it is in the provided range */
+    if (validity_end < start_order || validity_start > end_order)
```
zhoonit, line 109: Isn't it error if the tensor validity is out of provided range?

zhoonit: is the order inclusive?

nntrainer/tensor/tensor_pool.cpp
```diff
@@ -0,0 +1,214 @@
...
+      continue;
+    validity_start = std::max(validity_start, start_order);
+    validity_end = std::max(validity_end, end_order);
```
zhoonit, line 112: maybe one of this has to be min?

nntrainer/tensor/tensor_pool.cpp
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/**
+ * @brief     Check if the lifespan leads to long term valitidy
+ *
+ */
+bool TensorPool::isTensorLongTerm(const TensorLifespan &lifespan) {
+  switch (lifespan) {
+  case TensorLifespan::EPOCH_LIFESPAN:
+    /** fallthrough intended */
```
zhoonit, line 198: it's just minor, but we can have compiler friendly [[fallthrough]] as of c++17 :)

nntrainer/tensor/tensor_pool.cpp
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * @brief Allocate memory for all the managed tensors
+ */
+void TensorPool::allocate() {
+  mem_pool.allocate();
+
+  /** set the pointers using the token for all the tensors */
+  for (auto &entry : pool) {
+    auto &spec = entry.second;
+    spec.tensor->setData(mem_pool.getMemory(spec.token));
```
zhoonit, line 147: wonder if source tensor is going away?

test/unittest/unittest_nntrainer_tensor_pool.cpp
```diff
@@ -0,0 +1,364 @@
...
+  EXPECT_NO_THROW(
+    t2 = pool.requestTensor(nntrainer::TensorDim({1}), {},
+                            nntrainer::TensorLifespan::MAX_LIFESPAN, "abc2"));
+  EXPECT_NE(t2, nullptr);
+  EXPECT_FALSE(t2->isAllocated());
+
+  EXPECT_NE(t1, t2);
+
+  EXPECT_NO_THROW(pool.finalize(nntrainer::BasicPlanner(), 0, 2));
+  EXPECT_EQ(pool.minMemoryRequirement(), 0);
```
zhoonit, line 225: is this zero?

test/unittest/unittest_nntrainer_tensor_pool.cpp
```diff
@@ -0,0 +1,364 @@
...
+TEST(TensorPool, finalize_03_p) {
+  nntrainer::TensorPool pool;
+
+  EXPECT_NO_THROW(pool.finalize(nntrainer::BasicPlanner(), 0, 2));
+}
+
+/**
+ * @brief allocate
+ */
+TEST(TensorPool, allocate_deallocate_01_p) {
```
zhoonit, line 273: how about using something like reserve, evict?

test/unittest/unittest_nntrainer_tensor_pool.cpp
```diff
@@ -0,0 +1,364 @@
...
+  nntrainer::Tensor g1 = nntrainer::Tensor(nntrainer::TensorDim({100}));
+  g1.setRandNormal();
+  nntrainer::Tensor g2 = nntrainer::Tensor(nntrainer::TensorDim({100}));
+  g2.setRandNormal();
+
+  t1->copy(g1);
+  t2->copy(g2);
+
+  EXPECT_EQ(*t1, g1);
+  EXPECT_EQ(*t2, g2);
```
zhoonit, line 340: Wonder what this lines try to test? is this trying to check t1 and t2 does not clash?
If that's the case, I recommend to use Tensor::fill. copy creates a new tensor if it has empty data.

nntrainer/tensor/tensor_pool.cpp
kparichay: sounds good, updated.

nntrainer/tensor/tensor_pool.cpp
kparichay: No, its not error. When the model is created, the validity of a tensor is given for training as well as for inference. However, when it is during `finalize()` that the mode of execution is actually finalized. So, in inference, many tensors validity will always be out of provided range.

kparichay: > is the order inclusive for both side?

yes - https://github.com/nnstreamer/nntrainer/blob/3d6138f9c4c839d3087a16cf5e6478bb947d5fe9/nntrainer/tensor/tensor_pool.h#L88-L89

nntrainer/tensor/tensor_pool.cpp
kparichay: you are right, end must be determined using min - I will fix in #1531 to avoid rebasing issues - added https://github.com/nnstreamer/nntrainer/pull/1531/files#r716384622

nntrainer/tensor/tensor_pool.cpp
kparichay: done.

nntrainer/tensor/tensor_pool.cpp
kparichay: can you elaborate? If you are referring to src_tensor in tensor.h, then eventually that would be replaced with tensor_pool interface. But it would require all tensors to be created via tensor_pool, which would take some time.

test/unittest/unittest_nntrainer_tensor_pool.cpp
kparichay: Yes, it is zero because the tensors, although with a valid lifespan, have no usage. As they aren't expected to be ever used, they aren't allocated any memory.

test/unittest/unittest_nntrainer_tensor_pool.cpp
kparichay: These checks ensure that the memory of tensors t1 and t2 are not overlapped.

nntrainer/tensor/tensor_pool.cpp
zhoonit: I was wondering about the `src_tensor` in tensor.h will eventually go away. Thanks :)

test/unittest/unittest_nntrainer_tensor_pool.cpp
zhoonit: I see

test/unittest/unittest_nntrainer_tensor_pool.cpp
kparichay: for the name?

zhoonit: not for the name but the concept itself... it's because this sounds like memory allocation. but after the allocate call, tensor->isAllocate() is false.

nntrainer/tensor/tensor_pool.cpp
zhoonit: Cool

test/unittest/unittest_nntrainer_tensor_pool.cpp
zhoonit: Other than allocate will be good I think?

zhoonit (approved):
LGTM

nntrainer/tensor/tensor_pool.cpp
kparichay: Yes, I think it can eventually go away, and will be good to make tensor objects lighter.

test/unittest/unittest_nntrainer_tensor_pool.cpp
kparichay: > it's because this sounds like memory allocation. but after the allocate call, tensor->isAllocate() is false.

After the allocate call, tensor->isAllocate() is true. It sounds like memory allocation because it is. Let me know if I am missing what you are trying to convey here.

zhoonit: I see, I think I missed the bottom part :) Thanks for the explanation!

nntrainer/tensor/tensor_pool.cpp
jijoongmoon: I think it isn't easy to remove src_tensor. Because some of the ops ( as in RNN or GRU Layer ) is based on src_tensor. We need to check if it is ok when we remove src_tensor.

jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor_pool.cpp
kparichay: Yeah, we will have to check with all the existing usages when working on it.

nntrainer/tensor/tensor_pool.cpp
kparichay: https://github.com/nnstreamer/nntrainer/pull/1541/commits/858352d6852a7bf0e5fe71e616eaa512306f6700#r717205648

PR #1514, opened 2021-08-24
[unittest] backbone property name update

kparichay:
Update backbone property name for the layers unittests.
This patch fixes failing unittests for main.


zhoonit (approved):
Nice catch :)

jijoongmoon (approved):
LGTM

PR #1523, opened 2021-08-25
[Manager] Manager to use TensorPool for all requests

kparichay:
This patch updates manager to use TensorPool for all of its requests.
Other changes are as listed below:
- NetworkGraph now caches the inputs and labels Var_Grads which can
directly by the model before execution.
- setLabels and setInputs in the model have been updated. Further,
manual setting of inputs and labels has been removed.
- Introduce clear in memory pool to clear any allocations and requests
- TensorPool clears any requests made before making more requests in
finalize
- models unittest updated to not pass label when loss is not given in
the model. Passing label without loss in the model now results in error
- Manager updated to use tensor pool for all the memory requests. The
allocation, initialization and deallocation has been correspondingly simplified

Much needed cleanup will be done soon.

See Also #1127

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -429,7 +429,7 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
   auto allocated = tensor_manager->isAllocated();
 
   if (allocated)
-    tensor_manager->deallocateTensors();
+    deallocateTensors();
 
   for (auto iter = cbegin(); iter != cend(); iter++) {
     (*iter)->setBatch(batch_size);
@@ -446,10 +446,11 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
       }
     }
   }
+  /// resize input and output spec
   tensor_manager->setBatchSize(batch_size);
 
   if (allocated)
-    tensor_manager->allocateTensors();
+    allocateTensors();
 }
 
 sharedConstTensors NetworkGraph::forwarding(bool training) const {
@@ -712,6 +713,21 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
   const std::vector<Var_Grad *> &outputs =
     tensor_manager->requestOutputs(gnode, init_context.getOutputDimensions());
 
+  /**
+   * @note cache the labels and input var_grads to be able to fill them when
+   * running the graph.
+   */
+  if (gnode.getInputConnections().empty())
+    input_list.insert(input_list.end(), inputs.begin(), inputs.end());
+  /** @todo check compatibility of requireLabel() and
+   * getOutputConnections().empty() */
+  if (lnode->requireLabel())
+    label_list.insert(label_list.end(), outputs.begin(), outputs.end());
+
+  /**
+   * @note must use existing properties like name/trainable of run_context to
+   * create the new run_context
+   */
   lnode->configureRunContext(
     // TODO: update weights spec for trainable based on layer trainable prop
     tensor_manager->requestWeights(gnode, init_context.getWeightsSpec()),
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -724,10 +724,6 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
   if (lnode->requireLabel())
     label_list.insert(label_list.end(), outputs.begin(), outputs.end());
 
-  /**
-   * @note must use existing properties like name/trainable of run_context to
-   * create the new run_context
-   */
   lnode->configureRunContext(
     // TODO: update weights spec for trainable based on layer trainable prop
     tensor_manager->requestWeights(gnode, init_context.getWeightsSpec()),
--- a/nntrainer/graph/network_graph.h
+++ b/nntrainer/graph/network_graph.h
@@ -293,6 +293,15 @@ public:
    * @brief Allocate and initialize the weight variable
    */
   void initializeWeights() {
+    /**
+     * get the order of execution/usage order for the forwarding of the last
+     * layer and pass that as the max_exec_order ensuring that all weights with
+     * usage less than the max_exec_order are allocated.
+     *
+     * @note usage order of backwarding is not considered because weight has a
+     * separate memory pool for now. Later, if it shares pool with other
+     * tensors, then this must max of backward usage order.
+     */
     tensor_manager->initializeWeights(
       std::get<0>((*(cend() - 1))->getExecutionOrder()));
   }
@@ -303,10 +312,21 @@ public:
    */
   void initializeTensors(bool training) {
     if (!training)
+      /**
+       * get the order of execution/usage order for the forwarding of the last
+       * layer and pass that as the max_exec_order ensuring that all tensors
+       * with usage less than the max_exec_order are allocated.
+       */
       tensor_manager->initializeTensors(
         training, std::get<0>((*(cend() - 1))->getExecutionOrder()));
     else
       /** @todo update this to skip non-trainable layers */
+      /**
+       * get the order of execution/usage order for the backwarding of the first
+       * layer (as that will be the last layer to executed in the backwarding)
+       * and pass that as the max_exec_order ensuring that all tensors with
+       * usage less than the max_exec_order are allocated.
+       */
       tensor_manager->initializeTensors(
         training, std::get<1>((*(cbegin()))->getExecutionOrder()));
   }
--- a/nntrainer/graph/network_graph.h
+++ b/nntrainer/graph/network_graph.h
@@ -292,14 +292,23 @@ public:
   /**
    * @brief Allocate and initialize the weight variable
    */
-  void initializeWeights() { tensor_manager->initializeWeights(); }
+  void initializeWeights() {
+    tensor_manager->initializeWeights(
+      std::get<0>((*(cend() - 1))->getExecutionOrder()));
+  }
 
   /**
    * @brief Initialize the inputs/outputs/derivatives/gradients for the layers
    * @param[in] training If true, initialize derivates/gradients, else, do not.
    */
   void initializeTensors(bool training) {
-    tensor_manager->initializeTensors(training);
+    if (!training)
+      tensor_manager->initializeTensors(
+        training, std::get<0>((*(cend() - 1))->getExecutionOrder()));
+    else
+      /** @todo update this to skip non-trainable layers */
+      tensor_manager->initializeTensors(
+        training, std::get<1>((*(cbegin()))->getExecutionOrder()));
   }
 
   /**
@@ -336,11 +345,25 @@ public:
     for (auto const &w : tensor_manager->getWeights()) {
       const TensorDim &dim = w->getDim();
       std::vector<TensorDim> dims = cb(dim);
-      for (auto &dim : dims) {
-        w->addOptimizerVariable(dim);
-      }
+      w->setOptimizerVariables(tensor_manager->requestWeightOptimizerVariables(
+        dims, w->getName(), TensorLifespan::MAX_LIFESPAN,
+        Tensor::Initializer::ZEROS));
     }
   }
+
+  /**
+   * @brief Get the Input List for the graph
+   *
+   * @return const std::vector<Var_Grad *>& lists of inputs
+   */
+  const std::vector<Var_Grad *> &getInputList() { return input_list; };
+
+  /**
+   * @brief Get the Label List for the graph
+   *
+   * @return const std::vector<Var_Grad *>& lists of labels
+   */
+  const std::vector<Var_Grad *> &getLabelList() { return label_list; };
 
 private:
   std::map<std::string, std::string> sub_in_out; /** This is map to identify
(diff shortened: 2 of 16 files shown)


kparichay:
Use `requestExternallyAllocatedTensors` from #1541 to resolve the build issue here.

nntrainer/graph/network_graph.cpp
```diff
@@ -712,6 +713,21 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
...
+  if (gnode.getInputConnections().empty())
+    input_list.insert(input_list.end(), inputs.begin(), inputs.end());
+  /** @todo check compatibility of requireLabel() and
+   * getOutputConnections().empty() */
+  if (lnode->requireLabel())
+    label_list.insert(label_list.end(), outputs.begin(), outputs.end());
+
+  /**
+   * @note must use existing properties like name/trainable of run_context to
+   * create the new run_context
```
zhoonit, line 729: There is no run context before this, could you clarify the comment here?

kparichay: yes, this must be removed after rebase. I will remove it.

nntrainer/graph/network_graph.h
```diff
@@ -292,14 +292,23 @@ class NetworkGraph {
   /**
    * @brief Allocate and initialize the weight variable
    */
-  void initializeWeights() { tensor_manager->initializeWeights(); }
+  void initializeWeights() {
+    tensor_manager->initializeWeights(
+      std::get<0>((*(cend() - 1))->getExecutionOrder()));
```
zhoonit, line 297: Could you separate this to a variable or add a note? *(cend() - 1))->getExecutionOrder()

nntrainer/graph/network_graph.h
```diff
@@ -292,14 +292,23 @@ class NetworkGraph {
...
    */
   void initializeTensors(bool training) {
-    tensor_manager->initializeTensors(training);
+    if (!training)
+      tensor_manager->initializeTensors(
+        training, std::get<0>((*(cend() - 1))->getExecutionOrder()));
+    else
+      /** @todo update this to skip non-trainable layers */
+      tensor_manager->initializeTensors(
+        training, std::get<1>((*(cbegin()))->getExecutionOrder()));
```
zhoonit, line 311: same here.

nntrainer/models/neuralnet.cpp
```diff
@@ -162,31 +164,71 @@ int NeuralNetwork::initialize() {
...
-    for (unsigned int i = 0; i < layer_node->getNumOutputs(); i++) {
-      layer_node->getOutputGradUnsafe(i) = Tensor();
-    }
-  };
+  /// feed or clear label
+  for (unsigned int idx = 0; idx < data.size(); idx++) {
+    if (data.empty())
+      label_list[idx]->initializeGradient();
+    else if (data.size() == 1)
+      label_list[idx]->initializeGradient(data[0]);
```
zhoonit, line 181: maybe this is not needed?

nntrainer/models/neuralnet.cpp
```diff
@@ -162,31 +164,71 @@ int NeuralNetwork::initialize() {
...
 
-  auto clear_label = [](auto const &layer_node) {
-    for (unsigned int i = 0; i < layer_node->getNumOutputs(); i++) {
-      layer_node->getOutputGradUnsafe(i) = Tensor();
-    }
-  };
+  /// feed or clear label
+  for (unsigned int idx = 0; idx < data.size(); idx++) {
+    if (data.empty())
+      label_list[idx]->initializeGradient();
```
zhoonit, line 179: if data.empty(), this line cannot be executed?

nntrainer/models/neuralnet.cpp
```diff
@@ -162,31 +164,71 @@ int NeuralNetwork::initialize() {
...
-    auto const &lnode = *iter;
-    if (lnode->requireLabel()) {
-      label.empty() ? clear_label(lnode) : fill_label(lnode);
-    }
+  for (unsigned int idx = 0; idx < input_list.size(); idx++) {
+    input_list[idx]->initializeVariable(data[idx]);
+  }
+}
+
+static void setLabels(sharedConstTensors &data,
```
zhoonit, line 201: There is another setLabels above, one of them should depend on the other one?

nntrainer/tensor/manager.cpp
```diff
@@ -567,17 +529,7 @@ void Manager::allocateDerivatives() {
...
-      in->deallocateGradient();
-    }
-    for (auto &out : outputs_v2) {
-      out->deallocateGradient();
-    }
-    // for (auto &t : tensors_v2) {
-    //   t->deallocateGradient();
-    // }
-  } else {
+  if (!LAYER_V2) {
```
zhoonit, line 532: How about deleting this as well (maybe in later PR??)

nntrainer/tensor/manager.cpp
```diff
@@ -698,32 +633,20 @@ void Manager::initializeTensorsTrain() {
...
-    }
+    tensor_pool.finalize(BasicPlanner(), 0, max_exec_order_);
   }
 }
 
 /**
  * @brief Initialize the inputs/outputs/gradients/derivatives for the layer
  */
-void Manager::initializeTensors(bool training) {
+void Manager::initializeTensors(bool training, unsigned int max_exec_order_) {
```
zhoonit, line 643: What does max_exec_order_ do? is this initialize Until the given number?

nntrainer/tensor/manager.cpp
```diff
@@ -876,57 +799,64 @@ std::vector<Var_Grad *>
...
-    /** usage for inputs gradients (outgoing derivatives) */
-    tensor_exec_order[gname].push_back(std::get<2>(exec_order));
-
-    /** set tensor lifespan */
-    expandLifespan(vname, tspan);
-    expandLifespan(gname, tspan);
+    inputs_v2.emplace_back(std::make_unique<Var_Grad>(var, grad));
   }
 
+  std::transform(inputs_v2.begin() + current_size, inputs_v2.end(),
```
zhoonit, line 866: minor optimization: `ret.reserve(current_size)`

nntrainer/tensor/manager.h
```diff
@@ -365,7 +390,7 @@ class Manager {
    * will require full allocation than reusing memory allocated with inference
    * mode.
    */
-  void initializeTensors(bool training);
+  void initializeTensors(bool training, unsigned int max_exec_order);
```
zhoonit, line 393: Could you update doxygen? (also for line 639, 644)

nntrainer/tensor/var_grad.h
```diff
@@ -101,7 +101,12 @@ class Var_Grad {
    */
   explicit Var_Grad(Tensor *v, Tensor *g) :
     var(std::shared_ptr<Tensor>(v, [](void *) {})),
-    grad(std::shared_ptr<Tensor>(g, [](void *) {})) {}
+    grad(std::shared_ptr<Tensor>(g, [](void *) {})) {
+    if (!v)
+      var = std::make_shared<Tensor>();
+    if (!g)
+      grad = std::make_shared<Tensor>();
```
zhoonit, line 108: Are we making one if not present?
How about we let it null instead now? (maybe later??)
 Related to https://github.com/nnstreamer/nntrainer/pull/1522#discussion_r717299885

zhoonit (approved):
LGTM with minor comments.

nntrainer/graph/network_graph.h
kparichay: will do.

nntrainer/models/neuralnet.cpp
kparichay: This sets all the given labels to all the output nodes that require labels. Can you elaborate on your question?

nntrainer/models/neuralnet.cpp
kparichay: The purpose of this statement is to remove the labels/inputs which may have already been set.
For example, this clears the label set from the previous training during inference.

nntrainer/tensor/manager.cpp
kparichay: its cleaned up right in next PR :)

nntrainer/tensor/manager.cpp
kparichay: Yes, this tells the manager till which node is the network going to run. One example to exclude allocation gradients while inference as their execution order would be more than the provided max_exec_order.

nntrainer/tensor/var_grad.h
kparichay: We can't leave it null because the dimensions in layers depends on RunContext which depends on its outputs and inputs. So, for now, vars and grads objects do exist, but they can be empty or unallocated.

There will be yet another cleanup once the rebase is done to clean up VarGrad various interfaces.

nntrainer/models/neuralnet.cpp
kparichay: Yeah, its done with #1544. I will cherry-pick that change here.

nntrainer/models/neuralnet.cpp
zhoonit: this line seems to equivalent to the

```
label_list[idx]->initializeGradient(data[idx]);
```
so removing this else if statement?

nntrainer/tensor/var_grad.h
zhoonit: Sure :)

nntrainer/models/neuralnet.cpp
kparichay: aah, I understand the question. It is possible to receive just 1 label or input which must be set to all the labels/inputs. So, this `if` condition takes care of that special condition.

nntrainer/graph/network_graph.cpp
kparichay: done.

nntrainer/graph/network_graph.h
kparichay: done.

nntrainer/graph/network_graph.h
kparichay: done.

nntrainer/tensor/manager.cpp
kparichay: updated to `ret.reserve(inputs_dim.size())`.

nntrainer/tensor/manager.h
kparichay: Updated. 639 and 644 are removed in the next PR, so they are skipped.

nntrainer/models/neuralnet.cpp
zhoonit: if that's the case, could you check if the loop condition should be ?
```
for (unsigned int idx = 0; idx < label.size(); idx++) {
```

nntrainer/models/neuralnet.cpp
```diff
@@ -162,31 +164,71 @@ int NeuralNetwork::initialize() {
...
-      << " layer: " << layer_node->getName() << " label size: " << label.size()
-      << " requirements size: " << layer_node->getNumOutputs();
+static void setLabels(const std::vector<Tensor> &data,
+                      const std::vector<Var_Grad *> &label_list) {
 
-    for (unsigned int i = 0; i < layer_node->getNumOutputs(); i++) {
-      layer_node->getOutputGradUnsafe(i) = *label[i];
-    }
-  };
+  NNTR_THROW_IF(!data.empty() && data.size() != label_list.size(),
```
zhoonit, line 170: Data.size() == 1 is not an error?

nntrainer/tensor/manager.cpp
zhoonit: Awesome :)

nntrainer/tensor/manager.cpp
zhoonit: Cool

nntrainer/tensor/manager.h
zhoonit: Okay

nntrainer/models/neuralnet.cpp
zhoonit: https://github.com/nnstreamer/nntrainer/pull/1523#discussion_r718100920  I think this is this.

nntrainer/models/neuralnet.cpp
kparichay: You are right, its a bug for particular this function. It was resolved later with #1544, but I will update it here as well.

nntrainer/models/neuralnet.cpp
kparichay: This function will be slightly refactored when I borrow the latest version from #1544. Please refer to https://github.com/nnstreamer/nntrainer/pull/1544/files#diff-8136da3afc181f3163a8f8a15d86d132abee8c07de2c9b49bc2cf0759e380a69R833-R879 for the reference code. I will update as per this soon in this PR.

zhoonit: Cool, no worries. You can delay to reduce work at will :)

nntrainer/models/neuralnet.cpp
zhoonit: Cool, no worries. You might want to delay to reduce work at will :)

kparichay: its minor, so its fine.

nntrainer/models/neuralnet.cpp
kparichay: great, then I will let #1544 resolve it.

nntrainer/models/neuralnet.cpp
kparichay: done.

jijoongmoon (approved):
LGTM

PR #1531, opened 2021-08-27
[cleanup/fix] Cleanup + bugfix

kparichay:
Number of commits to review in this PR : 2

This patch provides cleanup for manager and related classes:
- cleanup for Manager
- cleanup for Var_Grad and Weights
Fixes:
- Graphcore,NetworkGraph,NeuralNet destructor fixed to not clear their
lists until they are destructed, as this caused bug in copy constructor
- initialize and allocation for tensors are merged for model, graph and
manager interface. This removes unnecessary confusion and possible bugs
from these classes
- Pass appropriate start and end execution orders to manager for
allocation
- Support setInputs to set multiple inputs
- clean inputs/labels which are allocated by the dataset, other manager
tries to clear them. This is temporary till dataset is not using the
manager
- memory pool to start giving tokens from 1 instead of 0. token 0 is
treated at a non-requested tensor memory
- add interface to check if the tensor pool and memory pool have
allocated memory
- initialize all the tensors token to 0, when finalizing multiple times
- tensor behavior update to not allow updateBatch() if it is allocated
- update setBatch to deallocate already allocated tensors, then update
the batch size, and then initialize and allocate the tensors

TODO:
- add a enableMemoryOptimization() option

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -142,8 +142,7 @@ int NeuralNetwork::initialize() {
     model_graph.requestOptimizerVariable(cb, true);
   }
 
-  // Allocate and initialize weights
-  model_graph.initializeWeights();
+  // Allocate weights
   model_graph.allocateWeights();
 
   if (in_place_optimization) {
@@ -162,7 +161,7 @@ int NeuralNetwork::initialize() {
 /**
  * @brief     free layers
  */
-NeuralNetwork::~NeuralNetwork() { model_graph.reset(); }
+NeuralNetwork::~NeuralNetwork() = default;
 
 static void setLabels(const std::vector<Tensor> &data,
                       const std::vector<Var_Grad *> &label_list) {
@@ -187,14 +186,18 @@ static void setLabels(const std::vector<Tensor> &data,
 static void setInputs(const std::vector<Tensor> &data,
                       const std::vector<Var_Grad *> &input_list) {
 
-  NNTR_THROW_IF(data.size() != input_list.size(), std::invalid_argument)
+  NNTR_THROW_IF(data.size() > 1 && data.size() != input_list.size(),
+                std::invalid_argument)
     << "input size does not match with the network requirements"
     << " input size: " << data.size()
     << " requirements size: " << input_list.size();
 
   /// feed or clear label
   for (unsigned int idx = 0; idx < input_list.size(); idx++) {
-    input_list[idx]->initializeVariable(data[idx]);
+    if (data.empty())
+      input_list[idx]->initializeVariable();
+    else
+      input_list[idx]->initializeVariable(data[idx]);
   }
 }
 
@@ -532,7 +535,7 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
   if (!validateInput(X))
     throw std::invalid_argument("Input validation failed.");
 
-  allocate(false);
+  allocate(ExecutionMode::INFERENCE);
 
   START_PROFILE(profile::NN_FORWARD);
   out = forwarding(X, {}, false);
@@ -545,6 +548,10 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
      * Weights of the model will be freed when the model is destroyed.
      */
     model_graph.deallocateTensors(false);
+
+  /** Clear the set inputs and labels */
+  setLabels({}, model_graph.getLabelList());
+  setInputs({}, model_graph.getInputList());
 
   return out;
 }
@@ -578,13 +585,10 @@ int NeuralNetwork::setDataset(const DatasetModeType &mode,
   return setDataBuffer(mode, std::static_pointer_cast<DataBuffer>(dataset));
 }
 
-int NeuralNetwork::allocate(bool trainable) {
-  model_graph.deallocateTensors(false);
+int NeuralNetwork::allocate(ExecutionMode mode) {
+  model_graph.deallocateTensors();
+  model_graph.allocateTensors(mode);
 
-  // TODO: directly replace this
-  model_graph.initializeTensors(trainable);
-
-  model_graph.allocateTensors();
   return ML_ERROR_NONE;
 }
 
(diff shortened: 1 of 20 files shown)


nntrainer/tensor/tensor_pool.cpp
```diff
@@ -0,0 +1,229 @@
...
+    if (isTensorLongTerm(spec.lifespan)) {
+      validity_start = start_order;
+      validity_end = end_order;
+    }
+
+    /** 2. for each tensor request if it is in the provided range */
+    if (validity_end < start_order || validity_start > end_order)
+      continue;
+    validity_start = std::max(validity_start, start_order);
+    validity_end = std::max(validity_end, end_order);
```
kparichay, line 118: this must std::min - refer https://github.com/nnstreamer/nntrainer/pull/1511#discussion_r716353620

kparichay: This is handled at https://github.com/nnstreamer/nntrainer/pull/1541/files#r717205648.

nntrainer/tensor/weight.h
```diff
@@ -178,13 +180,13 @@ class Weight : public Var_Grad {
    *
    * @note New dimension must maintain the shape of the variable
    */
-  void reset(const TensorDim &dim, const Tensor::Initializer init,
-             const WeightRegularizer reg, const float reg_const, bool ng) {
-    regularizer = reg;
-    regularizer_constant = reg_const;
+  // void reset(const TensorDim &dim, const Tensor::Initializer init,
```
kparichay, line 183: will be deleted with rebase.

nntrainer/tensor/var_grad.h
```diff
@@ -235,16 +212,16 @@ class Var_Grad {
    *
    * @return Cloned copy
    */
-  Var_Grad clone() const {
-    Var_Grad vg(*this);
+  // Var_Grad clone() const {
```
kparichay, line 215: will be deleted with rebase.

nntrainer/tensor/var_grad.h
```diff
@@ -254,17 +231,17 @@ class Var_Grad {
...
-      var->reshape(dim);
-    var->initialize(init);
-
-    if (ng && !grad->empty())
-      grad->reshape(dim);
-    else
-      grad = std::make_shared<Tensor>(grad->getName());
-    resetGradient();
-  }
+  // void reset(const TensorDim &dim, Tensor::Initializer init, bool ng) {
```
kparichay, line 234: will be deleted with rebase.

nntrainer/tensor/var_grad.h
kparichay: done.

nntrainer/tensor/var_grad.h
kparichay: done.

nntrainer/tensor/weight.h
kparichay: done.

nntrainer/models/neuralnet.cpp
```diff
@@ -187,14 +186,18 @@ static void setLabels(const std::vector<Tensor> &data,
...
     << " input size: " << data.size()
     << " requirements size: " << input_list.size();
 
   /// feed or clear label
   for (unsigned int idx = 0; idx < input_list.size(); idx++) {
-    input_list[idx]->initializeVariable(data[idx]);
+    if (data.empty())
+      input_list[idx]->initializeVariable();
+    else
+      input_list[idx]->initializeVariable(data[idx]);
```
zhoonit, line 200: as now idx is depending on input_list, it looks valid to do

```cpp
  else(data.size() == 1){
   input_list[idx]->initializeVariable(data[0]);
  }
```

nntrainer/models/neuralnet.cpp
```diff
@@ -792,6 +796,10 @@ int NeuralNetwork::train_run() {
                         update_eval_stat, eval_epoch_end);
   }
 
+  /** Clear the set inputs and labels */
+  setLabels({}, model_graph.getLabelList());
+  setInputs({}, model_graph.getInputList());
```
zhoonit, line 801: Wonder if we have to clear?

zhoonit: is this deallocated input,label?

nntrainer/tensor/manager.h
```diff
@@ -111,10 +111,7 @@ class Manager {
   /**
    * @brief     Constructor of Manager
    */
-  Manager(bool enable_gradient_memory_opt_ = true,
-          bool enable_derivative_memory_opt_ = false,
-          bool enable_activation_memory_opt_ = false,
-          bool enable_inference_inout_memory_opt_ = false);
+  Manager() = default;
```
zhoonit, line 114: static analyzer is likely to unlike this. (I think it was because of some primitive default initializers)

nntrainer/models/neuralnet.cpp
kparichay: This is done in the upcoming PR - https://github.com/nnstreamer/nntrainer/pull/1544/files#diff-8136da3afc181f3163a8f8a15d86d132abee8c07de2c9b49bc2cf0759e380a69R833-R845. If you don't mind, I will let it be resolved in the upcoming PR itself.

nntrainer/models/neuralnet.cpp
kparichay: The purpose is to unset the tensors which are not valid anymore as the dataset can free them. This is for maintaining the semantics.

nntrainer/tensor/manager.h
kparichay: All manager's members are now either vector or TensorPool and both have good default constructors, so I think it will be fine. If this causes issues, I will change.

nntrainer/models/neuralnet.cpp
zhoonit: I don't mind :)

nntrainer/models/neuralnet.cpp
zhoonit: Okay

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1534, opened 2021-08-28
[ini] Implement model save

zhoonit:
- [ini] Implement model save

```
This patch implements model section save in ini format

**Side Changes proposed in this PR:**
- Move `THROW_STATUS` and `RETURN_STATUS` to util_func.h

**Self evaluation:**

```

- [ini] Migrate to properties for model

```
This patch migrate to properties for neuralnet

**Semantic Changes**
- NeuralNetwork::batch_size does not change when actual batchsize change
from neuralnetwork instead, model graph keeps batch size separately.

- Move members to props in model
- Move inline declaration to declaration for some functions
- Remove unused member/functions from parse_util / neuralnet

**Self evaluation:**

```

- [ini] Prepare model properties

```
This patch prepare model properties in `model_common_properties`

**Additional Changes proposed in this PR:**
- Add PositiveIntegerProperty base class
- Fix copy assignment/ctor in base_properties

**Self evaluation:**

```

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -28,13 +28,13 @@
 
 #include <databuffer.h>
 #include <ini_interpreter.h>
+#include <ini_wrapper.h>
 #include <model_loader.h>
 #include <neuralnet.h>
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <node_exporter.h>
 #include <optimizer_context.h>
-#include <parse_util.h>
 #include <profiler.h>
 #include <util_func.h>
 
@@ -49,8 +49,8 @@ namespace nntrainer {
 
 NeuralNetwork::NeuralNetwork(AppContext app_context_, bool in_place_opt) :
   model_props(props::LossType()),
-  model_flex_props(props::Epochs(), props::BatchSize(), props::SavePath()),
-  continue_train(false),
+  model_flex_props(props::Epochs(), props::TrainingBatchSize(),
+                   props::SavePath(), props::ContinueTrain()),
   load_path(std::string()),
   epoch_idx(0),
   iter(0),
@@ -123,7 +123,8 @@ int NeuralNetwork::initialize() {
 
   ml_logd("initializing neural network, layer size: %d", n_layers);
 
-  model_graph.setBatchSize(std::get<props::BatchSize>(model_flex_props));
+  model_graph.setBatchSize(
+    std::get<props::TrainingBatchSize>(model_flex_props));
 
   status = model_graph.initialize();
   NN_RETURN_STATUS();
@@ -414,11 +415,27 @@ NeuralNetwork &NeuralNetwork::copy(NeuralNetwork &from) {
 }
 
 void NeuralNetwork::saveModelIni(const std::string &file_path) {
-  IniGraphInterpreter interpreter;
+  NNTR_THROW_IF(isFileExist(file_path), std::invalid_argument)
+    << "There is already a file, overriding to the exisiting file is not "
+       "permitted, path: "
+    << file_path;
 
-  /// @note this is to ensure permission checks are done
-  checkedOpenStream<std::ofstream>(file_path, std::ios::out);
-  /// @todo serialize model props
+  IniSection model_section("model");
+  model_section.setEntry("type", "NeuralNetwork");
+
+  Exporter e;
+  e.saveResult(model_props, ExportMethods::METHOD_STRINGVECTOR, this);
+  e.saveResult(model_flex_props, ExportMethods::METHOD_STRINGVECTOR, this);
+
+  const auto key_val_pairs = e.getResult<ExportMethods::METHOD_STRINGVECTOR>();
+  for (const auto &pair : *key_val_pairs) {
+    model_section.setEntry(pair.first, pair.second);
+  }
+
+  IniWrapper wrapper("model_saver", {model_section});
+  wrapper.save_ini(file_path);
+
+  IniGraphInterpreter interpreter;
   /// @todo serialize dataset props
   /// @todo serialize optimizer props
   interpreter.serialize(model_graph, file_path);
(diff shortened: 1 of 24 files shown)


nntrainer/models/neuralnet.h
```diff
@@ -518,6 +482,17 @@ class NeuralNetwork : public ml::train::Model {
...
+  using FlexiblePropTypes =
+    std::tuple<props::Epochs, props::BatchSize, props::SavePath>;
+  using RigidPropTypes = std::tuple<props::LossType>;
+
+  RigidPropTypes model_props;         /**< model props */
+  FlexiblePropTypes model_flex_props; /**< model train props */
+  props::ContinueTrain
+    continue_train; /**< true if epochs and iteration should be retained during
+                       separate model run */
+  std::string load_path; /**< path to load weights when initialize  */
```
kparichay, line 494: load_path and save_path are different?

nntrainer/models/neuralnet.cpp
```diff
@@ -245,13 +201,14 @@ sharedConstTensors NeuralNetwork::forwarding(bool training) {
 sharedConstTensors NeuralNetwork::forwarding(sharedConstTensors input,
                                              sharedConstTensors label,
                                              bool training) {
-
-  NNTR_THROW_IF(input[0]->batch() != batch_size ||
-                  (!label.empty() && label[0]->batch() != batch_size),
+  auto current_batch = model_graph.getBatchSize();
```
kparichay, line 204: this creates two cache values for batch size - one in props::BatchSize and another in model_graph.batch_size.
I don't think network_graph needs to cache batch_size.

nntrainer/models/neuralnet.h
zhoonit: load_path is from load(INI_WITH_BIN)

save_path can change later by properties, but should not override load_path.

nntrainer/models/neuralnet.cpp
zhoonit: so, props::BatchSize is a hint (for training). Single source of true value lies in model_graph.getBatchSize(). With this setup, model::inference doesn't change training batch size.

This will be needed if we are going to change only validation batchsize as well.

kparichay: if the batch size in the graph created (model graph) is cached for training, then it should be `training_batch_size`.
Given model_graph is abstracted from training/validation/inference, storing a batch_size which mismatches with the current set batch_size in the model_graph is not a good thing.
If you intend to cache the various types of batch size, IMO they are best done in neuralnetwork class.

nntrainer/models/neuralnet.h
kparichay: ok.

zhoonit: This won't be property and file check is being done at opening so I guess string fits here.

nntrainer/models/neuralnet.cpp
zhoonit: graph always stores a batch size with currently set batch size.
We can have various types of batch size cached in neuralnet (in the future) but that doesn't mean it removes the need of batchsize in the graph.
We need to query the batchsize from the model graph.

kparichay: the graph also needs to store the current execution mode in which it is (train, inference or validation). In the current implementation, graph is kept abstract from the execution mode, but to store the batch size, graph must then store this as well.

Also, if a graph/model runs inference with batch size 2, then the batch size of the graph/model is 2 (although the batch size of the last train might be 32).

zhoonit: we discussed offline to change name to training batchsize to make it clear.

jijoongmoon (approved):
LGTM.

jijoongmoon:
there might be differ between saved ini file and ini file from user. I think we need to describe or mention somewhere.

nntrainer/models/neuralnet.h
```diff
@@ -518,6 +482,17 @@ class NeuralNetwork : public ml::train::Model {
   void disableDynamicFineTuning() { dynamic_training_opt.disable(); }
 
 private:
+  using FlexiblePropTypes =
+    std::tuple<props::Epochs, props::BatchSize, props::SavePath>;
+  using RigidPropTypes = std::tuple<props::LossType>;
+
+  RigidPropTypes model_props;         /**< model props */
+  FlexiblePropTypes model_flex_props; /**< model train props */
+  props::ContinueTrain
```
lhs8928, line 491: Is there any reason that ContinueTrain is not included in FlexiblePropTypes?

nntrainer/utils/ini_wrapper.cpp
```diff
@@ -116,15 +118,12 @@ void IniWrapper::updateSections(const Sections &sections_) {
...
-  std::ofstream out(ini_name.c_str(), std::ios_base::out);
-  NNTR_THROW_IF(!out.good(), std::runtime_error) << "cannot open ini";
+  auto out = checkedOpenStream<std::ofstream>(ini_name, std::ios_base::app);
 
   for (auto &it : sections) {
     it.print(out);
     out << std::endl;
   }
-
-  out.close();
```
lhs8928, line 127: Is it intended?

lhs8928 (approved):
LGTM

nntrainer/models/neuralnet.h
zhoonit: No, it has to be part of model_flex_props, I'll fix

zhoonit:
> there might be differ between saved ini file and ini file from user. I think we need to describe or mention somewhere.

I'll update the nntrainer.h to wan this :)

nntrainer/utils/ini_wrapper.cpp
zhoonit: Yes, destructor calls this function so it's redaudant

nntrainer/models/neuralnet.h
zhoonit: Updated :) Nice catch!

kparichay (approved):
LGTM

PR #1535, opened 2021-08-28
[NeuralNet] Fix testset run

zhoonit:
- [NeuralNet] Fix testset run

```
This patch fixes testset running while cleaning up the code to reuse
repeating logics with callback

**Self evaluation:**

```

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -632,25 +632,26 @@ int NeuralNetwork::train_run() {
     return ML_ERROR_INVALID_PARAMETER;
   }
 
-  for (epoch_idx = epoch_idx + 1; epoch_idx <= epochs; ++epoch_idx) {
-    training.loss = 0.0f;
-
+  /**
+   * @brief run a single epoch with given callback, @a auto is used instead of
+   * std::function for performance measure
+   * @param buffer buffer to run
+   * @param shuffle whether to shuffle or not
+   * @param on_iteration_fetch function that will recieve reference to stat,
+   * buffer which will be called every time data is fetched and set
+   * @param on_epoch_end function that will recieve reference to stat,
+   * buffer which will be called on the epoch end
+   */
+  auto run_epoch = [this, &in, &label, &in_dims, &label_dims](
+                     DataBuffer *buffer, bool shuffle,
+                     auto &&on_iteration_fetch, auto &&on_epoch_end) {
+    /// @todo managing metrics must be handled here as well!! for now it is
+    /// handled in individual callbacks
+    RunStats stat;
     std::future<std::shared_ptr<IterationQueue>> future_iq =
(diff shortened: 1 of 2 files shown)


nntrainer/models/neuralnet.h
```diff
@@ -63,10 +63,12 @@ using DatasetModeType = ml::train::DatasetModeType;
  * @brief     Statistics from running or training a model
  */
 typedef struct RunStats_ {
-  float accuracy; /** accuracy of the model */
-  float loss;     /** loss of the model */
+  float accuracy;     /** accuracy of the model */
+  float loss;         /** loss of the model */
+  int count;          /** number of iterations done on this stat */
```
kparichay, line 68: rename to `num_iterations`?

nntrainer/models/neuralnet.h
```diff
@@ -63,10 +63,12 @@ using DatasetModeType = ml::train::DatasetModeType;
...
+  unsigned int right; /** number of right sample on this run */
```
kparichay, line 69: rename to `num_correct_predictions`? `right` is a bit vague.

nntrainer/models/neuralnet.cpp
```diff
@@ -632,25 +632,26 @@ int NeuralNetwork::train_run() {
...
+   * @brief run a single epoch with given callback, @a auto is used instead of
+   * std::function for performance measure
+   * @param buffer buffer to run
+   * @param shuffle whether to shuffle or not
+   * @param on_iteration_fetch function that will recieve reference to stat,
+   * buffer which will be called every time data is fetched and set
+   * @param on_epoch_end function that will recieve reference to stat,
+   * buffer which will be called on the epoch end
+   */
+  auto run_epoch = [this, &in, &label, &in_dims, &label_dims](
```
kparichay, line 645: 💯

kparichay: you can consider making it a private function itself than a lambda as there will be eventually many hooks supported for this.

zhoonit: I agree, But i'll delay for now :)

nntrainer/models/neuralnet.h
zhoonit: Done

nntrainer/models/neuralnet.h
zhoonit: Done

nntrainer/models/neuralnet.cpp
kparichay: IMO, the function will most likely be a static function which will additionally take `NetworkGraph` object, and return stats as now, so can remain in .cpp.

zhoonit: If it remains in .cpp with static, I guess this can stay here until we have the need.

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1539, opened 2021-09-02
[wait for #1531] [resnet] Enable resnet application + Ordered tensorlist

### LINKED ISSUE #1533: Enable resnet application
- solve the fail reason

This is need to verify #1127.

kparichay:
Number of commits to review for this patch : 2

This patch enables resnet application with some fixes:
- setBatchSize updated for handling gradient tensors
- added input layer for resnet
- passed arguments correctly for resnet test application
- tensor allocation for ENABLE_TEST mode updated

This patch also makes tensor list in tensorpool ordered.
This ensures that the requested tensors are initialized in the order of
their requests which is by the order of the sorted graph.

See Also #1127
Resolves #1533

This is now a trivial PR as these commits had been cherry-picked earlier while rebasing.

### DIFF
--- a/Applications/Resnet/jni/meson.build
+++ b/Applications/Resnet/jni/meson.build
@@ -21,7 +21,6 @@ e = executable('nntrainer_resnet18',
   install_dir: application_install_dir
 )
 
-
 if get_option('enable-long-test')
   testenv = environment()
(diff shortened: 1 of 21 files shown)


kparichay:
merged with #1540 as this has become trivial PR.

PR #1540, opened 2021-09-03
[planner] Add optimized v1 planner 

### LINKED ISSUE #1533: Enable resnet application

kparichay:
- This patch adds an optimized v1 planner for memory sharing.
- This planner assigns memory in the order of start of the validity,
and then by decreasing the order of the end of the validity. This matches
the memory use pattern while training the model.
This planner is supposed to work better for training than for inference.
- Also added unit tests for the optimized v1 planner.

Merged changes from #1539 (as it became a trivial PR after rebase):
- This patch enables resnet application with some fixes. tensor allocation for ENABLE_TEST mode updated.

See Also #1127
Resolves #1533

### DIFF
--- a/Applications/Resnet/jni/meson.build
+++ b/Applications/Resnet/jni/meson.build
@@ -21,7 +21,6 @@ e = executable('nntrainer_resnet18',
   install_dir: application_install_dir
 )
 
-
 if get_option('enable-long-test')
   testenv = environment()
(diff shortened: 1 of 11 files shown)


zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1541, opened 2021-09-03
[graph/manager] bug fix + Enable memory v1 optimizations 

kparichay:
- Flatten layer in-place implementation manipulates the tensor stored in
its input and output. This violates the execution order usage of the
tensors and does not work with the memory planning setup.
This patch makes flatten layer to work out-of-place with a copy.
There will soon be a patch to make flatten work in-place with proper
method.
- Added bug fix in the execution order calculation where some of the
execution values were being missed in the previous implementation.
- Optimized v1 planner is updated to reuse expired allocations which are
not at the top the sorted list. This is a heuristic which works well for
training memory usage scenario.
The unittest is also updated to be more generic when checking the
interval overlap with the allocations.
-This patch adds the corresponding bug fixes:
  - end validity for each requested tensor is fixed
  - execution order for requestInputs and requestOutputs are fixed
  - output is specially handled for activation layer as it uses output in
derivative calculation
  - gradient of the weights is kept valid for calcDerivative as well
because gradients are applied to weights after calcDerivative.
- This patch adds interface to enable memory optimizations with the neural
network. Enabling the interface changes the planner being used for the
memory allocation.
With this patch, OptimizedV1Planner is put to use when enabling
optimizations.
Unittest of models is updated to disable optimizations in the
non-optimized test cases.

See Also #1127
The memory usage comparison with this patch is noted at https://github.com/nnstreamer/nntrainer/issues/1127#issuecomment-911320879

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -129,7 +129,7 @@ MMapedMemory::~MMapedMemory() noexcept {
 
 void Manager::allocateWeights(unsigned int max_exec_order_) {
   if (!weight_pool.isAllocated()) {
-    weight_pool.finalize(BasicPlanner(), 0, max_exec_order_);
+    finalizeTensorPool(weight_pool, 0, max_exec_order_);
     weight_pool.allocate();
   }
 }
@@ -143,7 +143,7 @@ void Manager::allocateTensors(unsigned int max_exec_order_) {
   allocateWeights(max_exec_order_);
 
   if (!tensor_pool.isAllocated()) {
-    tensor_pool.finalize(BasicPlanner(), 0, max_exec_order_);
+    finalizeTensorPool(tensor_pool, 0, max_exec_order_);
     if (tensor_pool.minMemoryRequirement() > 0)
       tensor_pool.allocate();
   }
@@ -519,6 +519,14 @@ std::vector<Weight *> Manager::getWeights() {
   }
 
   return all_weights;
+}
+
+void Manager::finalizeTensorPool(TensorPool &pool, unsigned int start,
+                                 unsigned int end) {
+  if (enable_optimizations)
+    pool.finalize(OptimizedV1Planner(), start, end);
+  else
+    pool.finalize(BasicPlanner(), start, end);
 }
 
 } // namespace nntrainer
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -29,10 +29,12 @@
 #include <unistd.h>
 #include <vector>
 
+#include <activation_layer.h>
 #include <basic_planner.h>
 #include <layer_node.h>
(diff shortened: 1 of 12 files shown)


nntrainer/tensor/tensor_pool.cpp
```diff
@@ -115,7 +115,7 @@ void TensorPool::finalize(const MemoryPlanner &planner,
     if (validity_end < start_order || validity_start > end_order)
       continue;
     validity_start = std::max(validity_start, start_order);
-    validity_end = std::max(validity_end, end_order);
+    validity_end = std::min(validity_end, end_order);
```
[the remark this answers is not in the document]
zhoonit: awesome!

nntrainer/tensor/manager.cpp
```diff
@@ -254,10 +256,13 @@ std::vector<Weight *>
 Manager::requestWeights(const GraphNode &node,
                         const std::vector<Weight::Spec> &weights_spec) {
   const auto &exec_order = node.getExecutionOrder();
```
zhoonit, line 258: You can later consider below for brevity
```cpp
const auto [forwarding, calcGradient, calcDerivateiv] = node.getExecutionOrder();
```

nntrainer/tensor/manager.cpp
```diff
@@ -440,9 +446,13 @@ Manager::requestOutputs(const GraphNode &node,
                         const std::vector<TensorDim> &outputs_dim) {
   const auto &exec_order = node.getExecutionOrder();
   std::vector<unsigned int> var_exec_order(
-    {std::get<0>(exec_order), std::get<2>(exec_order)});
+    {std::get<0>(exec_order)}); /** forwarding */
+  if (node.getType() == ActivationLayer::type)
+    /** TODO: if removing this reduces memory consumption, resolve this */
+    var_exec_order.push_back(std::get<2>(exec_order)); /** calcDerivative */
```
zhoonit, line 452: I wonder what the bug was related to this?
does var only need to be saved for the output?

zhoonit (approved):
LGTM

nntrainer/tensor/manager.cpp
kparichay: activation layer depends on its output to reduce the computation, but a general layer depends on its input.

nntrainer/tensor/manager.cpp
kparichay: I will add this in a separate PR.

nntrainer/tensor/manager.cpp
zhoonit: I see

jijoongmoon (approved):
LGTM

nntrainer/tensor/manager.cpp
kparichay: done.

PR #1544, opened 2021-09-07
[wait for #1541] [inplace opt] Support in-place flatten and input layer 

kparichay:
- This patch updates the flatten layer to be a no-op layer. This is done
with the flatten layer setting the input and output shapes at finalize
time and making flatten layer execute in-place. Changes in this patch:
  1. requestPreallocatedTensor() in TensorPool now returns a new tensor
which will eventually share the memory with the preallocated tensor than
returning the preallocated tensor itself. This allows tensor metadata to
be changed (like name, shape, etc) which sharing the memory. This is
done by storing the dependency link between tensors in token.
Corresponding unittests are also added.
  2. Manager now supports giving shared tensors for outputs (shared with
some inputs) to support in-place running of some layers.
  3. Flatten layer is updated to be a basic no-op and to perform
flattening once at compile time.
  4. Update flatten layer supportBackwarding to true
  5. Input layer updated to not edit tensor mapping. Input layer will be
updated to be in-place in the next patch.

- This patch makes input layer work inplace. This is done by support of
externally allocated tensors in tensorPool, and making input of input
layer and labels to be externally allocated tensors.
Input layer is updated to work in-place.
Further the methodology to set inputs and labels has also been updated.


kparichay:
With #1531, certain temporary changes are made till externally allocated tensors are made available, undo them here.

test/unittest/memory/memory_planner_validate.cpp
```diff
@@ -171,8 +226,11 @@ TEST_P(MemoryPlannerValidate, partial_overlap) {
               std::accumulate(memory_size.begin(), memory_size.end(), 0u));
     EXPECT_TRUE(validateNoOverlap(memory_size, memory_offset, pool_size));
   } else {
-    EXPECT_EQ(pool_size,
+    EXPECT_GE(pool_size,
               *std::max_element(memory_size.begin(), memory_size.end()));
-    // EXPECT_TRUE(validateIntervalOverlap(memory_size, memory_offset));
+    EXPECT_LE(pool_size,
+              std::accumulate(memory_size.begin(), memory_size.end(), 0));
```
zhoonit, line 232: could you do s/0/0u/ here?

kparichay: I will.

nntrainer/tensor/manager.cpp
```diff
@@ -409,6 +416,28 @@ Manager::requestInputs(const GraphNode &node,
...
+#else
+      grad = tensor_pool.requestExternallyAllocateTensor(
+        dim,                              /// tensor dim
+        var_name + Var_Grad::grad_suffix, /// name
+        Tensor::Initializer::ZEROS        /// tensor initializer
+      );
+#endif
     }
 
     /**
```
zhoonit, line 443: Wonder if this is fixed?

kparichay: can you please check the link - its does not point to the selected change.

zhoonit: The line is right below this comment. (unable to comment on that line)
https://github.com/nnstreamer/nntrainer/blob/f2ab7f8b8cce3eaae296f22017138386fab6e6eb/nntrainer/tensor/manager.cpp#L444-L446

kparichay:
apply https://github.com/nnstreamer/nntrainer/pull/1541#discussion_r720322724

zhoonit (approved):
LGTM

test/unittest/memory/memory_planner_validate.cpp
kparichay: done.

nntrainer/tensor/manager.cpp
kparichay: done.

test/unittest/datasets/data_producer_common_tests.cpp
```diff
@@ -24,7 +24,6 @@ createSample(const std::vector<nntrainer::TensorDim> &input_dims,
   using namespace nntrainer;
   auto populate_tensor = [](const TensorDim &dim) {
     Tensor t(dim);
-    t.updateBatch(1);
```
zhoonit, line 27: could you let it be disabled? I'll fix it.

kparichay: removed.

jijoongmoon (approved):
LGTM

PR #1545, opened 2021-09-07
[wait for #1544] [conv] Update temporary memory requests 

kparichay:
This patch updates the request for temporary memory in the convolution
layer.
- im2col and col2im results are both the same size and used exclusive of
each other but both are requested for the backwarding. so, instead of
requesting both, they can share their memories.
- as the values in these tensors can be discarded between forwarding and
backwarding, two independent tensors are requested for forwarding and
backwarding so that the memory can be reused in the intermediate
duration.


jijoongmoon (approved):
LGTM

PR #1546, opened 2021-09-07
[Ini Save] input shape property

zhoonit:
- [Ini Save] input shape property

```
This patch includes input shape property generation

```

- [LayerNode] Change context to be RAII

```
This patch refactors context to be RAII to make layer node properties
dumb.

**Changes**
- init context is removed
   - layer node instead created and returned from layerNode::finalize(input_dims)
   - remove dependency to input dimension before initialize
   - layerNode now has input_shapes instead of setting directly from initContext
   - runcontext is used to query things
   - fix multiple bugs regarding the validity of layer node

- networkGraph::updateRunContext() -> finalizeContext() now
finalizes inside this function for brevity
- layerNode::updateRunContext() -> configureContext() to make it RAIIer.
- minor code cleans in networkGraph::initialize

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -141,8 +141,9 @@ createLayerNode(std::unique_ptr<nntrainer::Layer> &&layer,
 
 LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l) :
   layer(std::move(l)),
-  finalized(false),
   activation_type(ActivationType::ACT_NONE),
+  run_context(nullptr),
+  input_shapes(),
   layer_node_props(new PropsType(props::Name(), props::Flatten(),
                                  props::Distribute(), props::Trainable(), {})),
   loss(new props::Loss()),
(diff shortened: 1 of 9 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -373,7 +374,14 @@ void LayerNode::setInputLayers(const std::vector<std::string> &layers) {
...
 }
 
-bool LayerNode::hasInputShapeProperty() const { return !input_shapes.empty(); }
+bool LayerNode::hasInputShapeProperty() const {
+  auto &input_shapes =
+    std::get<std::vector<props::InputShape>>(*layer_node_props);
+
+  return !input_shapes.empty() &&
+         std::all_of(input_shapes.begin(), input_shapes.end(),
+                     [](const auto &input) { return !input.empty(); });
```
kparichay, line 383: is this possible? this means either setting shape of 0 is possible, or we are resizing out input shape without filling it.

zhoonit: This is 99.99% impossible, but as this function is not in critical path, I made it to cover some impossible scenario :p

nntrainer/layers/layer_node.cpp
```diff
@@ -432,17 +440,27 @@ void LayerNode::save(std::ofstream &file) const {
...
         << "calculated input dimension is different from given input_shape "
            "property";
     }
   } else {
-    actual_input_dims = prop_dims;
+    NNTR_THROW_IF(!hasInputShapeProperty(), std::invalid_argument)
+      << "if input dims not given, input shapes must be given by the user as "
+         "property";
+    NNTR_THROW_IF(prop_dims.size() != 1, std::invalid_argument)
+      << "input shapes must be one if connection is not given but given: "
```
kparichay, line 460: is not given but given?

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.cpp
zhoonit: Updated,
``` << "input shapes must be one if connection is not given but given input shapes size of: " ```

nntrainer/layers/layer_node.cpp
kparichay: cool.

zhoonit:
This is being closed to be merged with #1547

PR #1547, opened 2021-09-07
[props] Support enum properties

zhoonit:
- [props] Support enum properties

```
This patch aims to support enum based properties, please refer to the
test to see how it works.

**Self evaluation:**

```

- [Ini Save] input shape property

```
This patch includes input shape property generation

```

- [LayerNode] Change context to be RAII

```
This patch refactors context to be RAII to make layer node properties
dumb.

**Changes**
- init context is removed
   - layer node instead created and returned from layerNode::finalize(input_dims)
   - remove dependency to input dimension before initialize
   - layerNode now has input_shapes instead of setting directly from initContext
   - runcontext is used to query things
   - fix multiple bugs regarding the validity of layer node

- networkGraph::updateRunContext() -> finalizeContext() now
finalizes inside this function for brevity
- layerNode::updateRunContext() -> configureContext() to make it RAIIer.
- minor code cleans in networkGraph::initialize

**Self evaluation:**

```

### DIFF
--- a/nntrainer/utils/base_properties.h
+++ b/nntrainer/utils/base_properties.h
@@ -261,6 +261,17 @@ private:
 };
 
 /**
+ * @brief enum property
+ *
+ * @tparam T underlying type info to query enum_info
+ */
+template <typename EnumInfo>
+class EnumProperty : public Property<typename EnumInfo::Enum> {
+public:
+  static EnumInfo enum_info_;
+};
+
+/**
  * @brief abstract class for positive integer
  *
  */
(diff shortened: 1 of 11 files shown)


nntrainer/utils/base_properties.h
```diff
@@ -260,6 +260,23 @@ template <typename T> class Property {
...
+ */
+template <typename T> struct enum_info {};
+
+/**
+ * @brief enum property
+ *
+ * @tparam T underlying type to query enum_info
+ */
+template <typename T>
+class EnumProperty : public Property<typename enum_info<T>::Enum> {};
```
zhoonit, line 278: This should be changed just EnumInfo types, and injected to the properties to make it reusable.

test/unittest/unittest_base_properties.cpp
```diff
@@ -106,6 +106,38 @@ class PtrOfBanana : public nntrainer::Property<int *> {
...
+  enum class Enum {
+    Cavendish = 0,
+    Plantain = 1,
+    Manzano = 2,
+  };
+
+  static constexpr std::initializer_list<Enum> EnumList = {
+    Enum::Cavendish, Enum::Plantain, Enum::Manzano};
+
+  static constexpr const char *EnumStr[] = {"Cavendish", "Plantain", "Manzano"};
```
jijoongmoon, line 128: I wonder we need to set this every enum variable then.

jijoongmoon (approved):
LGTM with minor comment.

test/unittest/unittest_base_properties.cpp
zhoonit: Yes, there is no way to get string information except dirty methods (using some hacky macros)

kparichay: this is ok for now but consider moving to map/unoredered_map for enums where values are not continuously defined (which happens in layer types).

kparichay (approved):
LGTM

test/unittest/unittest_base_properties.cpp
zhoonit: Right, to iterate over it,
below information is forced to be inserted as well.
```
static constexpr std::initializer_list<Enum> EnumList = {
    Enum::Cavendish, Enum::Plantain, Enum::Manzano};
```

I used two primitives to make it constexpr but maps also works :)

lhs8928 (approved):
LGTM
By the way, can we remove validate function in layer context then?

zhoonit:
> By the way, can we remove validate function in layer context then?
let's not remove it until we are sure :)

PR #1548, opened 2021-09-07
[Wait for #1547] [Props] Implement activation property

zhoonit:
**[Pending commit: #1547]**

- [Props] Implement activation property

```
This patch implements activation property by enum

**Additional changes**
- TimeDistribute property is now handled at finalize()

**Self evaluation:**

```

### DIFF
--- a/nntrainer/layers/activation_layer.h
+++ b/nntrainer/layers/activation_layer.h
@@ -15,10 +15,17 @@
 #define __ACTIVATION_LAYER_H__
 #ifdef __cplusplus
 
+#include <memory>
+
 #include <acti_func.h>
 #include <layer_devel.h>
 
 namespace nntrainer {
+
+namespace props {
+class ActivationType;
+
+} // namespace props
 
 /**
  * @class   Activation Layer
@@ -30,9 +37,7 @@ public:
   /**
    * @brief     Constructor of Activation Layer
    */
-  ActivationLayer() : Layer() {
-    acti_func.setActiFunc(ActivationType::ACT_NONE);
-  }
+  ActivationLayer();
 
   /**
    * @brief     Destructor of Activation Layer
@@ -62,8 +67,7 @@ public:
   /**
    * @copydoc Layer::exportTo(Exporter &exporter, ExportMethods method)
    */
-  void exportTo(Exporter &exporter,
-                const ExportMethods &method) const override {}
+  void exportTo(Exporter &exporter, const ExportMethods &method) const override;
 
   /**
    * @copydoc Layer::getType()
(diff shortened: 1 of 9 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -184,75 +180,10 @@ LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l) :
...
-    break;
-  }
-  default:
-    return false;
-  }
-
-  return true;
+  left_properties =
+    loadProperties(left_properties, *layer_node_props_realization);
+  layer->setProperty(left_properties);
```
jijoongmoon, line 186: AFAIK, TimeDistLayer is added when the property is set. Looks like this is deleted.

nntrainer/layers/layer_node.cpp
```diff
@@ -474,6 +411,25 @@ InitLayerContext LayerNode::finalize(const std::vector<TensorDim> &input_dims) {
...
 
+  /** manipulate layers if required */
+  if (getType() == ActivationLayer::type) {
+    auto &act_prop = std::get<props::Activation>(*layer_node_props_realization);
+    if (!act_prop.empty()) {
+      layer->setProperty({"activation=" + to_string(act_prop)});
+    }
+  }
+  if (getType() != TimeDistLayer::type && getDistribute()) {
+    std::unique_ptr<TimeDistLayer> dlayer(new TimeDistLayer());
```
jijoongmoon, line 422: If we do this here, is it ok? I mean the order of execution is already set before this layer is finalized.

jijoongmoon (approved):
Overall LGTM except for minor comment.

nntrainer/layers/common_properties.h
```diff
@@ -10,18 +10,29 @@
...
+#include <string>
 
-#ifndef __COMMON_PROPERTIES_H__
-#define __COMMON_PROPERTIES_H__
+#include <base_properties.h>
 
 namespace nntrainer {
 
+/**
+ * @brief     Enumeration of activation function type
```
kparichay, line 25: add a note that if these enum are assigned values manually, update ActivationTypeInfo correspondingly,

nntrainer/layers/layer_node.h
```diff
@@ -607,29 +604,25 @@ upon final creation. Editing properties of the layer after init will not the
 properties in the context/graph unless intended. */
 
   using PropsType =
-    std::tuple<props::Name, props::Flatten, props::Distribute, props::Trainable,
+    std::tuple<props::Name, props::Distribute, props::Trainable,
                std::vector<props::InputLayer>, std::vector<props::InputShape>>;
+
+  using RealizationPropsType = std::tuple<props::Flatten, props::Activation>;
+  /** Props that will be used to realize a
```
kparichay, line 611: update description - these realization properties results in addition of new layers, hence skipped in generation of model architecture as the correspondingly layer itself is added. Distribute is also a property which is realized, but as it doesn't add new layer, it is saved.

nntrainer/layers/activation_layer.h
```diff
@@ -83,61 +87,11 @@ class ActivationLayer : public Layer {
...
-   * @param[in] std::function<Tensor(Tensor const &, Tensor &)> activation_fn
-   * activation function to be used
-   * @param[in] std::function<Tensor(Tensor const &, Tensor &)>
-   * activation_prime_fn activation_prime_function to be used
-   * @retval #ML_ERROR_NONE when successful
-   */
-  int setActivation(
-    std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
-    std::function<Tensor &(Tensor &, Tensor &)> const &activation_prime_fn);
+  std::unique_ptr<PropTypes> activation_props; /**< activation props */
```
kparichay, line 92: is this needed now? as layer_node caches the activation property and getActivationType is returned by layer_node itself.

nntrainer/layers/layer_node.cpp
kparichay: dist layer changes at least one property of the layer - supportInPlace(). If this is done here, supportInPlace must be returned simply not by the layer, but by the dist layer even before finalize (as inPlaceOptimization has been moved to compile time).

nntrainer/layers/layer_node.cpp
zhoonit: Yes, if we set TimeDistLayer when setProperty,
setProperty({"timedist=true", "timedist=true"}) is a bug.

So it's moved to `finalize`

nntrainer/layers/layer_node.cpp
zhoonit: Time dist layer does not change the order of execution so it should be okay.

> dist layer changes at least one property of the layer - supportInPlace(). If this is done here, supportInPlace must be returned simply not by the layer, but by the dist layer even before finalize (as inPlaceOptimization has been moved to compile time).

There is two (potential) bugs related with this.

1. `getLayer()` is returning inside layer, supportInPlace is not queried.
```
nntrainer::Layer *LayerNode::getLayer() {
  if (getDistribute())
    return ((TimeDistLayer *)(layer.get()))->getDistLayer();
  else
    return layer.get();
}
```

2. We will need another end point just like finalize() to invalidate changing properties after connection has been set. 

This will need more fundamental changes but I'll make a quick fix out of this PR.

nntrainer/layers/common_properties.h
zhoonit: Updated

nntrainer/layers/layer_node.h
zhoonit: Good idea :)

nntrainer/layers/activation_layer.h
zhoonit: This is needed to support below format.

[relu1]
type=activation
activation=relu

lhs8928 (approved):
LGTM

nntrainer/layers/activation_layer.h
kparichay: the `activation` can be added by layer node just like for every other layer.

zhoonit: In that case, we will need dynamic_cast to activation layer because we have to propagate activation somehow to the actual layer.

zhoonit:
Closing this to be merged with #1549

PR #1555, opened 2021-09-08
[wait for #1545]

kparichay:
- This patch adds support for in-place activation. Further, the in-place
optimization has been moved to the model compile phase.
This patch also adds support for running the model in in-place as well
as out-of-place mode.
- This patch optimized batch norm memory usage by updating the exec order
for its input as its not used in calcDerivative.
- This patch provides support for in-place batch normalization.

See Also #1127


jijoongmoon:
Could you update the title?

kparichay:
> LGTM. I wonder about the difference in peak memory consumption before and after. ( Resnet affects a lot, I guess )

I will measure and add here.

PR #1564, opened 2021-09-10
[Fix] Fix semantic tests were not running

zhoonit:
- [Fix] Fix semantic tests were not running

```
This patch fixes semantic tests were not running

**Self evaluation:**

```

### DIFF
--- a/Applications/Custom/LayerPlugin/meson.build
+++ b/Applications/Custom/LayerPlugin/meson.build
@@ -45,6 +45,7 @@ rnnt_loss_layer_dep = declare_dependency(
 
 if get_option('enable-test')
   test_target = [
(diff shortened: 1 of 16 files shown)


zhoonit:
semantic test wasn't running mainly because
https://stackoverflow.com/questions/54994175/test-p-tests-in-library-do-not-run-when-instantiate-test-case-p-is-called-in-a-s

We can make it link everything with static lib, but that is not desirable solution so tests are unpacked

kparichay (approved):
LGTM

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1567, opened 2021-09-13
[application] Add resnet18 tensorflow + update application @open sesame 09/23 13:59

### LINKED ISSUE #904: Evaluate ResNet

### LINKED ISSUE #1533: Enable resnet application

kparichay:
- Add resnet18 tensorflow version to match against nntrainer.
- Update resnet application to match tensorflow implementation.
- This patch reduces the number of operations required for tensor sum by
accumulating the consecutive axes in order to increase the precision of
the operation.
- This patch adds test check on the training of the resnet application.
This test can be enabled with enable-long-test and enable-test.

Note: reduction of operations for sum was intended to reduce the number of operations to result in higher precision but can very well result in speedup as well.

resolves #904
Resolves #1533

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -740,9 +740,9 @@ Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
   return sum(axes, ret, alpha);
 }
 
-void Tensor::merge_axis(unsigned int axis1, unsigned int axis2) {
+void Tensor::mergeAxis(unsigned int axis1, unsigned int axis2) {
   if (axis2 != axis1 + 1)
-    throw std::invalid_argument("Axis to be merged must be continuous.");
+    throw std::invalid_argument("axis2 must be axis1 + 1 for merging.");
 
   dim.setTensorDim(axis2, dim.getTensorDim(axis1) * dim.getTensorDim(axis2));
   dim.setTensorDim(axis1, 1);
@@ -761,7 +761,7 @@ Tensor &Tensor::sum(const std::vector<unsigned int> &axes, Tensor &output,
     std::vector<unsigned int> new_axes = {axes[0]};
     for (unsigned int i = 1; i < axes.size(); ++i) {
       if (axes[i] == axes[i - 1] + 1) {
-        new_reshaped.merge_axis(axes[i - 1], axes[i]);
+        new_reshaped.mergeAxis(axes[i - 1], axes[i]);
         new_axes.back() = axes[i];
       } else {
         new_axes.push_back(axes[i]);
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -214,6 +214,8 @@ bool Tensor::operator==(const Tensor &rhs) const {
   const float *rdata = rhs.getData();
 
   for (size_t i = 0; i < len; ++i) {
+    /** not checking sign change is intentional to avoid float calculation
+     * errors around 0 */
     if (std::isnan(data[i]) || std::isnan(rdata[i]) ||
         std::fabs(data[i] - rdata[i]) > epsilon)
       return false;
@@ -738,6 +740,14 @@ Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
   return sum(axes, ret, alpha);
 }
 
+void Tensor::merge_axis(unsigned int axis1, unsigned int axis2) {
+  if (axis2 != axis1 + 1)
+    throw std::invalid_argument("Axis to be merged must be continuous.");
+
+  dim.setTensorDim(axis2, dim.getTensorDim(axis1) * dim.getTensorDim(axis2));
+  dim.setTensorDim(axis1, 1);
+}
+
 Tensor &Tensor::sum(const std::vector<unsigned int> &axes, Tensor &output,
                     float alpha) const {
   if (axes.empty())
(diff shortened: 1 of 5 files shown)


Applications/Resnet/tensorflow/transLayer.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env python3
```
zhoonit, line 1: Could you use `sys.path.insert(1, {gen_input directory} )`; import transLayer as TL rather than including this file?

nntrainer/tensor/tensor.cpp
```diff
@@ -738,6 +740,14 @@ Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
   return sum(axes, ret, alpha);
 }
 
+void Tensor::merge_axis(unsigned int axis1, unsigned int axis2) {
```
zhoonit, line 743: maybe `mergeAxis`?

nntrainer/tensor/tensor.cpp
```diff
@@ -738,6 +740,14 @@ Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
...
+  if (axis2 != axis1 + 1)
+    throw std::invalid_argument("Axis to be merged must be continuous.");
```
zhoonit, line 745: is it also considered continuous if axis1 > axis2 ?

nntrainer/tensor/tensor.cpp
```diff
@@ -738,6 +740,14 @@ Tensor Tensor::sum(const std::vector<unsigned int> &axes, float alpha) const {
...
+
+  dim.setTensorDim(axis2, dim.getTensorDim(axis1) * dim.getTensorDim(axis2));
+  dim.setTensorDim(axis1, 1);
```
zhoonit, line 748: Can we call Tensor::reshape instead?

nntrainer/tensor/tensor.cpp
```diff
@@ -746,12 +756,22 @@ Tensor &Tensor::sum(const std::vector<unsigned int> &axes, Tensor &output,
   if (axes.size() == 1) {
     this->sum(axes[0], output, alpha);
   } else {
-    Tensor ret = this->sum(axes[0], alpha);
+    /** club axes together */
+    Tensor new_reshaped = *this;
+    std::vector<unsigned int> new_axes = {axes[0]};
+    for (unsigned int i = 1; i < axes.size(); ++i) {
+      if (axes[i] == axes[i - 1] + 1) {
```
zhoonit, line 763: note) it is not guaranteed that axes are always in sorted order. This is just an optimization so I think it's fine though

nntrainer/tensor/tensor.cpp
```diff
@@ -746,12 +756,22 @@ Tensor &Tensor::sum(const std::vector<unsigned int> &axes, Tensor &output,
...
+        new_reshaped.merge_axis(axes[i - 1], axes[i]);
+        new_axes.back() = axes[i];
+      } else {
+        new_axes.push_back(axes[i]);
+      }
+    }
 
-    for (unsigned int i = 1; i < axes.size() - 1; ++i)
+    Tensor ret = new_reshaped.sum(new_axes[0]);
```
zhoonit, line 771: wonder if this does not conflict with line 757?

Applications/Resnet/tensorflow/transLayer.py
kparichay: done.

nntrainer/tensor/tensor.cpp
zhoonit, line 771: we might need more unittest for Tensor::sum

nntrainer/tensor/tensor.cpp
kparichay: done.

nntrainer/tensor/tensor.cpp
kparichay: Update the error message to be more precise. We only merge is axis2 == axis1 + 1.

Applications/Resnet/tensorflow/transLayer.py
zhoonit: I guess this file can be deleted!

nntrainer/tensor/tensor.cpp
kparichay: To use reshape, there has to be a switch making this slower in critical section depending on the location of axis.

nntrainer/tensor/tensor.cpp
zhoonit: Cool

zhoonit (approved):
LGTM :)

nntrainer/tensor/tensor.cpp
kparichay: Yes, its not gauranteed. Developer must try to give it in order for optimization. We can add sort and unique but IMO its overkill.

nntrainer/tensor/tensor.cpp
kparichay: I will add an issue for this.

nntrainer/tensor/tensor.cpp
zhoonit: If we have change the reshape implementation, this will be likely to be untouched. If we have more detailed tests, I guess it will be fine :)

Applications/Resnet/tensorflow/transLayer.py
kparichay: it is :).

jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor.cpp
kparichay: #1581

nntrainer/tensor/tensor.cpp
kparichay: Yes, will include tests for this with #1581.

Issue #1562, opened 2021-09-10
[Coverage] Update missing tests


zhoonit:
- Optimizer tests
- layer semantic test has problem
- Pluggable object test

zhoonit:
There is no coverage issue (in terms of quality bar) as of now, closing this

PR #1574, opened 2021-09-16
[unittest/models] Fix matching for models unittest

kparichay:
This patch fixes the matching for the last layer of each model for the
unittest of the models.

See also #1562

### DIFF
--- a/test/unittest/unittest_nntrainer_models.cpp
+++ b/test/unittest/unittest_nntrainer_models.cpp
@@ -152,7 +152,7 @@ public:
    * @param iteration iteration
    * @return nntrainer::sharedConstTensor
    */
-  void forward(int iteration, NodeWatcher &next_node);
+  void forward(int iteration, bool verify);
 
   /**
    * @brief backward pass of the node with verifying inputs/gradients/outputs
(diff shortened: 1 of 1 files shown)


test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -423,18 +429,23 @@ void GraphWatcher::compareFor(const std::string &reference,
...
       nn.backwarding(label, iteration);
 
-      for (auto it = nodes.rbegin(); it != nodes.rend() - 1; it++) {
-        if (it == nodes.rend())
-          it->backward(iteration, true, !optimize);
-        else
+      for (auto it = nodes.rbegin(); it != nodes.rend(); it++) {
+        if (it == nodes.rend() - 1) {
+          /** check last layer backwarding only when not input layers */
+          if (it->getNodeType() != nntrainer::InputLayer::type)
```
zhoonit, line 444: Could you change this to query `getNumInputConnections()`?

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -423,18 +429,23 @@ void GraphWatcher::compareFor(const std::string &reference,
                   nntrainer::Tensor::epsilon);
     }
 
-    for (auto it = nodes.begin(); it != nodes.end() - 1; ++it) {
-      it->forward(iteration, *(it + 1));
+    auto it = nodes.begin();
+    for (; it != nodes.end() - 1; ++it) {
+      it->forward(iteration, !(it + 1)->supportInPlace());
     }
+    it->forward(iteration, true);
```
zhoonit, line 436: Nice catch :)

lhs8928 (approved):
LGTM

test/unittest/unittest_nntrainer_models.cpp
kparichay: This is intentional. First layer can be convolution but we want to verify backwarding for it.
Updated rather to supportBackwarding.

zhoonit: I see :)

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

Issue #1575, opened 2021-09-17
Migrate golden test to pytorch


zhoonit:
As pytorch package is available in gbs, we don't have to save the golden data in our repo rather we can generate on the runtime so that we can test and match bigger model.

PR #1578, opened 2021-09-17
[resnet/unittest] Add unittest for resnet @open sesame 09/23 13:59

kparichay:
- This patch adds resnet18 unittest model generation with genModelTests.
Further, the input data range is changed from 0 to x from -x to x as
relu based models work better with 0 to x data range to prevent loss of
information.
- This patch adds resnet models unittest for resnet18.
The verification has been done offline for 2 iterations for the output
of all layers with precision of 1.1e-4.
Derivaitves and gradients have higher error because of relu: when some
value is close to 0, it can be positive or negative with some error (of
the order of e-7). Although this error is way within the error limit.
however, this exacerbates the error in backwarding where derivatives (which
are significant in values) can flow if the relu value was over 0, and
not flow if under zero. This is manageable in smaller models but
difficult to avoid in unittests for larger models.

Other bug fixes in this patch:
- max error reported by unittest_nntrainer_models has been fixed
- error reporting now includes layer type as well
- ModelTestOption MINIMUM has been renamed to NO_THROW_RUN

Resnet unittest is disabled as the golden data exceeds 70MB. This will be enabled when the golden data can be generated at runtime.

### DIFF
--- a/test/input_gen/recorder.py
+++ b/test/input_gen/recorder.py
@@ -72,7 +72,8 @@ def _rand_like(tensorOrShape, scale=1):
     except AttributeError:
         shape = tensorOrShape
 
-    t = np.random.randint(-10, 10, shape).astype(dtype=np.float32)
+    # for relu based models, range of 0 to x is better than -x to x
+    t = np.random.randint(0, 10, shape).astype(dtype=np.float32)
     return tf.convert_to_tensor(t) * scale
 
 
--- a/test/unittest/unittest_nntrainer_models.cpp
+++ b/test/unittest/unittest_nntrainer_models.cpp
@@ -66,7 +66,10 @@ void verify(const nntrainer::Tensor &actual, const nntrainer::Tensor &expected,
       std::cout << "\033[1;33mdifference\033[0m " << diff;
       std::cout << "number of data: " << diff.size() << std::endl;
       std::cout << "\033[4;33mMAX DIFF: "
-                << *std::max_element(diff_data, diff_data + diff.size())
+                << *std::max_element(diff_data, diff_data + diff.size(),
+                                     [](auto v1, auto v2) {
+                                       return std::fabs(v1) < std::fabs(v2);
+                                     })
                 << "\033[0m\n";
     }
     std::stringstream ss;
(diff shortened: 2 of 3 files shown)


jijoongmoon (approved):
LGTM

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -1464,6 +1569,9 @@ INSTANTIATE_TEST_CASE_P(
 
       /**< multi output test */
       mkModelTc(multiple_output_model, "3:1:1:10", 10, ModelTestOption::COMPARE) // Todo: Enable option to ALL
+      /** resnet model */
+      // this must match training (verify only forwarding output values) for 2 iterations with tolerance 1.2e-4
+      // mkResNet18Tc(2, ModelTestOption::COMPARE)
```
zhoonit, line 1574: I guess this is intended for now?

kparichay: Yes, uncomment this, and generate its corresponding golden file from input_gen. This will be resolved with #1575.

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -580,7 +589,9 @@ class nntrainerModelTest
   int getIteration() { return iteration; };
   nntrainer::TensorDim getLabelDim() { return label_dim; }
 
-  bool shouldCompare() { return options & ModelTestOption::COMPARE; }
+  bool shouldCompare() {
+    return (options & ModelTestOption::COMPARE) == ModelTestOption::COMPARE;
```
zhoonit, line 593: I wonder about this change,` ModelTestOption::COMPARE` is 1 << 0 so it should work though.

zhoonit (approved):
LGTM

test/unittest/unittest_nntrainer_models.cpp
kparichay: Later, we might fine-grain COMPARE to COMPARE_FOR and COMPARE_BACK. This will be needed for resnet kind of models where the error in derivative is significantly outside the error threshold. Hence added this.

zhoonit: Alright :)

test/input_gen/recorder.py
```diff
@@ -72,7 +72,8 @@ def _rand_like(tensorOrShape, scale=1):
     except AttributeError:
         shape = tensorOrShape
 
-    t = np.random.randint(-10, 10, shape).astype(dtype=np.float32)
+    # for relu based models, range of 0 to x is better than -x to x
+    t = np.random.randint(0, 10, shape).astype(dtype=np.float32)
```
zhoonit, line 76: Should this be the case for others? we will not get less change to test if certain value is negative, eg) leaky relu.

kparichay: true. I think input range must be chosen based on model to model. I will update it as a separate PR to [0,10] for resnet and [-10,10] as default.

PR #1587, opened 2021-09-24
[Test] Add bn layer inference test

zhoonit:

<details><summary>[Test] Add bn layer inference test</summary><br />

- bn layer inference mode test
- add option to skip backwarding golden test

**Self evaluation:**

</details>

### DIFF
--- a/test/unittest/layers/layers_common_tests.h
+++ b/test/unittest/layers/layers_common_tests.h
@@ -79,11 +79,23 @@ protected:
   bool must_fail;
 };
 
+typedef enum {
+  SKIP_CALC_GRAD = 1 << 0,  /**< skip calculating gradient and compare */
+  SKIP_CALC_DERIV = 1 << 1, /**< skip calculating derivative and compare */
+
+  FORWARD_MODE_INFERENCE =
+    1 << 2, /**< set if layer should be forwarded with inference mode */
+
+  DEFAULT =
+    0, /**< default set up, compare forward, backward in training mode */
+} LayerGoldenTestParamOptions;
+
 using LayerGoldenTestParamType =
   std::tuple<LayerFactoryType /**< Layer creator */,
              std::vector<std::string> /**< Properties */,
              const char *, /**< Input Tensor dimensions representation */
-             const char * /**< Golden file name */>;
+             const char * /**< Golden file name */,
+             int /**< LayerGoldenTestParamOptions */>;
 
 /**
  * @brief Golden Layer Test with designated format
@@ -108,6 +120,28 @@ public:
    *
    */
   virtual void TearDown();
+
+  /**
+   * @brief check if given test suite should be forwarded with inference mode
+   * enabled
+   *
+   * @return bool layer should be forwarded with inference
+   */
+  bool shouldForwardWithInferenceMode();
+
+  /**
+   * @brief check if given test suite should skip calculating derivative
+   *
+   * @return bool true if should skip calculating derivative
+   */
+  bool shouldSkipCalcDeriv();
+
+  /**
+   * @brief check if given test suite should skip calculating Gradient
+   *
+   * @return bool true if should skip calculating Gradient
+   */
+  bool shouldSkipCalcGrad();
 };
 
 #endif // __LAYERS_COMMON_TESTS_H__
--- a/test/unittest/layers/layers_golden_tests.cpp
+++ b/test/unittest/layers/layers_golden_tests.cpp
@@ -145,10 +145,11 @@ static RunLayerContext prepareRunContext(const TensorPacks &packs) {
   return rc;
 }
 
-static void compareRunContext(RunLayerContext &rc, std::ifstream &file) {
+static void compareRunContext(RunLayerContext &rc, std::ifstream &file,
+                              bool skip_grad, bool skip_deriv) {
   file.seekg(0, std::ios::beg);
   auto compare_tensors = [&file](unsigned length, auto tensor_getter, auto pred,
-                                 const std::string &name) {
+                                 bool skip_compare, const std::string &name) {
     for (unsigned i = 0; i < length; ++i) {
       if (!pred(i)) {
         continue;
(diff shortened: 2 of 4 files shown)


test/unittest/layers/layers_common_tests.h
```diff
@@ -79,11 +79,24 @@ class LayerSemantics
...
 
+typedef enum {
+  SKIP_CALC_GRAD = 1 << 0,  /**< skip calculating gradient and compare */
+  SKIP_CALC_DERIV = 1 << 1, /**< skip calculating derivative and compare */
+
+  FORWARD_MODE_INFERENCE =
+    1 << 2, /**< set if layer should be forwarded with inference mode */
+
+  DEFAULT = 0, /**< default set up */
+  ONLY_FORWARD =
```
kparichay, line 90: ONLY_FORWARD -> FORWARD_TRAIN_MODE?

test/unittest/layers/layers_common_tests.h
```diff
@@ -79,11 +79,24 @@ class LayerSemantics
...
 };
 
+typedef enum {
+  SKIP_CALC_GRAD = 1 << 0,  /**< skip calculating gradient and compare */
+  SKIP_CALC_DERIV = 1 << 1, /**< skip calculating derivative and compare */
+
+  FORWARD_MODE_INFERENCE =
+    1 << 2, /**< set if layer should be forwarded with inference mode */
+
+  DEFAULT = 0, /**< default set up */
```
kparichay, line 89: is default only set up or setup and run both forwarding and backwarding?

test/unittest/layers/layers_golden_tests.cpp
```diff
@@ -157,36 +158,38 @@ static void compareRunContext(RunLayerContext &rc, std::ifstream &file) {
...
   compare_tensors(rc.getNumWeights(),
-                  [&rc](unsigned idx) { return rc.getWeight(idx); }, always,
-                  "initial_weights");
+                  [&rc](unsigned idx) { return rc.getWeight(idx); },
+                  always_read, skip_compare, "initial_weights");
   compare_tensors(rc.getNumInputs(),
-                  [&rc](unsigned idx) { return rc.getInput(idx); }, always,
-                  "inputs");
+                  [&rc](unsigned idx) { return rc.getInput(idx); }, always_read,
+                  !skip_compare, "inputs");
```
kparichay, line 180: check the layer for `supportInPlace`. If it's true, skip checking input as it would be overwritten, or create two modes to configure it and check for both (maybe in later PR).

kparichay (approved):

test/unittest/layers/layers_common_tests.h
zhoonit: It's bit flag for presetting SKIP_CALC_GRAD, SKIP_CALC_DERIV, 

I think, FORWARD_TRAIN_MODE | FORWARD_MODE_INFERENCE would be hard to infer.

test/unittest/layers/layers_common_tests.h
zhoonit: Done :)

test/unittest/layers/layers_common_tests.h
kparichay: ONLY_FORWARD should clarify if it is train or inference, which it is currently not clarifying.

test/unittest/layers/layers_golden_tests.cpp
kparichay: if the layer is set to operate in `in-place`, then the layer/model need not perform such checks.
I guess this will be clarified with https://github.com/nnstreamer/nntrainer/blob/de0ae0d01127ef62afda12caa2185d396b27e138/nntrainer/layers/layer_context.h#L279-L284. 

> Shouldn't we have to make sure supportInPlace must not override the input if underlying memory does not clashing each other?

`context.executeInPlace()` tells the layer if the layer is expected to run in-place. So, this test can start with execute-in-place to be always false, but must test the layers which `support in-place` operations with execute-in-place set to true as well.

zhoonit: Awesome, I'll make sure to make one in later PR

test/unittest/layers/layers_common_tests.h
zhoonit: I mean, It is intentionally not clarifying the mode because it doesn't care. It's bit flag preset. It's not setting any bit to the mode.

kparichay: I meant not from the perspective of bit flag or bit mode but for readability.

zhoonit: Well, this is used like https://github.com/nnstreamer/nntrainer/pull/1587/files#diff-c7e99e86961362f866b90c337b8af3b91a70b8f84a84bb5e0f386d81f0a619a2R26-R27 this. 

Now it will become `bn_inference_option = FORWARD_TRAIN_MODE | FORWARD_MODE_INFERENCE` which looks not intuitive IMO.

zhoonit: How about deleting this if this preset seems to cause some misunderstanding?

zhoonit: Done

test/unittest/layers/layers_golden_tests.cpp
```diff
@@ -157,36 +158,38 @@ static void compareRunContext(RunLayerContext &rc, std::ifstream &file) {
...
+                  always_read, !skip_compare, "outputs");
   compare_tensors(rc.getNumWeights(),
                   [&rc](unsigned idx) { return rc.getWeightGrad(idx); },
-                  only_trainable, "gradients");
+                  only_read_trainable, skip_grad, "gradients");
   compare_tensors(rc.getNumWeights(),
-                  [&rc](unsigned idx) { return rc.getWeight(idx); }, always,
-                  "weights");
+                  [&rc](unsigned idx) { return rc.getWeight(idx); },
+                  always_read, skip_compare, "weights");
```
lhs8928, line 189: !skip_compare?

zhoonit: Thanks updated

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1584, opened 2021-09-23
[Test] Add bn layer training

zhoonit:
## Commits to be reviewed in this PR

<details><summary>[Test] Add bn layer training</summary><br />

**Changes proposed in this PR:**
- Add batchnormalization layer test (for channel, width axis)

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

</details>

PR #1590, opened 2021-09-24
[batchnorm] Optimize batch norm for memory and speed @open sesame 09/28 18:42

### LINKED ISSUE #1598: Q. Shouldn't CalcGradient be skipped when non-trainable?
https://github.com/nnstreamer/nntrainer/blob/51a871bd7b6e63ddcac6185e55ae802c804d1d70/nntrainer/models/neuralnet.cpp#L244-L245

Current upstream code does not skip calcGradient when `node->trainable() == false`, I wonder if this is because some calculation is depending on calcGradient() (which they shouldn't)?

kparichay:
- This patch optimizes batch normalization implementation.
Reduces the temporary memory allocation and provide speedup for the
layer execution. With this patch, resnet18 runtime improves by approx
5%.
  - update the tensor usage method to give the output tensor. Although the methods where already allocated output can be passed were available, they were not being used. This patch adds the corresponding support.
- This patch optimizes batch norm layer and tries to share the
calculations performed in calcGradient and calcDerivative.
    - reuse dbeta and dgamma calculations
    - reduce number of required temporary variables
    - create all the required tensor variables with context
    - add support for checking if the layer is trainable or not via run
    context
    - support average operation with the output tensor already allocated
    - this patch reduces as much as memory as possible without sacrificing
    speed. more memory optimization is possible at the expense of speed but
    has been ommitted for now.

See also  #1591

resolves #1598

### DIFF
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -121,7 +121,8 @@ void BatchNormalizationLayer::forwarding(RunLayerContext &context,
     Tensor cmu = input_.average(axes_to_reduce);
     input_.subtract(cmu, deviation);
 
-    cvar = deviation.pow(2.0f).average(axes_to_reduce);
+    Tensor t1 = deviation.pow(2.0f);
+    cvar = t1.average(axes_to_reduce);
 
     mu.multiply_i(momentum);
     mu.add_i(cmu, 1 - momentum);
@@ -129,14 +130,14 @@ void BatchNormalizationLayer::forwarding(RunLayerContext &context,
     var.add_i(cvar, 1 - momentum);
 
     cvar.add_i(epsilon);
-    invstd = cvar.pow(-0.5f);
+    cvar.pow(-0.5f, invstd);
   } else {
-    deviation = input_.subtract(mu);
-    invstd = var.add(epsilon);
+    input_.subtract(mu, deviation);
+    var.add(epsilon, invstd);
     invstd.pow_i(-0.5f);
   }
 
-  hidden_ = deviation.multiply(invstd, hidden_);
+  deviation.multiply(invstd, hidden_);
   hidden_.multiply_i(gamma);
   hidden_.add_i(beta);
 }
@@ -148,21 +149,16 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
   Tensor &dx = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
   Tensor &deviation = context.getTensor(wt_idx[BNParams::deviation]);
 
-  int N = 1;
-  const Tensor &input = context.getInput(SINGLE_INOUT_IDX);
-  const TensorDim &in_dim = input.getDim();
-  for (auto &axis : axes_to_reduce) {
-    N *= in_dim.getTensorDim(axis);
-  }
-
   Tensor dx_1 = gamma.multiply(invstd);
-  Tensor dx_2 = deriv.multiply(N);
-  dx_2.subtract_i(deriv.sum(axes_to_reduce));
-  dx_2.subtract_i(deviation.divide(cvar).multiply(
-    deviation.multiply(deriv).sum(axes_to_reduce)));
+  Tensor dx_2 = deriv.subtract(deriv.average(axes_to_reduce));
 
-  dx = dx_2.multiply(dx_1, dx);
-  dx.divide_i(N);
+  Tensor t1 = deviation.multiply(deriv);
+  Tensor t2 = t1.average(axes_to_reduce);
+  deviation.divide_i(cvar);
+  deviation.multiply_i(t2);
+  dx_2.subtract_i(deviation);
+
+  dx_2.multiply(dx_1, dx);
 }
 
 void BatchNormalizationLayer::calcGradient(RunLayerContext &context) {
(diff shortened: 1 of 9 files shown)


kparichay:
Rebase with #1584 and ensure that batch size 2 unittest is passing, and also check for batch size 4.

nntrainer/layers/bn_layer.cpp
```diff
@@ -89,6 +110,18 @@ void BatchNormalizationLayer::finalize(InitLayerContext &context) {
...
+  wt_idx[BNParams::invstd] =
+    context.requestTensor(dim, context.getName() + ":invstd",
+                          Tensor::Initializer::NONE, false, ITERATION_LIFESPAN);
+  wt_idx[BNParams::t_reduced] =
+    context.requestTensor(dim, context.getName() + ":tensor_reduced",
+                          Tensor::Initializer::NONE, false, ITERATION_LIFESPAN);
+  wt_idx[BNParams::t_full_fw] = context.requestTensor(
+    in_dim, context.getName() + ":tensor_full_fw", Tensor::Initializer::NONE,
+    false, FORWARD_FUNC_LIFESPAN);
+  wt_idx[BNParams::t_full_bw] = context.requestTensor(
```
jijoongmoon, line 122: Just for checking, if we set FORWARD_FUNC_LIFESPAN here, is it null in backwarding?

jijoongmoon (approved):
LGTM

nntrainer/layers/bn_layer.cpp
```diff
@@ -147,35 +185,49 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
...
-  Tensor dx_1 = gamma.multiply(invstd);
-  Tensor dx_2 = deriv.multiply(N);
-  dx_2.subtract_i(deriv.sum(axes_to_reduce));
-  dx_2.subtract_i(deviation.divide(cvar).multiply(
-    deviation.multiply(deriv).sum(axes_to_reduce)));
+  dx = deriv.subtract(t_reduced);
 
-  dx = dx_2.multiply(dx_1, dx);
-  dx.divide_i(N);
+  t_full = deviation.multiply(deriv);
```
zhoonit, line 206: deviation.multiply(deriv, t_full); ?

nntrainer/layers/bn_layer.cpp
```diff
@@ -147,35 +185,49 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
...
-  Tensor &dgamma = context.getWeightGrad(wt_idx[BNParams::gamma]);
   Tensor &dbeta = context.getWeightGrad(wt_idx[BNParams::beta]);
   Tensor &deriv = context.getIncomingDerivative(SINGLE_INOUT_IDX);
-  Tensor &deviation = context.getTensor(wt_idx[BNParams::deviation]);
 
-  dbeta = deriv.sum(axes_to_reduce);
-  Tensor dev = deviation.multiply(invstd);
-  dev.multiply_i(deriv);
-  dgamma = dev.sum(axes_to_reduce);
+  deriv.sum(axes_to_reduce, dbeta);
```
zhoonit, line 230: only beta is calculated here and gamma is calculated in calcDeriv?

nntrainer/layers/bn_layer.cpp
```diff
@@ -147,35 +185,49 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
...
+  } else {
+    t_reduced = deriv.average(axes_to_reduce);
   }
 
-  Tensor dx_1 = gamma.multiply(invstd);
-  Tensor dx_2 = deriv.multiply(N);
-  dx_2.subtract_i(deriv.sum(axes_to_reduce));
-  dx_2.subtract_i(deviation.divide(cvar).multiply(
-    deviation.multiply(deriv).sum(axes_to_reduce)));
+  dx = deriv.subtract(t_reduced);
```
zhoonit, line 204: deriv.subtract(t_reduced, dx)?

nntrainer/layers/bn_layer.cpp
kparichay: currently, it is in available backward (run context doesn't know when function is calling it), but later we will have a check so `context.requestTensor()` will throw in this case.

kparichay: I have added a corresponding todo in layer context.

nntrainer/layers/bn_layer.cpp
kparichay: good catch, done.

nntrainer/layers/bn_layer.cpp
kparichay: good catch, done.

nntrainer/layers/bn_layer.cpp
kparichay: yes.

nntrainer/layers/bn_layer.cpp
```diff
@@ -89,6 +110,18 @@ void BatchNormalizationLayer::finalize(InitLayerContext &context) {
   wt_idx[BNParams::deviation] =
     context.requestTensor(in_dim, context.getName() + ":deviation",
                           Tensor::Initializer::NONE, false, ITERATION_LIFESPAN);
+  wt_idx[BNParams::invstd] =
+    context.requestTensor(dim, context.getName() + ":invstd",
+                          Tensor::Initializer::NONE, false, ITERATION_LIFESPAN);
+  wt_idx[BNParams::t_reduced] =
```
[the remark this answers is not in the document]
kparichay: added in the comments.

zhoonit (approved):
LGTM!

PR #1593, opened 2021-09-28
[batch norm] reduce batch norm memory utilization further

### LINKED ISSUE #1591: batch norm optimization
batch norm layer needs some optimization to reduce the memory consumption of its temporary tensors, as well as reuse calculations to provide speedup.

kparichay:
- Reduce the memory comsumption of batch norm for forwarding by re-using
the output tensor as temporary memory than requesting new memory.
- Remove the extra full size extra memory requirement as the cost of the
reduced memory. The difference in memory requirmement can be
significant. Earlier memory requirement was b*c*h*w vs now it is just c
where the assumption is that batch norm if is normalizing along
axis=channel.

This is achieved by reordering of the operations.
Note: this change has no performance impact.

Resolves #1591

### DIFF
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -32,17 +32,7 @@ namespace nntrainer {
 
 static constexpr size_t SINGLE_INOUT_IDX = 0;
 
-enum BNParams {
-  mu,
-  var,
-  gamma,
-  beta,
-  deviation,
-  invstd,
-  t_full_fw,
-  t_full_bw,
-  t_reduced
-};
+enum BNParams { mu, var, gamma, beta, deviation, invstd, t_full_bw, t_reduced };
 
 BatchNormalizationLayer::BatchNormalizationLayer(int axis_) :
   Layer(),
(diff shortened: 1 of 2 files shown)


nntrainer/layers/bn_layer.cpp
```diff
@@ -198,28 +187,15 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
...
   Tensor &t_reduced = context.getTensor(wt_idx[BNParams::t_reduced]);
-  Tensor &t_full = context.getTensor(wt_idx[BNParams::t_full_bw]);
-  Tensor cvar = t_reduced;
-
-  if (context.getTrainable()) {
-    /**
-     * This implementation depends on the pre-calculated dbeta calculated.
-     */
-    Tensor &dbeta = context.getWeightGrad(wt_idx[BNParams::beta]);
-    t_reduced = dbeta.divide(divider);
```
kparichay, line 211: this is a bus in original implementation. This assumes that the t_reduced is not needed anymore, but it is as it shared memory with c_var. This is working for now because this statement creates new memory.

nntrainer/layers/bn_layer.cpp
```diff
@@ -228,8 +204,19 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
     Tensor &dgamma = context.getWeightGrad(wt_idx[BNParams::gamma]);
     t_full.multiply_i(invstd);
     t_full.sum(axes_to_reduce, dgamma);
+
+    /**
+     * This implementation depends on the pre-calculated dbeta calculated.
+     */
+    Tensor &dbeta = context.getWeightGrad(wt_idx[BNParams::beta]);
+    t_reduced = dbeta.divide(divider);
```
kparichay, line 212: this new memory creation is solved universally for all layers at once in the new PR I am working on.
I verified locally that reusing this memory has no issue.

zhoonit (approved):
LGTM!

Issue #1599, opened 2021-10-01
Support attention layer


kparichay:
- [x] Support basic attention layer with query and value
  - [x] Support forwarding and setup - #1600
  - [x] Support backwarding - #1602
- [x] Support attention layer with query, key and value #1613
- [ ] Add unit tests for the same - Done - added within each PR.
- [ ] Extend to support MoL attention layer - https://arxiv.org/pdf/2009.02725.pdf

PR #1600, opened 2021-10-01
[layer] Add initial support for attention layer

kparichay:
- This patch adds bug fix for finalize of the layer node. The checks of
the inputs dimensions and input shapes has been fixed when multiple
inputs are expected to be set.
- This patch adds the initial commit for attention layer.
  - add class description
  - add basic forwarding
  This implements the common form of attention layer where key and value
are the same tensor. The other format will be supported soon.
- This patch adds basic unittest for attention layer.
To achieve, the existing tests are modified to support multiple inputs
in the test format.
- This patch adds backwarding for attention layer. Corresponding unittests
will be added in the next patch.

See Also #1599

### DIFF
--- a/nntrainer/layers/attention_layer.cpp
+++ b/nntrainer/layers/attention_layer.cpp
@@ -19,7 +19,7 @@ namespace nntrainer {
 
 static constexpr size_t SINGLE_INOUT_IDX = 0;
 
-enum AttentionParams { query = 0, value = 1 };
+enum AttentionParams { query = 0, value = 1, score, weights };
 
 void AttentionLayer::finalize(InitLayerContext &context) {
   if (context.getNumInputs() != 2)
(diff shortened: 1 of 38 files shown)


nntrainer/layers/attention_layer.cpp
```diff
@@ -28,28 +28,55 @@ void AttentionLayer::finalize(InitLayerContext &context) {
...
+
+  Tensor &dquery = context.getOutgoingDerivative(AttentionParams::query);
+  Tensor &dvalue = context.getOutgoingDerivative(AttentionParams::value);
+  Tensor &weights = context.getTensor(wt_idx[AttentionParams::weights]);
+
+  Tensor t1;
+  sm.run_prime_fn(weights, t1, derivative_);
+
+  Tensor t2 = value.dot(t1);
+  dquery = t2.dot(value).dot(derivative_);
```
zhoonit, line 77: Could you double check if this is possibly multiply(derivative_)? (and also for line 79)

zhoonit (approved):
LGTM

nntrainer/layers/attention_layer.cpp
kparichay: No, this is dot. The next PR verifies this implementaion with TF. Refer https://www.tensorflow.org/api_docs/python/tf/keras/layers/Attention for more details.

zhoonit: Cool :)

PR #1601, opened 2021-10-05
[ API ] Add Inference in CCAPI to get loss value

jijoongmoon:
Add Inference API to get the loss value

**Self evaluation:**


nntrainer/models/neuralnet.cpp
```diff
@@ -568,7 +601,21 @@ std::vector<float *> NeuralNetwork::inference(std::vector<float *> &input,
...
+    label_tensors.reserve(label.size());
+    for (unsigned int idx = 0; idx < label_dim.size(); idx++) {
+      label_dim[idx].batch(batch_size);
+      label_tensors.emplace_back(MAKE_SHARED_TENSOR(
+        Tensor::Map(label[idx], label_dim[idx].getDataLen() * sizeof(float),
+                    label_dim[idx], 0)));
+    }
+    output_tensors = inference(input_tensors, label_tensors, false);
+  } else {
+    output_tensors = inference(input_tensors, false);
```
kparichay, line 616: can let this be called with label_tensors as empty `sharedConstTensors`, and have only 1 call to inference.

nntrainer/models/neuralnet.cpp
```diff
@@ -556,9 +556,42 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
...
+  /** Clear the set inputs and labels */
+  setLabels({}, model_graph.getLabelList());
+  setInputs({}, model_graph.getInputList());
+
+  return out;
+}
+
+std::vector<float *> NeuralNetwork::inference(unsigned int batch_size,
+                                              std::vector<float *> &input,
+                                              std::vector<float *> &label) {
```
zhoonit, line 593: fyi) if we change the signature here, nnstreamer nntrainer filter has to change as well.

zhoonit (approved):
LGTM

nntrainer/models/neuralnet.cpp
kparichay: Lets set `label = {}` for default. With https://github.com/nnstreamer/nntrainer/pull/1601#discussion_r721831285, this can be supported.

PR #1607, opened 2021-10-06
[layer] Add checks for layer tensor overwrite bug

kparichay:
This patch adds a check to ensure that when layer tensors are created
and overwrites existing tensors. These checks are enabled only in DEBUG
mode to ensure that they only run in CI mode, and are called after each
operation - forwarding, calcGradient, and calcDerivative.


jijoongmoon (approved):
LGTM

Applications/Custom/pow.cpp
```diff
@@ -112,7 +112,7 @@ void PowLayer::calcDerivative(nntrainer::RunLayerContext &context) {
     context.getIncomingDerivative(SINGLE_INOUT_IDX);
   nntrainer::Tensor &dx = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
 
-  dx = derivative_.multiply(exponent);
+  derivative_.multiply(exponent, dx);
```
zhoonit, line 115: in those types of signature, let's remove returning reference at some time.

nntrainer/layers/layer_context.cpp
```diff
@@ -308,4 +333,86 @@ bool RunLayerContext::readyToUse() const {
...
+   * tensors are not set mistakenly by verifying their unique names
+   */
+#ifdef ENABLE_TEST
+  if (tensor_map.empty() || !tensor_map[inputs[0]->getName()]) {
+    auto filler = [this](const Var_Grad *val) {
+      tensor_map[val->getName()] = val->getVariableRef().getData();
+      tensor_map[val->getGradientName()] = val->getGradientRef().getData();
+    };
+
+    auto filler_w = [this, filler](const std::vector<Weight *> &vec) {
```
zhoonit, line 354: Code clean idea:

```suggestion
    auto filler_w = [this, filler](const auto &vec) { 
```

and removing filler_vw

nntrainer/layers/layer_context.cpp
```diff
@@ -308,4 +333,86 @@ bool RunLayerContext::readyToUse() const {
...
+ *
+ * @return true if ready, else false
+ */
+bool RunLayerContext::validate(bool skip_input, bool skip_label) {
+  /**
+   * @note a common mistake when using run_context is re-assigning the tensor
+   * references which leads to nasty bugs. This validation ensures that the
+   * tensors are not set mistakenly by verifying their unique names
+   */
+#ifdef ENABLE_TEST
```
zhoonit, line 347: How about using special option to check this? we commonly use ENABLE_TEST in packagin, I think it is easy to mistakenly use this in optimized build.

nntrainer/layers/layer_context.cpp
```diff
@@ -308,4 +333,86 @@ bool RunLayerContext::readyToUse() const {
...
+  if (tensor_map.empty() || !tensor_map[inputs[0]->getName()]) {
```
zhoonit, line 348: How about extracting this logic to `prepareValidate()`?

nntrainer/layers/layer_context.cpp
```diff
@@ -308,4 +333,86 @@ bool RunLayerContext::readyToUse() const {
...
+    };
+
+    auto matcher_w = [this, matcher](const std::vector<Weight *> &vec) {
+      auto ret = true;
+      for (auto const &val : vec)
+        ret &= matcher(val);
+      return ret;
+    };
+
+    auto matcher_vw = [this, matcher](const std::vector<Var_Grad *> &vec,
```
zhoonit, line 396: how about `std::all_of` instead of this lambda?

zhoonit: This is still fine though :)

kparichay:
V2: This patch enables debug mode for the CI build for both ubuntu and
tizen. This enables all the debug tests to be done in the CI which were
disabled till now.
Fixes required to enable the DEBUG mode is also added.

PR #1680, opened 2021-10-28
[do not merge] link ruy and implement

zhoonit:
this patch link ruy and implement dot with it.


kparichay:
@zhoonit Please add the final results we have here for logging purposes.

PR #1684, opened 2021-11-01
[layer] Simplify naming scheme for requested tensors 

kparichay:
- All layers were required to make unique name for their requests.
This is now rather done by layer context by appending layer name as a
prefix to the requested name.
- This patch allows support for prefix for sharing weight names across
layers, so that layers which should share their weights can do so by
using the same prefix.

### DIFF
--- a/nntrainer/layers/dropout.cpp
+++ b/nntrainer/layers/dropout.cpp
@@ -26,9 +26,9 @@ void DropOutLayer::finalize(InitLayerContext &context) {
 
   mask_idx.reserve(input_dims.size());
   for (auto &t : input_dims) {
-    mask_idx.push_back(context.requestTensor(
-      t, context.getName() + ":Mask", Tensor::Initializer::NONE, false,
-      TensorLifespan::ITERATION_LIFESPAN));
+    mask_idx.push_back(
+      context.requestTensor(t, "Mask", Tensor::Initializer::NONE, false,
+                            TensorLifespan::ITERATION_LIFESPAN));
   }
 }
 
--- a/nntrainer/layers/embedding.cpp
+++ b/nntrainer/layers/embedding.cpp
@@ -61,9 +61,9 @@ void EmbeddingLayer::finalize(InitLayerContext &context) {
   dim.width(out_dim);
   dim.batch(1);
 
-  weight_idx = context.requestWeight(
-    dim, weight_initializer, weight_regularizer, weight_regularizer_constant,
-    context.getName() + ":Embedding", true);
+  weight_idx =
+    context.requestWeight(dim, weight_initializer, weight_regularizer,
+                          weight_regularizer_constant, "Embedding", true);
 }
 
 void EmbeddingLayer::setProperty(const std::vector<std::string> &values) {
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -47,10 +47,13 @@ public:
     input_dim(dim),
     in_place(in_place_),
     num_outputs(num_out),
-    name(n) {
+    name(n),
+    prefix("") {
     NNTR_THROW_IF(!validate(), std::invalid_argument)
       << "Invalid init context name: " << name
       << " num inputs: " << getNumInputs();
+    if (prefix.empty())
+      prefix = name; // default prefix is the name
   }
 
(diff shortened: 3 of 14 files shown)


lhs8928 (approved):
LGTM

nntrainer/layers/layer_context.h
```diff
@@ -142,7 +142,8 @@ class InitLayerContext {
                              const Tensor::Initializer init,
                              const WeightRegularizer reg, const float reg_const,
                              const std::string &name, bool trainable = true) {
-    weights_spec.emplace_back(dim, init, reg, reg_const, trainable, name);
+    weights_spec.emplace_back(dim, init, reg, reg_const, trainable,
+                              getName() + ":" + name);
```
jijoongmoon: In this way, we need to guarantee those tensors are only used inside of layer then. Just for sure, Is there no case that call or expose beyond the layer such as network??

nntrainer/layers/layer_context.h
```diff
@@ -47,10 +47,13 @@ class InitLayerContext {
...
     in_place(in_place_),
     num_outputs(num_out),
-    name(n) {
+    name(n),
+    prefix("") {
     NNTR_THROW_IF(!validate(), std::invalid_argument)
       << "Invalid init context name: " << name
       << " num inputs: " << getNumInputs();
+    if (prefix.empty())
+      prefix = name; // default prefix is the name
```
jijoongmoon, line 56: I think we need to give this info say the default name prefix is name to custom layer developer. else if they are confused a little bit.

jijoongmoon (approved):
LGTM with minor comment

nntrainer/layers/dropout.cpp
```diff
@@ -26,9 +26,9 @@ void DropOutLayer::finalize(InitLayerContext &context) {
 
   mask_idx.reserve(input_dims.size());
   for (auto &t : input_dims) {
-    mask_idx.push_back(context.requestTensor(
-      t, context.getName() + ":Mask", Tensor::Initializer::NONE, false,
-      TensorLifespan::ITERATION_LIFESPAN));
+    mask_idx.push_back(
+      context.requestTensor(t, "Mask", Tensor::Initializer::NONE, false,
```
zhoonit, line 30: not related to the PR it self but maybe change "Mask -> "mask"??

nntrainer/layers/embedding.cpp
```diff
@@ -61,9 +61,9 @@ void EmbeddingLayer::finalize(InitLayerContext &context) {
   dim.width(out_dim);
   dim.batch(1);
 
-  weight_idx = context.requestWeight(
-    dim, weight_initializer, weight_regularizer, weight_regularizer_constant,
-    context.getName() + ":Embedding", true);
+  weight_idx =
+    context.requestWeight(dim, weight_initializer, weight_regularizer,
+                          weight_regularizer_constant, "Embedding", true);
```
zhoonit, line 66: not related to the PR it self but maybe change "Embedding-> "embedding"??

nntrainer/layers/layer_context.h
zhoonit: how about s/prefix/scope

and if scope is not given, we don't set scope?

kparichay: Yes, we need to make a document/guide on how to make a custom layer, and it must include this information.

zhoonit (approved):
LGTM with minor comments

nntrainer/layers/layer_context.h
kparichay: The tensors can be used externally as well. This patch changes the name which is given to the tensor at creation time. So if any code caches the name or uses it, they will use the correct name.

nntrainer/layers/dropout.cpp
kparichay: I will update here itself.

nntrainer/layers/embedding.cpp
kparichay: same here.

nntrainer/layers/layer_context.h
kparichay: Some scope will have to be set in order to make the names unique.

PR #1687, opened 2021-11-01
[graph] Skip backwarding bug fix + memory optimize

### LINKED ISSUE #1664: ENABLE_TEST=false does not work
ENABLE_TEST=false does not work on some models. 

Error is requesting gradient for non trainable tensors.

I think if there is a intermediat non trainable tensor, it seems affecting it.

### LINKED ISSUE #1686: [graph] Skipping backwarding is faulty
The current code for skipping backwarding assumes a linear model graph, and skips the backwarding only for the layers at the start of the model which dont need backwardng. However, this does not support for complicated graphs where intermediate layers also dont need backwarding.

kparichay:
Current implementation assumes that the graph is always linear and skips
the backwarding for contiguous set of layers at the start of the model
which should not do the backwarding.

This patch makes a fix to this issue but allowing layers in the middle
of the sorted graph to skip backwarding.

Resolves #1686
Resolves #1664

Note: this will provide minor speedup for complex models where some more layers backwarding can be skipped.

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -300,7 +300,7 @@ int NetworkGraph::checkCompiledGraph() {
   return ML_ERROR_NONE;
 }
 
-int NetworkGraph::checkInitializedGraph() {
+void NetworkGraph::markNodesForBackwarding() {
   /** accumulate all the nodes which must support backwarding */
   std::unordered_set<std::string> must_support_backwarding;
 
@@ -313,24 +313,21 @@ int NetworkGraph::checkInitializedGraph() {
     if (lnode->getTrainable() ||
         must_support_backwarding.find(lnode->getName()) !=
           must_support_backwarding.end()) {
+      lnode->needsCalcGradient(true);
+#ifdef ENABLE_TEST
+      if (lnode->supportBackwarding() && !optimize_memory) {
+        lnode->needsCalcDerivative(true);
+      }
+#endif
       for (auto const &out_layer : lnode->getOutputLayers()) {
         must_support_backwarding.insert(out_layer);
       }
     }
   }
 
-  /** verify all the required nodes support backwarding */
-  try {
-    for (auto const &node_name : must_support_backwarding)
-      LNODE(graph.getNode(node_name))->needsBackwarding(true);
-  } catch (std::exception &e) {
-    ml_loge(
-      "Backwaring required from layer which doesn't support backwarding: %s",
-      e.what());
-    return ML_ERROR_INVALID_PARAMETER;
-  }
-
-  return ML_ERROR_NONE;
+  /** mark all the required nodes support backwarding */
+  for (auto const &node_name : must_support_backwarding)
+    LNODE(graph.getNode(node_name))->needsCalcDerivative(true);
 }
 
 int NetworkGraph::realizeGraph() {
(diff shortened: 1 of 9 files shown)


kparichay:
V2:
- apply bug fix for lstmcell
- Currently, the memory was allocated assuming all the nodes were running
backwarding, and all the layers are trainable. But that is not true.
This patch applies the corresponding optimization and allocates memory
corresponding to the functions which will run on these nodes.

nntrainer/graph/network_graph.cpp
```diff
@@ -313,6 +313,12 @@ int NetworkGraph::checkInitializedGraph() {
     if (lnode->getTrainable() ||
         must_support_backwarding.find(lnode->getName()) !=
           must_support_backwarding.end()) {
+      lnode->needsCalcGradient(true);
+#ifdef ENABLE_TEST
```
zhoonit, line 317: This test flag is likely to be broken. Could we stick to less optimized version, and later optimize + fix test along instead?

nntrainer/graph/network_graph.cpp
```diff
@@ -313,6 +313,12 @@ int NetworkGraph::checkInitializedGraph() {
...
+      if (lnode->supportBackwarding() && !optimize_memory) {
+        lnode->needsCalcDerivative(true);
```
zhoonit, line 319: This function name suggest to check and not to touch the node property. If it is now, how about split this to different function or change the function name?

nntrainer/graph/network_graph.cpp
```diff
@@ -313,6 +313,12 @@ int NetworkGraph::checkInitializedGraph() {
     if (lnode->getTrainable() ||
         must_support_backwarding.find(lnode->getName()) !=
           must_support_backwarding.end()) {
+      lnode->needsCalcGradient(true);
```
zhoonit, line 316: If we are going to do this, I think layerNode::calcGradient should not implicitly make calcGradient noop.

below is related snippet.
```cpp
  if (getTrainable())
    layer->calcGradient(*run_context);
```

nntrainer/graph/network_graph.cpp
kparichay: There are two needsCalcDerivative(). One is a getter, one is a setter. Do you want different names for the two?

nntrainer/graph/network_graph.cpp
kparichay: updated  `if (getTrainable())` to `if (needs_calc_gradient)`.

nntrainer/graph/network_graph.cpp
zhoonit: no it's about `checkInitializedGraph()` seems not clear to me.

it sounds like a sanity check but actually writing something

nntrainer/graph/network_graph.cpp
kparichay: In general, we want to run optimized version where we want to skip running the nodes for backwarding where its not needed, and so not allocate memory for it either.
With ENABLE_TEST, we disable such optimizations. The issue is not the memory optimization here but performance as calcDerivative runs for nodes which ideally should not.

nntrainer/graph/network_graph.cpp
kparichay: I understand, I will split it into two functions.

nntrainer/graph/network_graph.cpp
zhoonit: I agree, my concern is that by we have `#ifndef ENABLE_TEST` part is never being tested, so from #1664 there is symptoms of a bug (android build is enabling this flag to true for now).

So I am suggesting to comment this out, remove #ifdef flag and make some changes to the test while switching to the optimized version.

jijoongmoon: > performance as calcDerivative runs for nodes which ideally should not.

I wonder which nodes are? You mean the feature extract cases which only support forwarding? If it is, then it's ok.
But other than that... None of the cases comes to my mind.

nntrainer/layers/layer_node.h
```diff
@@ -661,11 +661,47 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    */
   std::unique_ptr<LayerNode> cloneConfiguration();
 
+  /**
+   * @brief Set if the layer needs to do derivative calculation
+   *
+   * @param nb true if the layer needs to do backwarding, else false
+   */
+  void needsCalcDerivative(bool nb) {
```
jijoongmoon, line 669: so.. we have..
1. supportBackwarding
2. needsBackwarding --> needsCalcDerivative & needsCalCGradient?

nntrainer/graph/network_graph.cpp
kparichay: If the first few layers are non-trainable but support backward (for example, input layer followed by multi-out layer), the multi-out layer's calcDerivative should not be called. 
This patch ensures that multi-out layer's calcDerivative is not called, and the corresponding memory is also not allocated.

kparichay: There are only three locations where ENABLE_TEST is being used (2 of which are from this PR itself). I will check the issue from 1664 and see what is the problem, but ideally, we need to run some tests with ENABLE_TEST as well.

nntrainer/layers/layer_node.h
kparichay: supportBackwarding is acutally supportCalcDerivative (didn't change this as its easier to understand for the layer developers), but internally, needsBackwarding is created and split into two.

jijoongmoon: Thanks for the explanation.

jijoongmoon (approved):
LGTM

zhoonit (approved):
LGTM

nntrainer/graph/network_graph.cpp
kparichay: Marking resolves #1664 as this patch solves that issue.

nntrainer/graph/network_graph.cpp
kparichay: renamed to `markNodesForBackwarding`.

PR #1689, opened 2021-11-02
[wait for #1687] [layer] Inplace support for multiout layer 

kparichay:
This patch adds support for in-place execution of the multiout layer.
Corresponding unittests are also added.


nntrainer/graph/network_graph.cpp
```diff
@@ -696,7 +702,7 @@ bool NetworkGraph::canExecuteInPlace(const std::shared_ptr<LayerNode> &lnode) {
    * 2. if the layer is not supporting backwarding, there is no dependency
    * requirement with other nodes for backwarding.
    */
-  if (no_op(lnode) || !lnode->supportBackwarding())
+  if (no_op(lnode) || no_op_shared(lnode) || !lnode->supportBackwarding())
```
jijoongmoon, line 705: need to update the comment. 1, 2, and 3. Need to update no-op with shared.

nntrainer/graph/network_graph.cpp
```diff
@@ -736,6 +747,25 @@ void NetworkGraph::inPlaceOptimize() {
...
+ * @brief Set the Inplace Shared Memory Config By Layer object
+ *
+ * @param lnode layer node object
+ * @param shared_var if the variable should be shared
+ * @param shared_grad if the gradient should be shared
+ */
+static void
+setInplaceSharedMemoryConfigByLayer(const std::shared_ptr<LayerNode> &lnode,
+                                    bool &shared_var, bool &shared_grad) {
+  if (lnode->getType() == MultiOutLayer::type) {
```
jijoongmoon, line 760: what else layers are in this kind? split? 
and the concat, addition layer can be in shared_var and shared_grad=true? 
Activation can be share every thing. 
Could you listed up in comment?

kparichay: I will add as a comment.

nntrainer/graph/network_graph.cpp
kparichay: done.

nntrainer/graph/network_graph.cpp
```diff
@@ -696,27 +697,59 @@ bool NetworkGraph::canExecuteInPlace(const std::shared_ptr<LayerNode> &lnode) {
...
+   * @note Conditions to decide if this layer node can be in-place:
    * 1. if the layer is a no-op, then it can operate in-place as it is not
    * modifying its input/output tensors and does not need to check its
    * neighboring nodes for dependency.
    * 2. if the layer is not supporting backwarding, there is no dependency
    * requirement with other nodes for backwarding.
+   *
+   * @note Conditions to decide the type of inplace for this layer:
+   * 1. if the previous layers were restricting, then this layer will also be
+   * restricting.
```
zhoonit, line 709: what's the meaning of restricting?

nntrainer/layers/layer_node.h
```diff
@@ -52,6 +52,18 @@ class Activation;
 class SharedFrom;
 } // namespace props
 
+/**
+ * @brief Enum class for the various types of inplace modes supported by layer
+ *
+ */
+enum class InPlace {
+  NONE,           /**< layer is not inplace */
+  RESTRICTING,    /**< layer is in-place and does place restriction on layers
```
zhoonit, line 61: how about PREEMPTIVE to convey meaning?

nntrainer/graph/network_graph.cpp
kparichay: restricting means that making the current layer in-place will add restrictions on making the next layer in-place.

nntrainer/layers/layer_node.h
kparichay: I don't think that the meaning of preemptive applies here. Check https://github.com/nnstreamer/nntrainer/pull/1689#discussion_r746186559 for more details.

zhoonit: Cool.

zhoonit (approved):
LGTM

nntrainer/layers/layer_node.h
zhoonit: Idea to refine this.

1. There is 2 properties of requestSpec()
 - clean inplace and dirty inplace
 - if clean (meaning that tensor has first set and not overwritten) it is allowed to inplace (clean or dirty)
 - if dirty (meaning that tensor was overwritten) it is allowed to clean inplace but cannot do dirty inplace

| current tensor state | trying tensor state | can in place? |
|-----------------------|---------------------|------------|
| clean                        | clean inplace         | O |
| clean | dirty inplace | O |
| dirty | clean inplace | O |
| dirty | dirty inplace | X |

kparichay: Yes, clean and dirty names sound good, but have a slightly different concepts. `clean` and `dirty` seem to give the notion if the layer operation will modify the input tensor. But instead, there are two concepts:
- if the layer modifies the input tensor
- does the layer require its input/output tensor in its backward operation

So, given a state associated to the output from the previous layer output, how does the current state modify its input with inplace operation (1) and what is the requirement of the current layer wrt to its backward operation (2), both needs to be represented. Lets define two concepts:
a. Layer operation : modifies/not-modifies
b. Layer backwarding requirement : needed/non-needed

Based on (b), an output of a layer can be assumed to be needed/non-needed. It will be non-needed for out of place layers. If the inputs for a layer not-needed (b), then this layer operation can be always performed in-place (if possible) (a) and the tensor state will be updated based on (b).  If the inputs for a layer are needed, then only not-modifies (b) layer operation can work in-place, independent of layer backwarding requirement. If the layer operation is modifies (b), then it will have to work out-of-place.

As in-place optimization is performed at the compile time, tensor objects are not yet created, and its better to associate these properties to layer (defining the way they are operating) and can be checked with its neighbors based on their connections.

PS. there can be better names than `modifies` and `needed`.

zhoonit: having `modified`, `needed` sounds good, for needed we won't need another boolean but we can just check the max execution order is less then new min execution order.

Right, I am thinking to extend the concept to the normal tensor (not limited to the inputs, outputs).

| previously modified | needed after my execution order | view type | actually requested |
|-|-|-|-|
| - | - | READ_ONLY_VIEW | `view()` |
| x | x | MAYBE_MODIFIYING_VIEW | `view()` |
| o | o | MAYBE_MODIFIYING_VIEW |  `request()` |
| o | x | MAYBE_MODIFIYING_VIEW |  `view()` |
| x | o | MAYBE_MODIFIYING_VIEW |  `request()` |

Which contracts to only checking `needed after my execution order`

In this way, finalize does not need to query execute inplace, nor graph need to check whether it should inplace optimize but we have defined rule to hint graph how it should be initialized.

Any thoughts?

fyi) please check out #1700 as a proposal.

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_node.h
jijoongmoon: I prefer the modified and needed rather than clean and dirty. I agree it is different. I wonder requrest() means create()??

PR #1692, opened 2021-11-03
[tensor] Add checks for non-contiguous tensor before operations @open sesame 11/19 23:17

kparichay:
This patch adds checks for contiguous tensors ensuring that operations
which don't support non-contiguous tensors throw on receiving the
non-contiguous tensors.

This is in preparation for making the concat and split layer inplace.

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -214,6 +214,12 @@ bool Tensor::operator==(const Tensor &rhs) const {
   const float *data = getData();
   const float *rdata = rhs.getData();
 
+  if (contiguous != rhs.contiguous)
+    return false;
+
+  if (strides != rhs.strides)
+    return false;
+
   for (size_t i = 0; i < len; ++i) {
     /** not checking sign change is intentional to avoid float calculation
      * errors around 0 */
@@ -226,6 +232,9 @@ bool Tensor::operator==(const Tensor &rhs) const {
 }
 
 template <typename T> void Tensor::setDist(T dist) {
+  NNTR_THROW_IF(!contiguous, std::invalid_argument)
+    << getName() << " Tensor is not contiguous, cannot set distribution";
+
   float *data = getData();
   unsigned int len = size();
   for (unsigned int i = 0; i < len; ++i) {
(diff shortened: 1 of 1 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -687,27 +688,59 @@ bool NetworkGraph::canExecuteInPlace(const std::shared_ptr<LayerNode> &lnode) {
...
+   * @note Conditions to decide if this layer node can be in-place:
    * 1. if the layer is a no-op, then it can operate in-place as it is not
    * modifying its input/output tensors and does not need to check its
    * neighboring nodes for dependency.
    * 2. if the layer is not supporting backwarding, there is no dependency
    * requirement with other nodes for backwarding.
+   *
+   * @note Conditions to decide the type of inplace for this layer:
+   * 1. if the previous layers were restricting, then this layer will also be
+   * restricting.
```
jijoongmoon, line 700: It might be good if you give some examples here. the previous layer is restricted means that the previous layer is in-place layer. so.. the next layer of in place layer will be in place as well?

jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -226,6 +232,9 @@ bool Tensor::operator==(const Tensor &rhs) const {
 }
 
 template <typename T> void Tensor::setDist(T dist) {
+  NNTR_THROW_IF(!contiguous, std::invalid_argument)
+    << getName() << " Tensor is not contiguous, cannot set distribution";
```
zhoonit, line 236: can we use apply and remove this?

zhoonit (commented):
Wonder if this is a temporary measure for now for some operations? 

eg) tensor::read, tensor::save, setDist should be allowed by iterating element by element

lhs8928 (approved):
LGTM

nntrainer/graph/network_graph.cpp
kparichay: This was part of the previous PR, where it was expanded upon.

kparichay:
> Wonder if this is a temporary measure for now for some operations?
> 
> eg) tensor::read, tensor::save, setDist should be allowed by iterating element by element

Tensor::read, save, setDist etc will eventually we working with the original tensor than a strided view of the tensor. This check ensures that all the tensors operations do not unintentionally work on tensors assuming them to be non-strided.

zhoonit (approved):
Cool. LGTM :)

zhoonit:
Please hold merging this till today (svace issue)

PR #1700, opened 2021-11-10
[Tp] Propose new tensor spec

zhoonit:

<details><summary>[Tp] Propose new tensor spec</summary><br />

This patch proposes new tensor specification

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/tensor/manager.h
+++ b/nntrainer/tensor/manager.h
@@ -19,11 +19,14 @@
 
 #ifndef __MANAGER_H__
 #define __MANAGER_H__
+#include "tensor.h"
+#include "tensor_wrap_specs.h"
 #ifdef __cplusplus
 
 #include <functional>
 #include <memory>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 #include <basic_planner.h>
@@ -108,6 +111,22 @@ private:
 class Manager {
 
 public:
+  /**
+   * @brief Tensor Group Type
+   * @note this is not mutually exclusive list, a tensor might be identified as
+   * input as well as output
+   *
+   */
+  enum TensorGroupType {
+    INPUT = 0,   /**< Input of an operation */
+    OUTPUT = 1,  /**< Output of an operation */
+    WEIGHT = 2,  /**< Weight of an operation */
+    TENSORS = 3, /**< Extra states of an operation */
+  };
+
+  constexpr inline static unsigned NUM_TENSOR_GROUP_TYPE =
+    4; /**< number of tensor group type */
+
   /**
    * @brief     Constructor of Manager
    */
@@ -361,6 +380,28 @@ public:
     }
   }
 
+  /**
+   * @brief request Tensor with weight specification
+   *
+   * @param spec specification
+   * @param identify_as identify as tensor as a group
+   * @return Tensor* tensor
+   */
+  Tensor *requestTensor(const WeightSpecV2 &spec, TensorGroupType identify_as);
+
+  /**
+   * @brief request Tensor with variable + gradient specification
+   *
+   * @param spec specification
+   * @param identify_as identify as tensor as a group
+   * @param exec_order execution order to refer to
+   * @param scope common scope to attach in front of current specification name
+   * @return Tensor* tensor
+   */
+  Tensor *requestTensor(const VarGradSpecV2 &spec, TensorGroupType identify_as,
+                        const GraphNode::ExecutionOrder &exec_order,
+                        const std::string &scope = "");
+
 private:
   /** @todo: merge this list to one */
   std::vector<std::unique_ptr<Weight>>
@@ -371,6 +412,9 @@ private:
     outputs_v2; /**< outputs for the layers */
   std::vector<std::unique_ptr<Var_Grad>>
     tensors_v2; /**< extra tensors required by the layers */
+
+  std::array<std::vector<std::unique_ptr<Var_Grad>>, NUM_TENSOR_GROUP_TYPE>
+    tensor_book; /**< reference to tensor book kept */
 
   TensorPool weight_pool; /**< tensor pool to request tensors */
   TensorPool tensor_pool; /**< tensor pool to request tensors */
--- a/nntrainer/tensor/tensor_pool.h
+++ b/nntrainer/tensor/tensor_pool.h
@@ -197,7 +197,7 @@ public:
    * while extending the tensor's life according to the given arguments.
    * @note Created (or extended) tensor is considered identical and managed. It
    * is invalid to create a tensor with lifespan::UNMANAGED or dimension and
-   * initializer is different unon extension.
+   * initializer is different upon extension.
    *
    * @param name Name of the tensor
    * @param dim dimension
--- a/nntrainer/tensor/tensor_wrap_specs.h
+++ b/nntrainer/tensor/tensor_wrap_specs.h
@@ -14,6 +14,7 @@
 #ifndef __TENSOR_WRAP_SPECS_H__
 #define __TENSOR_WRAP_SPECS_H__
 
+#include <memory>
 #include <tuple>
 
 #include <tensor.h>
@@ -35,19 +36,23 @@ enum class WeightRegularizer {
  *
  */
 enum class TensorLifespan {
-  UNMANAGED = 0b0, /**< tensor with no lifespan, will not be allocated */
-  FORWARD_FUNC_LIFESPAN = 0b01,  /**< tensor must not be reset before during the
-                            forward function call, eg. temporary tensors
-                            needed during forward operations */
-  BACKWARD_FUNC_LIFESPAN = 0b10, /**< tensor must not be reset before during the
-                            backward function call, eg. temporary tensors
-                            needed during backward operations */
-  ITERATION_LIFESPAN = 0b11,     /**< tensor must not be reset until the owning
-                            layer     finishes its execution in the current
-                            iteration,     eg. hidden memory/cells of RNN */
-  EPOCH_LIFESPAN = 0b111, /**< tensor must not be reset before the epoch ends */
-  MAX_LIFESPAN = 0b1111,  /**< tensor must not be reset until the end of the
-                   model  execution, eg. layer weights */
+  UNMANAGED = 0b000, /**< tensor with no lifespan, will not be allocated */
+  FORWARD_FUNC_LIFESPAN = 0b001, /**< tensor must not be reset before during the
+                           forward function call, eg. temporary tensors
+                           needed during forward operations */
+  CALC_DERIV_LIFESPAN = 0b010,   /**< must be valid during calcDerivative() */
+  CALC_GRAD_LIFESPAN = 0b100, /**< tensor must be valid during calcGradient() */
+  CALC_GRAD_DERIV_LIFESPAN = 0b110, /**< tensor must not be reset before during
+                             the calc_grad and clac_deriv call, eg. temporary
+                             tensors needed during backward operations */
+  BACKWARD_FUNC_LIFESPAN =
+    CALC_GRAD_DERIV_LIFESPAN, /**< Alias of CALC_GRAD_DERIV_LIFESPAN */
+  ITERATION_LIFESPAN = 0b111, /**< tensor must not be reset until the owning
+                        layer finishes its execution in the current
+                        iteration, eg. hidden memory/cells of RNN */
+  EPOCH_LIFESPAN = 0b1111,    /**< tensor must be valid before the epoch ends */
+  MAX_LIFESPAN = 0b11111,     /**< tensor must not be reset until the end of the
+                      model  execution, eg. layer weights */
 };
 
 /**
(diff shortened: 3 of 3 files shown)


nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -70,6 +74,40 @@ typedef std::tuple<TensorDim, Tensor::Initializer, bool, const std::string,
                    TensorLifespan>
   VarGradSpec;
 
+/**
+ * @brief Tensor Specification which describes how this tensor should be
+ * allocated and managed
+ *
+ */
+struct TensorSpecV2 {
+  enum class RequestType {
```
jijoongmoon, line 83: So.. every request*() make this struct and used in manager to set the memory pool and planner?

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -70,6 +74,40 @@ typedef std::tuple<TensorDim, Tensor::Initializer, bool, const std::string,
...
+    PLACEHOLDER,
+    UNIQUE,
+    READ_ONLY_VIEW,
+    MAYBE_MODIFYING_VIEW,
+    SHARED,
```
jijoongmoon, line 88: Need more explanation what each key means.

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -70,6 +74,40 @@ typedef std::tuple<TensorDim, Tensor::Initializer, bool, const std::string,
...
+/**
+ * @brief Tensor Specification which describes how this tensor should be
+ * allocated and managed
+ *
+ */
+struct TensorSpecV2 {
+  enum class RequestType {
+    PLACEHOLDER,
+    UNIQUE,
+    READ_ONLY_VIEW,
```
kparichay, line 86: how about keeping just 1 view and 1 shared:
VIEW: non-modifiable internal memory structure (shared with source tensor) with size <= source tensor size
SHARED: modifiable internal memory structure (shared with source tensor) with size <= source tensor size

This keeps view semantics same as string_view (https://en.cppreference.com/w/cpp/string/basic_string_view).

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -70,6 +74,40 @@ typedef std::tuple<TensorDim, Tensor::Initializer, bool, const std::string,
...
 
+/**
+ * @brief Tensor Specification which describes how this tensor should be
+ * allocated and managed
+ *
+ */
+struct TensorSpecV2 {
+  enum class RequestType {
+    PLACEHOLDER,
+    UNIQUE,
```
kparichay, line 85: maybe UNIQUE -> NEW? All tensors are themselves unique anyway but sharing internal memory structure.

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -35,19 +36,22 @@ enum class WeightRegularizer {
...
-                            needed during forward operations */
-  BACKWARD_FUNC_LIFESPAN = 0b10, /**< tensor must not be reset before during the
-                            backward function call, eg. temporary tensors
+  UNMANAGED = 0b000, /**< tensor with no lifespan, will not be allocated */
+  FORWARD_FUNC_LIFESPAN = 0b001, /**< tensor must not be reset before during the
+                           forward function call, eg. temporary tensors
+                           needed during forward operations */
+  CALC_DERIV_LIFESPAN = 0b010,
+  CALC_GRAD_LIFESPAN = 0b100,
+  BACKWARD_FUNC_LIFESPAN = 0b110, /**< tensor must not be reset before during
```
kparichay, line 45: CALC_GRAD_DERIV_LIFESPAN?

nntrainer/tensor/manager.h
```diff
@@ -355,6 +358,12 @@ class Manager {
     }
   }
 
+  Tensor *requestTensor(const WeightSpecV2 &spec,
+                        unsigned /*<or enum*/ identify_as);
+
+  Tensor *requestTensor(const VarGradSpecV2 &spec,
+                        unsigned /*<or enum*/ identify_as);
```
kparichay, line 365: can you expand on identify_as?

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: That is can't be done. we need three concept. 

1. view changing data (MAYBE_MODIFYING_VIEW)
2. view not changing data but changing meta (READ_ONLY_VIEW)
3. share everything (SHARED)

We can seperate this concept out to another dimension but it will be dirtier.

nntrainer/tensor/manager.h
zhoonit: identify_as will be input, output, weight, tensors or weight / tensors.

This might not be needed afterward but this will work as a bridge for now.

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: yes, init context will have this specification to be passed to manager.

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: Yes, I'll add as a comment here :)

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: I like the concept. Let me delay this until there seems a way to represent the third state.

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: This is to represent the ownership semantics of `UNIQUE_`ptrvs `SHARED_`ptr, I don't mind NEW although please review this idea once again :)

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: BACKWARD_FUNC_LIFESPAN is already in use so I guess this has to be done with in a separate PR, if we mean to.

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -35,19 +36,22 @@ enum class WeightRegularizer {
...
-                            forward function call, eg. temporary tensors
-                            needed during forward operations */
-  BACKWARD_FUNC_LIFESPAN = 0b10, /**< tensor must not be reset before during the
-                            backward function call, eg. temporary tensors
+  UNMANAGED = 0b000, /**< tensor with no lifespan, will not be allocated */
+  FORWARD_FUNC_LIFESPAN = 0b001, /**< tensor must not be reset before during the
+                           forward function call, eg. temporary tensors
+                           needed during forward operations */
+  CALC_DERIV_LIFESPAN = 0b010,
+  CALC_GRAD_LIFESPAN = 0b100,
```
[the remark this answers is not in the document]
zhoonit: Yes, BACKWARD_FUNC_LIFESPAN is combination of CALC_DERIV & CALC_GRAD.

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -70,6 +74,40 @@ typedef std::tuple<TensorDim, Tensor::Initializer, bool, const std::string,
...
+    MAYBE_MODIFYING_VIEW,
+    SHARED,
+  };
+  RequestType request_type = RequestType::UNIQUE;
+  std::string name;
+  TensorDim dim;
+  TensorLifespan ls;
+  Tensor::Initializer initializer = Tensor::Initializer::NONE;
+
+  /** ONLY USED FOR READ_ONLY_VIEW, READ_WRITE_VIEW */
```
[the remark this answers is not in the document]
zhoonit: Will correct this :)

nntrainer/tensor/manager.cpp
```diff
@@ -278,31 +278,23 @@ std::vector<Weight *> Manager::requestWeights(
...
 
       if (trainable && need_gradient) {
-        grad = tensor_pool.requestPrerequestedTensor(
-          dim, grad_exec_order, grad_ls,
-          name + Var_Grad::grad_suffix,        /// name
-          shared_name + Var_Grad::grad_suffix, /// shared name
-          Tensor::Initializer::ZEROS           /// tensor initializer
-        );
+        grad = tensor_pool.view(name + Var_Grad::grad_suffix,
+                                shared_name + Var_Grad::grad_suffix, dim,
```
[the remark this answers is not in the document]
zhoonit: This is likely to be changed to `shared`. For now this is changing `requestPrerequested` to `view` without deep thinking :).

nntrainer/tensor/manager.cpp
```diff
@@ -355,27 +347,23 @@ Manager::requestTensors(const GraphNode &node,
...
       if (need_grad && tspan > TensorLifespan::FORWARD_FUNC_LIFESPAN) {
-        grad = tensor_pool.requestPrerequestedTensor(
-          dim, grad_exec_order, tspan,
-          name + Var_Grad::grad_suffix, /// name
-          shared_name + Var_Grad::grad_suffix,
-          Tensor::Initializer::ZEROS /// tensor initializer
-        );
+        grad = tensor_pool.view(name + Var_Grad::grad_suffix,
+                                shared_name + Var_Grad::grad_suffix, dim,
+                                grad_exec_order, tspan);
```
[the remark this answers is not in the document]
zhoonit: yes, view assumes that it is initialized by someone else :)

nntrainer/tensor/manager.cpp
```diff
@@ -601,9 +539,9 @@ std::vector<Tensor *> Manager::requestWeightOptimizerVariables(
   ret.reserve(dims.size());
 
   for (unsigned int idx = 0; idx < dims.size(); idx++)
-    ret.push_back(tensor_pool.requestTensor(dims[idx], exec_order, lifespan,
-                                            name + ":opt" + std::to_string(idx),
-                                            initializer));
+    ret.push_back(tensor_pool.request(name + ":opt" + std::to_string(idx),
```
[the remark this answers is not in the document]
zhoonit: opt is currently requested to the tensor pool.  Either should work the same, I think, though

nntrainer/tensor/tensor_pool.cpp
```diff
@@ -207,56 +194,174 @@ void TensorPool::deallocate() {
   }
 }
 
+const std::vector<unsigned int> &
+TensorPool::getExecutionOrder(const std::string &name) {
+  return std::get<SourceDetails>(getSourceSpec(name).details).exec_order;
```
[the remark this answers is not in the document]
zhoonit: Yes, it is hard restriction that source details must include all the dependent tensors's execution order.

nntrainer/tensor/tensor_pool.h
```diff
@@ -198,60 +118,211 @@ class TensorPool {
...
+  Tensor *
+  requestOrExtend(const std::string &name, const TensorDim &dim,
+                  const std::vector<unsigned int> &exec_order,
+                  TensorLifespan lifespan,
+                  const Tensor::Initializer &init = Tensor::Initializer::NONE);
+
+  /**
+   * @brief reidentify the source of already created tensor (or view).
+   * @note if @a dest tensor is a view of another tensor, the old source tensor
+   * of the view will become a view of @a new_src.
```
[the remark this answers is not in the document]
zhoonit: Yes! old source tensor will be come view of new source tensor.

A(source) -> B(View)

When reidentifySource(A, C)
C(new source) -> A (view) -> B (view).

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: I updated, and made an alias for the backward compat

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: Done

nntrainer/tensor/manager.h
zhoonit: added as 

```cpp

  /**
   * @brief Tensor Group Type
   * @note this is not mutually exclusive list, a tensor might be identified as
   * input as well as output
   *
   */
  enum TensorGroupType {
    INPUT = 0,   /**< Input of an operation */
    OUTPUT = 1,  /**< Output of an operation */
    WEIGHT = 2,  /**< Weight of an operation */
    TENSORS = 3, /**< Extra states of an operation */
  };
```

nntrainer/tensor/tensor_wrap_specs.h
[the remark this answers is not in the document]
zhoonit: Done

nntrainer/tensor/manager.cpp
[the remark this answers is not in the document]
zhoonit: This is done from #1707 and added as a merge commit here

nntrainer/tensor/manager.cpp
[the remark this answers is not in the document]
jijoongmoon: Oh.. I see, thanks.

nntrainer/tensor/manager.cpp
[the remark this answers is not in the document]
jijoongmoon: In optimization point, I think if the opt tensor is temporal, then it might be good to be tensor pool, but if it is like weight, then I guess it need to be placed in weight. but WM and WV need to be stores like weight. Aren't they?

jijoongmoon (approved):
LGTM! nice work!

nntrainer/tensor/manager.cpp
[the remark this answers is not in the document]
zhoonit: @kparichay Any idea on this?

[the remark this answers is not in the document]
kparichay: The reason to have weight_pool separate is to have weights allocate at initialize time. But tensor_pool is allocated at train/inference time. As optimizer weights are not needed at initialize time, I am not sure if its a good idea to merge them to weight_pool.

Independent of the above comment, the weight_pool and tensor_pool will later be merged (with some ammendments to Model::Save() and Model::Load()).

nntrainer/tensor/manager.h
```diff
@@ -108,6 +111,22 @@ class MMapedMemory {
 class Manager {
 
 public:
+  /**
+   * @brief Tensor Group Type
+   * @note this is not mutually exclusive list, a tensor might be identified as
+   * input as well as output
```
kparichay, line 117: in order to identify as both input and output, do we keep a list of this enum?
IMO INPUT and OUTPUT can be combined to one as a tensor allocated by tensor_pool and is an INPUT will always be OUTPUT to some other layer. So, all INPUT tensors will be OUTPUT for some layers (ignoring the special cases of the last layer where OUTPUT will not be INPUT).

kparichay: If the purpose is to just identify the request, then consider renaming to TensorRequestUsageType?

nntrainer/tensor/manager.h
```diff
@@ -372,6 +413,9 @@ class Manager {
   std::vector<std::unique_ptr<Var_Grad>>
     tensors_v2; /**< extra tensors required by the layers */
 
+  std::array<std::vector<std::unique_ptr<Var_Grad>>, NUM_TENSOR_GROUP_TYPE>
```
kparichay, line 416: how will you store std::unique_ptr<Weight> in this array?

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -35,19 +36,23 @@ enum class WeightRegularizer {
...
-  EPOCH_LIFESPAN = 0b111, /**< tensor must not be reset before the epoch ends */
-  MAX_LIFESPAN = 0b1111,  /**< tensor must not be reset until the end of the
-                   model  execution, eg. layer weights */
+  UNMANAGED = 0b000, /**< tensor with no lifespan, will not be allocated */
+  FORWARD_FUNC_LIFESPAN = 0b001, /**< tensor must not be reset before during the
+                           forward function call, eg. temporary tensors
+                           needed during forward operations */
+  CALC_DERIV_LIFESPAN = 0b010,   /**< must be valid during calcDerivative() */
+  CALC_GRAD_LIFESPAN = 0b100, /**< tensor must be valid during calcGradient() */
+  CALC_GRAD_DERIV_LIFESPAN = 0b110, /**< tensor must not be reset before during
```
kparichay, line 45:
```suggestion
  CALC_GRAD_DERIV_LIFESPAN = 0b110, /**< tensor must not be reset during
```

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -70,6 +75,75 @@ typedef std::tuple<TensorDim, Tensor::Initializer, bool, const std::string,
...
+  enum class RequestType {
+    PLACEHOLDER, /**< Placeholder defines that nntrainer should never care about
+                    the memory inside the particualar tensor */
+    UNIQUE, /**< Unique means a simple tensor that will be owned explicitly the
+               current request */
+    READ_ONLY_VIEW, /**< Readonly view defines a view of which ownership of @a
+                       underlying memory is at another tensor, also hinting
+                       nntrainer that the operation upon this particular tensor
+                       will never change value of the underlying memory */
+    MAYBE_MODIFYING_VIEW, /**< Maybe modifying view defines a (possible) view of
```
kparichay, line 99: consider the following nomenclature:
VIEW: read-only access to memory with size <= source size
SHARED: read-write access to memory with size <= source size
LINK: (inspired by symlink) link to the original tensor

kparichay (approved):
LGTM other than some new minor comments

nntrainer/tensor/manager.h
zhoonit: Weight is Var_Grad.
When accessing weight, we can static_cast before return.

nntrainer/tensor/tensor_wrap_specs.h
zhoonit: I'll open a issue regarding this.

nntrainer/tensor/manager.h
zhoonit: > in order to identify as both input and output, do we keep a list of this enum?
IMO INPUT and OUTPUT can be combined to one as a tensor allocated by tensor_pool and is an INPUT will always be OUTPUT to some other layer. So, all INPUT tensors will be OUTPUT for some layers (ignoring the special cases of the last layer where OUTPUT will not be INPUT).

I agree, we can have only input tensors, let's migrate to having only single concept for this.

TensorUsageType also sounds good.

PR #1704, opened 2021-11-12
[layer] Support MoL attention layer forwarding

kparichay:
- This patch prepares scaffolding for MoL Attention Layer.
Attention layer has also been updated for MoLAttentionLayer to extend
AttentionLayer.
- This patch provides support for all the properties required for MoL
attention layer.
- This patch provides the forwarding implementation for the mol attention
layer. New underlying requirements have been added with this which
requires more tensor operations to support strided operations like
softmax which will be supported soon.
- add basic validation unittests + forwarding fix for mol
attention layer.
Further a validation test for mol layer is also added for larger size,
and nntrainer model is fixed to support multiple inputs.
- Add support for batched dot operaiton. Provide cleanup for both the
attention layers.\

### DIFF
--- a/nntrainer/layers/acti_func.cpp
+++ b/nntrainer/layers/acti_func.cpp
@@ -156,7 +156,7 @@ Tensor &ActiFunc::softmax(Tensor const &t, Tensor &output) {
   float *dp;
   float *rp;
 
-  /** TODO: fix this check */
+  /** TODO: support strided operations */
   if (t.size() == output.size() && t.getStrides() != output.getStrides())
     throw std::invalid_argument(
       "Softmax does not support operating on strided tensors");
@@ -194,7 +194,7 @@ Tensor &ActiFunc::softmax(Tensor const &t, Tensor &output) {
 
 Tensor &ActiFunc::softmaxPrime(Tensor const &x, Tensor &output,
                                Tensor const &derivative) {
-  /** TODO: fix this check */
+  /** TODO: support strided operations */
   if ((x.size() == output.size() && x.getStrides() != output.getStrides()) ||
       (x.size() == derivative.size() &&
        x.getStrides() != derivative.getStrides()))
--- a/nntrainer/layers/acti_func.cpp
+++ b/nntrainer/layers/acti_func.cpp
@@ -156,6 +156,11 @@ Tensor &ActiFunc::softmax(Tensor const &t, Tensor &output) {
   float *dp;
   float *rp;
 
+  /** TODO: fix this check */
+  if (t.size() == output.size() && t.getStrides() != output.getStrides())
+    throw std::invalid_argument(
+      "Softmax does not support operating on strided tensors");
+
   Tensor divisor = t.clone();
 
   dp = divisor.getData();
@@ -189,6 +194,13 @@ Tensor &ActiFunc::softmax(Tensor const &t, Tensor &output) {
 
 Tensor &ActiFunc::softmaxPrime(Tensor const &x, Tensor &output,
                                Tensor const &derivative) {
+  /** TODO: fix this check */
+  if ((x.size() == output.size() && x.getStrides() != output.getStrides()) ||
+      (x.size() == derivative.size() &&
+       x.getStrides() != derivative.getStrides()))
+    throw std::invalid_argument(
+      "SoftmaxPrime does not support operating on strided tensors");
+
   unsigned int batch = x.batch();
   unsigned int channel = x.channel();
   unsigned int height = x.height();
--- a/nntrainer/layers/mol_attention_layer.cpp
+++ b/nntrainer/layers/mol_attention_layer.cpp
@@ -42,7 +42,13 @@ void MoLAttentionLayer::finalize(InitLayerContext &context) {
   tanh.setActiFunc(ActivationType::ACT_TANH);
   sigmoid.setActiFunc(ActivationType::ACT_SIGMOID);
 
+  NNTR_THROW_IF(std::get<props::Unit>(mol_props).empty(), std::invalid_argument)
+    << "Number of units not provided for layer " << context.getName();
   auto unit = std::get<props::Unit>(mol_props).get();
+
+  NNTR_THROW_IF(std::get<props::MoL_K>(mol_props).empty(),
+                std::invalid_argument)
+    << "MoL_K property not provided for layer " << context.getName();
   auto mol_k = std::get<props::MoL_K>(mol_props).get();
 
   TensorDim mlp_w_dim = {query_dim.width(), unit};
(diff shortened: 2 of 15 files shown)


nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is MoL Attention Layer Class for Neural Network
+ *
+ */
+
+#include <mol_attention_layer.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+
```
zhoonit, line 19: please add <layer_context.h>, related to #1724

kparichay: done.

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -0,0 +1,47 @@
...
+  finalizeCommon(context);
+  /** NYI */
+}
+
+void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
+  /** NYI */
+}
+
+void MoLAttentionLayer::calcDerivative(RunLayerContext &context) { /** NYI */ }
+
```
jijoongmoon, line 133: Even if we are not going to implement the calcGradient for now, how about adding it?

jijoongmoon (approved):
LGTM!

nntrainer/layers/mol_attention_layer.cpp
kparichay: I will add.

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -37,7 +38,8 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
 void MoLAttentionLayer::calcDerivative(RunLayerContext &context) { /** NYI */ }
 
 void MoLAttentionLayer::setProperty(const std::vector<std::string> &values) {
-  /** NYI */
+  auto remain_props = loadProperties(values, mol_props);
```
zhoonit, line 135: Could you also implement export_to as well?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -24,18 +25,103 @@ MoLAttentionLayer::~MoLAttentionLayer() {}
...
+  softmax.setActiFunc(ActivationType::ACT_SOFTMAX);
+  tanh.setActiFunc(ActivationType::ACT_TANH);
+  sigmoid.setActiFunc(ActivationType::ACT_SIGMOID);
+
+  auto unit = std::get<props::Unit>(mol_props).get();
+  auto mol_k = std::get<props::MoL_K>(mol_props).get();
+
+  TensorDim mlp_w_dim = {query_dim.width(), unit};
+  wt_idx[AttentionParams::mlp_w] =
+    context.requestWeight(mlp_w_dim, Tensor::Initializer::XAVIER_UNIFORM,
```
zhoonit, line 58: wonder if the initializers are fixed?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -24,18 +25,103 @@ MoLAttentionLayer::~MoLAttentionLayer() {}
...
+
+  /** @note kappa_hat, beta_hat, alpha_hat are strided */
+  Tensor kappa_hat = mlp_proj_out.getSharedDataTensor({batch, mol_k}, 0, false);
+  Tensor beta_hat =
+    mlp_proj_out.getSharedDataTensor({batch, mol_k}, mol_k, false);
+  Tensor alpha_hat =
+    mlp_proj_out.getSharedDataTensor({batch, mol_k}, mol_k * 2, false);
+
+  Tensor kappa = kappa_hat.apply(static_cast<float (*)(float)>(&std::exp));
+  Tensor beta = beta_hat.apply(static_cast<float (*)(float)>(&std::exp));
```
zhoonit, line 99: I guess we can use std::expf to avoid casting would happen inside function?

nntrainer/layers/acti_func.cpp
```diff
@@ -156,6 +156,11 @@ Tensor &ActiFunc::softmax(Tensor const &t, Tensor &output) {
   float *dp;
   float *rp;
 
+  /** TODO: fix this check */
```
zhoonit, line 159: Could you elaborate on this comment? Is this enable strided softmax?

test/unittest/unittest_nntrainer_models.cpp
```diff
@@ -741,6 +741,15 @@ INI multiout_model(
   }
 );
 
+INI mol_attention(
+  "mol_attention",
+  {nn_base + "batch_size = 3",
+   sgd_base + "learning_rate = 1",
+   I("in0") + input_base + "input_shape = 1:1:256",
+   I("in1") + input_base + "input_shape = 1:84:256",
+   I("in2") + input_base + "input_shape = 1:1:5", // this shape should match mol_k
```
zhoonit, line 750: If shape should match mol_k, can we use input_dim.width() to infer mol_k instead of setting it?

nntrainer/layers/mol_attention_layer.cpp
kparichay: it coming in the new PR.

nntrainer/layers/mol_attention_layer.cpp
kparichay: Its updated to use LayerImpl in the upcoming PR.

nntrainer/layers/mol_attention_layer.cpp
kparichay: Will use expf (not std::expf) in upcoming PR.

test/unittest/unittest_nntrainer_models.cpp
kparichay: It seems like mol_k should be able to interpret, but has been kept as a parameter as per the reference python implementation. If it is always fixed, will remove the property and interpret it from the input dimension later.

nntrainer/layers/acti_func.cpp
kparichay: Yes, it is to enable strided softmax in future. Updated the comment.

test/unittest/unittest_nntrainer_models.cpp
zhoonit: I see, thanks for the explanation :)

zhoonit (approved):
LGTM

PR #1710, opened 2021-11-15
[fixes] Fixes to layer and graph @open sesame 11/19 10:10

kparichay:
- Concat layer has properties which were not being saved. This patch adds
the corresponding support.
- Below bugs have been fixed for the slice realizer:
  - end layers were being added using children which were not set yet
  - the dfs_stack never ignored nodes which had already been visited which
was leading to infinite loop
  - an invalid graph given with loops would be stuck in infinite loop in
the DFS
- Slice realizer was storing the path for each node from the start nodes
to themselves. This can be big, and also comes with a overhead of
storing and copying these paths for each node.
This patch updates the implementation of the dfs to a recursive approach
which maintains the stack of the nodes in the memory which represents
the path till now.

### DIFF
--- a/nntrainer/compiler/slice_realizer.cpp
+++ b/nntrainer/compiler/slice_realizer.cpp
@@ -9,6 +9,7 @@
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
  * @bug No known bugs except for NYI items
  */
+
 #include <layer_node.h>
 #include <slice_realizer.h>
 
@@ -33,19 +34,15 @@ SliceRealizer::realize(const GraphRepresentation &reference) {
       to_be_added(false) {}
     std::shared_ptr<LayerNode> node; /**< set this if not visited */
     bool is_visited;                 /**< set this if visited */
-    bool to_be_added;                   /**< set this if it is to be added */
+    bool to_be_added;                /**< set this if it is to be added */
     std::vector<std::string> children;
-    std::vector<std::string> path;
-    /**< path is the tracing result from start to current node
-      eg) if traversal has started from a -> b -> c -> d.
-      The path has {"a", "b", "c", "d"} */
 
     LayerNode *operator->() { return node.get(); }
   };
 
   /** @note mp has to be ordered map to keep the ordering of the nodes in the
    * graph */
-  std::map<std::string, NodeInfo> mp; /// map point
+  std::unordered_map<std::string, NodeInfo> mp; /// map point
 
   std::transform(
     reference.begin(), reference.end(), std::inserter(mp, mp.end()),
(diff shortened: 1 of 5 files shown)


nntrainer/compiler/slice_realizer.cpp
```diff
@@ -30,20 +30,19 @@ SliceRealizer::realize(const GraphRepresentation &reference) {
...
-      eg) if traversal has started from a -> b -> c -> d.
-      The path has {"a", "b", "c", "d"} */
 
     LayerNode *operator->() { return node.get(); }
   };
 
-  std::unordered_map<std::string, NodeInfo> mp; /// map point
+  /** @note mp has to be ordered map to keep the ordering of the nodes in the
+   * graph */
+  std::map<std::string, NodeInfo> mp; /// map point
```
zhoonit, line 44: Could you elaborate on this? this will be ordered by name of node, which I think wouldn't mean anything.

zhoonit, line 44: Could you elaborate on this? this will be ordered by name of nodes, plus in this stage should not assume the representation is sorted.

nntrainer/compiler/slice_realizer.cpp
```diff
@@ -53,39 +56,38 @@ SliceRealizer::realize(const GraphRepresentation &reference) {
...
-                [&mp](std::shared_ptr<LayerNode> node) {
-                  auto node_name = node->getName();
-                  for (auto &parent : node->getInputLayers()) {
-                    mp.at(parent).children.push_back(node_name);
-                  };
-                });
-
-  GraphRepresentation processed;
+  if (cur_start_layers.empty()) {
+    throw std::runtime_error("No start layer is found, graph has a loop.");
```
zhoonit, line 81: nitpicking: even if start/end layers are not empty, it is possible that there is a loop, wonder if this is considered?

test/unittest/compiler/unittest_realizer.cpp
```diff
@@ -163,10 +163,10 @@ TEST(SliceRealizer, slice_p) {
   std::vector<LayerRepresentation> after = {
     {"fully_connected", {"name=a1"}},
     {"fully_connected", {"name=b1", "input_layers=a1"}},
+    {"fully_connected", {"name=b2", "input_layers=a2"}},
```
zhoonit, line 166: Could you add a negative test this patch is trying to solve as a regression?

nntrainer/compiler/slice_realizer.cpp
```diff
@@ -89,58 +85,64 @@ SliceRealizer::realize(const GraphRepresentation &reference) {
...
 
-    dfs_stack.insert(dfs_stack.end(), children.begin(), children.end());
+  /** run dfs from all the starting layers */
+  for (auto &name : cur_start_layers) {
+    dfs(name);
   }
 
-  GraphRepresentation processed;
+  /** created the subgraph */
+  GraphRepresentation subgraph;
```
zhoonit, line 134: Term looks good

nntrainer/compiler/slice_realizer.cpp
```diff
@@ -89,58 +85,64 @@ SliceRealizer::realize(const GraphRepresentation &reference) {
...
+       * to the subgraph */
+      if (node_info.to_be_added || is_end_node(name)) {
+        std::for_each(dfs_stack.begin(), dfs_stack.end(), update_processed);
+        update_processed(name);
+      }
 
-    auto &path = node_info.path;
-    node_info.is_visited = true;
-    path.push_back(node_info->getName());
+      /** if node is not visited, return */
```
zhoonit, line 114: node is visited?

nntrainer/compiler/slice_realizer.cpp
```diff
@@ -53,39 +56,38 @@ SliceRealizer::realize(const GraphRepresentation &reference) {
...
 
-  auto update_processed = [&processed, &mp](const std::string &name) {
-    auto &node_info = mp.at(name);
-    if (!node_info.is_added) {
-      processed.push_back(node_info.node);
-      node_info.is_added = true;
-    }
-  };
+  if (cur_end_layers.empty()) {
+    throw std::runtime_error("No start layer is found, graph has a loop.");
```
jijoongmoon, line 85: %s/start/end?

nntrainer/compiler/slice_realizer.cpp
kparichay: If the input_layers is not specified, then the connections are automatically added to the previous layer. Hence ordering is required. This is still buggy as the ordering is by name than the order of insertion. Will update.

nntrainer/compiler/slice_realizer.cpp
kparichay: No, that isn't checked. That check should be done outside the slice realizer.

nntrainer/compiler/slice_realizer.cpp
zhoonit: I recommend you to use unordered_map, this is not related if mp is ordered or not.

We rely on the order of GraphRepresentation to figure out this anyway.

nntrainer/compiler/slice_realizer.cpp
kparichay: done.

nntrainer/compiler/slice_realizer.cpp
kparichay: done.

test/unittest/compiler/unittest_realizer.cpp
kparichay: This change is just maintaining the orders of the node as defined in `before`.
This is the limitation of the `graphEqual` - https://github.com/nnstreamer/nntrainer/blob/988a6346d189b65c638fcefe7adf609f090ab700/test/unittest/compiler/compiler_test_util.cpp#L35

zhoonit: Yes, not about this line. But a regression test which this patch resolves.

nntrainer/compiler/slice_realizer.cpp
kparichay: yeah, unordered_map is right but the ordering still has to be maintained.
The GraphRepresentation is a vector of layers where ordering of the layers is part of the graph.
Selecting nodes using DFS can change the ordering.
Corresponding code to ensure the ordering is added.

test/unittest/compiler/unittest_realizer.cpp
kparichay: done.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1731, opened 2021-11-22
[grucell] Implement grucell

lhs8928:
Commit 1:     [bugfix] Bugfix on tensor sum
    
     - Check beta is null when copy the origin data on sum function and added unittest
     - Remove argument beta on outplace sum function
     - Makes bias_hh to zero instead of bias_ih in recurrent models when convert pytorch to nntrainer
     - Fix typos
    
    Self evaluation:
    
Commit 2:     [grucell] Implement grucell
    
     - Support add with non-contiguous tensor
     - Implement grucell which is run only 1 timestep
     - Todo:
       1. Make it more efficient with strided tensor(Reduce copy tensor)
       2. Reduce temporary tensor or do inplace operation
    
    Self evaluation:
    
Commit 3:     [unittest] Implement grucell unittest
    
     - Verify grucell with tensorflow by layer unittest
     - Added grucell model unittest to verify multi unit/timestep, stacked, unroll situation with pytorch
     - Todo: Finds other way without copying when convert gate order of pytorch grucell
    
    Self evaluation:
    

test/input_gen/transLayer_v2.py
```diff
@@ -64,6 +64,37 @@ def transpose_(weight):
...
+        return (weight[0], weight[1].transpose(1, 0))
+
+    # resetgate, inputgate, newgate -> inputgate, resetgate, newgate
+    def reorder_weight(weight_):
+        if (weight_[1].dim() == 2):
+            hidden_size = int(weight_[1].shape[1] / 3)
+        else:
+            hidden_size = int(weight_[1].shape[0] / 3)
+
+        import copy
```
lhs8928, line 83: If there is a better way let me know

nntrainer/tensor/tensor.cpp
```diff
@@ -755,9 +754,9 @@ Tensor Tensor::sum_by_batch() const {
 /**
  * @brief Calculate sum according to the axis.
  */
-Tensor Tensor::sum(unsigned int axis, float alpha, float beta) const {
+Tensor Tensor::sum(unsigned int axis, float alpha) const {
```
zhoonit, line 812: Q. beta is not supported?

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -1637,6 +1637,396 @@ TEST(nntrainer_Tensor, sum_02_p) {
...
+    nntrainer::Tensor result_0_2_2 = input.sum(0, output_0_2_2, 2, 2);
+    nntrainer::Tensor result_1_2_2 = input.sum(1, output_1_2_2, 2, 2);
+    nntrainer::Tensor result_2_2_2 = input.sum(2, output_2_2_2, 2, 2);
+    nntrainer::Tensor result_3_2_2 = input.sum(3, output_3_2_2, 2, 2);
+
+    for (unsigned int i = 0; i < result_0_2_2.batch(); ++i) {
+      for (unsigned int l = 0; l < result_0_2_2.channel(); ++l) {
+        for (unsigned int j = 0; j < result_0_2_2.height(); ++j) {
+          for (unsigned int k = 0; k < result_0_2_2.width(); ++k) {
+            if (ans_0_2_2[i][l][j][k] != result_0_2_2.getValue(i, l, j, k)) {
```
zhoonit, line 1976: this is float so we can't compare like this. 

Tensor has operator== defined so we can use this directly.

test/input_gen/genModelsRecurrent_v2.py
```diff
@@ -38,17 +38,12 @@ def __init__(self, unroll_for=1, num_rnn=1, input_size=1, hidden_size=1):
             ]
         )
         for rnn in self.rnns:
-            rnn.bias_ih.data.fill_(0.0)
-            rnn.bias_ih.requires_grad=False
+            rnn.bias_hh.data.fill_(0.0)
```
zhoonit, line 41: nice catch :)

nntrainer/layers/grucell.cpp
```diff
@@ -0,0 +1,414 @@
...
+  weight_hh,
+  bias_h,
+  hidden_state,
+  zrg,
+  dropout_mask
+};
+
+#define ENABLE_BIAS_IH 1
+// Todo: enable bias_hh
+#define ENABLE_BIAS_HH 0
```
zhoonit, line 53: shouldn't this be a property?

test/input_gen/transLayer_v2.py
zhoonit: you would want to use np.hsplit + np.hstack  (torch should have the same method defined as well)

test/input_gen/genModelsRecurrent_v2.py
```diff
@@ -38,17 +38,12 @@ def __init__(self, unroll_for=1, num_rnn=1, input_size=1, hidden_size=1):
...
+            rnn.bias_hh.requires_grad=False
```
kparichay, line 42: good job

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -1637,6 +1637,396 @@ TEST(nntrainer_Tensor, sum_02_p) {
...
+    nntrainer::Tensor result_0_1_0 = input.sum(0, 1);
+    nntrainer::Tensor result_1_1_0 = input.sum(1, 1);
+    nntrainer::Tensor result_2_1_0 = input.sum(2, 1);
+    nntrainer::Tensor result_3_1_0 = input.sum(3, 1);
+
+    for (unsigned int i = 0; i < result_0_1_0.batch(); ++i) {
+      for (unsigned int l = 0; l < result_0_1_0.channel(); ++l) {
+        for (unsigned int j = 0; j < result_0_1_0.height(); ++j) {
+          for (unsigned int k = 0; k < result_0_1_0.width(); ++k) {
+            if (ans_0_1_0[i][l][j][k] != result_0_1_0.getValue(i, l, j, k)) {
```
kparichay, line 1680: recommend to wrap `ans_*_1_0` with a tensor and directly compare tensors.
you can use a 4D vector of floats instead of 4D float array or wrap array with vector or add constructor in Tensor for 4D array.

nntrainer/tensor/tensor.cpp
kparichay: can you please elaborate what is the bug here?

nntrainer/layers/grucell.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+#define NUM_GATE 3
```
kparichay, line 55: consider using `static constexpr unsigned int NUM_GATE = 3`.

nntrainer/layers/grucell.cpp
```diff
@@ -0,0 +1,414 @@
...
+    zrg_gate.getSharedDataTensor({batch_size, unit}, 2 * unit, false);
+
+  Tensor weight_hh_zr;
+  Tensor weight_hh_g;
+  weight_hh_zr.copy_with_stride(
+    weight_hh.getSharedDataTensor({1, 1, unit, unit * 2}, 0, false));
+  weight_hh_g.copy_with_stride(
+    weight_hh.getSharedDataTensor({1, 1, unit, unit}, unit * 2, false));
+
+  zr_gate.add_i_strided(prev_hidden_state.dot(weight_hh_zr));
```
kparichay, line 209: consider wrapping this statement with `if (!timestep)` than creating a zero valued tensor.

nntrainer/layers/grucell.cpp
```diff
@@ -0,0 +1,414 @@
...
+  hidden_states_derivatives.reshape({max_timestep, 1, batch_size, unit});
+  Tensor hidden_state_derivative =
+    hidden_states_derivatives.getBatchSlice(timestep, 1);
+  hidden_state_derivative.reshape(incoming_derivative.getDim());
+  if (timestep + 1 == max_timestep) {
+    djdweight_xh.setZero();
+    djdweight_hh.setZero();
+    djdbias_ih.setZero();
+    hidden_state_derivative.copyData(incoming_derivative);
+    zrg_gates_derivatives.setZero();
```
kparichay, line 312: I think this is redundant. Please check.

test/unittest/layers/unittest_layers_grucell.cpp
```diff
@@ -0,0 +1,35 @@
...
+                        ::testing::Values(semantic_grucell));
+
+auto grucell_single_step = LayerGoldenTestParamType(
+  nntrainer::createLayer<nntrainer::GRUCellLayer>,
+  {"unit=5", "max_timestep=1", "timestep=0"}, "3:1:1:7",
+  "gru_single_step.nnlayergolden", LayerGoldenTestParamOptions::DEFAULT);
+
+INSTANTIATE_TEST_CASE_P(GRUCell, LayerGoldenTest,
+                        ::testing::Values(grucell_single_step
+                                          //   grucell_single_step_seq,
```
kparichay, line 34: is there another test to be added?

test/input_gen/transLayer_v2.py
```diff
@@ -64,6 +64,37 @@ def transpose_(weight):
     new_params = [transpose_(params[0]), transpose_(params[1]), bias]
     yield from new_params
 
+@register_for_((torch.nn.GRUCell))
+def gru_translate(model):
+    params = [(name, tensor.detach()) for name, tensor in model.named_parameters()]
+    bias = ("bias", params[2][1] + params[3][1])
+
+    # hidden, input -> input, hidden
```
kparichay, line 72: the code says the equation: `hidden, input -> hidden, input^T`. Can you please check

test/input_gen/transLayer_v2.py
lhs8928: Thanks 👍

test/unittest/unittest_nntrainer_tensor.cpp
lhs8928: Done

test/unittest/unittest_nntrainer_tensor.cpp
lhs8928: Done

nntrainer/layers/grucell.cpp
lhs8928: Eventually we will support bias_hh and then this will be a property.
But for now just think it as a comment.

nntrainer/layers/grucell.cpp
lhs8928: Make it as class member variable as you recommend

nntrainer/layers/grucell.cpp
lhs8928: Good catch 👍

nntrainer/layers/grucell.cpp
lhs8928: Good catch 👍

test/unittest/layers/unittest_layers_grucell.cpp
lhs8928: No it was my mistake

test/input_gen/transLayer_v2.py
lhs8928: No.
In here the weight[0] is the name and weight[1] is the tensor which has the shape [hidden, input]
I wrap it with the square bracket to clarify it as a dimension.

nntrainer/tensor/tensor.cpp
lhs8928: In sum function the argument beta was supposed to amplify the returned tensor by beta.
But in this case the returned tensor is newly made so the beta always should be zero so beta does not needed.
Correct me if I'm wrong.

nntrainer/layers/grucell.cpp
zhoonit: Alright

kparichay (approved):
LGTM

zhoonit (approved):
LGTM :) nice work!

nntrainer/tensor/tensor.cpp
kparichay: Right.

nntrainer/layers/grucell.cpp
```diff
@@ -0,0 +1,413 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 hyeonseok lee <hs89.lee@samsung.com>
```
jijoongmoon, line 3: 21?

jijoongmoon (approved):
LGTM

PR #1733, opened 2021-11-22
[trivial] Add note on the freq

zhoonit:

<details><summary>[trivial] Add note on the freq</summary><br />

This patch add note on the frequency map

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/compiler/multiout_realizer.cpp
+++ b/nntrainer/compiler/multiout_realizer.cpp
@@ -56,6 +56,12 @@ MultioutRealizer::realize(const GraphRepresentation &reference) {
   for (auto &[con_name, freq] : freq_map) {
     props::InputConnection con;
     from_string(con_name, con);
+
+    /// @note freq < 1 should never happen as the map entry is not created.
+    /// but if it happens multiout realizer is not interested in checking if it
+    /// is a dangled or actually an output. So there is no assurance done at
+    /// this point. Some other class must check if the given graph is formed in
+    /// a correct way.
     if (freq <= 1) {
       continue;
     }


nntrainer/compiler/remap_realizer.h
```diff
@@ -28,7 +28,28 @@ namespace nntrainer {
...
-  RemapRealizer(std::function<void(std::string &)> remap_function);
+  /**
+   * @brief Construct a new Remap Realizer object (connection mode)
+   *
+   * @param remap_connection_function remap connection function, with this
+   * constructor, only connections will be remapped eg) if you are inserting a
+   * new layer node in between, only connections need remapping
+   */
+  RemapRealizer(
+    std::function<void(std::string & /**< identifier */,
```
kparichay, line 39: remap_function was remapping the given identifier and modify the string.
Just to clarify, remap_connection_function will modify the index and the string both right?

zhoonit: yes :)

jijoongmoon (approved):
LGTM! Great Work!

nntrainer/compiler/multiout_realizer.cpp
```diff
@@ -9,16 +9,99 @@
...
 namespace nntrainer {
 MultioutRealizer::~MultioutRealizer() {}
 
 GraphRepresentation
 MultioutRealizer::realize(const GraphRepresentation &reference) {
+  GraphRepresentation processed(reference.begin(), reference.end());
+
+  std::unordered_multiset<std::string> freq_map;
+  std::unordered_set<std::string> node_names;
+  std::unordered_set<std::string> connection_names;
```
kparichay, line 33: consider using `std::unordered_map<std::string, unsigned int>` as a combined struct for `freq_map` and `connection_names`.

nntrainer/compiler/multiout_realizer.cpp
```diff
@@ -9,16 +9,99 @@
...
+  /// 2. for each connection names, if a connection is referenced multiple
+  /// times, create multioutput node and remap to multi output node index
+  std::unordered_multimap<std::string /**< original id */,
+                          std::shared_ptr<LayerNode> /**< created node */>
+    multiout_nodes;
+
+  for (auto &con_name : connection_names) {
+    props::InputConnection con;
+    from_string(con_name, con);
+    if (freq_map.count(con_name) <= 1) {
```
kparichay, line 60: `== 1`? < 1 should be error?

nntrainer/compiler/multiout_realizer.cpp
```diff
@@ -9,16 +9,99 @@
...
+      continue;
+    }
+
+    std::string id = con.get().getName();
+    auto idx = con.get().getIndex();
+
+    std::stringstream ss;
+    /// {connection_name}/generated_out_{index}
+    ss << id << "/generated_out_" << idx;
```
kparichay, line 69: `/s/generated_out_/multi_out_/`

nntrainer/compiler/multiout_realizer.cpp
```diff
@@ -9,16 +9,99 @@
...
+    while (node_names.count(ss.str()) != 0) {
```
kparichay, line 70: layer_name will be made unique if the name is already taken. IMO this can be skipped.

nntrainer/compiler/multiout_realizer.cpp
```diff
@@ -9,16 +9,99 @@
...
+        node->getInputConnectionName(i), node->getInputConnectionIndex(i)));
+      auto uniq_name = to_string(c);
+      connection_names.emplace(uniq_name);
+      freq_map.emplace(uniq_name);
+    }
+  }
+
+  /// 2. for each connection names, if a connection is referenced multiple
+  /// times, create multioutput node and remap to multi output node index
+  std::unordered_multimap<std::string /**< original id */,
```
kparichay, line 53: multimap allows multiple entries of equivalent keys, but L70 ensures that the key is unique. So, this can be unordered_map?

Applications/ProductRatings/res/product_ratings_model.ini
```diff
@@ -22,13 +22,13 @@ Type = split
 axis = 3 # split by width
 
 [user_embed]
-input_layers = split
+input_layers = split(0)
```
kparichay, line 25: please add a todo here to remove the split layer as we now support multiple inputs.

nntrainer/compiler/multiout_realizer.cpp
zhoonit: Yeah, that's more neat. Done

nntrainer/compiler/multiout_realizer.cpp
zhoonit: con_name == 0 means it's input, output or dangled. Multiout realizer cannot tell if it's error or intended.

nntrainer/compiler/multiout_realizer.cpp
zhoonit: This is checking if `{connection_name}/generated_out_{index}` is already taken from the given representation. At this stage  we need to ensure the name is unique else remapping will be wrong.

consider below ini.

```ini
[fc]
type=input
input_shape=3

[fc/generated_out_0]
type=input
input_shape=1

[a]
input_layers=fc

[b]
input_layers=fc
```
will be turned to below without this check.
```ini
[fc]
type=input
input_shape=3

[fc/generated_out_0]
type=input
input_shape=1

[fc/generated_out_0] 
type=multiout
input_layers=fc(0)

[a]
input_layers=fc/generated_out_0

[b]
input_layers=fc/generated_out_0
```

nntrainer/compiler/multiout_realizer.cpp
zhoonit: there will be 1:n mapping between ${id(layer name)}:${multiout_name}

consider below ini

```
[lstm]
type=lstmcell

[a]
input_layers=lstm(0)

[b]
input_layers=lstm(0)

[c]
input_layers=lstm(1)

[c]
input_layers=lstm(1)
```

key lstm will have lstm/generated_out_0, and lstm/generated_out_1

Applications/ProductRatings/res/product_ratings_model.ini
zhoonit: Yeah, done

nntrainer/compiler/multiout_realizer.cpp
kparichay: Let's just add a note here that `<1` should not happen and someone must check for this and report error.

nntrainer/compiler/multiout_realizer.cpp
kparichay: yeah, its better to have same semantics everywhere. So, renaming here is fine then.

FYI, in this case, the list of layers is not representative of the ini. 
In ini, two layers with same name gets interpreted by the ini file format as redefinition of the same layer.
But in the layer list, as layer is already defined, the layer name will automatically be updated.

zhoonit: Yeah, right.

What we can guarantee is the exact ini is always interpreted the same way

nntrainer/compiler/multiout_realizer.cpp
zhoonit: Done.

```cpp
    /// @note freq < 1 should never happen as the map entry is not created.
    /// but if it happens multiout realizer is not interested in checking if it
    /// is a dangled or actually an output. So there is no assurance done at
    /// this point. Some other class must check if the given graph is formed in
    /// a correct way.
```

zhoonit: note that #1734 which was ahead of this PR was merged. This PR will contain only review changes, and won't be referenced at other PR's

kparichay: thanks.

nntrainer/compiler/multiout_realizer.cpp
kparichay: This is fine, I thought the the struct is `multimap` of `string : [node]` than `string : node`.
Thanks for the example.

kparichay (approved):
LGTM

PR #1737, opened 2021-11-24
[layer] Embedding layer new property + bug fix @open sesame 11/29 16:00

### LINKED ISSUE #1729: Embedding padding concept
Embedding layer introduces the concept of padding where the input index 0 is reserved for not filling the output. I have two doubts:
1. If filling the output is skipped, this results in uninitialized values. Is this intended?
2. Currently 0 is always assumed to be not filling the output but should be enabled/disabled with boolean. Further, this padding_idx should be allowed to be set as a property.

kparichay:
- Embedding layer should assume that the input data is written in int
format than in float data. Reading data as float can lead to wrong
values when typecasting to int if std::lround() is not properly used.
Further, as other frameworks require using integer data for embedding,
its best if follow the same input formats.
- Embedding layer support zero mask index which if set will pass zero
values for that particular index in the forward propagation and will not
be updated with gradient.
Further minor updates to tensor related to getting tensor values.

resolves #1729

### DIFF
--- a/Applications/ProductRatings/jni/main.cpp
+++ b/Applications/ProductRatings/jni/main.cpp
@@ -82,10 +82,11 @@ bool getData(std::ifstream &F, float *input, float *label, unsigned int id) {
     return false;
 
   std::istringstream buffer(temp);
-  float x;
+  uint *input_int = (uint *)input;
+  uint x;
   for (unsigned int j = 0; j < feature_size; ++j) {
     buffer >> x;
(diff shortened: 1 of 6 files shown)


jijoongmoon (approved):
LGTM!

zhoonit (approved):
LGTM :)

zhoonit:
Please rebase to see if ci pass after #1747  :)

PR #1739, opened 2021-11-24
[weights] Support clip gradient by global norm

### LINKED ISSUE #1714: Support gradient clipping
Support gradient clipping by norm per layer

kparichay:
This patch supports clipping of gradient by global norm:
- Add the property clip gradient by norm and propagate the property to
each weight by the layer.
ClipGradByNorm property will clip the gradients by the global norm of
the weights with this value set. Each layer's weight can have different
scaling clip value.
- Extend the execution order for the gradients if it used for clipping by
global norm. The gradient is extended by the max execution order where
the norm of the gradients will be calculated and used to update and
apply the gradients.
This is done while requesting the gradient itself at the first time as
the layer properties are finalized by then.
- skip applying gradients if it is to be clipped by the global norm.
- apply the clipped gradients:
  - Calculate the global norm for the gradients which needs to be clipped
  - Clip the gradients with the calculated global norm
  - apply the clipped gradients

~TODO: add related unittest~ Unittests added

Resolves #1714 


nntrainer/graph/network_graph.cpp
```diff
@@ -494,6 +511,47 @@ void NetworkGraph::backwarding(
...
+
+  /** perform clipping of the gradients by global norm if any */
+  auto clip_weights = tensor_manager->getWeights([](Weight *w) {
+    return w->hasGradient() && w->isGradientLastAccess() &&
+           w->isGradientClipByGlobalNorm();
+  });
+  /** calculate the global norm */
+  unsigned int global_norm = 0;
+  for (auto w : clip_weights) {
+    global_norm += std::pow(w->getGradientNorm(), 2);
```
zhoonit, line 523: I recommend to make a tensor to store the vector to calculate global_norm
and global_norm = norms.l2norm()

There are two major factors behind.

1. Accumulation can go out of float range easily, so this might be unstable.
2. Speed (it was observed that this is significantly slower then Tensor::l2norm from the eariler version of nntrainer)

nntrainer/graph/network_graph.cpp
```diff
@@ -494,6 +511,47 @@ void NetworkGraph::backwarding(
     backwarding_op(ln, iteration);
     END_PROFILE(profile_keys.at(ln->getType()));
   }
+
+  /** perform clipping of the gradients by global norm if any */
+  auto clip_weights = tensor_manager->getWeights([](Weight *w) {
+    return w->hasGradient() && w->isGradientLastAccess() &&
+           w->isGradientClipByGlobalNorm();
+  });
```
zhoonit, line 519: nitpicking: can we cache this vector as a member to avoid this calculation every iteration?

nntrainer/graph/network_graph.h
```diff
@@ -147,14 +148,15 @@ class NetworkGraph {
...
    * @note if it is not the last of the gradient access, this is noop
+   * @note if the gradient is to be clipped by norm, this is noop
    *
    * @param node node to try apply gradient
    * @param apply_func apply function
    */
-  void applyGradientsOnLastAccess(LayerNode *node,
-                                  std::function<void(Weight &)> apply_func);
+  static void applyGradients(LayerNode *node,
+                             std::function<void(Weight &)> apply_func);
```
zhoonit, line 159: nitpicking: can we change apply_func to reference?

nntrainer/tensor/manager.cpp
```diff
@@ -570,14 +578,16 @@ std::vector<Tensor *> Manager::requestWeightOptimizerVariables(
   return ret;
 }
 
-std::vector<Weight *> Manager::getWeights() {
-  std::vector<Weight *> all_weights;
+std::vector<Weight *>
+Manager::getWeights(const std::function<bool(Weight *)> &condition) {
+  std::vector<Weight *> conditional_weights;
```
zhoonit, line 583: if (condition == nullptr) return allWeights()?

nntrainer/tensor/manager.h
```diff
@@ -241,11 +245,14 @@ class Manager {
...
+   * @brief     Get all the weights which match the above condition
    *
-   * @return    return all the weights
+   * @return    return the weights with satisfying the above condition
    */
-  std::vector<Weight *> getWeights();
+  std::vector<Weight *>
+  getWeights(const std::function<bool(Weight *)> &condition = [](Weight *) {
+    return true;
+  });
```
zhoonit, line 255:
```suggestion
  getWeights(const std::function<bool(Weight *)> &condition = nullptr;
```

nntrainer/graph/network_graph.cpp
kparichay: fair enough, I will update.

nntrainer/graph/network_graph.cpp
kparichay: will add.

nntrainer/graph/network_graph.h
kparichay: updated.

nntrainer/tensor/manager.h
kparichay: Return true was for expressiveness but nullptr works too.

zhoonit: I think this is just minor performance measure. but we will need to handle nullptr case anyway inside the implemenation :)

zhoonit (approved):
LGTM :)

nntrainer/tensor/manager.h
kparichay: nullptr support is added :)

kparichay:
V2: 
gradient clipping works by adding extending the execution order of the
gradients to the last node, where the global norm is calculated and the
gradients are clipped and applied.
However, weight sharing use of gradients also relies on the last access
of the gradient and gradient clipping disturbs the balance of gradient
last access.
As a quick fix, if gradient clip is enabled, the last access is replaced
with second to last access.

A better way would be for clipping to be layer, and then last access by
clipping layer would be a valid access and balance to the system can be
maintained.

Unittests for gradient clipping is added with and without weight
sharing.

nntrainer/graph/network_graph.cpp
```diff
@@ -1039,8 +1039,21 @@ int NetworkGraph::initialize(
                                           first_grad_access)) {
           rc.getWeightObject(i).setAsGradientFirstAccess();
         }
+        /**
+         * if the gradient is to be clipped by global norm, then the last access
+         * is by clipping itself. However, as clipping is not a layer and does
+         * not contain any weights, such weights never get assigned
+         * gradient_last_access. This is a quick hotfix.
```
zhoonit, line 1046: Would've been hard to pin point. Doesn't adding max execorder solve this prob?

test/input_gen/recorder_v2.py
```diff
@@ -84,6 +84,10 @@ def record_iteration(write_fn):
 
         optimizer.zero_grad()
         loss.backward()
+        if clip:
+            print([(name, param.grad, torch.norm(param.grad)) for name, param in model.named_parameters()])
+            norm = torch.nn.utils.clip_grad_norm_(model.parameters(), 0.0001)
+            print(norm)
```
zhoonit, line 90: maybe remove print?

zhoonit (approved):
LGTM :)

jijoongmoon (approved):
Great! LGTM! Even if we need more memory to save the gradient, but it is enabled in a pretty simple way though!

test/input_gen/recorder_v2.py
kparichay: removed prints

PR #1743, opened 2021-11-25
[fixes] Few minor fixes

kparichay:
- Add shape mismatch check in reshape layer.
- Permute layer backward bug in the calculation of the reverse
direction of the permute is fixed.
- Add missing layer constructor for permute.

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -376,6 +376,14 @@ Attention(const std::vector<std::string> &properties = {}) {
 }
 
 /**
+ * @brief Helper function to create Permute Layer
+ */
+inline std::unique_ptr<Layer>
+Permute(const std::vector<std::string> &properties = {}) {
+  return createLayer(LayerType::LAYER_PERMUTE, properties);
+}
+
+/**
  * @brief Helper function to create activation layer
  */
 inline std::unique_ptr<Layer>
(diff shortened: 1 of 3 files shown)


nntrainer/layers/permute_layer.cpp
```diff
@@ -57,7 +57,7 @@ void PermuteLayer::finalize(InitLayerContext &context) {
 
     /*** @todo deprecate this */
     direction_str = buildTrasposeString(direction);
-    rdirection_str = buildTrasposeString(direction);
+    rdirection_str = buildTrasposeString(reverse_direction);
```
zhoonit, line 60: wonder why it did not have any problem.

zhoonit (approved):
LGTM

nntrainer/layers/permute_layer.cpp
kparichay: it doesn't have unittest, so never got caught I guess.

api/ccapi/include/layer.h
```diff
@@ -375,6 +375,14 @@ Attention(const std::vector<std::string> &properties = {}) {
...
 }
 
+/**
+ * @brief Helper function to create Permute Layer
+ */
+inline std::unique_ptr<Layer>
+Permute(const std::vector<std::string> &properties = {}) {
+  return createLayer(LayerType::LAYER_PERMUTE, properties);
+}
+
```
jijoongmoon, line 385: nice catch!

jijoongmoon (approved):
LGTM

Issue #1721, opened 2021-11-18
[proposal] derivative support for all tensor operations


kparichay:
Layer implementation requires implementation of forwarding as well as backwarding by the layer developer. Implementation of forwarding is inevitable but backwarding could be simplified with the support of derivative support for all tensor operations.

Consider the dot operation interface.
```c++
Tensor &dot(Tensor const &m, Tensor &output, bool trans = false,
              bool trans_m = false, float beta = 0.0f) const;
```

A dot_derivative can be provided by the tensor itself to compute the derivative with respect to either of the two inputs, given the derivative of the output. Consider the below two interfaces:
```c++
/**
 * @brief compute the derivative of this in the current tensor
 * @param m same as given to the dot()
 * @param output_deriv the derivative of the output 
 * @param[in] trans same as given to the dot()
 * @param[in] trans_m same as given to the dot()
 * @param[in] beta same as given to the dot()
 * @note This will compute the derivative in-place and will overwrite existing data in the tensor
 * @todo will have to see if beta effects this computation
 */
Tensor &dot_deriv_wrt_1(Tensor const &m, Tensor const &output_deriv, bool trans = false,
              bool trans_m = false, float beta = 0.0f);

/**
 * @brief compute the derivative wrt m in the m tensor
 * @param m_deriv tensor where derivative wrt m will be stored
 * @param output_deriv the derivative of the output 
 * @param[in] trans same as given to the dot()
 * @param[in] trans_m same as given to the dot()
 * @param[in] beta same as given to the dot()
 * @note The caller tensor must be the same tensor as the one which called the dot() product. 
 */
Tensor &dot_deriv_wrt_2(Tensor &m, Tensor const &output, bool trans = false,
              bool trans_m = false, float beta = 0.0f) const;
```

Given the above two interfaces, the fc layer's forward, calcDerivative and calcGradient looks like below:
```diff

void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
  Tensor &weight = context.getWeight(weight_idx[FCParams::weight]);
  Tensor &bias = context.getWeight(weight_idx[FCParams::bias]);

  Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
  Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);

  input_.dot(weight, hidden_);
  hidden_.add_i(bias);
}

void FullyConnectedLayer::calcDerivative(RunLayerContext &context) {
[...]

Issue #1614, opened 2021-10-08
Support MoL attention layer


kparichay:
Support Mixture of Logistics attention layer - refer https://arxiv.org/pdf/2009.02725.pdf

kparichay:
Additionally refer to https://arxiv.org/pdf/1906.01083.pdf and https://arxiv.org/pdf/1701.05517.pdf for further details.

PR #1750, opened 2021-11-29
[layer] Support MoL backwarding

kparichay:
- Add derivatives for the dot operation as an easier interface to
calculate derivative for a dot operation used in the forward for both
the inputs.
Add the corresponding interface for dot_batched as well.
- Use the dot derivative operation in both fc and attention layer.
- This patch adds bug fix for copy with stride where both the source and
destination tensors are now allowed to be strided tensors.
- Bug fix to work with fc layer with disabled bias.
- Prepare for multiple inheritance for layers by marking public virtual of
the base class.
- Add MoL attention layer backwarding, exportTo, setbatch member
functions.
- Add MoL layer golden test.

See Also #1721
See Also #1614


kparichay:
Note: MoL backwarding layer has not been optimized to minimize memory usage yet. This implementation focuses on correctness.

nntrainer/tensor/tensor.cpp
```diff
@@ -978,6 +978,68 @@ Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
...
 }
+/**
+ * @brief compute the derivative of this in the current tensor
+ * @todo will have to see if beta effects this computation
+ */
+Tensor &Tensor::dot_deriv_wrt_1(Tensor const &m, Tensor const &output_deriv,
+                                bool trans, bool trans_m, float beta) {
+  bool deriv_trans_m = true;
+  bool deriv_trans = false;
+  /** @todo handle all cases of trans and trans_m */
```
zhoonit, line 989: if this is not supported, how about let it throw to prevent unintended use.

I recommend to add tensor test case for all. If this is wrong, we will have many time to figure out.

nntrainer/tensor/tensor.cpp
```diff
@@ -978,6 +978,68 @@ Tensor Tensor::dot(Tensor const &m, bool trans, bool trans_m) const {
 
   return output;
 }
+/**
+ * @brief compute the derivative of this in the current tensor
+ * @todo will have to see if beta effects this computation
+ */
+Tensor &Tensor::dot_deriv_wrt_1(Tensor const &m, Tensor const &output_deriv,
```
zhoonit, line 985: following naming scheme of camel case? (and for others in this commit)

nntrainer/layers/fc_layer.cpp
```diff
@@ -103,7 +103,7 @@ void FullyConnectedLayer::calcDerivative(RunLayerContext &context) {
   Tensor &derivative_ = context.getIncomingDerivative(SINGLE_INOUT_IDX);
   Tensor &ret_ = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
 
-  derivative_.dot(weight, ret_, false, true);
+  ret_.dot_deriv_wrt_1(weight, derivative_, false, false);
```
zhoonit, line 106: could you rename `ret_` to `d_input_` or sort for clarity?

nntrainer/layers/mol_attention_layer.h
```diff
@@ -16,14 +16,15 @@
...
+#include <layer_impl.h>
 
 namespace nntrainer {
 
 /**
  * @class   MoL Attention Layer
  * @brief   Mixture of Logistics Attention Layer
  */
-class MoLAttentionLayer : public AttentionLayer {
+class MoLAttentionLayer : public AttentionLayer, public LayerImpl {
```
zhoonit, line 27: There seems no codebits in this class using AttentionLayer methods, could you explain why trying to enable multiple inheritance?

I think we shouldn't enable multiple inheritance for non-interface classes unless we absolutely need to

kparichay:
V2:
[layer] perform layer context check on first forward
Enabled layer context check on the first forward itself,
which revealed a bug in forward which was earlier being shown in the
calcGradient in mol attention layer.
Added the corresponding fix for mol attention layer.
Added corresponding fixes to other locations as well.

nntrainer/layers/mol_attention_layer.h
kparichay: multiple inheritance will be used once common code bits are removed from mol attention layer code and reused from attention layer.

nntrainer/layers/fc_layer.cpp
kparichay: will do.

nntrainer/tensor/tensor.cpp
kparichay: Over time, the naming scheme has mixed up camel case (strided, and now this).
I will add an issue to fix the naming scheme for all the cases which is out of scope for this PR.

nntrainer/layers/mol_attention_layer.h
zhoonit: In that case, we can favor composition, or higher order class. My concern is making LayerImpl virtual, where we are opening a hatch all kinds of multiple inheritance will be possible.

zhoonit: eg1) below will not work without try catch.

```
 AttentionLayer::setProperty(remain_props);
 LayerImpl::setProperty(remain_props);
 exporter.saveResult(mol_props, method, this);
```

eg2) attention layer(or other potential) starts to inherit LayerImpl, this will be wrong

```
 AttentionLayer::exportTo(exporter, method);
 LayerImpl::exportTo(exporter, method);
 exporter.saveResult(mol_props, method, this);
```

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -72,34 +137,55 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
 
   const TensorDim &input_dim = query.getDim();
   unsigned int batch = input_dim.batch();
   auto mol_k = std::get<props::MoL_K>(mol_props).get();
 
-  Tensor mlp_out = query.dot(mlp_w);
+  /** reset helper state */
+  helper_exec = false;
+
+  fc_out = query.dot(fc_w);
```
zhoonit: query.dot(fc_w, fc_out) to maintain reference?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -72,34 +137,55 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+  fc_out.add_i(fc_bias);
 
-  Tensor mlp_tanh;
-  tanh.run_fn(mlp_out, mlp_tanh);
+  tanh.run_fn(fc_out, fc_tanh);
 
-  Tensor mlp_proj_out = mlp_tanh.dot(mlp_proj_w);
+  fc_proj_out = fc_tanh.dot(fc_proj_w);
```
zhoonit: same here?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -72,34 +137,55 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
-  Tensor kappa = kappa_hat.apply(static_cast<float (*)(float)>(&std::exp));
-  Tensor beta = beta_hat.apply(static_cast<float (*)(float)>(&std::exp));
+  kappa_src.apply_i(&expf);
+  beta_src.apply_i(&expf);
+  Tensor kappa = kappa_src;
+  Tensor beta = beta_src;
 
   Tensor alpha;
-  softmax.run_fn(alpha_hat, alpha);
+  softmax.run_fn(alpha_src, alpha);
```
zhoonit, line 181: Doesn't this throw considering below?
```
Tensor &ActiFunc::softmax(Tensor const &t, Tensor &output) {
  /**
   * shiftx_logit = logit - max_batch(logit)
   * softmax = exp(shiftx_logit) / (sum(exp(shiftx_logit)))
   *
   * @note softmax is applied on the last dimension
   */
  unsigned int fixed_dim = t.getDim().getDataLen() / t.width();
  float *dp;
  float *rp;

  /** TODO: support strided operations */
  if (t.size() == output.size() && t.getStrides() != output.getStrides())
    throw std::invalid_argument(
      "Softmax does not support operating on strided tensors");      ```

File: nntrainer/layers/mol_attention_layer.cpp [hunk 8]
```diff
@@ -72,34 +137,55 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+    fc_proj_out.getSharedDataTensor({batch, 1, 1, mol_k}, mol_k, false));
+  alpha_src.copy_with_stride(
+    fc_proj_out.getSharedDataTensor({batch, 1, 1, mol_k}, mol_k * 2, false));
 
-  Tensor kappa = kappa_hat.apply(static_cast<float (*)(float)>(&std::exp));
-  Tensor beta = beta_hat.apply(static_cast<float (*)(float)>(&std::exp));
+  kappa_src.apply_i(&expf);
+  beta_src.apply_i(&expf);
+  Tensor kappa = kappa_src;
+  Tensor beta = beta_src;
```
zhoonit, line 178: out of curiosity, wonder why we need this copy?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -72,34 +137,55 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+  Tensor &u_neg_div = context.getTensor(wt_idx[AttentionParams::u_neg_div]);
+  Tensor &u_pos_div = context.getTensor(wt_idx[AttentionParams::u_pos_div]);
 
   const TensorDim &input_dim = query.getDim();
   unsigned int batch = input_dim.batch();
   auto mol_k = std::get<props::MoL_K>(mol_props).get();
 
-  Tensor mlp_out = query.dot(mlp_w);
+  /** reset helper state */
+  helper_exec = false;
```
zhoonit, line 158: what does this do?

nntrainer/layers/layer_context.cpp
```diff
@@ -13,6 +13,7 @@
 #include "nntrainer_error.h"
 #include <functional>
 
+#include <iostream>
```
zhoonit, line 16: unnecessary header?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -116,19 +202,159 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+  alpha.copy_with_stride(
+    fc_proj_out.getSharedDataTensor({batch, 1, 1, mol_k}, mol_k * 2, false));
+
+  Tensor dscores = Tensor(TensorDim({value.batch(), 1, 1, value.height()}));
+  dscores.dot_batched_deriv_wrt_1(value, derivative);
+  dscores.reshape(TensorDim({scores.batch(), 1, scores.width(), 1}));
+
+  Tensor dprob_scaled = Tensor(TensorDim({batch, 1, value.height(), mol_k}));
+  dprob_scaled.setZero();
+  dprob_scaled.add_i(dscores);
```
zhoonit, line 258: Throwing some list for double checking
1. scores.batch() == batch ?
2. scores.width() == values.height()?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -116,19 +202,159 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+
+  Tensor dprob_left = dprob;
+  Tensor dprob_right = dprob.multiply(-1);
+
+  Tensor beta_eps = beta.add(1e-8f);
+  Tensor du_neg_div, du_pos_div;
+  sigmoid.run_prime_fn(prob_right, du_neg_div, dprob_right);
+  Tensor du_neg_m = du_neg_div.divide(beta_eps);
+  Tensor dm_neg = du_neg_m.multiply(-1).sum(2);
+  Tensor dbeta_eps_neg = du_neg_m.multiply(u_neg_div).multiply(-1).sum(2);
```
zhoonit, line 271: considering) u_neg_div = u_neg_m/beta_eps

dbeta_eps = -d_u_neg_div * u_neg_m / (beta_eps ^ 2) ?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -116,19 +202,159 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+
+  sigmoid.run_prime_fn(prob_left, du_pos_div, dprob_left);
+  Tensor du_pos_m = du_pos_div.divide(beta_eps);
+  Tensor dm_pos = du_pos_m.multiply(-1).sum(2);
+  Tensor dbeta_eps_pos = du_pos_m.multiply(u_pos_div).multiply(-1).sum(2);
```
zhoonit, line 276: same here??

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -116,19 +202,159 @@ void MoLAttentionLayer::forwarding(RunLayerContext &context, bool training) {
...
+  Tensor &derivative = context.getIncomingDerivative(SINGLE_INOUT_IDX);
+
+  Tensor &fc_proj_out = context.getTensor(wt_idx[AttentionParams::fc_proj_out]);
+  Tensor &dfc_proj_out =
+    context.getTensor(wt_idx[AttentionParams::fc_proj_out]);
+  Tensor &scores = context.getTensor(wt_idx[AttentionParams::scores]);
+  Tensor &prob = context.getTensor(wt_idx[AttentionParams::prob]);
+  Tensor &prob_left = context.getTensor(wt_idx[AttentionParams::prob_left]);
+  Tensor &prob_right = context.getTensor(wt_idx[AttentionParams::prob_right]);
+  Tensor &u_neg_div = context.getTensor(wt_idx[AttentionParams::u_neg_div]);
```
zhoonit, line 237: how about make this trainable tensor and use TensorGrad?

nntrainer/layers/mol_attention_layer.cpp
```diff
@@ -137,7 +363,21 @@ void MoLAttentionLayer::setProperty(const std::vector<std::string> &values) {
...
+  context.updateTensor(wt_idx[AttentionParams::prob_left], batch);
+  context.updateTensor(wt_idx[AttentionParams::prob_right], batch);
+  context.updateTensor(wt_idx[AttentionParams::u_neg_div], batch);
+  context.updateTensor(wt_idx[AttentionParams::u_pos_div], batch);
+}
+
+void MoLAttentionLayer::exportTo(Exporter &exporter,
+                                 const ExportMethods &method) const {
+  AttentionLayer::exportTo(exporter, method);
+  LayerImpl::exportTo(exporter, method);
```
zhoonit, line 379: setProperty will need this as well

test/unittest/models/unittest_models.cpp
```diff
@@ -40,11 +40,34 @@ IniWrapper reduce_mean_last("reduce_mean_last",
...
                             });
 
+static std::unique_ptr<NeuralNetwork> makeMolAttention() {
+  std::unique_ptr<NeuralNetwork> nn(new NeuralNetwork());
+  nn->setProperty({"batch_size=3"});
+
+  auto outer_graph = makeGraph({
+    {"input", {"name=in3", "input_shape=1:1:5"}},
+    {"input", {"name=in2", "input_shape=1:4:6"}},
+    {"input", {"name=in1", "input_shape=1:1:6"}},
```
zhoonit, line 50: Could we change this to fc to check if derivatives are correct?

test/input_gen/genModelTests_v2.py
```diff
@@ -32,4 +69,10 @@ def forward(self, inputs, labels):
         name="reduce_mean_last",
     )
 
-    # inspect_file("lstm_single.nnmodelgolden")
```
zhoonit, line 35: Could you leave this at the last line for the reference?

test/input_gen/genModelsRecurrent_v2.py
```diff
@@ -190,6 +190,5 @@ def forward(self, inputs, labels):
         input_dims=[(3, 2)],
         label_dims=[(3, 2, 2)],
         name="grucell_stacked",
-    )
```
zhoonit, line 193: Maybe deleted by mistake?

nntrainer/layers/mol_attention_layer.cpp
kparichay: this is done in the last commit.

nntrainer/layers/mol_attention_layer.cpp
kparichay: because alpha_src is not a shared strided tensor but a contiguous tensor with data copied in it.

nntrainer/layers/mol_attention_layer.cpp
kparichay: This is just for name-keeping and understanding. This will just be a shallow copy.

nntrainer/layers/mol_attention_layer.cpp
kparichay: Ensures that `calcGradient()` and `calcDerivative()` both dont call `calcDerivativeHelper()` twice.

nntrainer/layers/layer_context.cpp
kparichay: will remove.

nntrainer/layers/mol_attention_layer.cpp
kparichay: A lot of tensors are locally allocated in calcDerivativeHelper(). I have a todo to optimize for it and will use this suggestion when optimizing.

nntrainer/layers/mol_attention_layer.cpp
kparichay: added.

test/input_gen/genModelsRecurrent_v2.py
kparichay: yes, fixed.

test/input_gen/genModelTests_v2.py
kparichay: done.

test/unittest/models/unittest_models.cpp
kparichay: I will update the #1758 unit test to include fc layer for derivatives.

nntrainer/layers/mol_attention_layer.cpp
kparichay: your formula is right. To reduce the amount of computation, consider the below subsitutions and simplified result used in the code:
```
du_neg_m = d_u_neg_div/beta_eps // given
u_neg_div = u_neg_m/beta_eps // given
dbeta_eps/du_neg_div = -d_u_neg_div * u_neg_m / (beta_eps ^ 2) = (d_u_neg_div/beta_eps) * (u_neg_m/beta_eps) * -1 = du_neg_m * u_neg_div
```

nntrainer/layers/mol_attention_layer.cpp
kparichay: same as above.

nntrainer/layers/mol_attention_layer.cpp
kparichay: 1. Skipped the batch size check as batch is assumed to be same across all inputs.
2. scores.width() is manually set to values.height() :) - https://github.com/nnstreamer/nntrainer/blob/775d4418ad208bcd9ba2a76d6c67f8c685a1c97e/nntrainer/layers/mol_attention_layer.cpp#L108

But the dimensions checks were completely missed in finalizing. Added dimensionality checks for values and query.

nntrainer/layers/mol_attention_layer.h
kparichay: Agreed, dual inheritance is creating issues with setProperty as pointed in the first example but IMO, LayerImpl should not throw on remaining properties but let the caller throw.
But this is beyond the scope of this PR, so removed inheritance to attention layer.

nntrainer/layers/mol_attention_layer.cpp
kparichay: Let me know if I didnt answer your question.

nntrainer/layers/mol_attention_layer.h
zhoonit: Thanks :)

nntrainer/layers/mol_attention_layer.cpp
zhoonit: Cool

nntrainer/layers/mol_attention_layer.cpp
zhoonit: I see, thanks

nntrainer/layers/mol_attention_layer.cpp
zhoonit: That's clever. Thanks for the explanation :)

nntrainer/layers/mol_attention_layer.cpp
zhoonit: 👍

nntrainer/layers/mol_attention_layer.cpp
zhoonit: I wonder alpha.size() is empty?

zhoonit (approved):
LGTM. Amazing job done :)

nntrainer/layers/mol_attention_layer.cpp
kparichay: alpha is indeed empty, but it does not throw as empty tensor is not strided

zhoonit: Ah missed that part. Thanks :)

jijoongmoon (approved):
LGTM

PR #1752, opened 2021-11-30
[Fix] Setting batch after finalize/initialize

zhoonit:

<details><summary>[Fix] Setting batch after finalize/initialize</summary><br />

This patch delays setting batch after finalize/initialize.

Setting batch must be done after runcontext has made, the reason is that
we have semantics that finalize should be independent of batch size, but
if batch size is set before output_dims batch must be set according to
input_dims batch, which is not desirable.

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -13,7 +13,6 @@
  */
 
 #include <cmath>
-#include <sstream>
 #include <stdexcept>
 #include <string>
 
@@ -262,10 +261,12 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
     deallocateTensors();
 
   for (auto iter = cbegin(); iter != cend(); iter++) {
-    (*iter)->setBatch(batch_size);
     if ((*iter)->isFinalized()) {
+      /// resize tensors spec
+      /// @todo remove below, if cutsom tensor needs to change dimension
+      /// according to the tensor, it must be done explicitly, or at least have
+      /// a property to control the behavior
       const RunLayerContext &context = (*iter)->getRunContext();
-      // resize tensors spec
       for (unsigned int idx = 0; idx < context.getNumTensors(); idx++) {
         auto const &ts = context.getTensor(idx);
         tensor_manager->setBatchSize(ts.getName(), ts.getDim().batch());
@@ -275,6 +276,8 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
                                        ts_grad.getDim().batch());
         }
       }
+      /// override setting batch as per request
+      (*iter)->setBatch(batch_size);
     }
   }
   /// resize input and output spec
@@ -688,7 +691,7 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
 
   /**
    * Request manager for either a pre-allocated input as output or a newly
-   * allocated input. This is necesary for manager to know when this output
+   * allocated input. This is neccesary for manager to know when this output
    * node is going to be used with in-place optimizations.
    */
   const std::vector<Var_Grad *> &outputs =
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -126,7 +126,7 @@ void BatchNormalizationLayer::finalize(InitLayerContext &context) {
    * as the output of this layer need not be stored all the time.
    */
   wt_idx[BNParams::t_full] =
-    context.requestTensor(in_dim, "tesnor_full", Tensor::Initializer::NONE,
+    context.requestTensor(in_dim, "tensor_full", Tensor::Initializer::NONE,
                           false, TensorLifespan::BACKWARD_FUNC_LIFESPAN);
   /**
    * caches variance + epsilon as well.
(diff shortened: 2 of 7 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -194,6 +191,9 @@ int NeuralNetwork::initialize() {
   status = model_graph.initialize(input_layers, label_layers);
   NN_RETURN_STATUS();
 
+  model_graph.setBatchSize(
```
kparichay, line 194: :100:

kparichay (approved):
LGTM after CI fix
You might need to fix bugs in some layer if setBatch() is not properly supported by them

zhoonit:
> LGTM after CI fix You might need to fix bugs in some layer if setBatch() is not properly supported by them

Yeah, there were some (hidden) bugs related to this. Please take another look :)

nntrainer/layers/layer_node.cpp
```diff
@@ -586,20 +587,10 @@ void LayerNode::calcGradient() {
...
 void LayerNode::setBatch(unsigned int batch) {
-  /** @todo we won't going to need Layer::setBatch(InitLayerContext), remove it
-   */
-  if (hasInputShapeProperty()) {
-    auto &input_shapes =
-      std::get<std::vector<props::InputShape>>(*layer_node_props);
-    for (auto &input_shape : input_shapes) {
-      input_shape.get().batch(batch);
-    }
-  }
```
zhoonit, line 597: @note: we don't (and should not) query input shape property for querying (current and actual) input dimension, by this we can segregate + detect bugs earlier which would happen by assuming input_shape has the batch dimension as well.

nntrainer/layers/rnncell.cpp
```diff
@@ -248,7 +249,12 @@ void RNNCellLayer::setBatch(RunLayerContext &context, unsigned int batch) {
   const unsigned int max_timestep = std::get<props::MaxTimestep>(rnncell_props);
   context.updateTensor(wt_idx[RNNCellParams::hidden_state],
                        batch * max_timestep);
-  context.updateTensor(wt_idx[RNNCellParams::dropout_mask], batch);
+
+  const float dropout_rate = std::get<props::DropOutRate>(rnncell_props);
+  if (dropout_rate > epsilon) {
+    /// @note default value of wt_idx[dropout_mask] is 0
+    context.updateTensor(wt_idx[RNNCellParams::dropout_mask], batch);
+  }
```
zhoonit, line 257: FYI) @lhs8928

nntrainer/layers/bn_layer.cpp
```diff
@@ -256,6 +256,14 @@ void BatchNormalizationLayer::setBatch(RunLayerContext &context,
                                        unsigned int batch) {
   context.updateTensor(wt_idx[BNParams::deviation], batch);
   context.updateTensor(wt_idx[BNParams::t_full], batch);
+
+  /// reset divider
+  divider = 1;
+  auto input_dim = context.getInput(0).getDim();
+  input_dim.batch(batch);
```
kparichay, line 263: This is already done by layer node and can be skipped.

nntrainer/graph/network_graph.cpp
```diff
@@ -398,6 +398,8 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
                                        ts_grad.getDim().batch());
         }
       }
+      /// override setting batch as per request
+      (*iter)->setBatch(batch_size);
```
kparichay, line 402: Are you sure to move it here? If this is moved here, then the changes requested here by the layers will not be reflected (which is done from line 392 to 399)

zhoonit: Yeah, this is intended. 
for example below should be override default behavior.
```cpp
  context.updateTensor(wt_idx[RNNCellParams::hidden_state],
                       batch * max_timestep);
```

Please correct me if I got something wrong :)

kparichay: You are right that context.updateTensor() should decide how to update the tensor shapes, but IMU this updated information should reach tensor_pool, which is done from L392 to L399.

kparichay: The memory is always de-allocated with setBatch(), so `context.updateTensor()` alone is sufficient.
L392 to L399 can be removed IMO as its purpose was to pass the information to the tensor_pool if not deallocated, please check.

zhoonit: Yeah, the function guarantee's it is deallocated so it should be fine. But I'll delay removing the part as we don't have a proper test scenario that batch is being changed and run train or inference. (this will have to be added to a model test scenario). I am suspecting some layers requires setBatch() but did not create one, I'll leave an issue instead for now.

kparichay: We do have a model test with ccapi :) - https://github.com/nnstreamer/nntrainer/blob/c09bbf73ae234e9c3bd989c59d3fb7af220679b7/test/ccapi/unittest_ccapi.cpp#L364-L386

zhoonit: Awesome :)

nntrainer/layers/bn_layer.cpp
zhoonit: Please note that, input_dim.batch is not updated when this function is being called. We might want to update inputdims before coming here later.

jijoongmoon (approved):
LGTM

PR #1765, opened 2021-12-06
[all] Mixed commits for fixes

kparichay:
- concat to not loop over all inputs multiple times.
- gradient accumulation
- support only 1 output in case state update is not used
- bug fix in backwarding
- activation to work out of place
- Bug fix for activaiton realizer to find the realizer where the layer is
looping to itself with name change.
TODO: also do for flatten and other realizers which modified the graph
in similar fashion.
- zoneout lstm bug fix for handling the scenario where mask rate is set to
zero for either hidden or cell.
- Undo earlier changes due to bug in them.
The changes will be proposed again with more unittests.
The changes have been commented for now.
- Add clip gradient norm property for the the model as a simple interface
to set for all the layers.
- support recorder to read data from file than always generating random
data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 # ignore build directory
 /build
 /builddir
+.cache/
 
 # jni build files
 iniparser/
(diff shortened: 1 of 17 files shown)


zhoonit (approved):
LGTM!

jijoongmoon (approved):
LGTM!. So many fixes!

PR #1767, opened 2021-12-07
[manager] Remove activation input exec order from backwarding

### LINKED ISSUE #1725: Bug in memory optimization
For the activation layer, the forward output tensor of the activation layer is used in `calcDerivative` - this is added by the manager.

However, the forward input tensor is not used in the `calcDerivative` and this usage is not removed from the manager.

kparichay:
This patch removes activation input exec order from backwarding as input
of the activation layer is not used in the backwarding.

This leads to change in the unittests as we cannot check all the
outputs, especially close to the end of the model. So, with optimization
enabled, only the output layer's forwarding is checked.

Resolves #1725

### DIFF
--- a/test/unittest/models/models_test_utils.cpp
+++ b/test/unittest/models/models_test_utils.cpp
@@ -376,7 +376,7 @@ void GraphWatcher::compareFor(const std::string &reference,
 
     auto it = nodes.begin();
     for (; it != nodes.end() - 1; ++it) {
-      it->forward(iteration, !(it + 1)->supportInPlace());
+      it->forward(iteration, it->isOutputNode() | !optimize);
     }
     it->forward(iteration, true);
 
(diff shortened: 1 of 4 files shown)


test/unittest/models/models_test_utils.cpp
```diff
@@ -376,7 +376,7 @@ void GraphWatcher::compareFor(const std::string &reference,
 
     auto it = nodes.begin();
     for (; it != nodes.end() - 1; ++it) {
-      it->forward(iteration, !(it + 1)->supportInPlace());
+      it->forward(iteration, it->isOutputNode() | !optimize);
```
zhoonit, line 379: was this the reason for test fail? :p

zhoonit (approved):
LGTM :)

test/unittest/models/models_test_utils.cpp
kparichay: yes :)

kparichay: with these optimizations, we cannot always match all the outputs as they get overwritten before they can be verified.

jijoongmoon (approved):
LGTM

Issue #1769, opened 2021-12-07
Support skipping of use of bias in layer


kparichay:
Support of skipping of bias in a layer is needed from the finetuning purpose.
If the pretrained model does not use the bias, then nntrainer must not create a non-zero bias after training as it cannot be loaded with the pretrained model.

kparichay:
#1770 and #1780 supported this.

PR #1770, opened 2021-12-07
[layer] Support disable bias for fc and conv

kparichay:
- Add disable bias property for the layer impl class.
- This patch enables disable bias for fully connected and convolution layer.

Support for disable bias in other classes extending layer impl class will be done later.
See Also #1769

### DIFF
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -107,6 +107,21 @@ public:
 };
 
 /**
+ * @brief BiasInitializer Initialization Enumeration Information
+ *
+ */
+class DisableBias : public nntrainer::Property<bool> {
+public:
+  /**
+   * @brief Construct a DisableBias object
+   *
+   */
+  DisableBias(bool val = false) : nntrainer::Property<bool>(val) {}
+  using prop_tag = bool_prop_tag;
+  static constexpr const char *key = "disable_bias";
+};
+
+/**
  * @brief Normalization property, normalize the input to be in range [0, 1] if
  * true
  *
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -265,6 +265,7 @@ void Conv2DLayer::finalize(InitLayerContext &context) {
   auto &weight_initializer =
     std::get<props::WeightInitializer>(*layer_impl_props);
   auto &bias_initializer = std::get<props::BiasInitializer>(*layer_impl_props);
+  auto &disable_bias = std::get<props::DisableBias>(*layer_impl_props);
 
   unsigned int filter_size = std::get<props::FilterSize>(conv_props);
   auto &kernel_size =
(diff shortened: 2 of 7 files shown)


lhs8928 (approved):
LGTM

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -281,8 +282,11 @@ void Conv2DLayer::finalize(InitLayerContext &context) {
   wt_idx[ConvParams::weight] =
     context.requestWeight(dim, weight_initializer, weight_regularizer,
                           weight_regularizer_constant, "filter", true);
-  wt_idx[ConvParams::bias] = context.requestWeight(
-    bias_dim, bias_initializer, WeightRegularizer::NONE, 1.0f, "bias", true);
+
+  if (disable_bias.empty() || disable_bias.get() == false) {
+    wt_idx[ConvParams::bias] = context.requestWeight(
```
zhoonit, line 287: note) wt_idx[ConvParams::bias] will be default to be 0.

If someone tries to access this without requesting anything, this will falsely point to weights.

We will need to use default value of -1 or some big number to make it explicit error.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -484,7 +490,11 @@ void Conv2DLayer::calcGradient(RunLayerContext &context) {
   }
 
   delK.reshape(filter_dim);
-  derivative.sum({0, 2, 3}, delBias);
+  if (auto &disable_bias = std::get<props::DisableBias>(*layer_impl_props);
+      disable_bias.empty() || disable_bias.get() == false) {
```
zhoonit, line 494: nit picking) constructor of props::DisableBias that it is not empty so IMO we wouldn't need to check disable_bias.empty() for all.

nntrainer/layers/common_properties.h
```diff
@@ -106,6 +106,21 @@ class Trainable : public nntrainer::Property<bool> {
   using prop_tag = bool_prop_tag;
 };
 
+/**
+ * @brief BiasInitializer Initialization Enumeration Information
```
lhs8928, line 110: Please modify this

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -404,9 +407,13 @@ void Conv2DLayer::forwarding(RunLayerContext &context, bool training) {
...
-  status = hidden_.add_i(bias_kernel);
-  if (status != ML_ERROR_NONE) {
-    throw std::invalid_argument("[Conv2D] adding bias failed");
+  if (auto &disable_bias = std::get<props::DisableBias>(*layer_impl_props);
+      disable_bias.empty() || disable_bias.get() == false) {
+    Tensor &bias_kernel = context.getWeight(wt_idx[ConvParams::bias]);
+    status = hidden_.add_i(bias_kernel);
+    if (status != ML_ERROR_NONE) {
+      throw std::invalid_argument("[Conv2D] adding bias failed");
+    }
```
zhoonit, line 416: I know you are just transpiling but want to ask for the opinion.
should this be added for fc as well(it's missing)? or as we assures dimension matches, we don't check this?

nntrainer/layers/conv2d_layer.cpp
kparichay: I will make initialized to std::numeric_limits<unsigned int>::max() for all layers.

nntrainer/layers/conv2d_layer.cpp
kparichay: it's not needed. `add_i` always returns ML_ERROR_NONE from tensor.cpp. Even this can be removed but I will just leave it here.

zhoonit: Cool

nntrainer/layers/conv2d_layer.cpp
zhoonit: Alright :) I suppose it's gotta be out of scope of this PR

zhoonit (approved):
LGTM :)

jijoongmoon (approved):
LGTM

PR #1774, opened 2021-12-09
[layer] bug fix for array initializer

### LINKED ISSUE #1773: weight indices default value buggy in each layer
weight indices default values in each layer is defaulted to 0 but if not requested, instead of giving out of bounds access will return a different weight. The default value must be set to out of bounds.

kparichay:
Bug fix for array initializer storing the indices of the requested
weight so that accessing weights not requested always causes error.

resolves #1773

### DIFF
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -107,7 +107,7 @@ public:
 };
 
 /**
- * @brief BiasInitializer Initialization Enumeration Information
+ * @brief DisableBias to disable the bias
  *
  */
 class DisableBias : public nntrainer::Property<bool> {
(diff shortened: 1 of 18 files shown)


zhoonit (approved):
great work :)

nntrainer/layers/common_properties.h
```diff
@@ -107,7 +107,7 @@ class Trainable : public nntrainer::Property<bool> {
 };
 
 /**
- * @brief BiasInitializer Initialization Enumeration Information
+ * @brief DisableBias to disable the bias
```
lhs8928, line 110: 👍

lhs8928 (commented):
LGTM

jijoongmoon (approved):
LGTM

Issue #1776, opened 2021-12-09
Learning Rate Scheduler support


kparichay:
Support for an independent learning rate scheduler interface is needed.
- [x] create learning rate scheduler interface - must be easy to support custom scheduler #1776
- [x] support existing learning rate schedulers (constant, and exponential) with this interface #1776
- [x] register with appcontext for creation by type and later register other schedulers as they come and add unittests #1778
- [x] decouple learning rate for optimizer
  - [x] support new ini section `LearningRateScheduler` with properties `type=constant/exponential`, `learning_rate=1.0`, .... #1783
  - [x] for backward compatibility, the properties and ini interface must support this and pass the request to lr scheduler interface #1782
  - [x] remove learning rate member functions from optimizer but the properties must be supported for backward compatibility #1782
    - [x] create optimizer_scheduler - containing both optimizer and learning rate scheduler - OptimizerWrapped
    - [x] optimizer_scheduler will inherit ml::train::optimizer and will composite nntrainer::optimizer - OptimizerWrapped
- [x] support piecewise constant learning rate type #1784 
- [x] support C++ API #1786
- [x] support C API #2183

kparichay:
For the ini interface/API, there are two options:
```ini
# Option 1
[optimizer]
type = adam
learning_rate_scheduler=cosine
learning_rate=0.1
decay_steps=100

# Option 2
[optimizer]
type = adam
# for backward compatibility, learning_rate, decay_steps and decay_rate properties
# will be supported here, which will internally result in creation of constant/exponential
# learning rate schedule

[LearningRateScheduler]
type=cosine
decay_steps=100
learning_rate=0.1
```

The c/c++ API will look like below:
```c++
// option 1
auto optimizer = ml::train::optimizer::Adam({"learning_rate_scheduler=cosine", "learning_rate=0.1", "decay_steps=100"});

// option 2
auto optimizer = ml::train::optimizer::Adam(); 
auto lr = ml::train::learning_rate_scheduler::Cosine({"learning_rate=0.1", "decay_steps=100"});
optimizer.setLearningRate(lr);
// optimizer.setProperty({"learning_rate=0.1", "decay_steps=100"}); will be available for constant/exponential learning rate as an alternative to optimizer.setLearningRate() for backward compatility
```

I recommend option 2 as its more clearer and decouples the learning-rate scheduler from optimizer details.
Please share your opinions.

jijoongmoon:
[...]

PR #1782, opened 2021-12-10
[optimizer] Add optimizer wrapped 

kparichay:
- Add optimizer wrapped which wraps the opitmizer and the learning rate
scheduler.
In order to be backward compatible, each optimizer must support setting
the learning rate, decay rate and decay steps, even for new optimizers.
To make this extensible without each optimizer storing this information
and merging with the learning rate schedulers, and not creating new
interfaces, optimizer wrapped is added.
Optimizer wraps around optimizer, and owns both the optimizer and
learning rate scheduler. If the properties of LR or decay are passed to
the optimizer, they are intercepted by the optimizer wrapped and passed
to the learning rate scheduler appropriately.
- Update to use the optimizer wrapped instead of optimizer directly, and
prepare for the using learning rate scheduler.
- Update the optimizer to use the learning rate scheduler in the training
and weight updation.
- cleanup the optimizers.

See Also #1776


nntrainer/optimizers/optimizer_wrapped.cpp
```diff
@@ -0,0 +1,144 @@
...
+  auto remain_props = loadProperties(values, props);
+  optimizer->setProperty(remain_props);
+}
+
+double OptimizerWrapped::getLearningRate(size_t iteration) {
+  return lr_sched->getLearningRate(iteration);
+}
+
+void OptimizerWrapped::applyGradient(RunOptimizerContext &context) {
+  // optimizer->applyGradient(context);
```
zhoonit: comment intended?

nntrainer/optimizers/optimizer_wrapped.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void OptimizerWrapped::exportTo(Exporter &exporter,
+                                const ExportMethods &method) const {
+  // optimizer->exportTo(exporter, method);
```
zhoonit: comment intended?

nntrainer/optimizers/optimizer_wrapped.cpp
kparichay: yes, this will be enabled once the optimizer is set in upcoming PR.

nntrainer/optimizers/optimizer_wrapped.cpp
kparichay: yes, this will be enabled once the optimizer is set in upcoming PR.

zhoonit (approved):
LGTM, could you share briefly

lhs8928 (approved):
LGTM

jijoongmoon:
This pr is included in #1871

PR #1783, opened 2021-12-10
[wait for #1782] [modelloader] Support learning rate scheduler

kparichay:
Support learning rate scheduler with model loader along with a unittest.


nntrainer/models/model_loader.cpp
```diff
@@ -68,6 +99,7 @@ int ModelLoader::loadOptimizerConfigIni(dictionary *ini, NeuralNetwork &model) {
     std::shared_ptr<ml::train::Optimizer> optimizer =
       createOptimizerWrapped(opt_type, properties);
     model.setOptimizer(optimizer);
+    loadLearningRateSchedulerConfigIni(ini, optimizer);
```
zhoonit, line 102: what if learning rate scheduler exist but optimizer does not exist?

test/ccapi/unittest_ccapi.cpp
```diff
@@ -196,7 +196,7 @@ static nntrainer::IniSection outputlayer("outputlayer",
  */
 TEST(nntrainer_ccapi, train_with_config_01_p) {
   std::unique_ptr<ml::train::Model> model;
-  ScopedIni s("test_train_01_p",
+  ScopedIni s("ccapi_test_train_01_p",
```
zhoonit, line 199: nice finding :)

nntrainer/models/model_loader.cpp
kparichay: Then it will have no effect as the learning rate scheduler is used only by the optimizer.

zhoonit: I think it's reasonable. Could we issue a warning for this case?

If someone add sch in the ini file and add optimizer with model api, they might expect sch is added.

zhoonit (commented):
Oh and could you update ini saving as well? since we don't have a test for this, physically checking if learning rate section is created will be more than enough.

nntrainer/models/model_loader.cpp
kparichay: There is an existing warning: `There is no [Optimizer] section in the given ini file. This model can only be used for inference`. Do you want one more warning?

zhoonit: I see. Maybe adding `[Learning Schedule] section will be ignored if any` their would work. Thanks :)

lhs8928 (approved):
LGTM

jijoongmoon:
This pr is included in #1871

PR #1785, opened 2021-12-13
[layer] mse layer optimize for memory and compute

kparichay:
- reduce memory usage for mse forwarding.
- optimize mse backwarding calculation by merging multiple and divide into
a single call.
- optimize mse layer to use scaled l2norm instead of manually calculating
the l2norm.


zhoonit (approved):
LGTM, nice one :)

jijoongmoon (approved):
LGTM

PR #1788, opened 2021-12-14
[API] Add layer visitor for the model

zhoonit:

<details><summary>[API] Add layer visitor for the model</summary><br />

This patch add visitor for the model api

**Self evaluation:**

</details>

### DIFF
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -29,6 +29,9 @@
 #include <optimizer.h>
 #include <tensor_dim.h>
 
+namespace nntrainer {
+class RunLayerContext;
+}
 /** Define more aliases for the model in the API */
 namespace ml {
 namespace train {
@@ -191,6 +194,22 @@ public:
     const std::vector<std::string> &start_layers,
     const std::vector<std::string> &end_layers, ReferenceLayersType type,
     const std::vector<std::string> &type_properties = {}) = 0;
+
+  /**
+   * @brief Visit each layer inside model
+   * @param fn function to be called for each layer
+   * - call param
+   *   @param layer layer the call back is visiting
+   *   @param rc run layer context reference clubbed with layer
+   *   @param user_data user data passed along the function
+   * @param user_data user data to pass along the callback
+   */
+  virtual void
+  forEachLayer(std::function<void(Layer & /**< layer */,
+                                  nntrainer::RunLayerContext & /**< rc */,
+                                  void * /**< user_data */)>
+                 fn,
+               void *user_data = nullptr) = 0;
 
   /**
    * @brief     set optimizer for the neural network model
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -21,6 +21,7 @@
  *
  */
 
+#include "layer_context.h"
 #include <cmath>
 #include <cstring>
 #include <fstream>
@@ -1024,5 +1025,14 @@ void NeuralNetwork::print(std::ostream &out, unsigned int flags,
     (*iter)->printPreset(out, layerPrintPreset);
 
   /// @todo Add status to check neuralnet has been run. #290
+}
+
+void NeuralNetwork::forEachLayer(
+  std::function<void(ml::train::Layer &, RunLayerContext &, void *)> fn,
+  void *user_data) {
+  for (auto iter = model_graph.cbegin(); iter != model_graph.cend(); iter++) {
+    auto ln = std::static_pointer_cast<LayerNode>(*iter).get();
+    fn(*ln, std::forward<RunLayerContext &>(ln->getRunContext()), user_data);
+  };
 }
 } /* namespace nntrainer */
--- a/nntrainer/models/neuralnet.h
+++ b/nntrainer/models/neuralnet.h
@@ -293,6 +293,17 @@ public:
                  std::shared_ptr<ml::train::Dataset> dataset) override;
 
   /**
+   * @copydoc void forEachLayer(std::function<void(Layer &,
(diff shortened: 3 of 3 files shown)


api/ccapi/include/model.h
```diff
@@ -192,6 +195,20 @@ class Model {
...
+   * @brief Visit each layer inside model
+   * @param fn function to be called for each layer
+   * - call param
+   *   @param layer layer the call back is visiting
+   *   @param rc run layer context clubbed with layer, null if not initialized
+   * yet
+   *
+   */
+  virtual void
+  forEachLayer(std::function<void(Layer & /**< layer */,
```
kparichay, line 208: You should note a few things with opening this API:
- layer_context.h must be opened with this
- add a `void *user_data` which will passed to forEachLayer, and will be included in the function signature.
- if this is being opened, it would be good to also open `registerCallback` with 3 modes (before forwarding, before backwarding, after backwarding) which will also receive RunLayerContext

Regarding this particular function, if not initialized, I recommend to not allowed to be run for the layers as finalize of the layers is itself not run as each layer will have to check for this.

jijoongmoon (approved):
LGTM

api/ccapi/include/model.h
zhoonit: > layer_context.h must be opened with this

yes, layer_context.h is also used for the custom devel so already opened. We will need to move it to ml::train {}.

> add a void *user_data which will be passed to forEachLayer

std::function already can capture states so we won't need this. This will be needed for the capi front.

> it would be good to also open registerCallback with 3 modes 

Yeah, we will need to have some entry point for the model api. Let's discuss this when fixating model c++ api.

> Regarding this particular function, if not initialized, I recommend not allow to be run for the layers as finalize of the layers is itself not run as each layer will have to check for this. As the name/properties of the layer, ordering of execution, count of layers, and model itself change before and after compilation, it would be tough for the user to understand what is happening.

I think this is fair point. I'll not allow this and update documentation as per suggestion.

kparichay: > > add a void *user_data which will be passed to forEachLayer

> std::function already can capture states so we won't need this. This will be needed for the capi front.

If there will be a `void *user_data` for the c API anyway, I recommend to add here itself.

api/ccapi/include/model.h
```diff
@@ -192,6 +195,20 @@ class Model {
...
+                                  nntrainer::RunLayerContext * /**< rc */)>
```
kparichay, line 209: how about `RunLayerContext &` - match the API for a layer developer.

zhoonit: Yeah :) if this is after finalized, it will be Context &

api/ccapi/include/model.h
zhoonit: IMO, adding void * user_data is likely to encourage bad practice to the c++ api user. There is a functor, which is type safe and broader way to alter void *.

Let's have in-depth discussion later though, I'll add `void *` for now :)

lhs8928 (approved):
LGTM

api/ccapi/include/model.h
```diff
@@ -192,6 +195,22 @@ class Model {
     const std::vector<std::string> &end_layers, ReferenceLayersType type,
     const std::vector<std::string> &type_properties = {}) = 0;
 
+  /**
+   * @brief Visit each layer inside model
+   * @param fn function to be called for each layer
+   * - call param
+   *   @param layer layer the call back is visiting
+   *   @param rc run layer context clubbed with layer, null if not initialized
```
kparichay, line 203: update doxygen as per updated signature and behavior - reference for run layer context, and add user_data.

nntrainer/models/neuralnet.cpp
```diff
@@ -1025,4 +1026,13 @@ void NeuralNetwork::print(std::ostream &out, unsigned int flags,
 
   /// @todo Add status to check neuralnet has been run. #290
 }
+
+void NeuralNetwork::forEachLayer(
+  std::function<void(ml::train::Layer &, RunLayerContext &, void *)> fn,
+  void *user_data) {
+  for (auto iter = model_graph.cbegin(); iter != model_graph.cend(); iter++) {
```
kparichay, line 1033: you can add a quick check that model if model is not initialize, then throw.
or maybe ln->getRunContext() will throw?

zhoonit: Yes it throws so I skipped adding it

api/ccapi/include/model.h
zhoonit: done.

nntrainer/models/neuralnet.h
```diff
@@ -292,6 +292,17 @@ class NeuralNetwork : public ml::train::Model {
   int setDataset(const DatasetModeType &dt,
                  std::shared_ptr<ml::train::Dataset> dataset) override;
 
+  /**
+   * @copydoc void forEachLayer(std::function<void(Layer &,
+   * nntrainer::RunLayerContext &), void *user_data> fn);
+   *
+   */
+  void forEachLayer(
+    std::function<void(ml::train::Layer & /**< layer */,
```
kparichay, line 301: This interface will be limited to running the layer with ml::train::layer interface. If you want more functionality, `ml::train::Layer` can be `nntrainer::Layer` (it's not just a typecast anymore). Feel free to do later.

kparichay (approved):
LGTM

nntrainer/models/neuralnet.cpp
kparichay: cool.

api/ccapi/include/model.h
kparichay: cool.

PR #1790, opened 2021-12-14
[optim] Add pytorch reference for adam @open sesame 12/22 19:01

kparichay:
Add pytorch reference mode for adam which references adam implementation
of pytorch.


jijoongmoon (approved):
LGTM

PR #1808, opened 2021-12-20
[tensor] Tensor pool request bug fix

### LINKED ISSUE #1807: tensor pool max execution order bug
Tensor pool is including the max execution order in the execution order of some tensors.

kparichay:
If the max execution order for the tensor pool was less than the
largest execution order for a tensor in the tensor pool, then the max
execution order was being included in the execution order of that
tensor.
This patch resolves this issue for end order, as well as for the start
order.

resolves #1807

### DIFF
--- a/nntrainer/tensor/tensor_pool.cpp
+++ b/nntrainer/tensor/tensor_pool.cpp
@@ -112,11 +112,23 @@ void TensorPool::finalize(const MemoryPlanner &planner,
     }
     details->token = 0;
 
-    /** 1. create the validity ranges for the all the requested tensors */
-    unsigned int validity_start =
-      *std::min_element(details->exec_order.begin(), details->exec_order.end());
-    unsigned int validity_end =
-      *std::max_element(details->exec_order.begin(), details->exec_order.end());
+    /**
+     * 1. create the validity ranges for the all the requested tensors.
+     * validity_start/validity_end should be a value in the exec order of the
+     * given tensor or a value out of range so as to not request memory for this
+     * tensor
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/tensor_pool.cpp
```diff
@@ -131,8 +143,6 @@ void TensorPool::finalize(const MemoryPlanner &planner,
     /** 2. for each tensor request if it is in the provided range */
     if (validity_end < start_order || validity_start > end_order)
       continue;
```
zhoonit, line 145: Shouldn't this always be false (so never continue) because validity_end = validity_start is always >= start_order.

Could you check if initial value of `validity_end` = `start_order - 1` ?

kparichay: this is a sanity check in case the previous calculations does do anything wrong.

zhoonit: I mean sanity check will always pass because validity_end < start_order will be always false, if this is intended, I think this line should be removed instead.

kparichay: validity_end can be less than start_order, when the full range of validity is less than the start_order itself.

zhoonit: Resolved offline that `validity_end < start_order` remains always false but will be here for completeness.

zhoonit (approved):
LGTM

jijoongmoon (approved):
LGTM!

PR #1821, opened 2022-01-06
[model] add train log path property

lhs8928:
- Added SaveTrainLogPath property in model to save train status to file

Self evaluation:

### DIFF
--- a/nntrainer/models/model_common_properties.h
+++ b/nntrainer/models/model_common_properties.h
@@ -74,6 +74,17 @@ public:
 };
 
 /**
+ * @brief train log save path property
+ *
+ */
+class SaveTrainLogPath : public Property<std::string> {
+public:
+  static constexpr const char *key =
+    "save_train_log_path";       /**< unique key to access */
+  using prop_tag = str_prop_tag; /**< property type */
+};
+
+/**
  * @brief model batch size property
  *
  */
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -62,7 +62,8 @@ NeuralNetwork::NeuralNetwork(AppContext app_context_) :
   model_props(props::LossType(), {}, {}, props::ClipGradByGlobalNorm()),
   model_flex_props(props::Epochs(), props::TrainingBatchSize(),
                    props::SavePath(), props::ContinueTrain(),
-                   props::SaveBestPath(), props::MemoryOptimization()),
+                   props::SaveBestPath(), props::SaveTrainLogPath(),
+                   props::MemoryOptimization()),
   load_path(std::string()),
   epoch_idx(0),
   iter(0),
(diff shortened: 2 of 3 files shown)


jijoongmoon (approved):
LGTM

nntrainer/models/neuralnet.cpp
```diff
@@ -704,13 +713,18 @@ int NeuralNetwork::train_run() {
...
-  auto train_epoch_end = [this](RunStats &stat, DataBuffer &buffer) {
+  auto train_epoch_end = [this, &train_log_file](RunStats &stat,
+                                                 DataBuffer &buffer) {
     stat.loss /= static_cast<float>(stat.num_iterations);
     auto &save_path = std::get<props::SavePath>(model_flex_props);
     if (!save_path.empty()) {
       save(save_path, ml::train::ModelFormat::MODEL_FORMAT_BIN);
     }
 
+    if (train_log_file.good()) {
```
zhoonit, line 724: I think it's better to check if prop exists and try this while check if train_log_file is bad, we throw

nntrainer/models/model_common_properties.h
```diff
@@ -73,6 +73,17 @@ class SaveBestPath : public Property<std::string> {
   using prop_tag = str_prop_tag; /**< property type */
 };
 
+/**
+ * @brief train log save path property
+ *
+ */
+class SaveTrainLogPath : public Property<std::string> {
```
zhoonit, line 80: For now it looks good, but I think this should be a callback so that the user has some control over the properties(https://www.tensorflow.org/guide/keras/custom_callback?hl=en)

nntrainer/models/neuralnet.cpp
```diff
@@ -629,6 +630,14 @@ int NeuralNetwork::train_run() {
 
   auto batch_size = std::get<props::TrainingBatchSize>(model_flex_props);
 
+  std::ofstream train_log_file;
+  auto &save_train_log_path =
+    std::get<props::SaveTrainLogPath>(model_flex_props);
+  if (!save_train_log_path.empty()) {
+    train_log_file =
+      checkedOpenStream<std::ofstream>(save_train_log_path, std::ios::out);
+  }
```
kparichay, line 639: consider adding an else case where `train_log_file` points to `/dev/null` (`train_log_file.open( "/dev/null", FLAGS);`).  Then you dont need to add checks on `train_log_file` or `save_train_log_path` property is valid or exists

kparichay (approved):
LGTM other than a cleanup comment

songgot (approved):
Looks good to me

jihochu (approved):

DonghakPark (approved):
LGTM

SeoHyungjun:
LGTM!

myungjoo:
@lhs8928 Are you still working on it?

PR #1823, opened 2022-01-10
[ Network Graph ] Temporal Fix in getInputGrad

jijoongmoon:
When we turn on the inPlaceOptimization, it will fail if we have
mulitiple inputs: input layer ( can be inplace operatation ), normal
layers. Then, we are not allocate the grad between input layer and
this layer and it cause the error during calcDeriviate which requires
the grad tensor of input layer.

This patch fix this temporally by creating tensor buffer it reqruies.

and it includes modification of mem_check script to generate proper
output.

**Self evaluation:**

### DIFF
--- a/Applications/utils/mem_usage.sh
+++ b/Applications/utils/mem_usage.sh
@@ -10,14 +10,34 @@
 
 "$@" &
 process_id=$!
+peak=0
+mem_peak=0
+log="\nPID - $process_id :"
+outlog=""
+count=0
+while true; do
+    if [ -d "/proc/$process_id" ]; then
+	output=`grep VmHWM /proc/$process_id/status`
+	IFS=' ' read -ra mem_str <<< "$output"
+	count=$((count + 1))
+	mem_usage=$((${mem_str[1]}))
+	outlog="$outlog$count $mem_usage kB\n"
+	let mem_peak='mem_usage > mem_peak ? mem_usage : mem_peak'
+	sleep 1s # configure the sleep time if the program does not run long enough
+    else
+	break
+    fi
+done
 
 if [ -d "/proc/$process_id" ]; then
-  sleep 30s # configure the sleep time if the program does not run long enough
-  output=`grep VmHWM /proc/$process_id/status`
-  IFS=' ' read -ra mem_str <<< "$output"
-  mem_usage=$((${mem_str[1]}))
+    kill -9 $process_id
 fi
-echo "Mem Usage: $mem_usage"
 
-kill -9 $process_id
+log="$log: $mem_peak kB"
+log="$log\n-------------------------------------\n"
+log="$log$outlog"
+
+echo -e "$log"
+
+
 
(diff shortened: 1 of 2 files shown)


zhoonit (approved):
LGTM :)

Applications/utils/mem_usage.sh
```diff
@@ -10,14 +10,34 @@
...
+outlog=""
+count=0
+while true; do
+    if [ -d "/proc/$process_id" ]; then
+	output=`grep VmHWM /proc/$process_id/status`
+	IFS=' ' read -ra mem_str <<< "$output"
+	count=$((count + 1))
+	mem_usage=$((${mem_str[1]}))
+	outlog="$outlog$count $mem_usage kB\n"
+	let mem_peak='mem_usage > mem_peak ? mem_usage : mem_peak'
```
kparichay, line 25: 💯

Applications/utils/mem_usage.sh
```diff
@@ -10,14 +10,34 @@
...
+log="\nPID - $process_id :"
+outlog=""
+count=0
+while true; do
+    if [ -d "/proc/$process_id" ]; then
+	output=`grep VmHWM /proc/$process_id/status`
+	IFS=' ' read -ra mem_str <<< "$output"
+	count=$((count + 1))
+	mem_usage=$((${mem_str[1]}))
+	outlog="$outlog$count $mem_usage kB\n"
```
kparichay, line 24: later you can consider printing the log `$count $mem_usage kB\n` here itself to give user some feedback while the script progresses.

nntrainer/layers/layer_context.cpp
```diff
@@ -274,9 +274,20 @@ const Tensor &RunLayerContext::getInput(unsigned int idx) const {
...
+    //       the getInputGrad for multiple inputs ( input 1: Input Layer,
+    //       input 2: Normal Input ). In this situation, if we enable
+    //       inPlaceOptimization, it will fail.
+    //
+    // throw std::invalid_argument(
+    //   "Requesting gradient for a non-trainable tensor.");
+    //
+    Tensor var = inputs[idx]->getVariableRef();
+    Tensor t_grad = var;
+    inputs[idx]->initializeGradient(t_grad);
```
kparichay, line 289: This will make gradient share memory with the variable itself which can cause errors in certain cases.
I recommend to consider `return Tensor(inputs[idx]->getDim());`

jijoongmoon: So, you mean we have to create tensor every time, then?

kparichay: > So, you mean we have to create tensor every time, then? 

yes, we would have to. The current implementation can create issues if the `input->getVariableRef()` is used later, as the gradient is sharing the memory and can overwrite the tensor data. 

However, as this is just a temporary fix, if its working, lets keep and add this as a note/comment here.

kparichay (approved):
LGTM

nntrainer/layers/layer_context.cpp
jijoongmoon: fixed

PR #1824, opened 2022-01-10
[Android] Delegate option control to android.mk

zhoonit:

<details><summary>[Android] Delegate option control to android.mk</summary><br />

This patch delegates option control in android mk to meson for debug and
optimized build

**Self evaluation:**

</details>

### DIFF
--- a/jni/meson.build
+++ b/jni/meson.build
@@ -1,6 +1,16 @@
 and_conf = configuration_data()
 
-and_conf.set('MESON_CFLAGS', ' '.join(extra_defines))
+flags = []
+if get_option('debug')
+    flags += '-g'
+endif
+
+flags += '-O@0@'.format(get_option('optimization'))
+message('compile flags are:' + ' '.join(flags))
+
+flags = extra_defines + flags
+
+and_conf.set('MESON_CFLAGS', ' '.join(flags))
 and_conf.set('MESON_NNTRAINER_SRCS', ' '.join(nntrainer_sources))
 and_conf.set('MESON_NNTRAINER_INCS', ' '.join(nntrainer_inc_abs))
 and_conf.set('MESON_CCAPI_NNTRAINER_SRCS', ' '.join(ccapi_src))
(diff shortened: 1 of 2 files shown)


jni/meson.build
```diff
@@ -1,6 +1,16 @@
 and_conf = configuration_data()
 
-and_conf.set('MESON_CFLAGS', ' '.join(extra_defines))
+flags = []
+if get_option('debug')
+    flags += '-g'
+endif
+
+flags += '-O@0@'.format(get_option('optimization'))
```
kparichay, line 8: note that meson in-built optimization does not `fast` (0 < 1 < 2 < 3 < fast. In my understanding, the only way to enable `-Ofast` is to pass the flag manually.

zhoonit: Yeah, I have checked that https://mesonbuild.com/Builtin-options.html#core-options. I think we shouldn't use Ofast anyway so it should be okay. (fyi) in the reference Ofast is not particularly faster then O3 as well :p )

kparichay: Our experiment results are with or without -Ofast? if they are without `-Ofast` then its fine, otherwise all the experiments timings must be measured again with this PR.

zhoonit: They are (probably) with Ofast.
For the paper purpose I agree, but for the deployment purpose I think we don't need to do the Ofast benchmark.

Reason 1. O3 (seems) faster or at least equal to Ofast
Reason 2. O3 and Ofast has 0.2 initial loss difference (implying that Ofast is anyway wrong).

It would be better if we can test both, but we don't have much room to do broader test for now.

jijoongmoon (approved):
LGTM

jni/meson.build
kparichay: For now, as we are using `O3` in ubuntu and tizen builds, lets stick to O3 in android as well.

kparichay (approved):
LGTM

PR #1829, opened 2022-01-14
[lstm] remove timestep property

lhs8928:
[lstm] remove timestep property
    
     - Remove timestep property from lstm layer.
       This will disable unrolling the lstm layer.
     - Adjust recurrent unittest to simple lstm unittest.
    

nntrainer/layers/lstm.cpp
```diff
@@ -189,23 +189,11 @@ void LSTMLayer::forwarding(RunLayerContext &context, bool training) {
   const float dropout_rate = std::get<props::DropOutRate>(lstm_props).get();
   const unsigned int max_timestep =
     std::get<props::MaxTimestep>(lstm_props).get();
-  const props::Timestep timestep = std::get<props::Timestep>(lstm_props);
```
jijoongmoon, line 192: Is there any reason to remove the time step? That means the height of input is always 1?

nntrainer/layers/lstm.cpp
```diff
@@ -225,25 +213,16 @@ void LSTMLayer::forwarding(RunLayerContext &context, bool training) {
...
+  cs.setZero();
 
   for (unsigned int batch = 0; batch < batch_size; ++batch) {
     const Tensor input_batch = inputs.getBatchSlice(batch, 1);
     Tensor hs_batch = hs.getBatchSlice(batch, 1);
     Tensor cs_batch = cs.getBatchSlice(batch, 1);
     Tensor ifgo_batch = ifgos.getBatchSlice(batch, 1);
 
-    for (unsigned int t = start_timestep; t < end_timestep; ++t) {
+    for (unsigned int t = 0; t < max_timestep; ++t) {
```
jijoongmoon, line 225: I wonder time step always starts from 0?

jijoongmoon:
how about change the tile of the PR with more details .. intead of ~~~/1 and ~~~/2.?

test/input_gen/genModelsRecurrent_v2.py
```diff
@@ -56,40 +56,39 @@ def forward(self, inputs, labels):
         return ret, loss
 
 class LSTMStacked(torch.nn.Module):
```
kparichay, line 58: I wonder why change this unit test than add another one?

kparichay (approved):

nntrainer/layers/lstm.cpp
lhs8928: No. The height of input can be greater than 1.
Timestep property was used to unrolling the lstm layer like lstmcell layer. 
AFAIK this property is introduced because we did not implemented lstmcell layer but now we have so I removed this property.

nntrainer/layers/lstm.cpp
lhs8928: Yes. In the lstm layer timestep always starts from 0.

test/input_gen/genModelsRecurrent_v2.py
lhs8928: Previous unittest was test lstm layer works well when unrolling the lstm layer like lstmcell layer which is not applicable with this pr.

lhs8928:
> how about changing the tile of the PR with more details .. instead of ~~~/1 and ~~~/2.?

Thanks for your notice :)

jijoongmoon (approved):
LGTM

PR #1834, opened 2022-01-20
[ SAVE/LOAD ] save / load optimizer variables

jijoongmoon:
Enable save / load optimizer variables such as M and V for adam
optimizer

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -415,24 +415,46 @@ void LayerNode::exportTo(Exporter &exporter,
   layer->exportTo(exporter, method);
 }
 
-void LayerNode::read(std::ifstream &file) {
+void LayerNode::read(std::ifstream &file, bool opt_var) {
   NNTR_THROW_IF(!run_context, std::runtime_error)
     << __func__ << " layer needs to be finalized first!";
-  for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
-    /// @note shared weights are only be read at the first acecss
-    if (run_context->isGradientLastAccess(i)) {
-      run_context->getWeight(i).read(file);
+  if (opt_var) {
+    for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
+      if (run_context->isGradientLastAccess(i)) {
+        // @note read optimizer variables
+        for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
+          run_context->getWeightOptVar(i, j).read(file);
+        }
+      }
+    }
+  } else {
+    for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
+      /// @note shared weights are only be read at the first acecss
+      if (run_context->isGradientLastAccess(i)) {
+        run_context->getWeight(i).read(file);
+      }
     }
   }
 }
 
-void LayerNode::save(std::ofstream &file) const {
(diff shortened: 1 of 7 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -422,6 +422,10 @@ void LayerNode::read(std::ifstream &file) {
     /// @note shared weights are only be read at the first acecss
     if (run_context->isGradientLastAccess(i)) {
       run_context->getWeight(i).read(file);
+      // @note read optimizer variables
+      for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
```
kparichay, line 426: This PR will not provide backward compatibility and will break with existing saved model files. In my understanding, there are at least a couple of solutions:
1. loading the optimizer variables to be controlled by a property/config - which can be disabled by default providing backward compatibility to this PR
2. the optimizer variables are located at the end of the file, and must be loaded/saved after all the weights - this provides backward compatibility but requires lots of changes.

I would recommend to go with option 1.

nntrainer/models/neuralnet.cpp
```diff
@@ -203,10 +203,6 @@ int NeuralNetwork::initialize() {
 
   initialized = true;
 
-  if (!load_path.empty()) {
```
kparichay, line 206: this will break the inference API. I would recommend loading the weights here and then loading again at training for the optimizer.

To handle the optimizer variables not being allocated, you can move the file pointer ahead by the size than reading those values.

nntrainer/layers/layer_node.cpp
jijoongmoon: I think it is not the kind of feature to keep the backward compatibility. It's a bug. It might be better that we notice and encourage to change.

nntrainer/models/neuralnet.cpp
jijoongmoon: I think it is time to consider generating the model and bin file for the inference and training separately. For the temporary fixes, we could move ahead as you said for the case of inference. On second thought, need to handle different optimizers. for example, the model is saved with ADAM but load with SGD. We need to write the opt type in the model file. Definitely need some kind of header.

kparichay: IMO the inference and training saved file should be compatible for both training and inference.
Overall, we need a structure to the model file - flatbuffer/json format or some header.

nntrainer/layers/layer_node.cpp
kparichay: This discussion is pretty much similar to the other comment, both realizing the need for a model file structure.
Also, I think with this change, we need to update our release version number as this is a breaking change.

Regarding model compatibility, an even easier option is to support a new enum BIN_V2 with this if needed.

jijoongmoon: Right. I agree. But it is more than this pr is for.. and we need to deal with separate pr and issue. For temporal fixes, I will manage and update this PR with a small scope. Not a new release of NNTrainer or BIN_V2.

nntrainer/models/neuralnet.cpp
jijoongmoon: right. compatible but no need to be the same file.

kparichay (approved):
LGTM after CI pass

lhs8928 (approved):
LGTM :)

PR #1840, opened 2022-02-11
[layernode] Bug fix for gradient access @open sesame 02/11 18:12

kparichay:
Add bug fix to check if gradient is available before accessing gradient
related variables.

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -420,7 +420,8 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
     << __func__ << " layer needs to be finalized first!";
   if (opt_var) {
     for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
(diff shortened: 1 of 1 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -430,7 +431,8 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
   } else {
     for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
       /// @note shared weights are only be read at the first acecss
-      if (run_context->isGradientLastAccess(i)) {
+      if (run_context->weightHasGradient(i) &&
```
jijoongmoon, line 434: I think this shouldn't check the weightHasGradient. we need to load the weight for the forwarding only.

kparichay: The reason to add this is that `isGradientLastAccess` is wrong or not properly set for weights that don't have a gradient.

kparichay:
This is covered in #1842.

PR #1841, opened 2022-02-17
[capi/test] Example of batchsize set at train

kparichay:
Update capi test to include an example where batchsize is set at the
training time and not at compile time.


lhs8928 (approved):
LGTM

myungjoo (approved):
LGTM

PR #1842, opened 2022-02-17
[Load/Save] Update to load and save the adam momentum variables

jijoongmoon:
In this PR, Load and save is updated for the adam optimizer variables.
In order to do it, optimizer load twice, one for the weight, and one
for the optimizer variables.

- Added TOC generator for README.md

Resolves:

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -420,10 +420,13 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
     << __func__ << " layer needs to be finalized first!";
   if (opt_var) {
     for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
-      if (run_context->isGradientLastAccess(i)) {
+      if (run_context->isGradientLastAccess(i) && getTrainable()) {
         // @note read optimizer variables
-        for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
-          run_context->getWeightOptVar(i, j).read(file);
+        if (run_context->weightHasGradient(i)) {
+          for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i);
+               ++j) {
+            run_context->getWeightOptVar(i, j).read(file);
+          }
         }
       }
     }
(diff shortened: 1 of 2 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -205,7 +205,10 @@ int NeuralNetwork::initialize() {
 
   initialized = true;
 
-  if (!load_path.empty()) {
+  // @note we need check loadedWeight for the case of multiple call of load to
+  // load weight. Only the weight needs to be loaded here. Becuase the buffer
+  // for the optimizer is not allocated yet.
+  if (!load_path.empty() && !loadedWeight) {
```
kparichay, line 211: is this possible? in the current design, the model cannot load weights before initialization as initialization allocates weight buffers.

nntrainer/layers/layer_node.cpp
```diff
@@ -420,10 +420,13 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
     << __func__ << " layer needs to be finalized first!";
   if (opt_var) {
     for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
-      if (run_context->isGradientLastAccess(i)) {
+      if (run_context->isGradientLastAccess(i) && getTrainable()) {
         // @note read optimizer variables
-        for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
-          run_context->getWeightOptVar(i, j).read(file);
+        if (run_context->weightHasGradient(i)) {
```
kparichay, line 425: weightHasGradient() and getTrainable() have the same effect?

jijoongmoon: it is not. for the batch normalization layer, it has 4 weights and two of them do not have gradient even if trainable is set true.

nntrainer/models/neuralnet.cpp
jijoongmoon: model_graph.allocateWeights() is called at line 204. so, the buffer for weight is allocated and so, it is possible to load weight.

nntrainer/layers/layer_node.cpp
kparichay: thanks for the explanation.

nntrainer/models/neuralnet.cpp
kparichay: My concern is `loadedWeight` will always be false here, as the model cannot be initialized twice and NeuralNetwork::load() needs model to be initialized.

kparichay (approved):
LGTM

nntrainer/models/neuralnet.cpp
```diff
@@ -381,18 +385,22 @@ void NeuralNetwork::load(const std::string &file_path,
...
       model_file.seekg(bin_file_pos);
 
-      std::string opt_type;
-      opt_type = readString(model_file);
-      if (istrequal(opt_type, "adam") && istrequal(opt->getType(), "adam")) {
-        for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
-             iter++) {
-          (*iter)->read(model_file, true);
+      if (istrequal(opt->getType(), "adam")) {
+        char opt_type[4];
```
lhs8928, line 397: Does the adam is the only case which loads variables for optimizer?
If it is not, then I think the previous way is more suitable for general cases.

lhs8928 (approved):
LGTM except minor comment

PR #1843, opened 2022-02-17
[lstm] implement calcGradient for bidirectional lstm @open sesame 05/06 15:07

lhs8928:
- Implement calcGradient for bidirectional lstm
 - Added test case for bidirectional lstm

Self evaluation:


songgot (approved):
Looks good to me

jijoongmoon (approved):
LGTM!

test/unittest/models/unittest_models_recurrent.cpp
```diff
@@ -175,46 +175,46 @@ static std::unique_ptr<NeuralNetwork> makeStackedLSTM() {
...
+  for (auto &node : outer_graph) {
+    nn->addLayer(node);
+  }
+
+  nn->setOptimizer(ml::train::createOptimizer("sgd", {"learning_rate = 0.1"}));
+  return nn;
+}
+
+static std::unique_ptr<NeuralNetwork> makeStackedBidirectionalLSTM() {
+  std::unique_ptr<NeuralNetwork> nn(new NeuralNetwork());
```
jihochu, line 198: How about using `make_unique()` ?
```suggestion
  auto nn = std::make_unique<NeuralNetwork>();
```

please check the reference: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3588.txt

> make_unique shares two advantages with make_shared (excluding the third advantage, increased
efficiency). First, unique_ptr<LongTypeName> up(new LongTypeName(args)) must
mention LongTypeName twice, while auto up = make_unique<LongTypeName>(args)
mentions it once. Second, make_unique prevents the unspecified-evaluation-order
leak triggered by expressions like foo(unique_ptr<X>(new X),
unique_ptr<Y>(new Y)). (Following the advice "never say new" is simpler than
"never say new, unless you immediately give it to a named unique_ptr".)

lhs8928: Thank you for your notice :)

jihochu (approved):
LGTM

jijoongmoon:
rebase is required to fix the ci.

PR #1844, opened 2022-02-18
Weight decay fix

kparichay:
- weight decay should be applied before calling the optimizer
this was not detected earlier as it was tested with sgd
- Enable weighte regularization and decay with weiht clipping.


jijoongmoon (approved):
LGTM

PR #1851, opened 2022-03-03
Load Optimizer Variables @open sesame 3/8 11:35

jijoongmoon:
In this PR,
 1. new property of adam optimizer, "laod_var" is added to set loading
    momentum variables
 2. update the read and save binary file
    using  ml::train::OptimizerType
 3. update read in layer_node to skip the optimizer variables if
    load_var is set as "false"
 4. is_laod_var() is added in optimizer_devel

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -333,7 +333,16 @@ void NeuralNetwork::save(const std::string &file_path,
       (*iter)->save(model_file);
     }
 
+    model_file.write((char *)&epoch_idx, sizeof(epoch_idx));
+    model_file.write((char *)&iter, sizeof(iter));
+
     opt->save(model_file);
+    unsigned int write_op_var = 0;
+    if (istrequal(opt->getType(), "adam")) {
+      write_op_var = 1;
+    }
+
+    model_file.write((char *)&write_op_var, sizeof(write_op_var));
 
     if (istrequal(opt->getType(), "adam")) {
       for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
@@ -342,12 +351,10 @@ void NeuralNetwork::save(const std::string &file_path,
       }
     }
 
-    model_file.write((char *)&epoch_idx, sizeof(epoch_idx));
-    model_file.write((char *)&iter, sizeof(iter));
-
     model_file.close();
     break;
   }
+
   case ml::train::ModelFormat::MODEL_FORMAT_INI:
     saveModelIni(file_path);
     break;
@@ -396,21 +403,24 @@ void NeuralNetwork::load(const std::string &file_path,
       /// read. so, after this line, additional read shouldn't be called
       model_file.seekg(bin_file_pos);
 
-      if (istrequal(opt->getType(), "adam")) {
-        char opt_type[4];
-        model_file.read(opt_type, 4);
-        if (istrequal(opt_type, "adam")) {
-          for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
-               iter++) {
-            (*iter)->read(model_file, true);
-          }
-        }
-      }
-
       checkedRead(model_file, (char *)&epoch_idx, sizeof(epoch_idx),
                   "[NeuralNetwork::readModel] failed to read epoch_idx");
       checkedRead(model_file, (char *)&iter, sizeof(iter),
                   "[NeuralNetwork::readModel] failed to read iteration");
+
+      opt->read(model_file);
+      unsigned int exist_op_var = 0;
+
+      checkedRead(model_file, (char *)&exist_op_var, sizeof(exist_op_var));
+
+      if (istrequal(opt->getType(), "adam") && exist_op_var) {
+        bool load_opt_var = opt->is_load_var();
+        for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
+             iter++) {
+          (*iter)->read(model_file, true, load_opt_var);
+        }
+      }
+
     } catch (...) {
       std::cerr << "failed to read epoch idx, proceeding with default index\n";
     }
--- a/nntrainer/optimizers/optimizer_devel.cpp
+++ b/nntrainer/optimizers/optimizer_devel.cpp
@@ -30,7 +30,19 @@ void Optimizer::setProperty(const std::vector<std::string> &values) {
 }
 
 void Optimizer::read(std::ifstream &file) {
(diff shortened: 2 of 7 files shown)


nntrainer/optimizers/optimizer_devel.cpp
```diff
@@ -30,14 +30,33 @@ void Optimizer::setProperty(const std::vector<std::string> &values) {
 }
 
 void Optimizer::read(std::ifstream &file) {
-  std::string loaded_type = readString(file);
+  std::string loaded_type;
+  unsigned int opt_type = ml::train::OptimizerType::UNKNOWN;
+  checkedRead(file, (char *)&opt_type, sizeof(opt_type));
+  switch (opt_type) {
+  case ml::train::OptimizerType::ADAM:
+    loaded_type = "adam";
```
kparichay, line 38: loaded_type = Adam::type;

nntrainer/optimizers/optimizer_devel.cpp
```diff
@@ -30,14 +30,33 @@ void Optimizer::setProperty(const std::vector<std::string> &values) {
...
+    break;
+  case ml::train::OptimizerType::SGD:
+    loaded_type = "sgd";
```
kparichay, line 41: loaded_type = SGD::type;

nntrainer/optimizers/optimizer_devel.cpp
```diff
@@ -30,14 +30,33 @@ void Optimizer::setProperty(const std::vector<std::string> &values) {
...
   if (loaded_type != getType()) {
     throw std::runtime_error(
       "[Optimizer::read] written type unmatches with set type");
   }
 }
 
-void Optimizer::save(std::ofstream &file) { writeString(file, getType()); }
+void Optimizer::save(std::ofstream &file) {
+  unsigned int opt_type = ml::train::OptimizerType::UNKNOWN;
+  if (istrequal(getType(), "adam"))
```
kparichay, line 55: "adam" -> Adam::type

nntrainer/optimizers/optimizer_devel.cpp
```diff
@@ -30,14 +30,33 @@ void Optimizer::setProperty(const std::vector<std::string> &values) {
...
+    opt_type = ml::train::OptimizerType::ADAM;
+  if (istrequal(getType(), "sgd"))
```
kparichay, line 57: "sgd" -> SGD::type

kparichay (approved):
Overall LGTM

nntrainer/optimizers/optimizer_devel.cpp
jijoongmoon: fixed.

nntrainer/optimizers/optimizer_devel.cpp
jijoongmoon: fixed.

nntrainer/optimizers/optimizer_devel.cpp
jijoongmoon: fixed.

nntrainer/optimizers/optimizer_devel.cpp
jijoongmoon: fixed.

nntrainer/models/neuralnet.cpp
```diff
@@ -396,14 +403,16 @@ void NeuralNetwork::load(const std::string &file_path,
...
-          for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
-               iter++) {
-            (*iter)->read(model_file, true);
-          }
+      opt->read(model_file);
+      unsigned int exist_op_var = 0;
+
+      checkedRead(model_file, (char *)&exist_op_var, sizeof(exist_op_var));
+
+      if (istrequal(opt->getType(), "adam") && exist_op_var) {
```
lhs8928, line 411: I think 
`istrequal(opt->getType(),` "adam")`
part have to move line 415 like
(*iter)->read(model_file, true, istrequal(opt->getType(), "adam") && load_opt_var);

Think of the case which trained with adam and save optimizer variables and read with sgd optimizer.

nntrainer/models/neuralnet.cpp
```diff
@@ -334,6 +334,12 @@ void NeuralNetwork::save(const std::string &file_path,
     }
 
     opt->save(model_file);
+    unsigned int write_op_var = 0;
+    if (istrequal(opt->getType(), "adam")) {
```
lhs8928, line 338: As kparichay mentioned it is better to modify "adam" to Adam::type

nntrainer/models/neuralnet.cpp
jijoongmoon: in the case of sgd and other optimizer, we could not load the variables because it does not have variables. so.. I guess we need save it always at the end of binary file. so that we could just finish read if the variables does not need.

lhs8928 (approved):
LGTM

nntrainer/models/neuralnet.cpp
```diff
@@ -333,7 +333,16 @@ void NeuralNetwork::save(const std::string &file_path,
       (*iter)->save(model_file);
     }
 
+    model_file.write((char *)&epoch_idx, sizeof(epoch_idx));
+    model_file.write((char *)&iter, sizeof(iter));
```
zhoonit, line 337: This changes file reading order, which was reading opt type -> epoch idx -> iter number now it's epoch idx -> iter -> opt type -> write opt var(new). this should be updated with where this is being used.

nntrainer/models/neuralnet.cpp
```diff
@@ -396,21 +403,24 @@ void NeuralNetwork::load(const std::string &file_path,
...
-          }
-        }
-      }
-
       checkedRead(model_file, (char *)&epoch_idx, sizeof(epoch_idx),
                   "[NeuralNetwork::readModel] failed to read epoch_idx");
       checkedRead(model_file, (char *)&iter, sizeof(iter),
                   "[NeuralNetwork::readModel] failed to read iteration");
+
+      opt->read(model_file);
```
zhoonit, line 411: this is used to be string of "adam", now it's single value, this is non backward compatible change which needs attention.

zhoonit (commented):
There is at least one more problem. After this patch, reference model does not train as expected. Until the issue is found and resolved (+ favorably, adding a test as well), the patch should remain pending.

Issue #1849, opened 2022-02-25
reduce restrictions for setting optimizer/dataset of a model


kparichay:
The public API requires the optimizer/dataset to be set before ml_train_model_compile. However, these restrictions can be relaxed and both can be later right until the training is being done:
1. dataset - current implementation allows this already, only documentation changes are needed
2. optimizer - minor implementation changes are needed (remove the check for the model is compiled when setting the optimizer and move the initialization of the optimizer to the training itself)

Thank you @pkosko for pointing these out.

PR #1854, opened 2022-03-11
Change loading meta information behavior @open sesame 03/14 11:08

zhoonit:

<details><summary>Change loading meta information behavior</summary><br />

**Before this PR**

optimizer variable loaded from load_path every time.
Calling model->train(); in a row became unintuitive

1. model->train() load from original load path
thus iteration number roll back to the first one.
2. Same happens for the adam weight
3. model->load(); after model->initialize(); is noop
because loadedWeight becomes true

**After this PR**

1. model load from load_path only at initialize time
2. model->load is not implicitly overriden

**Additional Changes**

1. optimizer weight became part of weights. Now available after initialize()
2. Save format became coherent with load format
3. Some unused variables deleted

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/layers/layer_context.cpp
+++ b/nntrainer/layers/layer_context.cpp
@@ -172,9 +172,6 @@ Tensor &RunLayerContext::getWeightGrad(unsigned int idx) const {
  */
 Tensor &RunLayerContext::getWeightOptVar(unsigned int idx,
                                          unsigned int jdx) const {
-  if (!weights[idx]->hasGradient())
-    throw std::invalid_argument(
-      "Requesting gradient for a non-trainable weight.");
   return weights[idx]->getOptimizerVariableRef(jdx);
 }
 
@@ -185,9 +182,6 @@ Tensor &RunLayerContext::getWeightOptVar(unsigned int idx,
  * @return int Number of the weight optimizer variable
  */
 unsigned int RunLayerContext::getNumWeightOptVar(unsigned int idx) const {
-  if (!weights[idx]->hasGradient())
-    throw std::invalid_argument(
-      "Requesting gradient for a non-trainable weight.");
   return weights[idx]->getNumOptVariable();
 }
 
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -421,12 +421,9 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
   if (opt_var) {
     for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
       if (run_context->isGradientLastAccess(i) && getTrainable()) {
-        // @note read optimizer variables
-        if (run_context->weightHasGradient(i)) {
-          for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i);
-               ++j) {
-            run_context->getWeightOptVar(i, j).read(file);
-          }
+        /// @note read optimizer variables
+        for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
+          run_context->getWeightOptVar(i, j).read(file);
         }
       }
     }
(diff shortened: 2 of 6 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -421,12 +421,9 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
...
       if (run_context->isGradientLastAccess(i) && getTrainable()) {
-        // @note read optimizer variables
-        if (run_context->weightHasGradient(i)) {
-          for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i);
-               ++j) {
-            run_context->getWeightOptVar(i, j).read(file);
-          }
+        /// @note read optimizer variables
+        for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
+          run_context->getWeightOptVar(i, j).read(file);
```
jijoongmoon, line 426: I wonder whether this is working for the bn layer without weightHasGradient check. Cause I had error last time.

nntrainer/models/neuralnet.cpp
```diff
@@ -381,22 +375,13 @@ void NeuralNetwork::load(const std::string &file_path,
...
+    for (auto iter = model_graph.cbegin(); iter != model_graph.cend(); iter++) {
+      (*iter)->read(model_file);
     }
     try {
       /// this is assuming that the failure is allowed at the end of the file
       /// read. so, after this line, additional read shouldn't be called
-      model_file.seekg(bin_file_pos);
-
-      if (istrequal(opt->getType(), "adam")) {
+      if (opt && istrequal(opt->getType(), "adam")) {
```
jijoongmoon, line 384: opt is sgd.. and if the file is saved with adam.. then, I guess we need to skip to read opt adam variables.. and try to read iteration number and epoch.

nntrainer/models/neuralnet.cpp
```diff
@@ -205,11 +205,7 @@ int NeuralNetwork::initialize() {
 
   initialized = true;
 
-  // @note we need check loadedWeight for the case of multiple call of load to
-  // load weight. Only the weight needs to be loaded here. Becuase the buffer
-  // for the optimizer is not allocated yet.
-  // loadedWeight check is just for the duplicate load of weight.
-  if (!load_path.empty() && !loadedWeight) {
+  if (!load_path.empty()) {
```
jijoongmoon, line 208: At this time, the opt variables are not allocated, I guess. correct me if I'm wrong.

nntrainer/layers/layer_node.cpp
zhoonit: bn mu and var is non-trainable so it should not allocate any weiht opt var (making run_context->getNumWeightOptVar(i) == 0

https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/layers/bn_layer.cpp#L105-L116

When I tested there was no problem regarding this matter, but if there is we will need to look at :)

nntrainer/models/neuralnet.cpp
zhoonit: Yes, it can be solved by changing file format (eg, reading iteration and epoch before optimizer section) for now, the patch is done in a way not breaking the file format. I guess the strategy in #1851 this can be fixed although it is reverted for other reason.

nntrainer/models/neuralnet.cpp
zhoonit: Yes it used to be, in this patch optimizer variable is allocated when weight is being allocated(manager.cpp:608), we may argue this is not the optimized strategy though in case of inference mode.

nntrainer/layers/layer_context.cpp
```diff
@@ -172,9 +172,6 @@ Tensor &RunLayerContext::getWeightGrad(unsigned int idx) const {
  */
 Tensor &RunLayerContext::getWeightOptVar(unsigned int idx,
                                          unsigned int jdx) const {
-  if (!weights[idx]->hasGradient())
```
kparichay, line 175: I wonder if it's safe to remove this check. If weight is not trainable, its optimizer variables would not be allocated memory.

nntrainer/models/neuralnet.cpp
kparichay: this is fine for now but can be done lazily with #1849, where optimizer will be determined at training time and not at initialize.

kparichay (approved):
Overall LGTM other than a minor comment

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_context.cpp
zhoonit: we need to check 1) weight index is in range, 2) weight opt index is in range. Having gradient or not is orthogonal in terms of this function. As other functions don't check bounds for performance reason, I think we can let this be. But I'll add a comment in the header to state that this function does not check any bounds.

kparichay: The issue is not the bounds, but if the tensor will be empty or allocated. 
Optimizer requests optimizer variables for all the weights even if they are trainable or not.
However, the optimizer requested tensor will not be allocated (in my understanding) if the weight is not trainable.
The lines in discussion performs the check if the weight is trainable or not by checking if a gradient was requested for the weight.

PR #1859, opened 2022-03-17
[fix] check return value

lhs8928:
- Added missing statement to check called function works properly


myungjoo (approved):

jijoongmoon (approved):
LGTM

Issue #1853, opened 2022-03-10
meson build fails on clean install


jrgeraci:
On Ubuntu 20.04 WSL2 AND normal Ubuntu 20.04 
After following the instructions on 
https://github.com/nnstreamer/nntrainer/blob/main/docs/getting-started.md#build-with-meson
meson build fails to find
Run-time dependency capi-ml-inference found: NO (tried pkgconfig and cmake)
and thereby fails with
nntrainer/layers/meson.build:58:4: ERROR: Problem encountered: NNStreamer CAPI dependency not found

CC:@myungjoo This may be an nnstreamer api related issue as nnstreamer-examples also fails to find this.

```
(base) jrgeraci@MSI:~/nntrainer$ meson build
The Meson build system
Version: 0.53.2
Source dir: /home/jrgeraci/nntrainer
Build dir: /home/jrgeraci/nntrainer/build
Build type: native build
Project name: nntrainer
Project version: 0.3.0
C compiler for the host machine: cc (gcc 9.4.0 "cc (Ubuntu 9.4.0-1ubuntu1~20.04) 9.4.0")
C linker for the host machine: cc ld.bfd 2.34
C++ compiler for the host machine: c++ (gcc 9.4.0 "c++ (Ubuntu 9.4.0-1ubuntu1~20.04) 9.4.0")
C++ linker for the host machine: c++ ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64
Compiler for C supports arguments -Wredundant-decls: YES
Compiler for C++ supports arguments -Wredundant-decls: YES
Compiler for C supports arguments -Wwrite-strings: YES
Compiler for C++ supports arguments -Wwrite-strings: YES
Compiler for C supports arguments -Wformat: YES
Compiler for C++ supports arguments -Wformat: YES
Compiler for C supports arguments -Wformat-nonliteral: YES
Compiler for C++ supports arguments -Wformat-nonliteral: YES
Compiler for C supports arguments -Wformat-security: YES
Compiler for C++ supports arguments -Wformat-security: YES
Compiler for C supports arguments -Winit-self: YES
Compiler for C++ supports arguments -Winit-self: YES
Compiler for C supports arguments -Waddress: YES
Compiler for C++ supports arguments -Waddress: YES
Compiler for C supports arguments -Wno-multichar -Wmultichar: YES
Compiler for C++ supports arguments -Wno-multichar -Wmultichar: YES
[...]

PR #1861, opened 2022-03-18
meson: dependency on ml-api should be optional @open sesame 03/31 17:58

myungjoo:
Users should be able to build nntrainer in a system without
ML-API and nnstreamer by default.

ML-API and nnstreamer dependency should only be mandated
in related systems (a few embedded systems).

Let's keep it "auto" so that most external users can forget
about nnstreamer and ML-API.

Addresses #1853

Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -54,6 +54,7 @@ warning_c_flags = [
   '-Wno-error=varargs'
 ]
 
+
 foreach extra_arg : warning_flags
   if cc.has_argument (extra_arg)
     add_project_arguments([extra_arg], language: 'c')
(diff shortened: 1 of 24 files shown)


meson.build
```diff
@@ -202,22 +223,7 @@ if not iniparser_dep.found()
...
-    warning('capi-nnstreamer version is too old, we do not know if it works with older nnstreamer version')
-  endif
-endif
-
-ml_api_common_dep = dummy_dep
-
-if get_option('platform') != 'android'
-  ml_api_common_dep = dependency(get_option('capi-ml-common-actual'), required: true)
-else
+if get_option('platform') == 'android'
```
jijoongmoon, line 226: like the below, if we will download ml-api-inference, then, we need to set -DML_API_COMMON=1 for android.

jijoongmoon:
LGTM, after android ci build test.

jijoongmoon:
I think there are CI server issues. I will fix it and let you know.

myungjoo:
Android build issue resolved. PTAL: @jijoongmoon @songgot @jihochu @lhs8928

jijoongmoon (approved):
CI issue is fixed.

PR #1873, opened 2022-04-06
[neuralnet] Add null character when load optizmier type from bin file

lhs8928:
- Added missing null character when load optimizer type from bin file

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -329,7 +329,8 @@ void NeuralNetwork::save(const std::string &file_path,
       (*iter)->save(model_file);
     }
     if (opt && istrequal(opt->getType(), "adam")) {
-      model_file.write("adam", 4);
+      std::string adam = "adam";
+      model_file.write(adam.c_str(), adam.size());
       for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
            iter++) {
         (*iter)->save(model_file, true);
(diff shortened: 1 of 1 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -382,7 +382,8 @@ void NeuralNetwork::load(const std::string &file_path,
       /// this is assuming that the failure is allowed at the end of the file
       /// read. so, after this line, additional read shouldn't be called
       if (opt && istrequal(opt->getType(), "adam")) {
-        char opt_type[4];
+        char opt_type[5];
+        opt_type[4] = '\0';
         model_file.read(opt_type, 4);
```
songgot, line 387: How about check the return value if 0 or ssize_t ?

nntrainer/models/neuralnet.cpp
```diff
@@ -382,8 +382,9 @@ void NeuralNetwork::load(const std::string &file_path,
       /// this is assuming that the failure is allowed at the end of the file
       /// read. so, after this line, additional read shouldn't be called
       if (opt && istrequal(opt->getType(), "adam")) {
-        char opt_type[4];
-        model_file.read(opt_type, 4);
+        std::string opt_type;
```
jijoongmoon, line 385: you might need to change the save function as well.

lhs8928: Done

jijoongmoon (approved):
LGTM

PR #1881, opened 2022-04-18
[ Compiler ] add batch norm layer realizer

jijoongmoon:
This patch includes skeleton code for bn realizer which removes the
batch normalization layer for inference

**Self evaluation:**

### DIFF
--- a/nntrainer/compiler/bn_realizer.h
+++ b/nntrainer/compiler/bn_realizer.h
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2022 Jijoong Moon <jijoong.moon@samsung.com>
+ *
+ * @file bn_realizer.h
+ * @date 13 April 2022
+ * @brief NNTrainer graph realizer which remove batch normalization layer for
+ * inference
+ * @see	https://github.com/nnstreamer/nntrainer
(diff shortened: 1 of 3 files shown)


lhs8928:
Can we restore bn_layer after batch norm layer realizer is applied?

jijoongmoon:
> Can we restore bn_layer after batch norm layer realizer is applied?

Well, this is only for inference. We will call this realizer when we save the model for the inference. It does not modify the original model described by user.

songgot (approved):
Looks good to me

lhs8928 (approved):
LGTM

nntrainer/compiler/bn_realizer.h
```diff
@@ -0,0 +1,58 @@
...
+   * @brief Destroy the Graph Realizer object
+   *
+   */
+  ~BnRealizer();
+
+  /**
+   * @brief graph realizer creates a shallow copied graph based on the reference
+   * @note bn realizer removes batch normalization layers from
+   * GraphRepresentation
+   * @throw std::invalid_argument if graph is ill formed
```
lhs8928, line 50: @param is omitted

jijoongmoon: fixed in later pr

PR #1882, opened 2022-04-19
[Wait for #1881] [ Compiler ] Implement bn realizer with test 

jijoongmoon:
This patch completes the bn realizer for the inference.
This path is only for the inference and therefore it is supposed to 1 to
1 connection ( one input and one output for the bn layer ) in the
model graph. That means if there are multiple connections, then
multi-out layer is followed, and make sure the bn layer has 1 to 1
in/output during compilation.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -223,9 +223,10 @@ void LayerNode::setOutputConnection(unsigned nth, const std::string &name,
   }
 
   auto &con = output_connections[nth];
-  NNTR_THROW_IF(con, std::invalid_argument)
-    << "cannot override connection, this slot is reserved for "
-    << con->toString();
+  // Should be override connection for the batch normalization realizer
+  // NNTR_THROW_IF(con, std::invalid_argument)
+  //   << "cannot override connection, this slot is reserved for "
+  //   << con->toString();
 
   con = std::make_unique<Connection>(name, index);
 }
--- a/test/nntrainer_test_util.cpp
+++ b/test/nntrainer_test_util.cpp
@@ -218,6 +218,30 @@ makeGraph(const std::vector<LayerRepresentation> &layer_reps) {
   return graph_rep;
 }
 
+nntrainer::GraphRepresentation
+makeGraph_V2(const std::vector<LayerRepresentation> &layer_reps) {
+  static auto &ac = nntrainer::AppContext::Global();
+
+  nntrainer::GraphRepresentation graph_rep;
+  auto model_graph = nntrainer::NetworkGraph();
+  for (auto &layer_representation : layer_reps) {
+    std::shared_ptr<nntrainer::LayerNode> layer = nntrainer::createLayerNode(
+      ac.createObject<nntrainer::Layer>(layer_representation.first),
+      layer_representation.second);
+    model_graph.addLayer(layer);
+  }
+  // compile with loss
+  model_graph.compile("mse");
+
+  for (auto &node : model_graph.getLayerNodes()) {
+    graph_rep.push_back(node);
+  }
+
+  // remove loss layer
+  graph_rep.pop_back();
+  return graph_rep;
+}
+
 void sizeCheckedReadTensor(nntrainer::Tensor &t, std::ifstream &file,
                            const std::string &error_msg) {
   unsigned int sz = 0;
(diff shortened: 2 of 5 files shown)


songgot (approved):
Looks good to me

test/nntrainer_test_util.cpp
```diff
@@ -218,6 +218,30 @@ makeGraph(const std::vector<LayerRepresentation> &layer_reps) {
...
+  nntrainer::GraphRepresentation graph_rep;
+  auto model_graph = nntrainer::NetworkGraph();
+  for (auto &layer_representation : layer_reps) {
+    std::shared_ptr<nntrainer::LayerNode> layer = nntrainer::createLayerNode(
+      ac.createObject<nntrainer::Layer>(layer_representation.first),
+      layer_representation.second);
+    model_graph.addLayer(layer);
+  }
+  // compile with loss
+  model_graph.compile("mse");
```
lhs8928, line 234: What if user already added loss layer?

test/nntrainer_test_util.cpp
```diff
@@ -218,6 +218,30 @@ makeGraph(const std::vector<LayerRepresentation> &layer_reps) {
...
+
+  for (auto &node : model_graph.getLayerNodes()) {
+    graph_rep.push_back(node);
+  }
+
+  // remove loss layer
+  graph_rep.pop_back();
```
lhs8928, line 241: Can we guaranteed that only the last layer is loss layer? (what about the multiple loss layer?)

test/nntrainer_test_util.cpp
```diff
@@ -218,6 +218,30 @@ makeGraph(const std::vector<LayerRepresentation> &layer_reps) {
   return graph_rep;
 }
 
+nntrainer::GraphRepresentation
+makeGraph_V2(const std::vector<LayerRepresentation> &layer_reps) {
```
lhs8928, line 222: How about rename it to explicitly let users knows what this would do? Such as makeCompledGraph?

nntrainer/layers/layer_node.cpp
```diff
@@ -223,9 +223,10 @@ void LayerNode::setOutputConnection(unsigned nth, const std::string &name,
   }
 
   auto &con = output_connections[nth];
-  NNTR_THROW_IF(con, std::invalid_argument)
-    << "cannot override connection, this slot is reserved for "
-    << con->toString();
+  // Should be override connection for the batch normalization realizer
+  // NNTR_THROW_IF(con, std::invalid_argument)
```
lhs8928, line 227: Why don't we make a warning or make a another function such as resetOutputConnection?

test/nntrainer_test_util.cpp
jijoongmoon: no. It is only for the test case and It is the only way to define the loss layer is by the defining model property in normal case.

nntrainer/layers/layer_node.cpp
jijoongmoon: there are two situation we call the realizer from now on. one is for the training  and one for the inference. This kind of waring or error is the case for the training, not for the inference. For bn normalization case, the input of this bn realizer is graph representation which is already compiled and we need the modify or override the connection due to removed bn layer. In this case, it is not the error.

test/nntrainer_test_util.cpp
lhs8928: If it will be only used in unittest then I think reside in util.cpp is not a good way I guess?
I agreed that most of user will defining the loss by passing the loss type to model property but we allowed it to define by layer aren't we?

nntrainer/layers/layer_node.cpp
jijoongmoon: no. it is not. There might be another cases like bn realize which modifies connections.

test/nntrainer_test_util.cpp
jijoongmoon: Why it is not good? This is the code for the test only. I don't agree with you saying good way. Any case it needs compiled connection whose inputs do not follow the loss layer.. then it might be help. Besides, if the user define the loss with model property and set the loss layer.. it causes the error and the same is here.

lhs8928: Oh sorry. I was confused.
I think the file is util_func.cpp but it was in test_unit.cpp actually.
Most of my comments coming from this confusion. My bad.

lhs8928 (approved):
LGTM

PR #1883, opened 2022-04-20
[Wait for #1882] [ Interpreter ] add batch normalziation realizer @open sesame 04/26 10:37

jijoongmoon:
This patch removes the batch normalization layers with bn_realizer in
the GraphRepresentation for TfliteInterperter which is for the
inference.

**Self evaluation:**

### DIFF
--- a/nntrainer/compiler/bn_realizer.cpp
+++ b/nntrainer/compiler/bn_realizer.cpp
@@ -21,21 +21,61 @@
 #include <unordered_map>
 
 namespace nntrainer {
+
+static constexpr size_t SINGLE_INOUT_IDX = 0;
+
 BnRealizer::BnRealizer() {}
 
 BnRealizer::~BnRealizer() {}
 
 GraphRepresentation BnRealizer::realize(const GraphRepresentation &reference) {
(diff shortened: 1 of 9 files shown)


songgot (approved):
Looks good to me

nntrainer/compiler/bn_realizer.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <stdexcept>
+#include <unordered_map>
+
+namespace nntrainer {
+
+static constexpr size_t SINGLE_INOUT_IDX = 0;
+
+BnRealizer::BnRealizer() {}
+
+BnRealizer::~BnRealizer() {}
```
mhs4670go, line 29: Just curiosity, is there any reason why these default constructor and destructor are declared even though they are implicitly done?

jijoongmoon: No reason. It's just from the skeleton code. I will remove them in a later pr. Thanks.

mhs4670go: If it is just convention things, don't bother removing them:) Anyway looks good to me.

lhs8928 (approved):
LGTM

PR #1886, opened 2022-04-20
[Wait for #1883] [ TEST ] add more test cases for batch normalization realizer

jijoongmoon:
This patch adds the test case for batch normalization realizer with
kind of resnet basic block which includes the multiout layer.

**Self evaluation:**

### DIFF
--- a/test/nntrainer_test_util.cpp
+++ b/test/nntrainer_test_util.cpp
@@ -26,6 +26,7 @@
 #include <climits>
 #include <iostream>
 #include <layer_node.h>
+#include <multiout_realizer.h>
 #include <nntrainer_error.h>
 #include <random>
 #include <regex>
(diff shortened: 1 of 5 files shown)


songgot (approved):
Look good to me

test/unittest/compiler/unittest_realizer.cpp
```diff
@@ -779,33 +782,55 @@ TEST(BnRealizer, bn_realizer_p) {
...
+    {"conv2d", {"name=a2", "kernel_size=3,3", "input_layers=ac1"}},
+    {"conv2d", {"name=b1", "kernel_size=3,3", "input_layers=ac0"}},
+    {"addition", {"name=c1", "input_layers=a2,b1"}},
+    {"activation", {"name=ac2", "activation=relu", "input_layers=c1"}},
+    {"fully_connected", {"name=fc3", "input_layers=ac2"}},
   };
-  BnRealizer r({});
-  compileAndRealizeAndEqual(r, before, after);
+  std::vector<std::unique_ptr<nntrainer::GraphRealizer>> realizers;
+  realizers.emplace_back(new nntrainer::MultioutRealizer());
```
lhs8928, line 833: AFAIK MultioutRealizer is already included in compile. And I tested that without this realizer unittest pass the test. Correct me if I am wrong :)

jijoongmoon: no. it is not. it is done before compile.

test/nntrainer_test_util.cpp
```diff
@@ -218,6 +219,40 @@ makeGraph(const std::vector<LayerRepresentation> &layer_reps) {
...
+  for (auto &realizer : realizers) {
+    graph_rep = realizer->realize(graph_rep);
+  }
+
+  for (auto &layer : graph_rep) {
+    model_graph.addLayer(layer);
+  }
+
+  // Compile with loss
+  model_graph.compile("mse");
```
lhs8928, line 246: This would add mse loss layer to return graph representation

lhs8928: Again this was coming from my confusion

test/unittest/compiler/unittest_realizer.cpp
lhs8928: Got it. Thanks for your explanation

lhs8928 (approved):
LGTM

lhs8928 (approved):
LGTM

PR #1888, opened 2022-04-21
[Fix] Clean code for resnet application @open sesame 04/26 10:15

jihochu:
- fix typo
- Use const for catched exception
- use EXIT_SUCCESS macro
- use error code

### DIFF
--- a/Applications/Resnet/jni/main.cpp
+++ b/Applications/Resnet/jni/main.cpp
@@ -197,8 +197,8 @@ ModelHandle createResnet18() {
   ModelHandle model = ml::train::createModel(ml::train::ModelType::NEURAL_NET,
                                              {withKey("loss", "cross")});
 
-  for (auto layers : createResnet18Graph()) {
-    model->addLayer(layers);
+  for (auto layer : createResnet18Graph()) {
+    model->addLayer(layer);
   }
 
   return model;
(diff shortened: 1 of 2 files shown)


jijoongmoon:
Could check the https://github.com/nnstreamer/nntrainer/blob/main/docs/contributing.md and use clang-format?

Applications/Resnet/jni/main.cpp
```diff
@@ -27,6 +27,7 @@
 #include <optimizer.h>
 
 #include <cifar_dataloader.h>
+#include <nntrainer_error.h>
```
jijoongmoon, line 30: Examples like this resnet supposes to use ccapi which is open for the public. And unfortunately, nntrainer_error.h is not opened to the public and such defined macro such as ML_ERROR_NONE is not available here. We could consider if we can open nntrainer_error.h.. but for now, it might be better not to use them.

songgot (approved):
Looks good to me. but let's remove ML_ERROR_NONE.

Applications/Resnet/jni/main.cpp
jihochu: thanks. it modified.

myungjoo (approved):
LGTM

lhs8928 (approved):
LGTM

jijoongmoon:
because of the ci server url issue, you need the rebase the main branch which includes the fix.

jijoongmoon (commented):
LGTM

PR #1891, opened 2022-04-22
[model] add callback in train @open sesame 05/06 15:07

lhs8928:
Commit 1: [model] support stop training

     - During the training the function will check every epoch to decide stop
       the training or not.
    
Commit 2: [model] added getter for current epoch_idx

     - Through this function user can get epoch_idx
    
Commit 3:  [neuralnet] set epoch_idx to zero only if train from scratch

     - Until now, the epoch property had a meaning of how many epoch to run,
       but from now it will have a meaning of final epochs to run
    
### DIFF
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -141,12 +141,20 @@ public:
   /**
    * @brief     Run Model training and validation
    * @param[in] values hyper parameters
+   * @param[in] stop_cb callback function to decide stop training or not
+   * ~~~~~
+   * @a user_data user_data to be used in stop_cb
+   * @a bool true if stop the training
+   * ~~~~~
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    * @details   This function accepts vector of properties in the format -
    *  { std::string property_name, void * property_val, ...}
    */
-  virtual int train(const std::vector<std::string> &values = {}) = 0;
+  virtual int train(const std::vector<std::string> &values = {},
+                    std::function<bool(void *)> stop_cb = [](void *user_data) {
+                      return false;
+                    }) = 0;
 
   /**
    * @brief     Run Model train with callback function by user
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -268,6 +268,12 @@ public:
                          ml_train_summary_type_e verbosity) = 0;
 
   /**
+   * @brief     Get current epoch_idx
+   * @retval    epoch_idx
+   */
+  virtual unsigned int getCurrentEpoch() = 0;
+
+  /**
    * @brief     Get Loss
    * @retval    loss value
    */
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -613,7 +613,8 @@ int NeuralNetwork::deallocate() {
   return ML_ERROR_NONE;
 }
 
-int NeuralNetwork::train(const std::vector<std::string> &values) {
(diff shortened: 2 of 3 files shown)


nntrainer/models/neuralnet.h
```diff
@@ -251,10 +257,14 @@ class NeuralNetwork : public ml::train::Model {
   /**
    * @brief     Run NeuralNetwork train
    * @param[in] values hyper parameters
+   * @param[in] stop_cb callback function to decide stop training or not
```
myungjoo, line 260: Please describe the semantics of 
1. stop_cb's return
2. stop_cb's input parameter
in the header function doxygen.

api/ccapi/include/model.h
```diff
@@ -141,12 +141,16 @@ class Model {
   /**
    * @brief     Run Model training and validation
    * @param[in] values hyper parameters
+   * @param[in] stop_cb callback function to decide stop training or not
```
myungjoo, line 144: Describe the semantics of stop_cb's return and input argument here.

nntrainer/models/neuralnet.cpp
```diff
@@ -739,16 +740,19 @@ int NeuralNetwork::train_run() {
     stat.num_iterations++;
   };
 
-  auto train_epoch_end = [this](RunStats &stat, DataBuffer &buffer) {
+  auto train_epoch_end = [this, stop_cb](RunStats &stat, DataBuffer &buffer) {
```
kparichay, line 742: Food for thought: from the on-device perspective, the number of epochs can be few (maybe even 1), and user might want to stop the training after a few iterations.
It might be worthwhile to design the implementation to later support stop_cb at iteration level as well.

api/ccapi/include/model.h
```diff
@@ -141,12 +141,16 @@ class Model {
...
    * @retval #ML_ERROR_NONE Successful.
    * @retval #ML_ERROR_INVALID_PARAMETER invalid parameter.
    * @details   This function accepts vector of properties in the format -
    *  { std::string property_name, void * property_val, ...}
    */
-  virtual int train(const std::vector<std::string> &values = {}) = 0;
+  virtual int train(const std::vector<std::string> &values = {},
+                    std::function<bool(void *)> stop_cb = [](void *) {
+                      return false;
```
jijoongmoon, line 152: It might be better to use user callback with user data.

nntrainer/models/neuralnet.h
lhs8928: According to metal's reply on https://stackoverflow.com/questions/54483956/how-to-document-a-function-that-takes-a-callback, Doxygen doesn't support document the functor itself in a nested fashion. So I workaround using tildes. Please let me know if there is a better way :)

api/ccapi/include/model.h
lhs8928: Added

nntrainer/models/neuralnet.h
myungjoo: You can just describe it in normal human language. :)

myungjoo (commented):
I'm good with the desciption. LGTM.

jijoongmoon:
rebase is required to fix the ci.

nntrainer/models/neuralnet.cpp
jijoongmoon: how about leaving this idea as todo and also adding the early_stopping  support.

jijoongmoon (approved):
LGTM

jihochu (approved):
LGTM

songgot (approved):
LGTM

PR #1895, opened 2022-05-02
[ Unit Test ] remove mse loss layer  in makeCompiledGraph @open sesame 05/04 10:00

jijoongmoon:
This patch remove the mse loss layer in makeCompiledGraph

**Self evaluation:**

### DIFF
--- a/test/nntrainer_test_util.cpp
+++ b/test/nntrainer_test_util.cpp
@@ -242,8 +242,7 @@ nntrainer::GraphRepresentation makeCompiledGraph(
     model_graph.addLayer(layer);
   }
 
-  // Compile with loss
-  model_graph.compile("mse");
+  model_graph.compile("");
 
   graph_rep.clear();
   for (auto &node : model_graph.getLayerNodes()) {


test/unittest/compiler/unittest_interpreter.cpp
```diff
@@ -191,6 +191,7 @@ TEST(nntrainerInterpreterTflite, simple_fc) {
               << "failed, reason: " << strerror(errno);
   }
 }
+
 #endif
```
myungjoo, line 195: remove this change from the commit.

test/nntrainer_test_util.cpp
```diff
@@ -242,8 +242,7 @@ nntrainer::GraphRepresentation makeCompiledGraph(
     model_graph.addLayer(layer);
   }
 
-  // Compile with loss
-  model_graph.compile("mse");
+  model_graph.compile("");
```
myungjoo, line 245: Could you please describe the reason of this removal in the commit message?

If you just wanted to test the case without ```mse```, can you add another case without mse and keep this case? (If you've removed mse because it's "incorrect" to have mse here, it's ok.)

test/unittest/compiler/unittest_interpreter.cpp
jijoongmoon: I will. Thanks.

songgot (approved):
Looks good to me

lhs8928 (approved):
LGTM

Issue #1879, opened 2022-04-14
Exporter for TFLite 


jijoongmoon:
- NNTrainer needs to export the trained model to TfLite.
- Currently, we have PoC Code for Fullyconnected layer, but we need to developer tflite interpreter to support more layers.
- First target of this issue is serialize the Resnet 18 Model from NNTrainer to TfLite.
- To-Do lists are below,
    - [x] Move the hard-coded function to get the builtin options for Fully Connected #1877 
    - [x] Batch Normalization layer Remove Realizer #1883 
    - [ ] Conv2D Exporter
    - [ ] Check Channel-First & Channel-Last
    - [ ] Addition Exporter
    - [ ] Activation Layer ( Relu )
    - [ ] Export Resnet Block

mhs4670go:
While I'm drafting codes, I've found out things that need to be implemented.

https://github.com/nnstreamer/nntrainer/blob/71415f893217c4ff3c4d0d31a82610cf06cd67a7/nntrainer/compiler/tflite_opnode.cpp#L81-L89

To create flatbuffers `Operator`, `BuiltinOptions` needs to be created first and above codes does it. But to create each `BuiltinOptions`, some attributes are needed. For example, these are the codes that make `FullyConnectedOptions`.

```cxx
inline flatbuffers::Offset<FullyConnectedOptions> CreateFullyConnectedOptions(
    flatbuffers::FlatBufferBuilder &_fbb,
    tflite::ActivationFunctionType fused_activation_function = tflite::ActivationFunctionType_NONE,
    tflite::FullyConnectedOptionsWeightsFormat weights_format = tflite::FullyConnectedOptionsWeightsFormat_DEFAULT,
    bool keep_num_dims = false,
    bool asymmetric_quantize_inputs = false) {
  FullyConnectedOptionsBuilder builder_(_fbb);
  builder_.add_asymmetric_quantize_inputs(asymmetric_quantize_inputs);
  builder_.add_keep_num_dims(keep_num_dims);
  builder_.add_weights_format(weights_format);
  builder_.add_fused_activation_function(fused_activation_function);
  return builder_.Finish();
}
```

It means current codes should be changed to like..
```cxx
switch (op_type) { 
   case tflite::BuiltinOperator_FULLY_CONNECTED:
    return tflite::CreateFullyConnectedOptions(f, fused_activation_function, weights_format, keep_num_dims, asymmetric_quantize_inputs).Union();
[...]

PR #1902, opened 2022-05-12
[ Compiler ] add loss layer realizer

mhs4670go:
This patch includes loss realizer implementation and tests which removes the loss layer for exports.

**Self evaluation:**

Related: #1879 

### DIFF
--- a/nntrainer/compiler/loss_realizer.cpp
+++ b/nntrainer/compiler/loss_realizer.cpp
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2022 seongwoo <mhs4670go@naver.com>
+ *
+ * @file loss_realizer.h
+ * @date 4 May 2022
+ * @brief NNTrainer graph realizer which remove loss layer for inference
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author seongwoo <mhs4670go@naver.com>
+ * @bug No known bugs except for NYI items
+ */
+#include <algorithm>
+#include <cassert>
+#include <connection.h>
+#include <layer_node.h>
+#include <loss_realizer.h>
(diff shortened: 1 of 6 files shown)


nntrainer/compiler/loss_realizer.cpp
```diff
@@ -0,0 +1,74 @@
...
+  std::unordered_map<std::string, LayerNode *> existing_nodes;
+  std::vector<LayerNode *> loss_layers;
+
+  std::transform(
+    reference.begin(), reference.end(),
+    std::inserter(existing_nodes, existing_nodes.end()),
+    [](auto &node) { return std::pair(node->getName(), node.get()); });
+
+  for (auto &node : reference) {
+    if (loss_type.find(node->getType()) != loss_type.end()) {
```
lhs8928, line 44: Consider requireLabel function of layer.
According to the document of requireLabel, it means, for now, that it is endpoint of a graph(numOutlayers == 0).
So all of loss layer will return true for requireLabel

lhs8928 (approved):
LGTM

nntrainer/compiler/loss_realizer.cpp
mhs4670go: @lhs8928 Thank you for your information. The thing that you said is exactly what I wanted. But it will be one of assert condition because we should list supported loss layer rather than supporting all of them. Please refer to above @note section.
```c++
/// @note Some layers need to consider not removing all semantics.
/// For example, When CrossEntropySigmoidLossLayer needs to be removed,
/// sigmoid computation shouldn't be removed.
```

songgot (approved):
LGTM

jijoongmoon (approved):
The name of the realizer isn't quite fit for the purpose. Actually, it removes the loss layer. For now, it's ok, but plz consider change the name like rm_loss_realizer?

PR #1903, opened 2022-05-12
[layers] flatten layer consider target_shape

mhs4670go:
This patch makes a flatten layer consider target_shape and use it during exporting.

FYI, flatten layer exports to `Reshape` operator in tflite. That is why the flatten layer needs to know about target shape.

**Self evaluation:**

Related: #1879 

### DIFF
--- a/nntrainer/layers/flatten_layer.cpp
+++ b/nntrainer/layers/flatten_layer.cpp
@@ -23,19 +23,24 @@ namespace nntrainer {
 static constexpr size_t SINGLE_INOUT_IDX = 0;
 
 void FlattenLayer::finalize(InitLayerContext &context) {
-  ReshapeLayer::setProperty({"target_shape=-1"});
+  const TensorDim &in_dim = context.getInputDimensions()[0];
+
+  std::string target_shape =
+    "target_shape=1:1:" + std::to_string(in_dim.getFeatureLen());
+  ReshapeLayer::setProperty({target_shape});
+
   /** @note the output dimension is in invalid state till finalize of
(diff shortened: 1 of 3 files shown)


nntrainer/layers/flatten_layer.cpp
```diff
@@ -23,26 +23,33 @@ namespace nntrainer {
...
    * reshape_layer is finished */
   ReshapeLayer::finalize(context);
 
-  const TensorDim &in_dim = context.getInputDimensions()[0];
   if (in_dim.channel() == 1 && in_dim.height() == 1) {
     ml_logw("Warning: the flatten layer is redundant");
   }
 }
 
 void FlattenLayer::setProperty(const std::vector<std::string> &values) {
```
lhs8928, line 41: Why don't we just simply call Reshape::setProperty?

lhs8928 (approved):
LGTM

nntrainer/layers/flatten_layer.cpp
mhs4670go: @lhs8928 Umm.. I think 
1. if an error is occurred, we will be able to see where the error happened.
```c++
std::string msg = "[FlattenLayer] Unknown Layer Properties count " + /// please note `FlattenLayer`
                      std::to_string(values.size());
```
2. for extensibility. it would make it easy to extend some logic if it has its own codes in the first place.

Please let me know if I am wrong.

songgot (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1892, opened 2022-04-26
[DRAFT] Exporter for TFLite @open sesame 05/06 15:07

mhs4670go:
This draft is for TFLite Exporter.

Related: #1879 
Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #1904, opened 2022-05-12
[ ccapi ] Add exports interface

mhs4670go:
This patch adds "exports" interface and implementation of neuralnet which literally exports the model according to given export method.

**Self evaluation:**

Related: #1879 


nntrainer/models/neuralnet.cpp
```diff
@@ -1191,4 +1194,24 @@ void NeuralNetwork::forEachLayer(
...
+
+void NeuralNetwork::exports(nntrainer::ExportMethods method,
+                            const std::string file_path) {
+  switch (method) {
+  case nntrainer::ExportMethods::METHOD_TFLITE: {
+    nntrainer::TfliteInterpreter interpreter;
+
+    /// We will call "serialize" method for the model which is already trained
+    /// or allocated. So, we need to call deallocateTensors first to make sure
+    /// `dealloc_weights == false`
```
mhs4670go, line 1206: This comments are from https://github.com/nnstreamer/nntrainer/pull/1892#discussion_r868712315.

nntrainer/layers/layer_devel.h
```diff
@@ -31,14 +31,14 @@ namespace ml::train {
 class Layer;
 }
 
+enum class ExportMethods;
```
myungjoo, line 34: Why do you need to declare ```ExportMethods``` in different headers?

Can you define it in a single header file (e.g., ```nntrainer-api.common.h```) and include this file directly or indirectly from the required headers/sources?

myungjoo: Besides, as long as ExportMethods is no longer using "namespace", you don't need enum class. For such enum, ```typedef enum ...``` is enough.

myungjoo:
In the commit message body, please explain, why ExportMethod should be out of namespace.

api/ccapi/include/model.h
```diff
@@ -290,6 +290,14 @@ class Model {
    * @retval    loss value
    */
   virtual float getValidationLoss() = 0;
+
+  /**
+   * @brief     export the model according to given export method
```
jihochu, line 295: too many spaces..

myungjoo (approved):
LGTM

jijoongmoon (approved):
LGTM!

PR #1917, opened 2022-05-29
[trivial] fix typo @open sesame 06/09 22:09

lhs8928:
- Fix typo
 - Delete duplicated code

### DIFF
--- a/nntrainer/layers/layer_context.cpp
+++ b/nntrainer/layers/layer_context.cpp
@@ -473,7 +473,7 @@ Tensor &RunLayerContext::getLabel(unsigned int idx) {
 bool RunLayerContext::readyToUse() const {
   /**
    * assumption:
-   * 1. there must be atleast 1 input
+   * 1. there must be at least 1 input
    * 2. the setter set everything at once
    */
   if (inputs.empty())
--- a/nntrainer/optimizers/optimizer_wrapped.cpp
+++ b/nntrainer/optimizers/optimizer_wrapped.cpp
@@ -131,7 +131,7 @@ void OptimizerWrapped::setLearningRateScheduler(
   lr_sched = std::unique_ptr<nntrainer::LearningRateScheduler>(ptr);
 }
 
-nntrainer::LearningRateScheduler *OptimizerWrapped::setLearningRateScheduler() {
+nntrainer::LearningRateScheduler *OptimizerWrapped::getLearningRateScheduler() {
   return lr_sched.get();
 }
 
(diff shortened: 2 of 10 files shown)


nntrainer/layers/layer_context.cpp
```diff
@@ -65,7 +65,7 @@ void InitLayerContext::setOutputDimensions(
   std::vector<VarGradSpecV2> specs;
   specs.reserve(out_dim.size());
 
-  for (unsigned i = 0u, sz = out_dim.size(); i < sz; ++i) {
+  for (unsigned i = 0u; i < out_dim.size(); ++i) {
```
myungjoo, line 68: is calling std::vector<TensorDim>.size() repeatedly ok? (performance-wise)

songgot (approved):
Looks good to me

nntrainer/layers/layer_context.cpp
lhs8928: According to the cpp reference the complexity of size function is constant. So I guess it will be okay.
Please refer (https://en.cppreference.com/w/cpp/container/vector/size)

DonghakPark: Since the length can change inside the for loop, I think it's better to assign the length to a new variable for safety reasons.
In addition, i wonder in the code below, the length is changed through push_back, but wouldn't the result be different if sz is not defined?

```.cpp
  for (unsigned i = 0u, sz = out_specs.size(); i < sz; ++i) {
    auto &spec = out_specs.at(i);
    ....    
    ....
    output_specs.push_back(std::move(spec)); <--- change out_specs size 
  }
```

nntrainer/optimizers/optimizer_wrapped.cpp
```diff
@@ -131,7 +131,7 @@ void OptimizerWrapped::setLearningRateScheduler(
   lr_sched = std::unique_ptr<nntrainer::LearningRateScheduler>(ptr);
 }
 
-nntrainer::LearningRateScheduler *OptimizerWrapped::setLearningRateScheduler() {
+nntrainer::LearningRateScheduler *OptimizerWrapped::getLearningRateScheduler() {
```
jijoongmoon, line 134: nice catch!

jijoongmoon (approved):
LGTM

nntrainer/layers/layer_context.cpp
lhs8928: The out_dim is const value so I guess length will not change inside the for loop but I will assign the length to a new variable any way.

PR #1920, opened 2022-06-07
[ Packaging ] Packaging for Tizen 6.0

jijoongmoon:
This patch includes fixes to support Tizne 6.0 build
 . Fix .spec & meson.build

**Self evaluation:**

### DIFF
--- a/packaging/nntrainer.spec
+++ b/packaging/nntrainer.spec
@@ -46,6 +46,10 @@
 
 %endif # 0%{tizen_version_major}%{tizen_version_minor} >= 65
 
+%if 0%{tizen_version_major}%{tizen_version_minor} < 65
+BuildRequires: flatbuffers-devel
+%endif
+
 Name:		nntrainer
 Summary:	Software framework for training neural networks
 Version:	0.3.0
@@ -120,7 +124,9 @@ BuildRequires:	nnstreamer-devel
 %define enable_nnstreamer_tensor_filter -Denable-nnstreamer-tensor-filter=true
 
 %if 0%{?unit_test}
+%if 0%{tizen_version_major}%{tizen_version_minor} > 60
 BuildRequires:	nnstreamer-test-devel
+%endif
 BuildRequires:	gst-plugins-good-extra
 BuildRequires:	python
 %endif #unit_test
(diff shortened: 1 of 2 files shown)


lhs8928 (approved):
LGTM

packaging/nntrainer.spec
```diff
@@ -106,11 +110,19 @@ Requires:	nnstreamer-tensorflow2-lite
...
+%endif
 %endif # support_tflite_backbone
 
 %if 0%{?support_tflite_interpreter}
+%if 0%{tizen_version_major}%{tizen_version_minor} <= 60
+BuildRequires: tensorflow-lite-devel
+%else
 BuildRequires: tensorflow2-lite-devel
+%endif
 %endif # support_tflite_interpreter
```
myungjoo, line 126: Create a macro for Tizen < 6.5 (== Tizen <= 6.0).

```
%if 0%{?tizen_version_major}%{?tizen_version_minor} < 65
%define tizen65plus 0
%else
%define tizen65plus  1
%endif
```

Merge line 112-126.

```
%if 0%{?support_tflite_backbone} || 0%{?support_tflite_interpreter}
%if 0%{?tizen65plus}
BuildRequires: tensorflow2-lite-devel
%else
BuildRequires: tensorflow-lite-devel
%endif
%endif
```

meson.build
```diff
@@ -15,7 +15,12 @@ extra_defines = ['-DMIN_CPP_VERSION=201703L']
 cc = meson.get_compiler('c')
 cxx = meson.get_compiler('cpp')
 
+tizen_6_0 = false
```
myungjoo, line 18: Suggestion:

Add a meson_options with ```tflite_package_name``` with the default value of ```tensorflow2-lite```. And let ```tflite_dep = dependency('tensorflow-lite', required: false)``` use this option variable instead.

Then, with the minimal changes, you can control this at .spec file.

This removes the knowledge dependency on Tizen versions from meson code, reducing inter-module coupling! (better design)

test/meson.build
```diff
@@ -40,7 +40,7 @@ if enable_ccapi
   subdir('unittest')
 endif
 
-if get_option('enable-nnstreamer-tensor-filter')
+if get_option('enable-nnstreamer-tensor-filter') and not tizen_6_0
```
myungjoo, line 43: Is this because of the dependency on ```nnstreamer-test-devel``` package? If so, it appears that you need to check it with meson's dependency check (or by checking the file existance).

packaging/nntrainer.spec
jijoongmoon: this will be removed after tensorflow2 spec in tizen 6.0 is fixed.

meson.build
jijoongmoon: this will be removed after tensorflow2 spec in tizen 6.0 is fixed.

test/meson.build
jijoongmoon: this will be removed after tensorflow2 spec in tizen 6.0 is fixed.

songgot (approved):
LGTM

PR #1921, opened 2022-06-07
[bug_fix] bug fix zoneout lstmcell layer

lhs8928:
- This patch enable zoneout even though the zoneout rate is smaller than epsilon


jijoongmoon (approved):
LGTM

PR #1932, opened 2022-06-15
[Spec] Add flatbuffers-devel BuildRequires as default

again4you:
The compiler module always uses flatc command but 'flatbuffers-devel'
dependency is added in case of the Tizen version is lower than 6.5.
Because of this reason, the buildbreak issue occurs when building for
Tizen 7.0. This patch fixes this bug.

Signed-off-by: Sangjung Woo <sangjung.woo@samsung.com>

### DIFF
--- a/packaging/nntrainer.spec
+++ b/packaging/nntrainer.spec
@@ -46,10 +46,6 @@
 
 %endif # 0%{tizen_version_major}%{tizen_version_minor} >= 65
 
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM!

DonghakPark (approved):
LGTM!

lhs8928 (approved):
LGTM

PR #1942, opened 2022-06-21
[neuralnet] check tflite interpreter is enabled when export

lhs8928:
- Added ifdef statement to check tflite interpreter is enabled
 - Added override specifier in get/setWeights

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -1203,6 +1203,7 @@ void NeuralNetwork::exports(const ml::train::ExportMethods &method,
                             const std::string file_path) {
   switch (method) {
   case ml::train::ExportMethods::METHOD_TFLITE: {
+#ifdef ENABLE_TFLITE_INTERPRETER
     nntrainer::TfliteInterpreter interpreter;
 
     /// We will call "serialize" method for the model which is already trained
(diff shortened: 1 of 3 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -1203,16 +1203,20 @@ void NeuralNetwork::exports(const ml::train::ExportMethods &method,
...
+
+      /// We will call "serialize" method for the model which is already trained
+      /// or allocated. So, we need to call deallocateTensors first to make sure
+      /// `dealloc_weights == false`
+      model_graph.deallocateTensors();
+      model_graph.allocateTensors(ExecutionMode::INFERENCE);
+      interpreter.serialize(graph_representation, file_path);
+      model_graph.deallocateTensors();
+    #else
+      throw std::runtime_error{"ENABLE_TFLITE_INTERPRETER is not defined"};
```
songgot, line 1217: the log indicates how to correct the error.
I think that app set 'METHOD_TFLITE' and  it is defined in enum class (common.h) as below
enum class ExportMethods {
  METHOD_STRINGVECTOR = 0, /**< export to a string vector */
  METHOD_TFLITE = 1,       /**< epxort to tflite */
  METHOD_UNDEFINED = 999,  /**< undefined */
};
If we focus on 'METHOD_TFLITE', how about change error log?
```suggestion
throw std::runtime_error{"METHOD_TFLITE is not supported"};
throw std::runtime_error{"METHOD_TFLITE is not supported , please define ENABLE_TFLITE_INTERPRETER"};
```

DonghakPark (approved):
LGTM!

nntrainer/models/neuralnet.cpp
lhs8928: Done :)

songgot (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1946, opened 2022-06-28
[conv] support dilation property

### LINKED ISSUE #1922: Conv1D, Conv2D Dilation Support
NNTrainer needs to support the Dilation option for conv1d and conv2d.

lhs8928:
- Support dilation property in conv1d/conv2d layer
 - Added unittest with dilation

Close #1922


jijoongmoon (approved):
LGTM. Now we support dilation in convolution layers completely. Please check if it is ok if we set the stride and dilate during padding. If it is ok, then I think it is good to be merged. After this pr, it might be better if we have an application example with dilation. Wavenet with conv1d might be great.

Simple implementation of wavenet with keras is
``` python
for dilation in (1,2,4,8,16):
   Conv1D(filters=20, kernel_size=3, padding="casual", dilation_rate=rate)
```
Of course, casual padding must be supported. (https://stopspoon.tistory.com/48, https://paperswithcode.com/method/dilated-causal-convolution)

jihochu (approved):
LGTM

songgot (approved):
LGTM

DonghakPark (approved):
LGTM!

PR #1949, opened 2022-06-30
[BUILD] fix supported gmock version

jihochu:
google mock API is changed from 1.10.0,
and it does not support MOCK_METHOD under 1.10.0 version.
'profiler' test is only enabled for proper gmock.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -262,7 +262,7 @@ if get_option('enable-tflite-interpreter')
   extra_defines += '-DENABLE_TFLITE_INTERPRETER=1'
 endif
 
-gmock_dep = dependency('gmock', static: true, required: false)
+gmock_dep = dependency('gmock', static: true, main: false, required: false)
 gtest_dep = dependency('gtest', static: true, main: false, required: false)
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -262,7 +262,7 @@ if get_option('enable-tflite-interpreter')
   extra_defines += '-DENABLE_TFLITE_INTERPRETER=1'
 endif
 
-gmock_dep = dependency('gmock', static: true, required: false)
+gmock_dep = dependency('gmock', version : '>=1.10.0', static: true, required: false)
```
jijoongmoon, line 265: How about if gmock_dep is false, then skip the unittest for profile in line 54 at test/unittest/meson.build
``` bash
  if get_option('enable-profile') and gmock_dep
     test_target += [['unittest_nntrainer_profiler', []]] 
  endif
```

jihochu: thanks. it changed to skip only profile test.

DonghakPark (approved):
LGTM!

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #1951, opened 2022-07-06
[Wait for #1946] Support causal padding in conv1d

### LINKED ISSUE #1947: Support 'padding=casual'
We need support padding=casual. It could be done by aliasing padding with an unsigned int numbers

lhs8928:
- Replace padding2d to padding1d in conv1d
     - Enable causal property in conv1d
     - Added unittest for causal padding in conv1d
    
    Close #1947
    

nntrainer/layers/common_properties.cpp
```diff
@@ -220,17 +220,47 @@ bool Padding1D::isValid(const std::string &v) const {
...
   if (istrequal(padding_repr, "valid")) {
     return {0, 0};
+  } else if (istrequal(padding_repr, "same")) {
+
+    auto pad_horizontal =
+      calculate_padding(input_dim.width(), kernel, stride, dilation);
+
+    auto pad_left = pad_horizontal / 2;
+
+    return {pad_left, pad_horizontal - pad_left};
```
jihochu, line 247: pad_horizontal - pad_left = pad_horizontal - (pad_horizontal / 2) = pad_horizontal / 2 = pad_left
so,,,
return {pad_left, pad_left} ??

lhs8928: That's right if the pad_horizontal is even number.
If the pad_horizontal is odd number, it will return {pad, left, pad_left + 1}

jihochu: thanks for checking.

nntrainer/layers/common_properties.cpp
```diff
@@ -197,15 +197,15 @@ Padding2D::compute(const TensorDim &input, const TensorDim &kernel,
 
 bool Padding1D::isValid(const std::string &v) const {
 
-  /// case 1, 2: padding has string literal
-  if (istrequal(v, "valid") || istrequal(v, "same")) {
+  /// case 1, 2, 3: padding has string literal
+  if (istrequal(v, "valid") || istrequal(v, "same") || istrequal(v, "causal")) {
```
jijoongmoon, line 201: I wonder if the casual option only works in 1D.

nntrainer/layers/conv1d_layer.cpp
```diff
@@ -57,17 +68,10 @@ void Conv1DLayer::finalize(InitLayerContext &context) {
...
-                "1," +
-                  std::to_string(std::get<props::Stride>(conv_props).get()));
-  setPropertyKV(props::Padding2D::key,
-                std::get<props::Padding2D>(conv_props).get());
-  setPropertyKV(props::Dilation::key,
-                "1," +
-                  std::to_string(std::get<props::Dilation>(conv_props).get()));
+  setPropertyKV(props::KernelSize::key, "1," + std::to_string(kernel_size));
+  setPropertyKV(props::Stride::key, "1," + std::to_string(stride));
+  setPropertyKV(props::Padding2D::key, padding_str);
```
jijoongmoon, line 73: I wonder if it is right. It does not consistent with line 103. Here, property is Padding2D, but line 103, it is Padding1D. Please check once again.

nntrainer/layers/common_properties.cpp
lhs8928: According to the tensorflow con2d documentation, one of "valid" or "same" is available whreas "valid", "same" or "causal" is available in conv1d.
Please refer padding section in conv1d, conv2d
(conv1d: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Conv1D, conv2d: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Conv2D)

FYI in pytorch either a string {"valid", "same"} or tuple of int value is available in conv1d ("causal" is not available even in conv1d).

nntrainer/layers/conv1d_layer.cpp
lhs8928: For such issues described in https://github.com/nnstreamer/nntrainer/pull/1951#discussion_r917596381, conv1d needs Padding1D, not the Padding2D. (That is the reason for Padding1D at line 103)

But for now conv1d uses conv2d layer so we need to convert Padding1D property to Padding2D property. (line 73)

nntrainer/layers/conv1d_layer.cpp
```diff
@@ -30,7 +30,6 @@ static constexpr size_t SINGLE_INOUT_IDX = 0;
 
 Conv1DLayer::Conv1DLayer(const std::array<unsigned int, 2> &padding_) :
   LayerImpl(),
-  padding(padding_),
```
jijoongmoon, line 33: if it is not used, then we need to remove the `padding_ ` in param.

nntrainer/layers/common_properties.cpp
jijoongmoon: right. thanks.

nntrainer/layers/conv1d_layer.cpp
jijoongmoon: right. thanks.

nntrainer/layers/conv1d_layer.cpp
lhs8928: Fixed

jijoongmoon (approved):
LGTM

songgot (approved):
LGTM

jihochu (approved):
LGTM

PR #1956, opened 2022-07-13
[multi head attention] Added multi head attention scaffold @open sesame 8/22 13:32

lhs8928:
- Added calcCommonDerivative
     - Added finalizeCommon
     - Replace torch_ref property to reference_framework
    
### DIFF
--- a/nntrainer/layers/common_properties.cpp
+++ b/nntrainer/layers/common_properties.cpp
@@ -333,6 +333,20 @@ void GenericShape::set(const TensorDim &value) {
   Property<TensorDim>::set(ret);
 }
 
+NumHeads::NumHeads(unsigned int value) { set(value); }
+
+ProvideAttentionMask::ProvideAttentionMask(bool value) { set(value); }
+
+ReturnAttentionWeight::ReturnAttentionWeight(const std::string &value) {
+  set(value);
+}
+bool ReturnAttentionWeight::isValid(const std::string &value) const {
+  if (value == "" || value == "before" || value == "after") {
+    return true;
+  }
+  return false;
+}
+
 } // namespace props
 
 template <>
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -1118,6 +1118,119 @@ public:
 };
 
 /**
+ * @brief NumHeads property, NumHeads is number of head in multi head attention
+ *
+ */
+class NumHeads : public PositiveIntegerProperty {
+public:
+  /**
+   * @brief Construct a new NumHeads object with default value 1
+   *
+   */
+  NumHeads(unsigned int value = 1);
+  static constexpr const char *key = "num_heads"; /**< unique key to access */
+  using prop_tag = uint_prop_tag;                 /**< property type */
+};
+
+/**
+ * @brief ProjectedKeyDim property, projected key dim per head in multi head
+ * attention
+ * @details Correspond with key_dim of tensorflow
+ *
+ */
+class ProjectedKeyDim : public PositiveIntegerProperty {
+public:
+  static constexpr const char *key =
+    "projected_key_dim";          /**< unique key to access */
+  using prop_tag = uint_prop_tag; /**< property type */
+};
+
+/**
+ * @brief ProjectedValueDim property, projected value dim per head in multi head
+ * attention
+ * @details Correspond with value_dim of tensorflow
+ *
+ */
+class ProjectedValueDim : public PositiveIntegerProperty {
+public:
+  static constexpr const char *key =
+    "projected_value_dim";        /**< unique key to access */
+  using prop_tag = uint_prop_tag; /**< property type */
+};
+
+/**
+ * @brief OutputShape property, output shape of multi head
+ * attention
+ * @details Correspond with output_shape of tensorflow
+ *
+ */
+class OutputShape : public PositiveIntegerProperty {
+public:
+  static constexpr const char *key =
+    "output_shape";               /**< unique key to access */
+  using prop_tag = uint_prop_tag; /**< property type */
+};
+
+/**
+ * @brief ProvideAttentionMask, attention mask is provided or not
+ *
+ */
+class ProvideAttentionMask : public Property<bool> {
+public:
+  /**
+   * @brief Construct a new ProvideAttentionMask object
+   *
+   */
+  ProvideAttentionMask(bool val = false);
+  static constexpr const char *key =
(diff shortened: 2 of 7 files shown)


nntrainer/layers/multi_head_attention_layer.h
```diff
@@ -0,0 +1,143 @@
...
+
+#include <acti_func.h>
+#include <layer_impl.h>
+
+namespace nntrainer {
+
+/**
+ * @class   Multi Head Attention Layer
+ * @brief   Implementation of multi head attention which is described in paper
+ * "Attention is all you need"
```
jihochu, line 27: How about adding url for paper? https://arxiv.org/abs/1706.03762

nntrainer/layers/multi_head_attention_layer.h
```diff
@@ -0,0 +1,143 @@
...
+public:
+  /**
+   * @brief     Constructor of MultiHeadAttention Layer
+   */
+  MultiHeadAttentionLayer();
+
+  /**
+   * @brief     Destructor of MultiHeadAttention Layer
+   */
+  ~MultiHeadAttentionLayer();
```
jihochu, line 39: If MultiHeadAttentionLayer is not final layer, it's better to make destructor to virtual method.

nntrainer/layers/multi_head_attention_layer.h
lhs8928: Already described at line 8

nntrainer/layers/multi_head_attention_layer.h
lhs8928: I guess MultiHeadAttentionLayer would be the final layer

nntrainer/layers/common_properties.h
```diff
@@ -80,6 +80,27 @@ class Name : public nntrainer::Property<std::string> {
   bool isValid(const std::string &v) const override;
 };
 
+/**
+ * @brief     Enumeration of reference framework
+ */
+struct ReferenceFrameworkInfo {
+  enum class Enum { pytorch, tensorflow };
```
jijoongmoon, line 87: I think `torch_ref` property of adam optimizer needs to be fixed to be compatible with it. Also need to check if it is ok to change the this signature in API as well.

nntrainer/layers/common_properties.h
```diff
@@ -1115,6 +1136,138 @@ class MoL_K : public PositiveIntegerProperty {
...
+  static constexpr const char *key = "value_dim"; /**< unique key to access */
+  using prop_tag = uint_prop_tag;                 /**< property type */
+};
+
+/**
+ * @brief ProjectedKeyDim property, projected key dim in multi head attention
+ * @details Correspond with key_dim of tensorflow
+ *
+ */
+class ProjectedKeyDim : public PositiveIntegerProperty {
```
jijoongmoon, line 1192: I wonder what is the difference betwwen `KeyDim` and `ProjectedKeyDim`?

lhs8928: It is not a alias and used for actual different purpose.
In multi head attention layer each input(query, key, value) will be projected by 3 different fc layer.(query_fc, key_fc, value_fc)
The KeyDim means input key width dimension which also means input size of key_fc layer.
The projectedKeyDim means projected key width dimension which also means unit size of key_fc layer.

There is 2 different algorithm in multi head attention layer of torch/tensorflow.
1. If the return_attention_weight is set tensorflow return attention weight before applying dropout whereas pytorch will return the attention weight which is already applied the dropout.
2. The true/false value behaves differently in attention mask. In tensorflow 1 indicates attention and 0 indicates no attention whereas 1 indicates no attention and 0 indicates attention in pytorch

nntrainer/layers/common_properties.h
lhs8928: In this commit torch_ref property of adam optimizer already compatible with ReferenceFrameworkInfo.
Let me check is it okay to change the API.

nntrainer/layers/common_properties.h
jijoongmoon: Like we discussted off-line, nntrainer does not need to work as other frameworks. The compatibility with other frameworks can be considered in the model loader and if it is possible to convert them in model loader level, then we don't need to implement like this. Optimizer Adam is different. It is the case which is different the algorithm itself.

lhs8928: Done

nntrainer/layers/common_properties.cpp
```diff
@@ -333,6 +333,17 @@ void GenericShape::set(const TensorDim &value) {
...
 }
 
+NumHeads::NumHeads(unsigned int value) { set(value); }
+
+ProvideAttentionMask::ProvideAttentionMask(bool value) { set(value); }
+
+ReturnAttentionWeight::ReturnAttentionWeight(bool value) { set(value); }
+
+ReturnAttentionWeightBeforeDropout::ReturnAttentionWeightBeforeDropout(
+  bool value) {
```
jijoongmoon, line 343: What if those two properties set at the same time?

lhs8928: Discussed offline to use only ReturnAttentionWeight property.
ReturnAttentionWeight will be string type property which will have one of "empty", "before", "after" value.

jijoongmoon (approved):
LGTM. One thing small thing is, we need to add more detail explanation about `ReturnAttentionWeight` Property. Such as, `"before"`  means `"return weight before mask is applied."`

lhs8928:
> LGTM. One thing small thing is, we need to add more detail explanation about `ReturnAttentionWeight` Property. Such as, `"before"` means `"return weight before mask is applied."`

I describe it in https://github.com/nnstreamer/nntrainer/pull/1956/files#diff-f94da3514caf5dd909901e06dd23b5e75b237b9cf702d288862cd0918ccff7e7R1192

DonghakPark (approved):
LGTM

songgot (approved):
LGTM

jijoongmoon:
> > LGTM. One thing small thing is, we need to add more detail explanation about `ReturnAttentionWeight` Property. Such as, `"before"` means `"return weight before mask is applied."`
> 
> I describe it in https://github.com/nnstreamer/nntrainer/pull/1956/files#diff-f94da3514caf5dd909901e06dd23b5e75b237b9cf702d288862cd0918ccff7e7R1192

What I mean is for the APIs for user's documentation.

PR #1963, opened 2022-07-19
[multi_head_attention] implement calcDerivative, calcGradient

lhs8928:
- Implement multi head attention calcDerivative, calcGradient
     - Needs to support attention mask
    

jijoongmoon (approved):
I think most of all looks good to me. One think I'd like mention is that plz consider splitting the layers later.

songgot (approved):
LGTM

jihochu (approved):
LGTM

PR #1978, opened 2022-08-11
[Application] Add profile for VGG

jijoongmoon:
This patch adds the profile in VGG Application

**Self evaluation:**

### DIFF
--- a/Applications/VGG/jni/main.cpp
+++ b/Applications/VGG/jni/main.cpp
@@ -20,6 +20,7 @@
 
 #include <cifar_dataloader.h>
 #include <model.h>
+#include <profiler.h>
 
 /**
  * @brief     Data size for each category
@@ -86,6 +87,13 @@ int main(int argc, char *argv[]) {
     std::cout << "./nntrainer_vgg vgg.ini resource\n";
     exit(-1);
   }
+
+  std::shared_ptr<nntrainer::profile::GenericProfileListener> listener;
+
+#ifdef PROFILE
+  listener = std::make_shared<nntrainer::profile::GenericProfileListener>();
+#endif
+  PROFILE_BEGIN(listener);
 
   seed = time(NULL);
   srand(seed);
@@ -182,6 +190,8 @@ int main(int argc, char *argv[]) {
     std::cerr << "Error during train: " << e.what() << std::endl;
     return 1;
   }
+
+  PROFILE_END(listener);
 
   return 0;
 }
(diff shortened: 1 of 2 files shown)


Applications/VGG/jni/main.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include <cifar_dataloader.h>
 #include <model.h>
+#ifdef PROFILE
+#include <profiler.h> // disable including this in android as profiler is not exposed yet to devel header
+#endif
```
myungjoo, line 25: You may have
```
#ifdef PROFILE
#define PROFILE_BEGIN auto listener = std::make....
#define PROFILE_END std::cout << *listener;
#else
#define PROFILE_BEGIN do { } while (0)
#define PROFILE_END do { } while (0)
#endif
```
in ```profiler.h``` and remove #ifdef/#endif from .cpp files.

jijoongmoon: we might need `listener in BEGIN and END as an parameter. That means that we need the #ifdef as well.
``` c++
#ifdef PROFILE
auto listener = std::make_shared<nntrainer::profile::GenericProfileListener>();
PROFILE_BEGIN(listener);
#endif
....
PROFILE_END(listener);
```
Otherwise, there is compile error saying "listener is not defined" in PROFILE_END.

DonghakPark (approved):
LGTM

jihochu (approved):
LGTM

Applications/VGG/jni/main.cpp
```diff
@@ -87,6 +88,12 @@ int main(int argc, char *argv[]) {
     exit(-1);
   }
 
+#ifdef PROFILE
+  auto listener =
+    std::make_shared<nntrainer::profile::GenericProfileListener>();
+  PROFILE_BEGIN(listener);
```
jihochu, line 94: I'm not sure,
But removing PROFILE definition check would be better.
We only need to waste making used shared object In non-using PROFILE case, and get clean code benefit.

or,,,

```
std::shard_ptr<nntrainer::profile::ProfileListener> listener;

#ifdef PROFILE
listener = std::make_shared<nntrainer::profile::GenericProfileListener>();
#endif
```

songgot (approved):
LGTM

Applications/VGG/jni/main.cpp
jijoongmoon: done. Thanks.

PR #1980, opened 2022-08-11
[ LSTM ] Optimize LSTM Gradient calculation

jijoongmoon:
Gradient computaion of LSTM takes over 60% of computation. This patch
includes the optimization using the tensor dimenson which only has
width.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/lstmcell_core.cpp
+++ b/nntrainer/layers/lstmcell_core.cpp
@@ -15,6 +15,10 @@
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 
+#ifdef USE_BLAS
+#include <cblas.h>
+#endif
+
 namespace nntrainer {
 
 void lstmcell_forwarding(const unsigned int batch_size, const unsigned int unit,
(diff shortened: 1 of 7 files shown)


nntrainer/layers/lstmcell_core.cpp
```diff
@@ -125,8 +125,40 @@ void lstmcell_calcGradient(
...
+    for (unsigned int i = 0; i < d_weight_hh.height(); ++i) {
+      unsigned int out_width = d_weight_hh.width();
+      float in_hh = prev_hidden_state.getValue(i);
+
+      float *d_weight_hh_address = d_weight_hh.getAddress(i * out_width);
+
+      float *d_ifgo_address = d_ifgo.getData();
+
+      for (unsigned int j = 0; j < out_width; ++j) {
+        d_weight_hh_address[j] += d_ifgo_address[j] * in_hh;
```
DonghakPark, line 157: wouldn't it be possible to optimize performance more by using ```loop rolling```?

DonghakPark (approved):
LGTM

nntrainer/layers/lstmcell_core.cpp
jijoongmoon: I change it to use BLAS Lib. Thanks.

songgot (approved):
LGTM

PR #1981, opened 2022-08-11
[ Tensor ] add neon support for ARM Architecture 

jijoongmoon:
This patch includes implementation of sgemv function with ARM NEON.
Later, more arm neon optimized function will be added

**Self evaluation:**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -154,6 +154,10 @@ endif
 if get_option('enable-blas')
   extra_defines += '-DUSE_BLAS=1'
 
+  if get_option('enable-neon')
+     extra_defines += '-DUSE_NEON=1'
+  endif
+
   if get_option('platform') == 'android'
     message('preparing blas')
     run_command(meson.source_root() / 'jni' / 'prepare_openblas.sh', meson.build_root(), check: true)
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -14,6 +14,10 @@
 #include <blas_interface.h>
 #include <nntrainer_error.h>
 
+#ifdef USE_NEON
+#include <blas_neon.h>
+#endif
+
 #include <cmath>
 
 #define sgemv_loop(ci, cj, cM, cN)           \
@@ -246,6 +250,19 @@ void sgemv(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, const unsigned int M,
 #ifdef USE_BLAS
 #ifdef BLAS_NUM_THREADS
   openblas_set_num_threads(BLAS_NUM_THREADS);
+#endif
+
+#ifdef USE_NEON
+  if (TransA == CblasNoTrans) {
+    if (incX == 1 && incY == 1 && (N % 16 == 0 || N % 8 == 0 || N % 4 == 0)) {
+      return nntrainer::neon::sgemv_neon(A, X, Y, M, N, alpha, beta);
+    }
+  } else if (TransA == CblasTrans) {
+    if (incX == 1 && incY == 1 && (N % 16 == 0 || N % 8 == 0 || N % 4 == 0)) {
+      return nntrainer::neon::sgemv_transpose_neon(A, X, Y, M, N, alpha, beta);
+    }
+  }
+
 #endif
   return cblas_sgemv(order, TransA, M, N, alpha, A, lda, X, incX, beta, Y,
                      incY);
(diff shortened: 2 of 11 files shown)


meson.build
```diff
@@ -139,6 +139,10 @@ endif
 if get_option('enable-blas')
   extra_defines += '-DUSE_BLAS=1'
 
+  if get_option('enable-neon')
+     extra_defines += '-DUSE_NEON=1'
+  endif
+
```
myungjoo, line 145: You need to check the availability of neon (and others, including blas, too) at meson, before actually compiling the code.

Then, you can use "feature" instead of "boolean" for actual features (such as neon, blas, ...), which allows users to have "auto" options.

nntrainer/tensor/blas_interface.cpp
```diff
@@ -246,6 +250,19 @@ void sgemv(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, const unsigned int M,
 #ifdef USE_BLAS
 #ifdef BLAS_NUM_THREADS
   openblas_set_num_threads(BLAS_NUM_THREADS);
+#endif
+
+#ifdef USE_NEON
+  if (TransA == CblasNoTrans) {
+    if (incX == 1 && incY == 1 && (N % 16 == 0 || N % 8 == 0 || N % 4 == 0)) {
```
lhs8928, line 257: The (N % 16) and (N % 8) cases are included in (N % 4)

meson.build
```diff
@@ -139,6 +139,10 @@ endif
 if get_option('enable-blas')
   extra_defines += '-DUSE_BLAS=1'
 
+  if get_option('enable-neon')
```
lhs8928, line 142: If we are going to use neon in arm architecture only then we need to check that platform is android

nntrainer/tensor/blas_interface.cpp
jijoongmoon: Yes. but if 16 case is checked first and if it is done, then it returns without go further.

meson.build
jijoongmoon: neon is only available in arm architecture. it also should work in tizen platform in raspberry pi.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -465,7 +466,7 @@ void Conv2DLayer::calcDerivative(RunLayerContext &context) {
   /// col2im(column matrix) to reconstruct the original image
 
   auto compute_derivative = [&](unsigned int s, unsigned int e,
-                                void *user_data) {
+                                unsigned int pid, void *user_data) {
```
jihochu, line 469: Added **pid** is not used.

nntrainer/tensor/blas_neon.cpp
```diff
@@ -0,0 +1,267 @@
...
+        y8_11 = vmlaq_f32(y8_11, wvec8_11, x);
+        y12_15 = vmlaq_f32(y12_15, wvec12_15, x);
+
+        vst1q_f32(&y[0], y0_3);
+        vst1q_f32(&y[4], y4_7);
+        vst1q_f32(&y[8], y8_11);
+        vst1q_f32(&y[12], y12_15);
+      }
+    }
+    return;
```
jihochu, line 192: It seems to be unnecessary for `return` for all if else statement

jijoongmoon: I think it needs. %16 == 0 also true for %8 and %4.

meson.build
jijoongmoon: For blas lib, we check the openblase dependency like `dependency('openblas')` and if it is not true, then return the error  before compiling. If is the openblas is false in meson_options.txt, then we just use numeric op in nntrainer itself. And for the ARM or android platform, we download the precompiled blas lib for arm from web and use it. 

However, we might need to check the target architecture and check if it supports neon (and float16 op support). 
( Moreover, openblas is necessary for the neon option case, because neon implementation only works for the division by 4 case. If the matrix is not divided by 4, then we need to run with cblas lib. This will be improved soon and then we could separate the dependency with blas )

Considering all the cases in nntrainer, I will make separate pr for the feature and auto options in meson script.

songgot (approved):
Looks good to me

nntrainer/tensor/blas_neon.cpp
jihochu: I mean it can return naturally in line 264, and 'return' in line 264 seems unnecessary, too.

SeoHyungjun:
LGTM!

DonghakPark (approved):
LGTM!

myungjoo:
What is "#cache_pool#.cpp"?

PR #1989, opened 2022-08-25
[positional encoding] Implement positional encoding layer

lhs8928:
Commit1:     [positional encoding] implement positional encoding layer
    
     - Positional encoding just needed to be calculated only once
       so make its lifespan as max lifespan
    
Commi2:     [unittest] generate positional encoding unittest
    
     - Generate positional encoding layer/model unittest.
    

nntrainer/layers/positional_encoding_layer.cpp
```diff
@@ -0,0 +1,120 @@
...
+  input.add(pe_partial, output);
+}
+
+void PositionalEncodingLayer::calcDerivative(RunLayerContext &context) {
+  const nntrainer::Tensor &incoming_derivative =
+    context.getIncomingDerivative(SINGLE_INOUT_IDX);
+  nntrainer::Tensor &outgoing_derivative =
+    context.getOutgoingDerivative(SINGLE_INOUT_IDX);
+
+  outgoing_derivative.copyData(incoming_derivative);
```
jijoongmoon, line 78: This Positional Encoding Layer is not trainable. then.

nntrainer/layers/positional_encoding_layer.cpp
```diff
@@ -0,0 +1,120 @@
...
+  unsigned int max_token_size =
+    std::get<props::MaxTimestep>(positional_encoding_props);
+
+  unsigned int model_dim = context.getInput(SINGLE_INOUT_IDX).getDim().width();
+
+  nntrainer::Tensor &pe = context.getTensor(
+    weight_idx[PositionalEncodingParams::positional_encoding]);
+
+  float value;
+  for (unsigned int i = 0; i < max_token_size; ++i) {
```
jijoongmoon, line 92: It might be better to use std::transform to accelerate.

jijoongmoon (approved):
overall, LGTM

songgot (approved):
LGTM

nntrainer/layers/positional_encoding_layer.cpp
lhs8928: Exactly

nntrainer/layers/positional_encoding_layer.cpp
lhs8928: To calculate positional encoding value, it needs position, index information which is i, j in this for loop.
However in std::transform there is no way to get this information.
Please correct me if I'm wrong.

jihochu (approved):
LGTM

PR #1994, opened 2022-09-02
[TF Exporter] FullyConnected Layer Weights Transpose for TFLite Export @open sesame 10/28 10:14

DonghakPark:
This PR for TFlite export
Transpose FullyConnected Layer's weights (NCHW --> NHWC)

in shape 1:3:2:4 & FC with unit 1

Before reordering 
0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23

After reordering
0 8 16 1 9 17 2 10 18 3 11 19 4 12 20 5 13 21 6 14 22 7 15 23

The Result of ```input - flatten - FC``` predicted as 
sum(W<sub>i</sub> *  X<sub>i</sub>) + bias ==> ```4325``` and the ouput also ```4325```

Related : #1912
Close : #1973 
Signed-off-by: DonghakPark <donghak.park@samsung.com>

### DIFF
--- a/Applications/Resnet/jni/main.cpp
+++ b/Applications/Resnet/jni/main.cpp
@@ -195,7 +195,7 @@ std::vector<LayerHandle> createResnet18Graph() {
 /// @todo update createResnet18 to be more generic
 ModelHandle createResnet18() {
   ModelHandle model = ml::train::createModel(ml::train::ModelType::NEURAL_NET,
-                                             {withKey("loss", "cross")});
+                                             {withKey("loss", "mse")});
 
   for (auto layer : createResnet18Graph()) {
     model->addLayer(layer);
@@ -258,6 +258,9 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
                     std::move(dataset_valid));
 
   model->train();
+
+  model->exports(ml::train::ExportMethods::METHOD_TFLITE, "resnet_test.tflite");
+
 #if defined(ENABLE_TEST)
   training_loss = model->getTrainingLoss();
   validation_loss = model->getValidationLoss();
--- a/Applications/Resnet/jni/main.cpp
+++ b/Applications/Resnet/jni/main.cpp
@@ -194,8 +194,14 @@ std::vector<LayerHandle> createResnet18Graph() {
 
 /// @todo update createResnet18 to be more generic
 ModelHandle createResnet18() {
+/// @todo support "LOSS : cross" for TF_Lite Exporter
+#if defined(ENABLE_TEST)
   ModelHandle model = ml::train::createModel(ml::train::ModelType::NEURAL_NET,
                                              {withKey("loss", "mse")});
+#else
+  ModelHandle model = ml::train::createModel(ml::train::ModelType::NEURAL_NET,
+                                             {withKey("loss", "cross")});
+#endif
 
   for (auto layer : createResnet18Graph()) {
     model->addLayer(layer);
@@ -259,9 +265,8 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
 
   model->train();
 
-  model->exports(ml::train::ExportMethods::METHOD_TFLITE, "resnet_test.tflite");
-
 #if defined(ENABLE_TEST)
+  model->exports(ml::train::ExportMethods::METHOD_TFLITE, "resnet_test.tflite");
   training_loss = model->getTrainingLoss();
   validation_loss = model->getValidationLoss();
 #endif
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -286,7 +286,7 @@ public:
   }
 
 private:
-  float empty_buffer[0]; /**< reserved unintialized tensor points to this
+  float empty_buffer[0]; /**< reserved uninitialized tensor points to this
                             buffer */
 
   std::tuple<BidirectionalIndexMap<const float *, Buffer>,   /**< buffer map
@@ -327,47 +327,12 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
   for (auto &n : nodes) {
     auto tf_node = n.get();
 
-    auto now_node_type = tf_node->getOptionType();
-    if (now_node_type ==
-        tflite::BuiltinOptions::BuiltinOptions_FullyConnectedOptions) {
-
-      if (node_count != 0) {
-
-        auto previous_input_shape =
-          nodes.at(node_count - 1).get()->getInputs()[0];
-
-        const unsigned int UNIT = tf_node->getOutputs()[0]->height();
-        const unsigned int CHANNEL = previous_input_shape->channel();
-        const unsigned int HEIGHT = previous_input_shape->height();
-        const unsigned int WIDTH = previous_input_shape->width();
-        const unsigned int UNITLEN = (CHANNEL * WIDTH * HEIGHT);
-
-        auto weight_data = tf_node->getWeights()[0]->getData();
-        auto *ptr = const_cast<float *>(weight_data);
-
-        std::vector<float> old_value_list;
-
-        for (unsigned int i = 0; i < (UNIT * CHANNEL * HEIGHT * WIDTH); i++) {
-          old_value_list.push_back(weight_data[i]);
-        }
-
-        for (unsigned int u = 0; u < UNIT; u++) {
-          for (unsigned int h = 0; h < HEIGHT; h++) {
-            for (unsigned int w = 0; w < WIDTH; w++) {
-              for (unsigned int c = 0; c < CHANNEL; c++) {
-
-                int next_position = (u * CHANNEL * HEIGHT * WIDTH) +
-                                    c * (HEIGHT * WIDTH) + h * WIDTH + w;
-                int now_position = (u * CHANNEL * HEIGHT * WIDTH) +
-                                   h * (WIDTH * CHANNEL) + w * CHANNEL + c;
-
-                ptr[now_position] = old_value_list[next_position];
-              }
-            }
-          }
-        }
-      }
+    if (tf_node->getOptionType() ==
+          tflite::BuiltinOptions::BuiltinOptions_FullyConnectedOptions &&
+        node_count != 0) {
+      tf_node->setNeedReorderWeight();
     }
+
     node_count++;
   }
 
@@ -391,6 +356,17 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
           ->get();
       tf_node->setArg(index, layer_to_tf.find(input_layer_node)->second);
     }
+  }
+
+  node_count = 0;
+  for (auto &n : nodes) {
+    auto tf_node = n.get();
+
+    if (tf_node->isNeedReorder()) {
+      tf_node->weightReorder(node_count);
+    }
+
+    node_count++;
   }
 
   return nodes;
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -13,7 +13,9 @@
 
 #include <algorithm>
 #include <fstream>
+#include <map>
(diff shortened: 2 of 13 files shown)


Applications/Resnet/jni/main.cpp
```diff
@@ -258,6 +258,9 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
                     std::move(dataset_valid));
 
   model->train();
+
+  model->exports(ml::train::ExportMethods::METHOD_TFLITE, "resnet_test.tflite");
```
jijoongmoon, line 262: I think it might be better to enable this depending options.

nntrainer/layers/flatten_layer.h
```diff
@@ -57,6 +57,11 @@ class FlattenLayer : public ReshapeLayer {
    */
   void setProperty(const std::vector<std::string> &values) override;
 
+  /**
+   * @copydoc Layer::supportInPlace()
+   */
+  bool supportInPlace() const override { return false; }
```
jijoongmoon, line 63: flatten layer need to support the inplace. so.. it should return `true`.

nntrainer/tensor/tensor.h
```diff
@@ -684,19 +684,23 @@ class Tensor {
...
-   * @param[in] direction to transpose ex) 0:2:1
-   * @retval    Calculated Tensor
+   * @brief Transpose Tensor
+   *
+   * @param direction to transpose ex) 0:2:1
+   * @param is_FC if layer == FC only transpose shape and clone data
+   * @return Tensor
    */
-  Tensor transpose(const std::string &direction) const;
+  Tensor transpose(const std::string &direction, bool is_FC = false) const;
```
jijoongmoon, line 693: we do not need to add this is_FC. Tensor class must not know the type of Data.

nntrainer/utils/node_exporter.cpp
```diff
@@ -108,12 +108,14 @@ void Exporter::saveTflResult(const std::tuple<props::Unit> &props,
   auto weight_transform = [](std::vector<const Tensor *> &weights) {
     std::vector<Tensor> new_weights;
     new_weights.reserve(weights.size());
-    new_weights.push_back(weights[0]->transpose("0:2:1"));
+
+    new_weights.push_back(weights[0]->transpose("0:2:1", true));
```
jijoongmoon, line 112: We just do the operation properly here. and update new_weights. No need to use transpose function here.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -320,6 +323,54 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
     layer_to_tf.insert({ln.get(), nodes.back().get()});
   }
 
+  int node_count = 0;
```
jijoongmoon, line 326: I think what we have to do here is just identify which fully connected layer node needs to be reshaped. and do the reshaping in somewhere like saveTfLite

jijoongmoon:
I think commit title also need to change. when this commit is merged.. it is not draft any more.

DonghakPark:
> 

Fixed

DonghakPark:
> I think commit title also need to change. when this commit is merged.. it is not draft any more.

Fixed (Title and Comment)

nntrainer/utils/node_exporter.cpp
DonghakPark: Since all other layers have been transposed, I think FC layer should also be transposed. Is it unnecessary?

nntrainer/tensor/tensor.h
DonghakPark: then how about add these function to tensor.cpp

```.cpp
Tensor &Tensor::transpose_shape(const std::string &direction, Tensor &out) const {

  NNTR_THROW_IF(!contiguous, std::invalid_argument)
    << getName() << " is not contiguous. Cannot transpose.";

  if (out.getData() == getData()) {
    Tensor tmp = clone();
    return tmp.transpose_shape(direction, out);
  }

  out.reshape(dim.transpose(direction));

  out.copyData(clone());

  return out;

}

Tensor Tensor::transpose_shape(const std::string &direction) const {
  Tensor result(dim);
  transpose_shape(direction, result);
  return result;
}
```

DonghakPark:
```.py

    new_weights.push_back(*weights[0]);
    new_weights.push_back(*weights[1]);
    new_weights[0].reshape(new_weights[0].getDim().transpose("0:2:1"));
```

myungjoo:
The Ubuntu error might be related with #2013

Applications/Resnet/jni/main.cpp
jijoongmoon, line 262: If it needs to run the test , How about 
``` c++
#if defined(ENABLED_TEST)
 model->exports(ml::train::ExportMethods::METHOD_TFLITE, "resnet_test.tflite");
#endif
```

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -9,12 +9,14 @@
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
  * @bug No known bugs except for NYI items
  */
-#include <tflite_interpreter.h>
-
 #include <algorithm>
 #include <fstream>
+#include <iostream>
```
jijoongmoon, line 14: I think it is ok to remove iostream header.

nntrainer/compiler/tflite_opnode.h
```diff
@@ -138,6 +158,21 @@ class TfOpNode {
    */
   bool isOutputNode() const { return is_output; }
 
+  /**
+   * @brief check if this op node is virtual node
+   *
+   * virtual node is a node that will not be exported
```
jijoongmoon, line 164: Could you plz add more examples of virtual node?

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -324,13 +324,21 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
   }
 
   int node_count = 0;
+  bool is_local_first = true;
```
jijoongmoon, line 327: Could you add here what this ```bool is_local_first``` means?

nntrainer/layers/flatten_layer.h
```diff
@@ -57,11 +57,6 @@ class FlattenLayer : public ReshapeLayer {
    */
   void setProperty(const std::vector<std::string> &values) override;
 
-  /**
-   * @copydoc Layer::supportInPlace()
-   */
-  bool supportInPlace() const override { return false; }
```
jijoongmoon: I think you shouldn't remove this and it should be set as true.

nntrainer/tensor/tensor.cpp
```diff
@@ -1283,6 +1284,11 @@ Tensor &Tensor::transpose(const std::string &direction, Tensor &out) const {
 
   out.reshape(dim.transpose(direction));
 
+  if (is_FC) {
+    out.copyData(clone());
```
jihochu: I'm not sure, but why doesn't it copy data from `*this`? It may occur double copy.

nntrainer/tensor/tensor.h
```diff
@@ -675,19 +675,23 @@ class Tensor {
...
 
   /**
    * @brief Transpose Tensor
    * @param direction to transpose ex) 0:2:1
    * @param[out] Tensor to save to, dimension is always reshaped.
+   * @param is_FC if layer == FC only transpose shape and clone data
```
jihochu, line 690: It's strange. I think transpose operation cannnot distinguish layer....
If you needs this, add option like `shape_only`, and then copyData from it in outer code would be right.

test/unittest/compiler/unittest_interpreter.cpp
```diff
@@ -294,3 +297,110 @@ GTEST_PARAMETER_TEST(nntrainerAutoInterpreterTest, nntrainerInterpreterTest,
   mkTc(makeGraph({fc0, flatten}), "simple_fc_backbone.ini", ini_interpreter)
 ));
 // clang-format on
+
+/**
+ * @brief Fully Connected Layer weights transpose(NCHW -> NHWC) unittest
+ *
+ */
+
+TEST(nntrainerInterpreterTflite, simple_flatten) {
```
jihochu, line 306: What's the relation with line 131's `#ifdef ENABLE_TFLITE_INTERPRETER`

jihochu, line 306: I'm not sure, but can we extract common part of this to the parameterized tests (line 294) ??

```
GTEST_PARAMETER_TEST(nntrainerAutoInterpreterTest, nntrainerInterpreterTest,
                        ::testing::Values(
  mkTc(makeGraph({fc0, flatten}), "simple_fc.ini", ini_interpreter),
  mkTc(makeGraph({fc0, flatten}), "simple_fc_backbone.ini", ini_interpreter)
  mkTc(makeGraph({fc0, flatten}), "simple_fc.ini", tflite_interpreter),
  mkTc(makeGraph({fc0, flatten}), "simple_fc_backbone.ini", tflite_interpreter)
));
```

nntrainer/compiler/tflite_opnode.cpp
```diff
@@ -61,12 +112,78 @@ void TfOpNode::setLayerNode(const LayerNode &layer) {
...
+
+    auto previous_input_shape = input_nodes[0]->getInputs()[0];
+
+    const unsigned int UNIT = outputs[0]->height();
+    const unsigned int CHANNEL = previous_input_shape->channel();
+    const unsigned int HEIGHT = previous_input_shape->height();
+    const unsigned int WIDTH = previous_input_shape->width();
+
+    auto weight_data = weights[0]->getData();
+    auto *ptr = const_cast<float *>(weight_data);
```
jihochu, line 129:
```suggestion
    flaot *ptr = weights[0]->getData();
```
Isn't it possible? float is default for getData.

nntrainer/compiler/tflite_opnode.cpp
```diff
@@ -61,12 +112,78 @@ void TfOpNode::setLayerNode(const LayerNode &layer) {
...
+
+    std::vector<float> old_value_list;
+    old_value_list.reserve((UNIT * CHANNEL * HEIGHT * WIDTH));
```
jihochu, line 132:
```suggestion
    std::vector<flaot> old_value_list(UNIT * CHANNEL * HEIGHT * WIDTH);
```

Is there any type for `std::vector<flaot>` ?

jijoongmoon (approved):
LGTM

test/unittest/compiler/unittest_interpreter.cpp
jihochu: OK.

nntrainer/compiler/tflite_opnode.cpp
jihochu: There is non-const version of getData. please check.
https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/tensor/tensor.h#L1197

nntrainer/compiler/tflite_opnode.cpp
```diff
@@ -43,10 +73,31 @@ void TfOpNode::setLayerNode(const LayerNode &layer) {
...
+   * A1) the tflite uses NHWC format; nntrainer uses NCHW
+   *
+   * Q2) Why are only output tensors reshaped?
+   * A2) the tflite needs only one tensor between nodes. Therefore,
+   *basically, outputs are used for tflite tensors
+   **/
+  auto create_variables_with_NCHW_to_NHWC = [](auto tensor_getter,
+                                               unsigned size) {
+    Variables v;
+    v.reserve(size);
```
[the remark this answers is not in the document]
jihochu: I think it's same as "tflite_opcode.cpp:132"
please check.

jihochu (approved):
LGTM. but please check some comments.

PR #2007, opened 2022-09-27
[unittest] remove unittest_nntrainer_layers @open sesame 10/18 13:39

### LINKED ISSUE #2002: unittest_nntrainer_layers is disabled
- This unittest have been disabled from commit f519056 in #1281.  (Last commit that this unittest works is fe87882c)
 - Now we have done refactoring the layers, it needs to be enabled but it is using outdated interfaces and we develop new unittest for layers.

Following unit test is considerable.
 - negative setProperty unittest
 - with activation
 - with optimizer
 - with loss

lhs8928:
Commit1:    [unittest] migrate unittest from unittest_nntrainer_layers to unittest_layers
    
     - Migrate unittest for test negative property from unittest_nntrainer_layer to
       unittest_layers
     - Added LayerPropertySemantics which is inherited from LayerSemantics
       to test negative property only
    
Commit2:     [unittest] remove unittest_nntrainer_layers
    
     - Remove unittest_nntrainer_layers.cpp cause this test have been disable.
     - Rename unittest_layers_v2 to unittest
    
    Close #2002
    
Commit3:     [unittest] refine unittest
    
     - Use epsilon to compare float variable
     - Use unused enum LayerCreateSetPropertyOptions
     - fix typos
    

test/unittest/layers/unittest_layers_attention.cpp
```diff
@@ -16,9 +16,10 @@
 #include <attention_layer.h>
 #include <layers_common_tests.h>
 
-auto semantic_attention =
-  LayerSemanticsParamType(nntrainer::createLayer<nntrainer::AttentionLayer>,
-                          nntrainer::AttentionLayer::type, {}, 0, false, 2);
+auto semantic_attention = LayerSemanticsParamType(
+  nntrainer::createLayer<nntrainer::AttentionLayer>,
+  nntrainer::AttentionLayer::type, {},
+  LayerCreateSetPropertyOptions::AVAILABLE_FROM_APP_CONTEXT, false, 2);
```
jihochu, line 22: I couldn't find any reason to replace 0 to 1 (LayerCreateSetPropertyOptions::AVAILABLE_FROM_APP_CONTEXT).
Could you check this?

lhs8928: Previous option value was 0 which makes every layers to create object(layers_dependent_common_test.cpp:24) and if it throws then it register the layer to app context. In other words create object was used to test layer is already register or not.
But we already know which layer is registered by default so by giving the option LayerCreateSetPropertyOptions::AVAILABLE_FROM_APP_CONTEXT it can be skipped.

jihochu: So, we can check it's defaultly registered or not..
Thanks for explaining.

jihochu (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2009, opened 2022-09-29
Add Cache pool test

jihochu:
Legacy MemoryPool test is modifed to both MemoryPool and CachePool.
Parameterized test is used for this. CachePool specific tests are also
created.

It depends on: #1965


test/unittest/memory/memory_planner_validate.h
```diff
@@ -1,51 +0,0 @@
...
-#include <memory>
-#include <string>
-
-#include <memory_planner.h>
-
-/**
- * @class MemoryPlannerValidate
- * @brief Memory planning validation
- */
-class MemoryPlannerValidate : public ::testing::TestWithParam<std::string> {
```
jijoongmoon, line 26: One of the reason we make this Validate class separate is there could be multiple Memory Planner. So.. I think it might be better if we maintain as it is for later usage.

jihochu: The reason for moving this file to unittest_memory_planner_validate.cpp file is that it can hold all operation related to memory planner, using parameterized test. I think it doesnot need to divide file, which are used only for unittest_memory_planner.

jijoongmoon: I see. Thanks.

test/unittest/memory/unittest_cache_pool.cpp
```diff
@@ -0,0 +1,328 @@
...
+  EXPECT_CALL(*pool, invalidate).Times(testing::AtLeast(1));
+
+  std::shared_ptr<nntrainer::MemoryData<float>> mem;
+  auto idx = pool->requestMemory(1, 4, 5);
+  EXPECT_NO_THROW(pool->planLayout(nntrainer::BasicPlanner()));
+  EXPECT_NO_THROW(pool->allocate());
+
+  EXPECT_NO_THROW(mem = pool->getMemory(idx));
+  EXPECT_NE(mem, nullptr);
+  EXPECT_EQ(mem->getAddr(), nullptr);
```
jijoongmoon, line 89: I think it could be great if more details are provided.

jijoongmoon (approved):
LGTM except minor comment

test/unittest/memory/unittest_cache_pool.cpp
jihochu: thanks. comments which describes each test case are added.

DonghakPark (approved):
LGTM!

PR #2010, opened 2022-09-29
[neuralnet] add log about training

lhs8928:
- Added log when start/finish training
 - Added log when get current epoch is called

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -113,6 +113,13 @@ int NeuralNetwork::loadFromConfig(const std::string &config) {
   return ML_ERROR_NONE;
 }
 
+unsigned int NeuralNetwork::getCurrentEpoch() {
+#ifdef DEBUG
+  ml_logd("[NNTrainer] Current epoch: %d", epoch_idx);
+#endif
+  return epoch_idx;
+};
+
 void NeuralNetwork::setProperty(const std::vector<std::string> &values) {
   auto left_props = loadProperties(values, model_props);
   setTrainConfig(left_props);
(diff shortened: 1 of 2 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -774,6 +782,11 @@ int NeuralNetwork::train_run(std::function<bool(void *userdata)> stop_cb) {
                 << " - Training Loss: " << stat.loss;
       ml_logi("# %d / %d - Training Loss: %f", epoch_idx, getEpochs(),
               stat.loss);
+      ml_logd("[NNTrainer] Training epoch %d / %d finished successfully.",
+              epoch_idx, getEpochs());
+    } else {
+      ml_logd("[NNTrainer] Training stopped unexpectedly during epoch %d.",
+              epoch_idx);
```
jijoongmoon, line 789: I wonder why it is "unexpectedly"

nntrainer/models/neuralnet.cpp
```diff
@@ -113,6 +113,11 @@ int NeuralNetwork::loadFromConfig(const std::string &config) {
   return ML_ERROR_NONE;
 }
 
+unsigned int NeuralNetwork::getCurrentEpoch() {
```
jijoongmoon, line 116: If application developer need this ```getCurrentEpoch()```, then we need to add C/C++ API for this.

nntrainer/models/neuralnet.cpp
```diff
@@ -113,6 +113,11 @@ int NeuralNetwork::loadFromConfig(const std::string &config) {
...
+  ml_logd("[NNTrainer] Current epoch: %d", epoch_idx);
```
jijoongmoon, line 117: I wonder, why do we need the log here? It means... everytime we call ```getCurrentEpoch()```, the log appear.
If application developer needs this, then he can generate the log in application. Not inside of NNTrainer.

nntrainer/models/neuralnet.cpp
lhs8928: Changed to "stopped by stop callback function"

jijoongmoon: Great.

jijoongmoon (approved):
LGTM

PR #2012, opened 2022-10-04
[trivial] fix ahub issues

lhs8928:
- Remove noexcept in copy function cause it throws when data is non contiguous
 - Initialize char array
 - Delete array not element("[]" was missing)


jijoongmoon (approved):
LGTM

songgot (approved):
LGTM

PR #2015, opened 2022-10-14
[neuralnet] check stop_cb by every layer

lhs8928:
- Check stop_cb before doing forwarding/calcGrading/calcDerivative by every layer to stop training more quickly.

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -329,8 +329,10 @@ void NetworkGraph::applyGradients(
   }
 }
 
-sharedConstTensors NetworkGraph::forwarding(bool training) {
-  for (auto iter = cbegin(); iter != cend(); iter++) {
+sharedConstTensors
+NetworkGraph::forwarding(bool training,
+                         std::function<bool(void *userdata)> stop_cb) {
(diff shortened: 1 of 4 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -329,8 +329,10 @@ void NetworkGraph::applyGradients(
   }
 }
 
-sharedConstTensors NetworkGraph::forwarding(bool training) const {
-  for (auto iter = cbegin(); iter != cend(); iter++) {
+sharedConstTensors
+NetworkGraph::forwarding(bool training,
+                         std::function<bool(void *userdata)> stop_cb) const {
+  for (auto iter = cbegin(); iter != cend() && !stop_cb(nullptr); iter++) {
```
jijoongmoon, line 335: Although It think it could be considered in later pr, but the default fo stop_cb is return false. But how about changing it as nullptr and if the stop_cb is not null the call stop_cb so that we can skip the check if there is stop_cb is not set.

jijoongmoon (approved):
LGTM except some comments for later pr.

jihochu (approved):
LGTM!

PR #2033, opened 2022-11-07
Add Task and TaskExecutor @open sesame 11/11 10:50

jihochu:
Task has several properties for its state and working information.
There are two types for Task, Synchronous and Asynchronous. and, their
behaviors will be differ when they are executed by TaskExecutor.

TaskExecutor class manages task's execution by their type.
Synchoronous tasks are exeucuted immediately, while asynchoronous tasks
are executed in thread manner.

For testing above two classes, unittest codes are also included.
The mock class for TaskExecutor is introduced to check work thread is
correctly working, and completion check is also working correctly.
Many tests are added to check asynchronous tasks' correctness.

Issue (#2034)

### DIFF
--- a/nntrainer/tensor/task_executor.h
+++ b/nntrainer/tensor/task_executor.h
@@ -1,0 +1,175 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2022 Jiho Chu <jiho.chu@samsung.com>
+ *
+ * @file   task_executor.h
+ * @date   04 Nov 2022
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jiho Chu <jiho.chu@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  Task executor class
+ *
+ */
+
+#ifndef __TASK_EXECUTOR_H__
+#define __TASK_EXECUTOR_H__
+
+#include <atomic>
+#include <functional>
+#include <future>
+#include <map>
+#include <memory>
+#include <unistd.h>
+
+#include <nntrainer_error.h>
+#include <task.h>
+
+namespace nntrainer {
+
+/**
+ * @class   TaskExecutor
+ * @brief   task executor class
+ */
+class TaskExecutor {
+public:
+  /**
+   * @brief Complete error types
+   */
+  enum CompleteStatus {
+    SUCCESS,
+    FAIL_CANCEL,
+    FAIL_TIMEOUT,
+    FAIL,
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/task_executor.h
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * @brief Check task is completed
+   *
+   * @param data user data
+   */
+  template <typename T>
+  void completeChecker(int id, std::shared_ptr<TaskAsync<T>> task) {
+    // wait until tasks are fully mapped
+    while (tasks.find(id) == tasks.end())
```
jijoongmoon, line 135: I wonder if there is possiblity to loop forever when the id is greater than  ```map::size()```. better to check if the id is less than ```map::size()```

jihochu: Thanks.
timeout logic is added to. max timeout is set as 1 sec.

nntrainer/tensor/task_executor.h
```diff
@@ -0,0 +1,162 @@
...
+class TaskExecutor {
+public:
+  enum CompleteStatus {
+    SUCCESS,
+    FAIL_CANCEL,
+    FAIL_TIMEOUT,
+    FAIL,
+  };
+
+  using CompleteCallback =
```
jijoongmoon, line 42: Could you add more explanation as an comment?

nntrainer/tensor/task_executor.h
```diff
@@ -0,0 +1,162 @@
...
+
+    NNTR_THROW_IF(tasks.find(id) == tasks.end(), std::runtime_error)
+      << "tasks was not correctly mapped";
+
+    auto status = std::get<0>(tasks[id]).wait_for(T(task->getTimeout()));
+    handleCompleteStatus(id, status);
+
+    task->setState(Task::State::DONE);
+  }
+
```
jijoongmoon, line 150: add doxygen comment for this function?

jijoongmoon (approved):
LGTM except minor comment.

nntrainer/tensor/task_executor.h
jihochu: thanks. it added.

nntrainer/tensor/task_executor.h
jihochu: doxy comments are added.

DonghakPark (approved):
LGTM!

songgot (approved):
LGTM

PR #2041, opened 2022-11-15
(Wait for #2055) [SWAP] Add CacheLoader class @open sesame 11/17 16:55

jihochu:
This patch adds cache loader class.

Cache loader supports loading cache elements methods in both synchoronous and
asynchronous manners. Asynchoronous methods are using task executor, and
loods whole cache elements with execution orders in other thread.

### DIFF
--- a/nntrainer/tensor/cache_loader.cpp
+++ b/nntrainer/tensor/cache_loader.cpp
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2022 Jiho Chu <jiho.chu@samsung.com>
+ *
+ * @file   cache_loader.cpp
+ * @date   10 Nov 2022
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jiho Chu <jiho.chu@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  Cache loader class
+ *
+ */
+
+#include "cache_loader.h"
+#include "ml-api-common.h"
+#include "task.h"
+#include "task_executor.h"
+
+#include <cache_pool.h>
+#include <climits>
+#include <cstdint>
+#include <exception>
+#include <memory>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+
+namespace nntrainer {
+
+CacheLoader::CacheLoader(std::shared_ptr<CachePool> cache_pool) :
+  pool(cache_pool),
+  task_executor(nullptr) {}
+
+CacheLoader::~CacheLoader() {
+  if (task_executor)
+    delete task_executor;
+}
+
+void CacheLoader::init() {
+  if (task_executor)
+    return;
+
+  task_executor = new TaskExecutor(pool->getName());
+}
+
+void CacheLoader::finish() {
+  delete task_executor;
+  task_executor = nullptr;
+}
+
+void CacheLoader::load(unsigned int order) { pool->loadExec(order); }
+
+int CacheLoader::loadAsync(unsigned int order,
+                           TaskExecutor::CompleteCallback complete) {
+  return loadAsync(order, complete, LONG_MAX);
+}
+
+int CacheLoader::loadAsync(unsigned int order,
+                           TaskExecutor::CompleteCallback complete,
+                           long timeout_ms) {
+  if (!task_executor) {
+    ml_loge("init is needed");
+    return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  Task::Work work = [&](std::atomic_bool &running, void *data) {
+    unsigned int exe_order = (unsigned int)(std::uintptr_t)data;
+
+    pool->flushExcept({exe_order - 1, exe_order});
+    pool->loadExec(exe_order);
+
+    return ML_ERROR_NONE;
+  };
+
+  auto task =
+    std::make_shared<TaskAsync<>>(work, (void *)(std::uintptr_t)order);
+  task->setTimeout(timeout_ms);
+
+  return task_executor->run(task, complete);
+}
+
+int CacheLoader::cancelAsync(int id) {
+  try {
+    task_executor->cancel(id);
+  } catch (const std::exception &e) {
+    ml_loge("CacheLoader(%s): failed to cancel(%d): %s",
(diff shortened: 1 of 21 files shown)


nntrainer/tensor/cache_loader.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  task_executor = new TaskExecutor(pool->getName());
+}
+
+void CacheLoader::finish() {
+  delete task_executor;
+  task_executor = nullptr;
+}
+
+void CacheLoader::load(unsigned int order) { pool->loadExec(order); }
```
jijoongmoon, line 48: it calls ```pool->loadExec(order)```, and in ```loadExec(order)```, start searching the ```elems```,  ```CacheElems``` and check if the ```elem``` in ```elems``` has same execution order as ```order```. It's kind of something to be done definitely but I'm thinking, it only needs to be done first iteration. and rest of training processes, we do not need to compute which elem needs be loaded  at specific ```order```. Becuase it will be always same once we compute. 
So how about ```CachePool``` has those scheduled information according to execution order.. and if it is set, then just load. I think we do not validate either.

nntrainer/tensor/cache_loader.cpp
```diff
@@ -0,0 +1,91 @@
...
+                           long timeout_ms) {
+  if (!task_executor) {
+    ml_loge("init is needed");
+    return -1;
+  }
+
+  Task::Work work = [&](std::atomic_bool &running, void *data) {
+    unsigned int exe_order = (unsigned int)(std::uintptr_t)data;
+
+    pool->flushExcept({exe_order - 1, exe_order});
```
jijoongmoon, line 66: looks like it is not defined correctly. I wonder how the ci is passed. Could you check?

nntrainer/tensor/cache_loader.h
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Copyright (C) 2022 Jiho Chu <jiho.chu@samsung.com>
+ *
+ * @file   cache_loader.h
+ * @date   10 Nov 2022
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jiho Chu <jiho.chu@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  Cache loader class
+ *
```
jijoongmoon, line 11: It might be better to understand, how it cooperate with task executed & task.
so.. ```CacheLoader``` initialized with ```TaskExecutoer``` which has ```Task``` maps and ```Task``` defines work (functional with atomic running) and type and priority.

test/unittest/memory/unittest_cache_loader.cpp
```diff
@@ -0,0 +1,417 @@
...
+  pool->flush();
+}
+
+/**
+ * @brief load asynchronously
+ */
+TEST_F(CacheLoaderTest, load_async_02_p) {
+  std::shared_ptr<nntrainer::MemoryData<float>> mem1, mem2, mem3;
+  auto idx1 = pool->requestMemory(4, 1, 5, {1, 2, 3, 4, 5});
+  auto idx2 = pool->requestMemory(4, 3, 8, {3, 4, 5, 6, 7, 8});
```
jijoongmoon, line 202: I think it might be good if you could add about non-continuous execution order lists later.

nntrainer/tensor/cache_loader.cpp
```diff
@@ -0,0 +1,91 @@
...
+                           TaskExecutor::CompleteCallback complete) {
+  return loadAsync(order, complete, LONG_MAX);
+}
+
+int CacheLoader::loadAsync(unsigned int order,
+                           TaskExecutor::CompleteCallback complete,
+                           long timeout_ms) {
+  if (!task_executor) {
+    ml_loge("init is needed");
+    return -1;
```
jijoongmoon, line 60: How about using throw or ML_ERROR_* in nntrainer_error.h?

nntrainer/tensor/cache_loader.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  Task::Work work = [&](std::atomic_bool &running, void *data) {
+    unsigned int exe_order = (unsigned int)(std::uintptr_t)data;
+
+    pool->flushExcept({exe_order - 1, exe_order});
+    pool->loadExec(exe_order);
+
+    return 0;
```
jijoongmoon, line 69: ML_ERROR_NONE? maybe?

nntrainer/tensor/cache_loader.cpp
jihochu: Thanks for sharing idea.
It's implemented in #2055, and you may be already checked while reviweing #2055.

nntrainer/tensor/cache_loader.cpp
jihochu: I think it's right. {exe_order - 1, exe_order} are corrently needed.

nntrainer/tensor/cache_loader.h
jihochu: Maybe I'm misunderstang, but task executor already has task queue...

nntrainer/tensor/cache_loader.cpp
jihochu: thanks. it's fixed.

nntrainer/tensor/cache_loader.cpp
jihochu: ditto

test/unittest/memory/unittest_cache_loader.cpp
jihochu: thanks. `load_async_03_p` case is added.

nntrainer/tensor/cache_loader.cpp
jijoongmoon: Could you explain in more detail? I guess we flush the memory whose execution order is less than the current execution order.

nntrainer/tensor/cache_loader.h
jijoongmoon: yes. What I'm asking is more details about their relation.

jijoongmoon (approved):
LGTM! Thanks for Great Work!

baek2sm (approved):
LGTM

PR #2043, opened 2022-11-18
[Logger] Modify initialization error message of logger instance

songgot:
[Logger] Modify initialization error message of logger instance
- Runtime error occurs when an app is executed in a place other
  than the home directory excepted Android and Tizen
- Currently, Android and Tizen don't make logger
- If an user runs an app using nntrainer in a place where
  the logger cannot be initialized, the path is shown for help
- In case of Ubunut, logger unable to initialize on '/usr/bin/',
  /usr/lib/' and etc

**Self evaluation:**

Signed-off-by: hyunil park <hyunil46.park@samsung.com>

### DIFF
--- a/nntrainer/nntrainer_logger.cpp
+++ b/nntrainer/nntrainer_logger.cpp
@@ -30,6 +30,7 @@
 #include <sstream>
 #include <stdarg.h>
 #include <stdexcept>
+#include <unistd.h>
 
 namespace nntrainer {
 
@@ -77,7 +78,12 @@ Logger::Logger() {
      << std::setw(2) << now->tm_sec << ".out";
   outputstream.open(ss.str(), std::ios_base::app);
   if (!outputstream.good()) {
-    throw std::runtime_error("Unable to initialize the Logger!");
+    char buf[256];
+    char *ret = getcwd(buf, 256);
+    std::string cur_path = std::string(buf);
+    std::string err_msg =
+      "Unable to initialize the Logger on path(" + cur_path + ")";
+    throw std::runtime_error(err_msg);
   }
 }
 


songgot:
Use getcwd() instead of std::filesystem::current_path().

/var/www/html/android/android-ndk-r21e/sources/cxx-stl/llvm-libc++/include/filesystem:1458: undefined reference to `std::__ndk1::__fs::filesystem::__current_path(std::__ndk1::error_code*)'

nntrainer/nntrainer_logger.cpp
```diff
@@ -77,7 +78,12 @@ Logger::Logger() {
      << std::setw(2) << now->tm_sec << ".out";
   outputstream.open(ss.str(), std::ios_base::app);
   if (!outputstream.good()) {
-    throw std::runtime_error("Unable to initialize the Logger!");
+    char buf[256];
+    char *ret = getcwd(buf, 256);
```
myungjoo, line 82: https://en.cppreference.com/w/cpp/filesystem/current_path might be better for C++ code.

songgot: okay, thank you.

songgot: I've already tried that 12days ago.
We don't use filesystem::current_path becase of android c++ version.
nntrainer already use getcwd().
so I used getcwd() instead of std::filesystem::current_path()
error :
/var/www/html/android/android-ndk-r21e/sources/cxx-stl/llvm-libc++/include/filesystem:1458: undefined reference to `std::__ndk1::__fs::filesystem::__current_path(std::__ndk1::error_code*)'

jijoongmoon (approved):
LGTM

DonghakPark (approved):
LGTM!

lhs8928 (approved):
LGTM

PR #2050, opened 2022-11-23
[utils] Add get_realpath function

baek2sm:
Since "realpath" function is not available in the Windows operating system, "get_realpath" function has been added and replaced so that it can be used in the Windows operating system.

- Add get_realpath function to utils.

**Self evaluation:**

Signed-off-by: Seungbaek Hong <sb92.hong@samsung.net>


nntrainer/utils/util_func.cpp
```diff
@@ -197,4 +197,12 @@ bool istrequal(const std::string &a, const std::string &b) {
   });
 }
 
+char *get_realpath (const char *__restrict __name, char *__restrict __resolved) {
```
lhs8928, line 200: This is trivial but stick to use camel case for function name instead of snake case?

nntrainer/utils/util_func.cpp
```diff
@@ -197,4 +197,12 @@ bool istrequal(const std::string &a, const std::string &b) {
...
+#ifdef _WIN32
```
jijoongmoon, line 201: How about change the def for the consistency.
for Android, nntrainer uses ```__ANDROID___```, Tizen, ```___TIZEN__```.
So.. for Windows, how about ```___WIN___``` ?

nntrainer/utils/util_func.h
```diff
@@ -23,6 +23,10 @@
 #ifndef __UTIL_FUNC_H__
 #define __UTIL_FUNC_H__
 
+#ifdef _WIN32
+#define MAX_PATH_LENGTH 1024
```
jijoongmoon, line 27: I wonder where it is used?

jijoongmoon:
Plz, Check the CI.

nntrainer/utils/util_func.cpp
baek2sm: Thanks, I will reflect the modification.

nntrainer/utils/util_func.cpp
baek2sm: Thank you, I didn't know that there was a predefined keyword called "__WIN32__". I will fix it!

nntrainer/utils/util_func.h
baek2sm: It is used in "_fullpath" function at "util_func.cpp"(line 202). Would it be better to move this definition to a cpp file?

DonghakPark (approved):
LGTM!

nntrainer/utils/util_func.cpp
myungjoo: _WIN32 is predefined macro defined by toolchain. __ANDROID__ and __TIZEN__ are macro defined by nntrainer. If you want __WIN32__, you will need to do it in a common header: ```#define __WIN32__ (_WIN32)``` (do we really need this?)

nntrainer/utils/util_func.h
```diff
@@ -271,6 +275,16 @@ template <typename T, typename C = int> T enum_class_or(T e1, T e2) {
...
 
+/**
+ * @brief Convert a relative path into an absolute path.
+ *
+ * @param __name relative path
+ * @param __resolved variable to store the result value.
+ *
+ * @return absolute path
+ */
+char *getRealpath(const char *__restrict __name, char *__restrict __resolved);
```
jihochu, line 286: Should we use '__' prefix for parameter name ?

nntrainer/utils/util_func.cpp
baek2sm: This modification has not been reflected yet. If it is necessary to modify the keyword `_WIN32` into `__WIN32__`, then I will modify and reflect it again.

nntrainer/utils/util_func.h
baek2sm: Thank you for your good point. In fact, it just followed definition of the original realpath function at stdlib. However, It would be better to remove the underbar from the parameter name because it is a custom function. I will reflect that modification. Thanks a lot!

nntrainer/utils/util_func.cpp
myungjoo: OR... maybe https://en.cppreference.com/w/cpp/filesystem/absolute is better?

jijoongmoon: > _WIN32 is predefined macro defined by toolchain 
if it is defined by Visual Studio, then it is ok to go.

nntrainer/utils/util_func.h
jijoongmoon: > It is used in "_fullpath" function at "util_func.cpp"(line 202). Would it be better to move this definition to a cpp file?

I think it is. Thanks.

nntrainer/utils/util_func.cpp
baek2sm: > OR... maybe https://en.cppreference.com/w/cpp/filesystem/absolute is better?

The "absolute" function seems to work well in windows. It seems to be best way to use the absolute function. I will reflect this modification.

SeoHyungjun:
LGTM

nntrainer/utils/util_func.h
```diff
@@ -271,6 +271,16 @@ template <typename T, typename C = int> T enum_class_or(T e1, T e2) {
...
 
+/**
+ * @brief Convert a relative path into an absolute path.
+ *
+ * @param name relative path
+ * @param resolved variable to store the result value.
+ *
+ * @return absolute path
+ */
+char *getRealpath(const char *__restrict name, char *__restrict resolved);
```
jihochu, line 282: `__restrict` is optmization directive. Should we use this in here?
It doesnot needs to have disjoint property. 
please check: https://en.cppreference.com/w/c/language/restrict

baek2sm: I didn't know about it. The PR has been modified. Thank you very much for letting me know.

nntrainer/utils/util_func.cpp
baek2sm: Modifications were reflected to use the "absolute" function. At this time, the existing PR has been modified as it is, but I wonder if it would be better to upload new PR in the future if it is completely different from the existing PR.

baek2sm: Filesystem including absolute function does not seem to be fully supported in android r21 (supported from r22) : https://github.com/android/ndk/issues/609#issuecomment-631918497. Therefore, I would like to modify it again by defining the "getRealpath" function as before.

nntrainer/utils/util_func.h
baek2sm: @jihochu I've removed `__restrict` in that function. Thank you :)

nntrainer/utils/util_func.cpp
baek2sm: I uploaded it reflecting the work situation so far.

PR #2055, opened 2022-11-29
(Wait for #2098) [SWAP] Implement cache policy

jihochu:
cache policy determines whether swap in/out operation is needed
when cache data is requested. The policies are inferred from
tensor lifespan.

### DIFF
--- a/nntrainer/tensor/cache_pool.cpp
+++ b/nntrainer/tensor/cache_pool.cpp
@@ -31,8 +31,9 @@ namespace {
  *
  * @param lifespand tensor lifespan
  * @return cache policy
- * @note cache policy is defined by tensor's lifetime. If it needs to be maintained its value,
- * ALWAYS_SYNCED or ITERATION_CONSIST is proper. If not, TEMPORAL doesnot keep its value.
+ * @note cache policy is defined by tensor's lifetime. If it needs to be
+ * maintained its value, ALWAYS_SYNCED or ITERATION_CONSIST is proper. If not,
+ * TEMPORAL doesnot keep its value.
  */
 inline const CachePolicy
 convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
@@ -43,7 +44,7 @@ convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
     policy = CachePolicy::ALWAYS_SYNCED;
     break;
   case TensorLifespan::FORWARD_FUNC_LIFESPAN:
-    policy = CachePolicy::ALWAYS_SYNCED;
+    policy = CachePolicy::TEMPORAL;
     break;
   case TensorLifespan::CALC_DERIV_LIFESPAN:
     policy = CachePolicy::TEMPORAL;


nntrainer/tensor/cache_pool.cpp
```diff
@@ -11,51 +11,81 @@
...
+inline const CachePolicy
+convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
+  CachePolicy policy;
+
+  switch (lifespan) {
+  case TensorLifespan::UNMANAGED:
+    policy = CachePolicy::ALWAYS_SYNCED;
+    break;
+  case TensorLifespan::FORWARD_FUNC_LIFESPAN:
+    policy = CachePolicy::ALWAYS_SYNCED;
```
jijoongmoon, line 38: I'm not following why ```FORWARD_FUNC_LIFESPAN``` is needs to be ```ALWAYS_SYNCED```?

nntrainer/tensor/cache_pool.cpp
```diff
@@ -11,51 +11,81 @@
...
+  case TensorLifespan::CALC_GRAD_LIFESPAN:
+    policy = CachePolicy::TEMPORAL;
+    break;
+  case TensorLifespan::CALC_GRAD_DERIV_LIFESPAN:
+    policy = CachePolicy::TEMPORAL;
+    break;
+  case TensorLifespan::FORWARD_GRAD_LIFESPAN:
+    policy = CachePolicy::ITERATION_CONSIST;
+    break;
+  case TensorLifespan::FORWARD_DERIV_LIFESPAN:
```
jijoongmoon, line 52: This assumes ```calcDerivatives"` includes ```Apply Gradient``` ?

nntrainer/tensor/cache_pool.cpp
```diff
@@ -11,51 +11,81 @@
...
-void CacheElem::swapIn() {
-  std::lock_guard<std::mutex> lock(device_mutex);
-  void *buf = device->getBuffer(offset, length);
-  mem_data->setAddr((float *)buf);
-  mem_data->setValid(true);
-  active = true;
+namespace {
+
+inline const CachePolicy
+convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
```
jijoongmoon, line 30: It might be better if you give more detail explanation about how those lifespan converts to CachePolicy?

nntrainer/tensor/cache_pool.cpp
```diff
@@ -11,51 +11,81 @@
...
-  active = false;
+std::atomic_int pool_id = 0;
 
-  PROFILE_MEM_DEALLOC(buf);
-}
+} // namespace
 
-CachePool::CachePool(const std::string &name) :
-  swap_device(std::make_shared<SwapDevice>(name + std::to_string(getpid()))) {}
+CachePool::CachePool(const std::string &n) :
```
jijoongmoon, line 76: Does the n is for path of swap device? if it is could you add the comment in ```CachePool``` constructor in header?

nntrainer/tensor/cache_pool.cpp
```diff
@@ -106,23 +136,41 @@ void CachePool::invalidate(unsigned int id) {
...
+unsigned int CachePool::requestMemory(size_t bytes, unsigned int start_time,
+                                      unsigned int end_time,
+                                      std::vector<unsigned int> exec_order,
+                                      TensorLifespan lifespan) {
+  auto id = MemoryPool::requestMemory(bytes, start_time, end_time, exec_order,
+                                      lifespan);
+
+  const CachePolicy policy = convertTensorLifespanToCachePolicy(lifespan);
+
+  policies.push_back(policy);
```
jijoongmoon, line 148: I'm not sure whether the id is not considered here.

nntrainer/tensor/cache_pool.cpp
```diff
@@ -106,23 +136,41 @@ void CachePool::invalidate(unsigned int id) {
...
+
 std::shared_ptr<MemoryData<float>> CachePool::getMemory(unsigned int id) {
   NNTR_THROW_IF(!swap_device->isOperating(), std::invalid_argument)
     << "Allocate memory before allocation";
 
-  int offset = getMemoryOffset().at(id - 1);
+  size_t offset = getMemoryOffset().at(id - 1);
   size_t len = getMemorySize().at(id - 1);
   auto exe_order = getMemoryExecOrder().at(id - 1);
+  auto policy = getCachePolicy().at(id - 1);
```
jijoongmoon, line 163: It expects the ```id``` is set in an order.

nntrainer/tensor/cache_pool.cpp
```diff
@@ -133,69 +181,111 @@ std::shared_ptr<MemoryData<float>> CachePool::getMemory(unsigned int id) {
...
+    auto id = elem->getId();
+    auto exe_order = exe_orders.at(id - 1);
     auto found = std::find(exe_order.begin(), exe_order.end(), order);
     if (found == exe_order.end()) {
-      elem->swapOut();
+      /**
+       * We assumes that flushExcept will be called in front of each execution
+       * order, and the order is incremental. So, we can conclude that, if the
+       * order passes by the max order of the cache element, it was LAST access
+       * of the element.
```
jijoongmoon, line 205: For now, it is ok. However, in extreme optimization cases, we know and can save the list that needs to be flushed after the first iteration so that we can memorize them and use them until the training end without finding them again.

nntrainer/tensor/cache_pool.cpp
jijoongmoon, line 163: how about add ```getCachePolicy(unsigned int id)```?

nntrainer/tensor/cache_pool.cpp
jihochu: Known var/grad tensors donot have this lifespan. To avoid error for unconsidered tensors, it sets ALWAYS_SYNCED for this.

nntrainer/tensor/cache_pool.cpp
jihochu: apply gradient is not implement yet in this PR.

nntrainer/tensor/cache_pool.cpp
jihochu: thanks. it added.

nntrainer/tensor/cache_pool.cpp
jihochu: thanks. it added.

nntrainer/tensor/cache_pool.cpp
jihochu: It's for simplying the implementaion. We already knows that the `id` is size of each information's length(validity, exec_order).
To check the consititency, it checks the `id` with policies length (line 155).

nntrainer/tensor/cache_pool.cpp
jihochu: please check the above comment.

nntrainer/tensor/cache_pool.cpp
jihochu: Thanks. I'll consider it in later.

nntrainer/tensor/cache_pool.cpp
jihochu: OK. It's also for getMemorySize,and others. I'll modify all apis in later PR.

nntrainer/tensor/cache_pool.cpp
jijoongmoon: Well, for the var tensor for non-trainable layer, we could have FORWARD_FUNC_LIFESPAN and we do not need to be ALWAYS_SYNCED.

nntrainer/tensor/cache_pool.cpp
jijoongmoon: Yes. I see. I just ask to confirm.

nntrainer/tensor/cache_pool.cpp
jihochu: Current implementation for non-trainable does not distinguished by lifespan, it's excluding execution order when checking  lifespan.

jihochu: It changed to TEMPORAL as it is.

jijoongmoon (approved):
LGTM

PR #2061, opened 2022-12-06
[swap_device] change lseek return type from int to off_t @open sesame 12/20 15:37

lhs8928:
- Return value of lseek is offset of requested position. The 4byte int type cannot represent more than 3GB so any request more than 3GB offset cause overflow.

### DIFF
--- a/nntrainer/tensor/swap_device.cpp
+++ b/nntrainer/tensor/swap_device.cpp
@@ -25,32 +25,33 @@
 namespace nntrainer {
 
 void SwapDevice::start(size_t size) {
-  int ret;
-
   if (fd > 0)
     return;
 
   fd = open(dev_path.c_str(), O_RDWR | O_CREAT | O_TRUNC, (mode_t)0666);
   NNTR_THROW_IF(fd < 0, std::runtime_error) << "open file: " << dev_path;
 
+  off_t off;
+
   /* make sparse file */
-  ret = lseek(fd, size - 1, SEEK_SET);
-  NNTR_THROW_IF(ret < 0, std::runtime_error) << "seek file: " << dev_path;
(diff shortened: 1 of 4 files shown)


jijoongmoon (approved):
LGTM

nntrainer/tensor/swap_device.cpp
```diff
@@ -108,11 +109,13 @@ void SwapDevice::putBuffer(void *ptr) {
 
   auto [offset, size] = allocated[ptr];
 
-  ret = lseek(fd, offset, SEEK_SET);
-  NNTR_THROW_IF(ret < 0, std::runtime_error) << "seek file: " << dev_path;
+  off_t off;
```
jihochu, line 112: it's better to move line 107

nntrainer/tensor/swap_device.cpp
```diff
@@ -108,11 +109,13 @@ void SwapDevice::putBuffer(void *ptr) {
...
+  off = lseek(fd, offset, SEEK_SET);
+  NNTR_THROW_IF(off < 0, std::runtime_error) << "seek file: " << dev_path;
 
-  ret = write(fd, ptr, size);
-  NNTR_THROW_IF(ret != size, std::runtime_error) << "write file: " << dev_path;
+  ssize_t len;
```
jihochu, line 116: ditto.

DonghakPark (approved):
LGTM!

jihochu (approved):
LGTM.

nntrainer/tensor/swap_device.cpp
lhs8928: Done

nntrainer/tensor/swap_device.cpp
lhs8928: Done

PR #2065, opened 2022-12-09
[DataGen] Fix RandomDataLoader parms

DonghakPark:
Fix RandomDataLoader parm (iteration_for_one_epoch) --> data_size
- in RandomDataLoader it make data as much as data_size
- if developer set iteration_for_one_epoch as iterate size it doesn't work
- for example developer want to make BATCH 64 and ITER 1 it should be 64 in iteration_for_one_epoch so it doesn't make sense
- for more clearly change it's name

Signed-off-by: DonghakPark <donghak.park@samsung.com>


jijoongmoon (approved):
LGTM

SeoHyungjun:
LGTM!

songgot (approved):
LGTM

PR #2066, opened 2022-12-14
[meson] enable when tflite related meson option is off @open sesame 12/20 13:50

lhs8928:
- An applications that needs tflite to run is will not going to be build if option is unset.
 - Tflite interpreter unittest will be working only if tflite option is set.
 - Added default exporter constructor in case of tflite option is unset.

### DIFF
--- a/nntrainer/utils/node_exporter.cpp
+++ b/nntrainer/utils/node_exporter.cpp
@@ -9,26 +9,28 @@
  * @author Jihoon Lee <jhoon.it.lee@samsung.com>
  * @bug No known bugs except for NYI items
  */
-#include <node_exporter.h>
 
-#ifdef ENABLE_TFLITE_INTERPRETER
 #include <activation_layer.h>
 #include <bitset>
 #include <common_properties.h>
 #include <fc_layer.h>
 #include <map>
 #include <node_exporter.h>
+
+#ifdef ENABLE_TFLITE_INTERPRETER
 #include <tf_schema_generated.h>
 #include <tflite_opnode.h>
 #endif
 
 namespace {
 
+#ifdef ENABLE_TFLITE_INTERPRETER
 tflite::Padding tflite_padding(const std::string &padding) {
   std::map<std::string, tflite::Padding> m = {{"same", tflite::Padding_SAME},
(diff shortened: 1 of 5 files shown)


DonghakPark (approved):
LGTM!!

nntrainer/utils/node_exporter.cpp
```diff
@@ -37,11 +39,26 @@ namespace nntrainer {
...
 /**
  * @brief Construct a new Exporter object
  *
  */
-Exporter::Exporter() : fbb(nullptr), stored_result(nullptr), is_exported(false) {}
+Exporter::Exporter() : stored_result(nullptr), is_exported(false) {}
+
+#endif
 
 #ifdef ENABLE_TFLITE_INTERPRETER
```
SeoHyungjun, line 63: Isn't it possible to include it in the #ifdef on line 42?

nntrainer/utils/node_exporter.cpp
```diff
@@ -37,11 +39,26 @@ namespace nntrainer {
...
+#ifdef ENABLE_TFLITE_INTERPRETER
+
+/**
+ * @brief Construct a new Exporter object
+ *
+ */
+Exporter::Exporter() :
+  fbb(nullptr),
+  stored_result(nullptr),
+  is_exported(false) {}
```
jijoongmoon, line 51: How about making parent class of Exporter and making tflite_node_exporter inheriting the Exporter?

nntrainer/utils/node_exporter.cpp
```diff
@@ -37,11 +39,26 @@ namespace nntrainer {
...
+
+#else
+
 /**
  * @brief Construct a new Exporter object
  *
  */
-Exporter::Exporter() : fbb(nullptr), stored_result(nullptr), is_exported(false) {}
+Exporter::Exporter() : stored_result(nullptr), is_exported(false) {}
+
```
myungjoo, line 60: Please dont' use ```#if``` especially when you can handle it with OOP concepts or runtime conditions.

jihochu (approved):
LGTM

PR #2068, opened 2022-12-16
[model] Change the default value of parameter in inference function @open sesame 12/20 16:40

baek2sm:
Change the default value of the "free_mem" parameter to false 
in the inference function.

Since the "free memory" option can be used only when the model 
is in the training mode, the desired results may not be obtained 
if the "free_mem" parameter is set to true when we run inference
without training the model.

Therefore, it is better to set "free_mem" option to true only when
this option is needed.

**Self evaluation:**

Signed-off-by: Seungbaek Hong <sb92.hong@samsung.net>

### DIFF
--- a/nntrainer/models/neuralnet.h
+++ b/nntrainer/models/neuralnet.h
@@ -311,18 +311,20 @@ public:
   /**
    * @brief     Run NeuralNetwork inference
    * @param[in] X input tensor
+   * @param[in] free_mem true to free memory. used only in training mode.
    * @retval shared_ptr<const Tensor>
    */
-  sharedConstTensors inference(sharedConstTensors X, bool free_mem = true);
+  sharedConstTensors inference(sharedConstTensors X, bool free_mem = false);
 
   /**
    * @brief     Run NeuralNetwork inference
    * @param[in] X input tensor
    * @param[in] label label tensor
+   * @param[in] free_mem true to free memory. used only in training mode.
    * @retval shared_ptr<const Tensor>
    */
   sharedConstTensors inference(sharedConstTensors X, sharedConstTensors label,
-                               bool free_mem = true);
+                               bool free_mem = false);
 
   /**
    * @brief     Run the inference of the model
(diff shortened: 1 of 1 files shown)


nntrainer/models/neuralnet.h
```diff
@@ -313,7 +313,7 @@ class NeuralNetwork : public ml::train::Model {
    * @param[in] X input tensor
    * @retval shared_ptr<const Tensor>
    */
-  sharedConstTensors inference(sharedConstTensors X, bool free_mem = true);
+  sharedConstTensors inference(sharedConstTensors X, bool free_mem = false);
```
DonghakPark, line 316: LGTM!!
how about add commnet about function for "user can find default parms"

DonghakPark (approved):

nntrainer/models/neuralnet.h
myungjoo: Please verify backward-compatibility for test-cases and sample applications.
You may start by removing the default value and try to vuild all related test cases and applications. (it will generate build errors in such codes so that you can look at them)

baek2sm: @DonghakPark Thanks! I will add a description of "free_mem" parameter.

baek2sm: @myungjoo Thank you, I will try to verify backward-compatibility!

jijoongmoon (approved):
Nice Catch! LGTM!

myungjoo (approved):

lhs8928:
AFAIK all the layer inputs and outputs are allocated in tensor_pool.
So free_mem values should be true to deallocate these tensors even in inference mode.
As we discussed in offline if the loss layer is added to the model the output is validate even the free_mem value is true.

So in my opinion I think it is better to make free_mem value as false for now but leave it as issue.

baek2sm:
> AFAIK all the layer inputs and outputs are allocated in tensor_pool. So free_mem values should be true to deallocate these tensors even in inference mode. As we discussed in offline if the loss layer is added to the model the output is validate even the free_mem value is true.
> 
> So in my opinion I think it is better to make free_mem value as false for now but leave it as issue.

I will open the issue about the relevant contents, thank you!

songgot (approved):
LGTM

PR #2069, opened 2022-12-19
[Trivial] Fix svace issue @open sesame 12/20 15:37

SeoHyungjun:
- Added variable \'cur\' used for erase.
Current position is stored in cur, and the increment operation is performed on the iterator.
If an erase is required, cur is deleted and the iterator is unaffected.

- The for loop runs only once due to a return.
To resolve this issue, we added a conditional statement to run return when it is not ML_ERROR_NONE.

- The parameters of the functions \'ml_tensors_data_destroy\' and \'ml_tensors_info_destroy\' are both of the types \'ml_tensors_data_h\'.
However, several code are using address values.
Added dereference operator to match parameter type.

### DIFF
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -1286,7 +1286,10 @@ int ml_train_model_get_weight(ml_train_model_h model, const char *layer_name,
   for (unsigned int i = 0; i < dims.size(); ++i) {
     status = ml_tensors_data_set_tensor_data(
       *weight, i, w[i], dims[i].getDataLen() * sizeof(float));
-    return status;
+    if (status != ML_ERROR_NONE) {
+	    ml_tensors_data_destroy(weight);
+      return status;
+    }
   }
 
   return status;
--- a/api/capi/src/nntrainer.cpp
+++ b/api/capi/src/nntrainer.cpp
@@ -189,7 +189,7 @@ static int ml_train_dataset_create(ml_train_dataset_h *dataset,
 template <typename... Args>
 static int ml_train_dataset_add_(ml_train_dataset_h dataset,
                                  ml_train_dataset_mode_e mode,
-                                 ml::train::DatasetType type, Args &&...args) {
+                                 ml::train::DatasetType type, Args &&... args) {
   check_feature_state();
   std::shared_ptr<ml::train::Dataset> underlying_dataset;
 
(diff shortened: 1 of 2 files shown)


SeoHyungjun:
Warning from tuple is not resolved.

api/capi/src/nntrainer.cpp
```diff
@@ -1286,7 +1286,10 @@ int ml_train_model_get_weight(ml_train_model_h model, const char *layer_name,
   for (unsigned int i = 0; i < dims.size(); ++i) {
     status = ml_tensors_data_set_tensor_data(
       *weight, i, w[i], dims[i].getDataLen() * sizeof(float));
-    return status;
+    if (status != ML_ERROR_NONE) {
+	    ml_tensors_data_destroy(weight);
```
myungjoo, line 1290:
```
int ml_tensors_data_destroy (ml_tensors_data_h data)
```
and
```
ml_tensors_data_h *weight
```

Is this valid???

myungjoo: Have you turned compiler-warning-error off?

myungjoo (changes requested):
Incorrect pointer usage. LINE 1282 and 1290 has pointer errors.

DonghakPark (commented):
There is ```Merge <> into <>``` commit
How about using rebase instead leaving such commit message, just make the commit message meaningful

jijoongmoon (approved):
LGTM

lhs8928 (approved):
LGTM

api/capi/src/nntrainer.cpp
SeoHyungjun: The parameters of ml_tensors_data_destroy' and 'ml_tensors_info_destroy' are being used in the wrong type, so we modified it.

nntrainer/tensor/task_executor.cpp
```diff
@@ -48,9 +48,10 @@ void TaskExecutor::cancelAll(void) {
 void TaskExecutor::clean(void) {
   for (auto it = tasks.begin(); it != tasks.end();) {
     auto running = std::get<std::atomic_bool>(it->second).load();
-    if (running == false)
-      tasks.erase(it);
+    auto cur = it;
```
jihochu, line 51:
```suggestion
    if (running == false)
      it = tasks.erase(it);
    else
      it++;
```
It's a bug. please check.

SeoHyungjun: I modified it. Thank you.

jihochu (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2071, opened 2022-12-20
[Tensor] Remove calcGrad step for non-trainable layer

jihochu:
This patch is for implementing tarinable property behavior.

If a layer is set as non-trainable, it doesnot need to execute calcGrad step, so we can remove it from execution order, also skip gradient calculation.

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -85,15 +85,17 @@ int NetworkGraph::compile(const std::string &loss_type) {
 }
 
 void NetworkGraph::setExecutionOrder() {
-  auto max_count = graph.size() * 3;
-  /** @todo: remove backwarding count for non-trainble layers */
-  for (auto iter = cbegin(); iter != cend(); iter++) {
+  auto backward_order = graph.size();
+  for (auto iter = getBackwardingBeginIter(); iter != getBackwardingEndIter();
+       iter++) {
     auto &node = *iter;
-    auto order_idx = iter - cbegin();
+    auto order_idx = getBackwardingEndIter() - iter - 1;
     auto forward_order = order_idx;
-    auto calc_gradient_order = max_count - ((order_idx + 1) * 2);
-    /** calc derivative is called right after calc_gradient */
-    auto calc_derivative_order = calc_gradient_order + 1;
+    auto calc_gradient_order = backward_order;
+    if (node->getTrainable())
+      backward_order++;
+    auto calc_derivative_order = backward_order++;
+
     node->setExecutionOrder(
       {forward_order, calc_gradient_order, calc_derivative_order});
   }
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -824,7 +824,7 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
                                    lnode->getTrainable(), shared_weight_names),
     inputs, outputs,
     tensor_manager->requestTensors(gnode, init_context.getTensorsSpec(),
-                                   shared_tensor_names));
+                                   lnode->getTrainable(), shared_tensor_names));
 
   return outputs;
 }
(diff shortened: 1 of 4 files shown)


jihochu:
This PR is pending until @baek2sm 's  non-trainable layer test.

jihochu:
It passed #2077

nntrainer/graph/network_graph.cpp
```diff
@@ -85,14 +85,26 @@ int NetworkGraph::compile(const std::string &loss_type) {
...
   auto max_count = graph.size() * 3;
-  /** @todo: remove backwarding count for non-trainble layers */
+  size_t not_trainables = 0;
+
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    if (!(*iter)->getTrainable())
+      not_trainables++;
+  }
+
   for (auto iter = cbegin(); iter != cend(); iter++) {
```
lhs8928, line 95: How about we iterate from crbegin to crend? IMO iterate from end to begin will make logic more simple.
Please check https://gist.github.com/lhs8928/5610124f0de7787d1eb1b7248159f864

nntrainer/tensor/manager.cpp
```diff
@@ -355,9 +355,8 @@ std::vector<Weight *> Manager::requestWeights(
     node.getExecutionOrder();
 
   std::vector<unsigned int> var_exec_order(
-    {forwarding_order, calcGradient_order, calcDerivative_order});
-  std::vector<unsigned int> default_grad_exec_order(
-    {calcGradient_order, calcDerivative_order});
+    {forwarding_order, calcDerivative_order});
```
lhs8928, line 358: Can you please describe why did you remove calcGradient_order from var_exec_order?
I think weights are also needed in some cases when calculate gradient.
Please check https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/layers/grucell.cpp#L472

nntrainer/tensor/manager.cpp
```diff
@@ -369,6 +368,10 @@ std::vector<Weight *> Manager::requestWeights(
     auto &[dim, t_initializer, w_reg, w_reg_const, decay, clip_by_global_norm,
            need_gradient, name] = weights_spec.at(i);
     auto grad_exec_order = default_grad_exec_order;
+
+    if (trainable)
```
lhs8928, line 372: This is trivial but I think grad_exec_order already used only the layer is trainable so this change will not affect the weights lifespan.

nntrainer/graph/network_graph.cpp
```diff
@@ -85,14 +85,26 @@ int NetworkGraph::compile(const std::string &loss_type) {
...
-    auto calc_gradient_order = max_count - ((order_idx + 1) * 2);
+    auto calc_gradient_order =
+      max_count - ((order_idx + 1) * 2) - not_trainables;
     /** calc derivative is called right after calc_gradient */
     auto calc_derivative_order = calc_gradient_order + 1;
+    if (!node->getTrainable()) {
+      not_trainables--;
+      calc_gradient_order++;
+      calc_derivative_order = calc_gradient_order;
+    }
```
jijoongmoon, line 107: if we set ```calc_derivative_order = calc_gradient_order```, we might needs more test cases to confirm.

nntrainer/graph/network_graph.cpp
```diff
@@ -85,14 +85,26 @@ int NetworkGraph::compile(const std::string &loss_type) {
 
 void NetworkGraph::setExecutionOrder() {
   auto max_count = graph.size() * 3;
-  /** @todo: remove backwarding count for non-trainble layers */
+  size_t not_trainables = 0;
```
jijoongmoon, line 88: I think, We need to check if the max_count is the same with ```std::get<2>(backward_iter_end->getExecutionOrder()``` . Because the max is used at ```tensor_manager->allocateTensors(std::get<2>(backward_iter_end->getExecutionOrder())```

nntrainer/tensor/manager.cpp
lhs8928: Yeah you're right.
I was confused.

lhs8928 (approved):
LGTM

nntrainer/graph/network_graph.cpp
jihochu: it's changed to iterate reversely with @lhs8928 's comment.  please check.

jijoongmoon (approved):
LGTM

PR #2077, opened 2022-12-22
[test] add test cases that a specific layer is non-trainable

baek2sm:
Add test cases that a specific layer is non-trainable.

- Test based on the pytoch model with two fc hidden layers.
- Add a test case when the first hidden layer is set to non-trainable.
- Add a test case when the second hidden layer is set to non-trainable.

**Self evaluation:**


test/unittest/models/unittest_models.cpp
```diff
@@ -882,6 +932,10 @@ GTEST_PARAMETER_TEST(
                  "transformer_pseudo_bool_attn_mask", ModelTestOption::ALL_V2),
     mkModelIniTc(fc_relu_decay, DIM_UNUSED, NOT_USED_,
                  ModelTestOption::COMPARE_V2),
+    mkModelTc_V2(non_trainable_fc_idx1, "non_trainable_fc_idx1",
```
jihochu, line 935: How about makeing names simliar with others ?  "make...."

test/unittest/models/unittest_models.cpp
```diff
@@ -50,6 +50,56 @@ IniWrapper fc_relu_decay(
...
+get_func_to_create_non_trainable_fc(int idx) {
+
+  char fc1_trainable[17] = "trainable=true";
+  char fc2_trainable[17] = "trainable=true";
+
+  if (idx == 1) {
+    std::strcpy(fc1_trainable, "trainable=false");
+  } else if (idx == 2) {
+    std::strcpy(fc2_trainable, "trainable=false");
+  }
```
jihochu, line 68: This looks good, but Isn't thre any simple code this?

test/unittest/models/unittest_models.cpp
baek2sm: Thank you for your good point. I will correct it!

DonghakPark (approved):
LGTM!!

jihochu (approved):
LGTM

PR #2075, opened 2022-12-22
(Wait for #2041) [SWAP] Implement lookahead(preloading) behavior

jihochu:
It implement lookahead(preloading) behavior.
Preloading is working with cache flush synchroneously, due to
cache flush is done when each execution order finishes.

Signed-off-by: Jiho Chu <jiho.chu@samsung.com>

PR #2078, opened 2022-12-23
[Wait for #2075]  [Planner] Add Weight Gradient Planning

jijoongmoon:
This patch includes the memory optimization for weight
gradient. Weight Gradient can be reused under the not execution order
confiliction. In order to do it, memory pool need to have the data
structure for identification of weight gradient. Therefore, it is
marked when the tensor is request during weight request in mananger
except the weight sharing situation (when is_dependent is true).

Notice, we are not gaurantee the gradient is always initialized
property. So, layer developer, or the user must aware it can be
dirty.

**Self evaluation:**


nntrainer/tensor/manager.cpp
```diff
@@ -387,6 +390,10 @@ std::vector<Weight *> Manager::requestWeights(
                                         var_ls, t_initializer);
 
       if (trainable && need_gradient) {
+        /** We cannot use the tensor shedulding for weight gradient if the
```
baek2sm, line 393: It's trivial, but there's a typo error in the spelling of scheduling.

nntrainer/tensor/optimized_v1_planner.cpp
```diff
@@ -192,6 +227,65 @@ size_t OptimizedV1Planner::planLayout(
     sorted_req.push_back(&req);
   }
 
+  if (wgrad_requests.size()) {
+    /** TODO: We donot need to start from memeory_req. We might find proper
```
baek2sm, line 231: there is a typo error in memory_req !

baek2sm (approved):
LGTM

nntrainer/tensor/manager.cpp
jijoongmoon: fixed. Thanks.

nntrainer/tensor/optimized_v1_planner.cpp
jijoongmoon: fixed. Thanks.

jihochu (approved):
LGTM

baek2sm (approved):
LGTM

PR #2081, opened 2022-12-26
[Utils] Add trace feature

jihochu:
Trace feature enables easily trace information through training step. It can include memory and time tracing, and futher informations could be included.
The memory trace is defaultly written to memory_trace_$PID.log, while time trace is written to time_trace_$PID.log.


nntrainer/utils/tracer.cpp
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2022 Jiho Chu <jhoon.it.lee@samsung.com>
```
DonghakPark, line 3: There is Typo in Copyright
```<jihoon.it.lee@samsung.com> --> <jiho.chu@samsung.com>```

**Review by DonghakPark — APPROVED** [review]
LGTM!

File: nntrainer/utils/tracer.cpp (hunk 1 above)
Comment (jihochu) [reply nntrainer/utils/tracer.cpp]: Thanks. It fixed.

**Comment by SeoHyungjun** [comment]
LGTM!!

**Review by jijoongmoon — COMMENTED** [review]
LGTM, But I wonder how about using Profiler rather than making another Tracer for consistancy. Maybe it might need to make another profiler listener in profile.

File: nntrainer/utils/meson.build [hunk 2]
```diff
@@ -1,6 +1,7 @@
 util_sources = [
   'util_func.cpp',
   'profiler.cpp',
+  'tracer.cpp',
```
myungjoo, line 4: add tracer.cpp only when enable-trace is true. That helps build-modularizing.

nntrainer/utils/tracer.h
```diff
@@ -0,0 +1,222 @@
...
+#include <exception>
+#include <fstream>
+#include <list>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <tuple>
+#include <type_traits>
+
+namespace nntrainer {
```
myungjoo, line 26: Can you hide the class Tracer if TRACER is not defined? (it will make the resulting binary even lighter)

nntrainer/utils/meson.build
jihochu: Thanks. It modifed.

nntrainer/utils/tracer.h
jihochu: It's wrapped with ifdef.

jihochu:
> LGTM, But I wonder how about using Profiler rather than making another Tracer for consistancy. Maybe it might need to make another profiler listener in profile.

Profiler is little heavy and it used to analyze more detailed behaviors, like average, swap info... , so it also consumes more resources. Tracer overcomes it with ligtweight implementation. It only trace each trace point with little information.
I modulized Tracer as @myungjoo 's gudie, so it doesnot compiled when it turned off, and it would be helpful for light profiling.

myungjoo (approved):

PR #2082, opened 2022-12-26
[API] set default value on createModel function

baek2sm:
the default value of the "type" parameter of the "createModel" function is set to NEURAL_NET.

the createModel function receives the model type as a parameter, but the only value that can be set is NEURAL_NET.
(KNN is also not supported by the createModel function in API)

Thus, it is inefficient to set the parameter value every time because there is no other option.

**Self evaluation:**


jijoongmoon (approved):
LGTM

myungjoo (approved):

songgot (approved):
LGTM

Issue #2086, opened 2022-12-28
[graph] Undefined behavior


SeoHyungjun:
there are undefined behavior in cbegin() and cend() of graph/graph_core.h.

cbegin() and cend() return graph_const_iterator<T>(&(*node_list.cbegin())) and graph_const_iterator<T>(&(*node_list.cend())).

But the dereference to end iterator is undefined behavior.

If Begin is also empty, it is the same as end, which may cause problems. However, it does not occur in begin because model_graph.empty() is checked.

PR #2088, opened 2022-12-29
[Graph] Fix cend function

SeoHyungjun:
- Fix undefined behavior in cend() of \'graph_core.h\'

  The cend() function dereference node_list.cend(). This is undefined
  behavior because it accesses memory in an area that is not actually
  allowed. So changed cend() to operate normally through cbegin() + size.

  However, these changes also include problems. Usually, check that
  node_list is empty, and if it is, it will be treated as an exception.
  if use cbegin() and cend() when node_list is empty, there might be a
  problem. So, Added this caution at @note.

- Fixed a minor typo.

Related : #2086

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/graph_core.h
+++ b/nntrainer/graph/graph_core.h
@@ -101,6 +101,9 @@ public:
   /**
    * @brief     get begin iterator for the forwarding
    * @retval    const iterator marking the begin of forwarding
+   * @note      this function should not be used when node_list is empty.
+   * if node_list is empty, end iterator is dereferenced,
+   * This action is undefined behavior.
    */
   template <
     typename T = GraphNode,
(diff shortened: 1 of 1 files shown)


DonghakPark (approved):
LGTM!

nntrainer/graph/graph_core.h
```diff
@@ -114,16 +117,19 @@ class GraphCore {
 
   /**
    * @brief     get end iterator for the forwarding
-   * @retval    const iterator marking the emd of forwarding
+   * @retval    const iterator marking the end of forwarding
+   * @note      this function should not be used when node_list is empty.
+   * if node_list is empty, end iterator is dereferenced,
```
jijoongmoon, line 122: how about we check the node_list.empty() and do something before return.

jihochu: Usually, empty check is user's responsiblity.
https://cplusplus.com/reference/vector/vector/cbegin/
```suggestion
   * if node_list is empty, end iterator shall not be dereferenced,
```

jijoongmoon (approved):
LGTM

PR #2089, opened 2022-12-29
[test] add test case when a output fc layer is non-trainable

baek2sm:
Add a test case when a output fc layer is non-trainable.

**Self evaluation:**


jijoongmoon (approved):
LGTM.

SeoHyungjun:
LGTM!

myungjoo (commented):
unittest_models_v2.tar.gz ??? what's this?

baek2sm:
> unittest_models_v2.tar.gz ??? what's this?

The compressed file includes binary files in which correct answer values for model verification are stored. For example, the result values obtained after executing the same code in the pytorch are made into binary files and used to check if they are the same as the result values obtained from the NNTrainer. Previously, there were 52 binary test files, and I added the following 2 binary files for this test case.

- non_trainable_fc_idx1. nnmodelgolden
- non_trainable_fc_idx2. nnmodelgolden

jihochu (approved):
LGTM

myungjoo (approved):

PR #2093, opened 2023-01-02
[Memory Planner] Update the Memory Planner 

jijoongmoon:
This PR includes,
  1. assigning right execution order depending on layer type.
     : We do need to move those fixes in to each layer
  2. Update memory Planner to use the memory with smaller size

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/graph_node.h
+++ b/nntrainer/graph/graph_node.h
@@ -39,7 +39,8 @@ public:
    * This ensures that the operations are executed in the order of their
    * listing.
    */
-  typedef std::tuple<unsigned int, unsigned int, unsigned int, unsigned int> ExecutionOrder;
+  typedef std::tuple<unsigned int, unsigned int, unsigned int, unsigned int>
+    ExecutionOrder;
 
   /**
    * @brief     Destructor of Layer Class
@@ -69,6 +70,13 @@ public:
    * @return const std::string type representation
    */
   virtual const std::string getType() const = 0;
+
+  /**
+   * @brief     Get the trainable parameter
+   *
+   * @return bool true / false
+   */
+  virtual bool getTrainable() const = 0;
 
   /**
    * @brief     Get the input connections for this node
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -234,7 +234,7 @@ public:
    */
   static VarGradSpecV2
   outSpec(const TensorDim &dim, const std::string &name = "out",
-          TensorLifespan ls = TensorLifespan::FORWARD_GRAD_LIFESPAN,
+          TensorLifespan ls = TensorLifespan::FORWARD_FUNC_LIFESPAN,
           TensorLifespan grad_ls = TensorLifespan::BACKWARD_FUNC_LIFESPAN);
 
   /**
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -235,7 +235,7 @@ public:
   static VarGradSpecV2
   outSpec(const TensorDim &dim, const std::string &name = "out",
           TensorLifespan ls = TensorLifespan::FORWARD_FUNC_LIFESPAN,
-          TensorLifespan grad_ls = TensorLifespan::BACKWARD_FUNC_LIFESPAN);
+          TensorLifespan grad_ls = TensorLifespan::CALC_GRAD_DERIV_LIFESPAN);
 
   /**
    * @brief request outputs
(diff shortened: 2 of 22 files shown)


DonghakPark (approved):
LGTM!

baek2sm (approved):
LGTM

nntrainer/graph/graph_node.h
```diff
@@ -39,7 +39,8 @@ class GraphNode {
    * This ensures that the operations are executed in the order of their
    * listing.
    */
-  typedef std::tuple<unsigned int, unsigned int, unsigned int, unsigned int> ExecutionOrder;
+  typedef std::tuple<unsigned int, unsigned int, unsigned int, unsigned int>
```
SeoHyungjun, line 42: Wouldn't it be more concise if changed unsigned int to uint?

SeoHyungjun (approved):
LGTM!

nntrainer/graph/graph_node.h
jijoongmoon: we are using unsigned int so many point in NNTrainer. may be separate PR might better to change everywhere.

nntrainer/tensor/optimized_v1_planner.cpp
```diff
@@ -192,6 +215,65 @@ size_t OptimizedV1Planner::planLayout(
...
+        if (merge) {
+          req.offset = sr.mem_req->offset;
+          memory_offset[req.loc] = req.offset;
+          replace_and_fill = true;
+          wgrad_sorted_req[idx].start_end.push_back(
+            std::make_pair(req.start, req.end));
+          break;
+        } else {
+          replace_and_fill = false;
+        }
```
jihochu: else loop is unnecessary.

nntrainer/tensor/manager.cpp
```diff
@@ -534,9 +541,18 @@ Manager::requestInputs(const GraphNode &node,
...
       node.getType() == MultiOutLayer::type or
       node.getType() == BatchNormalizationLayer::type or
-      node.getType() == LayerNormalizationLayer::type)
+      node.getType() == LayerNormalizationLayer::type or !node.getTrainable())
     var_common_spec.ls = TensorLifespan::FORWARD_FUNC_LIFESPAN;
 
+  if (node.getType() == MSELossLayer::type or
+      node.getType() == CrossEntropySoftmaxLossLayer::type or
+      node.getType() == CrossEntropySigmoidLossLayer::type)
+    var_common_spec.ls = TensorLifespan::FORWARD_DERIV_LIFESPAN;
```
jihochu, line 550: Looks good.

nntrainer/tensor/optimized_v2_planner.cpp
```diff
@@ -282,15 +286,11 @@ size_t OptimizedV2Planner::planLayout(
       wgrad_sorted_req.push_back(WGradMemoryRequest(&req));
       wgrad_sorted_req.back().start_end.push_back(
         std::make_pair(req.start, req.end));
+#ifdef DEBUG
       new_grad_cnt++;
       new_grad_size += req.size;
+#endif
     }
-
-    ml_logd("Total Requested Memory(OPTV2): %lf MiB>>>>>>>> \n - new mem for "
```
jihochu: Isn't it better to keep this log for debugging?

nntrainer/layers/layer_context.h
```diff
@@ -235,7 +235,7 @@ class InitLayerContext {
   static VarGradSpecV2
   outSpec(const TensorDim &dim, const std::string &name = "out",
           TensorLifespan ls = TensorLifespan::FORWARD_FUNC_LIFESPAN,
-          TensorLifespan grad_ls = TensorLifespan::BACKWARD_FUNC_LIFESPAN);
+          TensorLifespan grad_ls = TensorLifespan::CALC_GRAD_DERIV_LIFESPAN);
```
jihochu, line 238: `BACKWARD_FUNC_LIFESPAN` contains `apply gradient` step.
IMHO, Isn't it better to remain it as default while `calc grad` is set?

jijoongmoon: In that case, we will have unnecessary execution order for the input tensor, which is not desirable.  We do not need the outgoing gradient to compute the apply gradient.

nntrainer/tensor/optimized_v2_planner.cpp
jijoongmoon: well, may be, it's for later pr, then.

nntrainer/tensor/optimized_v1_planner.cpp
jijoongmoon: removed. Thanks.

nntrainer/layers/layer_context.h
jihochu: OK. thanks fo checking.

jihochu (approved):
LGTM

Issue #2031, opened 2022-10-31
NNTrainer - Windows build Test (in Visual Studio)


DonghakPark:
for support windows system

- [x] Dependency Test
- [x] meson build Test
- [x] Application/API Test

PR #2096, opened 2023-01-10
[Windows] Modify for Windows porting

SeoHyungjun:
**Modify grammar that is not available in Visual Studio**

- The original variable is unsigned int type. It is impossible to put a minus sign. So changed it to multiplying by -1.

- Visual Studio determines the Variable Argument(args...) in the macro as error. Change 'args...' and '#args' to '...' and '_VA_ARGS__' specified in C99.

- POSIX names for getpid and close are deprecated. Use compatible _getpid and _close in ISO C and C++.

- Add dllexport to use function of nntrainer.dll in application

**Add header and data type for windows**

- In visual studio, must add a string header to use std::string. On Linux, can use it without a string header.

- In Windows, can use io.h instead of unistd.h.

- dlfcn.h is a header for linux.

- Need numeric.h to use accumulate and iota.

- ssize_t and mode_t are not defined in windows. Map SSIZE_T to ssize_t (SSIZE_T is defined in BaseTsd.h). Map int to mode_t.

- Disable malloc_trim in windows.

- Add two submodules to use the code of mman and iniparser.

**Update submodule iniparser**

 - Add header and data type for windows

**Submodule iniparser and mman were added for testing. I will add windows-resource repository later (ex. nnstreamer-android-resource for Android) and remove it.**

Related : #2031

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/dataset.h
+++ b/api/ccapi/include/dataset.h
@@ -80,7 +80,7 @@ public:
  * @param properties property representations
  * @return std::unique_ptr<Dataset> created dataset
  */
-std::unique_ptr<Dataset>
+ML_API std::unique_ptr<Dataset>
 createDataset(DatasetType type,
               const std::vector<std::string> &properties = {});
 
@@ -92,7 +92,7 @@ createDataset(DatasetType type,
  * @param properties property representations
  * @return std::unique_ptr<Dataset> created dataset
  */
-std::unique_ptr<Dataset>
+ML_API std::unique_ptr<Dataset>
 createDataset(DatasetType type, const char *path,
               const std::vector<std::string> &properties = {});
 
@@ -105,7 +105,7 @@ createDataset(DatasetType type, const char *path,
  * @param properties property representations
  * @return std::unique_ptr<Dataset> created dataset
  */
-std::unique_ptr<Dataset>
+ML_API std::unique_ptr<Dataset>
 createDataset(DatasetType type, datagen_cb cb, void *user_data = nullptr,
               const std::vector<std::string> &properties = {});
 } // namespace train
--- a/api/ccapi/include/common.h
+++ b/api/ccapi/include/common.h
@@ -19,6 +19,17 @@
 
 #include <nntrainer-api-common.h>
 
+/**
+ * @brief Defining Macros for Exporting Dynamic Library in Windows.
+ * @note Macro is meaningful only if platform is windows.
+ */
+#ifdef _WIN32
+#define ML_API __declspec(dllexport)
+#else
+#define ML_API
+#endif
+
+
 namespace ml {
 namespace train {
 
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -12,7 +12,9 @@
  *
  */
 #include <dirent.h>
+#ifndef _WIN32
 #include <dlfcn.h>
+#endif
 #include <iostream>
 #include <sstream>
 #include <string>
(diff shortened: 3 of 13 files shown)


api/ccapi/include/dataset.h
```diff
@@ -80,6 +80,9 @@ class Dataset {
  * @param properties property representations
  * @return std::unique_ptr<Dataset> created dataset
  */
+#ifdef _WIN32
```
jaeyun-jung, line 83: Add macro in common header?
e.g., 
```
#ifdef _WIN32
NNTRAINER_API __declspec(dllexport)
#else
NNTRAINER_API
#endif
```

SeoHyungjun: Added macro and modified code.

api/nntrainer-api-common.h
```diff
@@ -239,6 +239,16 @@ typedef enum {
          required for inference and training. */
 } ml_train_model_format_e;
 
+/**
+ * @brief Defining Macros for Exporting Dynamic Library in Windows.
+ * @note Macro is meaningful only if platform is windows.
+ */
+#ifdef _WIN32
+#define ML_API __declspec(dllexport)
```
jaeyun-jung, line 247: let's check this declaration after paper work is done.
nntrainer-api-common.h is header for Tizen native API set, so I think this may raise an error when checking Tizen release process.

Recommend: add this declaration in nntrainer internal header file, not in Tizen header.

SeoHyungjun: Changed the position of the define to api/ccapi/include/common.h.

songgot (approved):
LGTM

DonghakPark (approved):
LGTM!

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "tools/mman_win32"]
+	path = tools/mman_win32
+	url = https://github.com/SeoHyungjun/mman-win32.git
+[submodule "tools/iniparser_win32"]
+	path = tools/iniparser_win32
+	url = https://github.com/SeoHyungjun/iniparser.git
```
myungjoo, line 6: 1. .gitmodules is unconditionally activated regardless of OS. So, please get these sources in a Windows build script. (or a script called by Windows build script.)
2. Don't use your personal git repo for this.

SeoHyungjun: 1. Submodules have been added to conduct unit tests in Windows first. So I tried to make a script or batch file later. 
Let me write a script first. Is there no problem if I make it with python?

2. If I can't use personal git repo, can add a window-resource repo to nnstreamer?

myungjoo: 1. No. python is an overkill and you are creating another dependency. Use something already have: Windows Shell (batch file) or VS Project Script.

2. Yes.

myungjoo: Anyway, you really need to mark a git commit ID for both packages regardless whether you use gitmodules or not.

You may use gitmodules and fix it later or add a note to gitmodules and README that gitmoduels are required for Windows builds only.

myungjoo (approved):
@SeoHyungjun please read the last comment before merging it.

jihochu (approved):
LGTM

jijoongmoon:
We open windows resource repo here and please use it.
https://github.com/nnstreamer/nnstreamer-windows-resource

SeoHyungjun:
We are going to change to use nnstreamer-windows-resource instead of submodule.

I'll fix this PR after the windows-resource repo update.

PR #2098, opened 2023-01-25
[SWAP] task executor: Remove stop and clean feature

jihochu:
It's changed to use work queue to manage executions. There is some overhead for stop and cancel behaviors, so they are removed temporaly now, due to the performance.

### DIFF
--- a/nntrainer/tensor/task.h
+++ b/nntrainer/tensor/task.h
@@ -152,14 +152,14 @@ public:
    *
    * @param time timeout in T
    */
-  virtual void setTimeout(int time) { timeout = T(time); }
+  virtual void setTimeout(int64_t time) { timeout = T(time); }
 
   /**
    * @brief Get timeout
    *
    * @param time timeout in T
    */
-  virtual int getTimeout(void) const { return timeout.count(); }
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/task.h
```diff
@@ -152,14 +152,14 @@ class TaskAsync : public Task {
...
-  virtual void setTimeout(int time) { timeout = T(time); }
+  virtual void setTimeout(long time) { timeout = T(time); }
 
   /**
    * @brief Get timeout
    *
    * @param time timeout in T
    */
-  virtual int getTimeout(void) const { return timeout.count(); }
+  virtual long getTimeout(void) const { return timeout.count(); }
```
myungjoo, line 162: Why long? Is this coming from ```__kernel_time_t``` or ```time.h```?

jihochu: Thanks for checking that. 
I expect std::chrono::xxx::count(). (https://en.cppreference.com/w/cpp/chrono/duration)
so, I modifed to use int64_t.

jijoongmoon (approved):
LGTM

PR #2102, opened 2023-02-01
[Application] Add Simple Layers NNtrainer example(FC, LSTM, Conv)

DonghakPark:
Add Simple Layers NNtrainer example(FC, LSTM, Conv)  with Dummy Data
- Linear(Fully-Connected)
- Conv
- LSTM

We conduct Memory & Latency Benchmark test based on these code It's Loss possibly show inf cause It's DataSet is Dummy If you want to get actual Loss set your own Dataset
- It support only training if user want they can set dataset, validation loss, test loss... etc

**Self evaluation:**

Signed-off-by: DongHak Park <donghak.park@samsung.com>

### DIFF
--- a/Applications/Layers/README.md
+++ b/Applications/Layers/README.md
@@ -1,0 +1,46 @@
+# Layers
+
+In this example, we demonstrate training Single Layers model with dummy dataset.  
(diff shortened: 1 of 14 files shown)


myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #2103, opened 2023-02-01
[activation] add swish activation function @open sesame 02/28 16:20

baek2sm:
Added the swish activation function.

It needs input of activation and output of activation
to calculate derivative value.

So, I overload run_prime_fn function to use
input and output for calculating derivative.

- add swish activation
- add test case about swish activation function

**Issue**
Currently, this swish activation function is NOT SUPPORTED
in RNN, LSTM, GRU and Attention layers. 

I opened issue about it : #2118
I will resolve this problem later. 

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/acti_func.cpp
+++ b/nntrainer/layers/acti_func.cpp
@@ -39,6 +39,24 @@ int ActiFunc::setActivation(
   std::function<Tensor &(Tensor &, Tensor &, Tensor const &)> const
     &activation_prime_fn) {
   _act_fn = activation_fn;
+  _act_prime_fn =
+    [activation_prime_fn](Tensor const &t_in, Tensor &t_out,
+                          Tensor &outgoing_derivative,
+                          Tensor const &incoming_derivative) -> Tensor & {
+    return activation_prime_fn(t_out, outgoing_derivative, incoming_derivative);
+  };
+
+  return ML_ERROR_NONE;
+}
+
+int ActiFunc::setActivation(
+  std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
+  std::function<Tensor &(Tensor const &, Tensor const &, Tensor &,
+                         Tensor const &)> const &activation_prime_fn) {
+  if (in_place)
+    return ML_ERROR_INVALID_PARAMETER;
+
+  _act_fn = activation_fn;
   _act_prime_fn = activation_prime_fn;
 
   return ML_ERROR_NONE;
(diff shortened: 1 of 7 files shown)


nntrainer/layers/acti_func.cpp
```diff
@@ -39,6 +39,24 @@ int ActiFunc::setActivation(
...
+  };
+
+  return ML_ERROR_NONE;
+}
+
+int ActiFunc::setActivation(
+  std::function<Tensor &(Tensor const &, Tensor &)> const &activation_fn,
+  std::function<Tensor &(Tensor const &, Tensor const &, Tensor &,
+                         Tensor const &)> const &activation_prime_fn) {
+  if (in_place)
```
jijoongmoon, line 56: this means that this function is only for the non-inplace ActiFunc?  I'm not sure why we have to set this limitation for this function. It seems to me, something like it removes the opportunity to use this for other ActiFunc for later use.

nntrainer/layers/acti_func.cpp
```diff
@@ -299,6 +332,54 @@ float ActiFunc::leakyReluPrime(float x) {
...
+Tensor &ActiFunc::swish(Tensor const &t_in, Tensor &t_out) {
+  unsigned int batch = t_in.batch();
+  unsigned int channel = t_in.channel();
+  unsigned int height = t_in.height();
+  unsigned int width = t_in.width();
+
+  for (unsigned int b = 0; b < batch; ++b) {
+    for (unsigned int c = 0; c < channel; ++c) {
+      for (unsigned int h = 0; h < height; ++h) {
+        for (unsigned int w = 0; w < width; ++w) {
```
jijoongmoon, line 344: This might be better to use Tensor Op. 
```c++
t_out.apply_i([&](float x){ return sigmoid(x);});
t_out.multiply_i(t_in);
```

nntrainer/layers/acti_func.cpp
baek2sm: Yes, this function is only for non-inplace activation function. Each of the four parameters of activation_prime_fn means input tensor, output tensor, outgoing derivation, and incoming derivation. In addition, the fact that input and output tensors are independent during backwarding means that they do not support in-place operation(if it supports in-place, then they do not need input and output tensors independently). So, I don't think it makes sense to support the in-place operation here.

nntrainer/layers/acti_func.cpp
baek2sm: Thanks, I correct it.

nntrainer/layers/acti_func.cpp
```diff
@@ -39,6 +39,24 @@ int ActiFunc::setActivation(
   std::function<Tensor &(Tensor &, Tensor &, Tensor const &)> const
     &activation_prime_fn) {
   _act_fn = activation_fn;
+  _act_prime_fn =
+    [activation_prime_fn](Tensor const &t_in, Tensor &t_out,
+                          Tensor &outgoing_derivative,
+                          Tensor const &incoming_derivative) -> Tensor & {
+    return activation_prime_fn(t_out, outgoing_derivative, incoming_derivative);
```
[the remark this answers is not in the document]
baek2sm: When calling run_prime_fn in calcDerivative of activation_layer.cpp, we do not know the type of function called, so I passed on four parameters considering that four parameters are required. For this reason, a function prototype of _act_fn was created based on four parameters. If we want to divide it with the case of using 3 parameters, we need to change the activation_type variable to public, check activation_type in calcDerive, and then decide whether to deliver 3 parameters or 4 parameters. I'm not sure which way is better.

nntrainer/layers/acti_func.cpp
```diff
@@ -299,6 +332,47 @@ float ActiFunc::leakyReluPrime(float x) {
...
+
+  for (unsigned int b = 0; b < batch; ++b) {
+    for (unsigned int c = 0; c < channel; ++c) {
+      for (unsigned int h = 0; h < height; ++h) {
+        for (unsigned int w = 0; w < width; ++w) {
+          outgoing_derivative.setValue(b, c, h, w,
+                                       t_out.getValue(b, c, h, w) +
+                                         sigmoid(t_in.getValue(b, c, h, w)) *
+                                           (1 - t_out.getValue(b, c, h, w)));
+        }
```
jijoongmoon, line 366: I guess you can also write these function with Tensor Op.

baek2sm: > I guess you can also write these function with Tensor Op. Cause, when you use the tensor op, you can get more speedup due to the blas op in Tensor.

thanks, I correct these operations!

jijoongmoon (approved):
LGTM

DonghakPark (approved):
LGTM!

PR #2110, opened 2023-02-10
[Application] Save model trained from pytorch as nntrainer model format @open sesame 03/08 07:58

baek2sm:
I modified resnet example to save model trained from pytorch as nntrainer binary model format.
"translayer_v2.py" is renamed to "torchconverter.py" and moved to tools folder so that it can be used for common modules.

- the pytorch resnet example model is modified to be the same as the nntrainer resnet example model.

### DIFF
--- a/Applications/Resnet/PyTorch/main.py
+++ b/Applications/Resnet/PyTorch/main.py
@@ -16,6 +16,22 @@ from torch.utils.data import DataLoader
 from torchvision import datasets
 from torchvision.transforms import ToTensor, transforms
 
+import sys
+import os
+
+# get pyutils path using relative path
+def get_util_path():
+    current_path = os.path.abspath(os.path.dirname(__file__))
+    parent_path = os.path.abspath(os.path.dirname(current_path))
+    target_path = os.path.abspath(os.path.dirname(parent_path))
+    return os.path.dirname(target_path) + '/tools/pyutils/'
+
+# add pyutils path to sys.path
+sys.path.append(get_util_path())
+print(get_util_path())
+from torchconverter import save_bin
+
+
 DEVICE = "cpu"
 print(f"Using {DEVICE} device")
 print(f"PyTorch version: {torch.__version__}")
@@ -35,25 +51,27 @@ class BasicBlock(nn.Module):
     def __init__(self, in_planes, planes, stride=1):
         super(BasicBlock, self).__init__()
         self.conv1 = nn.Conv2d(
-            in_planes, planes, kernel_size=3, stride=stride, padding=1, bias=False
+            in_planes, planes, kernel_size=3, stride=stride, padding=1, bias=True
         )
         self.bn1 = nn.BatchNorm2d(planes)
-        self.conv2 = nn.Conv2d(planes, planes, kernel_size=3, stride=1, padding=1, bias=False)
-        self.bn2 = nn.BatchNorm2d(planes)
+        self.conv2 = nn.Conv2d(planes, planes, kernel_size=3, stride=1, padding=1, bias=True)
 
         self.shortcut = nn.Sequential()
         if stride != 1 or in_planes != self.expansion * planes:
             self.shortcut = nn.Sequential(
                 nn.Conv2d(
-                    in_planes, self.expansion * planes, kernel_size=1, stride=stride, bias=False
+                    in_planes, self.expansion * planes, kernel_size=1, stride=stride, bias=True
                 ),
-                nn.BatchNorm2d(self.expansion * planes),
+                # nn.BatchNorm2d(self.expansion * planes),
             )
+        
+        self.bn2 = nn.BatchNorm2d(planes)
 
     def forward(self, x):
         out = F.relu(self.bn1(self.conv1(x)))
-        out = self.bn2(self.conv2(out))
+        out = self.conv2(out)
         out += self.shortcut(x)
+        out = self.bn2(out)
         out = F.relu(out)
         return out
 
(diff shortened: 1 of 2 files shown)


Applications/Resnet/PyTorch/recorder_v2.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+##
+# Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
```
jihochu, line 4: 2023  Seungbaek Hong ?

Applications/Resnet/PyTorch/main.py
```diff
@@ -16,6 +16,14 @@
 from torchvision import datasets
 from torchvision.transforms import ToTensor, transforms
 
+import sys
+import os
+import numpy as np
+from collections.abc import Iterable
+
+from recorder_v2 import _get_writer
```
jihochu, line 24: `recorder_v2` seems not to want to expose `_get_writer`.

DonghakPark (commented):
As you commented in commit message ```reorder_v2.py``` and ```transLayer_v2.py``` , these files are copy of ```test/input_gen/{}```
: **How about move them to ```utils```, so we can use it on other Application**

Applications/Resnet/PyTorch/recorder_v2.py
DonghakPark: these files are copy of test/input_gen --> so i think there is no change

baek2sm: I copied that python script temporarily. I will fix it to import original python script today. thanks!

Applications/Resnet/PyTorch/main.py
baek2sm: That's a good point. I will consider to make independent utility for this task.

baek2sm:
> As you commented in commit message `reorder_v2.py` and `transLayer_v2.py` , these files are copy of `test/input_gen/{}` : **How about move them to `utils`, so we can use it on other Application**

I think that is a good idea. I'll reflect it.

DonghakPark (approved):
LGTM!

Applications/Resnet/PyTorch/transLayer_v2.py
```diff
@@ -0,0 +1,208 @@
...
+# Copyright (C) 2021 Jihoon Lee <jhoon.it.lee@samsung.com>
+#
+# @file transLayer_v2.py
+# @date 19 October 2021
+# @brief Rewrite parameters in the order consistent with nntrainer for the torch model
+# @author Jihoon lee <jhoon.it.lee@samsung.com>
+
+import torch
+from collections.abc import Iterable
+from zoneout import Zoneout
```
lhs8928, line 13: We don't need zoneout in resnet so we could delete this

test/input_gen/transLayer_v2.py
```diff
@@ -56,6 +56,12 @@ def bn1d_translate(model):
     mu, var, _ = [(name, tensor.detach()) for name, tensor in model.named_buffers()]
     yield from [mu, var, gamma, beta]
 
+@register_for_(torch.nn.BatchNorm2d)
```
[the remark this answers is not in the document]
baek2sm: > If bn2d_translate is identical with bn1d_translate please register only once. Please refer line 91

I didn't know there was a way to handle it at once. thanks :)

Applications/Resnet/PyTorch/transLayer_v2.py
baek2sm: > We don't need zoneout in resnet so we could delete this

I will reflect your comment. thanks :)

jihochu (approved):
LGTM

test/input_gen/transLayer_v2.py
```diff
@@ -50,8 +50,8 @@ def transpose_(weight):
         new_params = [transpose_(params[0])]
     yield from new_params
 
-@register_for_(torch.nn.BatchNorm1d)
-def bn1d_translate(model):
+@register_for_((torch.nn.BatchNorm1d, torch.nn.BatchNorm2d))
+def bn_translate(model):
```
jijoongmoon, line 54: Is n't it same file with pyutils/transLayer_v2.py? If it is then it might be better to place it into tools with proper naming. 
How about TorchConverter?

Applications/Resnet/PyTorch/main.py
```diff
@@ -90,6 +107,17 @@ def forward(self, x):
         out = self.linear(out)
         return out
 
+    def save_bin(self, name):
```
jijoongmoon, line 110: This could be in transLayer_v2.py.

baek2sm: > This could be placed in transLayer_v2.py.

That's a good idea. I'll move it to transLayer_v2.py. thanks!

test/input_gen/transLayer_v2.py
baek2sm: > Is n't it same file with pyutils/transLayer_v2.py? If it is then it might be better to place it into tools with proper naming. How about TorchConverter?

Okay, then I will move this script to tools/pyutils/TorchConverter.py for using common module. thanks!

baek2sm: It's done. thank you :)

jijoongmoon:
The input generator for unittest also uses this. So how about changing it to use this in later PR?

baek2sm:
> The input generator for unittest also uses this. So how about changing it to use this in later PR?

I didn't know about it. Then, I will leave original translayer_v2.py script for input generator.

Then, I will try to integrate the translayer scripts in the next PR. Thanks.

SeoHyungjun (approved):
LGTM

Applications/Resnet/PyTorch/main.py
```diff
@@ -109,7 +127,7 @@ def train(dataloader, model, loss_fn, optimizer):
         optimizer.zero_grad()
         loss.backward()
         optimizer.step()
-
+
```
jijoongmoon, line 130: remove tab?

baek2sm: i removed it. thanks!

PR #2130, opened 2023-02-28
[GitAction] Fix Gitaction Auto Labeler

DonghakPark:
### Problem

- Request for removal when there is no label or request for addition when there is a label

### Fix

- Add is_Contain Variable to upload & in labeler they check it already contain label or not
- In gitaction document their api use name : "Need Review" not labels : ["Need Review"]
- Remove unused Debug line
- Reorder step


jijoongmoon (approved):
LGTM

Issue #2108, opened 2023-02-09
Refactoring for several issues


jihochu:
There were serveral issues while reviwing PRELOADING PRs.
Some of them are postponed, this issue handles them.

- [x]  forwarding_op
        Issue: https://github.com/nnstreamer/nntrainer/pull/2075#discussion_r1070893524 
        fix: #2131 
- [x]  booking execution order items
       Issue: https://github.com/nnstreamer/nntrainer/pull/2036#discussion_r1029281223  
       fix: #2041

jihochu:
All todo is cleared.

PR #2131, opened 2023-03-03
[NEURALNET] Refine forwarding operation @open sesame 03/03 16:35

jihochu:
This patch refines forwading operation in neural network class.

The code depth is not matched for forward and backward operations. For backwarding operation, there is backwarding_op and it pasded to the graph, which can handle the operation.
To keep same depth for forwarding, this patch adds forwarding_op function, and passed to the graph class.

Releated Issue:
\#2108

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -344,19 +344,15 @@ void NetworkGraph::applyGradients(
   }
 }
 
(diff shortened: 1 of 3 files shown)


baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM!

PR #2132, opened 2023-03-03
[Application] add object detection example using pytorch

baek2sm:
Add YOLO v2 object detection example using pytorch.

It was implemented in the same way as the YOLO v2 model, but only the backbone model was made into a simpler cnn model to make it easier to configure the initial version of YOLO to be implemented by NNTrainer.

In order to clarify the parts to be implemented by NNTrainer, it is a code that constructs only the basic structure proposed in the original yolo paper, excluding several third-part training techniques.

### DIFF
--- a/Applications/YOLO/PyTorch/dataset.py
+++ b/Applications/YOLO/PyTorch/dataset.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2023 Seungbaek Hong <sb92.hong@samsung.com>
+#
+# @file dataset.py
+# @date 8 March 2023
+# @brief Define dataset class for yolo
+#
+# @author Seungbaek Hong <sb92.hong@samsung.com>
+
+import glob
+import numpy as np
+import torch
+from torch.utils.data import Dataset
+from torch.utils.data.dataloader import default_collate
+from PIL import Image
+
+##
+# @brief dataset class for yolo
+# @note Need annotation text files corresponding to the name of the images.    
+class YOLODataset(Dataset):
+    def __init__(self, img_dir, ann_dir):
+        super().__init__()
+        img_list = glob.glob(img_dir)
+        ann_list = glob.glob(ann_dir)
+        img_list.sort(), ann_list.sort()
+    
+        self.length = len(img_list)
+        self.input_images = []
+        self.bbox_gt = []
+        self.cls_gt = []
+
+        for i in range(len(img_list)):
+            img = np.array(Image.open(img_list[i])) / 255
+            label_bbox = []
+            label_cls = []
+            with open(ann_list[i], 'rt') as f:
+                for line in f.readlines():
+                    line = [int(i) for i in line.split()]
+                    label_bbox.append(np.array(line[1:], dtype=np.float32) / 416)
+                    label_cls.append(line[0])
+                    
+            self.input_images.append(img)
+            self.bbox_gt.append(label_bbox)
+            self.cls_gt.append(label_cls)
+        
+        self.input_images = np.array(self.input_images)
+        self.input_images = torch.FloatTensor(self.input_images).permute((0, 3, 1, 2))
+        
+    def __len__(self):
+        return self.length
+    
+    def __getitem__(self, idx):
+        return self.input_images[idx], self.bbox_gt[idx], self.cls_gt[idx]
+    
+##
+# @brief collate db function for yolo
+def collate_db(batch):
+    """
+    @param batch list of batch, (img, bbox, cls)
+    @return collated list of batch, (img, bbox, cls)
+    """
+    items = list(zip(*batch))
+    items[0] = default_collate(items[0])
+    items[1] = list(items[1])
+    items[2] = list(items[2])
+    return items
--- a/Applications/YOLO/PyTorch/main.py
+++ b/Applications/YOLO/PyTorch/main.py
@@ -1,0 +1,128 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2023 Seungbaek Hong <sb92.hong@samsung.com>
+#
+# @file main.py
+# @date 8 March 2023
+# @brief Implement training for yolo
+#
+# @author Seungbaek Hong <sb92.hong@samsung.com>
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.optim as optim
+import torch.nn.functional as F
+from torch.utils.data import DataLoader
+
+from yolo import YoloV2_light
+from yolo_loss import YoloV2_LOSS
+from dataset import YOLODataset, collate_db
+
+
+# set config
+out_size = 13
+num_classes = 5
+num_anchors = 5
+
+epochs = 1000
+batch_size = 8
+
+img_dir = './custom_dataset/images/*'
+ann_dir = './custom_dataset/annotations/*'
+
+
+# load data
+dataset = YOLODataset(img_dir, ann_dir)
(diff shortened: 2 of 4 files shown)


DonghakPark (approved):
LGTM!!
But There was doxygen fail, please check

baek2sm:
> LGTM!! But There was doxygen fail, please check

Thanks, I'll update annotation of these codes. :)

Applications/YOLO/PyTorch/main.py
```diff
@@ -0,0 +1,131 @@
...
+        (11.2364, 10.0071)]
+    )
+
+    outsize = (13, 13)
+    width, height = outsize
+    
+    # restore cell pos to x, y    
+    for w in range(width):
+        for h in range(height):
+            bbox_pred[0, height*h + w, :, 0] += w
```
lhs8928, line 101: I guess this code will only process first batch isn't it?

baek2sm: You are right. thanks! I'll fix it.

Applications/YOLO/PyTorch/yolo_loss.py
```diff
@@ -0,0 +1,187 @@
...
+            (11.2364, 10.0071)]
+        )
+                
+        self.mse = nn.MSELoss()
+        self.bbox_loss, self.iou_loss, self.cls_loss = None, None, None       
+        
+    def forward(self, bbox_pred, iou_pred, prob_pred, bbox_gt, cls_gt):
+        """
+        param bbox_pred: shape(batch_size, cell_h x cell_w, num_anchors, 4)
+        param iou_pred: shape(batch_size, 1)
```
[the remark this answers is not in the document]
baek2sm: It should be (batch_size, cell_h x cell_w, num_anchors, 1). thanks.

Applications/YOLO/PyTorch/yolo_loss.py
```diff
@@ -0,0 +1,187 @@
...
+        param prob_pred: shape(batch_size, cell_h x cell_w, num_anchors, num_classes)
+        param bbox_gt: shape(batch_size, num_bbox, 4), data range(0~1)
+        param cls_gt: shape(batch_size, 1)
```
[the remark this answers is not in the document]
baek2sm: That's right. thanks!

lhs8928 (approved):
LGTM

Applications/YOLO/PyTorch/dataset.py
```diff
@@ -0,0 +1,49 @@
...
+        self.input_images = np.array(self.input_images)
+        self.input_images = torch.FloatTensor(self.input_images).permute((0, 3, 1, 2))
+        
+    def __len__(self):
+        return self.length
+    
+    def __getitem__(self, idx):
+        return self.input_images[idx], self.bbox_gt[idx], self.cls_gt[idx]
+    
+def collate_db(batch):
```
DonghakPark, line 44: For CI PASS in  ```Applications/YOLO/PyTorch/dataset.py``` 
you need function doxygen tag 
For example)
```.py
"""
@param batch batch size
@return items items
"""
```

Applications/YOLO/PyTorch/yolo_loss.py
```diff
@@ -0,0 +1,205 @@
...
+        _cls_mask = torch.zeros((hw, num_anchors, 1))
+                        
+        # find best anchors
+        _bbox_gt_wh = _bbox_gt.clone()[:, 2:]        
+        best_anchors = find_best_ratio(self.anchors, _bbox_gt_wh)
+        
+        # normalize x, y pos based on cell coornindates
+        cell_w = float(self.img_shape[0]) / self.outsize[0] 
+        cell_h = float(self.img_shape[1]) / self.outsize[1]
+        cx = _bbox_gt[:, 0] * self.img_shape[0] / cell_w
```
lhs8928, line 180: How about this?
`cx = _bbox_gt[:, 0] * self.outsize[0]`
I think cell_w, cell_h does not needed

baek2sm: > How about this? `cx = _bbox_gt[:, 0] * self.outsize[0]` I think cell_w, cell_h does not needed

That's a good idea. thanks!

Applications/YOLO/PyTorch/dataset.py
```diff
@@ -0,0 +1,66 @@
+##
```
jijoongmoon, line 1: please add SPDX-License-Identifier: Apache-2.0 
as in https://github.com/nnstreamer/nntrainer/blob/6bf3610820817d298185d9dce8eda9d84a3a71ac/Applications/MNIST/PyTorch/main.py#L2

Applications/YOLO/PyTorch/main.py
```diff
@@ -0,0 +1,128 @@
+##
```
jijoongmoon, line 1: same here

Applications/YOLO/PyTorch/main.py
```diff
@@ -0,0 +1,128 @@
...
+# load data
+dataset = YOLODataset(img_dir, ann_dir)
+loader = DataLoader(dataset, batch_size=batch_size, collate_fn=collate_db, shuffle=True, drop_last=True)
+
+
+# set model, loss and optimizer
+model = YoloV2_light(num_classes=5)
+criterion = YoloV2_LOSS(num_classes=5)
+optimizer = optim.Adam(model.parameters(), lr=1e-3)
+scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=10, eta_min=0)
```
jijoongmoon, line 43: If we need to use this lr_scheduler, then we need to implement it into NNTrainer

Applications/YOLO/PyTorch/yolo_loss.py
```diff
@@ -0,0 +1,203 @@
+##
```
jijoongmoon, line 1: same here.

Applications/YOLO/PyTorch/yolo_loss.py
```diff
@@ -0,0 +1,203 @@
...
+    b1 = np.divide(anchors[:, 0], anchors[:, 1])
+    b2 = np.divide(bbox[:, 0], bbox[:, 1])
+    similarities = np.abs(b1.reshape(-1, 1) - b2)
+    best_match = np.argmin(similarities, axis=0)
+    return best_match
+
+##
+# @brief loss class for yolo
+class YoloV2_LOSS(nn.Module):
+    """Yolo v2 loss"""
```
jijoongmoon, line 63: when you implement this loss into NNTrainer, plz consider where these computations should be done among calDerivatives, calGradient.

jijoongmoon (approved):
LGTM! Great Work!

Applications/YOLO/PyTorch/yolo_loss.py
baek2sm: thank you, I will consider your comment and work on it.

Applications/YOLO/PyTorch/dataset.py
baek2sm: I added this line. thanks!

Applications/YOLO/PyTorch/main.py
baek2sm: actually, It's an optional element. so I think it's okay to add it later. thanks!

Applications/YOLO/PyTorch/main.py
baek2sm: thank you!

Applications/YOLO/PyTorch/yolo_loss.py
baek2sm: thank you

PR #2139, opened 2023-03-13
[Windows] Add Visual Studio Solution and Project

SeoHyungjun:
Added Solution and Project to create nntrainer.dll in windows.

Added a script that automatically receives the necessary external Library from nnstreamer-windows-resource and changes the path.

Reflecting the latest commit can cause additional modifications, so I created pr in the test environment.

### DIFF
--- a/nntrainer/compiler/remap_realizer.h
+++ b/nntrainer/compiler/remap_realizer.h
@@ -15,6 +15,9 @@
 #include <functional>
 #include <memory>
 #include <vector>
+#ifdef _WIN32
+#include <string>
+#endif
 
 #include <realizer.h>
 
--- a/nntrainer/dataset/databuffer.cpp
+++ b/nntrainer/dataset/databuffer.cpp
@@ -39,6 +39,9 @@
 #include <stdlib.h>
 #include <thread>
 #include <util_func.h>
+#ifdef _WIN32
+#include <numeric>
+#endif
 
 namespace nntrainer {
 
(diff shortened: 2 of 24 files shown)


nntrainer/compiler/remap_realizer.h
```diff
@@ -15,6 +15,9 @@
 #include <functional>
 #include <memory>
 #include <vector>
+#ifdef _WIN32
+#include <string>
```
jihochu, line 19: string is common in stl. how about remove ifdef ?

nntrainer/dataset/databuffer.cpp
```diff
@@ -39,6 +39,9 @@
 #include <stdlib.h>
 #include <thread>
 #include <util_func.h>
+#ifdef _WIN32
+#include <numeric>
```
jihochu, line 43: ditto

nntrainer/layers/common_properties.h
```diff
@@ -22,6 +22,10 @@
 #include <tensor.h>
 #include <tensor_wrap_specs.h>
 
+#ifdef _WIN32
+#define uint unsigned int
```
jihochu, line 26: I cannot find `uint` used.

SeoHyungjun: At line 496 uint is used.

nntrainer/compiler/remap_realizer.h
SeoHyungjun: Removed ifdef.

PR #2141, opened 2023-03-14
[tensor] split tensor by given sizes

lhs8928:
- Until now tensor was splited evenly by given size, but from now on split operation can split the tensor by given sizes


DonghakPark (approved):
LGTM!

nntrainer/tensor/tensor.cpp
```diff
@@ -745,32 +745,59 @@ Tensor Tensor::getSharedDataTensor(const TensorDim dim_, unsigned int offset,
   return ret;
 }
 
-std::vector<Tensor> Tensor::split(unsigned num_size, int axis) {
+std::vector<Tensor> Tensor::split(std::vector<unsigned int> sizes, int axis) {
```
jijoongmoon, line 748: how about using offset? and make an additional function that uses this?
``` c++
std::vector<Tensor> Tensor::split(std::vector<size_t> offset, int axis){...};

// This is to split tensor evenly.
std::vector<Tensor> Tensor::split(unsigned int num_size, int axix){
    std::vector< size_t> offsets.emplace_back(num_size) ;
   offsets[0] = 0;
   size_t size = dim.getTensorDim(axis) /num_size;
   for(unsigned int j = 1; j<num_size ;j++)
        offsets[j] = offsets[i] + size;
   
   return split( offset,  axis); 
}


```

jijoongmoon, line 748: and also, it is better to use ```size_t``` rather than ```unsigned int```

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -3393,7 +3393,7 @@ TEST(nntrainer_Tensor, split_01_p) {
                              110, 111, 112, 113, 114, 115, 116, 117, 118, 119};
       answer.emplace_back(ml::train::TensorDim{1, 2, 4, 5}, answer_data);
     }
-    EXPECT_EQ(t.split(3, 0), answer);
+    EXPECT_EQ(t.split({3}, 0), answer);
```
jijoongmoon, line 3396: you don't need to change these unit test once you provide ```td::vector<Tensor> Tensor::split(unsigned int num_size, int axis)```

nntrainer/tensor/tensor.cpp
lhs8928: Added additional function as you recommend.
Split can be done by either way of using offset, sizes. (numpy use offset, tensorflow, pytorch use size).
If there is specific reason to using offset please let me know. I will change that way.

lhs8928: Done

test/unittest/unittest_nntrainer_tensor.cpp
lhs8928: Reverted

nntrainer/tensor/tensor.cpp
jijoongmoon: If other frameworks use the size, then we could go with it then.

baek2sm (approved):
LGTM

PR #2147, opened 2023-03-17
[Application] yolo example moving weight from torch to nntrainer

baek2sm:
Add yolo example moving weight from pytorch to nntrainer.


baek2sm:
I made a mistake while rebase

PR #2149, opened 2023-03-20
[ Application ] Add android resnet application example @open sesame 03/20 13:48

jijoongmoon:
This PR includes the android resnet application example with cifar100 dataset. The dataset is not included becuase of the size of dataset, but user can download and set properly into asset direcotry in application.

This applicaiton demonstrate how nntrainer is used in android application through JNI interface and train, test, stop during processing, and inferencing.

**Self evaluation:**


DonghakPark (approved):
LGTM!

baek2sm (approved):
great work.. LGTM

PR #2150, opened 2023-03-20
Hotfix/tizen7.0 ahub

lhs8928:
Commit 1:     replace strerror with strerror_r
    
     - Replace strerror with strerror_r to make thread safe.
    
Commit 2:     [ahub] fix ahub issue
    
     - cast unsigned value to signed value and check its value it overrides sign-bit.
    

jijoongmoon (approved):
LGTM

PR #2153, opened 2023-03-22
[Application] dataloader for yolo @open sesame 04/07 20:08

baek2sm:
Add detection dataloader for yolo example.

- Set target directory, then this dataloader loads dataset
from "images" and "annotations" folders.
- Currently, It only supports "bmp" image format.
- For variable label data, it adds zero padding to label.

### DIFF
--- a/Applications/YOLO/jni/det_dataloader.cpp
+++ b/Applications/YOLO/jni/det_dataloader.cpp
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2023 Seungbaek Hong <sb92.hong@samsung.com>
+ *
+ * @file   det_dataloader.h
+ * @date   22 March 2023
+ * @brief  dataloader for object detection dataset
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Seungbaek Hong <sb92.hong@samsung.com>
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#include "det_dataloader.h"
+
+#include <cstring>
+#include <filesystem>
+#include <fstream>
+#include <nntrainer_error.h>
+#include <random>
+
+namespace nntrainer::util {
+
+// It supports bmp image file only now.
+DirDataLoader::DirDataLoader(const char *directory_, unsigned int max_num_label,
+                             unsigned int c, unsigned int w, unsigned int h,
+                             bool is_train_) :
+  max_num_label(max_num_label),
+  channel(c),
+  height(h),
+  width(w),
+  is_train(is_train_) {
+  dir_path.assign(directory_);
+
+  // set data list
+  std::filesystem::directory_iterator itr(dir_path + "images");
+  while (itr != std::filesystem::end(itr)) {
+    // get image file name
+    std::string img_file = itr->path().string();
+
+    // check if it is bmp image file
+    if (img_file.find(".bmp") == std::string::npos) {
+      itr++;
+      continue;
+    }
+
+    // set label file name
+    std::string label_file = img_file;
+    label_file.replace(label_file.find(".bmp"), 4, ".txt");
+    label_file.replace(label_file.find("/images"), 7, "/annotations");
+
+    // check if there is paired label file
+    if (!std::filesystem::exists(label_file)) {
+      itr++;
+      continue;
+    }
+
+    // set data list
+    data_list.push_back(make_pair(img_file, label_file));
+    itr++;
+  }
+
+  // set index and shuffle data
+  idxes = std::vector<unsigned int>(data_list.size());
+  std::iota(idxes.begin(), idxes.end(), 0);
+  if (is_train)
+    std::shuffle(idxes.begin(), idxes.end(), rng);
+
+  data_size = data_list.size();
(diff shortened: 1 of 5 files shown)


DonghakPark (approved):
LGTM!

Applications/YOLO/jni/det_dataloader.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include <fstream>
+#include <iostream>
+#include <nntrainer_error.h>
+#include <random>
+#include <sstream>
+
+namespace nntrainer::util {
+
+DirDataLoader::DirDataLoader(const char *directory_, unsigned int max_num_label,
+                             unsigned int c, unsigned int w, unsigned int h,
```
jijoongmoon, line 26: It might be better to describe saying it supports bmp only for now.

jijoongmoon (approved):
LGTM. You need to rebase the main to resolve the CI issue for android.

Applications/YOLO/jni/det_dataloader.cpp
baek2sm: I added it, thanks.

baek2sm:
> LGTM. You need to rebase the main to resolve the CI issue for android.

Thanks, I rebased it, but there is still a problem.
There seems to be a problem with my code. I'll try to find the cause.

Applications/YOLO/jni/main.cpp
```diff
@@ -254,6 +253,17 @@ int main(int argc, char *argv[]) {
     model->summarize(std::cout, ML_TRAIN_SUMMARY_MODEL);
 
     model->train();
+
+    // // simple inference example
+    // model->load("./best_model.bin");
```
lhs8928, line 258: I think we need to add load function to load weight bin file before train as well

Applications/YOLO/jni/main.cpp
```diff
@@ -254,6 +253,17 @@ int main(int argc, char *argv[]) {
     model->summarize(std::cout, ML_TRAIN_SUMMARY_MODEL);
 
     model->train();
+
+    // // simple inference example
```
lhs8928, line 257: Is there specific reason to leave this code as comment?

Applications/YOLO/jni/main.cpp
```diff
@@ -217,7 +215,8 @@ int main(int argc, char *argv[]) {
   // create train and validation data
   std::array<UserDataType, 2> user_datas;
   try {
-    user_datas = createFakeDataGenerator(batch_size, data_size, data_split);
+    user_datas =
+      createDetDataGenerator("./train_dir/", "./valid_dir/", 5, 3, 416, 416);
```
lhs8928, line 219: Please make this values like image height/width size as constant variable

DonghakPark (commented):
I think you should make ```Andrioid.mk``` file for android file or make if statement in meson.build file , for avoid android check

if you want to make ```Android.mk``` file in you can just reference ```/Application/AlexNet/jni/Android.mk``` and Application.mk file

Applications/YOLO/jni/main.cpp
baek2sm: It was simply a reference code for inference using dummy data. I removed these codes. thanks!

Applications/YOLO/jni/main.cpp
baek2sm: I made it using constant variables. thanks!

Applications/YOLO/jni/main.cpp
baek2sm: I think there should be a pre-trained weights bin file to load before train.

baek2sm:
> I think you should make `Andrioid.mk` file for android file or make if statement in meson.build file , for avoid android check
> 
> if you want to make `Android.mk` file in you can just reference `/Application/AlexNet/jni/Android.mk` and Application.mk file

Thank you so much, I will try to update it!!

Applications/YOLO/jni/main.cpp
```diff
@@ -217,7 +215,14 @@ int main(int argc, char *argv[]) {
   // create train and validation data
   std::array<UserDataType, 2> user_datas;
   try {
-    user_datas = createFakeDataGenerator(batch_size, data_size, data_split);
+    const char *train_dir = "./train_dir/";
+    const char *valid_dir = "./valid_dir/";
+    const int max_num_label = 5;
+    const int channel = 3;
+    const int width = 416;
```
lhs8928, line 222: Please declare this variables at the beginning of this code.
width, height values are also needed in input layer.
AFAIK max_num_label is 4 isn't is?

jijoongmoon:
you need to resolve the ci android issue : http://ci.nnstreamer.ai/nntrainer/ci/repo-workers/pr-checker/2153-202303301343320.37945890426636-78839f1307ed3e1ccfa85eae9e060523d011550e/report/build_log_2153_android_error.txt

SeoHyungjun (approved):
LGTM!

Can't use extensions other than .bmp when load data?
For example, can insert a file extension into a function's factor

baek2sm:
> LGTM!
> 
> Can't use extensions other than .bmp when load data? For example, can insert a file extension into a function's factor

now it only supports bmp image files. we have to support other formats later.

lhs8928 (approved):
LGTM

PR #2159, opened 2023-03-29
[Application] Update yolo example of torch for tracking gradients

baek2sm:
For tracking the gradients of Loss class,
I removed in-place operation in the Loss class.

And, I added hook_variable function for tracking specific tensors.

We can register specific variable with name using hook_variable
function, then we can check the gradient values using
print_hook_variable function after backwarding.


DonghakPark (approved):
LGTM!

jihochu (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2163, opened 2023-03-31
[Application] Update yolo v2 model similar to original paper @open sesame 04/20 15:09

baek2sm:
Yolo v2 model was updated similar to original yolo v2 model.

This model was intended to be implemented in accordance with
the original paper of Yolo v2 as much as possible,
but now average pooling is temporarily used instead of the
re-organization module.

If only the average pooling is replaced with the re-organization
module in the future, the rest is the same as the original paper
in Yolo v2.

Both the PyTorch version and the NNTrainer version updated the model
structure and verified that the same results could be obtained
by loading trained weights from PyTorch.

**Self evaluation:**

### DIFF
--- a/Applications/YOLO/jni/main.cpp
+++ b/Applications/YOLO/jni/main.cpp
@@ -29,6 +29,8 @@ using LayerHandle = std::shared_ptr<ml::train::Layer>;
 using ModelHandle = std::unique_ptr<ml::train::Model>;
 using UserDataType = std::unique_ptr<nntrainer::util::DirDataLoader>;
 
+const int num_classes = 4;
+
 int trainData_cb(float **input, float **label, bool *last, void *user_data) {
   auto data = reinterpret_cast<nntrainer::util::DirDataLoader *>(user_data);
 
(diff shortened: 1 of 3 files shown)


Applications/YOLO/jni/main.cpp
```diff
@@ -177,21 +179,50 @@ ModelHandle YOLO() {
   blocks.push_back(yoloBlock("conv7", "conv6", 128, 1, false));
   blocks.push_back(yoloBlock("conv8", "conv7", 256, 3, true));
   blocks.push_back(yoloBlock("conv9", "conv8", 512, 3, false));
-  blocks.push_back(yoloBlock("conv10", "conv9", 256, 1, true));
+  blocks.push_back(yoloBlock("conv10", "conv9", 256, 1, false));
+  blocks.push_back(yoloBlock("conv11", "conv10", 512, 3, false));
+  blocks.push_back(yoloBlock("conv12", "conv11", 256, 1, false));
+  blocks.push_back(yoloBlock("conv13", "conv12", 512, 3, false));
+
+  blocks.push_back({createLayer(
```
jijoongmoon, line 187: plz, leave a comment this will be replaced with residual in a later pr.

jijoongmoon (approved):
LGTM

Applications/YOLO/jni/main.cpp
baek2sm: thanks, I added a comment at "conv_b_pool(average pooling)", line 200.

DonghakPark (approved):
LGTM!

SeoHyungjun (approved):
LGTM!

PR #2165, opened 2023-03-31
[ TensorDim ] Add tensor format in TensorDim

jijoongmoon:
In order to support Channel Last & Channel First at the same time, we need to define the Tensor Format in TensorDim Class.

According to the format of tensorDim, it return proper value for the APIs such as bacch(), channel(), height() and width().

The default format is Channel First as it is now.

If nothing is provided when a Tensor is constructured, then it is set NCHW.

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/tensor_dim.h
+++ b/api/ccapi/include/tensor_dim.h
@@ -34,6 +34,11 @@ public:
   static constexpr const size_t MAXDIM = 4;
 
   /**
+   * @brief     Tensor Formant : Default is NCHW
+   */
+  enum class Format { NCHW, NHWC };
+
+  /**
    * @brief Get the Num Dim object
    *
    * @return unsigned int fixed value of MAXDIM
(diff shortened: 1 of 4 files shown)


nntrainer/utils/node_exporter.cpp
```diff
@@ -201,7 +202,10 @@ void Exporter::saveTflResult(
     new_inputs.reserve(inputs.size() + 1 /** perm **/);
     new_inputs.push_back(*inputs[0]);
     // create "perm" tensor for Transpose operator
-    TensorDim perm_dim{std::bitset<4>(0b0001)};
+    // @todo : This NCHW formant set is just temporal, it needs to be set by
```
baek2sm, line 205: typo error in format

baek2sm (approved):
LGTM

nntrainer/utils/node_exporter.cpp
jijoongmoon: fixed. thanks.

SeoHyungjun (approved):
LGTM!

DonghakPark (approved):
LGTM!

nntrainer/tensor/tensor_dim.cpp
```diff
@@ -36,7 +38,8 @@ TensorDim::TensorDim(const std::bitset<MAXDIM> &eff_dim_flag_,
   feature_len = 0;
 }
 
-TensorDim::TensorDim(std::initializer_list<size_t> dims) : TensorDim() {
+TensorDim::TensorDim(std::initializer_list<size_t> dims, Format fm) :
+  TensorDim() {
```
jihochu, line 42:
```suggestion
  TensorDim(), format(fm) {
```
it would be better.

api/ccapi/include/tensor_dim.h
```diff
@@ -43,27 +48,32 @@ class TensorDim {
   /**
    * @brief Construct a new Tensor Dim object
    *
+   * @param fm format NCHW | HNWC
    * @param eff_dim_flag_ effective dimension flag (1 means it's effective)
    * @param dyn_dim_flag_ dynamic dimension flag (1 means it's unspecified)
    */
-  explicit TensorDim(const std::bitset<MAXDIM> &eff_dim_flag_ = 0b1111,
+  explicit TensorDim(Format fm = Format::NCHW,
```
jihochu, line 55: If we move the `Format` to last order, we can assume it's NCHW when it's not provided as other APIs.

jijoongmoon: yes. but there is also a need for ```eff_dim_flag_``` and ```dyn_dim_flag_```. so.. if we set the TensorDim with Format.. we have to define these flags as well. I think it is quite bothering.

nntrainer/tensor/tensor_dim.cpp
jijoongmoon: I will fix it in later PR as ```TensorDim(fm)```. Thanks.

api/ccapi/include/tensor_dim.h
jihochu: OK. `eff_dim_flag_` and `dyn_dim_flag_` are not usually changed, original one would be better.

jihochu (approved):
LGTM 👍

PR #2175, opened 2023-04-06
[AHub] Fix AHub Defect @open sesame 04/10 09:22

DonghakPark:
## In this PR

Fix AHub Defect
- make some exception statement
- change auto element -> auto &element

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -335,7 +335,8 @@ void NeuralNetwork::backwarding(int iteration,
         node->calcGradient();
 
       /**
-       * If optimization off, or gradient must be applied, then this will be true
+       * If optimization off, or gradient must be applied, then this will be
+       * true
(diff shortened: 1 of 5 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -838,7 +839,12 @@ int NeuralNetwork::train_run(std::function<bool(void *userdata)> stop_cb,
 
   auto train_epoch_end = [this, stop_cb, user_data](RunStats &stat,
                                                     DataBuffer &buffer) {
-    stat.loss /= static_cast<float>(stat.num_iterations);
+    try {
+      stat.loss /= static_cast<float>(stat.num_iterations);
```
jijoongmoon, line 843: There is no way to be zero but we just need check if (stat.num_iterations == 0)

DonghakPark: [FIX] Thank you for Review

jijoongmoon (approved):
LGTM

PR #2185, opened 2023-04-14
[Wait For #2184][TFLite Export] Add Realized Path for Fused Op

DonghakPark:
## In This PR

**For Fused OP Made Realized Path**

1. Check Trainable 
- check node is trainable or not for Fusing

2. Conv + ReLU Fusing
- For Trainable (Conv) with Batch Normalization it can Fused 
**- Fused weight(conv) = weight(conv) * (weight(bn) / sqrt(var(bn) + eps))**
**- Fused Bias(cov) = (bias(conv) - mean) * (weight(bn) / sqrt(var(bn) + eps)) + bias(bn)**

3. Batch Normalization Fusing
- If non-trainable with batchnorm, it will split in to MUL + ADD nodes with below 
**- y = x  * (gamma / sqrt(variance + epsilon)) + (beta - mean * gamma / sqrt(variance + epsilon) )**

PR #2184, opened 2023-04-14
[TFLite Export] Add variabel, functions TfOpNodes for Fused OP export

DonghakPark:
## In this PR

for Export Tflite format with Fused Op add some Variable and Function

1. Add getter, setter, replace to weights, inputs, outputs
- for Fused Op we need to adjust Variables after made Opnode

2. Add isToBeRemove variable
- After made Opnode, check condition and mark as to be remove

3. Add additional_props
- for BatchNormalization Fused Op we need additional props from nntrainer
- made vector<float> variable for save additional data

PR #2186, opened 2023-04-14
[Wait for #2185][TFLite Export] Update node_exporter @open sesame 05/10 11:14

DonghakPark:
## In This PR

For adjust exist node, we need some information from old node, and in Flatbuffer it is hard to bring. so made some variables for saving information

Add Epsilon Props to additional_props for fusing
 - For Fusing we need Epsilon for batch norm
Add padding, stride props to props_vector
 - For Conv Fusing we need to made new BuiltinOption and for building new BuiltinOption with FUSED activation we need padding,stride

**Related PR**
#2184 #2185 #2204


baek2sm (approved):
LGTM

jijoongmoon (commented):
LGTM

jijoongmoon (approved):
LGTM

PR #2187, opened 2023-04-17
[capi] add identity layer to capi

lhs8928:
- Added identity layer enum to capi


baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM. but we need capi unittest for this identity layer.

PR #2199, opened 2023-04-27
[Layer] Add checkpoint property

jihochu:
This patch adds CheckPoint property to layer node.
Default value is checkpointed, which can store the layer data for the
backward step. laoded and unloaded state are for unchecked nodes. The
data would be unloaded from memory, and reloaded when it's needed when
the backward step.


baek2sm (approved):
LGTM

PR #2200, opened 2023-04-27
(Wait for #2199) [TENSOR] Initial implement for temp pool

jihochu:
temp pool is for temporal tensor which is used for checkpoint mechanism.
It reduces peak memory consumption while training, it only store
checkpointed activation for the backward phase. unchecked activations
will be recalculated using checkpointed values.


baek2sm (approved):
LGTM

PR #2204, opened 2023-05-09
[Wait for #2186][TFLite Export] Add Unit Test

DonghakPark:
## In this PR

**Add Unit Test case For Test Fused Op in Tensorflow lite export** 

Test Model consist like below

```Input -> Conv2D -> Batch Norm -> ReLU -> Flatten```

- make this model with nntrainer and export to tensorflow lite
- set input {1,3,4,4} Dim -> in Tensorflow lite native make same input with 3,4,4 and transpose this input

**latency**
(nntrainer exported has transpose layer)
10000 forward tflite native (sec)              : **0.0069193975830078125**
10000 forward nntrainter exported (sec) : **0.0070416927337646484**

**Related PR**
#2184 #2185 #2186

### DIFF
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -406,7 +406,7 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
           tflite::BuiltinOperator::BuiltinOperator_MUL &&
         nodes.at(node_count + 2).get()->getOpType() ==
           tflite::BuiltinOperator::BuiltinOperator_RELU) {
-      // Fuse Conv2D + Mul + ReLU to Conv2D
+      // Fuse Conv2D + Mul(Batch Norm) + ReLU to Conv2D
 
       auto props = tf_node->getProps();
       auto tf_padding = tflite::Padding_SAME;
@@ -462,7 +462,6 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
         conv_new_weights.push_back(&conv_weight);
         conv_new_weights.push_back(&conv_bias);
         nodes.at(node_count - 1).get()->setWeights(conv_new_weights);
-
         // set mul node to be removed (mul mean batch normalization)
         n->setToBeRemoved(true);
       }
@@ -706,30 +705,28 @@ TfOpNodes buildRealizedOpNodes(TfOpNodes &nodes,
   unsigned int node_count = 0;
 
   for (auto &node : nodes) {
-
     if (set_input) { // if front node is new added node set input output
       node->setArg(0, realized_nodes.back().get());
       realized_nodes.back()->setOutputs(node->getInputs());
       set_input = false;
     }
 
-    if (node->isToBeRemoved() == true) { // Remove node
+    if (node->isToBeRemoved() == true) {
+      // Remove node, Assume that Input Node is not removed
       realized_nodes.back().get()->setOutputs(
-        nodes.at(node_count + 1)->getInputs());
+        nodes.at(node_count)->getOutputs());
       nodes.at(node_count + 1)->setArg(0, realized_nodes.back().get());
-      nodes.at(node_count + 1)
-        ->setInputs(realized_nodes.back().get()->getOutputs());
+      nodes.at(node_count + 1)->setInputs(nodes.at(node_count)->getInputs());
     } else {
       realized_nodes.push_back(std::move(node));
 
       if (realized_nodes.back().get()->getOpType() ==
           tflite::BuiltinOperator_MUL) { // Fused MUL ADD (Non Trainable)
-
-        // remove weights (In .tflite this mean INPUTS)
+        /**
+          y = x * (gamma / sqrt(variance + epsilon)) +
+          (beta - mean * gamma / sqrt(variance + epsilon))
+        */
         auto removed_weights = realized_nodes.back().get()->getWeights();
-        // y = x
-        // * (gamma / sqrt(variance + epsilon))
-        // + (beta - mean * gamma / sqrt(variance + epsilon) )
         auto mul_mean = removed_weights.at(0)->clone();
         auto mul_variance = removed_weights.at(1)->clone();
         auto mul_gamma = removed_weights.at(2)->clone();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: nntrainer/compiler/tflite_interpreter.cpp [hunk 1]
```diff
@@ -323,8 +323,13 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
...
     Exporter e(&fbb);
     ln->exportTo(e, ml::train::ExportMethods::METHOD_TFLITE);
+    auto export_output = e.getResult<ml::train::ExportMethods::METHOD_TFLITE>();
+
+    if (export_output.get()->getWeights().size() == 0) {
+      export_output.get()->setTrainable(false);
+    }
 
-    nodes.emplace_back(e.getResult<ml::train::ExportMethods::METHOD_TFLITE>());
+    nodes.emplace_back(move(export_output));
```
jijoongmoon, line 332: I wonder do we need this move here? AFAIK, getResult return moved object already.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -391,6 +397,73 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
       tf_node->weightReorder(node_count);
     }
 
+    if (tf_node->getOpType() ==
```
jijoongmoon, line 400: Not this pr, but later, it might be better to separate these graph traversals as compiler optimizers.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -391,6 +397,73 @@ TfOpNodes buildOpNodes(const GraphRepresentation &representation,
...
+          tflite::BuiltinOperator::BuiltinOperator_CONV_2D &&
+        nodes.at(node_count + 1).get()->getOpType() ==
+          tflite::BuiltinOperator::BuiltinOperator_MUL &&
+        nodes.at(node_count + 2).get()->getOpType() ==
+          tflite::BuiltinOperator::BuiltinOperator_RELU) {
```
jijoongmoon, line 405: Is it possible to apply this fused op for sigmoid? If it is, it might be better to refactor the codes for more general cases.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -622,29 +695,127 @@ buildSubGraphs(const TfOpNodes &nodes, const TfOpIdxMap &map,
 
 } // namespace
 
+TfOpNodes buildRealizedOpNodes(TfOpNodes &nodes,
```
jijoongmoon, line 698: Is it only for the bn fusing? if it is, it is better to change the function name.

nntrainer/compiler/tflite_opnode.cpp
```diff
@@ -111,12 +112,33 @@ void TfOpNode::setLayerNode(const LayerNode &layer) {
...
 }
 
 void TfOpNode::setWeightTransformFn(TransformFn fn) { weight_transform = fn; }
 
 void TfOpNode::setInputTransformFn(TransformFn fn) { input_transform = fn; }
 
+void TfOpNode::setWeights(Variables weights_) {
+  unsigned int cnt = 0;
+  for (auto &w : weights_) {
+    const unsigned int UNIT = w->batch();
```
jijoongmoon, line 128: let's not use capital letters for common variables.

nntrainer/compiler/tflite_opnode.cpp
```diff
@@ -111,12 +112,33 @@ void TfOpNode::setLayerNode(const LayerNode &layer) {
...
+    const unsigned int CHANNEL = w->channel();
+    const unsigned int HEIGHT = w->height();
+    const unsigned int WIDTH = w->width();
+
+    auto weight_data = weights.at(cnt)->getData();
+    auto *ptr = const_cast<float *>(weight_data);
+    memcpy(&ptr[0], &w->getData()[0],
+           sizeof(float) * (UNIT * CHANNEL * HEIGHT * WIDTH));
+    auto weight_data2 = weights.at(cnt)->getData();
```
jijoongmoon, line 137: remove weight_data2?

nntrainer/compiler/tflite_opnode.cpp
```diff
@@ -111,12 +112,33 @@ void TfOpNode::setLayerNode(const LayerNode &layer) {
...
+  unsigned int cnt = 0;
+  for (auto &w : weights_) {
+    const unsigned int UNIT = w->batch();
+    const unsigned int CHANNEL = w->channel();
+    const unsigned int HEIGHT = w->height();
+    const unsigned int WIDTH = w->width();
+
+    auto weight_data = weights.at(cnt)->getData();
+    auto *ptr = const_cast<float *>(weight_data);
+    memcpy(&ptr[0], &w->getData()[0],
```
jijoongmoon, line 135: how about using tensor op rather than using memcpy?

nntrainer/compiler/tflite_interpreter.cpp
DonghakPark: I think yes.
getResult move object! but for some check i already move object to export_output, so we need to move again to nodes

nntrainer/compiler/tflite_interpreter.cpp
DonghakPark: I agree with you, Thank you for Review

nntrainer/compiler/tflite_interpreter.cpp
DonghakPark: Yes. it possible and Conv + MUL + Activation is method that tflite fusing

nntrainer/compiler/tflite_interpreter.cpp
DonghakPark: Later. i will add more fusing path in this function.
it is better to change function name for now ?

nntrainer/compiler/tflite_opnode.cpp
DonghakPark: FIX

nntrainer/compiler/tflite_opnode.cpp
DonghakPark: FIX

nntrainer/compiler/tflite_opnode.cpp
DonghakPark: can i use ```weights.at(cnt)->putData(weights_)```?

jijoongmoon: you can use https://github.com/nnstreamer/nntrainer/blob/65d5f04ab8cf666d4ec784d663a949103035d59b/nntrainer/tensor/tensor.h#L1055 
or 
https://github.com/nnstreamer/nntrainer/blob/65d5f04ab8cf666d4ec784d663a949103035d59b/nntrainer/tensor/tensor.h#L1061

baek2sm (approved):
LGTM

Hyoungsukim (approved):
lgtm

djeong20 (approved):
LGTM

SeoHyungjun (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2205, opened 2023-05-15
[model] Add epoch complete callback

songgot:
- Called the end of an epoch
- Users can do what they need at the end of each epoch. e.g. get RunStats.


jijoongmoon (approved):
LGTM

DonghakPark (approved):
LGTM!

PR #2212, opened 2023-05-26
[Wait for #2209][WIP][POC] Implement picoGPT/GPT2 Encoder @open sesame 10/10 18:42

DonghakPark:
**Implementing picoGPT/GPT2's Encoder in CPP**
- Using nlohman/json.hpp file so we need to add or make some path to compile json parser
- Need to check utf-8 encoding result


myungjoo:
It appears that you are direct importing files from other repositories, right?

DonghakPark:
> It appears that you are direct importing files from other repositories, right?

Yes. is it better to make some script files to prepare this kinds of files?

myungjoo:
Great! We are going to personalize GPT on devices!

jijoongmoon (changes requested):
Instead of directly include the external files like json.hpp or ctr-unicode.hpp, we could place them in gihtub.com/nnstreamer/nnstreamer-android-resource/raw/master/external/ repository and download them using shell script like jni/prepare_openblas.sh.

meson.build
```diff
@@ -172,6 +172,13 @@ if get_option('enable-cublas')
   extra_defines += '-DUSE_CUBLAS=1'
 endif
 
+if get_option('platform') != 'tizen'
```
jijoongmoon, line 175: I don't think we always use encoder. It only depends on the application Pico. So please remove the running shell script here and add it to the applications.

meson_options.txt
```diff
@@ -33,6 +33,7 @@ option('reduce-tolerance', type: 'boolean', value: true)
 option('enable-long-test', type: 'boolean', value: false)
 
 # backend options
+option('enable-encoder', type: 'boolean', value: true)
```
jijoongmoon, line 36: We don't need to add enable-encoder here.

packaging/nntrainer.spec
```diff
@@ -12,7 +12,7 @@
 %define         gen_input $(pwd)/test/input_gen/genInput.py
 %define         support_data_augmentation_opencv 1
 %define         configure_subplugin_install_path -Dnnstreamer-subplugin-install-path=%{nnstreamer_subplugin_path}
-
+%define         use_encoder 1
```
jijoongmoon, line 15: We don't need to define use_encoder here.

packaging/nntrainer.spec
```diff
@@ -347,6 +347,11 @@ NNSteamer tensor trainer static package for nntrainer to support inference.
 %define enable_cblas -Denable-blas=true
 %endif
 
+# enable encoder
+%if 0%{?use_encoder}
```
jijoongmoon, line 351: This encoder depends on Applicaiton PicoGPT.. so.. please run the script when we build Applicatin PicoGPT

meson.build
DonghakPark: Thank you for Review - Fix

meson_options.txt
DonghakPark: Thank you for Review - Fix

packaging/nntrainer.spec
DonghakPark: Thank you for Review - Fix

packaging/nntrainer.spec
DonghakPark: Thank you for Review - Fix

jijoongmoon (approved):
LGTM!

baek2sm (approved):
LGTM

Hyoungsukim (approved):
lgtm

djeong20 (approved):
LGTM

PR #2216, opened 2023-06-01
[LOG] print output dim instead of input dim in model summary  @open sesame 06/13 14:48

baek2sm:
When we print the model architecture using summarize method, nntrainer prints input dimension of each layer.

But, tensorflow and pytorch are printing output dimmension of each layer in the summary, thus it is inconvenient to compare each layer with tf and torch models.

Thus, I suggest to print output dimension of each layer instead of input dimension in the model summary.

**Self evaluation:**


DonghakPark (approved):
LGTM

myungjoo (approved):

songgot (approved):
LGTM

PR #2219, opened 2023-06-02
[Application] Fix Resnet18 @open sesame 06/13 10:05

SeoHyungjun:
Fix it because the results of the computation of pytorch and nntrainer are different.

padding was written as same when stride was 2 in NNtrainer Resnet code. In pytorch, this is set to error. 
In addition, padding was not applied normally in nntrainer, so the results were different. 

To solve this problem, the parameters of a1 layer (conv layer) have been modified.

Additionally, momentum and epsilon were added to batch_norm layer.

### DIFF
--- a/Applications/Resnet/jni/main.cpp
+++ b/Applications/Resnet/jni/main.cpp
@@ -113,15 +113,18 @@ std::vector<LayerHandle> resnetBlock(const std::string &block_name,
   };
 
   /** residual path */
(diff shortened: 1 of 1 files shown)


myungjoo (approved):

jijoongmoon (approved):
LGTM

baek2sm (approved):
LGTM

DonghakPark (approved):
LGTM!

DonghakPark (commented):
Please Check CI in Tizen&Ubuntu Build

PR #2232, opened 2023-06-19
[Application] Transfer learning example on Resnet-18

baek2sm:
I added transfer learning option to resnet-18 example.

If this option is enabled, then load pre-trained weights and freeze the weights of backbone(feature extractor). (It just a simple transfer learning, and default option is "false").

You can make pre-trained weights using save_bin function from our pytorch resnet-18 example.

**Self evaluation:**


myungjoo (approved):

jijoongmoon (approved):
LGTM.

PR #2234, opened 2023-06-20
[DataLoader] Add exception handling condition

baek2sm:
In the "isLabelAvailable" function, exception processing is required if the index is larger than the size of the array(outputs).

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/layer_context.cpp
+++ b/nntrainer/layers/layer_context.cpp
@@ -445,6 +445,9 @@ Weight &RunLayerContext::getWeightObject(unsigned int idx) {
  * @return true if label is available else false
  */
 bool RunLayerContext::isLabelAvailable(unsigned int idx) const {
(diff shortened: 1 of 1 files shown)


songgot (approved):
LGTM

myungjoo (approved):

nntrainer/layers/layer_context.cpp
```diff
@@ -445,6 +445,9 @@ Weight &RunLayerContext::getWeightObject(unsigned int idx) {
  * @return true if label is available else false
  */
 bool RunLayerContext::isLabelAvailable(unsigned int idx) const {
+  if (idx >= outputs.size())
```
jijoongmoon, line 448: How about adding a warning message here?

SeoHyungjun (approved):
LGTM

PR #2248, opened 2023-07-17
[graph_node] handle deprecated stl iterator

lhs8928:
- Explicitly provide the parameter as default parameter for stl iterator is deprecated.


SeoHyungjun (approved):
LGTM

PR #2251, opened 2023-07-20
[graph_node] remove deprecated stl iterator

lhs8928:
- Remove inheriting the stl iterator as this is deprecated in c++17
 - Added no-* warning flags. These flags should be removed.


jijoongmoon (approved):
LGTM, but we need check whether these options are really needed.

baek2sm (approved):
LGTM

PR #2264, opened 2023-08-01
[FIX] modified for checking weight grad

jihochu:
This path checks requested memory is weight gradient which information will be used for planning.


DonghakPark (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2265, opened 2023-08-01
[Wait for #2212][PoC] incremental inference

lhs8928:
Commit 1:     [PoC] incremental inference
    
     - PoC of incremental inference
     - Only works if batch, channel size is 1
     - For the concat layer, inference step only works if axis dimension is width axis
    
Commit 2:     [concat] enable incremental forwarding with multi threads
    
     - Each threads will copy the data with batchwise direction
    

jijoongmoon:
it includes in #2286

PR #2273, opened 2023-08-09
[SWAP] Fix cache pool policy

jihochu:
calculating gradient and derivative for a tensor are operated
in a different layer. So, invalidation of data between each
calculation can be possible. It fixes to keep data instead of
throwing away.

### DIFF
--- a/nntrainer/tensor/cache_pool.cpp
+++ b/nntrainer/tensor/cache_pool.cpp
@@ -56,7 +56,7 @@ convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
     policy = CachePolicy::TEMPORAL;
     break;
   case TensorLifespan::CALC_GRAD_DERIV_LIFESPAN:
-    policy = CachePolicy::TEMPORAL;
+    policy = CachePolicy::ITERATION_CONSIST;
     break;
   case TensorLifespan::CALC_GRAD_DERIV_AGRAD_LIFESPAN:
     policy = CachePolicy::ITERATION_CONSIST;
@@ -65,7 +65,7 @@ convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
     policy = CachePolicy::ITERATION_CONSIST;
     break;
   case TensorLifespan::FORWARD_GRAD_AGRAD_LIFESPAN:
-    policy = CachePolicy::ITERATION_CONSIST;
+    policy = CachePolicy::ALWAYS_SYNCED;
     break;
   case TensorLifespan::FORWARD_DERIV_LIFESPAN:
     policy = CachePolicy::ALWAYS_SYNCED;


nntrainer/tensor/cache_pool.cpp
```diff
@@ -65,7 +65,7 @@ convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
     policy = CachePolicy::ITERATION_CONSIST;
     break;
   case TensorLifespan::FORWARD_GRAD_AGRAD_LIFESPAN:
-    policy = CachePolicy::ITERATION_CONSIST;
+    policy = CachePolicy::ALWAYS_SYNCED;
```
myungjoo, line 68: LGTM.

However, if this inline function is access frequently, consider refactoring it as an array:

```
static const CachePolicy policy[TensorLifespan::MAX] = {
  [UNMANAGED] = ALWAYS_SYNCED,
  ...
};
```

nntrainer/tensor/cache_pool.cpp
```diff
@@ -56,7 +56,7 @@ convertTensorLifespanToCachePolicy(const TensorLifespan lifespan) {
     policy = CachePolicy::TEMPORAL;
     break;
   case TensorLifespan::CALC_GRAD_DERIV_LIFESPAN:
-    policy = CachePolicy::TEMPORAL;
+    policy = CachePolicy::ITERATION_CONSIST;
```
jijoongmoon, line 59: calc_grad_deriv is part of the backwarding only. It seems not the case of Iteration_consist. 
Do you define iteration differently? Seems like we need more fine-grained matching though.

jihochu: `ITERATION_CONSIST` in swap cache means its value keeps its consistency during iteration from the start to the end of its life. So, in this case (CALC_GRAD_DERIV_LIFESPAN), it keeps its value from GRAD to DERIV. and it abandons its value after DERIV.
 Sorry for the little difference meaning of `ITERATION`. I'll add comments for it.

nntrainer/tensor/cache_pool.cpp
jihochu: Thanks. I'll modify it.

jihochu: I checked the array initializer you mentioned, but it does not be supported  in c++17(currently used)
```
sorry, unimplemented: non-trivial designated initializers not supported
```
The designated initializer is supported after c++20.
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0329r4.pdf

I think it's not the time to apply the new standard yet.

myungjoo (approved):

SeoHyungjun:
LGTM!

DonghakPark (approved):
LGTM!

PR #2280, opened 2023-08-22
[gtest] positional encoding, layer norm, batch norm fp16 verification

skykongkong8:
## Commits to be reviewd in this PR

<details><summary>   bugfix </summary><br />

- bugfix for invalid data generation behavior

Signed-off-by:skykongkong8<ss.kong@samsung.com>
</details>

<details><summary>   positional encoding  </summary><br />

- now positional encoding layer supports fp16 (fp16-only case)

Signed-off-by:skykongkong8<ss.kong@samsung.com>
</details>

<details><summary>   ln bn layers </summary><br />

- issue : adding cosine similarity check in fp32/fp16 revealed that there was unmatched cosine similarity Tensors in case of near-zero Tensors. Nevertheless, absolute value difference and mse pass our epsilon value. We would better to come back here for sanity check.
- I want to emphasize that this issue only occurs for near-zero (1e-8 or lower) Tensors, and layer test cases that uses such tensor data. In non-near zero cases, everything works fine so this might be a problem of floating point itself. Please leave your opinions. 

Signed-off-by:skykongkong8<ss.kong@samsung.com>
</details>

**Self evaluation:**


jijoongmoon (approved):
LGTM!

test/unittest/layers/layers_golden_tests.cpp
```diff
@@ -209,11 +224,18 @@ static void compareRunContext(RunLayerContext &rc, std::ifstream &file,
...
+            constexpr auto eps = 1e-1;
+            if (a < b)
+              std::swap(a, b);
+            return (a - b) < eps;
+          } else {
+            constexpr auto eps = 1e-2;
+            if (a < b)
+              std::swap(a, b);
+            return (a - b) < eps;
+          }
```
djeong20, line 238: minimize the code redundancy?

```suggestion
          constexpr auto eps =  (skip_cos_sim) ? 1e-1 : 1e-2;
          if (a < b)
            std::swap(a, b);
          return (a - b) < eps;
```

djeong20 (approved):
LGTM!

test/unittest/layers/layers_golden_tests.cpp
skykongkong8: I also tried it, but it throws:
```
../test/unittest/layers/layers_golden_tests.cpp:228:33: error: lambda capture of ‘skip_cos_sim’ is not a constant expression
  228 |           constexpr auto eps = (skip_cos_sim) ? 1e-1 : 1e-2;
      |                                 ^~~~~~~~~~~~
```

skykongkong8: That block is unavoidable for near-zero Tensor case... (though I don't like it either) Any better idea?

PR #2281, opened 2023-08-22
[gtest] Enable layers with fp16, verify with gtest

skykongkong8:
This PR includes:

[tensor] Fix some fp16-unfriendly functions
- copyData : add ifdef code block to support fp16
- fill : ditto
- reshape : instead of switching the entire TensorDim, only get the Tensor shape info

[data] gtest data binary files
- proper gtest data files that matches with nntrainer behavior

[layer] multi-headed attention fp16 
- activation function template used
- beware of reshape function change
- fwding & bwding but cos_sim diverge at near-zero Tensor (1e-8, which is not float16's scope)

[layer] embedding layer fp16
- verified for single & multi batch cases
- added fp32 embedding layer gtest and data  generation code (previously there was no embedding layer gtest implementation)

[layer] concat layer
- same issue encountered at Tensor::reshape. Should pass properly formualted TensorDim when Tensor::Map<T>

[layer] lstm layer
- reassess Tensor value-by-value eps
- idea: need to add more default constructors for TensorDim with TensorType param

**Self evaluation:**


nntrainer/layers/multi_head_attention_layer.cpp
```diff
@@ -184,73 +195,77 @@ void MultiHeadAttentionLayer::finalize(InitLayerContext &context) {
...
     weight_idx[AttentionParams::fc_bias] = context.requestWeight(
       fc_bias_dim, weight_initializer, weight_regularizer,
       weight_regularizer_constant, weight_decay, "fc_bias", true);
   }
 
   /** tensor for output of query fc */
   TensorDim projected_query_dim(
-    {batch_size, 1, query_height, num_heads * projected_query_dim_prop});
+    {batch_size, 1, query_height, num_heads * projected_query_dim_prop},
+    tensor_type);
```
jijoongmoon, line 254: I'm not sure if this is tensor_type or {context.getFormat(), context.getWeightDataType()}. plz check once again.

nntrainer/layers/multi_head_attention_layer.cpp
```diff
@@ -265,22 +280,23 @@ void MultiHeadAttentionLayer::finalize(InitLayerContext &context) {
...
     weight_idx[AttentionParams::dropout_mask] = context.requestTensor(
       dropout_mask_dim, "dropout_mask", Tensor::Initializer::NONE, false,
       TensorLifespan::ITERATION_LIFESPAN);
   }
 
   /** tensor for attention output */
   TensorDim attention_output_dim(
-    {batch_size, 1, query_height, num_heads * projected_value_dim_prop});
+    {batch_size, 1, query_height, num_heads * projected_value_dim_prop},
+    tensor_type);
```
jijoongmoon, line 299: I wonder if this is the tensor_type or {context.getFormat(), context.getActivationDataType()}. please check once again.

skykongkong8: Just fixed it. Also fixed setActivationFunc for the same reason.

myungjoo (approved):

skykongkong8:
This PR is too outdated.
Close for now

PR #2285, opened 2023-08-25
[Ahub] Fix Ahub issue @open sesame 09/20 12:27

SeoHyungjun:
- Fixed the part where buf is not initialized and used in nntrainer_logger.cpp. 
Array initialized with ascii 0 ('\0').

- changed 'auto' to 'auto &'

- The 'initialized' variable receives a pointer via malloc.
If malloc fails, it will be null.
However, since the exception is not handled, calling initialize[i] points to null + i.
Exception handling has been added to prevent this problem.

- Restore does not work as expected after changing ostream type.
Added because there is no code to restore the variable out.
Changed the code where the nesting depth was set incorrectly to the correct location.

- Previously, nntrainer only supported fp32. nntrainer didn't need to change the data type, but it needs to be changed to support fp16.
If the tensor type is fp16, get input_dim of InitLayerContext through getInputDimentions and call setDataType. 
For getInputDimentions, return a const object. Added getMutableInputDimentions function because return object of getInputDimentions cannot be modified.

- The second argument of tensor_dtype is used as std::regex(string).
But it didn't include handling for std::regex_error.
created a getRegex() function because it is used in a similar form in other codes. 
The getRegex() function takes a string and retrun a std::regex object.

### DIFF
--- a/Applications/meson.build
+++ b/Applications/meson.build
@@ -12,9 +12,9 @@ subdir('Resnet/jni')
 subdir('YOLO/jni')
 subdir('ReinforcementLearning/DeepQ/jni')
 subdir('TransferLearning/CIFAR_Classification/jni')
-if enable_capi
-  subdir('TransferLearning/Draw_Classification/jni')
-endif
+# if enable_capi
+#   subdir('TransferLearning/Draw_Classification/jni')
+# endif
 subdir('Custom')
 subdir('ProductRatings/jni')
 subdir('AlexNet/jni')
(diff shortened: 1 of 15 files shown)


baek2sm (approved):
LGTM

DonghakPark (approved):
LGTM!

skykongkong8 (approved):
LGTM!

DonghakPark:
@SeoHyungjun Please Rebase

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2266, opened 2023-08-01
[Wait for #2233][Wait for #2265][Application] incremental inference pico gpt in android

lhs8928:
[Application] apply incremental inference to pico gpt
    
     - Apply incremental inference to pico gpt
    
    Signed-off-by: hyeonseok lee <hs89.lee@samsung.com>

PR #2286, opened 2023-08-28
[ Wait for #2266] enable W16A16 for picoGPT Application

jijoongmoon:
### This PR includes fixes to run picoGPT application with

-   enable memory optimization options: to reduce memory creation during inference
-   adding FP16-FP16 options in model properties
-   remove unnecessary computation during the incremental_inference.
     : This will start with just one input at a time and accumulate the computation result in cache buffer
      at Multi_head attention layer. 
-   All tensor computation is conducted with NEON/SIMD implementation except GEMM.
-   Some fixes to compute the FP16
     : In layer normalization computation, it includes the variance computation which includes the pow operation.
       It is quite often to overflow and this pr includes by-pass implementation.

### DIFF
--- a/Applications/Android/PicoGPTJNI/app/src/main/jni/picogpt.cpp
+++ b/Applications/Android/PicoGPTJNI/app/src/main/jni/picogpt.cpp
@@ -13,12 +13,12 @@
  *
  */
 
-#include "picogpt.h"
 #include "encoder.hpp"
+#include "picogpt.h"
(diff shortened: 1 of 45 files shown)


hj210 (approved):
It looks good to me.

baek2sm (approved):
LGTM

Hyoungsukim (approved):
lgtm

PR #2288, opened 2023-08-29
[TensorDim] Fix TensorDim constructor

skykongkong8:
- Previously, there were lack of default constructors for 1-batch, 1-channel, 1-height, 1-width with currently added TensorType option (regardless of format).
- Since there are a lot of codes using such case in FP32 implementation, I had no choice but have to explicitly feed the tensor_type (which contains the fp16 info) to the previously defined tensor.
- For clean code, I would like to propose a new default constructor for better construction of TensorDim instance.

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/tensor_dim.h
+++ b/api/ccapi/include/tensor_dim.h
@@ -142,12 +142,51 @@ public:
    * @param c channel
    * @param h height
    * @param w width
-   * @param fm format NCHW | HNWC
+   * @param t_type format NCHW | HNWC , dataType FP32 | FP16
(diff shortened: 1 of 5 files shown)


DonghakPark (approved):
LGTM!,
It seems to be much more intuitive for users to use.

djeong20 (approved):
LGTM

baek2sm (approved):
LGTM

lhs8928 (approved):
LGTM

skykongkong8:
Nothing but rebased

myungjoo (approved):

Hyoungsukim (approved):
lgtm

PR #2300, opened 2023-08-31
[Wait for #2286] [SWAP] Implement inference mode

jihochu:
This patch is for inference mode for a swap device.
It adds "memory_swap_mode" property, which can handle inference mode when using swap device.
The weights are not modified while inferencing, therefore it doesn't need to swap out data to the device.


DonghakPark (approved):
LGTM!

myungjoo:
Please remove generated files including *.lock

jihochu:
> Please remove generated files including *.lock

thanks. It removed lock files.

PR #2301, opened 2023-09-04
[Tensor] Quantized Tensor (Int 4) with Scale

djeong20:
- Quantized Tensor is now available with Int 4 with scale.
- Two Int 4 values use one Int 8 which each use 4 bits
- Dequantization is performed by multiplying scaling factors with given index (b, c, h, w).
- Only read (getValueQint4), write (setValue), and dequantization operations are allowed.
    
**Self evaluation:**


Hyoungsukim (approved):
lgtm

jijoongmoon (approved):
LGTM!

PR #2302, opened 2023-09-04
[neon] Add padding function in sgemv_neon_fp16

skykongkong8:
- Previously, when the Tensor column length is not divisible by 8, we have been applying sgemv_loop, which is not an optimized solution.
- In this PR, I would like to propose adding arbitrary-sized zero-padding to the Tensor to fit in the neon intrinsic SIMD calculation and accelerate the computing process.

- todo : In case of large-columned Tensor we should discuss about better padding-applying strategy.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_neon.h
+++ b/nntrainer/tensor/blas_neon.h
@@ -62,6 +62,33 @@ void sgemv_neon_fp16(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t rows,
                      uint32_t cols, float alpha, float beta);
 
 /**
+ * @brief     sgemv computation with neon : Y = alpha*A*X + beta*Y using
+ * 0-padding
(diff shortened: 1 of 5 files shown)


nntrainer/tensor/blas_neon.h
```diff
@@ -15,6 +15,7 @@
 #ifdef __cplusplus
 
 #include <arm_neon.h>
+#include <vector>
```
djeong20, line 18:
```suggestion
```

remove?

skykongkong8: I intentionally leaved it since it is used in blas_neon.cpp
but I also think that it is better to be moved to .cpp file

DonghakPark (approved):
LGTM!

skykongkong8:
Higher PR beyond this had been already merged.
Closed.

PR #2307, opened 2023-09-11
[Tensor] Unsigned Quantized Tensor

djeong20:
- Quantized tensor values are unsigned with zero points
- Layer context dequantize quantized tensor when request weight
- Template dequantize function

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -183,16 +183,16 @@ void Tensor::allocate() {
       throw std::invalid_argument("Error: enable-fp16 is not enabled");
 #endif
     } else if (getDataType() == ml::train::TensorDim::DataType::QINT8) {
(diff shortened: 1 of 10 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -227,6 +241,31 @@ bool Tensor::operator==(const Tensor &rhs) const {
 #else
     throw std::invalid_argument("Error: enable-fp16 is not enabled");
 #endif
+  } else if (dim.getDataType() == ml::train::TensorDim::DataType::QINT8) {
+    const uint8_t *_data = getData<uint8_t>();
```
jijoongmoon, line 245: we might need to check the zero point and scale as well.

djeong20: applied.

Hyoungsukim (approved):
lgtm

songgot (approved):
LGTM

PR #2309, opened 2023-09-12
[Tensor] Dequantize tensor in forwarding FC layer 

djeong20:
- FC layer manually dequantizes quantized weight tensor in forwarding instead of automatically dequantizing in getWeight.
- Dequantize function takes the output axis as a parameter instead of the tensor having an axis variable.
- Add output axis in weight spec to identify the direction of scale and zero points multiplication.

**Self evaluation:**


skykongkong8 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2312, opened 2023-09-13
[Tensor] Read quantized tensor from a binary file

djeong20:
This patch adds functionality to read quantized tensors from a binary file. Details are as follows.

- Read the tensor in the following order (axis, scale factors, zero points, and values).
- `Tensor::read` takes an extra argument of datatype to identify the datatype of scale factors and read exact bytes.
- Fix QINT4 tensor print segfault issue.

**Self evaluation:**


baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2317, opened 2023-09-15
[Application] handle getcwd return null pointer @open sesame 09/15 16:46

lhs8928:
- Handle when getcwd return NULL to prevent dereference null pointer

### DIFF
--- a/Applications/SimpleShot/task_runner.cpp
+++ b/Applications/SimpleShot/task_runner.cpp
@@ -104,7 +104,9 @@ const std::string getcwd_() {
   const size_t bufsize = 4096;
   char buffer[bufsize];
 
-  return getcwd(buffer, bufsize);
+  char *cwd = getcwd(buffer, bufsize);
+  std::string ret = (cwd == NULL) ? "" : std::string(cwd);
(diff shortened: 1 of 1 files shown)


baek2sm (approved):
LGTM

Applications/SimpleShot/task_runner.cpp
```diff
@@ -104,7 +104,14 @@ const std::string getcwd_() {
...
   char buffer[bufsize];
 
-  return getcwd(buffer, bufsize);
+  char *cwd = getcwd(buffer, bufsize);
+  std::string ret;
+  if (cwd == NULL) {
+    ret = "";
+  } else {
+    ret = std::string(cwd);
+  }
```
djeong20, line 113: maybe in one line?
```suggestion
  std::string ret = (cwd == NULL) ? "" : std::string(cwd);
```

djeong20 (approved):
lgtm!

Applications/SimpleShot/task_runner.cpp
lhs8928: Applied

PR #2322, opened 2023-09-19
[Svace] Fix Ahub issues

SeoHyungjun:
Added exception handling when malloc fails.

The value of the idx variable has been changed to be initialized when declared.

In the tensor constructor, output_axis was changed to be initialized to 3.


DonghakPark (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2324, opened 2023-09-20
[tflite_export] add Error Message, Fix Application @open sesame 10/05 15:54

DonghakPark:
Now, an Error occurs when export resnet to tflite format

Previously, the prop of the layer was changed in the process of comparing the values with the Pytorch. But, Tflite only supports ```same``` and ```valid``` padding, but the current application use value as 1,1.

To correct this, a macro was added, and a more detailed error message was added to the tflite exporter.

### DIFF
--- a/nntrainer/utils/node_exporter.cpp
+++ b/nntrainer/utils/node_exporter.cpp
@@ -197,7 +197,12 @@ void Exporter::saveTflResult(
   auto strides = std::get<std::array<props::Stride, CONV2D_DIM>>(props);
   assert(strides.size() == CONV2D_DIM);
   auto padding = std::get<props::Padding2D>(props).get();
-  assert(padding == "same" || padding == "valid");
+  if (padding != "same" && padding != "valid") {
+    std::ostringstream ss;
+    ss << "Unsupported padding type; \"" << padding
+       << "\" is not supported. Use \"same\" or \"valid\".";
+    throw std::runtime_error(ss.str());
+  }
   auto options = tflite::CreateConv2DOptions(*fbb, tflite_padding(padding),
                                              strides.at(0), strides.at(1))
                    .Union();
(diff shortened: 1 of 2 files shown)


SeoHyungjun (approved):
LGTM!

baek2sm (approved):
LGTM.

But we need to check later whether implementation of ResNet is right because "stride 2 and padding same" is a little ambiguous.

DonghakPark:
> LGTM.
> 
> But we need to check whether implementation of ResNet is right later because "stride 2 and padding same" is a little ambiguous.

Yes, that right! But in tensorflow they only use "same", "valid" padding, so for export as tflite format we need this part.

nntrainer/utils/node_exporter.cpp
```diff
@@ -197,7 +195,10 @@ void Exporter::saveTflResult(
   auto strides = std::get<std::array<props::Stride, CONV2D_DIM>>(props);
   assert(strides.size() == CONV2D_DIM);
   auto padding = std::get<props::Padding2D>(props).get();
-  assert(padding == "same" || padding == "valid");
+  if (padding != "same" && padding != "valid") {
+    throw std::runtime_error("Unsupported padding type");
```
myungjoo, line 199: Give more info for the error. E.g., 
```
        std::ostringstream ss;
        ss << "Unsupported padding type; \"" << padding << "\" is not supported. Use \"same\" or \"valid\".";
        throw std::runtime_error(ss.str());
    }
```

myungjoo: If you have thrown an error, do you still need to do assert?

DonghakPark: Thank you for Review, Fix!!

myungjoo (approved):

PR #2311, opened 2023-09-13
[Bug] Fix build error in yolo with fp16 @open sesame 09/15 16:02

djeong20:
This patch fixes the build error in yolo_v2_loss with fp16 enabled. The following works are done.
- Template sqrtFloat and absFloat
- Remove unnecessary print in the fp16 unit test

**Self evaluation:**
1. Build test:   [X]Passed [ ]Failed [ ]Skipped
2. Run test:     [X]Passed [ ]Failed [ ]Skipped

PR #2325, opened 2023-10-05
[Wait for #2311] [blas/neon] Use inter-fp32 value in dimension shrinking computation in neon

skykongkong8:
- Previously, sgemm and sgemv in neon intrinsics was depdendent on two conditions
	1. should have 8-divisible column or row
	2. fully work with fp16 variables (which might cause precision loss while being accumulated)
- In this commit, I expect sgemm and sgemv to work like:
	1. support every column length with adaptive-length-compute optimization
	2. use temporal fp32 array to secure cumulative data value especially in large scale Tensor
	3. accelerate converting such fp32 array to fp16 Tensor and vice versa with neon to enhance time performance
	4. consider the number of register to avoid register spilling
- Moreover, with this new logic, we need to get rid of unittest cases that compares full fp16 variables with neon result. From now on, we should only compare our result with fp32 Tensors.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -6,6 +6,7 @@
  * @date   28 Aug 2020
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author Sungsik Kong <ss.kong@samsung.com>
  * @bug    No known bugs except for NYI items
  * @brief  This is dummy header for blas support
  *
@@ -32,16 +33,16 @@
     }                                        \
   } while (0);
 
-#define sgemv_loop_fp16(ci, cj, cM, cN)             \
-  do {                                              \
-    _FP16 y0;                                       \
-    unsigned int i, j;                              \
-    for (ci = 0; ci != cM; ci++) {                  \
-      y0 = Y[ci * incy] * static_cast<_FP16>(beta); \
-      for (cj = 0; cj != cN; cj++)                  \
-        y0 += A[i + j * lda] * X[cj * incx];        \
-      Y[ci * incy] = y0;                            \
-    }                                               \
+#define sgemv_loop_fp16(ci, cj, cM, cN)                                 \
+  do {                                                                  \
+    float y0;                                                           \
+    unsigned int i, j;                                                  \
+    for (ci = 0; ci != cM; ci++) {                                      \
+      y0 = static_cast<float>(Y[ci * incy] * static_cast<_FP16>(beta)); \
+      for (cj = 0; cj != cN; cj++)                                      \
+        y0 += static_cast<float>(A[i + j * lda] * X[cj * incx]);        \
+      Y[ci * incy] = static_cast<_FP16>(y0);                            \
+    }                                                                   \
   } while (0);
 
 #define saxpy_loop_fp16()                                                  \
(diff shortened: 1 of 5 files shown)


nntrainer/tensor/blas_interface.cpp
```diff
@@ -55,15 +56,15 @@
   do {                                                                    \
     for (unsigned int m = 0; m < M; ++m) {                                \
       for (unsigned int n = 0; n < N; ++n) {                              \
-        _FP16 c = 0;                                                      \
+        float c = 0;                                                      \
         _FP16 c_old = C[m * ldc + n];                                     \
         for (unsigned int k = 0; k < K; ++k) {                            \
           _FP16 a, b;                                                     \
           a = ((TransA == CblasTrans) ? A[k * lda + m] : A[m * lda + k]); \
           b = ((TransB == CblasTrans) ? B[n * ldb + k] : B[k * ldb + n]); \
```
djeong20, line 64:
```suggestion                                                 \
          _FP16 a = ((TransA == CblasTrans) ? A[k * lda + m] : A[m * lda + k]); \
          _FP16 b = ((TransB == CblasTrans) ? B[n * ldb + k] : B[k * ldb + n]); \
```
I know this isn't your work, though this might work better to possibly avoid Ahub issue.

nntrainer/tensor/blas_neon.cpp
```diff
@@ -278,250 +266,365 @@ void sgemv_transpose_neon(const float *A, const float *X, float *Y,
...
+      y0 = vfmaq_f32(y0, wvec8_11_f32, x8_11_f32);
+      y0 = vfmaq_f32(y0, wvec12_15_f32, x12_15_f32);
+      y0 = vfmaq_f32(y0, wvec16_19_f32, x16_19_f32);
+      y0 = vfmaq_f32(y0, wvec20_23_f32, x20_23_f32);
+      y0 = vfmaq_f32(y0, wvec24_27_f32, x24_27_f32);
+      y0 = vfmaq_f32(y0, wvec28_31_f32, x28_31_f32);
+
+      yVal = vaddvq_f32(y0);
+
+      Y32[j] += yVal;
```
djeong20, line 352:
```suggestion
      Y32[j] += vaddvq_f32(y0);
```
would this work? (removing all `float yVal`)

djeong20 (approved):
LGTM

nntrainer/tensor/blas_neon.cpp
skykongkong8: Good point. I will check and revise if it works.

nntrainer/tensor/blas_interface.h
```diff
@@ -35,77 +36,93 @@ enum CBLAS_TRANSPOSE {
 #include <helper_functions.h>
 #endif
 
+#include <cstdint>
 #include <tensor_dim.h>
-
 namespace nntrainer {
 
 #ifdef ENABLE_FP16
+/** @brief sscal computation */
```
jijoongmoon, line 44: It might be better to write the brief more formally.
For example,
/**
 * @brief     sgemm computation with neon : Y = alpha*op(A)*op(B) + beta*C,
 * where op(X) is one of X or X**T
 * @param[in] A __fp16 * for Matrix A
 * @param[in] B __fp16 * for Matrix B
 * @param[in] C __fp16 * for Matrix C
 * @param[in] M number of op(A)'s and C's row
 * @param[in] N number of op(B)'s and C's columns
 * @param[in] K number of op(A)'s and columns and op(B)'s rows
 * @param[in] alpha float number
 * @param[in] beta float number
 */

or 
you can use @copydoc as well.

skykongkong8: We have not been using @brief in blas_interface, but recent CI test is asking me to do so for some reaseon.
I temporally added a skeleton breif, but wondering if this is a good behavior or not.
Is it recommended to proceed with it?

jijoongmoon: We have to add doxygen comment in every function definition.

jijoongmoon (approved):
LGTM

PR #2326, opened 2023-10-06
[neon] Optmize sgemv

skykongkong8:
- Instead of declaring explicit register variable, declaring the function in inline code can save the number of register variable in use.
- This way, we can load more of variables to accelerate sgemv computation

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -521,131 +521,582 @@ void sgemv_transpose_neon_fp16(const __fp16 *A, const __fp16 *X, __fp16 *Y,
   for (; idx < cols; ++idx) {
     Y32[idx] = beta * Y[idx];
   }
+  if (rows % 16 == 0) {
+    for (unsigned int i = 0; i < rows; i += 16) {
+      float x = alpha * static_cast<float>(X[i]);
+      float x2 = alpha * static_cast<float>(X[i + 1]);
+      float x3 = alpha * static_cast<float>(X[i + 2]);
+      float x4 = alpha * static_cast<float>(X[i + 3]);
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM

nntrainer/tensor/blas_neon.cpp
```diff
@@ -479,131 +521,582 @@ void sgemv_transpose_neon_fp16(const __fp16 *A, const __fp16 *X, __fp16 *Y,
...
+      float x7 = alpha * static_cast<float>(X[i + 6]);
+      float x8 = alpha * static_cast<float>(X[i + 7]);
+      float x9 = alpha * static_cast<float>(X[i + 8]);
+      float x10 = alpha * static_cast<float>(X[i + 9]);
+      float x11 = alpha * static_cast<float>(X[i + 10]);
+      float x12 = alpha * static_cast<float>(X[i + 11]);
+      float x13 = alpha * static_cast<float>(X[i + 12]);
+      float x14 = alpha * static_cast<float>(X[i + 13]);
+      float x15 = alpha * static_cast<float>(X[i + 14]);
+      float x16 = alpha * static_cast<float>(X[i + 15]);
```
myungjoo, line 541: Does these lines get SIMD applied automatically by the compiler? (I'm not seeing ```vmulq_f32``` here...)

skykongkong8: These variables are not really declared for SIMD variables(e.g. float32x4_t) declaration, but used to minimize the function calls during the total iteration.
Here, we need horizontal calculation and horizontal accumulation of the values, rather than simd-by-simd calculation and vertical accumulation.
It has also proven by unittest that by this code, we could enjoy up to 10~20% time efficiency without accuracy loss.
Please let me know if you need anything more..!

PR #2327, opened 2023-10-11
[neon/sgemm] Partial accumulation using both fp16 and fp32 intrinsics

s-debadri:
**SGEMM** modifications made:

- Used mixed precision accumulations using both fp16 and fp32 intrinsics to enhance performance.
- Modified function calls to inline calls to reduce register spilling.
- Temporary fp32 storage used to enhance accuracy.

**Self evaluation:**


skykongkong8 (approved):
LGTM!

PR #2328, opened 2023-10-11
[Wait for #2327] [neon] Optmizie sgemm transB row-wise

skykongkong8:
- We can reduce the number of function calls by re-using the register variable both row-wise, col-wise.
- However, using both requires more-than-thirty-two (which is prohibited) register variables, which might cause register spilling. Thus, I only implemented row-wise optimization.
- This optimization is valid especially for large scale Tensor.
 
 **Self evaluation:**
    

jijoongmoon (approved):
LGTM

Hyoungsukim (approved):
lgtm

s-debadri (approved):
Nice optimizations. LGTM!

PR #2330, opened 2023-10-12
[#Wait for #2312][Tensor] Optimize dequantize operation    

djeong20:
- Perform dequantization by utilizing tensor operations instead of manual calculation.
- Tensor now contains two types of scale factors (FP32, FP16).
- Add flate function that copies tensor values of different data types. This function is temporal and should be later replaced by scopy to fasten up.

#### WTD
- Convert unsigned QINT to signed QINT to minimize computation (i.e., remove subtract)
- Current scopy for the quantized tensor only supports QINT4 to FP16. It needs to support QINT8-FP16, QINT4-FP32, QINT8-FP32

### DIFF
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -132,14 +132,7 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
     unsigned int axis =
       context.getWeightObject(weight_idx[FCParams::weight]).getOutputAxis();
 
-    if (dtype == nntrainer::Tdatatype::FP32)
-      weight.dequantize<float>(weight_, axis);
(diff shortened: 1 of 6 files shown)


skykongkong8 (approved):
LGTM!

baek2sm (approved):
LGTM

skykongkong8:
Is there any unittest cases for current scopy _int4_to_fp16 ?
I can implement the rest but it will be really helpful with it

djeong20:
> Is there any unittest cases for current scopy _int4_to_fp16 ? I can implement the rest but it will be really helpful with it

I'm afraid not. I haven't added it since there need to be several modifications to `scopy` with uint8_t anyways (e.g. replace uint8_t with int8_t).

we should later add to `unittest_nntrainer_tensor_neon_fp16.cpp` after changes are made.

PR #2331, opened 2023-10-12
[neon] Optimize sgemv 2.0

skykongkong8:
This PR includes:

- Optimize sgemv transpose with partial summation
 * 50% faster sgemv_transpose in case of large scale Tensors
 * Support every column length (bugFix)

- Optimize sgemv noTrans with more variables
 * 20% faster sgemv in case of large scale Tensors


DonghakPark (approved):
LGTM!

nntrainer/tensor/blas_neon.cpp
```diff
@@ -351,10 +413,8 @@ void sgemv_neon_fp16(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t rows,
       y = vfmaq_f16(y, wvec48_55, x48_55);
```
s-debadri, line 413: These wvecX_Y variables can also be written inline as previous. Will improve performance.

skykongkong8: Fixed. Thanks!

jijoongmoon (approved):
LGTM

PR #2335, opened 2023-10-16
[neon/bugFix] Classify which sgemv optimizatized code block to use by col length

skykongkong8:
- Since we are relying on float16, using large batch for float16 computation might hinder accuracy when it comes to small-sized Tensors.
- Through unittest, I discovered there is a huge round-off error when performing non-4-divisible column-ed Tensor sgemv with batch size of 16, while batch size 8 is almost identical to batch size 1. Thus, I temporally blocked batch size 16 computation when column is not divisible with 4.
- Arbitrary classification implemented grounded on unittest, but optimal row length classification is needed.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -276,6 +276,7 @@ void sgemv_transpose_neon(const float *A, const float *X, float *Y,
 
 void sgemv_neon_fp16(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t rows,
                      uint32_t cols, float alpha, float beta) {
+  const int batch = 0;
   const __fp16 *__restrict x;
   float Y32[rows];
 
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/blas_neon.cpp
```diff
@@ -276,6 +276,7 @@ void sgemv_transpose_neon(const float *A, const float *X, float *Y,
 
 void sgemv_neon_fp16(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t rows,
                      uint32_t cols, float alpha, float beta) {
+  const int batch = 0;
```
s-debadri, line 279: Should this be 0 ?

s-debadri (approved):
Please remove the std::cout used for testing. Other than that changes are fine. Nice additions!

nntrainer/tensor/blas_neon.cpp
skykongkong8: According to my current test, noTrans sgemv does not really bring significant round-off error with higher batches.
I think it is because there is more shrinking dimension per iter in SGEMV transpose than SGEMV noTrans.
However, it might cause potential error in the future (e.g. it might cause significant round-off error in unbalanced Tensors, sparse Tensors, or Tensors with big numbers) and we might some classification. Thus I temporally put zero in the code here for now

s-debadri: Okay got it.

s-debadri (approved):
LGTM

jijoongmoon (approved):
LGTM

baek2sm (approved):
LGTM

PR #2342, opened 2023-10-19
Use GEMM computation for initial sequence

jijoongmoon:
This PR uses GEMM to compute initial sequences.

also, make toknizer as defalut.

**Self evaluation:**
1. Build test:	 [X]Passed [ ]Failed [ ]Skipped
2. Run test:	 [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: jijoong.moon <jijoong.moon@samsung.com>

PR #2343, opened 2023-10-19
[Wait for #2342] [ Weight ] Add packed property and output axis in weight spec 

jijoongmoon:
In This PR includes,

. enable packed (bool) property for layer node. It is only for the
weight. If it is false, it follows the global activation datatype; if it is true, 
it will follow the global weight datatype.

. add output axis in Weight Spec and set the private variable in weight.
  it is to find the right direction for multiplying scales and zero-point

**Self evaluation:**


baek2sm (approved):
LGTM

baek2sm (approved):
LGTM

skykongkong8 (approved):
LGTM!

djeong20 (approved):
LGTM!

PR #2345, opened 2023-10-23
[neon] Support scopy for multiple dataTypes in neon

skykongkong8:
- scopy_int4_to_fp32
- scopy_int8_to_fp32
- scopy_int8_to_fp16
- scopy_int8_or_int4 : Since we use uint8_t for int4 Tensors, codes can be shared here
- vcvt_fp32_u32_bitwise : Faster optimization can be done with bitwise operation rather than elementwise operation

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -159,8 +159,8 @@ static void scopy_FP16(const unsigned int N, const _FP16 *X, const int incX,
 #endif
 }
 
-static void scopy_INT4(const unsigned int N, const uint8_t *X, const int incX,
-                       _FP16 *Y, const int incY) {
+static void scopy_int4_to_fp16(const unsigned int N, const uint8_t *X,
+                               const int incX, _FP16 *Y, const int incY) {
   unsigned int incy = abs(incY);
   unsigned int incx = abs(incX);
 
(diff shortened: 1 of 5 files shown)


skykongkong8:
To reviewers, I am wondering about renaming function name 'scopy'.
As far as I am concerned, 's' for scopy refers to single-precision, and thus it does not really represent current multiType-supporting behavior.
Please leave your opinions.. thanks!

nntrainer/tensor/blas_interface.cpp
```diff
@@ -302,9 +321,14 @@ void scopy(const unsigned int N, const _FP16 *X, const int incX, _FP16 *Y,
   scopy_FP16(N, X, incX, Y, incY);
 }
 
-void scopy(const unsigned int N, const uint8_t *X, const int incX, _FP16 *Y,
-           const int incY) {
-  scopy_INT4(N, X, incX, Y, incY);
+void scopy_int4_to_float16(const unsigned int N, const uint8_t *X,
+                           const int incX, _FP16 *Y, const int incY) {
+  scopy_int4_to_fp16(N, X, incX, Y, incY);
```
djeong20, line 326: any reason to use `scopy_int4_to_fp16` instead of having its implementation?

djeong20 (approved):
As far as there are unit tests to double-check the logic, LGTM

skykongkong8:
> As far as there are unit tests to double-check the logic, LGTM

I checked in separate branch with vanilla copying, but more check with proper dequantize function is required.

nntrainer/tensor/blas_interface.cpp
skykongkong8: I followed convention in blas_interface.cpp
I think it is because we can add-on cuBLAS or any other calculation-accelerating methods

djeong20: makes sense

nntrainer/tensor/blas_neon.cpp
```diff
@@ -1104,6 +1223,14 @@ void sscal_neon_fp16(const unsigned int N, __fp16 *X, const float alpha) {
     X[idx] = alpha * X[idx];
 }
 
+float32x4_t vcvtq_f32_u32_bitwise(uint32x4_t u32) {
+  constexpr uint32_t offsetValue = 0x4b000000;
+  const uint32x4_t offsetInt = vdupq_n_u32(offsetValue);
+  u32 = vorrq_u32(u32, offsetInt);
+  return vsubq_f32(vreinterpretq_f32_u32(u32),
```
wooksong, line 1230:
```suggestion
  return vsubq_f32(vreinterpretq_f32_u32(vorrq_u32(u32, offsetInt)),
```

skykongkong8: Fixed.

jijoongmoon (approved):
LGTM

PR #2355, opened 2023-10-30
[layer] Backwarding fp16 layers

skykongkong8:
- This PR includes backwarding implementations of concat, dropout, lstm, embedding, multiHeadAttention layer and their test suites.
- Renamed notation 'fp16fp16' to 'w16a16' for better understanding.

Self-evaluation:


baek2sm (approved):
LGTM

djeong20 (approved):
LGTM

SeoHyungjun (approved):
LGTM!

nntrainer/layers/concat_layer.cpp
```diff
@@ -118,25 +118,50 @@ void ConcatLayer::forwarding(RunLayerContext &context, bool training) {
...
-      /** loop over the concat dimension itself */
-      for (unsigned int count = 0; count < irh.height(); count++) {
-        Tensor dest_tensor = Tensor::Map<float>(
-          output.getAddress<float>(batch, 0, output_height_offset + count, 0),
-          data_copy_size * sizeof(float), {1, 1, 1, data_copy_size});
-        const Tensor source_tensor = Tensor::Map<float>(
-          input.getAddress(batch, 0, count, 0), data_copy_size * sizeof(float),
-          {1, 1, 1, data_copy_size});
-        dest_tensor.copy(source_tensor);
+    if (in_dim.getDataType() == TensorDim::DataType::FP32) {
```
SeoHyungjun, line 129: Personally, I’m curious.

This part seems almost similar to the code after line 238 below.
Wouldn’t it be possible to separate it into a function?!

skykongkong8: You mean using `ConcatLayer::forwarding` inside of `ConcatLayer::calcDerivative` ?
If your point is right, that would reduce code length. However, it is not the main purpose of this PR...
Maybe we can make a separate PR after double-checking..!

PR #2356, opened 2023-10-31
[Application] fix deepq to make it run

lhs8928:
- Make res directory and move DeepQ.ini file to res dir
 - If the input batch size does not match with batch size property of model graph, set model graph batch size property as input batch size in forwarding function
 - Allocate weight/tensor memory as train mode of mainNet, targetNet network
 - Comment "return 1" statement so as to make it as train from scratch


jijoongmoon (approved):
LGTM

baek2sm (approved):
LGTM

PR #2359, opened 2023-11-02
[Tensor] Support multiple data types in `copyData`

djeong20:
Previously, `copyData` only supported copying data of the same data type.
Copying data of different data types is needed with the increased demand for flexibility of mixed-precision.

- `copyData` supports copying data of different types with the use of NEON
- Remove the flate function
- utilize `copyData` in dequantize

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -3041,17 +3041,46 @@ void Tensor::copyData(const Tensor &from) {
   if (size() != from.size())
     throw std::invalid_argument("Size of tensor to copy must match");
 
-  if (getDataType() != from.getDataType())
-    throw std::invalid_argument("Data type of tensor to copy must match");
-
-  if (getDataType() == ml::train::TensorDim::DataType::FP32) {
-    copy(from.getData<float>());
-  } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
+  if (getDataType() == from.getDataType()) {
+    if (getDataType() == ml::train::TensorDim::DataType::FP32) {
+      copy(from.getData<float>());
+    } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
-    copy(from.getData<_FP16>());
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -3041,17 +3041,46 @@ void Tensor::copyData(const Tensor &from) {
...
+        scopy_int8_to_float32(from.size(), from.getData<uint8_t>(), 1,
+                              getData<float>(), 1);
+      } else if (from.getDataType() == ml::train::TensorDim::DataType::QINT4) {
+        scopy_int4_to_float32((from.size() + 1) / 2, from.getData<uint8_t>(), 1,
+                              getData<float>(), 1);
+      }
+    } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
+#ifdef ENABLE_FP16
+      if (from.getDataType() == ml::train::TensorDim::DataType::QINT8) {
+        scopy_int8_to_float16(from.size(), from.getData<uint8_t>(), 1,
```
skykongkong8, line 3074: Quick question:
I am considering renaming the term `scopy` to something else...
"s" here means "single precision(fp32)", so it really cannot be used in situation like this.
Any idea for the name of copying function of int8 <-> int4 <-> float16?

skykongkong8 (approved):
LGTM!

nntrainer/tensor/tensor.cpp
djeong20: I don't think there are standard prefixes for these data types. maybe we could remove 's' and go with `copy_int8_float16`.

skykongkong8: Think so too.
Thanks!

DonghakPark (approved):
LGTM!

SeoHyungjun:
LGTM

Hyoungsukim (approved):
lgtm

baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2361, opened 2023-11-03
[Utils] Conversion to/from half-precision floating point

djeong20:
This PR includes utility functions for conversion to/from 16-bit floating point number, in bit representation, from/to 32-bit floating point number in IEEE format.

**Self-evaluation:**


skykongkong8 (approved):
LGTM
Interface for half-precision will be so much better with this PR!

baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM! It is really useful for Mixed Precision Training as well.

PR #2363, opened 2023-11-07
[Tensor] Scaling functions for `Tensor`

skykongkong8:
- For mixed precision training, we must support loss scaling to preserve round-off accuracy from half-precision.
- Scale factor for such training is generally a model hyperparameter, but here we gave it as a Tensor parameter for more subtle handling in lower level. Refer to each commit for propsed changes.
- Loss Scaler, which will govern all scaling situations for Tensor, Weight, Optimizer will be propsed in the near future.
- Add unittest accordingly.

For easier review, followings are the functions I revised to convey/consider `scale_factors_fp32` :
* `sum_by_batch`
* `sum`
* `multiply`
* `divide`
* `add`
* `dot`
* `transpose`
* `clone`
* `rotate_180`
* `scale`
* `descale`

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -20,7 +20,10 @@
 #endif
 
 #include <cmath>
+<<<<<<< HEAD
 #include <iostream>
+=======
+>>>>>>> 011e1c23... [Tensor] Consider scale factor in math ops
 
 #define sgemv_loop(ci, cj, cM, cN)           \
   do {                                       \
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -15,7 +15,7 @@
 #include <blas_interface.h>
 #include <nntrainer_error.h>
 
-#if (defined USE__FP16 && defined USE_NEON)
+#if (defined(USE__FP16) && defined(USE_NEON))
 #include <blas_neon.h>
 #endif
 
@@ -73,14 +73,14 @@
 
 namespace nntrainer {
 
-#ifdef ENABLE_FP16
+#if defined(ENABLE_FP16)
 static void saxpy_FP16(const unsigned int N, const float alpha, const _FP16 *X,
                        const int incX, _FP16 *Y, const int incY) {
   if (incX < 0 or incY < 0)
     throw std::invalid_argument(
       "Error: negative inc not supported without cblas");
 
-#if (defined USE__FP16 && USE_NEON)
+#if (defined(USE__FP16) && defined(USE_NEON))
   // USE__FP16 is defined when platform is android
   if (incX == 1 && incY == 1) {
     nntrainer::neon::saxpy_neon_fp16(N, alpha, X, Y);
@@ -102,13 +102,13 @@ static void sgemv_FP16(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA,
   unsigned int incx = abs(incX);
 
   if (TransA == CblasTrans) {
-#if (defined USE__FP16 && USE_NEON)
+#if (defined(USE__FP16) && defined(USE_NEON))
     nntrainer::neon::sgemv_transpose_neon_fp16(A, X, Y, M, N, alpha, beta);
 #else
     sgemv_loop_fp16(i, j, N, M);
 #endif
   } else {
-#if (defined USE__FP16 && USE_NEON)
+#if (defined(USE__FP16) && defined(USE_NEON))
     nntrainer::neon::sgemv_neon_fp16(A, X, Y, M, N, alpha, beta);
 #else
     sgemv_loop_fp16(j, i, M, N);
@@ -125,7 +125,7 @@ static _FP16 sdot_FP16(const unsigned int N, const _FP16 *X,
 
   _FP16 ret = 0;
 
-#if (defined USE__FP16 && USE_NEON)
+#if (defined(USE__FP16) && defined(USE_NEON))
   if (incX == 1 && incY == 1) {
     ret = nntrainer::neon::sdot_neon_fp16(N, X, Y);
   } else {
@@ -146,7 +146,7 @@ static void scopy_FP16(const unsigned int N, const _FP16 *X, const int incX,
   unsigned int incy = abs(incY);
   unsigned int incx = abs(incX);
 
-#if (defined USE__FP16 && USE_NEON)
+#if (defined(USE__FP16) && defined(USE_NEON))
   if (incX == 1 && incY == 1) {
     nntrainer::neon::scopy_neon_fp16(N, X, Y);
   } else {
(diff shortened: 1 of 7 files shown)


nntrainer/tensor/blas_interface.cpp
```diff
@@ -235,22 +235,24 @@ static void scopy_int8_to_fp16(const unsigned int N, const uint8_t *X,
 }
 
 static void ewvm_FP16(const unsigned int N, const _FP16 *X, const _FP16 *Y,
-                      _FP16 *Z) {
+                      _FP16 *Z, const float alpha) {
 #if (defined USE__FP16 && USE_NEON)
```
myungjoo, line 239: This preprocessor expression is not recommended.

https://stackoverflow.com/questions/1312132/boolean-in-ifdef-is-ifdef-a-b-the-same-as-if-defineda-definedb

skykongkong8: Then, as I presume:
`defined USE__FP16 && USE_NEON` to `defined USE__FP16 && define USE_NEON` ?

djeong20: what's the difference b/w `USE__FP16` and `USE_NEON` in the first place?

skykongkong8: First, my apologies for lack of understanding with those... but as far as I am concerned:

it is to differentiate fp16-enabled true/false with android/non-android build
neon is only valid for ARM.

djeong20: `USE_NEON` is currently enabled for both Android/non-android in `meson.build`. maybe we should fix this.

```
if get_option('enable-fp16')
   arch = target_machine.cpu_family()
   if get_option('platform') == 'android'
     extra_defines += '-DUSE_NEON=1'
   elif arch == 'aarch64'
     extra_defines += '-DUSE_NEON=1'
   else
     has_avx512fp16 = cc.has_argument('-mavx512fp16')
     if (has_avx512fp16)
       # add_project_arguments(['-mavx512fp16'], language: ['c','cpp'])
       message ('Float16 for x86_64 enabled. Modern gcc-x64 genrally supports float16 with _Float16. -mavx512fp16 added for hardware acceleration')
     else
       warning ('Float16 for x86_64 enabled. However, software emulation is applied for fp16, making it slower and inconsistent. Use GCC 12+ for AVX512 FP16 support. This build will probably fail unless you bring a compiler that supports fp16 for x64.')
     endif
   endif  
endif
```

djeong20: Also, I think we can replace `defined USE__FP16 && defined USE_NEON` to just `#if USE_NEON` since `USE__FP16` is inclusive

skykongkong8: First of all, `aarch64` also supports neon since it is 64bit ARM I guess?
Second of all, we still need it because neon also do work for float32.

For example,
1. fp16  / fp32 with no neon -> raw BLAS function or something else ( cBLAS / OpenCL, ... )
2. fp16 / fp32 with neon -> neon BLAS function

sometimes we might want to build with option such as: using neon but not fp16 (though it is not a typical case for now)

skykongkong8: I mean we are making more options like:
fp32-exclusive / fp16fp32 both  || neon / non-neon

djeong20: thanks for clarifying! to summarize, would things be like this?

- [x] USE_NEON=1 & USE__FP16=1: fp16/fp32 with neon BLAS
- [ ] (WTD) USE_NEON=1 & USE__FP16=0: fp32 only with neon BLAS 
- [ ] (WTD) USE_NEON=0 & USE__FP16=1: fp16/fp32 only with cBlas, openCL
- [ ] (WTD) USE_NEON=0 & USE__FP16=0: fp32 only with cBlas, openCL

skykongkong8: Happy to help :) That is correct.

myungjoo: Change ```#if defined KEYWORD``` --> ```#if defined (KEYWORD)```.
And, for ```#if defined X && defined Y```, use ```if defined(X) && defined(Y)```, for the readability and compatibility with compilers.

skykongkong8: Thank you for your clear guide!
Applied suggested changes.

nntrainer/tensor/blas_interface.cpp
```diff
@@ -182,7 +182,7 @@ static void scopy_float16_to_float32(const unsigned int N, const _FP16 *X,
   unsigned int incy = abs(incY);
   unsigned int incx = abs(incX);
 
-#if (defined USE__FP16 && USE_NEON)
+#if (defined(USE__FP16) && USE_NEON)
```
myungjoo, line 185: Because you've added ```-DUSE_NEON=1``` in meson.build, it should work, but in case you didn't do so in Android build, (I don't see -DUSE_NEON in jni directory) please correct this, too.

myungjoo (commented):
Other than the one line not corrected, it looks good.

nntrainer/tensor/blas_interface.cpp
skykongkong8: Apparently I missed it..
Fixed.

djeong20 (approved):
LGTM

djeong20 (commented):
maybe we could use negative test cases as well? (e.g. try scaling FP32 tensor)

skykongkong8:
> maybe we could use negative test cases as well? (e.g. try scaling FP32 tensor)

That would be meaningful, but I think it would be much  more efficient to work after applying your local work.
If we are able to separate Tensor files by type, it is quite straight-forward to do so I guess.
Or maybe you can temporally modify `Tensor::scale()` to check its `tensor_type` and throw `invalid_argument` to observe similar effect.

jijoongmoon:
@djeong20 , I think we are also considering blas operation when we restructure the tensor for multi-type 
```
Tensor --> FloatTensor --> fp32blas --> naive
                                    --> openBLAS
       --> HalfTensor  --> fp16blas  --> naive
                                     --> NEON
```

Hyoungsukim (approved):
lgtm

nntrainer/tensor/blas_interface.cpp
```diff
@@ -73,14 +73,14 @@
 
 namespace nntrainer {
 
-#ifdef ENABLE_FP16
+#if defined(ENABLE_FP16)
```
SeoHyungjun, line 76: Wouldn't ifdef be more concise if it only has one argument?

SeoHyungjun (approved):

jijoongmoon (approved):
LGTM

myungjoo (approved):
Good to go. Please rebase and move on.

skykongkong8:
This PR is deprecated. Close.

PR #2364, opened 2023-11-07
[trivial/bugfix] Add `inference_only_option` in multi_head_attention

skykongkong8:
- We were using `inference_only_option` for multi_head_attention fp16 unittest since we do not have loss scaling implementation yet.
- However, I discovered that there was a missing declaration of such option which might make malfunction when build, and added accordingly.

**Self evaluation:**


myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #2367, opened 2023-11-10
[Tensor] HalfTensor class for 16-bit floating point

djeong20:
This PR includes creating the HalfTensor class which separates 16-bit floating point calculation from nntrainer::Tensor.

- Create a HalfTensor class that only handles 16-bit floating point calculation.
- Remove operations for Quantized Tensor.

### DIFF
--- a/nntrainer/tensor/half_tensor.cpp
+++ b/nntrainer/tensor/half_tensor.cpp
@@ -1,0 +1,2291 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2023 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file	 half_tensor.cpp
+ * @date	 09 November 2023
+ * @brief	 This is HalfTensor class for calculation
+ * @see		 https://github.com/nnstreamer/nntrainer
+ * @author	 Donghyeon Jeong <dhyeon.jeong@samsung.com>
+ * @bug		 No known bugs except for NYI items
+ */
+
+#include <algorithm>
+#include <assert.h>
+#include <cmath>
+#include <cstring>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <iterator>
+#include <numeric>
+#include <regex>
+#include <sstream>
+#include <stdexcept>
+#include <stdio.h>
+
+#include <half_tensor.h>
+#include <util_func.h>
+
+#define transposeloop(cl, ci, cj, ck, sl, si, sj, sk)                 \
+  do {                                                                \
+    unsigned int i, j, k, l;                                          \
+    int inidx = 0, outidx = 0;                                        \
+    for (cl = 0; cl < sl; cl++)                                       \
+      for (ci = 0; ci < si; ci++)                                     \
+        for (cj = 0; cj < sj; cj++)                                   \
+          for (ck = 0; ck < sk; ck++) {                               \
+            outidx = si * sj * sk * cl + sj * sk * ci + sk * cj + ck; \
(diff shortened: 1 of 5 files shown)


skykongkong8:
Could you share source codes of `Tensor` with virtual functions as well?

djeong20:
> Could you share source codes of `Tensor` with virtual functions as well?

sure I'll make a separate PR when the work is settled.

nntrainer/tensor/meson.build
```diff
@@ -34,6 +35,7 @@ arch = target_machine.cpu_family()
 
 if get_option('platform') == 'android' or arch == 'aarch64'
   tensor_sources += 'blas_neon.cpp'
+  tensor_sources += 'half_tensor.cpp'
```
jijoongmoon, line 38: I think this is added when enable_fp16 is true.

nntrainer/tensor/half_tensor.cpp
```diff
@@ -0,0 +1,2292 @@
...
+#include <numeric>
+#include <regex>
+#include <sstream>
+#include <stdexcept>
+#include <stdio.h>
+
+#include <half_tensor.h>
+#include <util_func.h>
+
+#define transposeloop(cl, ci, cj, ck, sl, si, sj, sk)                 \
```
jijoongmoon, line 30: This define can be shared through out the tensor. so.. place it in Tensor Class?

nntrainer/tensor/meson.build
djeong20: Applied!

nntrainer/tensor/half_tensor.cpp
djeong20: You're right. In fact, I think the `transposeloop` and `transpose` functions should be in the future operation class (which includes multiply, sum, etc). For now, I left the operation functions intentionally for simplicity in verifying the operation class though it will be removed soon.

jijoongmoon (approved):
LGTM

nntrainer/tensor/half_tensor.cpp
jijoongmoon: Agreed. Thanks.

Hyoungsukim (approved):
lgtm

nntrainer/tensor/half_tensor.h
```diff
@@ -0,0 +1,1835 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2023 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file	 half_tensor.h
+ * @date	 09 November 2023
+ * @brief	 This is HalfTensor class for calculation
+ * @see		 https://github.com/nnstreamer/nntrainer
+ * @author	Donghyeon Jeong <dhyeon.jeong@samsung.com>
```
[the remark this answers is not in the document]
djeong20: Fixed!

nntrainer/tensor/half_tensor.h
```diff
@@ -0,0 +1,1835 @@
...
+    HE_NORMAL,      /** He normal initialization */
+    HE_UNIFORM,     /** He uniform initialization */
+    NONE            /** No initialization */
+  };
+
+  /**
+   * @brief     Basic Constructor of HalfTensor
+   */
+  HalfTensor(std::string name_ = "", Tformat fm = Tformat::NCHW,
+             Tdatatype d_type = Tdatatype::FP16) :
```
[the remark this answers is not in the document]
djeong20: Yes. maybe this isn't necessary. I'll try to figure out if I could remove this. thanks :)

PR #2368, opened 2023-11-16
[Tensor] Initial Draft of Tensor Version 2

djeong20:
In this PR, the initial working version of Tensor V2 is included.

- Create a TensorV2 class that follows a type erasure design pattern.
- Create a new SrcSharedTensorV2, which is a class template.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor_v2.cpp
+++ b/nntrainer/tensor/tensor_v2.cpp
@@ -1,0 +1,176 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2023 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file	tensor_v2.cpp
+ * @date	16 November 2023
+ * @brief	This is Tensor class in Type erasure design
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Donghyeon Jeong <dhyeon.jeong@samsung.com>
+ * @bug		No known bugs except for NYI items
+ */
+
+#include <algorithm>
+#include <assert.h>
+#include <cmath>
+#include <cstring>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <iterator>
+#include <numeric>
+#include <regex>
+#include <sstream>
+#include <stdexcept>
+#include <stdio.h>
+
+#include <lazy_tensor.h>
+#include <tensor_v2.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+TensorV2::TensorV2(std::string name_, Tformat fm, Tdatatype d_type) {
+  if (d_type == Tdatatype::FP32) {
+    object = std::shared_ptr<TensorBase<FloatTensor>>(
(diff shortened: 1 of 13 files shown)


nntrainer/tensor/tensor_v2.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <stdio.h>
+
+#include <lazy_tensor.h>
+#include <tensor_v2.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+TensorV2::TensorV2(std::string name_, Tformat fm, Tdatatype d_type) {
+  if (d_type == Tdatatype::FP32) {
```
jihochu, line 34: It may need to use a trick to get the benefit from the type erasure pattern.
Checking each type is not desirable, using template type deduction would be.

I'm not sure, but pseudocode can be,

```
  class HalfTensor;                                                                     
  class FloatTensor;
  namespace TensorType {
    HalfTensor *FP16 = nullptr;                                                                                                                                         
    FloatTensor *FP32 = nullptr;                                                                                                                                        
  };                                                                                                                                                                    
                                                                                                                                                                        
  template<typename T=FloatTensor>                                                                                                                                                  
  class Tensor {
  public:
   Tensor(A a, B b, T *) {
     object = new TesnorBase<T>(T(a, b));              
    }                                                                                                                                                           
  
  private:
   TensorBase<T> *object;                                                                                 
  };                                                                                                       
  
///////////////////////////////
  auto m = new Tensor(a, b);
auto n = new Tensor(a, b, TensorType::FP16);

```

djeong20: that's really a good point! I'll utilize the TensorBase constructor instead of handling data type in the TensorV2 constructor and update if it works.

djeong20: That's a good point. To explain the constructor implementation, the classic way of declaring TensorV2 (from the type erasure pattern) would be as follows. 

 ```
TensorV2 fp32_tensor = TensorV2(FloatTensor( ... ));
TensorV2 fp16_tensor = TensorV2(HalfTensor( ... ));
```
There's no need to make TensorV2 a class template.

However, declaring TensorV2 would differ from the original Tensor.

 `TensorV2(FloatTensor(const TensorDim &d, const void *buf = nullptr))`
vs.
 `Tensor(const TensorDim &d, const void *buf = nullptr))`
 
I wasn't sure which would be preferable, and I chose to implement TensorV2 constructors that are the same as Tensor constructors. In this case, checking each type is inevitable. What would be a better choice for the TensorV2 constructor?

jihochu: As we discussed offline, it is required to keep codes and structures as many as possible (as Jijong's comment).
There will be many constraints with the template class, so current code would be good.

jihochu (approved):
LGTM

DonghakPark (approved):
LGTM!

SeoHyungjun (approved):
LGTM

skykongkong8 (approved):
LGTM

PR #2378, opened 2023-12-04
[Tensor] Refactored TensorV2 class

djeong20:
In this PR, the TensorV2 structure is refactored instead of following a Type erasure pattern.

- TensorV2 is a target, expected for a user to use, that contains a TensorBase pointer.
- TensorBase is an abstract class that provides default infrastructure code.
- FloatTensor class inherits the TensorBase class and overrides the pure virtual methods with 32-bit floating point calculation.
- HalfTensor class inherits the TensorBase class and overrides the pure virtual methods with 16-bit floating point calculation.

**Note**
This is a skeleton to build the structure with no implementation.

**Self-evaluation:**

### DIFF
--- a/debian/nntrainer-dev.install
+++ b/debian/nntrainer-dev.install
@@ -9,6 +9,10 @@
 # tensor headers
 /usr/include/nntrainer/memory_data.h
 /usr/include/nntrainer/tensor.h
+/usr/include/nntrainer/tensor_v2.h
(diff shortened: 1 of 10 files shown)


Hyoungsukim (approved):
lgtm

jijoongmoon (approved):
LGTM!

PR #2386, opened 2023-12-14
[embed] match temporary weight data type with weight data type

lhs8928:
- Create weight dim to handle weight data type and output data type does not match

### DIFF
--- a/nntrainer/layers/embedding.cpp
+++ b/nntrainer/layers/embedding.cpp
@@ -91,6 +91,7 @@ void EmbeddingLayer::forwarding(RunLayerContext &context, bool training) {
   Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
   TensorDim out_tensor_dim =
     TensorDim({1, 1, 1, out_dim}, hidden_.getTensorType());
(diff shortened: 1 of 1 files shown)


djeong20 (approved):
LGTM

nntrainer/layers/embedding.cpp
```diff
@@ -104,7 +105,7 @@ void EmbeddingLayer::forwarding(RunLayerContext &context, bool training) {
       }
 
       Tensor cur_weight =
-        weight.getSharedDataTensor(out_tensor_dim, out_dim * embed_idx);
+        weight.getSharedDataTensor(weight_dim, out_dim * embed_idx);
```
jijoongmoon, line 108: I think embedding weight can differ from the output tensor. So it is correct to handle both of them differently as you pointed out. But check once again if the ```out_tensor.copyData(cur_weight)```is working properly for different tensors.  for example, ```FP16.copyData(Qint8)```

skykongkong8 (approved):
LGTM

SeoHyungjun (approved):
LGTM!

PR #2387, opened 2023-12-15
[ Ahub ] Fix svace issue

skykongkong8:
- Fixes : TensorV2 may not initialize itensor
- Fixes : After dynamic memory allocation of itensor, there is no destructor

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor_v2.cpp
+++ b/nntrainer/tensor/tensor_v2.cpp
@@ -24,7 +24,11 @@ TensorV2::TensorV2(std::string name_, Tformat fm, Tdatatype d_type) {
 #else
     throw std::invalid_argument("Error: enable-fp16 is not enabled");
 #endif
-  }
+  } else
+    throw std::invalid_argument(
+      "Error: TensorV2 cannot be constructed because the given d_type is not "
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/tensor_v2.cpp
```diff
@@ -24,7 +24,8 @@ TensorV2::TensorV2(std::string name_, Tformat fm, Tdatatype d_type) {
 #else
     throw std::invalid_argument("Error: enable-fp16 is not enabled");
 #endif
-  }
+  } else
+    throw std::invalid_argument("Error : TensorV2 itensor not initialized");
```
myungjoo, line 28: Error messages should be able to let the readers correct their codes without much effort.
Provide more info about what's wrong and how to correct it.

E.g., ```Error: TensorV2 itensor cannot be initialized because the given d_type is not compatible. The supported d_types are: FP32, FP16 (if built with ENABLE_FP16)).```

skykongkong8: Got it. Will fix it asap

lhs8928 (approved):
LGTM

myungjoo (approved):

PR #1964, opened 2022-07-19
[multi head attention] added unittest

lhs8928:
- Added layer/model unittest for multi head attention
    
    Signed-off-by: hyeonseok lee <hs89.lee@samsung.com>

PR #2391, opened 2024-01-02
[Tensor] Comparison operator overloading

djeong20:
This PR includes the implementation of comparison operators for TensorV2-related classes.

- TensorBase comparison operator compares Tensor information such as TensorDim.
- Float/HalfTensor comparison operator checks Tensor data.
- Destructor implementation is removed and set to default due to the rule of five.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -87,6 +87,20 @@ FloatTensor::FloatTensor(
   }
 }
 
+bool FloatTensor::operator==(const FloatTensor &rhs) const {
+  const float *_data = (float *)getData();
+  const float *_rdata = (float *)rhs.getData();
+  for (size_t i = 0; i < size(); ++i) {
+    /** not checking sign change is intentional to avoid float calculation
+     * errors around 0 */
+    if (std::isnan(_data[i]) || std::isnan(_rdata[i]) ||
+        std::fabs(_data[i] - _rdata[i]) > epsilon)
+      return false;
+  }
+
+  return true;
+}
+
 /// @todo support allocation by src_tensor
 void FloatTensor::allocate() {
   if (empty() || data)
--- a/nntrainer/tensor/tensor_base.h
+++ b/nntrainer/tensor/tensor_base.h
@@ -83,6 +83,20 @@ public:
     TensorBase(d, true) {}
 
   /**
+   * @brief     Comparison operator overload
+   * @param[in] rhs Tensor to be compared with
+   * @note      Only compares Tensor information
+   */
+  bool operator==(const TensorBase &rhs) const;
+
+  /**
+   * @brief     Comparison operator overload
+   * @param[in] rhs Tensor to be compared with
+   * @note      Only compares Tensor information
+   */
+  bool operator!=(const TensorBase &rhs) const { return !(*this == rhs); }
+
+  /**
    * @brief Basic Destructor
    */
   virtual ~TensorBase() {}
@@ -260,6 +274,8 @@ public:
    */
   TensorBase *getSharedDataTensor(const TensorDim dim_, size_t offset,
                                   bool reset_stride, const std::string &name_);
+
+  static constexpr float epsilon = 1e-5;
 
 protected:
   TensorDim dim;
--- a/nntrainer/tensor/tensor_v2.cpp
+++ b/nntrainer/tensor/tensor_v2.cpp
@@ -90,6 +90,27 @@ TensorV2::TensorV2(
 }
 #endif
 
+bool TensorV2::operator==(const TensorV2 &rhs) const {
+  /// compares tensor information
+  if (*itensor == *rhs.itensor) {
+    /// compares tensor data
+    if (getDataType() == Tdatatype::FP32) {
+      return *dynamic_cast<FloatTensor *>(itensor) ==
+             *dynamic_cast<FloatTensor *>(rhs.itensor);
+    } else if (getDataType() == Tdatatype::FP16) {
+#ifdef ENABLE_FP16
+      return *dynamic_cast<HalfTensor *>(itensor) ==
+             *dynamic_cast<HalfTensor *>(rhs.itensor);
+#else
+      throw std::invalid_argument(
+        "Error: HalfTensor cannot be created or used when FP16 is not enabled. "
+        "Please check if the tensor data type is set properly.");
+#endif
+    }
+  }
+  return false;
+}
+
 void TensorV2::allocate() { itensor->allocate(); }
 
 void TensorV2::deallocate() { itensor->deallocate(); }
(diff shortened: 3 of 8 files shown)


nntrainer/tensor/tensor_v2.cpp
```diff
@@ -90,6 +90,23 @@ TensorV2::TensorV2(
...
+    } else if (getDataType() == Tdatatype::FP16) {
+#ifdef ENABLE_FP16
+      return *dynamic_cast<HalfTensor *>(itensor) ==
+             *dynamic_cast<HalfTensor *>(rhs.itensor);
+#endif
+    }
+  }
+  return false;
+}
+
```
DonghakPark, line 109: How about using switch-case statement ? 
```.cpp
bool TensorV2::operator==(const TensorV2& rhs) const {  
    // Compare tensor information  
    if (*itensor == *rhs.itensor) {  
        // Compare tensor data based on data type  
        switch (getDataType()) {  
            case Tdatatype::FP32:  
                return *dynamic_cast<FloatTensor*>(itensor) ==   
                       *dynamic_cast<FloatTensor*>(rhs.itensor);  
            case Tdatatype::FP16:  
#ifdef ENABLE_FP16  
                return *dynamic_cast<HalfTensor*>(itensor) ==   
                       *dynamic_cast<HalfTensor*>(rhs.itensor);  
#else  
                break;  
#endif  
            default:  
                break;  
        }  
    }  
    return false;  
}  
```

The "Tensor == operation" can be used frequently and I believe that a switch case statement is more efficient when it comes to compiler optimization. Furthermore, I think that a switch case statement may be more appropriate semantically.

DonghakPark (approved):
It is good except for one suggestion!

nntrainer/tensor/tensor_v2.cpp
SeoHyungjun: In general cases, if statements can be faster if there are two conditions.
There isn't much difference in speed on recent computers, so wouldn't it be better to have code that is easy to read?

SeoHyungjun (approved):
LGTM!

nntrainer/tensor/tensor_base.h
```diff
@@ -242,6 +275,8 @@ class TensorBase {
   TensorBase *getSharedDataTensor(const TensorDim dim_, size_t offset,
                                   bool reset_stride, const std::string &name_);
 
+  static constexpr float epsilon = 1e-5;
```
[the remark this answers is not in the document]
djeong20: epsilon is used in the comparison operator of FloatTensor and HalfTensor to compare float values!

nntrainer/tensor/float_tensor.cpp
```diff
@@ -87,6 +87,21 @@ FloatTensor::FloatTensor(
   }
 }
 
+bool FloatTensor::operator==(const FloatTensor &rhs) const {
+  const float *_data = (float *)getData();
+  const float *_rdata = (float *)rhs.getData();
+  for (size_t i = 0; i < size(); ++i) {
+    /** not checking sign change is intentional to avoid float calculation
+     * errors around 0 */
+    if ((std::isnan(_data[i]) && !std::isnan(_rdata[i])) ||
```
jihochu, line 96: if both Nan case ??

nntrainer/tensor/tensor_v2.cpp
```diff
@@ -90,6 +90,23 @@ TensorV2::TensorV2(
...
 
+bool TensorV2::operator==(const TensorV2 &rhs) const {
+  /// compares tensor information
+  if (*itensor == *rhs.itensor) {
+    /// compares tensor data
+    if (getDataType() == Tdatatype::FP32) {
+      return *dynamic_cast<FloatTensor *>(itensor) ==
+             *dynamic_cast<FloatTensor *>(rhs.itensor);
+    } else if (getDataType() == Tdatatype::FP16) {
+#ifdef ENABLE_FP16
```
jihochu, line 101: it needs to generate exception when datatype is not supported.

nntrainer/tensor/float_tensor.cpp
djeong20: This condition is previously changed from `std::isnan(data[i]) || std::isnan(rdata[i])` in #1964. It states it is to pass if both tensor has a value.

djeong20: however, I doubt subtracting nan from nan will output zero.

nntrainer/tensor/tensor_v2.cpp
djeong20: This function will never be reached if enable-fp16 is not enabled since HalfTensor will never be created in the first place.
Therefore, I removed `throw std::invalid_argument("Error: enable-fp16 is not enabled");`
would it be better to throw an exception with a message stating this should not be reached?

Hyoungsukim (approved):
lgtm

nntrainer/tensor/tensor_v2.cpp
jihochu: exception means it should not happen.
As you mentioned, the case should not be happened, but a developer can make mistakes while maintaining.

nntrainer/tensor/float_tensor.cpp
lhs8928: @SeoHyungjun, @baek2sm, @djeong20 and I discussed offline about this problem.
For the case that comparing the 2 tensors which has a nan value as a element then we cannot say that these 2 tensors are identical even though this element is located at same position .

nntrainer/tensor/tensor_v2.cpp
djeong20: gotcha. I'll add an exception with a detailed msg. thanks 👍

nntrainer/tensor/float_tensor.cpp
jihochu: This would help for this.
https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.2.3

```
NaN is unordered, so:

The numerical comparison operators <, <=, >, and >= return false if either or both operands are NaN ([§15.20.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.20.1)).

The equality operator == returns false if either operand is NaN.

In particular, (x<y) == !(x>=y) will be false if x or y is NaN.

The inequality operator != returns true if either operand is NaN ([§15.21.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.21.1)).

In particular, x!=x is true if and only if x is NaN.
```

Hyoungsukim (approved):
lgtm

nntrainer/tensor/float_tensor.cpp
djeong20: Fixed to `std::isnan(data[i]) || std::isnan(rdata[i])`.

jijoongmoon (approved):
LGTM

PR #2392, opened 2024-01-03
[Tensor] Support additional weight initialization @open sesame 1/5 11:58

djeong20:
This PR enables additional weight initializers with a probability distribution.

- Functions to set tensor with random distribution are implemented.
- Tensor now supports various initializers besides zero and one.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -168,10 +168,44 @@ void FloatTensor::setZero() {
   }
 }
 
-/// @todo support additional initializer
+void FloatTensor::setRandNormal(float mean, float stddev) {
+  setDist<std::normal_distribution<float>>(
+    std::normal_distribution<float>(mean, stddev));
+}
+
+void FloatTensor::setRandUniform(float min, float max) {
+  setDist<std::uniform_real_distribution<float>>(
+    std::uniform_real_distribution<float>(min, max));
+}
+
+void FloatTensor::setRandBernoulli(float probability) {
+  setDist<std::bernoulli_distribution>(
+    std::bernoulli_distribution(probability));
+}
+
 void FloatTensor::initialize() {
   if (empty() || !isAllocated())
     return;
+
+  unsigned int fan_in, fan_out;
+
+  /// @fixme: when unit is equal to one, this does not work, we need to rely on
+  /// effective dimension then actual numbers here. For now, some heuristics
+  /// added to infer what would be fan_in/fan_out
+  if (dim.batch() * dim.channel() * dim.height() == 1) {
+    fan_out = fan_in = dim.width();
+  } else if (dim.batch() * dim.channel() == 1) { /// fc layer - 2-D tensor
+    fan_in = dim.height();
+    fan_out = dim.width();
+  } else { /// conv2d filters - 4d tensor, @todo extend this to > 4
+    auto field_size = dim.height() * dim.width();
+
+    // this also handles below cases.
+    // 1. fan_in = fan_out = 1 as well.
+    // 2. batch == 1, channel == 1 and height == 1, theoretical rank of 1
+    fan_in = dim.channel() * field_size;
+    fan_out = dim.batch() * field_size;
+  }
 
   switch (initializer) {
   case Initializer::ZEROS:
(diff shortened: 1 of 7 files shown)


nntrainer/tensor/half_tensor.h
```diff
@@ -116,6 +116,37 @@ class HalfTensor : public TensorBase {
...
+  /**
+   * @brief Set the Dist object
+   * @param dist distribution engine
+   */
+  template <typename Engine> void setDist(Engine dist) {
+    NNTR_THROW_IF(!contiguous, std::invalid_argument)
+      // << getName() << " Tensor is not contiguous, cannot set distribution";
+      << " Tensor is not contiguous, cannot set distribution";
+
+    _FP16 *data_ = (_FP16 *)getData();
```
skykongkong8, line 128: How are you planning to connect current TensorV2 with `blas_interface.h` ?
Would it have general function (e.g. gemv ) for `tensor_v2` or separate function (e.g. sgemv/hgemv ) for each `float_tensor`/`half_tensor`? Think I need to refactorize blas interface as well since new `getData` returns void pointer

skykongkong8 (approved):

nntrainer/tensor/float_tensor.cpp
```diff
@@ -168,18 +168,71 @@ void FloatTensor::setZero() {
   }
 }
 
-/// @todo support additional initializer
+void FloatTensor::setRandNormal(float mean, float std) {
```
DonghakPark, line 171: I understand that the variable name "std" is being used for and it was previously used in "tensor.cpp" and "tensor.h", but using the variable name "std" in C++ code does not seem like a good way.

Would it be possible to change all the existing variable names along with the new one we are writing now?
for example in tensorflow (keras) they use like below

tf.keras.initializers.RandomNormal

```
tf.keras.initializers.RandomNormal(
    mean=0.0, stddev=0.05, seed=None
)
```

or in numpy they use

```
random.normal(loc=0.0, scale=1.0, size=None)
```

We acknowledge that PyTorch uses the same name as us, but since it is based on Python, we believe it to be different.

DonghakPark (approved):

nntrainer/tensor/half_tensor.h
djeong20: both options would be valid. 
1. we could use the current form since we can manually cast it into a float of fp16 pointer. 
2. we can use different functions depending on the tensor class.

nntrainer/tensor/float_tensor.cpp
djeong20: 100% agree. will fix it

djeong20: fixed!

SeoHyungjun (approved):
looks good to me!

But now there appears to be a problem with CI.
Check it out!

SeoHyungjun:
https://github.com/nnstreamer/nnstreamer/blob/main/.TAOS-CI/.gbs.conf

It is said that if there is a problem with the values such as repo.base and repo.undefied in the above address, a similar error occurs in nnstreamer ci.

niley7464:
I'm not sure why the tiff version upgrade affects opencv
but there is a change in tiff library in [accepted/tizen/base/20240104.002457](https://review.tizen.org/gerrit/gitweb?p=platform/upstream/tiff.git;a=tag;h=refs/tags/accepted/tizen/base/20240104.002457)

I tested with libtiff version 4.4 and 4.4 version works well!

If you want to try build with libtiff 4.4 version, you can use this conf file

<details>
<summary>.gbs.conf</summary>

```
[general]
# Current profile name which should match a profile section name
profile = profile.tizen
tmpdir = /var/tmp
editor = vim
packaging_branch = tizen
workdir = .

[profile.tizen]
# Common authentication info for whole profile
# user = <WRITE_YOUR_TIZEN_ID>
# passwd = <WRITE_YOUR_TIZEN_PASSSWORD>
# obs = obs.tizen

# https://docs.tizen.org/platform/reference/gbs/gbs.conf/
# Note that a buildconf file of the last repo (e.g., /var/tmp/{USER}-gbs/tizen.of) is used by default.
repos = repo.base, repo.unified, repo.extra
buildroot = ~/GBS-ROOT-Snapshot/

[obs.tizen]
# OBS API URL pointing to a remote OBS.
url = https://api.tizen.org

[repo.base]
url = http://download.tizen.org/snapshots/TIZEN/Tizen/Tizen-Base/tizen-base_20231213.035537/repos/standard/packages/

[repo.unified]
url = http://download.tizen.org/snapshots/TIZEN/Tizen/Tizen-Unified/reference/repos/standard/packages/

[repo.extra]
url = http://download.tizen.org/live/devel:/Tizen:/6.0:/AI/Tizen_Unified_standard/

```
</details>

SeoHyungjun:
It appears that opencv requires libtiff.so!
(https://answers.opencv.org/question/35642/libtiff_40-link-errors/)

Hyoungsukim (approved):
lgtm

jijoongmoon (approved):
LGTM

PR #2395, opened 2024-01-08
[ BLAS ] Add trigonometric transformation functions

skykongkong8:
[ BLAS ] Add trigonometric transformation functions

- In order to accelerate trigonometric calculations, add such transformation functions in BLAS module with NEON SIMD
- Add zlib license file

[ Tensor ] Add trigonometric transformation functions in Tensor
    
- Add sin / cos transform function in Tensor : both for BLAS/raw
- Add unittest accordingly

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -838,4 +838,30 @@ unsigned int isamax(const unsigned int N, const float *X, const int incX) {
 #endif
 }
 
+void sine_transformation(const unsigned int N, float *X, float *Y,
+                         float alpha) {
+#ifdef USE_NEON
+  nntrainer::neon::sine_transformation_neon(N, X, Y, alpha);
+#else
+  unsigned int i = 0;
+  while (i < N) {
+    Y[i] = std::sin(alpha * X[i]);
+    ++i;
+  }
+#endif
+}
+
+void cosine_transformation(const unsigned int N, float *X, float *Y,
+                           float alpha) {
+#ifdef USE_NEON
+  nntrainer::neon::cosine_transformation_neon(N, X, Y, alpha);
+#else
+  unsigned int i = 0;
+  while (i < N) {
+    Y[i] = std::cos(alpha * X[i]);
+    ++i;
+  }
+#endif
+}
+
 } // namespace nntrainer
--- a/nntrainer/tensor/neon_mathfunc.h
+++ b/nntrainer/tensor/neon_mathfunc.h
@@ -1,0 +1,376 @@
+/**
+ * @file   neon_mathfunc.h
+ * @date   08 Jan 2024
+ * @brief  This is collection of sin, cos, exp, log function with NEON SIMD
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Julien Pommier
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+/** NEON implementation of sin, cos, exp and log
+
+   Inspired by Intel Approximate Math library, and based on the
+   
+   corresponding algorithms of the cephes math library
+*/
+
+/** Copyright (C) 2011  Julien Pommier
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+
+  (this is the zlib license)
+*/
+
+#ifndef __NEON_MATHFUNC_H_
+#define __NEON_MATHFUNC_H_
+#ifdef __cplusplus
+
+#include <arm_neon.h>
+
+typedef float32x4_t v4sf; // vector of 4 float
+typedef uint32x4_t v4su;  // vector of 4 uint32
+typedef int32x4_t v4si;   // vector of 4 uint32
+
+/**
+ * @def		c_inv_mant_mask extract the mantissa of a float by performing a
+ * bitwise negation on the binary representation of 0x7f800000.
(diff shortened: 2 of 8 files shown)


nntrainer/tensor/neon_mathfunc.h
```diff
@@ -0,0 +1,376 @@
...
+ * @param[in] x register variable (float32x4_t)
+ */
+v4sf cos_ps(v4sf x) {
+  v4sf ysin, ycos;
+  sincos_ps(x, &ysin, &ycos);
+  return ycos;
+}
+
+#endif
+#endif
```
jijoongmoon, line 376: please add new line.

skykongkong8: Fixed

nntrainer/tensor/tensor.cpp
```diff
@@ -3418,6 +3418,20 @@ Tensor &Tensor::erf(Tensor &out) const {
...
+    sine_transformation(this->size(), getData<float>(), out.getData<float>());
+  } else
+    throw std::invalid_argument("Error: sin_transform supports fp32 case only");
+}
+
+void Tensor::cos_transform(Tensor &out) {
+  if (getDataType() == ml::train::TensorDim::DataType::FP32) {
+    cosine_transformation(this->size(), getData<float>(), out.getData<float>());
+  } else
+    throw std::invalid_argument("Error: cos_transform supports fp32 case only");
```
jijoongmoon, line 3432: is it because of the accuracy or NYI?

jijoongmoon (approved):
LGTM!

nntrainer/tensor/tensor.cpp
```diff
@@ -3418,6 +3418,20 @@ Tensor &Tensor::erf(Tensor &out) const {
   return out;
 }
 
+void Tensor::sin_transform(Tensor &out) {
+  if (getDataType() == ml::train::TensorDim::DataType::FP32) {
+    sine_transformation(this->size(), getData<float>(), out.getData<float>());
```
djeong20, line 3423:
```suggestion
void Tensor::sin_transform(Tensor &out, float alpha) {
  if (getDataType() == ml::train::TensorDim::DataType::FP32) {
    sine_transformation(this->size(), getData<float>(), out.getData<float>(), alpha);
```

having an extra parameter `float alpha` seems reasonable since currently alpha is fixed to 1.0. (or will alpha always be 1.0?)

nntrainer/tensor/tensor.h
```diff
@@ -969,6 +969,20 @@ class Tensor {
    */
   Tensor &erf(Tensor &out) const;
 
+  /**
+   * @brief    sin transform function
+   * @param[out] out out to store the result
+   * @retval Calculated Tensor
```
djeong20, line 975:
```suggestion
```

think this should be removed since it's a void function

nntrainer/tensor/tensor.h
```diff
@@ -969,6 +969,20 @@ class Tensor {
...
+   */
+  void sin_transform(Tensor &out);
+
+  /**
+   * @brief    cos transform function
+   * @param[out] out out to store the result
+   * @retval Calculated Tensor
```
djeong20, line 982:
```suggestion
```
same here

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -4460,6 +4460,69 @@ TEST(nntrainer_Tensor, dequantize_05_n) {
...
+                      << result_sine.getValue(b, c, h, w) << std::endl;
+          if (cos_err > eps)
+            std::cout << cos_output.getValue(b, c, h, w) << " VS "
+                      << result_cosine.getValue(b, c, h, w) << std::endl;
+        }
+      }
+    }
+  }
+
+  EXPECT_EQ(flag, true);
```
djeong20, line 4523:
```suggestion
  EXPECT_EQ(sin_output, result_sine);
  EXPECT_EQ(cos_output, result_cosine);
```

quick question, any reason for manually checking each value instead of comparing two tensors?

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -4460,6 +4460,69 @@ TEST(nntrainer_Tensor, dequantize_05_n) {
   EXPECT_THROW({ input.dequantize(output, 1); }, std::invalid_argument);
 }
 
+TEST(nntrainer_neon_experimental, trigonometric_simd) {
```
djeong20, line 4463: just an idea. think it would be better to split this into two separate tests (cos and sin tests).

nntrainer/tensor/tensor.cpp
skykongkong8: alpha in trigonometric calculation is only used for rotary embedding (RoPE)
In other cases, it will be zero

nntrainer/tensor/tensor.h
skykongkong8: I missed it thanks!

test/unittest/unittest_nntrainer_tensor.cpp
skykongkong8: Since we are relying on `neon_mathfunc.h` for trigonometric computation, we need an epsilon value to assess

test/unittest/unittest_nntrainer_tensor.cpp
skykongkong8: Sounds more reasonable!
Fixed.

nntrainer/tensor/tensor.cpp
skykongkong8: But in a second thought, there is no reason for not having it as well.
I applied the change

nntrainer/tensor/blas_interface.cpp
```diff
@@ -838,4 +838,30 @@ unsigned int isamax(const unsigned int N, const float *X, const int incX) {
 #endif
 }
 
+void sine_transformation(const unsigned int N, float *X, float *Y,
```
SeoHyungjun, line 841: sine_transformation and cosine_transformation look very similar.
How about creating one conversion function and allowing it to be separated by input variables?
Wouldn't it be easy to support other transformations, such as tangent, later?

SeoHyungjun (approved):

baek2sm (approved):
LGTM

nntrainer/tensor/blas_interface.cpp
skykongkong8: Wouldn't it be more complex to add more hard-coded parameters for that?

djeong20 (approved):
LGTM :)

PR #2396, opened 2024-01-08
[ BLAS ] add inv squared root transformation inplace function for fp16

skykongkong8:
[ BLAS ] add inv squared root transormation inplace function for fp16
- add inv squared root transormation inplace function for fp16
- expect to accelerate rms normalization

### DIFF
--- a/nntrainer/tensor/blas_interface.h
+++ b/nntrainer/tensor/blas_interface.h
@@ -175,6 +175,14 @@ void ewva(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z);
  * @param[in] X __fp16 * for Vector X
  */
 unsigned int isamax(const unsigned int N, const _FP16 *X, const int incX);
+
+/**
+ * @brief squared root transformation inplace : X = sqrt(X)
+ *
+ * @param N size of X
+ * @param X __fp16 * for Vector X
+ */
+void inv_sqrt_inplace(const unsigned int N, _FP16 *X);
 #endif
 /**
  * @brief     sscal computation : X = alpha * X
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/blas_neon.cpp
```diff
@@ -2042,5 +2042,20 @@ void elementwise_vector_addition_neon_fp16(const unsigned int N,
...
+  unsigned int i = 0;
+  for (; N - i >= 8; i += 8) {
+    float16x8_t x0_7 = vld1q_f16(&X[i]);
+    float16x8_t x0_7_sqrt = vsqrtq_f16(x0_7);
+    float16x8_t ones = vmovq_n_f16(1);
+    float16x8_t x0_7_sqrt_div = vdivq_f16(ones, x0_7_sqrt);
+    vst1q_f16(&X[i], x0_7_sqrt_div);
+  }
+  while (i < N) {
+    X[i] = (1 / std::sqrt(static_cast<float>(X[i])));
```
jijoongmoon, line 2055: I just wonder if there is no accuracy loss when we use sqrt with fp16? is it ok because we use vsqrtq_f16?

skykongkong8: As far as I am concerned, we do not need to apply so-called 'enhanced-fp16' (using inter-fp32 values rather than full-fp16) for all fp16 functions. We have to optimize b/w time-accuracy only when it comes to "dimension shrinking" or some specific cases proven by a lot of papers ( BN, LN, sofrmax, ... )

I could observe the same model output with rms_norm layer with this function at glance, so I thought we could rely on it.
However indeed we should track on it becase it is a 'normalization' layer though.

Hyoungsukim (approved):
lgtm

DonghakPark (approved):
LGTM!

nntrainer/tensor/blas_interface.h
```diff
@@ -175,6 +175,14 @@ void ewva(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z);
  * @param[in] X __fp16 * for Vector X
  */
 unsigned int isamax(const unsigned int N, const _FP16 *X, const int incX);
+
+/**
+ * @brief squared root transformation inplace : X = sqrt(X)
```
djeong20, line 180: I'm confused. is this func $X = \sqrt{X}$ or $X = \frac{1}{\sqrt{X}}$?

skykongkong8: the latter.
I just found that I named it wrong :(

PR #2397, opened 2024-01-08
[ layer ] SIMD function for Rotary embedding incremental forwarding

skykongkong8:
[ util ] Add util_simd file
    
- Here I would like to introduce util_simd. I had no choice but to differentiate this file with blas for following reasons:
      1. This is actually not 'Basic Linear Algebra Subprogram' indeed
      2. This code is only used in very specific occasion.

[ layer ] Apply neon simd acceleration in rotary embedding computation
    
- Previous rotary embedding computation code was naively implemented with for-loop
- Using simd code, I expect this code to behave efficiently in time cost perspective, without precision loss
- Current implementation only supports Neon simd (ARMv8)

    **Self evaluation:**
    
### DIFF
--- a/nntrainer/layers/multi_head_attention_layer.h
+++ b/nntrainer/layers/multi_head_attention_layer.h
@@ -19,6 +19,7 @@
 #include <acti_func.h>
 #include <complex>
 #include <layer_impl.h>
+#include <util_simd.h>
 #include <utility>
 
 namespace nntrainer {
@@ -163,6 +164,10 @@ private:
       sin->assign(seq_len, std::vector<float>(dim, 0));
 
       for (unsigned int i = 0; i < seq_len; ++i) {
+#ifdef USE_NEON
+        calc_trigonometric_vals_dup(half_, freqs.data(), (*cos)[i].data(),
+                                    (*sin)[i].data(), i);
+#else
         for (unsigned int j = 0; j < half_; ++j) {
           float angle = i * freqs[j];
           (*cos)[i][j] = std::cos(angle);
@@ -171,8 +176,8 @@ private:
           (*sin)[i][j] = std::sin(angle);
           (*sin)[i][j + half_] = std::sin(angle); // repeated 2 times
         }
+#endif
       }
-
       freqs_cos = cos;
       freqs_sin = sin;
     }
@@ -197,10 +202,12 @@ private:
     std::vector<float> *sin_;
 
     if (from >= max_timestep) {
-      std::cout << from << " " << max_timestep << std::endl;
       cos_ = new std::vector<float>(dim);
       sin_ = new std::vector<float>(dim);
-
+#ifdef USE_NEON
+      calc_trigonometric_vals_dup(half_, freqs.data(), cos_->data(),
+                                  sin_->data(), from);
+#else
       for (unsigned int i = 0; i < half_; ++i) {
         float angle = from * freqs[i];
         (*cos_)[i] = std::cos(angle);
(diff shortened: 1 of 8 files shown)


nntrainer/tensor/blas_neon.cpp
```diff
@@ -2057,5 +2079,80 @@ void squared_root_transform_inplace_neon(const unsigned int N, __fp16 *X) {
   }
 }
 
+void compute_rotary_embedding_value_neon(unsigned int dim, unsigned int half_,
+                                         unsigned int w, __fp16 *in,
```
jijoongmoon, line 2083: It is really great to have this fuc with SIMD! Great work! One thing I would recommend is how about moving this function to util directory for management consistency.

jijoongmoon (approved):
LGTM except for minor.

nntrainer/tensor/blas_neon.cpp
skykongkong8: got it

nntrainer/layers/multi_head_attention_layer.h
```diff
@@ -161,7 +162,12 @@ class MultiHeadAttentionLayer : public LayerImpl {
...
       auto sin = new std::vector<std::vector<float>>();
       sin->assign(seq_len, std::vector<float>(dim, 0));
-
+#ifdef USE_NEON
+      for (unsigned int i = 0; i < seq_len; ++i) {
+        calc_trigonometric_vals_dup(half_, freqs.data(), (*cos)[i].data(),
+                                    (*sin)[i].data(), i);
+      }
+#else
       for (unsigned int i = 0; i < seq_len; ++i) {
```
djeong20, line 171:
```suggestion
      for (unsigned int i = 0; i < seq_len; ++i) {
#ifdef USE_NEON
        calc_trigonometric_vals_dup(half_, freqs.data(), (*cos)[i].data(),
                                    (*sin)[i].data(), i);
#else
```

this could reduce code

djeong20 (changes requested):
LGTM except for CI fails.

Please check `nntrainer.spec` and `.install` files to include newly added header files.

skykongkong8:
> LGTM except for CI fails.
> 
> Please check `nntrainer.spec` and `.install` files to include newly added header files.

Guess it must be `nntrainer-dev.install`

nntrainer/layers/multi_head_attention_layer.h
skykongkong8: Nice catch!
Will refactor immediately

djeong20 (approved):

myungjoo (commented):
Please minimize the usage of #if/#endif in .cpp/.c files; make them "unaware" of whether NEON is there or not.

E.g.,
```C
/** OOO.cpp */
if (simd_enabled())
   do_simd_func_X();
else
   do_some_general_func_X();

/** XXX.h */
#ifdef SIMD /* enable this at every ARM build. decide whether to use SIMD or not at runtime (with "if", not "#if"). */
void do_simd_func_X(); /** implementation in XXX.cpp */
#else
static void do_simd_func_X()
{ throw exception; }
#endif
```

skykongkong8:
> Please minimize the usage of #if/#endif in .cpp/.c files; make them "unaware" of whether NEON is there or not.
> 
> E.g.,
> 
> ```c
> /** OOO.cpp */
> if (simd_enabled())
>    do_simd_func_X();
> else
>    do_some_general_func_X();
> 
> /** XXX.h */
> #ifdef SIMD /* enable this at every ARM build. decide whether to use SIMD or not at runtime (with "if", not "#if"). */
> void do_simd_func_X(); /** implementation in XXX.cpp */
> #else
> static void do_simd_func_X()
> { throw exception; }
> #endif
> ```
> 
> Anyway, you can do this with next PRs, not in this PR. This PR itself is good. But try to revise the code (.cpp files) with too many #ifdefs. You are going to suffer from your own #if/#endif code soon if you don't fix them.

Totally agree.
Will try to refactorize asap

PR #2400, opened 2024-01-09
[ unittest ] Add unittest for inv_sqrt_i with fp16

skykongkong8:
- There was a request to add unittest for inv_sqrt_i in #2396 
- Conducted by comparing values between fp16 and fp32 Tensor with eps = 1e-3

**Self evaluation:**

### DIFF
--- a/test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
+++ b/test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
@@ -423,6 +423,60 @@ TEST(nntrainer_Tensor, dot_sgemm) {
   EXPECT_IN_RANGE((float)cosSimNeon, 0.99, 1);
 }
 
+TEST(nntrainer_Tensor, inv_sqrt_i_p) {
+  int batch = 4;
+  int channel = 10;
+  int height = 10;
+  int width = 10;
+
+  const int MOD = 10;
+  const float eps = 1e-3;
+
+  nntrainer::Tensor input(batch, channel, height, width,
+                          nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP16);
+  GEN_TEST_INPUT(input, (i * (batch * height) + j * (width) + k) % MOD + 1);
+
+  nntrainer::Tensor ground_truth(batch, channel, height, width,
+                                 nntrainer::Tformat::NCHW,
+                                 nntrainer::Tdatatype::FP32);
+  GEN_TEST_INPUT(ground_truth,
+                 (i * (batch * height) + j * (width) + k) % MOD + 1);
+
+  input.inv_sqrt_i();
+
+  for (int b = 0; b < batch; b++) {
+    for (int c = 0; c < channel; c++) {
+      for (int h = 0; h < height; h++) {
+        for (int w = 0; w < width; w++) {
+          ground_truth.setValue(
+            b, c, h, w, 1 / std::sqrt(ground_truth.getValue(b, c, h, w)));
+        }
+      }
+    }
+  }
+
+  bool flag = true;
+
+  for (int b = 0; b < batch; b++) {
+    for (int c = 0; c < channel; c++) {
+      for (int h = 0; h < height; h++) {
+        for (int w = 0; w < width; w++) {
+          double err = std::abs(input.getValue<__fp16>(b, c, h, w) -
+                                ground_truth.getValue(b, c, h, w));
+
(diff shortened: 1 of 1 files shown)


test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
```diff
@@ -423,6 +423,60 @@ TEST(nntrainer_Tensor, dot_sgemm) {
...
 }
 
+TEST(nntrainer_Tensor, inv_sqrt_i_p) {
+  int batch = 4;
+  int channel = 10;
+  int height = 10;
+  int width = 10;
+
+  const int MOD = 10;
+  const float eps = 1e-3;
```
djeong20, line 433: In tensor, epsilon is set to 1e-5 to compare values of two tensors. any reason why eps is 1e-3? 
along with that, I see epsilon values are inconsistent in neon_fp16 tests. are there any rules in determining the epsilon for comparison?

djeong20: maybe we can utilize things from here.
[How do you compare float and double while accounting for precision loss?](https://stackoverflow.com/questions/17333/how-do-you-compare-float-and-double-while-accounting-for-precision-loss)

skykongkong8: In half-precision, I could not find a strict eps value for accuracy validation.
There are many reasons:

1. As far as I am concerned, our current eps=1e-5 for float Tensor is grounded on decimal digit of IEEE float32 definition. I am currently using 1e-3 as a basic eps value for fp16 with similar reason.
2. However, I could observe some inconsistent accuracy with: large-scale Tensor, spare Tensor, Tensor with near-zero values, Tensor with really precise values (values that exceeds precision of its dataType )
3. Because of 2, I have been using inter-fp32 value to preserve 'sufficient' amount of accuracy(major reason of inconsistent eps), but this costs a little bit of latency. Thus, we should determine how accurate we want to be along with how fast we want to be...

skykongkong8: Although I agree with using consistent epsilon value, when it comes to half-precision, I am afraid I could not avoid using current style. I will definitely keep up with this issue.

To improve reliability, in case of more complex computation, I am using cos_sim and mse

skykongkong8: And one thing more:

In terms of this TC, I already conducted test case with llama

djeong20: Okay, so I did some research about this, and in comparing two values with the same precision, we should make the epsilon stick with the value of the positive difference between 1.0 and the next greater representable number in the corresponding type (i.e. the ulp of one). This could certainly verify the two floating point numbers are the same.

In the case of float32, this is already defined in the standard library with `std::numeric_limits<float>::epsilon()` or `FLT_EPSILON` which has a value of `0x1.0p-23f`, which is approximately 1.19209e-07.

If we apply the same methodology for half-precision, this value would be `0x1.0p-10f` (since 10 bits are used as a fraction) and this is approximately 0.000976562. 

If the value difference is greater than the difference with the next greater representable number, it would be difficult to say two values are the same.

However, like this unit test, in comparing fp32 and fp16 values, we can stick to using cos_sim and mse.

skykongkong8: Great research! I really appreciate it!

1. Glad to hear 0.000976562. is almost near 1e-3
2. And yes, in many (almost always) cases, we are comparing values with different precision, because we had a huge fail with model output when conducted with comparing the same precision. So maybe I could just stick to the current way, but I will definitely bear in mind the research you have done for me :)

skykongkong8: Interesting experience to share:

I could observe different model output even with cos_sim = 1.0, eps = 1e-3, near-zero mse GEMM/GEMV, when I revised the amount of using inter-fp32 values in BLAS level.  ( it did not have any effect on the metric that we are currently using, but effected the model output )
So I am always looking for a new metric to rely on... Before finding it, I guess I should use current metric

djeong20 (approved):

SeoHyungjun (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2403, opened 2024-01-10
Remove dependencies on FP16 support status.

myungjoo:
Most modules should NOT depend on the existnace of FP16 support. Let a single (or minimal) entity (a single class?) know and handle it.

This is a suggestion commit that shows an example how to refactor it. Most modules are not yet touched with this commit.

CC: @skykongkong8 (I've seen some #if/#endif in your commits.)

### DIFF
--- a/Applications/YOLO/jni/yolo_v2_loss.cpp
+++ b/Applications/YOLO/jni/yolo_v2_loss.cpp
@@ -178,11 +178,7 @@ calc_iou(nntrainer::Tensor &bbox1_x1, nntrainer::Tensor &bbox1_y1,
   if (type_intersection_width == ml::train::TensorDim::DataType::FP32) {
     intersection_width.apply_i<float>(nntrainer::ActiFunc::relu<float>);
   } else if (type_intersection_width == ml::train::TensorDim::DataType::FP16) {
-#ifdef ENABLE_FP16
     intersection_width.apply_i<_FP16>(nntrainer::ActiFunc::relu<_FP16>);
-#else
-    throw std::runtime_error("Not supported data type");
-#endif
   }
 
   intersection_y2.subtract(intersection_y1, intersection_height);
@@ -191,11 +187,7 @@ calc_iou(nntrainer::Tensor &bbox1_x1, nntrainer::Tensor &bbox1_y1,
   if (type_intersection_height == ml::train::TensorDim::DataType::FP32) {
     intersection_height.apply_i<float>(nntrainer::ActiFunc::relu<float>);
   } else if (type_intersection_height == ml::train::TensorDim::DataType::FP16) {
-#ifdef ENABLE_FP16
     intersection_height.apply_i<_FP16>(nntrainer::ActiFunc::relu<_FP16>);
-#else
-    throw std::runtime_error("Not supported data type");
-#endif
   }
 
   nntrainer::Tensor intersection =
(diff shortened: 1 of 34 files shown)


Applications/YOLO/jni/yolo_v2_loss.cpp
```diff
@@ -178,11 +178,9 @@ calc_iou(nntrainer::Tensor &bbox1_x1, nntrainer::Tensor &bbox1_y1,
   if (type_intersection_width == ml::train::TensorDim::DataType::FP32) {
     intersection_width.apply_i<float>(nntrainer::ActiFunc::relu<float>);
   } else if (type_intersection_width == ml::train::TensorDim::DataType::FP16) {
-#ifdef ENABLE_FP16
+    if (!is_fp16_enabled())
+      throw std::runtime_error("Not supported data type: fp16");
```
myungjoo, line 182: or maybe we can remove these checks and let it be cheked at apply_i() or its underlying function.

SeoHyungjun (approved):
LGTM!

api/ccapi/include/tensor_dim.h
```diff
@@ -21,15 +22,52 @@
...
+#include <cstdint>
 
 #ifdef ENABLE_FP16
 #ifdef USE__FP16
 #define _FP16 __fp16
 #else
 #define _FP16 _Float16
 #endif
+#else /* !ENABLE_FP16 */
+#define _FP16 uint16_t /* Keep the FP16 programming interface, but don't allow using it in run-time */
```
djeong20, line 34: after this fix, I think we'll need to add test cases to certify every part is well covered.
In the case of tensors, such as follows.

```
Tensor fp32_tensor(..., Tdatatype::FP32);
EXPECT_THROW({ fp32_tensor.getValue<FP16>(1); }, std::invalid_argument);
EXPECT_THROW(
  {
    fp32_tensor.apply(
      (std::function<_FP16(_FP16)>)[&](_FP16 in) { return idx++ % 2; });
  },
  std::invalid_argument);
```

previously, there was no need to check because `_FP16` cannot be used when fp16 is not enabled.
if everyone agrees, I'll work on adding test cases at the tensor level.

Applications/YOLO/jni/yolo_v2_loss.cpp
```diff
@@ -178,11 +178,8 @@ calc_iou(nntrainer::Tensor &bbox1_x1, nntrainer::Tensor &bbox1_y1,
   if (type_intersection_width == ml::train::TensorDim::DataType::FP32) {
     intersection_width.apply_i<float>(nntrainer::ActiFunc::relu<float>);
   } else if (type_intersection_width == ml::train::TensorDim::DataType::FP16) {
-#ifdef ENABLE_FP16
+    THROW_UNLESS_FP16_ENABLED;
```
myungjoo, line 181: I'll try to remove these lines by ensuring to throw in apply_i<_FP16>

myungjoo: Done

jihochu (approved):
LGTM :+1:

jijoongmoon:
We can consider this PR proposed for better readability and maintainability. We are in the middle of refactoring Tensor, BLAS, and Multi-Type Support and the codes will be changed a lot.

PR #2405, opened 2024-01-11
meson: do not force enable ml-api when it's not force enabled.

myungjoo:
The previous meson logic force enabled ml-api if it is not disabled and common headers are found.

The new logic disables ml-api even if common headers are found if ml-inference is not found.

This allows to build nntrainer in a system only common ml headers are available without any meson options.


jihochu (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2406, opened 2024-01-12
[TensorV2] Multiplication operation skeleton

djeong20:
This pull request adds a basic implementation of tensor multiplication operations to our codebase. 
The new functionality allows users to perform the multiplication of tensors by simply calling a function.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.h
+++ b/nntrainer/tensor/float_tensor.h
@@ -179,6 +179,30 @@ public:
                   TensorV2 &output) const override;
 
   /**
+   * @copydoc TensorV2::multiply_i(float const &value)
+   * @todo    Need implementation and unit tests
+   */
+  int multiply_i(float const &value) override { return 0; }
+
+  /**
+   * @copydoc TensorV2::multiply(float const &value, TensorV2 &out)
+   * @todo    Need implementation and unit tests
+   */
+  TensorV2 &multiply(float const &value, TensorV2 &out) const override {
+    return out;
+  }
+
+  /**
+   * @copydoc TensorV2::multiply(TensorV2 const &m, TensorV2 &output, const
+   * float beta = 0.0)
+   * @todo    Need implementation and unit tests
+   */
+  TensorV2 &multiply(TensorV2 const &m, TensorV2 &output,
+                     const float beta = 0.0) const override {
+    return output;
+  }
+
+  /**
    * @copydoc TensorV2::print(std::ostream &out)
    */
   void print(std::ostream &out) const override;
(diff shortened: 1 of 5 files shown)


skykongkong8 (approved):
LGTM!

nntrainer/tensor/tensor_v2.cpp
```diff
@@ -151,6 +151,31 @@ TensorV2 &TensorV2::apply(std::function<TensorV2 &(TensorV2, TensorV2 &)> f,
...
+TensorV2 &TensorV2::multiply(float const &value, TensorV2 &out) const {
+  return out;
+}
+
+int TensorV2::multiply_i(TensorV2 const &m, const float beta) {
+  return ML_ERROR_NONE;
+}
+
+TensorV2 TensorV2::multiply(TensorV2 const &m, const float beta) const {
+  TensorV2 t("", this->getFormat());
```
SeoHyungjun, line 170: Looking at Tensor's Basic Constructor, the default value of name_ is "".
Isn't it possible to omit it?

https://github.com/nnstreamer/nntrainer/blob/477a8aa065b0bd7bb885c87ac752ab508497a870/nntrainer/tensor/tensor_v2.h#L38

SeoHyungjun (approved):

nntrainer/tensor/tensor_v2.cpp
djeong20: discussed it in person. resolved.

nntrainer/tensor/float_tensor.h
```diff
@@ -178,6 +178,30 @@ class FloatTensor : public TensorBase {
...
+   * @todo    Need implementation and unit tests
+   */
+  int multiply_i(float const &value) override { return 0; }
+
+  /**
+   * @copydoc TensorV2::multiply(float const &value, TensorV2 &out)
+   * @todo    Need implementation and unit tests
+   */
+  TensorV2 &multiply(float const &value, TensorV2 &out) const override {
+    return out;
```
baek2sm, line 192: What if we add "throw std::logic_error("Not yet implemented");" in case implementation is not done yet? Not in this PR, but perhaps in a future PR.

baek2sm (approved):
LGTM

nntrainer/tensor/float_tensor.h
djeong20: good idea! I'll add it in later PRs with skeleton codes :)

jijoongmoon:
One quick comment, please keep in mind we do have support lazy tensor as well.

jijoongmoon (approved):
LGTM

PR #2402, opened 2024-01-10
[ Util ] Swish calculation implementation

skykongkong8:
[ NEON ] Refactor neon_mathfun

- For easier usage of neon_mathfun, refactor to avoid duplicated symbol error

[ util ] Implement swish function in util

- To accelerate swish activation function implement swish calculation function for:
  - neon fp32 / fp16
  - raw fp32 / fp16
- SIMD calculation of exponential function is grounded on neon_mathfun

**Self evaluation:**
1. Build test:     [X]Passed [ ]Failed [ ]Skipped
2. Run test:     [X]Passed [ ]Failed [ ]Skipped
    
 Signed-off-by: skykongkong8 <ss.kong@samsung.com>

PR #2410, opened 2024-01-16
[ Trivial ] Refactor trigonometric functions & inv_sqrt_i

skykongkong8:
[ Trivial ] Refactor trigonometric functions

- In case of non-contiguous Tensor, it is impossible to apply SIMD instructions. Add expection accordingly.
- Rename the function name for intuitiveness :
  -  sin_transform-> sin
  - cos_transform -> cos

[ Trivial ] Add exception in inv_sqrt_i function
    
- In case of non-contiguous Tensor, it is impossible to apply SIMD instructions. Add expection accordingly.

 [ Tensor ] Support non-contiguous case in sin, cos, inv_sqrt_i
    
- If it is not for BLAS, we can also support sin, cos, inv_sqrt_i functions for non-contiguous case as well.
- Fix related functions and add unittest accordingly.

Resolves: non-contiguous Tensor case of function usage raised from #2402 and #2410 

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -847,10 +847,10 @@ unsigned int isamax(const unsigned int N, const float *X, const int incX) {
 #endif
 }
 
-void sine_transformation(const unsigned int N, float *X, float *Y,
+void sine(const unsigned int N, float *X, float *Y,
                          float alpha) {
 #ifdef USE_NEON
-  nntrainer::neon::sine_transformation_neon(N, X, Y, alpha);
+  nntrainer::neon::sine_neon(N, X, Y, alpha);
 #else
   unsigned int i = 0;
   while (i < N) {
@@ -860,10 +860,10 @@ void sine_transformation(const unsigned int N, float *X, float *Y,
 #endif
 }
 
-void cosine_transformation(const unsigned int N, float *X, float *Y,
+void cosine(const unsigned int N, float *X, float *Y,
                            float alpha) {
 #ifdef USE_NEON
-  nntrainer::neon::cosine_transformation_neon(N, X, Y, alpha);
+  nntrainer::neon::cosine_neon(N, X, Y, alpha);
 #else
   unsigned int i = 0;
   while (i < N) {
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -127,8 +127,7 @@ public:
   SrcSharedTensor() : src(nullptr), off(0) {}
 
   SrcSharedTensor(const Tensor *tensor, size_t offset) :
-    src(tensor),
-    off(offset) {}
+    src(tensor), off(offset) {}
 
   /**
    * @brief   Get the allocated src tensor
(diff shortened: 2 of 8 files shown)


lhs8928 (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -3440,18 +3439,20 @@ Tensor &Tensor::erf(Tensor &out) const {
   return out;
 }
 
-void Tensor::sin_transform(Tensor &out, float alpha) {
+void Tensor::sin(Tensor &out, float alpha) {
+  NNTR_THROW_IF(!contiguous, std::invalid_argument)
+    << getName() << " is not contiguous, cannot calculate sin.";
```
djeong20, line 3445: shouldn't it also check for if the tensor size is equal? 
```
if (size() != out.size())
 throw std::invalid_argument("Size of tensor must match");
```

djeong20: or maybe we can compare TensorDim

skykongkong8: Makes sense

DonghakPark (approved):
LGTM!

nntrainer/tensor/blas_interface.cpp
```diff
@@ -847,10 +847,10 @@ unsigned int isamax(const unsigned int N, const float *X, const int incX) {
 #endif
 }
 
-void sine_transformation(const unsigned int N, float *X, float *Y,
+void sine(const unsigned int N, float *X, float *Y,
```
SeoHyungjun, line 850: Below, sin and cos are used.
However, sine and cosine are used here.
Is it impossible to unify as one?

nntrainer/tensor/blas_interface.cpp
```diff
@@ -847,10 +847,10 @@ unsigned int isamax(const unsigned int N, const float *X, const int incX) {
...
                          float alpha) {
 #ifdef USE_NEON
-  nntrainer::neon::sine_transformation_neon(N, X, Y, alpha);
+  nntrainer::neon::sine_neon(N, X, Y, alpha);
```
SeoHyungjun, line 853: If you look at the namespace, neon already exists.
Can't I just use sine or sin?

SeoHyungjun (approved):

nntrainer/tensor/blas_interface.cpp
skykongkong8: With this logic, every function in `blas_neon.cpp` must be revised.
Will do it in the near future.

jihochu (approved):
LGTM

djeong20 (approved):

jijoongmoon:
I'm not quite sure why we cannot compute the sin value if it is not continuous. We can compute sin element by element.

skykongkong8:
> I'm not quite sure why we cannot compute the sin value if it is not continuous. We can compute sin element by element. If we do need to use simd, it makes sense, but it is not, then we can just add element by element computation instead of generating error.

Think I can resolve by doing something like:

```C++
 if (getDataType() == ml::train::TensorDim::DataType::FP32) {
    if (!contiguous) {
      auto f = [alpha](float val) -> float { return std::cos(alpha * val); };
      apply<float>(f, out);
    } else {
      cosine(size(), getData<float>(), out.getData<float>(), alpha);
    }
  }
```

nntrainer/tensor/tensor.cpp
```diff
@@ -3440,33 +3440,56 @@ Tensor &Tensor::erf(Tensor &out) const {
...
-                        alpha);
+    if (!contiguous) {
+      auto f = [alpha](float val) -> float { return std::sin(alpha * val); };
+      apply<float>(f, out);
+    } else {
+      sine(size(), getData<float>(), out.getData<float>(), alpha);
+    }
   } else
-    throw std::invalid_argument("Error: sin_transform supports fp32 case only");
+    throw std::invalid_argument("Error: Tensor::sin supports fp32 case only.");
```
jihochu, line 3454: how about using NNTR_THROW_IF ?

PR #2413, opened 2024-01-17
util_simd: make typename consistent (__fp16 --> _FP16)

myungjoo:
_FP16 is the macro to unify different fp16 typenames across different architectures or libraries.

Note that util_simd.cpp has the correct name (_FP16) while the header has the incorrect naming (__fp16).

Although this does not break the build or execution, this is not good for readability and dependency clean-ups.

CC: @skykongkong8

### DIFF
--- a/nntrainer/utils/util_simd.h
+++ b/nntrainer/utils/util_simd.h
@@ -60,11 +60,11 @@ void compute_rotary_embedding_value(unsigned int dim, unsigned int half_,
  * @brief swish function : X = (Y / (1 + exp( -Y ))) * Z
  *
  * @param N number of elements in X
- * @param X __fp16 * for Vector X
- * @param Y __fp16 * for Vector Y
- * @param Z __fp16 * for Vector Z
+ * @param X _FP16 * for Vector X
+ * @param Y _FP16 * for Vector Y
+ * @param Z _FP16 * for Vector Z
  */
-void swish(const unsigned int N, __fp16 *X, __fp16 *Y, __fp16 *Z);
+void swish(const unsigned int N, _FP16 *X, _FP16 *Y, _FP16 *Z);
 #endif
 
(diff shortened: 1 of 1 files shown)


skykongkong8 (approved):
My apologies.
I definitely missed this point.

nntrainer/utils/util_simd.h
```diff
@@ -60,11 +60,11 @@ void compute_rotary_embedding_value(unsigned int dim, unsigned int half_,
...
  * @param N number of elements in X
- * @param X __fp16 * for Vector X
- * @param Y __fp16 * for Vector Y
- * @param Z __fp16 * for Vector Z
+ * @param X _FP16 * for Vector X
+ * @param Y _FP16 * for Vector Y
+ * @param Z _FP16 * for Vector Z
  */
-void swish(const unsigned int N, __fp16 *X, __fp16 *Y, __fp16 *Z);
+void swish(const unsigned int N, _FP16 *X, _FP16 *Y, _FP16 *Z);
```
jijoongmoon, line 67: @skykongkong8 , I'm not sure How the previous version of this code section compiled properly for the Debian build. Maybe the default of enable_fp16 is false for Debian. 
We need unittest about the swish for each data type ( fp32 and fp16 ) in the next pr.

jijoongmoon (approved):
LGTM

nntrainer/utils/util_simd.h
skykongkong8: I already conducted unittest on my local, but the reason why it is not here is because I moved this function to `util_simd.h` file from `tensor.h`.
I messed up with this typo issue while moving it.

jihochu (approved):
LGTM

djeong20 (approved):
lgtm

PR #2416, opened 2024-01-19
[TensorV2] Add support for applying operators with broadcasting 

djeong20:
This PR enables functionality to apply the given operator, such as multiply and divide, with broadcasting to the tensor.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -354,4 +354,66 @@ void FloatTensor::copy(const void *buf) {
   scopy(size(), (float *)buf, 1, (float *)getData(), 1);
 }
 
+void FloatTensor::apply_broadcast_util(
+  TensorV2 const &m,
+  std::function<void(const BroadcastInfoV2 &e, const float *, const float *,
+                     float *)>
+    v_func,
+  TensorV2 &output, const BroadcastInfoV2 &e, int cur_axis, size_t offset,
+  size_t m_offset) const {
+
+  const float *buf = (float *)this->getData();
+  const float *m_buf = m.getData<float>();
+  float *out_buf = output.getData<float>();
+
+  if (e.buffer_axis == cur_axis) {
+    v_func(e, buf + offset, m_buf + m_offset, out_buf + offset);
+    return;
+  }
+
+  cur_axis++;
+  unsigned int continuity[4] = {0, 1, 2, 3};
+  if (getFormat() == Tformat::NHWC) {
+    continuity[1] = 2;
+    continuity[2] = 3;
+    continuity[3] = 1;
+  }
+  for (unsigned int i = 0; i < dim.getTensorDim(continuity[cur_axis]); ++i) {
+    size_t next_offset = offset + i * strides[cur_axis];
+    size_t next_m_offset = m_offset + i * e.strides[cur_axis];
+    apply_broadcast_util(m, v_func, output, e, cur_axis, next_offset,
+                         next_m_offset);
+  }
+}
+
+void FloatTensor::apply_broadcast(
+  TensorV2 const &m,
+  std::function<void(const BroadcastInfoV2 &e, const float *, const float *,
+                     float *)>
+    v_func,
+  TensorV2 &output) const {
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/float_tensor.cpp
```diff
@@ -354,4 +354,66 @@ void FloatTensor::copy(const void *buf) {
   scopy(size(), (float *)buf, 1, (float *)getData(), 1);
 }
 
+void FloatTensor::apply_broadcast_util(
+  TensorV2 const &m,
```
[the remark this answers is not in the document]
djeong20: I think m is named after a matrix. As we previously discussed, it would be great to have a separate PR to unify all the naming conventions.

nntrainer/tensor/float_tensor.cpp
```diff
@@ -354,4 +354,66 @@ void FloatTensor::copy(const void *buf) {
...
+  const float *m_buf = m.getData<float>();
+  float *out_buf = output.getData<float>();
+
+  if (e.buffer_axis == cur_axis) {
+    v_func(e, buf + offset, m_buf + m_offset, out_buf + offset);
+    return;
+  }
+
+  cur_axis++;
+  uint continuity[4] = {0, 1, 2, 3};
```
jihochu, line 375: uint or unsigned int ? (line 381)

nntrainer/tensor/float_tensor.cpp
```diff
@@ -354,4 +354,66 @@ void FloatTensor::copy(const void *buf) {
...
+
+  /// shortcut to cover when dimension matches
+  /// note that buffer_size, the last stride is only used in v_func but it
+  /// might be changed
+  if (dim == m.getDim()) {
+    BroadcastInfoV2 e;
+    e.buffer_size = size();
+    e.strides[3] = 1;
+    e.tensor_type = getTensorType();
+    v_func(e, (float *)getData(), m.getData<float>(), output.getData<float>());
```
jihochu, line 412:
```suggestion
    v_func(e, getData<float>(), m.getData<float>(), output.getData<float>());
```

nntrainer/tensor/half_tensor.cpp
```diff
@@ -355,4 +355,59 @@ void HalfTensor::copy(const void *buf) {
...
+
+void HalfTensor::apply_broadcast_util(
+  TensorV2 const &m,
+  std::function<void(const BroadcastInfoV2 &e, const _FP16 *, const _FP16 *,
+                     _FP16 *)>
+    v_func,
+  TensorV2 &output, const BroadcastInfoV2 &e, int cur_axis, size_t offset,
+  size_t m_offset) const {
+
+  const _FP16 *buf = (_FP16 *)this->getData();
```
jihochu, line 395:
```suggestion
  const _FP16 *buf = getData<_FP16>();
```
same style would be more compatible.

nntrainer/tensor/float_tensor.cpp
djeong20: fixed to unsigned int :)

nntrainer/tensor/float_tensor.cpp
djeong20: In the Float/HalfTensor classes, getData() cannot be a template function since it overrides the TensorBase class method.
In other words, the virtual template function is not possible.

To keep the same style, the only option would be to make Tensor getData return a void pointer and manually cast it to the corresponding data type as follows. 

```
Tensor t(1,2,3,4);
float *data = (float *) t.getData();
```

I'm not sure which would be preferable for Tensor getData().
For now, I thought using the template function for the Tensor getData method would have more advantages in managing tensors. For instance,
```
Tensor t(1,2,3,4); // fp32 tensor
_FP16 *data = t.getData<_FP16>(); // possible to throw an exception with condition
_FP16 *data = (_FP16 *) t.getData(); // would be valid
```

jihochu: I misunderstood about new API. current is good. ^^

jihochu (approved):
LGTM

PR #2417, opened 2024-01-19
Make FP16 code more reusable. @open sesame 03/07 10:42

myungjoo:
This allows writing FP16-independent code without
breaking the current code.

This is a non-invasive subset of #2403 to
address the concern of https://github.com/nnstreamer/nntrainer/pull/2403#issuecomment-1899450849

I believe nntrainer can use this during the refactoring so that we do not introduce additional code divergences.

### DIFF
--- a/api/ccapi/include/tensor_dim.h
+++ b/api/ccapi/include/tensor_dim.h
@@ -9,6 +9,7 @@
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jijoong Moon <jijoong.moon@samsung.com>
  * @bug    No known bugs except for NYI items
+ * @todo   Move FP16 macros to somewhere common enough.
  *
  */
 
@@ -20,6 +21,7 @@
 #include <iosfwd>
 
 #include <bitset>
+#include <cstdint>
 #include <vector>
 
 #ifdef ENABLE_FP16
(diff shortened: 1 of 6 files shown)


api/ccapi/include/tensor_dim.h
```diff
@@ -28,8 +30,56 @@
 #else
 #define _FP16 _Float16
 #endif
+#else /* !ENABLE_FP16 */
+/* Keep the FP16 programming interface, but don't allow using it in run-time */
+#define _FP16 uint16_t
```
jijoongmoon, line 35: This is great. It definitely can be used for refactoring the multi type handle in NNTrainer.

nntrainer/layers/acti_func.h
```diff
@@ -54,6 +54,9 @@ class ActiFunc {
   template <typename T = float> void setActiFunc(ActivationType acti_type) {
     activation_type = acti_type;
 
+    if (typeid(T) == typeid(_FP16))
+      THROW_UNLESS_FP16_ENABLED;
```
jijoongmoon, line 58: this needs to be checked again. I don't think we need to throw an error here. These activation function type surely need to support FP16 datatype. To check, we have to enable the 'enable-fp16' option in meson_options.txt should be true.

SeoHyungjun (approved):
LGTM

nntrainer/layers/acti_func.h
myungjoo: This ensures thorwing an exception when FP16 activation is called without FP16 support.
By calling (conditional) THROW at the point where FP16 op actually exists, we can skip writing FP16 conditions at the callers (where you update codes mostly).

jihochu (approved):
LGTM

nntrainer/layers/acti_func.h
baek2sm: This is an exception for a situation where attempting to use FP16 even though FP16 is disabled in the Meson option, so it seems proper to handle the exception here.

baek2sm (approved):
LGTM

PR #2418, opened 2024-01-19
[TensorV2] Enable copying data from Tensor

djeong20:
This PR enables deep copy functionalities of a contiguous tensor with the following functions. `copy()`, `copyData()`, and `copy_with_strides()`.

The copy function completely copies the target tensor values regardless of the dimension of the input tensor. All elements and properties of the original tensor are copied to the new tensor. Therefore, if the copy function is used, a new tensor with the same size and shape as the original tensor is created.

On the other hand, the copyData function must match the size of the input and target tensors. This function only copies the data of the original tenor, so if the size or shape of the tensor is different, the copy may not be done properly.

Note that `copy` and `copyData` functions support copy data from multiple tensor data types while the `copy_with_strides` function only supports copying data from the same data type.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -293,6 +293,36 @@ TensorV2 &FloatTensor::apply(std::function<float(float)> f,
   return output;
 }
 
+void FloatTensor::copy(const TensorV2 &from) {
+  reshape(from.getDim());
+  copy(from.getData<float>());
+}
+
+void FloatTensor::copyData(const TensorV2 &from) {
+  NNTR_THROW_IF(!contiguous, std::invalid_argument)
+    << getName() << " is not contiguous, cannot copy.";
+
+  NNTR_THROW_IF(size() != from.size(), std::invalid_argument)
+    << "Size of tensor to copy must match";
+
+  switch (from.getDataType()) {
+  case ml::train::TensorDim::DataType::FP32:
+    copy(from.getData<float>());
+    break;
+  case ml::train::TensorDim::DataType::FP16:
+/// @todo remove #ifdef ENABLE_FP16
+#ifdef ENABLE_FP16
+    scopy(size(), from.getData<_FP16>(), 1, (float *)getData(), 1);
+#else
+    throw std::invalid_argument("Error: enable-fp16 is not enabled");
+#endif
+    break;
+  default:
+    throw std::invalid_argument("Error: Unsupported data type");
+    break;
+  }
+}
+
 void FloatTensor::print(std::ostream &out) const {
   printInstance(out, this);
   const float *data = (float *)getData();
@@ -342,10 +372,9 @@ void FloatTensor::print(std::ostream &out) const {
   out.copyfmt(init);
 }
 
-/// @todo include getName()
 void FloatTensor::copy(const void *buf) {
   NNTR_THROW_IF(!contiguous, std::invalid_argument)
-    << "Tensor is not contiguous, cannot copy.";
+    << getName() << " is not contiguous, cannot copy.";
 
   if (buf == getData()) {
     return;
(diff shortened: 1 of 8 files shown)


nntrainer/tensor/float_tensor.cpp
```diff
@@ -293,6 +293,37 @@ TensorV2 &FloatTensor::apply(std::function<float(float)> f,
...
+
+  if (size() != from.size())
+    throw std::invalid_argument("Size of tensor to copy must match");
+
+  switch (from.getDataType()) {
+  case ml::train::TensorDim::DataType::FP32:
+    copy(from.getData<float>());
+    break;
+  case ml::train::TensorDim::DataType::FP16:
+/// @todo remove #ifdef ENABLE_FP16
```
jijoongmoon, line 314: I think you can remove this in this PR.

nntrainer/tensor/half_tensor.cpp
```diff
@@ -293,6 +293,32 @@ TensorV2 &HalfTensor::apply(std::function<_FP16(_FP16)> f,
...
+void HalfTensor::copyData(const TensorV2 &from) {
+  if (!contiguous) {
+    throw std::runtime_error("Cannot copy non-contiguous tensor");
+  }
+
+  if (size() != from.size())
+    throw std::invalid_argument("Size of tensor to copy must match");
+
+  switch (from.getDataType()) {
+  case ml::train::TensorDim::DataType::FP32:
```
jijoongmoon, line 310: I think we do not need this FP32 case.

nntrainer/tensor/tensor_v2.cpp
```diff
@@ -180,6 +180,65 @@ void TensorV2::print(std::ostream &out) const { itensor->print(out); }
...
+    TensorV2 t = TensorV2(from.getDim(), true);
+    for (unsigned int b = 0; b < t.batch(); ++b) {
+      for (unsigned int c = 0; c < t.channel(); ++c) {
+        for (unsigned int h = 0; h < t.height(); ++h) {
+          for (unsigned int w = 0; w < t.width(); ++w) {
+            if (getDataType() == ml::train::TensorDim::DataType::FP32) {
+              t.setValue(b, c, h, w, from.getValue<float>(b, c, h, w));
+            } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
+              /// @todo remove #ifdef ENABLE_FP16
+#ifdef ENABLE_FP16
```
jijoongmoon, line 226: We can remove this ```ifdef ENABLE_FP16``` too.

jijoongmoon (approved):
LGTM!

nntrainer/tensor/float_tensor.cpp
djeong20: this case is fp16 --> float. float --> fp16 case will occur in the HalfTensor class.

nntrainer/tensor/tensor_v2.cpp
djeong20: yes. I'll remove every ifdef ENABLE_FP16 in the tensor level when enabling fp16 is always the case.

DonghakPark (approved):
LGTM!

SeoHyungjun (approved):
LGTM!

nntrainer/tensor/float_tensor.cpp
```diff
@@ -293,6 +293,37 @@ TensorV2 &FloatTensor::apply(std::function<float(float)> f,
   return output;
 }
 
+void FloatTensor::copy(const TensorV2 &from) {
+  reshape(from.getDim());
+  copy(from.getData<float>());
+}
+
+void FloatTensor::copyData(const TensorV2 &from) {
+  if (!contiguous) {
```
jihochu, line 302: Can't it use NNTR_IF.... ?

djeong20: fixed!

PR #2419, opened 2024-01-22
[TensorV2] Multiplication support

djeong20:
This PR adds support for performing the multiplication operation on two tensors.

- TensorV2 includes member functions to perform tensor multiplication.
- FloatTensor and HalfTensor take TensorV2 as input/output to perform multiplication.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -325,6 +325,56 @@ void FloatTensor::copyData(const TensorV2 &from) {
   }
 }
 
+int FloatTensor::multiply_i(float const &value) {
+  float *data = (float *)getData();
+  unsigned int len = size();
+
+  sscal(len, value, data, 1);
+
+  return ML_ERROR_NONE;
+}
+
+TensorV2 &FloatTensor::multiply(float const &value, TensorV2 &out) const {
+  auto f = std::bind(std::multiplies<float>(), std::placeholders::_1, value);
+  apply(f, out);
+  return out;
+}
+
+TensorV2 &FloatTensor::multiply(TensorV2 const &m, TensorV2 &output,
+                                const float beta) const {
+  auto f = [&](const BroadcastInfoV2 &e, const float *buf, const float *m_buf,
+               float *out_buf) {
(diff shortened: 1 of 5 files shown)


nntrainer/tensor/float_tensor.cpp
```diff
@@ -325,6 +325,56 @@ void FloatTensor::copyData(const TensorV2 &from) {
...
+  return out;
+}
+
+TensorV2 &FloatTensor::multiply(TensorV2 const &m, TensorV2 &output,
+                                const float beta) const {
+  auto f = [&](const BroadcastInfoV2 &e, const float *buf, const float *m_buf,
+               float *out_buf) {
+    if (e.strides[3] == 1 && output.getStrides()[3] == 1 && strides[3] == 1 &&
+        beta == 0.0) {
+      std::transform(buf, buf + e.buffer_size, m_buf, out_buf,
```
jijoongmoon, line 349: Even though we use this std::transform, we haven't tested it with other libs or implementations. Not in this PR, but we have to evaluate it.

nntrainer/tensor/float_tensor.cpp
```diff
@@ -325,6 +325,56 @@ void FloatTensor::copyData(const TensorV2 &from) {
...
+                     std::multiplies<float>());
+    } else {
+      for (unsigned int i = 0; i < e.buffer_size; ++i) {
+        *out_buf = *buf * *m_buf + beta * *out_buf;
+        buf += strides[3];
+        m_buf += e.strides[3];
+        out_buf += output.getStrides()[3];
```
jijoongmoon, line 356: If it is not continuous, we still can apply the multi-threading with something like omp. Not this PR, but we can also evaluate it.

jijoongmoon (approved):
LGTM except for some comments.

nntrainer/tensor/float_tensor.cpp
djeong20: gotcha. I'll add some testing to verify.

nntrainer/tensor/float_tensor.cpp
djeong20: agree. We can launch a thread using a function pointer or function objects in multithreading support.

SeoHyungjun (approved):
LGTM!

PR #2424, opened 2024-01-24
[CI] Add Pylint gitaction for gitaction CI

DonghakPark:
Add pylint yml file for python lint
- we move to gitaction from TAOS CI

- pylint.yml

**Self evaluation:**

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,50 @@
+# Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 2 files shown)


.github/workflows/pylint.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+
+    - name: Install dependencies
+      run: python -m pip install --upgrade pip && pip install pylint
```
jihochu, line 15: How about specifying pylint version ?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,17 @@
...
+    - name: Analysing the code with pylint
+      run: pylint $(git ls-files '*.py')
```
gichan-jang, line 17: How about checking the modified files instead of all files?

.github/workflows/pylint.yml
DonghakPark: i will fix !!

.github/workflows/pylint.yml
DonghakPark: good point !! i will fix

DonghakPark:
For all @nnstreamer/nntrainer 
- I set pylint as lastest version default option & you can check what's wrong in gitaction's detail
- if you want pylintrc file, please write pylintrc files and set them ```--rcfile=tools/pylintrc``` in yml files

gichan-jang (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2425, opened 2024-01-24
[CI] Add cpp file format checker

DonghakPark:
This patch adds a Github Action workflow to check cpp file format
- this file imported from deviceMLOps.MLAgent
- this file use cpp_linter marketplace actions

**Self evaluation:**

### DIFF
--- a/.github/workflows/cpp_linter.yml
+++ b/.github/workflows/cpp_linter.yml
@@ -1,0 +1,25 @@
+name: C++ Format Checker
+
+on: pull_request
+
+jobs:
+  cpp-linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cpp-linter/cpp-linter-action@v2
+        id: linter
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          style: file
+          extensions: 'cc'
+          version: 16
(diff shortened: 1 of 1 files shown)


djeong20 (commented):
quick question. current TAOS/pr-prebuild-clang specify the exact line to fix. would this work the same?

DonghakPark:
> quick question. current TAOS/pr-prebuild-clang specify the exact line to fix. would this work the same?

This action based on marketplace gitaction : https://github.com/marketplace/actions/c-c-linter
And according to explain they give line by line feedback !

SeoHyungjun (approved):
As far as I know, the official extension is yaml.
I understand that yml was created because the extension is limited to 3 characters in Windows.
How about changing it to yaml?

DonghakPark:
> As far as I know, the official extension is yaml. I understand that yml was created because the extension is limited to 3 characters in Windows. How about changing it to yaml?

I totally agree with you . The three-character limit is no longer a meaningful rule. 
However, since the default value for adding a new action in GitHub is defined as YML, all the GitAction files in NNTrafter have been unified to YML. I believe it is for wider compatibility and GitHub uses YML as the default value for GitAction.

What do you think about conforming to the default settings on Github or is it better to change official express ?

djeong20 (approved):

jijoongmoon (approved):
LGTM

PR #2427, opened 2024-01-26
[CI] Rename laber & upgrade node version & add workflow fail process

DonghakPark:
In order to improve the convenience and robustness of GitAction, make the following modifications:

1. In accordance with the guidelines to upgrade from Node 16 to Node 20, change gitaction-script version to v7
- ref : https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
2. If the check_count fails, alter it so that no additional actions are executed and the process terminates immediately.
3. adopt more descriptive and clear names for better understanding.

    renamed:    .github/workflows/Upload.yml -> .github/workflows/check_count.yml
    modified:   .github/workflows/labeler.yml

**Self evaluation:**

### DIFF
--- a/.github/workflows/check_count.yml
+++ b/.github/workflows/check_count.yml
@@ -1,4 +1,4 @@
-name: Upload Data
+name: Check Review
 
 on:
   pull_request:
@@ -7,7 +7,7 @@ on:
     types: [edited, dismissed, submitted]
 
 jobs:
-  label:
+  check_review:
     runs-on: ubuntu-latest
     steps:
       - name: review_count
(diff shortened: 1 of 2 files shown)


jijoongmoon (approved):
LGTM

djeong20 (commented):
I don't see the Need Review label in this PR. is this related to the change?

DonghakPark:
> I don't see the Need Review label in this PR. is this related to the change?

Yes, i related with this PR 

We are currently creating a "need review" label using two YAML files. 

1. "upload data," which checks the actual number of approvals based on the number of review approvals and the names of those who approved it, and saves it as an artifact. 

2. "Need Review Labeler," which reads this artifact and creates or deletes labels based on conditions. I wrote the code this way because of git action's permission issues. GitHub does not allow the creation or manipulation of labels on the main page for the files covered by the generated PR. 

**That is why the label is not working correctly in this PR.**

djeong20 (approved):

PR #2430, opened 2024-01-26
blas_neon: fix compiler errors in aarch64/Linux

myungjoo:
With stricter compilers, fp16 codes are not compilable.
To enable testing in non-android, fix type mismatches.

@s-debadri Note that there have been critical errors that Android compilers have ignored:
```C
for (; idx < (size - idx) >= 8; idx += 8) {
```
``` a < b >= c``` is not interpreted by C/C++ compilers as intended.
It is going to be interpreted as: ```( a < b ) ? ( 1 >= c ) : ( 0 >= c )```, not ``` (a < b) && (b >= c) ``` 
The compiler must have complained about this error/warning. Please DO NOT ignore or suppress them. Correct the code if there are warnings, please.

And ```float Y32[rows];``` (rows is a variable) is not in C/C++ syntax.

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -13,6 +13,7 @@
  */
 
 #include <blas_neon.h>
+#include <memory>
 #include <nntrainer_error.h>
 namespace nntrainer::neon {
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by jijoongmoon — APPROVED** [review]
LGTM

File: nntrainer/tensor/blas_neon.cpp [hunk 1]
```diff
@@ -1464,7 +1467,7 @@ void scopy_neon_fp16_to_fp32(const unsigned int N, const __fp16 *X, float *Y) {
     vst1q_f32(&Y[idx], y1);
   }
 
-  for (; idx < N; ++idx) {
+  for (; (unsigned int)idx < N; ++idx) {
```
djeong20, line 1470: can't `idx` be unsigned int?

djeong20 (approved):
LGTM

nntrainer/tensor/blas_neon.cpp
myungjoo: Updated

PR #2431, opened 2024-01-26
meson: fix fp16 support conditions for arm/aarch64

myungjoo:
According to GCC document,
https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/Half-Precision.html even if -mfp16-format=ieee is not given, aarch64 supports ieee fp16. Thus, for aarch64, even if the option is not available, try to built it with __fp16 type.

Then, add condition for arm: the final "else" is written for x64/x86 machines.

Need #2430 and test before merging.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -73,13 +73,35 @@ if get_option('enable-fp16')
      extra_defines += '-DUSE__FP16=1'
      extra_defines += '-DUSE_NEON=1'
    elif arch == 'aarch64'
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM

jihochu (approved):
LGTM

baek2sm (approved):
LGTM

SeoHyungjun (approved):
LGTM

DonghakPark (approved):
LGTM!

Issue #2435, opened 2024-01-29
Need to check version function


SeoHyungjun:
When the nntrainer library is loaded, it is difficult to check the version.
We need an API that can provide a version of nntrainer.

lhs8928:
@SeoHyungjun Can we close this issue?
I think #2436 resolves this issue.

PR #2436, opened 2024-01-29
[API] Add getVersion function

SeoHyungjun:
An API has been added to check the version of nntrainer through the project version of meson build.

Available via ml::train::getVersion.

issue : #2435

### DIFF
--- a/api/ccapi/src/factory.cpp
+++ b/api/ccapi/src/factory.cpp
@@ -24,6 +24,8 @@
 #include <optimizer.h>
 #include <optimizer_wrapped.h>
 
+#include <cstdlib>
+
 namespace ml {
 namespace train {
 
@@ -136,6 +138,16 @@ createLearningRateScheduler(const std::string &type,
                             const std::vector<std::string> &properties) {
   auto &ac = nntrainer::AppContext::Global();
   return ac.createObject<ml::train::LearningRateScheduler>(type, properties);
+}
+
+std::string getVersion() {
+  std::string version = std::to_string(VERSION_MAJOR);
+  version += ".";
+  version += std::to_string(VERSION_MINOR);
+  version += ".";
+  version += std::to_string(VERSION_MICRO);
+
+  return version;
 }
 
 } // namespace train
--- a/meson.build
+++ b/meson.build
@@ -10,6 +10,16 @@ project('nntrainer', 'c', 'cpp',
     'buildtype=release'
   ]
 )
+
+# Set version info
+nntrainer_version = meson.project_version()
+nntrainer_version_split = nntrainer_version.split('.')
+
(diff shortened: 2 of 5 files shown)


DonghakPark (approved):
LGTM! except android build error

meson.build
```diff
@@ -10,6 +10,16 @@ project('nntrainer', 'c', 'cpp',
     'buildtype=release'
   ]
 )
+
+# Set version info
+nntrainer_version = meson.project_version()
+nntrainer_version_split = nntrainer_version.split('.')
+
+add_project_arguments('-DVERSION="' + nntrainer_version + '"', language: ['c', 'cpp'])
+add_project_arguments('-DVERSION_MAJOR=' + nntrainer_version_split[0], language: ['c', 'cpp'])
```
jihochu, line 19: `NNTRAINER_VERSION_MAJOR`  or `VERSION_MAJOR` ?

EunjuYang:
LGTM!

lhs8928 (approved):
LGTM

meson.build
SeoHyungjun: The version information is stored as an argument in meson.build in the root.
However, in api/ccapi/meson.build, we define variables to add to the header.
It appears that the values defined in meson.build in the root cannot be used when creating headers.

project arguments
- add_project_arguments('-DVERSION_MAJOR=' + nntrainer_version_split[0], language: ['c', 'cpp'])

Definition for creating header
- version_conf.set('__NNTRAINER_VERSION_MAJOR__', nntrainer_version_split[0])

jihochu: > It appears that the values defined in meson.build in the root cannot be used when creating headers.

But,, project arguments are global var.  please check scope

```
Project arguments
Project arguments work similar to global arguments except that they are valid only within the current subproject. The usage is simple:

add_project_arguments('-DMYPROJ=projname', language : 'c')
This would add the compiler flags to all C sources in the current project.
```

SeoHyungjun: Modified to use only one.

In the case of andoird, separate settings were required, so it was added.

api/ccapi/src/factory.cpp
```diff
@@ -138,5 +140,15 @@ createLearningRateScheduler(const std::string &type,
...
 }
 
+std::string getVersion() {
+  std::string version = std::to_string(VERSION_MAJOR);
+  version += ".";
+  version += std::to_string(VERSION_MINOR);
+  version += ".";
+  version += std::to_string(VERSION_MICRO);
+
+  return version;
```
djeong20, line 150:
```suggestion
  return std::to_string(VERSION_MAJOR) + "."  + std::to_string(VERSION_MINOR) + "."  + std::to_string(VERSION_MICRO);
```
not a big deal, but a suggestion.

djeong20 (approved):
LGTM

skykongkong8 (approved):
Good PR!

PR #2441, opened 2024-01-31
[ Bug ] Fix coverity issues

skykongkong8:
- Fix non-const variables to const variables since their value is never changed in actual practice
- Use reference to avoid object copy
- Delete redundant variable

Resolves:
```
non-const type variable, but its value is never changed.
auto_causes_copy
```

**Self evaluation:**


djeong20 (approved):
LGTM

lhs8928 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2442, opened 2024-01-31
[LoRA] feat/ add LoRA to fc_layer @open sesame 03/07 10:42

EunjuYang:
(Update)
This PR contains commits to
* Add a new feature [LoRA](https://arxiv.org/abs/2106.09685) to `fc_layer`.
   * Add a new property : LoraRank (PositiveIntegerProperty)
   * Update forward func.
   * Update calcGradient func.
   * Update calcDerivative func.
* Update `node_exporter` of fully connected layer (to pass tflite test)

2024/03/06
* revise `forwarding` function
  * (W + A@B)x  ➡️  Wx + (A@B)x
* `forwarding_lora` is deleted
* Test with an application: two fc-layer w/ LoRA

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -180,7 +180,7 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
     Tensor &loraB = context.getWeight(lora_idx[LORAParams::loraB]);
     Tensor &weight_lora = context.getTensor(lora_idx[LORAParams::loraW]);
     Tensor &hidden_lora = context.getTensor(lora_idx[LORAParams::loraOut]);
-    loraA.dot(loraB, weight_lora); /* weight_lora = loraA @ loraB */
+    loraA.dot(loraB, weight_lora); 
     input_.dot(weight_lora, hidden_lora, false, false);
     hidden_.add_i(hidden_lora);
   }
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -41,7 +41,7 @@ enum LORAParams { loraA, loraB, loraW };
 
 FullyConnectedLayer::FullyConnectedLayer() :
   LayerImpl(),
-  fc_props(props::Unit(), props::LoraRank()){
+  fc_props(props::Unit(), props::LoraRank()) {
   weight_idx.fill(std::numeric_limits<unsigned>::max());
 }
 
@@ -58,7 +58,9 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
   auto &disable_bias = std::get<props::DisableBias>(*layer_impl_props);
 
   auto unit = std::get<props::Unit>(fc_props).get();
-  auto lora_rank = (std::get<props::LoraRank>(fc_props).empty())? 0 : std::get<props::LoraRank>(fc_props).get();
+  auto lora_rank = (std::get<props::LoraRank>(fc_props).empty())
+                     ? 0
+                     : std::get<props::LoraRank>(fc_props).get();
 
   NNTR_THROW_IF(context.getNumInputs() != 1, std::invalid_argument)
     << "Fully connected layer takes only one input";
@@ -106,11 +108,12 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
   }
 
   // create weights for LoRA
-  if(lora_rank){
+  if (lora_rank) {
 
     /** set LoRA specifications */
-    // loraA is a row vector (in_dim.width, lora_rank), loraB is a col vector of weight (lora_rank, out_dim),
-    // where shape of (loraA @ loraB) = shape of (W)
+    // loraA is a row vector (in_dim.width, lora_rank), loraB is a col vector of
+    // weight (lora_rank, out_dim), where shape of (loraA @ loraB) = shape of
+    // (W)
     TensorDim loraA_dim(
       1, is_nchw ? 1 : lora_rank, is_nchw ? in_dim.width() : 1,
       is_nchw ? lora_rank : in_dim.channel(),
@@ -133,10 +136,9 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
 
     // save weight_lora as a tensor to be updated by (loraA @ loraB)
     lora_idx[LORAParams::loraW] = context.requestTensor(
-      weight_dim, "weight_lora", Tensor::Initializer::NONE, true, 
+      weight_dim, "weight_lora", Tensor::Initializer::NONE, true,
       TensorLifespan::ITERATION_LIFESPAN);
   }
-
 }
 
 void FullyConnectedLayer::exportTo(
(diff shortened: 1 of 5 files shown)


nntrainer/layers/fc_layer.cpp
```diff
@@ -196,27 +252,54 @@ void FullyConnectedLayer::calcDerivative(RunLayerContext &context) {
...
+    // (cf) forward
+      // input_.dot(lora_weight, hidden) : hidden = input @ lora_weight
+      // lora_A.dot(lora_B, lora_weight) : lora_weight = lora_A @ lora_B
+    input_.dot_deriv_wrt_2(
+      djdlora_w, derivative_, false, false,
+      !context.isGradientFirstAccess(lora_idx[LORAParams::loraW]));
+    lora_A.dot_deriv_wrt_2(
+      djdlb, djdlora_w, false, false,
+      !context.isGradientFirstAccess(lora_idx[LORAParams::loraB]));
+    djdla.dot_batched_deriv_wrt_1(
```
lhs8928, line 299: Can you elaborate why did you use dot_batched_deriv_wrt_1 instead of dot_deriv_wrt_1?
As you described in comment lora_weight is result of dot product of lora_A and lora_B not dot_batch.

nntrainer/layers/fc_layer.cpp
```diff
@@ -115,6 +150,18 @@ void FullyConnectedLayer::setProperty(const std::vector<std::string> &values) {
   LayerImpl::setProperty(remain_props);
 }
 
+/**
+ * @brief forwarding function for lora. 
+ * It would be called during `forwarding` of FC layer.
+*/
+void FullyConnectedLayer::forwarding_lora(RunLayerContext &context, Tensor &weight){
```
lhs8928, line 157: If we are going to support LoRA this function will be needed all of layers not just fc layer so please consider to make it in parent class.

EunjuYang: Exactly! In this PR, I wrote the code only for fc layer, but it is required to make a parent class to cover LoRA in general way. I'm gonna reflect this comment. Thanks.

nntrainer/layers/fc_layer.cpp
```diff
@@ -115,6 +150,18 @@ void FullyConnectedLayer::setProperty(const std::vector<std::string> &values) {
...
+  Tensor &loraA = context.getWeight(lora_idx[LORAParams::loraA]);
+  Tensor &loraB = context.getWeight(lora_idx[LORAParams::loraB]);
+  Tensor &weight_lora = context.getTensor(lora_idx[LORAParams::loraW]);
+  loraA.dot(loraB, weight_lora);   // weight_lora = loraA @ loraB
+  weight.add(weight_lora, weight); // weight += weight_lora
```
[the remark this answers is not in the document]
EunjuYang: Thank you! I will replace it.

nntrainer/layers/fc_layer.cpp
EunjuYang: It was typo :) Thank you for your comment!

lhs8928 (approved):
LGTM

baek2sm (approved):
LGTM

skykongkong8 (approved):
LGTM!

nntrainer/layers/fc_layer.cpp
```diff
@@ -133,8 +179,12 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
       context.getWeightObject(weight_idx[FCParams::weight]).getOutputAxis();
 
     weight.dequantize(weight_, axis);
+    if (!std::get<props::LoraRank>(fc_props).empty())
+      forwarding_lora(context, weight_);
```
jijoongmoon, line 183: This is not the right way to implement LoRA. Overall, this is right but computing Lora's weight every time is not necessary. We only need this in finalize process of FC once. There is another way to implement it and I think it might be better to talk about it more.

jijoongmoon (changes requested):
Overall, It is ok but it might need further discussion.

nntrainer/layers/fc_layer.cpp
```diff
@@ -196,27 +246,54 @@ void FullyConnectedLayer::calcDerivative(RunLayerContext &context) {
...
+    Tensor &djdla = context.getWeightGrad(lora_idx[LORAParams::loraA]);
+    Tensor &djdlb = context.getWeightGrad(lora_idx[LORAParams::loraB]);
+    Tensor &djdlora_w = context.getTensorGrad(lora_idx[LORAParams::loraW]);
+
+    const Tensor &derivative_ = context.getIncomingDerivative(SINGLE_INOUT_IDX);
+    Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
+    Tensor &lora_A = context.getWeight(lora_idx[LORAParams::loraA]);
+    Tensor &lora_B = context.getWeight(lora_idx[LORAParams::loraB]);
+
+    // (cf) forward
```
SeoHyungjun, line 284: Is this a typo in FC?

nntrainer/layers/fc_layer.cpp
```diff
@@ -102,6 +106,39 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
       context.requestWeight(bias_dim, bias_initializer, WeightRegularizer::NONE,
                             1.0f, bias_decay, "bias", true);
   }
+
+  // create weights for LoRA
+  if (lora_rank) {
+
+    /** set LoRA specifications */
```
SeoHyungjun, line 113: How about using block comments?

SeoHyungjun (approved):
LGTM

nntrainer/layers/fc_layer.cpp
EunjuYang: Could you explain more details about _'separate training and inference'_? In `edc76c`, I revised the code in the way of separately forward baseline weight (W) and lora parameters (A@B). Still, I think there is more points that I have to consider to update the LoRA in terms of the _mode_. Thanks!

nntrainer/layers/fc_layer.cpp
EunjuYang: It meant 'confer (refer)'. It looks redundant,  so I removed it :)

EunjuYang:
> Overall, It is ok but it might need further discussion.

I reflected the off-line discussion in 
[68ec281](https://github.com/nnstreamer/nntrainer/pull/2442/commits/68ec281eea3592c9c839583cf3b03ad9a2358f96). Thanks  :)

nntrainer/layers/fc_layer.cpp
```diff
@@ -102,6 +105,46 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
...
+    TensorDim loraA_dim(
+      1, is_nchw ? 1 : lora_rank, is_nchw ? in_dim.width() : 1,
+      is_nchw ? lora_rank : in_dim.channel(),
+      TensorDim::TensorType(context.getFormat(), context.getWeightDataType()),
+      is_nchw ? 0b0011 : 0b0101);
+
+    /** loraB: (lora_rank, out_dim) */
+    TensorDim loraB_dim(
+      1, is_nchw ? 1 : unit, is_nchw ? lora_rank : 1,
+      is_nchw ? unit : in_dim.channel(),
```
djeong20, line 122: so there's no use of `lora_rank` in the channels last case for B?

DonghakPark (approved):
LGTM!
i will fix FC Layer can support export to tflite format later

nntrainer/layers/fc_layer.cpp
EunjuYang: It should be fixed! Thanks!!:)

SeoHyungjun (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2447, opened 2024-02-01
Clean up fp16 related build scripts and conditions

myungjoo:
1. unsigned / signed int type mismatch in blas_neon.cpp
2. build condition cleanup (meson.build)
3. explicitly disable fp16 in armv7l/Tizen. (not because of HW support, but because of the library usage of nntrainer)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -67,6 +67,8 @@ if get_option('enable-fp16')
      extra_defines += '-DENABLE_FP16=1'
      extra_defines += '-DUSE__FP16=1'
      extra_defines += '-DUSE_NEON=1'
+   elif arch == 'arm'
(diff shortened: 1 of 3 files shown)


skykongkong8 (approved):
I could find some more non-unsigned-int iterator variables in `blas_neon.cpp` file.
Should I make a Wait-for PR? If you are not working on it further about this specific problem, please let me know...!

myungjoo:
> I could find some more non-unsigned-int iterator variables in `blas_neon.cpp` file. Should I make a Wait-for PR? If you are not working on it further about this specific problem, please let me know...!

Fix them without waiting for other PRs. We can always merge manually.
We also need to reenable compiler warning-erros to prevent such.

jihochu (approved):
LGTM

djeong20 (approved):
LGTM

PR #2448, opened 2024-02-01
[ Tensor ] copyData to fp16 from fp32

skykongkong8:
- copyData function now supports fp16_fp32 conversion
- This is crucial for BatchNormalization layer in mixed precision training

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -3092,7 +3092,9 @@ void Tensor::copyData(const Tensor &from) {
       }
     } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
-      if (from.getDataType() == ml::train::TensorDim::DataType::QINT8) {
+      if (from.getDataType() == ml::train::TensorDim::DataType::FP32) {
+        scopy(size(), from.getData<float>(), 1, getData<_FP16>(), 1);
+      } else if (from.getDataType() == ml::train::TensorDim::DataType::QINT8) {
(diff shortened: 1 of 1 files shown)


djeong20 (approved):
LGTM

nntrainer/tensor/tensor.cpp
```diff
@@ -3092,7 +3092,9 @@ void Tensor::copyData(const Tensor &from) {
       }
     } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
-      if (from.getDataType() == ml::train::TensorDim::DataType::QINT8) {
+      if (from.getDataType() == ml::train::TensorDim::DataType::FP32) {
+        scopy(size(), from.getData<float>(), 1, getData<_FP16>(), 1);
```
jijoongmoon, line 3096: I wonder how it works for the unit16 case for (_FP16). In the end, we plan to make enable_fp16 true always and it also works for unit16 case as well.

jijoongmoon (approved):
LGTM except for minor comment.

nntrainer/tensor/tensor.cpp
skykongkong8: Adding function that works like : uint16 -> __fp16 -> fp32 would do for that.
But there is no such case for now maybe?

lhs8928 (approved):
LGTM

nntrainer/tensor/tensor.cpp
jijoongmoon: yes. not now. Thanks.

PR #2449, opened 2024-02-01
dist/Tizen: disable fp16 in Tizen

myungjoo:
NNTrainer FP16 implementation relies on NEON, which requires armv8.2-a ISA.

Tizen aarch64 is on armv8.0-a; thus it cannot support fp16-neon. Thus, disable fp16 in armv7l and aarch64.

Tizen x86/x64 does not support fp16, too.

This re-enabled Tizen build of nntrainer.
Please do not break the build in main branch!


skykongkong8 (approved):
LGTM

djeong20 (approved):
LGTM

PR #2450, opened 2024-02-01
[CI] Update Pylint - pip install with requirements.txt

DonghakPark:
In order to properly check the Python files used by NNTrainer, all modules used in Python must be downloaded.
For now, this part is missing in CI and this PR add and update the modules currently used by NNTrainer.
- fix Applications/LLaMA/PyTorch/weights_converter.py can got 10 score in pylint

- modified:   .github/workflows/pylint.yml
- modified:   Applications/LLaMA/PyTorch/weights_converter.py
- new file:   tools/requirements.txt

**Self evaluation:**


skykongkong8 (approved):
LGTM!
one quick question: how about the file that we recently discussed on off-line? `genLayerTests.py`
Can I refer to this file?

Applications/LLaMA/PyTorch/weights_converter.py
```diff
@@ -1,58 +1,63 @@
...
+        save_weight(params[layer_name + "self_attn.o_proj" + ".weight"].permute(1, 0))
 
-    def save_feed_forward(weights, layer_name):
-        save_weight(params[layer_name + 'post_attention_layernorm' + '.weight'])
-        save_weight(params[layer_name + 'mlp.up_proj' + '.weight'].permute(1, 0))
-        save_weight(params[layer_name + 'mlp.gate_proj' + '.weight'].permute(1, 0))        
-        save_weight(params[layer_name + 'mlp.down_proj' + '.weight'].permute(1, 0))
+    def save_feed_forward(layer_name):
+        save_weight(params[layer_name + "post_attention_layernorm" + ".weight"])
+        save_weight(params[layer_name + "mlp.up_proj" + ".weight"].permute(1, 0))
```
jihochu, line 36: Should we change this?

DonghakPark: Yes. This part is not essential for functionality, but it is a string representation method according to the PEP8 standard recommended by Python. Although it is a convention, I changed it because I thought it would be better to adhere to the latest standard. (I did not change it manually, but used Black to format it.)

DonghakPark:
> LGTM! one quick question: how about the file that we recently discussed on off-line? `genLayerTests.py` Can I refer to this file?

As. discuss offline please fix python code !

djeong20 (approved):
Good job!

jijoongmoon (approved):
LGTM. but we need to think this is only for the CI and pylint. how about making the configuration file as hidden or save in tools/ci?

DonghakPark:
> LGTM. but we need to think this is only for the CI and pylint. how about making the configuration file as hidden or save in tools/ci?

That right!! i will make ci dir!!

Applications/LLaMA/PyTorch/weights_converter.py
jihochu: PEP8 doe snot enforce double quoted for string.

>String Quotes
In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.

jijoongmoon (approved):
LGTM

PR #2451, opened 2024-02-01
[ fp16 ] Fix to avoid build error / compiler warnings

skykongkong8:
- Use unsigend int iterator to avoid compiler warnings
- Keeping in mind that NEON is properly supported on later version than ARMv8.2, modify sources to include.

### DIFF
--- a/nntrainer/utils/meson.build
+++ b/nntrainer/utils/meson.build
@@ -24,9 +24,14 @@ if get_option('enable-trace')
   util_headers += 'tracer.h'
 endif
 
-if get_option('enable-fp16')
-  util_sources += 'util_simd_neon.cpp'
-  util_headers += 'util_simd_neon.h'
+arch = host_machine.cpu_family()
+if get_option('enable-fp16') 
+  if arch == 'arm'
+    error ('FP16/ARM code (blas_neon.cpp) uses armv8.2 instructions. armv7 is not supported.')
+  elif arch == 'aarch64' or get_option('enable-neon')
+    util_sources += 'util_simd_neon.cpp'
+    util_headers += 'util_simd_neon.h'
+  endif
 endif
 
 foreach s : util_sources
--- a/meson.build
+++ b/meson.build
@@ -66,7 +66,7 @@ warning_c_flags = [
 
 
 if get_option('enable-fp16')
-   arch = target_machine.cpu_family()
+   arch = host_machine.cpu_family()
    if get_option('platform') == 'android'
      add_project_arguments('-mfp16-format=ieee', language: ['c', 'cpp'])
      extra_defines += '-DENABLE_FP16=1'
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -37,6 +37,7 @@ option('enable-blas', type: 'boolean', value: true)
 option('enable-fp16', type: 'boolean', value: false)
 option('enable-cublas', type: 'boolean', value: false)
 option('enable-openmp', type: 'boolean', value: true)
+option('enable-neon', type: 'boolean', value: false)
 
 # ml-api dependency (to enable, install capi-inference from github.com/nnstreamer/api )
 # To inter-operate with nnstreamer and ML-API packages, you need to enable this.
--- a/nntrainer/tensor/meson.build
+++ b/nntrainer/tensor/meson.build
@@ -34,10 +34,14 @@ tensor_headers = [
   'blas_interface.h'
 ]
 
-arch = target_machine.cpu_family()
-
-if get_option('platform') == 'android' or arch == 'aarch64'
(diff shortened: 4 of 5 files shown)


jihochu (approved):
LGTM

baek2sm (approved):
LGTM

EunjuYang:
LGTM!

jijoongmoon (approved):
LGTM

nntrainer/utils/meson.build
```diff
@@ -24,9 +24,13 @@ if get_option('enable-trace')
   util_headers += 'tracer.h'
 endif
 
-if get_option('enable-fp16')
+arch = target_machine.cpu_family()
+
+if get_option('platform') == 'android' or arch == 'aarch64'
```
myungjoo, line 29: Nope. If it's aarch64 with armv8.0, you cannot use fp16 simd-neon.
You need to depend on fp16 enable/disable explicitly

skykongkong8: Then how about:
1. fp16 using and android -> use neon
```meson
if get_option('enable-fp16') && get_option('platform') == 'android'
```
or,

2. make explicit option for using neon
```meson
if get_option('enable-neon')
```
?

Current option would include neon files for linux build as well. So we need a new option to separate them.

SeoHyungjun (approved):
LGTM

nntrainer/utils/meson.build
skykongkong8: ### Update 
During android build using `nntrainer/tools/package_android.sh` , 
`target_machine.cpu_family()` output is `x86_64` since we actually build on PC at this moment.
Thus think I should make a explicit option like:
```meson
if get_option('enable-neon')
```
and make such option true for ARM build.  

We were free from this problem because we just used 
```meson
if get_option('platform') == 'anrdoid'
```
which is not really a valid option.

This way, I think we can segregate the problem like:

$$ARM > NEON > FP16$$
> fyi, neon supports multiple dataTypes, so it is a bigger set than `ENABLE_FP16` option

nntrainer/utils/meson.build
```diff
@@ -24,9 +24,13 @@ if get_option('enable-trace')
...
 endif
 
-if get_option('enable-fp16')
+arch = target_machine.cpu_family()
+
+if get_option('enable-fp16') && arch == 'aarch64'
   util_sources += 'util_simd_neon.cpp'
   util_headers += 'util_simd_neon.h'
+elif arch == 'arm'
+  error ('FP16/ARM code (blas_neon.cpp) uses armv8.2 instructions. armv7 is not supported.')
```
myungjoo, line 33: You always get an error in arm with this code.

```
if get_option('enable-fp16') 
  if arch == 'aarch64'
    util_sources += 'util_simd_neon.cpp'
    util_headers += 'util_simd_neon.h'
  elif arch == 'arm'
    error ('FP16/ARM code (blas_neon.cpp) uses armv8.2 instructions. armv7 is not supported.')
  endif
```

skykongkong8: Yes, I am aware of it. (fixed in my local)
But the main problem is that `arch` during `package_android_build.sh` is `x86_64`.
codes like:
```meson
if arch == 'aarch64'
```
will not solve the current build fail.

So basically for `enable-fp16`:
```psuedo-code
if enable-fp16
  if enable-neon
    if arch == 'arm'
      error ( ... )
    else
      // add neon src files ?
```
In current android-build use, there is no case for `arch` to have `aarch64`, unless we build it from ARM based controller

myungjoo: Try "host machine" instead of "target machine", too.

Target machine == aarch64 means that the compiler you have built here will create aarch64 binary. (you are not compiling a compiler): https://mesonbuild.com/Cross-compilation.html

skykongkong8: it still says `x86_64`.
Think it is because we have been differentiating it with `-Dplatform=android`, and actually cross-compile with ndk-build.
Will check further ASAP

myungjoo: 1. test host-machine, target-machine in GBS. see if it gives the correct arch name
2. if 1 == correct, it means host-machine in Android build is the only issue.
   then you can do: (psuedo code)
```
if ANDROID
  build a dummy a.c.
  run "file a.out"
  read arch name from above
  set arch.
endif
```

myungjoo: And your code above is ok, too.

skykongkong8: Got it. Will let you know after trying that

skykongkong8: I checked that I could get correct  host-machine, target-machine conditions depending on `-A {$architecture}` option when GBS build.

PR #2455, opened 2024-02-02
Support mixed precision training @open sesame 03/08 07:57

jihochu:
It adds loss scale factor for removing invalid data while training.
The factor is dynamically calculated while gradient clipping step, And it initially disabled until 
loss scale proeprty is set.
fc/pooling/conv2d/softmax layers are modified for loss scale and mixed tensor type.

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -455,6 +455,18 @@ void NetworkGraph::backwarding(
     loss_scale = scale;
   };
 
+  auto check_weights = [](std::vector<Weight *> &weights) {
+    bool valid = true;
+    for (auto &w : weights) {
+      auto grad = w->getGradient();
+      if (grad.checkDataValidation(false) == false) {
+        grad.setZero();
+        valid = false;
+      }
+    }
+    return valid;
+  };
+
   // check first layer's derivative is valid
   // loss scale is adjusted between 1.0f ~ 256.0f
   // @todo provide max scale property
@@ -465,13 +477,15 @@ void NetworkGraph::backwarding(
     ml_logd(
       "Derivative validation failed. Skip applying gradient. loss_scale(%f)",
       scale);
+    check_weights(clip_weights);
     update_loss_scale(scale);
     return;
   } else {
     for (unsigned int idx = 0; idx < clip_weights.size(); idx++) {
       auto const &w = clip_weights[idx];
       w->applyScaler(loss_scale);
-      if (w->getGradient().checkDataValidation(false) == false) {
+
+      if (!check_weights(clip_weights)) {
         float scale = loss_scale > 1.5f ? loss_scale - 0.5f : 1.0f;
         ml_loge("gradient validation failed. skip update. loss_scale(%f)",
                 scale);
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -6,6 +6,7 @@
  * @date    19 Oct 2020
  * @see     https://github.com/nnstreamer/nntrainer
  * @author  Jijoong Moon <jijoong.moon@samsung.com>
+ * @author  Jiho Chu <jiho.chu@samsung.com>
  * @bug     No known bugs except for NYI items
  * @brief   This is Network Graph Class for Neural Network
  *
@@ -84,6 +85,14 @@ int NetworkGraph::compile(const std::string &loss_type) {
 
   status = checkCompiledGraph();
   NN_RETURN_STATUS();
+
+  /* @note It can be integrated with addLossLayer method
+   * if it removes adding loss layer to the model directly.
+   */
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    auto &ln = *iter;
+    ln->setLossScale(loss_scale);
+  }
 
   compiled = true;
 
@@ -353,10 +362,15 @@ sharedConstTensors NetworkGraph::forwarding(
   bool training,
   std::function<void(std::shared_ptr<LayerNode>, bool)> forwarding_op,
   std::function<bool(void *userdata)> stop_cb, void *userdata) {
+
+  for (auto w : clip_weights) {
+    w->applyMaster();
+  }
+
   for (auto iter = cbegin(); iter != cend() && !stop_cb(userdata); iter++) {
     auto &ln = *iter;
     PROFILE_TIME_START(profile_keys.at(ln->getType()));
-    forwarding_op(*iter, training);
+    forwarding_op(ln, training);
     PROFILE_TIME_END(profile_keys.at(ln->getType()));
   }
 
@@ -397,7 +411,7 @@ void NetworkGraph::backwarding(
   int iteration,
   std::function<void(std::shared_ptr<LayerNode>, int)> &backwarding_op,
   std::function<void(Weight &, int)> &apply_grad_clip_op,
-  std::function<bool(void *userdata)> stop_cb, void *userdata) const {
+  std::function<bool(void *userdata)> stop_cb, void *userdata) {
   /**
    * last layer backwarding is run out of this loop
    */
@@ -426,6 +440,46 @@ void NetworkGraph::backwarding(
   if (clip_weights.empty())
     return;
 
+  /**
+   * mixed precision trainging needs gradient clipping and loss scale,
+   * cause all weights are updated with clipping option.
+   * also, loss scale makes to avoid unexpected training result.
+   */
+  auto update_loss_scale = [&](float scale) {
+    ml_logd("set loss scale = %f", scale);
+    for (auto iter = cbegin(); iter != cend(); iter++) {
+      auto &ln = *iter;
+      ln->setLossScale(scale);
+    }
+    loss_scale = scale;
+  };
+
+  // check first layer's derivative is valid
+  // loss scale is adjusted between 1.0f ~ 256.0f
+  // @TODO provide max scale property
+  auto &ln = *(cbegin() + 1);
+  if (loss_scale != 0.0f && !ln->getRunContext().validateDerivatives()) {
+    // It will not apply train results if data is invalid
+    float scale = loss_scale > 1.5f ? loss_scale - 0.5f : 1.0f;
+    ml_logd(
+      "Derivative validation failed. Skip applying gradient. loss_scale(%f)",
+      scale);
+    update_loss_scale(scale);
+    return;
+  } else {
+    for (unsigned int idx = 0; idx < clip_weights.size(); idx++) {
+      auto const &w = clip_weights[idx];
+      w->applyScaler(loss_scale);
+      if (w->getGradient().checkDataValidation(false) == false) {
+        float scale = loss_scale > 1.5f ? loss_scale - 0.5f : 1.0f;
+        ml_loge("gradient validation failed. skip update. loss_scale(%f)",
+                scale);
+        update_loss_scale(scale);
+        return;
+      }
+    }
+  }
+
   /** calculate the global norm */
   Tensor global_norm_t(
     TensorDim({1u, 1u, 1u, (unsigned int)clip_weights.size()}));
@@ -434,6 +488,7 @@ void NetworkGraph::backwarding(
     auto const &w = clip_weights[idx];
     global_norm_data[idx] = w->getGradientNorm();
   }
+
   float global_norm = global_norm_t.l2norm();
   /** apply the gradient with the above global norm */
   for (auto w : clip_weights) {
@@ -442,6 +497,12 @@ void NetworkGraph::backwarding(
   /** apply the gradient with the above global norm */
   for (auto w : clip_weights) {
     apply_grad_clip_op(*w, iteration);
+  }
+
+  // update loss scale
+  if (loss_scale != 0.0f) {
+    float scale = loss_scale + 2.0f;
+    update_loss_scale(scale);
   }
 }
 
@@ -606,6 +667,14 @@ NetworkGraph::canExecuteInPlace(const std::shared_ptr<LayerNode> &lnode) {
     };
 
   /**
+   * if the layer's input and output type is not FP32, then it cannot be
+   * inplace. We assume that the input is always FP32.
+   */
+  if (lnode->getInputConnections().empty() &&
+      !istrequal(getTensorType()[2], "FP32"))
+    return InPlace::NONE;
+
+  /**
    * @note Conditions to decide if this layer node can be in-place:
    * 1. if the layer is a no-op, then it can operate in-place as it is not
    * modifying its input/output tensors and does not need to check its
@@ -684,15 +753,6 @@ NetworkGraph::canExecuteInPlace(const std::shared_ptr<LayerNode> &lnode) {
       return InPlace::NON_RESTRICTING;
 
     return InPlace::RESTRICTING;
-  }
-
-  /**
-   * if the layer's input and output type is not FP32, then it cannot be
-   * inplace. We assume that the input is always FP32.
-   */
-  if (lnode->getInputConnections().empty()) {
-    if (!istrequal(getTensorType()[2], "FP32"))
-      return InPlace::NONE;
   }
 
   return InPlace::NONE;
@@ -876,7 +936,11 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
   lnode->configureRunContext(
     // TODO: update weights spec for trainable based on layer trainable prop
     tensor_manager->requestWeights(gnode, init_context.getWeightsSpec(),
-                                   lnode->getTrainable(), shared_weight_names),
+                                   lnode->getTrainable(), shared_weight_names,
+                                   init_context.getActivationDataType() !=
+                                       init_context.getWeightDataType()
+                                     ? init_context.getActivationDataType()
+                                     : TensorDim::DataType::NONE),
     inputs, outputs,
     tensor_manager->requestTensors(gnode, init_context.getTensorsSpec(),
                                    lnode->getTrainable(), shared_tensor_names));
@@ -1551,6 +1615,7 @@ void NetworkGraph::flushCacheExcept(unsigned int order) {
 void NetworkGraph::requestOptimizerVariable(
   std::function<std::vector<TensorDim>(const TensorDim &)> cb,
   bool request_only_trainable) {
+  bool need_master = !istrequal(getTensorType()[1], getTensorType()[2]);
   for (auto const &w : tensor_manager->getWeights()) {
     if (w->isGradientLastAccess() && w->hasGradient()) {
       const TensorDim &dim = w->getDim();
@@ -1558,6 +1623,17 @@ void NetworkGraph::requestOptimizerVariable(
       w->setOptimizerVariables(tensor_manager->requestWeightOptimizerVariables(
         dims, w->getName(), TensorLifespan::MAX_LIFESPAN,
         w->isGradientClipByGlobalNorm(), Tensor::Initializer::ZEROS));
+      if (need_master) {
+        for (auto &dim : dims)
+          dim.setDataType(
+            str_converter<enum_class_prop_tag, nntrainer::TensorDataTypeInfo>::
+              from_string(getTensorType()[1]));
+        w->setOptimizerMasterVariables(
+          tensor_manager->requestWeightOptimizerVariables(
+            dims, w->getName(), TensorLifespan::MAX_LIFESPAN,
+            w->isGradientClipByGlobalNorm(), Tensor::Initializer::ZEROS,
+            need_master));
+      }
     }
   }
 }
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -86,7 +86,8 @@ int NetworkGraph::compile(const std::string &loss_type) {
   status = checkCompiledGraph();
   NN_RETURN_STATUS();
 
-  /* @note It can be integrated with addLossLayer method
+  /**
+   * @note It can be integrated with addLossLayer method
    * if it removes adding loss layer to the model directly.
    */
   for (auto iter = cbegin(); iter != cend(); iter++) {
@@ -456,7 +457,7 @@ void NetworkGraph::backwarding(
 
   // check first layer's derivative is valid
   // loss scale is adjusted between 1.0f ~ 256.0f
-  // @TODO provide max scale property
+  // @todo provide max scale property
   auto &ln = *(cbegin() + 1);
   if (loss_scale != 0.0f && !ln->getRunContext().validateDerivatives()) {
     // It will not apply train results if data is invalid
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -182,6 +182,11 @@ void BatchNormalizationLayer::forwarding(RunLayerContext &context,
   Tensor &cvar = context.getTensor(wt_idx[BNParams::cvar]);
 
   if (training) {
+    t_reduced.setZero();
+    deviation.setZero();
+    invstd.setZero();
+    cvar.setZero();
+
     input_.average(axes_to_reduce, t_reduced);
     input_.subtract(t_reduced, deviation);
 
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -112,6 +112,12 @@ void BatchNormalizationLayer::finalize(InitLayerContext &context) {
                           bias_decay, "beta", true);
 
   /**
+   * @note declare weigth dimention with activation datatype
+   */
+  TensorDim w_dim = dim;
+  w_dim.setDataType(in_dim.getDataType());
+
+  /**
    * caches the deviation -> input - avg(input)
    * @todo check if avoiding this storage and adding dependency on input (no
    * more in-place calculation) can save memory during memory optimization.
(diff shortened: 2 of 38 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -422,16 +424,43 @@ void NetworkGraph::backwarding(
     PROFILE_TIME_END(profile_keys.at(ln->getType()));
   }
 
+  auto update_loss_scale = [&](float scale) {
+    for (auto iter = iter_begin; iter != iter_end; iter++) {
+      auto &ln = *iter;
+      auto type = ln->getType();
+      if (type == CrossEntropyLossLayer::type ||
```
jijoongmoon, line 431: I'm worried when the developer uses his loss layer. If he wants to use it.. then this loss scale seems not to work properly.

jihochu:
> nntrainer/graph/network_graph.cpp:16:10 [clang-diagnostic-error]: nntrainer/graph/network_graph.cpp#L16
'node_exporter.h' file not found

@DonghakPark , please check cpp-linter 's behavior. It's not differ from clang which we are using with TAOS.

nntrainer/models/neuralnet.cpp
```diff
@@ -70,7 +70,7 @@ NeuralNetwork::NeuralNetwork() :
     props::Epochs(), props::TrainingBatchSize(), props::SavePath(),
     props::ContinueTrain(), props::SaveBestPath(), props::MemoryOptimization(),
     props::MemorySwap(), props::MemorySwapPath(), props::MemorySwapLookahead(),
-    props::TensorFormat(), props::ModelTensorDataType()),
+    props::TensorFormat(), props::ModelTensorDataType(), props::LossScale()),
```
jijoongmoon, line 73: I wonder whether props::LossScale is layer common property or network comment property. If it is network common property, then it should be set in ```model/model_common_properties.h```.

nntrainer/graph/network_graph.cpp
jihochu: It modified for loss layers to be grouped, and loss scale is controlled by the loss layers method.
Also, this type checking is removed for custom layer as you mentioned.

nntrainer/layers/layer_node.cpp
```diff
@@ -307,9 +308,8 @@ const std::vector<std::string> LayerNode::getInputLayers() const {
   names.reserve(input_connections.size());
   std::transform(
     input_connections.begin(), input_connections.end(),
-    std::back_inserter(names), [](const Connection &con) -> const auto & {
-      return con.getName();
-    });
+    std::back_inserter(names),
+    [](const Connection &con) -> const auto & { return con.getName(); });
```
jijoongmoon, line 312: no need to modify.

nntrainer/tensor/tensor.h
```diff
@@ -1923,6 +1924,48 @@ class Tensor {
...
+   */
+  bool checkDataValidation(bool print = false) const {
+    bool ret = true;
+
+    // It only support FP16 and FP32
+    if (getDataType() != Tdatatype::FP32 && getDataType() != Tdatatype::FP16)
+      return true;
+
+    std::string data;
+    for (unsigned int i = 0, len = size(); i < len; ++i) {
```
jijoongmoon, line 1958: not this pr, but later we can accelerate with SIMD or OMP.

nntrainer/models/neuralnet.cpp
jihochu: It used for both, so it may be better with just common property.
If a user set it as model property, train core will set the value to layer property dynamically.

nntrainer/graph/network_graph.cpp
jihochu: Applying loss scale needs to be performed while derivative calculation, so it changed to just provide a method to apply loss scale.

nntrainer/models/neuralnet.cpp
jihochu: > It used for both, so it may be better with just common property.
> If a user set it as model property, train core will set the value to layer property dynamically.

=> It moved to model property.

nntrainer/tensor/tensor.h
jihochu: SIMD support is really necessary!! it's too slow.

nntrainer/layers/layer_node.cpp
jihochu: It may be changed by clang-format checker;;

nntrainer/tensor/tensor.h
```diff
@@ -1923,6 +1944,53 @@ class Tensor {
...
+        }
+#ifdef ENABLE_FP16
+        else if (getDataType() == Tdatatype::FP16) {
+          data.append(std::to_string(static_cast<float>(getData<_FP16>()[i])) + ", ");
+        }
+#endif
+      }
+
+      if (getDataType() == Tdatatype::FP32) {
+        if (std::isnan(getData<float>()[i]) | std::isinf(getData<float>()[i])) {
```
myungjoo, line 1971: ```|``` ==> ```||```

nntrainer/tensor/tensor.h
myungjoo: https://github.com/VcDevel/std-simd

or

something like:
```
#include <iostream>
#include <experimental/simd>

using intV = std::experimental::fixed_size_simd<int, 8>;
using floatV = std::experimental::fixed_size_simd<float, 11>;


int main()
{
    int ia[] = {3, 1, -14, 1, 5, 9, -14, 9};
    float fa[] = {0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0};
    intV iv {[ia](const auto i) { return ia[i]; }};
    floatV fv {[fa](const auto i) { return fa[i]; }};

    iv += 100;
    fv += 10.0f;

    std::cout<<iv[0]<<std::endl;
    std::cout<<fv[0]<<std::endl;
}

```

nntrainer/tensor/tensor.h
myungjoo: ```||``` is often cheaper than ```|```

Use ```||``` or ```!std::isfinite()``` or ```std::fpclassify & (FP_NAN | FP_INFINITE)```

You may be in a big loop.

nntrainer/tensor/tensor.h
jihochu: I'll investigate this in later PR.

nntrainer/tensor/tensor.h
jihochu: It fixed to use isfinit()

nntrainer/graph/network_graph.cpp
```diff
@@ -426,14 +436,51 @@ void NetworkGraph::backwarding(
...
   Tensor global_norm_t(
     TensorDim({1u, 1u, 1u, (unsigned int)clip_weights.size()}));
   float *global_norm_data = global_norm_t.getData();
   for (unsigned int idx = 0; idx < clip_weights.size(); idx++) {
     auto const &w = clip_weights[idx];
+
+    if (loss_scale != 0.0f) {
+      w->applyScaler(loss_scale);
+      if (w->getGradient().checkDataValidation(false) == false) {
+        float scale = loss_scale - 0.5f;
```
jijoongmoon, line 473: I think 0.5f @here and 2.0f @L496 can be hyper-param.

nntrainer/graph/network_graph.cpp
```diff
@@ -426,14 +436,51 @@ void NetworkGraph::backwarding(
   if (clip_weights.empty())
     return;
 
+  // update loss scale
```
jijoongmoon, line 439: It might be worth mentioning that mixed-precision training needs to enable clip_weights. Well, this is not exactly the right comment but it is necessary to compute "apply gradient" after checking if the weight is NaN or Inf.

nntrainer/layers/bn_layer.cpp
```diff
@@ -175,31 +175,79 @@ void BatchNormalizationLayer::forwarding(RunLayerContext &context,
...
-
-    mu.multiply_i(momentum);
-    mu.add_i(t_reduced, 1 - momentum);
-
-    deviation.pow(2.0f, t_full);
-    t_full.average(axes_to_reduce, cvar);
-
-    var.multiply_i(momentum);
-    var.add_i(cvar, 1 - momentum);
+  const auto &in_type = input_.getDataType();
```
jijoongmoon, line 178: I wonder if it is different from #2462

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -56,7 +57,10 @@ static void col2im(const Tensor &col_matrix, const TensorDim &kdim,
                    const std::array<props::Stride, CONV2D_DIM> &mstride,
                    const std::array<props::Dilation, CONV2D_DIM> &dilation,
                    Tensor &image) {
-  auto [pt, pb, pl, pr] = padding;
+  auto pt = padding[0];
+  auto pb = padding[1];
+  auto pl = padding[2];
+  auto pr = padding[3];
```
jijoongmoon, line 63: Any reason to change?

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -84,32 +88,48 @@ static void col2im(const Tensor &col_matrix, const TensorDim &kdim,
...
+  };
+
+  if (image.getDataType() == nntrainer::Tdatatype::FP32) {
+    float val;
+    apply_data(&val);
+  }
+#ifdef ENABLE_FP16
+  else if (image.getDataType() == nntrainer::Tdatatype::FP16) {
+    _FP16 val;
+    apply_data(&val);
```
jijoongmoon, line 128: Looks Good!

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -409,26 +443,49 @@ void Conv2DLayer::forwarding(RunLayerContext &context, bool training) {
...
+  Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+
   Tensor &filter_kernel = context.getWeight(wt_idx[ConvParams::weight]);
+  Tensor &bias_kernel = context.getWeight(wt_idx[ConvParams::bias]);
+
+  auto &disable_bias = std::get<props::DisableBias>(*layer_impl_props);
+  bool enable_bias = disable_bias.empty() || disable_bias.get() == false;
+
+  const auto &in_type = input_.getDataType();
+  if (in_type == filter_kernel.getDataType()) {
```
jijoongmoon, line 472: I wonder if there is ```in_type != filter_kernel```.

nntrainer/layers/input_layer.h
```diff
@@ -82,7 +82,7 @@ class InputLayer : public Layer {
   /**
    * @copydoc Layer::supportInPlace()
    */
-  bool supportInPlace() const override { return true; }
+  bool supportInPlace() const override { return false; }
```
jijoongmoon, line 85: Any reason to set this false as a default?

nntrainer/layers/layer_context.cpp
```diff
@@ -326,6 +326,25 @@ Tensor &RunLayerContext::getOutgoingDerivative(unsigned int idx) {
   return getInputGrad(idx);
 }
 
+bool RunLayerContext::validateDerivatives() {
+  auto num_in = getNumInputs();
+  auto num_out = getNumOutputs();
+
+  for (unsigned int i = 0; i < num_in; ++i) {
```
jijoongmoon, line 333: If we are going to check this validationDerivatives in every layer, then we do not need to check input and output both.

nntrainer/layers/loss/cross_entropy_softmax_loss_layer.cpp
```diff
@@ -85,7 +85,14 @@ void CrossEntropySoftmaxLossLayer::calcDerivative(RunLayerContext &context) {
...
   // TODO: verify y and ret_derivative must not be same as loss layer is not
   // working in-place
-  ret.subtract(y2, ret_derivative);
+  if (ret.getDataType() != y2.getDataType()) {
+    ret.subtract(y2.clone(ret.getDataType()), ret_derivative);
+  } else {
+    ret.subtract(y2, ret_derivative);
+  }
+
+  applyLossScale(ret_derivative);
```
jijoongmoon, line 94: please leave the comment it is necessary to place here to enable mixed precision training.

nntrainer/layers/loss/mse_loss_layer.cpp
```diff
@@ -20,34 +21,76 @@ static constexpr size_t SINGLE_INOUT_IDX = 0;
...
 }
 
 void MSELossLayer::calcDerivative(RunLayerContext &context) {
   Tensor &ret_derivative = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
   const Tensor &y2 = context.getIncomingDerivative(SINGLE_INOUT_IDX);
   Tensor &y = context.getInput(SINGLE_INOUT_IDX);
 
-  y.subtract(y2, ret_derivative);
+  const auto &in_type = y.getDataType();
+  if (in_type != y2.getDataType()) {
```
jijoongmoon, line 68: I think it should be always ```in_type == y2.getDataType()```.

nntrainer/layers/lstm.cpp
```diff
@@ -625,91 +625,37 @@ void LSTMLayer::exportTo(Exporter &exporter,
...
-  const bool return_sequences =
-    std::get<props::ReturnSequences>(lstm_props).get();
-  const bool bidirectional = std::get<props::Bidirectional>(lstm_props).get();
-  const float dropout_rate = std::get<props::DropOutRate>(lstm_props).get();
-  const unsigned int max_timestep =
-    std::get<props::MaxTimestep>(lstm_props).get();
-
-  const unsigned int bidirectional_constant = bidirectional ? 2 : 1;
-  bool enable_dropout = dropout_rate > epsilon && training;
+static inline void forwarding_internal(
```
jijoongmoon, line 628: missing template?

nntrainer/layers/lstm.cpp
```diff
@@ -788,23 +730,205 @@ void LSTMLayer::forwarding(RunLayerContext &context, bool training) {
...
+    bidirectional ? context.getTensor(wt_idx[LSTMParams::reverse_hidden_state])
+                  : empty;
+  Tensor &reverse_cell_state =
+    bidirectional ? context.getTensor(wt_idx[LSTMParams::reverse_cell_state])
+                  : empty;
+  Tensor &reverse_ifgo =
+    bidirectional ? context.getTensor(wt_idx[LSTMParams::reverse_ifgo]) : empty;
+
+  const auto in_type = input.getDataType();
+  if (in_type != weight_ih.getDataType()) {
```
jijoongmoon, line 803: I think this should be the same always.

nntrainer/layers/lstm.cpp
```diff
@@ -864,57 +988,113 @@ void LSTMLayer::calcGradient(RunLayerContext &context) {
...
+      : empty;
+
+  const Tensor &reverse_ifgo =
+    bidirectional ? context.getTensor(wt_idx[LSTMParams::reverse_ifgo]) : empty;
+  Tensor &reverse_d_ifgo =
+    bidirectional ? context.getTensorGrad(wt_idx[LSTMParams::reverse_ifgo])
+                  : empty;
+
+  const auto in_type = input.getDataType();
+  if (in_type != d_weight_ih.getDataType()) {
```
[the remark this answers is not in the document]
jihochu: There are two options to keep the converted W16 weight for backwarding.
1) Globally maintaining copies of master W32.
2) Keeping them in the layer from forwarding to backwarding.

For maintaining, 1) would be better while blocking from mistakes.
2) would be a little better for memory usage.(with considering inference mode)

I think 1) would be a little better, but please share your opinions.

nntrainer/layers/layer_node.h
```diff
@@ -889,6 +889,11 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    */
   bool needsCalcGradient() { return needs_calc_gradient; }
 
+  /**
+   * @brief Set loss scale factor
+   */
+  void setLossSacle(float scale) { layer->setLossSacle(scale); }
```
djeong20, line 895:
```suggestion
  void setLossScale(float scale) { layer->setLossScale(scale); }
```

nntrainer/tensor/tensor.h
```diff
@@ -1588,6 +1588,20 @@ class Tensor {
     dim.setFormat(fm);
   }
 
+  /**
+   * @brief     set the memory format
+   * @param     fm format of Tensor
+   */
+  void convertDataType(TensorDim::DataType type) {
```
djeong20, line 1595: Where is this function used? (cannot find it)
We have a similar function `setDataType()` but will be deprecated. (please check the details in #2498)
If this functionality is essential, we have to consider another way to enable this after refactorization. (e.g., delete FloatTensor and recreate HalfTensor)

nntrainer/graph/network_graph.cpp
jihochu: fixed

nntrainer/layers/input_layer.h
jihochu: It's for checking dut to input layer's inplace check had a bug.
please chekc modified input layer check code.

nntrainer/layers/loss/cross_entropy_softmax_loss_layer.cpp
jihochu: comment is append.

nntrainer/graph/network_graph.cpp
```diff
@@ -85,6 +85,15 @@ int NetworkGraph::compile(const std::string &loss_type) {
   status = checkCompiledGraph();
   NN_RETURN_STATUS();
 
+  /**
+   * @note It can be integrated with addLossLayer method
+   * if it removes adding loss layer to the model directly.
+   */
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    auto &ln = *iter;
+    ln->setLossSacle(loss_scale);
```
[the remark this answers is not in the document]
jihochu: Nice catch!! thanks.

nntrainer/tensor/tensor.h
jihochu: Thanks. it was test code.
It removed now.

nntrainer/tensor/manager.h
```diff
@@ -191,14 +193,16 @@ class Manager {
...
+   * otherwise NONE. original source. if not shared pass empty vector
    *
    * @return created weights list
    */
   std::vector<Weight *>
   requestWeights(const GraphNode &node,
                  const std::vector<Weight::Spec> &weights_spec, bool trainable,
-                 const std::vector<std::string> &shared_names);
+                 const std::vector<std::string> &shared_names,
+                 TensorDim::DataType act_type = TensorDim::DataType::NONE);
```
djeong20, line 205: I don't quite understand why a `NONE` data type is added to check that the activation data type is the same as the weight.
The normal activation data type can be passed in while it is the same as weight.
Are there other use cases of `DataTye::NONE` besides this function or have a critical reason to add the NONE data type?

nntrainer/tensor/manager.cpp
```diff
@@ -416,28 +422,45 @@ std::vector<Weight *> Manager::requestWeights(
...
+    Tensor *var = nullptr, *grad = nullptr, *var_m = nullptr;
     bool is_dependent = !shared_names.empty();
+    TensorDim dim_a = dim;
     if (is_dependent) {
       /// shared_name is used and the orignal name is discarded
       const auto &shared_name = shared_names.at(i);
       /** case when shared names are given */
-      var = weight_pool.requestOrExtend(shared_name, dim, var_exec_order,
-                                        var_ls, t_initializer);
+      if (act_type == TensorDim::DataType::NONE) {
```
djeong20, line 432: can't this be replaced as follows?
```suggestion
      if (act_type == dim.getDataType()) {
```

nntrainer/layers/acti_func.h
```diff
@@ -154,11 +154,13 @@ class ActiFunc {
     unsigned int bch_size = input.getDim().getDataLen() / width;
 
     // copy will not executed in inplace case
+    // output.copyData(input);
     output.copy(input);
 
     T *output_data = output.getData<T>();
 
     // prevent overflow
+    // Tensor tmp(width, output.getTensorType());
```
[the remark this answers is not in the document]
jihochu: thanks. fixed.

nntrainer/layers/bn_layer.cpp
```diff
@@ -111,6 +111,12 @@ void BatchNormalizationLayer::finalize(InitLayerContext &context) {
     context.requestWeight(dim, bnparams_beta, WeightRegularizer::NONE, 1.0f,
                           bias_decay, "beta", true);
 
+  /**
+   * @note declare weith dimention with activation datatype
```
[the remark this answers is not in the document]
jihochu: :+1:

nntrainer/tensor/manager.cpp
jihochu: it used to check "not specified" state. so, it's necessary.

nntrainer/tensor/manager.h
jihochu: It is used only when the master weight is necessary. It's used  to checks activation type is different, and generates master weight.

myungjoo:
Recommentation:

Keep a PR with every related commits as a test basis and mark it "Do Not Merge" or "Draft PR"
Make a number of independent smaller PRs ("sub-PR of this PR") so that reviewers can actually read and understand, removing them from the "full PR", whenever they are merged.

nntrainer/layers/layer_node.cpp
```diff
@@ -464,16 +465,24 @@ void LayerNode::read(std::ifstream &file, bool opt_var) {
     for (unsigned int i = 0; i < run_context->getNumWeights(); ++i) {
       if (run_context->isGradientLastAccess(i) && getTrainable()) {
         /// @note read optimizer variables
+        auto num_w_opt_m = run_context->getNumWeightOptMasterVar(i);
         for (unsigned int j = 0; j < run_context->getNumWeightOptVar(i); ++j) {
-          run_context->getWeightOptVar(i, j).read(file);
+          if (num_w_opt_m > 0)
+            run_context->getWeightOptMasterVar(i, j).read(file);
+          else
+            run_context->getWeightOptVar(i, j).read(file);
```
jijoongmoon, line 473: This needs to be reversed. The base model data need to be saved in FP16. Not the FP32. We could read FP16 and save it to Master Wegith.

nntrainer/layers/lstm.cpp
```diff
@@ -509,21 +509,27 @@ void LSTMLayer::finalize(InitLayerContext &context) {
   }
 
   // hidden_state_dim : [ batch_size, 1, max_timestep, unit ]
-  const TensorDim hidden_state_dim(batch_size, 1, max_timestep, unit,
-                                   weight_tensor_type);
+  TensorDim hidden_state_dim(batch_size, 1, max_timestep, unit,
+                             weight_tensor_type);
+  hidden_state_dim.setDataType(context.getActivationDataType());
```
jijoongmoon, line 514: This could be just ```TensorDim hidden_state_dim(batch_size, 1, max_timestep, unit, context.getActivationDataType())```.

nntrainer/tensor/manager.cpp
```diff
@@ -148,13 +147,20 @@ void Manager::reinitialize() {
...
   if (!weight_pool.isAllocated()) {
     finalizeTensorPool(weight_pool, 0, max_exec_order_);
     weight_pool.allocate();
   }
 }
 
-void Manager::deallocateWeights() { weight_pool.deallocate(); }
+void Manager::deallocateWeights() {
+  weight_pool.deallocate();
+  weight_master_pool.deallocate();
```
jijoongmoon, line 162: I don't think we need a separate pool for the weight master.

nntrainer/tensor/manager.cpp
```diff
@@ -416,28 +423,45 @@ std::vector<Weight *> Manager::requestWeights(
...
-      var = weight_pool.requestOrExtend(shared_name, dim, var_exec_order,
-                                        var_ls, t_initializer);
+      if (act_type == TensorDim::DataType::NONE) {
+        var = weight_pool.requestOrExtend(shared_name, dim_a, var_exec_order,
+                                          var_ls, t_initializer);
+      } else {
+        dim_a.setDataType(act_type);
+        var = weight_pool.requestOrExtend(shared_name, dim_a, var_exec_order,
+                                          var_ls, t_initializer);
+        var_m = weight_master_pool.requestOrExtend(
```
jijoongmoon, line 440: I think tensor pool can manage if we just request to weight_pool.

nntrainer/tensor/manager.cpp
```diff
@@ -416,28 +423,45 @@ std::vector<Weight *> Manager::requestWeights(
...
-      var =
-        weight_pool.request(name, dim, var_exec_order, var_ls, t_initializer);
+      if (act_type == TensorDim::DataType::NONE) {
+        var = weight_pool.request(name, dim_a, var_exec_order, var_ls,
+                                  t_initializer);
+      } else {
+        dim_a.setDataType(act_type);
+        var = weight_pool.request(name, dim_a, var_exec_order, var_ls,
+                                  t_initializer);
+        var_m = weight_master_pool.request(name, dim, var_exec_order, var_ls,
```
jijoongmoon, line 462: The execution order of var_m should be ```applyGradient_order``` only.

nntrainer/graph/network_graph.cpp
```diff
@@ -353,10 +363,15 @@ sharedConstTensors NetworkGraph::forwarding(
   bool training,
   std::function<void(std::shared_ptr<LayerNode>, bool)> forwarding_op,
   std::function<bool(void *userdata)> stop_cb, void *userdata) {
+
+  for (auto w : clip_weights) {
```
jijoongmoon, line 367: I wonder if we have to enable gradient clip property also true to use mixed precision training.  I guess, this PR doesn't consider the case which enabled mixed + gradient clip.

jijoongmoon:
closed by  #2663

PR #2462, opened 2024-02-06
[ layer ] Mixed precision forwarding / backwarding for bn layer @open sesame 03/07 10:42

skykongkong8:
- According to recent papers in mixed precision, when it comes to computation of statistics in bn layer, we should use fp32 values
- input(fp16) <-> BN_layer(fp16 weight, but compute with fp32 and re-cast to fp16) <-> output(fp16)
- Although we need a bulky code block for this, I believe we can revisit here for cleaner code when TensorV2 becomes official

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -174,32 +174,125 @@ void BatchNormalizationLayer::forwarding(RunLayerContext &context,
   /** use hidden_ as temporary tensor before setting the result in hidden */
   Tensor t_full = hidden_;
   Tensor &cvar = context.getTensor(wt_idx[BNParams::cvar]);
+  if (input_.getDataType() == ml::train::TensorDim::DataType::FP16) {
+#ifdef ENABLE_FP16
+    TensorDim mu_dim = mu.getDim();
+    mu_dim.setDataType(ml::train::TensorDim::DataType::FP32);
+    Tensor mu32(mu_dim, true);
+    mu32.copyData(mu);
 
-  if (training) {
-    input_.average(axes_to_reduce, t_reduced);
-    input_.subtract(t_reduced, deviation);
+    TensorDim var_dim = var.getDim();
+    var_dim.setDataType(ml::train::TensorDim::DataType::FP32);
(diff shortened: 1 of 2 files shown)


nntrainer/layers/bn_layer.cpp
```diff
@@ -213,42 +306,151 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
 
   Tensor &t_reduced = context.getTensor(wt_idx[BNParams::t_reduced]);
   Tensor &t_full = context.getTensor(wt_idx[BNParams::t_full]);
+  if (deriv.getDataType() == ml::train::TensorDim::DataType::FP16) {
+#ifdef ENABLE_FP16
+    TensorDim gamma_dim = gamma.getDim();
```
djeong20, line 311: same here

skykongkong8: Applied!

baek2sm (approved):
LGTM

nntrainer/layers/bn_layer.cpp
```diff
@@ -213,42 +267,105 @@ void BatchNormalizationLayer::calcDerivative(RunLayerContext &context) {
...
+    deviation32.multiply(deriv32, t_full32);
+    t_full32.average(axes_to_reduce, t_reduced32);
+    t_reduced32.divide_i(cvar32);
+    deviation32.multiply_i(t_reduced32);
+
+    if (context.getTrainable()) {
+      /**
+       * This calculates dgamma tensor.
+       */
+      Tensor dgamma = context.getWeightGrad(wt_idx[BNParams::gamma]);
```
EunjuYang, line 290: is it `Tensor &dgamma = context.getWeightGrad(wt_idx[BNParams::gamma]);` ?

SeoHyungjun (approved):
LGTM

jijoongmoon (changes requested):
I dont think this is the right way to enable the mixed precision. please consider #2455

skykongkong8:
This PR is no longer needed. Close.

PR #2466, opened 2024-02-08
[Test] TensorV2 division unit test

djeong20:
This PR includes unit tests for verifying divide operation in the TensorV2 class.
Note that all unit tests are the same as tests in unittest_nntrainer_tensor.cpp file.

Self-evaluation:

### DIFF
--- a/test/unittest/unittest_nntrainer_tensor_v2.cpp
+++ b/test/unittest/unittest_nntrainer_tensor_v2.cpp
@@ -893,6 +893,502 @@ TEST(nntrainer_Tensor, multiply_strided_06_p) {
   EXPECT_EQ(status, ML_ERROR_NONE);
 }
 
+TEST(nntrainer_Tensor, divide_i_01_p) {
+  int status = ML_ERROR_NONE;
+  int batch = 3;
+  int channel = 1;
+  int height = 3;
+  int width = 10;
+
+  nntrainer::TensorV2 input(batch, channel, height, width);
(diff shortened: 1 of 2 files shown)


test/unittest/unittest_nntrainer_tensor_v2.cpp
```diff
@@ -748,6 +748,502 @@ TEST(nntrainer_Tensor, multiply_float_01_p) {
...
+TEST(nntrainer_Tensor, divide_02_n) {
+  int batch = 3;
+  int channel = 1;
+  int height = 3;
+  int width = 10;
+
+  nntrainer::TensorV2 input(batch, channel, height, width);
+  GEN_TEST_INPUT(input, i * (batch * height) + j * (width) + k + 1);
+
+  EXPECT_THROW({ input.divide(0.0); }, std::invalid_argument);
```
skykongkong8, line 857: Think it is great to check zero-division error, why won't we add TC for the Tensor case as well?
e.g.) `normal_tensor.divide(tensor_with_zero_value)`

skykongkong8 (approved):

test/unittest/unittest_nntrainer_tensor_v2.cpp
djeong20: the current tensor division allows dividing with a tensor with a zero value, which outputs an inf value. 
on the other hand, PyTorch and TensorFlow allow division by zero for both cases. 
there's an inconsistency in NNTrainer, so we should double-check on this.

myungjoo (approved):

PR #2468, opened 2024-02-08
[ bug ] Fix meson argument

skykongkong8:
- nntrainer_version_split is already a list of splitted strings, no need to join

**Self evaluation:**


lhs8928 (approved):
LGTM

djeong20 (approved):
To clarify, `nntrainer_version_split` is a list of strings, which means `nntrainer_version_split[0]` is a string so there's no need to join. is this correct?

skykongkong8:
> To clarify, `nntrainer_version_split` is a list of strings, which means `nntrainer_version_split[0]` is a string so there's no need to join. is this correct?

That is correct. Thanks for the clarification

jijoongmoon (approved):
Nice catch! LGTM

PR #2469, opened 2024-02-08
[TensorV2] Subtraction support @open sesame 02/20 11:30

djeong20:
This PR adds support for performing the subtraction operation on two tensors.

**Changes proposed**
- TensorV2 includes member functions to perform tensor subtraction.
- FloatTensor and HalfTensor take TensorV2 as input/output to perform subtraction.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -503,7 +503,8 @@ TensorV2 &FloatTensor::add(TensorV2 const &m, TensorV2 &output,
 }
 
 TensorV2 &FloatTensor::subtract(float const &value, TensorV2 &output) const {
-  throw std::logic_error("FloatTensor::subtract is not implemented yet");
+  auto f = std::bind(std::minus<float>(), std::placeholders::_1, value);
+  apply(f, output);
   return output;
 }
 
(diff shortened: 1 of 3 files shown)


skykongkong8 (approved):
LGTM!

nntrainer/tensor/float_tensor.cpp
```diff
@@ -503,7 +503,8 @@ TensorV2 &FloatTensor::add(TensorV2 const &m, TensorV2 &output,
 }
 
 TensorV2 &FloatTensor::subtract(float const &value, TensorV2 &output) const {
-  throw std::logic_error("FloatTensor::subtract is not implemented yet");
+  auto f = std::bind(std::minus<float>(), std::placeholders::_1, value);
+  apply(f, output);
```
jijoongmoon, line 507: Need to compare the performance. Inside of apply, we use std::transform. We need to check if it is faster than using SIMD Accelerator. If it is not, then it might be better to implement it.

jihochu (approved):
LGTM

nntrainer/tensor/float_tensor.cpp
djeong20: we are currently using std::transform for every main arithmetic operation. how should we compare with using a SIMD accelerator if there isn't any?

DonghakPark (approved):
LGTM!

PR #2470, opened 2024-02-08
[ rename/neon ] Fix neon-related functions name

skykongkong8:
- Every neon function resides in nntrainer::neon namespace, no need to be differentiated
- 's' in BLAS functions indicates the function works with single-precision (fp32). Incase of half-precision functions we should use 'h' instead of 's'
- In case of previous functions like : 'scopy_fp32_to_fp16', the function name itself shows the dataType it gets, so no need to add neither 's' nor 'h'
- There are more rooms for this, but still WIP ( e.g. `blas_interface.h` )

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -83,7 +83,7 @@ static void saxpy_FP16(const unsigned int N, const float alpha, const _FP16 *X,
 #if (defined USE__FP16 && USE_NEON)
   // USE__FP16 is defined when platform is android
   if (incX == 1 && incY == 1) {
-    nntrainer::neon::saxpy_neon_fp16(N, alpha, X, Y);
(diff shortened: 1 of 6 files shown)


lhs8928 (approved):
Although we can check the description on header file it would be nice to know the function meaning by its name so how about we name as like el_vec_add?
Or how about we remove prefix "elementwise"?
overall LGTM

skykongkong8:
> Although we can check the description on header file it would be nice to know the function meaning by its name so how about we name as like el_vec_add? Or how about we remove prefix "elementwise"? overall LGTM

Think it would be better to do it while fixing function names in `blas_interface.h`. This requires some trivial code fixes in `tensor.cpp` as well, so will do it in the next PR

jijoongmoon (approved):
LGTM

djeong20 (approved):
LGTM

PR #2473, opened 2024-02-15
[TensorV2] Add strided inputs and output

djeong20:
This PR adds the implementation of add operation that supports different strided inputs and output.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -326,7 +326,7 @@ TensorV2 FloatTensor::multiply_strided(TensorV2 const &m, TensorV2 &output,
     << output.getName() << " is not allocated";
 
   if (strides[3] != 1 || m.getStrides()[3] != 1 ||
-      output.getStrides()[3] != 1 || beta != 0.0) {
+      output.getStrides()[3] != 1 || std::fpclassify(beta) != FP_ZERO) {
     for (unsigned int b = 0; b < batch(); ++b) {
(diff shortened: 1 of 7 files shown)


skykongkong8 (approved):
LGTM!

nntrainer/tensor/float_tensor.cpp
```diff
@@ -476,6 +476,60 @@ TensorV2 &FloatTensor::divide(TensorV2 const &m, TensorV2 &output) const {
...
+                                   const float beta) const {
+  NNTR_THROW_IF(getData() == nullptr, std::invalid_argument)
+    << getName() << " is not allocated";
+  NNTR_THROW_IF(input.getData<float>() == nullptr, std::invalid_argument)
+    << input.getName() << " is not allocated";
+  NNTR_THROW_IF(output.getData<float>() == nullptr, std::invalid_argument)
+    << output.getName() << " is not allocated";
+
+  if (strides[3] != 1 || input.getStrides()[3] != 1 ||
+      output.getStrides()[3] != 1 || beta != 0.0) {
```
myungjoo, line 489: Are you really sure it's ok to use ```beta != 0.0``` with ```float beta```? 

Reference: https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison

jihochu (approved):
LGTM

nntrainer/tensor/float_tensor.cpp
djeong20: Good point. Fixed :)

PR #2474, opened 2024-02-19
[ BLAS ] Implement elementwise operations

skykongkong8:
Implemented elementwise vector operations
* ele_add (previously `ewva`)
* ele_sub
* ele_mul (previously `ewvm`)
* ele_div
with fp32 / fp16 precisions
for current BLAS, implement with raw version
for NEON SIMD, implement with SIMD version

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -387,7 +387,7 @@ void ele_mul(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
   nntrainer::neon::ele_mul(N, X, Y, Z, alpha, beta);
 #else
   for (unsigned int i = 0; i < N; ++i) {
-    if (beta != 0.f)
+    if (std::abs(beta) > __FLT_MIN__)
       Z[i] = static_cast<_FP16>(alpha) * X[i] * Y[i] +
              static_cast<_FP16>(beta) * Z[i];
     else
@@ -402,7 +402,7 @@ void ele_add(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
   nntrainer::neon::ele_add(N, X, Y, Z, alpha, beta);
 #else
   for (unsigned int i = 0; i < N; ++i) {
-    if (beta != 0.f)
+    if (std::abs(beta) > __FLT_MIN__)
       Z[i] = X[i] + static_cast<_FP16>(alpha) * Y[i] +
              static_cast<_FP16>(beta) * Z[i];
     else
@@ -417,7 +417,7 @@ void ele_sub(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
   nntrainer::neon::ele_sub(N, X, Y, Z, alpha, beta);
 #else
   for (unsigned int i = 0; i < N; ++i) {
-    if (beta != 0.f)
+    if (std::abs(beta) > __FLT_MIN__)
       Z[i] = X[i] - static_cast<_FP16>(alpha) * Y[i] +
              static_cast<_FP16>(beta) * Z[i];
     else
@@ -432,7 +432,7 @@ void ele_div(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
   nntrainer::neon::ele_div(N, X, Y, Z, alpha, beta);
 #else
   for (unsigned int i = 0; i < N; ++i) {
-    if (beta != 0.f)
+    if (std::abs(beta) > __FLT_MIN__)
       Z[i] = X[i] / (static_cast<_FP16>(alpha) * Y[i]) +
              static_cast<_FP16>(beta) * Z[i];
     else
@@ -943,7 +943,7 @@ void ele_mul(const unsigned int N, const float *X, const float *Y, float *Z,
   nntrainer::neon::ele_mul(N, X, Y, Z, alpha, beta);
 #else
   for (unsigned int i = 0; i < N; ++i) {
-    if (beta != 0.f)
+    if (std::abs(beta) > __FLT_MIN__)
       Z[i] = alpha * X[i] * Y[i] + beta * Z[i];
     else
       Z[i] = alpha * X[i] * Y[i];
@@ -957,7 +957,7 @@ void ele_add(const unsigned int N, const float *X, const float *Y, float *Z,
   nntrainer::neon::ele_add(N, X, Y, Z, alpha, beta);
 #else
   for (unsigned int i = 0; i < N; ++i) {
-    if (beta != 0.f)
+    if (std::abs(beta) > __FLT_MIN__)
       Z[i] = X[i] + alpha * Y[i] + beta * Z[i];
     else
       Z[i] = X[i] + alpha * Y[i];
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/blas_interface.cpp
```diff
@@ -400,12 +381,64 @@ void scopy_int8_to_float16(const unsigned int N, const uint8_t *X,
   copy_int8_to_fp16(N, X, incX, Y, incY);
 }
 
-void ewvm(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z) {
-  ewvm_FP16(N, X, Y, Z);
+void ele_mul(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
```
djeong20, line 384:
```suggestion
void elementwise_multiply(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
```
one suggestion. how about renaming it into something more clear?

nntrainer/tensor/blas_interface.cpp
```diff
@@ -400,12 +381,64 @@ void scopy_int8_to_float16(const unsigned int N, const uint8_t *X,
...
+             float alpha, float beta) {
+#if (defined USE__FP16 && USE_NEON)
+  nntrainer::neon::ele_mul(N, X, Y, Z, alpha, beta);
+#else
+  for (unsigned int i = 0; i < N; ++i) {
+    if (std::abs(beta) > __FLT_MIN__)
+      Z[i] = static_cast<_FP16>(alpha) * X[i] * Y[i] +
+             static_cast<_FP16>(beta) * Z[i];
+    else
+      Z[i] = static_cast<_FP16>(alpha) * X[i] * Y[i];
```
djeong20, line 394: any reason to differentiate beta == 0 case?
```suggestion
      Z[i] = static_cast<_FP16>(alpha) * X[i] * Y[i] + static_cast<_FP16>(beta) * Z[i];
```

nntrainer/tensor/blas_interface.h
```diff
@@ -152,22 +152,56 @@ void sgemv(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, const unsigned int M,
...
  * @param[in] N  length of the vector
- * @param[in] X __fp16 * for Vector X
- * @param[in] Y __fp16 * for Vector Y
- * @param[in] Z __fp16 * for Vector Z
- */
-void ewvm(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z);
+ * @param[in] X _FP16 * for Vector X
+ * @param[in] Y _FP16 * for Vector Y
+ * @param[in] Z _FP16 * for Vector Z
+ * @param[in] alpha scalar multiplier for input
```
djeong20, line 161: quick question. it seems like scalars are added only for Y and Z. wouldn't there be cases where X also needs a scalar?

djeong20: also, why is scalar added in the first place?

nntrainer/tensor/tensor.cpp
```diff
@@ -1076,7 +1076,7 @@ Tensor &Tensor::add(Tensor const &m, Tensor &output, float const alpha) const {
     auto f = [&](const BroadcastInfo &e, const float *buf, const float *m_buf,
                  float *out_buf) {
       if (e.strides[3] == 1 && strides[3] == 1 && strides[3] == 1 &&
-          alpha == 0) {
+          alpha == 1.f) {
```
djeong20, line 1079: could you explain why this condition has changed?

djeong20 (approved):
Overall, looks good to me. There are a few things to point out.
1. we should validate this code by replacing the current implementation of std::transform in the Tensor class with this, especially `divide()`.
2. after 1, we should check the performance difference for non-neon cases. in other words, check the effect of replacing std::transform() to for loop.

skykongkong8:
> Overall, looks good to me. There are a few things to point out.
> 
> 1. we should validate this code by replacing the current implementation of std::transform in the Tensor class with this, especially `divide()`.
> 2. after 1, we should check the performance difference for non-neon cases. in other words, check the effect of replacing std::transform() to for loop.
> 
> + please resolve the conflict!

I already conducted this with tensor unittests, and discovered that:
1) Although we use SIMD, the computational efficiency does not increase linearly (maybe because of loop-unrolling ? -> currently trying to clarify the reason)
2) latency in `std::transform()` is almost similar to for-loop. Maybe we can use it only for non-contiguous case only in later fixes.
And yes, I had some discussions with @jijoongmoon offline, considering when to apply for Tensor operations.

nntrainer/tensor/blas_interface.cpp
skykongkong8: I asked for many contributors' opinions offline, but shorter function name is quite preferred.
ele_* sounds clear enough for me.

nntrainer/tensor/blas_interface.cpp
skykongkong8: what if Z[i] != Z[i] ? (NaN in Z[i])

nntrainer/tensor/blas_interface.h
skykongkong8: 1. Good point. Ususally, strict BLAS functions only get X and Y, not Z. So it would go like : $X = X * \alpha (op) Y * \beta$. But current function usage in the NNTrainer requires all $X, Y, Z$  . 
2. from tensor operation, we differentiate whether to use std::transform or ele_* by stride and alpha / beta. For more broader use of this function. scalar multiplier is necessary.

nntrainer/tensor/tensor.cpp
skykongkong8: We have been using this code wrong.
`alpha` is an input scalar multiplier, so considering our total formula,

$$X = X  (op) Y * \alpha + Z * \beta$$

alpha == 0 is a nonsense. isn't it?

nntrainer/tensor/blas_interface.cpp
djeong20: regardless of beta, wouldn't it cause an error anyway?

djeong20: if beta is zero, Z = X * Y + beta * Z would be Z = X * Y.
if not, Z = X * Y + beta * Z.

nntrainer/tensor/blas_interface.h
djeong20: thanks for clarification :)

nntrainer/tensor/blas_interface.cpp
skykongkong8: isn't NaN * 0 = NaN ?

nntrainer/tensor/tensor.cpp
skykongkong8: As far as I am concerned, that is correct

nntrainer/tensor/blas_interface.cpp
djeong20: yeah, what I mean is having beta != 0 condition to avoid NaN or uninitialized error seems offbeat.

djeong20: maybe we could come up with a different way to handle these cases (e.g., check if tensor is initialized when using beta).

skykongkong8: Sounds more reasonable.
will try to fix it during resolving conflicts

nntrainer/tensor/tensor.cpp
jijoongmoon: Nice catch!

jijoongmoon (approved):
LGTM!

PR #2475, opened 2024-02-19
[TensorV2] Add Transpose Support for Tensor

djeong20:
This pull request adds the ability to transpose tensors, which allows users to flip or rotate their data along its axes.

- Added a new method called transpose() to the TensorV2 class, which takes direction argument: the axes to be swapped.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/half_tensor.cpp
+++ b/nntrainer/tensor/half_tensor.cpp
@@ -370,32 +370,6 @@ TensorV2 HalfTensor::multiply_strided(TensorV2 const &m, TensorV2 &output,
   return output;
 }
 
-void HalfTensor::copy(const TensorV2 &from) {
-  reshape(from.getDim());
-  copy(from.getData<_FP16>());
-}
-
-void HalfTensor::copyData(const TensorV2 &from) {
-  if (!contiguous) {
-    throw std::runtime_error("Cannot copy non-contiguous tensor");
-  }
-
-  if (size() != from.size())
-    throw std::invalid_argument("Size of tensor to copy must match");
-
-  switch (from.getDataType()) {
-  case ml::train::TensorDim::DataType::FP32:
-    scopy(size(), from.getData<float>(), 1, (_FP16 *)getData(), 1);
-    break;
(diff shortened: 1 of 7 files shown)


nntrainer/tensor/half_tensor.cpp
```diff
@@ -565,6 +539,98 @@ TensorV2 &HalfTensor::divide(TensorV2 const &m, TensorV2 &output) const {
...
+  if (!contiguous) {
+    throw std::runtime_error("Cannot copy non-contiguous tensor");
+  }
+
+  if (size() != from.size())
+    throw std::invalid_argument("Size of tensor to copy must match");
+
+  switch (from.getDataType()) {
+  case ml::train::TensorDim::DataType::FP32:
+    scopy(size(), from.getData<float>(), 1, (_FP16 *)getData(), 1);
```
skykongkong8, line 557: Using swich-case instead of previous if-else seems much better!

skykongkong8 (approved):

DonghakPark (approved):
LGTM!

jihochu (approved):
LGTM

nntrainer/tensor/tensor_base.h
```diff
@@ -21,6 +21,34 @@
 #include <tensor_dim.h>
 #include <util_func.h>
 
+#define transposeloop(cl, ci, cj, ck, sl, si, sj, sk)                 \
```
jijoongmoon, line 24: I'm not sure if it is working with omp directives in define. If it can, then we could try to apply it.

jijoongmoon (approved):
LGTM except minor comment.

djeong20:
Currently, the TAOS/CI clang-format version differs from Git action static checks. there's a mismatch that causes clang-format style error (the behavior of the clang-format varies for different versions). It is ok to merge.

nntrainer/tensor/tensor_base.h
djeong20: agree. I think we should test it after when openMP is enabled. I'll keep it in mind and come back to this later.

PR #2479, opened 2024-02-20
[ util ] Implement softmax calculation function in util

skykongkong8:
- Current activation functions are implemented as a function template, and fully computes with the function template parameter's precision unless using NEON intrinsics with inter-fp32-values explicitly.
- According to current papers, for safe convergence of mixed precision training, it is quite critical to calculate softmax with fp32 precision.
- This PR proposes a SIMD version of softmax calculation, and uses temporally higher precision when using half-precision
- For mathematical stability, applied linear translation (using minus values for the input of exponential function) to avoid precision overflow

### DIFF
--- a/nntrainer/utils/util_simd.cpp
+++ b/nntrainer/utils/util_simd.cpp
@@ -44,13 +44,8 @@ float max(const unsigned int N, float *X) {
 #ifdef USE_NEON
   return nntrainer::neon::max(N, X);
 #else
-  float ret = X[0];
-  unsigned int i = 1;
-  while (i < N) {
-    ret = std::fmax(ret, X[i]);
-    ++i;
-  }
-  return ret;
+  std::vector<float> v(X, X + N);
+  return *std::max_element(v.begin(), v.end());
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


djeong20 (approved):
LGTM

nntrainer/utils/util_simd.cpp
```diff
@@ -40,6 +40,39 @@ void swish(const unsigned int N, float *X, float *Y, float *Z) {
 #endif
 }
 
+float max(const unsigned int N, float *X) {
+#ifdef USE_NEON
+  return nntrainer::neon::max(N, X);
+#else
```
myungjoo, line 46: Use STL properly

```C
std::vector<float> v (X, X+N);
return *std::max_element(v.begin(), v.end());
```

And if you compile it properly, you may get x64/SIMD (AVX/SSE) for free:

https://stackoverflow.com/questions/59373900/why-is-there-no-simd-functionality-in-the-c-standard-library

myungjoo (approved):

baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM

nntrainer/utils/util_simd.cpp
```diff
@@ -40,6 +40,34 @@ void swish(const unsigned int N, float *X, float *Y, float *Z) {
...
+}
+
+void softmax(const unsigned int N, float *X, float *Y) {
+#ifdef USE_NEON
+  nntrainer::neon::softmax(N, X, Y);
+#else
+  unsigned int i = 0;
+  float sum = 0.f;
+  float max_x = max(N, X);
+  while (i < N) {
```
jijoongmoon, line 59: Not in this PR, but you can also optimize further.
using omp and temporal buffer to save X[i] - max_x.

PR #2480, opened 2024-02-20
[AVX] Added VCVT (between floating-point and integer) AVX support for x86 architecture

djeong20:
This pull request adds VCVT (between floating-point and integer) AVX support for the x86 architecture. 
This feature allows efficient conversion between single-precision floating-point and half-precision floating-point data types using the AVX instruction set.

- Added new VCVT instructions to the blas_interface
- Fix incorrect meson.build logic in checking gcc version and AVX support
- Add enable-avx option to enable AVX hardware acceleration

**Self-evaluation:**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -80,13 +80,16 @@ if get_option('enable-fp16')
    elif arch == 'arm'
      error ('FP16/ARM code (blas_neon.cpp) uses armv8.2 instructions. armv7 is not supported.')
    else
-     has_avx512fp16 = cc.has_argument('-mavx512fp16')
-     if (has_avx512fp16)
-       # add_project_arguments(['-mavx512fp16'], language: ['c','cpp'])
-       message ('Float16 for x86_64 enabled. Modern gcc-x64 genrally supports float16 with _Float16. -mavx512fp16 added for hardware acceleration')
+     if cc.version().version_compare('>=12.1.0')
+       message ('Float16 for x86_64 enabled. Modern gcc-x64 generally supports float16 with _Float16.')
        extra_defines += '-DENABLE_FP16=1'
+       if get_option('enable-avx')
+        extra_defines += '-DUSE_AVX=1'
+        add_project_arguments(['-march=native'], language: ['c','cpp'])
+        message('-march=native added for AVX hardware acceleration.')
+       endif
      else
-       warning ('Float16 for x86_64 enabled. However, software emulation is applied for fp16, making it slower and inconsistent. Use GCC 12+ for AVX512 FP16 support. This build will probably fail unless you bring a compiler that supports fp16 for x64.')
+       warning ('Float16 for x86_64 enabled. However, software emulation is applied for fp16, making it slower and inconsistent. Use GCC 12+ for FP16 support. This build will probably fail unless you bring a compiler that supports fp16 for x64.')
      endif
    endif  
 endif
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -19,6 +19,10 @@
 #include <blas_neon.h>
 #endif
 
+#if USE_AVX
+#include <blas_avx.h>
+#endif
+
 #include <cmath>
 
 #define sgemv_loop(ci, cj, cM, cN)           \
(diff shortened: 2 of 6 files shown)


skykongkong8 (approved):
Nice!

nntrainer/tensor/blas_interface.cpp
```diff
@@ -182,6 +186,13 @@ static void copy_float32_to_float16(const unsigned int N, const float *X,
...
+#elif USE_AVX
+  if (incX == 1 && incY == 1) {
+    nntrainer::avx::vcvt_f32_f16(N, X, Y);
+  } else {
+    for (unsigned int i = 0; i < N; ++i)
+      Y[i * incy] = static_cast<_FP16>(X[i * incx]);
+  }
 #else
   for (unsigned int i = 0; i < N; ++i)
     Y[i * incy] = static_cast<_FP16>(X[i * incx]);
```
myungjoo, line 198: This is not for this PR or this month, but for later versions (you are now having technical debts here):

maintain the same code in general cpp files across NEON/AVX/None.
centralize architectural dependencies to a corresponding header and its implementation only.

You may have a single header for all three cases and choose a corresponding implementation (.c/.cpp file) at build time, too.

myungjoo: Try the better approach from the next implementation (and fix this when you are ready)

skykongkong8: You mean something like:

```C++
// new file : blas_raw.cpp
...
void foo( ... ) {
...
} else {
    for (unsigned int i = 0; i < N; ++i)
      Y[i * incy] = static_cast<_FP16>(X[i * incx]);
  }
...
}
...
```
and use it for all `blas_interface.cpp`, `blas_neon.cpp`, `blas_avx.cpp` ?
?

djeong20: Thank you for bringing this up! I will keep it in mind for future improvement.

myungjoo: Or.. you may have a base class as a set of operators/functions/methods for general CPU operations compatible for ALL systems with virtual functions and a derived class as a set of operators/functions/methods for architecture-dependent (SIMD) operations overriding the virtual functions.

Then, if you don't have an implementation in the derived class (SIMD), the fallback method in the base class will be automatically chosen. If you have an implementation in the derived class and the class is chosen at run-time/build-time, it will be automatically chosen WITHOUT having #if/#endif or even general if statements. You can control that in the base class initializer or "get_instance" method of the singleton design.

myungjoo: Such refactoring needs planning and design before implementation. Don't just start without enough discussion and I don't want this refacotring effort hinder the immenent release.

Anyway, for your reference: https://github.com/nnstreamer/nntrainer/discussions/2482

myungjoo: The concept suggestion : https://github.com/myungjoo/nntrainer/tree/suggestion/refactoring/archdep
(this is not for actual PR)

myungjoo (approved):

meson.build
```diff
@@ -80,13 +80,33 @@ if get_option('enable-fp16')
...
-       message ('Float16 for x86_64 enabled. Modern gcc-x64 genrally supports float16 with _Float16. -mavx512fp16 added for hardware acceleration')
+     if cc.version().version_compare('>=12.1.0')
+       message ('Float16 for x86_64 enabled. Modern gcc-x64 generally supports float16 with _Float16.')
        extra_defines += '-DENABLE_FP16=1'
+
+       # check AVX-512
+       avx512_code = '''#include <immintrin.h>
+       int main (int argc, char *argv[]) { const __m256i vec = {0, 1, 2, 3}; __m512 v = _mm512_cvtph_ps(vec); return 0; }
+       '''
+       has_avx512 = cc.compiles(avx512_code,  args : '-march=native', name: 'avx512 support')
```
myungjoo, line 91: You may proceed and fix it later, but...

Are you sure that the build machine is AVX512 capabile while the target machine is also AVX512 capable? As we build ARM binaries in X64 machines, we may build X64/AVX512 binaries in X64/non-AVX512 machines. We may need to force-enable them (and you will be required to apply proper arch name, in this case, -march=native won't work.)

The current meson script is fine for developers and most users. 
And the current script is fine for this release because the automated build system is not applied for x64 targets. (we do it only for arm targets) But, you will need to consider this afterwards.

djeong20: That makes sense. I thought that if AVX512 is not supported on the build machine, it should not be enabled in the first place. Since there are no AVX512 instructions in our code, I think we can modify it now without any problems.

nntrainer/tensor/blas_interface.cpp
djeong20: both approaches seem great. I'll also research how other frameworks manage this. let's keep the discussion going

PR #2481, opened 2024-02-22
[SPEC] support neon enable option

jihochu:
It supports enable_neon only when aarch64 architecture.
And, neon option can be selectable for aarch64.


DonghakPark (approved):
LGTM!

myungjoo (approved):

PR #2487, opened 2024-02-26
[TensorV2] Average Tensor element by axes @open sesame 03/07 10:42

djeong20:
This pull request adds new functions to the TensorV2 that allow users to average tensor elements along specified axes. 
The functions take in an axis or list of axes as input and return a new tensor with elements replaced by their corresponding means. If no axis is provided, it returns a tensor value by averaging the elements by all axes.

**Self-evaluation:**


skykongkong8 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2489, opened 2024-02-27
[TensorV2] Add utility member functions to TensorV2 class @open sesame 03/07 10:42

djeong20:
This pull request adds several new utility member functions to the TensorV2 class, enabling users to perform various tasks with their tensors more easily and efficiently.
These include saving and loading tensors, updating batches, getting argmax and max absolute values, and more.
The implementation is based on the current Tensor class and aims to improve the overall usability and flexibility of the TensorV2 class.

- Added `save()` and `read()` methods to allow saving and loading of saved tensor data.
- Added `Map()` method to create a new Tensor object from a buffer.
- Added `argmax()` and `max_abs()` methods to retrieve the indices of max value by batch and the value of the maximum absolute element in a tensor.
- Added `updateBatch()` to update tensor batch size.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor_base.cpp
+++ b/nntrainer/tensor/tensor_base.cpp
@@ -40,12 +40,27 @@ bool TensorBase::operator==(const TensorBase &rhs) const {
   return true;
 }
 
+void TensorBase::setTensorVar(TensorDim d, void *buf, size_t offset) {
+  dim = d;
+  strides = d.computeStrides();
+  /// Tensor does not own the memory
+  data = std::shared_ptr<MemoryData>(new MemoryData((void *)buf),
+                                     std::default_delete<MemoryData>());
+  offset = offset;
+}
+
 void TensorBase::putData() const {
   if (!data)
     return;
 
   data->invalidate();
 }
+
+const std::shared_ptr<MemoryData> TensorBase::getMemoryData() const {
+  return data;
+}
+
+size_t TensorBase::getOffset() const { return offset; }
 
 void TensorBase::reshape(const TensorDim &d) {
   NNTR_THROW_IF(!contiguous, std::invalid_argument)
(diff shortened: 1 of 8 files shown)


nntrainer/tensor/tensor_v2.h
```diff
@@ -245,6 +245,35 @@ class TensorV2 {
...
+  /**
+   * @brief Construct a new Tensor object from a buffer
+   * This will not copy buffer to a new tensor but directly uses it
+   *
+   * @param[in] buf buffer
+   * @param[in] bytes buffer size in bytes
+   * @param[in] d tensor dim
+   * @param[in] offset offset to be used from current
+   * @return    Tensor object
+   * @throws    std::invalid_argument if buf is null
```
jihochu, line 257: Adding a note that the buffer is not owned by the mapped tensor would be better.
the user may be confused.

jihochu (approved):
LGTM

skykongkong8 (approved):

jijoongmoon (approved):
LGTM

nntrainer/tensor/tensor_v2.h
djeong20: applied!

nntrainer/tensor/tensor_base.cpp
```diff
@@ -40,13 +40,28 @@ bool TensorBase::operator==(const TensorBase &rhs) const {
...
 
+void TensorBase::setTensorVar(TensorDim d, void *buf, size_t offset) {
+  dim = d;
+  strides = d.computeStrides();
+  /// Tensor does not own the memory
+  data = std::shared_ptr<MemoryData>(new MemoryData((void *)buf),
+                                     std::default_delete<MemoryData>());
+  offset = offset;
+}
+
```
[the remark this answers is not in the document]
djeong20: this is simply a helper function to set certain TensorBase variables in the Tensor class.
since Map() is a static function, the possible way to set TensorBase variables would be using a Tensor public function. 
therefore, the SetTensorVar() function has been newly added.

baek2sm (approved):
LGTM

PR #2491, opened 2024-02-27
[ DRAFT ] Apply ele_* ops to Tensor

skykongkong8:
- `apply_broadcast` function would throw error when the input Tensor is not allocated.
- Implement ele_mul_fallback for default option
- Use ele_mul instead of std::transform with std::multiplies
- Note that this is just a draft PR


baek2sm (approved):
LGTM

myungjoo (approved):
Looks better than the previous. We can apply a proper design soon!

jijoongmoon (approved):
LGTM

PR #2492, opened 2024-02-28
[util] Add numpy file reader

DonghakPark:
In deep learning data feeding, there are often tasks that involve reading and processing numpy (.npy) files.

Therefore, it would be much more convenient to develop by entering the file name and returning the data in a vector instead of manually writing code to read it every time.

Function Signature
```
void read_npy_file(const char *file_path, std::vector<int> &dims,
                   std::vector<float> &values);
```

**Self evaluation:**

### DIFF
--- a/Applications/utils/npy_reader/npy_reader.cpp
+++ b/Applications/utils/npy_reader/npy_reader.cpp
@@ -1,0 +1,156 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Donghak Park <donghak.park@samsung.com>
+ *
+ * @file   npy_reader.cpp
+ * @date   28 Feb 2024
+ * @brief  reader for npy file
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Donghak Park <donghak.park@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#include "npy_reader.h"
+
+#include <cstring>
+#include <iostream>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <random>
+#include <string>
+#include <vector>
+
+namespace nntrainer::util {
+void NpyReader::read_npy_file(const char *file_path) {
+  FILE *file = fopen(file_path, "rb");
+
+  char magic[7] = {};
+  try {
+    if (!file) {
+      throw std::runtime_error("Failed to open file");
+    }
+
+    int bytes_read = fread(magic, 6, 1, file);
+
+    if (bytes_read != 1 || strcmp(magic, "\x93NUMPY") != 0) {
+      fclose(file);
+      throw std::runtime_error("Failed : this file is not a numpy file");
+    }
+  } catch (const std::exception &e) {
+    std::cerr << "Error: " << e.what() << std::endl;
+    return;
+  }
+
+  char major = 0;
+  if (fread(&major, 1, 1, file) != 1) {
+    fclose(file);
+  };
+
+  char major = 0;
+  if (fread(&major, 1, 1, file) != 1) {
+    fclose(file);
+    throw std::runtime_error("Failed to read major version");
+  }
+
+  char minor = 0;
+  if (fread(&minor, 1, 1, file) != 1) {
+    fclose(file);
+    throw std::runtime_error("Failed to read minor version");
+  }
+
+  uint16_t header_len;
+  if (fread(&header_len, 2, 1, file) != 1) {
+    fclose(file);
+    throw std::runtime_error("Failed to read header length");
+  }
+
+  char *header = (char *)malloc(header_len);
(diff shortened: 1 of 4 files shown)


djeong20 (approved):
LGTM!

is a reader the only thing we need? (no saver?)

Applications/utils/npy_reader/npy_reader.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * @param dims data shape
+ * @param values return value that contain npy's contents
+ */
+void read_npy_file(const char *file_path, std::vector<int> &dims,
+                   std::vector<float> &values) {
+  FILE *file = fopen(file_path, "rb");
+
+  char magic[7] = {};
+  fread(magic, 6, 1, file);
+  assert(strcmp(magic, "\x93NUMPY") == 0);
```
myungjoo, line 37: assert is not a good idea for a library if it fails because of invalid parameters from the app. return a proper error number or throw an exception.

this function also needs proper error handling (e.g., invalid fail_path) unless this is for test purpose only.

Applications/utils/npy_reader/npy_reader.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include <assert.h>
+#include <cstring>
+#include <iostream>
+#include <random>
+#include <string>
+#include <vector>
+
+namespace nntrainer::util {
+namespace {
```
myungjoo, line 23: Do you really intend to have an unnamed namespace (not allowing other files to access) for a utility function?

DonghakPark:
> LGTM!
> 
> is a reader the only thing we need? (no saver?)

i will add numpy DataLoader for nntrainer, and Save later!

Applications/utils/npy_reader/npy_reader.cpp
DonghakPark: 1. I replaced the assert statement with try catch and ml_error.
2. I added error checking for file operations.
3. make if statement that check fread count

Applications/utils/npy_reader/npy_reader.cpp
DonghakPark: Fixed!

Applications/utils/npy_reader/npy_reader.h
```diff
@@ -0,0 +1,31 @@
...
+ * @brief read numpy file from file_path
+ *
+ * @param file_path file_path that want to read
+ * @param dims data shape
+ * @param values return value that contain npy's contents
+ */
+void read_npy_file(const char *file_path, std::vector<int> &dims,
+                   std::vector<float> &values);
+
+} // namespace nntrainer::util
```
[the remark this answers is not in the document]
DonghakPark: Fixed!

Applications/utils/npy_reader/npy_reader.h
```diff
@@ -0,0 +1,43 @@
...
+
+public:
+  NpyReader(const char *file_path, std::vector<int> &dims,
+            std::vector<float> &values) :
+    file_path(file_path), dims(dims), values(values) {}
+  /**
+   *
+   * @brief read numpy file from file_path
+   */
+  void read_npy_file();
```
myungjoo, line 35: I don't think read_npy_file is required. The constructur should've done it.

DonghakPark: FIX! Thank you for your feedback.
your proposed method much better scalability!! later i will add reload, save, nntrainer_dataloader method to this util function!!

DonghakPark:
The failure of the clang-format check in TAOS is due to a version issue, and we plan to disable the TAOS CI.

skykongkong8 (approved):

Applications/utils/npy_reader/npy_reader.cpp
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * @brief read numpy file from file_path
+ *
+ * @param file_path file_path that want to read
+ * @param dims data shape
+ * @param values return value that contain npy's contents
+ */
```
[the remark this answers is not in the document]
DonghakPark: FIX! good point

Applications/utils/npy_reader/npy_reader.cpp
```diff
@@ -0,0 +1,119 @@
...
+  header = nullptr;
+
+  int total_entries = 1;
+  for (int i = 0; i < dims.size(); ++i) {
+    total_entries *= dims[i];
+  }
+
+  for (int i = 0; i < total_entries; ++i) {
+    float value;
+    fread(&value, 4, 1, file);
```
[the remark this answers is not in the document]
DonghakPark: umpy stores data in 4-byte increments for most types. It is necessary to check how many bytes are used for other data types.

myungjoo (approved):

PR #2498, opened 2024-03-07
[Layer] Remove Tensor setDataType() usage

djeong20:
In several layers, there are attempts to change the data type of a Tensor object after initializing it.
This is currently possible but can cause issues down the line (e.g., treat FloatTensor object as HalfTensor).
As such, the setDataType() method will be removed and considered not to be used in future updates.
Instead, users will need to provide the desired data type when creating a new tensor. 

**Self-evaluation:**

### DIFF
--- a/nntrainer/layers/layer_normalization_layer.cpp
+++ b/nntrainer/layers/layer_normalization_layer.cpp
@@ -227,7 +227,6 @@ void LayerNormalizationLayer::incremental_forwarding(RunLayerContext &context,
   deviation.multiply(inv_std_dev, output);
   output.multiply_i(gamma);
   output.add_i(beta);
-
 }
 
 void LayerNormalizationLayer::calcDerivative(RunLayerContext &context) {
@@ -236,8 +235,8 @@ void LayerNormalizationLayer::calcDerivative(RunLayerContext &context) {
   TensorDim::TensorType weight_tensor_type =
     context.getWeight(wt_idx[LNParams::gamma]).getTensorType();
 
-  Tensor empty;
-  empty.setTensorType(weight_tensor_type);
+  Tensor empty =
+    Tensor("empty", weight_tensor_type.format, weight_tensor_type.data_type);
 
   Tensor &outgoing_derivative = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
   const Tensor &incoming_derivative =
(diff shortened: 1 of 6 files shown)


skykongkong8 (approved):
Think this is quite important point.
I will keep this in mind in the future development of myself as well :)

myungjoo:
Question:

In a network, is the Tensor name required to be unique?
```
  Tensor tensor =
    Tensor("some_name", value.getFormat(), value.getDataType());
```

(I hope not..)

nntrainer/layers/layer_normalization_layer.cpp
```diff
@@ -236,8 +235,8 @@ void LayerNormalizationLayer::calcDerivative(RunLayerContext &context) {
   TensorDim::TensorType weight_tensor_type =
     context.getWeight(wt_idx[LNParams::gamma]).getTensorType();
 
-  Tensor empty;
-  empty.setTensorType(weight_tensor_type);
+  Tensor empty =
+    Tensor("empty", weight_tensor_type.format, weight_tensor_type.data_type);
```
jijoongmoon, line 239: We can set the format and data_type with weight_tensor_type. No need to set it. Also, we do not need to set the tensor name always.  I think when we create tensor like here, ```Tensor empty```, then it does not allocate the buffer. So.. I think it is ok to set the Tensor type here. The reason we set like this.. we do not want to allocate a buffer at this time.

djeong20:
> Question:
> 
> In a network, is the Tensor name required to be unique?
> 
> ```
>   Tensor tensor =
>     Tensor("some_name", value.getFormat(), value.getDataType());
> ```
> 
> (I hope not..)

I don't think the name has to be unique.

nntrainer/layers/layer_normalization_layer.cpp
djeong20: As we discussed offline, `Tensor empty;` would create FloatTensor as a default. 
I will make an additional Tensor constructor that only takes data type (e.g., `Tensor empty = Tensor(weight_tensor_type.data_type);`)

myungjoo:
> > Question:
> > In a network, is the Tensor name required to be unique?
> > ```
> >   Tensor tensor =
> >     Tensor("some_name", value.getFormat(), value.getDataType());
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (I hope not..)
> 
> I don't think the name has to be unique.

If this is for debugging and developer referencing, it's ok.

myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #1300, opened 2021-06-17
[ GRU ] Add GRU Unittest

jijoongmoon:
This commit includes,
  . unittests of gru layer
  . add keras code to compare

**Self evaluation:**
1. Build test:	 [X]Passed [ ]Failed [ ]Skipped
2. Run test:	 [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: jijoong.moon <jijoong.moon@samsung.com>

PR #2500, opened 2024-03-08
[Tensor] Refactorize Tensor Class to TensorV2 @open sesame 03/26 12:26

djeong20:
This pull request introduces a major refactorization of the Tensor class in our codebase by introducing the new TensorV2 class.
The previous Tensor class is removed, and all instances where it was used have been updated accordingly.
With this change, the overall functionality and stability of our system are expected to be increased. 

**Self-evaluation:**


skykongkong8 (approved):
I could find trivial typos or missed descriptions, but think it is not the major target of this PR.
Overall, LGTM! Thanks for your hard work! :)

djeong20:
> I could find trivial typos or missed descriptions, but think it is not the major target of this PR. Overall, LGTM! Thanks for your hard work! :)

please comment on the parts where it has typos or missed descriptions so I can fix it!

nntrainer/tensor/tensor_base.cpp
```diff
@@ -176,14 +176,14 @@ void TensorBase::getSharedDataTensor(const TensorDim dim_, size_t offset,
...
-TensorBase::computeBroadcastInfo(const TensorV2 &m) const {
+TensorBase::BroadcastInfo
+TensorBase::computeBroadcastInfo(const Tensor &m) const {
   if (m.size() > this->size())
     throw exception::not_supported("broadcasting *this is not supported");
 
   const TensorDim m_dim = m.getDim();
 
-  BroadcastInfoV2 e;
+  BroadcastInfo e;
```
SeoHyungjun, line 186: Is there a reason why BroadcastInfo's variable name is e?

SeoHyungjun (approved):
I think you just need to resolve the conflict!👍

nntrainer/tensor/tensor_base.cpp
djeong20: no particular reason! reused the previous name.

nntrainer/tensor/tensor.cpp
```diff
@@ -2148,1680 +734,401 @@ Tensor &Tensor::dot_batched_deriv_wrt_2(Tensor &m_deriv,
...
+    << "cannot split axis of axis: " << axis;
+
+  NNTR_THROW_IF(tensors.empty(), std::invalid_argument)
+    << "given tensor vector is empty";
+
+  Tensor output;
+  Tdatatype dtype = tensors.front().getDim().getDataType();
+
+  if (dtype == Tdatatype::FP32) {
+    output = FloatTensor::cat(tensors, axis);
```
jijoongmoon, line 832: Do we need this if branch?

nntrainer/tensor/tensor.cpp
```diff
@@ -2148,1680 +734,401 @@ Tensor &Tensor::dot_batched_deriv_wrt_2(Tensor &m_deriv,
...
+  }
 
-  } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
+  if (from.size() != 0 && size() == from.size() &&
+      getDataType() == from.getDataType()) {
+    // if tensor size and data type match, copy data
+    itensor->copy(from);
+  } else {
+    // replace with a new tensor that are the same with the given tensor
+    if (from.getDataType() == ml::train::TensorDim::DataType::FP32) {
```
jijoongmoon, line 878: Can we remove this if branch by calling itensor.copy?

nntrainer/tensor/tensor.cpp
```diff
@@ -2148,1680 +734,401 @@ Tensor &Tensor::dot_batched_deriv_wrt_2(Tensor &m_deriv,
...
     }
-  } else if (getDataType() == ml::train::TensorDim::DataType::FP16) {
+  } else {
+    // replace with a new tensor that has the same data as the given tensor
+    Tensor t = Tensor(from.getDim(), true);
+    for (unsigned int b = 0; b < t.batch(); ++b) {
+      for (unsigned int c = 0; c < t.channel(); ++c) {
+        for (unsigned int h = 0; h < t.height(); ++h) {
+          for (unsigned int w = 0; w < t.width(); ++w) {
+            if (getDataType() == ml::train::TensorDim::DataType::FP32) {
```
jijoongmoon, line 913: Ditto.

nntrainer/tensor/tensor.h
```diff
@@ -477,117 +274,92 @@ class Tensor {
...
   /**
-   * @brief    Deallocate memory for this tensor
-   * @note     This will not necessary free the memory as tensors share memory
+   * @brief     return Data pointer of Tensor
+   * @retval    template T pointer
    */
-  void deallocate() {
-    data = nullptr;
-    offset = 0;
+  template <typename T = float> T *getData() const {
```
jijoongmoon, line 302: I think eventually we should remove this getData() and getAddress() function.

nntrainer/tensor/tensor.cpp
djeong20: I left this code intentionally in the tensor class for the following reasons.

1. This condition checks the data type of the Tensor to copy. Therefore, even if this code is moved into Float/HalfTensor, we still need to check the source Tensor's data type, which eventually has the same code in both classes.
2. Current implementation uses swap() to perform copy. However, this won't be a valid option when it happens in the Float/HalfTensor class. This is because `swap(t, *this)` would be `swap(Tensor &lhs, FloatTensor &rhs)` or `swap(Tensor &lhs, HalfTensor &rhs)` instead of `swap(Tensor &lhs, Tensor &rhs)`. While Tensor has FloatTensor as a member variable,  I believe implementing a swap between Tensor and FloatTensor would require structural change.

Hope this clarifies and let me know if you have further questions.

djeong20: Also, if you have any suggestions, please bring them up!

nntrainer/tensor/tensor.cpp
djeong20: Same issue as `copy()`.

nntrainer/tensor/tensor.cpp
djeong20: I believe so. As far as I know, this function concatenates a vector of tensors by a given axis. 
However, the problem with this function is tensors are concatenated to the first tensor of the list, not the tensor that calls this function. Therefore, the FP16 tensor can concatenate a list of FP32 tensors.
While the input tensors have no relation with the tensor that calls this function, we need a decision point on whether to concatenate a list of FP32 tensors or FP16 tensors, and this should be made in the tensor class. 

I think there is plenty of room to improve the function itself and we can later newly design this function.

nntrainer/tensor/tensor.cpp
jijoongmoon: how about using template? Cause if we need to add more tensor type like int8, bf16, then there will be more branch.

nntrainer/tensor/tensor.cpp
jijoongmoon: It is minor, how about moving the if part to outside of the loop?

nntrainer/tensor/tensor.cpp
jijoongmoon: On second thought, we can just remove this branch and create Tensor T using copy constructor.

nntrainer/tensor/tensor.cpp
jijoongmoon: But we change the api.. like tensors[0]->itensor.cat(tensors, axis)?

djeong20: we cannot use it as `tensors[0]->itensor.cat(tensors, axis)` since `itensor` is a private variable. I'll modify this in another way and let you know.

nntrainer/tensor/tensor.cpp
djeong20: fixed!

nntrainer/tensor/tensor.cpp
djeong20: this is resolved now. thanks for pointing it out!

nntrainer/tensor/tensor.cpp
djeong20: I fixed this by creating a new member function of Tensor `concat()`, and the current `Tensor::cat()`, which is a static function, uses `concat()`. This should work exactly the same as how `tensors[0]->itensor.cat(tensors, axis)` would operate.

nntrainer/tensor/float_tensor.h
```diff
@@ -186,180 +192,203 @@ class FloatTensor : public TensorBase {
...
   /**
-   * @copydoc TensorV2::copyData(const TensorV2 &from)
+   * @copydoc Tensor::copyData(const Tensor &from)
    */
-  void copyData(const TensorV2 &from);
+  void copyData(const Tensor &from);
 
   /**
-   * @copydoc TensorV2::argmax()
+   * @brief      Copy the Tensor
```
EunjuYang, line 379: What does the `stride` stand for? (implicit stride is applied only when data types are heterogeneous) Is it used to copy tensor with different type only? Could you add more exaplanation on the comment?

nntrainer/tensor/float_tensor.h
```diff
@@ -186,180 +192,203 @@ class FloatTensor : public TensorBase {
...
+   * @param[in]  input Tensor to be copied
+   * @param[out] output output Tensor
+   */
+  void copy_with_stride(const Tensor &input, Tensor &output) override;
```
EunjuYang, line 383: What does the `stride` stand for? (implicit stride is applied only when data types are heterogeneous) Is it used to copy tensor with different type only? Could you add more explanation on the comment? If this copy is used only when tensor types are different, could you think about changing the function name? I feel it sounds to support strided copy from any tensor with any stride options.

djeong20: `copy_with_stride()` is not used to copy tensors with different data types. The function assumes it only takes the same data type. `copyData()` should be the function for copying different data types. 
I'm not sure what's the initial purpose for this function since there's not much description in https://github.com/nnstreamer/nntrainer/pull/1300, but it is currently used for copying uncontiguous tensors, which in most cases copies shared data tensors. We should discuss whether to keep the name or change it to a better one.

djeong20: Also, I'm working on a new PR for adding descriptions and unifying/changing naming. Please let me know if other functions seem unclear or need more explanations.

jijoongmoon (approved):
LGTM! Thanks for great work!

PR #2502, opened 2024-03-11
[ neon ] Hgemv transpose biggerkernel

skykongkong8:
1. Rather than using N - 8 and adding more chunk of code, using N8 can make the code shorter
2. In previous neon patch, I deliberately removed 16x8 kernel-loop for hgemv_transpose, since I could not find difference in terms of latency. However, it turns out to be efficient when multiple times of MV ops are used.

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -834,11 +834,12 @@ void hgemv_transpose(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t M,
     Y32[idx] = beta * Y[idx];
   }
   unsigned int i = 0;
+  unsigned int N8 = N & -8;
   for (; M - i >= 8; i += 8) {
     __fp16 x[8];
     vst1q_f16(&x[0], vmulq_n_f16(vld1q_f16(&X[i]), alpha));
 #pragma omp parallel for schedule(guided) num_threads(GEMV_NUM_THREADS)
-    for (unsigned int idx = 0; idx < N - 8; idx += 8) {
+    for (unsigned int idx = 0; idx < N8; idx += 8) {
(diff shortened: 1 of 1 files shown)


baek2sm (approved):
LGTM

nntrainer/tensor/blas_neon.cpp
```diff
@@ -834,11 +834,13 @@ void hgemv_transpose(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t M,
     Y32[idx] = beta * Y[idx];
   }
   unsigned int i = 0;
-  for (; M - i >= 8; i += 8) {
-    __fp16 x[8];
+  unsigned int N8 = N & -8;
```
myungjoo, line 837: For bitwise ops, "clean bits", ```N & ~0x7``` is more readable. (a negative number in an expression with unsigned is... irritating)

skykongkong8: Good point. Will apply right away.

myungjoo: ``` (N >> 3) << 3 ``` is also a popular expression for readability.

nntrainer/tensor/blas_neon.cpp
```diff
@@ -834,11 +834,13 @@ void hgemv_transpose(const __fp16 *A, const __fp16 *X, __fp16 *Y, uint32_t M,
     Y32[idx] = beta * Y[idx];
   }
   unsigned int i = 0;
-  for (; M - i >= 8; i += 8) {
-    __fp16 x[8];
+  unsigned int N8 = (N << 3) << 3;
```
myungjoo, line 837: No! ```(N >> 3) << 3```

skykongkong8: My bad... Fixed!

myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #2503, opened 2024-03-11
[OpenCL/GPU] Kernel binary caching @open sesame 04/04 09:22

s-debadri:
Changes for GPU kernel caching:

- Added utility for saving kernel as binary files using `clGetProgramInfo`.
- Added wrapper for `clCreateProgramWithBinary`.


nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,105 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+            error_code,
+            (GetProgramBuildInfo(device_id, CL_PROGRAM_BUILD_LOG)).c_str());
+    return false;
+  }
+
+  // since only one kernel is being saved via this function at a time
+  // kernel_names_size -> array with 1 element
+  kernel_names = (char *)malloc(kernel_names_size[0]);
+  error_code = clGetProgramInfo(program_, CL_PROGRAM_KERNEL_NAMES,
+                                kernel_names_size[0], kernel_names, NULL);
```
myungjoo, line 97: You will be accessing either a null pointer or a garbage value.

nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,105 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+  }
+
+  error_code = clGetProgramInfo(program_, CL_PROGRAM_BINARIES,
+                                binaries_ptr_alloc_size, binaries_ptr, NULL);
+
+  if (error_code != CL_SUCCESS) {
+    ml_loge("Failed to get program binary data. OpenCL error code: %d. %s",
+            error_code,
+            (GetProgramBuildInfo(device_id, CL_PROGRAM_BUILD_LOG)).c_str());
+    return false;
```
myungjoo, line 124: Unallocate everything before exiting with an error.

nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,105 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+  } else {
+    ml_logi("Saving kernel binary for: %s", std::string(kernel_names).c_str());
+  }
+
+  // Read the binary
+  size_t binaries_ptr_alloc_size = sizeof(unsigned char *) * num_devices;
+  binaries_ptr = (unsigned char **)malloc(binaries_ptr_alloc_size);
+
+  memset(binaries_ptr, 0, binaries_ptr_alloc_size);
```
myungjoo, line 112: Why memset here? (useless op)

nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,105 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+}
+
+/**
+ * @brief Utility to get program info and save kernel binaries
+ *
+ * @param device_id OpenCL device id
+ * @return true if successful or false otherwise
+ */
+bool Program::GetProgramInfo(cl_device_id device_id) {
+  // since only GPU is being used
```
myungjoo, line 60: You are writing the whole function as a C code without using convinient (making the code shorter, simpler, and more readable) C++ syntax. The pointers in this funcion are supposed to be freed on exit, try smart pointers (e.g., std::unique_ptr).

nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,105 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+  cl_int error_code = CL_SUCCESS;
+  size_t *binaries_size = NULL;
+  unsigned char **binaries_ptr = NULL;
+  char *kernel_names;
+
+  // Read the binary size
+  size_t binaries_size_alloc_size = sizeof(size_t) * num_devices;
+  binaries_size = (size_t *)malloc(binaries_size_alloc_size);
+  error_code = clGetProgramInfo(program_, CL_PROGRAM_BINARY_SIZES,
+                                binaries_size_alloc_size, binaries_size, NULL);
```
myungjoo, line 72: In C++, ```nullptr``` is proper for this pointer argument. Entering ```NULL``` may generate compiler errors in the future when the code gets extended.

nntrainer/opencl/opencl_program.cpp
myungjoo: And ```kernel_names_size[0]``` gives the impression that ```kernel_names_size``` is an array with multiple elements. When it's a pointer to a single value, please write ```*kernel_names_size```.

nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,105 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+  if (error_code != CL_SUCCESS) {
+    ml_loge("Failed to get program binary size. OpenCL error code: %d. %s",
+            error_code,
+            (GetProgramBuildInfo(device_id, CL_PROGRAM_BUILD_LOG)).c_str());
+    return false;
+  }
+
+  // Read the kernel name
+  size_t *kernel_names_size;
+  error_code = clGetProgramInfo(program_, CL_PROGRAM_KERNEL_NAMES, 0, NULL,
```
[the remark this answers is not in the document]
s-debadri: Thanks for these observations. I will modify it to have better, precise and  readable C++ paradigm. Will update the afterwards.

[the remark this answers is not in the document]
s-debadri: Added suggestions in the latest commit

nntrainer/opencl/opencl_program.cpp
```diff
@@ -41,6 +43,101 @@ bool Program::BuildProgram(cl_device_id device_id,
...
+    for (unsigned int i = 0; i < num_devices; ++i) {
+      delete[] binaries_ptr[i];
+    }
+    return false;
+  }
+
+  // Write the binary to file
+  // All kernels in the program will be saved in the binary file
+  for (unsigned int i = 0; i < num_devices; ++i) {
+    std::ofstream fs(std::string(kernel_names) + "_kernel.bin",
```
jihochu, line 130: It needs to get a parameter save path like `kernel_path + ... ` which can be defined by option.

nntrainer/opencl/opencl_program.cpp
```diff
@@ -29,8 +31,8 @@ namespace nntrainer::opencl {
  * @return true if successful or false otherwise
```
EunjuYang, line 31: It would be better to add a comment for `@param binaryCreated`

myungjoo (approved):

nntrainer/opencl/opencl_program.cpp
s-debadri: Added in latest commit.

nntrainer/opencl/opencl_program.cpp
s-debadri: Incorporated opencl-kernel-path pre-processor directive in latest commit

nntrainer/opencl/opencl_program.cpp
```diff
@@ -21,13 +21,20 @@
 
 #include <nntrainer_log.h>
 
+#ifdef OPENCL_KERNEL_PATH
+const std::string DEFAULT_KERNEL_PATH = OPENCL_KERNEL_PATH;
+#else
+const std::string DEFAULT_KERNEL_PATH = "";
```
myungjoo, line 27: In the next PR (you may go ahead with this PR and make another PR to fix it), please enable envvar or .ini file configuration for DEFAULT_KERNEL_PATH. Having the default file search path hardcoded is not recommended although it's fine if it can be overriden by envvar or .ini file.

And... it'd be more proper if it's a static-const property of the Program class.

meson.build
```diff
@@ -98,6 +98,11 @@ if get_option('enable-opencl')
   message ('OpenCL build is enabled. Will work only if OpenCL supported GPU is available.')
   extra_defines += '-DENABLE_OPENCL=1'
 endif
+
+if get_option('opencl-kernel-path') != ''
+  message ('OpenCL kernel path set to: @0@'.format(get_option('opencl-kernel-path')))
+  extra_defines += '-DOPENCL_KERNEL_PATH=\"@0@\"'.format(get_option('opencl-kernel-path'))
```
myungjoo, line 104: This is an over-engineering.

Refer to https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

Then, you don't need the #if/#endif in opencl_program.cpp.

s-debadri: Modified in latest commit

myungjoo:
@EunjuYang @jihochu PTAL. This is taking too long.

jijoongmoon:
could you check the CI?

DonghakPark (approved):
LGTM!! 
i will check CI!!

DonghakPark:
@s-debadri  @jijoongmoon  
```
nofile:0:0: information: Too many #ifdef configurations - cppcheck only checks 12 configurations. Use --force to check all configurations. For more details, use --enable=information. [toomanyconfigs]
```
It means that there are too many conditional compilation directives (#ifdef and others) in the code being analyzed. Cppcheck can handle up to 12 such configurations at a time by default

and i give --force option and run in my local environment --> There is no cppcheck error on this PR

jijoongmoon (approved):
LGTM

PR #2504, opened 2024-03-11
[build] Updated Android build script for OpenCl inclusion

s-debadri:
Added `enable-opencl` flag to `/tools/package_android.sh`.
This will compile OpenCL related changes for `libnntrainer.so`.

### DIFF
--- a/tools/package_android.sh
+++ b/tools/package_android.sh
@@ -16,13 +16,15 @@ pushd $TARGET
 if [ ! -d builddir ]; then
     #default value of openblas num threads is 1 for android
     #enable-tflite-interpreter=false is just temporally until ci system is stabel
-  meson builddir -Dplatform=android -Dopenblas-num-threads=1 -Denable-tflite-interpreter=false -Denable-tflite-backbone=false -Denable-fp16=true -Denable-neon=true -Domp-num-threads=1
+    #enable-opencl=true will compile OpenCL related changes or remove this option to exclude OpenCL compilations.
+  meson builddir -Dplatform=android -Dopenblas-num-threads=1 -Denable-tflite-interpreter=false -Denable-tflite-backbone=false -Denable-fp16=true -Denable-neon=true -Domp-num-threads=1 -Denable-opencl=true
 else
   echo "warning: $TARGET/builddir has already been taken, this script tries to reconfigure and try building"
   pushd builddir
     #default value of openblas num threads is 1 for android
     #enable-tflite-interpreter=false is just temporally until ci system is stabel  
-    meson configure -Dplatform=android -Dopenblas-num-threads=1 -Denable-tflite-interpreter=false -Denable-tflite-backbone=false -Denable-fp16=true -Denable-neon=true -Domp-num-threads=1
+    #enable-opencl=true will compile OpenCL related changes or remove this option to exclude OpenCL compilations.
+    meson configure -Dplatform=android -Dopenblas-num-threads=1 -Denable-tflite-interpreter=false -Denable-tflite-backbone=false -Denable-fp16=true -Denable-neon=true -Domp-num-threads=1 -Denable-opencl=true
     meson --wipe
   popd
 fi


skykongkong8 (approved):
LGTM

djeong20:
> Good work for fixing current build fail. But think there are a few things to check once more... @djeong20 and I are wondering why does excluding -DENABLE_OPENCL=1 and making option in `package_android.sh` 's enable-opencl as false is not working. And even with that option, I could find all opencl-related files are being included as src files while build. I want to check if this is a desired situtation because for example, all neon-related files are being included only when enable-neon option has set.

on top of this, if the Android compilation/unit test fails due to enable-opencl=false, fixing `package_android.sh` seems to be the incorrect way to fix the problem. it should work both ways (whether the option is true or false).

s-debadri:
> > Good work for fixing current build fail. But think there are a few things to check once more... @djeong20 and I are wondering why does excluding -DENABLE_OPENCL=1 and making option in `package_android.sh` 's enable-opencl as false is not working. And even with that option, I could find all opencl-related files are being included as src files while build. I want to check if this is a desired situation because for example, all neon-related files are being included only when enable-neon option has set.
> 
> on top of this, if the Android compilation/unit test fails due to enable-opencl=false, fixing `package_android.sh` seems to be the incorrect way to fix the problem. it should work both ways (whether the option is true or false).

Currently adding `enable-opencl=false` in either `meson_options.txt` or `package_android.sh` will have similar behaviour, i.e it will exclude OpenCL files and related changes. This will build the `libnntrainer.so` accordingly and not effect the unit tests.

But for unit tests, the build setup is in `nntrainer/test/jni/Android.mk`. So if we use any `#ifdef ENABLE_OPENCL` directive inside unit test codes then we have to add `DENABLE_OPENCL=1` to `LOCAL_CFLAGS` of that particular test. Currently `unittest_layers` build setup is using `DENABLE_OPENCL=1` by default. So `unittest_layers` is expecting symbols related to OpenCL inside `libnntrainer.so` and throwing build error if they are not present.

Removing `DENABLE_OPENCL=1` from `nntrainer/test/jni/Android.mk` of `unittest_layers` will build the unit test properly even if we have used `enable-opencl=false` while building `libnntrainer.so`. So there can be a single way to effect both `libnntrainer.so` and unit test build. Since, they are independent builds at this moment, we are getting build errors on unit test if it is trying to access OpenCL symbols from `libnntrainer.so`.

djeong20:
> > > Good work for fixing current build fail. But think there are a few things to check once more... @djeong20 and I are wondering why does excluding -DENABLE_OPENCL=1 and making option in `package_android.sh` 's enable-opencl as false is not working. And even with that option, I could find all opencl-related files are being included as src files while build. I want to check if this is a desired situation because for example, all neon-related files are being included only when enable-neon option has set.
> > 
> > 
> > on top of this, if the Android compilation/unit test fails due to enable-opencl=false, fixing `package_android.sh` seems to be the incorrect way to fix the problem. it should work both ways (whether the option is true or false).
> 
> Currently adding `enable-opencl=false` in either `meson_options.txt` or `package_android.sh` will have similar behaviour, i.e it will exclude OpenCL files and related changes. This will build the `libnntrainer.so` accordingly and not effect the unit tests.
> 
> But for unit tests, the build setup is in `nntrainer/test/jni/Android.mk`. So if we use any `#ifdef ENABLE_OPENCL` directive inside unit test codes then we have to add `-DENABLE_OPENCL=1` to `LOCAL_CFLAGS` of that particular test. Currently `unittest_layers` build setup is using `-DENABLE_OPENCL=1` by default. So `unittest_layers` is expecting symbols related to OpenCL inside `libnntrainer.so` and throwing build error if they are not present.
> 
> Removing `-DENABLE_OPENCL=1` from `nntrainer/test/jni/Android.mk` of `unittest_layers` will build the unit test properly even if we have used `enable-opencl=false` while building `libnntrainer.so`. So there can be a single way to effect both `libnntrainer.so` and unit test build. Since, they are independent builds at this moment, we are getting build errors on unit test if it is trying to access OpenCL symbols from `libnntrainer.so`.

To clarify, this change is made since there's a `-DENABLE_OPENCL=1` flag as a default in `Android.mk` file for `unittest_layers`. 
Also, removing `-DENABLE_OPENCL=1` from `Android.mk` would work properly without this change.
Is this correct?

s-debadri:
> > > > Good work for fixing current build fail. But think there are a few things to check once more... @djeong20 and I are wondering why does excluding -DENABLE_OPENCL=1 and making option in `package_android.sh` 's enable-opencl as false is not working. And even with that option, I could find all opencl-related files are being included as src files while build. I want to check if this is a desired situation because for example, all neon-related files are being included only when enable-neon option has set.
> > > 
> > > 
> > > on top of this, if the Android compilation/unit test fails due to enable-opencl=false, fixing `package_android.sh` seems to be the incorrect way to fix the problem. it should work both ways (whether the option is true or false).
> > 
> > 
> > Currently adding `enable-opencl=false` in either `meson_options.txt` or `package_android.sh` will have similar behaviour, i.e it will exclude OpenCL files and related changes. This will build the `libnntrainer.so` accordingly and not effect the unit tests.
> > But for unit tests, the build setup is in `nntrainer/test/jni/Android.mk`. So if we use any `#ifdef ENABLE_OPENCL` directive inside unit test codes then we have to add `-DENABLE_OPENCL=1` to `LOCAL_CFLAGS` of that particular test. Currently `unittest_layers` build setup is using `-DENABLE_OPENCL=1` by default. So `unittest_layers` is expecting symbols related to OpenCL inside `libnntrainer.so` and throwing build error if they are not present.
> > Removing `-DENABLE_OPENCL=1` from `nntrainer/test/jni/Android.mk` of `unittest_layers` will build the unit test properly even if we have used `enable-opencl=false` while building `libnntrainer.so`. So there can be a single way to effect both `libnntrainer.so` and unit test build. Since, they are independent builds at this moment, we are getting build errors on unit test if it is trying to access OpenCL symbols from `libnntrainer.so`.
> 
> To clarify, this change is made since there's a `-DENABLE_OPENCL=1` flag as a default in `Android.mk` file for `unittest_layers`. Also, removing `-DENABLE_OPENCL=1` from `Android.mk` would work properly without this change. Is this correct?

Yes, removing `-DENABLE_OPENCL=1` from `Android.mk` would work properly without this change. Basically, it will then work on both scenarios - `enable-opencl=false` / `enable-opencl=true` of `libnntrainer.so` since `unittest_layers` will **not expect** OpenCL symbols from `libnntrainer.so`.

myungjoo (approved):

djeong20 (approved):

PR #2512, opened 2024-03-21
[coverity] fix coverity issue

EunjuYang:
- This commit fixes some coverity issues of
	- AUTO_CAUSES_COPY
	- MSSING_LOCK

Self-evaluation:

### DIFF
--- a/nntrainer/dataset/iteration_queue.cpp
+++ b/nntrainer/dataset/iteration_queue.cpp
@@ -76,17 +76,17 @@ ScopedView<Sample> IterationQueue::requestEmptySlot() {
     current_iterator++;
   }
 
-  auto view =
-    ScopedView<Sample>(&(*current_iterator),
-                       [current_being_filed = this->being_filled] {
-                         current_being_filed->markSampleFilled();
-                       },
-                       [this, current_being_filled = this->being_filled] {
-                         std::unique_lock lg(empty_mutex);
-                         this->markEmpty(current_being_filled);
-                         num_being_filled--;
-                         notify_emptied_cv.notify_all();
-                       });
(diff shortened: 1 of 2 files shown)


nntrainer/dataset/iteration_queue.cpp
```diff
@@ -107,6 +107,7 @@ ScopedView<Iteration> IterationQueue::requestFilledSlot() {
     auto stop_request_state = FlowState::FLOW_STATE_STOP_REQUESTED;
     bool exchange_result = flow_state.compare_exchange_strong(
       stop_request_state, FlowState::FLOW_STATE_STOPPED);
+    std::scoped_lock lg(empty_mutex);
```
djeong20, line 110: We should consider whether this lock is necessary for this function.
As far as I know, this data race condition occurs due to variables access in `NNTR_THROW_IF` (e.g., `num_being_filled`). 
While `empty_mutex` is used to request empty slots, I don't think it is preferable for `requestFilledSlot` to hold two mutex locks just to print an error message. (please let me know if I'm wrong)

djeong20: One simple solution is to remove the critical section by changing the error message.

EunjuYang: I updated the code you mentioned ([a0b37ca](https://github.com/nnstreamer/nntrainer/pull/2512/commits/a0b37ca481f18c78e453b6b2c378358967234666)). Thanks for your comment! :)

lhs8928 (approved):
LGTM

SeoHyungjun (approved):
LGTM

PR #2529, opened 2024-04-02
[Tensor] CharTensor class with signed 8-bit integer

djeong20:
In this PR, a new type of tensor, the CharTensor class, is designed explicitly for handling signed 8-bit integer data types that have already undergone quantization. 
This new tensor class aims to provide users with more options when working with tensors and their respective data types. 
Currently, the CharTensor class does not support mathematical operations like multiplication or addition. However, these features will be added in future updates.

**Self-evaluation:**


nntrainer/tensor/char_tensor.cpp
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  switch (initializer) {
+  case Initializer::ZEROS:
+    setZero();
+    break;
+  case Initializer::ONES:
+    setValue(1.0f);
+    break;
+  case Initializer::LECUN_NORMAL:
```
jijoongmoon, line 234: Could you check whether these initializers also work for int datatype?
If it is not, then it gives proper notification.

jijoongmoon (approved):
overall LGTM

nntrainer/tensor/char_tensor.cpp
djeong20: currently, these initializers won't work for tensors with integer type and throw an error. I intentionally left the code for future work. however, if these initializers are not valid for integers, they can be removed such as follows.

```suggestion
  case default:
      throw std::invalid_argument("Initializer not valid in current data type");
```

I'll do some research and update this in the later PR!

djeong20: I figured both normal and uniform are not possible for integer types, so I made the changes. PTAL!

skykongkong8 (approved):
LGTM!

Issue #2488, opened 2024-02-26
[ GEMM ] HGEMM noTrans case


skykongkong8:
Latest NNTrainer (26.02.24) HGEMM does not consider TLB cache.
I believe this is one of the most suspicious reasons for having nonlinear latency in big sized GEMM computation.
I am currently implementing micro, and macro scale HGEMM kernels in order to resolve this issue : https://github.com/skykongkong8/nntrainer/commit/ac9cbb5507d215308214bf3fe62b0562ab26ea8f
and observing some improvements in latency without deterioration of accuracy (w.r.t. full-fp16 HGEMM).

[ outdated ]
* v5 : apply SIMD in Brute Force implementation
* v6 : v5 + loop unrolling
* v7 : v8 + bigger kernel
* v8 : v7 + cache blocking
* v9 : v8 + packing
* v10 : v9 +  bigger kernel + adaptive kernel use
*  v11 : v10 + discontinuous packing on B
* [ WIP ] [v10_modified : v10 + small kernels + adaptive data packing](https://github.com/skykongkong8/nntrainer/tree/gemm_kernel_debug_v10)

[ current status ] : #2531 , #2541 , #2578

1. continuous data packing
2. 4x4, 4x8, 8x8 hgemm kernel for f16 and f16-f32
3. software prefetching

skykongkong8:
Those who want to make comments / reviews on WIP branch, please leave it here, or let me know! :)

skykongkong8:
@s-debadri As we discussed, please get started from here :) Thanks a lot!

skykongkong8:
### Current Status : 08.04.2024
> Unittest output with Galaxy S23 with #2541

| GEMM dimension | fp32 | prev | 8x8 | f16-f32 8x16 | full-f16 |
| --- | --- | --- | --- | --- | --- |
| 4096 square | 2087 ms  | 7172 ms | ... | **1964 ms** | 1452 ms |
| 2048 square | 260 ms  | 413 ms | ... | **250 ms** | 185 ms |
| 1024 square | 34 ms  | 52 ms | ... | **30 ms** | 103 ms |
| 768 square | 13 ms | 18 ms | ... | **11 ms** | 10 ms |
| 256X1440X256 | 2869 mcrs | 3807 mcrs | ... | **2544 mcrs** | 2055 mcrs |
| 256X256X1440 | 2929 mcrs | 3950 mcrs | ... | **2467 mcrs** | 2523 mcrs |
[...]

PR #2531, opened 2024-04-03
[ hgemm ] Use optimized hgemm if possible

skykongkong8:
Through this PR, we can use optimized version of hgemm with following conditions:
	1. noTrans hgemm
	2. M, N, K is divisible with 4 or 8
	3. Row Major GEMM
	4. alpha = 1.0, beta = 0.0 (will be patched soon)
Otherwise, use previous version as a fallback.
- Note that there are a few optimization strategies are left for optimal hgemm. 
- This is just a WIP version of it, but still better than before.


skykongkong8:
This PR is about resolving #2488

jijoongmoon (approved):
LGTM

PR #2535, opened 2024-04-04
[GPU/OpenCL] Handling pre-compiled and existing kernels

s-debadri:
Handling existing kernels:

- Modified `clCreateKernel` to read pre-compiled kernel binaries.
- Handling already initialized kernels using global bitmask.
- Added enumerator for kernel names and a method to resolve as string.

### DIFF
--- a/nntrainer/layers/layer_context.cpp
+++ b/nntrainer/layers/layer_context.cpp
@@ -16,6 +16,7 @@
 #include <memory>
 #include <tensor_wrap_specs.h>
 
+#include <fstream>
 #include <iterator>
 #include <layer_context.h>
 #include <nntrainer_log.h>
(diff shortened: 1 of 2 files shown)


myungjoo:
LGTM.

Later, kernel path policy should be aware of security and privileges of Android and Tizen devices. In such devices, it should be able to accept different paths per app or the caller.

However, as long as the kernel path is configurable, it can wait until opencl support becomes mature.

nntrainer/layers/layer_context.cpp
```diff
@@ -567,35 +568,94 @@ bool RunLayerContext::validate(bool skip_input, bool skip_label) {
...
 bool RunLayerContext::clCreateKernel(std::string kernel_string,
-                                     std::string kernel_name) {
-  if (kernel_initialized) {
-    ml_logi("Kernel already initialized: %s", kernel_name.c_str());
+                                     LayerKernel layerKernel,
+                                     opencl::Kernel &kernel_) {
+
+  // checking bitmask for already initialized kernel. eg: 010 & 000 -> 0 but 010
+  // & 110 -> 010
+  if (layerKernel == (kernelInitializedMask & layerKernel)) {
```
dkjung, line 589:
```C
if (layerKernel & kernelInitializedMask) {
```

I think that this statement has the same effect.

dkjung (approved):
LGTM

nntrainer/layers/layer_context.cpp
s-debadri: Yes...nice observation.

nntrainer/layers/layer_context.cpp
```diff
@@ -567,35 +568,94 @@ bool RunLayerContext::validate(bool skip_input, bool skip_label) {
...
-    result =
-      program.CreateCLProgram(context_inst_.GetContext(),
-                              context_inst_.GetDeviceId(), kernel_string, "");
+
+    // reading binary
+    std::ifstream fs(opencl::Program::DEFAULT_KERNEL_PATH + "/" +
+                       getKernelName(layerKernel) + "_kernel.bin",
+                     std::ios::binary | std::ios::in);
+
+    if (fs) {
```
wooksong, line 607:
```suggestion
    if (fs.good()) {
```

Is there any case that ```fs``` can be null?

s-debadri: I don't think the `fs` object can be null.

jijoongmoon (approved):
LGTM

PR #2539, opened 2024-04-05
[LoRA] add scaling parameter to LoRA

EunjuYang:
- This commit adds `scaling` parameter to LoRA (fc)
- In the original paper, they adopted `alpha (int)` as a parameter to derive the scaling factor internally, i.e., scaling = alpha / rank.
- However, in this commit it directly takes `scaling` as a hyper-parameter.
- The updates are summarized as follows:
	- `common_properties.h` : add LoraScaling as a parameter.
	- `fc_layer.cpp`: update forwarding / calcGradient / calcDerivative func to apply scaling factor in LoRA computation
	- `fc_layer.h`: update to take LoraScaling as fc_props
	- `node_exporter.cpp/h`: add LoRA Scaling as a parameter in tf.export format of fc layer
- [TODO] update tflite format to support the updated fc_layer params.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -1340,7 +1340,19 @@ class LoraRank : public PositiveIntegerProperty {
 public:
   static constexpr const char *key = "lora_rank"; /**< unique key to access */
   using prop_tag = uint_prop_tag;                 /**< property type */
-  ;
+};
+
+/**
+ * @brief LoRA scaling parameter
+ * @details It is used to set the scaling factor of LoRA, which is calculated as
+ * `scaling = alpha / rank` in the original paper. Defulat = 1.
+ */
+class LoraScaling : public Property<float> {
+public:
+  LoraScaling(float value = 1.0) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key =
+    "lora_scaling";                /**< unique key to access */
+  using prop_tag = float_prop_tag; /**< property type */
 };
 
 /**
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -40,7 +40,8 @@ enum FCParams { weight, bias };
 enum LORAParams { loraA, loraB, loraTmp, loraOut };
 
 FullyConnectedLayer::FullyConnectedLayer() :
-  LayerImpl(), fc_props(props::Unit(), props::LoraRank()) {
+  LayerImpl(),
+  fc_props(props::Unit(), props::LoraRank(), props::LoraScaling()) {
   weight_idx.fill(std::numeric_limits<unsigned>::max());
 }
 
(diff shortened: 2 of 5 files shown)


myungjoo (approved):

nntrainer/layers/common_properties.h
```diff
@@ -1340,7 +1340,19 @@ class LoraRank : public PositiveIntegerProperty {
...
+
+/**
+ * @brief LoRA scaling parameter
+ * @details It is used to set the scaling factor of LoRA, which is calculated as
+ * `scaling = alpha / rank` in the original paper. Defulat = 1.
+ */
+class LoraScaling : public Property<float> {
+public:
+  LoraScaling(float value = 1.0) : nntrainer::Property<float>(value) {}
+  static constexpr const char *key =
```
wooksong, line 1353: Well, is ```const``` after ```constexpr``` meaningful?

EunjuYang: As I understood, `constexpr const char *` is equivalent to `const char* const`. In this code, it is conventionally used to make the key pointer not to be revised/replaced by another string. `const char *` is used to save an immutable string, and `constexpr` is used to make the pointer immutable. Please let me know if there’s something I missed. Thank you :)

jijoongmoon (approved):
LGTM

nntrainer/layers/common_properties.h
wooksong: ```constexpr``` implies ```const```, which means ```constexpr char *``` is equivalent to ```const char *```. ```constexpr char * const``` seems what you want.

EunjuYang: Thank you for your answer. I partially agree with your comment, and it is correct when it is applied to a general variable. However, in this case, I am applying the `constexpr` to the pointer, so it seems a little bit different. I think the `constexpr` is applied to the type (`const char*`) itself, not the value. (B.T.W. `const` right after `constexpr` is applied to the value) So, in this case, it seems okay to write `constexpr const char*`. Please refer to this link : https://www.reddit.com/r/cpp_questions/comments/15iwf22/static_constexpr_char_vs_static_constexpr_const/
Thnx!! :)

wooksong (approved):
LGTM!

nntrainer/layers/fc_layer.cpp
```diff
@@ -186,9 +187,11 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
     Tensor &loraB = context.getWeight(lora_idx[LORAParams::loraB]);
     Tensor &hidden_tmp_lora = context.getTensor(lora_idx[LORAParams::loraTmp]);
     Tensor &hidden_out_lora = context.getTensor(lora_idx[LORAParams::loraOut]);
+    auto lora_scaling = std::get<props::LoraScaling>(fc_props).get();
```
djeong20, line 190: could we use `const auto &` to prevent coverity issue
```suggestion
    const auto &lora_scaling = std::get<props::LoraScaling>(fc_props).get();
```

nntrainer/layers/fc_layer.cpp
```diff
@@ -249,8 +252,9 @@ void FullyConnectedLayer::calcDerivative(RunLayerContext &context) {
   if (!std::get<props::LoraRank>(fc_props).empty()) {
     Tensor &lora_A = context.getWeight(lora_idx[LORAParams::loraA]);
     Tensor &lora_B = context.getWeight(lora_idx[LORAParams::loraB]);
-    ret_.dot_deriv_wrt_1(weight.add(lora_A.dot(lora_B)), derivative_, false,
-                         false);
+    auto lora_scaling = std::get<props::LoraScaling>(fc_props).get();
```
djeong20, line 255: same here 
```suggestion
   const auto &lora_scaling = std::get<props::LoraScaling>(fc_props).get();
```

djeong20 (approved):
LGTM

nntrainer/layers/fc_layer.cpp
EunjuYang: Thanks!! I fix it!

PR #2541, opened 2024-04-05
[ hgemm ] hgemm noTrans with kernel 8x16

skykongkong8:
This commit proposes a 8x16 kernel for Half-precision GEMM
Note that this is not an '100%' optimized version of HGEMM, but still better than before.
Following is unittest output with f16-f32 partial accumulated HGEMM.
Fine accuracy with better latency.

> **mean latency ( TC = 20 )**

| GEMM dimension | fp32 (cblas) | prev | 8x8 | 8x16 |
| --- | --- | --- | --- | --- |
| 4096 square | 2087 ms  | 7172 ms | ... | **1964 ms** |
| 2048 square | 260 ms  | 413 ms | ... | **250 ms** |
| 1024 square | 34 ms  | 52 ms | ... | **30 ms** |
| 768 square | 13 ms | 18 ms | ... | **11 ms** |
| 256X1440X256 | 2869 mcrs | 3807 mcrs | ... | **2544 mcrs** |
| 256X256X1440 | 2929 mcrs | 3950 mcrs | ... | **2467 mcrs** |
| 8X1440X8 | 5 mcrs | **5 mcrs** | ... | 10 mcrs |
| 8X8X1440 | 5 mcrs | **4 mcrs** | ... | 8 mcrs |

### DIFF
--- a/nntrainer/tensor/hgemm/hgemm.cpp
+++ b/nntrainer/tensor/hgemm/hgemm.cpp
@@ -29,9 +29,7 @@ void hgemm_noTrans(const __fp16 *A, const __fp16 *B, float *C32, unsigned int M,
   if (alpha == 1.F && beta == 0.F) {
     if (M % 8 == 0 && N % 16 == 0 && K % 8 == 0) {
       hgemm_noTrans_8x16(M, N, K, A, K, B, N, C32, N, alpha, beta);
-    }
-    else
-    if (M % 8 == 0 && N % 8 == 0 && K % 8 == 0) {
+    } else if (M % 8 == 0 && N % 8 == 0 && K % 8 == 0) {
       hgemm_noTrans_8x8(M, N, K, A, K, B, N, C32, N, alpha, beta);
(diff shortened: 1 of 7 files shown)


nntrainer/tensor/hgemm/hgemm.cpp
```diff
@@ -27,7 +27,9 @@
 void hgemm_noTrans(const __fp16 *A, const __fp16 *B, float *C32, unsigned int M,
                    unsigned int N, unsigned int K, float alpha, float beta) {
   if (alpha == 1.F && beta == 0.F) {
-    if (M % 8 == 0 && N % 8 == 0 && K % 8 == 0) {
+    if (M % 8 == 0 && N % 16 == 0 && K % 8 == 0) {
```
SeoHyungjun, line 30: Can’t you do it with bitwise operations?
I know that bit operations are generally faster, but I don't know if there is actually a performance difference after optimization.
If so, I think we can improve our performance by changing it.

SeoHyungjun (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2542, opened 2024-04-05
[neuralnet] enable multi batch incremental inference

lhs8928:
- The output was not considered multi batch input in incremental inference. Now it will return multi batch output.


jijoongmoon (approved):
LGTM

PR #2548, opened 2024-04-17
[neuralnet] bugfix multi batch incremental inference

lhs8928:
- This commit will handle when the model activation datatype is fp32

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -857,12 +857,14 @@ std::vector<float *> NeuralNetwork::incremental_inference(
               last_out_buf_data + batch * out_t.width(), 1);
 
 #else
+        throw std::invalid_argument("Error: enable-fp16 is not set");
+#endif
+      } else if (out->getDataType() == ml::train::TensorDim::DataType::FP32) {
         const float *out_t_batch_ptr = out_t.getData() +
                                        batch * out_t.getDim().getFeatureLen() +
                                        step * out_t.getDim().width();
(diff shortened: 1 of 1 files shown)


skykongkong8 (approved):
lgtm!

nntrainer/models/neuralnet.cpp
```diff
@@ -857,12 +857,14 @@ std::vector<float *> NeuralNetwork::incremental_inference(
               last_out_buf_data + batch * out_t.width(), 1);
 
 #else
+        throw std::invalid_argument("Errro: enable-fp16 is not set");
```
djeong20, line 860: typo?
```suggestion
        throw std::invalid_argument("Error: enable-fp16 is not set");
```

nntrainer/models/neuralnet.cpp
```diff
@@ -857,12 +857,14 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
+#endif
+      } else {
```
djeong20, line 862: maybe it would be better to specify if the data type is fp32 
```suggestion
      } else if (out->getDataType() == ml::train::TensorDim::DataType::FP32) {
```

djeong20 (approved):
LGTM except for minor fixes

nntrainer/models/neuralnet.cpp
lhs8928: Done

nntrainer/models/neuralnet.cpp
lhs8928: Done

jijoongmoon (approved):
LGTM!

PR #2549, opened 2024-04-18
[ Tensor ] Refactor blas/math related files into cpu backend considering arch-dep @open sesame 10/02 13:19

skykongkong8:
While during the process of implementing additional features in NEON, I found myself making unnecessary code blocks.
This is a suggestion-draft of refactorization for current blas/math related files.

**DONE**
- build on x86 & unittest (w/ & w/o fp16)
- build on android & unittest
- for all fallbacks
- Temporarily, `cpu_backend.h` itself is done by now.
- Following step should be syncing this work with `TensorV2` : #2500 
- Remove CBLAS params in Tensor to get rid of unnecessary dependency
- Use valid params for functions like : `sgemv`, `sgemm` (StorageOrder, data addr, ... )
- This is already done for `tensor.cpp` in this PR, while intentionally eliminating ALL `TensorV2` related files
- Replace all ```#include <blas_interface.h>``` to ```#include <cpu_backend.h>```
- Replace all ```#include <blas_neon.h>``` to ```#include <neon_impl.h>``` and manage fp16 functions in `.cpp` files
- Include all paths for clean builds for Tizen

Final form of this PR would be like:

```bash
...
tensor
├─── cpu_backend
│   └─── cpu_backend.h (has all external functions from previous `blas_interface.h`)
│   └─── arm
│      └─── arm_compute_backend : copy of `cpu_backend.h`. selectively uses functions from `cblas_interface`, `neon_impl`, `fallback_internal`
│      └─── neon_impl_fp16 (For armv8.2+)
│      └─── neon_impl
│           ...
│   └─── x86
│      └─── x86_compute_backend : copy of `cpu_backend.h`. selectively uses functions from `cblas_interface`, `blas_avx`, `fallback_internal`
│      └─── blas_avx : AVX2_implemented (vcvt, cblas) 
│           ...
│   └─── fallback
│      └─── fallback ( !x86 & !arm ) : copy of `cpu_backend.h`. uses ALL functions from `fallback_internal`
│      └─── fallback_internal : all raw implementations without SIMD or external lib
│           ...
│   └─── cblas_interface
│      └─── cblas_interface : all cblas-related function interfaces, and params
...
```
and removing `blas_interface.h`

### In short,

1. Substitute `blas_interface.h` to `cpu_backend.h` which has virtual functions of `blas_interface.h`
2. Actual implementations are implemented at `arm_compute_backend`, `x86_compute_backend`, and `fallback`, and they are included considering target cpu architecture. `cblas.h` is used for both of them for fp32 computation.
3. There are some differences (unsupported intrinsics, or dataTypes ) along the versions, and they are managed under each `arm` or `x86` directory.


skykongkong8:
Currently working on syncing with current upstream. Will be introduced again in the close future.

PR #2552, opened 2024-04-20
Add Mish activation function

Boseong-Seo:
- Now, user can use Mish activation function like torch or tensorflow.

**Self evaluation**:

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -78,14 +78,17 @@ public:
       in_place = false;
       this->setActivation<Tensor>(gelu<T>, geluPrime<T>);
       break;
+    case ActivationType::ACT_ELU:
+      this->setActivation<T>(elu<T>, eluPrime<T>);
+      break;
     case ActivationType::ACT_SOFTPLUS:
       this->setActivation<T>(softplus<T>, softplusPrime<T>);
       break;
+    case ActivationType::ACT_MISH:
+      this->setActivation<T>(mish<T>, mishPrime<T>);
+      break;
     case ActivationType::ACT_NONE:
       this->setActivation<T>(no_op<T>, no_op_prime<T>);
-      break;
-    case ActivationType::ACT_ELU:
-      this->setActivation<T>(elu<T>, eluPrime<T>);
       break;
     case ActivationType::ACT_UNKNOWN:
     default:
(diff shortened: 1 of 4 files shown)


nntrainer/layers/acti_func.h
```diff
@@ -78,15 +78,18 @@ class ActiFunc {
       in_place = false;
       this->setActivation<Tensor>(gelu<T>, geluPrime<T>);
       break;
+    case ActivationType::ACT_ELU:
```
myungjoo, line 81: Do you have any reason to move this code block? It is not recommended to add unrelated code changes in a commit. It makes harder to track changes and bugs in the future.

myungjoo (commented):
LGTM except for the unrelated code move in a switch-case.

nntrainer/layers/acti_func.h
Boseong-Seo: @myungjoo Just for the code readability and uniformity(?).
All activation types except for `ACT_ELU` were placed on top of `ACT_NONE` in the switch-case. So, I moved the block to that place for some kind of uniformity(I wonder this kind of thing can be also said as unifomity...). But if you think it doesn't need to, then I'll rollback this change.

DonghakPark (approved):
LGTM!

test/unittest/unittest_nntrainer_activations.cpp
```diff
@@ -517,6 +517,59 @@ TEST(nntrainer_activation, eluPrime_01_p) {
...
+  int height = 1;
+  int width = 10;
+
+  float answer[30] = {
+    -0.18891649, -0.15113318, -0.10714479, -0.05678857, 0.00000000,
+    0.06317942,  0.13259898,  0.20800139,  0.28903052,  0.37524524,
+    -0.28396326, -0.24693604, -0.18891649, -0.10714479, 0.00000000,
+    0.13259898,  0.28903052,  0.46613652,  0.65969974,  0.86509836,
+    -0.30883580, -0.29565641, -0.24693604, -0.15113318, 0.00000000,
+    0.20800139,  0.46613652,  0.76120591,  1.07794595,  1.40337825};
```
djeong20, line 532: quick question. how are you getting this answer data? PyTorch, TensorFlow, or hand calculation?

Boseong-Seo: With PyTorch!
I got the answer data by using `torch.nn.Mish()`.

jijoongmoon (approved):
LGTM

EunjuYang (approved):
LGTM

skykongkong8 (approved):

djeong20 (approved):
LGTM

nntrainer/layers/acti_func.h
myungjoo: Uh.. then never mind. It's good to go.

myungjoo (approved):

PR #2555, opened 2024-04-24
[ hdot ] Use precision-enhanced hdot

skykongkong8:
**hdot : vec(1 x K ) x vec(K X 1 ) -> 1 (scalar)**

- Previous hdot was using full-fp16.
- Since this is also one of dimension-shrinking computations, should use inter-fp32 values to enhance precision.
- This has not been detected due to small dimension Tensor usage in unittest. Add bigger-sized test case accordingly.

### A. Accuracy w.r.t cblas-f32 (mse)
| dim | hdot-f16 (previous) | hdot-f16f32 (now) |
| --- | --- | --- |
| 1024 | 0.250061 | 3.72529e-09 |
| 2048 | 2.07308 | 0.00362171 |
| 4096 | 11.8475 | 0.195379 |

| dim | hdot-f16 | hdot-f16f32 | cblas-f32 |
| --- | --- | --- | --- |
| 1024 | 261 ns | 280 ns | 347 ns |
| 2048 | 328 ns | 359 ns | 1473 ns |
| 4096 | 508 ns | 562 ns | 909 ns |

Conclusion : Better accuracy with almost no latency deterioration

**Self evaluation:**
1. Build test:     [X]Passed [ ]Failed [ ]Skipped
2. Run test:     [X]Passed [ ]Failed [ ]Skipped

PR #2556, opened 2024-04-25
[ hnrm2 ] Use precision-enhanced hnrm2

skykongkong8:
Along with #2555 I inspected through every custom-made half-precision calculation function, and `hnrm2` is the only one left that needs f16-f32 precision supplement.

**Proposed Changes**
- Previous hnrm2 was using full-fp16.
- Since this is also one of dimension-shrinking computation, should use inter-fp32 values to enhance precision.
- This has not been detected due to small dimension Tensor usage in unittest. Add higher dimension test case accordingly.
- Note that this function is responsible for `Tensor::l2norm()`, frequently used for mse loss computation.

### Differnece w.r.t. fp32-cblas
| dim | f16 (prev) | f16f32 (now) |
| --- | --- | --- |
| 768 | 282.233 | 0.232574 |

| dim | f16 | f16f32 | fp32-cblas | 
| --- | --- | --- | --- |
| 768 | 508 ns | 579 ns | 4242 ns |

Better accuracy with almost no latency deterioration

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -281,24 +281,28 @@ void sscal(const unsigned int N, const float alpha, _FP16 *X, const int incX) {
 
 static _FP16 snrm2_FP16(const unsigned int N, const _FP16 *X, const int incX) {
   unsigned int incx = abs(incX);
-  _FP16 sum = 0;
+  _FP16 sum;
   _FP16 tmp;
 #if (defined USE__FP16 && USE_NEON)
   if (incX == 1) {
     sum = nntrainer::neon::hnrm2(N, X);
   } else {
+    float sum32 = 0;
     for (unsigned int i = 0; i < N; i++) {
       tmp = X[i * incx];
-      sum += tmp * tmp;
+      sum32 += tmp * tmp;
     }
+    sum = static_cast<_FP16>(sqrt(sum32));
   }
 #else
+  float sum32 = 0;
   for (unsigned int i = 0; i < N; i++) {
     tmp = X[i * incx];
-    sum += tmp * tmp;
+    sum32 += tmp * tmp;
   }
+  sum = static_cast<_FP16>(sqrt(sum32));
 #endif
-  return static_cast<_FP16>(sqrt(sum));
+  return sum;
 }
 
 static void sgemm_FP16(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA,
--- a/test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
+++ b/test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
@@ -150,9 +150,52 @@ TEST(nntrainer_Tensor, l2norm) {
   result_fp32 = input_fp32.l2norm();
 
   // absolute error
-  const float epsilon = 1e-2;
+  const float epsilon = 1e-3 * width;
 
   EXPECT_NEAR(result_neon, result_fp32, epsilon);
+}
+
+TEST(nntrainer_Tensor, l2norm_big768) {
+
+  nntrainer::TensorDim::TensorType t_type_nchw_fp16 = {
+    nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP16};
+
+  nntrainer::TensorDim::TensorType t_type_nchw_fp32 = {
+    nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP32};
+
+  size_t batch = 1;
+  size_t channel = 1;
+  size_t height = 768;
+  size_t width = 768;
+
+  nntrainer::Tensor input(
+    nntrainer::TensorDim(1, 1, height, width, t_type_nchw_fp16));
+
+  nntrainer::Tensor input_fp32(
+    nntrainer::TensorDim(1, 1, height, width, t_type_nchw_fp32));
+
+  const float alpha = 1e-1;
+  const int MOD = 10;
+
+  GEN_TEST_INPUT(input, ((i * j * (batch * height * channel) +
+                          j * (batch * height) + k * (width) + l + 1) %
+                         MOD) *
+                          alpha);
+  GEN_TEST_INPUT(input_fp32, ((i * j * (batch * height * channel) +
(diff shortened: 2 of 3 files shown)


jijoongmoon (approved):
Great ! LGTM

nntrainer/tensor/blas_interface.cpp
```diff
@@ -281,24 +281,28 @@ void sscal(const unsigned int N, const float alpha, _FP16 *X, const int incX) {
...
   _FP16 tmp;
 #if (defined USE__FP16 && USE_NEON)
   if (incX == 1) {
     sum = nntrainer::neon::hnrm2(N, X);
   } else {
+    float sum32 = 0;
     for (unsigned int i = 0; i < N; i++) {
       tmp = X[i * incx];
-      sum += tmp * tmp;
+      sum32 += tmp * tmp;
```
djeong20, line 293: just wondering. would having `tmp` a float instead of fp16 make a difference?

test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
```diff
@@ -150,11 +150,54 @@ TEST(nntrainer_Tensor, l2norm) {
...
+
+  __fp16 result_neon;
+  float result_fp32;
+
+  result_neon = input.l2norm();
+  result_fp32 = input_fp32.l2norm();
+
+  float ErrorNeon = abs(result_neon - result_fp32);
+
+  const float epsilon = 1e-3 * width;
```
djeong20, line 197: what's the reasoning for multiplying width? is it to sum errors of num of width?

test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
```diff
@@ -150,11 +150,54 @@ TEST(nntrainer_Tensor, l2norm) {
...
+                              MOD) *
+                               alpha);
+
+  __fp16 result_neon;
+  float result_fp32;
+
+  result_neon = input.l2norm();
+  result_fp32 = input_fp32.l2norm();
+
+  float ErrorNeon = abs(result_neon - result_fp32);
```
djeong20, line 195: maybe use camelCase or snake_case?
```suggestion
  float error_neon = abs(result_neon - result_fp32);
```

djeong20 (approved):
Nice work.

nntrainer/tensor/blas_interface.cpp
skykongkong8: Definitely yes, for several reasons.

1. You know, fp16 arithmetic can only hold up to 65504, so using fp16 for summing variable(?) is not really a good idea.
It is highly likely to make overflow if N is sufficiently big number.
2. For the same reason, it might cause a underflow as well. For example, it might just skip small decimal values into consideration. As far as I am concerned, even with `__fp16` * `__fp16` multiplication, compiler internally changes them into `float`. Thus, storing that value in `__fp16` variable is also not a clever way I guess. At glance, the effect might seem trivial, but eventually it was causing quite big error when compared to output of fp32-cblas computation.

test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
skykongkong8: Maybe for the next PR? That is not really a point of this commit, but I will definitely take that in my mind.

test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
skykongkong8: There is unavoidable round-off error that linearly increase w.r.t. tensor size in half-precision computation. Meanwhile, I am comparing half-precision result with fp32 result. If they were using the same precision, I could use mathematically-defined criteria, but it is not. So I have been using this kind of epsilon to compensate such error.
Strictly speaking, this function makes (M x N) Matrix to one single value, so something like:
```C++
  const float epsilon = 1e-3 * height * width;
  // or const float epsilon = sqrt(1e-3 * height * width); ???
``` 
would be better in terms of consistency, but I just left it this way this way for stricter condition.

nntrainer/tensor/blas_interface.cpp
skykongkong8: Ah, I misunderstood. Thought you were asking about `sum32`.
But like I mentioned above, the compiler will internally compute them with `float` anyway, so think it won't make huge difference I guess.
Additionally, personally think this is better way i terms of consistency as well. SIMD functions for this computation is implemented is like : "multiply with f16 - and accumulate with f32"

test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
djeong20: makes sense. if so, I think comparing fp16 with fp32 could be pointless since it would never be close (especially when the dimension increases). how about we compare the fp16 tensor l2 norm value with other frameworks such as PyTorch or Tensorflow?

skykongkong8: Nice approach...
Then, is there a efficient way to compare Pytorch output and ourcase in unittest codes?
I don't really prefer hard-coded unittest cases like we do now...

djeong20: I tested the added test case, `l2norm_big768`, with PyTorch and Tensorflow. The L2 norm value for the float32 tensor matches with both frameworks. However, there's a mismatch for float16.

#### NNTrainer
Before: 45.25
After: 410.25

#### PyTorch
156.7500

#### Tensorflow
inf

This fix gets the closest value to the float32 tensor, which is 409.999, so there's no need to compare with other frameworks. I think it is good to go :)

skykongkong8: Based on the output from PyTorch / Tensorflow, I think they are just computing the value with full-fp16 precision.
Mathematically it is not really a wrong value, but it is not actually desired way in our framework I guess.
Anyway, thanks for pointing this issue out! 👍

baek2sm (approved):
LGTM

PR #2559, opened 2024-04-27
Add SELU activation function

kimhan0515:
- Now, user can use SELU activation function like torch or tensor flow.

**Self evaluation:**

Signed-off-by: kimhan0515 <kimhan0515@gmail.com>

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -81,6 +81,9 @@ public:
     case ActivationType::ACT_ELU:
       this->setActivation<T>(elu<T>, eluPrime<T>);
       break;
+    case ActivationType::ACT_SELU:
+      this->setActivation<T>(selu<T>, seluPrime<T>);
+      break;
     case ActivationType::ACT_SOFTPLUS:
       this->setActivation<T>(softplus<T>, softplusPrime<T>);
       break;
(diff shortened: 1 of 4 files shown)


skykongkong8 (changes requested):

test/unittest/unittest_nntrainer_activations.cpp
```diff
@@ -517,6 +517,58 @@ TEST(nntrainer_activation, eluPrime_01_p) {
...
+  nntrainer::Tensor selu_prime_result =
+    input.apply<float>(nntrainer::ActiFunc::seluPrime<float>);
+  float *data = selu_prime_result.getData();
+  ASSERT_NE(nullptr, data);
+
+  for (int i = 0; i < batch * height * width; ++i) {
+    EXPECT_NEAR(data[i], answer[i], tolerance);
+  }
+}
+
```
skykongkong8, line 571: Along with this, I would like to personally recommend implementing a half-precision TC as well.
You might need to add a new file like : `unittest_nntrainer_activations.cpp`

skykongkong8 (approved):
Overall, lgtm

test/unittest/unittest_nntrainer_activations.cpp
kimhan0515: I agree with the point!
Maybe it would be better to implement half-precision TCs for other types of activations as well as SELU, in separate PR!

nntrainer/layers/acti_func.h
```diff
@@ -667,6 +694,8 @@ class ActiFunc {
 private:
   constexpr static inline float alpha = 1.0f; /**< alpha for elu */
   constexpr static inline float beta = 1.0f;  /**< beta for Softplus */
+  constexpr static inline float selu_alpha = 1.67326324f; /**< alpha for selu */
+  constexpr static inline float selu_scale = 1.05070098f; /**< scale for selu */
```
baek2sm, line 698: As you may already know, setting the parameters of the activation function in this way is not a good direction. I understand that existing codes, especially Leaky ReLU, have already been implemented in this manner. So I know why you implemented it this way and I understand you. As for the next task, rather than adding an activation function, how about improving the structure so that the parameters of the activation function can be processed by receiving the property as input? It is optional but not essential for you.

baek2sm (approved):
LGTM

nntrainer/layers/acti_func.h
kimhan0515: Thanks for your opinion! I think that the task will be quite worthwhile.

PR #2562, opened 2024-04-30
[ Loss Scale ]  Enable to define different type of Variable and Gradient for Loss Scale

jijoongmoon:
## In this PR
This PR enables separate definition of variable and gradient of Weight. It's because we have to support different type considering Mixed Precision Training or Quantization.  
For example, Variable should be FP16 and Gradient should be FP16 for FP16-FP16 Mixed Precision. 
Or Variable should support BCQ3 and Gradient FP16 for BCQ3-FP16 QAT.
Therefore, we need to set update the Variable weight as it is requested.. and for Gradient, we have to set same type of Activation. 

This PR enables to set different type between Variable and Gradient.

<details><summary>[Property] Add loss scale property </summary><br />

It add loss scale property as model common property.

</details>

<details><summary>[ Weight ] Add Loss Scale factor in Weight</summary><br />

This PR enables the loss scale factor in Weight.
. Change the WeightSpec to incluide the loss factor
. Add LossScaleForMixed Property as an layer common property, so that
  it can set the scale factor in initContext.
. Add Loss Scale in initContext
. Set the LossScaleForMixed Property when there is LossScale Model
  Property

Resolves:

**Self evaluation:**

</details>

<details><summary>[ Weight ] split variable dim and grad dim to set separately </summary><br />
This PR split the Variable and Gradient Dim in Var_Grad and Weight.
By this way we can set different Variable Type and Gradient in Wegiht.
. add dim_g for gradient in WeightSpec.
. manager need to update to support WeightSpec.
. Create Tensors according to dim_v and dim_g
. Create Weight chaged in Weight.h

Resolves:

**Self evaluation:**

</details>

<details><summary>[ NEURALNET ] change the loss scale property to Rigid Property</summary><br />
Loss Scale is more like Rigid Property of model, rather than flexible
property.

Resolves:

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -62,7 +62,8 @@ public:
                    const std::string &n = "", const std::string &prefix_ = "",
                    const float max_norm = 0.0,
                    std::array<std::string, 3> tensor_type_ = {"NCHW", "FP32",
-                                                              "FP32"});
+                                                              "FP32"},
+                   const float loss_scale = 0.0);
   /**
    * @brief   get Tensor Format of Layer
    *
@@ -189,7 +190,7 @@ public:
                              bool trainable = true, unsigned int out_axis = 3) {
     weights_spec.emplace_back(dim, init, reg, reg_const, decay,
                               clip_by_global_norm, trainable,
-                              prefix + ":" + name, out_axis);
+                              prefix + ":" + name, out_axis, loss_scale);
     return weights_spec.size() - 1;
   }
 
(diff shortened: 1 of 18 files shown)


nntrainer/layers/layer_context.h
```diff
@@ -171,7 +172,8 @@ class InitLayerContext {
   /**
    * @brief Request a new weight for the layer
    *
-   * @param dim dimension of the weight
+   * @param dim_v dimension of Variagble of the weight
```
EunjuYang, line 175:
```suggestion
   * @param dim_v dimension of Variable of the weight
```

Also, `@param dim_v` doesn't seem to match the function prototype.

jijoongmoon: yes. My bad. Fixed. Thanks!

SeoHyungjun (approved):
LGTM

myungjoo (approved):

PR #2563, opened 2024-05-02
[Wait for #2562 ] [ Weight ] Add Var32 Tensor in Weight @open sesame 05/03 15:14 

jijoongmoon:
## In this PR
We will add Var32 Tensor if the Variable Weight is not Full
precision (FP32). This eables the Weight Update with full precision
and only Apply Gradient Process ueses this Tensor. Therefore, the
lifespan of this tensor should be "ApplyGradient".

Also,  modify TensorPool to generate Weigth considering Mixed Precision.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/weight.h
+++ b/nntrainer/tensor/weight.h
@@ -45,7 +45,8 @@ public:
     regularizer_constant(1.0f),
     decay(0.0f),
     clip_by_global_norm(0.0f),
-    output_axis(3) {}
+    output_axis(3),
+    loss_scale(0.0) {}
 
   /**
    * @brief Construct a new Weight object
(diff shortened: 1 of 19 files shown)


DonghakPark (approved):
LGTM! (need rebase)

nntrainer/tensor/weight.h
```diff
@@ -316,6 +317,16 @@ class Weight : public Var_Grad {
...
   }
 
+  /**
+   * @brief Check if the variable type is not full precision
+   *
+   * @return true if it is not full precsion
+   * @return false otherwise
+   */
+  bool isMixedPrecision() const {
+    return var->getDataType() == ml::train::TensorDim::DataType::FP32;
```
djeong20, line 327: I'm quite confused here. if this function returns true when the variable is **not** in full precision, shouldn't it be `!=` instead of `==`?

djeong20: fixed in #2566

djeong20 (approved):
LGTM

EunjuYang (approved):
LGTM

myungjoo:
@jijoongmoon Let's rebase and move on.

jijoongmoon:
closed by  #2663

PR #2565, opened 2024-05-03
[Property] Add loss scale property

myungjoo:
It add loss scale property as model common property.

This is a partial PR from #2455.

@jihochu Doing PR per commit for this huge PR (#2455) will acclerate merging.


jijoongmoon:
I think this is already in the #2562.  Please take a look.

PR #2569, opened 2024-05-07
[Application] fix input data error on example of LLaMa

baek2sm:
in case of running without encoder, a problem has been fixed where invalid values are set during operation due to incorrect assignment of input data, and this causes word index related errors.

**Self evaluation:**


skykongkong8 (approved):
Nice catch!

myungjoo (approved):

PR #2577, opened 2024-05-09
[ hgemm/bugfix ] Consider small hgemm case @open sesame 05/13 21:35

skykongkong8:
- Small means, for here, is smaller than: (4x4) * (4x4) 
- With hard-coded sized (4x4, 8x8, ...) kernel hgemm, matrix length that are smaller than unit vector length might fail.
- Toss such case to fallback. Such small sized gemm will take a few nanoseconds and therefore negligible.

**Self evaluation:**


SeoHyungjun (approved):
LGTM

skykongkong8:
With different commit this PR seems to be no longer needed. Close.

PR #2580, opened 2024-05-11
[Wait for #2574] [ Context ] Add loss scale in Context & using mse loss

jijoongmoon:
## In this PR
This PR add loss scale parameter in RunLayerContext and use it to update
mse loss.

. Add Loss Scale Parameter in RunLayerContext Constructor
. Add applyLossScale func to update return derivative in Loss Layer
. Change MSE Loss Layer to apply the loss scale to return derivative

**Self evaluation:**
1. Build test:	 [X]Passed [ ]Failed [ ]Skipped
2. Run test:	 [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: jijoong.moon <jijoong.moon@samsung.com>

PR #2581, opened 2024-05-13
[Wait for #2580] [ Mixed Precision ] Enable Mixed Precision 

jijoongmoon:
## In this PR
This PR enables Mixed Precision Training. For now only FP16-FP32
is considered. Additional Test cases will be added.

. add getSortedLayerIdx to set the graph order for forwarding.
. change clip_weights to lazy_apply_weights to use both cases.
. add fowarding_op to run forwarding from that layer which has a
gradient with nan.
. add a while loop for re-run backwarding after resetting the loss scale.
. add setLossScale in RunLayerContext
. add check the gradient if mixed precision is enabled.

**Self evaluation:**

### DIFF
--- a/Applications/KNN/jni/meson.build
+++ b/Applications/KNN/jni/meson.build
@@ -15,4 +15,4 @@ e = executable('knn_sample',
   install_dir: application_install_dir
 )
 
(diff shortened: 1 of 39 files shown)


jijoongmoon:
closed by  #2663

PR #2585, opened 2024-05-14
[ hgemm ] Support beta case @open sesame 06/04 13:13

skykongkong8:
- This commit allows hgemm to get beta condition as well.
- Note that beta for here is as follows:
```math
C = alpha * A * B + beta * C
```
- In many cases, beta is 0.F. However, we can fuse calculation process if we have to do something like:
```C++
Tensor A, B, C;
Tensor tmp;
bool transA = false;
bool transB = false;
float scalar = 1.1F // anything you want

// redundant case 1
tmp = A.dot(B);
C.multiply_i(scalar);
C = tmp.add(C);

// use beta
C = A.dot(B, C, transA , transB, scalar);

// redundant case 2
C = A.dot(B);
C.add_i(scalar);

// use beta
C = C.ones();
C = A.dot(B, C, transA , transB, scalar);

```
- In addition add zero-init code for beta = 0.F case. According to recent model profiling result, even for initialization, minimizing instruction is quite helpful for overall model latency reduction.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -1555,8 +1555,10 @@ void hgemm(const __fp16 *A, const __fp16 *B, __fp16 *C, uint32_t M, uint32_t N,
   // performing beta*C
   unsigned int idx = 0;
   unsigned int size = M * N;
+  unsigned int size8 = (size >> 3) << 3;
+  unsigned int size4 = (size >> 2) << 2;
   if (beta != 0.F) {
-    for (; idx < (size - idx) && (size - idx) >= 8; idx += 8) {
+    for (; idx < size8; idx += 8) {
       float16x8_t c =
         vmulq_n_f16(vld1q_f16(&C[idx]), static_cast<__fp16>(beta));
 
@@ -1564,7 +1566,7 @@ void hgemm(const __fp16 *A, const __fp16 *B, __fp16 *C, uint32_t M, uint32_t N,
       vst1q_f32(&C32[idx + 4], vcvt_f32_f16(vget_high_f16(c)));
     }
     // remaining 4
-    for (; idx < (size - idx) && (size - idx) >= 4; idx += 4) {
+    for (; idx < size4; idx += 4) {
       float16x4_t c = vmul_n_f16(vld1_f16(&C[idx]), static_cast<__fp16>(beta));
 
       vst1q_f32(&C32[idx], vcvt_f32_f16(c));
@@ -1576,7 +1578,7 @@ void hgemm(const __fp16 *A, const __fp16 *B, __fp16 *C, uint32_t M, uint32_t N,
     }
   } else {
     float32x4_t zeros = vmovq_n_f32(0.F);
-    for (; idx < (size - idx) && (size - idx) >= 4; idx += 4) {
+    for (; idx < size4; idx += 4) {
       vst1q_f32(&C32[idx], zeros);
     }
     for (; idx < size; idx++) {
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/hgemm/hgemm.h
```diff
@@ -181,6 +181,26 @@ void hgemm_noTrans_8x8(unsigned int M, unsigned int N, unsigned int K,
...
                        float alpha = 1.F, float beta = 0.F);
 
+/**
+ * @brief hgemm noTrans computation with 4x4 kernel : C = A*B,
+ *
+ * @param M length of the row of matrix A
+ * @param N length of the col of matrix B
+ * @param K length of the col of matrix A
+ * @param A input matrix A
+ * @param lda length of the col of matrix C
```
SeoHyungjun, line 191: This is the column length of matrix A, right?

nntrainer/tensor/hgemm/hgemm.h
```diff
@@ -181,6 +181,26 @@ void hgemm_noTrans_8x8(unsigned int M, unsigned int N, unsigned int K,
...
+ * @param B input matrix B
+ * @param ldb length of the col of matrix C
```
SeoHyungjun, line 193: This is the same too

nntrainer/tensor/hgemm/hgemm.h
```diff
@@ -181,6 +181,26 @@ void hgemm_noTrans_8x8(unsigned int M, unsigned int N, unsigned int K,
                        unsigned int ldb, __fp16 *C, unsigned int ldc,
                        float alpha = 1.F, float beta = 0.F);
 
+/**
+ * @brief hgemm noTrans computation with 4x4 kernel : C = A*B,
+ *
+ * @param M length of the row of matrix A
+ * @param N length of the col of matrix B
```
SeoHyungjun, line 188: Already have lda and ldb. Does this mean row?

SeoHyungjun (approved):
LGTM

nntrainer/tensor/blas_neon.cpp
```diff
@@ -1555,22 +1555,35 @@ void hgemm(const __fp16 *A, const __fp16 *B, __fp16 *C, uint32_t M, uint32_t N,
...
-
-    vst1q_f32(&C32[idx], vcvt_f32_f16(vget_low_f16(c)));
-    vst1q_f32(&C32[idx + 4], vcvt_f32_f16(vget_high_f16(c)));
-  }
-  // remaining 4
-  for (; idx < (size - idx) && (size - idx) >= 4; idx += 4) {
-    float16x4_t c = vmul_n_f16(vld1_f16(&C[idx]), static_cast<__fp16>(beta));
+  unsigned int size8 = (size >> 3) << 3;
+  unsigned int size4 = (size >> 2) << 2;
+  if (beta != 0.F) {
```
djeong20, line 1560: might not be a good idea to compare floating points. 
```suggestion
  if (std::fpclassify(beta) != FP_ZERO) {
```

nntrainer/tensor/hgemm/hgemm.cpp
```diff
@@ -31,7 +31,7 @@
 
 void hgemm_noTrans(const __fp16 *A, const __fp16 *B, float *C32, unsigned int M,
                    unsigned int N, unsigned int K, float alpha, float beta) {
-  if (alpha == 1.F && beta == 0.F) {
+  if (alpha == 1.F) {
```
djeong20, line 34: same here

djeong20 (approved):
other than floating point comparison, LGTM

myungjoo (approved):

PR #2586, opened 2024-05-17
[Nnstreamer-subplugin] Add save_path to setProperty

songgot:
- Add save_path to setProperty to save the model for each epoch.
- Remove model->save() call to avoid saving the current epoch result to the model when current epoch is interrupted

**Self evaluation:**

### DIFF
--- a/nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc
+++ b/nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc
@@ -555,7 +555,8 @@ void NNTrainer::NNTrainerImpl::trainModel() {
   ml_logd("pid[%d], tid[%d]", pid, tid);
 
   try {
-    model->setProperty({"epochs=" + std::to_string(num_epochs)});
+    model->setProperty(
+      {"epochs=" + std::to_string(num_epochs), "save_path=" + model_save_path});
   } catch (const std::exception &e) {
     ml_loge("Error %s, %s", typeid(e).name(), e.what());
     return;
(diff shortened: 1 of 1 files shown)


nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc
```diff
@@ -574,14 +575,6 @@ void NNTrainer::NNTrainerImpl::trainModel() {
...
   }
 
-  try {
-    ml_logd("Save_model: %s", model_save_path.c_str());
-    model->save(model_save_path, ml::train::ModelFormat::MODEL_FORMAT_BIN);
-
-  } catch (const std::exception &e) {
-    ml_loge("Error %s, %s", typeid(e).name(), e.what());
-    return;
-  }
```
myungjoo, line 584: Then, when do you call ```save()``` ?

songgot: If set as a property, the model is updated at the end of each epoch.
When the stop callback is called, the current epoch is executed and the model is not saved.
```nntrainer``` continues to request data until the end of the current epoch, I double-checked with Choi Hyeongjun.

**Review by DonghakPark — APPROVED** [review]
LGTM!

**Review by SeoHyungjun — APPROVED** [review]
LGTM

File: nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc [hunk 2]
```diff
@@ -555,7 +555,8 @@ void NNTrainer::NNTrainerImpl::trainModel() {
   ml_logd("pid[%d], tid[%d]", pid, tid);
 
   try {
-    model->setProperty({"epochs=" + std::to_string(num_epochs)});
+    model->setProperty(
+      {"epochs=" + std::to_string(num_epochs), "save_path=" + model_save_path});
```
jijoongmoon, line 559: Not this PR, but I wonder how we can handle it if the user wants to set the "save_best_path".

PR #2587, opened 2024-05-17
[ RMSProp ] add rms prop optimizer

jijoongmoon:
## In this PR
This PR implements rms prop optimizer.
unittest for rms prop is required.
. implementation of RMSProp Class
. Separate the optimizer properties into optimizer common properties. and some fixes.
. add enum for rms props

**Self evaluation:**

### DIFF
--- a/nntrainer/optimizers/rmsprop.cpp
+++ b/nntrainer/optimizers/rmsprop.cpp
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Jijoong Moon <jijoong.moo@samsung.com>
+ *
+ * @file   rmsprop.cpp
+ * @date   17 May 2024
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is the RMSProp optimizer.
+ */
+
+#include <cmath>
+#include <fstream>
+
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <node_exporter.h>
+#include <rmsprop.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+RMSProp::RMSProp() :
+  rmsprop_props(props::Rho(), props::PropsEpsilon(), props::TorchRef()) {
+  /** default properties */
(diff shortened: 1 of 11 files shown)


nntrainer/optimizers/rmsprop.cpp
```diff
@@ -0,0 +1,100 @@
...
+  Tensor &wv = context.getOptimizerVariable(RMSPropParams::wv);
+
+  wv.multiply_i(rho);
+  wv.add_i(x_grad.multiply(x_grad), 1.0f - rho);
+
+  if (torch_ref) {
+    Tensor denom = wv.apply<float>(sqrtFloat<float>);
+    denom.add_i(epsilon);
+
+    wv.divide(denom, x_grad);
```
EunjuYang, line 86: Original code seems  to compute
`W = W - alpha * wv / (sqrt(wv) + epsilon)`, which should be
`W = W - alpha * x_grad / (sqrt(wv) + epsilon)` to be properly applied.
```suggestion
    x_grad.divide_i(denom);
```

DonghakPark (approved):
I think the CI-related error is a rebase issue. In my local environment, when I built with 2584 as the base, it passed build & test without any problems. Other than that, LGTM!!

nntrainer/optimizers/rmsprop.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    context.applyGradient(context.getLearningRate(), x_grad);
+
+  } else {
+    std::function<double(double)> sqrtEps = [epsilon](double f) {
+      return 1 / (sqrtDouble(f) + epsilon);
+    };
+
+    x_grad = wv.apply<float>(sqrtEps, x_grad);
+    context.applyGradient(context.getLearningRate(), x_grad);
```
lhs8928, line 96: RumOptimizerContext does not have a function which has a prototype of applyGradient(double, tensor) as the CI issued.

nntrainer/optimizers/rmsprop.cpp
lhs8928: It seems eunju is right.
Further more unlike the adam optimizer torch seems to work in the same way as other frameworks do.
Please refer https://pytorch.org/docs/stable/generated/torch.optim.RMSprop.html

lhs8928 (changes requested):

PR #2595, opened 2024-05-21
[unittest/gpu] Added LayerSemanticsGpu test suite

s-debadri:
Added unit test for layer semantics on GPU:
- `LayerSemanticsGpu` test suite added.
- Modified `layers_[standalone|dependent]_common_test.cpp` to incorporate the test cases.

Resolved a bug:
- Removed `run_context->setComputeEngine(...)` call from LayerNode since `run_context` doesn't get initialized in the beginning.
- Used `LayerNode` to store compute engine info which can be propagated as required.

> Key for new implemented layers on GPU will be different which can not be fetched from `app_context`. Hence these changes were made to fetch the keys from `cl_context` as required.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -152,7 +152,15 @@ createLayerNode(const ml::train::LayerType &type,
  */
 std::unique_ptr<LayerNode>
 createLayerNode(const std::string &type,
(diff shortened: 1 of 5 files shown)


jijoongmoon (approved):
LGTM

PR #2599, opened 2024-05-22
[blas/neon] isamax improvement for larger input length

### LINKED ISSUE #2593: Bug in `max_abs()` function in FP16 Tensor
Found an error in `Tensor::max_abs()` with FP16 dataType.
Strongly believe that this issue has to do with `isamax()` in `blas_neon.cpp`.
You can reproduce the same error with the TC below. 

```C++
TEST(nntrainer_Tensor, max_abs_768) {

nntrainer::TensorDim::TensorType t_type_nchw_fp16 = {
    nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP16};

  nntrainer::TensorDim::TensorType t_type_nchw_fp32 = {
    nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP32};

  size_t batch = 1;
  size_t channel = 1;
  size_t height = 768;
  size_t width = 768;

  nntrainer::Tensor input(
    nntrainer::TensorDim(batch, channel, height, width, t_type_nchw_fp16));

  nntrainer::Tensor input_fp32(
    nntrainer::TensorDim(batch, channel, height, width, t_type_nchw_fp32));

  const float alpha = 1e-1;
  const int MOD = 10;

  GEN_TEST_INPUT(input, ((k * l * (batch * height * channel) +
                          l * (batch * height) + k * (width) + l + 1) %
                         MOD) *
                          alpha);
  GEN_TEST_INPUT(input_fp32, ((k * l * (batch * height * channel) +
                               l * (batch * height) + k * (width) + l + 1) %
                              MOD) *
                               alpha);


  __fp16 result_neon = input.max_abs();
  float result_fp32 = input_fp32.max_abs();

  float absErrorNeon =
    std::abs(result_neon - result_fp32);

  const float epsilon = 1e-3;

  EXPECT_EQ(result_neon, result_fp32); // maybe a little bit different
  EXPECT_IN_RANGE(absErrorNeon, 0, epsilon); // but should reside in this range
}

```

s-debadri:
Fixes #2593 :
- Used `uint32_t` operations to process indices larger than `65535`.
- Added unittest of shape(1,1,768,768) for `max_abs()` which calls `isamax()`.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -1508,7 +1508,8 @@ unsigned int isamax(const unsigned int N, const __fp16 *X) {
   unsigned int idx = 8;
 
   // processing batch of 8
-  for (; (N - idx) >= 8; idx += 8) {
+  // stop before idx reaches UNIT16_MAX
+  for (; ((N - idx) >= 8) && ((idx + 8) <= UINT16_MAX); idx += 8) {
     float16x8_t values = vld1q_f16(&X[idx]);
     curr_index = vaddq_u16(curr_index, stride);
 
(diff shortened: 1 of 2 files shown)


djeong20 (approved):
nice work!
one question is why should the computation be separated under and over UINT16_MAX.
couldn't be combined into one case (under UINT32_MAX)?

skykongkong8:
> nice work! one question is why should the computation be separated under and over UINT16_MAX. couldn't be combined into one case (under UINT32_MAX)?

Think I can reply for that.
It is because using 8-length-vector would be x2 times faster than using 4-length-vector

nntrainer/tensor/blas_neon.cpp
```diff
@@ -1508,7 +1508,8 @@ unsigned int isamax(const unsigned int N, const __fp16 *X) {
   unsigned int idx = 8;
 
   // processing batch of 8
-  for (; (N - idx) >= 8; idx += 8) {
+  // stop before idx reaches UNIT16_MAX
```
myungjoo, line 1511: Why? (index is not stored in fp16...)

test/unittest/unittest_nntrainer_tensor_neon_fp16.cpp
```diff
@@ -384,6 +384,47 @@ TEST(nntrainer_Tensor, max_abs) {
   EXPECT_NEAR(result_neon, result_fp32, epsilon);
 }
 
+TEST(nntrainer_Tensor, max_abs_768) {
```
myungjoo, line 387: Please add test cases for isamax, verifying its results, where:
1. N == UINT16_MAX
2. N == UINT16_MAX - 1
3. N >> UINT16_MAX

skykongkong8 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #2606, opened 2024-05-23
[GPU/OpenCL] Initial version of Addition Layer with OpenCL ops

yashSingh0723:
Added initial version of Addition Layer for GPU. This is a basic implementation using naive kernel.
Changes added with this PR:

- `addition_layer_cl.cpp` added containing the new `AdditionLayerCL` class for OpenCL implementation.
- Updated `LayerKernel` enum inside `layer_context.h.`
- Added `unittest_layers_addition_cl.cpp` to test Addition Layer on GPU.

Signed-off-by: Yash Singh [yash.singh@samsung.com](mailto:s.debadri@samsung.com)

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -354,21 +354,11 @@ Reshape(const std::vector<std::string> &properties = {}) {
 /**
  * @brief Helper function to create addition layer
  */
-inline std::unique_ptr<Layer>
-Addition(const std::vector<std::string> &properties = {}) {
-  return createLayer(LayerType::LAYER_ADDITION, properties);
-}
-
-#ifdef ENABLE_OPENCL
-/**
- * @brief Helper function to create Addition layer for GPU
- */
-inline std::unique_ptr<Layer>
-AdditionCL(const std::vector<std::string> &properties = {},
-           const LayerComputeEngine &compute_engine = LayerComputeEngine::CPU) {
+inline std::unique_ptr<Layer> Addition(
+          const std::vector<std::string> &properties = {},
+          const LayerComputeEngine &compute_engine = LayerComputeEngine::CPU) {
   return createLayer(LayerType::LAYER_ADDITION, properties, compute_engine);
 }
-#endif
 
 /**
  * @brief Helper function to create concat layer
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -359,6 +359,17 @@ Addition(const std::vector<std::string> &properties = {}) {
   return createLayer(LayerType::LAYER_ADDITION, properties);
 }
 
+#ifdef ENABLE_OPENCL
+/**
+ * @brief Helper function to create Addition layer for GPU
+ */
+inline std::unique_ptr<Layer>
+AdditionCL(const std::vector<std::string> &properties = {},
+           const LayerComputeEngine &compute_engine = LayerComputeEngine::CPU) {
+  return createLayer(LayerType::LAYER_ADDITION, properties, compute_engine);
+}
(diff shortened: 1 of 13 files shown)


api/ccapi/include/layer.h
```diff
@@ -359,6 +359,17 @@ Addition(const std::vector<std::string> &properties = {}) {
   return createLayer(LayerType::LAYER_ADDITION, properties);
 }
 
+#ifdef ENABLE_OPENCL
+/**
+ * @brief Helper function to create Addition layer for GPU
+ */
+inline std::unique_ptr<Layer>
+AdditionCL(const std::vector<std::string> &properties = {},
```
s-debadri, line 367: Don't need to create `AdditionCL` separately. Existing `Addition` API can be modified with `LayerComputeEngine`. Since both of them use the same `LayerType::LAYER_ADDITION`.

nntrainer/layers/cl_layers/addition_layer_cl.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+  if (input.getDataType() == ml::train::TensorDim::DataType::FP32) {
+    unsigned int size = input.size();
+    const float *data = input.getData();
+    float *rdata = result.getData();
+
+    addition_cl(data, rdata, size, context);
+
+  } else
+    throw std::invalid_argument("Error: OpenCL fp16 is not supported yet.");
```
jijoongmoon, line 83: I think it is because we are not implementing FP16 yet. isn't it? If it is, then I think supporting FP16 is more important because the LLM is working based on FP16.

nntrainer/layers/cl_layers/addition_layer_cl.cpp
```diff
@@ -0,0 +1,210 @@
...
+    /** @todo check possibility for in-place of addition layer */
+    for (unsigned int idx = 0; idx < context.getNumInputs(); ++idx) {
+      const Tensor &input_ = context.getInput(idx);
+      TensorDim input_dim = input_.getDim();
+
+      TensorDim input_step_dim = input_dim;
+      input_step_dim.batch(1);
+      input_step_dim.height(to - from);
+
+      Tensor input_step = input_.getSharedDataTensor(
```
jijoongmoon, line 177: I wonder if we do like this way, we need to data transfer from CPU to GPU. Doesn't it affect the latency?

api/ccapi/include/layer.h
yashSingh0723: Updated in the new commit.
Thanks.

nntrainer/layers/cl_layers/addition_layer_cl.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+#include <addition_layer_cl.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <node_exporter.h>
+#include <util_func.h>
+
+#include <layer_context.h>
+
+std::string addition_cl_kernel_ =
```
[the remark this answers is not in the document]
yashSingh0723: Updated in the new commit.

nntrainer/layers/cl_layers/addition_layer_cl.cpp
yashSingh0723: Added FP16 support for Addition Layer in the latest commit.

s-debadri (approved):
LGTM!

myungjoo (approved):
@yashSingh0723 LGTM. However, please address Jijoong's concern before merging.

jijoongmoon (approved):
LGTM

PR #2591, opened 2024-05-20
[Wait for #2584] [ TEST ] add torch input and output test data for mixed precision 

jijoongmoon:
## In this PR
This PR adds torch mixed precision golden data generation and input and
output for the test.

. some fixes to test.

Resolves:

**Self evaluation:**
1. Build test:	 [X]Passed [ ]Failed [ ]Skipped
2. Run test:	 [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: jijoong.moon <jijoong.moon@samsung.com>

PR #2607, opened 2024-05-24
[Wait for #2591] [ TEST ] Add more unittest and fixes for mixed precision 

jijoongmoon:
## In this PR

This PR includes more unittest and fixes for mixed precision.
. Model Unittest
  . 2 fc layer which generates NaN or Inf Gradient from Troch.
  . MSE Loss and Check the whole procedure of the mixed precision training.
  . Even if the FC model only has one weight, it is good enough
  to validate the mixed precision.
  . Torch model also works similar way to NNTrainer.
  . Some fixes about the execution order of applying gradient when the
  mixed precision is on.
  . Update SGD to support Mixed Precision training

- Added TOC generator for README.md

Resolves:

**Self evaluation:**


nntrainer/layers/layer_node.cpp
```diff
@@ -180,6 +180,7 @@ LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l) :
   inplace(InPlace::NONE),
   needs_calc_derivative(false),
   needs_calc_gradient(false),
+
```
SeoHyungjun, line 183: It looks like an unnecessary line. Is there a reason you added it?

SeoHyungjun (approved):
LGTM

jijoongmoon:
closed by  #2663

PR #2608, opened 2024-05-28
[GPU/OpenCL] Check fp16(half) support

s-debadri:
Added check for OpenCL fp16 `half` support on device.
Will show proper message if not found.

To use fp16 enable `cl_khr_fp16` as following with the kernels.
```cpp
#pragma OPENCL EXTENSION cl_khr_fp16 : enable
    
 __kernel void kernel_name(...) { } 
```

### DIFF
--- a/nntrainer/opencl/opencl_context_manager.cpp
+++ b/nntrainer/opencl/opencl_context_manager.cpp
@@ -155,6 +155,31 @@ bool ContextManager::CreateDefaultGPUDevice() {
   device_id_ = devices[0];
   platform_id_ = platform_id_;
 
+#ifdef ENABLE_FP16
+  // check for fp16 (half) support available on device
+  // getting extensions
+  size_t extension_size;
+  status =
+    clGetDeviceInfo(device_id_, CL_DEVICE_EXTENSIONS, 0, NULL, &extension_size);
+  if (status != CL_SUCCESS) {
+    ml_loge("clGetDeviceInfo returned %d", status);
+    return false;
+  }
+
+  char extensions[extension_size];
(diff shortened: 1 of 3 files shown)


nntrainer/opencl/opencl_context_manager.cpp
```diff
@@ -155,6 +155,31 @@ bool ContextManager::CreateDefaultGPUDevice() {
   device_id_ = devices[0];
   platform_id_ = platform_id_;
 
+#ifdef ENABLE_FP16
+  // check for fp16 (half) support available on device
```
djeong20, line 159: which half-precision does it support? `_Float16`, `__fp16`, or `half`

djeong20 (approved):

nntrainer/opencl/opencl_context_manager.cpp
s-debadri: Ideally all of `_Float16`, `__fp16` and `half` can be used interchangeably inside kernels when `cl_khr_fp16` extension is enabled. I have tested with `__fp16` and `half` myself, they worked exactly the same. However I did not test with `__Float16` but it is supported by OpenCL 2.1 or later.

s-debadri: `half` is specifically designed for use in OpenCL kernels. Some are architecture specific  (e.g: `__fp16` for ARM). So `half` will be more portable and benefit from further optimizations..

s-debadri: `half` will only work inside OpenCL kernels. So I think adding it to tensor data type won't be useful in this scenario.

myungjoo (approved):

PR #2609, opened 2024-05-29
[GPU/OpenCL] Added fp16 support for FC layer on GPU

s-debadri:
FC Layer GPU kernels added for `fp16` operation:
- Added `blas_kernels_fp16.cpp` for BLAS `fp16` OpenCL kernels.
- Used `lda` for SGEMV computation for generalization.
- Unit tests added for FC Layer `fp16` support on GPU.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/cl_layers/blas_kernels.h
+++ b/nntrainer/layers/cl_layers/blas_kernels.h
@@ -25,8 +25,11 @@ namespace nntrainer {
  * @brief declaring global kernel objects
  */
 extern opencl::Kernel kernel_sgemv;
+extern opencl::Kernel kernel_sgemv_fp16;
 extern opencl::Kernel kernel_sgemm;
+extern opencl::Kernel kernel_sgemm_fp16;
 extern opencl::Kernel kernel_dot;
+extern opencl::Kernel kernel_dot_fp16;
 
 /**
  * @brief     sgemv computation : Y = A*X + Y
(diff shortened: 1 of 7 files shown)


nntrainer/layers/cl_layers/blas_kernels.cpp
```diff
@@ -115,7 +115,7 @@ void sgemv_cl(const float *matAdata, const float *vecXdata, float *vecYdata,
       break;
     }
 
-    result = kernel_sgemv.SetKernelArguments(4, &dim2, sizeof(int));
+    result = kernel_sgemv.SetKernelArguments(4, &lda, sizeof(int));
```
myungjoo, line 118: Is this a bugfix? It doesn't look like an fp16 implementation.

Could you please separate bugfix commit and feature-implementation commit?
Mixing up the two topics in a single commit confuses reviewers.

myungjoo (approved):
Please identify the changes in blas_kernels.cpp before merging. It appears unrelated to other changes.

PTAL: @skykongkong8 @lhs8928

EunjuYang (approved):
It's good to read your contributions in GPU enablement. One quick question. Do you have a plan to further improve the kernels? e.g., `sgemv_cl_kernel`'s parallel level is one thread per one component of out vector, which can be further parallelized. 
It would be great to know the current speed-up status compared to CPU.

nntrainer/layers/cl_layers/blas_kernels.cpp
s-debadri: I have used `lda`  in SGEMV kernel to generalize it for future uses.

skykongkong8:
> Please identify the changes in blas_kernels.cpp before merging. It appears unrelated to other changes.
> 
> PTAL: @skykongkong8 @lhs8928

It was one of my suggestions to use terms like lda, ldb, or ldc from previous reviews, although it might have been better to separate feature-implementation commit and bugfix commit. I could confirm current implementation is more desirable than before

skykongkong8 (approved):

jijoongmoon:
Not this PR, blas_kernel code needs to be under the tensor directory for better maintenance.

PR #2611, opened 2024-05-30
[layer] added start/end dimension in flatten layer

lhs8928:
- For now flatten layer flatten all dimension except batch. This commit will be able to flatten only the sub dimensions

Signed-off-by: hyeonseok <hs89.lee@samsung.com>

### DIFF
--- a/nntrainer/layers/flatten_layer.cpp
+++ b/nntrainer/layers/flatten_layer.cpp
@@ -6,6 +6,7 @@
  * @date   16 June 2020
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author hyeonseok Lee <hs89.lee@samsung.com>
  * @bug	   No known bugs except for NYI items
  * @brief  This is Flatten Layer Class for Neural Network
  *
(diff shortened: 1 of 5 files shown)


EunjuYang (approved):
LGTM!

nntrainer/layers/flatten_layer.cpp
```diff
@@ -49,6 +71,7 @@ void FlattenLayer::setProperty(const std::vector<std::string> &values) {
 
 void FlattenLayer::exportTo(Exporter &exporter,
                             const ml::train::ExportMethods &method) const {
+  exporter.saveResult(flatten_props, method, this);
```
DonghakPark, line 74: When exporting Tflite, since graphs and tensors are changed for export, it seems that if you call saveResult twice, it will not work properly.

DonghakPark: I will also look for a way to solve this problem.

SeoHyungjun (approved):
LGTM

skykongkong8 (approved):
LGTM

nntrainer/layers/flatten_layer.cpp
```diff
@@ -39,7 +60,8 @@ void FlattenLayer::finalize(InitLayerContext &context) {
 }
 
 void FlattenLayer::setProperty(const std::vector<std::string> &values) {
-  auto remain_props = loadProperties(values, reshape_props);
+  auto remain_props = loadProperties(values, flatten_props);
+  remain_props = loadProperties(remain_props, reshape_props);
```
myungjoo, line 64: Uh... overwriting results into the same variable? Why? The first call becomes useless...

djeong20 (approved):
LGTM

nntrainer/layers/flatten_layer.cpp
lhs8928: Removed this line

nntrainer/layers/flatten_layer.cpp
lhs8928: Removed this line

DonghakPark (approved):
LGTM!!!

nntrainer/layers/flatten_layer.cpp
myungjoo: You didn't remove it....

lhs8928: I'm sorry I was confused with @DonghakPark 's comment.
The function "loadProperties" loads only those properties from the first given list of properties that correspond to the second given argument list, and returns any non-matching arguments.
So first call is needed to load the "flatten properties" from the "values".

PR #2622, opened 2024-06-05
[WIP][Application] fix some settings for LLaMA @open sesame 06/07 14:56

baek2sm:
The ENABLE_ENCODER2 option allows you to use specific tokenizer file on server.

However, this option does not fit the general user usage scenario because the tokenizer varies from every llama version to be distributed. In addition, the situation in which tokenizer is not used at all does not conform to the general usage scenario, too.

So, now the default setting is to always use the tokenizer of the path specified by the user.

**Self evaluation:**

### DIFF
--- a/Applications/LLaMA/jni/main.cpp
+++ b/Applications/LLaMA/jni/main.cpp
@@ -30,18 +30,21 @@
 #include <swiglu.h>
 #include <transpose_layer.h>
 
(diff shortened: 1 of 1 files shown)


baek2sm:
I will close this pr because we can't use the tokenizer in tizen environment now, so we can't set tokenizer as default.

PR #2635, opened 2024-06-13
android: consistant ML_API_COMMON macro

myungjoo:
ML_API_COMMON macro has been inconsistent for Android build, where it is force-defined 1 in Android.mk while it may become 0 or 1 depending on build system in meson.

Because android build uses meson and Android.mk simultaneously, this must become consistent.


EunjuYang (approved):
LGTM

DonghakPark (approved):
LGTM!

PR #2637, opened 2024-06-13
action: Yocto devtool test

myungjoo:
Test if a pull request breaks Yocto build or not.

### DIFF
--- a/.github/workflows/yocto.yml
+++ b/.github/workflows/yocto.yml
@@ -1,0 +1,55 @@
+name: Yocto devtool test
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    name: Build with Yocto devtool on Ubuntu
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/yocto.yml
```diff
@@ -0,0 +1,55 @@
+name: Yocto devtool test
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    name: Build with Yocto devtool on Ubuntu
+    runs-on: ubuntu-20.04
```
DonghakPark, line 10: Will there be no job on Ubuntu 22.04?

DonghakPark (approved):

.github/workflows/yocto.yml
myungjoo: Yocto devtool is a build tool with isolated env; it's like "gbs" for "Yocto OS". We don't need to test with multiple Ubuntu versions.

skykongkong8 (approved):

PR #2642, opened 2024-06-18
[ layer ] Lower the computational code of lstmcell_core to BLAS level

skykongkong8:
Through this PR, 2 things:
1. Cleaner code : reduce if/def code block to tensor level
2. Acceleration : Use SIMD haxpy in fp16 case. Previously it was only using naive loop

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -1005,6 +1005,24 @@ int Tensor::add_i(Tensor const &m, float const alpha) {
   return ML_ERROR_NONE;
 }
 
+int Tensor::add_i_partial(unsigned int len, unsigned int addr_idx, Tensor &m,
+                          unsigned int incX, unsigned int incY,
+                          const Tensor alphas, unsigned int alpha_idx) {
+  if (dim.getDataType() == ml::train::TensorDim::DataType::FP32) {
+    saxpy(len, alphas.getValue<float>(alpha_idx), m.getData<float>(), incX, getAddress<float>(addr_idx),
+          incY);
+  } else if (dim.getDataType() == ml::train::TensorDim::DataType::FP16) {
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -1005,6 +1005,24 @@ int Tensor::add_i(Tensor const &m, float const alpha) {
...
 
+int Tensor::add_i_partial(unsigned int len, unsigned int addr_idx, Tensor &m,
+                          unsigned int incX, unsigned int incY,
+                          const Tensor alphas, unsigned int alpha_idx) {
+  if (dim.getDataType() == ml::train::TensorDim::DataType::FP32) {
+    saxpy(len, alphas.getValue<float>(alpha_idx), m.getData<float>(), incX,
+          getAddress<float>(addr_idx), incY);
+  } else if (dim.getDataType() == ml::train::TensorDim::DataType::FP16) {
+#ifdef ENABLE_FP16
+    saxpy(len, alphas.getValue<_FP16>(alpha_idx), m.getData<_FP16>(), incX,
```
SeoHyungjun, line 1016: 2. Acceleration : Use SIMD haxpy in fp16 case. Previously it was only using naive loop.

In your comment, you mentioned that you are using haxpy, do you mean saxpy? 
Or is SIMD haxpy equivalent to saxpy?

SeoHyungjun (approved):
LGTM

nntrainer/tensor/tensor.cpp
skykongkong8: It will internally use haxpy in `blas_neon.cpp` but the interface function of using that is named as `saxpy`.
This needs fixing in the near future

jijoongmoon (approved):
LGTM

PR #2641, opened 2024-06-17
[CI] Add fp-16 build in github action

heka1024:
Closes #2560. You can now see the build results from Fp16 in a Github action.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: heka1024 <heka1024@gmail.com>

Issue #2560, opened 2024-04-29
Add `-Denable-fp16=true` build to CI


heka1024:
Some code could break build with `-Denable-fp16=true` option, like https://github.com/nnstreamer/nntrainer/pull/2545. I think it would be helpful to add this build to CI to ease the burden on reviewers.

heka1024:
@skykongkong8 What do you think?

skykongkong8:
Good point.
As far as I am concerned, Reason why `-Denable-fp16` is not included in current CI is because it is relatively newer option than the others.
Even for now, I personally think that it is not really matured to be checked for every PR, because we have multiple plans to refactorize the structure of half-precision formats. I need more opinions about this. @jijoongmoon

myungjoo:
You may add enable-fp16 build test in CI.
You may limit test coverage or error reporting mechanisms for enable-fp16, too. (you can let github-action gracefully ignore failed TCs, too, so that we can see the number of failed cases but the CI says "OK".)

Anyone needing this may add it.

heka1024:
@skykongkong8 @myungjoo I'll add this to CI. Thanks for comment!

skykongkong8:
Great! Thanks for pointing this issue out :)

PR #2647, opened 2024-06-20
action/ubuntu: fp16 on/off handled by matrix

myungjoo:
To de-dup scripts handling fp16-enabled and fp16-disabled cases for meson-clean build, add another matrix entry.

Reference: https://github.com/nnstreamer/nntrainer/pull/2641/files#r1646986341

CC: @heka1024

### DIFF
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -11,6 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-20.04, ubuntu-22.04 ]
+        meson_options: [ "-Denable-fp16=false", "-Denable-fp16=true" ]
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


heka1024 (commented):
I think I read in https://github.com/nnstreamer/nntrainer/issues/2560#issuecomment-2082273173 that the Fp16 build was supposed to be optional. But I think this pull request make a build on fp16 as required. Should we make it required?

myungjoo:
> I think we said in [#2560 (comment)](https://github.com/nnstreamer/nntrainer/issues/2560#issuecomment-2082273173) that the Fp16 build should be optional. But I think this pull request make a build on fp16 as required. Should we make it required?

fp16 build should be kept optional. Users should be able to enable/disable fp16 for their own build.

fp16 integration should be kept mandatory. Any commit should NEVER break fp16 support, which will keep users being able to enable/disable fp16 for their own build. Without this mandate, users won't be able to enable fp16 in the near future.

heka1024 (approved):
It seems much nicer. Thank you!

DonghakPark (approved):
it's much better !! LGTM!!

djeong20 (approved):
LGTM

SeoHyungjun (approved):
LGTM!

PR #2651, opened 2024-06-26
[GPU/OpenCL] Initial version of Reshape Layer with OpenCL ops

niket-agarwal:
Added initial version of Reshape Layer for GPU. This is a basic implementation using naive kernel.

**Changes added with this PR:**

- `reshape_cl.cpp` added containing the new `ReshapeLayerCL` class for OpenCL implementation.
- Added `unittest_layers_reshape_cl.cpp` to test Reshape Layer on GPU.

Signed-off-by: Niket Agarwal [niket.a@samsung.com](mailto:niket.a@samsung.com)

### DIFF
--- a/nntrainer/layers/cl_layers/reshape_cl.cpp
+++ b/nntrainer/layers/cl_layers/reshape_cl.cpp
@@ -1,0 +1,325 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Niket Agarwal <niket.a@samsung.com>
+ *
+ * @file   reshape_cl.cpp
+ * @date   18 June 2024
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Niket Agarwal <niket.a@samsung.com>
+ * @bug	   No known bugs except for NYI items
+ * @brief  This is Reshape GPU Layer Implementation
+ */
+
+#include <iostream>
+#include <layer_context.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <node_exporter.h>
+#include <reshape_cl.h>
+
+std::string copy_cl_kernel_fp16_ =
+  R"(
+    #pragma OPENCL EXTENSION cl_khr_fp16 : enable
+    __kernel void copy_cl_fp16(__global const half* input, 
+                               __global half* output,
+                               const int batchsize, 
+                               const int channels, 
+                               const int height, 
+                               const int width) {
+
+    int i= get_global_id(0);
+    output[i] = input[i];
+    
+})";
+
+std::string copy_cl_kernel_ =
+  R"(__kernel void copy_cl(__global const float* input, 
+                               __global float* output,
+                               const int batchsize, 
+                               const int channels, 
+                               const int height, 
+                               const int width) {
+    
+    int i= get_global_id(0);
+    output[i] = input[i];
+
+})";
+
+namespace nntrainer {
+
+static constexpr size_t SINGLE_INOUT_IDX = 0;
+
+void ReshapeLayerCl::finalize(InitLayerContext &context) {
+  NNTR_THROW_IF(context.getNumInputs() != 1, std::invalid_argument)
+    << "Reshape only supports 1 input for now";
+
+  const TensorDim &in_dim = context.getInputDimensions()[0];
+
+  auto &target_shape = std::get<props::TargetShape>(reshape_props);
+  NNTR_THROW_IF(target_shape.empty(), std::invalid_argument)
+    << "Reshape layer must be provided with target shape";
+  TensorDim out_dim = target_shape.get();
+
+  if ((int)out_dim.getDataLen() == -1) {
+    out_dim.height(1);
+    out_dim.channel(1);
+    out_dim.width(in_dim.getFeatureLen());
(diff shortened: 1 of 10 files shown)


djeong20:
just wondering. why do we need this ReshapeLayer in the first place?
I think operations (e.g., forwarding, backwarding) would be a simple data copy that doesn't change any order of the data.

nntrainer/layers/cl_layers/reshape_cl.cpp
```diff
@@ -0,0 +1,358 @@
...
+        int input_index = batch_index * channels * height * width +
+                          input_channel * height * width +
+                          input_height * width +
+                          input_width;
+
+        int output_index = batch_index * elements_per_batch + element_index;
+        
+        output[output_index] = input[input_index];
+    }
+})";
```
djeong20, line 78: if the whole process is simply copying, couldn't it be as follows?

```
int i= get_global_id(0);
output[i] = input[i]
```

nntrainer/layers/cl_layers/reshape_cl.h
```diff
@@ -0,0 +1,154 @@
...
+  inline static const std::string type = "reshape";
+
+  static opencl::Kernel kernel_reshape;
+  static opencl::Kernel kernel_reshape_fp16;
+
+  /**
+   * @brief Process data and dimensions for reshape operation
+   * @param[in] input Tensor
+   * @param[in] result Tensor
+   * @param[in] RunLayerContext reference
```
SeoHyungjun, line 110: In the functions below, the same argument is written with a different explanation.

```suggestion
   * @param[in] context RunLayerContext reference
```

test/unittest/layers/unittest_layers_reshape_cl.cpp
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Niket Agarwal <niket.a@samsung.com>
+ *
+ * @file unittest_layers_reshape_cl.cpp
+ * @date 18th June 2024
+ * @brief Reshape Layer Test
+ * @see	https://github.com/nnstreamer/nntrainer
+ * @author Niket Agarwal <niket.a@samsung.com>
+ * @bug No known bugs except for NYI items
```
SeoHyungjun, line 10: The indentation of the code is slightly different.
In addition, it would be better to change the date format and the order of @brief for unity with other parts.

```suggestion
 * @file   unittest_layers_reshape_cl.cpp
 * @date   18 June 2024
 * @see    https://github.com/nnstreamer/nntrainer
 * @author Niket Agarwal <niket.a@samsung.com>
 * @bug    No known bugs except for NYI items
 * @brief  Reshape Layer Test
```

SeoHyungjun (approved):
LGTM

nntrainer/layers/cl_layers/reshape_cl.cpp
```diff
@@ -0,0 +1,358 @@
...
+
```
EunjuYang, line 79: This kernel only holds for the case, where `{dimsize,0,0}`.
If you intended this global-level parallelization only, @djeong20 's suggestion seems to be better.
If you find more generalized one, however, please update this kernel to use the indices you computed in this code block.

nntrainer/layers/cl_layers/reshape_cl.cpp
niket-agarwal: Thanks for the suggestion! I have tested and updated in the latest commit.

test/unittest/layers/unittest_layers_reshape_cl.cpp
niket-agarwal: Updated.

nntrainer/layers/cl_layers/reshape_cl.cpp
niket-agarwal: Updated with @djeong20's suggestion. Thanks.

djeong20 (commented):
Appreciate the hard work! One last suggestion is renaming the reshape_cl kernel as copy_cl. Since the kernel is basically copying data, I think copy_cl better reflects the function of the kernel (the final result would be `ReshapeLayerCl` uses the `copy_cl` kernel).

test/unittest/layers/unittest_layers_reshape_cl.cpp
```diff
@@ -0,0 +1,36 @@
...
+GTEST_PARAMETER_TEST(ReshapeGPU, LayerSemanticsGpu,
+                     ::testing::Values(semantic_reshape_gpu));
+
+auto reshape_basic_plain = LayerGoldenTestParamType(
+  nntrainer::createLayer<nntrainer::ReshapeLayerCl>, {"target_shape=-1"},
+  "2:3:3:3", "reshape.nnlayergolden",
+  LayerGoldenTestParamOptions::SKIP_CALC_DERIV |
+    LayerGoldenTestParamOptions::SKIP_CALC_GRAD |
+    LayerGoldenTestParamOptions::USE_INC_FORWARD,
+  "nchw", "fp32", "fp32");
```
djeong20, line 33: Please add the fp16 test case and negative test case (maybe in a later PR?)

djeong20 (approved):
Great work! LGTM

jijoongmoon (approved):
LGTM

PR #2654, opened 2024-06-28
[ hgemm ] Consider K=1 changes

skykongkong8:
- Current implementation is rooted on general cases, thus optimize only w.r.t. K accumulation.
- However, when it comes to M,1 x 1,N computation, all optimizations like packing, transposing is no use.
- Implementing a explicit kernel function for such case resolved the latency issue.

| dim = (576, 1) x (1, 1024) | fp16 | fp32 |
| --- | --- | --- |
| noTrans | **190834 ns** |  380525 ns |
| transA |  **173896 ns** | 387860 ns |
| transB | **180369 ns** | 382123 ns |
| transAB | **179263 ns** | 379238 ns |

Since this is K=1 case, we do not need to partial-accumulate w.r.t. K-direction with fp32, thereby accelerated approximately 200%

**Self evaluation:**


jijoongmoon (approved):
LGTM

SeoHyungjun (approved):
LGTM

PR #2655, opened 2024-07-01
[ hgemm ] Improve transposed B matrix computation and matrix padding @open seasame 07/15 20:33

skykongkong8:
Major changes:  
1. Implement transposed B matrix data packing-blocking-kernel sequences. Currently only supports for 8x16 kernel.
2. Implement matrix padding. When input data shape is not compatible with fixed-sized blocking and kernel, adding padding can facilitate them. Note that fine-grained blocking/kernel should be implemented for optimal performance(in terms of speed and *memory*), but padding can be one of the *sub-optimal* options for easier implementation.
3. Split hgemm interface, padding, kernel for easier maintenance.

Following is the unittest output conducted on Galaxy S23 with mean latency. (TC=100) 
Additionally, I tested further with experimental GEMM kernel that is quite faster, but less accurate. (not included in this PR, will be introduced in the near future)

| TCs | fp32 | fp16 before | fp16 after | fp16 experimental wip |
| --- | --- | --- | --- | --- |
| conv21 | 802 ms | 1484 ms | **692 ms** | (575 ms) |
| conv22 | 810 ms | 1561 ms | **745 ms** | (572 ms) |
| conv23 | 900 ms | 1652 ms | **756 ms** | (570 ms) |
| conv24_noTrans | 805 ms | 599 ms | **592 ms** | (431 ms) |
| conv24_transB  | 854 ms | 999 ms | **733 ms** | (584 ms) |
| conv768  | 22575 ms | 16448 ms | **15937 ms** | (11163 ms) |
| dot_gemm_fc  | 76083 ms | 131610 ms | **57408 ms** | (39968 ms) |

- Note that the ratio here might differ when M-K-N configuration changes!

| TCs | time consumed during padding | ratio (padding time / total time * 100) |
| --- | --- | --- |
| conv21 |  0.015 ms | 0.0021% |
| conv22 | 0.018 ms | 0.0024% |
| conv23 | 0.015 ms | 0.0019% |

**Self evaluation:**


nntrainer/tensor/blas_interface.cpp
```diff
@@ -326,8 +326,8 @@ static void sgemm_FP16(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA,
                        const unsigned int ldc) {
 
 #if (defined USE__FP16 && USE_NEON)
-  nntrainer::neon::hgemm(A, B, C, M, N, K, alpha, beta, TransA == CblasTrans,
-                         TransB == CblasTrans);
+  nntrainer::neon::custom_hgemm(A, B, C, M, N, K, alpha, beta,
```
jijoongmoon, line 329: Is there any reason for ```custom_```?

jijoongmoon:
It is not this PR, but we can also consider using multi-threading for the blocking loop. 
If it is ready, please remove do not merge tag.

nntrainer/tensor/blas_interface.cpp
skykongkong8: Because there can be many half-precision gemm solutions in the future, and `blas_interface` should be just an interface to choose from them.
On top of that, I just wanted to split `hgemm` function totally from `blas_interface`

skykongkong8:
> It is not this PR, but we can also consider using multi-threading for the blocking loop. If it is ready, please remove do not merge tag.

Not yet for the upstream merge. It would invoke some unittest fails because currently we are forcibly padding all matrices, while there are some cases of padding situations NYI. But still I think we can check on some model applications.

nntrainer/tensor/hgemm/hgemm_padding/hgemm_padding_a.h
```diff
@@ -0,0 +1,136 @@
...
+ *
+ */
+
+/**
+ * @brief Padding function for matrix A in HGEMM
+ *
+ * @param A src matrix to pad
+ * @param Ap dst matrix after padding
+ * @param M row length of matrix A
+ * @param K col length of matrix A
```
EunjuYang, line 20:
```suggestion
 * @param M the number of rows of matrix A
 * @param K the number of cols of matrix A
```
Please consider to revise this comment.
The phrase "row length" may lead some confusions.

skykongkong8: Got it!

EunjuYang:
Would it be possible to combine the `hgemm_padding` functions of matrices A and B by using row-wise / col-wise multiple padding arguments? (instead of using specific parameter for A and B)
I believe this would allow us to reuse the code for various padding scenarios. 

For instance, we could use a generalized version of 'hgemm_padding_A_noTrans_wrt_MK' to handle 'hgemm_padding_B_noTrans_wrt_KN' as well. The only differences between these two functions are in the naming of rows and columns as well as the specific padding sizes used. 
By combining them, we can achieve flexibility when dealing with different padding sizes. 
Currently, the implementation only supports row 8 / col 8 padding for Matrix A and row 8 / col 16 padding for Matrix B.

skykongkong8:
> Would it be possible to combine the `hgemm_padding` functions of matrices A and B by using row-wise / col-wise multiple padding arguments? (instead of using specific parameter for A and B) I believe this would allow us to reuse the code for various padding scenarios.
> 
> For instance, we could use a generalized version of 'hgemm_padding_A_noTrans_wrt_MK' to handle 'hgemm_padding_B_noTrans_wrt_KN' as well. The only differences between these two functions are in the naming of rows and columns as well as the specific padding sizes used. By combining them, we can achieve flexibility when dealing with different padding sizes. Currently, the implementation only supports row 8 / col 8 padding for Matrix A and row 8 / col 16 padding for Matrix B.

That is correct, and I am aware of that point. The reason why I implemented like this is just to code faster. Technically, all functions related to matrix padding will be deleted in the future. Adding padding is actually a nonsense in the first place with that pov

EunjuYang (approved):

nntrainer/tensor/blas_interface.cpp
jijoongmoon: Then, I think hgemm is more proper function name for the users and we can manage internal functions name according to the various hgemm version.

skykongkong8: Sure. will be refactored properly in the later patch

jijoongmoon (approved):
LGTM

skykongkong8:
Interesting information to share:
I found that even ARM Compute Library uses zero-padding in matrix packing!
https://developer.arm.com/documentation/109246/0100/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication/preprocess-r-function-overview

PR #2657, opened 2024-07-01
[FP16][Tensor] Remove unnecessary copy on save

DonghakPark:
There are unnecessary copy of tensor, in case of fp16

It seems that when developing previously, the tensor structure was not accurately established, so it attempted to save by forcibly converting to FP16.

Now, when performing getData<_FP16>(), it is automatically converted, so the process of putting every tensor one by one in the temp array is unnecessary and only slows down the speed.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -3155,12 +3155,7 @@ void Tensor::save(std::ostream &file) {
                  "[Tensor::save] operation failed");
   } else if (this->getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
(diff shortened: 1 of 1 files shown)


skykongkong8 (approved):
Don't really understand why we coded like this previously

lhs8928 (approved):
Nice catch 💯

SeoHyungjun (approved):
Good correction.
I think copying the previous code can have a slight impact on both memory and time.

EunjuYang (approved):
LGTM!👍

djeong20 (approved):
👍👍

myungjoo (approved):

PR #2662, opened 2024-07-02
BUG FIX : Swiglu fp16 GPU Layer test filename mismatch

niket-agarwal:
Modified the swiglufp16 filename in gen_layer_tests for unity with the name in unittest_layers_swiglu_cl.


test/input_gen/gen_layer_tests.py
```diff
@@ -909,6 +909,6 @@ def swiglu(inputs):
     record_single_fp16(
         swiglu_layer,
         [(2, 3, 3, 3), (2, 3, 3, 3)],
-        "swiglu",
+        "swiglufp16",
```
jijoongmoon, line 912: I wonder why the input is float data type?

skykongkong8: As far as I recall, we have been formulating fp16 layer test data like this. This is because I found that using float16 as input makes tf to compute literally everything in fp16, resulting in too much precision loss or even worse, NaN values.

djeong20 (approved):
LGTM

skykongkong8 (approved):

test/input_gen/gen_layer_tests.py
jijoongmoon: Thanks.

jijoongmoon (approved):
LGTM

PR #2664, opened 2024-07-03
[Tensor] Remove NaN check for integer

djeong20:
Fixed-sized integer formats do not have a way of explicitly indicating invalid data.
Every possible value of an int is a number. Therefore, removing NaN check for int values.

**Self-evaluation:**


myungjoo (approved):

skykongkong8 (approved):
Good point!

PR #2666, opened 2024-07-03
[GPU/OpenCL] Moving Addition kernel to Tensor Directory @open sesame 07/04 09:01

yashSingh0723:
Moved `addition_cl` kernel to Tensor directory and refactored it for generalization.
Naming convention changed for `fp16` in `addition_layer` unittests.

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernel_interface.h
+++ b/nntrainer/tensor/cl_operations/blas_kernel_interface.h
@@ -63,5 +63,13 @@ void dotBatchedCl(Tensor const &input, Tensor const &m, Tensor &result,
  */
 void multiplyCl(Tensor &input, float const &value, RunLayerContext &context);
 
+/**
+ * @brief Process data and dimensions for add operation
+ * @param[in] input Tensor
+ * @param[in] result Tensor
+ * @param[in] RunLayerContext reference
+ */
+void add_i_cl(Tensor const &input, Tensor &result, RunLayerContext &context);
+
 } // namespace nntrainer
 #endif /* __BLAS_KERNEL_INTERFACE_H__ */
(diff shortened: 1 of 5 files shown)


s-debadri (approved):
LGTM

nntrainer/tensor/cl_operations/blas_kernel_interface.h
```diff
@@ -63,5 +63,13 @@ void dotBatchedCl(Tensor const &input, Tensor const &m, Tensor &result,
  */
 void multiplyCl(Tensor &input, float const &value, RunLayerContext &context);
 
+/**
+ * @brief Process data and dimensions for add operation
+ * @param[in] input Tensor
+ * @param[in] result Tensor
+ * @param[in] RunLayerContext reference
+ */
+void add_i_cl(Tensor const &input, Tensor &result, RunLayerContext &context);
```
djeong20, line 72: this is not a big deal but how about renaming it as `add_cl()` for clarity since Tensor uses `_i` as in-place operations. (which in this case takes an output tensor)

EunjuYang: +) What about naming it as `addCl` to make it consistent with other kernel operations?

yashSingh0723: Actually I was following the cpu naming convention, for cpu implementation it is named as `add_i`. And for the GPU kernel as well I am doing the in-place operation. So inorder to add `input[0]` and `input[1]`, I am storing the result as `input[0] += input[1]`, not creating any other output tensor additionally and treating `input[0]` as output tensor. Followed the CPU implementation only.

But still if I need to change the name to `addCL`, Please tell and I'll do it accordingly.

jijoongmoon (approved):
LGTM

EunjuYang (approved):
LGTM

PR #2674, opened 2024-07-11
Remove dangerous dummy meson dep

### LINKED ISSUE #2673: Abusive usage of meson dependency system.
https://github.com/nnstreamer/nntrainer/blob/64bd12e479bc038a1bdf875e0d4bd9e735210858/meson.build#L204

This is dangerous usage, especially when ```a_dep``` is applied as a dependency while it's defined  by ```a_dep = found_dummy_dep```. Instead, we should declare dependencies based on the gathered information (e.g., include path, lib path, ...) when it's gathered by scripts.

I'll remove such usages.

myungjoo:
When a dependency library is installed with hardcoded scripts, declare dependency with as much information as possible from the installed package to detect dependency errors at build-time.

Don't add a dummy dependency for actual library dependencies.

Fixes #2673

### DIFF
--- a/api/capi/meson.build
+++ b/api/capi/meson.build
@@ -43,7 +43,7 @@ if get_option('platform') == 'tizen'
 endif
 
 if get_option('platform') == 'android'
-  nntrainer_capi_dep = found_dummy_dep
+  nntrainer_capi_dep = declare_dependency(include_directories: capi_inc)
 else
   shared_library('capi-nntrainer',
     capi_src,
--- a/api/ccapi/meson.build
+++ b/api/ccapi/meson.build
@@ -24,7 +24,7 @@ ccapi_deps = [
 ]
 
 if get_option('platform') == 'android'
-  nntrainer_ccapi_dep = found_dummy_dep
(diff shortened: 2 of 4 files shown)


SeoHyungjun (approved):
It looks good because it shows the name of the real dependency, not a dummy.

DonghakPark (approved):
Shouldn't this part be corrected as well?

```openmp_dep = dummy_dep```
at line 261 (meson.build)

lhs8928 (approved):
LGTM

myungjoo:
> **quick question)** Shouldn't dummy_dep be corrected as well like below ?
> 
> ```
> openmp_dep = dummy_dep
> opencv_dep = dummy_dep
> ```

Those might use https://mesonbuild.com/Reference-manual_returned_disabler.html for "initialization purpose".

skykongkong8 (approved):
LGTM

djeong20 (approved):
LGTM! Would this be for debugging purposes or to avoid any build errors?

myungjoo:
> LGTM! Would this be for debugging purposes or to avoid any build errors?

This allows earlier errors: i.e., meson errors rather than compiler or runtime errors.

PR #2683, opened 2024-07-26
fix: incorrect C/C++ preprocessor macro

myungjoo:
When -DENABLE_ENCODER is given, you do
 #ifdef ENABLE_ENCODER
not
 #ifdef DENABLE_ENCODER

CC: @baek2sm


djeong20 (approved):

jijoongmoon (approved):
LGTM

PR #2686, opened 2024-07-26
[ Layers ] Bug Fix for NHWC Support

jijoongmoon:
This PR fixes the bug about the inplace layers when the channel last is enabled.
Previously, the input var_grad tensor's format is not changed in inplace layers.

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -779,8 +779,10 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
           TensorSpecV2::RequestType::READ_ONLY_VIEW;
         if (lnode->getType() == IdentityLayer::type) {
           s.variable_spec.reference_name = inputs[i]->getName();
+          s.variable_spec.dim.setFormat(inputs[i]->getDim().getFormat());
         } else {
           s.variable_spec.reference_name = inputs[0]->getName();
+          s.variable_spec.dim.setFormat(inputs[0]->getDim().getFormat());
         }
       }
       if (shared_grad && s.gradient_spec) {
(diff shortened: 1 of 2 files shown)


DonghakPark (approved):
LGTM!

nntrainer/graph/network_graph.cpp
```diff
@@ -779,17 +779,21 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
           TensorSpecV2::RequestType::READ_ONLY_VIEW;
         if (lnode->getType() == IdentityLayer::type) {
           s.variable_spec.reference_name = inputs[i]->getName();
+          s.variable_spec.dim.setFormat(inputs[i]->getDim().getFormat());
```
djeong20, line 782: how about using `getFormat()` directly instead of calling `getDim()`?
```suggestion
          s.variable_spec.dim.setFormat(inputs[i]->getFormat());
```
using `Tensor::getFormat()` could also have a memory advantage since `getDim()` returns a copy of the TensorDim.

djeong20 (approved):
overall, LGTM

skykongkong8 (approved):
I will try testing on top of this PR. Good

nntrainer/layers/flatten_layer.cpp
```diff
@@ -33,21 +33,34 @@ void FlattenLayer::finalize(InitLayerContext &context) {
...
+    unsigned int flattened_size = 1;
+    for (unsigned int i = start_dimension; i <= end_dimension; ++i) {
+      flattened_size *= in_dim[i];
+      target_dim[i] = 1;
+    }
+    target_dim[end_dimension] = flattened_size;
+
+    target_shape = "target_shape=" + std::to_string(target_dim[1]) + ":" +
+                   std::to_string(target_dim[2]) + ":" +
+                   std::to_string(target_dim[3]);
```
SeoHyungjun, line 62: How about adding a getter like getShape() to tensor_dim?

```suggestion
   target_shape = "target_shape=" + target_dim.getShape()
```

In tensor_dim, MAXDIM is set to 4.

It would be beneficial if the getShape function could create and return a string that matches the dimensions.

SeoHyungjun (approved):

EunjuYang (approved):
LGTM

PR #2687, opened 2024-07-30
[Tensor] Update tensorbase for efficient creation of new tensor class.

djeong20:
This PR updates the TensorBase class to make mathematical operations that are not required to create a new tensor class.
This change allows developers to easily create new classes without implementing math operations.
Note that these functions should be implemented to utilize tensor operations fully.

- Change math operation function from pure virtual function to virtual function
- Add a private function to get the data type as a string

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor_base.cpp
+++ b/nntrainer/tensor/tensor_base.cpp
@@ -342,4 +342,236 @@ void TensorBase::calculateFlattenDot(
   ldc = (getFormat() == Tformat::NHWC) ? output.channel() : output.width();
 }
 
+/**
+ * Please note that the following functions need to be implemented in a child
+ * class to utilize tensor operations fully — operations such as addition,
+ * division, multiplication, dot production, data averaging, and so on.
+ */
+void TensorBase::setRandNormal(float mean, float stddev) {
+  throw std::invalid_argument(
+    "Tensor::setRandNormal() is currently not supported in tensor data type " +
+    getStringDataType());
+}
+
+void TensorBase::setRandUniform(float min, float max) {
+  throw std::invalid_argument(
+    "Tensor::setRandUniform() is currently not supported in tensor data type " +
+    getStringDataType());
+}
+
+void TensorBase::setRandBernoulli(float probability) {
+  throw std::invalid_argument("Tensor::setRandBernoulli() is currently not "
+                              "supported in tensor data type " +
+                              getStringDataType());
+}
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/tensor_base.h
```diff
@@ -734,6 +721,32 @@ class TensorBase {
                            unsigned int &input_last_axis, unsigned int &M,
                            unsigned int &N, unsigned int &K, unsigned int &lda,
                            unsigned int &ldb, unsigned int &ldc) const;
+
+  /**
+   * @brief  Get the Data Type String object
+   * @return std::string of tensor data type
+   */
+  std::string getStringDataType() const {
```
lhs8928, line 729: Instead of switch statement why don't we use function overriding as like other mathematical operations?

lhs8928: Suggested code would be as follows.
```
class TensorBase {
  virtual std::string getStringDataType() const {
    return "Undefined type";
  }
}

class FloatTensor : public TensorBase {
  std::string getStringDataType() const override {
    return "FP32";
  }
}

class HalfTensor : public TensorBase {
  std::string getStringDataType() const override {
    return "FP16";
  }
}

...

```

djeong20: Good point! This could definitely make a better code structure.

DonghakPark (approved):
From the perspective of clean code and maintainability, it looks much better.

lhs8928 (approved):
LGTM

skykongkong8 (approved):

nntrainer/tensor/tensor_base.cpp
```diff
@@ -342,4 +342,236 @@ void TensorBase::calculateFlattenDot(
...
+  return ML_ERROR_NOT_SUPPORTED;
+}
+
+Tensor &TensorBase::multiply(float const &value, Tensor &output) const {
+  throw std::invalid_argument(
+    "Tensor::multiply() is currently not supported in tensor data type " +
+    getStringDataType());
+  return output;
+}
+
```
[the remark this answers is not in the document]
djeong20: not really. If no overridden function exists in the derived class, it will throw an error.
On top of that, you must implement any mathematical tensor operations in the derived class (e.g., FloatTensor).

jijoongmoon (approved):
LGTM

PR #2695, opened 2024-08-02
[GPU/OPENCL] RMSNorm Accuracy Fix

pallaviNNT:
Accuracy issue observed while running SRLLM on GPU due to wrong gamma values being used.
Fixed the same


jijoongmoon:
Could check the static checker  CI issue? 
And also have to check wether we do need the header inclusion of RMSNormCL in layer.h https://github.com/nnstreamer/nntrainer/blob/32d901cee3c913a27e95ce81923266f6de7a063b/api/ccapi/include/layer.h#L314

djeong20 (approved):

EunjuYang (approved):
Nice catch!
What about updating the comment for the alpha as well? (one for each ~~channel~~)

skykongkong8 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2698, opened 2024-08-06
[ matrix_transpose/bugfix ] Prevent reading/saving data from/to unallocated memory

skykongkong8:
- Previous transpose kernel occasionally load/save unallocated memory, and then masked it.
- Now, it does not read them at the first place, but load with for-loop
- This would deteriorate speed of fp16 matrix transpose, but won't be dominant in total model latency

| dim | before | after |
| --- | --- | --- |
|  87,2049 | 884 ns | 16114 ns  |
| 2048, 86 | 34019 ns | 82258 ns  |

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/matrix_transpose_neon/matrix_transpose_kernels_neon.h
+++ b/nntrainer/tensor/matrix_transpose_neon/matrix_transpose_kernels_neon.h
@@ -73,14 +73,20 @@ static void transpose_kernel_mxn_neon_128(unsigned int N, const __fp16 *src,
                                           unsigned int ld_src, __fp16 *dst,
                                           unsigned int ld_dst) {
 
-  uint16x4_t bitmask_v8 =
-    vld1_u16(reinterpret_cast<const uint16_t *>(masks[N]));
   float16x4_t input[4];
   float16x4_t ZEROS = vmov_n_f16(0.F);
 
   unsigned i;
   for (i = 0; i < M; ++i) {
-    input[i] = vbsl_f16(bitmask_v8, vld1_f16(&src[i * ld_src]), ZEROS);
+    if (N == 4) {
+      input[i] = vld1_f16(&src[i * ld_src]);
+    } else {
+      float16x4_t tmp = ZEROS;
+      for (unsigned int n = 0; n < N; ++n) {
+        tmp[n] = src[i * ld_src + n];
+      }
+      input[i] = tmp;
+    }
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/matrix_transpose_neon/matrix_transpose_kernels_neon.h
```diff
@@ -73,14 +73,20 @@ static void transpose_kernel_mxn_neon_128(unsigned int N, const __fp16 *src,
...
   float16x4_t input[4];
   float16x4_t ZEROS = vmov_n_f16(0.F);
 
   unsigned i;
   for (i = 0; i < M; ++i) {
-    input[i] = vbsl_f16(bitmask_v8, vld1_f16(&src[i * ld_src]), ZEROS);
+    if (N == 4) {
+      input[i] = vld1_f16(&src[i * ld_src]);
+    } else {
+      float16x4_t tmp = ZEROS;
```
lhs8928, line 84: AFAIK attempting to read the unallocated memory only happens if the i == M - 1.
Please let me know if I'm wrong.

skykongkong8: That is correct, but `M` inside of this kernel does not mean global M. 
M here ranges to 1~8 (local row size).
And as you can see below, it will affect for the leftover local M only, since otherwise it will fall into fixed size kernels, or leftover kernels with template param `M = 4` or `M = 8 `
```C++
...
//   if (N % 8 > 0 && N % 8 < 4) {
        transpose_kernel_mxn_neon_128<4>(N - jb, &src[i * ld_src + jb], ld_src,
                                         &dst[i + jb * ld_dst], ld_dst);
...
//   } else {
      if (jb < N) {
        transpose_kernel_mxn_neon_256<8>(N - jb, &src[ib * ld_src + jb], ld_src,
                                         &dst[ib + jb * ld_dst], ld_dst);
      }
...
```

skykongkong8: And fyi, I think this won't harm the total GEMM latency, because 82258 ns is 0.082258 ms, and normal SGEMM computation using similar dimension would be range from 3 ~ 4 ms, which is quite trivial

jijoongmoon (approved):
LGTM

PR #2699, opened 2024-08-07
[ BUILD ] Add 16K shared lib package option for Android

jijoongmoon:
Android encourage to use 16KB package for the shared library. This PR add the 16KB package option and also recommand to use ndk which is higher or equal version of r27.

Resolves:

**Self evaluation:**

### DIFF
--- a/jni/Android.mk.in
+++ b/jni/Android.mk.in
@@ -36,6 +36,7 @@ LOCAL_MODULE := ml-api-inference
 
 LOCAL_SRC_FILES := @MESON_ML_API_COMMON_ROOT@/lib/arm64-v8a/libnnstreamer-native.so
 LOCAL_EXPORT_C_INCLUDES := @MESON_ML_API_COMMON_ROOT@/include
+LOCAL_LDFLAGS 	    += "-Wl,-z,max-page-size=16384"
 
 include $(PREBUILT_SHARED_LIBRARY)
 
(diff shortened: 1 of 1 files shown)


DonghakPark (approved):
LGTM

djeong20 (changes requested):
we should also update `Application.mk` to support 16 KB alignment with Android NDK version r27 and higher. [(link)](https://developer.android.com/guide/practices/page-sizes#compile-r26-lower)

DonghakPark:
> we should also update `Application.mk` to support 16 KB alignment with Android NDK version r27 and higher. [(link)](https://developer.android.com/guide/practices/page-sizes#compile-r26-lower)

added Flags for all Android.mk file. Please Take a look #2705

djeong20:
> added Flags for all Android.mk file. Please Take a look #2705

updating `Android.mk` is to compile with NDK version r26 or lower. The problem is whether we support NDK version r26 or lower or restrict to use of version r27 or higher.

> Regardless, it is recommended to move to Android NDK r27 or higher to update to the 16 KB ELF-aligned version of the C++ standard library. Otherwise, your app will fail to install on 16 KB devices today.

if we restrict it to use Android NDK version r27 or higher, we do not need to change `Android.mk` and only update `Application.mk`.

DonghakPark:
> > added Flags for all Android.mk file. Please Take a look #2705
> 
> updating `Android.mk` is to compile with NDK version r26 or lower. The problem is whether we support NDK version r26 or lower or restrict to use of version r27 or higher.
> 
> > Regardless, it is recommended to move to Android NDK r27 or higher to update to the 16 KB ELF-aligned version of the C++ standard library. Otherwise, your app will fail to install on 16 KB devices today.
> 
> if we restrict it to use Android NDK version r27 or higher, we do not need to change `Android.mk` and only update `Application.mk`.

Oh. i see !!
Would it not be necessary to set up in Android.mk as well for when the ndk version is below 27?

djeong20:
> Would it not be necessary to set up in Android.mk as well for when the ndk version is below 27?

you're right! If we plan to support the ndk version below 27, we should specify these flags. though, I'm not sure what would be the right choice. maybe we can allow ndk version 26 <= for now, then check the ndk version and print a proper message to upgrade to r27.

PR #2700, opened 2024-08-08
[ hgemm ] Redesign micro kernel to get amount of K direction accumulation as an argument

skykongkong8:
- Previous implementation naively used fixed-sized ukernels for the K-direction accumulation.
- Such kernels were excessively long, but had better performance than looping through single K-iteration.
- However, recent test results have shown that just stacking 4 K iters, and looping through such ukernel preserved the performance with better code readability.

**Self evaluation:**


djeong20 (approved):

jijoongmoon (approved):
LGTM

PR #2703, opened 2024-08-12
[Tensor] Add optional output tensor for tensor concatenation

djeong20:
This PR adds an optional feature in Tensor::cat to pass the output tensor to the function.
This change allows the user-given tensor to store the result of the concatenation without creating a new tensor.

- Add optional argument output (the output tensor) to the cat function.
- Add negative test cases for tensor concatenation.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -17,7 +17,6 @@
 #ifdef ENABLE_FP16
 #include <half_tensor.h>
 #endif
-
 namespace nntrainer {
 
 Tensor::Tensor(std::string name_, Tformat fm, Tdatatype d_type) {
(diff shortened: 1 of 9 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -1088,6 +1105,46 @@ void Tensor::setTensorVar(TensorDim d, void *buf, size_t offset) {
   itensor->setTensorVar(d, buf, offset);
 }
 
+TensorDim Tensor::calculateConcatOutputDim(const std::vector<Tensor> &tensors,
+                                           int axis) {
```
djeong20, line 1109: This newly added private static function is to calculate the output tensor dimension with tensor vectors and axis.
It is used for checking the given output tensor dimension and creating a new output tensor if not provided.

nntrainer/tensor/tensor.h
```diff
@@ -1202,9 +1202,12 @@ class Tensor {
    *
    * @param tensors tensors to be concatenated to the first tensor
    * @param axis axis
+   * @param output output tensor to store the result
    * @return Tensor concatenated tensor
+   *
+   * @note  This function should not be used directly. Please use cat() instead.
```
djeong20, line 1208: I need advice on structure and design here.

`Tensor::concat()` is designed for internal use and should not be used directly like this:
```
Tensor t;
t.concat(tensors, axis);
```

This is because the tensor t calling the concat has no relationship with the input and output tensors. Additionally, directly using concat cannot properly handle errors. So, I'm looking for a way to prevent the use of `concat`.

Currently, concat is used as follows.
1. The public static function `Tensor::cat()` utilizes the tensor member function `concat`.
2. `Tensor::concat()` makes use of a tensor private variable `itensor` to call Float/HalfTensor `concat`.

The issue here is that if concat is made private, `Tensor::cat()` wouldn't be able to utilize concat since it's a static function. Is there a way to restrict the usage of concat to only `Tensor::cat()`?

DonghakPark (approved):
I think it's a good direction for improvement.

Q) However, is it changing the entire signature?

Rather than immediately changing the function's signature, it may be better to add a new parm by overloading it and leaving the existing function as it is.

Overloading it to maintain compatibility, the old function is left out, and the internal functionality is changed into a new form.

djeong20:
> I think it's a good direction for improvement.
> 
> Q) However, is it changing the entire signature?
> 
> Rather than immediately changing the function's signature, it may be better to add a new parm by overloading it and leaving the existing function as it is.
> 
> Overloading it to maintain compatibility, the old function is left out, and the internal functionality is changed into a new form.

Thank you for pointing it out! To explain, the function `Tensor::cat()` is overloaded and does not change the function signature. However, the `concat()` function is changed but will not require any exterior changes since it is only used internally.

skykongkong8 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2704, opened 2024-08-12
[ Tensor ] Remove CBLAS params from Tensor related files.

skykongkong8:
- Remove cblas params from tensor related files since nntrainer is not fully-dependent on cblas anymore.
- Letting tensors to be aware of Cblas related parameters is a nonsense at the first place.
- CBLAS params will be declared only when functions from cblas is called.
- fyi) `TStorageOrder` : Tensor Storage Order

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -24,6 +24,12 @@
 #include <blas_avx.h>
 #endif
 
+#ifdef USE_BLAS
+extern "C" {
+#include <cblas.h>
+}
+#endif
+
 #include <cmath>
 
 #define sgemv_loop(ci, cj, cM, cN)           \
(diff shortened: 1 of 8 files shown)


skykongkong8:
This pr is related to known issue rasied from #2704

skykongkong8:
@s-debadri 
Please do not let cblas params to reside outside of actual cblas function interface.
I could observe some from `opencl` related works.

nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
```diff
@@ -170,20 +165,19 @@ void dotCl(Tensor const &input, Tensor const &m, Tensor &result,
     }
     /// case2: (M * K) X (K * 1)
     else if (N == 1) {
-      transA ? sgemv_cl(data, mdata, rdata, dim2, dim1, lda, context)
-             : sgemv_cl(data, mdata, rdata, dim1, dim2, lda, context);
+      trans ? sgemv_cl(data, mdata, rdata, dim2, dim1, lda, context)
```
SeoHyungjun, line 168: It seems that transA has been changed to trans and transB has been changed to trans_m.
Is there a reason why you didn't use trans_n or another name?

And in other functions, we use transA and transB.

SeoHyungjun (approved):

nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
skykongkong8: `trans_m` and `trans` are variables from the function parameter. I believe using them is more coherent.

nntrainer/tensor/float_tensor.cpp
```diff
@@ -493,8 +493,8 @@ void FloatTensor::sum_by_batch(Tensor &output) const {
 
   Tensor ones(1, 1, 1, feat_len, this->getFormat());
   ones.setValue(1.0);
-  sgemv(CblasRowMajor, CblasNoTrans, batch, feat_len, 1, data, feat_len,
-        ones.getData<float>(), 1, 0.0, out_data, 1);
+  sgemv((unsigned int)dim.getStorageOrder(), false, batch, feat_len, 1, data,
```
djeong20, line 496: This is just a suggestion! how about having a fixed value for storage orders like transpose?
although there's no difference in the result, I think it would be easier for us to understand the code and debug.

djeong20 (approved):
Great work! please take a look at the comments :)

nntrainer/tensor/blas_interface.cpp
```diff
@@ -93,8 +99,7 @@ static inline void transpose_fallback(unsigned int M, unsigned int N,
 static void saxpy_FP16(const unsigned int N, const float alpha, const _FP16 *X,
                        const int incX, _FP16 *Y, const int incY) {
   if (incX < 0 or incY < 0)
-    throw std::invalid_argument(
-      "Error: negative inc not supported without cblas");
+    throw std::invalid_argument("Error: negative inc not supported");
```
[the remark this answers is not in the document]
skykongkong8: `incX` and `incY` are indices, thus should be always positive. I think this would answer both questions!

nntrainer/tensor/blas_interface.cpp
```diff
@@ -755,10 +761,8 @@ void sgemm(CBLAS_ORDER order, CBLAS_TRANSPOSE TransA, CBLAS_TRANSPOSE TransB,
     cublasHandle_t handle;
     cublasCreate(&handle);
 
-    cublasOperation_t transA =
-      (TransA == CblasTrans) ? CUBLAS_OP_T : CUBLAS_OP_N;
-    cublasOperation_t transB =
-      (TransB == CblasTrans) ? CUBLAS_OP_T : CUBLAS_OP_N;
+    cublasOperation_t transA = (TransA) ? CUBLAS_OP_T : CUBLAS_OP_N;
+    cublasOperation_t transB = (TransB) ? CUBLAS_OP_T : CUBLAS_OP_N;
     cublasSgemm(handle, transA, transB, N, M, K, &alpha, d_B, N, d_A, K, &beta,
```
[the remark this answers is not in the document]
skykongkong8: Never knew it 😮 Thanks for pointing this out!

nntrainer/tensor/float_tensor.cpp
skykongkong8: I don't really get it... Could you elaborate a little bit more for me?
I think current implementation is quite similar to transpose cases.
With my understanding of your suggestion, do you mean we should have functions like:
```
sgemv_rowMaj(...)
...
sgemv_colMah(...)
....
```
?

nntrainer/tensor/blas_interface.cpp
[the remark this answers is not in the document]
skykongkong8: Indeed, it does take matrices as Col-Maj storage order !
https://stackoverflow.com/questions/56043539/cublassgemm-row-major-multiplication

nntrainer/tensor/float_tensor.cpp
djeong20: what I meant by having a fixed value is as follows.
```suggestion
  sgemv(TStorageOrder::ROW_MAJOR, false, batch, feat_len, 1, data,
```
same as we pass the transpose with true/false!

baek2sm (approved):

PR #2702, opened 2024-08-08
[Layer] Improve forwarding logic of ConcatLayer @open sesame 08/08 18:22

djeong20:
## Description

This pull request aims to optimize the performance of `forwarding()` in `ConcatLayer`.
This is accomplished by improving the current concatenation logic by utilizing the tensor operation `concat()` instead of `Map()`.
Note that this optimization is only for `forwarding()` and `calcDeriv()` would be improved in later PR.

#### Before
|key|avg|min|max|sum|
|------|---|---|---|---|
|calcDeriv(concat)|1.004153s|0.997947s|1.010360s|2.008307s|
|forward(concat)|1.083193s|1.042626s|1.010360s|2.166386s|
|concat|1.129055s|0.997965s|1.252510s|9.032440s|

#### After
|key|avg|min|max|sum|
|------|---|---|---|---|
|calcDeriv(concat)|1.134034s|1.131812s|1.136257s|2.268069s|
|forward(concat)|0.017167s|0.012304s|0.022031s|0.034335s|
|concat|0.627868s|0.012310s|1.346551s|5.022949s|

## Note

input 0: 1:1:196:256 [ FP16 : NCHW ]
input 1: 1:1:196:1024 [ FP16 : NCHW ]
output: 1:1:196:1280 [ FP16 : NCHW ]

PR #2706, opened 2024-08-14
[Layer] enhance ConcatLayer algorithms for efficient concatenation and split

djeong20:
## Description

This PR renovates reshape/concatenation algorithms to facilitate efficient concatenation and split in ConcatLayer.

Previously, dimension 2 (height) was set as a standard axis to operate concatenation.
However, this causes an overhead by copying a tensor size of 1 when the concat dimension is 3 (width).

The new algorithm consolidates all dimensions to the first and last axes based on the concat dimension, sets the standard axis to be 3, and performs concat and split.

- Revise creating helper dimension logic in `finalize()`.
- Update `forwarding()` and `calcDeriv()` workflow to be efficient.
- Add descriptions for the new concat algorithm.

#### Previous
|key|avg|min|max|sum|
|------|---|---|---|---|
|calcDeriv(concat)|1.004153s|0.997947s|1.010360s|2.008307s|
|forward(concat)|1.083193s|1.042626s|1.010360s|2.166386s|
|concat|1.129055s|0.997965s|1.252510s|9.032440s|

#### #2702 
|key|avg|min|max|sum|
|------|---|---|---|---|
|calcDeriv(concat)|1.134034s|1.131812s|1.136257s|2.268069s|
|forward(concat)|0.017167s|0.012304s|0.022031s|0.034335s|
|concat|0.627868s|0.012310s|1.346551s|5.022949s|

#### Current
|key|avg|min|max|sum|
|------|---|---|---|---|
|calcDeriv(concat)|0.002177s|0.002166s|0.002188s|0.004354s|
|forward(concat)|0.002414s|0.002321s|0.002508s|0.004829s|
|concat|0.002343s|0.002174s|0.002514s|0.018746s|

## Note

This is a profiling result with the following tensors.

- input 0: 1:1:196:256 [ FP16 : NCHW ]
- input 1: 1:1:196:1024 [ FP16 : NCHW ]
- output: 1:1:196:1280 [ FP16 : NCHW ]

### DIFF
--- a/nntrainer/layers/concat_layer.cpp
+++ b/nntrainer/layers/concat_layer.cpp
@@ -6,6 +6,7 @@
  * @date   27 Oct 2020
  * @see    https://github.com/nnstreamer/nntrainer
  * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author Donghyeon Jeong <dhyeon.jeong@samsung.com>
  * @bug    No known bugs except for NYI items
  * @brief  This is Concat Layer Class for Neural Network
  *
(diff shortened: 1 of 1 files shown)


nntrainer/layers/concat_layer.cpp
```diff
@@ -70,85 +71,128 @@ void ConcatLayer::finalize(InitLayerContext &context) {
...
+   * @example the following shows how the helper dimension will look with given
+   * inputs and concat dimension.
+   *
+   *          | cat_dim 1 | cat_dim 2 | cat_dim 3
+   *  --------|-----------|-----------|-----------
+   *  input0  |  2:1:2:3  |  1:2:1:3  |  1:2:2:3
+   *  input1  |  2:3:2:3  |  1:2:3:3  |  1:2:2:1
+   *  --------|-----------|-----------|-----------
+   *  helper0 |  2:1:1:6  |  2:1:1:3  |  4:1:1:3
+   *  helper1 |  2:1:1:18 |  2:1:1:18 |  4:1:1:1
```
lhs8928, line 93: For cat_dim 2 dimension of helper 1 would be 2:1:1:9

lhs8928 (approved):
LGTM

nntrainer/layers/concat_layer.cpp
djeong20: fixed!

skykongkong8 (approved):
Great work! It must have been really confusing...
one thing I want to point out is that how about adding a new Tensor function to get rid of #ifdefs and leave some rooms for future support for new datatypes?

DonghakPark (approved):
LGTM!

djeong20:
> Great work! It must have been really confusing... one thing I want to point out is that how about adding a new Tensor function to get rid of #ifdefs and leave some rooms for future support for new datatypes?

that's an excellent point! I'm actually working on `Tensor::cat()` to adapt this algorithm and replace it with the updated function 😂. Check it out for the next PR!

nntrainer/layers/concat_layer.cpp
```diff
@@ -70,85 +71,128 @@ void ConcatLayer::finalize(InitLayerContext &context) {
...
   /**
+   * Forwarding in ConcatLayer works as follows
+   *
+   *    in1        in2       in3                  output
+   * |---0---| |----3----| |--6--|      |---0---||----3----||--6--|
+   * |---1---| |----4----| |--7--|  =>  |---1---||----4----||--7--|
+   * |---2---| |----5----| |--8--|      |---2---||----5----||--8--|
+   *
+   * @note the number of each block indicates the order of copy to output
+   *
```
EunjuYang, line 140: Great work! Your comments are clear and provide valuable insights into how this code operates. To enhance comprehension even further, I recommend including descriptions of what each axis in the diagram signifies. (row = width / column = batch) By doing so, readers will have a clear understanding of the concepts presented.

EunjuYang (approved):
LGTM! Thank you for your detailed and clear comments! 👍

baek2sm (approved):
LGTM

PR #2710, opened 2024-08-16
[Layer] add "add layer"

baek2sm:
- added "add layer"
- added a "model unit test" for add layer. Since 'model-level unit test'
havn't been run for a long time, I diabled some test cases that causing
issues when running model unit test.

Many people gave great feedback, so I've improved the structure accordingly.
- An upper class called "OperationLayer" was added to reduce redundant code.
- Based on the number of input tensors, the behavior of "OperationLayer" has been classified into two types: unary and binary operations.
- Various additional code cleanups have also taken place.

- there is an issue where committing compressed files containing golden data for unit test prevents pushing changes to the remote server. (I've confirmed that all unit tests pass locally using that golden data.)

**Self evaluation:**


SeoHyungjun (approved):
LGTM

nntrainer/layers/add_layer.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    TensorDim input_dim = input0.getDim();
+    TensorDim input_step_dim = input_dim;
+    input_step_dim.batch(1);
+    input_step_dim.height(to - from);
+
+    Tensor input_step = input0.getSharedDataTensor(
+      input_step_dim, b * input_dim.getFeatureLen(), true);
+
+    input0.add(input1, hidden_step);
```
EunjuYang, line 71: It appears that the dimensions for `input0` and `hidden_step` do not match. Did you intend to use `input_step` instead of `input0`? I am uncertain whether `input1` can be applied here for the same reason. If it does not, the code related to `input_step` may appear redundant

baek2sm: It appears there was an error in the computation of 'hidden_step'. Thank you! I'll correct it.

skykongkong8 (approved):
LGTM

EunjuYang (approved):
LGTM!

djeong20 (approved):
LGTM!

nntrainer/layers/add_layer.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+void AddLayer::finalize(InitLayerContext &context) {
+  context.setOutputDimensions({context.getInputDimensions()[0]});
+}
+
+void AddLayer::forwarding(RunLayerContext &context, bool training) {
+  Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+
+  const Tensor &input0 = context.getInput(0);
+  const Tensor &input1 = context.getInput(1);
```
lhs8928, line 34: Does add layer only has 2 inputs?
Is there any possibility of having more than 2 inputs?

lhs8928 (commented):
I can't find the difference with addition layer.
Could you explain why did you make a add layer even though NNTrainer already has a addition layer?

nntrainer/layers/add_layer.cpp
baek2sm: @lhs8928 Good point! The addition layer, to be precise, implements a summation operation. This layer corresponds to four basic operations that take two inputs.

nntrainer/layers/add_layer.cpp
```diff
@@ -0,0 +1,94 @@
...
+    Tensor hidden_step = hidden_.getSharedDataTensor(
+      hidden_step_dim, b * hidden_dim.getFeatureLen(), true);
+
+    const Tensor &input0 = context.getInput(0);
+    const Tensor &input1 = context.getInput(1);
+
+    TensorDim input_dim = input0.getDim();
+    TensorDim input_step_dim = input_dim;
+    input_step_dim.batch(1);
+    input_step_dim.height(to - from);
```
EunjuYang, line 66: Do we need to add this code inside the batch loop?
What about moving the codes to the outside of the loop ?

nntrainer/layers/operation_layer.h
```diff
@@ -0,0 +1,142 @@
...
+#include <layer_context.h>
+#include <layer_devel.h>
+
+namespace nntrainer {
+
+/**
+ * @brief Type of layer depending on the number of inputs.
+ *
+ */
+enum class OperationType { NONE, UNARY, BINARY };
```
jijoongmoon, line 26: It is important to consider whether the operation can be performed in-place to save memory. I guess it might be good to give the user an option to set it when the operation layer is created.

baek2sm: A new pull request will be uploaded concerning inplace operations. This implementation has been updated based on the UnaryOperationLayer and BinaryOperationLayer classes! Thanks

jijoongmoon (approved):
LGTM

Issue #2715, opened 2024-08-23
Main branch error after merging #2704


djeong20:
The NNTrainer main branch has an error after merging #2704.
The build error was resolved by #2714, but issues with failing unit tests remain. The details are as follows.

```
The output from the failed tests:

23/41 unittest_nntrainer_tensor_fp16          FAIL     0.37 s (exit status 1)
...
[  PASSED  ] 125 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] nntrainer_Tensor.dot_06_p
[  FAILED  ] nntrainer_Tensor.dot_transpose_p

 2 FAILED TESTS
-------

27/41 unittest_layers                         FAIL     0.37 s (killed by signal 11 SIGSEGV)
...
[----------] 3 tests from FullyConnected16/LayerGoldenTest
[ RUN      ] FullyConnected16/LayerGoldenTest.run/0
 ** On entry to SGEMM  parameter number 13 had an illegal value
 ** On entry to SGEMM  parameter number 10 had an illegal value
../test/unittest/layers/layers_golden_tests.cpp:236: Failure
Expected: ((cos_sim)) >= ((1 - cos_sim_range)), actual: 0.766846 vs 0.99999
../test/unittest/layers/layers_golden_tests.cpp:241: Failure
Expected: ((mean_squared_error)) <= ((mse_range)), actual: 19.4233 vs 0.015
../test/unittest/layers/layers_golden_tests.cpp:296: Failure
Value of: compare_percentage_tensors(tensor, answer, match_percentage, skip_cos_sim)
  Actual: false
Expected: true
outputs at 0
../test/unittest/layers/layers_golden_tests.cpp:241: Failure
Expected: ((mean_squared_error)) <= ((mse_range)), actual: 772.4 vs 0.05
../test/unittest/layers/layers_golden_tests.cpp:241: Failure
Expected: ((mean_squared_error)) <= ((mse_range)), actual: 1.97871 vs 0.03
-------
```

PTAL @skykongkong8

djeong20:
Additionally, current GitHub Action [meson_test (ubuntu-20.04, -Denable-fp16=true)](https://github.com/nnstreamer/nntrainer/actions/runs/10347289811/job/28637279563#logs) is not functioning correctly. It is built and run without enabling fp16. (fp16 unit tests cannot be found)

PR #2716, opened 2024-08-23
[Wait for #2720][CI] Fix meson ubuntu ci build

DonghakPark:
Fix build bug
- Currently, there is a bug in the matrix used in CI where the first Meson build runs successfully but subsequent builds fail due to the presence of a 'build' folder. I would like to fix this issue.
- Before running the Meson build, ensure that any existing folders named 'build' are deleted.

Resolves:
- https://github.com/nnstreamer/nntrainer/issues/2715

**Self evaluation:**

Co-authored-by: hyeonseok <hs89.lee@samsung.com>
Co-authored-by: Donghyeon Jeong <dhyeon.jeong@samsung.com>

### DIFF
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -25,11 +25,23 @@ jobs:
       run: sudo apt-get update && sudo apt-get install -y gcc g++ pkg-config libopenblas-dev libiniparser-dev libjsoncpp-dev libcurl3-dev tensorflow2-lite-dev nnstreamer-dev libglib2.0-dev libgstreamer1.0-dev libgtest-dev ml-api-common-dev flatbuffers-compiler ml-inference-api-dev libunwind-dev
     - name: install additional packages for features
       run: sudo apt-get install -y python3-dev python3-numpy python3
+#    - name: add repo
+#      if: ${{matrix.os}} == 'ubuntu-22.04'
+#      run: sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+    - name: gcc version change
+      run: |
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+        sudo apt-get install build-essential
+        sudo apt update
+        sudo apt install -y gcc-13
+        sudo apt install -y g++-13
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-13 1000 
+        sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-13 1000
+        sudo update-alternatives --set gcc /usr/bin/gcc-13
+    - name: check gcc version
+      run: gcc --version
     - name: install build systems
       run: sudo apt install meson ninja-build
-    - run: meson setup build/
-      env:
-        CC: gcc
     - run: |
         meson \
           --buildtype=plain \
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -25,9 +25,6 @@ jobs:
       run: sudo apt-get update && sudo apt-get install -y gcc g++ pkg-config libopenblas-dev libiniparser-dev libjsoncpp-dev libcurl3-dev tensorflow2-lite-dev nnstreamer-dev libglib2.0-dev libgstreamer1.0-dev libgtest-dev ml-api-common-dev flatbuffers-compiler ml-inference-api-dev libunwind-dev
     - name: install additional packages for features
       run: sudo apt-get install -y python3-dev python3-numpy python3
-#    - name: add repo
-#      if: ${{matrix.os}} == 'ubuntu-22.04'
-#      run: sudo add-apt-repository ppa:ubuntu-toolchain-r/test
     - name: gcc version change
       run: |
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test
@@ -38,8 +35,6 @@ jobs:
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-13 1000 
         sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-13 1000
         sudo update-alternatives --set gcc /usr/bin/gcc-13
-    - name: check gcc version
-      run: gcc --version
     - name: install build systems
       run: sudo apt install meson ninja-build
     - run: |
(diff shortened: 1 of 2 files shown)


lhs8928 (approved):
LGTM

DonghakPark:
It is normal to fail because the CI has not been working properly so far. Additionally, for the current changes, instead of deleting directories, each name was changed differently, and the meson setup was removed.

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -49,7 +46,7 @@ jobs:
           -Dcapi-ml-inference-actual=capi-ml-inference \
           -Denable-capi=enabled \
           ${{ matrix.meson_options }} \
-          build
-    - run: ninja -C build
+          build-${{matrix.meson_options}}-${{matrix.os}}
```
myungjoo, line 49: After removal of ```meson setup build```, you won't see the build directory. Do you really need this?

(from meson manual, ```meson setup build``` == ```meson build```):
```
The setup command
       Using Meson is simple and follows the common two-phase process of most build systems. First you run Meson to configure your build:

       meson setup [ options ] [ build directory ] [ source directory ]

       Note that the build directory must be different from the source directory. Meson does not support building inside the source directory and attempting to do that leads to an error.

       After a successful configuration step you can build the source by running the actual build command in the build directory. The default backend of Meson is Ninja, which can be invoked like this.

       ninja [ target ]

       You only need to run the Meson command once: when you first configure your build dir. After that you just run the build command. Meson will autodetect changes in your source tree and regenerate all files needed to build  the
       project.

       The setup command is the default operation. If no actual command is specified, Meson will assume you meant to do a setup. That means that you can set up a build directory without the setup command like this:

       meson [ options ] [ build directory ] [ source directory ]

```

djeong20:
Currently, there's an issue with the GCC version. The default GCC version is gcc-9 for 20.04 and gcc-11 for 22.04
To use float16, we must install and set the gcc version higher than gcc-12 (GCC 13 is preferred).

myungjoo:
> Currently, there's an issue with the GCC version. The default GCC version is gcc-9 for 20.04 and gcc-11 for 22.04 To use float16, we must install and set the gcc version higher than gcc-12 (GCC 13 is preferred).


gcc-12 is officially available for Ubuntu 22.04: https://packages.ubuntu.com/jammy-updates/devel/ 
Install gcc-12, change the default setting, and let meson use gcc-12.

for official gcc-13 Ubuntu packages, move to 24.04, or use PPA: ppa:ubuntu-toolchain-r/test

SeoHyungjun (approved):
LGTM

.github/workflows/ubuntu_clean_meson_build.yml
DonghakPark: When running matrix jobs in GitHub Actions, they run concurrently,so I changed to using the format "build-~~" instead of just "build" as the directory name to avoid potential issues with executing subsequent commands if there were directories with the same name.

myungjoo: They are running separated.

When renaming is dropped, you may consider it's approved.

DonghakPark:
For clarity's sake, let me provide some comments.

So far, the reason why Meson build & Ninja tests have not been working as intended is due to the redundant usage of both 'meson setup build' and 'meson build'.

In addition to this, I initially thought that jobs in GitAction would run within the same container but with @myungjoo  help, I realized it was unnecessary to separate the names of build directories, so I made adjustments accordingly.

skykongkong8 (approved):

myungjoo:
> > Currently, there's an issue with the GCC version. The default GCC version is gcc-9 for 20.04 and gcc-11 for 22.04 To use float16, we must install and set the gcc version higher than gcc-12 (GCC 13 is preferred).
> 
> gcc-12 is officially available for Ubuntu 22.04: https://packages.ubuntu.com/jammy-updates/devel/ Install gcc-12, change the default setting, and let meson use gcc-12.
> 
> for official gcc-13 Ubuntu packages, move to 24.04, or use PPA: ppa:ubuntu-toolchain-r/test

Urgg....

https://github.com/actions/runner-images/issues/9866

They have recently removed GCC 13 packages from PPA. You won't be able to install gcc-13 with ppa, too.

myungjoo:
For gcc-13 in Ubuntu Jammy, use ```ppa:ubuntu-toolchain-r/test``` instead of ```ppa:ubuntu-toolchain-r/ppa```

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -25,11 +25,23 @@ jobs:
       run: sudo apt-get update && sudo apt-get install -y gcc g++ pkg-config libopenblas-dev libiniparser-dev libjsoncpp-dev libcurl3-dev tensorflow2-lite-dev nnstreamer-dev libglib2.0-dev libgstreamer1.0-dev libgtest-dev ml-api-common-dev flatbuffers-compiler ml-inference-api-dev libunwind-dev
     - name: install additional packages for features
       run: sudo apt-get install -y python3-dev python3-numpy python3
+#    - name: add repo
+#      if: ${{matrix.os}} == 'ubuntu-22.04'
+#      run: sudo add-apt-repository ppa:ubuntu-toolchain-r/test
```
djeong20, line 30:
```suggestion
```

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -25,11 +25,23 @@ jobs:
...
+    - name: gcc version change
+      run: |
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+        sudo apt-get install build-essential
+        sudo apt update
+        sudo apt install -y gcc-13
+        sudo apt install -y g++-13
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-13 1000 
+        sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-13 1000
+        sudo update-alternatives --set gcc /usr/bin/gcc-13
```
djeong20, line 40: let's set g++ with g++-13 as well.
```suggestion
        sudo update-alternatives --set gcc /usr/bin/gcc-13
        sudo update-alternatives --set g++ /usr/bin/g++-13
```

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -25,11 +25,23 @@ jobs:
...
+    - name: check gcc version
+      run: gcc --version
```
djeong20, line 42: I think this can be removed now!
```suggestion
```

djeong20 (approved):
This patch properly uses gcc-13 & build and run with `enable-fp16=true` 👍  I think it is good to go :) nice work!

.github/workflows/ubuntu_clean_meson_build.yml
DonghakPark: Thank you, FIxed!

PR #2720, opened 2024-08-28
[Tizen7.0] PR to support Tizen7.0

EunjuYang:
This pull request includes the following commits to backport NNTrainer for Tizen 7.0:
- Commit related to fp16 enablement is included: 9328d1f898953d08d5e2f48c3c9f500b1ad3fdee
- Commit for layer initialization is included : 4d7a8c4b9bfc998a71df94aec2b2a856aba71478
- Commit to disable nnstreamer nntrainer is included : 450000572e8a7e993d36f06e0413c2ae4abc2e5a
- Commit to support Tizen7.0 is included : 96bb3df9726e3975710817a82c2889f1c26b4484

This PR is tested with AuXFT example on QRB4210 board (w/ Tizen 7.0).

----
**Self evaluation:**

### DIFF
--- a/debian/nntrainer-dev.install
+++ b/debian/nntrainer-dev.install
@@ -24,6 +24,7 @@
 /usr/include/nntrainer/layer_context.h
 /usr/include/nntrainer/layer_devel.h
 /usr/include/nntrainer/layer_impl.h
+/usr/include/nntrainer/acti_func.h
 # custom layer kits
 /usr/include/nntrainer/app_context.h
 # logger
--- a/meson.build
+++ b/meson.build
@@ -88,6 +88,7 @@ if get_option('enable-fp16')
      # comaptible with armv8.0 machines.
      if cxx.has_argument('-mfp16-format=ieee')
        add_project_arguments('-mfp16-format=ieee', language: ['c', 'cpp'])
+       add_project_arguments('-march=armv8.2-a+fp16', language: ['c', 'cpp'])
      else
        message ('The compiler does not support -mfp16-format=ieee. However, according to https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/Half-Precision.html, gcc may use IEEE fp16 anyway. Thus, we will proceed without the option for FP16 support.')
      endif
--- a/nntrainer/layers/layer_devel.h
+++ b/nntrainer/layers/layer_devel.h
@@ -160,6 +160,8 @@ public:
    */
   virtual void finalize(InitLayerContext &context) = 0;
 
+  virtual void initialize(RunLayerContext &context){};
+
   /**
    * @brief     Forward Propagation of a layer
    * @param     context Context of the layer
--- a/nntrainer/layers/layer_devel.h
+++ b/nntrainer/layers/layer_devel.h
@@ -160,6 +160,9 @@ public:
    */
   virtual void finalize(InitLayerContext &context) = 0;
 
+  /**
+   * @brief    Initialize the layer
+   */
   virtual void initialize(RunLayerContext &context){};
 
   /**
--- a/packaging/nntrainer.spec
+++ b/packaging/nntrainer.spec
@@ -393,6 +393,11 @@ export CFLAGS+=" -fprofile-arcs -ftest-coverage"
 export CXXFLAGS+=" -fprofile-arcs -ftest-coverage"
 %endif
 
+%if 0%{?enable_fp16}
+export CFLAGS+=" -march=armv8.2-a+fp16"
+export CXXFLAGS+=" -march=armv8.2-a+fp16"
+%endif
+
 # Add backward competibility for tizen < 6
 %if 0%{tizen_version_major} < 6
 ln -sf %{_includedir}/nnstreamer/nnstreamer.h %{_includedir}/nnstreamer/ml-api-common.h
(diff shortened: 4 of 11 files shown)


meson_options.txt
```diff
@@ -40,7 +40,7 @@ option('enable-blas', type: 'boolean', value: true)
 option('enable-fp16', type: 'boolean', value: false)
 option('enable-cublas', type: 'boolean', value: false)
 option('enable-openmp', type: 'boolean', value: true)
-option('enable-neon', type: 'boolean', value: false)
+option('enable-neon', type: 'boolean', value: true)
```
SeoHyungjun, line 43: How about enabling the `enable-neon` option when necessary?
It doesn't seem to be mandatory, so it may not always need to be set as true. 👍

SeoHyungjun (approved):
LGTM

meson_options.txt
EunjuYang: Yes. You are right; it seems better to make the default Meson option more conservative. I will update it. Thank you.

baek2sm (approved):
LGTM

packaging/nntrainer.spec
```diff
@@ -408,8 +413,8 @@ meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} \
       %{enable_profile} %{enable_nnstreamer_backbone} %{enable_tflite_backbone} \
       %{enable_tflite_interpreter} %{capi_ml_pkg_dep_resolution} \
       %{enable_reduce_tolerance} %{configure_subplugin_install_path} %{enable_debug} \
-      -Dml-api-support=enabled -Denable-nnstreamer-tensor-filter=enabled \
-      -Denable-nnstreamer-tensor-trainer=enabled -Denable-capi=enabled \
+      -Dml-api-support=enabled \
```
jijoongmoon, line 416: I think nnstreamer related defines should be added. 
``` bash
%{enable_nnstreamer_tensor_trainer} %{enable_nnstreamer_tensor_filter}
```

nntrainer/layers/layer_devel.h
```diff
@@ -160,6 +160,8 @@ class Layer {
    */
   virtual void finalize(InitLayerContext &context) = 0;
 
+  virtual void initialize(RunLayerContext &context){};
```
djeong20, line 163: same here

packaging/nntrainer.spec
EunjuYang: They were already in the line 412; that's why I removed these redundant lines.

debian/nntrainer-dev.install
```diff
@@ -24,6 +24,7 @@
 /usr/include/nntrainer/layer_context.h
 /usr/include/nntrainer/layer_devel.h
 /usr/include/nntrainer/layer_impl.h
+/usr/include/nntrainer/acti_func.h
```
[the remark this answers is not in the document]
EunjuYang: I added `acti_func.h` in this file to  achieve consistency with the `nntrainer/layers/meson.build` and spec files, where `acti_func.h` is added as a header. I made this based on jijoong's previous commit. 
BTW, @jijoongmoon , could you please clarify if this modification allows for functions in `acti_func.h` from external sources? or If it is redundant, please let me know.

skykongkong8 (commented):
Shouldn't we consider arch-dep?

meson.build
```diff
@@ -88,6 +88,7 @@ if get_option('enable-fp16')
      # comaptible with armv8.0 machines.
      if cxx.has_argument('-mfp16-format=ieee')
        add_project_arguments('-mfp16-format=ieee', language: ['c', 'cpp'])
+       add_project_arguments('-march=armv8.2-a+fp16', language: ['c', 'cpp'])
```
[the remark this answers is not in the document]
EunjuYang: As far as I know, this option is required to support FP16-SIMD. If this option is missing, it cannot use FP16-SIMD.
I think we may need additional condition beforehand to add this option. Let me add more conditions for the arm version check to support  >= armv8.2 only. Thank you.

DonghakPark (approved):
LGTM!

packaging/nntrainer.spec
```diff
@@ -562,8 +567,14 @@ cp -r result %{buildroot}%{_datadir}/nntrainer/unittest/
 %{_includedir}/nntrainer/util_simd.h
 %if 0%{?enable_fp16}
 %{_includedir}/nntrainer/util_simd_neon.h
+%{_includedir}/nntrainer/blas_neon.h
+%{_includedir}/nntrainer/hgemm.h
+%{_includedir}/nntrainer/hgemm_util.h
 %endif
```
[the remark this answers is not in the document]
skykongkong8: I think refering to [here](https://github.com/nnstreamer/nntrainer/blob/b88301b1c2dfe224749c7e52f1432e2f2db2b443/packaging/nntrainer.spec) might be helpful for you

packaging/nntrainer.spec
jijoongmoon: Thanks.

jijoongmoon (approved):
LGTM

debian/nntrainer-dev.install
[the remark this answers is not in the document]
jijoongmoon: This is required to use activation function when the developer makes his layer.

packaging/nntrainer.spec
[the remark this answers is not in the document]
EunjuYang: In the current version, Meson build is designed to support Neon exclusively on AArch64 with FP16 enabled. To maintain consistency, an architectural condition was added before the block. Please examine and review the update.

jijoongmoon (approved):
LGTM

PR #2721, opened 2024-08-28
[GPU/OpenCL] Initial version of Rotary Embedding Kernel for GPU and generalization via Attention Interface @ open sesame 10/18 11:30

yashSingh0723:
Added initial version of Rotary Embedding kernel for GPU. This includes both` FP32` and `FP16` implementation for GPU kernel.
Changes added with this PR:

- Created `attention_kernel_interface` to handle pre processing.
- Implemeneted OpenCL `rotary_embedding` kernel for GPU for both `fp32` anf `fp16` usecase.
- Refactored `apply_rotary_emb_cl` and  `apply_rotary_emb_cl_fp16` for generalization. 
- Added `unittest_attention_kernels_cl.cpp` to test Attention kernels on GPU.
- Added `testing_rotary_emb` to test rotary embedding on CPU.
- Added `attention_kernel_strings.h` to handle attention kernels at one place.
- Modified attention kernels to remove `cl_context` related dependencies.
- Added `initAttentionCLKernels function to register default attention kernels.
`


nntrainer/tensor/cl_operations/attention_kernels.cpp
```diff
@@ -0,0 +1,273 @@
...
+            for (unsigned int k = 0; k < dim; k++) {
+              unsigned int span = w + k;
+              value = input[b * channel * height * width + c * height * width + h * width + span];
+              if (k < half_) {
+                transformed_value = -1.0f * input[b * channel * height * width + c * height * width + h * width + span + half_];
+              } else {
+                transformed_value = input[b * channel * height * width + c * height * width + h * width + span - half_];
+              }
+              value = value * cos_ptr[k] + transformed_value * sin_ptr[k];
+              // printf("GPU Batch: %u, Height: %u, Channel: %u, Width: %u, K: %u, Span: %u, Value: %f, Transformed Value: %f, cos_ptr[k]: %f, sin_ptr[k]: %f\n",  b, h, c, w, k, span, value, transformed_value, cos_ptr[k], sin_ptr[k]);
```
skykongkong8, line 62: how about removing unnecessary commented print codes

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
```diff
@@ -0,0 +1,195 @@
...
+    cos.assign(seq_len, std::vector<float>(dim, 0));
+
+    auto sin = std::vector<std::vector<float>>();
+    sin.assign(seq_len, std::vector<float>(dim, 0));
+
+    for (unsigned int i = 0; i < seq_len; ++i) {
+#ifdef USE_NEON
+      nntrainer::calc_trigonometric_vals_dup(half_, freqs.data(), cos[i].data(),
+                                             sin[i].data(), i);
+#else
```
skykongkong8, line 48: Is using NEON during cl operations possible?

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
```diff
@@ -0,0 +1,195 @@
...
+                  transformed_value =
+                    -1.0 * in.getValue<float>(b, c, h, span + half_);
+                } else {
+                  transformed_value = in.getValue<float>(b, c, h, span - half_);
+                }
+                value = value * cos_[k] + transformed_value * sin_[k];
+                // printf("CPU Batch: %u, Channel: %u, Height: %u, Width: %u, K:
+                // %u, Span: %u, Value: %f, Transformed Value: %f, cos_ptr[k]:
+                // %f, sin_ptr[k]: %f\n ",  b, c, h, w, k, span, value,
+                // transformed_value, cos_[k], sin_[k]);
```
skykongkong8, line 139: Think this should be removed

test/unittest/unittest_attention_kernels_cl.cpp
```diff
@@ -0,0 +1,233 @@
...
+                          MOD) *
+                           alpha);
+
+  B_fp32.copy(A_fp32);
+
+  // std::cout << "\nA_fp32 and B_fp32 before rotary embedding:" << std::endl;
+  // for (unsigned int i = 0; i < A_fp32.size(); ++i) {
+  //   std::cout << "Element " << i << " -> " << *(A_fp32.getData<float>() + i)
+  //   <<"\t"<<*(B_fp32.getData<float>() + i)<< std::endl;
+  // }
```
skykongkong8, line 72: Think this should be removed

nntrainer/tensor/cl_operations/attention_kernels_fp16.cpp
```diff
@@ -0,0 +1,279 @@
...
+    result = freqs_sinBuf.WriteData(context.command_queue_inst_,
+                                    freqs_sin_flat.data());
+    if (!result) {
+      printf("Failed to write sin data\n");
+      break;
+    }
+
+    result = cosBuf.WriteData(context.command_queue_inst_, cos_.data());
+    if (!result) {
+      printf("Failed to write cos data\n");
```
baek2sm, line 160: Like the "Failed to set ~" log, how about making the freqs_cosBuf and cosBuf logs different in the "Failed to write ~" log?

nntrainer/tensor/cl_operations/attention_kernels.cpp
yashSingh0723: Yes, thank you pointing it out. I've removed all the unecessary comments from the code.

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
yashSingh0723: No, actually I took this code for rotary emb from the felice repo. This file was just to test rotary emb on the cpu and to run the unittests. Regardless of it, I've updated and removed this code.

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
yashSingh0723: Yes, removed.

nntrainer/tensor/cl_operations/attention_kernels_fp16.cpp
yashSingh0723: Yeah sure, Made the necessary changes in the latest commit. Thanks for pointing it out.

baek2sm (approved):
LGTM

EunjuYang (commented):
Can we realize any performance gains from this kernel?  I believe there may be potential for parallelization that has not been exploited yet. Could you think of updating the kernel to make it more parallelizable?

yashSingh0723:
> Can we realize any performance gains from this kernel? I believe there may be potential for parallelization that has not been exploited yet. Could you think of updating the kernel to make it more parallelizable?

Yes I am looking a way to optimize the process better and to make it more parallel. So to follow the flow of `for loops` of cpu code, I came up with this so that `b,c,h,w` and `k` all follow the same pattern.
   ```
 unsigned int gid = get_global_id(0);

    unsigned int b = gid / (channel * height * width);
    unsigned int c = (gid % (channel * height * width)) / (height * width);
    unsigned int h = ((gid % (channel * height * width)) % (height * width)) / width;
    unsigned int w = (((gid % (channel * height * width)) % (height * width)) % width) / dim * dim;
    unsigned int k = (((gid % (channel * height * width)) % (height * width)) % width) % dim;
```

and for the above code snippet, the work_group_cnt I set it as below
```
unsigned int times = batch*channel*height*(width/dim)*dim // which is nothing but b*c*h*w
const int work_groups_count[3] = {(int)times, 1, 1};
```
But as opencl kernel don't go in a sequential way I was not able to get the unittests passed, I am still looking at a way to make it more parallel, I couldn't figure out why the values were not matching if doing it parallelly as according to me even if kernel executes in parallel, for every value of b,c,h,w,k and span, it will point to a individual index of both input and output. But regardless of this, I am looking to optimize the kernel. This is just the initial version. If any suggestions, please give.

**Comment by skykongkong8** [comment]
To make optimization strategy's effect more clear, I would like to suggest to share performance measure table in terms of latency per GPU implemented layers next time

File: nntrainer/tensor/cl_operations/attention_kernel_interface.cpp [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+      float angle = i * freqs[j];
+      cos[i][j] = std::cos(angle);
+      cos[i][j + half_] = std::cos(angle); // repeated 2 times
+
+      sin[i][j] = std::sin(angle);
+      sin[i][j + half_] = std::sin(angle); // repeated 2 times
+    }
+  }
+  freqs_cos = cos;
+  freqs_sin = sin;
```
djeong20, line 53: I don't think using `cos` and `sin` as variable names is a good idea. how about replacing it with `cos_vec`?

nntrainer/tensor/cl_operations/attention_kernel_interface.cpp
```diff
@@ -0,0 +1,136 @@
...
+  unsigned int half_ = dim / 2;
+  for (unsigned int i = 0; i < half_; ++i) {
+    freqs.push_back(1.0 / (std::pow(theta, (2 * i) / static_cast<float>(dim))));
+  }
+
+  auto cos = std::vector<std::vector<float>>();
+  cos.assign(seq_len, std::vector<float>(dim, 0));
+
+  auto sin = std::vector<std::vector<float>>();
+  sin.assign(seq_len, std::vector<float>(dim, 0));
```
djeong20, line 40: I don't think using `cos` and `sin` as variable names is a good idea. how about replacing it with `cos_vec`?

nntrainer/tensor/cl_operations/attention_kernel_interface.cpp
```diff
@@ -0,0 +1,136 @@
...
+ * @param[in] in input tensor
+ * @param[in] dim hidden dim size
+ * @param[in] from sequence order
+ * @param[in] max_timestep maximum timestep
+ */
+void apply_rotary_emb_cl(Tensor &in, unsigned int dim, unsigned int from,
+                         unsigned int max_timestep, RunLayerContext &context) {
+  nntrainer::Tensor out(in.getDim());
+  float value = 0;
+  float transformed_value = 0.0;
```
djeong20, line 67: it is not a big deal, but it would be better to initialize it with 0.0f. ([reference](https://stackoverflow.com/questions/5199338/what-is-the-significance-of-0-0f-when-initializing-in-c))
```suggestion
  float value = 0.0f;
  float transformed_value = 0.0f;
```

nntrainer/tensor/cl_operations/attention_kernel_interface.cpp
```diff
@@ -0,0 +1,136 @@
...
+  freqs_sin = sin;
+}
+
+/**
+ * @brief     apply rotary embedding
+ * @param[in] in input tensor
+ * @param[in] dim hidden dim size
+ * @param[in] from sequence order
+ * @param[in] max_timestep maximum timestep
+ */
```
djeong20, line 62: missing doc?

```suggestion
 * @param[in] context layer context to get the resource manager and queue id
 */
```

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
```diff
@@ -0,0 +1,181 @@
...
+ * @brief     compute frequency for rotary embedding
+ * @param[in] dim hidden dim size
+ * @param[in] seq_len sequency length
+ * @param[out] freqs_cos cosine of the frequencies
+ * @param[out] freqs_sin sine of the frequencies
+ * @param[out] freqs base frequencies array to be used in computation of cos and
+ * sin values for each position in sequence
+ * @param[in] theta rotary angle
+ */
+void precompute_freqs(int dim, unsigned int seq_len,
```
djeong20, line 27: This function shares the same implementation with precompute_freqs() in attention_kernel_interface.cpp. could you explain why you redefined it here?

djeong20 (commented):
Please check the comments!

EunjuYang:
> > Can we realize any performance gains from this kernel? I believe there may be potential for parallelization that has not been exploited yet. Could you think of updating the kernel to make it more parallelizable?
> 
> Yes I am looking a way to optimize the process better and to make it more parallel. So to follow the flow of `for loops` of cpu code, I came up with this so that `b,c,h,w` and `k` all follow the same pattern.
> 
> ```
> unsigned int gid = get_global_id(0);
> 
>  unsigned int b = gid / (channel * height * width);
>  unsigned int c = (gid % (channel * height * width)) / (height * width);
>  unsigned int h = ((gid % (channel * height * width)) % (height * width)) / width;
>  unsigned int w = (((gid % (channel * height * width)) % (height * width)) % width) / dim * dim;
>  unsigned int k = (((gid % (channel * height * width)) % (height * width)) % width) % dim;
> ```
> 
> and for the above code snippet, the work_group_cnt I set it as below
> 
> ```
> unsigned int times = batch*channel*height*(width/dim)*dim // which is nothing but b*c*h*w
> const int work_groups_count[3] = {(int)times, 1, 1};
> ```
> 
> But as opencl kernel don't go in a sequential way I was not able to get the unittests passed, I am still looking at a way to make it more parallel, I couldn't figure out why the values were not matching if doing it parallelly as according to me even if kernel executes in parallel, for every value of `b,c,h,w,k and span`, it will point to a individual index of both input and output. But regardless of this, I am looking to optimize the kernel. This is just the initial version. If any suggestions, please give.

First of all, thank you for trying additional optimization and sharing a detailed idea. 

Regarding the issue of unit test failures, it might be worth checking if the width is divisible by dim. Moreover, I believe you should be cautious regarding potential race conditions involving `cos_ptr` and `sin_ptr`. 

Furthermore, if your code is naively parallelized using the indices you have suggested, there exists a possibility that code lines 45 through 51 could potentially be executed multiple times by different threads with the same values for `b, c, and h`. Please check them as well. 

Looking forward to seeing your next commit! :)

nntrainer/tensor/cl_operations/attention_kernel_interface.cpp
yashSingh0723: yeah sure, I'll do this in the next commit.

nntrainer/tensor/cl_operations/attention_kernel_interface.cpp
yashSingh0723: Sure, I'll update this as well.

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
yashSingh0723: So Initially, testing_rotary_emb.cpp was to test the cpu code before writing the kernel. No specific reason for redefining it again. We can remove it as well, I didn't as initially in gpu kernel the paramters were different as compared to the cpu one. So thats why kept both. It was just for testing purposes.

yashSingh0723:
> > > Can we realize any performance gains from this kernel? I believe there may be potential for parallelization that has not been exploited yet. Could you think of updating the kernel to make it more parallelizable?
> > 
> > 
> > Yes I am looking a way to optimize the process better and to make it more parallel. So to follow the flow of `for loops` of cpu code, I came up with this so that `b,c,h,w` and `k` all follow the same pattern.
> > ```
> > unsigned int gid = get_global_id(0);
> > 
> >  unsigned int b = gid / (channel * height * width);
> >  unsigned int c = (gid % (channel * height * width)) / (height * width);
> >  unsigned int h = ((gid % (channel * height * width)) % (height * width)) / width;
> >  unsigned int w = (((gid % (channel * height * width)) % (height * width)) % width) / dim * dim;
> >  unsigned int k = (((gid % (channel * height * width)) % (height * width)) % width) % dim;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and for the above code snippet, the work_group_cnt I set it as below
> > ```
> > unsigned int times = batch*channel*height*(width/dim)*dim // which is nothing but b*c*h*w
> > const int work_groups_count[3] = {(int)times, 1, 1};
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But as opencl kernel don't go in a sequential way I was not able to get the unittests passed, I am still looking at a way to make it more parallel, I couldn't figure out why the values were not matching if doing it parallelly as according to me even if kernel executes in parallel, for every value of `b,c,h,w,k and span`, it will point to a individual index of both input and output. But regardless of this, I am looking to optimize the kernel. This is just the initial version. If any suggestions, please give.
> 
> First of all, thank you for trying additional optimization and sharing a detailed idea.
> 
> Regarding the issue of unit test failures, it might be worth checking if the width is divisible by dim. Moreover, I believe you should be cautious regarding potential race conditions involving `cos_ptr` and `sin_ptr`.
> 
> Furthermore, if your code is naively parallelized using the indices you have suggested, there exists a possibility that code lines 45 through 51 could potentially be executed multiple times by different threads with the same values for `b, c, and h`. Please check them as well.
> 
> Looking forward to seeing your next commit! :)

Yes, as per my discussion with Mr. Lee, it is given that dim is an divisor of width, although there is no condition to check that, it is a given thing as of now. I pointed this as well in the group. 
Yes I've taken care of the condition for cos_ptr and sin_ptr, the condition from + h < max_timestep, is to make sure that they are always in bound. 
For the parallelization, I am looking for another way. Will update in upcoming commits.

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
djeong20: makes sense! I think a comment or a brief note that this is a testing code for CPU results would be awesome :)

EunjuYang (commented):
The updated kernel seems to have applied our discussions and it looks better 👍  I left a comment on attention kernel interface. Could you please answer to this? Thank you!

nntrainer/tensor/cl_operations/attention_kernel_interface.cpp
```diff
@@ -0,0 +1,137 @@
...
+ * @param[in] context layer context to get the resource manager and queue id
+ */
+void apply_rotary_emb_cl(Tensor &in, unsigned int dim, unsigned int from,
+                         unsigned int max_timestep, RunLayerContext &context) {
+  nntrainer::Tensor out(in.getDim());
+  float value = 0.0f;
+  float transformed_value = 0.0f;
+  unsigned int half_ = dim / 2;
+
+  std::vector<std::vector<float>> freqs_cos = {};
```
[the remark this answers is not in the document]
EunjuYang: One possible way to update is to take freqs_cos and freqs_sin as input, and call the precompute if they are null or empty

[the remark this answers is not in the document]
yashSingh0723: So Initially I also kept freqs_cos, freqs_sin, and freqs as global vector and was only updating it inside precompute_freqs `if(freqs_cos.empty())`, but after looking at the `custom_multi_head_attention_layer.cpp` in felice repo, dim is passed as 
```
  apply_rotary_emb_tensor(projected_query_step, projected_query_dim_prop, _from);
  apply_rotary_emb_tensor(cache_key_step, projected_key_dim_prop, _from);
```
Thats why as per my understanding, I computed both the vectors repeatedly as per the `dim`.

And are you suggesting to pass freqs_cos anf freqs_sin as parameters in `apply_rotary_emb_tensor_cl` function or somewhere else ? If we pass both vectors as parameters then while calling precompute_freqs we can check whether they are empty or not, hence reducing the redundancy of the code.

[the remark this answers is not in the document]
EunjuYang: You're correct in that the dimension for the query/key might be different, but the block will be repeated (stacked several times!). Also, I think `apply_rotary_emb_tensor_cl` will be invoked whenever forwarding of the multi_head_attention layer is called. Thus, it is worth considering reducing unnecessary computation.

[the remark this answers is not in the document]
EunjuYang: It sounds reasonable to call `precompute_freqs()` when the `finalize()` is invoked ! 👍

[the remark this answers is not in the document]
yashSingh0723: I've added a @todo in the latest commit for the above.

EunjuYang (approved):

EunjuYang:
Memory unit test has been failed. Please check it :)

yashSingh0723:
> Memory unit test has been failed. Please check it :)

It is resolved now.

niket-agarwal:
LGTM!

nntrainer/tensor/cl_operations/testing_rotarty_emb.cpp
```diff
@@ -0,0 +1,182 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Yash Singh <yash.singh@samsung.com>
+ *
+ * @file	testing_rotary_emb.cpp
```
jijoongmoon, line 5: if this file is for the testing.. then it might be better to be in unittest dir.

yashSingh0723: Moved the file to unittest directory in the latest commit.

jijoongmoon (approved):
We might need more code restructuring, but we could do it in the later PR. LGTM

PR #2722, opened 2024-08-29
[bugfix] fix coverity issues

djeong20:
This PR resolves coverity issues in the ShortTensor class. Replace max_abs() implementation with maxValue() since the maximum absolute value of unsigned int equals to the maximum value.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/short_tensor.cpp
+++ b/nntrainer/tensor/short_tensor.cpp
@@ -244,6 +244,7 @@ void ShortTensor::copyData(const Tensor &from) {
   switch (from.getDataType()) {
   case ml::train::TensorDim::DataType::UINT16:
     copy(from.getData());
+    break;
   default:
     throw std::invalid_argument("Error: Unsupported data type");
     break;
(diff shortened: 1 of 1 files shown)


EunjuYang (approved):
LGTM. 
BTW, `ShortTensor` sounds like int16 not uint16 (It made me confused). Please think rename this when one adds int16 tensor :)

skykongkong8 (approved):
LGTM!

skykongkong8:
> LGTM. BTW, `ShortTensor` sounds like int16 not uint16 (It made me confused). Please think rename this when one adds int16 tensor :)

I agree. I think it would be better to have something that clarifies its characteristics s.t. `unsigend` or `quantized` or else

nntrainer/tensor/short_tensor.cpp
```diff
@@ -278,20 +279,7 @@ std::vector<unsigned int> ShortTensor::argmax() const {
...
-    if (cur_val > max_val) {
-      max_val = cur_val;
-    }
-  }
-
-  return max_val;
-}
+float ShortTensor::max_abs() const { return maxValue(); }
 
 float ShortTensor::maxValue() const {
```
skykongkong8, line 284: And just curious, if this tensor is containing 'unsigned' integer, should it be `max_abs()` ? it is already absolute.

djeong20: sorry I didn't quite understand. do you mean why `max_abs()` should be used?

skykongkong8: Yes. I think can't see any difference between maxValue and maxAbsoluteValue for this tensor

djeong20: You're right! There's no difference between `max_abs()` and `maxValue()` for this tensor class. That's why max_abs() is using maxValue(). The reason why there are `max_abs()` and `maxValue()` is to be compatible with the TensorBase and Tensor class. While max_abs() is currently a pure virtual function, it will get a compile error without implementing `max_abs()` in the ShortTensor class.

baek2sm (approved):
LGTM!

lhs8928 (approved):
LGTM

PR #2725, opened 2024-08-30
[Wait for #2724][Layer] add "multiply layer" @open sesame 11/12 10:52

baek2sm:
- added "multiply layer" for multiplication

**Self evaluation:**
1. Build test:   [X]Passed [ ]Failed [ ]Skipped
2. Run test:     [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Seungbaek Hong <sb92.hong@samsung.com>

PR #2726, opened 2024-08-30
[Wait for #2725][Layer] add "divide layer"

baek2sm:
- added "divide layer" for division.

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -40,6 +40,7 @@ enum LayerType {
   LAYER_ADD = ML_TRAIN_LAYER_TYPE_ADD,           /**< Add Layer type */
   LAYER_SUBTRACT = ML_TRAIN_LAYER_TYPE_SUBTRACT, /**< Subtract Layer type */
   LAYER_MULTIPLY = ML_TRAIN_LAYER_TYPE_MULTIPLY, /**< Multiply Layer type */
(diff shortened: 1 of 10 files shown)


nntrainer/layers/div_layer.cpp
```diff
@@ -0,0 +1,97 @@
...
+void DivLayer::finalize(InitLayerContext &context) {
+  context.setOutputDimensions({context.getInputDimensions()[0]});
+}
+
+void DivLayer::forwarding(RunLayerContext &context, bool training) {
+  Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+
+  const Tensor &input0 = context.getInput(0);
+  const Tensor &input1 = context.getInput(1);
+
```
EunjuYang, line 35: Maybe we need to check the case by dividing by 0?

skykongkong8: I'm not sure.. Is checking zero-division error at layer level recommendable?
I mean, I tried checking them value-by-value at computation backend level, but it gives a bit of latency overhead during computation process.
Where do you think we should put them? or, do we really need them? (just curious)

EunjuYang: Yes......! your comment seems reasonable. It might cause too much latency; Since this is an atomic tensor operation, it would be better not to perform the value-by-value calculation. It seems more reasonable to let the user handle it :)

EunjuYang (approved):

skykongkong8 (approved):

jijoongmoon (approved):
LGTM

PR #2727, opened 2024-09-03
[ GPGPU ] enable X86 OpenCL with some fixes

jijoongmoon:
This PR enables meson build with opencl for X86.
Also, fix some unittest case errors and types.

Resolves:

**Self evaluation:**

### DIFF
--- a/nntrainer/opencl/opencl_context_manager.cpp
+++ b/nntrainer/opencl/opencl_context_manager.cpp
@@ -166,7 +166,7 @@ bool ContextManager::CreateDefaultGPUDevice() {
     return false;
   }
 
-  char extensions[extension_size];
+  char *extensions = new char[extension_size];
   status = clGetDeviceInfo(device_id_, CL_DEVICE_EXTENSIONS, extension_size,
                            extensions, NULL);
   if (status != CL_SUCCESS) {
@@ -178,6 +178,7 @@ bool ContextManager::CreateDefaultGPUDevice() {
     ml_loge("fp16 (half) is not supported by device");
     return false;
   }
+  delete[] extensions;
 #endif
 
   return true;
--- a/nntrainer/opencl/opencl_program.cpp
+++ b/nntrainer/opencl/opencl_program.cpp
@@ -16,9 +16,11 @@
 #include <cstring>
 #include <fstream>
 #include <string>
+#include <vector>
 
 #include "opencl_loader.h"
 
+#include <memory>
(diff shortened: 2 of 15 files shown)


EunjuYang (approved):
LGTM

nntrainer/opencl/opencl_context_manager.cpp
```diff
@@ -178,6 +178,7 @@ bool ContextManager::CreateDefaultGPUDevice() {
     ml_loge("fp16 (half) is not supported by device");
     return false;
   }
+  delete[] extensions;
```
djeong20, line 181: This may potentially result in a memory leak. when the `status` is not CL_SUCCESS, it will exit from a function without freeing.

djeong20: maybe we can add `delete[] extensions` before `return false` for both if statements.

djeong20: or use `std::vector<char> extensions(extension_size)`.

nntrainer/opencl/opencl_program.cpp
```diff
@@ -95,29 +99,36 @@ bool Program::GetProgramInfo(cl_device_id device_id) {
...
   // Read the binary
   size_t binaries_ptr_alloc_size = sizeof(unsigned char *) * num_devices;
-  unsigned char *binaries_ptr[num_devices];
+
+  // unsigned char *binaries_ptr[num_devices];
+  std::vector<unsigned char *> binaries_ptr;
 
   for (unsigned int i = 0; i < num_devices; ++i) {
-    binaries_ptr[i] = new unsigned char[binaries_size[i]];
+    binaries_ptr.push_back(new unsigned char[binaries_size[i]]);
```
skykongkong8, line 126: I couldn't find a memory release code for this variable. Is it deleted somewhere else later on?

djeong20: it is released in the latter part!

https://github.com/nnstreamer/nntrainer/blob/656a231787c426e021f0c58479d2ffaabf3558b2/nntrainer/opencl/opencl_program.cpp#L155-L158

EunjuYang (commented):
I think `<tensor_dim.h>` is required to use `_FP16`.
Please add the header to `concat_cl.h`, `swiglu_cl.h` `reshape_cl.h`, etc.

myungjoo (approved):

PR #2728, opened 2024-09-05
[ LORA ] Update FC Layer to support LoRA's incremental forwarding & batch_size option.

EunjuYang:
This pull request (PR) consists of two commits:

1) Update 'incremental_forwarding' for 'FullyConnectedLayer'.
    - Code updated to support multiple batches .
    - Code modified to work with incremental forwarding using LoRA.

2) Fix bugs in 'fc_layer.cpp/h' handling of batch size with LoRA.
    - In the previous version, tensor dimensions used in LoRA computations did not account for batch size.
    - 'setBatch' function has been overrided to correctly update the batch size of tensors.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -230,20 +230,39 @@ void FullyConnectedLayer::incremental_forwarding(RunLayerContext &context,
     to = 1;
   }
 
+  input_step_dim.batch(1);
   input_step_dim.height(to - from);
+  hidden_step_dim.batch(1);
   hidden_step_dim.height(to - from);
 
-  // @todo: set reset stride as false. This implementation only works when batch
-  // size is 1
-  Tensor input_step = input_.getSharedDataTensor(input_step_dim, 0, true);
-  Tensor hidden_step = hidden_.getSharedDataTensor(hidden_step_dim, 0, true);
+  // @todo make it parallelized with batch axis
+  for (unsigned int b = 0; b < hidden_.batch(); ++b) {
+    Tensor input_step = input_.getSharedDataTensor(
+      input_step_dim, b * hidden_dim.getFeatureLen(), true);
+    Tensor hidden_step = hidden_.getSharedDataTensor(
+      hidden_step_dim, b * hidden_dim.getFeatureLen(), true);
 
-  input_step.dot(weight, hidden_step, false, false);
+    input_step.dot(weight, hidden_step, false, false);
 
-  if (auto &disable_bias = std::get<props::DisableBias>(*layer_impl_props);
-      disable_bias.empty() || disable_bias.get() == false) {
-    Tensor &bias = context.getWeight(weight_idx[FCParams::bias]);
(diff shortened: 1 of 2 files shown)


nntrainer/layers/fc_layer.cpp
```diff
@@ -230,20 +250,39 @@ void FullyConnectedLayer::incremental_forwarding(RunLayerContext &context,
...
+      Tensor &loraB = context.getWeight(lora_idx[LORAParams::loraB]);
+      Tensor &hidden_tmp_lora =
+        context.getTensor(lora_idx[LORAParams::loraTmp]);
+      Tensor &hidden_out_lora =
+        context.getTensor(lora_idx[LORAParams::loraOut]);
+
+      input_step.dot(loraA, hidden_tmp_lora, false, false);
+      hidden_tmp_lora.dot(loraB, hidden_out_lora, false, false);
+      hidden_out_lora.multiply_i(lora_scaling);
+      hidden_step.add_i(hidden_out_lora);
```
skykongkong8, line 278: Pure question:
won't there be any possibility for `lora_scaling` to be defined as a scalar by any chance?
Although it is not supported in the nntrainer yet, if it is a scalar,
Then we can do something like:
```C++
hidden_tmp_lora.dot(loraB, hidden_out_lora, false, false, lora_scaling);
```
in optimized GEMM case, computing everything in fused-ops..
(Or, even if it is a vector, I can implement a fused op for optimal performance)

To do so, 
We should add params like `alpha` and `beta` at `Tensor::dot()` -> AFAIK current nntr Tensor::dot supports `beta` only...

skykongkong8 (approved):

nntrainer/layers/fc_layer.cpp
EunjuYang: Thank you for asking a question !
The `lora_scaling` is nothing but $$\frac{\alpha}{rank}$$, which is a scalar (Please refer to https://arxiv.org/abs/2106.09685, Section 4.1). It is used for ease of finding the best rank $$r$$.
You're asking about this because there's a chance that we might need normal GEMM, right? It could be one of the cases where the generalized GEMM is used, but I'm not certain if it will be actively utilized.

e.g., The purspoe of using `lora_scaling` is to make a consistent of hyper-parameter across the various lora ranks. But I'm not sure it will be used in on-device training (rank will be fixed or the best rank might be explored in the PC-side)

djeong20 (approved):
nice work! LGTM

lhs8928 (approved):
Overall LGTM except minor comment.

nntrainer/layers/fc_layer.cpp
```diff
@@ -148,7 +159,7 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
                             true, TensorLifespan::FORWARD_DERIV_LIFESPAN);
```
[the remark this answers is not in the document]
EunjuYang: Exactly! I'm gonna update this. Thank you for your comment.

Issue #2723, opened 2024-08-29
[ GPU ] GPU Kernel creation time


EunjuYang:
For now, it seems `clCreateKernel` is called whenever any type of `_cl` wrapper function is invoked.

For example,

`FullyConnectedLayer::forwarding -> dotCl -> dot_cl -> clCreateKernel`

Considering the timing of calling `_cl` functions, this could potentially slow down performance. (Of course, it already avoids duplicate of registration. However, it may not be enough for speed up.)
Since NNTrainer already has a compilation phase, what about to move the kernel registration process into the compilation stage?
During the compilation step, we can identify which computational units are utilized by each layer and generate the corresponding kernels accordingly.

jijoongmoon:
I'm thinking to change the current implementation cl context structure.
I think we need cl kernel context and cl_context has it. and when it creates cl layer, it will set the cl kernel context..
And then, it might be possible to register custom cl kernel at finalize function of cl layer.
and also Tensor kernel which is developed by us ( like gemm cl kenels ) could be initialized when kernel context is 
created. (like default Kernel we provide)

s-debadri:
PR #2732 has been created for addressing this issue. Following is the plan:

- Added `registerClKernel` function at `cl_context` to register custom OpenCl kernels as well as in-house kernels.
- Used hash map to track created `Kernel` objects. `shared_ptr` was used t store inside the map.
- Modified `sscal` using above `Kernel` creation flow to remove dependency of `layer_context`.

> In progress: Removing `layer_context` dependency for existing kernels.

EunjuYang:
**[Suggestion / Need Discussion]**

As I understood, current GPU's ClContext is expected to be created with LayerNode.
What about moving its first creation time to AppContext creation by adding it as a member variable to AppContext?
By doing so, we can further abstract user API to create Layer without directly calling Cl Layer's name.


![image](https://github.com/user-attachments/assets/5f358853-7ddf-49d9-9fee-50f47b5c0067)

PR #2732, opened 2024-09-11
[gpu/enhance] Registering OpenCL kernels at cl_context

s-debadri:
Changes made in this PR:

- Added `registerClKernel` function to register custom OpenCl kernels as well as in-house kernels.
- Exposed `CommandQueueManager` and `ContextManager` instances of `cl_context`.
- Used hash map to track created `Kernel` objects.
- Modified `createLayer` API signature to accept compute engine information.
- Modified blas kernels to remove `cl_context` related dependencies.
- Removed kernel management logic from `layer_context`.
- Temporarily commented out `cl_layers` build and initialization to avoid build failure.

> To Do: Remove dependency of `layer_context` for all `cl_layers`. This can be done in subsequent PR.

**Self evaluation:**


nntrainer/cl_context.h
```diff
@@ -183,12 +191,33 @@ class ClContext {
...
+                                            std::string kernel_name);
+
+  /**
+   * @brief create OpenCl kernel
+   * @param kernel_string reference of implementation string
+   * @param kernel_name reference of kernel_name
+   * @param kernel_ptr_ reference of shared_ptr of Kernel
+   * @return true if successful, false otherwise
+   */
+  bool clCreateKernel(std::string &kernel_string, std::string &kernel_name,
```
EunjuYang, line 210: As I understood, it is not recommended call `clCreateKernel` directly, but should be call `registerClKernel`. Therefore, could we consider altering 'clCreateKernel' into a private member function?

EunjuYang (commented):
It seems beneficial to update the CL functions so they do not require `RunLayerContext` as a parameter! I believe this PR may be linked to Issue #2723. Would you mind briefly sharing your plans regarding this matter in the issue? I feel that it would assist the reviewers in understanding our intended course of action.

s-debadri:
> It seems beneficial to update the CL functions so they do not require `RunLayerContext` as a parameter! I believe this PR may be linked to Issue #2723. Would you mind briefly sharing your plans regarding this matter in the issue? I feel that it would assist the reviewers in understanding our intended course of action.

Added the course of action in Issee #2723

nntrainer/cl_context.h
s-debadri: Updated the commit with this change

jijoongmoon:
I thought we might need another kernel context to handle the kernels more efficiently. But embracing the kernels in cl context is also fine for me. However, I suggest handling the blas kernels as a kind of default kernels in NNTrainer if opencl option is enabled so that, we can create kernels at initialize time. These kernels are used and they might be better managed in the same directory. so.. make separate files (let's says blas_kernels.h) and save the blas-related kernel strings in this file and register them when the cl context is created. Then we can use them in cl tensors without clKernelCreate call.

jijoongmoon (approved):
Overall, LGTM

s-debadri:
> I thought we might need another kernel context to handle the kernels more efficiently. But embracing the kernels in cl context is also fine for me. However, I suggest handling the blas kernels as a kind of default kernels in NNTrainer if opencl option is enabled so that, we can create kernels at initialize time. These kernels are used and they might be better managed in the same directory. so.. make separate files (let's says blas_kernels.h) and save the blas-related kernel strings in this file and register them when the cl context is created. Then we can use them in cl tensors without clKernelCreate call.

Got it. Shall I make the changes related to default kernel registration in this same PR or different?

PR #2735, opened 2024-09-13
[ Tensor ] UInt Tensor : UInt8 / UInt16 / Uint32 @open sesame 09/23 16:50

### LINKED ISSUE #2733: [ Tensor ] more types of Tensors are required.
Our awesome TensorV2 has been successfully activated.
As of September 12th, 2024, we support the following types of tensors: 

FloatTensor : **`fp32`**, 
HalfTensor : **`fp16`**, 
CharTensor : **`qint8`**, 
ShortTensor : **`uint16`**.

For further expansion of our on-device training capabilities, we also need to incorporate tensors using unsigned 32 bit integers (**`uint32`**) and unsigned 8 bit integers (**`uint8`**).

EunjuYang:
- This commit is related to #2733
- This commit implements template of UIntTensor based on the ShortTensor class.
- Based on the template, this commit supports UInt8 / UInt16 / UInt32
- Implement UIntTensor Template
- Implement UInt8Tensor / UInt16Tensor / UInt32Tensor (ShortTensor is replaced)
- Unit tests for UInt8 and UInt32 are added

**Update 2024-09-23 14:54**
- Resolving NDK build error and reflecting some review comments.
- This PR contains 11044d7aaa44f0f2cc8f166df0bc153c791f20c9 and 591665032b282038b05fea948a9d5738387719cb
- 591665032b282038b05fea948a9d5738387719cb : removes some headers (`float_tensor`, `char_tensor`, `uint_tensor.h`), where their corresponding `.cpp` file imports `tneosr.h`. Reason of this change is
    - In order to clarify the files relationship.
    - uint_tensor.h should contain all implementations in one file, which is additional update in 11044d7aaa44f0f2cc8f166df0bc153c791f20c9.
- 11044d7aaa44f0f2cc8f166df0bc153c791f20c9 : defines template uint_tensor class based on ShortTensor. Since it is based on the template class, its implementation should be in one file (NDK build does not work with template class declaration)

Self evaluation:


nntrainer/tensor/uint_tensor.cpp
```diff
@@ -1,46 +1,54 @@
...
-  TensorBase(name_, fm, Tdatatype::UINT16) {}
+template class UIntTensor<uint8_t, Tdatatype::UINT8>;
+template class UIntTensor<uint16_t, Tdatatype::UINT16>;
+template class UIntTensor<uint32_t, Tdatatype::UINT32>;
 
-ShortTensor::ShortTensor(const TensorDim &d, bool alloc_now, Initializer init,
-                         std::string name) :
+template <typename T, Tdatatype D>
+UIntTensor<T, D>::UIntTensor(std::string name_, Tformat fm) :
+  TensorBase(name_, fm, D) {}
```
djeong20, line 27: It seems like `Tdatatype D` is used only in constructors. I think it could be removed.

```suggestion
template <typename T>
UIntTensor<T>::UIntTensor(std::string name_, Tformat fm) :
  TensorBase(name_, fm, checkTensorDataType()) {}
```

and define a private function `checkTensorDataType()`.
```
// uint_tensor.h
Tdatatype checkTensorDataType() const {
  if (typeid(T) == typeid(uint8_t))
    return Tdatatype::UINT8;
  else if (typeid(T) == typeid(uint16_t))
    return Tdatatype::UINT16;
  else if (typeid(T) == typeid(uint32_t))
    return Tdatatype::UINT32;
  else
    throw std::runtime_error("unsupported type");
}
```

api/ccapi/include/tensor_dim.h
```diff
@@ -49,12 +49,14 @@ class TensorDim {
...
    * @brief Tensor Data Type.
-   * Currently support QINT4, QINT8, UINT16, FP16 & FP32
+   * Currently support QINT4, QINT8, UINT8, UINT16, UINT32, FP16 & FP32
    */
   enum class DataType {
     QINT4,  /** quantized int 4*/
     QINT8,  /** quantized int 8*/
+    UINT8,  /** unsigned int 8 bit */
     UINT16, /** unsigned int 16 bit */
+    UINT32, /** unsigned int 32 bit */
```
djeong20, line 59: please check tensor_dim.cpp to reflect this patch. (e.g., `getDataTypeSize()`, `operator<<`)

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -250,6 +250,52 @@ TEST(nntrainer_Tensor, Tensor_05_p) {
...
+      for (size_t h = 0; h < tensor.height(); ++h) {
+        for (size_t w = 0; w < tensor.width(); ++w) {
+          size_t idx = tensor.getIndex(b, c, h, w);
+          ASSERT_EQ(idx, tensor.getValue<uint32_t>(idx));
+        }
+      }
+    }
+  }
+}
+
```
djeong20, line 298: one negative test case that comes to my mind is comparing UInt16Tensor t1 and UInt32Tensor t2 with the same size and values and checking for ASSERT_NE(t1, t2).

djeong20 (commented):
Great work overall! Please check the comments below.

nntrainer/tensor/uint_tensor.cpp
EunjuYang: Thank you for your suggestion. It sounds great! It seems much better to update the code like you recommended. I will update the code.

api/ccapi/include/tensor_dim.h
EunjuYang: I updated the  [`tensor_dim.cpp`](https://github.com/nnstreamer/nntrainer/pull/2735/files#diff-8a77082124d610f5fb8ab7e7cc4efab8ca088a9f98ecf8c5288cfbecd55d91f8). Thank you for the comment.

nntrainer/tensor/tensor.h
```diff
@@ -203,6 +197,8 @@ class Tensor {
    * @note      This constructor copies vector again. needs refactoring
    * @param[in] d data for the Tensor with batch size one
    * @param[in] t_type Tensor Type
+   * @todo      It is more desirable to move this implementaton into
+   *            `tensor.cpp`, for it requires half_tensor.h
```
djeong20, line 201: I see all other data type implementations have been moved to .cpp except for half tensor. could you explain why you left it as todo?

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -1,46 +1,46 @@
...
+ *          included by uint_tensor.h only.
  * @see		https://github.com/nnstreamer/nntrainer
  * @author	Donghyeon Jeong <dhyeon.jeong@samsung.com>
+ * @author	Eunju Yang <ej.yang@samsung.com>
  * @bug		No known bugs except for NYI items
  */
 
-#include <iomanip>
-#include <iostream>
+#ifdef __UINT_TENSOR_H__
```
djeong20, line 15: I'm not sure whether having uint_tensor.cpp file is good or not. is there a particular reason why you preserved the cpp file?

djeong20 (commented):
Thank you for the update 👍  I left some comments on the new patch. PTAL!

nntrainer/tensor/tensor.h
EunjuYang: I attempted to relocate it to 'tensor.cpp' with `#ifdef`, but encountered a linking error during compilation. Thus, I decided to leave it unchanged in the current version and will investigate soon.

nntrainer/tensor/uint_tensor.cpp
EunjuYang: You are correct that there is no need to separate the files, and it may actually be more natural to include everything in one header file. However, I believed doing so could enhance readability and attempted to reach a consensus with other tensor implementations. Please share your thoughts, as I plan to incorporate the reviewers' feedback into the PR.

djeong20: I think there's no right answer to this. I was just curious if there are any reasons behind this and I agree with you about the benefits of having a cpp file!

djeong20 (approved):
Appreciate the hard work! great job :) I think it is good to go 👍

DonghakPark (approved):
Great discussion & Changes. LGTM!!

baek2sm (approved):
great work. LGTM!

nntrainer/tensor/tensor.h
```diff
@@ -231,16 +227,41 @@ class Tensor {
     Tensor(std::vector<std::decay<decltype(d)>::type>{d}, t_type){};
 #endif
 
+  /**
+   * @brief     Constructor of Tensor
+   * @param[in] d data for the Tensor. It needs to set format properly.
+   * @param[in] t_type Tensor Type
+   */
+  Tensor(std::vector<std::vector<std::vector<std::vector<uint8_t>>>> const &d,
```
jijoongmoon, line 235: We might need a tensor constructer with zero and scale.

EunjuYang: What else do we need besides the constructor with zero and scale?
Do you expect other methods or functions to improve with the parameters?

jijoongmoon: For now, Zero and Scale (their vector type may need to be supported) are the only two I can think of, but there are more. @djeong20, let's discuss it. I think this can be done in a later PR.

jijoongmoon: Another think is that we need to implement the tensor operations like add, mul, etc.

PR #2736, opened 2024-09-18
[bugfix] Fix sgemv_cl function call from blas_kernel_interface @open sesame 14/10 10:42

s-debadri:
Fixed `sgemv_cl` function call. Was failing unittest after recent changes.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
@@ -138,8 +138,8 @@ void dotCl(Tensor const &input, Tensor const &m, Tensor &result, bool trans,
     /// = R^T = (K * N) ^T * (1 * K) ^T = (N * K) * (K * 1) = (N * K) * (1 * K)
     /// Effectively a translation of sgemv
     else if (M == 1) {
-      trans_m ? sgemv_cl(mdata, data, rdata, mdim2, mdim1, ldb)
-              : sgemv_cl(mdata, data, rdata, mdim1, mdim2, ldb);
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
```diff
@@ -139,8 +139,8 @@ void dotCl(Tensor const &input, Tensor const &m, Tensor &result,
     /// = R^T = (K * N) ^T * (1 * K) ^T = (N * K) * (K * 1) = (N * K) * (1 * K)
```
EunjuYang, line 139: This is just a simple question.
Why do you use `mdim1`, `mdim2` instead of `N` or  `K` ?
Because you computed `M`, `K`, and `N` accordance with various trans options, I wonder why you use `mdim1` and `mdim2` here, not the variables you already computed. (It is hard to match with your comment on line 137 for me😢).
Can you please explain this?

s-debadri: Here `M`, `K` and `N` will depend on both `trans` and `trans_m`. So while calling `sgemv_cl`, we might have to write more conditions and it might get difficult to manage. So, to keep things simple, `mdim1` and `mdim2` was used. I have followed the same paradigm and comments used in `tensor.cpp` for `dot` operation.

djeong20 (approved):

skykongkong8 (approved):
LGTM

EunjuYang (approved):
LGTM

baek2sm (approved):
LGTM

PR #2737, opened 2024-09-23
[Filter] util to get nth info

jaeyun-jung:
Use util function to get nth tensor-info ptr.
Also, remove unnecessary header include in filter subplugin.

### DIFF
--- a/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
+++ b/nnstreamer/tensor_filter/tensor_filter_nntrainer.cc
@@ -23,8 +23,6 @@
 
 #include <nntrainer_error.h>
 
-#include "ml-api-common.h"
-#include "nnstreamer.h"
 #include "nnstreamer_plugin_api.h"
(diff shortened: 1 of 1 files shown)


DonghakPark (approved):
LGTM!

djeong20 (approved):
LGTM! Is it because using the util function is safer than directly accessing nth info?

jaeyun-jung:
> LGTM! Is it because using the util function is safer than directly accessing nth info?

tensors-info struct has only 16 array slots, others (max number of tensors is 256) are allocated memory.
If the number of tensors are larger than 16, out-of-bound error.

skykongkong8 (approved):

DonghakPark:
Please Rebase (upstream/main), https://github.com/nnstreamer/nntrainer/pull/2741 resolve cacheLoader_test_case issue

lhs8928 (approved):
LGTM

baek2sm (approved):
LGTM

PR #2738, opened 2024-09-24
[gpu/enhance] Utility for registering Blas kernels during initialization @open sesame 10/02 11:44

s-debadri:
Changes made in this PR:

- Added `initBlasClKernels` function to register default Blas kernels.
- Updated `layer_node.cpp` to incorporate `initBlasClKernels` calls.
- `blas_kernel_string.h` added to handle Blas kernels at one place.
- Modified unit test to remove `layer_context` dependency.

Usage:

```cpp
auto &cc = nntrainer::ClContext::Global();
cc.initBlasClKernels();
```

**Self evaluation:**


test/unittest/unittest_blas_kernels_cl.cpp
```diff
@@ -27,17 +27,13 @@
...
 using namespace nntrainer;
 
-static RunLayerContext setUpGpuContext() {
-
+static void setUpGpuContext() {
   auto &ac = nntrainer::ClContext::Global();
-  auto rc = RunLayerContext();
-
-  return rc;
+  ac.initBlasClKernels();
```
EunjuYang, line 32: What about calling `initBlasClKernels()` inside of the `Global()` ?
If not, I think you need to add `cc.initBlasClKernels();` in `layer_node.cpp`  (line 159) as well.

nntrainer/tensor/cl_operations/blas_kernels.h
```diff
@@ -22,7 +22,7 @@
 namespace nntrainer {
 
 // get global cl_context to use in kernels
-extern ClContext &cl_context_ref;
+static ClContext &cl_context_ref = ClContext::Global();
```
EunjuYang, line 25: Don't we need to add `cl_context_ref.initBlasClKernels(); ` here?

EunjuYang (commented):
Overall, good to separate registerClKernel.
I wonder two things:
1) Isn't it better to call `initBlaseClKenelrs()` inside of the `Global()` ? Since initBlasClKenrnels() registers very basic blas kernels, it might be better to make it as a default registered at Global() context is created.

2) If (1) does not hold, don't we need to add some code lines to call `initBlasClKenrles` in your modification? e.g., `layer_node.cpp` / `blas_kernels.h`.

test/unittest/unittest_blas_kernels_cl.cpp
s-debadri: There is some issue while calling `initBlasKernels()` from `ClContext` functions itself. `clBuildProgram` was failing. I haven't got a concrete reason for that yet. I assume there is some memory related undefined behaviour when creating a static reference and calling these OpenCL functions within that.

I have kept it such that users can write:
```cpp
auto &cc = nntrainer::ClContext::Global();
cc.initBlasClKernels();
```
This will initialize all the default blas related kernels at the very beginning.
Also as you suggested, it can be added in `layer_node.cpp` as well which will initialize these kernels when any first `LayerNode` is created.

nntrainer/tensor/cl_operations/blas_kernels.h
s-debadri: Similar to what I mentioned above, if users call `initBlasClKernels()`, then the kernels will get registered at the beginning. Otherwise each kernel implementation will separately get registered if its not registered initially. Here `cl_context_ref.initBlasClKernels();` won't be required if users call `initBlasClKernels()` before. If not then individually the kernels will get registered whenever they are called.

s-debadri:
> Overall, it seems good to split `registerClKernel` separately! I wonder two things:
> 
> 1. Isn't it better to call `initBlaseClKenelrs()` inside of the `Global()` ? Since initBlasClKenrnels() registers very basic blas kernels, it might be better to make it as a default registered at Global() context is created.
> 2. If (1) does not hold, don't we need to add some code lines to call `initBlasClKenrles` in your modification? e.g., `layer_node.cpp` / `blas_kernels.h`.

Currently, there are two scenarios that might happen:
- If user calls `initBlasClKernels()` inside there application, then all the Blas kernels will get registered in the beginning.
- If user does not call `initBlasClKernels()`, then whenever there is any function call related to Blas kernel, it will get registered at that particular time.

As you suggested, if `initBlasClKernels()` is added to `layer_node.cpp` in that case users will not be required to call the function. The registration of the Blas kernels will happen at the beginning whenever any first GPU `LayerNode` is created.
Shall I incorporate the changes required for this?

nntrainer/tensor/cl_operations/blas_kernels.h
EunjuYang: Hmm.. It sounds `ClContext::Global()` can be called multiple times, and they return different global ClContext right?
I think I missed its flow. I'm gonna take a look at it again.

s-debadri: It will return the same static instance. I have slightly updated this line:
`static ClContext &cl_context_ref = ClContext::Global();` -> `static ClContext cl_context_ref;`

jijoongmoon (approved):
LGTM

PR #2740, opened 2024-09-26
[ccapi] add return last output option in incremental inference @open sesame 12/04 16:19

lhs8928:
- Added return last output option in incremental inference. To support backward compatibility, default value would be set as false.

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -851,7 +851,7 @@ sharedConstTensors NeuralNetwork::incremental_inference(
 std::vector<float *> NeuralNetwork::incremental_inference(
   unsigned int batch_size, const std::vector<float *> &input,
   const std::vector<float *> &label, unsigned int init_seq_len,
-  unsigned int from, unsigned int to) {
+  unsigned int from, unsigned int to, bool output_hidden_state) {
   sharedConstTensors input_tensors, output_tensors;
   auto in_dim = getInputDimension();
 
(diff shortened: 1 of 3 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -849,27 +849,33 @@ std::vector<float *> NeuralNetwork::incremental_inference(
   unsigned int step = from ? 0 : to - 1;
 
   for (auto &out : output_tensors) {
-    const auto &out_t = *out.get();
-    float *last_out_buf_data = new float[batch_size * out_t.width()];
+    auto out_t = *out.get();
```
djeong20, line 852: is there any reason why you changed it to `auto`? 
If out_t is not modified, `const auto &` would be a better option to avoid the coverity issue.

djeong20 (approved):
LGTM! please check the comments

skykongkong8 (approved):

nntrainer/models/neuralnet.cpp
```diff
@@ -816,7 +816,7 @@ sharedConstTensors NeuralNetwork::incremental_inference(
 std::vector<float *> NeuralNetwork::incremental_inference(
   unsigned int batch_size, const std::vector<float *> &input,
   const std::vector<float *> &label, unsigned int init_seq_len,
-  unsigned int from, unsigned int to) {
+  unsigned int from, unsigned int to, bool return_last_output_only) {
```
SeoHyungjun, line 819: How about setting it to "need_all_output" or "need_only_last" instead of "return_last_output_only"?
(I think the word 'return' is unnecessary and could be replaced with a better name)
It seems like an internal flag that determines the return value.

SeoHyungjun (approved):

DonghakPark:
Please Rebase (upstream/main), https://github.com/nnstreamer/nntrainer/pull/2741 resolve cacheLoader_test_case issue

nntrainer/models/neuralnet.cpp
baek2sm: I think the current parameter name is also good.
However, if there's a necessity for renaming it, I would recommend considering 'output_hidden_states'. Because this term has been utilized in HuggingFace for similar purposes. (if "output_hidden_states == false", then only return the last output values)

SeoHyungjun: If there is a similar case, it would be good to use the same name. 
It seems that developers will easily understand this.

EunjuYang (approved):

nntrainer/models/neuralnet.cpp
lhs8928: Renamed it as @baek2sm suggested.

nntrainer/models/neuralnet.cpp
lhs8928: Unfortunately there might be a chance that out_t could be change.
Please check line 876.

PR #2741, opened 2024-09-30
[bugfix] Fix memcheck in CacheLoader unit tests @open sesame 10/02 10:47

djeong20:
This pull request fixes the issue of failing Continuous Integration.
The new patch checks for nullptr after a flush operation on the cache pool.
This adjustment is expected to rectify the previous failures encountered during the CI process.

**Self-evaluation:**

### DIFF
--- a/test/unittest/memory/unittest_cache_loader.cpp
+++ b/test/unittest/memory/unittest_cache_loader.cpp
@@ -182,10 +182,12 @@ TEST_F(CacheLoaderTest, load_async_01_p) {
 
   // Check memory loading for each execution order
   int id = loader->loadAsync(1, complete);
-  EXPECT_EQ(mem->getAddr(), nullptr);
 
   future.wait();
   pool->flush();
+
+  EXPECT_NE(id, 0);
+  EXPECT_EQ(mem->getAddr(), nullptr);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


skykongkong8 (commented):
This is one of old problems in the nntrainer.. hope this patch solves the problem

DonghakPark (approved):
Great! LGTM

test/unittest/memory/unittest_cache_loader.cpp
```diff
@@ -221,13 +223,14 @@ TEST_F(CacheLoaderTest, load_async_02_p) {
   };
 
   int id = loader->loadAsync(1, complete1);
+  future.wait();
+  pool->flush();
+
```
djeong20, line 228: In synchronous loading tests, it does not check if `mem->getAddr()` is a nullptr after load.
Instead, it checks if the memory is properly loaded after load.

https://github.com/nnstreamer/nntrainer/blob/8104cbe29bce20019bb829edee5dd5d050e5498a/test/unittest/memory/unittest_cache_loader.cpp#L118-L122


On the other hand, asynchronous loading tests check if memory is properly loaded with execution order by using a lambda function `complete`. Current load_async test logic checks if all memory is swapped out after `loader->loadAsync()`. While no flush logic is found in `loadAsync()` (at least on my side), checking if the memory data is swapped out after loading and before the cache pool flushes seems incorrect. Therefore, I added `pool->flush()` before checking if the memory address is a nullptr.

skykongkong8 (approved):
It seems this PR solved the issue! Great work!

EunjuYang (approved):
LGTM! Great work! 👍

jijoongmoon (approved):
LGTM

PR #2745, opened 2024-10-04
    [GPU/OpenCL] Updated the SwiGLU, Reshape and Concat Layers with latest GPU pipeline changes @open sesame 10/04 17:23

niket-agarwal:
Updated the **SwiGLU**, **Reshape**, and **Concat** layer with the new shared_ptr flow.
    Replaced `clCreateKernel` with `registerClKernel` for all these layers.

       **Self evaluation:**
    
### DIFF
--- a/nntrainer/layers/cl_layers/concat_cl.h
+++ b/nntrainer/layers/cl_layers/concat_cl.h
@@ -15,6 +15,7 @@
 #define __CONCAT_LAYER_CL_H__
 #ifdef __cplusplus
 
+#include <cl_context.h>
 #include <common_properties.h>
 #include <layer_context.h>
 #include <layer_devel.h>
@@ -31,6 +32,10 @@ namespace nntrainer {
  * @brief   Concat Layer
  */
 class ConcatLayerCl : public Layer {
+
+private:
+  inline static ClContext cl_context_ref;
+
 public:
   /**
    * @brief     Constructor of Concat Layer
@@ -113,10 +118,8 @@ public:
    * @param[in] input1 Tensor
    * @param[in] input2 Tensor
    * @param[in] result Tensor
-   * @param[in] RunLayerContext reference
    */
-  void ConcatProcess(Tensor const &in1, Tensor const &in2, Tensor &result,
-                     RunLayerContext &context);
+  void ConcatProcess(Tensor const &in1, Tensor const &in2, Tensor &result);
 
   /**
    * @brief     concat computation for axis 3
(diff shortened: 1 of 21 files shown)


nntrainer/layers/cl_layers/concat_cl.h
```diff
@@ -129,13 +132,11 @@ class ConcatLayerCl : public Layer {
...
    * @param[in] input1_width   represents the width of the input tensor A
    * @param[in] input2_width   represents the width of the input tensor X
-   * @param[in] context RunLayerContext reference
    */
   void concat_cl_axis3(const float *matAdata, const float *vecXdata,
                        float *vecYdata, unsigned int input1_batch_size,
                        unsigned int input1_channels, unsigned int input1_height,
-                       unsigned int input1_width, unsigned int input2_width,
-                       RunLayerContext &context);
+                       unsigned int input1_width, unsigned int input2_width);
```
EunjuYang, line 140: Please add `ENABLE_FP16` condition to this header.

```suggestion
#ifdef ENABLE_FP16
```
Besides this function prototypes, all `concat_cl_axis*_fp16` should be imported only when `ENABLE_FP16` is defined.
Please consider this conditional compilation for `cpp` file as well.

nntrainer/layers/cl_layers/reshape_cl.cpp
```diff
@@ -115,185 +114,178 @@ void ReshapeLayerCl::ReshapeProcess(Tensor const &input, Tensor &output,
...
 
   do {
-    result = context.clCreateKernel(copy_cl_kernel_, context.LayerKernel::COPY,
-                                    ReshapeLayerCl::kernel_copy);
-    if (!result) {
+    ClContext::SharedPtrClKernel kernel_copy_ptr =
+      cl_context_ref.registerClKernel(copy_cl_kernel_fp16_, "copy_cl_fp16");
+    if (!kernel_copy_ptr) {
       break;
     }
```
EunjuYang, line 144: What about making this part as a separate function and letting the function to be called when the `clLayer` is registered? This opinion comes from #2723. Could you please think about this and share your opinion?

nntrainer/layers/cl_layers/concat_cl.h
```diff
@@ -148,15 +149,13 @@ class ConcatLayerCl : public Layer {
    * @param[in] input1_height   represents the height of the input tensor
    * @param[in] input1_width   represents the width of the input tensor A
    * @param[in] input2_width   represents the width of the input tensor X
-   * @param[in] context RunLayerContext reference
    */
   void concat_cl_axis3_fp16(const __fp16 *matAdata, const __fp16 *vecXdata,
```
EunjuYang, line 153: Please add `ENABLE_FP16` condition to this header.
```suggestion
#ifdef ENABLE_FP16
  void concat_cl_axis3_fp16(const __fp16 *matAdata, const __fp16 *vecXdata,
```
Besides this function prototypes, all concat_cl_axis*_fp16 should be imported only when ENABLE_FP16 is defined.
Please consider this conditional compilation for cpp file as well.

nntrainer/layers/cl_layers/reshape_cl.cpp
EunjuYang: About this comment, we may need additional discussions and apply it in another PR. As I understood, this commit aims to update the kernels to work with the recent commit. You can ignore this comment for this PR now :)

s-debadri:
Please rebase it so that it's in sync with PR #2738 (merged). make sure to add the kernel strings in `blas_kernel_strings.h`.

PR #2751, opened 2024-10-09
[GPU/OpenCL] Updated the SwiGLU, Reshape and Concat Layers with latest GPU pipeline changes

niket-agarwal:
Updated the **SwiGLU**, **Reshape**, and **Concat** layer with the new shared_ptr flow.
Replaced `clCreateKernel` with `registerClKernel` for all these layers.

   **Self evaluation:**


EunjuYang (approved):
LGTM

baek2sm (approved):
LGTM.
In the future, it would be better to split one layer into one commit.
Then it will be helpful for smooth review. Thanks :)

PR #2753, opened 2024-10-10
[ GPU ] fix return by reference to return by value

EunjuYang:
- In the previous code, `registerClKernel` function returned reference of `SharedPtrClKernel`, causing the possible two problems: 
	- returning by reference cannot return nullptr, leading an error in line 170 of `cl_context.cpp`
	- returning by reference of `shared_ptr` may not increase the reference counter, leading to early free problem.

### DIFF
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -153,7 +153,7 @@ void ClContext::initBlasClKernels() {
   blas_kernels_initialized = true;
 }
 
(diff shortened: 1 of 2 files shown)


lhs8928 (approved):
LGTM

baek2sm (approved):
LGTM

SeoHyungjun (approved):
LGTM

djeong20 (approved):
nice catch!

PR #2757, opened 2024-10-16
[Loss&Test] Implement KLD Loss & Add Unit Test on loss layers @open sesame 10/28 13:48

DonghakPark:
## In this PR
In this PR, performed three tasks as follows:

commit 1 - [[UnitTest] Add kld-loss unit test](https://github.com/nnstreamer/nntrainer/commit/dba998f84449a7fe80a7426e7da97c0812d69397) 
- To increase coverage, we added a KLD loss layer using LayerSemanticsParamType to the unit test case.

commit 2 - [[UnitTest] Add integration_tests on unit test](https://github.com/nnstreamer/nntrainer/commit/47ba1cafbd4db8a2323110bf12861f07b0b94895) 
- Considering the characteristics of NNTrainer, it was difficult to accurately test one loss or BN layer, so we created integration tests and wrote exception cases.

commit 3 - [[Loss] Implement KLD Loss Layer](https://github.com/nnstreamer/nntrainer/commit/5dbaa12a63aaced937c665cb30958b461af8d27c) 
- implemented KLD Loss in torch
image
- KLD_loss forwarding formula
image

**Self evaluation:**
    

SeoHyungjun:
There is a typo in 'implemented `KDL` Loss layer as below'.
It should be corrected to 'implemented `KLD` Loss layer as below.'

nntrainer/layers/loss/kld_loss_layer.cpp
```diff
@@ -20,9 +21,50 @@ KLDLossLayer::KLDLossLayer() {}
...
+  auto &temp = context.getTensor(temp_idx);
+  auto &before_sum = context.getTensor(before_sum_idx);
 
-void KLDLossLayer::calcDerivative(RunLayerContext &context) {}
+  mu.pow(2.0f, temp);                 // 1. temp = mu ^ 2
+  log_std.subtract(temp, before_sum); // 2. before_sum = log_std - temp
+  log_std.apply<float>(expf, temp);   // 3. temp = exp(log_std) - 1
+  temp.subtract_i(1.0f);
+  before_sum.subtract_i(temp);          // 4. before_sum = before_sum - temp
+  before_sum.sum({1, 2, 3}, ret, -0.5); // 5. sum * 0.5
```
SeoHyungjun, line 44: I don't fully understand the forward implementation of KLD Loss in commit 3. 
Are the written formulas and actual implementations the same?

I would like to know the difference between the formula on line 32 of kld_loss_layer.cpp and the PyTorch's KLD formula added to the PR.

In PyTorch, it is summarized as follows:
```
if not log_target: # default
    loss_pointwise = target * (target.log() - input)
else:
    loss_pointwise = target.exp() * (target - input)
```

SeoHyungjun, line 44: The comment says "sum * 0.5".
However, in the code, "-0.5" is used.
Is this correct?

nntrainer/layers/loss/kld_loss_layer.cpp
```diff
@@ -20,9 +21,50 @@ KLDLossLayer::KLDLossLayer() {}
...
+void KLDLossLayer::setProperty(const std::vector<std::string> &values) {
+  if (values.size()) {
+    throw std::invalid_argument(
+      "kld loss does not take any properties, but values given");
+  }
+}
 
-void KLDLossLayer::forwarding(RunLayerContext &context, bool training) {}
+void KLDLossLayer::forwarding(RunLayerContext &context, bool training) {
+  // -0.5 * sum(1 + log_std - pow(mu, 2) - exp(log_std))
```
SeoHyungjun, line 32: If it is the same as torch's KLD formula, it would be great if you could write down the derivation process of the formula like in the comment for the calcDerivative function.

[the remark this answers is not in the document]
DonghakPark: Thank you for your detailed review. Based on the formula below, 
https://stackoverflow.com/questions/74865368/kl-divergence-loss-equation

image

I have attached the Torch formula to assist in understanding. As Hyungjun and Eunju mentioned, it seems necessary to discuss which assumptions we need to implement this.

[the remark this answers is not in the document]
EunjuYang: Thank you for your detailed explanation. As i mentioned in the comment, however, it seems the forwarding of KLD loss only covers the special case of $KL(p(x)||q(x))$, where $q(x) \sim \mathcal{N}(\mu , \sigma^2)$ and $p(x) \sim \mathcal{N}(0,1)$.
The link you shared refers the code of a [page](https://ai.stackexchange.com/questions/26366/how-is-this-pytorch-expression-equivalent-to-the-kl-divergence), which considers the special case of KL loss for training VAE, ([here](https://ai.stackexchange.com/questions/26366/how-is-this-pytorch-expression-equivalent-to-the-kl-divergence)). I recommend you to consider updating this to make our KLD loss cover more general form :)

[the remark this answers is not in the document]
DonghakPark: Oh, i see 
i will update kld loss for more general cases

nntrainer/layers/loss/kld_loss_layer.cpp
```diff
@@ -16,13 +17,61 @@
...
+    temp.apply<float>(logf, temp);
+    output.fill(temp.sum({0, 1, 2, 3}));
+  }
+}
+
+void KLDLossLayer::calcDerivative(RunLayerContext &context) {
+  /**
+   * d/dQ = -P**2/(Q**2) * ln(P/Q)
+   */
+  nntrainer::Tensor &predicted = context.getInput(SINGLE_INOUT_IDX); // Q
```
EunjuYang, line 60: Could you check the derivative formula?
In my opinion,  $KLD(P| Q) = P \log(\frac{P}{Q}) = P \log P - P \log Q$, so $\frac{d}{dQ} KLD(P|Q) = - \frac{P}{Q}$.
(According to the way of reduction, the derivative can be updated as well)

test/unittest/integration_tests/integration_tests.cpp
EunjuYang, line 1: It is a simple question. Why is the name of the test `integration_tests`? I couldn't catch its meaning by the comment only.

nntrainer/layers/loss/kld_loss_layer.cpp
DonghakPark: We need to revisit our discussion on the formula! Thank you for your insightful feedback.

test/unittest/integration_tests/integration_tests.cpp
DonghakPark: In NNTrainer, there are limitations when conducting layer unit tests using only layers. 

For example, in the case of cross, errors need to be output if it is not softmax or sigmoid
however, with the current structure for parameter testing, it is difficult to reach that specific scenario.

Additionally, issues may arise during repeated iterations, unlike batch normalization (BN) or other layers where each layer performs one forward/backward pass without any difference. 

Therefore, this script was created to conduct layer-focused tests at the network level or larger units. 

**The name "integration test" reflects these intentions, but if you have a more suitable alternative, feel free to suggest it.**

EunjuYang: Oh I see. Thank you for the detailed explanation :)

test/unittest/integration_tests/integration_tests.cpp
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Donghak Park <donghak.park@samsung.com>
+ *
+ * @file unittest_loss_crossentropy.cpp
```
baek2sm, line 5: file name :)

DonghakPark: Nice Catch!! Thank you for Review

nntrainer/layers/loss/kld_loss_layer.cpp
DonghakPark: I have completed apply your review.
Thank you for providing additional check with pytorch example.

nntrainer/layers/loss/kld_loss_layer.cpp
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * 1. Output = label / predicted
+     * 2. Output = output * label
+     * 3. Output = log(output)
+     * 4. Output = sum(output)
+     */
+    label.divide(predicted, temp);
+    temp.multiply_i(label);
+    temp.apply<float>(logf, temp);
+    output.fill(temp.sum({0, 1, 2, 3}));
```
EunjuYang, line 41: The code seems like `(P * (P / Q)).log()` which is different from `P * (P/Q).log()`.
Please change the code like
```suggestion
    /**
     * 1. Output = label / predicted
     * 2. Output = log(Output)
     * 3. Output = Output * label
     * 4. Output = sum(output)
     */
    label.divide(predicted, temp);
    temp.apply<float>(logf, temp);
    temp.multiply_i(label);
    output.fill(temp.sum({0, 1, 2, 3}));
```

nntrainer/layers/loss/kld_loss_layer.cpp
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * d/dQ = -P/Q
+   */
+  nntrainer::Tensor &predicted = context.getInput(SINGLE_INOUT_IDX); // Q
+  nntrainer::Tensor &label = context.getLabel(SINGLE_INOUT_IDX);     // P
+  nntrainer::Tensor &deriv = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
+
+  label.multiply_i(-1.0f);
+  label.divide_i(predicted);
+  deriv.fill(label);
```
EunjuYang, line 55: I think it is just a coding style, but it is just a simple suggestion from my side:
```suggestion
  label.divide(predicted, deriv);
```

EunjuYang (commented):
Thank you for considering my previous comment.
Please check some suggestions below :)

nntrainer/layers/loss/kld_loss_layer.cpp
```diff
@@ -0,0 +1,58 @@
...
+    nntrainer::Tensor temp; // temp output
+    /**
+     * 1. Output = label / predicted
+     * 2. Output = output * label
+     * 3. Output = log(output)
+     * 4. Output = sum(output)
+     */
+    label.divide(predicted, temp);
+    temp.multiply_i(label);
+    temp.apply<float>(logf, temp);
```
skykongkong8, line 40: is this function only supported on FP32?

nntrainer/layers/loss/kld_loss_layer.cpp
DonghakPark: !! Thank you for check

nntrainer/layers/loss/kld_loss_layer.cpp
DonghakPark: Good Point, i will apply your review

nntrainer/layers/loss/kld_loss_layer.cpp
DonghakPark: For now, Yes
but i will apply FP16 in next PR

EunjuYang (approved):
LGTM!

skykongkong8 (approved):
I think it's good to go!

jijoongmoon (approved):
LGTM

PR #2759, opened 2024-10-17
[GPU/OpenCL] Broadcasting support added for GPU Addition kernel.

niket-agarwal:
Performing addition where dimensions of InputA and InputB vary.
Added broadcasting support only where number of batches vary and other dimensions are same for both inputs.
Number of batch of InputB must be 1.
Output of add_i_cl(A,B) is stored in A inplace.

Self evaluation:

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
@@ -200,38 +200,46 @@ void multiplyCl(Tensor &input, float const &value) {
   }
 }
 
-void add_i_cl(Tensor const &input, Tensor &result) {
+void add_i_cl(Tensor &result, Tensor const &input) {
 
-  CREATE_IF_EMPTY_DIMS(result, result.getDim());
-
-  NNTR_THROW_IF(result.getData() == nullptr, std::invalid_argument)
-    << result.getName() << " is not allocated";
   NNTR_THROW_IF(input.getData() == nullptr, std::invalid_argument)
     << input.getName() << " is not allocated";
+  NNTR_THROW_IF(result.getData() == nullptr, std::invalid_argument)
+    << result.getName() << " is not allocated";
 
-  if (input.getDim() != result.getDim()) {
-    throw std::invalid_argument(
-      "Error: Dimensions does not match for addition");
-  }
+  // Broadcasting done for the case where batch size vary for both inputs
+  // If batch size vary, batch size of input must be 1
+  if ((result.getDim() == input.getDim()) ||
+      (result.getDim() != input.getDim() && input.batch() == 1 &&
+       result.channel() == input.channel() &&
+       result.height() == input.height() && result.width() == input.width())) {
 
-  if (input.getDataType() == ml::train::TensorDim::DataType::FP32) {
-    unsigned int size = input.size();
-    const float *data = input.getData();
-    float *rdata = result.getData();
+    if (result.getDataType() == ml::train::TensorDim::DataType::FP32) {
+      unsigned int size_res = result.size();
+      unsigned int size_input = input.size();
+      float *data_res = result.getData();
+      const float *data_input = input.getData();
 
-    addition_cl(data, rdata, size);
+      addition_cl(data_input, data_res, size_input, size_res);
 
-  } else if (input.getDataType() == ml::train::TensorDim::DataType::FP16) {
+    } else if (result.getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
-    unsigned int size = input.size();
-    const _FP16 *data = input.getData<_FP16>();
-    _FP16 *rdata = result.getData<_FP16>();
+      unsigned int size_res = result.size();
+      unsigned int size_input = input.size();
+      _FP16 *data_res = result.getData<_FP16>();
+      const _FP16 *data_input = input.getData<_FP16>();
 
-    addition_cl(data, rdata, size);
+      addition_cl(data_input, data_res, size_input, size_res);
 
 #else
-    throw std::invalid_argument("Error: enable-fp16 is not enabled");
+      throw std::invalid_argument("Error: enable-fp16 is not enabled");
 #endif
+    }
+  }
+
+  else {
+    throw std::invalid_argument(
+      "Error: Broadcasting not supported for these dimensions!");
   }
 }
 
--- a/nntrainer/tensor/cl_operations/blas_kernel_interface.h
+++ b/nntrainer/tensor/cl_operations/blas_kernel_interface.h
@@ -64,11 +64,10 @@ void multiplyCl(Tensor &input, float const &value);
 
 /**
  * @brief Process data and dimensions for add operation
- * @param[in] input Tensor
  * @param[in] result Tensor
- * @param[in] RunLayerContext reference
+ * @param[in] input Tensor
  */
-void add_i_cl(Tensor const &input, Tensor &result);
+void add_i_cl(Tensor &result, Tensor const &input);
 
 /**
  * @brief Process data and dimensions for transpose operation
--- a/nntrainer/tensor/cl_operations/blas_kernel_strings.h
+++ b/nntrainer/tensor/cl_operations/blas_kernel_strings.h
@@ -118,11 +118,11 @@ static const std::string sgemm_cl_transAB_kernel_ =
     })";
 
 static const std::string addition_cl_kernel_ =
-  R"(__kernel void addition_cl(__global const float* input, __global float* output, const unsigned int size) {
+  R"(__kernel void addition_cl(const __global float* input, __global float* output, unsigned int size_input, unsigned int size_res) {
     #pragma printf_support
     size_t idx = get_global_id(0);
-    if (idx < size) {
-        output[idx] = output[idx] + input[idx];
+    if (idx < size_res) {
+        output[idx] = output[idx] + input[idx % size_input];
     }
   })";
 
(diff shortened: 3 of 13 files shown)


nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
```diff
@@ -200,38 +200,47 @@ void multiplyCl(Tensor &input, float const &value) {
   }
 }
 
-void add_i_cl(Tensor const &input, Tensor &result) {
+void add_i_cl(Tensor &inputA, Tensor const &inputB) {
 
-  CREATE_IF_EMPTY_DIMS(result, result.getDim());
+  CREATE_IF_EMPTY_DIMS(inputA, inputA.getDim());
```
baek2sm, line 205: I have a question. The result tensor before modification could be empty, but is it possible that inputA tensor is empty in the current modification?

niket-agarwal: Right! It shouldn't be. I'll modify it, thanks.

myungjoo:
PTAL: @baek2sm @skykongkong8 @EunjuYang

nntrainer/tensor/cl_operations/blas_kernel_interface.h
```diff
@@ -64,11 +64,10 @@ void multiplyCl(Tensor &input, float const &value);
...
 /**
  * @brief Process data and dimensions for add operation
- * @param[in] input Tensor
- * @param[in] result Tensor
- * @param[in] RunLayerContext reference
+ * @param[in] inputA Tensor
+ * @param[in] inputB Tensor
  */
-void add_i_cl(Tensor const &input, Tensor &result);
+void add_i_cl(Tensor &inputA, Tensor const &inputB);
```
djeong20, line 70: I'm assuming inputA is a result, and inputB is an input. is this correct?

nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
```diff
@@ -200,38 +200,47 @@ void multiplyCl(Tensor &input, float const &value) {
...
-
-  if (input.getDataType() == ml::train::TensorDim::DataType::FP32) {
-    unsigned int size = input.size();
-    const float *data = input.getData();
-    float *rdata = result.getData();
-
-    addition_cl(data, rdata, size);
-
-  } else if (input.getDataType() == ml::train::TensorDim::DataType::FP16) {
+void add_i_cl(Tensor &inputA, Tensor const &inputB) {
```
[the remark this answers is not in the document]
baek2sm: I feel the same way. I guess the intention of this naming was to express `"a = a + b"` more clearly. (before `"result(b) = input(a) + result(b)"`). So the inputA and inputB parameter names are also good, but I think they seem to be inconsistent with other functions.

nntrainer/tensor/cl_operations/blas_kernel_interface.h
niket-agarwal: Both are inputs, the addition is taking inplace and inputA is returned as output

nntrainer/tensor/cl_operations/blas_kernel_strings.h
```diff
@@ -106,11 +106,11 @@ static const std::string sgemm_cl_transAB_kernel_ =
...
 
 static const std::string addition_cl_kernel_ =
-  R"(__kernel void addition_cl(__global const float* input, __global float* output, const unsigned int size) {
+  R"(__kernel void addition_cl(const __global float* input, __global float* output, unsigned int size_input, unsigned int size_res) {
     #pragma printf_support
     size_t idx = get_global_id(0);
-    if (idx < size) {
-        output[idx] = output[idx] + input[idx];
+    if (idx < size_res) {
+        output[idx] = output[idx] + input[idx % size_input];
```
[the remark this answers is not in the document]
niket-agarwal: Yes correct.

EunjuYang, line 113: You are assuming the input have multiple batch with same dimension... is it correct? `input[idx % size_input]` was written for the purpose (broadcast input with multi-batch). Do I correctly understand? If so, what about adding some condition to check whether `size_input` is multiple of `size_res`?

nntrainer/tensor/cl_operations/blas_kernels.h
```diff
@@ -135,10 +136,11 @@ void sgemm_cl(bool TransA, bool TransB, const __fp16 *A, const __fp16 *B,
...
  * @param[in] input fp16 * for input
  * @param[in] res fp16 * for result/output
- * @param[in] size number of elements in input vector
- * @param[in] context RunLayerContext reference
+ * @param[in] size_input number of elements in input vector
+ * @param[in] size_res number of elements in result vector
  */
-void addition_cl(const __fp16 *input, __fp16 *res, unsigned int size);
+void addition_cl(const __fp16 *input, __fp16 *res, unsigned int size_input,
+                 unsigned int size_res);
```
EunjuYang, line 143:
```suggestion
void addition_cl(const _FP16 *input, _FP16 *res, unsigned int size_input,
                 unsigned int size_res);
```

nntrainer/tensor/cl_operations/blas_kernels_fp16.cpp
```diff
@@ -291,7 +291,8 @@ void sgemm_cl(bool TransA, bool TransB, const __fp16 *A, const __fp16 *B,
   } while (false);
 }
 
-void addition_cl(const __fp16 *input, __fp16 *res, unsigned int size) {
+void addition_cl(const __fp16 *input, __fp16 *res, unsigned int size_input,
+                 unsigned int size_res) {
```
EunjuYang, line 295: What about updating `__fp16` into `_FP16` ?
Also, since header file is defining this function for fp16 only under the specific condition, it should be updated following the same policy.

```suggestion
#ifdef ENABLE_FP16
void addition_cl(const _FP16 *input, _FP16 *res, unsigned int size_input,
                 unsigned int size_res) {
```

nntrainer/tensor/cl_operations/blas_kernels_fp16.cpp
```diff
@@ -333,12 +335,18 @@ void addition_cl(const __fp16 *input, __fp16 *res, unsigned int size) {
...
 
-    const int work_groups_count[3] = {(int)size, 1, 1};
+    result =
+      kernel_addition_fp16_ptr->SetKernelArguments(3, &size_res, sizeof(int));
+    if (!result) {
+      break;
+    }
+
+    const int work_groups_count[3] = {(int)size_res, 1, 1};
     const int work_group_size[3] = {32, 32, 1}; // test-value
```
EunjuYang, line 350: This is simple question.
How can we optimize this work_group_size? Do you have any idea or plan on this?

EunjuYang (commented):
Thank you for your contribution.
I left my opinions and questions bellow. Please check it. Thanks.

nntrainer/tensor/cl_operations/blas_kernel_interface.h
```diff
@@ -64,11 +64,11 @@ void multiplyCl(Tensor &input, float const &value);
 
 /**
  * @brief Process data and dimensions for add operation
- * @param[in] input Tensor
- * @param[in] result Tensor
+ * @param[in] inputA Tensor
```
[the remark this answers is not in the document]
niket-agarwal: okay will add that.

nntrainer/tensor/cl_operations/blas_kernel_strings.h
niket-agarwal: It will always be a multiple. Broadcasting is done only for the case where inputA has batch n, and inputB has batch to be unity. This fulfils our requirement of running addition for MHA usecase.

nntrainer/tensor/cl_operations/blas_kernels_fp16.cpp
niket-agarwal: This file will only be compiled for fp16, hence I think the ifdef check is not required. Updated __fp16 into _FP16 though.

nntrainer/tensor/cl_operations/blas_kernels_fp16.cpp
niket-agarwal: We are thinking on this. Will let you know. Thanks for reviewing.

nntrainer/tensor/cl_operations/blas_kernels_fp16.cpp
EunjuYang: Oh, I was confused it. It's my bad. Sorry for the confusion 🤣

nntrainer/tensor/cl_operations/blas_kernel_strings.h
EunjuYang: Oh, I see. I found the code checking the condition in `add_i_cl`.  However, what about adding codes to check its condition in `addition_cl` as well? The `addition_cl` might be used in other blas_cl_interface functions later. Isn't it? Please let me know if I missed something. BTW, there was an error in my previous comment. `size_res` should be multiple of `size_input` right? Sorry for the confusion.

EunjuYang (commented):
In the current main branch, `unittest_layers_addition_cl` is disabled.

https://github.com/nnstreamer/nntrainer/blob/a80a6e10bad3bdc689b85dd911393ea9b79a4a41/test/jni/Android.mk#L479

Please enable it and check the unittest_layers pass all unittest cases with `./tools/android_test.sh`. If you find *.nnlayergolden missing errors, please update `unittest_layers.tar.gz` as well. (c.f. https://github.com/nnstreamer/nntrainer/pull/2798)

EunjuYang:
Also, for you added new feature of broadcasting support for addition kernel, what about adding unit test for the case?

nntrainer/tensor/cl_operations/blas_kernel_strings.h
niket-agarwal: Yes right, size_res should be multiple of size_input, and this is already checked in add_i_cl. The addition_cl will be called via add_i_cl as far as I can understand. Could you clarify why another condition is needed in addition_cl?

EunjuYang: If it is called by the interface only, it doesn't need to be. If it is not, please consider it.

djeong20 (approved):
Overall, LGTM! 
As @EunjuYang mentioned, please add test cases for the newly added feature.

EunjuYang (approved):
LGTM!

PR #2761, opened 2024-10-21
[ GPU ] move `initBlaseClKernels()` to registerer

EunjuYang:
- With the current version code, a user should be care about the timing to call `initBlasClKernels()`. However, it is not desirable, causing some unexpected mistakes.
- Instead of that, this draft moves the `initBlasClKernels()` into registerer, which is called once in Global().`
- This PR is related to #2723 

Self evaluation:

### DIFF
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -55,6 +55,7 @@ static void add_default_object(ClContext &cc) {
 
 static void registerer(ClContext &cc) noexcept {
   try {
+    cc.initBlasClKernels();
     add_default_object(cc);
   } catch (std::exception &e) {
     ml_loge("cl_context: registering layers failed!!, reason: %s", e.what());
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -139,7 +139,6 @@ createLayerNode(const ml::train::LayerType &type,
 #ifdef ENABLE_OPENCL
   if (compute_engine == ml::train::LayerComputeEngine::GPU) {
     auto &cc = nntrainer::ClContext::Global();
-    cc.initBlasClKernels();
     return createLayerNode(cc.createObject<nntrainer::Layer>(type), properties,
                            compute_engine);
   }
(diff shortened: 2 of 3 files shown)


nntrainer/cl_context.cpp
```diff
@@ -55,6 +55,7 @@ static void add_default_object(ClContext &cc) {
 
 static void registerer(ClContext &cc) noexcept {
   try {
+    cc.initBlasClKernels();
```
s-debadri, line 58: There was some undefined behaviour when I tried to add `initBlasClKernels` here. It seemed that some `static` variables where initialized whereas some were not.
If you are also getting the same behaviour, just FYI this might be related to [static initialization order fiasco](https://isocpp.org/wiki/faq/ctors#static-init-order).

EunjuYang: Hmm.. I see! It is what you mentioned before.
How did you detect the **undefined behaviour**? I cannot check it only on the unittest (I passed all unittest for layer and blas_kernel_cl)

EunjuYang: Can you please tell me how you detect the issue? I'm going to check it as well.

s-debadri: While running `unittest_layers`, some `LayerSemantics` tests were failing with message like some key was missing. While checking logcat I saw that there was this log:
`cl_context: factory has registered with key: , int_key: 1`

Instead this should be:
`cl_context: factory has registered with key: fully_connected, int_key: 1`

This `fully_connected` is a static member `type` of `fc_layer_cl.h`. Every layer header files have them which is used during layer register.

EunjuYang: Oh, I see, but it is not reproduced in my end 😥
The log I can check:
```
07-05 06:48:48.708 16758 16758 D nntrainer: cl_context: factory has registered with key: fully_connected, int_key: 1
07-05 06:48:48.708 16758 16758 D nntrainer: cl_context: factory has registered with key: swiglu, int_key: 30
07-05 06:48:48.708 16758 16758 D nntrainer: cl_context: factory has registered with key: reshape, int_key: 505
07-05 06:48:48.708 16758 16758 D nntrainer: cl_context: factory has registered with key: rmsnorm, int_key: 503
07-05 06:48:48.708 16758 16758 D nntrainer: cl_context: factory has registered with key: concat, int_key: 8
```
If so, you think it's a dynamic error.... Could you test this PR as well?

s-debadri: This works fine. I re-checked from my end as well. Same issue was not reproduced.

s-debadri (approved):
LGTM

myungjoo (approved):

baek2sm (approved):
LGTM

djeong20 (approved):
LGTM!

PR #2764, opened 2024-10-21
[Unittest] re-activate the deactivated unittest

baek2sm:
I have reactivated the deactivated unittest except for the test cases that cause errors.

- enable mol_attention_layer unittest
- enable reduce_mean_layer unittest
- enable unittest_models except for test cases that cause errors

**Self evaluation:**

### DIFF
--- a/test/unittest/layers/meson.build
+++ b/test/unittest/layers/meson.build
@@ -50,6 +50,7 @@ test_target = [
   'unittest_layers_multiout.cpp',
   'unittest_layers_rnn.cpp',
   'unittest_layers_rnncell.cpp',
+  'unittest_layers_reduce_mean.cpp',
   'unittest_layers_lstm.cpp',
   'unittest_layers_lstmcell.cpp',
   'unittest_layers_gru.cpp',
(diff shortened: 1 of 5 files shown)


DonghakPark (approved):
LGTM!! Nice Catch

test/unittest/layers/meson.build
```diff
@@ -62,7 +63,7 @@ test_target = [
   'unittest_layers_attention.cpp',
   'unittest_layers_dropout.cpp',
   'unittest_layers_reshape.cpp',
-  # 'unittest_layers_mol_attention.cpp',
+  'unittest_layers_mol_attention.cpp',
```
skykongkong8, line 66: Is mol_attention a molecular attention?
If so, does keras support it? (for gtest purpose)

SeoHyungjun (approved):
LGTM

jijoongmoon (approved):
LGTM

test/unittest/layers/meson.build
baek2sm: @skykongkong8 I don't think we need to support mol_attention layer anymore. This layer has been included in the main branch, so the unittest is also necessary for now. However, it would be nice to discuss removing this layer together.

PR #2769, opened 2024-10-24
[ unittest/modelfile ] Fix backbone_n_15 @open sesame 10/17 17:36

skykongkong8:
- The original purpose of this TC is to make sure the model is not compiled when input layer is not clarified.
- As far as I am concerned, without any information about input layer, the model should fail to compile, and since it is not compiled, should not support initialization.
- Related commit hashes:
original purpose : 68e7dab5ed2f99e8b380ec8a5f599d76952b7fc3
wrong fix : e1c5915fba7d64db66062d8b62a722d6c1b653ad


jijoongmoon (approved):
LGTM

DonghakPark (approved):
LGTM!

lhs8928 (approved):
LGTM

djeong20 (approved):

PR #2773, opened 2024-10-25
[ GPU ] Update android_test.sh to enable unittests for GPU  @open sesame 11/04 09:08

EunjuYang:
- This commit updates `android_test.sh` to enable unittests for GPU
    - create `nntrainer_opencl_kernels` folder
    - add some command lines to push nngolden data to android
- unittest_layers.tar.gz is updated to add nnlayer golden data  for GPU layers.
- This commit updates `how-to-use-testcases.md` file, including guide for running test cases for OpenCL on Android.

**Note**
- This PR doesn't resolve all unit-test but aims to build a common environment to run GPU-related unittests.
- There are still 12 unittest failures. Please refer to #2774 

**Self evaluation**:


test/jni/Android.mk
```diff
@@ -458,7 +458,6 @@ LOCAL_SRC_FILES := \
 	 ../unittest/layers/unittest_layers_flatten.cpp \
 	 ../unittest/layers/unittest_layers_activation.cpp \
 	 ../unittest/layers/unittest_layers_addition.cpp \
-	 ../unittest/layers/unittest_layers_addition_cl.cpp \
```
DonghakPark, line 461: Question) is there any reason to remove this from unittest_layers build ?

EunjuYang: This `../unittest/layers/unittest_layers_addition_cl.cpp` should be added after #2752.
If not, NDK build returns error, for addition layer for cl kernel is not implemented in the current upstream. I wanted to make upstream run as it is.
By the way... it would be better to change this PR [Wait for #2752] and remove this change.

DonghakPark: i see, thank you for detail explain 

```By the way... it would be better to change this PR [Wait for https://github.com/nnstreamer/nntrainer/pull/2752], while removing this change.```  
Yes i would be good!! or i think you can simply rebase #2752 or cherry pick some commit's ( if there are some dependency)

rebase sample command
``` git pull --rebase upstream refs/pull/2752/head: ```

Comment (EunjuYang) [reply test/jni/Android.mk:461]: Sounds bood! I will rebase #2752. Thank you ☺

Comment (EunjuYang) [reply test/jni/Android.mk:461]: @DonghakPark 
I tested rebasing on #2752, but the PR itself is not well rebased on the main; It makes this PR jammed.
Instead of rebasing on it, I updated the `Android.mk` to leave line you mentioned as commented out. Could you please check the update?
If #2752 is merged or it resolves the rebasing issue, I will update this PR as well.

File: tools/android_test.sh [hunk 2]
```diff
@@ -42,5 +43,8 @@ fi
 # The steps are as follows.
 
 # $ meson build [flags...]
-# $ cd build
-# $ adb push res/ /data/local/tmp/nntr_android_test
+# meson build will unzip golden data for the unit tests
+cd ../../../
+meson build
```
djeong20, line 48: it would be nice to add some basic flags to get the proper golden data (e.g., `-Denable-fp16=true`, `-Denable-neon=true`, `-Denable-opencl=true`)

djeong20 (approved):

baek2sm:
Don't we need a script to make golden data as well?

test/jni/Android.mk
DonghakPark: If so, it seems to be good. It may reduce inconvenience

tools/android_test.sh
DonghakPark: Good point!! for more information we use below flags for build android_package

```
-Dplatform=android 
-Dopenblas-num-threads=1 
-Denable-tflite-interpreter=false 
-Denable-tflite-backbone=false 
-Denable-fp16=true 
-Denable-neon=true 
-Domp-num-threads=1 
-Denable-opencl=true 
-Dhgemm-experimental-kernel=false
```

EunjuYang: Thank you for the comments, all! It would be better to apply the same options for the android build. I will apply it 😀

EunjuYang:
> Don't we need a script to make golden data as well?

Thank you for your comment. The script for the golden data was already updated in `gen_lsyer_tests.py` in previous PRs. However, they were tested by creating the golden data manually and pushing them to test environment.
To make it easier and a common test tool, this PR updates `unittest_layers.tar.gz` and `android_test.sh`. If I missed some steps, please let me know.

FYI, I attach some links for the golden layer generation parts additionally added in this update :

https://github.com/nnstreamer/nntrainer/blob/63f987506ae8779bb31bf596c665b5fe99a95748/test/input_gen/gen_layer_tests.py#L927

https://github.com/nnstreamer/nntrainer/blob/63f987506ae8779bb31bf596c665b5fe99a95748/test/input_gen/gen_layer_tests.py#L917

https://github.com/nnstreamer/nntrainer/blob/63f987506ae8779bb31bf596c665b5fe99a95748/test/input_gen/gen_layer_tests.py#L885

tools/android_test.sh
EunjuYang: @DonghakPark I applied all the options you suggested except for the android option, for`meson.build` doesn't create resource for unittest with `-Dplatofrm=android` option

Please refer to 
https://github.com/nnstreamer/nntrainer/blob/63f987506ae8779bb31bf596c665b5fe99a95748/meson.build#L439

DonghakPark: Thank you for considering my opinion.
I think we need to ensure that testing is enabled in the future, even for cases of Android as well. (As there are increasingly more examples utilizing Android.)
I will create an issue for this matter. 

- #2775

baek2sm (approved):
LGTM

tools/android_test.sh
```diff
@@ -42,5 +43,15 @@ fi
 # The steps are as follows.
 
 # $ meson build [flags...]
-# $ cd build
-# $ adb push res/ /data/local/tmp/nntr_android_test
+# meson build will unzip golden data for the unit tests
+cd ../../../
+meson build -Dopenblas-num-threads=1  \
```
djeong20, line 48: there might be cases where the `build` directory already exists. we should handle that case!

EunjuYang: Great point! I will update the script. Thanks!

djeong20 (approved):
LGTM!

skykongkong8 (approved):
This resolves many inconveniences that I have been working on. Thanks!

DonghakPark (approved):

PR #2779, opened 2024-10-31
[Tensor] Update with scale factor support for quantized data types

djeong20:
This pull request aims to enhance the functionality of the Tensor class by enabling it to handle scale factors specifically designed for quantized data types. This patch ensures tensor class accurately represents and processes quantized data while maintaining original features. This update will provide developers with more flexibility when working with quantized models

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor_base.h
+++ b/nntrainer/tensor/tensor_base.h
@@ -189,6 +189,24 @@ public:
   virtual void *getData(size_t idx) const = 0;
 
   /**
+   * @copydoc Tensor::getScale()
+   */
+  virtual void *getScale() const {
+    throw std::invalid_argument(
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/tensor_base.h
```diff
@@ -568,7 +586,7 @@ class TensorBase {
    * @brief     Get size of current tensor
    * @retval    unsigned int size of the current tensor
    */
-  size_t size() const { return dim.getDataLen(); }
+  virtual size_t size() const { return dim.getDataLen(); }
```
djeong20, line 589: size() is now a virtual function since it will be overridden by a quantized tensor that utilizes data packing.

jijoongmoon (approved):
LGTM

PR #2780, opened 2024-11-01
Add AdamW `Optimizer` @open sesame 12/04 16:19

jangdan:
This commit implements the AdamW algorithm in the form of an `Optimizer`.

Some (potential) further tasks include:
- Merging the implementations of Adam and AdamW, so that `AdamW` inherits from `Adam`
- Adding more unit tests

### DIFF
--- a/api/ccapi/include/optimizer.h
+++ b/api/ccapi/include/optimizer.h
@@ -34,6 +34,7 @@ class LearningRateScheduler;
  */
 enum OptimizerType {
   ADAM = ML_TRAIN_OPTIMIZER_TYPE_ADAM,      /** adam */
+  ADAMW = ML_TRAIN_OPTIMIZER_TYPE_ADAMW,    /** AdamW */
   SGD = ML_TRAIN_OPTIMIZER_TYPE_SGD,        /** sgd */
   UNKNOWN = ML_TRAIN_OPTIMIZER_TYPE_UNKNOWN /** unknown */
 };
@@ -133,6 +134,14 @@ Adam(const std::vector<std::string> &properties = {}) {
 inline std::unique_ptr<Optimizer>
 SGD(const std::vector<std::string> &properties = {}) {
   return createOptimizer(OptimizerType::SGD, properties);
+}
+
+/**
+ * @brief Helper function to create AdamW Optimizer
+ */
+inline std::unique_ptr<Optimizer>
+AdamW(const std::vector<std::string> &properties = {}) {
+  return createOptimizer(OptimizerType::ADAMW, properties);
 }
 
 } // namespace optimizer
--- a/api/nntrainer-api-common.h
+++ b/api/nntrainer-api-common.h
@@ -91,7 +91,8 @@ typedef enum {
  * @since_tizen 6.0
  */
 typedef enum {
-  ML_TRAIN_OPTIMIZER_TYPE_ADAM = 0, /**< Adam Optimizer */
+  ML_TRAIN_OPTIMIZER_TYPE_ADAM = 0,  /**< Adam Optimizer */
+  ML_TRAIN_OPTIMIZER_TYPE_ADAMW = 2, /**< AdamW Optimizer */
   ML_TRAIN_OPTIMIZER_TYPE_SGD = 1, /**< Stochastic Gradient Descent Optimizer */
   ML_TRAIN_OPTIMIZER_TYPE_UNKNOWN = 999 /**< Unknown Optimizer */
 } ml_train_optimizer_type_e;
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -28,6 +28,7 @@
 #include <util_func.h>
 
 #include <adam.h>
+#include <adamw.h>
 #include <sgd.h>
 
 #include <activation_layer.h>
@@ -235,6 +236,8 @@ static void add_default_object(AppContext &ac) {
   ac.registerFactory(nntrainer::createOptimizer<SGD>, SGD::type, OptType::SGD);
   ac.registerFactory(nntrainer::createOptimizer<Adam>, Adam::type,
                      OptType::ADAM);
+  ac.registerFactory(nntrainer::createOptimizer<AdamW>, AdamW::type,
+                     OptType::ADAMW);
   ac.registerFactory(AppContext::unknownFactory<nntrainer::Optimizer>,
                      "unknown", OptType::UNKNOWN);
 
--- a/nntrainer/optimizers/adamw.cpp
+++ b/nntrainer/optimizers/adamw.cpp
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Daniel Jang <minhyukjang@snu.ac.kr>
+ *
+ * @file   adamw.cpp
+ * @date   3 November 2024
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author Parichay Kapoor <pk.kapoor@samsung.com>
+ * @author Daniel Jang <minhyukjang@snu.ac.kr>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is the AdamW Optimizer.
+ */
+
+#include <cmath>
+#include <fstream>
+
+#include <adamw.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <node_exporter.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+AdamW::AdamW() : adam_props(PropsB1(), PropsB2(), PropsEpsilon(), TorchRef()) {
+  /** default properties */
+  auto &[b1, b2, eps, torch_ref] = adam_props;
+  b1.set(0.9f);
+  b2.set(0.999f);
+  eps.set(1.0e-7f);
+  torch_ref.set(false);
+}
+
+AdamW::~AdamW() {}
+
+enum AdamParams { wm, wv };
+
+std::vector<TensorDim> AdamW::getOptimizerVariableDim(const TensorDim &dim) {
+  return {dim, dim};
+}
+
+void AdamW::exportTo(Exporter &exporter,
+                     const ml::train::ExportMethods &method) const {
+  exporter.saveResult(adam_props, method, this);
+  Optimizer::exportTo(exporter, method);
+}
+
+void AdamW::setProperty(const std::vector<std::string> &values) {
+  auto left = loadProperties(values, adam_props);
+  Optimizer::setProperty(left);
+}
+
+void AdamW::applyGradient(RunOptimizerContext &context) {
+  Tensor &x_grad = context.getGradient();
+
+  auto &beta1 = std::get<PropsB1>(adam_props).get();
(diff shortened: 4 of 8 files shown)


jijoongmoon:
This commit needs the commit body. Could you write details in the commit body?

myungjoo:
Please add a short description about the commit in the commit message body. This helps when we need to browse git logs to debug.

In case you are not familiar with git, you can edit the commit message:
```
$ git commit --amend
```

EunjuYang:
Thank you for your contribution 😀
NNTrainer follows a specific coding convention (i.e., specific clang format is provided), so you need to apply it to pass the CI. Please refer to [[this](https://github.com/nnstreamer/nntrainer/blob/main/docs/coding-convention.md)].

nntrainer/optimizers/adamw.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <adamw.h>
+#include <nntrainer_error.h>
+#include <nntrainer_log.h>
+#include <node_exporter.h>
+#include <util_func.h>
+
+namespace nntrainer {
+
+AdamW::AdamW() : adam_props(PropsB1(), PropsB2(), PropsEpsilon(), TorchRef()) {
```
DonghakPark, line 26: Adam case, when the torch_ref flag is provided, it handles separate operations to ensure that the results are as similar as possible to PyTorch.

I am curious whether Adamw also considered this case in adding torch_ref or simply did so to have the same properties as Adam. It seems like torch_ref has no meaning for AdamW.

```.cpp
  if (torch_ref) {
    Tensor denom = wv.apply<float>(sqrtFloat<float>);
    denom.divide_i(sqrtFloat(biasCorrection2));
    denom.add_i(epsilon);
    wm.divide(denom, x_grad);

    context.applyGradient(context.getLearningRate() / biasCorrection1);

  } else {
    std::function<double(double)> sqrtEps = [epsilon](double f) {
      return 1 / (sqrtDouble(f) + epsilon);
    };
```

test/ccapi/unittest_ccapi.cpp
```diff
@@ -147,6 +147,7 @@ TEST(ccapi_optimizer, construct_01_n) {
 TEST(ccapi_optimizer, construct_02_p) {
   EXPECT_NO_THROW(ml::train::optimizer::Adam());
   EXPECT_NO_THROW(ml::train::optimizer::SGD());
+  EXPECT_NO_THROW(ml::train::optimizer::AdamW());
```
DonghakPark, line 150: Have you ever conducted any actual training tests other than ccapi construct test?
when i tested on MNIST Application Example i got  ```segmentation fault (core dumped)``` error

tested on this code
```.cpp
  std::unique_ptr<ml::train::Model> model;
  model = createModel(ml::train::ModelType::NEURAL_NET);
  model->load(config, ml::train::ModelFormat::MODEL_FORMAT_INI_WITH_BIN);
  model->compile();
  model->initialize();
  model->setDataset(ml::train::DatasetModeType::MODE_TRAIN, dataset_train);
  model->setDataset(ml::train::DatasetModeType::MODE_VALID, dataset_val);
  model->train();
```
with this ```ini``` file
```.ini
# Network Section : Network

[Model]
Type = NeuralNetwork                # Network Type : Regression, KNN, NeuralNetwork
Epochs = 1500                       # Epochs
Loss = cross                        # Loss function : mse (mean squared error)
                                    # cross ( for cross entropy )
# Save_Path = "mnist_model.bin"     # model path to save / read
batch_size = 32                     # batch size

[Optimizer]
Type = adamw
beta1 = 0.9                         # beta 1 for adam
beta2 = 0.999                       # beta 2 for adam
epsilon = 1e-7                      # epsilon for adam

# Layer Section : Name
[inputlayer]
Type = input
Input_Shape = 1:28:28

# Layer Section : Name
[conv2d_c1_layer]
Type = conv2d
input_layers = inputlayer
kernel_size = 5,5
bias_initializer=zeros
Activation=sigmoid
weight_initializer = xavier_uniform
filters = 6
stride = 1,1
padding = 0,0

[pooling2d_p1]
Type=pooling2d
input_layers = conv2d_c1_layer
pool_size = 2,2
stride =2,2
padding = 0,0
pooling = average

[conv2d_c2_layer]
Type = conv2d
input_layers = pooling2d_p1
kernel_size = 5,5
bias_initializer=zeros
Activation=sigmoid
weight_initializer = xavier_uniform
filters = 12
stride = 1,1
padding = 0,0

[pooling2d_p2]
Type=pooling2d
input_layers = conv2d_c2_layer
pool_size = 2,2
stride =2,2
padding = 0,0
pooling = average

[flatten]
Type=flatten
input_layers = pooling2d_p2

[outputlayer]
Type = fully_connected
input_layers=flatten
Unit = 10                           # Output Layer Dimension ( = Weight Width )
weight_initializer = xavier_uniform
bias_initializer = zeros
Activation = softmax                # activation : sigmoid, softmax
```

api/ccapi/include/optimizer.h
```diff
@@ -35,6 +35,7 @@ class LearningRateScheduler;
 enum OptimizerType {
   ADAM = ML_TRAIN_OPTIMIZER_TYPE_ADAM,      /** adam */
   SGD = ML_TRAIN_OPTIMIZER_TYPE_SGD,        /** sgd */
+  ADAMW = ML_TRAIN_OPTIMIZER_TYPE_ADAMW,    /** AdamW */
   UNKNOWN = ML_TRAIN_OPTIMIZER_TYPE_UNKNOWN /** unknown */
 };
```
SeoHyungjun, line 40: Wouldn't it be better to write in alphabetical order unless there's a specific reason?

```suggestion
enum OptimizerType {
  ADAM = ML_TRAIN_OPTIMIZER_TYPE_ADAM,      /** adam */
  ADAMW = ML_TRAIN_OPTIMIZER_TYPE_ADAMW,    /** AdamW */
  SGD = ML_TRAIN_OPTIMIZER_TYPE_SGD,        /** sgd */
  UNKNOWN = ML_TRAIN_OPTIMIZER_TYPE_UNKNOWN /** unknown */
};
```

api/nntrainer-api-common.h
```diff
@@ -88,6 +88,7 @@ typedef enum {
 typedef enum {
   ML_TRAIN_OPTIMIZER_TYPE_ADAM = 0, /**< Adam Optimizer */
   ML_TRAIN_OPTIMIZER_TYPE_SGD = 1, /**< Stochastic Gradient Descent Optimizer */
+  ML_TRAIN_OPTIMIZER_TYPE_ADAMW = 2, /**< AdamW Optimizer */
```
SeoHyungjun, line 91: Ditto

nntrainer/app_context.cpp
```diff
@@ -29,6 +29,7 @@
 
 #include <adam.h>
 #include <sgd.h>
+#include <adamw.h>
```
SeoHyungjun, line 32: In our project, the option "SortIncludes" is enabled in the ".clang-format" file.
Therefore, the order of headers must be sorted unconditionally.

```suggestion
#include <adam.h>
#include <adamw.h>
#include <sgd.h>
```

SeoHyungjun (commented):
Thank you for your contribution!

Please check out the comments below. :art:

baek2sm (changes requested):
As others have pointed out, you need adding a commit body and making modifications to follow the clang format. Then you can pass the CI.

nntrainer/optimizers/adamw.cpp
jangdan: Indeed, the `"torch_ref"` property is not used (at all) in `AdamW`. The reasoning behind keeping it (and also retaining the identifier `adam_props`, to address @baek2sm's [comment](https://github.com/nnstreamer/nntrainer/pull/2780#discussion_r1825411607)) was that I thought it would be a very good idea to, in a later PR, have the `AdamW` class inherit from the `Adam` class, to reduce code duplication, as they share a lot of boilerplate code. Sorry for not making this clear.

Any thoughts on this? If this is not the way to go, I'd happily make the necessary changes.

FYI: PyTorch's [AdamW class](https://pytorch.org/docs/stable/_modules/torch/optim/adamw.html#AdamW) <ins>doesn't</ins> seem to inherit from its [Adam class](https://pytorch.org/docs/stable/_modules/torch/optim/adam.html#Adam).

api/ccapi/include/optimizer.h
jangdan: Good idea. It wasn't particularly clear that that was the case, with only two entries.

api/nntrainer-api-common.h
jangdan: Done.

jangdan: Done.

jangdan: Done.

nntrainer/app_context.cpp
jangdan: Fixed, with clang-format.

test/ccapi/unittest_ccapi.cpp
jangdan: The code should now actually work.

Thanks.

baek2sm (approved):

nntrainer/optimizers/adamw.cpp
```diff
@@ -0,0 +1,89 @@
...
+  Optimizer::exportTo(exporter, method);
+}
+
+void AdamW::setProperty(const std::vector<std::string> &values) {
+  auto left = loadProperties(values, adam_props);
+  Optimizer::setProperty(left);
+}
+
+void AdamW::applyGradient(RunOptimizerContext &context) {
+  Tensor &x_grad = context.getGradient();
```
EunjuYang, line 55: Code to process mixed precision is missing. Please refer to 
https://github.com/nnstreamer/nntrainer/blob/0d74457e18b97fa8d26702ec7663d6f6e9510ac2/nntrainer/optimizers/adam.cpp#L75-L84

EunjuYang (commented):
Thanks for your contribution to implement `AdamW` optimizer.
`AdamW` is a varient of `Adam` that applies weight decay.
However, I couldn't find the part for this weight decay. Could you explain if I missed something? Please check the comments below. 
----------------
I missed baek2sm's comment above. I will leave the updated review again.

nntrainer/optimizers/adamw.cpp
jijoongmoon: yes. this needs to be considered. Please check it.

nntrainer/optimizers/adamw.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  wv.divide_i(sqrtFloat(biasCorrection2));
+  std::function<double(double)> sqrtEps = [epsilon](double f) {
+    return 1 / (sqrtDouble(f) + epsilon);
+  };
+  Tensor &term = wv;
+  term.apply<float>(sqrtEps);
+  term.divide_i(biasCorrection1);
+  term.multiply_i(wm);
+  x_grad.add_i(term);
```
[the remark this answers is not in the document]
jangdan: I think you are correct. Thanks for the catch!

nntrainer/optimizers/adamw.cpp
jangdan: This, too, has been addressed.

jijoongmoon (approved):
LGTM

PR #2781, opened 2024-11-01
[Layer] Add conv2dTranspose layer @open sesame 12/04 16:19

sugsugsug:
This PR adds the [ConvTranspose2d](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html) class.

The implementation includes forwarding, calcDerivative, and calcGradient which use im2col for optimization.

In im2col_transpose function, first I caculated output matrix shape. While iterating through all locations in the matrix I put each input value in right position corresponding to the filter value to multiply with.

col2im was also implemented the same way as im2col. 

I followed the way in conv2d.cpp combining each function to make forwarding, calcDerivative, and calcGradient.

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -45,7 +45,9 @@ enum LayerType {
   LAYER_SWIGLU = ML_TRAIN_LAYER_TYPE_SWIGLU, /**< Swiglu Layer type */
   LAYER_BN = ML_TRAIN_LAYER_TYPE_BN, /**< Batch Normalization Layer type */
   LAYER_CONV2D = ML_TRAIN_LAYER_TYPE_CONV2D, /**< Convolution 2D Layer type */
-  LAYER_CONV2D_TRANSPOSE = ML_TRAIN_LAYER_TYPE_CONV2D_TRANSPOSE, /**< Convolution 2D Transpose Layer type */
+  LAYER_CONV2D_TRANSPOSE =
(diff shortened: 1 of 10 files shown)


jijoongmoon:
Please check the CI reviews (pre & post) and then we can review the code.

EunjuYang:
Thank you for your contribution 😀 
NNTrainer requires a contributor to follow a specific coding convention with signing off. Please refer to this contribution guide [[link](https://github.com/nnstreamer/nntrainer/blob/main/docs/contributing.md)]. For the details, please check the CI review.

DonghakPark (commented):
As @jijoongmoon @EunjuYang mentioned
Please review the following checklist items for compliance.
1. Check Clang-format 
2. Fill commit body (more detail explain)
3. check signed-off 
4. doxygen tag

baek2sm (changes requested):
Currently, there is an unittest-error in the compiler part. Specifically, the `tf_schema.fbs` script is causing problem. I don't think you need to work on the tf_schema part now, so it seems okay to exclude the `tf_schema.fbs` script from the commit.

If you exclude that script from the commit and perform a rebase (`nntrainer-api-common.h` file), then i think, you can pass the CI.

DonghakPark:
Please rebase upstream-main then it will pass all ci 
```
git rebase upstream/main
```

DonghakPark (approved):

baek2sm (approved):

PR #2784, opened 2024-11-04
[gpu/enhance] Shared memory implementation for Blas kernels

s-debadri:
Implementation of shared memory for Blas kernels.

Updates made:

- Added data pointer for `clCreateBuffer` calls using wrapper constructor
- Removed all `WriteData` calls
- Used `CL_MEM_USE_HOST_PTR` flag
- Used `MapBuffer` and `UnMapBuffer` wrappers to ensure cache consistency

> Note: GPU can cache buffer data on device memory which might not get flushed to host after kernel execution. To enforce this `MapBuffer` or `ReadData` wrapper should be used.

**Self evaluation:**


baek2sm (approved):
LGTM.

I have a simple question. All of these codes are running within a do-while statement, but why is it necessary to use a do-while loop even though the current while condition is always false?

s-debadri:
> LGTM.
> 
> I have a simple question. All of these codes are running within a do-while statement, but why is it necessary to use a do-while loop even though the current while condition is always false?

We want the code to return at the point of failure and don't proceed any further. Hence, to avoid multiple nested if-else conditions we are using a do-while loop to make the code readable and clean.

PR #2785, opened 2024-11-04
[ GPU/OpenCL ] Split kernel registration from forwarding method @open sesame 12/02 09:39

EunjuYang:
- This draft is a suggestion for #2723
- This draft splits kernel registration from forwarding function.
- This draft make kernelPtr as static member of layer to avoid redundant kernel registration.
- This draft contains example update for `concat_cl` , `reshape_cl`, and `fc_layer_cl` only.

Self evaluation:

### DIFF
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -33,30 +33,37 @@ std::once_flag global_cl_context_init_flag;
 
 static void add_default_object(ClContext &cc) {
 
-  FullyConnectedLayerCl::registerClKernels();
-  cc.registerFactory(nntrainer::createLayer<FullyConnectedLayerCl>,
-                     FullyConnectedLayerCl::type,
-                     ml::train::LayerType::LAYER_FC);
+  if (FullyConnectedLayerCl::registerClKernels()) {
+    cc.registerFactory(nntrainer::createLayer<FullyConnectedLayerCl>,
+                       FullyConnectedLayerCl::type,
+                       ml::train::LayerType::LAYER_FC);
+  }
 
   // cc.registerFactory(nntrainer::createLayer<AdditionLayerCL>,
   //                    AdditionLayerCL::type,
   //                    ml::train::LayerType::LAYER_ADDITION);
 
-  // cc.registerFactory(nntrainer::createLayer<SwiGLULayerCl>,
-  // SwiGLULayerCl::type,
-  //  ml::train::LayerType::LAYER_SWIGLU);
+  // @todo swiglulayercl also needs to be updated.
+  cc.registerFactory(nntrainer::createLayer<SwiGLULayerCl>,
+  SwiGLULayerCl::type,
+   ml::train::LayerType::LAYER_SWIGLU);
 
-  ReshapeLayerCl::registerClKernels();
-  cc.registerFactory(nntrainer::createLayer<ReshapeLayerCl>,
-                     ReshapeLayerCl::type, ml::train::LayerType::LAYER_RESHAPE);
+  if (ReshapeLayerCl::registerClKernels()) {
+    cc.registerFactory(nntrainer::createLayer<ReshapeLayerCl>,
+                       ReshapeLayerCl::type,
+                       ml::train::LayerType::LAYER_RESHAPE);
+  }
 
-  // cc.registerFactory(nntrainer::createLayer<RMSNormLayerCl>,
-  //  RMSNormLayerCl::type, ml::train::LayerType::LAYER_RMSNORM);
+  // @todo rmsnormlayercl also needs to be updated.
+  cc.registerFactory(nntrainer::createLayer<RMSNormLayerCl>,
+   RMSNormLayerCl::type, ml::train::LayerType::LAYER_RMSNORM);
 
-  ConcatLayerCl::registerClKernels();
-  cc.registerFactory(nntrainer::createLayer<ConcatLayerCl>, ConcatLayerCl::type,
-                     ml::train::LayerType::LAYER_CONCAT);
+  if (ConcatLayerCl::registerClKernels()) {
+    cc.registerFactory(nntrainer::createLayer<ConcatLayerCl>,
+                       ConcatLayerCl::type, ml::train::LayerType::LAYER_CONCAT);
+  }
 
+  // @todo transposlayercl also needs to be updated.
   cc.registerFactory(nntrainer::createLayer<TransposeLayerCl>,
                      TransposeLayerCl::type,
                      ml::train::LayerType::LAYER_TRANSPOSE);
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -33,6 +33,7 @@ std::once_flag global_cl_context_init_flag;
 
 static void add_default_object(ClContext &cc) {
 
+  FullyConnectedLayerCl::registerClKernels();
   cc.registerFactory(nntrainer::createLayer<FullyConnectedLayerCl>,
                      FullyConnectedLayerCl::type,
                      ml::train::LayerType::LAYER_FC);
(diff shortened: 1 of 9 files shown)


nntrainer/cl_context.cpp
```diff
@@ -32,25 +32,33 @@ std::once_flag global_cl_context_init_flag;
 
 static void add_default_object(ClContext &cc) {
 
-  cc.registerFactory(nntrainer::createLayer<FullyConnectedLayerCl>,
-                     FullyConnectedLayerCl::type,
-                     ml::train::LayerType::LAYER_FC);
+  if (FullyConnectedLayerCl::registerClKernels()) {
```
djeong20, line 35: currently, registerClKernels() always return true. are there going to be cases where it returns false?

nntrainer/layers/cl_layers/concat_cl.cpp
```diff
@@ -233,12 +233,54 @@ std::string concat_cl_axis1_kernel_ =
...
+    << "OpenCL Error: Fail to register concat_cl_axis2_fp16 kernel";
+  layer_kernel_ptrs.emplace_back(kernel_concat_ptr);
+
+  kernel_concat_ptr = cl_context_ref.registerClKernel(
+    concat_cl_axis3_kernel_fp16_, "concat_cl_axis3_fp16");
+  NNTR_THROW_IF(!kernel_concat_ptr, std::runtime_error)
+    << "OpenCL Error: Fail to register concat_cl_axis3_fp16 kernel";
+  layer_kernel_ptrs.emplace_back(kernel_concat_ptr);
+
+  return true;
```
djeong20, line 281: quick question! can't ConcatLayerCl::registerClKernels() be called twice?
assume it is called for a second time, would it throw a runtime error or return true?

nntrainer/layers/cl_layers/fc_layer_cl.h
```diff
@@ -101,12 +101,19 @@ class FullyConnectedLayerCl : public LayerImpl {
    */
   void setProperty(const std::vector<std::string> &values) override;
 
+  static bool registerClKernels() { return true; };
```
djeong20, line 104: Is it NYI or always true?

nntrainer/layers/cl_layers/fc_layer_cl.h
```diff
@@ -101,12 +101,19 @@ class FullyConnectedLayerCl : public LayerImpl {
...
+
   inline static const std::string type = "fully_connected";
 
 private:
   std::tuple<props::Unit>
     fc_props; /**< fc layer properties : unit - number of output neurons */
   std::array<unsigned int, 2> weight_idx; /**< indices of the weights */
+
+  const static int num_layer_kernels = 0; /** < number of layer kernels */
```
djeong20, line 113: it looks like num_layer_kernels is unused. what is it for?

nntrainer/layers/cl_layers/concat_cl.cpp
```diff
@@ -233,12 +233,54 @@ std::string concat_cl_axis1_kernel_ =
...
+    concat_cl_axis2_kernel_fp16_, "concat_cl_axis2_fp16");
+  NNTR_THROW_IF(!kernel_concat_ptr, std::runtime_error)
+    << "OpenCL Error: Fail to register concat_cl_axis2_fp16 kernel";
+  layer_kernel_ptrs.emplace_back(kernel_concat_ptr);
+
+  kernel_concat_ptr = cl_context_ref.registerClKernel(
+    concat_cl_axis3_kernel_fp16_, "concat_cl_axis3_fp16");
+  NNTR_THROW_IF(!kernel_concat_ptr, std::runtime_error)
+    << "OpenCL Error: Fail to register concat_cl_axis3_fp16 kernel";
+  layer_kernel_ptrs.emplace_back(kernel_concat_ptr);
```
djeong20, line 279: wouldn't this throw an error when fp16 is not enabled?

nntrainer/layers/cl_layers/fc_layer_cl.h
EunjuYang: There's no need to register kernel for `fc_layer_cl`, since it works with blas kernel only (cl blas kernels are registered separately). Thus, it is always true.

nntrainer/layers/cl_layers/fc_layer_cl.h
EunjuYang: oh it is the previous version lol! Thank you for your comment. I tried to add kerenel registration check through num_layer_kernels == len() thing... but I decided to remove it. I'm gonna update the code. Thank you.

nntrainer/cl_context.cpp
EunjuYang: Yeap. In current version, I just throw exception. But, I think it would be better to update the code to return false and register it if it fails to register (to allow nntrainer to retry it later).

nntrainer/layers/cl_layers/concat_cl.cpp
EunjuYang: Yes you're right. I just write the code based on the current kernel version. I will add a separate commit to update this. Thanks!

nntrainer/layers/cl_layers/concat_cl.cpp
EunjuYang: I assumed it is only called in `add_default_object` only, which is called in `registerer` ; the `registerer` is called only once. However, it seems better to check it. I will update it. 

```
ClContext &ClContext::Global() {
  static ClContext instance;

  // initializing commandqueue and context
  bool result = instance.clInit();

  if (!result) {
    ml_loge("cl_context: opencl command queue creation failed");
  }

  /// in g++ there is a bug that hangs up if caller throws,
  /// so registerer is noexcept although it'd better not
  /// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70298
  std::call_once(global_cl_context_init_flag, registerer, std::ref(instance));
  return instance;
}
```

EunjuYang: I added a condition to check it in the `registerClKernels()` as well.
https://github.com/nnstreamer/nntrainer/blob/6dc250f57117375d5d55a981976f59cdf4e8540c/nntrainer/layers/cl_layers/reshape_cl.cpp#L55

nntrainer/cl_context.cpp
EunjuYang: I updated the code to return true only when all the kernels are registered successfully.
Please refer to: 
https://github.com/nnstreamer/nntrainer/blob/6dc250f57117375d5d55a981976f59cdf4e8540c/nntrainer/layers/cl_layers/concat_cl.cpp#L302

EunjuYang:
📢 An additional commit to fix fp16-related issue in `concat_cl` is included : da185961adba6753be0f398f4e6107b1e744c211
🩹 @djeong20 's recommendation is applied. Please review it and give me feedbacks. Thanks.

nntrainer/layers/cl_layers/concat_cl.cpp
```diff
@@ -233,12 +233,82 @@ std::string concat_cl_axis1_kernel_ =
...
 
 static constexpr size_t SINGLE_INOUT_IDX = 0;
 static constexpr size_t INPUT_IDX_1 = 0;
 static constexpr size_t INPUT_IDX_2 = 1;
 
+bool ConcatLayerCl::registerClKernels() {
+
+  // check if already registered
+  if (!layer_kernel_ptrs.empty())
+    return true;
```
djeong20, line 246: quick question. if kernels are already registered, why does it return true?
Is there a case where kernels are registered but not the layer?

djeong20: I apologize for the confusion. 
My question is: If `layer_kernel_ptrs` is already filled, doesn't that mean the corresponding layer is also registered?
Assuming the layer is registered, wouldn't returning true in this condition register the layer more than once in the cl_context.cpp?

https://github.com/nnstreamer/nntrainer/blob/6dc250f57117375d5d55a981976f59cdf4e8540c/nntrainer/cl_context.cpp#L60-L63

EunjuYang: Oh I see, you mean it should be false, not true. I'm gonna update this.

djeong20 (approved):
Thank you for the hard work! 👍

nntrainer/layers/cl_layers/concat_cl.cpp
```diff
@@ -812,40 +866,36 @@ void ConcatLayerCl::concat_cl_axis2_fp16(
...
-    }
+    const auto &kernel_concat_ptr =
+      layer_kernel_ptrs[Kernels::CONCAT_CL_AXIS2_FP16];
 
-    int dim = int(input1_batch_size * input1_width * input1_height *
-                  (input1_channels + input2_channels));
+    int dim = int(input1_batch_size * input1_channels * input1_width *
+                  (input1_height + input2_height));
 
     opencl::Buffer inputA(cl_context_ref.context_inst_,
```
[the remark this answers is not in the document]
EunjuYang: Great ! It is required to support heterogeneous computing units with a common interface. I will create an issue and handle it later. Thank you for your opinion :)

jijoongmoon (approved):
LGTM

PR #2789, opened 2024-11-07
[Graph] rename variable & function names for clarity @open sesame 11/07 21:30

baek2sm:
currently, same term 'InPlace' is used with different meanings by script.

it could refer to a boolean type value indicating whether in-place operation is supported, or it could refer an enum type value representing the in-place type(type: None, Restriction, Non-Restriction).

To clarify this distinction, variables and functions related to the in-place type have been renamed from 'InPlace' to 'InPlaceType'.

Similarly, both functions for getting/setting whether it supports inplace (boolean) and those for getting/setting the type of inplace (enum) use the same name, "executeInPlace".

To clarify this, I've changed the names as follows..
- set the in-place flag (boolean) : 'executeInPlace' -> 'setInPlace'.
- get the in-place flag (boolean) : 'executeInPlace' -> 'getInplace'.
- set the type of in-place (enum) : 'executeInPlace' -> 'setInPlaceType'
- get the type of in-place (enum) : 'executeInPlace' -> 'getInPlaceType.'

**Self evaluation:**

### DIFF
--- a/Applications/Custom/mae_loss.cpp
+++ b/Applications/Custom/mae_loss.cpp
@@ -27,7 +27,7 @@ void MaeLossLayer::forwarding(nntrainer::RunLayerContext &context,
   nntrainer::Tensor &predicted = context.getInput(SINGLE_INOUT_IDX);
   nntrainer::Tensor &output = context.getOutput(SINGLE_INOUT_IDX);
 
-  if (!context.executeInPlace())
+  if (!context.getInPlace())
     output.fill(predicted);
 }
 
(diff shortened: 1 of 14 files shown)


djeong20 (approved):
Nice work! I think this change is good to not get confused.

SeoHyungjun (approved):
Currently, InPlaceType is divided into NONE, RESTRICTING, and NON_RESTRICTING.

In my opinion, I think DISABLED is clearer than NONE.
How about using DISABLED, RESTRICTING, and NON_RESTRICTING?

```cpp
enum class InPlaceType{
  DISABLED,       /**< layer is not inplace */
  RESTRICTING,    /**< layer is in-place and does place restriction on layers
                    ahead of it to be in-place */
  NON_RESTRICTING /**< layer is in-place and does NOT place restriction on the
                    layers ahead of it to be in-place */
};
```

baek2sm:
@SeoHyungjun Hmm.. since this enum is  "TYPE" of inplace, wouldn't it be more appropriate to use "none" rather than "disabled"?

PR #2794, opened 2024-11-13
[Graph] change of inplace-type setting method

baek2sm:
The method of setting the inplace-type has been redefined.

The reason why inplace processing becomes complicated is that since a `multi-out layer` shares output variables, so it needs to be considered whether or not inplace can be performed.

Actually, the layers that can perform inplace even after the `multi-out layer` are only `no-operation layers(no-op layers)`. These `no-op layers` include `identity`, `reshape`, and `flatten layers`.

For other layers, even if they support inplace, they cannot perform inplace when there is a `multi-out layer` in front of them.

Note that because `no-op layers` connected with `multi-out layer` share memory with the `multi-out layer`, so they have the same properties as the `multi-out layer`. This property is expressed as `RESTRICTING` in our script.

Based on these definitions, I've redesigned the method of setting inplace type.

1. By default, initialize the inplace type for each layer. If supportInPlace is true, it will be initialized as `NON_RESTRICTING`; otherwise, it will be initialized as `NONE`.
2. However, not all layers are initialized like this. For `multi-out layers` or `no-op layers`, if supportInPlace is true, they will be initialized as `RESTRICTING` types(However, the `no-op layer` will be changed to a `NON_RESTRICTING` type if that is not connected with the `multi-out layer` within the `network_graph.cpp`).
3. After initialization, confirm the input connections from the `network_graph.cpp` to determine the final inplace type. It's clearer to see the source code for this part.

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -642,7 +642,6 @@ void NetworkGraph::addLayer(std::shared_ptr<LayerNode> layer) {
 
 InPlaceType
 NetworkGraph::canExecuteInPlace(const std::shared_ptr<LayerNode> &lnode) {
-
   if (!lnode->supportInPlace()) {
     return InPlaceType::NONE;
   }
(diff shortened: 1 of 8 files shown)


nntrainer/graph/network_graph.cpp
```diff
@@ -796,7 +713,8 @@ setInplaceSharedMemoryConfigByLayer(const std::shared_ptr<LayerNode> &lnode,
     shared_var = true;
     shared_grad = true;
   }
-  /** @todo for addition layer, variables are not shared but gradients are */
+  /** @todo for addition layer, variables are not shared but gradients are
```
jijoongmoon, line 716: Could you check the indentation in clang-format?

nntrainer/layers/reshape_layer.h
```diff
@@ -78,6 +78,17 @@ class ReshapeLayer : public Layer {
    */
   bool supportInPlace() const override { return true; }
 
+  /**
+   * @brief Initialize the in-place type of the layer
+   * @return InPlaceType
+   */
+  InPlaceType initializeInPlaceType() override {
```
EunjuYang, line 85: Simple question!
It seems flatten_layer, which inherits ReshapeLayer, overrides the initializeInPlaceType() with the same code.
If you think it is always same, what about defining this function as `final` instead of `override` to reduce code replication?

nntrainer/graph/network_graph.cpp
baek2sm: This comment needed to be deleted. I've removed it. Thank you!

nntrainer/layers/layer_node.cpp
```diff
@@ -930,6 +930,20 @@ bool LayerNode::supportInPlace() const {
...
+ */
+InPlaceType LayerNode::initializeInPlaceType() {
+  inplace_type = layer->initializeInPlaceType();
+  return inplace_type;
+}
+
+/**
+ * @brief
+ *
+ */
```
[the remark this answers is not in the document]
baek2sm: @EunjuYang I removed it. Thanks!

nntrainer/layers/reshape_layer.h
baek2sm: @EunjuYang  That's a good point! I set that function of the reshape layer to be `final` instead of `override`, and removed that function from the flatten layer. Thank you.

EunjuYang (approved):
LGTM!

djeong20 (approved):
LGTM

PR #2795, opened 2024-11-14
[Graph] rename variable and remove non-working codes @open sesame 11/14 19:02

baek2sm:
- The variables with the same meaning are written as `in_place` or `is_inplace` by the script, so I unified it to use the term "is_inplace".

- Some layer's finalize function includes code that determines whether or not to support in-place depending on the tensor type. However, this code does not work. The reason it seems like this code is working is because there is a same purpose of code at the top of the `canExecuteInPlace` function within the `network_graph.cpp` and that code works.

It is meaningless to determine whether or not to support in-place within the `finalize` function because the `canExecuteInPlace` function, which decides how InPlace will behave, is called before the `finalize` function.

The canExecuteInPlace function is called during `compile` while the finalize function is called during `initialize` after `compile`. Therefore, setting supportInplace inside the finalize function does not work.

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/loss/loss_layer.cpp
+++ b/nntrainer/layers/loss/loss_layer.cpp
@@ -25,10 +25,6 @@ void LossLayer::finalize(InitLayerContext &context) {
                     nntrainer::TensorDataTypeInfo>::from_string("FP32"));
 
   context.setOutputDimensions(output_dim);
-
-  is_inplace = true;
-  if (context.getActivationDataType() != ml::train::TensorDim::DataType::FP32)
-    is_inplace = false;
 }
 
 void LossLayer::updateLoss(RunLayerContext &context, const Tensor &l) {
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -37,8 +37,8 @@ public:
    */
   template <typename T = float>
   ActiFunc(ActivationType at = ActivationType::ACT_NONE,
-           bool in_place_ = true) :
-    in_place(in_place_) {
+           bool is_inplace_ = true) :
+    is_inplace(is_inplace_) {
     setActiFunc<T>(at);
   }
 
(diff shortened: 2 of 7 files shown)


skykongkong8 (approved):
You have a point. Great work!

nntrainer/layers/loss/loss_layer.cpp
```diff
@@ -25,10 +25,6 @@ void LossLayer::finalize(InitLayerContext &context) {
                     nntrainer::TensorDataTypeInfo>::from_string("FP32"));
 
   context.setOutputDimensions(output_dim);
-
-  is_inplace = true;
-  if (context.getActivationDataType() != ml::train::TensorDim::DataType::FP32)
-    is_inplace = false;
```
EunjuYang, line 31: I understood that this code is meaningless (based on your description). Then, which part do this action? I cannot find the part replacing this code to support mixed precision. Could you explain this?

baek2sm: @EunjuYang Currently, processing for the input layer is being done within the network graph, but processing for the loss layer has not been implemented yet. The loss layer cannot be processed in the same way as the input layer. I'll try to set whether or not the loss layer is in-place and upload it as a separate PR. Very good point. Thanks a lot!

EunjuYang: Oh I see! Thank you the detailed explanation !👍

EunjuYang (approved):
LGTM. 
BTW, according this PR, the mixed-precision related feature in loss_layer was not properly applied as intended. It might need to be validated. @DonghakPark , could you see this?

baek2sm:
@EunjuYang As we discussed offline, I don't think this PR will affect the validity of mixed-precision because the default value for inplace in the loss layer was set to false. These `non-working codes` were intended to set in-place to false when using mixed precision. The only advantage of that code working as intended is that it allows for inplace operations on the loss layer when not using mixed precision (which would generally save just 4-bytes memory consumption). Of course, we need to update to support inplace operation in loss layer.

jijoongmoon (approved):
LGTM

PR #2796, opened 2024-11-14
[WIP][Wait for #2794, #2795][Graph] add inplace setting through layer property

baek2sm:
- now you can set in-place flag through layer property of tensor
operation layers.
- rename "initializeInPlaceType" func to "initializeInPlace". now
"is_inplace" property is set in that function, too.
- in some layers, support_backwarding flag may be changed by the
in-place setting.

**Self evaluation:**
1. Build test:   [X]Passed [ ]Failed [ ]Skipped
2. Run test:     [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Seungbaek Hong <sb92.hong@samsung.com>

PR #2797, opened 2024-11-15
[Wait for #2796][Graph] add inplace direction setting through layer property

baek2sm:
added inplace direction setting for binary tensor operation layer.

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -768,6 +768,9 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
         if (lnode->getType() == IdentityLayer::type) {
           s.variable_spec.reference_name = inputs[i]->getName();
           s.variable_spec.dim.setFormat(inputs[i]->getDim().getFormat());
+        } else if (lnode->getInPlaceDirection() == InPlaceDirection::RIGHT) {
+          s.variable_spec.reference_name = inputs[1]->getName();
+          s.variable_spec.dim.setFormat(inputs[1]->getDim().getFormat());
         } else {
           s.variable_spec.reference_name = inputs[0]->getName();
           s.variable_spec.dim.setFormat(inputs[0]->getDim().getFormat());
@@ -779,6 +782,9 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
         if (lnode->getType() == IdentityLayer::type) {
           s.gradient_spec->reference_name = inputs[i]->getGradientName();
           s.gradient_spec->dim.setFormat(inputs[i]->getDim().getFormat());
+        } else if (lnode->getInPlaceDirection() == InPlaceDirection::RIGHT) {
+          s.gradient_spec->reference_name = inputs[1]->getGradientName();
+          s.gradient_spec->dim.setFormat(inputs[1]->getDim().getFormat());
         } else {
           s.gradient_spec->reference_name = inputs[0]->getGradientName();
           s.gradient_spec->dim.setFormat(inputs[0]->getDim().getFormat());
@@ -924,6 +930,8 @@ NetworkGraph::refinalizeContext(const std::shared_ptr<LayerNode> &lnode,
           TensorSpecV2::RequestType::READ_ONLY_VIEW;
         if (lnode->getType() == IdentityLayer::type) {
           s.variable_spec.reference_name = inputs[i]->getName();
+        } else if (lnode->getInPlaceDirection() == InPlaceDirection::RIGHT) {
+          s.variable_spec.reference_name = inputs[1]->getName();
         } else {
           s.variable_spec.reference_name = inputs[0]->getName();
         }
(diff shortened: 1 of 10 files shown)


nntrainer/layers/divide_layer.h
```diff
@@ -77,6 +78,27 @@ class DivideLayer : public BinaryOperationLayer {
...
+  /**
+   * @brief Get the inplace direction for the tensor operation layer
+   *
+   * @return InPlaceDirection
+   */
+  /**
+   * @brief Get the inplace direction for the tensor operation layer
+   *
+   * @return InPlaceDirection
+   */
```
djeong20, line 90:
```suggestion
```
duplicated tag

nntrainer/layers/layer_devel.h
```diff
@@ -53,6 +53,16 @@ enum class InPlaceType {
                     layers ahead of it to be in-place */
 };
 
+/**
+ * @brief Emum class for the direction of inplace
+ *
+ */
+enum class InPlaceDirection {
+  NONE,  /**< default value */
```
djeong20, line 61: what does `NONE` indicate? does it mean it is out of place?

djeong20 (approved):

nntrainer/layers/layer_devel.h
baek2sm: @djeong20 The default value is set to `NONE` to use it as a default when the inplace option is turned off or when the inplace option is turned on but it's a unary operation, then there's no need to express direction using `LEFT` or `RIGHT`. However, if expressing it this way causes confusion, it would be fine to leave only `LEFT` and `RIGHT`.

djeong20: sorry, I'm a bit confused. when `is_inplace` is false, what should `getInPlaceDirection()` return?

baek2sm: @djeong20 if it is called when `is_inplace` is false, then it will return `NONE`.

EunjuYang: > @djeong20 The default value is set to `NONE` to use it as a default `when the inplace option is turned off` or `when the inplace option is turned on but it's a unary operation`, then there's no need to express direction using `LEFT` or `RIGHT`. However, if expressing it this way causes confusion, it would be fine to leave only `LEFT` and `RIGHT`.

If so, what about adding the explanation to the comment?

```suggestion
  NONE,  /**< default value; the inplace option is turned off */
```

nntrainer/layers/layer_devel.h
```diff
@@ -255,6 +265,15 @@ class Layer {
    */
   virtual bool supportInPlace() const { return is_inplace; }
 
+  /**
+   * @brief Get the inplace direction for the tensor operation layer
+   *
+   * @return InPlaceDirection
+   */
+  virtual InPlaceDirection getInPlaceDirection() {
+    return InPlaceDirection::NONE;
```
EunjuYang, line 274: Even though it is virtual, is it okay to return `InPlaceDirection::NONE` regardless of the `is_inplace`?
According to the meaning of the `InPlaceDirection::NONE`, it holds only when `is_inplace = false`.

EunjuYang (commented):
This pull request adds an in-place direction for right, which is set by an option. Can you provide an example for setting the in-place direction as right? I couldn't understand its advantages or examples yet.

nntrainer/layers/layer_devel.h
baek2sm: For a clearer explanation, the default value of the inplace direction is `None`, and it will be set to 'LEFT' or 'RIGHT' only when the type of the operation layer is `binary` and the `is_inplace` setting is `true`.

nntrainer/layers/layer_devel.h
baek2sm: As discussed offline, the `InPlaceDirection` option has meaning only when the layer type is `binary`. Therefore, even if the `is_inplace` option is set to `true`, it may return `NONE` if the layer type is not `binary`.

nntrainer/layers/layer_devel.h
EunjuYang: Oh I see. Thank you for the detail.

EunjuYang (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -768,6 +768,9 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
         if (lnode->getType() == IdentityLayer::type) {
           s.variable_spec.reference_name = inputs[i]->getName();
           s.variable_spec.dim.setFormat(inputs[i]->getDim().getFormat());
+        } else if (lnode->getInPlaceDirection() == InPlaceDirection::RIGHT) {
+          s.variable_spec.reference_name = inputs[1]->getName();
```
jijoongmoon, line 772: It seems strainge. Could you check if it is '1'?? or i??

nntrainer/layers/divide_layer.h
baek2sm: @djeong20 thanks!

jijoongmoon (approved):
LGTM! One more thing I recommand is adding more unittest about this in-place property for binary operation!

PR #2801, opened 2024-11-20
[layer] add pow operation layer @open sesame 12/06 20:13

baek2sm:
added a pow operation layer.

there was an example of a pow layer in the custom layers, so I modified the key value of the custom pow layer to "custom_pow" in order to avoid duplication of layer key value.

**Self evaluation:**


nntrainer/layers/pow_layer.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace nntrainer {
+
+void PowLayer::finalize(InitLayerContext &context) {
+  context.setOutputDimensions({context.getInputDimensions()[0]});
+}
+
+void PowLayer::forwarding_operation(const Tensor &input, Tensor &hidden) {
+  float exp = std::get<props::Exponent>(pow_props).get();
+  input.pow(exp, hidden);
+}
```
skykongkong8, line 32: Quick question! this isn't really has to do with PR but:
Is there any difference with multiplying itself if the given exponent is set to 2? ( or inv_sqrt function <-> pow with -0.5 exponent )
I find almost every `pow` or `pow_i` function usage in the nntrainer use 2.0, 0.5, or -0.5 as an exponent, and this is handled with naive loop Tensor member function.

skykongkong8 (approved):

nntrainer/layers/meson.build
```diff
@@ -9,6 +9,7 @@ layer_sources = [
   'subtract_layer.cpp',
   'multiply_layer.cpp',
   'divide_layer.cpp',
+  'pow_layer.cpp',
```
EunjuYang, line 12: This is a simple suggestion (not directly related to this PR though).
What about using prefix like 'op_' to operation layers?
It might reduce the confusion on the layer types (e.g., add_layer & addition_layer).

nntrainer/layers/pow_layer.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <layer_context.h>
+
+namespace nntrainer {
+
+void PowLayer::finalize(InitLayerContext &context) {
+  context.setOutputDimensions({context.getInputDimensions()[0]});
+}
+
+void PowLayer::forwarding_operation(const Tensor &input, Tensor &hidden) {
+  float exp = std::get<props::Exponent>(pow_props).get();
```
EunjuYang, line 30: Can we change the name `exp` ? (exp function is defined in <cmath>)

EunjuYang (approved):
LGTM

myungjoo:
```unittest_models_v2.tar.gz```
1. You have two .tar.gz files with the same name.
2. I don't see any scripts that access these .tar.gz files.

nntrainer/layers/meson.build
baek2sm: That sounds good idea. I'll reflect it in later pr. thanks!

nntrainer/layers/pow_layer.cpp
baek2sm: I've modified it(`exp` -> `exponent`). Thanks a lot!

baek2sm:
@myungjoo
1. The inclusion of the duplicate file was my mistake. I've removed one, thanks!
2. That file contains the golden data for unittest and it's extracted with `test/unittest/meson.build` script(It's already included, not this commit). Then `test/unittest/unittest_models.cpp` script utilizes these golden data files. Thanks a lot.

myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #2803, opened 2024-11-25
[Bugfix] Revert mixed precision case inplace set false

DonghakPark:
Currently, when using mixed precision mode with in-place set to true, the ninja test fails.

I think this failure is due to setting the in-place default value as true for all cases where it shouldn't have been.

Since loss layers and input layers haven't yet implemented non-FP32 operations, I've modified them so that they now specify in-place computations as false.

**Self evaluation:**


skykongkong8 (approved):

jijoongmoon (approved):
LGTM

PR #2805, opened 2024-11-26
[Trivial] Change Android Log type

DonghakPark:
Change Android Log type from ml_loge -> ml_logi

This log is not an error but simply indicates an iteration. 
However, it is being recorded as an error, so we will make changes accordingly.

**Self evaluation:**


skykongkong8 (approved):
nice catch!

jijoongmoon (approved):
LGTM

PR #2810, opened 2024-11-29
[ GPU ] split kernel registration from forwarding function in `addition_layer_cl` and `transpose_cl`

### LINKED ISSUE #2723: [ GPU ] GPU Kernel creation time

EunjuYang:
[ GPU/OpenCL ] change  `addition_layer_cl` and `transpose_cl` to inherit LayerImplCl

- This commit updates  `addition_layer_cl` and `transpose_cl`.cpp/h to inherit LayerImplCl.
- This commit implements registerClKernels() of  `addition_layer_cl` and `transpose_cl` layer.
- This commit update cl_context.cpp (applying  `addition_layer_cl` and `transpose_cl`'s update)
- This PR is continuation of https://github.com/nnstreamer/nntrainer/pull/2785
- This PR can close #2723 

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/cl_layers/addition_layer_cl.h
+++ b/nntrainer/layers/cl_layers/addition_layer_cl.h
@@ -18,6 +18,7 @@
 #include <cl_context.h>
 #include <common_properties.h>
 #include <layer_devel.h>
+#include <layer_impl_cl.h>
 
 namespace nntrainer {
 
@@ -25,12 +26,12 @@ namespace nntrainer {
  * @class   AdditionLayerCL
  * @brief   Addition Layer
  */
-class AdditionLayerCL : public Layer {
+class AdditionLayerCL : public LayerImplCl {
 public:
   /**
    * @brief     Constructor of Addition Layer
    */
-  AdditionLayerCL() : Layer(), add_props(props::Print()) {}
+  AdditionLayerCL() : LayerImplCl(), add_props(props::Print()) {}
 
   /**
    * @brief     Destructor of Addition Layer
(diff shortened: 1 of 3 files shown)


baek2sm (approved):
LGTM

nntrainer/layers/cl_layers/addition_layer_cl.h
```diff
@@ -93,10 +94,20 @@ class AdditionLayerCL : public Layer {
...
-  std::tuple<props::Print>
-    add_props; /**< fc layer properties : unit - number of output neurons */
+  /**
+   * @brief     registerClKernels for addition_layer_cl
+   * @details   registerClKernels for addition_layer_cl always returns true
+   * without any specific action for kernel registeration. It only uses
+   * cl_blas_kernels and there is no specific kernels for this. If there are
+   * specific kernels for this, it should be updated to register the kernels .
+   */
+  static bool registerClKernels() { return true; };
```
myungjoo, line 104: does it register ClKernels?
or
does this check if CLKernels are registered or not?

(Clareify at brief)

myungjoo:
And if a PR is waiting for too long and you think it's ready, please ping people.

nntrainer/layers/cl_layers/addition_layer_cl.h
EunjuYang: Missed your previous message—sorry for the delayed response!

This change intends to handle layer-specific CL kernel registrations. However, both the `addition` and `transpose` layers don't actually require any specific kernels, so we're skipping the registration step and simply returning true.

I've updated the `@brief` section accordingly. Please let me know if there's a better approach, and I'd be happy to implement it. Thanks again! 😊

skykongkong8 (approved):

djeong20 (approved):

PR #2816, opened 2024-12-04
[GPU] Global Buffer manager and optimization

s-debadri:
Optimized GPU pipeline by managing `Buffer` creations globally:

- Added singleton class `ClBufferManager` for global `Buffer` objects.
- Created 5 buffers of 256 mebibytes globally which can be used by all the kernels multiple times.
- Added wrappers for OpenCL APIs `clEnqueueReadBufferRect` and `clEnqueueWriteBufferRect`.
- Implemented `WriteDataRegion` and `ReadDataRegion` members of `Buffer` class to read and write to a particular region of device  buffer.
- Lazy initialization of buffers at `cl_context`.
- Tested with BLAS kernels.

These changes optimizes the current GPU pipeline by managing device buffers optimally.

> Note: Device buffer preset size can be modified later if required.

**Self evaluation:**

### DIFF
--- a/nntrainer/cl_buffer_manager.cpp
+++ b/nntrainer/cl_buffer_manager.cpp
@@ -22,20 +22,20 @@ ClBufferManager &ClBufferManager::getInstance() {
 // to-do: Implementation to be updated with array of Buffer objects if required
 // fp16 Buffer objects to be added in future
 void ClBufferManager::initBuffers() {
-  readBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  readBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  readBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  writeBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
-  writeBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
+  inBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
+  inBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
+  inBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
+  outBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
+  outBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
   ml_logi("ClBufferManager: Buffers initialized");
 }
 
 ClBufferManager::~ClBufferManager() {
-  delete readBufferA;
-  delete readBufferB;
-  delete readBufferC;
-  delete writeBufferA;
-  delete writeBufferB;
+  delete inBufferA;
+  delete inBufferB;
+  delete inBufferC;
+  delete outBufferA;
+  delete outBufferB;
   ml_logi("ClBufferManager: Buffers destroyed");
 }
 
--- a/nntrainer/cl_buffer_manager.cpp
+++ b/nntrainer/cl_buffer_manager.cpp
@@ -21,7 +21,7 @@ ClBufferManager &ClBufferManager::getInstance() {
 
 // to-do: Implementation to be updated with array of Buffer objects if required
 // fp16 Buffer objects to be added in future
-ClBufferManager::ClBufferManager() {
+void ClBufferManager::initBuffers() {
   readBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   readBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   readBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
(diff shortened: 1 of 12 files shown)


nntrainer/cl_buffer_manager.cpp
```diff
@@ -0,0 +1,42 @@
...
+// to-do: Implementation to be updated with array of Buffer objects if required
+// fp16 Buffer objects to be added in future
+void ClBufferManager::initBuffers() {
+  readBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
+  readBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
+  readBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
+  writeBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
+  writeBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
+  ml_logi("ClBufferManager: Buffers initialized");
+}
```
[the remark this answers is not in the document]
EunjuYang: I understood the concept. The point I want to ask is how may buffers should be created in advance. Also, as you mentioned, the buffer can be reused. Then, the manager should schedule the proper buffer by hiding the internal buffer assets.

nntrainer/cl_buffer_manager.h
```diff
@@ -0,0 +1,77 @@
...
+   *
+   * @return ClBufferManager&
+   */
+  static ClBufferManager &getInstance();
+
+  opencl::Buffer *readBufferA;
+  opencl::Buffer *readBufferB;
+  opencl::Buffer *readBufferC;
+  opencl::Buffer *writeBufferA;
+  opencl::Buffer *writeBufferB;
```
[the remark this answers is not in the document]
s-debadri: Added abstraction in `cl_buffer_manager` in the latest commit.

nntrainer/cl_buffer_manager.cpp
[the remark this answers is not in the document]
s-debadri: As per the PR, 5 buffers are created in advance each of 256 MiB. Also, as you suggested before I have added proper abstraction for `cl_buffer_manager`.

jijoongmoon (approved):
LGTM

EunjuYang (approved):
Overall updates look good :)

I have one question about the performance. 
I run the `unittest_blas_kernels_cl` to check its speedup; the result was interesting. 

This PR shows regularly slow (**545 ms** total in avg.).
The previous version showed slow for the first run (**754 ms**) and  (**243 ms** in avg.) for remaining trials.
Is it because the buffer manager create all buffers for every process in advance? I don't know why the previous version showed better performance except for the first run.

s-debadri:
> Overall updates look good :)
> 
> I have one question about the performance. I run the `unittest_blas_kernels_cl` to check its speedup; the result was interesting.
> 
> This PR shows regularly slow (**545 ms** total in avg.). The previous version showed slow for the first run (**754 ms**) and (**243 ms** in avg.) for remaining trials. Is it because the buffer manager create all buffers for every process in advance? I don't know why the previous version showed better performance except for the first run.

This PR contains new buffer implementation with only `sgemv` kernel as of now. So other unit tests were still creating additional buffers with this new PR. Which means extra buffers were being created with this new PR since all the BLAS kernels are not updated.

For proper calculation of latency, you can see the difference if you run the same kernel (`sgemv` for this PR) multiple times.

> Note: This PR mainly contains the buffer manager and is tested with `sgemv`. New changes for other blas kernels will be done once this PR gets merged otherwise one PR will have too many commits.

nntrainer/cl_context.h
```diff
@@ -272,6 +275,9 @@ class ClContext {
 
     // getContext() called inside createCommandQueue which creates clContext
     bool result = command_queue_inst_.CreateCommandQueue();
+    // initialize device buffers
+    clbuffInstance.initBuffers();
```
myungjoo, line 279: Can you gurantee that initBuffers() is called before any possible call to ClBufferManager::some-access-funtion?
Is there a reason not to do it at the constructor of ClBufferManager?

myungjoo (commented):
Other than ClBufferManager's constructor/initializer timing, LGTM.

nntrainer/cl_context.h
s-debadri: Yes `initBuffers()` will be called at the beginning when context is initialized before any `ClBufferManager` member is accessed.
The main reason of not doing it inside the constructor is to enforce lazy loading of the buffers. Otherwise buffers might get initialized before OpenCL context and command queue is initialized which might result in undefined behaviour when trying to read/write into the buffers. To avoid such ambiguity, it was removed from the constructor.

myungjoo: To guarantee the calling sequence (other memer functions are called after ```initBuffers()```), it should not depend on human programmer's plans, comments, documents, or promises. It is still prone to human errors (especially with open source contributors). It should be guaranteed by the code so that another developer who doesn't know about this in the open source community won't be able to reverse the sequence by mistake, thinking that this object is initialized already with getInstance or constructor.

At least, set the buffer pointers NULL at the constructor so that the caller is guaranteed to know that something's wrong if init wasn't called. And state (doxygen for potentially harmful member functions) that it will return NULL if init is not called.

s-debadri: Thanks for your insights. I have added initializers on the constructor and added relevant doc for the member functions.

myungjoo (approved):

PR #2818, opened 2024-12-09
[Tensor] Binary-code-based quantized tensor

djeong20:
This PR adds a Tensor class that supports the Binary-code-based quantization type.
Binary-code-based quantization ([BCQ](https://sites.google.com/view/insoochung/sub-3bit-quantization)) allows extremely low bit quantization and BCQTensor aims to store and manage such quantized data.
Note that the BiQGEMM is utilized for quantized inference and the BiQGEMM library is needed to use BCQTensor fully.
[Binary-code-based quantizer](https://github.com/insoochung/transformer_bcq) demonstrates how to perform binary-code-based quantazation.

**Self-evaluation:**

### DIFF
--- a/api/ccapi/include/tensor_dim.h
+++ b/api/ccapi/include/tensor_dim.h
@@ -54,6 +54,7 @@ public:
   enum class DataType {
     QINT4,  /** quantized int 4*/
     QINT8,  /** quantized int 8*/
+    BCQ,    /** binary-code-based quantized*/
     UINT8,  /** unsigned int 8 bit */
     UINT16, /** unsigned int 16 bit */
     UINT32, /** unsigned int 32 bit */
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -18,6 +18,11 @@
 #ifdef ENABLE_FP16
 #include <half_tensor.h>
 #endif
+
+#ifdef ENABLE_BIQGEMM
+#include <bcq_tensor.h>
+#endif
+
 namespace nntrainer {
 
 Tensor::Tensor(
@@ -80,6 +85,14 @@ Tensor::Tensor(std::string name_, Tformat fm, Tdatatype d_type) {
   } else if (d_type == Tdatatype::QINT8) {
     itensor = std::shared_ptr<CharTensor>(new CharTensor(name_, fm),
                                           std::default_delete<CharTensor>());
+  } else if (d_type == Tdatatype::BCQ) {
+#ifdef ENABLE_BIQGEMM
+    itensor = std::shared_ptr<BCQTensor>(new BCQTensor(name_, fm),
+                                         std::default_delete<BCQTensor>());
+#else
+    throw std::invalid_argument("Error: enable-biqgemm is not activated. "
+                                "Enable only if your system supports BiQGEMM.");
+#endif
   } else {
     throw std::invalid_argument(
       "Error: Tensor cannot be constructed because the given d_type is not "
@@ -120,6 +133,15 @@ Tensor::Tensor(const TensorDim &d, bool alloc_now, Initializer init,
     itensor =
       std::shared_ptr<CharTensor>(new CharTensor(d, alloc_now, init, name),
                                   std::default_delete<CharTensor>());
+  } else if (d.getDataType() == Tdatatype::BCQ) {
+#ifdef ENABLE_BIQGEMM
+    itensor =
+      std::shared_ptr<BCQTensor>(new BCQTensor(d, alloc_now, init, name),
+                                 std::default_delete<BCQTensor>());
+#else
+    throw std::invalid_argument("Error: enable-biqgemm is not activated. "
+                                "Enable only if your system supports BiQGEMM.");
+#endif
   } else {
     throw std::invalid_argument(
       "Error: Tensor cannot be constructed because the given d_type is not "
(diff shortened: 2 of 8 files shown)


nntrainer/tensor/tensor.cpp
```diff
@@ -249,6 +291,15 @@ bool Tensor::operator==(const Tensor &rhs) const {
...
              *std::dynamic_pointer_cast<CharTensor>(rhs.itensor);
+    } else if (getDataType() == Tdatatype::BCQ) {
+#ifdef ENABLE_BIQGEMM
+      return *std::dynamic_pointer_cast<BCQTensor>(itensor) ==
+             *std::dynamic_pointer_cast<BCQTensor>(rhs.itensor);
+#else
+      throw std::invalid_argument(
+        "Error: enable-biqgemm is not activated. "
+        "Enable only if your system supports BiQGEMM.");
+#endif
```
skykongkong8, line 302: How about `ENABLE_BCQ` ? it is not actually 'BiQGEMM' stuff I guess

api/ccapi/include/tensor_dim.h
```diff
@@ -54,6 +54,7 @@ class TensorDim {
   enum class DataType {
     QINT4,  /** quantized int 4*/
     QINT8,  /** quantized int 8*/
+    BCQ,    /** binary-code-based quantized*/
```
skykongkong8, line 57: trivial, but:
```C++
/** binary-code-based quantization*/
```
?

nntrainer/tensor/tensor.cpp
djeong20: Thank you for sharing your thoughts! To explain, there are two reasons for this.
1. It is designed BCQTensor can be used only when BiQGEMM is available. If BiQGEMM is not available, a user cannot utilize BCQTensor.
2. In meson.build, `enable-biqgemm` checks for the BiQGEMM library.

```
if get_option('enable-biqgemm')
  # check if BiQGEMM directory exist. otherwise, throw an error
  fs = import('fs')
  if fs.is_dir('../BiQGEMM')
    extra_defines += '-DENABLE_BIQGEMM=1'
    biqgemm_inc = include_directories('../BiQGEMM')
  else
    error ('BiQGEMM cannot be enabled without BiQGEMM library.')
  endif
endif
```

Hope this clarifies!

api/ccapi/include/tensor_dim.h
djeong20: I followed the upper naming conventions (quantized int 4, int 8).
If you want, we can change all of it :) 
```
QINT4,  /** int4 quantization*/
QINT8,  /** int8 quantization*/
BCQ,    /** binary-code-based quantization*/
```

nntrainer/tensor/tensor.cpp
```diff
@@ -9,6 +9,7 @@
  * @bug		No known bugs except for NYI items
  */
 
+#include <bcq_tensor.h>
```
EunjuYang, line 12: Doesn't it need `ENABLE_BIQGEMM` ?
```suggestion
#ifdef ENABLE_BIQGEMM
#include <bcq_tensor.h>
#endif
```

api/ccapi/include/tensor_dim.h
skykongkong8: Ah, I see

nntrainer/tensor/tensor.cpp
EunjuYang: Should it be coupled? Can't we decouple BCQTensor and BiQGEMM?

skykongkong8: Thanks for the explanation!
You have a point, but just wanted to make sure again that BCQ and BiQGEMM are quite different.

EunjuYang (commented):
Thank you for the hard work! I'm thrilled to review this PR.
Upon the comment by @skykongkong8 , what about decoupling BCQTensor and BiQGEMM. By doing so (detaching the BiQGEMM from BCQTensor), one may attach their own kernel for BCQTensor. This is just simple opinion by my side.

nntrainer/tensor/tensor.cpp
djeong20: > Should it be coupled? Can't we decouple BCQTensor and BiQGEMM?

You're right, it can be decoupled! Regarding this issue, I discussed this with @jijoongmoon the other day and we've decided to have BCQTensor can only be used when BiQGEMM is available.

nntrainer/tensor/tensor.cpp
djeong20: You're right! I'll fix it right away

nntrainer/tensor/tensor.cpp
djeong20: > To me, determining it with the word 'BiQGEMM' sounds quite unclear.

I get the point that BCQ and BiQGEMM are different. As @EunjuYang mentioned, it can be decoupled (BCQTensor can be created regardless of the presence of BiQGEMM). However, @jijoongmoon and I decided to restrict the use of BCQ-type by BiQGEMM availability. We can discuss what would be the best option!
Also, you're right about the use of BiQGEMM is arch-dep & type-dep! Current dot implementation is targeted only for full-precision computation. Half-precision computation support will be introduced in the later PR :)

skykongkong8: Got it.

skykongkong8 (approved):

EunjuYang (approved):

jijoongmoon (approved):
LGTM

PR #2823, opened 2024-12-10
[ Meson ] BCQTensor dependency is added for Android build

EunjuYang:
- This commit add BiQGEMM path to nntrainer_inc_abs to support Android build with enable-biqgemm option.
- The path for BiQGEMM corresponds to the one in the top `meson.build`

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/meson.build
+++ b/nntrainer/tensor/meson.build
@@ -80,6 +80,7 @@ if get_option('enable-biqgemm')
   tensor_headers += 'bcq_tensor.h'
   tensor_sources += 'bcq_tensor.cpp'
   nntrainer_inc += biqgemm_inc
+  nntrainer_inc_abs += meson.source_root() / '..' / 'BiQGEMM'
 endif
 
 if get_option('enable-opencl')


nntrainer/meson.build
```diff
@@ -11,6 +11,17 @@ nntrainer_inc_abs = [
   meson.source_root() / 'api' / 'ccapi' / 'include'
 ]
 
+# The code below feeds BiQGEMM absolute path to android.mk
+# The BiQGEMM absolute path is included only when biqgemm is enabled.
+if get_option('enable-biqgemm')
```
skykongkong8, line 16: Just to be sure, this won't affect `tools/package_android.sh`, isn't it?

skykongkong8 (approved):

nntrainer/meson.build
EunjuYang: Yes. This won't be affect to the current `tools/package_android.sh`. This aims for the case where `meson build -Dplatform=android -Denable-biqgemm=true`.

djeong20 (approved):
LGTM! 
One minor suggestion: How about adding it to the tensor/meson.build for consistency?

https://github.com/nnstreamer/nntrainer/blob/cd17a66b8eec45a2e91e543a884722763b5e608e/nntrainer/tensor/meson.build#L79-L83

EunjuYang:
> LGTM! One minor suggestion: How about adding it to the tensor/meson.build for consistency?
> 
> https://github.com/nnstreamer/nntrainer/blob/cd17a66b8eec45a2e91e543a884722763b5e608e/nntrainer/tensor/meson.build#L79-L83

Following the suggestion from @djeong20, I moved the code to `nntrainer/tensor/meson.build`.
Since the existence check for BiQGEMM folder is completed at the precedent meson file, the duplicate check code is removed.

jijoongmoon (approved):
LGTM

myungjoo:
```
meson.source_root() / '..' / 'BiQGEMM'
```
DO NOT HARDCODE EXTERNAL PATHS IN MESON SCRIPT!

If pkgconfig or cmake is not ready with BiQGEMM,
get such information via meson_options and specify such "default paths" at meson_options.

myungjoo:
And..

```
if get_option('enable-biqgemm')
  # check if BiQGEMM directory exist. otherwise, throw an error
  fs = import('fs')
  if fs.is_dir('../BiQGEMM')
    extra_defines += '-DENABLE_BIQGEMM=1'
    biqgemm_inc = include_directories('../BiQGEMM')
  else
    error ('BiQGEMM cannot be enabled without BiQGEMM library.')
  endif
endif
```
is completely non-sense. Other developers (e.g., nntrainer users in other departments) won't understand what's going on here.

You should try
1. get biqgemm info from pkgconfig/cmake.
2. (if 1 fails) try to load it from common path (/usr/include, /usr/include/biqgemm)
3. (if 2 fails) try to load it from user-defined path (meson_option)
4. (if 3 fails) try to load it from such hardcoded path. but specify such external hardcoded path at meson_options.txt, not in this script.

EunjuYang:
@myungjoo, Thank you for the detailed comment and guidance :) I will make additional PR to make it feasible following your suggestion. Sincerely.

PR #2824, opened 2024-12-11
Quantizer class to perform quantization

djeong20:
This pull request introduces a quantizer class allowing quantization and dequantization with different schemes. The goal is to offer users more choices when dealing with various types of quantization. Initial support targets include affine quantization (per tensor and per channel) and binary-code-based quantization. This pull request presents the basic structure of these classes, and further implementation details will be added in future updates.

**Self-evaluation:**


skykongkong8 (approved):
Nice draft! but think you need to add this file for tizen spec file

EunjuYang (approved):
LGTM except for CI.

myungjoo:
If you are trying factory method pattern so that
1. you can load a concrete quantizer class on demand 
2. adding a new concrete quantizer NEVER requires changes in the base code.


Make the base class (quantizer) a pure virtual class and design a proper creator class (or creator C module).
You may refer to tensor_filter_subplugin class of nnstreamer, which supports 1 and 2.

djeong20:
> If you are trying factory method pattern so that
> 
> 1. you can load a concrete quantizer class on demand
> 2. adding a new concrete quantizer NEVER requires changes in the base code.
> 
> Make the base class (quantizer) a pure virtual class and design a proper creator class. You may refer to tensor_filter_subplugin class of nnstreamer, which supports 1 and 2. Then, you will need a class that uses this pure vitual class, too; you may regard this as separating virtual part and concrete part of your initial quantizer design.
> 
> Then, you can let an application add arbitrary quantizers without changing nntrainer.

First of all, thank you for the kind guidance! I really appreciate it :)
I have a question and need a help regarding designing the creator class.
To explain, in the current implementation, each quantizer class has one parameter type. This could simplify the factory class.

```
class Quantization {
public:
  std::unique_ptr<Quantizer> createQuantizer(QScheme qscheme, Tdatatype dtype) {
    switch (qscheme) {
    case QScheme::PER_TENSOR_AFFINE:
      return std::make_unique<PerTensorAffineQuantizer>(dtype);
      break;
    case QScheme::PER_CHANNEL_AFFINE:
      return std::make_unique<PerChannelAffineQuantizer>(dtype);
      break;
    case QScheme::BINARY_CODE_BASED:
      return std::make_unique<BinaryCodeBasedQuantizer>(dtype);
      break;
    default:
      throw std::invalid_argument("Unsupported quantization scheme");
      break;
    }
  }
};
```
However, in later development, the parameters of the constructors will differ from one another as follows.

1. `PerTensorAffineQuantizer(Tdatatype dtype_, float scale_, int zero_point_)`
2. `PerChannleAffineQuantizer(Tdatatype dtype_, float* scales_, int* zero_points_)`
3. `BinaryCodeBasedQuantizer(Tdatatype dtype_, float* scales_)`

Also, scale could be fp16 and zero_point can be float for other quantizer.
In this case, I find it hard to think of implementing a method to get the object of type since parameters differ by type.
What would be the best pratice in such case?

myungjoo:
Having such a switch-case clauses for pre-defined subclasses is fine.
As you already know, for user-defined subclasses, it doesn't work.

You need to think the opposite way for such user-defined subclasses.
Reference:
- nnstreamer/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc (a derived/concrete class)
- nnstreamer/gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh (a base class)

A derived class is built as an independent shared object (.so), and when it's loaded,
in the example above, ```init_filter_snpe()```, is called: that's what ```__attribute__ ((constructor));``` means.
Then, ```tensor_filter_subplugin::register_subplugin<snpe_subplugin>``` function
registers the new derived class's empty instance so that applications may fetch this
newly registered derived class with the registered name.

You may use both "name" and "enum" for faster execution; however, you will need a proper
enum rule for that to avoid duplications. In nnstreamer, we use string name anyway.

myungjoo:
One more thing to add:

As long as the custom plugins (derived classes supplied by applications) are used only by the supplier (the application), your implementation does need to be as complex as my example. The example (tensor-filter-cppsubplugin) is to allow applications to access the derived classed from other entities (applications share derived classes).

Using enum instead of string as the key will be much easier with such conditions. You don't need things like ```__attribute__ ((constructor));``` in yours.

myungjoo:
Someone needs to check if similar changes are required for other classes in nntrainer.... but I believe most.. at least "classic".. "extensible" classes of nntrainer can get new subclasses from application without changes or rebuilding of nntrainer.

djeong20:
@myungjoo Thank you for sharing your expertise! You are the best 👍 
I made some modifications following your advice.
1. Quantization class, creator, is added to create various quantizers
2. The quantizer class contains a hash map & register method to support a user-defined quantizer.

I would really appreciate it if you could take a look and please let me know if any changes are needed.

jijoongmoon:
So, Quanized Tensor has quntizer then.

nntrainer/tensor/quantizer.h
```diff
@@ -0,0 +1,326 @@
...
+  /**
+   * @brief Register the user defined quantizer class
+   *
+   * @param qscheme Quantization scheme (use CUSTOM_QUANTIZER_#)
+   * @param quantizer quantizer class to register
+   *
+   * @note This function registers the custom quantizer class. User defined
+   * derived class must be registered with this function.
+   */
+  static void registerQuantizer(QScheme qscheme, Quantizer &quantizer) {
```
jijoongmoon, line 62: This needs to check the duplication of the map. If it is static, the only one object exists throughout the process. I just wonder wether it is possible to use same type of tensor can have different quantizer.

jijoongmoon (approved):
LGTM

PR #2828, opened 2024-12-16
[Quantizer] PerTensorAffineQuantizer operations

djeong20:
This PR adds initial PerTensorAffineQuantizer operation implementations.
This change allows users to quantize and dequantize tensors.
Note that the current implementation is naive and has limited features.
The optimized version will be introduced in the later PR. 

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/quantizer.cpp
+++ b/nntrainer/tensor/quantizer.cpp
@@ -8,9 +8,23 @@
  * @bug		No known bugs except for NYI items
  */
 
+#include <math.h>
 #include <quantizer.h>
 
 namespace nntrainer {
+
+/**
+ * @brief Helper function for clipping
+ *
+ * @tparam T data type
+ * @param val value to clip
+ * @param lower lower bound
+ * @param upper upper bound
+ * @return T cliped data
+ */
+template <typename T> T clip(const T &val, const T &lower, const T &upper) {
+  return std::max(lower, std::min(val, upper));
+}
 
 /**
  * @brief PerTensorAffineQuantizer class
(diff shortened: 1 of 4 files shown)


skykongkong8 (commented):
I really appreciate your current initiative work for new quantizer class in the nntrainer!
Additionally, I want to come up with a little bit of idea to share-
It is possible that in the future, a quantization scheme such as follows may be required:
```C++
// Types.hpp from mllm
...
typedef struct {
    float d;                  // delta
    int8_t qs[QK_K];          // quants
    int16_t bsums[QK_K / 16]; // sum of quants in groups of 16
} block_q8_K;
```

This is because at the computational kernel level, having such additional information helps to reduce its cost a lot!
Although this is not an urgent thing to consider, but I just wanted to inform you about this to avoid potential workload in the future.

djeong20:
> I really appreciate your current initiative work for new quantizer class in the nntrainer! Additionally, I want to come up with a little bit of idea to share- It is possible that in the future, a quantization scheme such as follows may be required:
> 
> ```c++
> // Types.hpp from mllm
> ...
> typedef struct {
>     float d;                  // delta
>     int8_t qs[QK_K];          // quants
>     int16_t bsums[QK_K / 16]; // sum of quants in groups of 16
> } block_q8_K;
> ```
> 
> This is because at the computational kernel level, having such additional information helps to reduce its cost a lot! Although this is not an urgent thing to consider, but I just wanted to inform you about this to avoid potential workload in the future.

Thank you for sharing the information! There are various quantization parameters for different quantization schemes. 
I wanna make it clear that we are on the same page!
Do you mean the current quantization schemes can also have the parameters (delta, quants, sum of quants) or there is another quantization scheme that has such q params?

skykongkong8:
> Thank you for sharing the information! There are various quantization parameters for different quantization schemes. I wanna make it clear that we are on the same page! Do you mean the current quantization schemes can also have the parameters (delta, quants, sum of quants) or there is another quantization scheme that has such q params?

Yes, I mean even with the same current quantization algorithm (uniform quantization, etc) qParam *might* be better to hold such additional information. Of course, there may be cases where this is not necessary, but at least in the case of matrix multiplication operations, it can be more effective to have such information for algorithm implementation. In a nutshell, even current algorithms sometimes require additional information like that, and sometimes they don't, so considering that there might be separate modifications in the future... It's just to inform you to prevent additional fixes in the future 😅

skykongkong8 (approved):

DonghakPark (approved):
Great Work!!! 
LGTM!

nntrainer/tensor/quantizer.cpp
```diff
@@ -8,10 +8,24 @@
...
 namespace nntrainer {
 
+/**
+ * @brief Helper function for clipping
+ *
+ * @tparam T data type
+ * @param val value to clip
+ * @param lower lower bound
+ * @param upper upper bound
+ * @return T cliped data
```
baek2sm, line 23: clipped?

baek2sm (approved):
Nice work. LGTM!

EunjuYang (approved):
LGTM

jijoongmoon (approved):
LGTM.

PR #2831, opened 2024-12-17
[GPU/OpenCL] Fused DotCL, Addition and RMS for optimization

yashSingh0723:
1. Added the Fused Layer for all above operation. `fused_fc_norm_cl.cpp`
2. Fused Process where existing kernels are getting used, Please refer the below files: `blas_kernel_interface.cpp` and `blas_kernels.cpp`.
3. Wrote Related Unittest for the above
4. Need to check for layer unittest as they are failing. Some issues with generating golden data, Output is correct but st some point not able to match it with generated golden data. Check file` test/unittest/layers/unittest_layers_fused_fc_rms_cl.cpp` and `test/input_gen/gen_layer_tests.py`. For Reference and Debugging check `test/unittest/layers/layers_golden_tests.cpp`.
5. Added `rms_norm_cl_kernel `in `blas_kernels `directory for generalization.

TO-DO
1. Add for FP16
2. Add for dotcl for case M==1 and N==1.

Signed-off-by: Yash Singh [yash.singh@samsung.com](mailto:yash.singh@samsung.com)

### DIFF
--- a/nntrainer/layers/cl_layers/fused_fc_norm_cl.cpp
+++ b/nntrainer/layers/cl_layers/fused_fc_norm_cl.cpp
@@ -1,0 +1,243 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Debadri Samaddar <s.debadri@samsung.com>
+ *
+ * @file	fused_fc_norm_cl.cpp
+ * @date	7 May 2024
+ * @brief	This is Fully Connected Layer Class for Neural Network with OpenCl
+ * implementation
+ * @see		https://github.com/nnstreamer/nntrainer
+ * @author	Debadri Samaddar <s.debadri@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ */
+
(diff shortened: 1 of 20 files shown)


nntrainer/layers/cl_layers/fused_fc_norm_cl.cpp
```diff
@@ -0,0 +1,243 @@
...
+  Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
+
+  // for rms
+  Tensor &gamma = context.getWeight(weight_idx[FC_RMSParams::gamma]);
+  auto &epsilon = std::get<props::Epsilon>(fc_rms_props).get();
+
+  auto disable_bias = std::get<props::DisableBias>(*layer_impl_props);
+  bool disable_bias_value = disable_bias.empty() || disable_bias.get() == false;
+  const Tensor &bias = context.getWeight(weight_idx[FC_RMSParams::bias]);
+  // printf("\n*************************************************************************************************************************************\n");
```
baek2sm, line 144: do we need these comments?

PR #2832, opened 2024-12-18
Pr/transpose/simd/fp32

skykongkong8:
This PR proposes neon simd kernel in fp32 matirix transpose.
Table below is conducted with Galaxy S24U, TC = 100.
Note that this kernel is more effective for sufficiently big matrices, but still better than before anyway.
I added additional unittest TC for validating this function with simple idea : (A.T.T = A) but please suggest any idea if there's more efficient way.

| dim | prev | neon |
| --- | --- | --- |
| 768x768 | 1.9 ms | 1.6 ~ 1.0 ms |
| 1440x1440 | 2.9 ms | 2.29 ms |
| 1920x1560 | 4.2 ~ 3.6 ms | 3.36 ~ 2.67 ms |
| 1560x2048 | 7.13 ~ 6.97 ms | 3.57 ~ 3.2 ms |
| 512x2048 | 2.75 ms  | 1.80 ~ 1.7 ms  |

**Self evaluation:**


nntrainer/tensor/blas_interface.cpp
```diff
@@ -659,6 +659,16 @@ static unsigned int isamax_raw(const unsigned int N, const float *X,
...
 
+void transpose_matrix(const unsigned int M, const unsigned int N,
+                      const float *src, unsigned int ld_src, float *dst,
+                      unsigned int ld_dst) {
+#ifdef USE_NEON
+  transpose_neon<float>(M, N, src, ld_src, dst, ld_dst);
+#else
+  transpose_fallback<float>(M, N, src, ld_src, dst, ld_dst);
+#endif
+}
```
myungjoo, line 670: Someday, all of such #ifdef (NEON, FP16, ...) need to be migrated to header files (preferably centralized into a single header) and code lines depending on such need to be separated as a class or a file (e.g., defined in ```blas_neon.cpp``` and whether to use functions/sub-classes in that file is determined at a header and build-script/option.

But for today, lets move on.

myungjoo (approved):

nntrainer/tensor/blas_interface.cpp
skykongkong8: Seems this issue is related to #2549
I am aware of it. Think we urgently have to discuss when to apply this

skykongkong8: In fact, one of the biggest problems of applying #2549 is that all android make file / tizen spec file in the current nntr is including `blas_interface.h` since this pr substitutes `blas_interface` into single `cpu_backend.h`, this affects quite a lot of files...

EunjuYang (approved):
LGTM.
Combining transpose with dot product, e.g., ($$(W \cdot I)^T = I^T \cdot W^T$$) can be another option for unit test for transpose. Anyway transpose of transpose seems good :)

skykongkong8:
> LGTM. Combining transpose with dot product, e.g., ( ( W ⋅ I ) T = I T ⋅ W T ) can be another option for transpose unittest. Anyway transpose of transpose seems good :)

Nice approach! Gotta try that one too

DonghakPark (approved):
LGTM!

PR #2835, opened 2024-12-23
[Unittest] FSU Unittest with Simple FC Model

DonghakPark:
Add Unittest for FSU
- Use Simple FC Model that consist with 6-FC layer
- FP16-FP16 / Lookahead 1

i will add more case of FSU Model at Benchmark

**Self evaluation:**

### DIFF
--- a/test/unittest/integration_tests/integration_test_fsu.cpp
+++ b/test/unittest/integration_tests/integration_test_fsu.cpp
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Donghak Park <donghak.park@samsung.com>
+ *
+ * @file   integration_test_fsu.cpp
+ * @date   20 Dec 2024
+ * @brief  Unit Test for Asynch FSU
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Donghak Park <donghak.park@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#include <app_context.h>
+#include <array>
+#include <chrono>
+#include <ctime>
+#include <gtest/gtest.h>
+#include <iostream>
+#include <layer.h>
+#include <memory>
+#include <model.h>
+#include <optimizer.h>
+#include <sstream>
+#include <vector>
+
+using LayerHandle = std::shared_ptr<ml::train::Layer>;
+using ModelHandle = std::unique_ptr<ml::train::Model>;
+
+template <typename T>
+static std::string withKey(const std::string &key, const T &value) {
+  std::stringstream ss;
+  ss << key << "=" << value;
+  return ss.str();
+}
+
+template <typename T>
+static std::string withKey(const std::string &key,
+                           std::initializer_list<T> value) {
+  if (std::empty(value)) {
+    throw std::invalid_argument("empty data cannot be converted");
+  }
+
+  std::stringstream ss;
+  ss << key << "=";
+
+  auto iter = value.begin();
+  for (; iter != value.end() - 1; ++iter) {
+    ss << *iter << ',';
+  }
+  ss << *iter;
+
+  return ss.str();
+}
+
+TEST(fsu, simple_fc) {
+
+  std::unique_ptr<ml::train::Model> model = ml::train::createModel(
+    ml::train::ModelType::NEURAL_NET, {withKey("loss", "mse")});
+
+  model->addLayer(ml::train::createLayer(
+    "input", {withKey("name", "input0"), withKey("input_shape", "1:1:320")}));
+  for (int i = 0; i < 6; i++) {
+    model->addLayer(ml::train::createLayer(
+      "fully_connected",
+      {withKey("unit", 1000), withKey("weight_initializer", "xavier_uniform"),
+       withKey("bias_initializer", "zeros")}));
+  }
+  model->addLayer(ml::train::createLayer(
+    "fully_connected",
+    {withKey("unit", 100), withKey("weight_initializer", "xavier_uniform"),
+     withKey("bias_initializer", "zeros")}));
+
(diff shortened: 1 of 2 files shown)


test/unittest/integration_tests/integration_test_fsu.cpp
```diff
@@ -0,0 +1,124 @@
...
+    {withKey("unit", 1000), withKey("weight_initializer", "xavier_uniform"),
+     withKey("bias_initializer", "zeros")}));
+  model->addLayer(ml::train::createLayer(
+    "fully_connected",
+    {withKey("unit", 1000), withKey("weight_initializer", "xavier_uniform"),
+     withKey("bias_initializer", "zeros")}));
+  model->addLayer(ml::train::createLayer(
+    "fully_connected",
+    {withKey("unit", 1000), withKey("weight_initializer", "xavier_uniform"),
+     withKey("bias_initializer", "zeros")}));
```
EunjuYang, line 81: What about updating this duplicate code blocks with loop?

test/unittest/integration_tests/integration_test_fsu.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  int status = model->compile(ml::train::ExecutionMode::INFERENCE);
+  EXPECT_EQ(status, ML_ERROR_NONE);
+
+  status = model->initialize(ml::train::ExecutionMode::INFERENCE);
+  EXPECT_EQ(status, ML_ERROR_NONE);
+
+  model->save("simplefc_weight_fp16_fp16_100.bin",
+              ml::train::ModelFormat::MODEL_FORMAT_BIN);
+  model->load("./simplefc_weight_fp16_fp16_100.bin");
```
EunjuYang, line 103: Simple question. Why do we need this `save` and `load` code? Is it because it only supports inference mode now?

test/unittest/integration_tests/integration_test_fsu.cpp
```diff
@@ -0,0 +1,124 @@
...
+    input[j] = j;
+
+  std::vector<float *> in;
+  std::vector<float *> l;
+  std::vector<float *> answer;
+
+  in.push_back(input);
+  l.push_back(label);
+
+  answer = model->inference(1, in, l);
```
EunjuYang, line 120: How can we check the FSU works from this test code?

EunjuYang (commented):
Overall LGTM :) I have some questions on this PR. Could you answer to this? Thank you.

test/unittest/integration_tests/integration_test_fsu.cpp
DonghakPark: Good idea i will fix that !

test/unittest/integration_tests/integration_test_fsu.cpp
DonghakPark: for test FSU, we need file that stored in storage. so save model file, and load files for now
--> in real Inference case, they have bin files already

test/unittest/integration_tests/integration_test_fsu.cpp
DonghakPark: You can check mem_usage. 
actually i add this case for FSU build test, not for detail works
--> working test will be add to benchmarks tool

test/unittest/integration_tests/integration_test_fsu.cpp
EunjuYang: Oh yes, that's what the FSU is !🤣 Thank you for the kind reply :)

myungjoo:
LGTM.

Question on FSU in general:
- Would it be possible to use FSU for inference without calling any C++ APIs except for the simple inference APIs?
e.g., (pseudo code)
```
h = open_model(modelfile_path);
h.configure(FSU=true);
output = h.run(input);
h = close_model();
```

DonghakPark:
> LGTM.
> 
> Question on FSU in general:
> 
> * Would it be possible to use FSU for inference without calling any C++ APIs except for the simple inference APIs?
>   e.g., (pseudo code)
> 
> ```
> h = open_model(modelfile_path);
> h.configure(FSU=true);
> output = h.run(input);
> h = close_model();
> ```
i think It is possible to activate FSU by simply setting Setproperty after loading the model that train has been completed.

skykongkong8 (approved):
LGTM!

EunjuYang (approved):

myungjoo:
> > LGTM.
> > Question on FSU in general:
> > 
> > * Would it be possible to use FSU for inference without calling any C++ APIs except for the simple inference APIs?
> >   e.g., (pseudo code)
> > 
> > ```
> > h = open_model(modelfile_path);
> > h.configure(FSU=true);
> > output = h.run(input);
> > h = close_model();
> > ```
> 
> i think It is possible to activate FSU by simply setting Setproperty after loading the model that train has been completed.

In the end, an app like the "f" should be able to do so.

myungjoo (approved):

DonghakPark:
> > > LGTM.
> > > Question on FSU in general:
> > > 
> > > * Would it be possible to use FSU for inference without calling any C++ APIs except for the simple inference APIs?
> > >   e.g., (pseudo code)
> > > 
> > > ```
> > > h = open_model(modelfile_path);
> > > h.configure(FSU=true);
> > > output = h.run(input);
> > > h = close_model();
> > > ```
> > 
> > 
> > i think It is possible to activate FSU by simply setting Setproperty after loading the model that train has been completed.
> 
> In the end, an app like the "f" should be able to do so.

That's a great idea. By raising the level of abstraction, allowing users with pre-trained weights to simply activate the swap functionality using something like model.Run(parm) would significantly enhance convenience during development. 
for this i will make issue

djeong20 (approved):
Nice work!

jijoongmoon (approved):
LGTM

PR #2843, opened 2024-12-26
[FSU] Modify the condition of LoadTensors

SeoHyungjun:
Previously, more layers were loaded into memory than the value of lookahead. 
I modified the code so that the correct number of layers are loaded.

In addition, I removed the conditional statement so that the same code is executed 
when the value of lookahead is 0 and when it is not.

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -393,20 +393,10 @@ sharedConstTensors NeuralNetwork::forwarding(
         unsigned int lookahead =
           std::get<props::MemorySwapLookahead>(model_flex_props);
 
-        if (lookahead != 0) {
-          if ((f) % (lookahead + 1) == lookahead - 1) {
-            std::cout << "request load tensor : " << f + lookahead + 1
-                      << std::endl;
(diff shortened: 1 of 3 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -393,17 +393,8 @@ sharedConstTensors NeuralNetwork::forwarding(
...
-                      << std::endl;
-            ml_logd("request load tensor for %d", f + 1);
-            model_graph.LoadTensors((f / (lookahead + 1) + 1) *
-                                    (lookahead + 1));
-          }
-        } else {
-          model_graph.LoadTensors(f);
-        }
+        if (f != 0 && lookahead != 0)
+            model_graph.LoadTensors(f);
```
DonghakPark, line 397: for lookahead = 0 case (on demand) 
doesn't we need this code ? 
```.cpp
else {
          model_graph.LoadTensors(f);
        }
```

SeoHyungjun: Removed the condition Lookahead != 0.

SeoHyungjun:
There is a problem when the lookahead is 0 in the current code.
Confirmation is required.

PR #2844, opened 2024-12-27
 [CharTensor] Enable memory data to store scale factors based on quantization schemes

djeong20:
This pull request aims to modify the existing codebase such that the memory data of CharTensor can now store scale factors based on different quantization schemes.
Additionally, this change allows the Tensor class to specify the desired quantization scheme while creating a new CharTensor instance.
The scale factors are determined either during the quantization process using a specific quantizer or they can be manually initialized if both the quantized data and the corresponding scale factors are provided as inputs.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/char_tensor.cpp
+++ b/nntrainer/tensor/char_tensor.cpp
@@ -17,18 +17,18 @@
 
 namespace nntrainer {
 
-CharTensor::CharTensor(std::string name_, Tformat fm) :
+CharTensor::CharTensor(std::string name_, Tformat fm, QScheme qscheme_) :
   TensorBase(name_, fm, Tdatatype::QINT8) {}
 
 CharTensor::CharTensor(const TensorDim &d, bool alloc_now, Initializer init,
-                       std::string name) :
-  TensorBase(d, alloc_now, init, name) {
+                       std::string name, QScheme qscheme_) :
+  TensorBase(d, alloc_now, init, name), qscheme(qscheme_) {
   if (alloc_now)
     allocate();
 }
 
-CharTensor::CharTensor(const TensorDim &d, const void *buf) :
-  CharTensor(d, true) {
+CharTensor::CharTensor(const TensorDim &d, const void *buf, QScheme qscheme_) :
+  CharTensor(d, true, Initializer::NONE, "", qscheme_) {
   if (d.getDataLen() != 0) {
     if (buf != nullptr)
       copy(buf);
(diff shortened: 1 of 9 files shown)


nntrainer/tensor/char_tensor.cpp
```diff
@@ -59,9 +59,14 @@ CharTensor::CharTensor(
...
   initializer = Initializer::NONE;
+  qscheme = qscheme_;
 
+  NNTR_THROW_IF(scales.size() != scale_size(), std::invalid_argument)
+    << "invalid scale factor size " << scales.size();
+
+  /// @note 4 * scale_size() assumes scale factors are in full-precision fp.
   MemoryData *mem_data =
-    new MemoryData((void *)(new int8_t[dim.getDataLen()]()));
+    new MemoryData((void *)(new int8_t[dim.getDataLen() + 4 * scale_size()]()));
```
EunjuYang, line 69: What about updating the code like:
```suggestion
    new MemoryData((void *)(new int8_t[dim.getDataLen() + sizeof(float) * scale_size()]()));
```
If you have a plan to support different type of `scales`, it would be better to update it.

nntrainer/tensor/char_tensor.cpp
```diff
@@ -141,6 +162,25 @@ void *CharTensor::getData(size_t idx) const {
...
+
+void *CharTensor::getScale(size_t idx) const {
+  NNTR_THROW_IF(idx > scale_size(), std::invalid_argument)
+    << "Tensor::getScale() index is not valid";
+
+  if (!data)
+    return nullptr;
+
+  data->validate();
+  return ((int8_t *)getScale()) + idx;
```
EunjuYang, line 181: Isn't it  `((float *)getScale()) + idx`?

EunjuYang (commented):
Good works :) Please check my review below:

nntrainer/tensor/char_tensor.cpp
djeong20: thank you for sharing your thoughts! `sizeof(float)` makes more sense :)

nntrainer/tensor/char_tensor.cpp
djeong20: yes, you're right! thank you 👍

EunjuYang (approved):
LGTM !

skykongkong8 (approved):
LGTM

jijoongmoon (approved):
LGTM

Issue #2837, opened 2024-12-24
FSU Test & Verification on CPU


DonghakPark:
FSU CPU side Test & Verification at Inference 

- [x] Inference test (Fully Connected Layer)
- [x] test swap mem_usage ( each lookahead )

work with @SeoHyungjun 

we will update test result & issues on this issue

DonghakPark:
@SeoHyungjun  My test Model is here 
```.cpp
std::vector<LayerHandle> createGraph() {
  using ml::train::createLayer;

  std::vector<LayerHandle> layers;

  layers.push_back(createLayer(
    "input", {withKey("name", "input0"), withKey("input_shape", "1:1:320")}));

  layers.push_back(createLayer("fully_connected",
                               {withKey("unit", 100),
                                withKey("weight_initializer", "xavier_uniform"),
                                withKey("bias_initializer", "zeros")}));

  for (int i = 0; i < 50; i++) {
    layers.push_back(createLayer(
      "fully_connected",
      {withKey("unit", 5000), withKey("weight_initializer", "xavier_uniform"),
       withKey("bias_initializer", "zeros")}));
  }

  layers.push_back(createLayer("fully_connected",
                               {withKey("unit", 100),
                                withKey("weight_initializer", "xavier_uniform"),
                                withKey("bias_initializer", "zeros")}));

  return layers;
}
```

DonghakPark:
@SeoHyungjun Please check below application code 

```.cpp
// SPDX-License-Identifier: Apache-2.0
/**
 * Copyright (C) 2024 Jijoong Moon <jijoong.moon@samsung.com>
 *
 * @file   main.cpp
 * @date   10 Dec 2024
 * @brief  Test Application for Asynch FSU
 * @see    https://github.com/nnstreamer/nntrainer
 * @author Jijoong Moon <jijoong.moon@samsung.com>
 * @bug    No known bugs except for NYI items
 */
#include <array>
#include <chrono>
#include <ctime>
#include <iostream>
#include <memory>
#include <sstream>
#include <vector>


#include <layer.h>
#include <model.h>
#include <optimizer.h>

#include <cifar_dataloader.h>

#ifdef PROFILE
#include <profiler.h>
#endif

using LayerHandle = std::shared_ptr<ml::train::Layer>;
using ModelHandle = std::unique_ptr<ml::train::Model>;

using UserDataType = std::unique_ptr<nntrainer::util::DataLoader>;

/** cache loss values post training for test */
[...]

PR #2846, opened 2024-12-27
[FSU] Modify the condition of LoadTensors

SeoHyungjun:
Can find out the number of currently loaded tensors through the getNumLoadedTensors function. 
When LoadTensors is executed once, getNumLoadedTensors increases twice.

Added the leave_lookahead argument to LoadTensors. 
Calculate leave_lookahead through getNumLoadedTensors and execute loadTensors.

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -383,31 +383,19 @@ sharedConstTensors NeuralNetwork::forwarding(
                the forwarding, ask load tensors for next n layers.
       **/
 
-      if (f == 0)
-        model_graph.LoadTensors(f);
+      unsigned int lookahead =
+        std::get<props::MemorySwapLookahead>(model_flex_props);
 
-      if (model_graph.checkLoadComplete(f)) {
-        node->forwarding(training);
-        ml_logd("Forwarding is done %d : %s", f, node->getName().c_str());
-
-        unsigned int lookahead =
-          std::get<props::MemorySwapLookahead>(model_flex_props);
-
-        if (lookahead != 0) {
-          if ((f) % (lookahead + 1) == lookahead - 1) {
-            std::cout << "request load tensor : " << f + lookahead + 1
-                      << std::endl;
-            ml_logd("request load tensor for %d", f + 1);
-            model_graph.LoadTensors((f / (lookahead + 1) + 1) *
-                                    (lookahead + 1));
-          }
-        } else {
-          model_graph.LoadTensors(f);
-        }
-
-        if (f != 0)
-          model_graph.UnloadTensors(f);
+      if (!((model_graph.getNumLoadedWeightPoolTensors() + 1) / 2 <
+            lookahead + 1)) {
+        model_graph.checkUnloadComplete(f - 1);
       }
+      model_graph.LoadTensors(
+        f, lookahead - (model_graph.getNumLoadedWeightPoolTensors() + 1) / 2);
+
+      model_graph.checkLoadComplete(f);
+      node->forwarding(training);
+      model_graph.UnloadTensors(f);
     }
   };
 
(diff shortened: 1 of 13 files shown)


DonghakPark:
Releated issue : #2837

DonghakPark (approved):
LGTM! but Please Check CI & Rebase onto main

jijoongmoon (approved):
LGTM

skykongkong8 (commented):
First things first, it seems you should apply some clang-format here.
And I really don't get "When LoadTensors is executed once, getNumLoadedTensors increases twice."
Was it a problem to fix? or is it a change proposed from this PR?

nntrainer/models/neuralnet.cpp
```diff
@@ -383,31 +383,19 @@ sharedConstTensors NeuralNetwork::forwarding(
                the forwarding, ask load tensors for next n layers.
       **/
```
djeong20, line 384: please update the brief!

nntrainer/models/neuralnet.cpp
```diff
@@ -383,31 +383,19 @@ sharedConstTensors NeuralNetwork::forwarding(
...
+        std::get<props::MemorySwapLookahead>(model_flex_props);
 
-        if (f != 0)
-          model_graph.UnloadTensors(f);
+      if (!((model_graph.getNumLoadedWeightPoolTensors() + 1) / 2 <
+            lookahead + 1)) {
+        model_graph.checkUnloadComplete(f - 1);
       }
+      model_graph.LoadTensors(
+        f, lookahead - (model_graph.getNumLoadedWeightPoolTensors() + 1) / 2);
```
djeong20, line 394: Would this logic still be applicable when the `lookahead` value is 0?

djeong20 (commented):
Please add or update the description for the newly added or updated function in the header file.

nntrainer/tensor/manager.cpp
```diff
@@ -813,9 +813,10 @@ void Manager::LoadTensors(unsigned int order) {
     async_load_tensor[o] = std::make_tuple(load_weight, load_tensor);
   };
 
-  for (unsigned int i = order; i < order + swap_lookahead + 1; ++i) {
-    if (i <= max_exec_order)
+  for (unsigned int i = order; i < order + remainder_lookahead + 1; ++i) {
+    if (i <= max_exec_order) {
       enqueTasks(i);
```
EunjuYang, line 818: Simple question. Does it act like `enqueue` ? I find this lambda function do two tasks: 1) neglecting a duplicate request and 2) directly executing  loadTensorAsync. In this sense, it seems there is no queue for request. Do I correctly understood?

nntrainer/models/neuralnet.cpp
```diff
@@ -383,31 +383,19 @@ sharedConstTensors NeuralNetwork::forwarding(
...
-          model_graph.LoadTensors(f);
-        }
+      unsigned int lookahead =
+        std::get<props::MemorySwapLookahead>(model_flex_props);
 
-        if (f != 0)
-          model_graph.UnloadTensors(f);
+      if (!((model_graph.getNumLoadedWeightPoolTensors() + 1) / 2 <
+            lookahead + 1)) {
+        model_graph.checkUnloadComplete(f - 1);
```
EunjuYang, line 391: Do we have to call `checkUnloadComplete` here? In my opinion, calling `forward()` should not be blocked by this unload process.

EunjuYang (commented):
As I understood, this PR updates the codes in the way that swap requests are enqueued depending on the `num_loaded_tensors`; the `num_loaded_tensors` is linked with  `swapIn`. It seems there's no queue but swap request is enqueued and executed only when it can be executable.  In this sense, `checkUnloadComplete` is called in `forwarding()` function to verify unload is completed well. However, it seems that this might hinder `forward()` process, lowering performance. I'm not sure I understood correctly. It would be great if I could get more explanation on this.

DonghakPark:
This commit merged by #2906 with some modify

PR #2847, opened 2024-12-27
[ Log ] Adding an error log when setOptimizer is called after initialization

### LINKED ISSUE #2830: Proposal for Exception Handling in NNTrainer's setOptimizer Method
I encountered an issue while using the NNTrainer to create a model with code. It seems that there is no clear guidance on the execution order for methods such as `NeuralNetwork::setOptimizer`.
I placed the call to `NeuralNetwork::setOptimizer` after `NeuralNetwork::initialize`. However, even though the code executed without any errors, it did not proceed with the actual training process.

I believe it would be beneficial if the `NeuralNetwork::setOptimizer` could throw an exception or trigger an error when called after initialization or `initialize` called without calling `NeuralNetwork::setOptimizer`. This way, users can easily identify and rectify any issues related to the incorrect execution sequence.

EunjuYang:
This PR aims to resolve #2830.

- This PR adds an error log when setOptimizer is called after initialization.
- This PR updates some undesirable codes of example application to check its setOptimizer status.

**Self evaluation:**

### DIFF
--- a/Applications/LLaMA/jni/main.cpp
+++ b/Applications/LLaMA/jni/main.cpp
@@ -668,9 +668,12 @@ void createAndRun(unsigned int epochs, unsigned int batch_size) {
                         withKey("save_path", "test_model.bin")});
 
   auto optimizer = ml::train::createOptimizer("sgd", {"learning_rate=0.001"});
-  g_model->setOptimizer(std::move(optimizer));
+  int status = g_model->setOptimizer(std::move(optimizer));
+  if (status) {
+    throw std::invalid_argument("failed to set optimizer!");
+  }
 
-  int status = g_model->compile();
+  status = g_model->compile();
   if (status) {
     throw std::invalid_argument("model compilation failed!");
   }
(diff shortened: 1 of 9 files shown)


DonghakPark (approved):
LGTM!!
how about print error log at setOptimizer() at neuralnet.cpp
```.cpp
int NeuralNetwork::setDataBuffer(const DatasetModeType &mode,
                                 std::shared_ptr<DataBuffer> data_buffer) {
  if (data_buffer == nullptr) {
    //print error log here ? 
  }

  this->data_buffers[static_cast<int>(mode)] = data_buffer;

  return ML_ERROR_NONE;
}
```

djeong20 (approved):
LGTM!

myungjoo (approved):
This is an application code, not a library code. Printing out error messages to std::err is enough, but this works as well..

(anyway, for a library code, I'd recommend to add more detailed error messages)

jijoongmoon (approved):
LGTM

PR #2849, opened 2024-12-30
Add ComputeEngine Property for choosing Engine

jijoongmoon:
## Add Compute Engine Property
This PR add ComputeEngine Enum Property.
Enum elements are "cpu", "gpu", "qnn" for now.
The property format is "engine=qnn".

**Self evaluation:**

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -115,14 +115,6 @@ enum LayerType {
 };
 
 /**
- * @brief     Enumeration of layer compute engine
- */
-enum LayerComputeEngine {
-  CPU, /**< CPU as the compute engine */
-  GPU, /**< GPU as the compute engine */
-};
-
-/**
  * @class   Layer Base class for layers
  * @brief   Base class for all layers
  */
@@ -261,16 +253,14 @@ public:
  */
 std::unique_ptr<Layer>
 createLayer(const LayerType &type,
-            const std::vector<std::string> &properties = {},
-            const LayerComputeEngine &compute_engine = LayerComputeEngine::CPU);
+            const std::vector<std::string> &properties = {});
 
 /**
  * @brief Factory creator with constructor for layer
  */
 std::unique_ptr<Layer>
 createLayer(const std::string &type,
-            const std::vector<std::string> &properties = {},
-            const LayerComputeEngine &compute_engine = LayerComputeEngine::CPU);
+            const std::vector<std::string> &properties = {});
 
 /**
  * @brief General Layer Factory function to register Layer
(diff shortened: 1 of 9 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -129,21 +129,73 @@ class SharedFrom : public Name {
...
 LayerNode::~LayerNode() = default;
+  
+/**
+ * @brief get the compute engine property from property string vector
+ *  : default is CPU
+ * @return LayerComputeEngine Enum : CPU, GPU, QNN
+ *
+ */
+ml::train::LayerComputeEngine
+getComputeEngine(const std::vector<std::string> &props) {
```
EunjuYang, line 140: Function prototype for `getComputeEngine` is missing in header. Isn't it required to be declared in header?

nntrainer/layers/layer_node.cpp
```diff
@@ -129,21 +129,73 @@ class SharedFrom : public Name {
...
-                           compute_engine);
-  }
+    return createLayerNode(cc.createObject<nntrainer::Layer>(type), properties);
+#else
+    throw std::invalid_argument(
+      "opencl layer creation without enable-opencl option");
 #endif
+  }
+
   auto &ac = nntrainer::AppContext::Global();
```
EunjuYang, line 199: Do you think it holds for both `ml::train::LayerComputeEngine::CPU` and `ml::train::LayerComputeEngine::QNN`?
If not, you may need to add more condition to check it.
I think it would be better to add a case to check `QNN` with NYI exception.

nntrainer/layers/layer_node.cpp
```diff
@@ -153,15 +205,18 @@ createLayerNode(const ml::train::LayerType &type,
...
```
EunjuYang, line 220: ditto.

EunjuYang (commented):
I'm glad to review this PR, initiating additional compute engine for QNN. 👍✨
This PR adds a new `LayerComputeEngine` property in `common.h` and includes it as a property. Here're some comments from my side.

myungjoo (approved):

nntrainer/layers/layer_node.cpp
jijoongmoon: Thanks for your comment. This getComputeEngine function will be removed in PR #2849. 
It moved and renamed as "parseComputeEngine"

nntrainer/layers/layer_node.cpp
jijoongmoon: No. it doesn't. This will changed in PR #2848 too. 
https://github.com/nnstreamer/nntrainer/blob/dbbde82f1042e4a34be01fae7ac9cdcdb95657c7/nntrainer/layers/layer_node.cpp#L140

EunjuYang (approved):

api/ccapi/include/layer.h
```diff
@@ -343,37 +333,35 @@ DivideLayer(const std::vector<std::string> &properties = {}) {
...
-}
+// /**
+//  * @brief Helper function to create RMS normalization layer for GPU
+//  */
+// inline std::unique_ptr<Layer>
+// RMSNormCl(const std::vector<std::string> &properties = {},
+//           const LayerComputeEngine &compute_engine = LayerComputeEngine::GPU)
+//           {
+//   return createLayer(LayerType::LAYER_RMSNORM, properties, compute_engine);
+// }
```
djeong20, line 357: let's remove it

djeong20 (approved):
LGTM

api/ccapi/include/layer.h
jijoongmoon: I think it might be better to remove this after evaluating the cl_context.

PR #2850, opened 2024-12-31
[CharTensor] Enable QINT8 multiplication feature

djeong20:
This pull request aims to enable the QINT8 element-wise multiplication feature in CharTensor.
This takes two tensors of the same dimensions and returns a matrix of the multiplied corresponding elements.
Please note that automatically determining the new scale factor will be added in a future update.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/char_tensor.cpp
+++ b/nntrainer/tensor/char_tensor.cpp
@@ -64,7 +64,6 @@ CharTensor::CharTensor(
   NNTR_THROW_IF(scales.size() != scale_size(), std::invalid_argument)
     << "invalid scale factor size " << scales.size();
 
-  /// @note 4 * scale_size() assumes scale factors are in full-precision fp.
   MemoryData *mem_data = new MemoryData(
     (void *)(new int8_t[dim.getDataLen() + sizeof(float) * scale_size()]()));
   data = std::shared_ptr<MemoryData>(mem_data, [](MemoryData *mem_data) {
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/char_tensor.cpp
```diff
@@ -268,6 +267,56 @@ void CharTensor::initialize(Initializer init) {
...
+    << ((bool)(this->getFormat()) ? "NHWC" : "NCHW") << " is not match. ("
+    << ((bool)(input.getFormat()) ? "NHWC" : "NCHW") << ")";
+
+  NNTR_THROW_IF(!contiguous || !input.getContiguous() ||
+                  !output.getContiguous(),
+                std::invalid_argument)
+    << getName() << " is not contiguous, cannot multiply";
+
+  float lhs_scale = *(float *)getScale();
+  float rhs_scale = *input.getScale<float>();
```
EunjuYang, line 291: Are they always scalars? Based on #2844, it seems scale can be an array.
Is it related to the condition in the note, i.e.,  *4. only per-tensor quantization qscheme is supported* ?
if so, we need to add a condition to verify if the case can be supported or not.

EunjuYang (commented):
LGTM except for the comment below.

nntrainer/tensor/char_tensor.cpp
djeong20: Thank you for the detailed review! Yes, you're correct. Currently, there isn't such a way to check the quantization scheme of the tensor. I'll create a new PR to do so and apply it.

EunjuYang (approved):

nntrainer/tensor/char_tensor.cpp
```diff
@@ -228,6 +267,56 @@ void CharTensor::initialize(Initializer init) {
...
+
+  for (unsigned int i = 0; i < size(); ++i) {
+    int32_t accum_val =
+      static_cast<int32_t>(lhs[i]) * static_cast<int32_t>(rhs[i]);
+
+    result[i] =
+      std::max(-128, std::min((int)std::lround(multiplier * accum_val), 127));
+  }
+
+  *output.getScale<float>() = scale;
```
skykongkong8, line 315: As you might already know, in order to add simd accelerated code and maintain it with vanilla code altogether, this should be reside at `blas_interface.cpp` level. I can do that later on, but would like to discuss one thing:
```C++
// scalar scale
void ele_mul(int8_t* lhs, int8_t* rhs, int8_t* res, float lhs_scale, float rhs_scale, float scale, unsigned int N);

// vector scale
void ele_mul(int8_t* lhs, int8_t* rhs, int8_t* res, float* lhs_scale, float* rhs_scale, float* scale, unsigned int N);
```

would function design like above valid for your intention?

skykongkong8 (approved):
Otherwise, great work!

nntrainer/tensor/char_tensor.cpp
djeong20: Thank you for sharing your opinion! To answer yes, the function design you suggested would be valid (although for vector scale, the length of output channels should be provided).
Maybe we could use a single kernel to support both scalar and vector scales.

```C++
void qmul_kernel(int8_t* lhs, int8_t* rhs, int8_t* res, unsigned int data_len,
                 float* lhs_scale, float* rhs_scale, float* res_scale, unsigned int scale_len)
```

skykongkong8: Sounds even better. I will refer to that.

nntrainer/tensor/char_tensor.cpp
```diff
@@ -228,6 +267,56 @@ void CharTensor::initialize(Initializer init) {
...
+    << ((bool)(this->getFormat()) ? "NHWC" : "NCHW") << " is not match. ("
+    << ((bool)(input.getFormat()) ? "NHWC" : "NCHW") << ")";
+
+  NNTR_THROW_IF(!contiguous || !input.getContiguous() ||
+                  !output.getContiguous(),
+                std::invalid_argument)
+    << getName() << " is not contiguous, cannot multiply";
+
+  float lhs_scale = *(float *)getScale();
+  float rhs_scale = *input.getScale<float>();
```
skykongkong8, line 291: One more thing to check...
Could I confirm whether there are no plans to incorporate zero point as a qParam for QInt8 Tensors at this moment?
I could find some computational kernels that uses zero points for both int8, and uint8 so I just want to make sure where we heading to.

baek2sm (approved):
Nice work. LGTM

dkjung (approved):
LGTM

PR #2852, opened 2025-01-02
[FSU][Bugfix] Bugfix on FSU Inference

DonghakPark:
at case of inference there are some bug
- when set lifespan to tensor swap & non-swap value is exchanged
- set swap case - forward_func_lifespan
- set non-swap case - forward_infer_lifespan

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -381,9 +381,9 @@ std::vector<Weight *> Manager::requestWeights(
     var_ls = TensorLifespan::MAX_LIFESPAN;
   } else {
     if (enable_swap) {
-      var_ls = TensorLifespan::FORWARD_INFER_LIFESPAN;
-    } else {
       var_ls = TensorLifespan::FORWARD_FUNC_LIFESPAN;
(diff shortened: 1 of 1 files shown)


myungjoo (commented):
Need a unit test case that verifies this fix.

Reason 1: prevent regression
Reason 2: verify that this patch is really required for bugfix. (or a case that verifies that this is a bug)

myungjoo (approved):
LGTM. But please consider a unit test case for this.

DonghakPark:
> Need a unit test case that verifies this fix.
> 
> Reason 1: prevent regression Reason 2: verify that this patch is really required for bugfix. (or a case that verifies that this is a bug)

good point !!

As suggested, I'll make unit test cases for accurate operation verification. 
more detail explain : in the case of INFER_LIFESPAN, all weights are set at once for speed purposes. However, since only the necessary amount needs to be configured during swap option enable, I reported this as a bug.

EunjuYang (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #2854, opened 2025-01-03
[Wait for #2846][FSU] Modify Load Logic at FSU @open sesame 01/07 09:50

DonghakPark:
The process of loading trained weights during inference in NNTrainer currently uses model->load(). (before call inference)

However, since FSU loads weights from flash memory according to the required layers up to the specified look-ahead value, we need to prevent weights from being loaded before inference.

To reflect this change, I modified the code in LayerNode and added a new parameter called swap_parm.

commit summary
commit 1 [[Application] Update FSU SimpleFC Application](https://github.com/nnstreamer/nntrainer/commit/c0569b52f0ceb4f54de85329fec028e279f1f4e2) : update SimpleFC Application for more easy test & actual environment
commit 2 [[FSU] update layer weight load logic at fsu](https://github.com/nnstreamer/nntrainer/commit/950c5d50efdb2a038d5bc228b11effb3e660d3c3) : update layernode not to load weight at load()

**Self evaluation:**

### DIFF
--- a/Applications/SimpleFC/jni/main.cpp
+++ b/Applications/SimpleFC/jni/main.cpp
@@ -15,17 +15,13 @@
 #include <iostream>
 #include <memory>
 #include <sstream>
+#include <fstream>
 #include <vector>
-
-#if defined(ENABLE_TEST)
-#include <gtest/gtest.h>
-#endif
 
 #include <layer.h>
 #include <model.h>
 #include <optimizer.h>
-
-#include <cifar_dataloader.h>
+#include <unistd.h>
 
 #ifdef PROFILE
 #include <profiler.h>
@@ -34,11 +30,6 @@
 using LayerHandle = std::shared_ptr<ml::train::Layer>;
 using ModelHandle = std::unique_ptr<ml::train::Model>;
 
-using UserDataType = std::unique_ptr<nntrainer::util::DataLoader>;
-
-/** cache loss values post training for test */
-float training_loss = 0.0;
-float validation_loss = 0.0;
 
 /**
  * @brief make "key=value" from key and value
@@ -84,16 +75,18 @@ std::vector<LayerHandle> createGraph() {
 
   std::vector<LayerHandle> layers;
 
-  layers.push_back(createLayer(
-    "input", {withKey("name", "input0"), withKey("input_shape", "1:1:320")}));
+  layers.push_back(
+    createLayer("input", {withKey("name", "input0"),
+                          withKey("input_shape", "1:1024:1440")}));
 
+  for (int i = 0; i < 5; i++) {
+    layers.push_back(createLayer(
+      "fully_connected",
+      {withKey("unit", 1440), withKey("weight_initializer", "xavier_uniform"),
+       withKey("bias_initializer", "zeros")}));
+  }
   layers.push_back(createLayer("fully_connected",
-                               {withKey("unit", 100),
-                                withKey("weight_initializer", "xavier_uniform"),
-                                withKey("bias_initializer", "zeros")}));
-
-  layers.push_back(createLayer("fully_connected",
-                               {withKey("unit", 100),
+                               {withKey("unit", 1),
                                 withKey("weight_initializer", "xavier_uniform"),
                                 withKey("bias_initializer", "zeros")}));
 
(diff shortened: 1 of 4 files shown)


nntrainer/layers/layer_node.cpp
```diff
@@ -519,24 +519,21 @@ void LayerNode::read(std::ifstream &file, bool opt_var,
...
+           * training, bn weight type is fixed with full precsion */
+
+          TensorDim dim = run_context->getWeight(i).getDim();
+          dim.setDataType(this->getWeightDataType());
+          Tensor T_read(dim, true);
+          T_read.read(file);
+          run_context->getWeight(i).copyData(T_read);
         } else {
-          run_context->getWeight(i).read(file);
+          if (!swap)
```
jijoongmoon, line 535: It might be better to leave the comment saying we do need to check the read for training. This hot fixes is temperral. We also need to check skipping read is ok for the FSU case.

DonghakPark: :+1:  i will leave comment on #2906.
Thank you for Review

skykongkong8 (approved):
Clear explanation really helped my understanding! Looks good to me👍

Applications/SimpleFC/jni/main.cpp
```diff
@@ -182,87 +149,68 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
...
-  return {std::move(train_data), std::move(valid_data)};
-}
+  // model->summarize(std::cout, ML_TRAIN_SUMMARY_MODEL);
 
-std::array<UserDataType, 2>
-createRealDataGenerator(const std::string &directory, unsigned int batch_size,
-                        unsigned int data_split) {
+  answer = model->inference(1, in, l);
+  std::cout << answer[0][0] << std::endl;
+  auto end = std::chrono::system_clock::now();
```
EunjuYang, line 171: What about changing the order of these code lines?

Applications/SimpleFC/jni/main.cpp
```diff
@@ -182,87 +149,68 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
...
-  UserDataType train_data(new nntrainer::util::RandomDataLoader(
-    {{batch_size, 1, 1, 320}}, {{batch_size, 1, 1, 100}},
-    simulated_data_size / data_split));
-  UserDataType valid_data(new nntrainer::util::RandomDataLoader(
-    {{batch_size, 1, 1, 320}}, {{batch_size, 1, 1, 100}},
-    simulated_data_size / data_split));
-
-  return {std::move(train_data), std::move(valid_data)};
-}
+  // model->summarize(std::cout, ML_TRAIN_SUMMARY_MODEL);
```
EunjuYang: It would be great if we could check the time here to separately measure warm-up time and computation time.

Applications/SimpleFC/jni/main.cpp
```diff
@@ -158,14 +156,16 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
   std::string filePath = "./simplefc_weight_fp16_fp16_100.bin";
   if (access(filePath.c_str(), F_OK) == 0) {
     model->load(filePath);
-
+    auto load_end = std::chrono::system_clock::now();
+    std::chrono::duration<double> load_elapsed_seconds = load_end - start;
+    std::time_t load_end_time = std::chrono::system_clock::to_time_t(load_end);
+    std::cout << "Load finished computation at " << std::ctime(&load_end_time)
+              << "elapsed time: " << load_elapsed_seconds.count() << "s\n";
```
EunjuYang, line 163: For debugging, it is okay to print this log in the middle. For performance measurement, it might be better to skip this cout :)

EunjuYang (approved):
LGTM! Thank you for the detailed explanations of the updates.
One question from my side. I your update, `load()` process is skipped when it is called at inference mode with swap=true. Then, a kinda warm-up (i.e., pre-loading for look-ahead numbers layers) is conducted when forward is called. Am I correct?

Applications/SimpleFC/jni/main.cpp
```diff
@@ -155,25 +132,13 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
...
-
-  // to test asynch fsu, we do need save the model weight data in file
-  model->save("simplefc_weight_fp16_fp16_100.bin",
-              ml::train::ModelFormat::MODEL_FORMAT_BIN);
-  model->load("./simplefc_weight_fp16_fp16_100.bin");
-
-  model->summarize(std::cout, ML_TRAIN_SUMMARY_MODEL);
-
-  unsigned int feature_size = 320;
+  uint feature_size = 1 * 1024 * 1440;
```
djeong20, line 135: let's not use `uint` as in #2890
```suggestion
  unsigned int feature_size = 1 * 1024 * 1440;
```

Applications/SimpleFC/jni/main.cpp
djeong20:
```suggestion
#ifdef DEBUG
    std::cout << "Load finished computation at " << std::ctime(&load_end_time)
              << "elapsed time: " << load_elapsed_seconds.count() << "s\n";
#endif
```

DonghakPark:
This commit merged by https://github.com/nnstreamer/nntrainer/pull/2906 with some change

PR #2857, opened 2025-01-07
[ neon ] Implement int8 mul neon simd kernel @open sesame 01/09 12:38

skykongkong8:
This PR proposes a accelerated function proposed from #2850
This will not affect the function immediately, but will introduce PR for that in the near future.

- Note that current int8 Tensor does not consider zero point as a qParam.
- Multiply with 16 to 32 bit widening, 32 bit scaling, and min-max saturation.
- Use nearest rounding for the result.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_neon.cpp
+++ b/nntrainer/tensor/blas_neon.cpp
@@ -626,6 +626,81 @@ void custom_scopy(const unsigned int N, const float *X, const int incX,
   }
 }
 
+void ele_qmul(int8_t *lhs, int8_t *rhs, int8_t *res, unsigned int data_len,
+              const float *lhs_scale, const float *rhs_scale,
+              const float *res_scale, unsigned int scale_len) {
+  if (scale_len == 1) {
+    return __ele_qmul_kernel(lhs, rhs, res, data_len, lhs_scale[0],
+                             rhs_scale[0], res_scale[0]);
+  } else {
+    return __ele_qmul_kernel(lhs, rhs, res, data_len, lhs_scale, rhs_scale,
+                             res_scale, scale_len);
+  }
+}
+
+void __ele_qmul_kernel(int8_t *lhs, int8_t *rhs, int8_t *res,
+                       unsigned int data_len, const float lhs_scale,
+                       const float rhs_scale, const float res_scale) {
+  float32x4_t multiplier = vdupq_n_f32(lhs_scale * rhs_scale / res_scale);
+  int8x16_t int8_max = vdupq_n_s8(127);
+  int8x16_t int8_min = vdupq_n_s8(-128);
+  unsigned int N16 = (data_len >> 4) << 4;
+  for (unsigned int n = 0; n < N16; n += 16) {
+    int16x8_t lhs0_7 = vmovl_s8(vld1_s8(lhs));
+    lhs += 8;
+    int16x8_t lhs8_15 = vmovl_s8(vld1_s8(lhs));
+    lhs += 8;
+    int16x8_t rhs0_7 = vmovl_s8(vld1_s8(rhs));
+    rhs += 8;
+    int16x8_t rhs8_15 = vmovl_s8(vld1_s8(rhs));
+    rhs += 8;
+
+    int32x4_t res0_3 = vmull_s16(vget_low_s16(lhs0_7), vget_low_s16(rhs0_7));
+    int32x4_t res4_7 = vmull_s16(vget_high_s16(lhs0_7), vget_high_s16(rhs0_7));
+    int32x4_t res8_11 = vmull_s16(vget_low_s16(lhs8_15), vget_low_s16(rhs8_15));
+    int32x4_t res12_15 =
+      vmull_s16(vget_high_s16(lhs8_15), vget_high_s16(rhs8_15));
+
+    float32x4_t res_f32_0 = vcvtq_f32_s32(res0_3);
+    float32x4_t res_f32_1 = vcvtq_f32_s32(res4_7);
+    float32x4_t res_f32_2 = vcvtq_f32_s32(res8_11);
+    float32x4_t res_f32_3 = vcvtq_f32_s32(res12_15);
+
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/blas_neon.cpp
```diff
@@ -626,6 +626,81 @@ void custom_scopy(const unsigned int N, const float *X, const int incX,
...
+    output = vminq_s8(output, int8_max);
+
+    vst1q_s8(res, output);
+    res += 16;
+  }
+  for (unsigned int n = N16; n < data_len; ++n) {
+    res[n] = std::max(
+      -128, std::min(127, (int)std::lround(lhs[n] * rhs[n] * lhs_scale *
+                                           rhs_scale / res_scale)));
+  }
```
EunjuYang, line 694: I cannot understand this block. Why does it overwrite the `res` array? I guess `res[n]` should be on the RHS.

nntrainer/tensor/blas_neon.h
```diff
@@ -183,6 +183,55 @@ bool isValid(const size_t N, const float *input);
...
+ * @param res  int8_t * for Vector res (output)
+ * @param data_len length of the vector
+ * @param lhs_scale scale factor of the lhs vector
+ * @param rhs_scale scale factor of the rhs vector
+ * @param res_scale scale factor of the result vector
+ * @param scale_len length of the scale factor
+ */
+void ele_qmul(int8_t *lhs, int8_t *rhs, int8_t *res, unsigned int data_len,
+              float *lhs_scale, float *rhs_scale, float *res_scale,
+              unsigned int scale_len);
```
EunjuYang, line 201: What about declaring the `lhs_scale`, `rhs_scale` and `res_scale` as const pointer?
```suggestion
void ele_qmul(int8_t *lhs, int8_t *rhs, int8_t *res, unsigned int data_len,
              const float *lhs_scale, const float *rhs_scale, const float *res_scale,
              unsigned int scale_len);
```

nntrainer/tensor/blas_neon.cpp
```diff
@@ -626,6 +626,81 @@ void custom_scopy(const unsigned int N, const float *X, const int incX,
...
+void ele_qmul(int8_t *lhs, int8_t *rhs, int8_t *res, unsigned int data_len,
+              float *lhs_scale, float *rhs_scale, float *res_scale,
+              unsigned int scale_len) {
+  if (scale_len == 1) {
+    return __ele_qmul_kernel(lhs, rhs, res, data_len, lhs_scale[0],
+                             rhs_scale[0], res_scale[0]);
+  } else {
+    return __ele_qmul_kernel(lhs, rhs, res, data_len, lhs_scale, rhs_scale,
+                             res_scale, scale_len);
+  }
```
EunjuYang, line 638: simple question. do you think there's more room if we separately define the kernel for the scalar case? if not, we might consider removing this condition after implementing the vector scale case.

EunjuYang (commented):
Happy to read the PR. Here're some comments from my side.

nntrainer/tensor/blas_neon.cpp
skykongkong8: I can't understand your point.. res is a resultant output, so shouldn't it be on the LHS?

nntrainer/tensor/blas_neon.h
skykongkong8: Sounds better!
AFAIK, current char tensor scale factor is stored like `std::vector<float> const &scales`, so should be fixed that way. Thanks!

nntrainer/tensor/blas_neon.cpp
EunjuYang: Oh I misunderstood. I thought it is overwrite the value, while it tries to clilp. It seems the code compute the remaining part. Sorry for confusion :)

nntrainer/tensor/blas_neon.cpp
skykongkong8: Do you mean is there a room for a unified kernel that can handle both the scalar and vector cases?
It is not impossible, but it will deteriorate the kernel functionality. For example, same scale factor vector will be loaded redundantly while scale factor can be pre-loaded in the scalar-only kernel like : `float32x4_t multiplier = vdupq_n_f32(lhs_scale * rhs_scale / res_scale);`

EunjuYang: I see, so you're saying that if we know it's a scalar, using a scalar-specific kernel would be better? In that case, it might be wiser to proceed as PRed. Thanks for your response!

EunjuYang (approved):
LGTM!

nntrainer/tensor/blas_neon.cpp
```diff
@@ -626,6 +626,81 @@ void custom_scopy(const unsigned int N, const float *X, const int incX,
...
+    float32x4_t res_f32_1 = vcvtq_f32_s32(res4_7);
+    float32x4_t res_f32_2 = vcvtq_f32_s32(res8_11);
+    float32x4_t res_f32_3 = vcvtq_f32_s32(res12_15);
+
+    res_f32_0 = vmulq_f32(res_f32_0, multiplier);
+    res_f32_1 = vmulq_f32(res_f32_1, multiplier);
+    res_f32_2 = vmulq_f32(res_f32_2, multiplier);
+    res_f32_3 = vmulq_f32(res_f32_3, multiplier);
+
+    /// @note: currently we use vcvtnq_s32_f32 instead of vcvtq_s32_f32
```
djeong20, line 674: could you explain this part?

djeong20 (approved):
Awesome 👍

nntrainer/tensor/blas_neon.cpp
skykongkong8: vcvtq_s32_f32 : round to zero
vcvtnq_s32_f32 : round to nearest

As you might already know, there are many rounding rules during quantized value quantization (stochasitc rounding, ... ), but AFAIK rounding to the nearest is the most typical one.

skykongkong8:
Should be raised again after #2930

PR #2859, opened 2025-01-07
[OpenCL/GPU] Optimized Blas and Attention kernels with the latest GPU Pipeline.

yashSingh0723:
- Upated the kernels as per the latest buffer generalized changes for both `FP32 `and `FP16`.
- Added unittest for Addition Kernel FP16 in `unittest_blas_kernels_cl.cpp`.
- `add_i `and `rotary_emb` ops are updated.

Signed-off-by: Yash Singh [yash.singh@samsung.com](mailto:yash.singh@samsung.com)

### DIFF
--- a/nntrainer/tensor/cl_operations/attention_kernel_strings.h
+++ b/nntrainer/tensor/cl_operations/attention_kernel_strings.h
@@ -15,7 +15,8 @@
 #define __ATTENTION_KERNEL_STRINGS_H__
 
 #include <string>
-
+// unsigned int offsetFeqsSin,
+//                                       unsigned int offsetSin
 namespace nntrainer {
 static const std::string rotary_emb_cl_kernel_ = R"(
 
@@ -34,10 +35,11 @@ __kernel void rotary_emb_cl(__global float *input,
                                       unsigned int dim,
                                       unsigned int half_,
                                       unsigned int max_timestep,
-                                      unsigned int from) {
+                                      unsigned int from,
+                                      unsigned int offsetFreqsSin,
+                                      unsigned int offsetSin) {
     __global float *cos_ptr = cos_;
     __global float *sin_ptr = sin_;
-
     float value = 0.0f;
     float transformed_value = 0.0f;
 
@@ -50,7 +52,7 @@ __kernel void rotary_emb_cl(__global float *input,
           unsigned idx = (from + h)*dim;
           for(unsigned int i = idx; i < idx + dim; i++){
             cos_ptr[i - idx] = freqs_cos[i];
-            sin_ptr[i - idx] = freqs_sin[i];
+            sin_ptr[i - idx + offsetSin] = freqs_sin[i + offsetFreqsSin];
           }
         }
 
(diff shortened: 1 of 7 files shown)


nntrainer/tensor/cl_operations/attention_kernel_strings.h
```diff
@@ -15,7 +15,8 @@
 #define __ATTENTION_KERNEL_STRINGS_H__
 
 #include <string>
-
+// unsigned int offsetFeqsSin,
+//                                       unsigned int offsetSin
```
djeong20, line 19:
```suggestion
```
let's remove it

nntrainer/tensor/cl_operations/attention_kernel_strings.h
```diff
@@ -50,7 +52,7 @@ __kernel void rotary_emb_cl(__global float *input,
           unsigned idx = (from + h)*dim;
           for(unsigned int i = idx; i < idx + dim; i++){
             cos_ptr[i - idx] = freqs_cos[i];
-            sin_ptr[i - idx] = freqs_sin[i];
+            sin_ptr[i - idx + offsetSin] = freqs_sin[i + offsetFreqsSin];
```
djeong20, line 55: could you explain this part?
1. why offsetSin and offsetFreqsSin (`cos_.size()` and `freqs_cos.size() * dim`) are added 
2. what is the intended behavior in this change?

Also, wouldn't this result in accessing invalid memory space for freqs_sin?

test/unittest/unittest_blas_kernels_cl.cpp
```diff
@@ -581,6 +581,59 @@ TEST(blas_kernels, addition_i) {
...
+  int height = 26;
+  int width = 26;
+
+  int batch_b = 1;
+
+  const float alpha = 1e-1;
+  const int MOD = 10;
+
+  // nntrainer::TensorDim::TensorType t_type_nchw_fp32 = {
+  //   nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP32};
```
djeong20, line 597:
```suggestion
```

djeong20 (approved):
Overall, LGTM

nntrainer/tensor/cl_operations/attention_kernel_strings.h
yashSingh0723: Hello, so as per the latest GPU pipeline changes we are using a genralized set if buffers instead of creating buffers everytime whenever we a kernel is called. As of now there are only 5 generalized buffers, 3 for input buffers and 2 for output buffers. 
As i need 5 input buffers so I am using an offset for both `freqs_sin` and `freqs_sin_flat`. Thats why in the code as well I've used the offset for both. 
Example, lets say there is bufferA of size 500, so from my 1-100, I am storing freqs_cos and from 100-200 I am storing freqs_sin, so when using the data of freqs_sin, I'll have to use an offset of 100 and that is what I am doing here.

Please refer this PR for more understanding: https://github.com/nnstreamer/nntrainer/pull/2816

nntrainer/tensor/cl_operations/attention_kernel_strings.h
yashSingh0723: I'll update in the latest commit.

test/unittest/unittest_blas_kernels_cl.cpp
yashSingh0723: I'll update in the next commit. Thanks for pointing it out.

nntrainer/tensor/cl_operations/attention_kernel_strings.h
djeong20: Thank you for the clarification!

dkjung:
@yashSingh0723 Could you rebase this PR?

PR #2858, opened 2025-01-07
[Tensor] Proposal for unifying tensor addition: remove add_i implementation

djeong20:
This pull request proposes the removal of the add_i implementation and its unification with the add function.
Currently, the use of two distinct kernels for tensor addition can result in performance inefficiencies.
To enhance overall performance, it is crucial to address this issue.
Additionally, we must undertake follow-up work to integrate ele_add and sxapy successfully.

**Self-evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test:   [X]Passed [ ]Failed [ ]Skipped

PR #2862, opened 2025-01-08
[ Tensor ] Integrate saxpy and ele_add

skykongkong8:
- This PR resolves the issue proposed from #2858
- Integrate for better maintenance and functionality from the Tensor level
- Add alpha, beta, gamma constant to differentiate the operation
- note :
    - axpy : a * x plus y
    - xpaypbz : x plus a * y plus b * z

- AS-IS
    1. ele_add : z = x + a * y + b * z
    2. axpy : y = a * x + y
- TO-BE
    1. ele_add : z = a * x + b * y + c * z
    2. axpy : y = a * x + y ( z = b * y + z from current fix )
    3. xpaypbz : z = x + a * y + b * z

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/blas_interface.cpp
+++ b/nntrainer/tensor/blas_interface.cpp
@@ -452,11 +452,11 @@ static void ele_mul_fallback(const unsigned int N, const _FP16 *X,
   }
 }
 
-static void ele_add_fallback(const unsigned int N, const _FP16 *X,
-                             const _FP16 *Y, _FP16 *Z, float alpha, float beta,
+static void xpbypcz_fallback(const unsigned int N, const _FP16 *X,
+                             const _FP16 *Y, _FP16 *Z, float beta, float gamma,
                              unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X + static_cast<_FP16>(alpha) * *Y + static_cast<_FP16>(beta) * *Z;
+    *Z = *X + static_cast<_FP16>(beta) * *Y + static_cast<_FP16>(gamma) * *Z;
     X += o_stride;
     Y += i_stride;
     Z += o_stride;
@@ -498,17 +498,30 @@ void ele_mul(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
     ele_mul_fallback(N, X, Y, Z, alpha, beta, i_stride, o_stride);
 }
 
-void ele_add(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
-             float alpha, float beta, unsigned int i_stride,
+void xpbypcz(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
+             float beta, float gamma, unsigned int i_stride,
              unsigned int o_stride) {
   if (i_stride == 1 && o_stride == 1) {
 #if (defined USE__FP16 && USE_NEON)
-    nntrainer::neon::ele_add(N, X, Y, Z, alpha, beta);
+    nntrainer::neon::xpbypcz(N, X, Y, Z, beta, gamma);
 #else
-    ele_add_fallback(N, X, Y, Z, alpha, beta, i_stride, o_stride);
+    xpbypcz_fallback(N, X, Y, Z, beta, gamma, i_stride, o_stride);
 #endif
   } else
-    ele_add_fallback(N, X, Y, Z, alpha, beta, i_stride, o_stride);
+    xpbypcz_fallback(N, X, Y, Z, beta, gamma, i_stride, o_stride);
+}
+
+void ele_add(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
+             float alpha, float beta, float gamma, unsigned int i_stride,
+             unsigned int o_stride) {
+  if (std::fpclassify(alpha) == FP_ZERO && gamma == 1.F) {
+    return saxpy(N, beta, Y, i_stride, Z, o_stride);
+  } else if (alpha == 1.F) {
+    return xpbypcz(N, X, Y, Z, beta, gamma, i_stride, o_stride);
+  } else {
+    std::invalid_argument("Addition for 3 term with non-unit alpha and 2 term "
+                          "with non-unit gamma is NYI");
+  }
 }
 
 void ele_sub(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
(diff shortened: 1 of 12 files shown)


skykongkong8:
PTAL @djeong20

jijoongmoon (approved):
LGTM

nntrainer/tensor/blas_interface.cpp
```diff
@@ -498,17 +498,30 @@ void ele_mul(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
...
-    ele_add_fallback(N, X, Y, Z, alpha, beta, i_stride, o_stride);
+    xpbypcz_fallback(N, X, Y, Z, beta, gamma, i_stride, o_stride);
+}
+
+void ele_add(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
+             float alpha, float beta, float gamma, unsigned int i_stride,
+             unsigned int o_stride) {
+  if (std::fpclassify(alpha) == FP_ZERO && gamma == 1.F) {
+    return saxpy(N, beta, Y, i_stride, Z, o_stride);
+  } else if (alpha == 1.F) {
```
djeong20, line 519: let's avoid using operator== for floating point numbers.

```suggestion
  if (std::fpclassify(alpha) == FP_ZERO && std::fpclassify(gamma- 1.f) == FP_ZERO) {
    return saxpy(N, beta, Y, i_stride, Z, o_stride);
  } else if (std::fpclassify(alpha - 1.f) == FP_ZERO) {
```

nntrainer/tensor/blas_interface.cpp
```diff
@@ -1268,4 +1257,17 @@ bool is_valid(const size_t N, ml::train::TensorDim::DataType d_type,
   return false;
 }
 
+void ele_add(const unsigned int N, const float *X, const float *Y, float *Z,
+             float alpha, float beta, float gamma, unsigned int i_stride,
+             unsigned int o_stride) {
+  if (std::fpclassify(alpha) == FP_ZERO && gamma == 1.F) {
+    return saxpy(N, beta, Y, i_stride, Z, o_stride);
+  } else if (alpha == 1.F) {
+    return xpbypcz(N, X, Y, Z, beta, gamma, i_stride, o_stride);
```
djeong20, line 1266: Here too.

nntrainer/tensor/blas_interface.h
```diff
@@ -162,20 +162,36 @@ void ele_mul(const unsigned int N, const _FP16 *X, const _FP16 *Y, _FP16 *Z,
...
  * @param[in] N  length of the vector
  * @param[in] X _FP16 * for Vector X
  * @param[in] Y _FP16 * for Vector Y
  * @param[in] Z _FP16 * for Vector Z
- * @param[in] alpha scalar multiplier for input
- * @param[in] beta scalar multiplier for output
+ * @param[in] beta scalar multiplier for input
+ * @param[in] gamma scalar multiplier for output
  * @param[in] i_stride input stride
  * @param[in] o_stride output stride
```
djeong20, line 174: let's clarify whether the input is X or Y.

```suggestion
 * @param[in] beta scalar multiplier for input Y
 * @param[in] gamma scalar multiplier for output
 * @param[in] i_stride input Y stride
 * @param[in] o_stride output stride
```

nntrainer/tensor/blas_neon.cpp
```diff
@@ -492,29 +492,29 @@ void ele_mul(const unsigned int N, const float *X, const float *Y, float *Z,
...
   unsigned int i = 0;
-  float32x4_t alpha_vec = vdupq_n_f32(alpha);
   float32x4_t beta_vec = vdupq_n_f32(beta);
+  float32x4_t gamma_vec = vdupq_n_f32(gamma);
   for (; N - i >= 4; i += 4) {
     float32x4_t x0_3 = vld1q_f32(&X[i]);
     float32x4_t y0_3 = vld1q_f32(&Y[i]);
-    if (alpha != 1.f) {
-      y0_3 = vmulq_f32(y0_3, alpha_vec);
+    if (beta != 1.f) {
```
djeong20, line 503:
```suggestion
    if (std::fpclassify(beta - 1.f) != FP_ZERO) {
```

nntrainer/tensor/tensor.cpp
```diff
@@ -576,10 +576,11 @@ int Tensor::add_i_partial(unsigned int len, unsigned int addr_idx, Tensor &m,
 
 Tensor Tensor::add(Tensor const &m, float const alpha) const {
   Tensor t("", getFormat(), getDataType());
-  return this->add(m, t, alpha);
+  return this->add(m, t, 1.F, alpha, 0.F);
 }
 
-Tensor &Tensor::add(Tensor const &m, Tensor &output, float const alpha) const {
+Tensor &Tensor::add(Tensor const &m, Tensor &output, float const alpha,
+                    float const beta, float const gamma) const {
```
djeong20, line 583: I don't believe it would be beneficial to add additional parameters (beta and gamma) at the tensor level unless there is a specific use case for these features. Furthermore, the introduction of these new parameters would create inconsistencies, as we currently have multiple addition functions within the tensor class. Instead, how about we maintain a consistent tensor interface and pass in the default values for each tensor class?

djeong20: https://pytorch.org/docs/stable/generated/torch.add.html

nntrainer/tensor/blas_interface.cpp
skykongkong8: Got it, but all blas-level fixes should be applied after merging #2930 unless super-urgent for now. Let me fix this afterwards.

nntrainer/tensor/tensor.cpp
skykongkong8: I agree that this might cause unexpected complexity that we do not really have to care about in the actual use case.
Then, however, how should we integrate

1. ele_add : z = x + a * y + b * z
2. axpy : y = a * x + y

in a single general function with out using z = a*x + b*y + c*z?

djeong20: I believe we should avoid performing these operations as a single function; instead, they should be executed one at a time. In my opinion, it's important to separate them from the perspective of ONNX. While combining these operations into the expression (z = ax + by + cz) may have its advantages, I question whether this format aligns with ONNX standards and reflects the intention of the nntrainer::AddLayer. @baek2sm, could you please check if \(z = ax + by + c*z\) is considered a valid single operation in ONNX?

baek2sm: @djeong20 @skykongkong8 In my opinion, the ONNX representation scheme seems to pursue generality rather than computational efficiency. In ONNX, `Add` is expressed extremely simply. 

`input: A, B, output: C, so C = A + B.`

Even the scale parameter that exists in PyTorch has restrictions when exporting to ONNX format.
The alpha value must be a constant tensor or a number to be exported, and after export, the alpha value appears to be merged with the B value (A=A, B=alpha*B).

Therefore, it does not seem appropriate to follow the optimization criteria for tensor operations according to the ONNX representation scheme.

However, unless the addition of `beta and gamma` values provides computational advantages in many general cases, wouldn't it be beneficial from a management perspective to maintain the most widely used PyTorch representation scheme as a standard(only support alpha)? (This is just a personal opinion.)

skykongkong8: I see. Even if we integrate them in a most-generalized form(z = ax + by + cz), 1. such format is barely used in real cases, 2. lower-level kernel should be implemented explicitly for each. Thus, I believe it following the PyTorch format makes more sense. Although we have `beta` option in current `Add`, AFAIK there is no Tensor op that directly calls it.
Thereafter, how about just closing this PR and stick to current implementation? I don't see actual benefit by integrating saxpy and ele_add. @djeong20

dkjung:
Is this going to be close?

skykongkong8:
I think enough discussions were already made, and we can come back to this issue later on.

PR #2865, opened 2025-01-10
[Subgraph] Implement `SubGraphBase`/`SubGraphCpu` to update `NetworkGraph`; adding `graph_scope` property

### LINKED ISSUE #2864: [Subgraph] Enabling a feature to manage a graph in subgraph level.
**Abstract**

To accommodate heterogeneous compute engines, it becomes necessary to separately administer distinct graph regions — referred to as *subgraphs* — each tailored to function according to specific compute engines. This approach differs from our current method of managing only one graph solely at the layer level; therefore, adding a new tier labeled 'subgraph' would allow us to structure our system hierarchically as follows: Graph > Subgraphs > Layers.

**Requirements for Subgraphs**

1. Each subgraph type should be able to define its own operations (e.g., forwarding, backwarding).
   - Currently, NNTrainer uses GraphCore to manage layer information within a graph, and the operations (e.g., forwarding, backwarding) in this graph are given from higher levels of the network.
    - These functions need to be modified so they can be managed at the subgraph level.
2. Subgraphs should be able to define their operations based on the compute engine type (e.g., CPU, NPU, GPU).
3. Both topological sorting at the subgraph level and at the layer level are needed.
4. Networks should be saved separately for each subgraph.
5. This feature should support backward compatibility.
6. Various subgraph type should be pluggable 

**Idea to implement**

- In order to fulfill the requirements, I'm planning to implement base class for subgraph that inherits `GraphNode`. By doing so, we can manage subgraphs with NetworkCore, which supports topological sort and node management.
- As an interface to add a subgraph, I'm planning to add a new property used when `createLayer` is called, with a default value.

EunjuYang:
Dear reviewers, 
- Sorry for the large PR.
- Please See #2864

**Abstract**
This PR aims to suggest the first draft to resolve #2864.
This PR implements `SubGraphBase` which is a virtual class with `GraphCore subgraph`; the `subgraph` manages the layers included in the subgraph. Also, This PR updats `NetworkGraph` to manage multiple subgraph usng `GraphCore graph`. To this end, it implemented the `SubGraphBase` to inherit `GraphNode`.
Please note that, this draft only considers to implement the skeleton, so it doesn't fully consider to support multiple subgraphs. The feature will be added in later PR. However, it would be happy to listen your idea on managing multiple subgraphs.

**List of commits**
- 4d66ea96dd8351e087bc6a5087425c4edebe9806:  Refactor: Move Defined Behavior from NetworkGraph to SubGraphBase
    - Moved the defined behavior from `NetworkGraph` to `SubGraphBase`.
    - Updated public methods of `NetworkGraph` to pass their behaviors to `SubGraphBase`.
    - Removed unused private methods in `NetworkGraph`.
- 708475e97d58169799ba975dd41e86bea6c6f4b2: Move forwarding/incremental_forwarding op from neuralnet to SubGraphBase
    - Update the definition of forwarding operations in neuralnet to be defined within each subgraph type.
    - Modify the implementation so that both forwarding and incremental_forwarding can now be defined individually for each subgraph type.
- d4c99223e8f1426913ba7e2c2e04216152d5789f: Move backwarding op from neuralnet to SubGraphBase
    - Update the definition of backwarding operations in neuralnet to be defined within each subgraph type.
    - `applyGradients` is also updated without taking any callback from
  outside.
- 66bd5b55477b488f15f2432b05bedca437265262: Update type of graph in NetworkGraph and relevant unit tests
    - Updated `SubGraphBase` to inherit `GraphNode`
    - Changed the `graph` type inside `NetworkGraph` from `SubGraphBase` to `GraphCore`, using `SubGraphBase` as its `GraphNode` type.
    - Allowed for the addition of multiple `SubGraphBase` graph nodes in the future.
    - Assumed that NetworkGraph always has a `default_subgraph` called default_subgraph, and updated the existing code to maintain compatibility.
    - Updated the relevant unit test codes.
- 3c4aecd32263164e76e42f0c5e9afd5522499d30: Split Input Setup and Label Setup for Subgraphs
    - Refactored the input and label setup logic for subgraphs by splitting them into separate functions: `setInputs()` and `setLabels()`.
    - Introduced new functions `setInputs()` and `setLabels()` to handle input and label setups independently for each subgraph.
- 17e0c8c24235f2c094c842c7fedcc9e542bbbeff: Update SubGraphBase to be virtual and Implement SubGraphCpu
    - Updated the base class `SubGraphBase` to become a virtual base class and implemented a derived class `SubGraphCpu`.
    - Set up `NetworkGraph` to default to creating and utilizing one instance of `SubGraphCpu`.
    - Declared several member functions related to memory management and computation engines as pure virtual functions (=0) in `SubGraphBase`.
- 87edf02a82b42e3db7644bf2b462d6b059a862e1: Add graph_scope property to designate subgraph of a layer
    - This commit adds a `graph_scope` property to designate the subgraph,
  which a layer belongs to.
    - This commit updates NetworkGraph's addLayer function to check the
  GraphName of the layer and create a subgraph if the graph doesn't
  exist.

**Self evaluation:**


skykongkong8 (approved):
If I understood correctly, this PR aims to move almost all the basic functionalities of networkGraph to some instances of subgraph, which can have different compute engine backends variants in order to govern them! (CPU, GPU, ... )
Detailed explanations per commit helped my understanding a LOT! Really appreciate it😁
Looks Good to me!

EunjuYang:
Please note that I've rebased the PR on main (2025-01-16 15:46 KST) !
Updated commits are 4d66ea96dd8351e087bc6a5087425c4edebe9806 and 17e0c8c24235f2c094c842c7fedcc9e542bbbeff, which were conflicted with #2872.

nntrainer/graph/network_graph.h
```diff
@@ -24,9 +25,15 @@
 #include <graph_core.h>
 #include <layer_node.h>
 #include <manager.h>
+#include <optimizer_wrapped.h>
+
+#include <subgraph_base.h>
+#include <subgraph_cpu.h>
```
jijoongmoon, line 31: I'm not sure whether we need to include subgraph_cpu.h here.

jijoongmoon, line 31: I'm not sure whether we need to include subgraph_cpu.h here.

EunjuYang: Thank you for the comment. First of all, in my opinion, we may not need to split subgraph_base and subgraph_cpu. I think the action on the CPU can be the subgraph_base's action. If so, we don't need to create `subgraph_cpu` separately, but merge it. What do you think? If so, let me remove the commits to split subgraph_base and subgraph_cpu.

About making subgraph pluggable, please let me make another PR if it is acceptable.

haehun (approved):
If I understood correctly, this pr implements subgraph and its CPU implementation.

dkjung (approved):
LGTM

EunjuYang:
Dear reviewers, thank you for reviewing this PR.
Based on your comments, I've updated the code (Please see #3018).
Just so that you know, the updated one is still under WIP.
In this context, it is inappropriate to merge this PR. Therefore, let me close this and make new PRs after the draft is completed. About the draft, I hope we can discuss it further in an offline meeting.

PR #2866, opened 2025-01-10
[Tensor] ShortTensor class with quantized signed 16-bit integer

djeong20:
In this PR, a new tensor type, the ShortTensor class, is designed explicitly for handling quantized signed 16-bit integer data types.
This new tensor class aims to give users more options when working with various data types.
Currently, the ShortTensor class does not support mathematical operations like multiplication or addition.
Additionally, more unit tests for the QINT16 type will be added in the later PRs.

**Self-evaluation:**


jijoongmoon (approved):
LGTM

nntrainer/tensor/short_tensor.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include <iostream>
+
+#include <blas_interface.h>
+#include <short_tensor.h>
+#include <tensor.h>
+
+namespace nntrainer {
+
+ShortTensor::ShortTensor(std::string name_, Tformat fm, QScheme qscheme_) :
+  TensorBase(name_, fm, Tdatatype::UINT16), qscheme(qscheme_) {}
```
EunjuYang, line 21:
```suggestion
  TensorBase(name_, fm, Tdatatype::QINT16), qscheme(qscheme_) {}
```

nntrainer/tensor/short_tensor.cpp
```diff
@@ -0,0 +1,448 @@
...
+    dim.setTensorDim(1, d[0].size());
+    dim.setTensorDim(2, d[0][0].size());
+    dim.setTensorDim(3, d[0][0][0].size());
+  } else {
+    dim.setTensorDim(2, d[0].size());
+    dim.setTensorDim(3, d[0][0].size());
+    dim.setTensorDim(1, d[0][0][0].size());
+  }
+
+  dim.setTensorType({fm, Tdatatype::UINT16});
```
skykongkong8, line 59:
```suggestion
  dim.setTensorType({fm, Tdatatype::QINT16});
```
Same here

EunjuYang, line 59: Ditto?
```suggestion
  dim.setTensorType({fm, Tdatatype::QINT16});
```

nntrainer/tensor/short_tensor.cpp
```diff
@@ -0,0 +1,448 @@
...
+  size_t feature_len = dim.getFeatureLen();
+
+  result.resize(batch_size);
+
+  for (unsigned int b = 0; b < batch_size; b++) {
+    auto max_iter =
+      std::max_element(data + b * feature_len, data + (b + 1) * feature_len);
+    result[b] = std::distance(data, max_iter) - (b * feature_len);
+  }
+  return result;
```
EunjuYang, line 317: It's a simple question. Doesn't it need to consider scale factor when computing `max` value?

nntrainer/tensor/short_tensor.cpp
djeong20: fixed! thanks :)

nntrainer/tensor/short_tensor.cpp
djeong20: fixed :)

nntrainer/tensor/short_tensor.cpp
djeong20: This is a tricky situation. I'm unsure if the argmax function needs to be implemented for the quantized tensor in the first place, especially since we typically use argmax for the output and label tensors, which are mostly floating-point tensors. What do you think about this?

nntrainer/tensor/short_tensor.h
```diff
@@ -0,0 +1,280 @@
...
+   *
+   * @param buf buffer to copy from
+   */
+  void copy(const void *buf);
+
+  /**
+   * @brief  Get the Data Type String object
+   * @return std::string of tensor data type (UINT16)
+   */
+  std::string getStringDataType() const override { return "UINT16"; }
```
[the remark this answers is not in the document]
djeong20: thanks!

nntrainer/tensor/short_tensor.cpp
```diff
@@ -0,0 +1,448 @@
...
+  // compare scale factors
+  const float *_scales = (float *)getScale();
+  const float *_rscales = (float *)rhs.getScale();
+  for (size_t i = 0; i < scale_size(); ++i) {
+    if (std::fabs(_scales[i] - _rscales[i]) > 1e-5)
+      return false;
+  }
+
+  return true;
+}
```
skykongkong8, line 116: Is there a specific reason for choosing 1e-5 as an epsilon here?
Otherwise, how about using `https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon` ?

skykongkong8 (approved):
Overall, LGTM!

nntrainer/tensor/short_tensor.cpp
EunjuYang: I see what you mean. For this version, returning the maximum value from the "data" region seems like a better option since `getData` or `getValue` gives us the *quantized* "data" values rather than the *dequantized* ones. Moreover, it's not certain that the tensor is strictly meant for quantized tensors alone. Besides, the present tensor interface lacks a provision to fetch *dequantized* values individually. Hence, syncing up argmax with this strategy looks feasible given the circumstance.

Nonetheless, scenarios could arise wherein utilizing the quantized tensor as an activation would require us to access the dequantized value and perform argmax / max based on those values eventually. If a model data type is given as `QINT8-QINT16`, then the last activation from the model might be the `QINT16` tensor I guess, which requires dequantized value in some sense.

EunjuYang (approved):
LGTM

nntrainer/tensor/short_tensor.cpp
djeong20: no particular reason. I follow to use epsilon from the tensor class.

https://github.com/nnstreamer/nntrainer/blob/d6d02c8ec8dd0304c4f3677142e7e2f20446dd7e/nntrainer/tensor/tensor.h#L1709

baek2sm (approved):
LGTM

PR #2869, opened 2025-01-13
[Windows] Avoid using dirent.h @open sesame 01/15 12:44

gkisalapl:
This PR start series of changes which should make possible compilation of nntrainer on windows platform.
Header dirent.h is not available on windows so to make code more portable I've replaced this functionality with std::filesystem

Self-evaluation:

### DIFF
--- a/nntrainer/tensor/char_tensor.cpp
+++ b/nntrainer/tensor/char_tensor.cpp
@@ -17,18 +17,18 @@
 
 namespace nntrainer {
 
-CharTensor::CharTensor(std::string name_, Tformat fm) :
+CharTensor::CharTensor(std::string name_, Tformat fm, QScheme qscheme_) :
   TensorBase(name_, fm, Tdatatype::QINT8) {}
 
 CharTensor::CharTensor(const TensorDim &d, bool alloc_now, Initializer init,
-                       std::string name) :
-  TensorBase(d, alloc_now, init, name) {
+                       std::string name, QScheme qscheme_) :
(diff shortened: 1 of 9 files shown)


jijoongmoon:
@gkisalapl , Thanks for the PR. It is good if NNTrainer can support Windows Operating System naturally and I'm sure you are working on it. Thanks aging. One thing I want to address is that this Windows Porting task shouldn't be only for windows. I mean this code change also should work in other platform such as android, Linux, etc. So It might be better to use separate build processes, files and defines.

DonghakPark (approved):
LGTM!
i will check CI ```unable to access 'https://github.com/nnstreamer/nntrainer/': The requested URL returned error: 503``` it seems like github access error

gkisalapl:
> @gkisalapl , Thanks for the PR. It is good if NNTrainer can support Windows Operating System naturally and I'm sure you are working on it. Thanks aging. One thing I want to address is that this Windows Porting task shouldn't be only for windows. I mean this code change also should work in other platform such as android, Linux, etc. So It might be better to use separate build processes, files and defines.

Yes I'm making those changes in the way it works both on windows and linux/android. Draft PR with experimental windows build is exception here - I've focused on proof of concept windows build in this draft PR.

DonghakPark:
Please use rebase command

```git rebase upstream/main```

gkisalapl:
> Please use rebase command
> 
> `git rebase upstream/main`

I've broke rebase operation and create new PR with manual rebase of those changes on top of current main #2884

PR #2870, opened 2025-01-13
[Windows] Use portable version of strerror

gkisalapl:
This PR is part of porting nntrainer to windows.
This PR replace using GNU specific function strerror_r with portable version std::strerror

Replace this:
https://linux.die.net/man/3/strerror_r 
With:
https://en.cppreference.com/w/cpp/string/byte/strerror

Self-evaluation:

### DIFF
--- a/Applications/utils/datagen/cifar/cifar_dataloader.cpp
+++ b/Applications/utils/datagen/cifar/cifar_dataloader.cpp
@@ -137,11 +137,8 @@ void Cifar100DataLoader::next(float **input, float **label, bool *last) {
   /// @note below logic assumes a single input and the fine label is used
 
   auto fill_one_sample = [this](float *input_, float *label_, int index) {
-    const size_t error_buflen = 102;
-    char error_buf[error_buflen];
     NNTR_THROW_IF(!file.good(), std::invalid_argument)
-      << "file is not good, reason: "
-      << strerror_r(errno, error_buf, error_buflen);
+      << "file is not good, reason: " << std::strerror(errno);
     file.seekg(index * Cifar100DataLoader::SampleSize, std::ios_base::beg);
 
     uint8_t current_label;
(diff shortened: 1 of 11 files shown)


skykongkong8 (approved):

EunjuYang (approved):
LGTM. Please check the CI.

skykongkong8 (commented):
You have to apply clang-format in order to pass the CI.
Please check the clang-format complaint, `unittest_tflite_export.cpp`

gkisalapl:
> You have to apply clang-format in order to pass the CI. Please check the clang-format complaint, `unittest_tflite_export.cpp`

I've run clang format, using following version:

$ clang-format --version
$ Ubuntu clang-format version 14.0.0-1ubuntu1.1

but it seems that on continues integration we're using some different version. 
Do you know which exactly ?

baek2sm (commented):
The strerror_r() function is similar to strerror(), but strerror_r() is thread safe function. 
In Windows, a thread-safe function named strerror_s() is provided separately.
Nevertheless, will there be no issue when this modification is applied?

gkisalapl:
> The strerror_r() function is similar to strerror(), but strerror_r() is thread safe function. In Windows, a thread-safe function named strerror_s() is provided separately. Nevertheless, will there be no issue when this modification is applied?

Good point, thank you for this finding. In order to be 100% sure that we keep current functionality we should do:

#ifdef _WIN32
strerror_s(...)
#else 
stderror_r(...)
#endif 

but If we stick with portable non thread safe version stderror() the only drawback is that subsequent calls to strerror() may overwrite the buffer addressed by the returned pointer and print incorrect error message in result.
In order to make it happen we would need to run report error by strerror() in 2 threads in similar time which is very unique situation I think. My recommendation is to go with portable version instead of making os specific calls.

myungjoo (approved):

djeong20 (approved):
LGTM! please rebase to pass the CI 💯

gkisalapl:
> LGTM! please rebase to pass the CI 💯

I've rebased this PR but it fail due to #2887

DonghakPark (approved):

Issue #2867, opened 2025-01-10
Unit Test Failures in Asynchronous Task Executor on Ubuntu


djeong20:
At present, the unit test for the asynchronous task executor experiences occasional failures during pdebuild on Ubuntu, which highlights the need for a thorough review and enhancement of our testing process. 

```
40/40 unittest_nntrainer_task          TIMEOUT        120.13s   killed by signal 15 SIGTERM
>>> MALLOC_PERTURB_=227 /build/nntrainer-0.5.0/build/test/unittest/unittest_nntrainer_task --gtest_output=xml:/build/nntrainer-0.5.0/build/unittest_nntrainer_task.xml
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
[==========] Running 6 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 6 tests from TaskExecutorTest
[ RUN      ] TaskExecutorTest.create_01_p
[       OK ] TaskExecutorTest.create_01_p (0 ms)
[ RUN      ] TaskExecutorTest.task_create_01_p
[       OK ] TaskExecutorTest.task_create_01_p (6 ms)
[ RUN      ] TaskExecutorTest.task_async_create_01_p
[       OK ] TaskExecutorTest.task_async_create_01_p (0 ms)
[ RUN      ] TaskExecutorTest.run_01_p
[       OK ] TaskExecutorTest.run_01_p (6 ms)
[ RUN      ] TaskExecutorTest.run_async_01_p
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――


Summary of Failures:

40/40 unittest_nntrainer_task          TIMEOUT        120.13s   killed by signal 15 SIGTERM
```

skykongkong8:
I could find the similar error occasionally outbreaks at Tizen x86_64 gbs build as well.

PR #2872, opened 2025-01-14
[ graph ] Divide addLossLayer process when compiling inference model graph @open sesame 01/15 14:27

skykongkong8:
- Inference model does not require label and loss layer.
- Therefore, we can just simply ignore addLossLayer during the inference mode.
- Add unittest to verify such.

**Self evaluation:**

FYI) Following is the output of : 
```C++
TEST(nntrainerGraphUnitTest, NoLossLayerWhenInferenceMode) {
...
   model->summarize(std::cout, ML_TRAIN_SUMMARY_MODEL);
...
}
```
```bash
================================================================================
          Layer name          Layer type    Output dimension         Input layer
================================================================================
              input0               input           1:1:1:256                    
--------------------------------------------------------------------------------
    fully_connected0     fully_connected          1:1:1:1024              input0
--------------------------------------------------------------------------------
    fully_connected1     fully_connected          1:1:1:1024    fully_connected0
--------------------------------------------------------------------------------
    fully_connected2     fully_connected          1:1:1:1024    fully_connected1
--------------------------------------------------------------------------------
    fully_connected3     fully_connected           1:1:1:100    fully_connected2
================================================================================
```

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -65,14 +65,21 @@ int NetworkGraph::compile(const std::string &loss_type) {
 
   graph.realizeInputOutputNode();
 
-  try {
-    /// @todo realize loss beforehand
-    status = addLossLayer(loss_type);
-    NN_RETURN_STATUS();
-  } catch (const std::exception &e) {
-    ml_loge("%s", e.what());
-    status = ML_ERROR_INVALID_PARAMETER;
-    NN_RETURN_STATUS();
+  if (exec_mode != ExecutionMode::INFERENCE) {
+    try {
+      /// @todo realize loss beforehand
+      status = addLossLayer(loss_type);
+      NN_RETURN_STATUS();
+    } catch (const std::exception &e) {
+      ml_loge("%s", e.what());
+      status = ML_ERROR_INVALID_PARAMETER;
+      NN_RETURN_STATUS();
+    }
+  } else {
+    if (!loss_type.empty()) {
+      ml_loge(
+        "Warning : Loss type is given in inference mode. Ignoring loss type.");
+    }
   }
 
   graph.topologicalSort();
--- a/test/unittest/unittest_nntrainer_graph.cpp
+++ b/test/unittest/unittest_nntrainer_graph.cpp
@@ -368,6 +368,51 @@ TEST(nntrainerGraphUnitTest, call_functions) {
   }
 }
 
+TEST(nntrainerGraphUnitTest, NoLossLayerWhenInferenceMode) {
+  std::unique_ptr<ml::train::Model> model =
+    ml::train::createModel(ml::train::ModelType::NEURAL_NET);
+
+  model->addLayer(ml::train::createLayer(
+    "input", {withKey("name", "input0"), withKey("input_shape", "1:1:256")}));
+
+  for (int i = 0; i < 3; ++i) {
+    model->addLayer(ml::train::createLayer(
+      "fully_connected",
+      {withKey("unit", 1024), withKey("weight_initializer", "xavier_uniform"),
+       withKey("bias_initializer", "zeros")}));
+  }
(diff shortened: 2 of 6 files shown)


test/unittest/unittest_nntrainer_graph.cpp
```diff
@@ -368,6 +368,55 @@ TEST(nntrainerGraphUnitTest, call_functions) {
...
+    "fully_connected",
+    {withKey("unit", 100), withKey("weight_initializer", "xavier_uniform"),
+     withKey("bias_initializer", "zeros")}));
+
+  model->setProperty({withKey("batch_size", 1), withKey("epochs", 1),
+                      withKey("memory_swap", "false"),
+                      withKey("model_tensor_type", "FP32-FP32")});
+
+  auto optimizer = ml::train::createOptimizer("sgd", {"learning_rate=0.001"});
+  model->setOptimizer(std::move(optimizer));
```
EunjuYang, line 394: I think the optimizer is unnecessary as well for the `INFERENCE` mode.

EunjuYang (approved):
Nice work! 👍

test/unittest/unittest_nntrainer_graph.cpp
skykongkong8: You have a point! Let me fix this one.

test/unittest/unittest_nntrainer_graph.cpp
```diff
@@ -368,6 +368,52 @@ TEST(nntrainerGraphUnitTest, call_functions) {
...
+                      withKey("model_tensor_type", "FP32-FP32")});
+
+  int status = model->compile(ml::train::ExecutionMode::INFERENCE);
+  EXPECT_EQ(status, ML_ERROR_NONE);
+
+  status = model->initialize(ml::train::ExecutionMode::INFERENCE);
+  EXPECT_EQ(status, ML_ERROR_NONE);
+
+  float input[256];
+  float label[1];
```
DonghakPark, line 400: is this label need ?

DonghakPark (approved):
Good Point!! LGTM!

test/unittest/unittest_nntrainer_graph.cpp
skykongkong8: I missed it😂 Thanks!

EunjuYang (approved):

skykongkong8:
It seems issue raised from #2867 is occurring here as well

djeong20:
> It seems issue raised from #2867 is occurring here as well

please re-run the action to pass as for now.

lhs8928 (approved):
LGTM

skykongkong8:
For the static checker error, it seems there's an similar problem that @djeong20 raised from  #2870 is occuring for the PR as well.
With the Static checkers and verifier's suggestion, I could not avoid C++ Format Checker, and with C++ Format Checker's suggesion, I could not avoid Static checkers and verifier error.

nntrainer/graph/network_graph.cpp
```diff
@@ -65,14 +65,21 @@ int NetworkGraph::compile(const std::string &loss_type) {
...
+      /// @todo realize loss beforehand
+      status = addLossLayer(loss_type);
+      NN_RETURN_STATUS();
+    } catch (const std::exception &e) {
+      ml_loge("%s", e.what());
+      status = ML_ERROR_INVALID_PARAMETER;
+      NN_RETURN_STATUS();
+    }
+  } else {
+    if (!loss_type.empty()) {
```
baek2sm, line 79: Wouldn't it be better to allow it to run even with loss layer in inference mode for evaluation?
I think someone can add a loss layer in inference mode to evaluate the performance of the model without training.

skykongkong8: Well, you have a point. Maybe we should discuss with @jijoongmoon about this point.

PR #2878, opened 2025-01-15
[CI] update static checker script to downgrade clang-format version.

djeong20:
This pull request updates the workflow script to run on Ubuntu 22.04 instead of the latest version, allowing the static checker to use clang-format version 14.

**Self-evaluation:**
2. Run test:   [ ]Passed [ ]Failed [X]Skipped

### DIFF
--- a/.github/workflows/cpp_linter.yml
+++ b/.github/workflows/cpp_linter.yml
@@ -4,7 +4,7 @@ on: pull_request
 
 jobs:
   cpp-linter:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
       - name: install dev packages
@@ -23,6 +23,7 @@ jobs:
           file-annotations: false
           step-summary: true
           format-review: true
+          tidy-checks: '-*' # disable clang-tidy checks.
 
       - name: failing fast
         if: steps.linter.outputs.clang-format-checks-failed > 0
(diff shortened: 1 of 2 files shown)


DonghakPark (approved):
Nice Catch!
And i think we can remove cpp_checker in static_check to avoid confusion ```.github/workflows/cpp_linter.yml``` seems to be working properly, except minor differences.

skykongkong8 (approved):
I suffered from this SO LONG!
Great work!

EunjuYang (approved):
LGTM 👍

djeong20:
> Nice Catch! And i think we can remove cpp_checker in static_check to avoid confusion `.github/workflows/cpp_linter.yml` seems to be working properly, except minor differences.

Thank you for your feedback! I have removed the clang-format checks from the static checkers and verifiers. Please take a look.

.github/workflows/cpp_linter.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           file-annotations: false
           step-summary: true
           format-review: true
+          tidy-checks: '-*' # disable clang-tidy checks.
```
djeong20, line 26: Disabling clang-tidy checks because the reports generated are just simple warnings and may provide incorrect results.

DonghakPark:
> > Nice Catch! And i think we can remove cpp_checker in static_check to avoid confusion `.github/workflows/cpp_linter.yml` seems to be working properly, except minor differences.
> 
> Thank you for your feedback! I have removed the clang-format checks from the static checkers and verifiers. Please take a look.

Thank you for quick apply!! LGTM again

baek2sm (approved):
LGTM

PR #2879, opened 2025-01-15
[Windows] Add multi os wrapper for loading dynamic libraries

gkisalapl:
Create class which make possible to use the same interface for loading dynamic libraries and symbols for multiple operating systems

**Self-evaluation:**

### DIFF
--- a/nntrainer/utils/dynamic_library_loader.h
+++ b/nntrainer/utils/dynamic_library_loader.h
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2025 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file   dynamic_library_loader.h
+ * @date   14 January 2025
+ * @brief  Wrapper for loading dynamic libraries on multiple operating systems
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Grzegorz Kisala <g.kisala@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#ifndef __DYNAMIC_LIBRARY_LOADER__
+#define __DYNAMIC_LIBRARY_LOADER__
+
+#include <string>
+
+#ifdef _WIN32
+#include "windows.h"
+
+// This flags are not used on windows. Defining those symbols for windows make
+// possible using the same external interface for loadLibrary function
+#define RTLD_LAZY 0
+#define RTLD_NOW 0
+#define RTLD_BINDING_MASK 0
+#define RTLD_NOLOAD 0
+#define RTLD_DEEPBIND 0
+#define RTLD_GLOBAL 0
+#define RTLD_LOCAL 0
+#define RTLD_NODELETE 0
+
+#else
+#include <dlfcn.h>
+#endif
+
+namespace nntrainer {
+
+/**
+ * @brief DynamicLibraryLoader wrap process of loading dynamic libraries for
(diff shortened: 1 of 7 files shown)


DonghakPark (commented):
Please Fill Commit body message on below commits for log & clear understanding of others 

[Fix Tizen GBS build](https://github.com/nnstreamer/nntrainer/pull/2879/commits/d1115721c911196b02ba325d7d79798a5c4a83fe) 


[Add new file to nntrainer.spec](https://github.com/nnstreamer/nntrainer/pull/2879/commits/8a8b97a1f5f880be74bd46591e5aae98944c7a06)

myungjoo (approved):

myungjoo:
@DonghakPark We are exporting every header files to .dev package. But, this is not a good practice and we need to choose header files that are needed by external packages and exclude other headers from .dev package. Opening up all the symbols to external packages is not a good design choice (you are declaring all functions as APIs by doing so.)

DonghakPark:
> @DonghakPark We are exporting every header files to .dev package. But, this is not a good practice and we need to choose header files that are needed by external packages and exclude other headers from .dev package. Opening up all the symbols to external packages is not a good design choice (you are declaring all functions as APIs by doing so.)

That's correct. 
Recently, during development, we haven't had enough time to sort out these aspects. we'll discuss this matter with other team members so that we can distinguish APIs (open or not) that will be released at the completion of FSU, SubGraph, and Op(ONNX) developments

myungjoo:
> > @DonghakPark We are exporting every header files to .dev package. But, this is not a good practice and we need to choose header files that are needed by external packages and exclude other headers from .dev package. Opening up all the symbols to external packages is not a good design choice (you are declaring all functions as APIs by doing so.)
> 
> That's correct. Recently, during development, we haven't had enough time to sort out these aspects. we'll discuss this matter with other team members so that we can distinguish APIs (open or not) that will be released at the completion of FSU, SubGraph, and Op(ONNX) developments

It's Saturday morning. Go somewhere else!

nntrainer/utils/dynamic_library_loader.h
```diff
@@ -0,0 +1,76 @@
...
+// This flags are not used on windows. Defining those symbols for windows make
+// possible using the same external interface for loadLibrary function
+#define RTLD_LAZY 0
+#define RTLD_NOW 0
+#define RTLD_BINDING_MASK 0
+#define RTLD_NOLOAD 0
+#define RTLD_DEEPBIND 0
+#define RTLD_GLOBAL 0
+#define RTLD_LOCAL 0
+#define RTLD_NODELETE 0
```
djeong20, line 31: Some of the flags, such as `RTLD_BINDING_MASK`, are currently unused. Will we use these unused flags in the future?

djeong20 (approved):
Nice work!

nntrainer/utils/dynamic_library_loader.h
gkisalapl: I've added all flags defined in bits/dlfcn.h linux header to ensure full compatibility.
I do not think that we will use all of them.

myungjoo: Or.. we may hide them by:

```C++
class DynamicLibraryLoader {
public:
  static void *loadLibrary(const char *path) {
#if defined(_WIN32)
    return LoadLibraryA(path);
#else
    return dlopen(path, RTLD_NOW | RTLD_LOCAL);
#endif
  }

```

jijoongmoon (approved):
LGTM!

PR #2883, opened 2025-01-16
[Windows] Simplify iniparser import @open sesame 01/21 17:38

gkisalapl:
Use most standard meson way to import cmake iniparser library This way of importing will work regardless of operating system

**Self-evaluation:**


DonghakPark (commented):
Please write more message on Commit body

[Fix android build and cleanup](https://github.com/nnstreamer/nntrainer/pull/2883/commits/88fc893bc286e5d8a01db5b50647f7a442ca3170)

myungjoo (changes requested):
You are breaking non Windows build here.

In most Linux distros, you are supposed to build with the packages available inside the distro, not from external repositories. For embedded deviecs, the principle is the same; you share the same library (iniparse in this case) with all the other packages in the system.

You may conditionally import the wrap-git if iniparser is not found by pkg-config (```dependnecy('iniparser', ...)```)

myungjoo (approved):

meson.build
```diff
@@ -1,7 +1,7 @@
 project('nntrainer', 'c', 'cpp',
   version: '0.5.0',
   license: ['apache-2.0'],
-  meson_version: '>=0.50.0',
```
EunjuYang, line 4: Regarding the Meson version, we need to update the version information in `getting-started.md` once this pull request is merged. Additionally, I've encountered the issues with previous versions of Meson myself, so I support upgrading to version 0.55.

EunjuYang (approved):
LGTM! 👍

meson.build
EunjuYang: It needs to be approved by maintainer about the meson version up.

EunjuYang (commented):
For the android build, I find it should be installed under the authorized path. Please check it.

meson.build
```diff
@@ -347,37 +347,20 @@ libdl_dep = cxx.find_library('dl') # DL library
...
+
+  # iniparser
+  iniparser_root = meson.source_root() / 'subprojects' / 'iniparser'
+  iniparser_options = cmake.subproject_options()
+  iniparser_options.add_cmake_defines({'BUILD_DOCS': false})
+  iniparser_options.add_cmake_defines({'BUILD_EXAMPLES': false})
+  iniparser_options.add_cmake_defines({'BUILD_TESTING': false})
+  iniparser_options.add_cmake_defines({'BUILD_SHARED_LIBS': false})
+  iniparser_subproject = cmake.subproject('iniparser', options: iniparser_options, required: true)
+  iniparser_dep = iniparser_subproject.dependency('iniparser-static')
```
myungjoo, line 363: pdebuild (Ubuntu platform build) and gbs/obs (Tizen platform build) does NOT allow referencing external repo. We need to keep the currentfallback mechanism because Ubuntu-iniparser does not have pkgconfig support and we still need to link to such packages.

If the platform (Ubuntu, Tizen, and so on) has the iniparser, we need to link to the platform-supplied shared library.
So, please restrict the usage of external repos to the case that really needs it.

(this is not related to meson versioning issue.)

meson.build
EunjuYang: Oh I see. Thank you for the confirmation!

myungjoo:
> Currently, the CI's Meson is installed via `apt install`, which defaults to version 0.53. It appears that updating the CI Meson version would also be required if the Meson version change is approved.
> 
> 


If this is successful, it will be fixed: https://code.launchpad.net/~myungjoo-ham/+recipe/meson-daily-devel

piotrrak (approved):

EunjuYang:
> > Currently, the CI's Meson is installed via `apt install`, which defaults to version 0.53. It appears that updating the CI Meson version would also be required if the Meson version change is approved.
> 
> If this is successful, it will be fixed: https://code.launchpad.net/~myungjoo-ham/+recipe/meson-daily-devel
> 
> Now, you are getting the most recent version of meson for Focal: https://code.launchpad.net/~myungjoo-ham/+recipe/meson-daily

Thank you so much, @myungjoo ! I checked the issue is resolved with your support!

BTW, @gkisalapl , you still have some issues on CI. 
1) Please check the comment about the fallback to reference internal repo only.
2) About the test case failure (#2900), it would be handled in #2901, so you can skip to handle the issue.

djeong20 (approved):
Nice work! LGTM 👍

dkjung:
> > > Currently, the CI's Meson is installed via `apt install`, which defaults to version 0.53. It appears that updating the CI Meson version would also be required if the Meson version change is approved.
> > 
> > 
> > If this is successful, it will be fixed: https://code.launchpad.net/~myungjoo-ham/+recipe/meson-daily-devel
> > Now, you are getting the most recent version of meson for Focal: https://code.launchpad.net/~myungjoo-ham/+recipe/meson-daily
> 
> Thank you so much, @myungjoo ! I checked the issue is resolved with your support!
> 
> BTW, @gkisalapl , you still have some issues on CI.
> 
> 1. Please check the comment about the fallback to reference internal repo only.
> 2. About the test case failure ([Possible FP16/FP32 Bug. Unittest fails #2900](https://github.com/nnstreamer/nntrainer/issues/2900)), it would be handled in [[Tensor] Fix improper vector initialization in Tensor::split function #2901](https://github.com/nnstreamer/nntrainer/pull/2901), so you can skip to handle the issue.

jijoongmoon (approved):
LGTM

PR #2886, opened 2025-01-20
[ graph ] Allow loss layer to be used in inference mode

skykongkong8:
- Intention of the c5db575d commit was to completely remove the usage of loss and the label in inference.
- However, there are cases where the loss layer is used in inference mode, such as the case of the loss layer being used as a metric which means that the formulation of the loss layer should not be determined by the execution mode, but the presence of the loss layer given by the user.
- Thus, when loss layer is given when the model is formulated, the model will always have the loss layer, and it will expect to have a label vector later on (since the presence of the label vector cannot be detected on compile time.
- On the contrary, when the loss layer is not given, the model will not have the loss layer, and it will not expect to have a label vector later on.

To summarize when inference mode,
|  loss | label | at inference time... |
| --- | --- | --- |
| O | O | Loss layer is formulated, loss will be computed w.r.t. label |
| O | X | Loss layer is formulated, but no label -> emit error |
| X | O | Loss layer is not formulated, but label is given -> emit error |
| X | X | Loss layer is not formulated, label is empty as well -> inference w/o loss |

Added TC will emit the output like below with model summary:
```bash
================================================================================
          Layer name          Layer type    Output dimension         Input layer
================================================================================
              input0               input           1:1:1:256                    
--------------------------------------------------------------------------------
    fully_connected0     fully_connected          1:1:1:1024              input0
--------------------------------------------------------------------------------
    fully_connected1     fully_connected          1:1:1:1024    fully_connected0
--------------------------------------------------------------------------------
    fully_connected2     fully_connected          1:1:1:1024    fully_connected1
--------------------------------------------------------------------------------
    fully_connected3     fully_connected           1:1:1:100    fully_connected2
--------------------------------------------------------------------------------
                mse0                 mse           1:1:1:100    fully_connected3
================================================================================
```

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -65,7 +65,7 @@ int NetworkGraph::compile(const std::string &loss_type) {
 
   graph.realizeInputOutputNode();
 
-  if (exec_mode != ExecutionMode::INFERENCE) {
+  if (!loss_type.empty()) {
     try {
       /// @todo realize loss beforehand
       status = addLossLayer(loss_type);
@@ -74,11 +74,6 @@ int NetworkGraph::compile(const std::string &loss_type) {
       ml_loge("%s", e.what());
       status = ML_ERROR_INVALID_PARAMETER;
       NN_RETURN_STATUS();
-    }
-  } else {
-    if (!loss_type.empty()) {
-      ml_loge(
-        "Warning : Loss type is given in inference mode. Ignoring loss type.");
     }
   }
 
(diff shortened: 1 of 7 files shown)


DonghakPark (approved):
LGTM! especially Consider various case, add unittest and fix tflite export!! 👍

EunjuYang (approved):
LGTM

dkjung:
It seems that this PR is tagged `DO NOT MERGE`. Will this be close?

PR #2890, opened 2025-01-21
[Windows] Replace uint with unsigned int @open sesame 01/22 16:36

gkisalapl:
We're currently using uint type which is not C++ fundamental type (see https://en.cppreference.com/w/cpp/language/types) but GNU define. This PR replace all occurrences of uint with unsigned int - which is compatible across multiple operating systems

**Self-evaluation:**

Signed-off-by: Grzegorz Kisala <gkisala@gmail.com>

### DIFF
--- a/Applications/Android/NNDetector/app/src/main/jni/dataloader.cpp
+++ b/Applications/Android/NNDetector/app/src/main/jni/dataloader.cpp
@@ -42,8 +42,8 @@ namespace simpleshot {
  * @param input_h height of input
  * @param new_dim dimension
  */
-void resize_input(float *input, float *resized_input, uint &input_w,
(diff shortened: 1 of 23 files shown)


EunjuYang (approved):
LGTM

DonghakPark (approved):
LGTM!

skykongkong8 (approved):
Good catch!

djeong20 (approved):

myungjoo (approved):
LGTM.

Anyway, using cstdint (stdint.h) is fine if you wish.

jijoongmoon (approved):
We should use "unsigned int" later on then. LGTM.

PR #2891, opened 2025-01-21
[ONNX] add onnx proto & meson setting

baek2sm:
- add onnx proto & compile it using meson protoc
- add "enable-onnx-interpreter" meson option.

I'll upload a separate pr for the onnx interpreter script.

**Self evaluation:**


baek2sm:
i'll check the doxygen issue (that compiled-onnx-schema script is automatically generated by the protocol buffer compiler with onnx schema file)

EunjuYang (commented):
I have a question about the onnx proto.
Isn't the proto already defined for onnx format? What about adding proto file (not pb.cc & pb.h) and using it by compilation?

DonghakPark:
> Member

For external file : i think we don't need consider doxygen
and for clang-format relate CI : you can use "// clang-format off ~~ // clang-format on " or just ignore that too

myungjoo:
If onnx.pb.cc/h files are generated by protoc, add a onnx.protofile and let meson script generate .cc/.h files from it.

Or is there a reason you cannot generate files in the build scripts?

Reference: https://github.com/nnstreamer/nnstreamer/commit/ff5c1fd34230e4d72eee0711118fedef64f3b8a1

baek2sm:
@DonghakPark Thank you for your detailed guide!!

myungjoo:
> @EunjuYang @myungjoo Actually, I just didn't want to have a dependency on protoc..! But if that's not important, I'll add a onnx.proto file and compile it using meson script. thanks!

Having auto-generated files commited in a git repo is not recommended. It is like having .so file in the git repo where you can compile that .so file easily from available open source codes.

You already have a dependency to protobuf-devel. Adding protoc dependency shouldn't be a problem.
For ubuntu, add the following to /debian/control :
```
 protobuf-compiler (>=3.12), libprotobuf-dev,
```
For Tizen:
```
BuildRequires: protobuf-devel >= 3.4.0
```
For Android, unfortunately, you may need to touch a few more scripts. (I guess this is why you added the auto-generated files.) But you don't have any dependency issue for this; you just need to enforce developers to install a proper protobuf-compiler here.

nntrainer/compiler/meson.build
```diff
@@ -16,6 +16,15 @@ compiler_headers = [
...
 ]
 
+if get_option('enable-onnx-interpreter')
+  protobuf_dep = dependency('protobuf', required: true)
+  nntrainer_base_deps += protobuf_dep
+
+  compiler_sources += [
+    'onnx.pb.cc',
+  ]
+endif
```
myungjoo, line 26: You need a script to generate onnx.pb.cc around here.

baek2sm:
```
protobuf_dep = dependency('protobuf', required: true)
nntrainer_base_deps += protobuf_dep

protoc = find_program('protoc', required: true)  
pb_gen = generator(
      protoc,
      output: ['@BASENAME@.pb.cc', '@BASENAME@.pb.h'],
      arguments: ['--cpp_out=@BUILD_DIR@', '-I=@CURRENT_SOURCE_DIR@', '@INPUT@'],
    )
pb_gen_src = pb_gen.process('onnx.proto')
// + shared_library setting using pb_gen_src
```
@myungjoo I referred to the given code and tried to use the `generator` as shown above, and I think this method is more appropriate for this purpose. However, since the same meson script uses `custom_target` instead of `generator` for FlatBuffer and TFLite, I uploaded the code using `custom_target` to unify the code within the same script. How about updating all parts to use `generator` instead of `custom_target` at a later time(when I have some spare time)?

myungjoo (approved):

EunjuYang (approved):
LGTM! Thank you for reflecting the opinion 👍

jijoongmoon:
We do have tflite scheme file (fb) in compiler directory as well. I recommend to make scheme dir and put those in the same dir for better managing.

skykongkong8 (approved):

baek2sm:
> We do have tflite scheme file (fb) in compiler directory as well. I recommend to make scheme dir and put those in the same dir for better managing.

@jijoongmoon Thank you for your feedback. then I'll make a new commit for making a new directory and put those schema files(onnx, tflite, nntrainer schema) in the same directory!

myungjoo:
@baek2sm please rebase to HEAD. let's see if that tc error goes away.

jijoongmoon (approved):
LGTM

PR #2895, opened 2025-01-23
[Tensor] Int4QTensor with quantized 4-bit integer data type

djeong20:
This pull request presents the class, a powerful solution for efficiently storing quantized 4-bit integer data. By packing each 4-bit integer into an 8-bit memory space, we utilize memory resources effectively—the first four bits hold the first 4-bit value, and the last four bits hold the second.


nntrainer/tensor/int4_tensor.cpp
```diff
@@ -0,0 +1,504 @@
...
+  output += value;
+
+  // if result value is out of range, clamp to max/min value
+  int8_t val = std::trunc(std::clamp((int)output, -8, 7));
+
+  // encode result value to int8 data
+  ((int8_t *)getData())[idx / 2] =
+    (idx % 2 == 0) ? (val << 4) | (((int8_t *)getData())[idx / 2] & 0x0f)
+                   : (((int8_t *)getData())[idx / 2] << 4) | (val & 0x0f);
+}
```
skykongkong8, line 248: Quick question:
Do we just expect the user to consider scale factor in input `float value` and `float beta`?
I am curious about how basic math in int4Q tensor goes...

skykongkong8 (approved):
Glad to have many negatigve unittest TCs as well. All good!

nntrainer/tensor/int4_tensor.cpp
djeong20: Thanks for asking! Currently, no. This is to modify the quantized value directly.

nntrainer/tensor/int4_tensor.cpp
```diff
@@ -0,0 +1,504 @@
...
+  output *= beta;
+  output += value;
+
+  // if result value is out of range, clamp to max/min value
+  int8_t val = std::trunc(std::clamp((int)output, -8, 7));
+
+  // encode result value to int8 data
+  ((int8_t *)getData())[idx / 2] =
+    (idx % 2 == 0) ? (val << 4) | (((int8_t *)getData())[idx / 2] & 0x0f)
+                   : (((int8_t *)getData())[idx / 2] << 4) | (val & 0x0f);
```
EunjuYang, line 247: As I understood, isn't the computation should be 
```suggestion
                   : (((int8_t *)getData())[idx / 2] & 0xf0) | (val & 0x0f);
```
I'm quite confused with it. Please let me know If I'm wrong :)

nntrainer/tensor/int4_tensor.cpp
```diff
@@ -0,0 +1,504 @@
...
+                           unsigned int w, float value) {
+  NNTR_THROW_IF(value < -8 || value > 7, std::out_of_range)
+    << "Value must be in range [-8, 7]. Input value: " << value;
+
+  auto const &idx = getIndex(b, c, h, w);
+  int8_t val = value;
+
+  ((int8_t *)getData())[idx / 2] =
+    (idx % 2 == 0) ? (val << 4) | ((int8_t *)getData())[idx / 2]
+                   : ((int8_t *)getData())[idx / 2] | (val & 0x0f);
```
EunjuYang, line 261: I find we need to clear out the space we want to append the value.
```suggestion
    (idx % 2 == 0) ? (val << 4) | (((int8_t *)getData())[idx / 2] & 0x0f)
                   : (((int8_t *)getData())[idx / 2] & 0xf0) | (val & 0x0f);
```

nntrainer/tensor/int4_tensor.cpp
djeong20: you're right! thanks for pointing it out :)

nntrainer/tensor/int4_tensor.cpp
djeong20: makes sense 👍

EunjuYang (approved):
LGTM!

It's a simple question. Is there any specific naming convention assumed for `Int4QTensor`? Considering the datatype is `QINT4`, I'm wondering if it would be better to rename it to `QInt4Tensor`.

jijoongmoon (approved):
LGTM

PR #2638, opened 2024-06-13
action: add risc-v test on Ubuntu (wait for #2902)

myungjoo:
RISC-V is going to be one of the officially supported archtiecture, soon.
Test it and reject a PR if it breaks RISC-V builds.

PR #2902, opened 2025-01-24
meson: avx condition fix.

myungjoo:
enable-avx=true is the default option, which breaks the build in other archs in general.

1. Ignore AVX if arch is not x64/x86.
2. Reading get_option should be done only once. Fixed for enable-avx.
2. Corrected the buggy AVX condition.

This fix is required by RISC-V: #2638

### DIFF
--- a/nntrainer/tensor/meson.build
+++ b/nntrainer/tensor/meson.build
@@ -52,9 +52,10 @@ tensor_headers = [
 
 arch = host_machine.cpu_family()
 
-if get_option('enable-avx') and get_option('platform') != 'android'
-    tensor_sources += 'blas_avx.cpp'
-    tensor_headers += 'blas_avx.h'
+
+if avx_enabled == true
+  tensor_sources += 'blas_avx.cpp'
+  tensor_headers += 'blas_avx.h'
 endif
 
 if get_option('enable-fp16') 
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/meson.build
```diff
@@ -52,9 +52,11 @@ tensor_headers = [
...
 arch = host_machine.cpu_family()
 
-if get_option('enable-avx') and get_option('platform') != 'android'
-    tensor_sources += 'blas_avx.cpp'
-    tensor_headers += 'blas_avx.h'
+
+if avx_enabled == true
+  tensor_sources += 'blas_avx.cpp'
+  tensor_headers += 'blas_avx.h'
+  avx_enabled = true
```
djeong20, line 59:
```suggestion
```
remove?

djeong20 (approved):
Overall, LGTM

skykongkong8 (approved):

skykongkong8 (commented):
This must be one of the reasons for other PR's CI

DonghakPark (approved):
LGTM!

nntrainer/tensor/meson.build
myungjoo: Removed. Thanks.

PR #2903, opened 2025-01-24
tizen spec: fix path macros

myungjoo:
Path macro has been incorrectly used.

%{_libdir} == /usr/lib, /usr/lib64, or ...
%{_bindir} == /usr/bin

Meson accepts postfixes for libdir and bindir.
Do not give meson absolute pathes for libdir/bindir. The relative library path working as postfix is %{_lib}.

Standard rpm macros are at:
https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/

### DIFF
--- a/packaging/nntrainer.spec
+++ b/packaging/nntrainer.spec
@@ -2,7 +2,7 @@
 %define         use_cblas 1
 %define         nnstreamer_filter 1
 %define         nnstreamer_trainer 1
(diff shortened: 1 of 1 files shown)


haehun (approved):
LGTM!

skykongkong8 (approved):
LGTM

EunjuYang (approved):
LGTM

PR #2868, opened 2025-01-13
[windows] Add experimental windows build

gkisalapl:
# PR Description

This is draft PR which contains experimental support for windows build (MSVC compiler) of nntrainer

PR itself is too big for review and SHOULD NOT be reviewed and SHOULD NOT be merged to main branch
Based on this draft PR I'll make series of PRs including small changes in code which in result add full support for windows build

PR #2904, opened 2025-01-24
Compile nntrainer on windows

gkisalapl:
[Windows] Compile nntrainer on windows

After this PR all nntrainer library files are possible to compile with MSVC compiler on Windows and most of MSVC warnigs are fixed.
It can be tested by using meson.build from #2868 

Following options should be disabled during build:

[project options]

enable-blas=false 
enable-tflite-backbone=false 
enable-memory-swap=false 
enable-nnstreamer-backbone=false 
enable-tflite-interpreter=false 
enable-app=false 
ml-api-support='disabled'
enable-test=false
install-app=false
enable-avx=false
enable-mmap=false

[built-in options]
werror=false

As a next step I'll modify build system in order to make both windows / linux / android / tizen build possible using the same meson scripts

Self-evaluation:

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -17,6 +17,7 @@ option('enable-memory-swap', type: 'boolean', value: false)
 option('memory-swap-path', type: 'string', value: '')
 option('test-timeout', type: 'integer', value: 60)
 option('opencl-kernel-path', type: 'string', value: 'nntrainer_opencl_kernels')
+option('enable-mmap', type: 'boolean', value: true)
 
 # dependency conflict resolution
 option('capi-ml-inference-actual', type: 'string', value: 'capi-ml-inference',
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -222,14 +222,6 @@ const std::string getFullPath(const std::string &path,
 
 std::mutex factory_mutex;
 
-/**
- * @brief finalize global context
- *
- */
-static void fini_global_context_nntrainer(void) __attribute__((destructor));
-
-static void fini_global_context_nntrainer(void) {}
-
 std::once_flag global_app_context_init_flag;
 
 static void add_default_object(AppContext &ac) {
--- a/nntrainer/tensor/manager.h
+++ b/nntrainer/tensor/manager.h
@@ -38,6 +38,8 @@
 
 namespace nntrainer {
 using ExecutionMode = ml::train::ExecutionMode;
+
+#if !defined(_WIN32)
 /**
  * @class MMappedMemory
  * @brief Memory Handler, that has mmaped memory with a file descriptor
@@ -104,6 +106,7 @@ private:
   size_t buf_size;  /**< buffer size */
   bool allocate_fd; /**< option to choose to allocate an fd */
 };
+#endif
 
 /**
  * @class   Manager
--- a/nntrainer/tensor/swap_device.h
+++ b/nntrainer/tensor/swap_device.h
@@ -18,15 +18,22 @@
 #include <map>
 #include <memory>
 #include <string>
-#include <sys/mman.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <system_error>
-#include <unistd.h>
 #include <utility>
 
-/* Uncomment this to use mmap for swap data */
-#define USE_MMAP
+#if defined(_WIN32)
+#include <io.h>
+#define O_SYNC 0UL
+#else
+#include <sys/mman.h>
+#include <unistd.h>
+#endif
+
+#if defined(_WIN32)
+using ssize_t = std::make_signed_t<size_t>;
+#endif
 
 namespace nntrainer {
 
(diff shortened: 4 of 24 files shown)


skykongkong8:
FYI) Rebasing main will resolve errors like:
```
[ 1489s] [ RUN      ] nntrainer_Tensor.split_01_p
[ 1489s] stderr:
[ 1489s] /usr/lib64/gcc/x86_64-tizen-linux-gnu/14.2.0/include/c++/bits/stl_vector.h:1130: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = ml::train::TensorDim; _Alloc = std::allocator<ml::train::TensorDim>; reference = ml::train::TensorDim&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
[ 1489s] ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
[ 1489s] 
```

from [here](https://github.com/nnstreamer/nntrainer/actions/runs/12954740719/job/36137324487?pr=2904)

myungjoo:
Great! Now, github-action script for windows build becomes possible after this!

myungjoo (approved):

skykongkong8 (approved):
LGTM

dkjung (approved):
LGTM

nntrainer/app_context.cpp
```diff
@@ -222,14 +222,6 @@ const std::string getFullPath(const std::string &path,
...
 std::mutex factory_mutex;
 
-/**
- * @brief finalize global context
- *
- */
-static void fini_global_context_nntrainer(void) __attribute__((destructor));
-
-static void fini_global_context_nntrainer(void) {}
-
```
dkjung, line 232: Does this cause a compile error for Windows target?

dkjung (approved):

nntrainer/app_context.cpp
gkisalapl: Yes __attribute__((destructor)) is GNU extension

EunjuYang (approved):
LGTM!

nntrainer/tensor/swap_device.h
```diff
@@ -18,15 +18,22 @@
...
-#include <sys/mman.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <system_error>
-#include <unistd.h>
 #include <utility>
 
-/* Uncomment this to use mmap for swap data */
-#define USE_MMAP
+#if defined(_WIN32)
```
jijoongmoon, line 26: The default value of swap device should be  USE_MMAP. I think if this feature is not supported in Windows, it might be better not to compile this swap_device.h

meson_options.txt
```diff
@@ -17,6 +17,7 @@ option('enable-memory-swap', type: 'boolean', value: false)
 option('memory-swap-path', type: 'string', value: '')
 option('test-timeout', type: 'integer', value: 60)
 option('opencl-kernel-path', type: 'string', value: 'nntrainer_opencl_kernels')
+option('enable-mmap', type: 'boolean', value: true)
```
jijoongmoon, line 20: This mmap should be enabled for Android. please add enable-mmap in tools/package_android.sh  https://github.com/nnstreamer/nntrainer/blob/main/tools/package_android.sh

nntrainer/tensor/manager.h
```diff
@@ -38,6 +38,8 @@
 
 namespace nntrainer {
 using ExecutionMode = ml::train::ExecutionMode;
+
+#if !defined(_WIN32)
```
jijoongmoon, line 42: Is there any reason to remove this Code for Windows?

jijoongmoon:
Over all, I think it is LGTM. Some comments about disabled options.
[project options]

- **enable-blas=false**  
    : Currently we are using OpenBlas as default. but you could use MKL if the X86 Architecture.  But still need to consider OpenBlas by default for the case of No MKL (Intel Math Kernel Library)

- **enable-tflite-backbone=false**
: Can be disabled when the Tensorflow lite is not available.

- **enable-memory-swap=false**
: We do have FSU (Flash Storage Utilization) with this option. This need to be supported later. 

- **enable-nnstreamer-backbone=false**
: We could disable this option for now. 

- **enable-tflite-interpreter=false**
: Can be disabled when the Tensorflow lite is not available. 

- **enable-app=false**
: This app should be enabled for integrated tests and for the demonstration. 

- **ml-api-support='disabled'**
: Can be disabled.

- **enable-test=false**
: This should be enabled for the unit test. Without it, we are not sure whether everything works or not. 

- **install-app=false**
: It can be disable this, if we do not want to install applications.

- **enable-avx=false**
: It could be enabled for the X86 Architecture.

- **enable-mmap=false**
: It is newally added option. but I'm not sure whether the MMap is not supported in Windows. If it is, then is there any way to implement similar feature?

[built-in options]
- **werror=false**
: we enable this for better code quality. is there any similar compiler option for Windows?

nntrainer/tensor/manager.h
gkisalapl: This code use mmap not supported on windows and it's .... not used.

gkisalapl:
> * nabled for the X86 Arch

I agree with this comment, I'm going to add support for missing functionalities incrementally

PR #2905, opened 2025-02-03
[Wait for #2891][Schema] add schema directory

baek2sm:
add schema build directory for nntrainer, tf and onnx schema.

**Self evaluation:**


djeong20 (approved):
LGTM!

DonghakPark (approved):
LGTM!

nntrainer/compiler/meson.build
```diff
@@ -16,6 +16,9 @@ compiler_headers = [
   'compiler_fwd.h'
 ]
 
+nntrainer_schema_dir = meson.build_root() / 'schema'
+run_command('mkdir', '-p', nntrainer_schema_dir)
```
EunjuYang, line 20: What about creating a `schema` directory and moving `nntrainer_schema.fbs` and `onnx.proto` to their?

EunjuYang (commented):
Current update uses the schema directory as an output directory for the schema compilation.
For me, it sounds more natural to save `.proto` and `nntrainer_schema.fbs` in `schema` directory and work on the directory. What do you think?

nntrainer/compiler/meson.build
baek2sm: I think so! that's why it is a work in progress. I'll create a `nntrainer/schema` directory and move all the schema files into that folder. Thanks for your comment!!

EunjuYang: Oh I missed the tag! Thanks for your comment :)

baek2sm: I've modified it. thanks a lot!

skykongkong8 (approved):
LGTM!

EunjuYang:
It seems good to move `.fbs` and `.proto` under `/schema`. In my opinion, however, I'm not sure it is the best option to move all `*.cpp` and `*.h` files related to their interpretation to the schema folder.

jijoongmoon:
> It seems good to move `.fbs` and `.proto` under `/schema`. In my opinion, however, I'm not sure it is the best option to move all `*.cpp` and `*.h` files related to their interpretation to the schema folder.

yes. I agree with it. It might be better not to include headers and sources in /schema.

jijoongmoon (approved):
LGTM

EunjuYang (approved):
LGTM!

PR #2908, opened 2025-02-04
[Util] Floating-point and fixed-point conversion

djeong20:
This pull request introduces utility functions that convert floating-point numbers to fixed-point numbers and exponents, and vice versa. Note that the fixed-point values are scaled to maximize precision in use. This enhancement allows for the integration of a third-party computation library.

**Self-evaluation:**

### DIFF
--- a/nntrainer/utils/util_func.cpp
+++ b/nntrainer/utils/util_func.cpp
@@ -237,4 +237,20 @@ std::regex getRegex(const std::string &str) {
   return result;
 }
 
+void floatToFixedPointAndExponent(float input, int &fixedpoint, int &exponent) {
+  exponent = 0;
+  // normalize the floating-point number into the form: mantissa * 2^exponent
+  float mantissa = std::frexp(input, &exponent);
+  // scale mantissa to a fixed-point range to maximize precision
+  fixedpoint = static_cast<int>(mantissa * std::numeric_limits<int>::max());
+}
+
+float fixedPointAndExponentToFloat(int fixedpoint, int exponent) {
+  // scale back to the normalized floating-point range
+  float mantissa =
+    static_cast<float>(fixedpoint) / std::numeric_limits<int>::max();
+  // reconstruct the floating-point number
+  return std::ldexp(mantissa, exponent);
+}
+
 } // namespace nntrainer
--- a/nntrainer/utils/util_func.h
+++ b/nntrainer/utils/util_func.h
@@ -309,6 +309,26 @@ tm *getLocaltime(tm *tp);
  */
 std::regex getRegex(const std::string &str);
 
+/**
+ * @brief  Convert a floating-point number into its fixed-point component (the
+ * mantissa) and its exponent component.
+ *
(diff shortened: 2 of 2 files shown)


nntrainer/utils/util_func.cpp
```diff
@@ -237,4 +237,20 @@ std::regex getRegex(const std::string &str) {
   return result;
 }
 
+void floatToFixedPointAndExponent(float input, int &fixedpoint, int &exponent) {
+  exponent = 0;
```
DonghakPark, line 241: Could you Please Explain why exponent initialize 0 here?

DonghakPark (approved):
Nice work! it will be useful to apply various method that optimize Network

nntrainer/utils/util_func.cpp
djeong20: No specific reason; I thought it better to set it to 0 instead of using a random initial value.

skykongkong8 (commented):
What third party lib are we targeting to with this PR?

nntrainer/utils/util_func.h
```diff
@@ -309,6 +309,26 @@ tm *getLocaltime(tm *tp);
...
 
+/**
+ * @brief  Convert a floating-point number into its fixed-point component (the
+ * mantissa) and its exponent component.
+ *
+ * @param[in] input floating point to convert
+ * @param[out] fixedpoint fixed-point
+ * @param[out] exponent exponent
+ */
+void floatToFixedPointAndExponent(float input, int &fixedpoint, int &exponent);
```
SeoHyungjun, line 320: I think floatToFixedPointAndExponent is a clear name, but what do you think about using convertFloatToFixedPointAndExponent? I believe a name that includes the action of 'convert' is more intuitive for understanding the function's behavior.

```suggestion
void convertFloatToFixedPointAndExponent(float input, int &fixedpoint, int &exponent);
```

nntrainer/utils/util_func.h
```diff
@@ -309,6 +309,26 @@ tm *getLocaltime(tm *tp);
...
+
+/**
+ * @brief Convert a fixed-point number and an exponent into a floating-point
+ * number.
+ *
+ * @param[in] fixedpoint fixed-point
+ * @param[in] exponent exponent
+ * @return floating point result
+ */
+float fixedPointAndExponentToFloat(int fixedpoint, int exponent);
```
SeoHyungjun, line 330: ditto

```suggestion
float convertFixedPointAndExponentToFloat(int fixedpoint, int exponent);
```

SeoHyungjun (approved):

SeoHyungjun (commented):
LGTM

nntrainer/utils/util_func.cpp
baek2sm: @DonghakPark `exponent` variable looks to be used solely for returning the result value.

baek2sm (approved):
LGTM

PR #2909, opened 2025-02-04
Add git submodules: ruy

djeong20:
This PR adds the ruy matrix multiplication library as a submodule.

**Self-evaluation:**
2. Run test:   [ ]Passed [ ]Failed [X]Skipped


EunjuYang (approved):

skykongkong8 (approved):
LGTM!

DonghakPark (approved):
LGTM!

gkisalapl (commented):
I think that we need to make architectural decision in terms of how we deal with 3rd party.

Currently iniparser is configured by jni/prepare_iniparser.sh script which is not standard way
googletest - after quick check I couldn't find how we use it (probably we assume that it's installed in system)
Now you're adding ruy with git submodules which is fine but it seems that in meson wraps are more standard way of adding dependencies: https://mesonbuild.com/Wrap-dependency-system-manual.html

In my opinion we should discuss it, select one way of using third party and stick with one method

djeong20:
Thank you for sharing your thoughts @gkisalapl ! I agree that our current use of iniparser is not ideal and needs to be improved.

Regarding GoogleTest, you are correct that we assume it is already installed on the system. While we utilize ndk-build for the Android build, which includes GoogleTest in the Android NDK, we still need to copy GoogleTest to the test/jni directory. Instructions for this can be found in the [documentation](https://github.com/nnstreamer/nntrainer/blob/main/test/jni/README.md). However, I believe it would be better to integrate GoogleTest as a submodule rather than using the current method.

As for adding Ruy as a submodule, I intentionally chose not to use Meson subprojects. This decision stems from the fact that Ruy itself contains submodules that are necessary for a proper build. Including these would require extra steps, which might confuse developers. In NNTrainer, my plan is not to build Ruy.

I think adding iniparser and GoogleTest as Git submodules would be a good choice. Please share your thoughts!

gkisalapl:
> Thank you for sharing your thoughts @gkisalapl ! I agree that our current use of iniparser is not ideal and needs to be improved.
> 
> Regarding GoogleTest, you are correct that we assume it is already installed on the system. While we utilize ndk-build for the Android build, which includes GoogleTest in the Android NDK, we still need to copy GoogleTest to the test/jni directory. Instructions for this can be found in the [documentation](https://github.com/nnstreamer/nntrainer/blob/main/test/jni/README.md). However, I believe it would be better to integrate GoogleTest as a submodule rather than using the current method.
> 
> As for adding Ruy as a submodule, I intentionally chose not to use Meson subprojects. This decision stems from the fact that Ruy itself contains submodules that are necessary for a proper build. Including these would require extra steps, which might confuse developers. In NNTrainer, my plan is not to build Ruy.
> 
> I think adding iniparser and GoogleTest as Git submodules would be a good choice. Please share your thoughts!

Thank you for explanation, no I understand Your motivation behind using git submodules

gkisalapl:
> Before proceed, please be aware:
> 
> 1. Tizen build system does NOT allow git submodule. (build will fail)
> 2. I'm not sure with ppa-launchpad-pdebuild for git submodule behaviors.
> 3. You won't use platform's standard libraries if you use git submodule. Thus, you will generate inefficient binaries with usual Linux distros (e.g., Ubuntu). Deployment process of the included module (ruy) should be defined, too. Is it a shared library?
> 
> If you want git-submodule for platform-supported packages (e.g., iniparser), make sure that builds for most Linux distro (Tizen and Ubuntu included) keep using platform packages and the git-submodule for such packages are activated in other operating systems only.

Based on what you write I would suggest to follow following scheme:

1. Use system library if possible
2. If system library not found use git submodules

Do You agree ?

SeoHyungjun:
What do you think about creating a script to build ruy instead of adding it as a submodule?

In my opinion, if we create a script and decide whether to build it based on options, we can maintain compatibility with the Tizen build system.
Also, isn't it possible without significantly changing the build process?

myungjoo:
> > Before proceed, please be aware:
> > 
> > 1. Tizen build system does NOT allow git submodule. (build will fail)
> > 2. I'm not sure with ppa-launchpad-pdebuild for git submodule behaviors.
> > 3. You won't use platform's standard libraries if you use git submodule. Thus, you will generate inefficient binaries with usual Linux distros (e.g., Ubuntu). Deployment process of the included module (ruy) should be defined, too. Is it a shared library?
> > 
> > If you want git-submodule for platform-supported packages (e.g., iniparser), make sure that builds for most Linux distro (Tizen and Ubuntu included) keep using platform packages and the git-submodule for such packages are activated in other operating systems only.
> 
> Based on what you write I would suggest to use following scheme:
> 
> 1. Use system library if possible
> 2. If system library not found use git submodules
> 
> Do You agree ?

Yes. That'd be good. (As in #2883)

meson.build
```diff
@@ -55,7 +55,8 @@ warning_flags = [
   '-Wdefaulted-function-deleted',
   '-ftree-vectorize',
   '-Wno-maybe-uninitialized',
-  '-Wno-unused-variable'
+  '-Wno-unused-variable',
+  '-Wno-comment'
```
djeong20, line 59: This addition was made to prevent an error in the Ruy source code during the GBS build.

djeong20:
> For Ruy, anyway, you may need to package ruy (to make it a system library in Tizen and Ubuntu) or have its sources packaged (.zip, .tar.gz) in this git repo until you get ruy packaged properly.

@myungjoo I've applied your suggestions! Please take a moment to review the changes. Thank you!

myungjoo (approved):

PR #2912, opened 2025-02-05
[Coverity] Fix DIVIDE_BY_ZERO issue

EunjuYang:
- This commit resolves coverity issue of DIVIDE_BY_ZERO.
- This commit updatees preprocess_l2norm_layer.cpp/.h
- In order to fix the DIVIDE_BY_ZERO issue, epsilon property is added to l2norm_layer; the value is used when the norm is divided.

Self evaluation:

### DIFF
--- a/nntrainer/layers/preprocess_l2norm_layer.cpp
+++ b/nntrainer/layers/preprocess_l2norm_layer.cpp
@@ -39,6 +39,7 @@ void PreprocessL2NormLayer::finalize(InitLayerContext &context) {
 
 void PreprocessL2NormLayer::forwarding(RunLayerContext &context,
                                        bool training) {
+  const float epsilon = std::get<props::Epsilon>(l2norm_props);
   auto &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
 
   auto &input_ = context.getInput(SINGLE_INOUT_IDX);
@@ -46,7 +47,7 @@ void PreprocessL2NormLayer::forwarding(RunLayerContext &context,
   for (unsigned int b = 0; b < input_.batch(); ++b) {
     auto input_slice = input_.getBatchSlice(b, 1);
     auto hidden_slice = hidden_.getBatchSlice(b, 1);
-    input_slice.multiply(1 / input_slice.l2norm(), hidden_slice);
+    input_slice.multiply(1 / (input_slice.l2norm() + epsilon), hidden_slice);
   }
 }
 
@@ -57,11 +58,10 @@ void PreprocessL2NormLayer::calcDerivative(RunLayerContext &context) {
 
 void PreprocessL2NormLayer::setProperty(
   const std::vector<std::string> &values) {
-  if (!values.empty()) {
-    std::string msg = "[FlattenLayer] Unknown Layer Properties count " +
-                      std::to_string(values.size());
-    throw exception::not_supported(msg);
-  }
+  auto remain_props = loadProperties(values, l2norm_props);
+  NNTR_THROW_IF(!remain_props.empty(), std::invalid_argument)
+    << "[PreprocessL2Norm Layer] Unknown Layer Properties count " +
+         std::to_string(remain_props.size());
 }
 
 } // namespace nntrainer
--- a/nntrainer/layers/preprocess_l2norm_layer.h
+++ b/nntrainer/layers/preprocess_l2norm_layer.h
@@ -31,7 +31,7 @@ public:
    * @brief Construct a new L2norm Layer object
    * that normalizes given feature with l2norm
    */
-  PreprocessL2NormLayer() : Layer() {}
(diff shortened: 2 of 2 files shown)


skykongkong8 (approved):
LGTM

nntrainer/layers/preprocess_l2norm_layer.cpp
```diff
@@ -39,14 +39,15 @@ void PreprocessL2NormLayer::finalize(InitLayerContext &context) {
...
+  const float epsilon = std::get<props::Epsilon>(l2norm_props);
   auto &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
 
   auto &input_ = context.getInput(SINGLE_INOUT_IDX);
 
   for (unsigned int b = 0; b < input_.batch(); ++b) {
     auto input_slice = input_.getBatchSlice(b, 1);
     auto hidden_slice = hidden_.getBatchSlice(b, 1);
-    input_slice.multiply(1 / input_slice.l2norm(), hidden_slice);
+    input_slice.multiply(1 / (input_slice.l2norm() + epsilon), hidden_slice);
```
myungjoo, line 50: One silly question: (don't know if you can use .divide for this)
```
input_slice.divide(input_slice.l2norm(), hidden_slice);
```
and let Tensor::divide to emit a large value if it is divided by zero with some additional 1) parameter, 2) property, or 3) method-name?
?

(around Line 467~514 of tensor.cpp)

1)
```
input_slice.divide(input_slice.l2norm(), hidden_slice, true);
```
2)
```
input_slice.allow_dbz = true;
input_slice.divide(input_slice.l2norm(), hidden_slice);
```
3)
```
input_slice.divide_prevent_dbz(input_slice.l2norm(), hidden_slice);
```

EunjuYang: Thank you for your opinion. Your suggestion was one of options I considered. The reason I updated to take epsilon in the layer side, however, is related to the layer's action.

In current, several ML frameworks take epsilon as a parameter of several normalization layers [[example](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html)]. Even their default values are different (keras = 0.001, pytorch = 1e-5f); so the layer's operation should be varied by the parameter from user to support accurate model conversion. In this context, I thought it would better to handle this dbz in this layer, not tensor level. What do you think?

By the way, I think we need to adopt your opinion in tensor side, since the current tensor simply throws an error when it faces dbz.

dkjung (approved):
LGTM

nntrainer/layers/preprocess_l2norm_layer.cpp
skykongkong8: Actually from the kernel level, we already discussed similar issue from #2466!
From the kernel's perspective, it is perfectly fine to get zero values (or near-zero) as denominators.
They will return with SIMD vector filled with inf, or -inf values. (sign determined by numerators)

myungjoo: I've leave this matter to you @EunjuYang @dkjung @skykongkong8 @DonghakPark  .

I just wanted to throw another possible option. Don't wait for my decisions, make a conclusion and apply it. :)

skykongkong8: If that's the case, I think we have following options:

1. remove all error-throwing codes related zero division. (compute kernel will return inf values anyway)
2. add certain epsilon to it like other frameworks do.
  - what should we do if, for example, divided output is bigger than `FP16` max range? Should we saturate it? 
  - if the denominator is given as a Tensor (not a scalar), should we check for all values to make sure is there any zero values to add eps?

djeong20, line 50: Wouldn't this impact cases where the L2 norm is not zero?

djeong20: We have chosen to block division by zero when working with scalar values. However, we allow division by tensors that contain zero values, which can lead to infinite results. 

As @skykongkong8 pointed out, permitting division by zero at the kernel level does not present any issues. Other frameworks handle both scenarios: dividing by zero and dividing by tensors that contain zero values, which also yield infinite results.

In my opinion, if we allow division by zero with scalar, it should return 'inf' rather than a large value.

EunjuYang: Good point. Let me check it in terms of our layer's definition.
I thought the preprocess_l2norm_layer is related to LayerNorm of pytorch [[link](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html)]. However, if it is not the layer norm but normalization function, I need to update it like
```suggestion
    input_slice.multiply(1 / max(input_slice.l2norm(), epsilon), hidden_slice);
```
Please let me check it and make another PR if it corresponds to the latter case.

PR #2913, opened 2025-02-05
[Coverity] Fix double_unlock issue in markFilled()

EunjuYang:
- This PR resolves coverity issue (number : 1831328) in recent report. I verified this PR resolves the coverity issue.
- The `unique_lock` was manually unlocked before going out of scope, causing a double unlock issue.
- This has been fixed by removing the explicit `unlock()` call.
- Additionally, `notify_emptied_cv.notify_all()` is now called without holding the lock to avoid unnecessary cntention.

FYI) I initially thought it would be fine to simply remove the unlock code line (no issue with notify_all) , but I decided to update it to a scoped lock just in case.

**Self evaluation:**

### DIFF
--- a/nntrainer/dataset/iteration_queue.cpp
+++ b/nntrainer/dataset/iteration_queue.cpp
@@ -153,10 +153,11 @@ void IterationQueue::notifyEndOfRequestEmpty() {
 }
 
 void IterationQueue::markFilled(MarkableIteration *iteration) {
-  std::unique_lock lg(empty_mutex);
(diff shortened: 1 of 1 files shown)


skykongkong8 (approved):
Never knew using parenthesis like this enables scoped lock. Thanks!

dkjung (approved):
LGTM

dkjung (approved):
LGTM

myungjoo (approved):

PR #2919, opened 2025-02-06
[Coverity] Fix Coverity issue at neuralnet destructor

DonghakPark:
Uncaught exception (UNCAUGHT_EXCEPT)

exn_spec_violation: An exception of type std::length_error is thrown but the exception specification /*implicit*/noexcept doesn't allow it to be thrown. This will result in a call to terminate().
- in general, throw exception at destrctor is not allowed
- so, add exception at deallocate Tensor

Resolves:

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -1008,9 +1008,14 @@ int NeuralNetwork::allocate(ExecutionMode mode) {
 }
 
 int NeuralNetwork::deallocate() {
-  model_graph.deallocateTensors(true);
-
-  return ML_ERROR_NONE;
+  try {
+    model_graph.deallocateTensors(true);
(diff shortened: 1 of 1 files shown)


dkjung (commented):
It looks weird that this function always returns `ML_ERROR_NONE` if an exception is thrown or not.

It should return an error code in case of an exception, shouldn't it?

DonghakPark:
> It looks weird that this function always returns `ML_ERROR_NONE` if an exception is thrown or not.
> 
> It should return an error code in case of an exception, shouldn't it?

your Right!! i will fix now!! Thank you for review

myungjoo (approved):

lhs8928 (approved):
LGTM

EunjuYang (approved):
LGTM!

PR #2906, opened 2025-02-03
[Wait For #2854][FSU] Modify SimpleFC Application for FSU-CPU Test @open sesame 02/10 14:56

DonghakPark:
Modiy Simple FC Application for FSU-CPU Side TEST 

commit summary
commit 1 [[Application] Update FSU SimpleFC Application](https://github.com/nnstreamer/nntrainer/commit/c0569b52f0ceb4f54de85329fec028e279f1f4e2) : update SimpleFC Application for more easy test & actual environment
commit 2 [[FSU] update layer weight load logic at fsu](https://github.com/nnstreamer/nntrainer/commit/950c5d50efdb2a038d5bc228b11effb3e660d3c3) : update layernode not to load weight at load()

commit 3 [[FSU] Modify Application for CPU side Test](https://github.com/nnstreamer/nntrainer/pull/2906/commits/ec4e82c9cac230dda47ef6503a786de36f61e5c6) : update Application

Self evaluation:

Build test: [X]Passed [ ]Failed [ ]Skipped
Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Donghak PARK [donghak.park@samsung.com](mailto:donghak.park@samsung.com)

PR #2927, opened 2025-02-06
[Wait For #2906] [FSU] file offset

lhs8928:
[fsu] add file offset info
     - Store file offset info
     - Added setMemorySwapPath
    
    Signed-off-by: hyeonseok <hs89.lee@samsung.com>

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -24,6 +24,8 @@
 #include <bcq_tensor.h>
 #endif
 
+#include <iostream>
+
 namespace nntrainer {
 
 Tensor::Tensor(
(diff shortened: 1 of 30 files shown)


jijoongmoon:
I think this is working but How about this?
when we load the weight, we can generate file offset without reading the binary file. We can just do the loop and get the weight for each layer and compute the offset. and then we can set the vector of weight offset of binary file directly. 

Then we don't need to change the optimizer and other files.

Also we do need to consider which file we have to use. If it is inference and FSU, we can use weight binary file at the load, but if it is training and FSU, then we do need to make copy of Swap path file  and compute file offset at the same time. We do not want to  overwrite the update weight into origitna weight binary file.

nntrainer/tensor/tensor.cpp
```diff
@@ -1143,6 +1145,8 @@ void Tensor::save(std::ostream &file) {
     << "save size: " << bytes() + scale_size() * sizeof(float)
     << " is too big. It cannot be represented by std::streamsize";
 
+  std::cerr << getData<_FP16>()[0] << "\n";
```
EunjuYang, line 1148: It seems okay to remove this line.

nntrainer/tensor/tensor.cpp
```diff
@@ -24,6 +24,8 @@
 #include <bcq_tensor.h>
 #endif
 
+#include <iostream>
```
EunjuYang, line 27: ditto.

EunjuYang (commented):
Overall LGTM. Please check the following comments:

DonghakPark:
Close This PR for reorganize and resubmit it so that can be reviewed conveniently. Closed By #2969

PR #2928, opened 2025-02-07
[Coverity] Fix Coverity issue on ini_interpreter (path nullptr excetion)

DonghakPark:
For Resolve below issue
- Return value of a function 'iniparser_getstring' is dereferenced at ini_interpreter.cpp:350 without checking for null, but it is usually checked for this function

add NNTR_THROW_IF() to check backbone_path is null when backbone_path is nullptr then it will print error msg

**Self evaluation:**


skykongkong8 (approved):
Nice catch!

haehun (approved):
This should fix Coverity issue

jijoongmoon (approved):
LGTM

PR #2897, opened 2025-01-23
Add arm-native build and test on Ubuntu (wait for #2910) @open sesame 02/05 17:05

myungjoo:
We can run arm natively with github-action.
Test fp16 natively on arm machines!

PR #2930, opened 2025-02-07
[ cpu_backend ] Refactor blas_interface considering arch-dep

skykongkong8:
Apologies for the HUGE pr, but this change is quite essential for the upcoming works for the nntrainer.
I left `README.md` file below `cpu_backend` directory, and the aim of the PR can be summarized in 3 lines like below:

1. Substitute `blas_interface.h` to `cpu_backend.h` which has virtual functions of `blas_interface.h`
2. Actual implementations are implemented at `arm_compute_backend`, `x86_compute_backend`, and `fallback`, and they are included considering target cpu architecture. `cblas.h` is used for both of them for fp32 computation.
3. There are some differences (unsupported intrinsics, or dataTypes ) along the versions, and they are managed under each `arm` or `x86` directory.

From now on, at least under cpu_backend, all fp16, and SIMD-related options should be managed "under" pre-determined architecture.

### AS-IS
```
tensor
├─── blas_interface.h : header interface to use from `tensor` & `layer`
├─── blas_interface.cpp : set of `cblas.h`, `blas_neon.h`, `blas_avx.h`, and static fallback functions, divided by `#ifdef`s
├─── blas_neon.h : header of neon impl
├─── blas_neon.cpp : source code of neon impl (with blas_neon.h)
├─── blas_avx.h : header of avx impl
├─── blas_avx.cpp : source code of avx2 impl (with immintrin.h)
├─── matrix_transpose_neon : (only for arm) neon SIMD matrix transpose impl
├─── hgemm : (only for arm) neon SIMD half-precision GEMM impl
```
### TO-BE
```
...
tensor
├─── cpu_backend
│   └─── cpu_backend.h (has all external function interfaces for tensor operations)
│   └─── arm
│      └─── arm_compute_backend : copy of `cpu_backend.h`. selectively uses functions from `cblas_interface`, `neon_impl`, `fallback_internal`
│      └─── neon_impl : custom implementation based on neon SIMD intrinsic
│      └─── neon_impl_fp16 (For armv8.2+)
│      └─── armv7_neon (For armv7l)
│      └─── matrix_transpose_neon
│      └─── hgemm
│           ...
│   └─── x86
│      └─── x86_compute_backend : copy of `cpu_backend.h`. selectively uses functions from `cblas_interface`, `avx2_impl`, `fallback_internal`
│      └─── avx2_impl : custom implementation based on avx2 SIMD intrinsic
│           ...
│   └─── fallback
│      └─── fallback ( !x86 & !arm ) : copy of `cpu_backend.h`. uses ALL functions from `fallback_internal`. This is for additional architecture (!x86 && !arm)
│      └─── fallback_internal : all raw implementations without SIMD or external lib
│           ...
│   └─── cblas_interface
│      └─── cblas_interface : all cblas-related function interfaces, and params
...
```

**Self evaluation:**


nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
```diff
@@ -0,0 +1,315 @@
...
+                      const unsigned int ldc) {
+  for (unsigned int m = 0; m < M; ++m) {
+    for (unsigned int n = 0; n < N; ++n) {
+      double c = 0.0;
+      float c_old = C[m * ldc + n];
+      for (unsigned int k = 0; k < K; ++k) {
+        float a, b;
+        a = ((TransA == true) ? A[k * lda + m] : A[m * lda + k]);
+        b = ((TransB == true) ? B[n * ldb + k] : B[k * ldb + n]);
+        c += a * b;
```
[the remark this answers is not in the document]
skykongkong8: Nope. `c` is defined as `double` variable already.

skykongkong8:
I could observe some format CI error even after applying clang-format...
Mine :
```
Ubuntu clang-format version 14.0.0-1ubuntu1.1
```
[.github/workflows/cpp_linter.yml](https://github.com/nnstreamer/nntrainer/blob/44d9efa14e3cb2d0c8118287e27ac022ba7a076b/.github/workflows/cpp_linter.yml)
```yml
...
version: 14
...
```

dkjung:
It seems that the PR does not pass `C++ Format Checker`. Could you check if the format?

dkjung (approved):
LGTM

skykongkong8:
> It seems that the PR does not pass `C++ Format Checker`. Could you check the format?

I applied clang-format 14.0.0 already.

baek2sm (approved):
great work.. LGTM

nntrainer/tensor/char_tensor.cpp
```diff
@@ -8,11 +8,10 @@
...
  */
 
+#include <char_tensor.h>
+#include <cpu_backend.h>
 #include <iomanip>
 #include <iostream>
-
-#include <blas_interface.h>
-#include <char_tensor.h>
 #include <tensor.h>
```
djeong20, line 15: I thought we usually separate the standard library and nntrainer headers, with the standard library at the top and the nntrainer headers at the bottom.

``` C++
#include <iomanip>
#include <iostream>

#include <char_tensor.h>
#include <cpu_backend.h>
#include <tensor.h
```

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -0,0 +1,210 @@
...
+ */
+#include <arm_compute_backend.h>
+#include <assert.h>
+#include <cblas_interface.h>
+#include <fallback_internal.h>
+#include <neon_impl.h>
+#include <nntrainer_error.h>
+
+#define ROW_MAJOR 0
+#define COL_MAJOR 1
```
djeong20, line 21: quick question! are these currently used?

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -0,0 +1,210 @@
...
+                                 float *sin_, unsigned int alpha) {
+  nntrainer::neon::calc_trigonometric_vals_dup(N_half, angle, cos_, sin_,
+                                               alpha);
+}
+
+void swiglu(const unsigned int N, float *X, float *Y, float *Z) {
+  nntrainer::neon::swiglu(N, X, Y, Z);
+}
+
+float max(const unsigned int N, float *X) { return nntrainer::neon::max(N, X); }
```
djeong20, line 35: IMO, using the same function name as standard types could lead to undefined behavior. Why not rename it to something different, like get_max or max_val?

jni/meson.build
```diff
@@ -25,6 +25,12 @@ and_conf.set('VERSION_MAJOR', nntrainer_version_split[0])
 and_conf.set('VERSION_MINOR', nntrainer_version_split[1])
 and_conf.set('VERSION_MICRO', nntrainer_version_split[2])
 
+arch = host_machine.cpu_family()
+and_conf.set('ARM', 1)
+if arch == 'arm'
+  and_conf.set('ARMV7', 1)
+endif
+
```
djeong20, line 33: This is a configuration used in Android.mk.in, so it is unnecessary.

```suggestion
```

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -0,0 +1,592 @@
...
+/**
+ * @brief     check if X array has NaN or inf
+ * @param[in] N  length of the vector
+ * @param[in] X float * for Vector X
+ * @param[out] bool false if not valid else true
+ */
+bool is_valid(const unsigned int N, const float *X);
+} /* namespace nntrainer */
+#endif /* __cplusplus */
+#endif /* __ARM_COMPUTE_BACKEND_H__ */
```
djeong20, line 592: Please review the function descriptions and include explanations for the newly added parameters (e.g., incX, incY).

djeong20, line 592: I know this idea requires a considerable amount of work and is not urgent, but I’d like to suggest reordering the parameters for the function that uses `incX` and `incY`. In most cases, both values are set to 1, but due to the current order, we always have to specify 1 when calling the function. 
``` C++
sdot(K, data, 1, mdata, 1);
```
By reordering the parameters, we can simplify usage and avoid confusing users of these functions.
``` C++
sdot(const unsigned int N, const float *X, const float *Y,
        unsigned int incX = 1, unsigned int incY = 1);

...

sdot(K, data, mdata);
```

djeong20 (commented):
I haven't finished the review yet. I will continue later, but please take a look at the comments in the meantime!

meson.build
```diff
@@ -74,29 +74,26 @@ arch = host_machine.cpu_family()
...
+  extra_defines += '-DUSE_NEON=1'
+  if arch == 'arm'
+    extra_defines += '-DARMV7=1'
   endif
+elif arch == 'x86' or arch == 'x86_64'
+  message('Build for X86 architecture')
+  add_project_arguments(['-march=native'], language: ['c','cpp'])
+  add_project_arguments(['-mavx2'], language: ['c','cpp'])
+  message('-march=native added for AVX hardware acceleration.')
+  extra_defines += '-DX86=1'
```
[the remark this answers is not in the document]
skykongkong8: You are right. But just like neon option in current android build, every target x86 devices have AVX2 intrinsics so I just let them work as a default option. Besides, if the user does not want to use additional HW accelerator, they can just build with fallback so that they can use pure vanilla compute codes.

nntrainer/tensor/char_tensor.cpp
skykongkong8: Didn't really aware of it. Thanks for pointing it out!

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
skykongkong8: Will be fixed!

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
skykongkong8: Good idea

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
skykongkong8: I totally agree, but postponed this work because it might evoke even more conflicts with on-going PRs. But still, I also feel quite uncomfortable about current function parameter orders. I will definitely come back to this afterwards, but maybe not in this PR.

meson.build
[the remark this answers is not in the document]
djeong20: Understood! If that's the case, let's take out the option `enable-avx` from the meson_options.txt file.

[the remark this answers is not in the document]
skykongkong8: Sure!

debian/nntrainer-dev.install
```diff
@@ -16,11 +16,14 @@
...
-/usr/include/nntrainer/blas_interface.h
+/usr/include/nntrainer/fallback_internal.h 
+/usr/include/nntrainer/cblas_interface.h
+/usr/include/nntrainer/x86_compute_backend.h
+/usr/include/nntrainer/cpu_backend.h
 /usr/include/nntrainer/var_grad.h
 /usr/include/nntrainer/weight.h
 /usr/include/nntrainer/quantizer.h
-/usr/include/nntrainer/blas_avx.h
+/usr/include/nntrainer/avx2_impl.h
```
djeong20, line 26: don't we have to include neon-related headers too?

djeong20: Wait, we should install these headers by its architecture. The current condition is strictly limited to x86_64. This won't be valid for ARM build on Ubuntu (#2897)

nntrainer/tensor/cpu_backend/arm/arm_compute_backend_fp16.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+_FP16 sdot(const unsigned int N, const _FP16 *X, const unsigned int incX,
+           const _FP16 *Y, const unsigned int incY) {
+  _FP16 ret = 0;
+  assert(incX > 0 && incY > 0);
+  if (incX == 1 && incY == 1) {
+    ret = nntrainer::neon::hdot(N, X, Y);
+  } else {
+    __fallback_sdot(N, X, incX, Y, incY);
```
djeong20, line 110: let's change the naming as well in the near future (sdot for single-precision, hdot for half-precision, etc)

nntrainer/tensor/cpu_backend/arm/matrix_transpose_neon/meson.build
```diff
@@ -0,0 +1,21 @@
...
+    matrix_transpose_neon_headers += 'matrix_transpose_kernels_neon.h'
+endif
+
+foreach s : matrix_transpose_neon_sources
+  nntrainer_sources += meson.current_source_dir() / s
+endforeach
+
+foreach h : matrix_transpose_neon_headers
+  nntrainer_headers += meson.current_source_dir() / h
+endforeach
```
djeong20, line 21: could you explain why matrix transpose is separated from other neon_impl?

nntrainer/tensor/cpu_backend/arm/meson.build
```diff
@@ -0,0 +1,38 @@
...
+    'neon_mathfun.h',
+    'neon_mathfun.hxx'
+]
+arm_compute_backend_sources = [
+    'arm_compute_backend.cpp',
+    'neon_impl.cpp'
+]
+
+
+subdir('matrix_transpose_neon')
```
djeong20, line 14: could you explain why matrix transpose is separated from other neon_impl?

debian/nntrainer-dev.install
skykongkong8: I also thought that current condition only considered x86_64, so I just referred to it.

nntrainer/tensor/cpu_backend/arm/matrix_transpose_neon/meson.build
skykongkong8: Nothing special. Just like hgemm, it needs too many additional codes unlike other functions do. On top of that, if we have to support 8bit transpose or other datatypes in the future, this will also expand as well like hgemm.

debian/nntrainer-dev.install
djeong20: I think we should handle this before #2897 is merged. I'll leave an issue regarding this and let's discuss it.

skykongkong8: I think we can add patch after merging this one. #2897 is not merged yet, and it would take more time to have solid conclusion from #2966.

nntrainer/tensor/tensor.h
```diff
@@ -23,7 +23,9 @@
 
 #include <cstddef>
 
-#include <blas_interface.h>
+#include <char_tensor.h>
+#include <cpu_backend.h>
+#include <float_tensor.h>
```
djeong20, line 28: no need to include char_tensor.h and float_tensor.h!

```suggestion
#include <cpu_backend.h>
```

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
```diff
@@ -0,0 +1,177 @@
...
+  return __cblas_snrm2(N, X, incX);
+}
+
+void sgemm(const unsigned int TStorageOrder, bool TransA, bool TransB,
+           const unsigned int M, const unsigned int N, const unsigned int K,
+           const float alpha, const float *A, const unsigned int lda,
+           const float *B, const unsigned int ldb, const float beta, float *C,
+           const unsigned int ldc) {
+  __cblas_sgemm(TStorageOrder, TransA, TransB, M, N, K, alpha, A, lda, B, ldb,
+                beta, C, ldc);
```
djeong20, line 144: Since we use CBLAS for this, it is dependent on CBlas. If we set `enable-blas` option as false, it throws an error.

```
/home/donghyeon/nntrainer/nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp:17:10: fatal error: cblas_interface.h: No such file or directory
   17 | #include <cblas_interface.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

djeong20: There are several options on how to handle this, let's discuss them. We need to define whether we will use OpenBLAS only on x86 or also on ARM and whether we will force the use of OpenBLAS (by removing the enable-blas option and making it always true) or add a fallback to the current condition. Which option would be better?

djeong20 (changes requested):

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
skykongkong8: I prefer the latter. We are perfectly relying on cblas lib for fp32 SIMD kernels. I personally believe that there is no point for checking enable-blas option.

djeong20: We are currently disabling the enable-blas options on Windows (please refer to #2904 and #2962). If this PR is merged, it may cause problems for those working on Windows.

skykongkong8: I see! Then I could think of 2 options--plz share your opinion.

1. Use single-lib per op/arch
According to @jijoongmoon 's comment on #2904 , he mentioned MKL for making enable-blas option alive. However, point of using MKL for our circumstance is to support different tensor dim order(channel-first), so we can actually include it with default openblas option, and minimize combinations of multiple options.

2. Allow multiple-lib backend options per op/arch
Maybe we can make a new file structure to manage multiple combinations of external BLAS-like libs.
However this might make unexpected complexity. For example, there might be some ops that some libs do support and some libs don't (maybe per-OP, per-datatype, ... ). In this case, we might have to expand such files per-OPs. I wonder if this is super urgent to fix this right now.
```
...
tensor
├─── cpu_backend
│   └─── cpu_backend.h (has all external function interfaces for tensor operations)
│   └─── arm
│   └─── x86
│    ...
│   └─── external_interface (new!)
│      └─── external_interface.h (let meson include either cblas_interface or mkl_interface with build option)
│      └─── cblas_interface
│      └─── mkl_interface (tentative)
│      └─── ruy_interface (tentative)'
│       ...
...
```

skykongkong8: I personally prefer jumping on to option#1 (First, I just want to simplify current cpu_backend build options...)
And gradually moving on to option#2 when we really have to support them in the future.

djeong20: I agree. Let's work on the first option and move on to the second option later on :)

nntrainer/tensor/cpu_backend/fallback/fallback_internal_fp16.cpp
```diff
@@ -0,0 +1,298 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Sungsik Kong <ss.kong@samsung.com>
+ *
+ * @file fallback_internal.cpp
+ * @date   23 April 2024
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Sungsik Kong <ss.kong@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  Single-precision computation functions based on NEON
```
EunjuYang, line 10: Please check the brief.

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
skykongkong8: Right. Then I think I should make `enable-blas` option to be true for a default option.
Anyway, thanks for sharing your thoughts actively although this PR is quite humongous😂 Please don't hesitate to point out any mistakes I made :)

EunjuYang (approved):
LGTM

nntrainer/tensor/cpu_backend/fallback/fallback_internal_fp16.cpp
skykongkong8: I double-checked briefs in other files as well. Thanks!

Applications/Android/NNDetector/app/src/main/jni/Android.mk
```diff
@@ -28,17 +28,15 @@ include $(CLEAR_VARS)
...
 SIMPLESHOT_DIR = .
 
-
 LOCAL_ARM_NEON := true
-LOCAL_CFLAGS += -std=c++17 -Ofast -mcpu=cortex-a53 -Ilz4-nougat/lib
-LOCAL_LDFLAGS += -Llz4-nougat/lib/obj/local/$(TARGET_ARCH_ABI)/
-LOCAL_CXXFLAGS += -std=c++17 -frtti -fexceptions
+LOCAL_CFLAGS += -std=c++17 -Ofast -mcpu=cortex-a53 -Ilz4-nougat/lib -DARM=1 
+LOCAL_LDFLAGS += -Llz4-nougat/lib/obj/local/$(TARGET_ARCH_ABI)/ -DARM=1 
+LOCAL_CXXFLAGS += -std=c++17 -frtti -fexceptions -fopenmp -static-openmp -DARM=1
```
jijoongmoon, line 34: how about using complier macro to check the architecture in code?
```c++
 #if defined(__aarch64__) 
    std::court << "ARM64 architecture"<<std::endl;
#elif define(__ARM_ARHC_7A__)
   std::cout << "ARMv7-A (32-bit) architecture"<<std::endl;
#elif define(__x86_64__)
  std::cour << "x86_64 (64-bit) architecture"<<std::endl;
#else 
    std::cout << "Unknown Architecture" <<std::endl;
#endif
```

Applications/Custom/LayerClient/jni/Android.mk
```diff
@@ -22,6 +22,9 @@ NNTRAINER_INCLUDES := $(NNTRAINER_ROOT)/nntrainer \
 	$(NNTRAINER_ROOT)/nntrainer/utils \
 	$(NNTRAINER_ROOT)/nntrainer/optimizers \
 	$(NNTRAINER_ROOT)/nntrainer/tensor \
+	$(NNTRAINER_ROOT)/nntrainer/tensor/cpu_backend \
+	$(NNTRAINER_ROOT)/nntrainer/tensor/cpu_backend/fallback \
+	$(NNTRAINER_ROOT)/nntrainer/tensor/cpu_backend/arm \
```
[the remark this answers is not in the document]
skykongkong8: Got it. Thanks for your review

Applications/Android/NNDetector/app/src/main/jni/Android.mk
skykongkong8: Fixed.

djeong20 (approved):
Excellent work! Appreciate all the hard work you put in 👍

myungjoo:
Please resolve cpp-linter error and merge quickly. A large PR quickly gets merge conflicts.

skykongkong8:
> Please resolve cpp-linter error and merge quickly. A large PR quickly gets merge conflicts.

I already applied all the formats and it passes on my local(no suggested changes when applying cpp-linter and clang-format). I wonder why this fails.

skykongkong8:
> Please resolve cpp-linter error and merge quickly. A large PR quickly gets merge conflicts.

Seems this is not an issue for this PR. Expecting #2983 to solve this!

PR #2932, opened 2025-02-10
[Tensor] Scale factors support for unsigned int type

djeong20:
This pull request modifies the unsigned int type Tensor class to include scale factors in memory.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/uint_tensor.cpp
+++ b/nntrainer/tensor/uint_tensor.cpp
@@ -15,20 +15,21 @@
 #ifdef __UINT_TENSOR_H__
 
 template <typename T>
-UIntTensor<T>::UIntTensor(std::string name_, Tformat fm) :
-  TensorBase(name_, fm, checkTensorDataType()) {}
+UIntTensor<T>::UIntTensor(std::string name_, Tformat fm, QScheme qscheme_) :
+  TensorBase(name_, fm, checkTensorDataType()), qscheme(qscheme_) {}
 
 template <typename T>
 UIntTensor<T>::UIntTensor(const TensorDim &d, bool alloc_now, Initializer init,
-                          std::string name) :
-  TensorBase(d, alloc_now, init, name) {
+                          std::string name, QScheme qscheme_) :
+  TensorBase(d, alloc_now, init, name), qscheme(qscheme_) {
   if (alloc_now)
     allocate();
 }
 
 template <typename T>
-UIntTensor<T>::UIntTensor(const TensorDim &d, const void *buf) :
-  UIntTensor(d, true) {
+UIntTensor<T>::UIntTensor(const TensorDim &d, const void *buf,
+                          QScheme qscheme_) :
+  UIntTensor(d, true, Initializer::NONE, "", qscheme_) {
   if (d.getDataLen() != 0) {
     if (buf != nullptr)
       copy(buf);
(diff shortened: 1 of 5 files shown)


jijoongmoon (approved):
LGTM

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -344,6 +370,36 @@ template <typename T> void UIntTensor<T>::print(std::ostream &out) const {
...
+void UIntTensor<T>::read_quantization_info(std::ifstream &file) {
+  checkedRead(file, (char *)&qscheme, sizeof(uint8_t),
+              "[CharTensor::read] failed to read quantization information");
+}
+
+template <typename T> size_t UIntTensor<T>::scale_size() const {
+  switch (qscheme) {
+  case QScheme::PER_TENSOR_AFFINE:
+    return 1;
+    break;
```
dkjung, line 389: This line is not reachable because of the line before, which returns

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -344,6 +370,36 @@ template <typename T> void UIntTensor<T>::print(std::ostream &out) const {
...
+  case QScheme::PER_CHANNEL_AFFINE:
+    return width();
+    break;
```
dkjung, line 392: This line is not reachable because of the line before

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -141,6 +148,25 @@ template <typename T> void *UIntTensor<T>::getData(size_t idx) const {
...
+
+template <typename T> void *UIntTensor<T>::getScale(size_t idx) const {
+  NNTR_THROW_IF(idx > scale_size(), std::invalid_argument)
+    << "Tensor::getScale() index is not valid";
+
+  if (!data)
+    return nullptr;
+
+  data->validate();
+  return ((T *)getScale()) + idx;
```
dkjung, line 167: The overload of this method `getScale()` does almost the same thing, 
so if you call `getScale()` here, you will make things done twice.

Null checking for `data` and `data->validate()` will be done twice.

How about this?
```c++
return (T *)getData() + size() + idx;
```

dkjung (approved):
Minor comments

nntrainer/tensor/uint_tensor.cpp
djeong20: You're right! Thanks :)

nntrainer/tensor/uint_tensor.cpp
djeong20: That's a good point. Thank you for your suggestion!

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -141,6 +148,25 @@ template <typename T> void *UIntTensor<T>::getData(size_t idx) const {
...
+
+template <typename T> void *UIntTensor<T>::getScale(size_t idx) const {
+  NNTR_THROW_IF(idx > scale_size(), std::invalid_argument)
+    << "Tensor::getScale() index is not valid";
+
+  if (!data)
+    return nullptr;
+
+  data->validate();
+  return (T *)getData() + size() + idx;
```
EunjuYang, line 167: Is the type of scale `T`? AFIK, the type of scale is float; thus it should be updated as :
```suggestion
  return (float*)((T *)getData() + size()) + idx;
```

EunjuYang (commented):
What about adding a unittest case with multiple scales ? (for the purpose of validating its initialization).

baek2sm (approved):
LGTM

nntrainer/tensor/uint_tensor.cpp
djeong20: You're absolutely correct! Thank you for pointing that out.

nntrainer/tensor/uint_tensor.h
```diff
@@ -221,7 +235,32 @@ template <typename T> class UIntTensor : public TensorBase {
...
+   */
+  void read_quantization_info(std::ifstream &file) override;
+
+  /**
+   * @copydoc Tensor::scale_size()
+   */
+  size_t scale_size() const override;
+
+  /**
+   * @copydoc Tensor::scale_size()
```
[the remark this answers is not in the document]
djeong20: Thanks :)

djeong20:
> What about adding a unittest case with multiple scales ? (for the purpose of validating its initialization).

UIntTensor is a work in progress, and the current design has not been finalized. I will definitely add several test cases with the final output!

EunjuYang (approved):
LGTM!

PR #2935, opened 2025-02-11
[Bug] fix multithreading issue in lstm layer

baek2sm:
In LSTM, when the multi-thread option is enabled, either only the d_bias_h (hidden) value is calculated or both d_bias_ih (input-hidden) and d_bias_hh (hidden-hidden) values are calculated, depending on whether the integrate_bias option is set. However, there was an error because all three values were being attempted to be calculated at the same time. Therefore, this patch adds a conditional statement to fix this bug.

**Self evaluation:**


jijoongmoon (approved):
LGTM!

EunjuYang (approved):
Nice catch! 👍

PR #2936, opened 2025-02-11
[FC] Add dequantize action to FC layer

EunjuYang:
- This commit is temporary one to realize a fc layer with QINT weight.
- This should be fixed in the way to move the dequantize operations under the tensor op.

### DIFF
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -162,6 +162,19 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
       context.requestTensor(loraOut_dim, "hidden_lora", Initializer::NONE, true,
                             TensorLifespan::FORWARD_FUNC_LIFESPAN);
   }
+
+  ///@todo this quantizaer should be moved to tensor, not layer!
+  switch (context.getWeightDataType()) {
+  case ml::train::TensorDim::DataType::QINT4:
+  case ml::train::TensorDim::DataType::QINT8:
+  case ml::train::TensorDim::DataType::QINT16:
+    quantizer =
+      Quantization::createQuantizer(nntrainer::QScheme::PER_TENSOR_AFFINE);
+    break;
+  default:
+    quantizer = nullptr;
+    break;
+  }
 }
 
 void FullyConnectedLayer::exportTo(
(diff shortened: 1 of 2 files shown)


nntrainer/layers/fc_layer.cpp
```diff
@@ -162,6 +162,19 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
       context.requestTensor(loraOut_dim, "hidden_lora", Initializer::NONE, true,
                             TensorLifespan::FORWARD_FUNC_LIFESPAN);
   }
+
+  ///@todo this quantizaer should be moved to tensor, not layer!
+  switch (context.getWeightDataType()) {
+  case ml::train::TensorDim::DataType::QINT4:
```
baek2sm, line 168: how about adding 'not supported yet' message till implementation?

baek2sm (approved):
LGTM

nntrainer/layers/fc_layer.cpp
EunjuYang: It intends to make a quantizerfor all QINT4, QINT8, QINT16 cases. AFIK, QINT4, 8, 16 all work with a quantizer. Please let me know if I missed something!

EunjuYang:
I tested the case where QINT8-FP32 is used (weight dtype-tensor dtype).
Please refer to the result below:

- dequantization overhead (tested with https://github.com/EunjuYang/nntrainer/commit/cfe0f2f57e2d178ee66dbe86a4f81b7807489331):

```
$  ./nntrainer_simplefc 100 false 0 FP32 FP32
$  ./nntrainer_simplefc 100 false 0 QINT8 FP32
$  ./nntrainer_simplefc 100 true 2 FP32 FP32
$  ./nntrainer_simplefc 100 true 2 QINT8 FP32
```

Without SWAP |   | With SWAP   (LA=2) |  |
-- | -- | -- | -- |
FP32-FP32 | QINT8-FP32 | FP32-FP32 | QINT8-FP32
0.162787 | 0.166505 | 0.0487955 | 0.127012

bin file size is 400 MB vs. 100 MB.
Result was not tested (no suitable initialization method for scale is implemented yet)

nntrainer/layers/fc_layer.cpp
```diff
@@ -189,19 +203,9 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
...
-       {weight.getFormat(), dtype}},
-      true);
-
-    unsigned int axis =
-      context.getWeightObject(weight_idx[FCParams::weight]).getOutputAxis();
-
-    // weight.dequantize(weight_, axis);
+  ///@todo This dequantization action should be moved to tensor.dot()
+  if (quantizer != nullptr) {
+    Tensor weight_ = quantizer->dequantize(weight, input_.getDataType());
```
skykongkong8, line 208: I wonder how should it be expected to work at `Tensor::dot()`.
Maybe something like:
```C++
Tensor::dot() {
...
int8_t data = getData();
int8_t input_data = input;
Tensor qinput = quantizer->dequantize(...);
GEMM(data, qinput.getData(), output);
...
}
```
?

EunjuYang: I thought it would be like what you mentioned. Specifically, the dequantize/quantize operation would be applied only to the quantized tensors (e.g., char_tensor) by overriding the dot method in the quantized tensor itself, while comparing the type between the input and the tensor's type. We may need additional discussion about it with @djeong20 later.

jijoongmoon (approved):
LGTM

dkjung (approved):
LGTM

PR #2942, opened 2025-02-12
[ Doc ] add a note about the gcc/g++ version (enable_fp16)

EunjuYang:
- In order to build nntrainer with -Denable-fp16=true option on Ubuntu, it is required to use gcc/g++ >= 13.
- This commit adds the comment to the related `.md` files.

Self-evaluation:

Build test: [ ]Passed [ ]Failed [X]Skipped
Run test: [ ]Passed [ ]Failed [X]Skipped


baek2sm (approved):
Nice catch! LGTM

myungjoo (approved):

PR #2949, opened 2025-02-13
[FSU] Apply Static Memory allocation at FSU(Swap)

DonghakPark:
Replaced the FSU Logic from dynamically allocating memory --> overwriting and sharing statica allocated memory.

1. In order to prevent multiple frees, I added a map to check whether the mem_address has already been processed. Previously, memory was allocated through buf, but now it is being allocated directly.

2. Update FSU forwarding logic
--> FSU will handle look ahead tensor inside of pool
--> so we don't need to call Loadtensor for f + i

3. inference we don't need optimizer at application : remove optimizer at application

**Self evaluation:**

Co-authored-by: jijoong.moon <jijoong.moon@samsung.com>

### DIFF
--- a/Applications/SimpleFC/jni/main.cpp
+++ b/Applications/SimpleFC/jni/main.cpp
@@ -111,13 +111,7 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
     model->setProperty({withKey("memory_swap_lookahead", look_ahaed)});
   }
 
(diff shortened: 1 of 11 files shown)


baek2sm (approved):
nice work. LGTM!

DonghakPark:
Close This PR for reorganize and resubmit it so that can be reviewed conveniently.

PR #2951, opened 2025-02-14
[ONNX] add basic onnx interpreter @open sesame 02/17 20:00

baek2sm:
A simplest basic structure of onnx_interpreter has been added.
In this implementation, only input layer, weight layer, add layer are supported.

Basically, it is implemented by reading and parsing the onnx file using protobuf, and then creating an NNTrainer graph by sequentially checking the nodes. Users can simply specify the path to the ONNX model file through the "loadONNX" NNTrainer API function.

The types of tensors handled during the process of creating a graph can be classified into input tensor, constant tensor, weight tensor.
Among these, the current implementation only supports input tensor and weight tensor.
In ONNX, weight tensors have initializers unlike other tensors, so it can be distinguished as weight tensors through initializers. In the current implementation, initializer list are managed separately as a vector, but this vector will be removed as the implementation is completed in the future.

This commit uploads a minimal working implementation, and the onnx interpreter needs to be continuously updated.
An example application of reading an onnx file to create an NNTrainer model using this interpreter, and a document setting up the execution environment by installing protobuf will be uploaded as separate commits.

**Self evaluation:**

### DIFF
--- a/nntrainer/compiler/meson.build
+++ b/nntrainer/compiler/meson.build
@@ -16,6 +16,15 @@ compiler_headers = [
   'compiler_fwd.h'
 ]
 
+if get_option('enable-onnx-interpreter')
+  compiler_sources += [
+    'onnx_interpreter.cpp',
+  ]
+  compiler_headers += [
+    'onnx_interpreter.h',
+  ]
+endif
+
 if get_option('enable-tflite-interpreter')
   compiler_sources += [
     'tflite_interpreter.cpp',
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -90,7 +90,8 @@ for inference and training without any configurations*/
     ML_TRAIN_MODEL_FORMAT_INI_WITH_BIN, /**< ini file with save_path defined
                                            where the binary will be saved */
   MODEL_FORMAT_FLATBUFFER =
-    ML_TRAIN_MODEL_FORMAT_FLATBUFFER, /**< flatbuffer file */
+    ML_TRAIN_MODEL_FORMAT_FLATBUFFER,             /**< flatbuffer file */
+  MODEL_FORMAT_ONNX = ML_TRAIN_MODEL_FORMAT_ONNX, /**< ONNX file */
 };
 
 /**
--- a/api/nntrainer-api-common.h
+++ b/api/nntrainer-api-common.h
@@ -263,7 +263,8 @@ typedef enum {
     2, /**< Ini with bin format file saves configurations with parameters
          required for inference and training. */
   ML_TRAIN_MODEL_FORMAT_FLATBUFFER =
-    3 /**< Flatbuffer format file saves model configurations and weights. */
+    3, /**< Flatbuffer format file saves model configurations and weights. */
+  ML_TRAIN_MODEL_FORMAT_ONNX = 4, /**< ONNX format file. */
 } ml_train_model_format_e;
 
 /**
--- a/meson.build
+++ b/meson.build
@@ -516,6 +516,10 @@ if get_option('enable-tflite-interpreter')
   extra_defines += '-DENABLE_TFLITE_INTERPRETER=1'
 endif
 
+if get_option('enable-onnx-interpreter')
+  extra_defines += '-DENABLE_ONNX_INTERPRETER=1'
+endif
+
 opencv_dep = dummy_dep
 
 if get_option('platform') != 'android'
--- a/nntrainer/compiler/onnx_interpreter.cpp
+++ b/nntrainer/compiler/onnx_interpreter.cpp
@@ -1,0 +1,152 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2025 SeungBaek Hong <sb92.hong@samsung.com>
+ *
+ * @file   onnx_interpreter.cpp
+ * @date   12 February 2025
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author SeungBaek Hong <sb92.hong@samsung.com>
+ * @bug	   No known bugs except for NYI items
+ * @brief  This is onnx converter interface for c++ API
+ */
+
+#include "layer_node.h"
+#include <onnx_interpreter.h>
+
+namespace nntrainer {
+
+void ONNXInterpreter::serialize(const GraphRepresentation &representation,
+                                const std::string &out){};
+
+GraphRepresentation ONNXInterpreter::deserialize(const std::string &in) {
+  // Load and parse onnx file with protobuf
+  std::ifstream file(in, std::ios::binary);
+  if (!file.is_open()) {
+    throw std::runtime_error("File does not exist: " + in);
+  }
+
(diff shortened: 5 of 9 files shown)


skykongkong8 (approved):

nntrainer/schema/onnx_interpreter.h
```diff
@@ -0,0 +1,249 @@
...
+   * @param path path of onnx model file.
+   */
+  std::unique_ptr<ml::train::Model> load(std::string path) {
+    // Load and parse onnx file with protobuf
+    std::ifstream file(path, std::ios::binary);
+    onnx_model.ParseFromIstream(&file);
+
+    // Create nntrainer model instance
+    nntrainer_model = ml::train::createModel();
+    std::vector<std::shared_ptr<ml::train::Layer>> layers;
```
EunjuYang, line 96: Is it okay to create a *model* in direct? In my opinion, we may need to provide common interface for model importing from other framework. In this context, `IniGraphInterpreter`  creates `GraphRepresentation` as its return value.

What about making this `ONNXInterpreter` work with `ModelLoader` like `IniGraphInterpreter`? (in terms of making consensus with other loader).
I think this is related to which interface we provide to user for importing model with onnx .

FYI)
https://github.com/nnstreamer/nntrainer/blob/859348b09776e2eb027a373a5910f15becc470d3/nntrainer/models/model_loader.cpp#L425-L428

nntrainer/schema/onnx_interpreter.h
```diff
@@ -0,0 +1,249 @@
...
+      }
+
+      if (node.op_type() == "Add") {
+        layerOutputMap.insert(
+          {cleanName(node.output()[0]), cleanName(node.name())});
+
+        layers.push_back(ml::train::createLayer(
+          "add",
+          {"name=" + cleanName(node.name()),
+           withKey("input_layers", inputNames[0] + "," + inputNames[1])}));
```
EunjuYang, line 157: This is just simple idea suggestion. We may need  to update this to remove if / else clause.
We may add `node.op_type()` as a duplicate key for its corresponding layer_name; Anyhow, we may need common interface to handle various type of layers.

EunjuYang (commented):
I'm thrilled to review this PR. This PR is a great starting point to support the ONNX format. The ONNXInterpreter returns a neuralnet model as its interpretation outcome, allowing the user to create a neuralnet directly.

In my opinion, however, instead of letting the ONNXInterpreter have control over model creation, what about giving the right to modelLoader instead of directly allowing the interpreter to create a model?

I feel this code allows the user to create a model:

```
auto onnx_interpreter = ONNXInterpreter();
auto model = onnx_interpreter(path);
model.compile();
model.initialize();
```

Instead:

```
auto &model = createModel();
model.load(file_path, ml::train::ModelFormat::ONNX);
```

I may have missed some points on how to use this ONNXInterpreter. Please let me know your idea on this :)

SeoHyungjun (approved):
LGTM

baek2sm:
> I'm thrilled to review this PR. This PR is a great starting point to support the ONNX format. The ONNXInterpreter returns a neuralnet model as its interpretation outcome, allowing the user to create a neuralnet directly.
> 
> In my opinion, however, instead of letting the ONNXInterpreter have control over model creation, what about giving the right to modelLoader instead of directly allowing the interpreter to create a model?
> 
> I feel this code allows the user to create a model:
> 
> ```
> auto onnx_interpreter = ONNXInterpreter();
> auto model = onnx_interpreter(path);
> model.compile();
> model.initialize();
> ```
> 
> Instead:
> 
> ```
> auto &model = createModel();
> model.load(file_path, ml::train::ModelFormat::ONNX);
> ```
> 
> I may have missed some points on how to use this ONNXInterpreter. Please let me know your idea on this :)

That's good point. but, user don't use the ONNXInterpreter class directly.
Instead, user uses "onnx.h" api to create & load ONNX model.
example code: 
```
#include <onnx.h>
auto &model = ml::train::loadONNX(file_path);
```

EunjuYang:
> > I'm thrilled to review this PR. This PR is a great starting point to support the ONNX format. The ONNXInterpreter returns a neuralnet model as its interpretation outcome, allowing the user to create a neuralnet directly.
> > In my opinion, however, instead of letting the ONNXInterpreter have control over model creation, what about giving the right to modelLoader instead of directly allowing the interpreter to create a model?
> > I feel this code allows the user to create a model:
> > ```
> > auto onnx_interpreter = ONNXInterpreter();
> > auto model = onnx_interpreter(path);
> > model.compile();
> > model.initialize();
> > ```

> >   
> > Instead:
> > ```
> > auto &model = createModel();
> > model.load(file_path, ml::train::ModelFormat::ONNX);
> > ```

> > I may have missed some points on how to use this ONNXInterpreter. Please let me know your idea on this :)
> 
> @EunjuYang That's good point. but, user don't use the ONNXInterpreter class directly. Instead, user uses "onnx.h" api to create & load ONNX model. example code:
> 
> ```
> #include <onnx.h>
> auto &model = ml::train::loadONNX(file_path);
> ```

I see! I missed #2958; the sample app seems good and it looks easy to use. One thing I have been worrying about is the different interface between other importers. If it doesn't matter, it would be okay to go.

nntrainer/schema/onnx_interpreter.h
baek2sm: @EunjuYang Unlike the existing API, I tried to improve user convenience by handling model creation internally.
However, as you mentioned, I think it can be integrated with ModelLoader.
I will upload this part as a refactoring commit again. Thanks a lot!!

dkjung:
@baek2sm Could you rebase?

dkjung:
> An example application of reading an onnx file to create an NNTrainer model using this interpreter, and a document setting up the execution environment by installing protobuf will be uploaded as separate commits.

Is the doc ready?

myungjoo (approved):

jijoongmoon:
First of all, thanks for the good feature. But, I don't think it is the right way to support onnx file. Currently we do have feature to read ini format and binary. I guess it also should work in a simillar way to load. By doing that, it does not need to add another load api in ccapis. It is also relate with capi for tizen and it could be more convinient adding options than adding another api.
please take a look https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/models/model_loader.cpp and https://github.com/nnstreamer/nntrainer/blob/main/nntrainer/models/model_loader.h

baek2sm:
@jijoongmoon okay then i'll try to integrate it with the previous api

baek2sm:
@EunjuYang @jijoongmoon  I've integrated with the previous interface using model loader. thanks!

nntrainer/compiler/meson.build
```diff
@@ -16,6 +16,15 @@ compiler_headers = [
   'compiler_fwd.h'
 ]
 
+if get_option('enable-onnx-interpreter')
+  compiler_sources += [
```
jijoongmoon, line 20: I think this should be enabled as a default.

jijoongmoon (approved):
LGTM!

EunjuYang:
Awesome patch! Thanks!

PR #2959, opened 2025-02-20
[Bug Fix] Fix static initialization order fiasco

gkisalapl:
Execution of static object constructors are undefined across multiple compilation units which result in random failing of unit tests (especially visible on windows). This PR replace using of:
inline static const std::string
with
static constexpr const char *
which fix std::string constructor order randomness

see https://en.cppreference.com/w/cpp/language/siof

**Self-evaluation:**

### DIFF
--- a/Applications/Custom/mae_loss.h
+++ b/Applications/Custom/mae_loss.h
@@ -88,7 +88,7 @@ public:
    */
   bool requireLabel() const { return true; }
 
-  inline static const std::string type = "mae_loss";
+  static constexpr const char *type = "mae_loss";
 };
 
 } // namespace custom
--- a/Applications/Custom/momentum.h
+++ b/Applications/Custom/momentum.h
@@ -78,7 +78,7 @@ public:
    */
   void setProperty(const std::vector<std::string> &values) override;
 
-  inline static const std::string type = "momentum";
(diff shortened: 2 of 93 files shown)


djeong20 (approved):
Great observation! I understand that the issue arises from the differing order of initialization for static objects, which can lead to problems if one depends on another. However, I’m not clear on why std::string should be replaced with a char array. Could you clarify this point for me?

dkjung (approved):
LGTM

skykongkong8 (approved):
LGTM!

jijoongmoon (approved):
LGTM.

jijoongmoon:
> > Great observation! I understand that the issue arises from the differing order of initialization for static objects, which can lead to problems if one depends on another. However, I’m not clear on why std::string should be replaced with a char array. Could you clarify this point for me?
> 
> I found that issue while running unittest_compiler test on windows. Problem happened in add_default_object method from app_context.cpp. We were performing registering factory objects but std::string object type for one of object was not initialized when function add_default_object from context was called which caused test failure.
> 
> When we replace std::string (which is object with constructor) with constexpr const char * (which are initialized at compile time) we're ensuring that all type values are initialized before running global constructors

and I think it copy only the pointer I believe.

PR #2962, opened 2025-02-21
Meson option build fixes

piotrrak:
Misc fixes for some of the options disabled by configuration options.
This is step in direction of fixing builds with various configuration options configurations.

**Self evaluation:**

**How to evaluate:**

meson setup <builddir> --native-file=disables.ini

`disables.ini` with contents:
```
[project options]

enable-blas=false 
enable-tflite-backbone=false 
enable-memory-swap=false 
enable-nnstreamer-backbone=false 
enable-tflite-interpreter=false 
enable-app=false 
ml-api-support='disabled'
enable-test=true
werror=false
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -583,7 +583,7 @@ if get_option('enable-app')
   endif
 endif
 
-if get_option('platform') != 'android' and host_machine.system() != 'windows'
+if get_option('platform') != 'android' and host_machine.system() != 'windows' and get_option('enable-nnstreamer-backbone')
   nnstreamer_dep = dependency('nnstreamer')
   message('building nnstreamer')
   subdir('nnstreamer')
--- a/test/unittest/compiler/meson.build
+++ b/test/unittest/compiler/meson.build
@@ -24,6 +24,7 @@ exe = executable(
   install_dir: application_install_dir
 )
 
+
 test(test_name, exe,
   args: '--gtest_output=xml:@0@/@1@.xml'.format(meson.build_root(), test_name),
   timeout: test_timeout
--- a/test/unittest/meson.build
+++ b/test/unittest/meson.build
@@ -104,4 +104,6 @@ subdir('compiler')
 subdir('layers')
 subdir('datasets')
 subdir('models')
-subdir('integration_tests')
+if get_option('enable-tflite-interpreter')
+  subdir('integration_tests')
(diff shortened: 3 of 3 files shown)


meson.build
```diff
@@ -504,7 +504,7 @@ if get_option('enable-app')
   endif
 endif
 
-if get_option('platform') != 'android' and host_machine.system() != 'windows'
+if get_option('platform') != 'android' and host_machine.system() != 'windows' and get_option('enable-nnstreamer-backbone')
```
baek2sm, line 507: This script contains the code `if get_option('enable-nnstreamer-backbone') and get_option('platform') != 'android'
  extra_defines += '-DENABLE_NNSTREAMER_BACKBONE=1'
endif`.

How about merging this part as well? It seems like it would be fine to move the part `extra_defines += '-DENABLE_NNSTREAMER_BACKBONE=1'` under this condition, too!

djeong20: FYI) line 407-409

djeong20 (approved):
LGTM!

meson.build
djeong20: FYI) line 407-409

piotrrak: @baek2sm @djeong20 thanks for suggestion, change updated/rebased.
I've dediced to move `host_machine.system() != 'windows'` there too, that probably won't be necessary in near future.

skykongkong8:
According to discussion made from #2930, I believe `enable-blas=true` should set by default for both x86 and arm situation. Please share your opinion if you find this fix inappropriate!

skykongkong8:
> Is "should" = "we never really shouldn't build with BLAS disabled on x86 or arm"?

Yes, I believe that would be more appropriate for the current status since we rely all single-precision computations on cblas.  And yes, I think we should really consider removing `enable-blas` option, but it actually needs more discussion because we might have to introduce multiple backend options per target architecture.

myungjoo (approved):

jijoongmoon (approved):
LGTM

meson.build
piotrrak: I've reverted back to not moving `subdir('nnstreamer')` to the earlier stage of meson.

skykongkong8 (approved):
LGTM

PR #2968, opened 2025-02-25
Add unit-tests for dotCL function, N=1, M=1 case

dkjung:
---
# Add two unit-tests that test `dotCL`

As there is no test for `dotCl` function in the case with `N=1, M=1`,
I added two unit-tests for that condition.

**Self evaluation:**

**How to evaluate:**
1. Run `./tools/android_test.sh` with an Android device attached to your machine
2. ```bash
    adb shell /data/local/tmp/nntr_android_test/unittest_blas_kernels_cl --gtest_filter=blas_kernels.dotCL_sgemv_N_1_M_1_*
    ```

### DIFF
--- a/test/unittest/unittest_blas_kernels_cl.cpp
+++ b/test/unittest/unittest_blas_kernels_cl.cpp
@@ -459,6 +459,97 @@ TEST(blas_kernels, dotCL_sgemv_n_fp16) {
   EXPECT_THROW(dotCl(A_fp16, B_fp16, transA, transB), std::runtime_error);
 }
 
+TEST(blas_kernels, dotCL_sgemv_N_1_M_1_1) {
+  setUpGpuContext();
+  int batch = 1;
+  int channel = 1;
+  int height = 1;
+  int width = 768;
+
+  int height_b = 1;
+  int width_b = 768;
+
+  bool transA = false;
+  bool transB = true;
+
+  const float alpha = 1e-1;
+  const int MOD = 10;
+
+  nntrainer::TensorDim::TensorType t_type_nchw_fp32 = {
+    nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP32};
+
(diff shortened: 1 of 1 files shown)


skykongkong8 (approved):
LGTM!

test/unittest/unittest_blas_kernels_cl.cpp
```diff
@@ -459,6 +459,97 @@ TEST(blas_kernels, dotCL_sgemv_n_fp16) {
   EXPECT_THROW(dotCl(A_fp16, B_fp16, transA, transB), std::runtime_error);
 }
 
+TEST(blas_kernels, dotCL_sgemv_N_1_M_1_1) {
+  setUpGpuContext();
```
djeong20, line 463: Do you know if we need to call this for every test? I see most of the test cases do not, but only two test cases call it. (`dotCL_sgemv_M_1_1`,  and `dotCL_sgemv_M_1_1_fp16`)

djeong20 (commented):
Will there also be half-precision unit tests for N=1, M=1 case?

baek2sm (approved):
LGTM

SeoHyungjun (approved):
LGTM

test/unittest/unittest_blas_kernels_cl.cpp
EunjuYang: `setUpGpuContext()` is expected to be called at least once. If it is called once, it will create a static context for CL and register CL-related kernels. I guess that's why the previous test code works even with the irregular call to setUpGpuContext().

myungjoo (approved):
LGTM.

But be aware, before the release, the SE will require you to have more negative test cases than positive test cases.

And you are adding positive test cases.

Thus, you are going to be very busy for the next release. (adding a lot of negative test cases in a few days just before the release/commercialization)

jijoongmoon (approved):
LGTM

PR #2977, opened 2025-03-05
[Wait for #2865] [Subgraph] rename `GraphRepresentation` to `GraphLayerNodeRepresentation`

EunjuYang:
- This PR is stacked upon #2865. 
- If you reviewed the previous one, the commit required to be review is : [5df2ca0](https://github.com/EunjuYang/nntrainer/commit/5df2ca03aaa6a03a463a36a723fe40244f8da04f)
- This commit renames the alias for `GraphRepresentation` to
  `GraphLayerNodeRepresentation`
- The reason we do this is to update the GraphRepresentation as the
  serialized SubGraphBase pointer, not LayerNode

**Self evaluation:**

Build test: [*]Passed [ ]Failed [ ]Skipped
Run test: [*]Passed [ ]Failed [ ]Skipped

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

PR #2979, opened 2025-03-05
[Wait for #2977] [Subgraph] Refactor: update GraphName (rename and update its property)

EunjuYang:
- This PR is stacked upon #2977.
- If you reviewed the previous one, the commit required to be reviewed is : [7ec9b2e](https://github.com/EunjuYang/nntrainer/commit/7ec9b2e9871216457ef6b8360aa40b003ce1ccee)
- Renamed `GraphName` to `SubGraphName`.
- Updated keyword for `SubGraphName` attribute to `subgraph_name`
- Made corresponding changes.

**Self evaluation:**


nntrainer/graph/network_graph.cpp
```diff
@@ -6,1115 +6,173 @@
...
-  }
-}
-
-void NetworkGraph::inPlaceOptimize() {
-  if (optimize_memory) {
-    for (unsigned int idx = 0; idx < graph.size(); ++idx) {
-      auto const &lnode = getSortedLayerNode(idx);
-      lnode->setInPlaceType(canExecuteInPlace(lnode));
-    }
+  const std::string &subgraph_name = layer->getSubGraphName();
```
SeoHyungjun, line 129: It feels strange that the graph name is stored in the layer.
However, changing the structure to graph -> subgraph -> layer in nntrainer would require too many changes.

SeoHyungjun (approved):

nntrainer/graph/network_graph.cpp
EunjuYang: That's a good point. It is related to how a user designate a subgraph for a specific layer.
For this PR, it assumed that a user call `model->addLayer()` with `subgraph_name` property, which is the only interface that user can handle.
It would be modified later when `createSubGraph` API is added and layer is added to a subgraph.

PR #2984, opened 2025-03-05
Context, Context Data and Context Allocator

jijoongmoon:
### In this PR
This PR enables the Engine, Context, Context Data and Context allocator.
#### Engine Class
In this PR, Engine Class which is managing the Context to create Layers for specicifc Accelerators.
AppContext create the CPU Layers, CLContext creates the GPU Layers. It also provides the Pluggable features, so that,
the developer can create it's own context to support various accelerators such as NPUs.
Thus, we have to use the ```getRegisterContext("cpu")``` call to get the proper Context. It also supports registerFactory to add custom layers as before.

#### Context Class
In this PR, Context Class is introduced. As mentioned in Engine class, it is possible to add custom context and this Context class might be the base class for all custom context. 
Context Class also has ```ContextData and MemAllocator```. ContextData can be used to create and used to utilize the global data for specific context after initialze the context and each specific accelerator need to spcific memory allocator. 

#### read and save the weight 
Previously layer_node handle the read and save the weight but, the read and save function can be differ depending on layer types such as batch normalization layer. So, In this pr, the read fuction is lowered to layer object itself. If there is no overwited read funciotn, it will use base class read, which is layer_devel.

#### RunContext with ContextData
In order to use the ContextData, additional parameter is set in RunLayerContext. In this way, it is possible to access in each layer. Also add compute_engine type inInitLayerContext as well.

#### Tensor and Weight Layer
This PR also introduce Tensor and Weight Layer which contain Tensors and Weights only. These layers act like container to provide the data.

#### FSU with shared Tensor
This PR also provides initial version of FSU with shared Weight Tensors. Most of the HW accelerators prefer the statically allocated memory rather than dynamically allocated memory. So modify function call of memory pool allocate and getMmeory but need to be tested more. For better latency, we also have to think the MAP_FIXED option in mmap and this also should be done later. For accurate controlablity of the Execution order, there is some fixes in manager as well.


nntrainer/utils/base_properties.h
```diff
@@ -19,9 +19,11 @@
 #include <string>
 #include <vector>
 
+#include <common.h>
 #include <nntrainer_error.h>
 #include <tensor_dim.h>
 #include <util_func.h>
+#include <iostream>
```
EunjuYang, line 26:
```suggestion
```

test/nntrainer_test_util.cpp
```diff
@@ -237,14 +237,19 @@ getResPath(const std::string &filename,
 
 nntrainer::GraphRepresentation
 makeGraph(const std::vector<LayerRepresentation> &layer_reps) {
-  static auto &ac = nntrainer::AppContext::Global();
+  static auto &eg = nntrainer::Engine::Global();
+  // #ifdef ENABLE_QNN
+  //   static auto &qc = nntrainer::QNNContext::Global();
+  // #endif
```
EunjuYang, line 243:
```suggestion
```

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  nntrainer::AppContext *app_context = new nntrainer::AppContext();
+  app_context->Global();
+
+  eg.registerContext("cpu", app_context);
+
+  // #ifdef ENALBE_OPENCL
+  //   eg.registererContext("gpu",
+  //                        nntrainer::ClContext(nntrainer::ClContext::Global()));
+  // #endif
```
EunjuYang, line 52: Can't this block be enabled?

meson.build
```diff
@@ -100,6 +100,7 @@ if get_option('enable-fp16')
      extra_defines += '-DENABLE_FP16=1'
      extra_defines += '-DUSE__FP16=1'
      extra_defines += '-DUSE_NEON=1'
+     extra_defines += '-DENABLE_QNN=1'
```
EunjuYang, line 103: It seems the option should be removed for this `meson.build`

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+std::string
+Engine::parseComputeEngine(const std::vector<std::string> &props) const {
+  for (auto &prop : props) {
+    std::string key, value;
+    int status = nntrainer::getKeyValue(prop, key, value);
+    if (nntrainer::istrequal(key, "engine")) {
+      constexpr const auto data =
+        std::data(props::ComputeEngineTypeInfo::EnumList);
+      for (uint i = 0; i < props::ComputeEngineTypeInfo::EnumList.size(); ++i) {
```
EunjuYang, line 82:
```suggestion
      for (unsigned int i = 0; i < props::ComputeEngineTypeInfo::EnumList.size(); ++i) {
```

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+  working_path_base = std::string(ret);
+  ml_logd("working path base has set: %s", working_path_base.c_str());
+  free(ret);
+}
+
+int Engine::registerContext(const std::string &library_path,
+                            const std::string &base_path) {
+  const std::string full_path = getFullPath(library_path, base_path);
+
+  void *handle = dlopen(full_path.c_str(), RTLD_LAZY | RTLD_LOCAL);
```
EunjuYang, line 153: `dlopen` won't work on Windows. Please replace `dlopen` with nntrainer's util function.
```suggestion
  void *handle = DynamicLibraryLoader::loadLibrary(full_path.c_str(), RTLD_LAZY | RTLD_LOCAL);
```

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+  const char *error_msg = dlerror();
```
EunjuYang, line 154: Ditto.
```suggestion
  const char *error_msg = DynamicLibraryLoader::getLastError();
```

nntrainer/layers/layer_context.h
```diff
@@ -265,15 +274,15 @@ class InitLayerContext {
...
+  unsigned int requestTensor(
+    const TensorDim &dim, const std::string &name,
+    const Initializer init = Initializer::NONE, bool trainable = false,
+    TensorLifespan lifespan = TensorLifespan::ITERATION_LIFESPAN,
+    bool private_ = true,
+    ml::train::LayerComputeEngine engine = ml::train::LayerComputeEngine::CPU) {
     const auto &prefix_ = private_ ? this->name : this->prefix;
     tensors_spec.emplace_back(dim, init, trainable, prefix_ + ":" + name,
-                              lifespan);
+                              lifespan, engine);
```
EunjuYang, line 285: Simple question. How is this engine info used for tensor management?

EunjuYang (commented):
It seems the PR contains some commits already merged to upstream. Please rebase the PR.

nntrainer/context.h
```diff
@@ -36,6 +36,19 @@
...
 namespace nntrainer {
 
+/**
+ * @class ContextData contains Data which generated by Context to use in Layers
+ * created by Context
+ * @brief Container to hold context data generated by Context. Eventaully
+ * RunLayerContext in layer_node will hold the ContextData, so that, Layer can
+ * access this Context Data.
+ */
+class ContextData {
```
piotrrak, line 47: We might want publicly inherit from `std::enable_shared_from_this<ContextData>`, since it will be used with `std::shared_pointer`. Upside would be that shared/weak atomic reference counters (shared_ptr control block) will become part of this object rather than allocated (most likely dynamic allocation).

nntrainer/layers/layer_node.cpp
```diff
@@ -129,6 +129,55 @@ class SharedFrom : public Name {
...
+
+/**
+ * @brief get the compute engine property from property string vector
+ *  : default is CPU
+ * @return LayerComputeEngine Enum : CPU, GPU, QNN
+ *
+ */
+ml::train::LayerComputeEngine
+getComputeEngine(const std::vector<std::string> &props) {
+  for (auto &prop : props) {
```
piotrrak: Is this common way to have to to enumerate all properties and filter on its value?

Such code hints that possible future improvement would be:
- to executed `nntrainer::getKeyValue`  once for all properties creating mapping
Example: std::map/std::unordered_map<property_key, property_value>/our implementation of future std::flat_map
- switching all such functions to accept such mapping rather doing O(n) filtering.

In the code execution it costs O(N*M) right now where N is number of specified properties and M is number of possible properties.

Other possilble benefits:
- It is harder to read and write such filtering probably;
- Would give an opportunity to 'case-normalize' keys of such collection

nntrainer/layers/common_properties.h
```diff
@@ -1586,6 +1586,65 @@ class PropsUserData final : public Property<void *> {
...
+                                            "epoch",
+                                            "forward_infer",
+                                            "max"};
+};
+
+/**
+ * @brief TensorLifeSpacn, return TensorLifespan
+ *
+ */
+class TensorLife : public EnumProperty<TensorLifeInfo> {
```
piotrrak, line 1634: Makes me wonder if TensorLife property is something that is temporary?

Seems like (may be wrong) something we could/should infer from network graph during planning rather than to be supplied by user.

Also bit worried if it won't be to much of a knowledge for user about what is best memory managment strategy.

nntrainer/layers/tensor_layer.cpp
```diff
@@ -0,0 +1,105 @@
...
+    << "Size of Dimensions, Types, Formats should be matched!";
+
+  tensor_idx.reserve(t_dims.size());
+
+  for (unsigned int i = 0; i < t_dims.size(); ++i) {
+    t_dims[i].setFormat(context.getFormat());
+    t_dims[i].setDataType(t_dtype[i]);
+    std::string name = context.getName()+"_t" + std::to_string(i);
+    if (!t_name.empty())
+      name = t_name[i];
```
piotrrak, line 76: `t_name.size()` is never checked if it is equal to `t_dims.size()` before being accessed here.

nntrainer/layers/tensor_layer.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  auto &t_dtype = std::get<std::vector<props::TensorDataType>>(tensor_props);
+  auto &t_name = std::get<std::vector<props::TensorName>>(tensor_props);
+  auto &t_life = std::get<std::vector<props::TensorLife>>(tensor_props);
+
+  NNTR_THROW_IF(!t_dims.size(), std::invalid_argument)
+    << "Tensor dimension is not provided";
+  n_tensor = t_dims.size();
+
+  if (!t_dtype.size()) {
```
piotrrak, line 49: Maybe `if (t_dtype.empty())`?

nntrainer/layers/tensor_layer.cpp
```diff
@@ -0,0 +1,105 @@
...
+    t_life.reserve(t_dims.size());
+    for (auto t : t_dims)
+      t_life.push_back(nntrainer::TensorLifespan::MAX_LIFESPAN);
+  }
+
+  auto engine = context.getComputeEngineType();
+
+  NNTR_THROW_IF(t_dims.size() != t_dtype.size(),
+                std::invalid_argument)
+    << "Size of Dimensions, Types, Formats should be matched!";
```
piotrrak, line 67: It seems, `InitLayerContext provides singular format provided by `InitLayerContext::getFormat()` for shapes (t_dims[i]) line 72;

Should the Format be mentioned here then?

nntrainer/layers/tensor_layer.cpp
```diff
@@ -0,0 +1,105 @@
...
+  n_tensor = t_dims.size();
+
+  if (!t_dtype.size()) {
+    ml_logi("Set Activation Tensor DataType");
+    t_dtype.reserve(t_dims.size());
+    for (auto t : t_dims)
+      t_dtype.push_back(context.getActivationDataType());
+  }
+
+  if (!t_life.size()) {
```
piotrrak, line 56: Again maybe `if (t_life.empty())`?

nntrainer/layers/tensor_layer.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  for (unsigned int i = 0; i < t_dims.size(); ++i) {
+    t_dims[i].setFormat(context.getFormat());
+    t_dims[i].setDataType(t_dtype[i]);
+    std::string name = context.getName()+"_t" + std::to_string(i);
+    if (!t_name.empty())
+      name = t_name[i];
+
+    tensor_idx.push_back(context.requestTensor(t_dims[i], name, Initializer::NONE,
+                                          true, t_life[i], true, engine));
```
piotrrak, line 79: `t_life.size()` was not verified to be equal to max possible value of `i` ie. `t_dims.size()`
We know it will be fine only if it was empty and we filled it ourselves, other codepath is not verified to have at least/or better exactly that many t_life values provided by user.

nntrainer/layers/weight_layer.cpp
```diff
@@ -40,20 +40,41 @@ void WeightLayer::finalize(InitLayerContext &context) {
...
 
-  const auto &weight_dim = std::get<props::TensorDimension>(weight_props).get();
+  auto &t_dtype = std::get<std::vector<props::TensorDataType>>(weight_props);
+  auto &t_name = std::get<std::vector<props::WeightName>>(weight_props);
+
+  NNTR_THROW_IF(!t_dims.size(), std::invalid_argument)
+    << "Weight dimension is not provided.";
+  n_weight = t_dims.size();
+
+  if (!t_dtype.size()) {
```
piotrrak, line 50: `empty()`?

nntrainer/layers/weight_layer.cpp
```diff
@@ -40,20 +40,41 @@ void WeightLayer::finalize(InitLayerContext &context) {
...
-  output_dims[0] = weight_dim;
+  for (unsigned int i = 0; i < t_dims.size(); ++i) {
+    t_dims[i].setFormat(context.getFormat());
+    t_dims[i].setDataType(t_dtype[i]);
+    std::string name = context.getName() + "_w" + std::to_string(i);
 
-  output_dims[0].setTensorType(
-    {context.getFormat(), context.getActivationDataType()});
+    if (!t_name.empty())
+      name = t_name[i];
```
piotrrak, line 70: `t_name.size()` was not verified to be `t_dims.size()`

nntrainer/layers/weight_layer.h
```diff
@@ -95,8 +112,12 @@ class WeightLayer : public LayerImpl {
   static constexpr const char *type = "weight";
 
 private:
-  std::tuple<props::TensorDimension> weight_props;
-  std::array<unsigned int, 1> weight_idx; /**< indices of the weights */
+  std::tuple<std::vector<props::TensorDimension>,
+             std::vector<props::TensorDataType>, std::vector<props::WeightName>>
+    weight_props;
```
piotrrak, line 109: Such helper template alias comes to mind:

```c++
template <typename... Types>
using VectorTuple = std::tuple<std::vector<Types>...>;
```
then this code would become:
```c++
VectorTuple<props::TensorDimension,
                        props::TensorDataType,
                        props::WeightName> weigt_props;
```

nntrainer/layers/weight_layer.h
```diff
@@ -95,8 +112,12 @@ class WeightLayer : public LayerImpl {
...
+
+  std::vector<unsigned int> weight_idx; /**< indices of the weights */
+  unsigned int n_weight;
```
piotrrak, line 112: Wouldn't invariant assert(n_weights == weight_idx.size()) be always true?
Could `n_weights` field be removed in favour of `weight_idx.size()`?

piotrrak (commented):
Interesting work, might be relevant also for how `openvino` could be supported. Will spend more time to review it, but some minor comments and inquries.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+ * 5) if @a base != "" && @a path != "", return @a base + "/" + path
+ *
+ * @param path path to calculate from base
+ * @param base base path
+ * @return const std::string
+ */
+const std::string getFullPath(const std::string &path,
+                              const std::string &base) {
+  /// if path is absolute, return path
+  if (path[0] == '/') {
```
piotrrak, line 111: Won't work on windows.
Best portable way is to use `std::filesystem::path` + `std::filesystem::is_absolute`

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+    return path;
+  }
+
+  if (base == std::string()) {
```
piotrrak, line 115: `base.empty()`?

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * @param path path to calculate from base
+ * @param base base path
+ * @return const std::string
+ */
+const std::string getFullPath(const std::string &path,
+                              const std::string &base) {
+  /// if path is absolute, return path
+  if (path[0] == '/') {
+    return path;
```
piotrrak, line 112: Shouldn't we throw if base wasn't an `empty()` string?
The rules how this function works are pretty complex, and non-empty base when passing absolute path is probably user error.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+const std::string Engine::getWorkingPath(const std::string &path) {
+  return getFullPath(path, working_path_base);
+}
+
+void Engine::setWorkingDirectory(const std::string &base) {
+  DIR *dir = opendir(base.c_str());
+
+  if (!dir) {
```
piotrrak, line 129: `if (!dir && errno == EPERM)`

Other issue might be that file descriptor limit per user/process/cgroup was reached.
Then `errno` would be `EMFILE` or `ENFILE`.
User getting such error for that case would be really puzzled/misguided, I imagine and would take probably `strace/gdb` to solve what is really going on.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  return path == std::string() ? base : base + "/" + path;
+}
+
+const std::string Engine::getWorkingPath(const std::string &path) {
+  return getFullPath(path, working_path_base);
+}
+
+void Engine::setWorkingDirectory(const std::string &base) {
+  DIR *dir = opendir(base.c_str());
```
piotrrak, line 127: we propably should assert !base.empty()

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  if (!dir) {
+    std::stringstream ss;
+    ss << func_tag << "path is not directory or has no permission: " << base;
+    throw std::invalid_argument(ss.str().c_str());
```
piotrrak, line 132: This is bug; Signature is `std::stringstream::str() -> std::string` - ie. returns temporary;
Lifetime of this temporary ends before `std::string::c_str() -> const char*` is passed to `std::invalid_argument` resulting passing pointer to memory that was already `delete[]` ed by `std::string::~string()` or if we are able to benefit from SVO (ie. string was small enough in good quality standard library) part (subobject) of std::string object that was destroyed.

`std::invalid_argument` has constructor `std::invalid_string(const std::string&)`
By rules of standard `const &` argument will extend lifetime of std::string temporary.
```
throw std::invalid_argument(ss.str());
```
Is correct and won't exhibit using use after free bug.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+  closedir(dir);
+
+  char *ret = getRealpath(base.c_str(), nullptr);
+
+  if (ret == nullptr) {
+    std::stringstream ss;
+    ss << func_tag << "failed to get canonical path for the path: ";
+    throw std::invalid_argument(ss.str().c_str());
```
piotrrak, line 141: BUG: please see my comment above, for explaination.

nntrainer/engine.h
```diff
@@ -0,0 +1,255 @@
...
+  static nntrainer::Context *getRegisteredContext(std::string name) {
+
+    std::transform(name.begin(), name.end(), name.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+
+    if (engines.find(name) == engines.end()) {
+      throw std::invalid_argument("not registered");
+    }
+    return engines.at(name);
+  }
```
piotrrak, line 108: Are we certain that we shouldn't take the mutex here?

It seems to me bit strange that mutator is protected by lock while `public static` observer of `engines` collection is not.

nntrainer/engine.h
```diff
@@ -0,0 +1,255 @@
...
+   * @param int_key key
+   * @param props property
+   * @return unitque_ptr<T> unique pointer to the LearningRateScheduler object
+   */
+  std::unique_ptr<ml::train::LearningRateScheduler>
+  createLearningRateSchedulerObject(
+    const int int_key, const std::vector<std::string> &properties = {}) {
+    auto ct = getRegisteredContext(parseComputeEngine(properties));
+    return ct->createLearningRateSchedulerObject(int_key, properties);
+  }
```
piotrrak, line 216: it might be wort to swtich this API's to template:
and providing have overloads for:
 `template <typename T> parse(std::vector<std::string> properties)`
 `ct->create<T>(int key, std::vector<std::string>)`
 `ct->create<T>(std::string type, std::string)`

Thus on this level removing all those mostly identical functions, that could be tag dispatched to implementations. Similar idea to how layers are already created in codebase;

nntrainer/engine.h
```diff
@@ -0,0 +1,255 @@
...
+   * path from current working directory
+   */
+  const std::string getWorkingPath(const std::string &path = "");
+
+  /**
+   * @brief Set Working Directory for a relative path. working directory is set
+   * canonically
+   * @param[in] base base directory
+   * @throw std::invalid_argument if path is not valid for current system
+   */
```
piotrrak, line 235: I don't really understand what does it mean to set WD for a relative path.
Also I do know what the cannonical path is (as defined by Posix), but not in context of second sentence in `@brief`, I didn't get what does it mean to set working directory cannonicaly

nntrainer/meson.build
```diff
@@ -17,6 +17,8 @@ nntrainer_headers = [
    meson.current_source_dir() / 'nntrainer_logger.h',
    meson.current_source_dir() / 'nntrainer_error.h',
    meson.current_source_dir() / 'app_context.h',
+   meson.current_source_dir() / 'context.h',
+   meson.current_source_dir() / 'engine.h',
```
piotrrak, line 21: This whole block could be rewritten as:
```
nntrainer_headers = files(
   'nntrainer_log.h',
   'nntrainer_logger.h',
   'nntrainer_error.h',
   'app_context.h',
   'context.h',
   'engine.h',
)
```
meson `files()` function creates list of files referencing current_source_dir()` those are files that are checked for existance in current source dir and returns list of files, rather than paths that may or may not exist.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <base_properties.h>
+#include <context.h>
+#include <engine.h>
+
+#ifdef ENABLE_OPENCL
+#include <cl_context.h>
+#endif
+
+static std::string solib_suffix = ".so";
+static std::string contextlib_suffix = "context.so";
```
piotrrak, line 31: Maybe dlopen OS wrapper should handle solib_suffix; It's no-go for Windows with '.so' exension.

Such function would require only filename stem adding platform specific extension '.dll' vs '.so' vs '.dylib' if we'll ever make MacOS port.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+std::once_flag global_engine_init_flag;
+
+void Engine::add_default_object(Engine &eg) {
+  /// @note all layers should be added to the app_context to guarantee that
+  /// createLayer/createOptimizer class is created
+
+  nntrainer::AppContext *app_context = new nntrainer::AppContext();
+  app_context->Global();
+
+  eg.registerContext("cpu", app_context);
```
piotrrak, line 47: Could we swith registerContext to `std::unique_ptr`?
May `AppContext::Global()` throw now or in the future?

Or at least 
```c++
auto app_context = std::make_unique<AppContext>();
app_context->Global();
eg.regiestContext("cpu", app_context.release());
```
To make it exception-safety future-proof.

nntrainer/engine.cpp
```diff
@@ -0,0 +1,181 @@
...
+    ml_loge("registering layer failed due to unknown reason");
+  }
+};
+
+Engine &Engine::Global() {
+  static Engine instance;
+  /// in g++ there is a bug that hangs up if caller throws,
+  /// so registerer is noexcept although it'd better not
+  /// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70298
+  std::call_once(global_engine_init_flag, registerer, std::ref(instance));
```
piotrrak, line 70: To still be able to throw an exception `registerer` method could have output argument of reference to `std::exception_ptr`

```c++
void Engine::registerer(Engine &eg, std::exception_ptr &eptr) noexcept {
     try {
          add_default_object(eg);
     } catch (...) {
         eptr = std::current_exception();
     }
}

Engine &Engine::Global() {
    static Engine instance{};
   
    // in g++ there is a bug that hangs up if caller throws, so registerer is noexcept 
    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70298
    // We return any exception if it occurs during std::once_call as exception_ptr
    // to  rethrow it later outside the scope of std::call_once call;
 
    std::exception_ptr eptr{};

    std::call_once(glabal_engine_init_flag, registerer, std::ref(instance), std::ref(eptr));
   
    if (eptr) std::rethrow_exception(eptr);
}
```

nntrainer/tensor/memory_pool.cpp
```diff
@@ -122,18 +133,22 @@ void MemoryPool::allocate() {
...
   // void *ptr;
   // allocators.at("qnn")->alloc(ptr, memory_size.at(idx-1), 1);
-  // std::cout << idx << " : "<<std::endl;
-  // if(ptr == nullptr)
-  //   std::cout << "alloc failed"<<std::endl;
 
-  auto mem_data = std::make_shared<MemoryData>((void *)ptr);
+#ifdef ENABLE_QNN
+  auto mem_data = std::make_shared<MemoryData>((void *)memory_ptrs.at(idx - 1));
+#else
```
piotrrak, line 152: `#if defined(ENABLE_QNN)` also `calloc`'ed `mem_pool`, but here result of that allocation is not checked

meson.build
DonghakPark: Nice Catch

test/nntrainer_test_util.cpp
DonghakPark: Thank you for Review, i will apply your suggestion at Next PR

nntrainer/tensor/memory_pool.cpp
DonghakPark: That's right Thank you for review

nntrainer/meson.build
DonghakPark: Good!

nntrainer/engine.cpp
piotrrak: I am not that sure anymore now that my analysis here is correct. The lieftime of this argument should be ok.
Nonetheless it makes no sense to call `c_str()` in case where is direct constructor from string.

nntrainer/engine.cpp
piotrrak: Probably not the bug.

nntrainer/layers/tensor_layer.cpp
DonghakPark: you are right! i will add!

nntrainer/models/neuralnet.cpp
```diff
@@ -649,15 +639,47 @@ void NeuralNetwork::load(const std::string &file_path,
...
+
   switch (format) {
   case ml::train::ModelFormat::MODEL_FORMAT_BIN: {
     NNTR_THROW_IF(!initialized, std::runtime_error)
       << "Cannot load if not initialized yet, path: " << file_path
       << " format: " << static_cast<unsigned>(format);
 
     auto model_file = checkedOpenStream<std::ifstream>(
-      file_path, std::ios::in | std::ios::binary);
+      (v.size() == 2) ? v[0] : v[1], std::ios::in | std::ios::binary);
```
lhs8928, line 682: What if v.size() is equal to 1?

DonghakPark:
This PR's Commit included in #3049, So Close This PR

PR #2989, opened 2025-03-06
[application] namespace duplication in llama application @open sesame 03/07 15:57

baek2sm:
Previously, the `multi_head_attention_layer` is seperated into a custom layer called `custom_multi_head_attention_layer` for refactoring purposes. At that time, the namespace of custom layer should be changed, but they were copied over as-is, causing duplication. I've renamed namespace.

This issue was discovered and reported by @gkisalapl . I always appreciate your efforts. thanks a lot!

Even after resolving this problem, the llama application still does not work properly now.
I'll try to resolve this issue as soon as possible.

**Self evaluation:**
2. Run test:     [ ]Passed [ ]Failed [X]Skipped

### DIFF
--- a/Applications/LLaMA/jni/custom_multi_head_attention_layer.cpp
+++ b/Applications/LLaMA/jni/custom_multi_head_attention_layer.cpp
@@ -22,7 +22,7 @@
 #include <thread>
 #include <vector>
 
-namespace nntrainer {
+namespace custom {
 
 MultiHeadAttentionLayer::MultiHeadAttentionLayer() :
   multi_head_attention_props(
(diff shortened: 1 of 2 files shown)


piotrrak:
Just a thought, but it seems that instead of using qualified names ie. `nntrainter::TensorDim`

```c++
namespace custom {
using namespace nntrainer;
///... rest mostly without changes
}
```
And leaving this code with unqualified names should be sufficient, with little change/none of the code churn.

baek2sm:
@piotrrak 

> Just a thought, but it seems that instead of using qualified names ie. `nntrainter::TensorDim`
> 
> ```c++
> namespace custom {
> using namespace nntrainer;
> ///... rest mostly without changes
> }
> ```
> 
> And leaving this code with unqualified names should be sufficient, with little change/none of the code churn.

I applied it, thanks :)

EunjuYang (approved):
LGTM

djeong20 (approved):
LGTM :)

jijoongmoon (approved):
LGTM

PR #2964, opened 2025-02-21
[Wait for #2963] Meson run command check

piotrrak:
This cleans up meson warning:

WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in meson 2.0.
         See also: https://github.com/mesonbuild/meson/issues/9300

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [*]Skipped
2. Run test: [ ]Passed [ ]Failed [*]Skipped

**How to evaluate:**
N/A

PR #2999, opened 2025-03-10
[meson] Remove bogus resource copy command

piotrrak:
This change removes copy run_command that was always failing without us noticing. This was part of change 931c586ac23
"[Application] Add Mixed Precision Application"
Path Application/MixedPrecision/res has never existed.

This change partly addresses build failure in #2964 on Ubuntu

**Self evaluation:**

### DIFF
--- a/Applications/MixedPrecision/jni/meson.build
+++ b/Applications/MixedPrecision/jni/meson.build
@@ -1,9 +1,3 @@
-build_root = meson.build_root()
-res_path = meson.current_source_dir() / '..' / 'res'
-
-nntr_mixed_resdir = nntr_app_resdir / 'MixedPrecision'
-run_command('cp', '-lr', res_path, nntr_mixed_resdir)
-
 mixed_sources = [
   'main.cpp',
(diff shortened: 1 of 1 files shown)


piotrrak:
@DonghakPark  Just FIY: you might have take a look, maybe those resource files were added to the other place or were forgotten to be added.

djeong20 (approved):
Nice catch! It seems like 'res' under contains `xx.ini` files.
@DonghakPark Could you confirm that the `res` directory is not required for this application?

DonghakPark (approved):
**Nice Catch, It's my mistake. Thank you for Detail work!!**

As @djeong20 mentioned while working by copying the folder of another application, the method used by the other application remained the same. 

MixedPrecision Application directly defines the model, there is no ```ini file`` and it is not necessary.

dkjung (approved):
LGTM

haehun (approved):
Remove unnecessary commands, LGTM

EunjuYang (approved):

skykongkong8 (approved):
LGTM!

PR #3008, opened 2025-03-12
[hgemm] Removal of assert in get_prev_mltpl_of_2p_n

djeong20:
This patch removes the assert statement in the function get_prev_mltpl_of_2p_n, which previously required that the input value N be a multiple of two. After review, I have identified several instances where N is not necessarily a multiple of two.

For instance, 
https://github.com/nnstreamer/nntrainer/blob/9c0c43e9f47ef7a0a94a316d570245ae44b0d79c/nntrainer/tensor/cpu_backend/arm/hgemm/hgemm_util.cpp#L46

Therefore, I've decided to remove this assert statement.

**Self-evaluation:**


skykongkong8 (approved):
LGTM

baek2sm (approved):
LGTM

dkjung (approved):
LGTM

PR #3012, opened 2025-03-12
[gtest] Remove duplicated main() in integration unit test

piotrrak:
the test/unittest/integration_tests/integration_tests.cpp already contains main, causing linking to fail on some Ubuntu 22.04
The suspicion is that are those with multilib installed, due to libc difference.
The root cause of this issue exposing on some systems is not confirmed.
But both gtest and this compilation unit contain symbol 'main' thus change of dependency is correct fix.

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed []Skipped

### DIFF
--- a/test/unittest/integration_tests/meson.build
+++ b/test/unittest/integration_tests/meson.build
@@ -26,7 +26,7 @@ exe = executable(
     test_target,
     include_directories: [include_directories('.'), model_util_include_dir],
     dependencies: [
-        nntrainer_test_main_deps,
+        nntrainer_test_deps,
         nntrainer_dep,
         nntrainer_ccapi_dep,
     ],


piotrrak:
Failure indeed seems related for configuration:
```
[Ubuntu Meson build & test / meson_test (ubuntu-22.04, -Denable-fp16=true)...]
```
Test that failed:
`28/43 integration_tests                   FAIL             0.82s   killed by signal 11 SIGSEGV`

Though, why and how it did pass for the other configuration like `-Denable-fp16=false` is puzzling.

piotrrak:
> Failure indeed seems related for configuration:
> 
> ```
> [Ubuntu Meson build & test / meson_test (ubuntu-22.04, -Denable-fp16=true)...]
> ```
> 
> Test that failed: `28/43 integration_tests FAIL 0.82s killed by signal 11 SIGSEGV`
> 
> Though, why and how it did pass for the other configuration like `-Denable-fp16=false` is puzzling.

Rerunning the tests made it pass, pointing to some regression exhibiting at random introduced by recent changes.
Unless someone observed such crash ealier, was first to me.

gkisalapl (approved):

dkjung (approved):
LGTM

skykongkong8 (approved):

DonghakPark (approved):
LGTM!

baek2sm (approved):

PR #3019, opened 2025-03-13
[Fix] Fix tensor data type check .

gkisalapl:
In tensor apply operation there were error in data type check. This PR fix checking output data type

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/tensor.h
+++ b/nntrainer/tensor/tensor.h
@@ -15,10 +15,10 @@
 
 #define MAKE_SHARED_TENSOR(...) std::make_shared<nntrainer::Tensor>(__VA_ARGS__)
 
-#define CREATE_IF_EMPTY_DIMS(tensor, ...) \
-  do {                                    \
-    if (tensor.empty())                   \
-      tensor = Tensor(__VA_ARGS__);       \
+#define CREATE_IF_EMPTY_DIMS(tensor, ...)                                      \
+  do {                                                                         \
+    if (tensor.empty())                                                        \
+      tensor = Tensor(__VA_ARGS__);                                            \
   } while (0);
 
 #include <cstddef>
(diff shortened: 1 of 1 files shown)


DonghakPark (approved):
LGTM!

DonghakPark:
@djeong20 Please take a look, There are too many Unittest Fail

djeong20:
The arguments passed in for `CREATE_IF_EMPTY_DIMS` are incorrect. My bad :( 
Could you change it as follows?
```C++
CREATE_IF_EMPTY_DIMS(output, itensor->getDim(), nullptr);
```
This fix will resolve the issue!

gkisalapl:
> ments passed in for `CREATE_IF_EMPTY_DIMS` are incorrect. My bad :(
> Could you change it as fo

Fixing this check shown that more tests on FP16 build fail and need to be fixed, I'm working on it

myungjoo (approved):
It is always better if you can emit an exception message with debuggable information.
E.g.,
```
[Tensor::apply] output format or data type does not match
```
==>
```
[Tensor::apply] output format/type, "3:3:2,uint8", is not compatible with the given tensor, "3:3:2,float32".
```

However, the required infra is not ready. LGTM.

nntrainer/tensor/tensor.h
```diff
@@ -15,10 +15,10 @@
...
 
-#define CREATE_IF_EMPTY_DIMS(tensor, ...) \
-  do {                                    \
-    if (tensor.empty())                   \
-      tensor = Tensor(__VA_ARGS__);       \
+#define CREATE_IF_EMPTY_DIMS(tensor, ...)                                      \
+  do {                                                                         \
+    if (tensor.empty())                                                        \
+      tensor = Tensor(__VA_ARGS__);                                            \
   } while (0);
```
piotrrak, line 22: Those macros should have least possible scope to not affect/have inference with all included headers:
- cpu_backend.h
- nntrainer_log.h
- tensor_base_h
- half_tensor.h
Could we please move them below?

gkisalapl: Change in those lines are not related with fix scope - it was only result of applying clang-format on this file

piotrrak: Sure, can live with that.

gkisalapl:
> > ments passed in for `CREATE_IF_EMPTY_DIMS` are incorrect. My bad :(
> > Could you change it as fo
> 
> Fixing this check shown that more tests on FP16 build fail and need to be fixed, I'm working on it

@djeong20 You were right in terms of fix, thank you for tip

djeong20 (approved):
LGTM!

SeoHyungjun (commented):
LGTM

PR #3022, opened 2025-03-14
[ Layer ] Fix Input layer to get the Low precision tensor

jijoongmoon:
Input Layer only takes the FP32 type. So we do need to quantize it if the output of input layer is low-bit. Until the
quantization/dequantization support Tensor Class inside, we do use calling quantize explicitly on layer to support.

Resolves:

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/input_layer.cpp
+++ b/nntrainer/layers/input_layer.cpp
@@ -43,10 +43,26 @@ void InputLayer::setProperty(const std::vector<std::string> &values) {
 }
 
 void InputLayer::forwarding(RunLayerContext &context, bool training) {
+
   Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+  std::unique_ptr<nntrainer::Quantizer> quantizer;
   if (!context.getInPlace()) {
     Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
-    hidden_.copyData(input_);
+    quantizer = nullptr;
+
+    switch (hidden_.getDataType()) {
+      // This will be supported in Tensor itself. Until then, we use this.
+    case Tdatatype::QINT16:
+    case Tdatatype::UINT16: {
+      quantizer =
+        Quantization::createQuantizer(nntrainer::QScheme::PER_TENSOR_AFFINE);
+      Tensor dq_i = quantizer->quantize(input_, hidden_.getDataType());
+      hidden_.copyData(dq_i);
+    } break;
+    default:
+      hidden_.copyData(input_);
+      break;
+    }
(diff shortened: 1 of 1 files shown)


myungjoo (approved):

nntrainer/layers/input_layer.cpp
```diff
@@ -43,10 +43,25 @@ void InputLayer::setProperty(const std::vector<std::string> &values) {
...
   Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+  std::unique_ptr<nntrainer::Quantizer> quantizer;
   if (!context.getInPlace()) {
     Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
-    hidden_.copyData(input_);
+    quantizer = nullptr;
+
+    switch (hidden_.getDataType()) {
+      // This will be supported in Tensor itself. Until then, we use this.
+    case Tdatatype::UINT16: {
```
EunjuYang, line 55: Does this patch support UINT16 only
In #2984, it seems QINT16 is also considered for an activation type.
How about adding QINT16 support as well?
```suggestion
    case Tdatatype::QINT16: 
    case Tdatatype::UINT16: {
```

EunjuYang (commented):
LGTM

baek2sm (approved):
LGTM

EunjuYang (approved):

nntrainer/layers/input_layer.cpp
jijoongmoon: Thanks. fixed.

nntrainer/layers/input_layer.cpp
```diff
@@ -43,10 +43,26 @@ void InputLayer::setProperty(const std::vector<std::string> &values) {
...
+
   Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
+  std::unique_ptr<nntrainer::Quantizer> quantizer;
   if (!context.getInPlace()) {
     Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
-    hidden_.copyData(input_);
+    quantizer = nullptr;
+
+    switch (hidden_.getDataType()) {
+      // This will be supported in Tensor itself. Until then, we use this.
```
djeong20, line 54: When it says "This will be supported in tensor," does it mean that copying data from a fp32 tensor to a uint16 tensor needs to be automatically quantized?

skykongkong8 (approved):

DonghakPark:
This PR Already Merge into main branch, so close this PR, 0ef0cb80

PR #3029, opened 2025-03-18
[Tensor] Add absolute & tangent tensor operation

baek2sm:
- Added absolute tensor operation for fp32 and fp16
- Added tangent tensor operation for fp32 and fp16

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -625,6 +625,12 @@ Tensor &FloatTensor::sum(unsigned int axis, Tensor &output, float alpha,
   return output;
 }
 
+Tensor &FloatTensor::abs(Tensor &output) const {
+  auto f = [](float in) { return std::abs(in); };
+  apply(f, output);
+  return output;
+}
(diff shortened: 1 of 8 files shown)


djeong20 (approved):
Nice work!

skykongkong8 (commented):
FYI) `abs` and `tan` can be accelerated via SIMD. This work can be set as higher priority if this function is frequently used, but can be introduced at later PR as well.

skykongkong8 (approved):

skykongkong8 (commented):
Just a simple idea. How about adding more unittest TCs after adding new tensor ops like these?

baek2sm:
> Just a simple idea. How about adding more unittest TCs after adding new tensor ops like these?

@skykongkong8 I agree. Adding unit tests is definitely necessary. I've added tangent unit tests at the layer level (in #3035), and I plan to add them at the tensor level as well. It's included in my work plan in 2 weeks. Thanks!

jijoongmoon (approved):
LGTM

PR #3030, opened 2025-03-18
[tensor] create bcq weight on read

lhs8928:
- Create BCQW on every dot opertion causes serious performance degradation.

### DIFF
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -16,8 +16,6 @@
 #include <tensor.h>
 #include <util_func.h>
 
-#include "BiQGEMM.h"
-
 namespace nntrainer {
 
 BCQTensor::BCQTensor(std::string name_, Tformat fm) :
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -285,6 +270,8 @@ void BCQTensor::read(std::ifstream &file) {
   checkedRead(file, (char *)getData(), sz,
               "[BCQTensor::read] operation failed");
   putData();
+
+  createBCQW();
```
EunjuYang, line 274: What if `read()` is not called ? Is there no dangling issue?

EunjuYang (commented):
It seems better to create the BCQ weight before calling the dot operation. 
However, I'm curious whether the weight creation is always guaranteed to be invoked before the dot operation.

nntrainer/tensor/bcq_tensor.cpp
lhs8928: Added createBCQW function in constructor

EunjuYang: Thank you for the comment and update! 👍

EunjuYang (approved):
LGTM

DonghakPark (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #3032, opened 2025-03-18
Ruy include only

piotrrak:
Since we never link ruy it is not required to build it.
It is not possible by cmake subproject in meson, but we can provide  own overlay using meson wrap system.

**Self evaluation:**

**How to evaluate:**
N/A - no functional change

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,6 +2,3 @@
 	path = subprojects/iniparser
 	url = https://github.com/ndevilla/iniparser.git
 
-[submodule "subprojects/ruy"]
-	path = subprojects/ruy
-	url = https://github.com/google/ruy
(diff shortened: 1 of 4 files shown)


djeong20:
Currently, the Ruy project is not linked to a general environment. This is because the Ruy interface is not yet ready, as it has been given low priority. However, linking is planned for the future. 

Regarding Android, the project has been linked within the Android environment, although I haven't yet worked on the Android test aspects.

https://github.com/nnstreamer/nntrainer/blob/5086fb991b15b5b40f0e13328e7acd3efc96d622/jni/Android.mk.in#L57-L82

https://github.com/nnstreamer/nntrainer/blob/5086fb991b15b5b40f0e13328e7acd3efc96d622/jni/Android.mk.in#L100

PR #3033, opened 2025-03-19
[Wait for #3029][Layer] Add SQRT Layer

baek2sm:
Added SQRT Layer and Unittest, and this PR contains below 4 commits.

- [layer] add sqrt layer structure
- [layer] implement forwarding & backwarding of sqrt layer
- [layer] optimize backwarding of sqrt layer
- [layer] add unittest for SQRT layer

**Self evaluation:**


nntrainer/tensor/float_tensor.cpp
```diff
@@ -677,11 +677,7 @@ void FloatTensor::tan(Tensor &output, float alpha) {
 }
 
 void FloatTensor::inv_sqrt(Tensor &out) {
-  if (!contiguous) {
-    apply([](float val) -> float { return 1 / std::sqrt(val); }, out);
-  } else {
-    inv_sqrt_inplace(out.size(), out.getData<float>());
-  }
```
djeong20, line 684: Could you explain why `inv_sqrt_inplace` is removed? It might make the computation slower.

djeong20 (approved):
Overall, LGTM!

dkjung (approved):
LGTM

nntrainer/tensor/float_tensor.cpp
baek2sm: That function caused unittest failures even for this very simple model, so I modified not to use it for now.
```
makeGraph({{"input", {"name=in", "input_shape=1:1:2"}},
               {"fully_connected", {"name=fc", "unit=2", "input_layers=in"}},
               {"sqrt", {"name=sqrt_layer", "input_layers=fc"}},
               {"mse", {"name=loss", "input_layers=sqrt_layer"}}});
```

jijoongmoon (approved):
LGTM! I also suggest we have to consider how we could support the fused op later.

nntrainer/layers/sqrt_layer.h
```diff
@@ -0,0 +1,124 @@
...
+  SQRTLayer(SQRTLayer &&rhs) noexcept = default;
+
+  /**
+   * @brief  Move assignment operator.
+   * @parma[in] rhs SQRTLayer to be moved.
+   */
+  SQRTLayer &operator=(SQRTLayer &&rhs) = default;
+
+  /**
+   * @copydoc Layer::finalize(InitLayerContext &context)
```
dkjung, line 56: Does this copy the super class' docstring?

dkjung (approved):
LGTM

nntrainer/tensor/float_tensor.cpp
dkjung: It seems that it has to work.

nntrainer/layers/sqrt_layer.h
baek2sm: That's right!

PR #3036, opened 2025-03-19
[BCQTensor] add argmin & bugfix in BCQTensor @open sesame 13:23

EunjuYang:
- Implement argmin() function for BCQTensor, which should be implemented
  to cover the virtual function (Based on @djeong20 's opinion, this patch changed argmin and argmax as non-pure virtual instead of implementing the op for BCQTensor in this PR)
- matrixDotMatrix takes reference (not pointer)

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -276,38 +276,6 @@ void BCQTensor::read(std::ifstream &file) {
   createBCQW();
 }
 
-std::vector<unsigned int> BCQTensor::argmax() const {
-  std::vector<unsigned int> result;
-  const uint32_t *data = (uint32_t *)getData();
-  size_t batch_size = batch();
-  size_t feature_len = dim.getFeatureLen();
-
-  result.resize(batch_size);
-
-  for (unsigned int b = 0; b < batch_size; b++) {
-    auto max_iter =
-      std::max_element(data + b * feature_len, data + (b + 1) * feature_len);
-    result[b] = std::distance(data, max_iter) - (b * feature_len);
-  }
-  return result;
-}
-
-std::vector<unsigned int> BCQTensor::argmin() const {
-  std::vector<unsigned int> result;
-  const uint32_t *data = (uint32_t *)getData();
-  size_t batch_size = batch();
-  size_t feature_len = dim.getFeatureLen();
-
-  result.resize(batch_size);
-
-  for (unsigned int b = 0; b < batch_size; b++) {
-    auto min_iter =
-      std::min_element(data + b * feature_len, data + (b + 1) * feature_len);
-    result[b] = std::distance(data, min_iter) - (b * feature_len);
-  }
-  return result;
-}
-
 void BCQTensor::save_quantization_info(std::ostream &file) {
   checkedWrite(file, (char *)&quantized_bit_size, sizeof(uint16_t),
                "[BCQTensor::save] failed to write quantization information");
@@ -410,9 +378,9 @@ void BCQTensor::createBCQW() {
   size_t size_of_clusters[] = {width()};
   const size_t number_of_cluster = 1;
 
-  /// @note hidden_tile_size should be set as a multiple of 32. This variable is
-  /// related to the speed of matrixDotMatrix. The optimal value should be found
-  /// with various values according to the usage environment.
+  /// @note hidden_tile_size should be set as a multiple of 32. This variable
+  /// is related to the speed of matrixDotMatrix. The optimal value should be
+  /// found with various values according to the usage environment.
   size_t hidden_tile_size = 32;
 
   bcq_weight = std::make_shared<BiQGEMM::BCQW>(
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -292,6 +292,22 @@ std::vector<unsigned int> BCQTensor::argmax() const {
   return result;
 }
 
+std::vector<unsigned int> BCQTensor::argmin() const {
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -292,6 +292,22 @@ std::vector<unsigned int> BCQTensor::argmax() const {
...
 
+std::vector<unsigned int> BCQTensor::argmin() const {
+  std::vector<unsigned int> result;
+  const uint32_t *data = (uint32_t *)getData();
+  size_t batch_size = batch();
+  size_t feature_len = dim.getFeatureLen();
+
+  result.resize(batch_size);
+
+  for (unsigned int b = 0; b < batch_size; b++) {
```
SeoHyungjun, line 303: Since the type of batch_size is size_t, I think it's better for b to also be of type size_t.

```suggestion
  for (size_t b = 0; b < batch_size; b++) {
```

nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -292,6 +292,22 @@ std::vector<unsigned int> BCQTensor::argmax() const {
...
+    auto min_iter =
+      std::min_element(data + b * feature_len, data + (b + 1) * feature_len);
+    result[b] = std::distance(data, min_iter) - (b * feature_len);
+  }
```
SeoHyungjun, line 307: There are multiple identical multiplication operations in the code. 
Generally, addition takes less time than multiplication; 
would it be better to modify the code like the example?

```suggestion
  for (unsigned int b = 0, auto start_position = 0; b < batch_size; b++, start_position += feature_len) {
    auto min_iter =
      std::min_element(data + start_position, data + start_position + feature_len);
    result[b] = std::distance(data, min_iter) - start_position;
  }
```

SeoHyungjun (approved):
LGTM

nntrainer/tensor/bcq_tensor.h
```diff
@@ -218,6 +218,11 @@ class BCQTensor : public TensorBase {
    */
   std::vector<unsigned int> argmax() const override;
 
+  /**
+   * @copydoc Tensor::argmax()
```
mwlasiuk, line 222: Shouldn't it be argmin() in the docs comment?

nntrainer/tensor/bcq_tensor.cpp
EunjuYang: I updated it. PTAL

nntrainer/tensor/bcq_tensor.cpp
EunjuYang: Thank you for the suggestion.
For the code, let's make the change in another PR; this `argmin()` op is defined in other tensor type as well. Thus, I've tried to match it with their implementation.
It would be better to update other tensors' argmin() implementation at the same time :)

nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -83,325 +83,336 @@ void *BCQTensor::getData() const {
 
   data->validate();
   return data->getAddr<uint32_t>() + offset;
-}
```
djeong20, line 86: This should not be removed! Seems like this is why there are changes in the indent below.

nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -83,325 +83,336 @@ void *BCQTensor::getData() const {
...
-  checkedRead(file, (char *)&quantized_bit_size, sizeof(uint16_t),
-              "[BCQTensor::read] failed to read quantization information");
-}
+    for (size_t b = 0; b < batch_size; b++) {
+      auto min_iter =
+        std::min_element(data + b * feature_len, data + (b + 1) * feature_len);
+      result[b] = std::distance(data, min_iter) - (b * feature_len);
+    }
+    return result;
+  }
```
djeong20, line 304: I've been thinking about whether argmax and argmin are necessary. Specifically, would they even be applicable for the BCQ type? Currently, the definitions of argmax and argmin refer to finding the minimum or maximum value of the packed value, which doesn’t seem appropriate for BCQ.

1. We could implement argmax and argmin as non-pure virtual functions and raise an error if users attempt to call them.
2. If we choose to keep argmax and argmin, their implementation should differ. We need to consider how this tensor is quantized. For example, if the quantized bit size is 3, we should examine the maximum or minimum based on its representation (e.g., 111, 110, 101, 100, etc.).

EunjuYang: Oh you're right. I missed the characteristic's of bcq.
I'd be better to change argmax and argmin to be non-pure first to implement suitable way of argmin/max for bcq tensor. Thank you for the comment!

EunjuYang:
According to the opinion from @djeong20 , I've updated the PR to make `argmin()` and `argmax()` as non-pure virtual function, instead of implementing the op for BCQTensor.

djeong20 (approved):
LGTM!

nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -200,7 +200,7 @@ void BCQTensor::initialize(Initializer init) {
 
 Tensor &BCQTensor::dot(Tensor const &input, Tensor &output, bool trans,
                        bool trans_in, float beta) const {
-  BiQGEMM::matrixDotMatrix(output.getData(), bcq_weight, input.getData(),
+  BiQGEMM::matrixDotMatrix(output.getData(), *bcq_weight, input.getData(),
```
lhs8928, line 203: Good point!
I missed it in last PR

lhs8928 (approved):
LGTM

PR #3037, opened 2025-03-20
Update QScheme data type and add total memory size retrieval

djeong20:
This PR includes two commits that enhance the functionality of the tensor class:

1. Update QScheme data type to 16-bit unsigned integer

This commit changes the QScheme data type to a 16-bit unsigned integer. This modification ensures consistency with the BCQTensor in reading and saving quantization information, which simplifies calculating the offset.

2. [Tensor] Total Memory Size Retrieval

This commit introduces a new method, getMemoryBytes(), to the tensor class, which retrieves the total size of the allocated memory in bytes. The key difference between the existing bytes() function and the new method is that bytes() only returns the storage size for quantized data in quantized tensors, without accounting for the memory used for scaling factors and zero points. However, for float-type tensors, both bytes() and getMemoryBytes() will return the same value.

These changes improve the accuracy and consistency of memory size calculations, making it easier to manage and optimize memory usage in tensor operations.

### DIFF
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -257,12 +257,10 @@ void BCQTensor::save(std::ostream &file) {
   /// @note Save quantization information
   save_quantization_info(file);
 
-  size_t tensor_bytes = bytes() + scale_size() * sizeof(float);
-
-  std::streamsize sz = static_cast<std::streamsize>(tensor_bytes);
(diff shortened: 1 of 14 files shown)


jijoongmoon (approved):
LGTM!

EunjuYang (approved):
LGTM

DonghakPark (approved):
LGTM!! this change make FSU Logic more Simple

PR #3038, opened 2025-03-20
[Save] Update save action (from layer_node -> to layer-wise)

EunjuYang:
- This patch move action of save from layer_node to every layer
- Instead of branchiing the action of save in the common layer_node wrapper, it is more desirable to define every save action in layer
- This update refers to #2984, which enabled similar features in `read()`

**Self evaluation:**

Build test: [*]Passed [ ]Failed [ ]Skipped
Run test: [*]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -376,4 +376,46 @@ void BatchNormalizationLayer::setBatch(RunLayerContext &context,
   }
 }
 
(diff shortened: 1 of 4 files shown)


jijoongmoon (approved):
LGTM!

DonghakPark (approved):
LGTM!

haehun (approved):
Move save action to layer impls, LGTM

PR #3042, opened 2025-03-20
[Fix] Fix nntrainer with BiQGEMM enabled compilation

gkisalapl:
BCQTensor class does not implement pure virtual argmin method from base class. So it's not possible to create instance of this class. It cause that nntrainer cannot be compiled with BIQGEMM enabled

**Self-evaluation:**


gkisalapl:
Problem already resolved on latest main

PR #3043, opened 2025-03-20
[bcq] swap bcq height and width

lhs8928:
- For now dot product of bcq only supports weight x input while nntrainer supports input x weight. To handle this issue swap bcq's height and width info on purpose.
 - This commit is SHOULD BE REMOVED.

### DIFF
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -116,7 +116,7 @@ void *BCQTensor::getScale(size_t idx) const {
 }
 
 void *BCQTensor::getAddress(unsigned int i) {
-  size_t index = getIndex(batch(), channel(), height(), width() / 32);
+  size_t index = getIndex(batch(), channel(), width(), height() / 32);
   if (i > index) {
     return nullptr;
   }
@@ -124,7 +124,7 @@ void *BCQTensor::getAddress(unsigned int i) {
 }
 
 const void *BCQTensor::getAddress(unsigned int i) const {
-  size_t index = getIndex(batch(), channel(), height(), width() / 32);
+  size_t index = getIndex(batch(), channel(), width(), height() / 32);
   if (i > index) {
     return nullptr;
   }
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -247,7 +247,7 @@ void BCQTensor::save(std::ostream &file) {
   std::streamsize sz = static_cast<std::streamsize>(getMemoryBytes());
 
   NNTR_THROW_IF(sz < 0, std::invalid_argument)
-    << "save size: " << getMemoryBytes()
+    << "save size: " << bytes()
```
EunjuYang, line 250: I guess the save size should be getMemoryBytes
```suggestion
    << "save size: " << bytes()
```

nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -259,7 +259,9 @@ void BCQTensor::read(std::ifstream &file) {
   /// @note Read quantization information
   read_quantization_info(file);
 
-  std::streamsize sz = static_cast<std::streamsize>(getMemoryBytes());
```
EunjuYang, line 262: I find it would be better to keep the previous version

EunjuYang (commented):
Thank you for this PR.
This patch needs to be updated later when the kernel is changed.
Eventhough it seems the patch seems weird (as you mentioned in the description), it is necessaey update which makes the current kernel run.
It would be great if this patch includes some notes on the comment about this. Thanks for your hard work!

nntrainer/tensor/bcq_tensor.cpp
lhs8928: fixed it

nntrainer/tensor/bcq_tensor.cpp
lhs8928: fixed it

lhs8928:
> Thank you for this PR. This patch needs to be updated later when the kernel is changed. Eventhough it seems the patch seems weird (as you mentioned in the description), it is necessaey update which makes the current kernel run. It would be great if this patch includes some notes on the comment about this. Thanks for your hard work!

Thank you for your thorough review!

EunjuYang (approved):
LGTM!!👍

skykongkong8 (approved):
Nicely done!

myungjoo (approved):

baek2sm (approved):
LGTM!

jijoongmoon:
This is applied in #3051

PR #3045, opened 2025-03-20
Upgrade to c++20

gkisalapl:
This PR upgrades C++ standard version used to build nntrainer to c++20 It will allow us to use modern features added in this version of standard

**Self-evaluation:**

### DIFF
--- a/nntrainer/compiler/tflite_interpreter.cpp
+++ b/nntrainer/compiler/tflite_interpreter.cpp
@@ -744,25 +744,29 @@ TfOpNodes buildRealizedOpNodes(TfOpNodes &nodes,
         auto mul_epsilon =
           realized_nodes.back().get()->getAdditionalProps().at(0);
 
-        Tensor new_mul_weight(mul_gamma.getDim());
-        new_mul_weight.allocate();
-        new_mul_weight.copy(mul_gamma);
+        std::unique_ptr<Tensor> new_mul_weight =
+          std::make_unique<Tensor>(mul_gamma.getDim());
+        new_mul_weight->allocate();
+        new_mul_weight->copy(mul_gamma);
 
         // new_mul_weight = (gamma / sqrt(variance + epsilon))
         mul_variance.add_i(mul_epsilon);
         mul_variance.pow_i(-0.5f);
-        new_mul_weight.multiply_i(mul_variance);
+        new_mul_weight->multiply_i(mul_variance);
 
         // beta =  (beta - mean * gamma / sqrt(variance + epsilon))
-        Tensor sub_result(new_mul_weight.getDim());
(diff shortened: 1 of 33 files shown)


dkjung:
Let's see if every CI test passes

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -744,25 +744,29 @@ TfOpNodes buildRealizedOpNodes(TfOpNodes &nodes,
         auto mul_epsilon =
           realized_nodes.back().get()->getAdditionalProps().at(0);
 
-        Tensor new_mul_weight(mul_gamma.getDim());
-        new_mul_weight.allocate();
-        new_mul_weight.copy(mul_gamma);
+        std::unique_ptr<Tensor> new_mul_weight =
```
dkjung, line 747: I think... it would just work as the same when you changed this as a unique_ptr.

https://www.reddit.com/r/cpp_questions/comments/pqx7x7/can_smart_pointers_point_to_normal_variables/

Because `new_mul_weight` is just a local variable, not a member variable of a class, I think Tensor's destructor would be invoked at the end of the block(scope of the local variable).

Not 100% sure.

dkjung (approved):
Minor comment

jijoongmoon:
I think we need to other platform like Tizen, Android, etc. to upgrade  to C++ 20.

nntrainer/compiler/tflite_interpreter.cpp
```diff
@@ -744,25 +744,29 @@ TfOpNodes buildRealizedOpNodes(TfOpNodes &nodes,
...
         sub_result.allocate();
-        sub_result.copyData(new_mul_weight);
+        sub_result.copyData(*new_mul_weight);
 
         mul_mean.multiply_i(sub_result);
         mul_beta.subtract_i(mul_mean);
-        new_mul_weight.setName("MUL");
+        new_mul_weight->setName("MUL");
+        for (auto weight : removed_weights) {
+          delete weight;
```
haehun, line 766: good catch

haehun (approved):
LGTM, now linux will use c++20 as well

myungjoo (approved):

skykongkong8 (approved):

gkisalapl:
> I think we need to other platform like Tizen, Android, etc. to upgrade to C++ 20.

Done

EunjuYang (approved):
LGTM

DonghakPark (approved):
I am very glad that this PR allows us to upgrade the C++ standard!!!

jijoongmoon:
I think it is a little bit early to upgrade c++20. We will think about it later.

PR #3046, opened 2025-03-21
[ bugfix ] Fix fp16 is_valid algorithm bug

skykongkong8:
- Followed by discussion made from #3001, naive reinterpret cast is an undefined behavior. (Thanks @piotrrak !)
- Instead, using std::memcpy will do.
- Meanwhile, SIMD instruction for reinterpret cast will work as the way it is expected.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl_fp16.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl_fp16.cpp
@@ -51,7 +51,8 @@ bool is_valid(const unsigned int N, const __fp16 *input) {
     if (val != val) {
       return false;
     }
-    uint16_t val_bits = reinterpret_cast<uint16_t &>(val);
+    uint16_t val_bits;
+    std::memcpy((char *)&val_bits, (const char *)&val, sizeof(__fp16));
     if (val_bits == 0x7C00 || val_bits == 0xFC00) {
       return false;
     }
--- a/nntrainer/tensor/swap_device.h
+++ b/nntrainer/tensor/swap_device.h
@@ -24,6 +24,7 @@
 #include <sys/types.h>
 #include <system_error>
 #include <utility>
+#include <vector>
 #if defined(_WIN32)
 #include <io.h>
 #define O_SYNC 0UL


nntrainer/tensor/swap_device.h
```diff
@@ -24,6 +24,7 @@
 #include <sys/types.h>
 #include <system_error>
 #include <utility>
+#include <vector>
```
skykongkong8, line 27: I intentionally added this because this evoked some android build error from my local

haehun (approved):
fix bug from reinterpret_cast, LGTM

nntrainer/tensor/cpu_backend/arm/neon_impl_fp16.cpp
```diff
@@ -51,7 +51,8 @@ bool is_valid(const unsigned int N, const __fp16 *input) {
     if (val != val) {
       return false;
     }
-    uint16_t val_bits = reinterpret_cast<uint16_t &>(val);
+    uint16_t val_bits;
+    std::memcpy((char *)&val_bits, (const char *)&val, sizeof(__fp16));
```
myungjoo, line 55: 1. wth is ```val != val``` ? (LINE 51)
2. std::memcpy is a bit too much. To avoid any casting issue, you may also try:
```
union {
  __fp16 val;
  uint16_t bits;
  } v;
  v.val = input[i];
  if (v.val != v.val) {
    return false;
 }
 if (v.bits == 0x7c00 || v.bits == 0xfc00) {
  return false;
}
```


```
00018 void *memcpy(void *dst, const void *src, size_t len)
00019 {
00020         size_t i;
00021 
00022         /*
00023          * memcpy does not support overlapping buffers, so always do it
00024          * forwards. (Don't change this without adjusting memmove.)
00025          *
00026          * For speedy copying, optimize the common case where both pointers
00027          * and the length are word-aligned, and copy word-at-a-time instead
00028          * of byte-at-a-time. Otherwise, copy by bytes.
00029          *
00030          * The alignment logic below should be portable. We rely on
00031          * the compiler to be reasonably intelligent about optimizing
00032          * the divides and modulos out. Fortunately, it is.
00033          */
00034 
00035         if ((uintptr_t)dst % sizeof(long) == 0 &&
00036             (uintptr_t)src % sizeof(long) == 0 &&
00037             len % sizeof(long) == 0) {
00038 
00039                 long *d = dst;
00040                 const long *s = src;
00041 
00042                 for (i=0; i<len/sizeof(long); i++) {
00043                         d[i] = s[i];
00044                 }
00045         }
00046         else {
00047                 char *d = dst;
00048                 const char *s = src;
00049 
00050                 for (i=0; i<len; i++) {
00051                         d[i] = s[i];
00052                 }
00053         }
00054 
00055         return dst;
00056 }
```

piotrrak: @myungjoo union-cast is Undefined Behavior. Once union member becomes initialized it becomes active member and one can only reference to this data using this member. It can lead to miss-compilation.
You can find lot of information about union type puning. Example: https://tttapa.github.io/Pages/Programming/Cpp/Practices/type-punning.html#why-can-t-i-use-a-union-for-type-punning

The only case it'd be possible if the leader types of all members are same:
```c++
struct Foo { int a; float b; };
struct Bar { int a; uint32_t b; };

union C {
  Foo foo;
  Bar bar; 
};

C c;
// both accesses are legal and still you'll have to inform compiler that you do it referencing non active union member
// - c.foo.a;
// - c.bar.a;
// You need to use c++17 std::launder to such access

// Referencing c.foo.b when c.bar.b was initialized or c.bar.b when c.foo.b was initialized won't be legal and will always be UB
```
`std::bit_cast` or `std::memcpy` are only valid was to do that.

myungjoo: Oh.. ok.. unlike C it's not allowed in C++ :(

piotrrak: It's really fine, compilers do know what std::memcpy is, thus it won't be really the copy in machine code level, just exactly what we do wan't eg. for arm should be something like `UBFX` or analogous.

myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #3047, opened 2025-03-21
[ neon ] Add qmul8 implementation

skykongkong8:
- Add missing int8 qmul neon implementation (missed due to current refact #2930)
- Refer to #2857 for details.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/armv7_neon.h
+++ b/nntrainer/tensor/cpu_backend/arm/armv7_neon.h
@@ -92,6 +92,14 @@ static inline uint32_t vaddvq_u32(uint32x4_t a) {
   return ret;
 }
 
+static inline int32x4_t vcvtnq_s32_f32(float32x4_t a) {
+  int32x4_t ret;
+  for (unsigned int i = 0; i < 4; ++i) {
+    ret[i] = std::lround(a[i]);
+  }
+  return ret;
+}
+
 #ifdef ENABLE_FP16
 /**
  * @brief macro for vfmaq_n_f16
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
@@ -22,6 +22,84 @@
 #endif
 
 namespace nntrainer::neon {
+static inline void __ele_qmul_kernel(int8_t *lhs, int8_t *rhs, int8_t *res,
+                                     unsigned int data_len,
+                                     const float lhs_scale,
+                                     const float rhs_scale,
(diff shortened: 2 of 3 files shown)


nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -22,6 +22,84 @@
...
+    res4_7 = vcvtnq_s32_f32(res_f32_1);
+    res8_11 = vcvtnq_s32_f32(res_f32_2);
+    res12_15 = vcvtnq_s32_f32(res_f32_3);
+
+    int8x16_t output = vcombine_s8(
+      vqmovn_s16(vcombine_s16(vqmovn_s32(res0_3), vqmovn_s32(res4_7))),
+      vqmovn_s16(vcombine_s16(vqmovn_s32(res8_11), vqmovn_s32(res12_15))));
+
+    output = vmaxq_s8(output, int8_min);
+    output = vminq_s8(output, int8_max);
```
myungjoo, line 71: output is already int8x16. why do you need max/min with int8min/max?

skykongkong8: I tested this issue on RPi. Skipping min/max clip will make values unstable.
Furthermore, it is quite common to add min/max saturation after computing with higher-precision and and converting it back to lower-precision like :
https://github.com/ggml-org/llama.cpp/blob/77f9c6bbe55fccd9ea567794024cb80943947901/examples/llava/clip.cpp#L411

nntrainer/tensor/cpu_backend/arm/armv7_neon.h
```diff
@@ -92,6 +92,14 @@ static inline uint32_t vaddvq_u32(uint32x4_t a) {
...
 }
 
+static inline int32x4_t vcvtnq_s32_f32(float32x4_t a) {
+  int32x4_t ret;
+  for (unsigned int i = 0; i < 4; ++i) {
+    ret[i] = std::lround(a[i]);
+  }
+  return ret;
+}
+
```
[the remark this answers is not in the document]
skykongkong8: Always nice to find more room for optimization!
I will refer to it, but just implemented a naive one just to avoid build fail for Tizen armv7l

jijoongmoon (approved):
LGTM

djeong20 (approved):
Quick question. Can we replace CharTensor::multiply's implementation with this function?

skykongkong8:
> Quick question. Can we replace CharTensor::multiply's implementation with this function?

Yes. It will be proposed later on.

PR #3051, opened 2025-03-22
Support Various Context and Memory Buffer

jijoongmoon:
### In this PR
This PR enables the Engine, Context, Context Data and Context allocator.
#### Engine Class
In this PR, Engine Class which is managing the Context to create Layers for specicifc Accelerators.
AppContext create the CPU Layers, CLContext creates the GPU Layers. It also provides the Pluggable features, so that,
the developer can create it's own context to support various accelerators such as NPUs.
Thus, we have to use the ```getRegisterContext("cpu")``` call to get the proper Context. It also supports registerFactory to add custom layers as before.

#### Context Class
In this PR, Context Class is introduced. As mentioned in Engine class, it is possible to add custom context and this Context class might be the base class for all custom context. 
Context Class also has ```ContextData and MemAllocator```. ContextData can be used to create and used to utilize the global data for specific context after initialze the context and each specific accelerator need to spcific memory allocator. 

#### read and save the weight 
Previously layer_node handle the read and save the weight but, the read and save function can be differ depending on layer types such as batch normalization layer. So, In this pr, the read fuction is lowered to layer object itself. If there is no overwited read funciotn, it will use base class read, which is layer_devel.

#### RunContext with ContextData
In order to use the ContextData, additional parameter is set in RunLayerContext. In this way, it is possible to access in each layer. Also add compute_engine type inInitLayerContext as well.

#### Tensor and Weight Layer
This PR also introduce Tensor and Weight Layer which contain Tensors and Weights only. These layers act like container to provide the data.

#### FSU with shared Tensor
This PR also provides initial version of FSU with shared Weight Tensors. Most of the HW accelerators prefer the statically allocated memory rather than dynamically allocated memory. So modify function call of memory pool allocate and getMmeory but need to be tested more. For better latency, we also have to think the MAP_FIXED option in mmap and this also should be done later. For accurate controlablity of the Execution order, there is some fixes in manager as well.

#### Change in memory allocation in memory_pool
For Android, rpcmem is the default allocator and this pr properly compute the memory buffer depending offset. If the planner compute different size of memory for same offset, we allocate biggest memory size and use for shared buffer.


nntrainer/tensor/memory_pool.cpp
```diff
@@ -25,6 +25,29 @@
 #include <profiler.h>
 #include <vector>
 
+#if defined(_WIN32)
```
DonghakPark, line 28: It looks much better and more maintainable.!!

DonghakPark (approved):
LGTM!! i will make unit test base on this PR

nntrainer/layers/tensor_layer.cpp
```diff
@@ -0,0 +1,107 @@
...
+    tensor_idx.push_back(context.requestTensor(
+      t_dims[i], name, Initializer::NONE, true, t_life[i], true, engine));
+  }
+
+  context.setOutputDimensions(t_dims);
+}
+
+void TensorLayer::forwarding(RunLayerContext &context, bool training) {
+  if (!context.getInPlace()) {
+    for (unsigned int i = 0; i < n_tensor; ++i) {
```
baek2sm, line 89: It seems weird that one tensor actually consists of several different tensors. I think a tensor should be the smallest unit..

meson.build
```diff
@@ -109,6 +109,8 @@ if get_option('enable-fp16')
      add_project_arguments('-mfp16-format=ieee', language: ['c', 'cpp'])
      extra_defines += '-DENABLE_FP16=1'
      extra_defines += '-DUSE__FP16=1'
+     extra_defines += '-DUSE_NEON=1'
```
skykongkong8, line 112: I wonder if this option is really being used.

nntrainer/layers/tensor_layer.cpp
jijoongmoon: This can be fixed in later PRs after more discussions. But for now, we could merge this PR.

meson.build
jijoongmoon: Will remove this in later PR.

baek2sm (approved):

Applications/SimpleFC/jni/main.cpp
```diff
@@ -139,7 +139,20 @@ void createAndRun(unsigned int epochs, unsigned int batch_size,
...
+
+    auto load_end = std::chrono::system_clock::now();
+    std::chrono::duration<double> load_elapsed_seconds = load_end - start;
+    std::time_t load_end_time = std::chrono::system_clock::to_time_t(load_end);
+    std::cout << "Load finished computation at " << std::ctime(&load_end_time)
+              << "elapsed time: " << load_elapsed_seconds.count() << "s\n";
+  } else {
+    model->save(filePath, ml::train::ModelFormat::MODEL_FORMAT_BIN);
+    model->load(filePath);
+  }
```
EunjuYang, line 155: Update to use `file_path` which is passed as argument:
```suggestion
  // to test asynch fsu, we do need save the model weight data in file
  if (std::filesystem::exists(file_path)) {
    model->load(file_path);

    auto load_end = std::chrono::system_clock::now();
    std::chrono::duration<double> load_elapsed_seconds = load_end - start;
    std::time_t load_end_time = std::chrono::system_clock::to_time_t(load_end);
    std::cout << "Load finished computation at " << std::ctime(&load_end_time)
              << "elapsed time: " << load_elapsed_seconds.count() << "s\n";
  } else {
    model->save(file_path, ml::train::ModelFormat::MODEL_FORMAT_BIN);
    model->load(file_path);
  }
```

nntrainer/tensor/cache_elem.h
```diff
@@ -134,7 +134,7 @@ class CacheElem {
   size_t length;                        /**< element size */
   CachePolicy policy;                   /**< cache policy */
   std::shared_ptr<MemoryData> mem_data; /**< allocated memory data */
-  void *memory_ptr;                     /** memory ptr to store read data*/
```
EunjuYang, line 137: Previous version looks better.

EunjuYang (approved):
LGTM 🚀

skykongkong8 (approved):

Issue #3055, opened 2025-03-25
Value mismatch issue in llama application


baek2sm:
In the (only public) llama application, there is an issue where the values differ when verifying the converted weights from PyTorch to NNTrainer.

The current issue has been identified, but the exact cause has not been determined yet (it may be due to an custom_multi_head_attention issue ?).

PR #3056, opened 2025-03-25
[Application] Refactoring llama application

baek2sm:
In recent LLaMA implementations, unused MULTIPLE_OF parameter has been
removed and replaced with an intermediate_size parameter in the same
format as HuggingFace.

Additionally, It was modified to pass the weight, vocab, and merge paths
as parameters during app execution and update document.

**Self evaluation:**

### DIFF
--- a/Applications/LLaMA/jni/main.cpp
+++ b/Applications/LLaMA/jni/main.cpp
@@ -11,8 +11,6 @@
 #include <string.h>
 
 #include <algorithm>
-#include <array>
-#include <chrono>
 #include <ctime>
 #include <iostream>
 #include <iterator>
@@ -48,10 +46,9 @@ ModelHandle g_model;
 
 // Hyper params for LLaMA
 int const DIM = 2304;
+int const INTERMEDIATE_SIZE = 2048;
 int const NUM_LAYERS = 28;
 int const NUM_HEADS = 18;
-
-int const MULTIPLE_OF = 256;
 
 float const NORM_EPS = 0.000001f;
 float const EPS = 0.000001f;
@@ -59,7 +56,7 @@ int const NUM_VOCAB = 96000;
 int MAX_SEQ_LEN = 1024;
 int NUM_TO_GENERATE = 100;
 
-constexpr unsigned int INIT_SEQ_LEN = 28;
+constexpr unsigned int INIT_SEQ_LEN = 2;
 unsigned int batch_size = 1;
 unsigned int epoch = 1;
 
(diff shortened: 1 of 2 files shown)


Applications/LLaMA/jni/main.cpp
```diff
@@ -568,10 +563,7 @@ void run(std::string text, bool apply_temperature) {
   unsigned int init_len;
 
 #if defined(ENABLE_ENCODER)
-  std::string vocab_file_name = "../Applications/LLaMA/jni/vocab.json";
-  std::string merge_file_name = "../Applications/LLaMA/jni/merges.txt";
-
-  auto tokenizer = unwrap(GPT2Encoder::load(vocab_file_name, merge_file_name),
+  auto tokenizer = unwrap(GPT2Encoder::load(vocab_file_path, merge_file_path),
                           "Error initializising GPT2 tokenizer\n");
```
djeong20, line 567: Typo
```suggestion
                          "Error Initializing GPT2 tokenizer\n");
```

Applications/LLaMA/jni/main.cpp
```diff
@@ -750,15 +740,28 @@ int main(int argc, char *argv[]) {
...
 #else
-    bool apply_temp = (strcasecmp("true", args[1].c_str()) == 0);
+    bool apply_temp = (strcasecmp("true", args[3].c_str()) == 0);
 #endif
-    createAndRun(epoch, batch_size);
 
-    run(text, apply_temp);
+    createAndRun(epoch, batch_size, weight_path);
+
+    run(text, vocab_file_name, merge_file_name, apply_temp);
```
djeong20, line 764: Although the ENABLE_ENCODER option is set to false, the user still needs to provide vocab and merge files. Is this intended?

djeong20 (approved):

jijoongmoon (approved):
LGTM

skykongkong8 (approved):

Applications/LLaMA/jni/main.cpp
baek2sm: @djeong20 I didn't expect that you can catch this part, but it's a great point!
In fact, from the user's perspective, this app has no meaning if ENABLE_ENCODER is not enabled.
Not enabling ENABLE_ENCODER is actually a debugging option for our nntrainer contributors.
Therefore, I'll plan to make it so that LLaMA can only be executed when ENABLE_ENCODER is set to true,
but for now, there is a value mismatch issue (#3055), so I want to debug the problem and remove all the codes related to the case where ENABLE_ENCODER is disabled.

Applications/LLaMA/jni/main.cpp
baek2sm: Nice catch!

PR #3063, opened 2025-03-27
[UINT16] Update network_graph output part to support UIN16-A @open sesame (2025-03-31)

EunjuYang:
- This patch update network_graph's forwarding.
- This patch aims to support AUI16.
- Please note that this loop-based value copy is slow.

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -384,10 +384,12 @@ sharedConstTensors NetworkGraph::forwarding(
   for (unsigned int i = 0; i < graph.getNumOutputNodes(); ++i) {
     auto const &output_layer_node = LNODE(graph.getOutputNode(i));
     for (unsigned int j = 0; j < output_layer_node->getNumOutputs(); ++j) {
-      // @todo we should determine what type to return
-      // out.push_back(MAKE_SHARED_TENSOR(
-      //   output_layer_node->getOutput(j).clone(TensorDim::DataType::FP32)));
-      out.push_back(MAKE_SHARED_TENSOR(output_layer_node->getOutput(j)));
+      auto out_j = output_layer_node->getOutput(j);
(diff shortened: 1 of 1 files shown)


skykongkong8 (commented):
Sorry to interrupt, but shouldn't code like these should reside at tensor-op level?

EunjuYang:
> Sorry to interrupt, but shouldn't code like these should reside at tensor-op level?

Yes it is. I was confused with the opposite case. I replaced the code with clone operation. Please check the update.

djeong20 (approved):
LGTM!

dkjung (approved):
LGTM

nntrainer/graph/network_graph.cpp
```diff
@@ -384,10 +384,12 @@ sharedConstTensors NetworkGraph::forwarding(
...
-      // @todo we should determine what type to return
-      // out.push_back(MAKE_SHARED_TENSOR(
-      //   output_layer_node->getOutput(j).clone(TensorDim::DataType::FP32)));
-      out.push_back(MAKE_SHARED_TENSOR(output_layer_node->getOutput(j)));
+      auto out_j = output_layer_node->getOutput(j);
+      if (out_j.getDataType() == Tdatatype::FP32)
+        out.push_back(MAKE_SHARED_TENSOR(out_j));
+      else
+        out.push_back(
+          MAKE_SHARED_TENSOR(out_j.clone(TensorDim::DataType::FP32)));
```
jijoongmoon, line 392: We are using the tensor buffer Input and label in loss layer allocated from outside when it is running for training. How about doing same way input and output for the case of inference? Then we could remove this tensor clone.

EunjuYang: I see. It sounds much better. Actually, even with this update, we may still need some ways to put this tensor out.
Let me apply the similar way of what training mode does to the inference mode as well. Then, we can simply return it.

PR #3064, opened 2025-03-27
Bug Fix: OpenCL Context Registration in Engine Class for NNTrainer GPU Support

djeong20:
This pull request addresses a bug in the Engine class related to the registration of the OpenCL context. The newly introduced Engine class lacked the functionality to register the OpenCL context, which prevented NNTrainer from utilizing the GPU engine. Additionally, I corrected a typo that caused the registerContext function to fail in checking the GPU context, preventing it from registering properly with the engine. I have also updated the functions for context registration.

### DIFF
--- a/nntrainer/engine.h
+++ b/nntrainer/engine.h
@@ -28,8 +28,11 @@
 
 #include <context.h>
 #include <mem_allocator.h>
-
 #include <nntrainer_error.h>
+
+#ifdef ENABLE_OPENCL
+#include <cl_context.h>
+#endif
 
 namespace nntrainer {
 
--- a/test/unittest/layers/layers_dependent_common_tests.cpp
+++ b/test/unittest/layers/layers_dependent_common_tests.cpp
@@ -124,12 +124,13 @@ TEST_P(LayerSemantics, setBatchValidateLayerNode_p) {
 
 #ifdef ENABLE_OPENCL
 TEST_P(LayerSemanticsGpu, createFromClContext_pn) {
-  auto &ac = nntrainer::ClContext::Global();
+  auto &eg = nntrainer::Engine::Global();
+  auto cc = static_cast<nntrainer::ClContext *>(eg.getRegisteredContext("gpu"));
   if (!(options & LayerCreateSetPropertyOptions::AVAILABLE_FROM_APP_CONTEXT)) {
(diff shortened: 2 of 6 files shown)


nntrainer/engine.h
```diff
@@ -64,6 +67,32 @@ class Engine {
...
+      throw std::invalid_argument(ss.str().c_str());
+    }
+    engines.insert(std::make_pair(name, context));
+
+    auto alloc = context->getMemAllocator();
+
+    allocator.insert(std::make_pair(name, alloc));
+  }
+#endif
+
```
EunjuYang, line 95: What about reusing `static void registerContext(std::string name, nntrainer::Context *context) `?
Both `AppContext` and `ClContext` inherit Context.

test/unittest/layers/layers_dependent_common_tests.cpp
```diff
@@ -124,12 +124,13 @@ TEST_P(LayerSemantics, setBatchValidateLayerNode_p) {
 
 #ifdef ENABLE_OPENCL
 TEST_P(LayerSemanticsGpu, createFromClContext_pn) {
-  auto &ac = nntrainer::ClContext::Global();
+  auto &eg = nntrainer::Engine::Global();
+  auto ac = static_cast<nntrainer::ClContext *>(eg.getRegisteredContext("gpu"));
```
EunjuYang, line 128: `ac` sounds like `appContext`. What about naming it `cc` instead of `ac`?
```suggestion
  auto cc = static_cast<nntrainer::ClContext *>(eg.getRegisteredContext("gpu"));
```

EunjuYang (commented):
Good work to support ClContext in engine.
What about reusing the existing method instead of making a duplicate action with type overriding?

nntrainer/engine.h
djeong20: Thank you for your review! It has been removed.

test/unittest/layers/layers_dependent_common_tests.cpp
djeong20: Makes sense!

EunjuYang (approved):
LGTM

skykongkong8 (approved):
Looks much better

jijoongmoon (approved):
LGTM!

PR #3066, opened 2025-03-27
[meson] [windows]Limit number of D9025 on windows msvc

piotrrak:
Fixes for overriden /MT /MD

**Self evaluation:**

**How to evaluate:**

Add signed-off message automatically by running **$git commit -s ...** command.

$ git push origin <your_branch_name>
```

### CODE REVIEW

**Comment by piotrrak** [comment]
@gkisalapl FYI

File: meson.build [hunk 1]
```diff
@@ -318,18 +319,27 @@ if get_option('enable-blas')
...
-        blas_options.append_compile_args('cpp', windows_compile_args)
-      endif
       blas_options.add_cmake_defines({'BUILD_TESTING': false})
       blas_options.add_cmake_defines({'BUILD_BENCHMARKS': false})
       blas_options.add_cmake_defines({'BUILD_SHARED_LIBS': host_machine.system() != 'windows'})
       blas_options.add_cmake_defines({'BUILD_WITHOUT_LAPACK': true})
       blas_options.add_cmake_defines({'NOFORTRAN': true})
+      if host_machine.system() == 'windows'
+        blas_options.add_cmake_defines(common_win_subproject_cmake_defines)
+        blas_options.add_cmake_defines({'MSVC_STATIC_CRT': '1'})
```
dkjung, line 332: What does this option do?

dkjung (approved):
I got `CMake` errors for `openblas` while running `meson build ...` on my Windows laptop, but this PR remedied the problem. I looked into this PR, but it's hard to how it works. Could you put some brief notes about this?

meson.build
piotrrak: Replaces `/MD` to `/MT` the old way in `CFLAGS` `CXXFLAGS` - by just looping over all options. This was the way cmake could do it for you.

meson.build
```diff
@@ -318,18 +319,27 @@ if get_option('enable-blas')
...
+        if meson.is_cross_build() and host_machine.system() == 'windows'
+          blas_options.add_cmake_defines({
+            'TARGET': 'generic',
+            'TARGET_CORE': 'generic',
+            'BUILD_COMPLEX': false,
+            'BUILD_COMPLEX16': false,
```
dkjung, line 338: What do these options do?

piotrrak:
> I got `CMake` errors for `openblas` while running `meson build ...` on my Windows laptop, but this PR remedied the problem. I looked into this PR, but it's hard see exactly to how it works. Could you put some brief notes about this?

It enables static crt, trying to ensure we don't pass option to enable dynamic and override to static.

Please see: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
Other thing it distinguishes between build_machine and host_machine where we didn't do it.
Later part is the patch that lets crosscompile openblass adding generic core, so we don't compile assembly optimized kernels.

crosscompiling using msvc on linux running it through wine is just so one can fix windows compile errors. There is problem with crosscompiling preprocessed assembly, there is meson bug with crosscompilation using msvc not using RSP. Build will fail due too many arguments. So this build is to not have to move sources, and experimental.

skykongkong8 (approved):

subprojects/packagefiles/openblas/0001-Aid-windows-cross-compile-wine-MSVC.patch
```diff
@@ -0,0 +1,141 @@
...
+@@ -145,6 +145,38 @@ if (DEFINED CORE AND CMAKE_CROSSCOMPILING AND NOT (${HOST_OS} STREQUAL "WINDOWSS
+       set(CGEMM3M_UNROLL_N 4)
+       set(ZGEMM3M_UNROLL_M 4)
+       set(ZGEMM3M_UNROLL_N 4)
++  elseif ("${TCORE}" STREQUAL "generic")
++    file(APPEND ${TARGET_CONF_TEMP}
++      "#define L1_DATA_SIZE\t24576\n"
++      "#define L1_DATA_LINESIZE\t64\n"
++      "#define L2_SIZE\t524288\n"
++      "#define L2_LINESIZE\t64\n"
```
jijoongmoon, line 64: Are those sizes different from CPU architecture? I wonder these affect the performance depending on the architectures.

dkjung: These values have to be queried from the system, don't they?

piotrrak: Yes, performance of OpenBLAS depends on system it's build on. We've already talked internally about checking different TARGET_CORE options, yet it is currently dominated by threading overheads, so bit premature to test.

jijoongmoon: If it is then, those have to be configurable.

PR #3073, opened 2025-03-31
Fix meson.build for Applications on Windows

dkjung:
While meson-setting is being run on Windows, `xcopy` fails when it overwrites

with the build dir already containing resource files (mostly by previous build). 

This commit fixes it by adding `/Y` option.

**Self evaluation:**

### DIFF
--- a/Applications/AlexNet/jni/meson.build
+++ b/Applications/AlexNet/jni/meson.build
@@ -8,7 +8,7 @@ if build_machine.system() == 'windows'
   if not fs.exists (build_app_res_dir_win)
     run_command([prog_win_cmd, '/C', 'mkdir', build_app_res_dir_win], check: true)
   endif
-  run_command([prog_win_cmd, '/C', 'xcopy', app_res_dir_win, build_app_res_dir_win], check: true)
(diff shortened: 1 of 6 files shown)


Applications/AlexNet/jni/meson.build
```diff
@@ -8,7 +8,7 @@ if build_machine.system() == 'windows'
   if not fs.exists (build_app_res_dir_win)
     run_command([prog_win_cmd, '/C', 'mkdir', build_app_res_dir_win], check: true)
   endif
-  run_command([prog_win_cmd, '/C', 'xcopy', app_res_dir_win, build_app_res_dir_win], check: true)
```
gkisalapl, line 11: Why did you replaced prog_win_cmd with xcopy ?
Wouldn't it be sufficient to add /y option only ?

dkjung: It didn't work when I tested it.
The error was something like `cmd.exe doesn't have option "/Y"`.

djeong20 (approved):

baek2sm (approved):

lhs8928 (approved):
LGTM

skykongkong8 (approved):

PR #3075, opened 2025-04-01
[ cpu_backend ] Add skeleton for gemm_q4_K

skykongkong8:
To reviewers, this PR might be helpful for those who want to add a new function interface on `cpu_backend.h`.
Those who want to involve in adding more features here can come back to this PR, and implement from the beginning.

Proposes:
- Add a skeleton implementation for q4_K quantized GEMM.
- Internal representation will be proposed later on.
- Function parameters introduced here might change.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -210,4 +210,9 @@ bool is_valid(const unsigned int N, const float *input) {
   return nntrainer::neon::is_valid(N, input);
 }
 
+void gemm_q4_K(const unsigned int M, const unsigned int N, const unsigned int K,
+               const float *A, const unsigned int lda, const void *B,
+               const unsigned int ldb, float *C, const unsigned int ldc) {
+  return __gemm_q4_K(M, N, K, A, lda, B, ldb, C, ldc);
+}
 } /* namespace nntrainer */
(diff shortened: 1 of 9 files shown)


jijoongmoon (approved):
LGTM

dkjung (approved):
LGTM

dkjung:
1) `extern` keyword for function declaration doesn't have any effect, does it?

2) If there are multiple declarations and each of them has their own _docstring_, which one will be shown in the IDE's tooltip?

skykongkong8:
> 1. `extern` keyword for function declaration doesn't have any effect, does it?
> 2. If there are multiple declarations and each of them has their own _docstring_, which one will be shown in the IDE's tooltip?

1. I think so too, but I followed to internal suggestion on offline discussion. I will share with you somehow.
2. Did not really care about it, maybe I should copydoc later on. Thanks!

EunjuYang:
> > 1. `extern` keyword for function declaration doesn't have any effect, does it?
> > 2. If there are multiple declarations and each of them has their own _docstring_, which one will be shown in the IDE's tooltip?
> 
> 1. I think so too, but I followed to internal suggestion on offline discussion. I will share with you somehow.
> 2. Did not really care about it, maybe I should copydoc later on. Thanks!

I wonder about the `extern` too.
Please share it with me as well :)

EunjuYang:
In my opinion, it is more desirable to use `cpu_backend.h` only.
In `cpu_backend.h`, it includes different headers with the same function prototype.
It seems unnatural for me.

In #3085, I made lot of duplicate code for different backend headers.
The duplication can be removed by making the common headers are defined in `cpu_backend.h` and only the unique functions are declared in each backend headers.
Each `*_backend .cpp` includes `cpu_backend.h`

djeong20 (approved):
LGTM!

EunjuYang (approved):

PR #3069, opened 2025-03-28
GPU Context Initialization and Layer Object Creation

djeong20:
This pull request addresses issues related to the GPU context when creating the global instance and the GPU layer object.

**Changes proposed in this PR:**
- Modify ClContext::Global to return the newly created object rather than a static instance.
- Override the Context::createLayerObject function to correctly create a GPU layer object.

**Self-evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test:   [X]Passed [ ]Failed [ ]Skipped

PR #3076, opened 2025-04-01
[Wait for #3069]Consolidation and Bug Fixes for OpenCL Context Management

djeong20:
This pull request (PR) addresses several OpenCL errors related to ClContext, OpenCL kernels, and other components. The patch consolidates the ClContexts into a single instance, which has now been integrated into the Engine object. Additionally, the GPU test cases have been refactored to enable tests when FP16 is not in use.

1. [Test] Enable OpenCL full precision only unit test
2. [ClContext] Streamlining OpenCL Context Management
3. Bug Fix: OpenCL Kernel Creation Issues in ClContext

**Self-evaluation:**

### DIFF
--- a/nntrainer/cl_buffer_manager.cpp
+++ b/nntrainer/cl_buffer_manager.cpp
@@ -16,6 +16,7 @@ namespace nntrainer {
 
 ClBufferManager &ClBufferManager::getInstance() {
   static ClBufferManager instance;
+  instance.initBuffers();
   return instance;
 }
 
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -29,8 +29,6 @@ namespace nntrainer {
 
 std::mutex cl_factory_mutex;
 
-std::once_flag global_cl_context_init_flag;
-
 static void add_default_object(ClContext &cc) {
 
   if (FullyConnectedLayerCl::registerClKernels()) {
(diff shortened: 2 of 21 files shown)


EunjuYang (commented):
Great! This patch fixes duplicate context bug.
By the way, for the layer, they have common cc, but not for the kernels.
Is there any way to update them to have common interface?
Can we add static clcontext variable in `opencl_kernel.h`?

djeong20:
> Great! This patch fixes duplicate context bug. By the way, for the layer, they have common cc, but not for the kernels. Is there any way to update them to have common interface? Can we add static clcontext variable in `opencl_kernel.h` to share the GlobalCC with the same variable ?

Thank you for sharing your opinion! Could you explain why we need ClContext for the Kernel class?
AFAIK, the ClContext class creates OpenCL kernels with kernel strings and has an `ocl_kernel_map` variable to store each created kernel. As a result, one unique ClContext variable (GlobalCC) will have these registered kernels. Meanwhile, this patch fixed `blas_kernels.h` to use GlobalCC to register kernels. Please correct me if I'm misunderstanding.

EunjuYang:
> > Great! This patch fixes duplicate context bug. By the way, for the layer, they have common cc, but not for the kernels. Is there any way to update them to have common interface? Can we add static clcontext variable in `opencl_kernel.h` to share the GlobalCC with the same variable ?
> 
> Thank you for sharing your opinion! Could you explain why we need ClContext for the Kernel class? AFAIK, the ClContext class creates OpenCL kernels with kernel strings and has an `ocl_kernel_map` variable to store each created kernel. As a result, one unique ClContext variable (GlobalCC) will have these registered kernels. Meanwhile, this patch fixed `blas_kernels.h` to use GlobalCC to register kernels. Please correct me if I'm misunderstanding.

Yes! you're correct. The action itself remains unchanged even with the suggested opinion. What I intended was to make one common cc varialbe for `attention_cc` and `blas_cc`. What I suggested was not making a context for Kernel class but making one common static variable for kernel implementation. Do you think it is common to create a separate file for new kernel like attention?

EunjuYang (approved):

dkjung (approved):
LGTM

jijoongmoon (approved):
LGTM

djeong20:
> Yes! you're correct. The action itself remains unchanged even with the suggested opinion. What I intended was to make one common cc varialbe for `attention_cc` and `blas_cc`. What I suggested was not making a context for Kernel class but making one common static variable for kernel implementation. Will it be common to create a separate file for new kernel like attention?

I understand your point now. Your suggestion can be addressed during the refactoring process. I believe there is no need to create separate files for BLAS kernels and attention kernels. Once I resolve the ongoing GPU execution issues, I will focus on creating a unified interface for both attention and BLAS kernels, which will ultimately utilize a single static variable for these kernels.

Issue #3065, opened 2025-03-27
[Tensor] various types support for copyData


EunjuYang:
In NNTrainer, we plan to support more diverse types of activations. To achieve this, the `input_layer` should support type casting between float tensors and other types of tensors. 

Currently, `copyData()`, which is one of the simplest methods for conversion, is not fully supported. For instance, `UIntTensor::copyData` only supports copying from UINT16 to UINT16 (or possibly from UINT16 to other UINT tensors - this should be validated as well). 

It would be beneficial if `copyData` was supported at the tensor level. Then, we could update the input layer's action in #3053 as well.

PR #3080, opened 2025-04-01
[Tensor] copy_fp32 implementation and its application in tensor copyData() @open sesame

EunjuYang:
- This commit implements template function copy_fp32 in x86_backend / fallback / arm_backend with __fallback
- This commit updates tensors' copyData() in uint/short/char tensor
- The reason I used template for `copy_fp32` is to use same signature function (overriden) in UIntTensor, which is implemented with template. In order to avoid the case where the type is not yet implemented, I implemented copy_fp32 as template as well.
- Please leave comment if you have better and simpler way to update
- This patch updates unittest for tensor: 1) update two negative cases as positive 2) add two more negative cases
- see also https://github.com/nnstreamer/nntrainer/issues/3065

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/char_tensor.cpp
+++ b/nntrainer/tensor/char_tensor.cpp
@@ -384,6 +384,9 @@ void CharTensor::copyData(const Tensor &from) {
   case ml::train::TensorDim::DataType::QINT8:
     copy(from.getData());
     break;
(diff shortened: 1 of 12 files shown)


skykongkong8 (approved):
Well done!

dkjung (approved):
LGTM

PR #3083, opened 2025-04-02
Bug Fix: OpenCL Kernel Registration Error through Lazy Initialization

djeong20:
This commit fixes an error that occurred during the registration of OpenCL kernels. The issue was caused by the static initialization order, with `std::call_once` in the Engine class being invoked before the kernel strings had been initialized. This patch introduces functions to lazily initialize the kernel strings, ensuring that they are fully initialized before they are used.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernels.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernels.cpp
@@ -27,10 +27,10 @@ void sgemv_cl(const float *matAdata, const float *vecXdata, float *vecYdata,
 
     if (TransA) {
       kernel_sgemv_ptr =
-        blas_cc->registerClKernel(sgemv_cl_kernel_, "sgemv_cl");
+        blas_cc->registerClKernel(getSgemvClKernel(), "sgemv_cl");
     } else {
-      kernel_sgemv_ptr =
-        blas_cc->registerClKernel(sgemv_cl_noTrans_kernel_, "sgemv_cl_noTrans");
+      kernel_sgemv_ptr = blas_cc->registerClKernel(getSgemvClNoTransKernel(),
+                                                   "sgemv_cl_noTrans");
     }
 
     if (!kernel_sgemv_ptr) {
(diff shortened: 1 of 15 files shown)


skykongkong8 (approved):

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -27,10 +27,10 @@ void sgemv_cl(const float *matAdata, const float *vecXdata, float *vecYdata,
...
     if (TransA) {
       kernel_sgemv_ptr =
-        blas_cc->registerClKernel(sgemv_cl_kernel_, "sgemv_cl");
+        blas_cc->registerClKernel(getSgemvClKernel(), "sgemv_cl");
     } else {
-      kernel_sgemv_ptr =
-        blas_cc->registerClKernel(sgemv_cl_noTrans_kernel_, "sgemv_cl_noTrans");
+      kernel_sgemv_ptr = blas_cc->registerClKernel(getSgemvClNoTransKernel(),
+                                                   "sgemv_cl_noTrans");
     }
```
EunjuYang, line 34: What about creating `getClKernel()` instead of calling `registerClKernel`?
As I understood, we are expecting the kernels are already registered in the initialization time.
Then it might be more natural to call `getClKernel("sgemv_cl");` when using kernels. With this code, it will create too much logs like: `Kernel already registered and initialized`.

dkjung (approved):
LGTM

nntrainer/tensor/cl_operations/blas_kernels.cpp
djeong20: Yes, you're right. Currently, there's an issue with registering kernels; some kernels are not properly registered during initialization. After I resolve the issue in the following PRs, I'll create a new function `getRegisteredClKernel()`!

EunjuYang (approved):
LGTM! I'm pleased to read your PRs enabling OpenCL support! Thanks!

PR #3090, opened 2025-04-03
Enhance OpenCL Context: Memory Allocator and Kernel Registration Improvements

djeong20:
This patch introduces a Memory Allocator for the OpenCL context, which was previously lacking, and ensures that the Context data captures information about GPU layers created by ClContext.
It fixes an error related to the initialization order of a static member variable, which affected the use of registered kernels in OpenCL Layers.
Additionally, the patch addresses issues in OpenCL kernel strings by modifying the concatenation logic and providing a temporary fix for SwiGLU and RMSNorm half-precision operations, ensuring the correct registration of half-precision kernels when FP16 is enabled.

### DIFF
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -74,6 +74,8 @@ static void add_default_object(ClContext &cc) {
 
 static void registerer(ClContext &cc) noexcept {
   try {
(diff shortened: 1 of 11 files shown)


EunjuYang (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #3093, opened 2025-04-03
[Softmax] change to online softmax

EunjuYang:
- This PR updates the softmax into online softmax
- This optimization fuses two loops into one
- I tested the performance improvement for several cases: 
    - b,c,h,w = 3, 1, 2048, 2048 -> 14.25 % speed-up 
    - b,c,h,w = 1, 1, 4096, 1024->  16.73 % speed-up 
    - b,c,h,w = 1, 1, 1024, 4096 ->  27.25% speed-up 

**Table**. average execution time (ms)
|b,c,h,w | BEFORE | AFTER|
|-- | -- | --|
|(3,1,2048,2048) | 149.5 | 128.2|
|(1,1,4096,1024) | 55.6 | 46.3|
|(1,1,1024,4096) | 52.1 | 37.9|

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -186,29 +186,25 @@ public:
 
     // prevent overflow
     Tensor tmp(width, input.getTensorType());
+    T max_val, sum_exp, max_prev;
     for (unsigned int i = 0; i < bch_size; i++) {
       T *ptr = output_data + i * width;
-
-      // find max value and subtract it
-      T max_value = *std::max_element(ptr, ptr + width);
-
-      tmp.setValue(max_value);
+      max_val = ptr[0];
+      sum_exp = 1;
+      for (unsigned int j = 1; j < width; ++j) {
+        max_prev = max_val;
+        max_val = std::max(max_prev, ptr[j]);
+        sum_exp = sum_exp * exp_util<T>(max_prev - max_val) +
+                  exp_util<T>(ptr[j] - max_val);
+      }
+      tmp.setValue(max_val);
+      // x-max
       saxpy(width, -1, tmp.getData<T>(), 1, ptr, 1);
+      // take exp : exp(x-max)
+      std::transform(ptr, ptr + width, ptr, [](T x) { return exp_util<T>(x); });
+      // normalize : exp(x-max)/sum_exp
+      sscal(width, 1.0 / sum_exp, ptr, 1);
     }
-
-    // take exp
-    output.apply<T>(exp_util<T>, output);
-
-    // take sum over the last dimension
-    Tensor sum = output.sum(3);
-
-    for (unsigned int i = 0; i < bch_size; i++) {
-      T *ptr = output_data + i * width;
-      std::transform(ptr, ptr + width, ptr,
-                     std::bind(std::divides<T>(), std::placeholders::_1,
-                               sum.getValue<T>(i)));
-    }
-
     return output;
   }
 
--- a/test/unittest/unittest_nntrainer_activations.cpp
+++ b/test/unittest/unittest_nntrainer_activations.cpp
@@ -97,6 +97,54 @@ TEST(nntrainer_activation, softmax_prime_01_p) {
   }
 }
 
+TEST(nntrainer_activation, softmax_big_test_1) {
+  int batch = 3;
+  int channel = 1;
+  int height = 2048;
+  int width = 2048;
+
+  nntrainer::Tensor input(batch, channel, height, width);
+  nntrainer::Tensor softmax_result;
+  GEN_TEST_INPUT(input, (i * (width) + k + l));
+
+  for (int i = 0; i < 10; ++i) {
+    EXPECT_NO_THROW(
+      input.apply(nntrainer::ActiFunc::softmax<float>, softmax_result));
+  }
+}
+
+TEST(nntrainer_activation, softmax_big_test_2) {
+  int batch = 1;
+  int channel = 1;
+  int height = 4096;
+  int width = 1024;
+
+  nntrainer::Tensor input(batch, channel, height, width);
+  nntrainer::Tensor softmax_result;
+  GEN_TEST_INPUT(input, (i * (width) + k + l));
+
+  for (int i = 0; i < 10; ++i) {
+    EXPECT_NO_THROW(
+      input.apply(nntrainer::ActiFunc::softmax<float>, softmax_result));
+  }
+}
+
+TEST(nntrainer_activation, softmax_big_test_3) {
+  int batch = 1;
+  int channel = 1;
+  int height = 1024;
+  int width = 4096;
+
+  nntrainer::Tensor input(batch, channel, height, width);
+  nntrainer::Tensor softmax_result;
+  GEN_TEST_INPUT(input, (i * (width) + k + l));
+
+  for (int i = 0; i < 10; ++i) {
+    EXPECT_NO_THROW(
+      input.apply(nntrainer::ActiFunc::softmax<float>, softmax_result));
+  }
+}
+
 TEST(nntrainer_activation, softmax_prime_02_n) {
   int batch = 3;
   int channel = 1;


nntrainer/layers/acti_func.h
```diff
@@ -186,29 +186,25 @@ class ActiFunc {
...
+        sum_exp = sum_exp * exp_util<T>(max_prev - max_val) +
+                  exp_util<T>(ptr[j] - max_val);
+      }
+      tmp.setValue(max_val);
+      // x-max
       saxpy(width, -1, tmp.getData<T>(), 1, ptr, 1);
+      // take exp : exp(x-max)
+      std::transform(ptr, ptr + width, ptr, [](T x) { return exp_util<T>(x); });
+      // normalize : exp(x-max)/sum_exp
+      sscal(width, 1.0 / sum_exp, ptr, 1);
```
skykongkong8, line 206: Guess we have a missing commit.. I implemented simd-accelerated exp function on util_simd.cpp It contains scaling option so above two function can be fused into a single function. LGTM for this PR and still excited to find more room for optimization!

skykongkong8 (approved):
LGTM! but how did you evaluated latency result?

EunjuYang:
> LGTM! but how did you evaluated latency result?

I added unit test case and check its running time. I don't think it's good way but just simply used to check the improvement.
```

TEST(nntrainer_activation, softmax_big_test_1) {
  int batch = 3;
  int channel = 1;
  int height = 2048;
  int width = 2048;

  nntrainer::Tensor input(batch, channel, height, width);
  nntrainer::Tensor softmax_result;
  GEN_TEST_INPUT(input, (i * (width) + k + 1));


  for(int i = 0; i < 10; ++i){
    EXPECT_NO_THROW(input.apply(nntrainer::ActiFunc::softmax<float>,softmax_result));
  }
}
```

```
[ RUN      ] nntrainer_activation.softmax_big_test_1
[       OK ] nntrainer_activation.softmax_big_test_1 (856ms)
```

nntrainer/layers/acti_func.h
EunjuYang: Sounds awesome! 👍👍

skykongkong8:
> > LGTM! but how did you evaluated latency result?
> 
> I added unit test case and check its running time. I don't think it's good way but just simply used to check the improvement.
> 
> ```
> 
> TEST(nntrainer_activation, softmax_big_test_1) {
>   int batch = 3;
>   int channel = 1;
>   int height = 2048;
>   int width = 2048;
> 
>   nntrainer::Tensor input(batch, channel, height, width);
>   nntrainer::Tensor softmax_result;
>   GEN_TEST_INPUT(input, (i * (width) + k + 1));
> 
> 
>   for(int i = 0; i < 10; ++i){
>     EXPECT_NO_THROW(input.apply(nntrainer::ActiFunc::softmax<float>,softmax_result));
>   }
> }
> ```
> 
> ```
> [ RUN      ] nntrainer_activation.softmax_big_test_1
> [       OK ] nntrainer_activation.softmax_big_test_1 (856ms)
> ```

Still seems nice for standalone test! Why did not you left it on the PR? I think it will be useful for me if you left it as the way it is.

EunjuYang:
@skykongkong8 , I added the test cases!

skykongkong8:
You are the best👍👍

nntrainer/layers/acti_func.h
dkjung: Um, definition of a function should be placed in `.c` or `.cpp`, shouldn't it?

dkjung (approved):
Looks good, minor comments:
- Set of _softmax_git_test_s doesn't seem to be checking the correctness, does it? If so, could you make such a test?
- This static function declaration is supposed to be in header while its definition is supposed to be in source file, isn't it?

EunjuYang:
> Looks good, minor comments:
> 
> * Set of _softmax_git_test_s doesn't seem to be checking the correctness, does it? If so, could you make such a test?
> * This static function declaration is supposed to be in header while its definition is supposed to be in source file, isn't it?
>   
>   * If so, its compiled objective code could be put in .obj file of other source file that include this header.

1) Exactly 🤣 I just made it to quickly compare the speed only, for the accuracy test is done in other test cases. (so I didn't contain it in this PR at the first time) As you mentioned, however, it would be much better to check its accuracy as well. Let me do it in another PR, for I have no room for the update for this PR right now :)

2) I think the reason why it is defined in header is it is implemented as a template function. Is there better way to separate it in .c or .cpp function? Then it would be better to update the `acti_func.h`! :)

dkjung:
Thx, I'm gonna check performance with this PR in.

dkjung:
I ran LLM on my laptop with/out this PR applied,

token gen. speed:
without: 12.052 TPS 
=>
with: 11.927 TPS

skykongkong8:
> I ran LLM on my laptop with/out this PR applied,
> 
> token gen. speed: without: 12.052 TPS => with: 11.927 TPS

Just curious, does softmax operation take a huge portion from the total model latency?

nntrainer/layers/acti_func.h
skykongkong8: For function template, isn't it `.h`?
Maybe explicitly instantiated function template can reside at `.cpp`

dkjung:
Windows, Core 9 185H 2.50GHz,

[ RUN      ] nntrainer_activation.softmax_big_test_1
[       OK ] nntrainer_activation.softmax_big_test_1 (1232 ms)
[ RUN      ] nntrainer_activation.softmax_big_test_2
[       OK ] nntrainer_activation.softmax_big_test_2 (405 ms)
[ RUN      ] nntrainer_activation.softmax_big_test_3
[       OK ] nntrainer_activation.softmax_big_test_3 (407 ms)

EunjuYang:
I updated the unittest case -> input generation was not good to compare btw/ big_test2 and 3.
Based on @dkjung 's analysis, I updated the input. 
By the way, with this update, unit test time increased.... so let me reduce the size if you think it hinders CI.

baek2sm (approved):
Nice work!!

EunjuYang:
> I ran LLM on my laptop with/out this PR applied,
> 
> token gen. speed: without: 12.052 TPS => with: 11.927 TPS

I think, for llama, softmax is applied to `attention_weight_step`, where the width is not that big (upto max seq_len).
Could you test with prefill as well?

dkjung:
Windows, Core 9 185H 2.50GHz,

|b,c,h,w | BEFORE | AFTER|
|-- | -- | --|
|(3,1,2048,2048) | 3103 | 3672 |
|(1,1,4096,1024) | 1001 | 1191|
|(1,1,1024,4096) | 1079 | 1231|


It seems it gets slower on Windows..😔

dkjung (changes requested):
The unit tests reported that the speed gets slower on Windows.
Let's have some more time to study for getting it run faster on Windows.

skykongkong8:
In a second glance, I just remembered I implemented 
```C++
void softmax_util(const unsigned int N, float *X, float *Y) {
  softmax(N, X, Y);
}
```
at util_simd.cpp😂
It is neon-accelerated function, but not accelerated with avx2.

skykongkong8:
Internally something like:
```C++
void softmax(const unsigned int N, float *X, float *Y) {
  unsigned int i = 0;
  float sum = 0.f;
  float max_x = max_val(N, X);
  float32x4_t max_x_v = vmovq_n_f32(max_x);
  for (; N - i >= 4; i += 4) {
    float32x4_t x0_3 = vld1q_f32(&X[i]);
    x0_3 = vsubq_f32(x0_3, max_x_v);
    float32x4_t exp0_3 = exp_ps(x0_3);
    sum += vaddvq_f32(exp0_3);
  }
  while (i < N) {
    sum += std::exp(X[i] - max_x);
    ++i;
  }

  i = 0;
  float32x4_t sum_vec = vmovq_n_f32(sum);
  for (; N - i >= 4; i += 4) {
    float32x4_t x0_3 = vld1q_f32(&X[i]);
    x0_3 = vsubq_f32(x0_3, max_x_v);
    float32x4_t exp0_3 = exp_ps(x0_3);
    float32x4_t softmax0_3 = vdivq_f32(exp0_3, sum_vec);
    vst1q_f32(&Y[i], softmax0_3);
  }
  while (i < N) {
    Y[i] = std::exp(X[i] - max_x) / sum;
    ++i;
  }
}
```

EunjuYang:
Dear reviewers (@dkjung , @skykongkong8 ),
I just tested this PR on Android; I checked this PR makes the softmax slower.
I think my test result on ubuntu cannot be generalized. Let me close this PR for this time.

EunjuYang:
@lhs8928 , As we discussed in offline, let me close this PR.

PR #3095, opened 2025-04-04
Remove Half-Precision Floating Point Support Check and Adjust Kernel Execution Work Sizes

djeong20:
This PR combines two commits that address improvements in our OpenCL implementation:

1. Removal of Half-Precision Floating Point Support Check: The first commit removes the requirement to check for half-precision floating point support on devices. In CUDA, enabling the cl_khr_fp16 pragma/feature macro allows for basic operations using half data types, including addition, subtraction, multiplication, and division, with the newer compiler. 

2. Adjust Kernel Execution Global and Local Work Sizes: The second commit updates the global and local work sizes used when dispatching OpenCL kernels. Previously set to a fixed value of 32, this could lead to errors when the input size was less than 32. The work sizes have now been adjusted to align with the specific requirements of each kernel, ensuring more efficient and accurate execution.

### DIFF
--- a/nntrainer/cl_buffer_manager.cpp
+++ b/nntrainer/cl_buffer_manager.cpp
@@ -16,7 +16,6 @@ namespace nntrainer {
 
 ClBufferManager &ClBufferManager::getInstance() {
   static ClBufferManager instance;
-  instance.initBuffers();
   return instance;
 }
 
--- a/nntrainer/layers/cl_layers/concat_cl.cpp
+++ b/nntrainer/layers/cl_layers/concat_cl.cpp
@@ -314,7 +314,8 @@ void ConcatLayerCl::concat_cl_axis3(const float *matAdata,
     }
 
     const int work_groups_count[3] = {dim, 1, 1};
-    const int work_group_size[3] = {32, 32, 1}; // test-value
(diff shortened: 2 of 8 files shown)


nntrainer/cl_buffer_manager.cpp
```diff
@@ -16,7 +16,6 @@ namespace nntrainer {
 
 ClBufferManager &ClBufferManager::getInstance() {
   static ClBufferManager instance;
-  instance.initBuffers();
```
dkjung, line 19: Is this not necessary?

dkjung (approved):
LGTM

nntrainer/cl_buffer_manager.cpp
djeong20: This already exists in `ClContext::clInit()`.

https://github.com/nnstreamer/nntrainer/blob/070b4eb5f2a722726f62b42e34006eabcca6793b/nntrainer/cl_context.h#L291

So, I'm removing it here.

jijoongmoon (approved):
LGTM!

SeoHyungjun (approved):
LGTM

skykongkong8 (approved):

PR #3098, opened 2025-04-07
[layer] add cast layer

baek2sm:
added cast layer (type casting).

- there is no model-unittest yet.
- In some cases, optimization may be considered (such as fusing only one
layer in front), but for now, only the basic implementation is included.
optimization will be applied through ONNX graph optimization work,
later.

**Self evaluation:**


skykongkong8 (approved):

jijoongmoon:
I couldn't get rid of thinking about the unary operation layer for all. It seems like repeating the same code in every layer. I think it is ok to have various properties or using some templete to set each functions.

baek2sm:
> I couldn't get rid of thinking about the unary operation layer for all. It seems like repeating the same code in every layer. I think it is ok to have various properties or using some templete to set each functions.

@jijoongmoon Okay, I will prepare a commit to merge Unary layers. However, this is not limited to Unary layers. From that perspective, most existing NNTrainer's layers should also be merged into two or three category."

PR #3101, opened 2025-04-08
[OpenCL] Load clFinish to make synchronization point

djeong20:
This PR adds loading clFinish to block until all previous queued OpenCL commands have been completed.

**Self-evaluation:**

### DIFF
--- a/nntrainer/opencl/opencl_command_queue_manager.cpp
+++ b/nntrainer/opencl/opencl_command_queue_manager.cpp
@@ -362,6 +362,8 @@ bool CommandQueueManager::DispatchCommand(
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


EunjuYang (approved):

EunjuYang (approved):
LGTM! Not directly related to this PR. Could you please check if we need to use `clReleaseKernel`? In opencl_loader.h, there is no extern for these functions, which means we don't use clReleaseKernel.

skykongkong8 (approved):
LGTM! Explanations from offline helped me to understand a lot :)

jijoongmoon (approved):
LGTM

PR #3104, opened 2025-04-09
[tensor] modify dotBatch for gqa

lhs8928:
[tensor] Enhance dotBatch to enable GQA(Grouped Query Attention)
    
     - In GQA(Grouped Query Attention) given two tensors have a different batch sizes.
       So we need to figured out which samples are corresponding to each other.
       LCM(least common multiple) is used as a key point to figure this out.
       Each sample performs a dot operation one by one with corresponding group samples.
    
       In case of the batch size of the given tensor does not equal, the bigger on should be a multiple of the smaller one.

    [test] Add dot batch unittest
    
     - Added dot batch unittest for following cases
      (1) If the batch sizes of the given two tensors does not have common multiple.
      (2) If the batch sizes of the bigger one is not a multiple of the smaller one.
      (3) If the batch sizes of the bigger one is a multiple of the smaller one.

### DIFF
--- a/nntrainer/tensor/tensor.h
+++ b/nntrainer/tensor/tensor.h
@@ -1361,7 +1361,9 @@ public:
   /**
    * @copydoc Tensor::dot(Tensor const &input, Tensor &output, bool trans,
               bool trans_in, float beta) const
-   * @details performs dot operation over a batch of inputs
+   * @details performs dot operation over a batch of inputs. If the batch sizes
+   of the given two tensors are different, the bigger one should be a multiple
+   of the smaller one.
    */
   Tensor &dotBatched(Tensor const &input, Tensor &result, bool trans = false,
                      bool trans_in = false, float beta = 0.0f) const;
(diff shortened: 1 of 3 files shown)


jijoongmoon:
I wonder what '[Temp]' means in title.

lhs8928:
> I wonder what '[Temp]' means in title.

I means it was under work in process.
I wondered that it could be improved by perform dot operation on group of batches. 
But it has some issues when the rhs operand has bigger batch size.
And I added a unit test to verify enhanced dot batch.

nntrainer/tensor/tensor.h
```diff
@@ -1361,7 +1361,9 @@ class Tensor {
   /**
    * @copydoc Tensor::dot(Tensor const &input, Tensor &output, bool trans,
               bool trans_in, float beta) const
-   * @details performs dot operation over a batch of inputs
+   * @details performs dot operation over a batch of inputs. If the batch sizes
+   of the given two tensors are different, the bigger one should be a multiple
+   of the smaller one.
```
dkjung, line 1366: What will happen if two sizes are not the same? It would turn more helpful if you left some comments.

dkjung (approved):

nntrainer/tensor/tensor.h
lhs8928: As mentioned in comment if the two batch size of given tensor does not equal, then the bigger one should be a multiple of  the smaller one.
Otherwise it will throw an error.

EunjuYang (approved):
LGTM

PR #3071, opened 2025-03-28
[FSU] Update FSU getBuffer (working with memcpy) @open sesame

EunjuYang:
- The previous version works with CPU.
- However, it does not properly work with other memalloc manner.
- Until fixing the issue and validating the action, we will use this temporary action, i.e., memcpy

**Self evaluation:**

Build test: [X]Passed [ ]Failed [ ]Skipped
Run test: [X]Passed [ ]Failed [ ]Skipped

PR #3105, opened 2025-04-09
[FSU] Update FSU getBuffer (memcpy)

DonghakPark:
In this PR
----

[FSU] Update FSU getBuffer (memcpy)

Update MMAP offset, read size calc for support various weight size
- it will use memcpy for support MMAP from not aligned size

This PR include Previous @EunjuYang change 
- #3071 

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/swap_device.cpp
+++ b/nntrainer/tensor/swap_device.cpp
@@ -61,44 +61,39 @@ void *SwapDevice::getBuffer(off_t offset, size_t size, void *memory_ptr,
     << "SwapDevice: Device is not started";
 
 #ifdef USE_MMAP
-  // page aligned
   if (execution_mode == ml::train::ExecutionMode::INFERENCE) {
+    // FSU Load Weights
     auto len_offset = weight_offset.at(offset_index);
+    size_t off =
+      (len_offset.first / sysconf(_SC_PAGE_SIZE)) * sysconf(_SC_PAGE_SIZE);
+    size_t diff = len_offset.first - off;
+    size_t len = len_offset.second + diff;
 
-    /**
-     * @todo
-     * Please note that the following branch is for the case where using ION
-     * memory under the __ANDROID__ flag. It is not a desirable way to branch
-     * the code. We will fix this after checking the direct use of mmap with ION
-     * memory. With this option, we temporarily use memcpy.
-     */
-#if defined(__ANDROID__)
-    void *ptr =
-      mmap(nullptr, len_offset.second, PROT_READ | PROT_WRITE | PROT_EXEC,
-           MAP_PRIVATE, fd, len_offset.first);
-#else
-    void *ptr =
-      mmap(memory_ptr, len_offset.second, PROT_READ | PROT_WRITE | PROT_EXEC,
-           MAP_SHARED | MAP_FIXED, fd, len_offset.first);
-#endif
+    char *ptr = static_cast<char *>(mmap(
+      nullptr, len, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_SHARED, fd, off));
 
-    // mmap check
     const size_t error_buflen = 100;
     char error_buf[error_buflen];
     NNTR_THROW_IF(ptr == (void *)-1, std::runtime_error)
       << "SwapDevice: mmap: "
       << std::string(strerror_r(errno, error_buf, error_buflen));
 
-#if defined(__ANDROID__)
-    std::memcpy(memory_ptr, ptr, len_offset.second);
-    munmap(ptr, len_offset.second);
-#else
-    NNTR_THROW_IF(ptr != memory_ptr, std::runtime_error)
-      << "SwapDevice: mmap: ptr!= memory_ptr";
-#endif
+    // MADVISE can be used to improve performance.
+    // madvise(ptr, len, MADV_SEQUENTIAL);
+
+    void *buf = static_cast<void *>(ptr + diff);
+
+    memcpy(memory_ptr, buf, len_offset.second);
+    munmap(ptr, len);
 
     ++offset_index;
     ++num_loaded_tensors;
+
+    // @todo : need to check at cache_loader & check multi thread execution
+    if (offset_index >= (int)weight_offset.size()) {
+      offset_index = 0;
+    }
+
     return memory_ptr;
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/swap_device.cpp
```diff
@@ -61,33 +61,42 @@ void *SwapDevice::getBuffer(off_t offset, size_t size, void *memory_ptr,
...
       << "SwapDevice: mmap: "
       << std::string(strerror_r(errno, error_buf, error_buflen));
-
     NNTR_THROW_IF(ptr != memory_ptr, std::runtime_error)
       << "SwapDevice: mmap: ptr!= memory_ptr";
 
     ++offset_index;
     ++num_loaded_tensors;
-    return ptr;
+    if (offset_index >= (int)weight_offset.size()) {
```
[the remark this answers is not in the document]
DonghakPark: That's right, i will leave comment & fix next PR

jijoongmoon (approved):
LGTM

dkjung (approved):
LGTM

baek2sm (approved):

nntrainer/tensor/swap_device.cpp
```diff
@@ -61,33 +61,40 @@ void *SwapDevice::getBuffer(off_t offset, size_t size, void *memory_ptr,
...
       << std::string(strerror_r(errno, error_buf, error_buflen));
 
-    NNTR_THROW_IF(ptr != memory_ptr, std::runtime_error)
-      << "SwapDevice: mmap: ptr!= memory_ptr";
+    // MADVISE can be used to improve performance.
+    // madvise(ptr, len, MADV_SEQUENTIAL);
+
+    void *buf = static_cast<void *>(ptr + diff);
+
+    memcpy(memory_ptr, buf, len_offset.second);
```
SeoHyungjun, line 86: As far as I know, the initial concept of FSU was to allocate mmap directly to a memory pool.
However, looking at the current code, the process of reading files through mmap and copying data using memcpy has been added.
Do you know why the method has changed?

SeoHyungjun (approved):

nntrainer/tensor/swap_device.cpp
DonghakPark: Yeah. Initially we target reuse allocate from memory pool ( now FSU case use ```allocateFSU()``` -> Please Check manager logic). but some reason we changed our logic to this **(  still aim to improve the structure and reuse memory in the end. )**

The method you mentioned refers to mapping a file to memory that has already been allocated in a memory pool with a fixed format using mmap. There are several restrictions ( For detail check this docs : https://man7.org/linux/man-pages/man2/mmap.2.html).
```        
 mmap() creates a new mapping in the virtual address space of the
       calling process.  The starting address for the new mapping is
       specified in addr.  The length argument specifies the length of
       the mapping (which must be greater than 0).

       If addr is NULL, then the kernel chooses the (page-aligned)
       address at which to create the mapping; this is the most portable
       method of creating a new mapping.  If addr is not NULL, then the
       kernel takes it as a hint about where to place the mapping; on
       Linux, the kernel will pick a nearby page boundary (but always
       above or equal to the value specified by
       /proc/sys/vm/mmap_min_addr) and attempt to create the mapping
       there.  If another mapping already exists there, the kernel picks
       a new address that may or may not depend on the hint.  The address
       of the new mapping is returned as the result of the call.

       The contents of a file mapping (as opposed to an anonymous
       mapping; see MAP_ANONYMOUS below), are initialized using length
       bytes starting at offset offset in the file (or other object)
       referred to by the file descriptor fd.  offset must be a multiple
       of the page size as returned by sysconf(_SC_PAGE_SIZE).

       After the mmap() call has returned, the file descriptor, fd, can
       be closed immediately without invalidating the mapping.

       The prot argument describes the desired memory protection of the
       mapping (and must not conflict with the open mode of the file).
       It is either PROT_NONE or the bitwise OR of one or more of the
       following flags:

MAP_FIXED
              Don't interpret addr as a hint: place the mapping at
              exactly that address.  addr must be suitably aligned: for
              most architectures a multiple of the page size is
              sufficient; however, some architectures may impose
              additional restrictions.  If the memory region specified by
              addr and length overlaps pages of any existing mapping(s),
              then the overlapped part of the existing mapping(s) will be
              discarded.  If the specified address cannot be used, mmap()
              will fail.

              Software that aspires to be portable should use the
              MAP_FIXED flag with care, keeping in mind that the exact
              layout of a process's memory mappings is allowed to change
              significantly between Linux versions, C library versions,
              and operating system releases.  Carefully read the
              discussion of this flag in NOTES!

```
- need to make length aligned to page size 
- need to make offset aligned to page size
- need aligned allocated memory ptr ( that's why ```allocateFSU()``` use aligend alloc )

We intend to support the FSU function on Linux, Android, and Windows, and we want to allocate memory using RPCmem_allocate on Android. -> make FSUallocate()

But there is still some problem

if weight size is less than page_size -> it will be fail -> we need to save weight file always greater than page_size. even if weight size is 100-> for support this we need to change all tensor's save method 

So, for now, we receive the size, check how far it is from the page size alignment, and use the difference to perform memcpy.

if you have any more question. always welcome to come and discuss

SeoHyungjun: Thank you for understanding.
So, if we just align the save and read operations to match the page size, will work correctly?

DonghakPark: Yes. we need to support all tensor save with aligned size (it will change depend on System-information)
for example, one layer has one weight and it's data is look like ```{1, 2, 3, 4} ``` then we need to fill padding (page_size - weight size) = ex)4096-4 : 4092!!!

so. we need to discuss this issue, is this correct way? memcpy consume that much cost ?

SeoHyungjun: It seems like we need to conduct a performance comparison.

DonghakPark: Yes. When implement new profile function, Please check that with SimpleFC Application!!

SeoHyungjun: memcpy is known to be well optimized. 
Let's discuss any additional changes that might be needed while using the current code.

PR #3108, opened 2025-04-10
[FSU] Add readFSU to createFSU Tensor (BCQ)

DonghakPark:
In This PR
----
The current method of reading FSU weights is as follows (Simple Description):
1. In read, if FSU, it does not read during the initial load process.
2. During inference, depending on the execution order and look-ahead, the weight file is read through the FSU function at the moment it is needed and loaded.

There is a problem with this method. For logic like BCQ that reads information during loading and uses it to create tensors, the information needs to be known before memory allocation, so there is need to pass only that information. so I created a new readFSU function for cases in the future where settings for FSU are needed for various tensors.

Add readFSU function & Change BCQ Create Signature
- prepare Tensor that has other information (except Weight)
- Change bcq tensor create signature

**Self evaluation:**
4. Run test:	 [X]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -401,15 +401,11 @@ void BCQTensor::createBCQW() {
   /// found with various values according to the usage environment.
   size_t hidden_tile_size = 32;
 
-  // bcq_weight = std::make_shared<BiQGEMM::BCQW>(
-  //   (uint32_t *)getData(), (float *)getScale(), width(), height(),
-  //   number_of_cluster, qbit_of_clusters, size_of_clusters, hidden_tile_size);
-
   bcq_weight = std::make_shared<BiQGEMM::BCQW>(
     (uint32_t *)(data->getAddr<uint32_t>()),
     (float *)((uint32_t *)(data->getAddr<uint32_t>()) + size()), width(),
     height(), number_of_cluster, qbit_of_clusters, size_of_clusters,
-    hidden_tile_size, true);
+    hidden_tile_size);
 }
 
 } // namespace nntrainer
--- a/nntrainer/tensor/bcq_tensor.cpp
+++ b/nntrainer/tensor/bcq_tensor.cpp
@@ -255,6 +255,8 @@ void BCQTensor::save(std::ostream &file) {
   putData();
 }
 
+void BCQTensor::readFSU() { createBCQW(); }
+
 void BCQTensor::read(std::ifstream &file) {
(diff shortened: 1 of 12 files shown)


nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -399,9 +401,15 @@ void BCQTensor::createBCQW() {
...
 
+  // bcq_weight = std::make_shared<BiQGEMM::BCQW>(
+  //   (uint32_t *)getData(), (float *)getScale(), width(), height(),
+  //   number_of_cluster, qbit_of_clusters, size_of_clusters, hidden_tile_size);
+
   bcq_weight = std::make_shared<BiQGEMM::BCQW>(
-    (uint32_t *)getData(), (float *)getScale(), width(), height(),
-    number_of_cluster, qbit_of_clusters, size_of_clusters, hidden_tile_size);
+    (uint32_t *)(data->getAddr<uint32_t>()),
+    (float *)((uint32_t *)(data->getAddr<uint32_t>()) + size()), width(),
```
jijoongmoon, line 410: (float*) , is this mean only bcq 32 is supported for  now?

jijoongmoon (approved):
LGTM

EunjuYang (approved):
LGTM

nntrainer/tensor/bcq_tensor.cpp
```diff
@@ -399,9 +401,15 @@ void BCQTensor::createBCQW() {
   /// found with various values according to the usage environment.
   size_t hidden_tile_size = 32;
 
+  // bcq_weight = std::make_shared<BiQGEMM::BCQW>(
+  //   (uint32_t *)getData(), (float *)getScale(), width(), height(),
+  //   number_of_cluster, qbit_of_clusters, size_of_clusters, hidden_tile_size);
```
djeong20, line 406:
```suggestion
  // bcq_weight = std::make_shared<BiQGEMM::BCQW>(
  //   (uint32_t *)getData(), (float *)getScale(), width(), height(),
  //   number_of_cluster, qbit_of_clusters, size_of_clusters, hidden_tile_size);
```

let's remove this line

djeong20 (approved):
LGTM!

nntrainer/tensor/bcq_tensor.cpp
DonghakPark: Okay. i will remove

skykongkong8 (approved):

PR #3112, opened 2025-04-11
Build gglm cpu

gkisalapl:


subprojects/ggml.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+url = https://github.com/ggml-org/ggml.git
+directory=ggml
+revision = HEAD
```
dkjung, line 4: How about fixing to a specific version?

gkisalapl: Unfortunately ggml project does not have any tags

PR #3120, opened 2025-04-16
[FSU] update offset calc & Window page_size

DonghakPark:
[FSU] update offset calc & Window page_size

- Update offset calc with getmemoryBytes()
- Update Window page_size to  sysInfo.dwAllocationGranularity;

ref : https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile

[in] dwFileOffsetLow

A low-order DWORD of the file offset where the view is to begin. The combination of the high and low offsets must specify an offset within the file mapping. They must also match the virtual memory allocation granularity of the system. That is, the offset must be a multiple of the VirtualAlloc allocation granularity. To obtain the VirtualAlloc memory allocation granularity of the system, use the [GetSystemInfo](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsysteminfo) function, which fills in the members of a [SYSTEM_INFO](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/ns-sysinfoapi-system_info) structure.

[in] dwNumberOfBytesToMap

The number of bytes of a file mapping to map to the view. All bytes must be within the maximum size specified by [CreateFileMapping](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-createfilemappinga). If this parameter is 0 (zero), the mapping extends from the specified offset to the end of the file mapping.

**Self evaluation:**

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -340,11 +340,21 @@ NeuralNetwork::~NeuralNetwork() {
   }
 }
 
+void NeuralNetwork::setupFSU() { model_graph.setupFSU(); }
+
 /**
  * @brief     forward propagation using layers object which has layer
  */
 sharedConstTensors NeuralNetwork::forwarding(
   bool training, std::function<bool(void *userdata)> stop_cb, void *userdata) {
+
+  auto fsu_enable = std::get<props::Fsu>(model_flex_props);
+  if (fsu_enable) {
+    auto look_ahead = std::get<props::FsuLookahead>(model_flex_props);
+    for (unsigned int order = 0; order < look_ahead; order++) {
+      model_graph.LoadTensors(order, look_ahead);
+    }
+  }
 
   std::function<void(std::shared_ptr<LayerNode>, bool)> forwarding_op =
     [this, stop_cb, userdata](std::shared_ptr<LayerNode> node,
@@ -385,10 +395,10 @@ sharedConstTensors NeuralNetwork::forwarding(
                and if it is loaded, then run forwarding. Every time it finishes
                the forwarding, ask load tensors for next n layers.
       **/
-
-      model_graph.LoadTensors(f);
-      model_graph.checkLoadComplete(f);
+      auto look_ahead = std::get<props::FsuLookahead>(model_flex_props);
+      model_graph.checkFsuLoadComplete(f);
       node->forwarding(training);
+      model_graph.LoadTensors(f + look_ahead, look_ahead);
     }
   };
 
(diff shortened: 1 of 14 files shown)


myungjoo (approved):

DonghakPark:
Please ignore ```yocto devtool / Build``` for now 
it's come from retirement of runner : 
```
[Build with Yocto devtool on Ubuntu](https://github.com/nnstreamer/nntrainer/actions/runs/14488317606/job/40638687415)
This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
```

nntrainer/models/neuralnet.cpp
```diff
@@ -340,12 +340,26 @@ NeuralNetwork::~NeuralNetwork() {
   }
 }
 
+void NeuralNetwork::setupFSU() {
+  auto fsu_enalbe = std::get<props::Fsu>(model_flex_props);
```
dkjung, line 344: typo: `fsu_enable`

nntrainer/models/neuralnet.cpp
```diff
@@ -385,10 +399,10 @@ sharedConstTensors NeuralNetwork::forwarding(
                and if it is loaded, then run forwarding. Every time it finishes
                the forwarding, ask load tensors for next n layers.
       **/
-
-      model_graph.LoadTensors(f);
-      model_graph.checkLoadComplete(f);
+      auto look_ahead = std::get<props::FsuLookahead>(model_flex_props);
+      model_graph.checkFsuLoadComplete(f);
```
dkjung, line 403: Does this function wait until the load is complete?

nntrainer/tensor/manager.cpp
```diff
@@ -598,7 +598,12 @@ Manager::requestInputs(const GraphNode &node,
   using RT = TensorSpecV2::RequestType;
 
   TensorSpecV2 var_common_spec, grad_common_spec;
-  var_common_spec.ls = TensorLifespan::FORWARD_GRAD_LIFESPAN;
+  bool is_train_mode = (exec_mode == ExecutionMode::TRAIN) ? true : false;
```
dkjung, line 601:
```c++
bool is_train_mode = exec_mode == ExecutionMode::TRAIN;
```

nntrainer/models/neuralnet.cpp
DonghakPark: Fixed! Thank you for Review

nntrainer/models/neuralnet.cpp
DonghakPark: Yes. if load is not completed. but we wish already loaded by async load

nntrainer/tensor/manager.cpp
DonghakPark: FIxed!! Thank you for Review

PR #3132, opened 2025-04-23
[Tensor] Add Q4_K Tensor class @open sesame (2025-05-08 15:17)

djeong20:
This PR introduces a new type of Tensor class, Q4_K_Tensor.
This tensor class aims to store Q4_K Quantized data types.
The initial implementation is targeted only at Q4_Kx8 types.
Please note that only data storage and dot operations will be valid for this class.

**Self-evaluation:**


djeong20:
@skykongkong8 Please check if anything is missing or if the implementation is incorrect.

nntrainer/tensor/q4_k_tensor.cpp
```diff
@@ -0,0 +1,78 @@
...
+  Uint4QTensor(name_, fm, QScheme::Q4_Kx8) {}
+
+Q4_K_Tensor::Q4_K_Tensor(const TensorDim &d, bool alloc_now, Initializer init,
+                         std::string name, QScheme qscheme_) :
+  Uint4QTensor(d, alloc_now, init, name, qscheme_) {
+  NNTR_THROW_IF(d.batch() != 1 || d.channel() != 1 ||
+                  (d.height() * d.width()) % 256 != 0,
+                std::invalid_argument)
+    << "Q4_K_Tensor must be 2 dimensional tensor with batch size 1 and "
+       "height * width must be divisible by 256";
```
skykongkong8, line 27: Think I made you confused a little bit:
```suggestion
                  (d.height() % 256 != 0 || d.width()% 256 != 0),
                std::invalid_argument)
    << "Q4_K_Tensor must be 2 dimensional tensor with batch size 1 and "
       "height * width must be divisible by 256";
```

nntrainer/tensor/q4_k_tensor.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void Q4_K_Tensor::copy(const void *buf) {
+  NNTR_THROW_IF(!contiguous, std::invalid_argument)
+    << getName() << " is not contiguous, cannot copy.";
+
+  if (buf == getData()) {
+    return;
+  }
+  // copy tensor data
+  scopy(size(), (uint8_t *)buf, 1, (uint8_t *)getData(), 1);
```
skykongkong8, line 75: I don't think this will copy internal data properly, because the Tensor consists of:
```C++
struct block_q4_Kx8{
  int16_t d[8];     // super-block scale for quantized scales
  int16_t dmin[8];  // super-block scale for quantized mins
  uint8_t scales[96]; // scales and mins, quantized with 6 bits
  uint8_t qs[1024];   // 4--bit quants
};
```
I think in ggml, it is being copied like [this](https://github.com/ggml-org/llama.cpp/blob/658987cfc9d752dca7758987390d5fb1a7a0a54a/ggml/src/ggml-cpu/ops.cpp#L1011)

nntrainer/tensor/q4_k_tensor.cpp
djeong20: thanks!

djeong20: To clarify, is there one block_q4_Kx8 for each 256 elements or 256 x 256 elements?

nntrainer/tensor/q4_k_tensor.cpp
djeong20: Isn't the total size `sizeof(uint8_t) * 114 * height() * width() / 256`?
`sizeof(uint8_t) * 114` is `sizeof(block_q4_Kx8)`

nntrainer/tensor/q4_k_tensor.h
```diff
@@ -0,0 +1,104 @@
...
+   * @copydoc Tensor::isValid()
+   */
+  bool isValid() const override { return true; }
+
+}; // class Q4_K_Tensor
+
+} // namespace nntrainer
+
+#endif /* __cplusplus */
+#endif /* __Q4_K_TENSOR_H__ */
```
EunjuYang, line 104: Please add a new line at the end of the file

nntrainer/tensor/q4_k_tensor.cpp
skykongkong8: Yup, but I am afraid that `scopy` function might fail to sucessfully copy `int16_t` data internally. Is it connected with memcpy?

nntrainer/tensor/q4_k_tensor.cpp
skykongkong8: Hope this solves your question:

- 256 float values - q4_K quantization -> 1 block_q4_K
- 8 block_q4_k - repack -> block_q4_Kx8

Thus one block_q4_Kx8 represent 8*256 original float values, AFAIK

nntrainer/tensor/tensor.cpp
```diff
@@ -180,9 +181,14 @@ Tensor::Tensor(const TensorDim &d, bool alloc_now, Initializer init,
...
-      std::shared_ptr<Uint4QTensor>(new Uint4QTensor(d, alloc_now, init, name),
-                                    std::default_delete<Uint4QTensor>());
+    if (qscheme != QScheme::Q4_Kx8)
+      itensor = std::shared_ptr<Uint4QTensor>(
+        new Uint4QTensor(d, alloc_now, init, name),
+        std::default_delete<Uint4QTensor>());
+    else
+      itensor =
+        std::shared_ptr<Q4_K_Tensor>(new Q4_K_Tensor(d, alloc_now, init, name),
+                                     std::default_delete<Q4_K_Tensor>());
```
EunjuYang, line 191: Can I add `Q4_K` as type? like 
https://github.com/nnstreamer/nntrainer/blob/7e29114159947f7f63c693b9308a35a57a70770b/api/ccapi/include/tensor_dim.h#L55-L60

In #3135, I introduced Q4_K as a new data type and updated the tensor creation logic accordingly. Could you verify if this change aligns with the tensor-side philosophy?

nntrainer/tensor/q4_k_tensor.cpp
djeong20: Thanks for the clarification! I get it now. However, I don't understand why both height and width should be divisible by 256. If one block of q4x8 represents 2048 float values, shouldn't the tensor size be divisible by 2048?

djeong20: Thanks for the clarification!

nntrainer/tensor/tensor.cpp
djeong20: As we discussed online, let's proceed with the current implementation for now, and later we can explore ways to improve it.

nntrainer/tensor/q4_k_tensor.h
```diff
@@ -0,0 +1,108 @@
...
+ * This struct is not for use, only for reference.
+ */
+struct block_q4_K {
+  int16_t d[1];       // super-block scale for quantized scales
+  int16_t dmin[1];    // super-block scale for quantized mins
+  uint8_t scales[12]; // scales and mins, quantized with 6 bits
+  uint8_t qs[128];    // 4--bit quants
+};
+
+#define Q4_K_SIZE 114    // sizeof(block_q4_K)
```
EunjuYang, line 32: Do you mean 144?
```suggestion
#define Q4_K_SIZE 144    // sizeof(block_q4_K)
```

djeong20: Thanks :)

EunjuYang (commented):
Is it okay to directly inherit Uint4QTensor's save and load in Q4K_Tensor?

EunjuYang (approved):
LGTM 🚀

EunjuYang:
Please check the Tizen CI issue.

jijoongmoon (approved):
LGTM

PR #3146, opened 2025-04-30
Bugfix: GGML Build Error in NNTrainer on Ubuntu 20.04

djeong20:
This pull request addresses an error that occurred during the GGML build in NNTrainer.
The error reported unsupported instructions from AVX512, which happens exclusively on Ubuntu 20.04 with assembler version 2.34.
This change restricts the use of any AVX-512 instructions, allowing only up to AVX2 level instructions.

**Self-evaluation:**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -473,6 +473,22 @@ if get_option('enable-ggml')
   ggml_options.add_cmake_defines({'GGML_BUILD_EXAMPLES': false})
   ggml_options.add_cmake_defines({'CMAKE_INSTALL_LIBDIR': nntrainer_libdir / 'arm64-v8a'})
   ggml_options.append_compile_args('c', ['-Wno-error=maybe-uninitialized', '-Wno-error=pointer-to-int-cast'])
+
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -468,6 +468,22 @@ if get_option('enable-ggml')
   ggml_options.add_cmake_defines({'BUILD_SHARED_LIBS': host_machine.system() != 'windows'})
   ggml_options.add_cmake_defines({'GGML_BUILD_TESTS': false})
   ggml_options.add_cmake_defines({'GGML_BUILD_EXAMPLES': false})
+
+  if arch == 'x86' or arch == 'x86_64'
```
myungjoo, line 472: x86 is 32bit x86 and x86_64 is 64bit x86.
Don't add ```-march=x86-64``` for x86.

djeong20: Thank you! It has now been fixed.

EunjuYang (approved):

skykongkong8 (approved):

PR #3127, opened 2025-04-17
[WIP] AVX2 SWIGLU

piotrrak:
---
# [Template] PR Description

In general, github system automatically copies your commit message for your convenience.
Please remove unused part of the template after writing your own PR description with this template.
```bash
$ git commit -s filename1 filename2 ... [enter]

Summarize changes in around 50 characters or less

More detailed explanatory text, if necessary. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical;
various tools like `log`, `shortlog` and `rebase` can get confused 
if you run the two together.

Further paragraphs come after blank lines.

**Changes proposed in this PR:**
- Bullet points are okay, too
- Typically a hyphen or asterisk is used for the bullet, preceded
  by a single space, with blank lines in between, but conventions vary here.

Resolves: #123
See also: #456, #789

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [*]Skipped
2. Run test: [ ]Passed [ ]Failed [*]Skipped

**How to evaluate:**
1. Describe how to evaluate in order to be reproduced by reviewer(s).

Add signed-off message automatically by running **$git commit -s ...** command.

$ git push origin <your_branch_name>
```

PR #3153, opened 2025-05-07
[x86_64] AVX approximation of swiglu

piotrrak:
This uses custom expf() approximation with maximal error <= 5e-6. Speedup of 9x-12x. Differs by compiler 8x wrt. scalar should be expected.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -12,12 +12,295 @@
  *
  */
 
-#include <avx2_impl.h>
+#include "avx2_impl.h"
+#include <array>
+#if __has_include(<bit>)
+#include <bit>
+#endif
 #include <cassert>
 #include <cmath>
 #include <cstdint>
+#include <cstring>
 #include <immintrin.h>
 #include <limits>
+#if __has_include(<numbers>)
+#include <numbers>
+#endif
+#include <type_traits>
+#if __has_include(<version>)
+#include <version>
+#endif
+
+#if !defined(__has_constexpr_builtin)
+#define __has_constexpr_builtin(x) (0)
+#endif
+
+#if !defined(__has_cpp_attribute)
+#define __has_cpp_attribute(x) (0)
+#endif
+
+// VECTORCALL calling-conv (default for x86_64-linux-gnu)
+#if _MSC_VER >= 1700
+#define _nnt_CC_VECTORCALL __vectorcall
+#else
+#define _nnt_CC_VECTORCALL
+#endif
+
+// Flatten attribute
+#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::flatten)
+#define _nnt_ATTR_FLATTEN [[msvc::flatten]]
+#elif __has_cpp_attribute(gnu::flatten)
+// clang, g++
+#define _nnt_ATTR_FLATTEN [[gnu::flatten]]
+#else
+#define _nnt_ATTR_FLATTEN
+#endif
+
+#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::noinline)
+#define _nnt_ATTR_NOINLINE [[msvc::noinline]]
+#elif __has_cpp_attribute(gnu::flatten)
+// clang, g++
+#define _nnt_ATTR_NOINLINE [[gnu::noinline]]
+#else
+#define _nnt_ATTR_NOINLINE
+#endif
+
+#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::forceinline)
+#define _nnt_ATTR_ALWAYS_INLINE [[msvc::forceinline]]
+#elif __has_cpp_attribute(gnu::always_inline)
+#define _nnt_ATTR_ALWAYS_INLINE [[gnu::always_inline]]
+#endif
+
+#if __has_cpp_attribute(unikely)
+#define UNLIKELY [[unlikely]]
+#else
+#define UNLIKELY
+#endif
+
+#if !defined(_MSC_VER) && !defined(__clang__)
+#pragma GCC diagnostic ignored "-Wattributes"
+#endif
+
+namespace {
+
+template <typename To_, typename From_>
+constexpr inline bool concept17_BinaryCastable =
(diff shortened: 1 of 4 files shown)


skykongkong8 (commented):
Implementation itself looks fine, but how did you evaluate the acceleration profit?

piotrrak:
> Implementation itself looks fine, but how did you evaluate the acceleration profit?

Yes, I've profiled it. As mentioned in PR its 9-12x faster than scalar depending on compiler. More than 800% speed up should be expected in any case. It should do bit better on newer cores. Due to sub_ps/add_ps latency decresed to 2 from cycles since Alder Lake/Saphaire Rapids - that's the reason round is done by bithack. Though it is not big win.
It's not fair comparison since it calculates with limited precision (approaching that required for FP16). It also does away with handling INF for x above ~88 in exp(x) how would real libmath exp be implemented.
This approximation is good enough, to be used in real models. Until proven otherwise. There are some tricks possible to increase the precision. It is about 2-3x faster than my previous attempt: https://github.com/nnstreamer/nntrainer/pull/3127#issuecomment-2827424194

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -12,12 +12,298 @@
...
+inline constexpr auto exp2_table_type_v = Exp2Table<N_, Ty_>::calculate();
+
+// Scalar version of expf() approximation with 3-th deg polynominal of
+// fractional part
+//
+// The error with regards to std::expf less than 5e-6
+// It is valid in range [-87, +88.37) - not handling +INF, NaN etc.
+// The function domain is clamped to valid function range.
+//
+// The strategy picked is to approximate exp as 2^K*2*F
```
piotrrak, line 181: 2^K*2^F

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -12,12 +12,298 @@
...
+
+  constexpr static uint32_t N_MASK = uint32_t(N_ - 1U);
+  constexpr static auto EXP2_TBL = exp2_table_v<N_>;
+  constexpr static unsigned MANTISSA_BITS =
+    std::numeric_limits<float>::digits - 1;
+
+  // Ensure arg in range [exp_arg_min, exp_arg_max]
+  xs = _mm256_max_ps(xs, _mm256_set1_ps(EXP_ARG_MIN));
+  // Would clamp to EXP_ARG_MAX but we move it after multiplication for IPC
+  xs = _mm256_min_ps(xs, _mm256_set1_ps(EXP_ARG_MAX));
```
piotrrak, line 233: Should be commented out, clamp to upper bound is handled below after multiplication.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -12,12 +12,298 @@
...
+  }
+};
+
+#endif
+
+template <unsigned N_> // requires PowerOfTwo<N_>
+inline constexpr auto exp2_table_v = Exp2Table<N_>::calculate();
+
+template <unsigned N_, typename Ty_> // requires PowerOfTwo<N_>
+inline constexpr auto exp2_table_type_v = Exp2Table<N_, Ty_>::calculate();
```
piotrrak, line 172: Unused.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -12,12 +12,298 @@
...
+                                 0x3f75fed7U, 0x3f7504f3U, 0x3f75672aU,
+                                 0x3f7744fdU, 0x3f7ac0c7U};
+    return t;
+  }
+};
+
+#endif
+
+template <unsigned N_> // requires PowerOfTwo<N_>
+inline constexpr auto exp2_table_v = Exp2Table<N_>::calculate();
```
piotrrak, line 169: requires aligned_array or `alignas(16)`

piotrrak: https://gist.github.com/piotrrak/1e5c5026358c21d0038495535295c740 aligned_array implementation, but it is c+20, so alignas for now.

skykongkong8 (approved):

baek2sm (approved):

gkisalapl:
3x speedup on Windows (msvc)
0.047s -> 0.016s

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -12,12 +12,295 @@
...
+    return t;
+  }
+};
+
+#endif
+
+template <unsigned N_> // requires PowerOfTwo<N_>
+alignas(__m256) inline constexpr auto exp2_table_v = Exp2Table<N_>::calculate();
+
+// Scalar version of expf() approximation with 3-th deg polynominal of
```
gkisalapl, line 171: Could You please add link to publication which was base for this implementation ?

gkisalapl:
> 3x speedup on Windows (msvc) 0.047s -> 0.016s

3x speedup on Windows (clang) 0.092s -> 0.031s

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
piotrrak: Guess: https://nic.schraudolph.org/pubs/Schraudolph99.pdf with classic Cody Waite range reduction.
"Software Manual for the Elementary Functions"  William J. Cody, Jr. and William Waite
Some ideas from https://research.ibm.com/publications/fast-exponential-computation-on-simd-architectures

piotrrak:
> > 3x speedup on Windows (msvc) 0.047s -> 0.016s
> 
> 3x speedup on Windows (clang) 0.092s -> 0.031s

Guess that did worse than previous implementation on windows.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
gkisalapl: > e ideas fro

Please add it as comment to code.

gkisalapl (approved):

PR #3155, opened 2025-05-09
[FSU] Optimize FSU Feature

DonghakPark:
In this PR
---

There are several Point for improvement in the current FSU Logic. the following improvements have been made:

1. Change the InActivate timing - now, inActivate is being performed within forwarding after each layer execution, but this does not make sense within a single inference. Therefore, we have changed it to inActivate after one complete inference is finished.
2. early exit in Task Cancle
3. Change map to unordered_map - we don't need sorted list for task executor

**Self evaluation:**
5. Run test:	 [X]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -425,7 +425,7 @@ sharedConstTensors NeuralNetwork::forwarding(sharedConstTensors input,
   model_graph.setInputsLabels(input, label);
   auto fsu_enable = std::get<props::Fsu>(model_flex_props);
   if (fsu_enable) {
-    model_graph.inActive(0);
+    model_graph.inActiveAll();
   }
   return forwarding(training);
 }
@@ -1021,7 +1021,7 @@ sharedConstTensors NeuralNetwork::incremental_inference(
   model_graph.setInputsLabels({}, {});
   auto fsu_enable = std::get<props::Fsu>(model_flex_props);
   if (fsu_enable) {
-    model_graph.inActive(0);
+    model_graph.inActiveAll();
   }
   return out;
 }
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -395,7 +395,7 @@ sharedConstTensors NeuralNetwork::forwarding(
       **/
       model_graph.checkLoadComplete(f);
       node->forwarding(training);
-      model_graph.inActive(f);
+      // model_graph.inActive(f);
       model_graph.LoadTensors(f + lookahead);
     }
   };
@@ -424,7 +424,10 @@ sharedConstTensors NeuralNetwork::forwarding(sharedConstTensors input,
     << " target_batch: " << current_batch;
 
   model_graph.setInputsLabels(input, label);
-
+  auto fsu_enable = std::get<props::Fsu>(model_flex_props);
+  if (fsu_enable) {
+    model_graph.inActive(0);
+  }
   return forwarding(training);
 }
 
@@ -454,7 +457,7 @@ sharedConstTensors NeuralNetwork::incremental_forwarding(
     } else {
       model_graph.checkLoadComplete(f);
       node->incremental_forwarding(from, to, training);
-      model_graph.inActive(f);
+      // model_graph.inActive(f);
       model_graph.LoadTensors(f + lookahead);
     }
   };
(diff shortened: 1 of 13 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -395,7 +395,7 @@ sharedConstTensors NeuralNetwork::forwarding(
       **/
       model_graph.checkLoadComplete(f);
       node->forwarding(training);
-      model_graph.inActive(f);
+      // model_graph.inActive(f);
```
baek2sm, line 398: Wouldn't it be better just to delete it?

baek2sm (approved):
LGTM

nntrainer/models/neuralnet.cpp
```diff
@@ -994,7 +997,10 @@ sharedConstTensors NeuralNetwork::incremental_inference(
   /** @todo: deallocate tensor after incremental inference **/
   /** Clear the set inputs and labels */
   model_graph.setInputsLabels({}, {});
-
+  auto fsu_enable = std::get<props::Fsu>(model_flex_props);
+  if (fsu_enable) {
+    model_graph.inActive(0);
```
SeoHyungjun, line 1002: In the previous code, variable f was used in the inActive. 
However, in the modified code, only the value 0 is used.
What is the reason behind this change?

nntrainer/tensor/cache_loader.cpp
```diff
@@ -195,25 +195,30 @@ int CacheLoader::cancelAsync(int id) {
...
+  for (auto id : ids) {
+    auto exec_id = id.second;
+    for (auto &id : exec_id) {
+      std::shared_ptr<CacheElem> elem = pool->getCacheElem(id);
+      int load_task_id = elem->getLoadTaskID();
+      if (load_task_id >= 0) {
+        load_task_executor->releaseTask(load_task_id);
+        elem->setLoadTaskID(-1);
+        states[id] = LoadState::Unloading;
+      }
```
SeoHyungjun, line 211: In the current code, the variable name "id" is duplicated.
How about changing the variable names?

```suggestion
  for (auto id : ids) {
    auto exec_ids = id.second;
    for (auto &exec_id: exec_ids) {
      std::shared_ptr<CacheElem> elem = pool->getCacheElem(id);
      int load_task_id = elem->getLoadTaskID();
      if (load_task_id >= 0) {
        load_task_executor->releaseTask(load_task_id);
        elem->setLoadTaskID(-1);
        states[exec_id] = LoadState::Unloading;
      }
```

SeoHyungjun (approved):

nntrainer/models/neuralnet.cpp
DonghakPark: Done Thank you for Review

nntrainer/tensor/cache_loader.cpp
DonghakPark: your right !!! thank you for review

nntrainer/models/neuralnet.cpp
DonghakPark: In the previous logic, elements were inactivated according to the execution order. However, given the current operation method of FSU, it is more efficient to handle this once per inference rather than processing it layer by layer. This change was made for efficiency.

The part you mentioned is a remnant of the previous method where it was handled order by order. 
To prevent confusion, I will correct it immediately.

DonghakPark: Signature Change Done! 👍

nntrainer/tensor/cache_loader.cpp
```diff
@@ -194,26 +194,31 @@ int CacheLoader::cancelAsync(int id) {
...
+        load_task_executor->releaseTask(load_task_id);
+        elem->setLoadTaskID(-1);
+        states[element_id] = LoadState::Unloading;
+      }
+      elem->inActive();
     }
-    actives.remove(elem);
-    elem->inActive();
   }
+  // std::set<unsigned int> exec_id = pool->getExecIDs(order);
```
djeong20, line 215:
```suggestion
```

djeong20 (approved):
LGTM!

myungjoo:
@DonghakPark Please rebase. I guess the test fail will disappear.

skykongkong8 (approved):

PR #3160, opened 2025-05-12
[TensorDim] Allow setting dimension values to 0 in TensorDim

baek2sm:
enables dimension values to be set to 0 in `TensorDim`

- It can be used for index-based slicing property for `Slicing layer`.
- It allows handling cases where dimensions are dynamically coded in
other libraries (e.g., -1), but the dimensions are actually static
values. Then we can set the dynamic dimension values to 0, and then its
dimension is resolved during the finalize step by confirming the input
layer's output tensor dimensions.


songgot (approved):
LGTM

baek2sm:
I'll think of a better way

PR #3162, opened 2025-05-14
[Load]Optimize load with Async

### LINKED ISSUE #3159: Load() Optimize
Optimize Load()

- [x] Use Thread for Load()

DonghakPark:
In This PR 
----
This PR improves the loading logic so that model weight files are read asynchronously and in parallel.
By calculating each weight’s file offset, multiple threads can simultaneously read weights, and to enable this, a start_offset parameter was added to read function interfaces, ensuring that file reading uses the right offset.

Loading performance improved (for 16GB, about 4500ms → 1600ms).

``` 
Linux
CPU info : 11th i7-11700
memory Performance 28693.32 MB/s

Disk Random 16.0 Read 240.06 MB/s 8.0
Disk Sequential 64.0 Read 835.56 MB/s 8.3
Disk Sequential 64.0 Write 287.56 MB/s 7.7

load : 16GB bin file, 4543 ms --> 1446 ms,
```

 The implementation uses futures and async so that multiple layers or weights are loaded at the same time, and read function signatures across tensor and layer classes were modified to support this. Also, utility functions were updated to handle file offsets according to these changes.

Resolves
- #3159 

close #3159 

**Self evaluation:**

### DIFF
--- a/nntrainer/layers/bn_layer.cpp
+++ b/nntrainer/layers/bn_layer.cpp
@@ -423,7 +423,7 @@ void BatchNormalizationLayer::read(std::ifstream &file,
                                    ml::train::ExecutionMode mode,
                                    bool trainable,
                                    TensorDim::DataType definedWeightDataType,
-                                   bool fsu) {
+                                   bool fsu, size_t start_offset) {
   if (opt_var) {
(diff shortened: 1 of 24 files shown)


jijoongmoon (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #3170, opened 2025-05-15
[ggml] Add Q6_K dot product operation

mnalewaj:
WIP - Do not merge

Add Q6_K dot product operation

---
# [Template] PR Description

In general, github system automatically copies your commit message for your convenience.
Please remove unused part of the template after writing your own PR description with this template.
```bash
$ git commit -s filename1 filename2 ... [enter]

Summarize changes in around 50 characters or less

More detailed explanatory text, if necessary. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical;
various tools like `log`, `shortlog` and `rebase` can get confused 
if you run the two together.

Further paragraphs come after blank lines.

- Bullet points are okay, too
- Typically a hyphen or asterisk is used for the bullet, preceded
  by a single space, with blank lines in between, but conventions vary here.

Resolves: #123
See also: #456, #789

**Self evaluation:**

**How to evaluate:**

Add signed-off message automatically by running **$git commit -s ...** command.

$ git push origin <your_branch_name>
```


test/unittest/unittest_nntrainer_cpu_backend.cpp
```diff
@@ -0,0 +1,590 @@
...
+      // std::cout << "[INFO] dot_q6_K_q8_K : " << dt.count() << " ns " <<
+      // std::endl;
+      if (i >= 0) { // skip the first run
+        q6_k_time += dt;
+      }
+    }
+    // printf("res: %f  res: %f, diff: %f\n", result, ref_result, result -
+    // ref_result);
+    EXPECT_NEAR(result, ref_result, 0.25 * K / 256);
+  }
```
skykongkong8, line 525: I think we have a different point of view here.
I believe that 
```C++
ntrainer::quantize_row_q8_K(activation.data(), v_q8_activation.data(), K);
```
should be placed inside of 
```C++
nntrainer::dot_q6_K_q8_K(K, q6_K_weight.data(), v_q8_activation.data());
```
like
```C++
 nntrainer::gemm_q4_K(M, N, K, activations, K, (void *)repacked_qWeight.data(),
                       N, dst.data(), N);
```

compute kernel function's activation input should be `float *`, not `void *`, meaning, not q8K. It should be runtime-quantized.

skykongkong8 (commented):
I could reproduce a fair result with q6K dot kernel with this PR! Thanks for you hard work!
Followingly, please check my review below:
Moreover, like we discussed in the different channel, could you please work for computing GEMM using this dot function?
It won't be that complex I believe, somethine like:
```C++
void matmul(float* C, float* A, float* B, unsigned int M, unsigned int N, unsigned int K) {  
    // Matrix multiplication using sdot  
    // Assumptions:  
    // - Matrix A[M][K] :
    // - Matrix B[K][N] : (in our case, this is B[N][K] but just for example)
    // - Output Matrix C[M][N]  

    for(unsigned int i = 0; i < M; ++i) {  
        for(unsigned int j = 0; j < N; ++j) {  
            // Calculate C[i][j] = dot product between  
            // i-th row of A and j-th column of B  
            C[i * N + j] = sdot(  
                &A[i * K],      // Start address of i-th row  
                &B[j * K],      // Start address of j-th column  
                K               // Vector length  
            );  
        }  
    }  
}  
```
will do. Or you can refer to:
`ggml_compute_forward_mul_mat_one_chunk()`

skykongkong8:
Moreover, we need to introduce q6K to float dequantize function as well.

PR #3175, opened 2025-05-19
[FSU] Bugfix on FSU-OFF & Look_ahead Given Case

DonghakPark:
In This PR
---- 
Fix FSU Off & fsu_look_ahead given case -> try to load weight from disk,flash bug

For accurate operation, it is necessary to ensure that the FSU option is enabled and load the data up to the Look Ahead value. However, in the current code, even if FSU is turned off, if the user has set a Look_Ahead value, it attempts to load it. Although the actual loading action is not performed because the Cache_loader is not created, code handling this logic is also required at the Neuralnet level.

**Self evaluation:**


jijoongmoon (approved):
LGTM

PR #3185, opened 2025-05-20
Avoid rebuilding OpenBLAS

gkisalapl:
On systems where there no OpenBLAS provided we're trying to compile own version and it's running every time we make modifications in meson.build With this PR we can reuse previous build of OpenBLAS

**Self-evaluation:**


jijoongmoon (approved):
LGTM. But Could you make the script with shell (eg. bash) for the various platform support?

gkisalapl:
> LGTM. But Could you make the script with shell (eg. bash) for the various platform support?

I think that having the same script for multiple platforms (python) is advantage here. Why would you like to have separate bash / powershell scripts for each platform ?

jijoongmoon:
> > LGTM. But Could you make the script with shell (eg. bash) for the various platform support?
> 
> I think that having the same script for multiple platforms (python) is advantage here. Why would you like to have separate bash / powershell scripts for each platform ?

It does not support Android build. Script includes just cmake command, so I think you can also use meson run_command, so that we can compile openblas when we need for all the platform we support.

Or, you can propose alternatives to compile the openblas for all the platfom we support and make sure it is compiled properly.

gkisalapl:
> > > LGTM. But Could you make the script with shell (eg. bash) for the various platform support?
> > 
> > 
> > I think that having the same script for multiple platforms (python) is advantage here. Why would you like to have separate bash / powershell scripts for each platform ?
> 
> It does not support Android build. Script includes just cmake command, so I think you can also use meson run_command, so that we can compile openblas when we need for all the platform we support.
> 
> Or, you can propose alternatives to compile the openblas for all the platfom we support and make sure it is compiled properly.
> 
> also it could be applicable for ruy as well. please take a look.

OK I'll try to modify it to use meson run_command

gkisalapl:
> OK I'll try to modify it to use meson run_command

done

gkisalapl:
> From what I've gathered, this patch manually builds OpenBLAS using CMake, resulting in the creation of a 'build' directory under the OpenBLAS project. The build results will then be used to link to NNTrainer instead of utilizing the Meson CMake subproject. Please correct me if I'm wrong.

Yes You're right. Advantage from our side is that we're checking if subprojects/OpenBLAS/build directory exist and if so skip rebuilding OpenBLAS in next meson runs

skykongkong8 (approved):

myungjoo (approved):

PR #3187, opened 2025-05-21
[ q4K ] Enable non-4-divisible-M q4K GEMM case

skykongkong8:
- DONE : Compute all q4K GEMM rows with q4K GEMV to avoid divisibility
- TODO : Compute with q4K GEMM kernel first, and compute remaining rows with q4K GEMV + apply multithreading on the total sequence
- Add unittest for that (M = 457, 458, 459) accordingly
- With recent upstream main, leftover rows for 4 after GEMM kernel, left zero-filled rows

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.cpp
+++ b/nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.cpp
@@ -182,7 +182,37 @@ void __ggml_q4_K_8x8_q8_K_GEMM(const unsigned int M, const unsigned int N,
                                 (void *)((char *)B + M_step_start * B_step),
                                 QA.data(), M, M_step_end - M_step_start);
     }
-  } else { // GEMM
+  } else if (M % 4 != 0) {
+    int n_threads = 4;
+    if (K < 1592 && N < 1592)
+      n_threads = 1;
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.cpp
```diff
@@ -182,7 +182,38 @@ void __ggml_q4_K_8x8_q8_K_GEMM(const unsigned int M, const unsigned int N,
                                 (void *)((char *)B + M_step_start * B_step),
                                 QA.data(), M, M_step_end - M_step_start);
     }
-  } else { // GEMM
+  }
+  if (M % 4 != 0) {
```
EunjuYang, line 186: Do you intend for the case M == 1 to go through this block as well? 
Should it be an else if instead?

```suggestion
  else if (M % 4 != 0) {
```

skykongkong8: Woah, my mistakes! thanks!

EunjuYang (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #3188, opened 2025-05-21
[Tensor] Add Q6_K Tensor class

djeong20:
This PR introduces a new type of Tensor class, Q6_K_Tensor.
This tensor class aims to store Q6_K Quantized data types.
Please note that only data storage and dot operations will be valid for this class.

**Self-evaluation:**


nntrainer/tensor/q6_k_tensor.cpp
```diff
@@ -0,0 +1,90 @@
...
+namespace nntrainer {
+
+Q6_K_Tensor::Q6_K_Tensor(std::string name_, Tformat fm) :
+  TensorBase(name_, fm) {}
+
+Q6_K_Tensor::Q6_K_Tensor(const TensorDim &d, bool alloc_now, Initializer init,
+                         std::string name) :
+  TensorBase(d, false, init, name) {
+  NNTR_THROW_IF(d.batch() != 1 || d.channel() != 1 ||
+                  (d.height() % 256 != 0 && d.width() % 256 != 0),
```
EunjuYang, line 24: Simple question: Should Q6_K also be a multiple of 256?
I successfully converted my own model, which is not a multiple of 256, into Q6_K.
Should we add padding to ensure it meets the Q6_K requirements?

If so, what about creating a padded tensor instead of throwing an error?

EunjuYang (commented):
Could you add some simple unit tests for the Q6_K tensor? Additionally, we need to ensure support for Q6_K tensors where the dimension is not a multiple of 256.

nntrainer/tensor/q6_k_tensor.cpp
skykongkong8: Previous 256-divisibility was dependent on Q4_K"x8" weight, not Q4_K itself, technically. I think @djeong20 just refered to previous format.
AFAIK too, Q6_K won't have such constraint.

djeong20: What would be the number of elements that a single q6_k block represents?

```C++
struct block_q6_K {
  uint8_t ql[128];   // quants, lower 4 bits
  uint8_t qh[64];    // quants, upper 2 bits
  int8_t scales[16]; // scales, quantized with 8 bits
  uint16_t d;        // super-block scale
};
```

djeong20: I thought a single Q4_K should be a multiple of 256, and Q4_Kx8 should be a multiple of 2,048. Please correct me if I'm wrong.

skykongkong8: 1. For quantization datatype's perspective, for Q4_K block, AFAIK, width might be divisible by 256, but not for height. This is because Q4_K block itself takes 8 blocks of 32 elements (total **256 continuous** FP32 elements per-superblock) to formulate itself. 
2. For kernel's perspective, For normal dot kernels for Q4_K and Q6_K, I coud find assertion code like:
```C++
void ggml_vec_dot_q4_K_q8_K(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, size_t bx, const void * GGML_RESTRICT vy, size_t by, int nrc) {
    assert(n % QK_K == 0);
...
void ggml_vec_dot_q6_K_q8_K(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, size_t bx, const void * GGML_RESTRICT vy, size_t by, int nrc) {
    assert(n % QK_K == 0);
...
```
But nothing for height.
In contrast, for q4k 8x8 gemm kernel:
```C++
static void ggml_gemm_q4_K_8x8_q8_K(int n, float * GGML_RESTRICT s, size_t bs, const void * GGML_RESTRICT vx, const void * GGML_RESTRICT vy, int nr, int nc) {
    const int qk = QK_K;
...
    assert (n % qk == 0);
    assert (nr % 4 == 0);
...
```

Hope this helps your understanding

skykongkong8: > Previous 256-divisibility was dependent on Q4_K"x8" weight, not Q4_K itself, technically. I think @djeong20 just refered to previous format. AFAIK too, Q6_K won't have such constraint since Q6_K does **NOT** have such packed format like Q6_Kx8.

And this was my misunderstanding. M should be divisible by 4 for pure GEMM kernel, not by 256! This was fixed in #3187  FYI) @EunjuYang @jijoongmoon

EunjuYang: As I understood,

- The width (i.e., `row_size` in ggml code) should be a multiple of 256.
- The height should be divisible by 4. (This was related to a previous assertion issue, which has been fixed in #3187)

Am I correct?

If so, it seems that the current tensor cannot support `lm_head`, which has a size of size 3072 x 105900.
One thing I'm curious about is whether we will quantize the transposed of lm_head. Then, can it be quantized to Q6_K?

skykongkong8: If you are talking about Q6_K, I believe **not**. Height will not have any divisibility-dependency, but width, yes, should be divisible by 256.
This can be also shown from:
```C++
void quantize_row_q6_K(const float * GGML_RESTRICT x, void * GGML_RESTRICT vy, int64_t k) {
    assert(k % QK_K == 0);
    block_q6_K * GGML_RESTRICT y = vy;
    quantize_row_q6_K_ref(x, y, k);
}
...
void dequantize_row_q6_K(const block_q6_K * GGML_RESTRICT x, float * GGML_RESTRICT y, int64_t k) {
    assert(k % QK_K == 0);
...
}
```

nntrainer/tensor/q6_k_tensor.cpp
```diff
@@ -0,0 +1,89 @@
...
+Q6_K_Tensor::Q6_K_Tensor(std::string name_, Tformat fm) :
+  TensorBase(name_, fm) {}
+
+Q6_K_Tensor::Q6_K_Tensor(const TensorDim &d, bool alloc_now, Initializer init,
+                         std::string name) :
+  TensorBase(d, false, init, name) {
+  NNTR_THROW_IF(d.batch() != 1 || d.channel() != 1 || d.width() % 256 != 0,
+                std::invalid_argument)
+    << "Q6_K_Tensor must be 2 dimensional tensor with batch size 1 and "
+       "height or width must be divisible by 256";
```
EunjuYang, line 26:
```suggestion
    << "Q6_K_Tensor must be 2 dimensional tensor with batch size 1 and "
       "width must be divisible by 256";
```

nntrainer/tensor/q6_k_tensor.h
```diff
@@ -0,0 +1,238 @@
...
+  void initialize(Initializer init) {
+    throw std::invalid_argument("Q6_K_Tensor::initialize() is not supported.");
+  }
+
+  /**
+   * @copydoc Tensor::initialize()
+   */
+  void initialize() {
+    throw std::invalid_argument("Q6_K_Tensor::initialize() is not supported.");
+  }
```
EunjuYang, line 152: Is it okay to throw an exeception for `initialize()`?

EunjuYang (commented):
Thank you for implementing new tensor type, Q6_K.
Could you make unittests for this tensor?

nntrainer/tensor/q6_k_tensor.h
EunjuYang: I think we cannot use this tensor, since `allocate()` requires initialize() which raises an error by default. Could you update this?

nntrainer/tensor/q6_k_tensor.cpp
djeong20: fixed!

nntrainer/tensor/q6_k_tensor.h
djeong20: thank you! it is now fixed

djeong20:
> Thank you for implementing new tensor type, Q6_K. Could you make unittests for this tensor?

Added!

EunjuYang (approved):
LGTM

skykongkong8 (approved):
Good to go!

dkjung (approved):
LGTM

PR #3193, opened 2025-05-22
[ cpu_backend ] Add q6_K dequantize function

skykongkong8:
- arm compute backend, x86 compute backend now supports q6K dequantize function
- fallback backend will emit error, but since ggml has naive implementations as well, it should work at the the final form of this commit

**Self evaluation:**


skykongkong8:
This is introduced in #3189 after this PR is proposed.

PR #3196, opened 2025-05-23
Enable Using Pre-Built CLBlast Packages

djeong20:
This PR allows the use of pre-built CLBlast packages.
It uses the installed CLBlast and compiles it from source if it does not exist, potentially reducing compilation time.

Note that CLBlast can be installed with:
```
sudo apt install libclblast-dev
```

**Self-evaluation:**


EunjuYang (approved):
LGTM

PR #3197, opened 2025-05-23
[ q4K ] Optimize non-4-divisible M case for GEMM and leftover GEMV

skykongkong8:
- Optimize q4K GEMM leftover computation by computing output rows greatest multiple of 4 that does not exceed M with GEMM kernel
- On top of that, compute output leftover 1 ~ 3 rows with GEMV kernel
- Apply multithreading strategy for both kernels

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.cpp
+++ b/nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.cpp
@@ -183,21 +183,51 @@ void __ggml_q4_K_8x8_q8_K_GEMM(const unsigned int M, const unsigned int N,
                                 QA.data(), M, M_step_end - M_step_start);
     }
   } else if (M % 4 != 0) {
-    int n_threads = 4;
-    if (K < 1592 && N < 1592)
-      n_threads = 1;
-    int blocks_per_row = (K + QK_K - 1) / QK_K;
-    int qa_size = sizeof(block_q8_K) * blocks_per_row;
+    int n_threads = std::thread::hardware_concurrency();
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.cpp
```diff
@@ -183,36 +183,66 @@ void __ggml_q4_K_8x8_q8_K_GEMM(const unsigned int M, const unsigned int N,
...
+
+      src0_start =
+        (src0_start % 8) ? src0_start + 8 - (src0_start % 8) : src0_start;
+      src0_end = (src0_end % 8) ? src0_end + 8 - (src0_end % 8) : src0_end;
+
+      ::ggml_gemm_q4_K_8x8_q8_K(K, (float *)(C + src0_start), ldc,
+                                (void *)((char *)B + src0_start * B_step),
+                                QA.data(), M4 * 4, src0_end - src0_start);
+    }
+
```
EunjuYang, line 222: I believe this code can be merged with `else` clause, which indicates that it contains some redundancy. What do you think about merging them?

EunjuYang (commented):
I'm pleased that we've optimized the non-4-divisible M case code. As I mentioned in the comment below, I believe there is potential to eliminate redundant code. What are your thoughts?

EunjuYang (approved):
LGTM

DonghakPark (approved):
LGTM!

jijoongmoon (approved):
LGTM. Removing redundant code could be done in other PR.

PR #3199, opened 2025-05-23
[Windows] Add windows clang build to CI

gkisalapl:


skykongkong8 (approved):

.github/workflows/windows_clang.yml
```diff
@@ -0,0 +1,30 @@
+name: "Build Test - Windows Meson"
```
DonghakPark, line 1: Please add compiler name like ```"Build Test - Windows Meson Clang"

DonghakPark (approved):
Nice Work! but please write PR message

.github/workflows/windows_clang.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: windows-2025
```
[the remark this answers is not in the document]
gkisalapl: windows-2022 builder got clang 18 which fail to build our project due to llvm bug fixed in 19: https://github.com/llvm/llvm-project/issues/96275 (that's why I've added only windows-2025 builder)

.github/workflows/windows_clang.yml
gkisalapl: > Windows Meson Clang

done

djeong20 (approved):

PR #3201, opened 2025-05-26
[Test] Disable DefaultEnvironmentPath Test

djeong20:
This PR disables the DefaultEnvironmentPath test. Currently, NNTrainer uses a single engine class to manage the context. However, the existing plugin tests utilize a separate AppContext class to verify this context. As a result, this discrepancy causes an error when creating an object.

Please note that `layer_plugin_test.cpp` is not used. Therefore, removing it.

**Self-evaluation:**


Applications/Custom/OptimizerPlugin/optimizer_plugin_common_test.cpp
```diff
@@ -38,13 +38,13 @@ TEST_P(OptimizerPluginCommonTest, DlRegisterDirectory_n) {
                std::invalid_argument);
 }
```
songgot, line 40: How about adding ?
```
// TODO: Re-enable after resolving engine context issue
``

songgot (approved):
LGTM

dkjung (approved):
Just for check:
- `Engine` adopts singleton pattern.
- Plugin tests create `Engine` instances, but it is not realistic.

Am I right?

jijoongmoon (approved):
LGTM

PR #3213, opened 2025-05-29
Add getProperty function

SeoHyungjun:
Implemented a getProperty function that allows checking the value using the key of a set property.
The actual implementation must be added in the layer. 
If not implemented, it returns an empty string ("").

To use getProperty, inherit from layer_impl and insert the property tuple existing in the layer into LayerImpl::for_each function.

If the key exists in the tuple, it returns the value. If the property is not set and there is no default value, it returns "empty".

### DIFF
--- a/nntrainer/layers/layer_devel.h
+++ b/nntrainer/layers/layer_devel.h
@@ -252,6 +252,13 @@ public:
   virtual void setProperty(const std::vector<std::string> &values) = 0;
 
   /**
+   * @brief     Get property value of layer
+   * @param     key Property key to retrieve
+   * @return    Property value as string (empty string if not found)
+   */
+  virtual std::string getProperty(const std::string &key) { return ""; }
+
+  /**
    * @brief this function helps exporting the layer in a predefined format,
    * while workarounding issue caused by templated function type eraser
    *
--- a/nntrainer/layers/layer_impl.h
+++ b/nntrainer/layers/layer_impl.h
@@ -78,6 +78,11 @@ public:
   virtual void setProperty(const std::vector<std::string> &values) override;
 
   /**
+   * @copydoc Layer::getProperty(const std::string &key)
+   */
+  virtual std::string getProperty(const std::string &key) override;
+
+  /**
    * @copydoc Layer::exportTo(Exporter &exporter, const ml::train::ExportMethods
    * &methods)
    */
--- a/nntrainer/layers/layer_node.cpp
+++ b/nntrainer/layers/layer_node.cpp
@@ -228,6 +228,24 @@ void LayerNode::setProperty(const std::vector<std::string> &properties) {
   }
 }
 
+std::string LayerNode::getProperty(const std::string &key) {
+  if (layer_node_props) {
+    std::string result = find_in_tuple(*layer_node_props, key);
+    if (!result.empty()) {
+      return result;
+    }
+  }
+
+  if (layer_node_props_realization) {
+    std::string result = find_in_tuple(*layer_node_props_realization, key);
+    if (!result.empty()) {
+      return result;
+    }
+  }
+
+  return layer->getProperty(key);
+}
+
 void LayerNode::setWeights(const std::vector<float *> weights) {
   NNTR_THROW_IF(!run_context, std::runtime_error)
     << __func__ << " layer needs to be finalized first!";
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -195,6 +195,18 @@ public:
   virtual void setProperty(const std::vector<std::string> &values) = 0;
 
   /**
+   * @brief     Get property value of layer
+   * @param[in] key Property key to retrieve
+   * @retval    std::string Property value as string
+   * @retval    Empty string if property not found
+   * @details   For layers derived from layer_impl: Property lookup is handled
+   * automatically.
+   * @details For layers derived from layer_devel: The getProperty() function
+   * must be overridden to enable property retrieval.
+   */
+  virtual std::string getProperty(const std::string &key) = 0;
+
+  /**
    * @brief     Get name of the layer
    * @retval    name of the layer
    * @note      This name is unique to this layer in a model
--- a/nntrainer/layers/layer_impl.cpp
+++ b/nntrainer/layers/layer_impl.cpp
@@ -37,6 +37,14 @@ void LayerImpl::setProperty(const std::vector<std::string> &values) {
          std::to_string(values.size());
 }
 
+std::string LayerImpl::getProperty(const std::string &key) {
+  if (!layer_impl_props)
+    return "";
+
+  std::string result = find_in_tuple(*layer_impl_props, key);
+  return !result.empty() ? result : "";
+}
+
 void LayerImpl::exportTo(Exporter &exporter,
                          const ml::train::ExportMethods &method) const {
   exporter.saveResult(*layer_impl_props, method, this);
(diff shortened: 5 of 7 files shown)


api/ccapi/include/model.h
```diff
@@ -93,7 +93,9 @@ for inference and training without any configurations*/
     ML_TRAIN_MODEL_FORMAT_FLATBUFFER,             /**< flatbuffer file */
   MODEL_FORMAT_ONNX = ML_TRAIN_MODEL_FORMAT_ONNX, /**< ONNX file */
 
-  MODEL_FORMAT_QNN = ML_TRAIN_MODEL_FORMAT_QNN /**< qnn binary file */
+  MODEL_FORMAT_QNN = ML_TRAIN_MODEL_FORMAT_QNN, /**< qnn binary file */
+  MODEL_FORMAT_KVCACHE = ML_TRAIN_MODEL_FORMAT_KVCACHE,
+  MODEL_FORMAT_LORA_BIN = ML_TRAIN_MODEL_FORMAT_LORA_BIN
```
jijoongmoon: I wonder do we need this?

jijoongmoon (changes requested):
May be we don't need to set the enum values for KVCACHE and LORA BIN.

nntrainer/layers/layer_impl.h
```diff
@@ -84,6 +88,35 @@ class LayerImpl : public virtual Layer {
...
+                  result = to_string(elem);
+                } else {
+                  result = "empty";
+                }
+              }
+            }(),
+            0));
+
+    return result;
+  }
```
dkjung: Can you help me understand how how this function works since I've never used `...` for template parameters.

In addition, argument `f` doesn't look to be used. If so, can we remove it?

dkjung (approved):
minor comments

dkjung (commented):
CI's format checked failed. Can you see?

DonghakPark (commented):
I think here's some mixed code with https://github.com/nnstreamer/nntrainer/pull/3200
how about just seperate getProperty code only in This PR

api/ccapi/include/model.h
SeoHyungjun: It was an unnecessary Enum, so I deleted it.

SeoHyungjun:
> I think here's some mixed code with #3200 how about just seperate getProperty code only in This PR

I removed unnecessary commits. Thank you!

DonghakPark (commented):
Could You Please 
1. Add Doxygen at header files ?
2. Apply clang-format

nntrainer/layers/layer_devel.h
```diff
@@ -251,6 +251,8 @@ class Layer {
    */
   virtual void setProperty(const std::vector<std::string> &values) = 0;
 
+  virtual std::string getProperty(const std::string &key) { return ""; }
```
EunjuYang, line 259: It should raise the error in this function.

nntrainer/layers/layer_impl.h
```diff
@@ -77,13 +77,46 @@ class LayerImpl : public virtual Layer {
    */
   virtual void setProperty(const std::vector<std::string> &values) override;
 
+  virtual std::string getProperty(const std::string &key) override {
+    return "";
```
EunjuYang: It should check the key is in `layer_impl_props`.

nntrainer/layers/layer_node.cpp
```diff
@@ -229,6 +229,10 @@ void LayerNode::setProperty(const std::vector<std::string> &properties) {
   }
 }
 
+std::string LayerNode::getProperty(const std::string &key) {
+  return layer->getProperty(key);
```
EunjuYang, line 246: Before returning layer->getProperty(key), it should try to getProperty from `layer_node_props`.
If the key is not in `layer_node_props`, then it should return `layer->getProperty(key)`.

nntrainer/layers/layer_impl.h
DonghakPark: i think this work like below. is this right ? @SeoHyungjun 

1. `std::string result = "";`: Initializes a string variable `result` to an empty string. This will store the outcome.
2. `(..., (lambda_expression, 0));`: This is a C++17 fold expression. This syntax executes the lambda expression for each index in the `std::index_sequence<ls...>`.
- The lambda expression `[&] { ... }()`:
    - It captures the `result` variable by reference (`[&]`), allowing the lambda to modify `result`.
    - `auto &&elem = std::get<ls>(t);`: Retrieves the element at the current index from the tuple `t` using `std::get<ls>(t)`. `auto&&` preserves the value category (lvalue/rvalue) of the element, similar to perfect forwarding. `ls`
    - `if (strcmp(getPropKey(elem), key.c_str()) == 0)`:
        - `getPropKey(elem)`: This assumes there's a function `getPropKey` that takes a tuple element `elem` and returns its "property key." This function is expected to return a `const char*` (since , a C-style string comparison function, is used). This `getPropKey` function is not defined in the provided code and must be defined elsewhere by the user. `strcmp`
        - It compares the retrieved property key with the input `key`.

    - If the keys match:
        - `if (!elem.empty())`: This assumes that the element `elem` has an `empty()` member function, which is called to check if the element is empty. This `empty()` method must also be defined for the element's type.
        - `result = to_string(elem);`: If the element is not empty, the `to_string(elem)` function is called to convert the element to a string, which is then stored in `result`. This `to_string` function must also be defined or overloaded for the element's type.
        - `else { result = "empty"; }`: If the element is empty, `result` is set to the string "empty".

3. If multiple elements match the given `key`, the `result` variable will be overwritten with the value of the last matching element found during the iteration.
4. If no element matches the `key`, `result` will remain its initial value, an empty string (`""`).
5. `return result;`: Returns the final `result` string.

Ref) at here : https://modoocode.com/290 get some help to understand ```variadic template```, ```parameter pack```, ```fold expression```

EunjuYang (commented):
Since we are considering making `getProperty` a pure virtual interface for layer, we have to thoroughly consider the layer property hierarchy.

nntrainer/layers/layer_impl.h
SeoHyungjun: @dkjung 
At first, I tried to pass a compare function to enable comparison.
However, I realized there are only two cases: empty or having a value, so I didn't use it.
I removed the unused f.

nntrainer/layers/layer_devel.h
SeoHyungjun: getProperty is not a function used by all layers.
Since it's only needed for specific layers, making it a pure virtual function would require all layers inheriting from layer_devel to implement it.

SeoHyungjun: To avoid this, should provide an empty body, but since a return value is required, we added a body that returns an empty string.
This way, all layers inheriting from layer_devel won't need to override the getProperty function

nntrainer/layers/layer_impl.h
SeoHyungjun: It is modified.

SeoHyungjun: Fixed.

nntrainer/layers/layer_node.cpp
SeoHyungjun: Fixed.

nntrainer/layers/layer_impl.h
SeoHyungjun: @DonghakPark 
Thank you for explaining it instead.

nntrainer/layers/layer_devel.h
EunjuYang: Yes, so what about raising error like NYI in this function since this body will be called only for layers not overriding this?

DonghakPark (approved):

PR #3218, opened 2025-05-30
[OpenCL] Add Q6_K GEMV kernel

djeong20:
This PR implements the Q6_K GEMV OpenCL kernel.
Note that the current implementation loads input data into global memory every time.
A preloading version is a work in progress.

**Self-evaluation:**

#### 1. Linux-x86 (NVIDIA GeForce GTX 1650)
|Engine|avg time|
|------|---|
|GPU (preload) | 0.02982 ms|
|GPU| 2.21088 ms|
|CPU| 6.31663 ms|

#### 2. Android (Qualcomm Adreno 740)
|Engine|avg time|
|------|---|
|GPU (preload) | 0.18577 ms|
|GPU| 2.981 ms|
|CPU| 16.050 ms|

## Note

The following matrix sizes are used for the benchmark and profile.
- M: 1
- K: 3072
- N: 105900

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
@@ -16,6 +16,132 @@
 
 namespace nntrainer {
 
+const std::string &getQ6KSgemvClKernel() {
+  static const std::string q6_k_sgemv_cl_kernel_ =
+    R"(
+    #pragma OPENCL EXTENSION cl_khr_fp16 : enable
+
+    #define QK_K 256
+    #define N_SIMDWIDTH 16
+    #define N_SIMDGROUP 2
+    #define N_DST 1
+    #define BLOCK_STRIDE (N_SIMDWIDTH / 16)
+
+    typedef char int8_t;
+    typedef uchar uint8_t;
+    typedef short int16_t;
+    typedef ushort uint16_t;
+    typedef int int32_t;
+    typedef uint uint32_t;
+
+    typedef struct {
+        uint8_t ql[QK_K / 2];
+        uint8_t qh[QK_K / 4];
+        int8_t  scales[QK_K / 16];
+        half d;
+    } block_q6_K;
+
+    kernel void kernel_mul_mv_q6_K_f32(
+        global void * src0,
+        ulong offset0,
+        global float * src1,
+        ulong offset1,
+        global float * dst,
+        ulong offsetd,
+        int ne00,
+        int ne01,
+        int ne02,
+        int ne10,
+        int ne12,
+        int ne0,
+        int ne1,
+        int r2,
+        int r3
+    ) {
+        __local float reduction_buf[N_SIMDGROUP][N_SIMDWIDTH];
+
+        src0 = (global void*)((global char*)src0 + offset0);
+        src1 = (global float*)((global char*)src1 + offset1);
+        dst = (global float*)((global char*)dst + offsetd);
+
+        int nb = ne00 / QK_K;
+
+        int r0 = get_group_id(0);
+        int r1 = get_group_id(1);
+        int im = get_group_id(2);
+        int lid = get_local_id(0);
+        int lsize = get_local_size(0);
+
+        int row_group = lid / N_SIMDWIDTH;
+        int lane = lid % N_SIMDWIDTH;
+        int row = r0 * N_SIMDGROUP + row_group;
+
+        int i12 = im % ne12;
+        int i13 = im / ne12;
+
+        ulong offset_src0 = (i12 / r2) * (nb * ne01) + (i13 / r3) * (nb * ne01 * ne02);
+
+        global block_q6_K * x = (global block_q6_K *) src0 + row * nb + offset_src0;
+        global float      * yy = (global float     *) src1 + r1 * ne10 + im * ne00 * ne1;
+
+        uchar kmask1 = 0x03, kmask2 = 0x0C, kmask3 = 0x30, kmask4 = 0xC0;
+
+        int tid  = lane / BLOCK_STRIDE;
+        int ix   = lane % BLOCK_STRIDE;
+        int ip   = tid / 8;
+        int il   = tid % 8;
+        int n    = 4;
+        int l0   = n * il;
+        int is   = 8 * ip + l0 / 16;
+
+        int y_offset = 128 * ip + l0;
+        int q_offset_l = 64 * ip + l0;
+        int q_offset_h = 32 * ip + l0;
+
+        float sumf = 0.0f;
+
+        for (int i = ix; i < nb; i += BLOCK_STRIDE) {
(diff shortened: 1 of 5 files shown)


skykongkong8:
Just for informatiojn, CPU Q6_K GEMV computation for current updated version gives - gemm_q6_K: 6.145672 ms

nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
```diff
@@ -16,6 +16,132 @@
...
+    typedef char int8_t;
+    typedef uchar uint8_t;
+    typedef short int16_t;
+    typedef ushort uint16_t;
+    typedef int int32_t;
+    typedef uint uint32_t;
+
+    typedef struct {
+        uint8_t ql[QK_K / 2];
+        uint8_t qh[QK_K / 4];
```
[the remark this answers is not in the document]
djeong20: Yes, in a 6-bit quantized value, ql represents the low 4 bits, while qh represents the high 2 bits

nntrainer/tensor/cl_operations/blas_kernels.h
```diff
@@ -29,6 +29,17 @@ static ClContext *blas_cc =
...
 
+/**
+ * @brief     Q6_K sgemv computation : Y = A*X
+ * @param[in] matAdata void * for Matrix A
+ * @param[in] vecXdata float * for Vector X
+ * @param[in] vecYdata float * for Vector Y
+ * @param[in] M number of rows in matrix A
+ * @param[in] N number of columns in matrix A
+ */
+void sgemv_q6_k_cl(const void *matAdata, const float *vecXdata, float *vecYdata,
```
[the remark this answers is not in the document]
djeong20: I'm not sure if returning a boolean value would be useful. For kernels that return a single value, we set the return type accordingly. For kernels that return a vector or matrix, we keep the return type as void and pass the output pointer as a parameter. If we decide to change it to return a boolean value, we should also pass the output pointer for the function that returns a single value. What do you think is the best approach?

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -19,6 +19,191 @@
...
+
+  kernel_q6_k_sgemv_ptr =
+    blas_cc->registerClKernel(getQ6KSgemvClKernel(), "kernel_mul_mv_q6_K_f32");
+
+  if (!kernel_q6_k_sgemv_ptr) {
+    ml_loge("Failed to register kernel_q6_k_sgemv_ptr");
+    return;
+  }
+
+  int q6k_size = 210 * M * N / 256;
```
EunjuYang, line 36: This is a minor comment.
In terms of code coherence, it appears that variables with the same name are being used for different purposes.
Please refer to the following link for context:

https://github.com/nnstreamer/nntrainer/blob/4cbe7ef88f63d4c91bcffeff48952eb7ea4dbb35/nntrainer/tensor/q6_k_tensor.h#L32

Additionally, instead of directly using constants, what about moving `ggml-common.h` to `ggml_interface.h` and importing `ggml_interface.h` in both q6_k_tensor and the BLAS kernels? While this may not be the best approach, it could serve as a starting point for finding a clearer solution.

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -19,6 +19,191 @@
...
+  int ne12 = 1; // number of channels in vector A
+  int ne13 = 1; // number of channels in vector A (Need to check)
+  int ne0 = N;  // number of rows in output vector Y
+  int ne1 = 1;  // number of columns in output vector Y
+
+  int r2 = 1; // number of batches in vector A
+  int r3 = 1; // number of batches in matrix X
+
+  int nth0 = 2;
+  int nth1 = 16;
```
EunjuYang, line 66: unused variables
```suggestion
```

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -19,6 +19,191 @@
...
+  int ne02 = 1; // number of channels in matrix X
+  int ne10 = M; // number of rows in vector A
+  int ne11 = 1; // number of columns in vector A
+  int ne12 = 1; // number of channels in vector A
+  int ne13 = 1; // number of channels in vector A (Need to check)
+  int ne0 = N;  // number of rows in output vector Y
+  int ne1 = 1;  // number of columns in output vector Y
+
+  int r2 = 1; // number of batches in vector A
+  int r3 = 1; // number of batches in matrix X
```
EunjuYang, line 63: What about making them const?

nntrainer/tensor/cl_operations/blas_kernels.cpp
EunjuYang, line 36: I believe we may need to verify the multiplication of QK_K.

nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
```diff
@@ -16,6 +16,132 @@
...
+        int ne12,
+        int ne0,
+        int ne1,
+        int r2,
+        int r3
+    ) {
+        __local float reduction_buf[N_SIMDGROUP][N_SIMDWIDTH];
+
+        src0 = (global void*)((global char*)src0 + offset0);
+        src1 = (global float*)((global char*)src1 + offset1);
```
EunjuYang, line 64: Simple question. Is it okay to cast it char * to reflect the `offset1`?
What's this for?

nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
```diff
@@ -16,6 +16,132 @@
...
+                reduction_buf[row_group][lane] += reduction_buf[row_group][lane + offset];
+            }
+            barrier(CLK_LOCAL_MEM_FENCE);
+        }
+
+        if (lane == 0) {
+            int global_row = r0 * N_SIMDGROUP + row_group;
+            dst[r1 * ne0 + im * ne0 * ne1 + global_row] = reduction_buf[row_group][0];
+
+            // dst[r1 * ne0 + im * ne0 * ne1 + row] = reduction_buf[row_group][0];
```
EunjuYang, line 137:
```suggestion
```

EunjuYang (approved):
Thanks for your hard work! LGTM 👍

nntrainer/tensor/cl_operations/blas_kernels.cpp
skykongkong8: This is out of topic, but anyone can refer to here : https://github.com/skykongkong8/nntrainer/blob/d38dc87e62bef3055e006b65ec0ab505277898a7/nntrainer/tensor/cpu_backend/ggml_interface/ggml_cpu_impl.cpp#L1920

to understand how these params work and delivered to the ggml kernels compared to the nntrainer

PR #3221, opened 2025-05-30
[model] update time step

lhs8928:
- Until now attention and input/output for autoregressive model is allocated by maximum time step. This commit will allocate memory for the input timestep.

### DIFF
--- a/api/ccapi/include/model.h
+++ b/api/ccapi/include/model.h
@@ -323,6 +323,21 @@ public:
                         unsigned int to, bool output_hidden_state = false) = 0;
 
   /**
+   * @brief     reset input dimensions of a model
+   * @param[in] dims input dimensions
+   * @note Similar to reinitialize, the resetInputDimension API is used for
+   * modifying input dimensions after model initialization. The reinitialize
+   * function should be the officially called API when changing input
+   * dimensions, as it properly recalculates weights, tensors, and outputs for
+   * each layer. On the other hand, resetInputDimension is a specialized API
+   * created to modify only specific dimensions (specifically height values)
+   * within input/output dimensions. Since this API uniformly adjusts the height
+   * across all model layers, developers must verify that every layer in their
+   * model architecture can safely accommodate such height modifications.
+   */
+  virtual void resetInputDimension(std::vector<ml::train::TensorDim> dims) = 0;
+
+  /**
    * @brief     Summarize the model
    * @param out std::ostream to get the model summary
    * @param verbosity verbosity of the summary
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -331,6 +331,28 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
     label_dims[idx] = tensor_manager->getTensor(label_list[idx])->getDim();
 }
 
+void NetworkGraph::resetInputDimension(std::vector<TensorDim> dims) {
+  auto allocated = tensor_manager->isAllocated();
+
+  if (allocated)
+    deallocateTensors();
+
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    if ((*iter)->isFinalized()) {
+      (*iter)->updateTensorsByInputDimensions(dims);
+    }
+  }
+
+  if (allocated)
+    allocateTensors(exec_mode);
+
+  /** update input and label dimensions */
+  for (unsigned int idx = 0; idx < input_list.size(); idx++)
+    input_dims[idx] = tensor_manager->getTensor(input_list[idx])->getDim();
+  for (unsigned int idx = 0; idx < label_list.size(); idx++)
+    label_dims[idx] = tensor_manager->getTensor(label_list[idx])->getDim();
+}
+
 void NetworkGraph::applyGradients(
   LayerNode *node, const std::function<void(Weight &)> &apply_func) {
   if (!node->getTrainable())
--- a/nntrainer/graph/network_graph.h
+++ b/nntrainer/graph/network_graph.h
@@ -173,6 +173,21 @@ public:
   void setBatchSize(unsigned int batch_size);
 
   /**
+   * @brief     reset input dimensions of a model
+   * @param[in] dims input dimensions
+   * @note Similar to reinitialize, the resetInputDimension API is used for
+   * modifying input dimensions after model initialization. The reinitialize
+   * function should be the officially called API when changing input
+   * dimensions, as it properly recalculates weights, tensors, and outputs for
+   * each layer. On the other hand, resetInputDimension is a specialized API
+   * created to modify only specific dimensions (specifically height values)
+   * within input/output dimensions. Since this API uniformly adjusts the height
+   * across all model layers, developers must verify that every layer in their
+   * model architecture can safely accommodate such height modifications.
+   */
+  void resetInputDimension(std::vector<TensorDim> dims);
+
+  /**
    * @brief try apply gradient if possible
    * @note if it is not the last of the gradient access, this is noop
    * @note if the gradient is to be clipped by norm, this is noop
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -824,6 +824,30 @@ public:
   void updateTensor(unsigned int idx, unsigned int batch);
 
   /**
+   * @brief Update the dimensions for a requested input
+   *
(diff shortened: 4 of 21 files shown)


nntrainer/graph/network_graph.h
```diff
@@ -172,6 +172,8 @@ class NetworkGraph {
    */
   void setBatchSize(unsigned int batch_size);
 
+  void updateTimeStep(unsigned int max_timestep);
```
jijoongmoon: Needs doxygen tag to explain about this function.

nntrainer/layers/layer_context.h
```diff
@@ -823,6 +823,10 @@ class RunLayerContext {
    */
   void updateTensor(unsigned int idx, unsigned int batch);
 
+  void updateTensor(unsigned int idx, unsigned int dim, unsigned int value);
```
jijoongmoon: Same here

nntrainer/layers/layer_node.h
```diff
@@ -349,6 +349,8 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
    */
   void setBatch(unsigned int batch);
 
+  void updateTimeStep(unsigned int max_timestep);
```
jijoongmoon: here as well.

nntrainer/models/neuralnet.cpp
```diff
@@ -1010,10 +1013,25 @@ sharedConstTensors NeuralNetwork::incremental_inference(
...
+        new_X.emplace_back(MAKE_SHARED_TENSOR(Tensor::Map(
+          X[i]->getData(), in_dim.getDataLen() * sizeof(float), in_dim, 0)));
+      }
+      X = new_X;
+      model_graph.updateTimeStep(init_seq_len + num_to_generate);
+    }
+    model_graph.allocateTensors(ExecutionMode::INFERENCE);
   }
 
+  model_init_seq_len = init_seq_len;
```
EunjuYang: Simple question.
Is it acceptable to update the `model_init_seq_len` even when the `init_seq_len` is unrelated to the model_graph's time step?
 If incremental forwarding is called multiple times (i.e., the prefill is invoked several times by the service API), it might lead to redundant allocation actions.

nntrainer/tensor/tensor_base.cpp
```diff
@@ -124,6 +124,30 @@ void TensorBase::updateBatch(unsigned int batch) {
...
 
+void TensorBase::updateDimension(unsigned int axis, unsigned int value) {
+  if (dim.getTensorDim(axis) == value) {
+    return;
+  }
+
+  if (isAllocated())
+    throw std::invalid_argument("Cannot update tensor dimension for an allocated tensor");
+
+  switch (axis) {
```
EunjuYang: Does the axis always consider NCHW ?

EunjuYang (commented):
Overall looks good!

nntrainer/graph/network_graph.h
lhs8928: Added

nntrainer/layers/layer_context.h
lhs8928: ditto

nntrainer/layers/layer_node.h
lhs8928: ditto

nntrainer/models/neuralnet.cpp
lhs8928: 1. Incremental inference is designed exclusively for auto-regressive models. In other words, it always assumes that init_seq_len means model graph's time step.
2. The allocation overhead is not that significant, so it can tolerate multiple calls of prefill.

nntrainer/tensor/tensor_base.cpp
lhs8928: Axis it not related to tensor format.
For example, let's assumed that user wants to modify height of tensor. With the NCHW format user will set axis to be 2 and will set axis to be 1 for NHWC.

nntrainer/models/neuralnet.cpp
jijoongmoon: How about an explicit call of this function from outside? Just let the model developer or app developer handle this. Just call this whenever they want. Then, we do not need this num_to_generate parameter change.
Say.. ``` resetInputDimension(Dim)```

api/ccapi/include/model.h
```diff
@@ -313,14 +313,16 @@ class Model {
...
    * @retval list of output as float *
    * @note The output memory must not be freed by the caller
    */
   virtual std::vector<float *>
   incremental_inference(unsigned int batch, const std::vector<float *> &input,
                         const std::vector<float *> &label,
                         unsigned int init_seq_len, unsigned int from,
-                        unsigned int to, bool output_hidden_state = false) = 0;
+                        unsigned int to, bool output_hidden_state = false,
+                        unsigned int num_to_generate = 512) = 0;
```
djeong20: Quick question. Does the default value for the argument apply to the function that overrides it?

djeong20 (approved):
LGTM

nntrainer/tensor/tensor_base.cpp
EunjuYang: I see. In this case, if the user sets 1 for NHWC, the channel dimension will be updated, which differs from the user's intention. Is that correct? I mean, I'm wondering about the following part:
```
case 1:
   dim.channel(value);
   break;
```

nntrainer/models/neuralnet.cpp
EunjuYang: @lhs8928 
I see. It makes sense if it is acceptable to call it multiple times.
What I want to suggest is: what about updating the `model_init_seq_len` only when its allocation is updated?
Could you give me your opinion on this?

nntrainer/layers/layer_devel.h
```diff
@@ -276,6 +276,16 @@ class Layer {
...
   virtual void setBatch(RunLayerContext &context, unsigned int batch) {}
 
+  /**
+   * @brief Update the time step for the layer
+   * @param     context Context of the layer
+   * @param     max_timestep timestep value to be update
+   * @details Update the run context based on the updated max_timestep if
+   * required
+   */
+  virtual void updateTimeStep(RunLayerContext &context,
```
jijoongmoon: I think this function name should be changed for more generic purpose. The goal of this function is changing the tensor dimensions by input tensor dimensions. Not only for the time step. So I recommend to change this
``` updateTensorsByInputDim (RunLayerContext &context, TensorDim inDim) ```

File: nntrainer/graph/network_graph.cpp [hunk 8]
```diff
@@ -331,6 +331,34 @@ void NetworkGraph::setBatchSize(unsigned int batch_size) {
...
+  tensor_manager->updateInputOutput(tensor_format == "NCHW" ? 2 : 1,
+                                    height);
+
+  for (auto iter = cbegin(); iter != cend(); iter++) {
+    if ((*iter)->isFinalized()) {
+      (*iter)->updateTimeStep(height);
+    }
+  }
+
+  if (allocated)
```
myungjoo, line 346: Check the condition.

If not allocated yet, you won't allocate????

api/ccapi/include/model.h
lhs8928: Correct

nntrainer/models/neuralnet.cpp
lhs8928: > How about an explicit call of this function from outside? Just let the model developer or app developer handle this. Just call this whenever they want. Then, we do not need this num_to_generate parameter change. Say.. ` resetInputDimension(Dim)`

Applied

lhs8928: As per the method @jijoongmoon suggested (introducing resetInputDimension), model_init_seq_len is no longer used.

nntrainer/tensor/tensor_base.cpp
lhs8928: Correct

nntrainer/graph/network_graph.cpp
lhs8928: Yes.
resetInputDimension is a function that resets the input dimension as the name suggests.
If the input tensors are already allocated, it will reallocate them. Otherwise, it will only modify the tensor specification without performing reallocation.

nntrainer/layers/layer_devel.h
lhs8928: Updated as you recommended

PR #3222, opened 2025-05-30
[Fix] Fix ggml assertion in q6K quantization test

gkisalapl:
Fix ggm assertion in unit test caused by row value not divided by 256

**Self-evaluation:**

### DIFF
--- a/test/unittest/unittest_nntrainer_tensor.cpp
+++ b/test/unittest/unittest_nntrainer_tensor.cpp
@@ -874,7 +874,7 @@ TEST(nntrainer_Tensor, Q_Tensor_17_p) {
   /// with Q6_K tensor format
 
   const unsigned int M = 1;
-  uint32_t K = 5012; // height
+  uint32_t K = 5120; // height
   uint32_t N = 3072; // width
 
   // Float tensor
@@ -919,7 +919,7 @@ TEST(nntrainer_Tensor, QTensor_18_p) {
 
   const unsigned int M = 1;
   uint32_t K = 3072;
-  uint32_t N = 5012;
+  uint32_t N = 5120;
 
   std::vector<float> weight = generate_random_vector<float>(K * N, -0.05, 0.05);
   std::vector<float> activation =


skykongkong8 (approved):
LGTM

test/unittest/unittest_nntrainer_tensor.cpp
```diff
@@ -874,7 +874,7 @@ TEST(nntrainer_Tensor, Q_Tensor_17_p) {
   /// with Q6_K tensor format
 
   const unsigned int M = 1;
-  uint32_t K = 5012; // height
+  uint32_t K = 5120; // height
```
EunjuYang, line 877: (cc. @skykongkong8 )
As I understand it, the constraint on tensor dimensions seems to apply only to the row size. 
Do we need to use tensors where both the height and width are divisible by 256 for Q6_K?

gkisalapl: That's right

skykongkong8: > (cc. @skykongkong8 ) As I understand it, the constraint on tensor dimensions seems to apply only to the row size. Do we need to use tensors where both the height and width are divisible by 256 for Q6_K?

@EunjuYang That's a great question.
For `K`, yes. for `M` and `N`, no.
If `K` is not divisible by 256, `Q6_K` quantization is not possible in the first place. Or if it worked, such tensor must have been quantized in an invalid manner.

EunjuYang: Hmm.. Then, I need to double-check the term. Does K denote height or width?
I noticed that these two different unit tests misused N and K. (Sorry for the bugs in the code I wrote!) The update in this patch attempts to fix the height to be a multiple of 256.
If it makes an assertion, can we resolve it by updating line 898 instead of N and K?

gkisalapl: @skykongkong8 you're right only K needs to be divisible by 256. I've change value 5012 to 5120 in both unit test to keep some consistency.

skykongkong8: > Hmm.. Then, I need to double-check the term. Does K denote height or width? I noticed that these two different unit tests misused N and K. (Sorry for the bugs in the code I wrote!) The update in this patch attempts to fix the height to be a multiple of 256. If it makes an assertion, can we resolve it by updating line 898 instead of N and K?

@EunjuYang 
I just double-checked. I think you can refer to the `unittest_nntrainer_cpu_backend.cpp` for clear understanding.
To be a little bit more specific, nntrainer expects gemm_q6_K to behave something identical to:
```C++
  nntrainer::sgemm(0, false, true, M, N, K, 1.F, activation.data(), K,
                   weight.data(), K, 0.F, ref_dst.data(), N);
```
Here, as you can see, the leading dimension of B(Weight in gemm_q6_K) matrix(aka, ldb), is K.
This shows that K is column of the matrix, and in this case(transB GEMM's rhs matrix, RowMaj 2D Tensor), it denotes width.

In different point of view, you can also deep dive into ggml kernel as well:
```C++
size_t __ggml_quantize_q6_K(const float *src, void *dst, int64_t nrow,
                            int64_t n_per_row, const float *quant_weights) {
  return ::quantize_q6_K(src, dst, nrow, n_per_row, quant_weights);
}

void __ggml_quantize_row_q6_K(const float *src, void *dst, int64_t k) {
  ::quantize_q6_K(src, dst, 1, k, nullptr);
}
```
Observing the usage of `quantize_q6_K` gives an intuitive takeaway.
Nevertheless, it might differ from time-to-time. For example, you can refer to : https://github.com/ggml-org/llama.cpp/blob/master/media/matmul.png If the GEMM problem is defined in different format, the height and row, or width and height can be switched.

PR #3230, opened 2025-06-02
Enable OpenCL Support on Windows

djeong20:
This PR enables the use of OpenCL on Windows.
Please note that **OpenCL must be installed** in order to use OpenCL.

### DIFF
--- a/nntrainer/opencl/opencl_loader.cpp
+++ b/nntrainer/opencl/opencl_loader.cpp
@@ -44,7 +44,12 @@ bool LoadOpenCL() {
   }
 
   void *libopencl = nullptr;
+
+#if defined(_WIN32)
+  static const char *kClLibName = "OpenCL.dll";
+#else
   static const char *kClLibName = "libOpenCL.so";
+#endif
 
   libopencl =
     DynamicLibraryLoader::loadLibrary(kClLibName, RTLD_NOW | RTLD_LOCAL);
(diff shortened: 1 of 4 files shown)


dkjung:
I think we need to check if CI Windows build test passes with `-Denable-opencl = true` is given for `meson setup`. How about postpone merging this PR until then?

nntrainer/opencl/opencl_loader.cpp
```diff
@@ -44,7 +44,12 @@ bool LoadOpenCL() {
   }
 
   void *libopencl = nullptr;
+
+#if defined(_WIN32)
+  static const char *kClLibName = "OpenCL.dll";
+#else
```
dkjung, line 50: Thank you for the bug fix

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -14,8 +14,10 @@
 #include <blas_kernel_strings.h>
 #include <blas_kernels.h>
 
+#if !defined(_WIN32)
```
dkjung, line 17: We have to enable `clblast` for Windows, don't we?

dkjung (changes requested):
Let's discuss about this PR more face to face.

dkjung (commented):
Ah, this disables `clblast` for Windows.

dkjung:
I think I got to know that how to set dependency `clblast` on a Windows machine. Have a chat later.

PR #3232, opened 2025-06-03
[Quality] Fix all warnings reported by clang

gkisalapl:
This PR fix all warnings reported by clang compiler It will also ensure that continues integration will fail in case someone introduce any new warnings

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/quantizer.h
+++ b/nntrainer/tensor/quantizer.h
@@ -249,12 +249,7 @@ public:
   /**
    * @brief Basic Constructor of a PerChannelAffineQuantizer
    */
-  PerChannelAffineQuantizer() :
-    UniformQuantizer(),
-    scales(nullptr),
-    zero_points(nullptr),
-    quant_min(0),
-    quant_max(0) {}
+  PerChannelAffineQuantizer() : UniformQuantizer() {}
 
   /**
    * @copydoc Quantizer::create()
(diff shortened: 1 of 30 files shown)


nntrainer/tensor/quantizer.h
```diff
@@ -286,11 +281,6 @@ class PerChannelAffineQuantizer : public UniformQuantizer {
   QScheme qscheme() const override;
 
 private:
-  float *scales;
-  int *zero_points;
-  long int quant_min = 0;
-  long int quant_max = 0;
```
EunjuYang, line 292: How about allowing them to comment instead of removing them? These are not yet utilized because the quantizer is not yet implemented, even though they need to be declared.

EunjuYang (approved):
LGTM!
Thank you for this PR. It must have been an annoying and tedious task to fix this for clang compilation. Appreciate your contribution!

skykongkong8 (approved):
This will handle a bunch of potential coverity issues as well! 👍👍

nntrainer/tensor/quantizer.h
gkisalapl: ok I'll do it

gkisalapl: done

jijoongmoon (approved):
LGTM

PR #3234, opened 2025-06-04
[Tensor][Read] Change Tensor read from file_offset

DonghakPark:
## Dependency of the PR
- None

<details><summary>[Tensor][Read] Change Tensor read from file_offset</summary><br />

    [Tensor][Read] Change Tensor read from file_offset
    
    Previously, the tensor would receive parameters and, if the flag to read from the file offset was true, it would read from the passed start_offset.
    Now, since the tensor holds its own fileOffset, if the value is size_t::max, it will default to reading from the stored fileOffset.
    
    **Self evaluation:**
    
</details>

<details><summary>[Load] Refactoring Model Load Async</summary><br />

    [Load] Refactoring Model Load Async
    
    The existing method, which passed the offset through the execution order for each layer using a separate start_offset,
    has been modified to store and read the start_offset calculated by the FSU at the tensor level.
    
    **Self evaluation:**
    
</details>

### Summary

- The tensor now stores its own fileOffset and defaults to reading from this stored offset if the value is ```size_t::max.```
- After this pr, read will work as below 
```.cpp
read(std::ifstream &file, bool opt_var, ml::train::ExecutionMode mode, bool fsu, size_t start_offset = 0, bool read_from_offset = false)
```
**case 1. read(file, opt_var, exec_mode, fsu)** 
- will read from 0 to end

**case 2. read(file, opt_var, exec_mode, fsu,std::numeric_limits<size_t>::max(), true)**
- will read from tensor's fileOffset

**case 3. read(file, opt_var, exec_mode, fsu, 1050, true)**
- will read from 1050 to 1050 + Tensor's size

----

## Motivation & History of Change 

To help reviewers better understand, I will explain the motivation and background behind the changes.

**1. Sequential Loading Issue:**
- During inference, NNTrainer loaded weights by sequentially reading files based on their size. As modern AI models (especially LLMs) grew larger, this approach caused significant delays in loading large files. To address this, an **asynchronous loading** mechanism was introduced.

**2. Preloading Weight Sizes for Async Support:**
- To enable async loading, the starting point of each file segment needed to be determined beforehand. Thus, the system was modified to **pre-scan layer weights** and calculate their sizes before loading.

**3. Handling Quantization Info:**
- With support for diverse tensor types, certain cases (e.g., quantized tensors) required special handling due to embedded quantization metadata. This necessitated separate processing logic for such tensors.

**4. Offset Adjustment for Quantized Tensors:**
- For tensors with quantization info, the system first reads the metadata before creating/setting the tensor. The actual data is loaded afterward, requiring adjustments to the starting offset during the read operation.

**5. Dynamic LoRA Support:**
- While async loading improved speed, dynamic adaptations (e.g., LoRA) required tensors to **store their original offsets** instead of relying on runtime parameters. The default behavior was updated to read from these stored offsets.

**6. Flexible Offset Handling for LoRA:**
- To support LoRA, the read() function was modified to accept a **user-specified offset** when needed. Otherwise, it defaults to the tensor’s pre-stored offset for standard weight loading.

**Key Improvements:**
- **Faster loading** via async I/O and pre-scanning.
- **Enhanced flexibility** for quantized tensors and dynamic adaptations like LoRA.
- **Backward-compatible** offset management for both default and custom read operations.

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -671,12 +671,10 @@ void NeuralNetwork::load(const std::string &file_path,
   const std::regex reg_("\\s*\\:\\s*");
   auto v = split(file_path, reg_);
 
-  std::vector<size_t> start_offsets = {0};
-  std::vector<std::pair<size_t, size_t>> file_offset;
   size_t start_from = 0;
+  std::vector<std::pair<size_t, size_t>> file_offset;
   for (auto iter = model_graph.cbegin(); iter != model_graph.cend(); iter++) {
     auto weights = (*iter)->getRunContext().getWeights();
-    auto local_offset = start_offsets.back();
     for (auto weight : weights) {
       size_t size = weight->getVariable().getMemoryBytes();
       auto tensor_data_type = weight->getDim().getDataType();
@@ -694,9 +692,7 @@ void NeuralNetwork::load(const std::string &file_path,
       }
       file_offset.emplace_back(std::make_pair(start_from, size));
       start_from += size;
-      local_offset += size;
     }
-    start_offsets.push_back(local_offset);
   }
 
   if (exec_mode == ExecutionMode::INFERENCE && fsu_mode) {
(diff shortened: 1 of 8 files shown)


skykongkong8 (approved):
Nice work!
I think it would be a little bit more easier for other reviewers to understand if you attach the motivation of this change! I guess it would have came from some recent changes in variant Tensor classes?

DonghakPark:
> Nice work! I think it would be a little bit more easier for other reviewers to understand if you attach the motivation of this change! I guess it would have came from some recent changes in variant Tensor classes?

Okay !! that's good idea. i will add more detail history & motivation of this change at PR message

djeong20 (approved):

PR #3237, opened 2025-06-05
Windows opencl ci

gkisalapl:


djeong20:
Normally, when this happens, the output values are all zeros. 
1. Could you temporarily edit the `unittest_attention_kernels_cl.cpp` file to print output tensors?
2. Also, would it be possible to check the log file from the current build? It should be created under `builddir/logs` as `log_nntrainer_**.out`. If running `cat *.out` is possible after the build failure, we could definitely see what the problem is.

PR #3239, opened 2025-06-06
Enable OpenCL and CLBlast Support on Windows

gkisalapl:
This PR fix CLBlast build on windows for both msvc and clang
Fixing CLBlast build was complicated because prior this PR we were using static libraries for windows and static CLBlast was dependant on OpenCL library which we load in dynamic way.
To resolve this problem I unified build system of all cmake submodules and they're now linking in the same way as on linux


meson.build
```diff
@@ -169,21 +169,73 @@ if get_option('enable-mmap')
   extra_defines += '-DUSE_MMAP=1'
 endif
 
+dummy_dep = dependency('', required: false)
+
+# CLBlast
+clblast_dep = dummy_dep
+
```
djeong20:
```suggestion
```

I believe this can be removed

meson.build
```diff
@@ -169,21 +169,73 @@ if get_option('enable-mmap')
...
-    clblast_proj = cmake.subproject('clblast', options: clblast_options, required: true)
-    clblast_dep = clblast_proj.dependency('clblast')
+    clblast_dep = dependency('clblast', required : false)
+
+    if not clblast_dep.found()
+      message('Cannot find CLBlast dependency in system')
+      clblast_dir_relative = 'subprojects' / 'CLBlast'
+      clblast_include_dir_relative = clblast_dir_relative / 'include'
+      clblast_dir_absolute = meson.current_source_dir() / clblast_dir_relative
+      clblast_build_dir_relative = clblast_dir_relative / 'build'
```
djeong20: I'm considering whether it's a good idea to have two build directories, `build_msvc` and `build_clang`. This way, we wouldn't need to delete the existing build directory when switching compilers.

djeong20 (changes requested):
First of all, thank you for the awesome work! I was able to successfully compile CLBlast on both Windows and Linux. However, there's a runtime error, unfortunately. To explain, there's no CLBlast integration in the main branch, so there should be no runtime error. However, when I cherry-picked the commits from the PR that uses CLBlast (#3174 + #3239), there was an error. Without this patch (only #3174), it works perfectly fine. FYI, I ran `unittest_blas_kernels_cl`. We should resolve this before this PR is merged.

meson.build
gkisalapl: I was also thinking about it but decided to leave one build directory to save disc space. Please discuss it with the Team and let me know which option you preffer

meson.build
gkisalapl: We're following this pattern for all other dependencies

gkisalapl:
I've done following test:
1. Switch to #3239
2. Cherry-pick #3174 
3. Build project on windows (using clang) and everything goes well with the build

commit 70c0ebe9e715c235bf72f7c603e69fc8b6c50486 (HEAD -> windows_clblast)
Author: Donghyeon Jeong <dhyeon.jeong@samsung.com>
Date:   Mon May 19 10:12:24 2025 +0900

    Fix Android Packaging for CLBlast: Ensuring Static Library Build

    This commit addresses the Android packaging issue in CLBlast, ensuring that CLBlast is correctly packaged as a static library. Please note that to successfully build CLBlast, you will need an Android OpenCL shared object and headers. The necessary files can be downloaded from the NNStreamer Android resources.

    **Self-evaluation:**

    Signed-off-by: Donghyeon Jeong <dhyeon.jeong@samsung.com>

commit daab87aa5665adea24a811e68105cec0db0a3cb8 (origin/windows_clblast)
Author: Grzegorz Kisala <gkisala@gmail.com>
Date:   Fri Jun 6 12:15:20 2025 +0200

    Enable OpenCL and CLBlast Support on Windows

    This PR fix CLBlast build on windows for both msvc and clang

    **Self-evaluation:**

    Signed-off-by: Grzegorz Kisala <gkisala@gmail.com>

meson.build
djeong20: I mean, it looks like the build should work without this code. I don't see why this is necessary. let me know if I'm wrong

meson.build
djeong20: sounds fair :)

meson.build
gkisalapl: you're right I've added it here because it's done for other subproject and by doing this we can avoid doing this:

if get_option('enable-opencl')
  nntrainer_base_deps += clblast_dep
endif

PR #3241, opened 2025-06-09
Remove leftover unused function

mwlasiuk:
- Removed dot(q6, float) copy that was never used because there is second function doing same job that is acctually called

**Self evaluation:**


djeong20 (approved):

skykongkong8 (approved):
This function is indeed unused in CPU, but I heard similar behavior occurs on GPU side.
It can be removed for now, but there might be a 'possibility' to re-introduce this function for debugging purpose.

PR #3253, opened 2025-06-13
Tensor/topk

EunjuYang:
## Dependency of the PR
None

<details><summary>[Tensor] implement topK op for float tensor</summary><br />

This commit introduces the `topK` function to compute the top-K maximum values along the width dimension for each batch, channel, and height slice. The results are returned as separate tensors for values and indices, preserving the original tensor format (NCHW/NHWC).

Key changes:
- Added `topK` method to `Tensor` class in `tensor.h/cpp`
- Updated `TensorBase` class in `tensor_base.h/cpp` to support virtual `topK` method
- Implemented `topK` logic in `FloatTensor` class in `float_tensor.h/cpp`

Notes:
- Indices are returned as a separate tensor of type `UINT32`
- Parallelized using OpenMP for batch-channel-height processing
- Throws `std::invalid_argument` for invalid inputs (e.g., k=0, unsupported data types)

**Self evaluation:**

</details>

<details><summary>[Tensor] Add comprehensive unit tests for top-K functionality</summary><br />

This commit adds a suite of unit tests to validate the correctness and robustness of the `topK` functionality across various scenarios. The tests cover basic usage, multi-batch, multi-channel, edge cases, and unsupported data types.

- Key changes:
- Added tests for basic NHWC and NCHW formats
- Added tests for multi-batch and multi-channel scenarios
- Added tests for edge cases (e.g., k equals width, invalid k values)
- Added tests for unsupported data types
**Self evaluation:**

</details>

### Summary

- Implement topK operation : interface in tensor, tensor_base ; actual implementation for float_tensor
- Implement unit test cases for topK operation


nntrainer/tensor/float_tensor.cpp
```diff
@@ -881,6 +881,74 @@ std::vector<unsigned int> FloatTensor::argmin() const {
...
+            input_idx = b * input_strides[0] + h * input_strides[1] +
+                        w * input_strides[2] + c;
+          }
+          elements[w] = {input_buffer[input_idx], w};
+        }
+
+        // Sort first k elements in descending order
+        std::partial_sort(
+          elements.begin(), elements.begin() + k, elements.end(),
+          [](const auto &a, const auto &b) { return a.first > b.first; });
```
SeoHyungjun: Is it okay to disregard the second of elements?

EunjuYang: Strictly speaking, it doesn’t need to consider it (any of the topK is acceptable—if the values are the same, it’s okay to pick any of them). However, it seems worthwhile to consider the indices' values during sorting if we need to provide a consistent index order for the same value.
Do you think this should be supported by default?

baek2sm: Why do we need to consider the indices during sorting?
I think we can only use first element(logit) for picking top-k words (the current implementation seems to work well).

baek2sm (approved):
LGTM

nntrainer/tensor/float_tensor.cpp
SeoHyungjun: If the value of first is the same, it can be used regardless of second, right?
In that case, I think the current formula is sufficient!
I was asking because I wasn’t entirely sure.

SeoHyungjun (approved):

nntrainer/tensor/float_tensor.cpp
EunjuYang: Thank you for the clarification. I’ll make a note of it and consider adding an option based on your suggestion as part of the follow-up work.

nntrainer/tensor/float_tensor.cpp
```diff
@@ -881,6 +881,74 @@ std::vector<unsigned int> FloatTensor::argmin() const {
...
+            input_idx = b * input_strides[0] + c * input_strides[1] +
+                        h * input_strides[2] + w;
+          } else {
+            // NHWC: [b][h][w][c]
+            input_idx = b * input_strides[0] + h * input_strides[1] +
+                        w * input_strides[2] + c;
+          }
+          elements[w] = {input_buffer[input_idx], w};
+        }
+
```
jijoongmoon, line 920: instead of this.. creating new array copy.. just playing with index is good enough I think.
```c++
Tensor B = getSharedDataTensor(TensorDim({1,1,1,width, getTensorType()}, b*input_strides[0]+c*input_strdies[1]+ h*input_strides[2], true);
std::nth_element(idx.begin(), index.begin()+k, indx.end(), [*](size_t a, size_t b){ 
    return B[a] > B[b];
});
for(int i=0;i<k;++i)
   value[i] = B[idx[i]];
```

EunjuYang: It seems much better. I’ll proceed with updating the patch in this direction.

skykongkong8 (approved):

myungjoo:
https://chatgpt.com/share/686b34d8-67e4-800f-aa11-33950141039d

Just for ur reference

myungjoo (approved):
LGTM.

jijoongmoon (approved):
LGTM

PR #3255, opened 2025-06-17
Update OpenCL headers

mwlasiuk:
- Remove old OpenCL headers directory (replace with submodules for C and C++ bindings)
- Add new OpenCL headers [C / C++] as submodules
- Update include path to use <> for OpenCL
- Use fixed target version of 300 for now

**Self evaluation:**


mwlasiuk:
> Can you explain the background of this update? Why is it necessary to update the OpenCL headers?

- Using OpenCL C/C++ submodules from Khronos is the easiest way to provide headers and give possibility to control their version. 
- Using OpenCL C++ bindings will make it easier to implement platform / device enumeration (less code and more readable code) which is required for Piotr to implement features needed for SVM discussed in [this](https://github.com/nnstreamer/nntrainer/discussions/3256) thread and other means of communication.
- Using newest headers gives us more control over OpenCL loading in future (using ICD)

djeong20:
> > Can you explain the background of this update? Why is it necessary to update the OpenCL headers?
> 
> * Using OpenCL C/C++ submodules from Khronos is the easiest way to provide headers and give possibility to control their version.
> * Using OpenCL C++ bindings will make it easier to implement platform / device enumeration (less code and more readable code) which is required for Piotr to implement features needed for SVM discussed in [this](https://github.com/nnstreamer/nntrainer/discussions/3256) thread and other means of communication.
> * Using newest headers gives us more control over OpenCL loading in future (using ICD)

I doubt that this is a good choice when considering Android.
In NNTrainer, header files are installed for the Android build. (check `builddir/android_build_result/include/nntrainer`)
This patch would require installing the headers of two subprojects, and I don't think it is a good idea.
I also see CI/CD failure for the Android build, and this is the main reason why it fails.

djeong20:
#3258 would be enough.

meson.build
```diff
@@ -171,7 +171,14 @@ endif
 
 if get_option('enable-opencl')
   message ('OpenCL build is now enabled and will function only if an OpenCL-supported GPU is available. Using CLBlast as the OpenCL backend.')
+
   extra_defines += '-DENABLE_OPENCL=1'
+
+  extra_defines += '-DCL_TARGET_OPENCL_VERSION=300'
+  extra_defines += '-DCL_MINIMUM_OPENCL_VERSION=300'
+  extra_defines += '-DCL_HPP_TARGET_OPENCL_VERSION=300'
+  extra_defines += '-DCL_HPP_MINIMUM_OPENCL_VERSION=300'
```
djeong20, line 180: This should be changed. Some target devices (e.g., Tizen) do not support OpenCL 3.0. That's why we target OpenCL 2.0.

djeong20 (changes requested):
I get an error when I test this PR with `./tools/android_test.sh`

```
/home/donghyeon/nntrainer/nntrainer/opencl/opencl_program.h:19:10: fatal error: 'CL/opencl.hpp' file not found
   19 | #include <CL/opencl.hpp>
      |          ^~~~~~~~~~~~~~~
1 error generated.
```
Please verify the unit test on the Android device.

gkisalapl:
> nghyeon's comment is critical.
> Please check his comm

As discussed during our meeting let's move on with #3258 this week, once we fix Android we will swap to implementation from this PR

PR #3254, opened 2025-06-16
Tensor/get batch slice with indices

EunjuYang:
## Dependency of the PR
None

## Commits to be reviewed in this PR

<details><summary>[Tensor] Add batch slicing for a given indices vector</summary><br />

- Implement `Tensor::getBatchSlice(std::vector<unsigned int>) for MoE
  layer token routing (MoE token routing is irrelevant with this patch
though)
- Validate tensor contiguity before slicing operations
- Add dimension checks for out-of-range indices
- Please note that this operation does not use sharedFrom but do the
  copy.
- This operation would be used in scatter

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

</details>

<details><summary>[Tensor] Add unittest for getBatchSlice</summary><br />

- Verify FP32/UINT16 data type handling
- Add positive test cases:
   - Single/Multi-batch selection
   - Large tensor stress test
- Add negative test cases
   - out of range index validation
- Include parallel execution stability test
   - 100x concurrent access check

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

</details>

### Summary

- This PR aims to implement `getBatchSlice(std::vector<unsigned int> indices)` and create its unit test cases
- This PR's operation is implemented to support scatter operation in MoE layer
- Please also see #3252 

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

PR #3257, opened 2025-06-17
[Feat] MoE layer (inference mode only)

EunjuYang:
## Dependency of the PR

#3253 
#3254 

<details><summary>[MoE] Implement MoE Layer</summary><br />

- This patch implements moe layer.
- This implementation does not support backwarding yet.

**Self evaluation:**

</details>

### Summary

- Please note that this is a very naive version of MoE layer.
- It only implements inference part and construct MoE action following the procedure of python code.
- This implementation is slow and have memory redundancy due to getBatchSlice with memory copy.
- This implementation will be updated in another PR.
- This patch can be used for the model construction purpose. (inference only)

### DIFF
--- a/nntrainer/layers/moe_layer.cpp
+++ b/nntrainer/layers/moe_layer.cpp
@@ -34,8 +34,16 @@ static constexpr size_t SINGLE_INOUT_IDX = 0;
 
 MoELayer::MoELayer() :
   LayerImpl(),
+  num_experts(0),
+  topk(0),
   moe_props(props::NumExperts(), props::NumExpertsPerToken(), props::Unit(),
-            props::MoEActivation()) {}
+            props::MoEActivation()),
+  expert_gate_proj_indices({}),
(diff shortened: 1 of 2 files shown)


nntrainer/layers/moe_layer.cpp
```diff
@@ -0,0 +1,297 @@
...
+  ///@todo acti_func.run_fn(gate_out, gate_out); -> this inplace operation
+  /// doesn't work
+  acti_func.run_fn(gate_out, acti_out);
+
+  // Up projection: [tokens,1,1,H] x [H,I] -> [tokens,1,1,I]
+  Tensor up_out(tokens, 1, 1, intermediate_size);
+  input.dot(up_proj, up_out);
+
+  // Multiply: silu(gate_out) * up_out
+  acti_out.multiply_i(up_out);
```
EunjuYang, line 256: These can be replaced with `nntrainer::swiglu()`, which is implemented in backend!

PR #3265, opened 2025-06-26
[OpenCL] Enable SVM for SGEMV Kernel

djeong20:
This patch updates FloatTensor to allocate SVM (Shared Virtual Memory) for unmanaged memory.
This change enables the use of SVM in the SGEMV (Single Precision General Matrix-Vector Multiplication) kernel.

**Self-evaluation:**

#### 1. Linux-x86 (NVIDIA GeForce RTX 1650)
| |Time|
|------|---|
|Before|47 ms|
|After|19.15 ms|

#### 2. Linux-x86 (Intel(R) UHD Graphics 770)
| |Time|
|------|---|
|Before|89.7 ms|
|After|45.25 ms|

#### 3. Android (Qualcomm Adreno 740)
| |Time|
|------|---|
|Before|145.8 ms|
|After|108.85ms|

## Note

The following matrix sizes are used for the benchmark and profile.
- M: 1
- K: 3072
- N: 21180

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
@@ -116,8 +116,8 @@ void dotCl(Tensor const &input, Tensor const &m, Tensor &result, bool trans,
     (input.getFormat() == Tformat::NHWC) ? result.channel() : result.width();
 
   if (input.getDataType() == ml::train::TensorDim::DataType::FP32) {
-    const float *data = input.getData();
-    const float *mdata = m.getData();
+    float *data = input.getData();
+    float *mdata = m.getData();
     float *rdata = result.getData();
 
     /// shortcut handling in case of vector
--- a/nntrainer/tensor/cl_operations/blas_kernels.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernels.cpp
@@ -204,10 +204,8 @@ void sgemv_q6_k_cl(void *matAdata, float *vecXdata, float *vecYdata,
   }
 }
 
-void sgemv_cl(const float *matAdata, const float *vecXdata, float *vecYdata,
-              bool TransA, unsigned int dim1, unsigned int dim2,
-              unsigned int lda) {
-
+void sgemv_cl(float *matAdata, float *vecXdata, float *vecYdata, bool TransA,
+              unsigned int dim1, unsigned int dim2, unsigned int lda) {
   bool result = false;
 
   auto *blas_cc =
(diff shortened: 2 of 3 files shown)


skykongkong8 (approved):
Nice work!

nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
```diff
@@ -116,8 +116,8 @@ void dotCl(Tensor const &input, Tensor const &m, Tensor &result, bool trans,
     (input.getFormat() == Tformat::NHWC) ? result.channel() : result.width();
 
   if (input.getDataType() == ml::train::TensorDim::DataType::FP32) {
-    const float *data = input.getData();
-    const float *mdata = m.getData();
+    float *data = input.getData();
+    float *mdata = m.getData();
```
dkjung, line 120: Does it make an error if `const` is not removed?

nntrainer/tensor/cl_operations/blas_kernels.h
```diff
@@ -40,9 +40,8 @@ static ClBufferManager &clbuffInstance = ClBufferManager::getInstance();
  * @param[in] lda number of X's columns
  * @param[in] context RunLayerContext reference
  */
-void sgemv_cl(const float *matAdata, const float *vecXdata, float *vecYdata,
-              bool TransA, unsigned int dim1, unsigned int dim2,
-              unsigned int lda);
+void sgemv_cl(float *matAdata, float *vecXdata, float *vecYdata, bool TransA,
+              unsigned int dim1, unsigned int dim2, unsigned int lda);
```
dkjung, line 50: `vecYdata` is the output parameter of the function and its content is to be overwritten, while `matAdata`, `vecXdata` are read-only.

Therefore, It is allowed to put `const` to `matAdata`, `vecXdata`, isn't it?

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -98,8 +85,8 @@ void sgemv_cl(const float *matAdata, const float *vecXdata, float *vecYdata,
       break;
     }
 
-    result = clbuffInstance.getOutBufferA()->ReadDataRegion(
-      blas_cc->command_queue_inst_, dim1_size, vecYdata);
+    result =
+      blas_cc->command_queue_inst_.enqueueSVMMap(vecYdata, dim1_size, true);
```
dkjung, line 279: FYI
I did ask chatGPT about `enqueueSVMMap`, `enqueueSVMUnmap`.

1. clEnqueueSVMMap:
   - Purpose: This function is used to map a region of SVM for the host to access. It ensures that any pending operations on the device involving this memory are completed before the host accesses it.
   - When to Invoke:
     - Before the host reads from or writes to an SVM buffer that has been used by a kernel on the device.
     - When you need to ensure consistency of data between the host and device after kernel execution.
     - Example: If a kernel writes results to an SVM buffer, and you want to read these results on the host, you should call clEnqueueSVMMap before accessing the buffer.

2. clEnqueueSVMUnmap:
   - Purpose: This function unmaps a previously mapped SVM region, allowing the device to access it again.
   - When to Invoke:
     - After the host has finished accessing an SVM buffer that was previously mapped with clEnqueueSVMMap.
     - Before enqueuing any further commands that involve the device accessing this memory.
     - Example: Once you've read or modified data in an SVM buffer on the host, call clEnqueueSVMUnmap to allow subsequent kernels to access the updated data.

dkjung:
Which unit test did you run to measure time? Can you share its name?

nntrainer/tensor/cl_operations/blas_kernel_interface.cpp
djeong20: Yes, if the pointer is const, it throws an error during map/unmap.

nntrainer/tensor/cl_operations/blas_kernels.h
djeong20: ditto

PR #3270, opened 2025-06-27
[Fix] Fix checking for negative inf when validating float numbers

gkisalapl:
When we're checking if float / half float value is valid we should consider also checking for negative infinity
This PR introduce common function which check float / half float validity


EunjuYang (approved):
LGTM

baek2sm (approved):
LGTM

DonghakPark (approved):
LGTM!

skykongkong8 (approved):

PR #3284, opened 2025-07-03
[Fix] Fix nan values problem while beta calculation.

gkisalapl:
This PR address following issues:

1. Workaround for problem with nan when calculating beta * nan

In many functions we're calculating:

Z[i] = X[i] - alpha * Y[i] + beta * Z[i];

if beta is zero we're expecting to have 0 as results of beta * Z[i] but it's not always true because when Z is not initialized it can be nan. This PR add workaround for this and fix problem when beta is 0

2. Fix problem of not initialized tensors in tests (model_test_utils.cpp) - we were running tests with uninitialized memory in tensors and expect zeros (which was not always true). This PR initialize tensors memory

### DIFF
--- a/nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
+++ b/nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
@@ -26,7 +26,10 @@
     float y0;                                                                  \
     unsigned int i, j;                                                         \
     for (ci = 0; ci != cM; ci++) {                                             \
-      y0 = Y[ci * incY] * beta;                                                \
+      y0 = 0.0f;                                                               \
+      if (beta != 0.0f) {                                                      \
+        y0 = Y[ci * incY] * beta;                                              \
+      }                                                                        \
       for (cj = 0; cj != cN; cj++)                                             \
         y0 += A[i + j * lda] * X[cj * incX];                                   \
       Y[ci * incY] = y0;                                                       \
@@ -198,8 +201,9 @@ void __fallback_sgemm(const unsigned int TStorageOrder, bool TransA,
         c += a * b;
       }
       C[m * ldc + n] = alpha * c;
-      if (beta != 0.0)
+      if (beta != 0.0f) {
         C[m * ldc + n] += beta * c_old;
+      }
     }
   }
 }
@@ -258,7 +262,7 @@ void __fallback_ele_mul(const unsigned int N, const float *X, const float *Y,
                         float *Z, float alpha, float beta,
                         unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X * alpha * *Y + beta * *Z;
+    *Z = *X * alpha * *Y + ((0.0f == beta) ? 0.0f : beta * *Z);
     X += o_stride;
     Y += i_stride;
     Z += o_stride;
@@ -269,7 +273,7 @@ void __fallback_ele_add(const unsigned int N, const float *X, const float *Y,
                         float *Z, float alpha, float beta,
                         unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X + alpha * *Y + beta * *Z;
+    *Z = *X + alpha * *Y + ((0.0f == beta) ? 0.0f : beta * *Z);
     X += o_stride;
     Y += i_stride;
     Z += o_stride;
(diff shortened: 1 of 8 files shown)


nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
```diff
@@ -269,7 +274,8 @@ void __fallback_ele_add(const unsigned int N, const float *X, const float *Y,
                         float *Z, float alpha, float beta,
                         unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X + alpha * *Y + beta * *Z;
+    *Z = *X + alpha * *Y + ((0.0f == beta) ? 0.0f : beta * *Z);
+    ;
```
mwlasiuk: Remove additional semicolon

nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
```diff
@@ -291,7 +298,8 @@ void __fallback_ele_div(const unsigned N, const float *X, const float *Y,
                         float *Z, float alpha, float beta,
                         unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X / (alpha * *Y) + beta * *Z;
+    *Z = *X / (alpha * *Y) + ((0.0f == beta) ? 0.0f : beta * *Z);
+    ;
```
mwlasiuk: Remove additional semicolon

nntrainer/tensor/cpu_backend/fallback/fallback_internal_fp16.cpp
```diff
@@ -238,7 +247,10 @@ void __fallback_ele_sub(const unsigned N, const _FP16 *X, const _FP16 *Y,
                         _FP16 *Z, float alpha, float beta,
                         unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X - static_cast<_FP16>(alpha) * *Y + static_cast<_FP16>(beta) * *Z;
+    *Z = *X - static_cast<_FP16>(alpha) * *Y +
+         ((0.0f == beta) ? static_cast<_FP16>(0.0f)
+                         : static_cast<_FP16>(beta) * *Z);
+    ;
```
mwlasiuk: Remove

nntrainer/tensor/cpu_backend/fallback/fallback_internal_fp16.cpp
```diff
@@ -249,7 +261,10 @@ void __fallback_ele_div(const unsigned N, const _FP16 *X, const _FP16 *Y,
                         _FP16 *Z, float alpha, float beta,
                         unsigned int i_stride, unsigned int o_stride) {
   for (unsigned int i = 0; i < N; ++i) {
-    *Z = *X / (static_cast<_FP16>(alpha) * *Y) + static_cast<_FP16>(beta) * *Z;
+    *Z = *X / (static_cast<_FP16>(alpha) * *Y) +
+         ((0.0f == beta) ? static_cast<_FP16>(0.0f)
+                         : static_cast<_FP16>(beta) * *Z);
+    ;
```
mwlasiuk: Remove

mwlasiuk (changes requested):

mwlasiuk (approved):

skykongkong8 (approved):

nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
```diff
@@ -26,7 +26,10 @@
     float y0;                                                                  \
     unsigned int i, j;                                                         \
     for (ci = 0; ci != cM; ci++) {                                             \
-      y0 = Y[ci * incY] * beta;                                                \
+      y0 = 0.0f;                                                               \
+      if (beta != 0.0f) {                                                      \
```
myungjoo, line 30: 1. This doesn't recognize -0.0f (0.0f != -0.0f) (ref: https://en.cppreference.com/w/cpp/numeric/math/FP_categories )
2. Do you want to ignore round off errors?

gkisalapl: My understanding is that beta parameter is not result of some previous computations but rather value set to 0.0f or 1.0f when we call particular functions (for example in dot operation). So I do not think that we should do more advanced checks like checking for -0.0 or checking if beta is gather than some epsilon.

myungjoo (approved):

jijoongmoon (approved):
LGTM

PR #3299, opened 2025-07-08
Replace repeating FP16/FP32 code with template

mwlasiuk:
Replace repeating FP16/FP32 code with template

- These kernels have exactly the same structure with only thing that differs is used kernel (its registered name) and type dependent sizeof()-s

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernels_templates.h
+++ b/nntrainer/tensor/cl_operations/blas_kernels_templates.h
@@ -1,0 +1,446 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Debadri Samaddar <s.debadri@samsung.com>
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/cl_operations/blas_kernels_templates.h
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  result = clbuffInstance.getOutBufferA()->ReadDataRegion(
+    blas_cc->command_queue_inst_, dim_size, res);
+  if (!result) {
+    return;
+  }
+}
+
+} // namespace nntrainer
```
skykongkong8: Check EOF. You should leave one blank line at the end of the file for format

djeong20 (approved):
Nice :)

PR #3302, opened 2025-07-08
[Layer] Unsqueeze Layer Added

abhinav-nn:
Co-authored-by: Abhinav Dwivedi abhinav.d@samsung.com
Co-authored-by: Sachin Singh sachin.3@samsung.com

Added unsqueeze layer based on torch implementation.

finalize, forwarding, calcDerivative and setProperty are implemented.

Used props::Original_ndim as a number of dimensions in its input. This prop has to be passed correctly during interpreting onnx.

**Self evaluation:**

Signed-off-by: {Abhinav Dwivedi} <{abhinav.d@samsung.com}>

</details>


test/input_gen/genModelTests_v2.py
```diff
@@ -590,7 +590,17 @@ def forward(self, inputs, labels):
         loss = self.loss(out, labels[0])
         return out, loss
 
+class UnsqueezeOperation(torch.nn.Module):
```
baek2sm, line 593: Is it correct that you couldn't add the model unit test because of the input data type(integer type)? I haven't reviewed the model unit-test part yet. So if possible, it would be great if you could work on modifying the model unit test to support integer types!

abhinav-nn: No that was not the case. We just checked that all the models defined in the unit test were set for float inputs.
One thing that we cant do is add tests for negative axis as the Axis property takes unsigned int. So negative access support isn't there yet.
Generic axis prop should be changed to take both positive and negative values.

baek2sm: @abhinav-nn I am also aware that the current `model unittest` has been designed to accept only float inputs. It is likely that the existing structure does not support integer inputs. Therefore, I would like to request an improvement to the `model unittest system` itself to support integer types! This task does not need to be completed within this PR; I would appreciate it if you could submit it as a separate PR.

abhinav-nn: Sure we will modify the current unittest system to accept int inputs as well and will submit another PR for that as soon as possible. Meanwhile please review this PR (op implementation part) and merge.

nntrainer/layers/unsqueeze_layer.cpp
```diff
@@ -0,0 +1,95 @@
...
+                std::invalid_argument)
+    << "Unsqueeze layer must have axis in range [" << (-1 - input_ndim) << ","
+    << (input_ndim) << "] for input_ndims " << input_ndim << " ."
+    << " Your axis is " << input_axis;
+
+  if (axis < 0) {
+    axis.set(input_axis + input_ndim + 1); // negative axis handling
+  }
+
+  axis.set((in_dim.getNumDim() + input_axis - input_ndim -
```
baek2sm, line 51: Is it correct to always overwrite it with the negative axis condition right above it? Also, it seems like the axis variable is being set but not actually used for setting out_dim.

EunjuYang: I agree with his comment.

nntrainer/layers/unsqueeze_layer.cpp
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2020 Jijoong Moon <jijoong.moon@samsung.com>
```
baek2sm, line 3: copyright?

nntrainer/layers/common_properties.h
```diff
@@ -356,6 +356,28 @@ class Axis : public nntrainer::PositiveIntegerProperty {
...
+  /**
+   * @brief check if given value is valid
+   *
+   * @param v value to check
+   * @retval true if it is greater equal to 0 and smaller than
+   * ml::train::TensorDim::MAXDIM
+   * @retval false if it is smaller than 0 or greater than
+   * ml::train::TensorDim::MAXDIM
+   */
+  // bool isValid(const unsigned int &value) const override;
```
baek2sm, line 378: If it's not needed, it's better to delete it..

PR #3304, opened 2025-07-08
[Layer] Add Neg layer

sumon-98:
Added Neg Tensor op, layer implementation and unit tests.

Commits:
    [Tensor] Add Neg tensor operation
    [LAYER] Add Neg layer
    [TEST] Add unittest for Neg layer

Self evaluation:
    1. Build test:  [x] Passed [ ] Failed [ ] Skipped
    2. Run test:    [x] Passed [ ] Failed [ ] Skipped

Signed-off-by: Sumon Nath <[sumon.nath@samsung.com](mailto:sumon.nath@samsung.com)>


skykongkong8 (commented):
Solid layer PR with unittest included! I think this is nicely done overall!
Please check the clang-format though.

nntrainer/tensor/tensor.cpp
```diff
@@ -841,6 +841,22 @@ Tensor &Tensor::sqrt(Tensor &output) const {
...
+
+Tensor &Tensor::neg(Tensor &output) const {
+  if (size() != output.size() || getDataType() != output.getDataType() ||
+      getFormat() != output.getFormat())
+    throw std::invalid_argument(
+      "Error: Tensor::sqrt requires output tensor to be same size, data type "
+      "and format as input tensor.");
+
+  itensor_->neg(output);
+  return output;
```
EunjuYang: Instead of implementing `neg` for tensorbase and every tensor types, what about just calling multiply?
For example, 
```suggestion
  itensor_->multiply(-1, output);
  return output;
```

EunjuYang (commented):
What about implementing tensor's neg opration simply callying multiply instead of implementing every neg operation for base tensors?

nntrainer/tensor/tensor.cpp
sumon-98: Yes 100%, we could do that. 
However, I believe negation will be more efficient than multiplying -1. In addition to that, the tensor op might be useful in future in terms of ease of use. 
Let me know what you think. Thanks!

EunjuYang: If the `neg` operation (i.e., `-in`) is significantly faster, I completely agree with creating a dedicated `neg` function. However, if the performance improvement is marginal, it might be more efficient to use multiply instead. I believe the compiler could optimize it effectively.

EunjuYang: By the way, I agree with implementing the neg operation for tensors. However, my suggestion is to reuse the multiply operation instead of implementing the neg operation for tensor_base and its derived tensor types if the impact is marginal :)

sumon-98: I have updated the commit with your suggested changes. I compared the test results for both neg and multiply (-1) and they came out to be comparable for a large tensor. Thanks for the suggestion :)

skykongkong8 (approved):

EunjuYang (approved):
LGTM

PR #3307, opened 2025-07-08
[WIP] Add backend operations for multi head attention layer

mnalewaj:
Do not marge it.
In this PR:
- Implementation several backend operations for multi head attention layer.
- Refactoring AVX2 implementation.
- Porting AVX2 to NEON.
- Not finished yet.

In the next PR I will implement FP16 version on NEON.


nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -908,4 +909,417 @@ void exp_i(const unsigned int N, float *X) {
...
+  y = vfmaq_f32(c0, y, r); // final y = (((...r+...)*r+...)*r + 1)
+
+  // Reconstruct exp(x) = 2^n * exp(r)
+  int32x4_t emm0 = vcvtq_s32_f32(fx); // There are other options of conversion
+  emm0 = vaddq_s32(emm0, vdupq_n_s32(127));
+  emm0 = vshlq_n_s32(emm0, 23);
+  float32x4_t pow2n = vreinterpretq_f32_s32(emm0);
+
+  return vmulq_f32(y, pow2n);
+}
```
skykongkong8: I think nntrainer already has this kind of function. Is there any significant difference compared to `exp_ps()` from `neon_mathfun.h`? If not, I think this code can be removed.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -908,4 +909,417 @@ void exp_i(const unsigned int N, float *X) {
...
+      float32x4_t b = vld1q_f32(&inout[i1]);
+
+      if (only_convert_to_fp16) {
+        if (out_type == OutputType::FP16) {
+          float16x4_t a_fp16 = vcvt_f16_f32(a);
+          float16x4_t b_fp16 = vcvt_f16_f32(b);
+
+          vst1_f16(static_cast<float16_t *>(output) + i0, a_fp16);
+          vst1_f16(static_cast<float16_t *>(output) + i1, b_fp16);
+        }
```
skykongkong8: I'm curious about `float16_t`. NNTrainer has been using `__fp16` instead of this. Why did you choose to use this type?

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -908,4 +909,417 @@ void exp_i(const unsigned int N, float *X) {
...
+  }
+  for (; i < chunk_size; ++i) {
+    temp_row[i] = convert_scalar(b_src[i]);
+  }
+}
+
+template <>
+void compute_kcaches(const float *A, const uint16_t *B, float *output,
+                     int num_rows, int N, int chunk_size, int group_size,
+                     int tile_size, bool process_all) {
```
skykongkong8: Although this is the correct way to implement for now, (choosing `uint16_t` as B type) we might consider using `__fp16` for this in arm in the future.

skykongkong8 (commented):
Seems everything is going well! But please check the following comments👍

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
mnalewaj: The types float16_t and __fp16 have the same binary representations.
They are often used interchangeably.
Compiler can compile both type names.
So I replaced float16_t to __fp16.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
mnalewaj: The method exp_ps() looks similar but is has a little other implementation.
I have checked both methods in unit tests and they both give acceptable results.
So I removed exp256_f32x4 method.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
mnalewaj: Yes, you are right. 
Currently I leave it as is to have the same api in AVX2 and NEON implementation.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -578,4 +580,502 @@ void ele_add(const unsigned int N, const float *X, const float *Y, float *Z,
...
+          for (; i < chunk_size; ++i)
+            sum += a_ptr[i] * b_row[i];
+
+          output[(row_tile_start + row) * N * group_size + n * group_size + g] =
+            sum / sqrt((float)chunk_size);
+        }
+      }
+    }
+  }
+}
```
Paaaark: `process_all` is not used in the method. We should leave that as a todo.

mnalewaj: I'm not sure what you mean.
I took this method from a custom layer, so I don't know if this parameter will be used in the future.
Do you know or someone else knows if the parameter `process_all` will be used in the future?
If yes should I @todo comment in doxygen header for this method?

Paaaark: If I recall correctly, during inference, we would be using `process_all = false` to prevent recomputing the kcache all over again.

mnalewaj: I have analyzed this method and usage of it and I see that parameter `process all` it is not needed.
I every usage of this method computes only one row. Probably the parameter will be not used in the future.
Thus I removed this parameter.

Paaaark: Hello! I reviewed the concept and realized I made a mistake in my previous comment.

I now get what you mean by it computing only one row. That is the expected behavior of attention score during inference. However, during training, we would want to compute attention score for each K against all Q, effectively increasing the dimension of the output to `num_rows * num_rows * N * group_size` or `(num_rows * (num_rows + 1)) / 2 * N * group_size` if you remove repeated computations. I think this should be left as a @todo to implement that case.

mnalewaj: Hello! I still have some doubts about this.

From the point of view of the user of this method during inference, the pseudo code looks like this:
```
CustomMHACoreV2Layer::compute_kcaches(..., uint32_t from, uint32_t to, ...) {
	if (from == 0) { // Prefill stage:
		PARALLEL:
		for (r in range(from, to)) {
			nntrainer::compute_kcaches<>(..., r, ...); <<--- compute one row
		}
	} else { // Generation stage:
		nntrainer::compute_kcaches<>(..., from + 1, ...); <<--- compute one row (to - from == 1)
	}
}
```
If my understanding is correct, I think a similar situation occurs during training.
I am not sure if I am right, so could you confirm that?
If yes the parameter `process_all` is not needed in the future.

Paaaark: Oh I see so nntrainer::compute_kcaches is used as a helper function for the custom layer compute_kcaches function. If that is the case, I think the code now makes sense :)

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -371,4 +371,57 @@ void repack_q4_K_to_q4_K_8(void *W, void *repacked_W, size_t data_size,
...
+void compute_fp16vcache_fp32_transposed(int iter, const float *in,
+                                        const uint16_t *vcache, float *output,
+                                        int seq, int num_cache_head,
+                                        int gqa_size, int head_dim,
+                                        bool process_all) {
+#ifdef ARMV7
+  __fallback_compute_fp16vcache_fp32_transposed(iter, in, vcache, output, seq,
+                                                num_cache_head, gqa_size,
+                                                head_dim, process_all);
+#else
```
skykongkong8: nntrainer does not enjoy adding #ifdefs like `ARMV7` in the `*.cpp` source code. There is file named `armv7_neon.cpp ` I think is is better for code maintenance, to move this functionality to there.

And I wonder why did you separate this code in the first place becasue:
1. if you implemented  vcache in a `uint16_t` manner -> I think armv7l also has neon intrinsics to handle `uint16x8_t` and `float32x4_t` so it does not have to be separated with the current implementation.
2. if you implemented vcache in a `__fp16` manner -> armv7 does not support `__fp16` in the first place, so we also do not expect it to be computed in that way. In this case, we rather put this kernels in the `neon_impl_fp16.cpp` to manage `__fp16`-related code in a separate file, for easier maintenance and build.
3. if the reason why you separated this code is because there are no certain intrinsics that `armv7l` does not  support, we are currently just implementing them in a naive manner in `armv7_neon.cpp`, so please take a look. I think it is quite straight-forward.

Personally I believe, if you have both 1 & 2 implemenations, how about just adding two different functions like 
`compute_fp16vcache_fp32_transposed()` and `compute_u16vcache_fp32_transposed` ?

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -371,4 +371,57 @@ void repack_q4_K_to_q4_K_8(void *W, void *repacked_W, size_t data_size,
...
+}
+
+template <>
+void compute_kcaches(const float *A, const uint16_t *B, float *output,
+                     int num_rows, int N, int chunk_size, int group_size,
+                     int tile_size) {
+#ifdef ARMV7
+  __fallback_compute_kcaches<uint16_t>(A, B, output, num_rows, N, chunk_size,
+                                       group_size, tile_size);
+#else
```
skykongkong8: Ditto

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -888,6 +908,78 @@ void repack_q4_0_to_q4_0_8(void *W, void *repacked_W, size_t data_size,
...
+ * @param iter
+ * @param in
+ * @param vcache
+ * @param output
+ * @param seq
+ * @param num_cache_head
+ * @param gqa_size
+ * @param head_dim
+ * @param process_all
+ */
```
skykongkong8: Please fill in the doxygen tag param explanations

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -888,6 +908,78 @@ void repack_q4_0_to_q4_0_8(void *W, void *repacked_W, size_t data_size,
...
+ * @tparam BType
+ * @param A
+ * @param B
+ * @param output
+ * @param num_rows
+ * @param N
+ * @param chunk_size
+ * @param group_size
+ * @param tile_size
+ */
```
skykongkong8: Please fill in the doxygen tag param explanations

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -888,6 +908,78 @@ void repack_q4_0_to_q4_0_8(void *W, void *repacked_W, size_t data_size,
...
+ * @brief
+ * @param width
+ * @param dim
+ * @param half_
+ * @param inout
+ * @param output
+ * @param cos_
+ * @param sin_
+ * @param only_convert_to_fp16
+ */
```
skykongkong8: Please fill in the doxygen tag param explanations

nntrainer/tensor/cpu_backend/arm/neon_impl.h
```diff
@@ -558,6 +558,79 @@ void ele_div(const unsigned N, const float *X, const float *Y, float *Z,
...
                       const float *src, unsigned int ld_src, float *dst,
                       unsigned int ld_dst);
+
+/**
+ * @brief
+ * @param qk_out float * input/output values
+ * @param start_row start row number
+ * @param end_row end row number
+ * @param num_heads heads number
+ */
```
skykongkong8: I think you missed the brief, and the same for the other functions in this file.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
mnalewaj: I have faced compilation errors for ARMV7.
As a fix I moved the problematic 3 methods to neon_impl_fp16.cpp.
Regarding `compute_fp16vcache_fp32_transposed()` and `compute_u16vcache_fp32_transposed()` for arm I have only implementation method using intrinsics for conversion between fp16 and fp32.

I hope the current version meets your expectations.

Current implementation is not supported ARMV7 devices.
If you need please let me know but it will be cost additional time of implementation and probably it will be slow.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
mnalewaj: Fixed

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
mnalewaj: Done

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
mnalewaj: Done

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
mnalewaj: Done

nntrainer/tensor/cpu_backend/arm/neon_impl.h
mnalewaj: Done

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
skykongkong8: I see. Then let's leave it as it is for now (or rollback to previous version), and come back here afterwards.

PR #3309, opened 2025-07-09
Work on OpenCL compiler options and rotary_emb_cl

mwlasiuk:
- Add possibility to pass OpenCL compiler options from registerKernel() (plan to use in future to configure kernels via #defines which is helpful in case of Qn_K kernels)
- Do not copy data in rotary_emb_cl - use reference

**Self evaluation:**

### DIFF
--- a/nntrainer/cl_context.h
+++ b/nntrainer/cl_context.h
@@ -208,10 +208,13 @@ public:
    * @brief register or return already present OpenCl kernel pointer
    * @param kernel_string kernel implementation string
    * @param kernel_name kernel name
+   * @param compiler_options compiler options for OpenCL kernel compiler
    * @return std::shared_ptr<opencl::Kernel>
    */
-  const SharedPtrClKernel registerClKernel(std::string kernel_string,
-                                           std::string kernel_name);
+  const SharedPtrClKernel
+  registerClKernel(const std::string &kernel_string,
+                   const std::string &kernel_name,
+                   const std::string &compiler_options = "");
 
   /**
    * @brief Initialize and register all blas OpenCl kernels
@@ -300,11 +303,14 @@ private:
    * @brief create OpenCl kernel
    * @param kernel_string reference of implementation string
    * @param kernel_name reference of kernel_name
-   * @param kernel_ptr_ reference of shared_ptr of Kernel
+   * @param kernel_ptr reference of shared_ptr of Kernel
+   * @param compiler_options compiler options for OpenCL kernel compiler
    * @return true if successful, false otherwise
    */
-  bool clCreateKernel(std::string &kernel_string, std::string &kernel_name,
-                      const SharedPtrClKernel &kernel_ptr_);
+  bool clCreateKernel(const std::string &kernel_string,
+                      const std::string &kernel_name,
+                      const SharedPtrClKernel &kernel_ptr,
+                      const std::string &compiler_options);
 };
 
 /**
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -176,8 +176,9 @@ void ClContext::initAttentionClKernels() {
 }
 
 const ClContext::SharedPtrClKernel
-ClContext::registerClKernel(std::string kernel_string,
-                            std::string kernel_name) {
+ClContext::registerClKernel(const std::string &kernel_string,
+                            const std::string &kernel_name,
+                            const std::string &compiler_options) {
   // check if created before
   if (ocl_kernel_map.find(kernel_name) != ocl_kernel_map.end()) {
     ml_logi("Kernel already registered and initialized: %s",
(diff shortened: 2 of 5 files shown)


nntrainer/cl_context.h
```diff
@@ -215,8 +215,10 @@ class ClContext : public Context {
    * @param kernel_name kernel name
    * @return std::shared_ptr<opencl::Kernel>
```
djeong20, line 212: Can you also add a description here?

```suggestion
   * @param compiler_options compiler options for the OpenCL kernel compiler.
   * @return std::shared_ptr<opencl::Kernel>
```

nntrainer/cl_context.h
```diff
@@ -298,11 +300,14 @@ class ClContext : public Context {
    * @brief create OpenCl kernel
    * @param kernel_string reference of implementation string
    * @param kernel_name reference of kernel_name
-   * @param kernel_ptr_ reference of shared_ptr of Kernel
+   * @param kernel_ptr reference of shared_ptr of Kernel
+   * @param compiler_options compiler options for OpenCL kernel compiler
```
djeong20, line 307: can you also explain here why compiler option is added and how it will be used?

djeong20 (approved):
LGTM :)

nntrainer/cl_context.h
mwlasiuk: Sample usage - removing duplicated (no transpose vs transpose) code by defining how memory is being read based on compiler option:

For Qn_K it will be helpful to configure kernels using defines based on device capabilities instead of passing arguments - this can make kernel more optimal as gives compiler possibility to utilize concepts like [constant folding](https://en.wikipedia.org/wiki/Constant_folding) (for loops with index limited by defined constant vs by argument).

In CPP:

```cpp
  registerClKernel(getSgemvClKernel(), "sgemv_cl");
  registerClKernel(getSgemvClKernel(), "sgemv_cl_noTrans", "-DNO_TRANS");
  // registerClKernel(getSgemvClNoTransKernel(), "sgemv_cl_noTrans");
```

In OpenCL:
```cpp
  static const std::string sgemv_cl_kernel_ =
    R"(__kernel void 
        #ifdef NO_TRANS
          sgemv_cl_noTrans
        #else    
          sgemv_cl
        #endif 
        (const __global float* A, const __global float* X,
                          __global float* Y, unsigned int N, unsigned int lda) {                                            
            unsigned int i;
            i = get_global_id(0);        
            float y0 = 0.0f; 

            #ifdef NO_TRANS
              for (unsigned int j = 0; j < N; j++)                         
                  y0 += A[j + i * lda] * X[j];             
            #else                
              for (unsigned int j = 0; j < N; j++)                         
                  y0 += A[i + j * lda] * X[j]; 
            #endif

            Y[i] = y0;                            
              
        })";
  return sgemv_cl_kernel_;
}

const std::string &getSgemvClNoTransKernel() {
  static const std::string sgemv_cl_noTrans_kernel_ =
    R"(   //__kernel void sgemv_cl_noTrans(const __global float* A, const __global float* X,
          //                 __global float* Y, unsigned int N, unsigned int lda) {                                            
          //   unsigned int i;
          //   i = get_global_id(0);                         
          //   float y0 = 0.0f;
          //   for (unsigned int j = 0; j < N; j++)                         
          //       y0 += A[j + i * lda] * X[j]; 
          //   Y[i] = y0;                            
          //     
          //})";
  return sgemv_cl_noTrans_kernel_;
}
```

nntrainer/cl_context.h
mwlasiuk: done

nntrainer/cl_context.h
mwlasiuk: Also possible modification:

```cpp
const ClContext::SharedPtrClKernel
ClContext::registerClKernel(const std::string &kernel_string,
                            const std::string &kernel_name,
                            const std::string &compiler_options) {

  std::string map_name = kernel_name + compiler_options;

  // check if created before
  if (ocl_kernel_map.find(map_name) != ocl_kernel_map.end()) {
    ml_logi("Kernel already registered and initialized: %s",
            map_name.c_str());
    return ocl_kernel_map[map_name];
  }

  // creating shared_ptr for kernel object
  SharedPtrClKernel kernelPtr = std::make_shared<opencl::Kernel>();
  if (!clCreateKernel(kernel_string, kernel_name, kernelPtr,
                      compiler_options)) {
    ml_loge("Failed to register kernel %s", map_name.c_str());
    return nullptr;
  }
  // add to map
  ocl_kernel_map.emplace(map_name, kernelPtr);
  return ocl_kernel_map[map_name];
}
```

While registering kernel register it by name and compiler options - this will remove necessity of using :
```cpp
        #ifdef NO_TRANS
          sgemv_cl_noTrans
        #else    
          sgemv_cl
        #endif 
```
in kernels.

Every instance of the kernel will have __kernel__ function called 'kernel_name' but we will be able to access different configurations based on defines.

PR #3310, opened 2025-07-09
Address a few small memory leaks

p-pokorski-samsung:
`tflite::label_image::read_bmp` allocated memory for the image, which requires manual freeing. In some cases the resultant buffer was not freed. I changed this to return `std::vector<uint8_t>` so no manual memory management is needed.

**Self evaluation:**
3. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: p-pokorski-samsung p.pokorski@samsung.com

### DIFF
--- a/Applications/KNN/jni/main.cpp
+++ b/Applications/KNN/jni/main.cpp
@@ -176,7 +176,7 @@ int main(int argc, char *argv[]) {
       img += total_label[i] + std::to_string(j + 1) + ".bmp";
       printf("%s\n", img.c_str());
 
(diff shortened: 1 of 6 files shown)


DonghakPark (approved):
Nice Catch!! 
but Please fill commit message body for history ?

djeong20 (approved):
nice :)

skykongkong8 (approved):
LGTM!

myungjoo:
The style errors are not coming from this fix. Let's ignore them.

PR #3318, opened 2025-07-10
Fallback to cpu context when OpenCL device is not available in runtime

p-pokorski-samsung:
## Dependency of the PR
None

<details><summary>Fallback to cpu context when OpenCL device is not available in runtime</summary><br />

1. Don't register OpenCL context in Engine when initialization fails
2. Change fetching GPU context to make it optional by maybeGetRegisteredContext
3. In MemoryPool use calloc when OpenCL context is unavailable
4. Skip tests using OpenCL if OpenCL is unavailable

**Self-evaluation:**

Signed-off-by: Piotr Pokorski <p.pokorski@samsung.com>
</details>

### Summary

- Enable CPU fallback when OpenCL is enabled at compile-time, but unavailable during runtime (For example - no detected device)
- Don't register OpenCL context when OpenCL initialization fails
- Modify MemoryPool to use calloc when OpenCL is not available
- Change some OpenCL context fetching (by `getRegisteredContext`) to make it optional (`maybeGetRegisteredContext`)
- Skip tests when OpenCL is not available

### DIFF
--- a/nntrainer/engine.h
+++ b/nntrainer/engine.h
@@ -138,16 +138,33 @@ public:
    * @return Context Pointer : for register Object factory, casting might be
    * needed.
    */
-  nntrainer::Context *getRegisteredContext(std::string name) const {
+  nntrainer::Context *getRegisteredContext(const std::string &name) const {
+    auto context = maybeGetRegisteredContext(name);
+    if (context == nullptr) {
+      throw std::invalid_argument("[Engine] " + name +
+                                  " Context is not registered");
+    } else {
+      return context;
+    }
+  }
 
+  /**
+   * @brief get registered a Context
+   *
+   * @param name Registered Context Name
+   * @return Context Pointer : for register Object factory, casting might be
+   * needed. nullptr if no registered context with this name
+   */
+  nntrainer::Context *maybeGetRegisteredContext(std::string name) const {
     std::transform(name.begin(), name.end(), name.begin(),
                    [](unsigned char c) { return std::tolower(c); });
 
-    if (engines.find(name) == engines.end()) {
-      throw std::invalid_argument("[Engine] " + name +
-                                  " Context is not registered");
+    auto it = engines.find(name);
+    if (it == engines.end()) {
+      return nullptr;
+    } else {
+      return it->second;
     }
-    return engines.at(name);
   }
 
   std::unordered_map<std::string, std::shared_ptr<nntrainer::MemAllocator>>
--- a/nntrainer/layers/cl_layers/layer_impl_cl.h
+++ b/nntrainer/layers/cl_layers/layer_impl_cl.h
@@ -62,7 +62,7 @@ public:
 
 protected:
   inline static ClContext *global_cl_context =
-    static_cast<ClContext *>(Engine::Global().getRegisteredContext("gpu"));
+    static_cast<ClContext *>(Engine::Global().maybeGetRegisteredContext("gpu"));
   inline static ClBufferManager &clbuffInstance =
     ClBufferManager::getInstance();
 };
--- a/nntrainer/opencl/opencl_command_queue_manager.h
+++ b/nntrainer/opencl/opencl_command_queue_manager.h
@@ -38,7 +38,7 @@ class CommandQueueManager {
    * @brief Private constructor to prevent object creation
    *
    */
-  CommandQueueManager(){};
+  CommandQueueManager() {};
 
 public:
   /**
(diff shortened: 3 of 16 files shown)


nntrainer/engine.h
```diff
@@ -138,16 +138,33 @@ class Engine {
...
+  }
 
+  /**
+   * @brief get registered a Context
+   *
+   * @param name Registered Context Name
+   * @return Context Pointer : for register Object factory, casting might be
+   * needed. nullptr if no registered context with this name
+   */
+  nntrainer::Context *maybeGetRegisteredContext(std::string name) const {
```
djeong20, line 158: Do you think this function is necessary? Perhaps we utilize `getRegisteredContext`.

djeong20: How about implementing a function like `checkRegisteredContext()`, and start it from there?

nntrainer/layers/cl_layers/layer_impl_cl.h
```diff
@@ -62,7 +62,7 @@ class LayerImplCl : public LayerImpl {
 
 protected:
   inline static ClContext *global_cl_context =
-    static_cast<ClContext *>(Engine::Global().getRegisteredContext("gpu"));
+    static_cast<ClContext *>(Engine::Global().maybeGetRegisteredContext("gpu"));
```
djeong20, line 65: Check if the GPU context is registered or not. If not registered, let's not have a global_cl_context variable at all.

nntrainer/opencl/opencl_command_queue_manager.h
```diff
@@ -206,9 +206,9 @@ class CommandQueueManager {
   /**
    * @brief Get the OpenCL Command Queue object
    *
-   * @return const cl_command_queue
+   * @return cl_command_queue
    */
-  const cl_command_queue GetCommandQueue();
+  cl_command_queue &GetCommandQueue();
```
djeong20, line 211: Is this to avoid a copy?

djeong20 (commented):
I left some comments, PTAL!
Also, please check [C++ Format Checker](https://github.com/nnstreamer/nntrainer/actions/runs/16194010357?pr=3318)

nntrainer/engine.h
EunjuYang: I agree with Donghyeon. I find it odd that there is no clear distinction in purpose between `maybeGetRegisteredContext` and `getRegisteredContext`. (some code uses `maybeGetRegisteredContext` and some uses `getRegisteredContext`) It seems they serve similar functions, which could lead to code redundancy and make the codebase less efficient. Can we explore an alternative approach to address this issue?

nntrainer/layers/cl_layers/layer_impl_cl.h
EunjuYang: This could potentially result in accessing a `nullptr`.

nntrainer/engine.h
p-pokorski-samsung: The idea was to have an optional way to retrieve a context in cases where it can be nullable. Mostly it's an workaround around global variables. where there's no option to retrieve from an exception.

nntrainer/layers/cl_layers/layer_impl_cl.h
p-pokorski-samsung: > This could potentially result in accessing a `nullptr`.

My understanding is that if methods of `LayerImplCl` are used then OpenCL context is available (as OpenCL context actually registers implementations).

Generally I had an issue with global variables fetching context which may be unavailable. I think there are 3 solutions to this problem:
1. Always use `getRegisteredContext("gpu")`. This is quite cumbersome but simplest solution to the problem.
2. Do it as I did - so optional lookup of a context which will NOT fail when functions are actually called, because if they are called then the context is available.
3. Have some kind of singleton pattern which fetches and caches context on demand, something like:
```
ClContext* getClContext() {
    static ClContext* context = static_cast<ClContext *>(Engine::Global().getRegisteredContext("gpu"));
    return context;
}
```
but I think it's duplication of already existing `getRegisteredContext`

Other than 2 I think 1 could be second best, but we would have to remove all global variables fetching `ClContext` (which is a good thing, actually).
What are your thoughts?

nntrainer/engine.h
p-pokorski-samsung: > How about implementing a function like `checkRegisteredContext()`, and start it from there?

Yes, we could do it like this, but I'd like to discuss the approach to global variables first, as I think it will determine the best coarse of action.

nntrainer/opencl/opencl_command_queue_manager.h
p-pokorski-samsung: To be able to get an address as `clblast::Scal<float>` requires an address. Although it's more of a convenience - we could create a copy of command queue and pass an address to this copy as well.

PR #3321, opened 2025-07-11
Lets make singletons great again.

gkisalapl:
1. Unify the way we create noncopyable and nonmovable objects
2. Unify the way we create singletons
3. Prevent copy of move of selected classes
4. Avoid using global static variables
5. Generate .def file for nntrainer symbols (windows only)
6. Build nntrainer as shared library on windows

### DIFF
--- a/generate_def.py
+++ b/generate_def.py
@@ -1,0 +1,118 @@
+"""
+SPDX-License-Identifier: Apache-2.0
+Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved.
+
+@file build_windows_resources.py
+@date 16 July 2025
+@brief Extract symbols from obj files to build def file
+@author Grzegorz Kisala <gkisala@gmail.com>
+"""
+import sys
+import os
+import re
+import subprocess
+import argparse
+from collections import defaultdict
+
+def extract_syms(symfile, defs):
+    print('Opening ', symfile)
+    with open(symfile, 'r') as f:
+        for line in f:
+            if not line.strip():
+                continue
+            line = re.sub(r'notype \(\)', 'func', line)
+            line = re.sub(r'notype', 'data', line)
+            pieces = line.split()
+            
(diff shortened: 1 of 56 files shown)


nntrainer/nntrainer_logger.cpp
```diff
@@ -141,6 +141,8 @@ void Logger::log(const std::string &message,
   outputstream << ss.str() << " " << message << std::endl;
 }
 
+} /* namespace nntrainer */
```
djeong20, line 144: Is this intentional?

generate_def.py
```diff
@@ -0,0 +1,118 @@
+"""
+SPDX-License-Identifier: Apache-2.0
+Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved.
+
+@file build_windows_resources.py
+@date 16 July 2025
+@brief Extract symbols from obj files to build def file
```
djeong20, line 7: Could you explain in detail what this file does and how it is used?

djeong20 (approved):
Overall, LGTM. Great work!

nntrainer/nntrainer_logger.cpp
gkisalapl: yes

generate_def.py
gkisalapl: On Windows os if we want to use some functions from dll files they need to be exported by adding __declspec(dllexport) keyword. Since in nntrainer we do not have all functionality exposed by C/CPP API and custom layers design require exporting a lot of functions it's not convenient to use __declspec(dllexport) for most of nntrainer functions.
Instead of that we can simulate similar functionality to CMake https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html
by generating def file with all exported symbols and using such file in meson.
This is what generate_def.py script does - generation of def file with all exported symbols

piotrrak (approved):

PR #3322, opened 2025-07-11
Fix race conditions in CacheLoader

p-pokorski-samsung:
1. Add synchronization to SwapDevice (to mapped field)
2. Remove synchronization in CacheElement
3. Add synchronization to actives field in CachePool, rewrite it a little
4. Fix cache loader tests

**Self-evaluation:**

<details><summary>Fix race conditions in CacheLoader</summary><br />

{commit message}

Fix race conditions in CacheLoader

1. Add synchronization to SwapDevice (to mapped field)
2. Remove synchronization in CacheElement
3. Add synchronization to actives field in CachePool, rewrite it a little
4. Fix cache loader tests

**Self-evaluation:**

</details>

### Summary

There were a few race conditions:

1. Accessing `mapped` in `SwapDevice`, because different `CacheElem` could access the same `SwapDevice` for read and write
2. `active` in `CachePool` was not synchronized.

I introduced synchronization to `SwapDevice` to reduce synchronization to minimum (for example `mmap` doesn't need to happen under mutex)
Also I dropped mutexes from `CacheElem` - `CacheElem` is either swapped in or out. By contract of `CacheLoader` it won't happen at the same time, so it doesn't require synchronization per se.


nntrainer/tensor/cache_pool.cpp
```diff
@@ -218,9 +220,9 @@ std::shared_ptr<MemoryData> CachePool::getMemory(unsigned int id) {
     id, std::bind(&CachePool::validate, this, std::placeholders::_1),
     std::bind(&CachePool::invalidate, this, std::placeholders::_1), memory_ptr);
 
-  auto elem = std::make_shared<CacheElem>(swap_device, id, offset, len,
+  auto elem = std::make_unique<CacheElem>(swap_device, id, offset, len,
                                           mem_data, policy, memory_ptr);
```
mwlasiuk: Why not merge instead of create and move?

nntrainer/tensor/cache_pool.cpp
```diff
@@ -171,15 +172,16 @@ void CachePool::deallocate() {
 void CachePool::validate(unsigned int id) {
   if (!elems[id]->isActive()) {
     elems[id]->swapIn();
-    actives.push_back(elems[id]);
+    std::lock_guard<std::mutex> lock(mutex);
+    actives.insert(id);
```
DonghakPark, line 176:
```.cpp
  int load_task_id = load_task_executor->submit(
    [this, id](void *data) {
      std::lock_guard<std::mutex> lock(this->state_mutex);
      pool->loadTensor(id);
      this->states[id] = LoadState::Loaded;
    },
    (void *)(std::uintptr_t)id);
```

we lock here and 
pool->loadTensor(id) just call validate(id)
```.cpp
void CachePool::loadTensor(unsigned int id) { validate(id); }
```

should we need lock at validate with mutex again ?

p-pokorski-samsung: Thanks! I moved this mutex in `submit` to check something, but then I forgot to move it back. I moved it and it still seems to be correct. Good catch!

nntrainer/tensor/cache_pool.cpp
p-pokorski-samsung: Merged it with second line.

skykongkong8:
Occasionally, you should re-run timeout-ed build tests manually. It happens, and we are looking into it currently.

DonghakPark (approved):
LGTM! Thank you for handle this issue

skykongkong8 (approved):
LGTM!

DonghakPark:
> > Occasionally, you should re-run timeout-ed build tests manually. It happens, and we are looking into it currently.
> 
> Ok, thanks. How can I do that? Strangely I can't see a rerun job button...

It could happen : CI action (only repo(org) members can run, maybe)

i will rerun CI build if failed

DonghakPark:
@p-pokorski-samsung Could you Please rebase on upstream/main ?

i tested this code on my side & it work well

PR #3329, opened 2025-07-17
build/install: install path of nnstreamer subplugins

myungjoo:
The default option should follow the prefix.
Don't make a hardcoded absolute path as the default.

This often make problems in Github workflow action script testing.


jijoongmoon (approved):
LGTM

PR #3342, opened 2025-07-21
cblas: num-thread config code clean

myungjoo:
1. Do not call set-num-thread for every call. Call it once. It's a global configuration.
2. Add some warning messages about num-threads.


skykongkong8 (approved):
Guess this could be one of the possible reasons for too many function calls related to threads from openBLAS! LGTM!

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
```diff
@@ -33,6 +33,17 @@ void init_backend() {
...
+  // Do not repeatedly call set_num_threads. It's a global config.
+  openblas_set_num_threads(BLAS_NUM_THREADS);
+#else
+  /// Without openblas_set_num_threads,
+  /// it's set std::thread::hardware_concurrency()
+  /// It can be too high especially when the given blas function is small
+  /// or if there are other threads already created (nntrainer thread pool)
+  /// With big-little & threadboost, hardware_concurrency might be not good.
+  /// @todo configure this! (4?)
+#endif
```
skykongkong8: Calling `openblas_set_num_threads()` here might cause build fail because `x86_compute_backend.cpp` and `arm_compute_backend.cpp` only include `cblas_interface.h`, not `cblas.h`
`cblas.h` is only included at `cblas_interface.cpp`... Will it be ok to move `#include <cblas.h>` to be included at `cblas_interface.h`?

myungjoo: There is a issue of refering to cblas_interface.h even if enable-cblas is false.

I didn't fix that issue. I added an explicit compiler error for such cases.
Otherwise, it would emit link errors.

myungjoo: Anyway, openblas_set_num_thread is called from cblas_interface.cpp, not from compute_backend.cpp in this version.

skykongkong8: > There is a issue of refering to cblas_interface.h even if enable-cblas is false.
> 
> I didn't fix that issue. I added an explicit compiler error for such cases. Otherwise, it would emit link errors.

That was intentional. At the beginning of cpu_backend refactorization I assumed the cblas library to be a always-existing external lib since it supports both x86 and arm. (And I was planning to remove `enable-blas` option in the near future.)
However, like we discussed offline, there are some cases that `cblas.h` should not be included in the first place. I am currently handling this issue and such patch will be included at https://github.com/nnstreamer/nntrainer/pull/3350 by today or tmrw.

djeong20 (approved):

nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.cpp
```diff
@@ -15,11 +15,25 @@
 #include <cblas_interface.h>
 
 namespace nntrainer {
+
+void __openblas_set_num_threads(int num_threads) {
+  if (num_threads == -1) {
```
songgot: I think it would be a good idea to check the range.
```
  if (num_threads <= 0 && num_threads != -1) {
    throw std::invalid_argument("Number of threads must be positive or -1 for default");
  }
```

nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.cpp
```diff
@@ -15,11 +15,30 @@
...
+  #endif
+  } else {
+    openblas_set_num_threads(num_threads);
+  }
+}
 void __cblas_saxpy(const unsigned int N, const float alpha, const float *X,
                    const unsigned int incX, float *Y, const unsigned int incY) {
-#ifdef BLAS_NUM_THREADS
-  openblas_set_num_threads(BLAS_NUM_THREADS);
-#endif
```
jijoongmoon, line 22: I'm not sure we have to check if the num thread is propery set in every unpack function. We could just delete the openblas_set_num_threads in every function. and we just set it when we create the Network.
But for now, It is ok. We could remove the __openblase_set_num_threads call in unpack function later.

jijoongmoon (approved):
LGTM

PR #3347, opened 2025-07-22
[Emergency Fix] Revert "memory leak fix (valgrind error): Application/KNN"

myungjoo:
5731541cb254b2e579ba1a9edc0cc4b8e9ab1d2c has already fixed the issue in a different way, which creates a build error with this commit.

This reverts commit bc19f1af859acde591509d826708bd0b64ddf1ea.


skykongkong8 (approved):
Approve for quick fix

baek2sm (approved):
LGTM. 

Regardless of the content of this PR, I think it would be better to remove the KNN example. It doesn't seem like KNN is an application of NNTrainer(It isn't even a neural network, so it doesn't use our neuralnet graph system).

PR #3350, opened 2025-07-22
Pr/ggml/arm/q4048 gemm

skykongkong8:
# This PR includes:

1. Optimized q4_0 GEMM kernel for ARM -> previously it was using unoptimized for-loop fallback kernel
2. repacking function to support q4_0_4_8 gemm kernel and refactorize accordingly
3. bs threadpool version of q6_K GEMM (fine-grained)

**Self evaluation:**
4. Run test:     [X]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -245,7 +245,7 @@ void gemm_q4_0(const unsigned int M, const unsigned int N, const unsigned int K,
                const float *A, const unsigned int lda, const void *B,
                const unsigned int ldb, float *C, const unsigned int ldc) {
 #ifdef ENABLE_GGML
-  return __ggml_q4_0_8x8_q8_0_GEMM(M, N, K, A, lda, B, ldb, C, ldc);
+  return __ggml_q4_0_4x8_q8_0_GEMM<float>(M, N, K, A, lda, B, ldb, C, ldc);
 #else
   return __fallback_gemm_q4_0(M, N, K, A, lda, B, ldb, C, ldc);
 #endif
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -364,13 +364,13 @@ void repack_q4_0_to_q4_0_8(void *W, void *repacked_W, size_t data_size,
 }
 
 void repack_q4_0(void *W, void *repacked_W, size_t data_size,
-                        const unsigned int M, const unsigned int N){
-                          #ifdef ENABLE_GGML
+                 const unsigned int M, const unsigned int N) {
+#ifdef ENABLE_GGML
   __ggml_repack_q4_0_to_q4_0_4(W, repacked_W, data_size, M, N);
 #else
   __fallback_repack_q4_0_to_q4_0_4(W, repacked_W, data_size, M, N);
 #endif
-                        }
+}
 
 void repack_q4_K_to_q4_K_8(void *W, void *repacked_W, size_t data_size,
                            const unsigned int M, const unsigned int N) {
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -363,6 +363,15 @@ void repack_q4_0_to_q4_0_8(void *W, void *repacked_W, size_t data_size,
 #endif
 }
 
+void repack_q4_0(void *W, void *repacked_W, size_t data_size,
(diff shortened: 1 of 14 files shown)


skykongkong8:
To reviewers, I found that there are some more room for refactorization afterwards:
- making all _FP16 functions to function template
- Combining repeated function declarations in omp and bstp
Just to let you guys know that I am also aware of it!

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -184,20 +186,33 @@ void ele_div(const unsigned N, const float *X, const float *Y, float *Z,
 
 void saxpy(const unsigned int N, const float alpha, const float *X,
            const unsigned int incX, float *Y, const unsigned int incY) {
+#ifdef USE_BLAS
   __cblas_saxpy(N, alpha, X, incX, Y, incY);
+#else
+  __fallback_saxpy(N, alpha, X, incX, Y, incY);
```
jijoongmoon, line 242: It might be better to leave @todo comments here saying we have to optimize this fallback function with AVX2, NEON, OpenCL.

nntrainer/tensor/cpu_backend/cpu_backend.h
```diff
@@ -611,13 +642,14 @@ extern bool is_valid(const unsigned int N, const float *X);
...
  * @param ldc Leading dimension of C
  */
+template <typename T = float>
 extern void gemm_q4_0(const unsigned int M, const unsigned int N,
-                      const unsigned int K, const float *A,
-                      const unsigned int lda, const void *B,
-                      const unsigned int ldb, float *C, const unsigned int ldc);
+                      const unsigned int K, const T *A, const unsigned int lda,
+                      const void *B, const unsigned int ldb, T *C,
+                      const unsigned int ldc);
```
jijoongmoon, line 683: We also need to provide function for the std::vector<void*> B as well.

nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.h
```diff
@@ -203,6 +275,40 @@ void __ggml_repack_q4_0_to_q4_0_8(void *W, void *repacked_W, size_t data_size,
  */
 void __ggml_repack_q4_K_to_q4_K_8(void *W, void *repacked_W, size_t data_size,
                                   const unsigned int M, const unsigned int N);
+
+#ifdef ENABLE_FP16
```
jijoongmoon, line 343: Let's consider to make something like ggml_interface_fp16.h or use template.

jijoongmoon:
Overall, LGTM except CI.

EunjuYang:
I encountered the following error during the initial build of this PR. I used the `-Wno-error` option during the subsequent build, which is why the message below appears as a warning instead of an error.

```
warning: converting to ‘_Float16’ from ‘float’ with greater conversion rank
  161 |       y[i * qk + j] = x[i].qs[j] * d;
      |                       ~~~~~~~~~~~^~~
```

nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface.h
skykongkong8: Yep. Like I mentioned here:

> To reviewers, I found that there are some more room for refactorization afterwards:
> 
> * making all _FP16 functions to function template
> * Combining repeated function declarations in omp and bstp
>   Just to let you guys know that I am also aware of it!

Those functions will be re-implemented to function template.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
skykongkong8: Yep, definitely 👍

nntrainer/tensor/cpu_backend/cpu_backend.h
skykongkong8: I wonder why would we need such thing? For QKV / FFN simultaneous computation?

skykongkong8:
> I encountered the following error during the initial build of this PR. I used the `-Wno-error` option during the subsequent build, which is why the message below appears as a warning instead of an error.
> 
> ```
> warning: converting to ‘_Float16’ from ‘float’ with greater conversion rank
>   161 |       y[i * qk + j] = x[i].qs[j] * d;
>       |                       ~~~~~~~~~~~^~~
> ```

From which file?
I don't see such code from newly added files... I used : 
```C++
...
y[i * qk + j] = static_cast<_FP16>(x[i].qs[j] * d);
...
```
to avoid it.

skykongkong8:
@EunjuYang
Fixed in current push

nntrainer/tensor/cpu_backend/cpu_backend.h
jijoongmoon: For better latency as in #3367

PR #3353, opened 2025-07-23
[Q6_K] Update Q6_K dot opration.

EunjuYang:

<details><summary>[Q6_K] Update Q6_K dot opration</summary><br />

- This patch updatese Q6_K's dot operation to support transpose.
- In pricinple, Q6_K doesn't support transpose, but here, we aim to support trans option for dimension change in tensor for dot operation.

**Self evaluation:**

</details>

### Summary
- This PR updates FP32.dot(Q6_K).
- In the previous implementation, it assumes 'transposed' in dimension by default.
- This PR fixes the operation, while supporting in_transpose.
- Please note that this in_transpose support does not transpose of the matrix data but only change the dimension of height and weight.


songgot (approved):
There doesn't seem to be any problem with the code.
LGTM.

PR #3359, opened 2025-07-24
[Layer] Unsqueeze Layer Added

sachin-nntrainer:
Added Unsqueeze layer and its Unit test case

Commits:
[Layer] Add Unsqueeze layer
[TEST] Add unittest for Unsqueeze layer

Self evaluation:

Co-authored-by: Sachin Singh [sachin.3@samsung.com](mailto:sachin.3@samsung.com)
Co-authored-by: Abhinav Dwivedi [abhinav.d@samsung.com](mailto:abhinav.d@samsung.com)


baek2sm:
First, thank you for implementing the code neatly.

In NNTrainer, since the dimension is statically four-dimensional, there are constraints in implementing unsqueeze. Therefore, it is difficult to implement it in 100% compatibility with other frameworks.

So, what if we make some assumptions and implement it so that it works correctly under those conditions?

1. Let’s not consider the original input dimension from other frameworks like ONNX, and instead focus on the four-dimensional input dimension calculated by NNTrainer. 
2. Normally, when performing an Unsqueeze operation, a dimension with a value of 1 is added to the specified axis. However, since we are fixed to four dimensions, the definition of adding a dimension can be ambiguous. So, why don’t we implement it in a way similar to PyTorch by shifting dimensions?
    1. For example, if the shape is (1, 1, 2), then either unsqueeze(1) or unsqueeze(2) would result in (1, 1, 1, 2). In NNTrainer, both would result in (1, 1, 1, 2), so nothing happens.
        1. [PyTorch] (1, 1, 2) -> unsqueeze(1) or unsqueeze(2) -> (1, 1, 1, 2)
        2. [NNTrainer] (1, 1, 1, 2) -> unsqueeze(1) or unsqueeze(2) -> (1, 1, 1, 2)
    2. If the shape is (1, 1, 2), then unsqueeze(3) would result in (1, 1, 2, 1). This seems feasible to implement in NNTrainer as (1, 1, 1, 2) -> (1, 1, 2, 1).
        1. [PyTorch] (1, 1, 2) -> unsqueeze(3) -> (1, 1, 2, 1)
        2. [NNTrainer] (1, 1, 1, 2) -> unsqueeze(3) -> (1, 1, 2, 1)
3. By considering various scenarios like this, we can define the following:
    1. Among the (Batch, Channel, Height, Width) dimensions, Batch is always fixed and should not be affected by unsqueeze.
    2. As shown in the example above ((1, 1, 1, 2) -> (1, 1, 2, 1)), when unsqueeze operates in a form where Height or Width values shift to the left, the channel dimension must always be 1. Otherwise, an error should be returned because the total number of data elements will change.
    3. Conversely, when unsqueeze operates in a form where Channel or Height values shift to the right ((1, 1, 2, 1) -> (1, 1, 1, 2)), the width dimension must always be 1. Otherwise, an error should be returned because the total number of data elements will change.

Please review this and feel free to propose a better direction if you have any ideas—we can proceed in that direction as well. And thank you so much for your work!

sachin-nntrainer:
Thankyou @baek2sm for reviewing the PR and providing your feedback. 

As for 2., this unsqueeze layer operates as per pytorch implementation. Only the axis provided will be relative to four-dimensional input dimension as defined by NNTrainer. The example explained above will work. 

For 3.,

i. I will add a check to prevent batch size from changing. 
ii. & iii. I am unable to understand this point. Unsqueeze op doesnt consider tensor format. Could you please provide examples with respect to pytorch as well. It will help me gain more clarity. 

Again, Thanks a lot for providing your valuable feedback :)

nntrainer/layers/unsqueeze_layer.cpp
```diff
@@ -0,0 +1,73 @@
...
+void UnsqueezeLayer::finalize(InitLayerContext &context) {
+  NNTR_THROW_IF(context.getNumInputs() != 1, std::invalid_argument)
+    << "Unsqueeze only supports 1 input";
+
+  auto &axis = std::get<props::Axis>(unsqueeze_props);
+  unsigned int input_axis = axis.get();
+
+  TensorDim out_dim = context.getInputDimensions()[0];
+
+  for (unsigned int idx = 0; idx < input_axis; idx++) {
```
sachin-nntrainer, line 33: Shifting logic, determining the output dimension.

PR #3361, opened 2025-07-25
[OpenCL] Enable Q4_0 Block Processing

djeong20:
This commit introduces OpenCL kernels for flattening and restoring the q4_0 block.
These kernels will convert an array of q4_0 blocks into a block of arrays, enabling the utilization of the GPU's memory read throughput.

**Self-evaluation:**


dkjung (approved):
LGTM.

When does these functions become necessary?

dkjung:
`nntrainer/tensor/cl_operations/blas_kernel_strings.cpp` is getting too large.

We'd better split it into parts.

dkjung:
> LGTM.
> 
> When does these functions become necessary?

I just found you already put some notes.

PR #3366, opened 2025-07-28
Wrap nntrainer::ErrorNotification with warning suppressor

mwlasiuk:
- Added macros disabling warning (throwing exception in class destructor) to silence compiler warnings (as this throw is intended here)

Not proper way to fix it but if original code is written that way by intention lets at least suppress this warning

**Self evaluation:**


songgot (approved):
LGTM.

gkisalapl (commented):
This warning is really annoying but I think that we should fix this problem instead of silence compiler

PR #3372, opened 2025-07-30
[Layer] Add Reduce sum layer

sumon-98:
Added Reduce Sum layer implementation and unit tests.

Commits:
[Layer] Add Reduce sum layer

[Test] Add unittest for reduce sum layer

Self-evaluation:
1. Build test: [x] Passed [ ] Failed [ ] Skipped
2. Run test: [x] Passed [ ] Failed [ ] Skipped

Signed-off-by: Sumon Nath <[sumon.nath@samsung.com](mailto:sumon.nath@samsung.com)>


nntrainer/layers/reduce_sum_layer.cpp
```diff
@@ -0,0 +1,76 @@
...
+  const TensorDim &in_dim = context.getInputDimensions()[0];
+  TensorDim out_dim = in_dim;
+
+  /**
+   * if reduce axis is not provided, reduction is performed across all the
+   * dimensions except the batch
+   */
+  auto &reduce_axis = std::get<props::ReduceDimension>(reduce_sum_props);
+  if (reduce_axis.empty()) {
+    out_dim = TensorDim({1, 1, 1, 1});
```
baek2sm: How about raising error if `reduce_axis` prop is null?

sumon-98: We can do that; I did this to maintain conformity with reduce mean op. 
Another thing that can be done, is to have a special number which tells to reduce all axes.

baek2sm: Okay, it seems good. Then, how about reducing only "channel, height, and width" to 1, while keeping the batch size the same as the input size?

sumon-98: Yes, that seems like a good idea. I'll check and update.

sumon-98: I didn't check for empty axis earlier; it throws an error when the axis is empty as you were asking earlier. 
And it seems like the code where I handle empty axis is dead. Although, I set the axis to null explicitly in the forwarding function and checked the behavior and it is as you mentioned, it reduces all the dimensions except the batch size. 
I thought it would work as it is done the same way in reduce mean. What's your take on this?

baek2sm: Thanks to you, I also confirmed that the same part exists in the `reduce_mean_layer.cpp` script.
However, even though the execution result maintains the batch size as we intended and sets all other dimensions to 1, I still have some doubts about this script. Rather, since it was set to (1, 1, 1, 1) but operates as (batch_size, 1, 1, 1), could this be a bug? (This doesn’t mean your code is wrong, but rather that another part of the framework might be implemented incorrectly.)
I think explicitly implementing it as (batch_size, 1, 1, 1) would align better with our intent. Thank you for checking it.

sumon-98: Yes, you are right. But the main issue is, passing null axis throws an. It will be better if we remove the logic for handling null axis, as it is never invoked.

baek2sm: That makes sense. It would be better to remove the related code!

baek2sm (approved):
LGTM (please remove only the condition `reduce_axis.empty()`)

nntrainer/layers/reduce_sum_layer.cpp
sumon-98: Updated!

nntrainer/layers/reduce_sum_layer.h
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * @copydoc Layer::getType()
+   */
+  const std::string getType() const override { return ReduceSumLayer::type; };
+
+  static constexpr const char *type = "reduce_sum";
+
+private:
+  /** TODO: support scalar multiplier to simulate reduce_sum */
```
EunjuYang: What do you mean by this?

EunjuYang (approved):
LGTM

nntrainer/layers/reduce_sum_layer.h
sumon-98: Ah, that's my mistake. Forgot to remove the old comment while copying the header. I'll clean it up.

nntrainer/layers/reduce_sum_layer.cpp
```diff
@@ -0,0 +1,67 @@
...
+  context.getInput(SINGLE_INOUT_IDX)
+    .sum(reduce_axis, context.getOutput(SINGLE_INOUT_IDX));
+}
+
+void ReduceSumLayer::calcDerivative(RunLayerContext &context) {
+  auto &deriv = context.getIncomingDerivative(SINGLE_INOUT_IDX);
+  auto &ret_deriv = context.getOutgoingDerivative(SINGLE_INOUT_IDX);
+
+  ret_deriv.setZero();
+  ret_deriv.add_i(deriv);
```
djeong20, line 50: Why not copy?

```suggestion
  ret_deriv.copyData(deriv);
```

sumon-98: Thanks! Updated it.

sumon-98: @djeong20 I have reverted it to setZero and add_i. 
After rebasing on the latest main, the tests were failing while using copyData, as apparently copyData is not broadcasting the upstream gradients to the downstream gradients. Add_i is working fine.

PR #3379, opened 2025-07-31
Fix setting kernel args

gkisalapl:
We should use cl_mem instead of opencl::Buffer*
when setting kernel arguments with clSetKernelArg


jijoongmoon (approved):
LGTM

PR #3381, opened 2025-07-31
Unify tensor memory allocation.

gkisalapl:
1. This PR unify the code for tensor data memory allocation.
Instead of repeating code for data allocation and data access 11 times (for every tensor type) we've got now single function in tensor base which is responsible for that.
2. This PR also fix problem with not allocated scale and zero point data

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -68,13 +68,8 @@ void FloatTensor::allocate() {
     /** as this memory is shared, do NOT initialize */
   } else {
     /// allocate new memory for the tensor data
-    MemoryData *mem_data;
 
-    mem_data = new MemoryData((void *)(new float[dim.getDataLen()]{}));
-    data = std::shared_ptr<MemoryData>(mem_data, [](auto *mem_data) {
-      delete[] mem_data->template getAddr<float>();
-      delete mem_data;
-    });
+    allocateInternal();
 
     offset = 0;
     initialize();
@@ -84,22 +79,6 @@ void FloatTensor::allocate() {
 void FloatTensor::deallocate() {
   data = nullptr;
   offset = 0;
-}
-
-void *FloatTensor::getData() const {
-  if (!data)
-    return nullptr;
-
-  data->validate();
-  return data->getAddr<float>() + offset;
-}
-
-void *FloatTensor::getData(size_t idx) const {
-  if (!data)
-    return nullptr;
-
-  data->validate();
-  return data->getAddr<float>() + offset + idx;
 }
 
 void *FloatTensor::getAddress(unsigned int i) {
@@ -182,6 +161,8 @@ void FloatTensor::setRandBernoulli(float probability) {
 void FloatTensor::initialize() {
   if (empty() || !isAllocated())
     return;
+
+  TensorBase::initialize();
 
   unsigned int fan_in, fan_out;
 
--- a/nntrainer/tensor/q4_0_tensor.h
+++ b/nntrainer/tensor/q4_0_tensor.h
@@ -81,19 +81,6 @@ public:
   }
 
   /**
-   * @copydoc Tensor::getData()
-   */
-  void *getData() const override;
-
-  /**
-   * @copydoc Tensor::getData()
-   */
-  void *getData(size_t idx) const override {
-    throw std::invalid_argument(
-      "Q4_0_Tensor::getData() is not supported. Use getData() instead.");
-  }
-
-  /**
    * @copydoc Tensor::getAddress()
    */
   void *getAddress(unsigned int i) override {
@@ -230,6 +217,10 @@ private:
    * @copydoc Tensor::isValid()
    */
   bool isValid() const override { return true; }
+
+  std::size_t getDataTypeBitsSize() const override {
+    return sizeof(uint8_t) * CHAR_BIT;
+  }
 };
 
 } // namespace nntrainer
--- a/nntrainer/tensor/q4_k_tensor.cpp
+++ b/nntrainer/tensor/q4_k_tensor.cpp
@@ -45,6 +45,17 @@ Q4_K_Tensor::Q4_K_Tensor(const TensorDim &d, const void *buf,
   }
 }
 
+bool Q4_K_Tensor::operator==(const Q4_K_Tensor &rhs) const {
+  if (qscheme != rhs.qscheme)
+    return false;
+
+  // compare quantized data
+  const uint8_t *_data = (uint8_t *)getData();
+  const uint8_t *_rdata = (uint8_t *)rhs.getData();
+
+  return std::equal(_data, _data + size(), _rdata);
+}
+
 void Q4_K_Tensor::allocate() {
   if (empty() || data)
     return;
@@ -55,13 +66,7 @@ void Q4_K_Tensor::allocate() {
     /** as this memory is shared, do NOT initialize */
   } else {
     /// allocate new memory for the tensor data
-    MemoryData *mem_data;
-
-    mem_data = new MemoryData((void *)(new uint8_t[size()]{}));
-    data = std::shared_ptr<MemoryData>(mem_data, [](auto *mem_data) {
-      delete[] mem_data->template getAddr<uint8_t>();
-      delete mem_data;
-    });
+    allocateInternal();
 
     offset = 0;
     initialize();
--- a/nntrainer/tensor/q4_k_tensor.h
+++ b/nntrainer/tensor/q4_k_tensor.h
@@ -72,9 +72,25 @@ public:
   Q4_K_Tensor(TensorBase &rhs) : Uint4QTensor(rhs, QScheme::Q4_Kx8) {}
 
   /**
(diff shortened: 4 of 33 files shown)


nntrainer/tensor/tensor_base.h
```diff
@@ -184,49 +188,102 @@ class TensorBase {
...
+  virtual void *getData(size_t idx) const;
 
   /**
    * @copydoc Tensor::getScale()
    */
-  virtual void *getScale() const {
-    throw std::invalid_argument(
-      "Tensor::getScale() is not supported in tensor data type " +
-      getStringDataType());
+  virtual float *getScale() const {
```
p-pokorski-samsung, line 218: Is scale always `float`? Are there other variants?

nntrainer/tensor/tensor_base.h
```diff
@@ -184,49 +188,102 @@ class TensorBase {
...
+    if (hasScale()) {
+      if (!data) {
+        return nullptr;
+      }
+
+      data->validate();
+      return (float *)((std::byte *)getData() + getDataBytesSize());
```
p-pokorski-samsung, line 225: Is this address always guaranteed to be 32-bit aligned?

nntrainer/tensor/tensor_base.h
```diff
@@ -184,49 +188,102 @@ class TensorBase {
...
+    } else {
+      throw std::invalid_argument(
+        "Tensor::getScale() is not supported in tensor data type " +
+        getStringDataType());
+    }
```
p-pokorski-samsung, line 230:
```suggestion
  virtual float *getScale() const {
    NNTR_THROW_IF(!hasScale()) << "Tensor::getScale() is not supported in tensor data type " <<
        getStringDataType();
      if (!data) {
        return nullptr;
      }

      data->validate();
      return reinterpret_cast<float*>((reinterpret_cast<std::byte*>(getData()) + getDataBytesSize()));
    }
```

nntrainer/tensor/tensor_base.h
```diff
@@ -184,49 +188,102 @@ class TensorBase {
...
+      NNTR_THROW_IF(idx > scale_size(), std::invalid_argument)
+        << "Tensor::getScale() index is not valid";
+
+      return getScale() + idx;
+    } else {
+      throw std::invalid_argument(
+        "Tensor::getScale() is not supported in tensor data type " +
+        getStringDataType());
+    }
   }
```
p-pokorski-samsung, line 247:
```suggestion
  virtual float *getScale(size_t idx) const {
      NNTR_THROW_IF(!hasScale()) << "Tensor::getScale() is not supported in tensor data type " <<
        getStringDataType();
      NNTR_THROW_IF(idx > scale_size(), std::invalid_argument)
        << "Tensor::getScale() index is not valid";

      return getScale() + idx;
  }
```

nntrainer/tensor/tensor_base.h
```diff
@@ -184,49 +188,102 @@ class TensorBase {
...
+      }
+
+      data->validate();
+      return (unsigned int *)((std::byte *)getData() + getDataBytesSize() +
+                              getScaleBytesSize());
+    } else {
+      throw std::invalid_argument(
+        "Tensor::getZeroPoint() is not supported in tensor data type " +
+        getStringDataType());
+    }
```
p-pokorski-samsung, line 275:
```suggestion
      NNTR_THROW_IF(!hasZeroPoint()) <<
        "Tensor::getZeroPoint() is not supported in tensor data type " <<
        getStringDataType();
      if (!data) {
        return nullptr;
      }

      data->validate();
      return reinterpret_cast<unsigned int*>((reinterpret_cast<std::byte*>(getData()) + getDataBytesSize() + getScaleBytesSize()));
```

nntrainer/tensor/tensor_base.h
```diff
@@ -184,49 +188,102 @@ class TensorBase {
...
   }
 
   /**
    * @copydoc Tensor::getZeroPoint(size_t idx)
    */
   virtual unsigned int *getZeroPoint(size_t idx) const {
-    throw std::invalid_argument(
-      "Tensor::getZeroPoint() is not supported in tensor data type " +
-      getStringDataType());
+    if (hasZeroPoint()) {
```
p-pokorski-samsung, line 282: Same as up

nntrainer/tensor/float_tensor.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include <iomanip>
 #include <iostream>
+#include <memory>
```
p-pokorski-samsung: It seems unnecessary.

nntrainer/tensor/q4_0_tensor.h
```diff
@@ -230,6 +217,10 @@ class Q4_0_Tensor : public TensorBase {
    * @copydoc Tensor::isValid()
    */
   bool isValid() const override { return true; }
+
+  std::size_t getDataTypeBitsSize() const override {
+    return sizeof(uint8_t) * CHAR_BIT;
```
p-pokorski-samsung, line 222: I don't know many about `Q4_0`, but is it right for sure?

nntrainer/tensor/q4_k_tensor.h
```diff
@@ -111,6 +127,10 @@ class Q4_K_Tensor : public Uint4QTensor {
    */
   bool isValid() const override { return true; }
 
+  std::size_t getDataTypeBitsSize() const override {
+    return sizeof(uint8_t) * CHAR_BIT;
```
p-pokorski-samsung, line 131: Same as with Q4_0 - is it right?

nntrainer/tensor/q4_k_tensor.cpp
```diff
@@ -45,6 +45,21 @@ Q4_K_Tensor::Q4_K_Tensor(const TensorDim &d, const void *buf,
...
+
+  // compare quantized data
+  const uint8_t *_data = (uint8_t *)getData();
+  const uint8_t *_rdata = (uint8_t *)rhs.getData();
+  for (size_t i = 0; i < (size() + 1) / 2; ++i) {
+    if (_data[i] != _rdata[i])
+      return false;
+  }
+
+  return true;
```
p-pokorski-samsung:
```suggestion
  return std::equal(_data, _data + (size() + 1) / 2, _rdata);
```

nntrainer/tensor/tensor_base.h
gkisalapl: It's always float

nntrainer/tensor/tensor_base.h
gkisalapl: It's not guaranteed and the same problem exists before this PR.
We've got two choices here:
1. Keep memory continues and not guarantee proper alignment for scale and zero point
2. Do not keep continues memory and ensure alignment

I would like to address this problem in other PR

nntrainer/tensor/q4_0_tensor.h
gkisalapl: yes Q4 use unit8_t as underlying type

nntrainer/tensor/q4_k_tensor.h
gkisalapl: same here

nntrainer/tensor/tensor_base.h
gkisalapl: I preffer previous version of getScale() from one reason. I expect that some day we will disable NNTR_THROW_IF macro for non debug builds due to performance. In such case we will not check if tensor have scale at all

nntrainer/tensor/q4_k_tensor.cpp
gkisalapl: good tip

nntrainer/tensor/tensor_base.h
gkisalapl: same comment as for scale

nntrainer/tensor/tensor_base.h
p-pokorski-samsung: Ok, I see. Fair point.

nntrainer/tensor/q4_k_tensor.cpp
```diff
@@ -45,6 +45,17 @@ Q4_K_Tensor::Q4_K_Tensor(const TensorDim &d, const void *buf,
...
 
+bool Q4_K_Tensor::operator==(const Q4_K_Tensor &rhs) const {
+  if (qscheme != rhs.qscheme)
+    return false;
+
+  // compare quantized data
+  const uint8_t *_data = (uint8_t *)getData();
+  const uint8_t *_rdata = (uint8_t *)rhs.getData();
+
+  return std::equal(_data, _data + size(), _rdata);
```
p-pokorski-samsung, line 56: Before you iterated over `(size() + 1) / 2`, now over `size()`. Which one is correct?

gkisalapl: size() is correct you found a bug in previous version

JanuszBedkowski:
ready

mwlasiuk (approved):

PR #3389, opened 2025-08-05
Write kernel arguments in bulk

mwlasiuk:
- Added methods to set multiple kernel arguments "at once" (with one kernel->... call)
- Use new feature to reduce OpenCL codebase complexity

**Self evaluation:**


nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -69,111 +69,26 @@ void sgemv_q6_k_cl(void *matAdata, float *vecXdata, float *vecYdata,
...
+                                               {13, &r2, sizeof(int)},
+                                               {14, &r3, sizeof(int)}});
   if (!result) {
-    ml_loge("Failed to set kernel argument 13 for kernel_q6_k_sgemv_ptr");
     return;
   }
 
-  result = kernel_q6_k_sgemv_ptr->SetKernelArguments(14, &r3, sizeof(int));
-
+  result = kernel_q6_k_sgemv_ptr->SetKernelSVMArguments(
```
djeong20, line 592: How about combining `SetKernelArguments` and `SetKernelSVMArguments` into a single function?

For instance, 

```C++
kernel_q6_k_sgemv_ptr->SetKernelArguments({{0, matAdata, 0 /** size */},
                                           {1, &offset0, sizeof(cl_ulong)},
                                           {2, vecXdata,  0 /** size */},
                                           {3, &offset1, sizeof(cl_ulong)},
...
```

If size is 0, then set it as an SVM argument.

mwlasiuk: In my opinion it might be ambiguous and I would rather prefer to know that I set SVM arg explicitly. But spec says that zero size for normal arg is invalid so only other option that makes sense is SVM arg - so if other reviewers accept this I think that we may use your proposed solution.

nntrainer/opencl/opencl_kernel.h
```diff
@@ -15,12 +15,30 @@
...
 #include "opencl_program.h"
 
 namespace nntrainer::opencl {
 
+/**
+ * @brief Parameters needed to set OpenCL kernel argument
+ */
+struct KernelArgument {
+  uint32_t index = {};
+  const void *value = {};
```
gkisalapl: I would feel better if we initialize pointers with nullptr instead of {}.
I'm not sure if it's the same prior c++23

nntrainer/opencl/opencl_kernel.h
```diff
@@ -15,12 +15,30 @@
...
+  uint64_t size = {};
+};
+
+/**
+ * @brief Parameters needed to set OpenCL kernel SVM argument
+ */
+struct KernelSVMArgument {
+  uint32_t index = {};
+  const void *value = {};
```
gkisalapl: Same here

nntrainer/tensor/cl_operations/blas_kernels.cpp
gkisalapl: I would alslo preffer separate functions for setting SVM args

nntrainer/tensor/cl_operations/blas_kernels.cpp
```diff
@@ -1179,35 +868,13 @@ void transpose_32_16(float *data, int M, int K) {
...
-  if (!result)
-    throw std::runtime_error(
-      "Failed to set kernel argument 3 for kernel_transpose_32_16");
-
-  result = kernel_ptr->SetKernelArguments(arg++, &padded_height, sizeof(int));
-  if (!result)
-    throw std::runtime_error(
-      "Failed to set kernel argument 4 for kernel_transpose_32_16");
+  kernel_ptr->SetKernelSVMArguments(
+    {{9, data}, {1, clbuffInstance.getSVMInput()}});
```
djeong20: typo?
```suggestion
    {{0, data}, {1, clbuffInstance.getSVMInput()}});
```

PR #3390, opened 2025-08-05
Add possibility to get OpenCL error code as string.

mwlasiuk:
- Add OpenCL error code string to logged messages for easier debugging

**Self evaluation:**

### DIFF
--- a/nntrainer/opencl/opencl_loader.cpp
+++ b/nntrainer/opencl/opencl_loader.cpp
@@ -154,9 +154,6 @@ const char *OpenCLErrorCodeToString(const cl_int code) {
   default:
     return "(unknown)";
   }
-
-  return "(unknown)";
-
 #undef SWITCH_CASE_RETURN
 }
 


djeong20 (approved):
Looks good! This will be very helpful for debugging. For the next step, do you think we could actually throw an error instead of just logging it?

nntrainer/opencl/opencl_loader.cpp
```diff
@@ -65,6 +65,101 @@ bool LoadOpenCL() {
...
+#endif
+#ifdef CL_VERSION_2_2
+    SWITCH_CASE_RETURN(CL_INVALID_SPEC_ID);
+    SWITCH_CASE_RETURN(CL_MAX_SIZE_RESTRICTION_EXCEEDED);
+#endif
+  default:
+    return "(unknown)";
+  }
+
+  return "(unknown)";
```
songgot: The return "(unknown)"; statement at the end of the function will never be executed, because it has already been returned from the default case of the switch statement.
Let's remove line 158.

mwlasiuk: Done

skykongkong8 (approved):
Clearer logging for debug

PR #3401, opened 2025-08-07
[OpenCL] Optimized Q4_0 GEMM Kernel

djeong20:
This PR integrated the q4_0 matrix multiplication OpenCL kernel to nntrainer.
The following steps are required:
1. Flatten the q4_0x8 (CPU weights) into quantized values and scales, using AVX2 instructions for acceleration.
2. Transpose the quantized values and scales using an OpenCL kernel.
3. Transpose the activations and convert them to FP16 using another OpenCL kernel.
4. Perform the q4_0 8x4 GEMM operation.

The original kernel utilized an OpenCL image to optimize memory reads on Qualcomm Adreno GPUs.
However, on Intel GPUs, using shared virtual memory yields the highest performance compared to buffers and images.

There are several additional optimizations to consider:
1. Minimize synchronization delays. Currently, we synchronize after every execution.
2. Execute asynchronously on the GPU by employing either an out-of-order queue or multiple in-order queues.


skykongkong8 (approved):
This PR would handle several GPU unittest fails

PR #3405, opened 2025-08-11
[refact] rename local variable of network_graph.cpp

baek2sm:
The variable name "input_dims" was already declared as a class member
variable in the header file. But it reused as a local variable name in
the class method func, so the name of local variable was changed from
"input_dims" to "input_dimensions".

**Self evaluation:**


gkisalapl (commented):
We can avoid such errors if we start naming class variables with underscore suffix (use input_dims_ name for class variable in this case)

songgot (approved):
LGTM

djeong20 (approved):

baek2sm:
> We can avoid such errors if we start naming class variables with underscore suffix (use input_dims_ name for class variable in this case)

Thank you so much, I've renamed class variable from "input_dims" to "input_dims_" !

gkisalapl (approved):

jijoongmoon (approved):

PR #3412, opened 2025-08-13
Fix ggml linux build

gkisalapl:
1. Apply ggml patch in meson.build for all platforms
2. Other fixes

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -561,16 +561,19 @@ ggml_dep = dummy_dep
 if get_option('enable-ggml')
   message('preparing ggml')
 
+  if get_option('platform') != 'android' and get_option('platform') != 'tizen'
+    ggml_dir_relative = subprojects_dir_relative / 'ggml'
+    ggml_patch_relative = subprojects_dir_relative / 'packagefiles' / 'ggml' / '0001-nntrainer-ggml-patch.patch'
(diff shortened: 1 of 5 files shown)


meson.build
```diff
@@ -561,16 +561,19 @@ ggml_dep = dummy_dep
 if get_option('enable-ggml')
   message('preparing ggml')
 
+  if get_option('platform') != 'android' and get_option('platform') != 'tizen'
+    ggml_dir_relative = subprojects_dir_relative / 'ggml'
+    ggml_patch_relative = subprojects_dir_relative / 'packagefiles' / 'ggml' / '0001-nntrainer-ggml-patch.patch'
+    ggml_patch_command = ['git', 'apply', '--directory=' + ggml_dir_relative, ggml_patch_relative]
+    run_command(ggml_patch_command, check: false)
```
songgot, line 568: Changing `check: true` will cause the build to stop if the patch fails. 
Is this what you intended?

skykongkong8:
I picked your fixes with commit#a883ffb on #3409 as a Co-authored commit.
Other fixes proposed here seems to be duplicated fixes, so I think you can close this PR after merging #3409. Thanks for finding this issue out!👍👍

PR #3408, opened 2025-08-12
Optimize OpenCL Addition

gkisalapl:
Optimize OpenCL Addition

Unify opencl addition for FP16/32
Add possibility to pass nullptr as local work size

Speedup: 

Before:
FP32 ADD : N: 12 C: 1 H: 2048 W: 2048
 - time : CPU = 14135 us
 - time : GPU = 114108 us

After:
FP32 ADD : N: 12 C: 1 H: 2048 W: 2048
 - time : CPU = 13866 us
 - time : GPU = 74545 us

PR #3415, opened 2025-08-14
Fix and optimize rmsnorm OpenCL implementation, inspired by GGML opecl kernel.

p-pokorski-samsung:
This commit fixes and optimizes RMSNorm implementation of OpenCL kernel. Details:

1. `rmsnorm_cl` function is added to `blas_kernels.h`
2. RMSNormLayerCL now only calls `rmsnorm_cl` instead of setting up the kernel
3. `rmsnorm_cl` optionally uses SVM
4. The implementation is change inspired by GGML implementation - there's one subgroup per workgroup and each workgroup cooperatively calculates sum of elements in each row and then normalizes the row. Performance was tested for `3072x3072` tensor.
5. Currently there's an additional, commented implementation of kernel which uses 64 size workgroup (2 subgroups). On my machine (Intel UHD Graphics) there's no difference in performance between them, but on something other there can be.

| CPU      | GPU Before | GPU After         |
|----------|------------|-------------------|
| 25-40 ms | ~200 ms    | 3-5 ms (with SVM) | 

**Self-evaluation:**
6. Run test:   [X]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
@@ -879,33 +879,94 @@ const std::string &getConcatClAxis1Kernel() {
 
 const std::string &getRMSNormClKernel() {
   static const std::string rmsnorm_cl_kernel_ =
-    R"(__kernel void rmsnorm_cl(
+    R"(
+  #pragma OPENCL EXTENSION cl_intel_subgroups : enable  
+  #pragma OPENCL EXTENSION cl_intel_required_subgroup_size : enable
+
+ __attribute__((intel_reqd_sub_group_size(32)))
+ // __attribute__((reqd_work_group_size(32, 1, 1))) // For some reason gives worse performance, although should be necessary
+  __kernel void rmsnorm_cl(
     __global const float *input,  // Input tensor
     __global float *output,    // Output tensor
     __global const float *alpha,  // Alpha values (one for each width)
     float epsilon,
-    int B,                  // Number of batches
-    int C,                  // Number of channels
     int H,                  // Height of feature map
     int W                   // Width of feature map
 ) {
     // Compute the corresponding batch, height, and channel indices
-    int n = get_global_id(0) / C;
-    int c = get_global_id(0) % C;
-    int h = get_global_id(1);
-    int index = ((n * C + c) * H + h) * W;
+    int h = get_group_id(0);
+    int index = h * W;
     // Calculate RMS norm for the current channel, height, and batch
-    float sum_squares = 0.0f;
-    for (int j = 0; j < W; ++j) {
-        sum_squares += input[index+j] * input[index+j];
+    __global const float4 *in = (__global const float4*)(input + index);
+    float4 sum_squares_4 = 0.0f;
+    for (int i = get_local_id(0); i < W / 4; i += get_local_size(0)) {
+        sum_squares_4 += in[i] * in[i];
     }
-    sum_squares /= W;
-    float rms_norm = sqrt(sum_squares + epsilon);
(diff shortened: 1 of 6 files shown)


nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
```diff
@@ -717,33 +717,94 @@ const std::string &getConcatClAxis1Kernel() {
 
 const std::string &getRMSNormClKernel() {
   static const std::string rmsnorm_cl_kernel_ =
-    R"(__kernel void rmsnorm_cl(
+    R"(
+  #pragma OPENCL EXTENSION cl_intel_subgroups : enable  
+  #pragma OPENCL EXTENSION cl_intel_required_subgroup_size : enable
+
+ __attribute__((intel_reqd_sub_group_size(32)))
```
djeong20: This feature is only compatible with Intel GPUs. Let's add a condition for non-Intel GPUs.

```suggestion
    #ifdef cl_intel_required_subgroup_size
    #pragma OPENCL EXTENSION cl_intel_subgroups : enable
    #pragma OPENCL EXTENSION cl_intel_required_subgroup_size : enable
    #define INTEL_GPU 1
    #define REQD_SUBGROUP_SIZE_32 __attribute__((intel_reqd_sub_group_size(32)))
    #elif defined(cl_qcom_reqd_sub_group_size)
    #pragma OPENCL EXTENSION cl_qcom_reqd_sub_group_size : enable
    #define ADRENO_GPU 1
    #define REQD_SUBGROUP_SIZE_64 __attribute__((qcom_reqd_sub_group_size("half")))
    #endif
    
    #ifdef INTEL_GPU
    REQD_SUBGROUP_SIZE_32
    #elif defined(ADRENO_GPU)
    REQD_SUBGROUP_SIZE_64
    #endif
```

nntrainer/tensor/cl_operations/blas_kernels_templates.h
```diff
@@ -319,6 +319,90 @@ addition_cl_internal(ClContext::SharedPtrClKernel kernel, const T *input,
...
+  unsigned size_gamma = dim_gamma * sizeof(T);
+
+  auto *blas_cc =
+    static_cast<ClContext *>(Engine::Global().getRegisteredContext("gpu"));
+
+  if (use_svm) {
+    blas_cc->command_queue_inst_.enqueueSVMMap(const_cast<T *>(input), size_in,
+                                               true);
+    blas_cc->command_queue_inst_.enqueueSVMMap(const_cast<T *>(gamma),
+                                               size_gamma, true);
```
djeong20: As I mentioned in #3408, the usage of "Map" and "Unmap" is incorrect. Please address and correct this issue.

djeong20 (approved):
Please check the comments!

FYI, the following is the result on the Galaxy Book 5 Ultra 5
RMSNorm time : GPU = 2.82 ms
RMSNorm time : CPU = 27.24 ms

nntrainer/tensor/cl_operations/blas_kernel_strings.cpp
p-pokorski-samsung: OK, good idea. I'll do it, thanks :)

p-pokorski-samsung: Done.

nntrainer/tensor/cl_operations/blas_kernels_templates.h
p-pokorski-samsung: Done, thanks :)

p-pokorski-samsung:
Rebased onto main.

jijoongmoon (approved):
LGTM

PR #3423, opened 2025-08-20
[ ggml ] Add weight-vector function in Q4_K openmp thread backend

skykongkong8:
This PR adds Q4_K GEMM & GEMV for multiple weights in openMP thread backend

**Self evaluation:**


songgot:
#include <algorithm> is duplicated. I think it would be good to fix that as well.

PR #3424, opened 2025-08-20
Update FloatTensor Class for Q4_0 GEMM Support and Optimize GEMV Performance

djeong20:
This pull request updates the FloatTensor class to enable Q4_0 GEMM computation.
Note that the Q4_0 CPU kernel is utilized for GEMV operations.
Additionally, this patch restores the number of OpenMP threads to 6, as a decrease in performance was observed

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -723,12 +723,20 @@ void FloatTensor::dot(std::vector<Tensor *> input, std::vector<Tensor *> output,
   std::vector<float *> rdatas;
 
   for (unsigned int i = 0; i < input.size(); ++i) {
-    Ns.push_back(input[i]->getDim().width());
-    mdatas.push_back((void *)input[i]->getData<uint8_t>());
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/float_tensor.cpp
```diff
@@ -723,12 +723,20 @@ void FloatTensor::dot(std::vector<Tensor *> input, std::vector<Tensor *> output,
...
+#ifdef ENABLE_OPENCL
+    if (M == 1) {
+      gemm_q4_0(M, N, K, data, K, (void *)mdata, N, rdata, N);
+    } else {
+      gemm_q4_0_cl((void *)mdata, data, rdata, M, N, K);
+    }
+#else
+    /// @todo Support multi-weight q4_0 for x64
+    gemm_q4_0(M, N, K, data, K, (void *)mdata, N, rdata, N);
+#endif
```
djeong20, line 738: Please note that this is currently fixed to q4_0. To properly support this, we need to identify whether the input tensor type is q4_0 or q4_k.

PR #3407, opened 2025-08-12
Fix SwiGLU layer + utilize local group.

mwlasiuk:
- swiglu layer was setting cl_mem object instead pointer to cl_mem object as input / outpt arg causing error '-38'
- this PR fixes failing SwiGLU layer tests

``` bash
Running main() from ./googletest/src/gtest_main.cc
Note: Google Test filter = *Swiglu*
[==========] Running 12 tests from 3 test suites.
[----------] Global test environment set-up.
[----------] 10 tests from SwigluGPU/LayerSemanticsGpu
[ RUN      ] SwigluGPU/LayerSemanticsGpu.createFromClContext_pn/0
[       OK ] SwigluGPU/LayerSemanticsGpu.createFromClContext_pn/0 (472 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.setPropertiesInvalid_n/0
[       OK ] SwigluGPU/LayerSemanticsGpu.setPropertiesInvalid_n/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.finalizeValidateLayerNode_p/0
[       OK ] SwigluGPU/LayerSemanticsGpu.finalizeValidateLayerNode_p/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.getTypeValidateLayerNode_p/0
[       OK ] SwigluGPU/LayerSemanticsGpu.getTypeValidateLayerNode_p/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.gettersValidateLayerNode_p/0
[       OK ] SwigluGPU/LayerSemanticsGpu.gettersValidateLayerNode_p/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.setBatchValidateLayerNode_p/0
[       OK ] SwigluGPU/LayerSemanticsGpu.setBatchValidateLayerNode_p/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.setProperties_n/0
[       OK ] SwigluGPU/LayerSemanticsGpu.setProperties_n/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.gettersValidate_p/0
[       OK ] SwigluGPU/LayerSemanticsGpu.gettersValidate_p/0 (0 ms)
[ RUN      ] SwigluGPU/LayerSemanticsGpu.finalizeValidate_p/0
[...]

PR #3428, opened 2025-08-20
Combine GPU kernels #3424 + #3415 + #3408 + #3407

gkisalapl:


gkisalapl:
Not needed after merging most of PR-s to main

PR #3429, opened 2025-08-20
[ggml] Imported frequently used ggml functions into nntrainer

p-debski2:
### Summary

- imported GGML functions into `nntr_ggml_impl` directory and used those instead of the actual GGML in the `ggml_interface` functions
- removed GGML includes from `ggml_interface`

Signed-off-by: p-debski2 <p.debski2@samsung.com>


jijoongmoon:
I think it is OK to include the code to remove the ggml submodule.

djeong20 (changes requested):
This patch would not work correctly for all platforms.
For instance, you need to modify the Android.mk file under jni and test/jni. There are still ggml dependencies left.

For instance, 
https://github.com/nnstreamer/nntrainer/blob/bce948a22a57eab14524f6f856006585bcc3c9f0/jni/Android.mk.in#L95-L131
 
https://github.com/nnstreamer/nntrainer/blob/bce948a22a57eab14524f6f856006585bcc3c9f0/test/jni/Android.mk#L57-L62

In case of Tizen, you need to remove `ggml.tar.gz` under the packaging directory, modify nntrainer.spec.

https://github.com/nnstreamer/nntrainer/blob/bce948a22a57eab14524f6f856006585bcc3c9f0/packaging/nntrainer.spec#L445-L448

https://github.com/nnstreamer/nntrainer/blob/bce948a22a57eab14524f6f856006585bcc3c9f0/packaging/nntrainer.spec#L601-L604

https://github.com/nnstreamer/nntrainer/blob/bce948a22a57eab14524f6f856006585bcc3c9f0/packaging/nntrainer.spec#L801-L810

djeong20:
Additionally, I don't think the `enable_ggml` option is needed anymore. Let's remove the meson option and the ENABLE_GGML macro as well.

packaging/nntrainer.spec
```diff
@@ -442,11 +436,6 @@ ln -sf %{_libdir}/pkgconfig/capi-nnstreamer.pc %{_libdir}/pkgconfig/capi-ml-comm
 # Setup Ruy
 tar -xf packaging/ruy.tar.gz -C subprojects
 
-# Setup GGML
-%if 0%{?enable_ggml}
-tar -xf packaging/ggml.tar.gz -C subprojects
```
djeong20, line 447: Please delete the `packaging/ggml.tar.gz` file as well.

p-debski2:
All PR comments have been addressed, and I believe it should work correctly for all platforms now. 

However, I keep having trouble with Ubuntu Meson tests on CI after removing the `enable-ggml` flag. Building process is correct, but when it comes to launching unit tests, it always times-out on the `cpu-backend` one. 

Locally it works fine for me, and different builds also pass the unit tests phase (Ubuntu pdebuild & Ubuntu Meson with Clang). It's also concerning that for those workflows that pass the CI, this test usually takes ~20s to finish, while here it times out in 60s.

PR #3419, opened 2025-08-19
[mha_core] Add support for FP16 in mha_core layer

mnalewaj:
This PR adds support for FP16  data in `mha_core` layer.
This PR expands the scope of use the backend method: compute_rotary_emb_value() for FP16 data.
This PR adds unit tests for that methods.
Correctness of `mha_core` layer (for FP32 and FP16) verified on Android device by running LLM model.

**Self-evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test:   [X]Passed [ ]Failed [ ]Skipped

PR #3433, opened 2025-08-21
[Wait for #3419][Backend] Add local attention in backend methods for mha core v2 layer

mnalewaj:
This PR adds support for local attention backend methods used by mha_core v2 layer.

    [Backend] Add local attention in backend methods for mha core v2 layer

    **Self-evaluation:**

    Signed-off-by: Maciej Nalewaj <m.nalewaj@samsung.com>


EunjuYang (commented):
I found that this PR itself doesn't work (generates wrong output).
This should be reflected only after the mha_core layer is updated.

PR #3448, opened 2025-08-27
GitHub workflow - CPU OpenCL.

mwlasiuk:
This PR adds OpenCL support on CI GitHub build via pocl and for CI builds it updates build artifacts (compressed build directory) so it is possible to view exact logs from tests that were run.

**Self evaluation:**

Signed-off-by: mwlasiuk <testmailsmtp12345@gmailc.com>

### DIFF
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -25,6 +25,11 @@ jobs:
       run: sudo apt-get update && sudo apt-get install -y gcc g++ pkg-config libopenblas-dev libiniparser-dev libjsoncpp-dev libcurl3-dev tensorflow2-lite-dev nnstreamer-dev libglib2.0-dev libgstreamer1.0-dev libgtest-dev ml-api-common-dev flatbuffers-compiler ml-inference-api-dev libunwind-dev
     - name: install additional packages for features
       run: sudo apt-get install -y python3-dev python3-numpy python3
+    - name: install Opencl requirements
+      run: |
+        sudo apt install -y libpocl-dev
+        sudo apt install -y opencl-headers
+        sudo apt install -y ocl-icd-opencl-dev
     - name: gcc version change
       run: |
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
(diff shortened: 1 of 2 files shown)


.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-22.04, ubuntu-24.04 ]
-        meson_options: [ "-Denable-fp16=false", "-Denable-fp16=true" ]
+        meson_options: [ "-Denable-fp16=false", "-Denable-fp16=true", "-Denable-opencl=true" ]
```
djeong20: I think we should set enable-opencl=true as a default.

```suggestion
        meson_options: [ "-Denable-fp16=false", "-Denable-fp16=true" ]
```

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -57,10 +62,21 @@ jobs:
           -Dcapi-ml-common-actual=capi-ml-common \
           -Dcapi-ml-inference-actual=capi-ml-inference \
           -Denable-capi=enabled \
+          -Dci-build=true \
```
djeong20:
```suggestion
          -Denable-opencl=true \
```

nntrainer/opencl/opencl_context_manager.cpp
```diff
@@ -225,6 +226,8 @@ bool ContextManager::CreateDefaultGPUDevice() {
 
   if ((nullptr == platform_id_) || (nullptr == device_id_)) {
     ml_loge("No suitable platform / device found - using default (first)");
+    ml_loge("No suitable platform / device found - %p / %p",
+            platform_device_pairs[0].first, platform_device_pairs[0].second);
```
songgot, line 236: Is it necessary to include the code below?
```
if (platform_device_pairs.empty()) {  
    ml_loge("No OpenCL platforms/devices available");  
    return false;  
}
```

djeong20:
@mwlasiuk This pull request cannot be merged until the failing tests are fixed. Please include a commit with the solution.

nntrainer/opencl/opencl_context_manager.cpp
myungjoo: It is necessary.
When num_platforms > 0 && num_devices == 0 for all platforms, it's an empty vector.

mwlasiuk:
> This PR is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 3 days.

OK ...

PR #3449, opened 2025-08-28
Pr 3447+rmsnorm for pr

skykongkong8:
#3447 + rms normalization for 2 different ways:

- get FP32 input, compute with FP32, return FP32
- get FP32 input, compute with FP16, return FP32


skykongkong8:
Tizen build fail is nothing once #3429 is merged. But we should check for License issues.

PR #3451, opened 2025-09-01
Optimizing Memory Management in Asynchronous Tasks

djeong20:
This commit updates the mapping and unmapping of the model file within each async task and releases it immediately after reading.
This change allows file-backed pages to be dropped sooner, reducing peak RSS and working set during parallel loads.

**Self-evaluation:**


nntrainer/models/neuralnet.cpp
```diff
@@ -720,70 +720,85 @@ void NeuralNetwork::load(const std::string &file_path,
...
                        std::numeric_limits<size_t>::max(), true);
           } else {
-            node->read(mmaped, false, exec_mode, fsu_mode,
+#if defined(_WIN32)
+            // Map per-task, then unmap immediately after: enables early release
+            // of pages
+            HANDLE hFile =
+              CreateFileA(f_path.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
+                          OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+            NNTR_THROW_IF((hFile == INVALID_HANDLE_VALUE), std::runtime_error)
```
dkjung, line 748: It's well organized with exceptions being thrown properly. 

This PR just keeps me from being annoyed, thanks 👍

nntrainer/models/neuralnet.cpp
```diff
@@ -720,70 +720,85 @@ void NeuralNetwork::load(const std::string &file_path,
...
+              CreateFileMapping(hFile, NULL, PAGE_READONLY, 0, 0, NULL);
+            NNTR_THROW_IF((hMap == NULL), std::runtime_error)
+              << "CreateFileMapping failed";
+
+            char *view =
+              static_cast<char *>(MapViewOfFile(hMap, FILE_MAP_READ, 0, 0, 0));
+            NNTR_THROW_IF((view == nullptr), std::runtime_error)
+              << "MapViewOfFile failed";
+
+            node->read(view, false, exec_mode, fsu_mode,
```
dkjung, line 761: I have one question.

It seems that `view` always points the start of the file.
How does calling of `node->read` with the same `view` read the right position of the file? I tried to view inside `node->read`, but it's not easy to navigate to the real implementation because it's overridden.

dkjung (approved):
Can we measure loading time with this fix? I wonder if there is any change with/out this change.

nntrainer/models/neuralnet.cpp
djeong20: It's not particularly handled here in neuralnet.cpp. The call to node->read() currently uses `std::numeric_limits<size_t>::max()` as the starting offset. In TensorBase::read(), the reading process utilizes the file_offset when it is set to max().

dkjung: I didn't expect setting offset as `std::numeric_limits<size_t>::max()` would work like that.

jijoongmoon (commented):
LGTM

songgot (commented):
LGTM :+1:

Issue #3444, opened 2025-08-26
[CausalLM] support gpt-oss-20b


EunjuYang:
- CausalLM needs to support gpt-oss-20b [[model](https://huggingface.co/openai/gpt-oss-20b)]

# WTD

- [x] gpt-oss-20b model conversion script @EunjuYang 
- [x] gpt-oss-20b causallm.h / cpp @EunjuYang [[78f1182](https://github.com/nnstreamer/nntrainer/pull/3452/commits/78f1182e465aa4e938a9e4e05cc2e446ab28ad20)]
- [x] update fc layer to support bias with activation type @EunjuYang [[547059b](https://github.com/nnstreamer/nntrainer/pull/3452/commits/547059bbb4265486119c81af51fac39cb676dc57)]
- [x] rope scaling should be applied @EunjuYang  [[779cfdc](https://github.com/nnstreamer/nntrainer/pull/3452/commits/779cfdce68a9fe426ec81c3777c40df80da55683)]
- [x] yarn ROPE should be implemented (x86 only) @EunjuYang  [[779cfdc](https://github.com/nnstreamer/nntrainer/pull/3452/commits/779cfdce68a9fe426ec81c3777c40df80da55683)]
- [x] yarn ROPE support for neon should be updated @EunjuYang [[02bf8d9](https://github.com/nnstreamer/nntrainer/pull/3452/commits/02bf8d925028396b222d454cf32cffef78b4427e)]
- [x] update MoE layer to support bias @EunjuYang [[920d101](https://github.com/nnstreamer/nntrainer/pull/3452/commits/920d101e8e48e55aa44a43ff6937d4573f7a79a4)]
- [x] update attention layer to support sink @skykongkong8 
- [x] add gpt-oss-specific MoE layer @EunjuYang [[920d101](https://github.com/nnstreamer/nntrainer/pull/3452/commits/920d101e8e48e55aa44a43ff6937d4573f7a79a4)]
- [x] swiglu with alpha should be added @EunjuYang [[8a7dec2](https://github.com/nnstreamer/nntrainer/pull/3452/commits/8a7dec23cb290bdf088bdbfa6d853b2b35361e58)]
- [x] fast clamp operation should be added
- [x] Convert gpt-oss-20b weight to FP32-Q4_0_Q4_0 format @EunjuYang 
- [x] Fix attention performance degradation
- [x] Fix FP16 enabled Android issue
- [x] Support FP16 swiglu with alpha @EunjuYang 

[...]

PR #3452, opened 2025-09-01
[Draft] Causallm/gpt oss 20b

EunjuYang:
## Dependency of the PR
#3444


nntrainer/tensor/cpu_backend/arm/neon_impl_fp16.cpp
```diff
@@ -1451,8 +1515,194 @@ void softmax_row_inplace(__fp16 *qk_out, size_t start_row, size_t end_row,
...
+  } else {
+    return softmax_row_inplace_with_fp16_sink(qk_out, start_row, end_row,
+                                              num_heads, sink);
+  }
+}
+
+// New function: softmax_row_inplace with __fp16 input and float sink
+static void softmax_row_inplace_with_fp32_sink(__fp16 *qk_out, size_t start_row,
+                                               size_t end_row, size_t num_heads,
+                                               float *sink) {
```
skykongkong8, line 1533: Like we discussed offline, this is redundant behavior in terms of latency. Let's rather inject half-precision sink instead from the layer level, although I understand everything like these can be resolved once we introduce half-precision activation flow... Nice try btw!👍

EunjuYang: Let me use `T*qk_out , ... , T*sink` instead of this. Thank you for the comment!

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -854,6 +854,29 @@ void swiglu(const unsigned int N, float *X, float *Y, float *Z) {
   }
 }
 
+void swiglu(const unsigned int N, float *X, float *Y, float *Z, float alpha) {
+  unsigned int i = 0;
+  float32x4_t alpha_vec = vmovq_n_f32(alpha);
+  float32x4_t neg_alpha_vec = vmovq_n_f32(-alpha);
+
```
skykongkong8: `alpha_vec` is never being used here

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -814,11 +814,33 @@ void transpose_matrix(const unsigned int M, const unsigned int N,
...
+      vst1q_f32(&sin_[j], sin_vals);
+    }
+    while (j < N_half) {
+      cos_[j] *= attention_scaling;
+      sin_[j] *= attention_scaling;
+      ++j;
+    }
+  }
+
+  // Copy values to second half (duplicate)
```
skykongkong8: There is a huge redundancy in this implementation.
Trigonometric values are being loaded/stored twice. Although I can fix this simply, not really sure how to integrate such fix in this PR

EunjuYang: Thank you for the comment! Sure. It should be fixed.
Let's use this draft like a branch. Please checkout this PR and make a draft. Then I'm gonna cherry-pick and test. Is it reasonable?

PR #3453, opened 2025-09-01
Cache built kernel binaries and load them on subsequent runs

### LINKED ISSUE #3427: OpenCL Program Build with Kernel Binary
**Workflow Overview:**

1. If the kernel binary is not found:  
   - Build the program using the source code.  
   - Save the kernel binary.

2. If kernel binary is found:  
   - Build the program using the binary.

**Clarifications:**

`ClContext::clCreateKernel()` selects whether to build the program with a binary or source code. (DONE)

- Building the program with source code: (DONE)  
- Saving the kernel binary: (WORK TO DO)

- Building the program with the binary: (DONE, but needs verification)

p-pokorski-samsung:
Closes #3427 

**Self-evaluation:**

## Dependency of the PR
None
### Summary

- When building a kernel, a hash will be taken of the kernel code and build options. Hash will be the binary's filename
- Upon context initialization, the program will look for a binary. If it is found, binary will be used to create program.
- This change should be immune to changes in code and command line options. Not to driver and device version for now.

A quick tests I did showed the time to complete `nntrainer:unittests / unittest_blas_kernels_cl` reduced from 6.5 seconds without cache to 0.17 seconds with cache (for me this test fails after first test case but context initializes correctly).

### DIFF
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -25,6 +25,8 @@
 #include <swiglu_cl.h>
 #include <transpose_cl.h>
 
+#include <filesystem>
+
 #if defined(_WIN32)
 #include <windows.h>
 #endif
@@ -33,11 +35,41 @@ namespace nntrainer {
 
 std::mutex cl_factory_mutex;
 
+std::vector<std::byte> readBinaryFile(const std::string &path) {
+  // reading binary
+  std::ifstream fs(path, std::ios::binary | std::ios::in);
+
+  if (fs.good()) {
+    fs.seekg(0, std::ios::end);
+    size_t binary_size = fs.tellg();
+    fs.seekg(0, std::ios::beg);
+
+    std::vector<std::byte> data(binary_size);
+    fs.read(reinterpret_cast<char *>(data.data()), binary_size);
+    return data;
+  } else {
+    return {};
+  }
+}
+
+bool writeBinaryFile(const std::string &path,
+                     const std::vector<std::byte> &data) {
+  std::ofstream fs(path, std::ios::out | std::ios::binary);
+  if (!fs) {
+    ml_loge("Failed to open file for writing: %s", path.c_str());
+    return false;
+  }
+
+  fs.write(reinterpret_cast<const char *>(data.data()), data.size());
+  return true;
+}
+
 void ClContext::initialize() noexcept {
   try {
     if (!clInit()) {
       ml_loge("cl_context: opencl command queue creation failed");
     }
+    std::filesystem::create_directories(opencl::Program::DEFAULT_KERNEL_PATH);
 
     setMemAllocator(std::make_shared<MemAllocator>());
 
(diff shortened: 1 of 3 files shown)


nntrainer/cl_context.cpp
```diff
@@ -33,11 +35,41 @@ namespace nntrainer {
...
+  if (fs.good()) {
+    fs.seekg(0, std::ios::end);
+    size_t binary_size = fs.tellg();
+    fs.seekg(0, std::ios::beg);
+
+    std::vector<std::byte> data(binary_size);
+    fs.read(reinterpret_cast<char *>(data.data()), binary_size);
+    return data;
+  } else {
+    return {};
```
songgot, line 51: How about adding a log?

p-pokorski-samsung: Log is currently in the `else` path after `readBinaryFile` is called.

nntrainer/cl_context.cpp
```diff
@@ -219,45 +251,46 @@ bool ClContext::clCreateKernel(std::string &kernel_string,
...
+    if (result) {
+      auto binary = program.GetProgramBinary(
+        opencl::ContextManager::Global().GetDeviceId());
+
+      if (binary.empty()) {
+        ml_loge("Failed retrieving binary for kernel %s", kernel_name.c_str());
+        result = false;
+      } else {
+        result &= writeBinaryFile(binary_file_path, binary);
+      }
```
djeong20, line 285: I think we need to properly handle the error here when the result is false. When there's a kernel with syntax errors, it consistently tries to compile from the source. Let's throw an error when `CreateCLProgram` fails.

nntrainer/cl_context.cpp
```diff
@@ -219,45 +251,46 @@ bool ClContext::clCreateKernel(std::string &kernel_string,
...
-
-    if (!result) {
-      break;
+  opencl::Program program;
+
+  // reading binary
+  std::string binary_file_path =
+    opencl::Program::DEFAULT_KERNEL_PATH + "/" +
+    std::to_string(program.GetKernelHash(kernel_string, "")) + ".cl.bin";
+  auto binary_data = readBinaryFile(binary_file_path);
```
djeong20, line 260: It would be beneficial to provide an option that updates the existing binary. In some cases, the kernel may be updated, but the current configuration could continue to use the outdated version. Could we implement a check to verify whether the kernel string has been updated?

djeong20 (commented):
Please find below comments

nntrainer/cl_context.cpp
p-pokorski-samsung: Do you mean to throw an actual exception?
Or do you mean we should just cache some empty compilation result and not try to recompile?

Currently if `CreateCLProgram` fails the `GetProgramBinary` will not be called as `result` will be `false`.
Personally I'd love to see more exceptions being thrown as it would clean-up a lot of error handling where errors should anyway NOT be recoverable.

nntrainer/cl_context.cpp
p-pokorski-samsung: Yes, this is currently the case! If a kernel changes then `program.GetKernelHash` will return a different value.
`program.GetKernelHash` takes compilation options, kernel code and some default compilation options (inside `Program` class) and calculates the hash value. The hash value is used as a filename for binary. So if compilation options, kernel code change -> hash changes -> filename changes -> binary file not found -> recompilation.

Currently changes in hardware and driver version are not taken into account, but should be.

nntrainer/cl_context.cpp
djeong20: Yes, something like below

```C++
if (!result) {
  throw std::runtime_error(
    "Error: Failed to create OpenCL program " + kernel_name);
}
```

djeong20: + It would be nice to show what the problem is as follows.

https://github.com/nnstreamer/nntrainer/blob/bce948a22a57eab14524f6f856006585bcc3c9f0/nntrainer/opencl/opencl_program.cpp#L48-L51

djeong20: Let's not use `ml_loge` since it only logs the error message.

p-pokorski-samsung: I think it's a great idea. To take it a step further we could refactor all (or maybe not all) calls to OpenCL functions to throw  an exception in case of failure. 
For example:
`clCreateProgramWithSource` -> not recoverable in case of failure
`clCreateProgramWithBinary` -> not recoverable in case of failure

etc. But I'd leave it up to the next PR to make it consistent, what do you think?

nntrainer/cl_context.cpp
djeong20: Make sense! In that case, how is the outdated kernel binary deleted?

nntrainer/cl_context.cpp
djeong20: Sounds good :)

nntrainer/cl_context.cpp
p-pokorski-samsung: > Make sense! In that case, how is the outdated kernel binary deleted?

Currently it's only deleted by a rebuild. Maybe it is a little wasteful but in the production setting the kernels will not change too often and binaries themselves are not that big - several kB max.

Also I didn't have a good idea how to implement this nicely - we'd need to have some kind of cache registry to monitor "current cache" and purge anything not in "current cache". This would require some implementation and testing.

djeong20 (approved):
Awesome job! Let's move forward with the next phase of the work.

nntrainer/cl_context.cpp
songgot: Thank you.

jijoongmoon (approved):
LGTM

PR #3456, opened 2025-09-03
[layer] enable fp16 dtype for input tensor

lhs8928:
- Enable fp16 data type for input tensor of input & embedding layer

### DIFF
--- a/nntrainer/layers/layer_node.h
+++ b/nntrainer/layers/layer_node.h
@@ -1048,7 +1048,8 @@ properties in the context/graph unless intended. */
                std::vector<props::InputConnection>,
                std::vector<props::InputShape>, props::SharedFrom,
                props::ClipGradByGlobalNorm, props::Packed, props::WeightDtype,
-               props::LossScaleForMixed, props::ComputeEngine>;
+               props::LossScaleForMixed, props::ComputeEngine,
+               props::InputTensorDataType>;
 
   using RealizationPropsType = std::tuple<props::Flatten, props::Activation>;
   /** these realization properties results in addition of new layers, hence
--- a/nntrainer/utils/base_properties.h
+++ b/nntrainer/utils/base_properties.h
@@ -707,6 +707,17 @@ struct TensorTypeInfo {
 namespace props {
 
 /**
+ * @brief Input Data Type Enumeration Information
+ *  This property is differentiated with TensorDataType in that it doesn't have
+ *  default value
+ */
+class InputTensorDataType final : public EnumProperty<TensorDataTypeInfo> {
+public:
+  using prop_tag = enum_class_prop_tag;
+  static constexpr const char *key = "input_dtype";
+};
+
+/**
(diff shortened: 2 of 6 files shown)


nntrainer/layers/layer_node.h
```diff
@@ -1048,7 +1048,8 @@ properties in the context/graph unless intended. */
                std::vector<props::InputConnection>,
                std::vector<props::InputShape>, props::SharedFrom,
                props::ClipGradByGlobalNorm, props::Packed, props::WeightDtype,
-               props::LossScaleForMixed, props::ComputeEngine>;
+               props::LossScaleForMixed, props::ComputeEngine,
+               props::InputTensorDataType>;
```
jijoongmoon, line 1052: I think we can use the InputShape property to set the Tensor Data Type. 
InputShape property is from the GenericShape and GenericShape is from Property<TensorDim>. 
As you can see, TensorDim class has TensorDataType.

api/ccapi/include/tensor_dim.h
```diff
@@ -39,6 +40,14 @@ namespace train {
  */
 class TensorDim {
 public:
+#ifdef ENABLE_FP16
+  using IO_TensorType = std::variant<float *, uint32_t *, uint16_t *, uint8_t *,
+                                     int16_t *, int8_t *, _FP16 *>;
+#else
+  using IO_TensorType = std::variant<float *, uint32_t *, uint16_t *, uint8_t *,
+                                     int16_t *, int8_t *>;
+#endif
```
djeong20, line 49: Just a heads up: the `_FP16` type can be represented as `uint16_t` on Windows. Please note that this feature is still a work in progress.

nntrainer/layers/layer_node.h
lhs8928: As you described we can use the input shape property to set input tensor data type.
But in this case we need to specify all the input tensor even this data type is identical with model activation data type.
Let me explain why.
Currently in layer node we get the input shape and set the data type with model activation data type.
The way I thought was set the input tensor data type with model activation data type if input data type is not provided. In this way we only need to specify the data type only when the input tensor type differs form the model activation type.
But if we are going to use input shape property to set input tensor data type then we need to specify all the input tensor data type.
So these are the differences, please let me know which one is better then I will do that way.

nntrainer/utils/base_properties.h
```diff
@@ -706,6 +706,17 @@ struct TensorTypeInfo {
...
 
+/**
+ * @brief Input Data Type Enumeration Information
+ *  This property is differentiated with TensorDataType in that it doesn't have
+ *  default value
+ */
+class InputTensorDataType final : public EnumProperty<TensorDataTypeInfo> {
+public:
+  using prop_tag = enum_class_prop_tag;
+  static constexpr const char *key = "input_dtype";
```
EunjuYang, line 717: How about setting 'FP32' as the default value for backward compatibility?

PR #3417, opened 2025-08-18
[Build] support opencl disablement in Android

EunjuYang:
## Commits to be reviewed in this PR

<details><summary>[Build] fix android build to support opencl disablement</summary><br />

- Fix #3416
- This patch updates build script for Android (ggml-related script)
- This patch updates meson build and Android mk files to support enable-opencl=false option

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Eunju Yang <ej.yang@samsung.com>
</details>

### Summary

- Fix #3416
- This patch updates build script for Android (ggml-related script)
- This patch updates meson build and Android mk files to support enable-opencl=false option
- One can do the android build with `./tools/package_android.sh -Denable-opencl=true / false`

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

PR #3457, opened 2025-09-04
Enable OpenCL Support on Android and Resolve Issues

djeong20:
This PR addresses issue #3417 by enabling OpenCL support on Android, which was previously disabled.

This patch verifies build arguments for android_test and ensures OpenCL libraries are correctly built and included.

**Self-evaluation:**


test/jni/meson.build
djeong20, line 1: The meson.build file is unused since we rely on ndk-build, so it is being removed.

songgot (approved):
LGTM 👍

PR #3468, opened 2025-09-09
Add Fp16 support on windows (depends on #3429)

gkisalapl:
This PR make possible to perform windows build with FP16
    enabled, uint16_t is used as FP16 storage only type


gkisalapl:
Changes moved to #3498

PR #3477, opened 2025-09-12
Restricting OpenCL Context Usage When Unavailable

djeong20:
This PR addresses issues where the OpenCL context only considered scenarios in which OpenCL was available.
As a result, there was no proper handling of cases where OpenCL was not available.
This patch now restricts the use of the OpenCL context accordingly.
Please note that additional work may still be required.

**Self-evaluation:**

### DIFF
--- a/nntrainer/cl_context.h
+++ b/nntrainer/cl_context.h
@@ -289,6 +289,11 @@ private:
 
     // getContext() called inside createCommandQueue which creates clContext
     bool result = command_queue_inst_.CreateCommandQueue();
+
+    // Return false when creating command queue fails
+    if (!result)
+      return result;
+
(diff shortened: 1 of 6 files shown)


nntrainer/cl_context.h
```diff
@@ -289,6 +289,11 @@ class ClContext : public Context, public Singleton<ClContext> {
 
     // getContext() called inside createCommandQueue which creates clContext
     bool result = command_queue_inst_.CreateCommandQueue();
+
+    // Return false when creating command queue fails
+    if (!result)
+      return result;
```
songgot, line 295: How about adding a log?
```
ml_loge("opencl command queue creation failed");
```

PR #3479, opened 2025-09-12
[tflite] Set disabling tflite as the default option

baek2sm:
I think there's no need to have tflite dependency
as the default option, so I disabled it.

**Self evaluation:**


songgot (approved):
LGTM

PR #3481, opened 2025-09-15
[LLM] Update MoE cache policy

DonghakPark:
## Dependency of the PR

*No external dependencies introduced in this PR.*

<details><summary>4de43dbd324c582a1db77c3502a279a3255710bc</summary><br />

[FSU][GPT] Update Cache Policy & run Parallel

Update Cache Policy & run Parallel
- update topk policy at expert cache
- run activate & deactivate parallel
 
**Self evaluation:**

</details>

<details><summary>52fc0f1569fb2c11984a7ab165d4848d8ae9a244</summary><br />

[FSU][Qwem] Update Cache Policy & run parallel

Update Cache Policy & run Parallel
- update topk policy at expert cache
- run activate & deactivate parallel

**Self evaluation:**

</details>

### Summary

- Added extra Top‑K handling and mutex‑protected LRU cache updates to improve cache hit rate and thread‑safety for GPT MoE layer.
- Refactored GPT & Qwen MoE layer with similar extra Top‑K logic, dynamic scheduling, and synchronized cache eviction, resulting in better parallel performance.

### Qwen 30B
--- 

#### Before
```
=================[ LLM with NNTrainer ]===================
prefill: 33 tokens, 1206 ms, 27.3632 TPS
generation: 487 tokens, 54409 ms, 8.95073 TPS
==========================================================
Max Resident Set Size: 5495144 KB
```
#### After | Extra TopK = 5 | Cache Size = 32
``` 
=================[ LLM with NNTrainer ]===================
prefill: 33 tokens, 1123 ms, 29.3856 TPS
generation: 487 tokens, 45918 ms, 10.6059 TPS
==========================================================
Max Resident Set Size: 5357796 KB
``` 
### GPT OSS 20B
---

#### Before
```
=================[ LLM with NNTrainer ]===================
prefill: 78 tokens, 6101 ms, 12.7848 TPS
generation: 1024 tokens, 132296 ms, 7.74022 TPS
==========================================================
Max Resident Set Size: 6712968 KB
```
#### After | Extra TopK = 3 | Cache Size = 8
```
=================[ LLM with NNTrainer ]===================
prefill: 78 tokens, 3676 ms, 21.2187 TPS
generation: 1024 tokens, 117344 ms, 8.72648 TPS
==========================================================
Max Resident Set Size: 6458396 KB
```

### DIFF
--- a/Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
+++ b/Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
@@ -250,7 +250,6 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
     input.dot(gate_weights, router_logits);
     router_logits.apply(nntrainer::ActiFunc::softmax<float>, router_logits);
 
-
     // get extra topK
     auto extra_topk_result = router_logits.topK(topk + 3);
     auto extra_topk_values = std::get<0>(extra_topk_result);
@@ -260,8 +259,8 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
     const uint32_t *extra_indices_data = extra_topk_indices.getData<uint32_t>();
 
     // get extra topk
-    for (int i = 0; i < static_cast<int>(total_tokens); ++i) {
-      for (int k = topk; k < static_cast<int>(topk + 3); ++k) {
+    for (int i = static_cast<int>(total_tokens) - 1; i >= 0; --i) {
+      for (int k = 0; k < static_cast<int>(topk + 3); ++k) {
         unsigned expert_idx = extra_indices_data[i * topk + k];
         extra_top_k.push_back(expert_idx);
       }
@@ -353,11 +352,6 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
 #endif
         {
           std::lock_guard<std::mutex> lock(cache_mutex);
-          if (iteration_map.find(expert_idx) != iteration_map.end()) {
-            loaded_expert_deque.erase(iteration_map[expert_idx]);
-            loaded_expert_deque.push_back(expert_idx);
-            iteration_map[expert_idx] = --loaded_expert_deque.end();
-          }
           hit_count += 1;
         }
 
--- a/Applications/CausalLM/layers/qwen_moe_layer_cached.cpp
+++ b/Applications/CausalLM/layers/qwen_moe_layer_cached.cpp
@@ -328,8 +328,8 @@ void CachedSlimMoELayer::incremental_forwarding(
     const uint32_t *extra_indices_data = extra_topk_indices.getData<uint32_t>();
 
     // get extra topk
-    for (int i = 0; i < static_cast<int>(total_tokens); ++i) {
-      for (int k = topk; k < static_cast<int>(topk + 5); ++k) {
+    for (int i = static_cast<int>(total_tokens) - 1; i >= 0; --i) {
+      for (int k = 0; k < static_cast<int>(topk + 5); ++k) {
         unsigned expert_idx = extra_indices_data[i * topk + k];
         extra_top_k.push_back(expert_idx);
       }
@@ -414,11 +414,6 @@ void CachedSlimMoELayer::incremental_forwarding(
 #endif
         {
           std::lock_guard<std::mutex> lock(cache_mutex);
-          if (iteration_map.find(expert_idx) != iteration_map.end()) {
-            loaded_expert_deque.erase(iteration_map[expert_idx]);
-            loaded_expert_deque.push_back(expert_idx);
-            iteration_map[expert_idx] = --loaded_expert_deque.end();
-          }
           hit_count += 1;
         }
 


Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
```diff
@@ -249,6 +249,24 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
...
+    for (int i = 0; i < static_cast<int>(total_tokens); ++i) {
+      for (int k = topk; k < static_cast<int>(topk + 3); ++k) {
+        unsigned expert_idx = extra_indices_data[i * topk + k];
+        extra_top_k.push_back(expert_idx);
+      }
+    }
+
     auto topk_result = router_logits.topK(topk);
     auto topk_values = std::get<0>(topk_result);
     auto topk_indices = std::get<1>(topk_result);
```
EunjuYang, line 271: Instead of calling `topK()` again, how about reusing the `extra_topk_values` and `extra_topk_indices` with a shared tensor?

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
```diff
@@ -249,6 +249,24 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
...
+    // get extra topK
+    auto extra_topk_result = router_logits.topK(topk + 3);
+    auto extra_topk_values = std::get<0>(extra_topk_result);
+    auto extra_topk_indices = std::get<1>(extra_topk_result);
+    std::deque<int> extra_top_k = {};
+    extra_topk_values.divide_i(extra_topk_values.sum(3));
+    const uint32_t *extra_indices_data = extra_topk_indices.getData<uint32_t>();
+
+    // get extra topk
+    for (int i = 0; i < static_cast<int>(total_tokens); ++i) {
```
EunjuYang: As far as I understood, you put higher priority to the first element of `extra_top_k`. (This patch updates `loaded_expert_dequeue` in reverse order of `extra_top_k`) If so, the order must be reversed. (lateset token's expert should have higher priority)
```suggestion
    for (int i = static_cast<int>(total_tokens) - 1; i >= 0; --i) {
```

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
```diff
@@ -287,103 +306,107 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
...
+        context.getWeight(expert_down_bias_indices[expert_idx]).activate();
 
-    auto t1_miss = high_resolution_clock::now();
+        {
+          std::lock_guard<std::mutex> lock(cache_mutex);
+          loaded_expert_deque.push_back(expert_idx);
+          iteration_map[expert_idx] = --loaded_expert_deque.end();
+          need_load[expert_idx] = false;
+          miss_count += 1;
+        }
```
EunjuYang, line 335: If these are added to the deque before extra experts are included, the priority of these actually used experts will decrease. Why don't we update the cache all at once? Additionally, if we update the cache using multi-threading, the order might not be preserved as intended.

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
EunjuYang: I think we don't need to separate `topk_values` and `extra_topk_values`.
What do you think?

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
```diff
@@ -270,6 +288,7 @@ void CachedSlimGptOssMoELayer::incremental_forwarding(
 
     // Parallel processing for multiple tokens with many active experts
     std::vector<nntrainer::Tensor> expert_outputs(num_experts);
+#pragma omp parallel for schedule(static)
```
skykongkong8, line 290: I think we should look for other layers for the same idea too.
If tensor init declaration is taking too long, unnecessary amount of latency overhead might have been leaking from somewhere else. We were only focused on only GEMM so far, there might be more things like this. (I inspected with simpleperf and vtune though..)

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
DonghakPark: That's right!!

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
DonghakPark: That's right, I was thinking the same thing too.
But when I tried getting the topK + a, saved just the indices, and then calculated the rest up to the topK, the answer didn't come out right.
It seems like when i fetch the values or indices, it uses the tensor exactly as it is. 

Do you know why that might be happening?

EunjuYang: Could you share what you've tried?

Applications/CausalLM/layers/gpt_oss_moe_layer_cached.cpp
DonghakPark: Good Point! i will fix this

EunjuYang (approved):
LGTM.
I checked it works and its speedup is valid. Let's optimize further later :) Good work👍

PR #3494, opened 2025-09-17
Fix: Tail Handling Bug in SwiGLU AVX2 Kernel

djeong20:
## Dependency of the PR

None

<details><summary>Fix: Tail Handling Bug in SwiGLU AVX2 Kernel</summary><br />

This pull request addresses a bug related to tail handling in the SwiGLU AVX2 kernel. The following issues in the implementation have been identified:
- Block conditions incorrectly use < instead of <=, causing exact multiples of 16 or 8 to be skipped.
- The final remainder check is written as i > N, but it should be i < N.
**Self evaluation:**

Signed-off-by:  Donghyeon Jeong <dhyeon.jeong@samsung.com>

</details>

### Summary

- Fix: Tail Handling Bug in SwiGLU AVX2 Kernel

Signed-off-by:  Donghyeon Jeong <dhyeon.jeong@samsung.com>

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -892,40 +892,44 @@ void transpose_matrix(const unsigned int M, const unsigned int N,
 void swiglu(const unsigned int N, float *X, const float *Y, const float *Z) {
   size_t i = 0;
 
-  auto oldcsr = _mm_getcsr();
-  // We don't need denormals, enable:
-  // DAZ = Denormals Are Zero
-  // FTZ = Flush To Zero
-  _mm_setcsr(oldcsr | 0x8040);
+  const auto oldcsr = _mm_getcsr();
+  _mm_setcsr(oldcsr | 0x8040); // DAZ | FTZ
 
-  for (; i + 16 < N; i += 16) {
-    auto y0 = _mm256_loadu_ps(Y + i);
-    auto y1 = _mm256_loadu_ps(Y + i + 8);
-    auto z0 = _mm256_loadu_ps(Z + i);
-    auto z1 = _mm256_loadu_ps(Z + i + 8);
+  // 16-wide blocks
+  for (; i + 16 <= N; i += 16) {
+    const __m256 y0 = _mm256_loadu_ps(Y + i);
+    const __m256 y1 = _mm256_loadu_ps(Y + i + 8);
+    const __m256 z0 = _mm256_loadu_ps(Z + i);
+    const __m256 z1 = _mm256_loadu_ps(Z + i + 8);
 
     _mm256_storeu_ps(X + i, avx2_approx_swiglu(y0, z0));
     _mm256_storeu_ps(X + i + 8, avx2_approx_swiglu(y1, z1));
   }
 
-  if (i + 8 < N)
-    UNLIKELY {
-      auto y0 = _mm256_loadu_ps(Y + i);
-      auto z0 = _mm256_loadu_ps(Z + i);
-      _mm256_storeu_ps(X + i, avx2_approx_swiglu(y0, z0));
-      i += 8;
-    }
+  // One 8-wide block if available
+  if (i + 8 <= N) {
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -892,80 +892,92 @@ void transpose_matrix(const unsigned int M, const unsigned int N,
...
 
-  if (i > N)
-    UNLIKELY {
-      alignas(64) int mask[] = {-1, -1, -1, -1, -1, -1, -1, -1,
-                                0,  0,  0,  0,  0,  0,  0,  0};
+  // Remaining 1..7 elements via maskload/maskstore
+  if (i < N) {
+    const int remain = static_cast<int>(N - i); // 1..7
+
+    alignas(64) static const int mtab[16] = {-1, -1, -1, -1, -1, -1, -1, -1,
```
dkjung: I think `static` is not necessary for `mtab` at best.

It could possibly worsen the performance by placing `mtab` far from other local variables in terms of address space.
I think just declaring it as a local variable helps increase spatial locality.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -892,80 +892,92 @@ void transpose_matrix(const unsigned int M, const unsigned int N,
...
-  for (; i + 16 < N; i += 16) {
-    auto y0 = _mm256_loadu_ps(Y + i);
-    auto y1 = _mm256_loadu_ps(Y + i + 8);
-    auto z0 = _mm256_loadu_ps(Z + i);
-    auto z1 = _mm256_loadu_ps(Z + i + 8);
+  const auto oldcsr = _mm_getcsr();
+  _mm_setcsr(oldcsr | 0x8040); // DAZ | FTZ
+
+  // 16-wide blocks
+  for (; i + 16 <= N; i += 16) {
```
dkjung, line 899: At first, `<=` looks unfamiliar,
but it is correct.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -892,80 +892,92 @@ void transpose_matrix(const unsigned int M, const unsigned int N,
...
-    }
+  // One 8-wide block if available
+  if (i + 8 <= N) {
+    const __m256 y0 = _mm256_loadu_ps(Y + i);
+    const __m256 z0 = _mm256_loadu_ps(Z + i);
+    _mm256_storeu_ps(X + i, avx2_approx_swiglu(y0, z0));
+    i += 8;
+  }
 
-  if (i > N)
```
dkjung, line 919: `void swiglu(X, Y, Z)`
and
`void swiglu(X, Y, Z, a)`
didn't use the same algorithm.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
dkjung: FYI: the follow is what Google AI has given to me.


A static variable declared within a block (such as a function or a method) is stored in the data segment of the program's address space.
Specifically:
    Initialized static variables
    are stored in the .data segment. This segment contains initialized global and static variables.
    Uninitialized static variables
    are stored in the .bss segment (Block Started by Symbol). This segment is used for uninitialized global and static variables, which are automatically initialized to zero by the system before the program starts. 

Unlike local variables, which are typically stored on the stack and have a lifetime limited to the block in which they are declared, static variables persist throughout the entire execution of the program. They are allocated memory once when the program starts and retain their value across multiple calls to the function or block where they are declared.

dkjung:
This PR is interesting to review. Thanks!

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
djeong20: Fixed!

jijoongmoon (approved):
LGTM!

PR #3497, opened 2025-09-17
Poc/arm/qai8dxp qsi4cxp

skykongkong8:
## commits to be reviewed

1. 3d2fdfd1a9efa6e7282146d4c2f8218eb35b31f4

- This commit introduces simd optimized kleidiai kernel
- This commit will be a foundational reference for introducing more kernels from kleidiai
- This time, I added qai8dxp-qsi4cxp Matmul and qs4cx_f32 quantization

**Self evaluation:**

2. 40d4d6b15096f3030ebd88edb89593ddc6c394ea

- Reduces gemm unittests from cpu backend

**Self evaluation:**


dkjung:
What does `qai8dxp-qsi4cxp` stand for?

dkjung (approved):
LGTM

nntrainer/tensor/cpu_backend/arm/arm_compute_backend_fp16.cpp
```diff
@@ -459,4 +460,22 @@ void rms_norm_wrt_width_fp16_intrinsic(const float *__restrict X,
...
 }
 
+void nntr_quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
+                          void *rhs_native_mtx_qs4cx, void *rhs_scales_f32,
+                          bool transB) {
+  __fallback_nntr_quant_qs4cx_f32(n, k, rhs_native_mtx_f32,
+                                  rhs_native_mtx_qs4cx, rhs_scales_f32, transB);
+}
+
+template <>
```
dkjung, line 470: This is for what?

skykongkong8: To my knowledge, current openCL 4-bit integer kernels in the nntrainer expects to get half-precision activation flow to compute. On the other hand, this CPU function do the similar job, but expects single-precision activation.
Just In case to sync the same behavior with GPU, for easier future development, I intentionally made this function as a function template.
You can also see:
```C++
// arm_compute_backend.h
template <typename T = float>
void nntr_gemm_qai8dxp_qsi4cxp(size_t m, size_t n, size_t k,
                               void *lhs_native_mtx, void *rhs_native_mtx_qs4cx,
                               void *rhs_scales, T *dst_mtx, bool transB = true,
                               T lower_bound = std::numeric_limits<T>::lowest(),
                               T upper_bound = std::numeric_limits<T>::max());
```

jijoongmoon (approved):
LGTM

PR #3500, opened 2025-09-18
[Draft] Enable INT4-FP16 MatMul

djeong20:
This PR is to enable INT4-FP16 MatMul on GPU.
Note that the Int4QTensor class is temporarily modified to use 16-bit for scales.


nntrainer/tensor/float_tensor.cpp
```diff
@@ -705,6 +705,10 @@ Tensor &FloatTensor::dot(Tensor const &input, Tensor &output, bool trans,
   case Tdatatype::Q4_0:
     dotQnK(input, output, trans, trans_in, beta, input.getDataType());
     break;
+  case Tdatatype::QINT16:
+  case Tdatatype::QINT8:
+  case Tdatatype::QINT4:
+    dotQInteger(input, output, trans, trans_in, beta, input.getDataType());
```
djeong20, line 713: missing break;

PR #3507, opened 2025-09-23
nntrainer static build - enablement

mwlasiuk:
- This PR makes it possible to build nntrainer statically
- Note that static build requires PLUGGABLE to be disabled

**Self evaluation:**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -220,6 +220,7 @@ if get_option('enable-opencl')
   else 
     clblast_options = cmake.subproject_options()
     clblast_options.add_cmake_defines({'CMAKE_POLICY_VERSION_MINIMUM': '3.10'})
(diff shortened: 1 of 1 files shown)


api/ccapi/meson.build
```diff
@@ -26,21 +26,24 @@ ccapi_deps = [
...
+    )
+  else
+    nntrainer_ccapi_lib = shared_library('ccapi-nntrainer',
+      ccapi_src,
+      dependencies: ccapi_deps,
+      include_directories: ccapi_inc,
+      install: true,
+      install_dir: nntrainer_libdir,
+    )
+  endif
```
myungjoo, line 46: Can we use ```library()``` instead of ```shared_library``` and ```shared_library```? ref: https://mesonbuild.com/Reference-manual_functions.html#library

PR #3515, opened 2025-09-26
Make dispatching readable

mwlasiuk:
- Use struct with named fields instead of 2 arrays of 3 int
- This way dispatching kernels is more readable and less error prone

**Self evaluation:**

### DIFF
--- a/nntrainer/opencl/opencl_command_queue_manager.cpp
+++ b/nntrainer/opencl/opencl_command_queue_manager.cpp
@@ -324,31 +324,17 @@ bool CommandQueueManager::enqueueSVMUnmap(void *svm_ptr, cl_event *event) {
  * @brief Function to initiate execution of the command queue.
  *
  * @param kernel OpenCL kernel
- * @param work_groups_count Total number of work items that will execute the
- * kernel function
- * @param work_group_size Number of work items that make up a work group
+ * @param sizes Sizes for the dispatch
  * @param event Object that identifies this command and can be used to query
  * or wait for this command to complete
  * @return true if command queue execution is successful or false otherwise
  */
 bool CommandQueueManager::DispatchCommand(
-  Kernel kernel, const int (&work_groups_count)[3],
-  const int (&work_group_size)[3], cl_event *event,
-  std::vector<cl_event> events_to_wait) {
+  Kernel kernel, const DispatchSize &sizes, cl_event *event,
+  const std::vector<cl_event> &events_to_wait) {
 
-  // work_dim of 2 has been hardcoded, might be modified later based on
-  // requirements
-
-  // setting the local_work_size referred to as the size of the
-  // work-group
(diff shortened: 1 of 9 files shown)


skykongkong8 (commented):
Please check for Android build failure

mwlasiuk:
> Please check for Android build failure

Done

nntrainer/opencl/opencl_command_queue_manager.h
```diff
@@ -21,6 +21,21 @@
...
+ * @brief Sizes for OpenCL kernel enqueue function
+ *
+ */
+struct DispatchSize {
+  size_t local_x = 0;
+  size_t local_y = 0;
+  size_t local_z = 0;
+  size_t global_x = 0;
+  size_t global_y = 0;
+  size_t global_z = 0;
```
djeong20, line 36: How about we set the default values to 1? This way, you won't need to assign all the variables.
For instance

```C++
opencl::DispatchSize sizes{};
    sizes.local_x = chosen_local;
    sizes.local_y = 1;
    sizes.local_z = 1;
    sizes.global_x = dim;
    sizes.global_y = 1;
    sizes.global_z = 1;
```
The code above can be simplified.
```C++
opencl::DispatchSize sizes{};
    sizes.local_x = chosen_local;
    sizes.global_x = dim;
```

nntrainer/opencl/opencl_command_queue_manager.h
```diff
@@ -21,6 +21,21 @@
 
 namespace nntrainer::opencl {
 
+/**
+ * @struct DispatchSize contains wrappers for managing OpenCL command
+ * queue
+ * @brief Sizes for OpenCL kernel enqueue function
+ *
+ */
+struct DispatchSize {
```
djeong20, line 30: As you might know, there are restrictions on both local and global work sizes. In OpenCL 1.x, if you specify local_work_size when calling clEnqueueNDRangeKernel, then the global_work_size for each dimension must be an integer multiple of the corresponding local size. However, in OpenCL 2.0, this requirement can be relaxed if the device supports it. 

For future work, it would be beneficial for `DispatchSize` to accommodate these scenarios.

nntrainer/opencl/opencl_command_queue_manager.cpp
```diff
@@ -365,24 +351,21 @@ bool CommandQueueManager::DispatchCommand(
...
 
+/**
+ * @brief Overloaded function to initiate execution of the command queue.
+ *
+ * @param kernel_ptr reference of OpenCL kernel shared_ptr
+ * @param sizes Sizes for the dispatch
+ * @param event Object that identifies this command and can be used to query
+ * or wait for this command to complete
+ * @return true if command queue execution is successful or false otherwise
+ */
```
songgot, line 362: There is no description for '@events_to_wait'.

PR #3517, opened 2025-09-29
[ONNX] Multi token qwen model with attention mask

sachin-nntrainer:
Multi token Qwen 1.7B model with attention mask

This Qwen model is able to generate multi token output. Attention mask size is fixed (Upto to max length) and passed as input to the model. For every token generation mask remains constant. This is done to handle dynamic inputs therefore fixing input size and mask size before first token generation itself.  

Currently we are generating a fixed number of tokens. EOS handling is to be added.

Self evaluation:

Build test: []Passed [ ]Failed [X]Skipped
Run test: []Passed [ ]Failed [X]Skipped

Signed-off-by: Sachin Singh [sachin.3@samsung.com](mailto:sachin.3@samsung.com)


Applications/ONNX/python/qwen3/generateResult.py
```diff
@@ -0,0 +1,85 @@
+import torch
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from modelling_qwen3 import  NNTrainerQwen3ForCausalLM, Qwen3RotaryEmbedding
```
songgot, line 3: Please check if there are any typos in the module name.
'modeling'?

Applications/ONNX/python/qwen3/generateResult.py
```diff
@@ -0,0 +1,85 @@
...
+
+enc = tokenizer(prompt, return_tensors="pt", truncation=True, max_length=max_len)
+cur_len = enc.input_ids.size(1)
+
+input_ids = torch.full((1, max_len), tokenizer.pad_token_id)  # preallocate with PAD
+input_ids[0, :cur_len] = enc.input_ids[0]
+
+generated = input_ids.clone()
+
+for step in range(20):  # generate 20 tokens
```
songgot, line 29: What if we defined the magic number 20 as a constant?
```NUM_TOKENS_TO_GENERATE = 20  ```

PR #3519, opened 2025-10-02
[ cpu_backend ] Enable practical use of qsi4cxp_qs4cxs1s0 GEMM with openMP multithreading and automatic ukernel candidate selection

skykongkong8:
## Dependency of the PR

None.

### Summary

This Pull Request introduces even more optimized version of qsi4cxp_qs4cxs1s0 GEMM first propsed from #3497 
Optimization technique s.t.:
- neon SIMD
- openMP based multithreading
- automatic ukernel selection

As a result, this patch accelerated approximately x4 times faster GEMM computation latency:
```bash
// Tested on Galaxy S25U
[ RUN      ] nntrainer_cpu_backend_standalone.qai8dxp_qsi4cxp_512x768x2048
BEFORE : 8639844 ns 8639 us 8 ms
AFTER : 2524531 ns 2524 us 2 ms
```
In my inspection, this is more than 2~3 times faster than previous `Q4_0`
FYI) `gemm_q4_0: 6986094 ns 6986 us 6 ms`

For practical using sample, refer to `unittest_nntrainer_cpu_backend_fp16.cpp`

    **Self evaluation:**
    

nntrainer/tensor/cpu_backend/arm/kai/matmul_clamp_f32_qai8dxp_qsi4cxp/kai_matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_4x4x32_neon_i8mm.c
```diff
@@ -0,0 +1,307 @@
...
+  KAI_ASSERT((k_internal % 2) == 0);
+
+  return kai_nr * ((k_internal / 2) + kai_num_bytes_multiplier_rhs +
+                   kai_num_bytes_sum_rhs);
+}
+
+size_t
+kai_get_m_step_matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_4x4x32_neon_i8mm(void) {
+  return kai_m_step;
+}
```
myungjoo, line 68: is this function really exported to other modules?

myungjoo: If this value is accessed by external modules in the middle of critical calculation (performance impacting), you can define this functin as a static-inline function in the header and have kai_m_step declared in the header. Moreoever, for conventional C++ principles, this is better to be a static value of a public class, removing all the worries.

myungjoo: What I worry: calling this function in a loop or other module.

nntrainer/tensor/cpu_backend/fallback/fallback_fp16.cpp
```diff
@@ -234,4 +234,58 @@ void clamp(const _FP16 *input, _FP16 *output, size_t length, _FP16 lower_bound,
...
+  __fallback_nntr_quant_qs4cx_f32(n, k, rhs_native_mtx_f32,
+                                  rhs_native_mtx_qs4cx, rhs_scales_f32, transB);
+}
+
+template <>
+uint32_t nntr_gemm_qai8dxp_qsi4cxp_unpacked(
+  size_t m, size_t n, size_t k, void *lhs_native_mtx_f32,
+  void *rhs_native_mtx_qs4cx, void *rhs_scales_f32, float *dst_mtx_f32,
+  bool transB, float lower_bound, float upper_bound) {
+  __fallback_nntr_gemm_qai8dxp_qsi4cxp_unpacked(
```
myungjoo, line 250: Missing return?

nntrainer/tensor/float_tensor.cpp
```diff
@@ -971,6 +968,21 @@ Tensor &FloatTensor::dotQInteger(Tensor const &input, Tensor &output,
   unsigned int K = getDim().width();
   unsigned int N = output.getDim().width();
 
+#ifndef ENABLE_OPENCL
+#ifdef ENABLE_FP16
+  if (input.q_scheme() == QScheme::PER_CHANNEL_AFFINE) {
+    uint32_t opt_kernel_idx = (M == 1) ? 1 : 5;
+    nntr_gemm_qai8dxp_qsi4cxp_packed(
```
myungjoo, line 975: Missing return statement?

throw of LINE 983 is always called after this!

nntrainer/tensor/cpu_backend/fallback/fallback.h
```diff
@@ -24,6 +24,93 @@
...
+ * groups.
+ *
+ * @param n N length of the matrix
+ * @param k K length of the matrix
+ * @param rhs_native_mtx_f32 matrix data before quantization to load
+ * @param rhs_native_mtx_qs4cx matrix data after quantization to stroe
+ * @param rhs_scales_f32 matrix quant scale after quantization to stroe
+ * @param transB
+ */
+void nntr_quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
```
myungjoo, line 40: double definition?

nntrainer/tensor/cpu_backend/fallback/fallback.h
```diff
@@ -411,6 +498,93 @@ void inv_sqrt_inplace(const unsigned int N, _FP16 *X);
...
```
myungjoo, line 515: double definition? (not built in our github action scenarios?)

skykongkong8: Thanks for your thorough review! I wonder why this problem haven't detected previously..!

nntrainer/tensor/float_tensor.cpp
skykongkong8: Aha, I missed this one! It should be

```C++
#else
  throw std::runtime_error(
    "Error: FP16 should be enabled for QINT4 Dot on CPU");
 #endif
```

Not
```C++
#endif
  throw std::runtime_error(
    "Error: FP16 should be enabled for QINT4 Dot on CPU");
```

nntrainer/tensor/cpu_backend/arm/kai/matmul_clamp_f32_qai8dxp_qsi4cxp/kai_matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_4x4x32_neon_i8mm.c
skykongkong8: I see your worries and let me explain.

First of all, `kai_get_m_step_matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_4x4x32_neon_i8mm` is not actually being called from anywhere in the current status.
Then why should we keep this function? -> Because it is component of `kai_matmul_ukernel_f32_qa8dxp_qs4cxp` struct, which is general format of using kleidiai GEMM kernels, and we imported only some of them, and planning to introduce more in the future. In that case, we might need  some `m_step` calling functions, but even in such case, it will not be called in a loop, or in other modules except for `neon_kleidiai.cpp`

nntrainer/tensor/cpu_backend/fallback/fallback_fp16.cpp
skykongkong8: Fixed! Sorry for trivial misses!

nntrainer/tensor/cpu_backend/arm/kai/matmul_clamp_f32_qai8dxp_qsi4cxp/kai_matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_4x4x32_neon_i8mm.c
skykongkong8: If you are worried about the latency-perspective, I think it won't affect the total GEMM latency meaningfully...
However, if you are worried about SW-design perspective, maybe we should undergo major refactoring for detaching all

```C++
struct kai_matmul_clamp_f32_qai8dxp_qsi4cxp_ukernel {
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_m_step_func_t get_m_step;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_n_step_func_t get_n_step;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_mr_func_t get_mr;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_nr_func_t get_nr;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_nr_func_t get_kr;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_sr_func_t get_sr;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_lhs_packed_offset_func_t
    get_lhs_packed_offset;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_rhs_packed_offset_func_t
    get_rhs_packed_offset;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_dst_offset_func_t get_dst_offset;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_get_dst_size_func_t get_dst_size;
  kai_matmul_clamp_f32_qai8dxp_qsi4cxp_run_matmul_func_t run_matmul;
};
``` 

struct into static value of a public class since [kleidiai](https://github.com/ARM-software/kleidiai) is getting some const values with just normal functions.

PR #3525, opened 2025-10-14
Int4 working

gkisalapl:


nntrainer/tensor/int4_tensor.cpp
```diff
@@ -391,8 +391,8 @@ void Int4QTensor::read(ReadSource src, size_t start_offset,
   }
 
   checkedRead(src, (char *)getData(), sz,
-              "[Int4QTensor::read] operation failed", start_offset,
-              read_from_offset);
+              "[Int4QTensor::read] operation failed",
+              start_offset - sizeof(uint16_t), read_from_offset);
```
djeong20, line 395: FYI, you can remove `read_quantization_info` instead of modifying the offset.

PR #3528, opened 2025-10-20
Fix static windows build

gkisalapl:
Make possible to build nntrainer in static or shared version on
Windows platform

Use following commandt for static build:

meson setup --wipe --native-file windows-native.ini builddir -Ddefault_library=static -Denable-pluggable=false -Denable-opencl=true
meson compile -C builddir

### DIFF
--- a/Applications/CausalLM/meson.build
+++ b/Applications/CausalLM/meson.build
@@ -50,6 +50,12 @@ endif
 tokenizer_dependencies = []
 tokenizer_path = meson.current_source_dir() / 'lib' / 'libtokenizers_c.a'
 
+cpp_args = []
+
+if get_option('default_library') == 'shared'
+    cpp_args += [ '-DPLUGGABLE' ]
+endif
+
 causallm = shared_library('causallm',
     causallm_src,
     dependencies: [
@@ -63,7 +69,7 @@ causallm = shared_library('causallm',
     install: true,
     install_dir: application_install_dir,
     link_args: [tokenizer_path],
-    cpp_args: '-DPLUGGABLE'
+    cpp_args: cpp_args
 )
 
 causallm_dep = declare_dependency(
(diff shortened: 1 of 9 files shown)


Applications/CausalLM/meson.build
```diff
@@ -63,7 +69,7 @@ causallm = shared_library('causallm',
     install: true,
     install_dir: application_install_dir,
     link_args: [tokenizer_path],
-    cpp_args: '-DPLUGGABLE'
+    cpp_args: cpp_args
```
djeong20, line 72: Is it okay to have the same naming?

meson_options.txt
```diff
@@ -63,3 +63,6 @@ option('nnstreamer-subplugin-install-path', type: 'string', value: 'lib/nnstream
 # application related options
 option('enable-transformer', type: 'boolean', value: false)
 option('libiomp_root', type:'string', value: './libiomp_win')
+
+# use layers as pluggable modules
+option('enable-pluggable', type: 'boolean', value: true)
```
djeong20: how about adding `default_library=static` and `enable-pluggable=false` to windows-native.ini?

djeong20: Could the following condition be possible? 
default_library is static, and enable-pluggable is true; default_library is not static, and enable-pluggable is false.
If this is not feasible, we might need to remove the new option.

djeong20 (approved):
Overall, LGTM!
Please check the comments

Applications/CausalLM/meson.build
gkisalapl: I think it's ok, it's key - value relation

meson_options.txt
gkisalapl: configuration in which default_library is static, and enable-pluggable is true would not work

djeong20: In that case, we could remove this option and revise it.

```
cpp_args = []

if get_option('default_library') != 'static'
    cpp_args += [ '-DPLUGGABLE' ]
endif
```

gkisalapl: Sounds good

jijoongmoon (approved):
LGTM

djeong20:
This work is included in #3541.

PR #3531, opened 2025-10-22
[Int4] Add support in OpenCL GEMV/GEMM kernels for unaligned data.

mnalewaj:
Before this PR:
- the K dimension should be divisible by scale_group_size (32).
- and N dimension should be divisible by 32 (it is connected with kernel input format).
- the dimenstion M can have any value.

After application this PR:
- the K dimension should be divisible by 4.
- the N dimension should be divisible by 8 (it is connected with kernel input format).
- the dimenstion M can have any value.

But quantization weights should be generated in the following way.
NOTE: These limitations are connected with input data format which GEMV/GEMM kernels expects as input.

Weights:
	- data in K dimention should be aligned to scale_group_size (32)
	- data in N dimention should be aligned to N_BLOK_SIZE (32)
	- values of padded data should be equal 0
Scales:
	- data are in column major order when column size N is aligned to 32
	- values of padded data should be equal 0

Inputs (activations):
	- data should be not aligned without any padding

For reference of quantization please check method:
void Int4Utils::quantizeAndRepack()

In that method we are using dynamic padding adding complex conditions (it is working and tested so we left it as is),
but there is a simpler way to quantization of weights. We suggest do the quantization in the following way:
1. Add padding to unaligned weights with 0 values in columns and rows (K aligned to 32, N aligned to 32).
2. Then quantize these aligned weights by normal algorithm (even without padding).
These steps should give a shape of data that the kernels are expecting.

### DIFF
--- a/nntrainer/tensor/cl_operations/blas_kernels.cpp
+++ b/nntrainer/tensor/cl_operations/blas_kernels.cpp
@@ -18,6 +18,14 @@
 
 namespace nntrainer {
 
+static inline unsigned int ceil_div(unsigned int a, unsigned int b) {
+  return (a + b - 1) / b;
+};
+
+static inline unsigned int align(unsigned int a, unsigned int b) {
+  return (a % b == 0) ? a : a - a % b + b;
+};
+
 void gemv_int4_async_cl(std::vector<void *> weights,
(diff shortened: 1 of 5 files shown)


djeong20:
This work is included in #3539.

PR #3538, opened 2025-10-29
Update Neural Network Load Regex for Windows Compatibility

djeong20:
This pull request updates the neuralnet load regex to use a semicolon (;) instead of a colon (:).
This is to have compatibility with the Windows environment.
Additionally, this patch adds a fallback in the FloatTensor dot method for handling multiple inputs.

**Self-evaluation:**


jijoongmoon (approved):
LGTM

PR #3546, opened 2025-11-04
Configurable kernels cache dir

gkisalapl:
1. Add API for setting kernels cache dir
2. Compile kernels in lazy way

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -21,7 +21,7 @@ option('opencl-kernel-path', type: 'string', value: 'nntrainer_opencl_kernels')
 option('enable-mmap', type: 'boolean', value: true)
 option('enable-npu', type: 'boolean', value: false)
 option('mmap-read', type: 'boolean', value: true)
-option('enable-kernel-caching', type: 'boolean', value: false)
+option('enable-kernel-caching', type: 'boolean', value: true)
 
 # dependency conflict resolution
 option('capi-ml-inference-actual', type: 'string', value: 'capi-ml-inference',
--- a/nntrainer/cl_context.cpp
+++ b/nntrainer/cl_context.cpp
@@ -73,15 +73,9 @@ void ClContext::initialize() noexcept {
       ml_loge("Error: ClContext::initialize() failed");
       return;
     }
-    if (KERNEL_CACHE_ENABLED) {
-      std::filesystem::create_directories(opencl::Program::DEFAULT_KERNEL_PATH);
-    }
 
-    initBlasClKernels();
-    initAttentionClKernels();
     add_default_object();
     setMemAllocator(std::make_shared<MemAllocator>());
-
   } catch (std::exception &e) {
     ml_loge("cl_context: registering layers failed!!, reason: %s", e.what());
   } catch (...) {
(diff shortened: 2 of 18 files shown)


nntrainer/cl_context.h
```diff
@@ -249,19 +244,15 @@ class ClContext : public Context, public Singleton<ClContext> {
...
-  bool attention_kernels_initialized = false;
-
   FactoryMap<nntrainer::Layer> factory_map;
 
   template <typename Args, typename T> struct isSupportedHelper;
 
   // global map to store initialized opencl::Kernel
   inline static OclKernelMap ocl_kernel_map;
 
+  std::string kernels_cache_path_ = std::filesystem::current_path().string();
```
djeong20: let's use the default kernel path here.

```suggestion
  std::string kernels_cache_path_ = stringify(OPENCL_KERNEL_PATH);
```

nntrainer/cl_context.cpp
```diff
@@ -73,15 +73,9 @@ void ClContext::initialize() noexcept {
...
       return;
     }
-    if (KERNEL_CACHE_ENABLED) {
-      std::filesystem::create_directories(opencl::Program::DEFAULT_KERNEL_PATH);
-    }
 
-    initBlasClKernels();
-    initAttentionClKernels();
     add_default_object();
     setMemAllocator(std::make_shared<MemAllocator>());
```
djeong20, line 78: Although we are trying to initialize in a lazy way, I think it would be great to have a separate function to initialize kernels, something like `initializeKernels()`. What do you think?

nntrainer/cl_context.h
gkisalapl: ok

nntrainer/cl_context.cpp
gkisalapl: I agree, do you think that adding such functions should be part of this PR ?

nntrainer/cl_context.cpp
```diff
@@ -263,8 +228,11 @@ bool ClContext::clCreateKernel(std::string &kernel_string,
 
   // reading binary
   std::string binary_file_path =
-    opencl::Program::DEFAULT_KERNEL_PATH + "/" +
-    std::to_string(program.GetKernelHash(kernel_string, "")) + ".cl.bin";
+    std::filesystem::path(kernels_cache_path_)
```
jijoongmoon, line 231: We might need to set the default path for binary in case of enable-kernel-caching=true and no given path by user.

nntrainer/cl_context.cpp
```diff
@@ -312,4 +280,22 @@ template const int ClContext::registerFactory<nntrainer::Layer>(
   const FactoryType<nntrainer::Layer> factory, const std::string &key,
   const int int_key);
 
+void ClContext::setKernelsCachePath(const std::string &kernels_cache_path) {
+  NNTR_THROW_IF(cl_kernels_initialized_, std::runtime_error)
+    << "OpenCL kernels already initialized kernels path should be set before "
+       "initialization";
```
songgot, line 286:
```suggestion
<< "OpenCL kernels already initialized. Kernels path should be set before initialization.";
```

meson_options.txt
```diff
@@ -21,7 +21,7 @@ option('opencl-kernel-path', type: 'string', value: 'nntrainer_opencl_kernels')
 option('enable-mmap', type: 'boolean', value: true)
 option('enable-npu', type: 'boolean', value: false)
 option('mmap-read', type: 'boolean', value: true)
-option('enable-kernel-caching', type: 'boolean', value: false)
+option('enable-kernel-caching', type: 'boolean', value: true)
```
djeong20, line 24: @jijoongmoon Should caching be enabled by default?

jijoongmoon: If it is working properly and could save the latency, yes. We can merge it and fix later if we need.

PR #3551, opened 2025-11-07
Apply RAII for ClContext internals

gkisalapl:
Make ContextManager, CommandQueueManager and BufferManager owned by ClContext instead of global singleton objects It gives us better control over lifetime of this objects

**Self-evaluation:**

### DIFF
--- a/nntrainer/cl_buffer_manager.cpp
+++ b/nntrainer/cl_buffer_manager.cpp
@@ -20,17 +20,17 @@ namespace nntrainer {
 // to-do: Implementation to be updated with array of Buffer objects if required
 // fp16 Buffer objects to be added in future
 void ClBufferManager::initBuffers() {
-  inBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  inBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  inBufferC = new opencl::Buffer(context_inst_, unused_buffer_bytes, true);
-  outBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
-  outBufferB = new opencl::Buffer(context_inst_, unused_buffer_bytes, false);
+  inBufferA = new opencl::Buffer(context_manager_, buffer_size_bytes, true);
+  inBufferB = new opencl::Buffer(context_manager_, buffer_size_bytes, true);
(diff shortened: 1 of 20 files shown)


nntrainer/cl_buffer_manager.cpp
```diff
@@ -43,11 +43,11 @@ ClBufferManager::~ClBufferManager() {
...
 
-  context_inst_.releaseSVMRegion(data_input);
+  context_manager_.releaseSVMRegion(data_input);
   for (unsigned int i = 0; i < max_qs; ++i) {
-    context_inst_.releaseSVMRegion(scale_vec[i]);
-    context_inst_.releaseSVMRegion(quant_vec[i]);
-    context_inst_.releaseSVMRegion(output_vec[i]);
+    context_manager_.releaseSVMRegion(scale_vec[i]);
+    context_manager_.releaseSVMRegion(quant_vec[i]);
+    context_manager_.releaseSVMRegion(output_vec[i]);
```
myungjoo, line 50: Is it possible to call ```~ClContext()``` before this point so that ClBufferManager (```context_manager_.ReleaseContext();```) is already deleted at this point?

gkisalapl: Thank You for catching this, I've already fixed that issue. Now order of destruction of ClContext members will ensure that BufferManager will be destructed after CLContextManager.

nntrainer/cl_context.h
```diff
@@ -55,30 +55,32 @@ class ClContext : public Context, public Singleton<ClContext> {
...
-
   /**
    * @brief   Default constructor
    */
-  ClContext() : Context(std::make_shared<ContextData>()) {}
+  ClContext() :
+    Context(std::make_shared<ContextData>()),
+    context_manager_(),
+    command_queue_manager_(context_manager_),
+    buffer_manager_(context_manager_) {}
```
jijoongmoon, line 65: Let's consider there are multiple networks in one main. In this case, we call the engine once, then everything releated with this is just call_once. only one object exists. I think engine, context for the managing the binary is ok to be in one object, but the queue manager and buffer manager should be only one for one network. Let me know your opinion.

gkisalapl: ok

PR #3552, opened 2025-11-10
Enhancement: Thread Pool Reset Functionality

djeong20:
This pull request adds a feature to reset the thread pool in the manager class.
This patch allows the reduction of thread pool usage when it is not needed.

**Self-evaluation:**

### DIFF
--- a/nntrainer/utils/bs_thread_pool_manager.hpp
+++ b/nntrainer/utils/bs_thread_pool_manager.hpp
@@ -35,23 +35,42 @@ public:
   std::size_t select_k_quant_thread_count(unsigned int M, unsigned int N,
                                           unsigned int K);
 
-  BS::thread_pool<> &getThreadPool() { return pool_; }
+  BS::thread_pool<> &getThreadPool() {
+    if (reset_) {
+      reset_ = false;
+      pool_->reset(std::thread::hardware_concurrency());
+    }
+
+    return *pool_;
+  }
 
   /**
    * @brief Construct a new Thread Pool Manager object
    *
    */
-  ThreadPoolManager() : pool_(std::thread::hardware_concurrency()) {}
+  ThreadPoolManager() : reset_(false) {
+    pool_ =
+      std::make_unique<BS::thread_pool<>>(std::thread::hardware_concurrency());
+  }
+
   /**
    * @brief Destroy the Thread Pool Manager object
-   *
    */
   ~ThreadPoolManager() = default;
 
+  void shutdown() {
+    if (!pool_)
+      return;
+
+    pool_->purge();
+    pool_->reset();
+    reset_ = true;
+  }
+
 private:
-  BS::thread_pool<> pool_;
+  std::unique_ptr<BS::thread_pool<>> pool_;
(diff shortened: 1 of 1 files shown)


nntrainer/utils/bs_thread_pool_manager.hpp
```diff
@@ -35,23 +35,42 @@ class ThreadPoolManager : public Singleton<ThreadPoolManager> {
...
   ~ThreadPoolManager() = default;
 
+  void shutdown() {
+    if (!pool_)
+      return;
+
+    pool_->purge();
+    pool_->reset();
+    reset_ = true;
+  }
```
myungjoo, line 68: Q1. Who's calling shutdown()?
Q2. What's purge()? Can this be called automatically by the automatic sequence? (unique_ptr --> class destructor --> purge)

Adding shutdown() for a unique_ptr is very wierd.

djeong20: The `shutdown()` method should be called by the object that initializes the `ThreadPoolManager`. This method does not destroy the existing thread pool; instead, it eliminates the threads that are currently in the pool. 

The `purge()` method is used to remove all tasks waiting in the queue; however, it does not affect tasks that are currently running. On the other hand, the `reset()` method is designed to destroy all threads in the pool.

myungjoo (approved):

nntrainer/utils/bs_thread_pool_manager.hpp
```diff
@@ -35,23 +35,42 @@ class ThreadPoolManager : public Singleton<ThreadPoolManager> {
...
 
-  BS::thread_pool<> &getThreadPool() { return pool_; }
+  BS::thread_pool<> &getThreadPool() {
+    if (reset_) {
+      reset_ = false;
+      pool_->reset(std::thread::hardware_concurrency());
+    }
+
+    return *pool_;
+  }
```
songgot, line 45: Doesn't this cause a race condition in a multithreaded environment?
If so, what about the code below?
```
#include <mutex>

std::mutex reset_mutex;

BS::thread_pool<> &getThreadPool() {
  std::lock_guard<std::mutex> lock(reset_mutex);
  if (reset_) {
    reset_ = false;
    pool_->reset(std::thread::hardware_concurrency());
  }
  return *pool_;
}

void shutdown() {
  std::lock_guard<std::mutex> lock(reset_mutex);
  if (!pool_)
    return;

  pool_->purge();
  pool_->reset();
  reset_ = true;
}
```

myungjoo: @djeong20 PTAL

djeong20: Thank you for the review! :)

I want to clarify that this pull request (PR) addresses an issue related to the BS thread pool. However, I have discovered that the `reset` functionality alone is insufficient to resolve the problem I’m facing. As a result, I may need to completely revise this work. I will definitely consider this option when it comes to destroying the thread pool itself.

PR #3543, opened 2025-11-04
[PR #2897] Add arm-native build and test on Ubuntu @open sesame 12/19 09:42

djeong20:
https://github.com/nnstreamer/nntrainer/pull/2897#issue-2805918149

PR #3558, opened 2025-11-12
Enable ARM64 Support for Windows Builds

djeong20:
This pull request enables Windows builds for nntrainer on ARM64.
It also updates the Windows resources to support ARM64 and resolves build errors.

**Self-evaluation:**

### DIFF
--- a/.github/workflows/windows_arm.yml
+++ b/.github/workflows/windows_arm.yml
@@ -1,0 +1,34 @@
+name: "Build Test - Windows on ARM Meson"
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ windows-11-arm ]
+
+    name: Windows Meson build & test 
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: ${{ github.event.pull_request.commits }}
+      - name: Install submodules
+        run: git submodule sync && git submodule update --init --recursive
+      - name: Install Python Dependencies
+        run: pip install meson==1.7.2 ninja
+      - name: Prepare MSVC
+        uses: bus1/cabuild/action/msdevshell@v1
+        with:
+          architecture: arm64
+      - name: Prepare Build
+        run: meson setup --native-file windows-native.ini -Denable-blas=false builddir
+      - name: Run Build
+        run: meson compile -C builddir
+      - name: Run Test Suite
+        run: meson test -C builddir


nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -20,6 +20,15 @@
 #include <stdexcept>
 #include <tensor_dim.h>
 
+#if defined(_WIN32)
+#ifdef max
```
myungjoo, line 25: In case both fallback.h and arm_compute_backend.h are included in a c/cpp file,
add ```#ifndef NOMINMAX```.

djeong20:
> It would be good to add Windows arm build to github actions which seems to be possible now:
> 
> https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

Nice :) I'll add it to https://github.com/nnstreamer/nntrainer/pull/3543 after resolving the build issue.

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.h
```diff
@@ -26,7 +26,9 @@
 #undef max
 #undef min
 #endif
+#ifndef NOMINMAX
 #define NOMINMAX
+#endif
```
skykongkong8: Is this has to do with something with ARM build?

djeong20: No, Myungjoo recommends this patch.

.github/workflows/windows_arm.yml
```diff
@@ -0,0 +1,34 @@
...
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ windows-11-arm ]
+
+    name: Windows Meson build & test
```
songgot, line 14: How about removing the space at the end of the line?

nntrainer/tensor/cpu_backend/fallback/fallback.h
```diff
@@ -21,6 +21,17 @@
...
+#if defined(_WIN32)
+#ifdef max
+#undef max
+#undef min
+#endif
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#include <windows.h>
+#endif
```
songgot, line 33: The same Windows-related preprocessor code is repeated in multiple header files.
Would you be able to make a header file?
```
/* windows_compat.h */
#ifndef __WINDOWS_COMPAT_H__
#define __WINDOWS_COMPAT_H__

#if defined(_WIN32)
#ifdef max
#undef max
#undef min
#endif
#ifndef NOMINMAX
#define NOMINMAX
#endif
#include <windows.h>
#endif

#endif // __WINDOWS_COMPAT_H__
```
And use.
```
#include "windows_compat.h"
```

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -402,54 +415,69 @@ void copy_int4_to_fp32(const unsigned int N, const uint8_t *X, float *Y) {
   // keep in mind that : len(X) = N, and len(Y) = 2*N
 
   // processing batch of 16
-  float32x4_t y0, y1, y2, y3;
-  float32x4_t y4, y5, y6, y7;
-
-  uint8_t low0, low1, high0, high1;
+  float y0_array[4], y1_array[4], y2_array[4], y3_array[4];
+  float y4_array[4], y5_array[4], y6_array[4], y7_array[4];
```
jijoongmoon, line 419: Is it just a compiler preprocessing matter, or will there be performance degradation? guessing.. just compiler preprocessing. Isn't it?

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -402,54 +415,69 @@ void copy_int4_to_fp32(const unsigned int N, const uint8_t *X, float *Y) {
...
       } else {
-        y7[2 * (i - 6)] = high0;
-        y7[2 * (i - 6) + 1] = high1;
+        y7_array[2 * (i - 6)] = high0;
+        y7_array[2 * (i - 6) + 1] = high1;
       }
     }
+
+    // Store arrays back to vectors
+    float32x4_t y0 = vld1q_f32(y0_array);
```
jijoongmoon, line 472: Oh.. this is additional computation then.

.github/workflows/windows_arm.yml
djeong20: sure :)

nntrainer/tensor/cpu_backend/fallback/fallback.h
djeong20: We could, but I don’t think it’s a good idea. This patch is meant to remove `min` and `max`, and there could be cases where `min` and `max` are still going to be used.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
djeong20: Yes, this change is necessary.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
myungjoo: If a code with ```#include "arm_compute_backend.h"``` has already included ```windows.h```, the definitions for max/min from ```windows.h``` will be lost and ```windows.h``` in this lines will be ignored, losing max/min.

This should be:

```C
#if defined(_WIN32)
#ifndef NOMINMAX

#ifdef max
#undef max
#undef min
#endif

#define NOMINMAX
#endif

#include <windows.h> // This assume that there is duplicated-include check in windows.h

#endif
```

djeong20: Applied! Thanks 👍

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.h
myungjoo: It is not to do #undef again. If it's done twice, the second include windows.h will be skipped and max/min will be lost.

.github/workflows/windows_arm.yml
```diff
@@ -0,0 +1,34 @@
...
+      matrix:
+        os: [ windows-11-arm ]
+
+    name: Windows Meson build & test 
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: -${{ github.event.pull_request.commits }}
```
myungjoo: negative fetch-depth? isn't it "undefined behavior"?

meson.build
```diff
@@ -447,7 +455,7 @@ extra_defines += '-DNNTR_NUM_THREADS=@0@'.format(get_option('nntr-num-threads'))
 message('set nntrainer num threads=@0@'.format(get_option('nntr-num-threads')))
 openmp_dep = dummy_dep
 if get_option('enable-openmp')
- if cxx_compiler_id == 'msvc'
+ if cxx_compiler_id == 'msvc' and (arch == 'x86' or arch == 'x86_64')
```
myungjoo, line 459: With arm64, you are passing to 
```
 else
     openmp_dep = dependency('openmp')
 endif
```

Is it ok?

.github/workflows/windows_arm.yml
djeong20: I'm not too sure about this. I used the same script as the current Windows action, changed the OS and build option.

meson.build
djeong20: Yes. To explain, the current work isn’t about defining OpenMP, but about using OpenMP 5.0 instead of the older OpenMP 2.0 standard. On arm64, it would use the OpenMP 2.0 standard.

myungjoo: Add such info as a comment in the script.

.github/workflows/windows_arm.yml
myungjoo: From actions/checkout:

https://github.com/actions/checkout/blob/1af3b93b6815bc44a9784bd300feb67ff0d1eeb3/src/input-helper.ts#L105-L109
```ts
  result.fetchDepth = Math.floor(Number(core.getInput('fetch-depth') || '1'))
  if (isNaN(result.fetchDepth) || result.fetchDepth < 0) {
    result.fetchDepth = 0
  }
  core.debug(`fetch depth = ${result.fetchDepth}`)
```

It's always 0.

gkisalapl (approved):

PR #3560, opened 2025-11-13
Disable log on Windows

djeong20:
This pull request disables the logging functionality for Windows.
There's an issue with the Windows application in creating and accessing the log file directory using nntrainer_logger.
As a result, I have decided to turn off the logging feature on Windows.

**If anyone has a better solution for logging that does not involve nntrainer_logger, please let me know.**

### DIFF
--- a/nntrainer/nntrainer_log.h
+++ b/nntrainer/nntrainer_log.h
@@ -40,42 +40,49 @@
 #elif defined(__ANDROID__)
 #include <android/log.h>
 
-#define ml_logi(...) \
+#define ml_logi(...)                                                           \
   __android_log_print(ANDROID_LOG_INFO, TAG_NAME, __VA_ARGS__)
 
-#define ml_logw(...) \
+#define ml_logw(...)                                                           \
   __android_log_print(ANDROID_LOG_WARN, TAG_NAME, __VA_ARGS__)
 
-#define ml_loge(...) \
+#define ml_loge(...)                                                           \
   __android_log_print(ANDROID_LOG_ERROR, TAG_NAME, __VA_ARGS__)
 
-#define ml_logd(...) \
+#define ml_logd(...)                                                           \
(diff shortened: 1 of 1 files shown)


myungjoo:
Actually, nntrainer_logger needs to get envvar or .ini value of "log-level" and ignore ALL logging that is lower than the required "log-level".

E.g., (nntrainer_logger.cpp)
```C++
void __nntrainer_log_print(nntrainer_loglevel loglevel,
                           const std::string format_str, ...) {
+ if (loglevel > required_loglevel)
+   return;
  int final_n, n = ((int)format_str.size()) * 2;
  std::unique_ptr<char[]> formatted;
  va_list ap;
  while (1) {
    formatted.reset(new char[n]);
    std::strncpy(&formatted[0], format_str.c_str(), format_str.size());
    va_start(ap, format_str);
    final_n = vsnprintf(&formatted[0], n, format_str.c_str(), ap);
    va_end(ap);
    if (final_n < 0 || final_n >= n)
      n += abs(final_n - n + 1);
    else
      break;
  }

```

Anyway, you may start with 
```C++
#elif defined(_WIN32) && !defined(DEBUG_MODE_YOU_MAY_DEFINE)

#define ml_logi(format, ...)
#define ml_logw(format, ...)
#define ml_loge(format, ...)
#define ml_logd(format, ...)
```

nntrainer/nntrainer_log.h
```diff
@@ -40,42 +40,49 @@
...
-#define ml_logd(...) \
+#define ml_logd(...)                                                           \
   __android_log_print(ANDROID_LOG_DEBUG, TAG_NAME, __VA_ARGS__)
 
+#elif defined(_WIN32)
+
+#define ml_logi(format, ...)
+#define ml_logw(format, ...)
+#define ml_loge(format, ...)
+#define ml_logd(format, ...)
```
songgot, line 60: Except for this code, the code diff was generated due to line breaks and indentation alignment.

PR #3562, opened 2025-11-13
Add CUDA context support and build configuration

dkjung:
This PR adds CUDA context management files (cuda_context.h and cuda_context.cpp) that provide similar functionality to the existing OpenCL context. 

The changes include: 
- Implementation of CudaContext class inheriting from Context and Singleton 
- CUDA kernel management and execution interfaces 
- Build system updates to support CUDA with enable-cuda option 
- Conditional linking of CUDA runtime library for both Windows and Linux 
- Addition of enable-cuda option in meson_options.txt

### DIFF
--- a/nntrainer/cuda_context.cpp
+++ b/nntrainer/cuda_context.cpp
@@ -40,18 +40,11 @@ void CudaContext::initialize() noexcept {
   } catch (...) {
     ml_loge("cuda_context: registering layer failed due to unknown reason");
   }
-};
+}
 
 void CudaContext::add_default_object() {
-  // Register default layers that support CUDA
-  registerFactory(nntrainer::createLayer<FullyConnectedLayer>,
-                  FullyConnectedLayer::type, ml::train::LayerType::LAYER_FC);
-
-  registerFactory(nntrainer::createLayer<AdditionLayer>, AdditionLayer::type,
-                  ml::train::LayerType::LAYER_ADDITION);
-
-  registerFactory(nntrainer::createLayer<ReshapeLayer>, ReshapeLayer::type,
-                  ml::train::LayerType::LAYER_RESHAPE);
+  // todo: implementation is needed
+  return;
 }
 
 template <typename T>
--- a/nntrainer/cuda_context.h
+++ b/nntrainer/cuda_context.h
@@ -57,10 +57,14 @@ public:
     if (cuda_initialized) {
       // Release CUDA resources
       if (stream_) {
-        cudaStreamDestroy(stream_);
+        cudaError_t err = cudaStreamDestroy(stream_);
+        if (err != cudaSuccess) {
+          ml_loge("Failed to destroy CUDA stream: %s", cudaGetErrorString(err));
+        }
+        stream_ = nullptr;
       }
     }
-  };
+  }
 
   /**
    * @brief Factory register function, use this function to register custom
--- a/nntrainer/cl_context.h
+++ b/nntrainer/cl_context.h
@@ -88,7 +88,7 @@ public:
    * @param key key to access the factory, if key is empty, try to find key by
    * calling factory({})->getType();
    * @param int_key key to access the factory by integer, if it is -1(default),
-   * the function automatically unsigned the key and return
+   * the function automatically deassigns the key and return
    * @return const int unique integer value to access the current factory
    * @throw invalid argument when key and/or int_key is already taken
    */
@@ -109,7 +109,7 @@ public:
    * @param key key to access the factory, if key is empty, try to find key by
    * calling factory({})->getType();
    * @param int_key key to access the factory by integer, if it is -1(default),
-   * the function automatically unsigned the key and return
+   * the function automatically deassigns the key and return
    * @return const int unique integer value to access the current factory
    * @throw invalid argument when key and/or int_key is already taken
    */


nntrainer/cuda_context.cpp
```diff
@@ -0,0 +1,129 @@
...
+void CudaContext::add_default_object() {
+  // Register default layers that support CUDA
+  registerFactory(nntrainer::createLayer<FullyConnectedLayer>,
+                  FullyConnectedLayer::type, ml::train::LayerType::LAYER_FC);
+
+  registerFactory(nntrainer::createLayer<AdditionLayer>, AdditionLayer::type,
+                  ml::train::LayerType::LAYER_ADDITION);
+
+  registerFactory(nntrainer::createLayer<ReshapeLayer>, ReshapeLayer::type,
+                  ml::train::LayerType::LAYER_RESHAPE);
```
djeong20: This is a duplicate of the AppContext registration. For OpenCL, there's a `FullyConnectedLayerCl`.

nntrainer/cuda_context.h
```diff
@@ -0,0 +1,260 @@
...
+
+namespace nntrainer {
+
+extern std::mutex cuda_factory_mutex;
+
+/**
+ * @class CudaContext contains user-dependent configuration for CUDA support
+ * @brief CUDA support for app context
+ */
+class CudaContext : public Context, public Singleton<CudaContext> {
```
djeong20, line 46: I'm not sure if it’s a good idea to have a CudaContext alongside OpenCL.
Are we planning to allow OpenCL and CUDA to run at the same time?

djeong20: Maybe we could create a GPUContext as a base class and have ClContext and CudaContext inherit from it.

nntrainer/meson.build
```diff
@@ -37,6 +37,35 @@ if get_option('enable-opencl')
...
+      # Fallback to manual library specification for Linux
+      cuda_lib = cc.find_library('cudart', required: false)
+      if cuda_lib.found()
+        nntrainer_base_deps += cuda_lib
+      endif
+    else
+      nntrainer_base_deps += cuda_dep
+    endif
+  endif
+endif
```
djeong20, line 67: Shouldn’t it give an error if CUDA isn’t installed?

nntrainer/cuda_context.cpp
```diff
@@ -0,0 +1,129 @@
...
+  return assigned_int_key;
+}
+
+bool CudaContext::cudaInit() {
+  // if already initialized
+  if (cuda_initialized)
+    return true;
+
+  // Initialize CUDA context
+  cudaError_t err = cudaSetDevice(0);
```
songgot, line 98: It would be nice to be able to set the device ID when implementing additional features.
How about adding a member variable?
```
class CudaContext {
private:
  int device_id = 0;
  // ...
};
```
```
cudaError_t err = cudaSetDevice(device_id)
```

nntrainer/cuda_context.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    add_default_object();
+    setMemAllocator(std::make_shared<MemAllocator>());
+
+  } catch (std::exception &e) {
+    ml_loge("cuda_context: registering layers failed!!, reason: %s", e.what());
+  } catch (...) {
+    ml_loge("cuda_context: registering layer failed due to unknown reason");
+  }
+};
```
songgot: I think it would be good to delete the ';'.

nntrainer/cuda_context.h
```diff
@@ -0,0 +1,260 @@
...
+  /**
+   * @brief Factory register function, use this function to register custom
+   * object
+   *
+   * @tparam T object to create. Currently Layer is supported
+   * @param factory factory function that creates std::unique_ptr<T>
+   * @param key key to access the factory, if key is empty, try to find key by
+   * calling factory({})->getType();
+   * @param int_key key to access the factory by integer, if it is -1(default),
+   * the function automatically unsigned the key and return
```
songgot, line 78: Maybe typo? 'unsigned' -> 'assigns'

nntrainer/cuda_context.h
```diff
@@ -0,0 +1,260 @@
...
+  CudaContext() : Context(std::make_shared<ContextData>()) {}
+
+  /**
+   * @brief destructor to release cuda context
+   */
+  ~CudaContext() override {
+    if (cuda_initialized) {
+      // Release CUDA resources
+      if (stream_) {
+        cudaStreamDestroy(stream_);
```
songgot: How about checking the error?
```
    cudaError_t err = cudaStreamDestroy(stream_);
    if (err != cudaSuccess) {
      ml_loge("Failed to destroy CUDA stream: %s", cudaGetErrorString(err));
    }
    stream_ = nullptr;
```

nntrainer/meson.build
dkjung: It is activated only if `enable-cuda` option is on. 
It's expected to issue an error if CUDA isn't installed.

nntrainer/cuda_context.h
dkjung: This is copied from 
https://github.com/nnstreamer/nntrainer/blob/ba47d7fc37bd236f98400ddc18149118cd94e06c/nntrainer/cl_context.h#L91.

I think that `deassign` is more correct.

PR #3564, opened 2025-11-14
Make Engine owner of ThreadPoolManager

gkisalapl:
After this change Engine class is owner of ThreadPoolManager object. When we call Engine.release() function we can release ThreadPoolManager

**Self-evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -16,6 +16,7 @@
 #include <thread>
 #include <vector>
 
+#include <engine.h>
 #include <fp16.h>
 #include <layer_context.h>
 #include <mha_core.h>
@@ -480,7 +481,8 @@ void MHACoreLayer::one_batch_incremental_forwarding(
 
   /** 1. Load Input Tensors of this batch : b_ denotes a Tensor for this batch
    * **/
-  auto &pool = nntrainer::ThreadPoolManager::Global().getThreadPool();
+  auto &pool =
+    nntrainer::Engine::Global().getThreadPoolManager()->getThreadPool();
 
   nntrainer::Tensor b_cache_key_step = cache_key.getSharedDataTensor(
     cache_key_step_dim,
@@ -556,7 +558,8 @@ void MHACoreLayer::one_batch_incremental_forwarding(
 
   /** 1. Load Input Tensors of this batch : b_ denotes a Tensor for this batch
    * **/
-  auto &pool = nntrainer::ThreadPoolManager::Global().getThreadPool();
+  auto &pool =
+    nntrainer::Engine::Global().getThreadPoolManager()->getThreadPool();
 
   nntrainer::Tensor b_cache_key_step = cache_key.getSharedDataTensor(
     cache_key_step_dim,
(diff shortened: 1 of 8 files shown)


nntrainer/engine.h
```diff
@@ -133,6 +136,14 @@ class Engine : public Singleton<Engine> {
...
   }
 
+  ThreadPoolManager *getThreadPoolManager() {
+    if (!thread_pool_manager_) {
+      thread_pool_manager_ = std::make_unique<ThreadPoolManager>();
+    }
+
+    return thread_pool_manager_.get();
+  }
+
```
songgot: Are there any race conditions? If so, how about using "call_once" or a "mutex"?
```
std::once_flag init_flag;

ThreadPoolManager *getThreadPoolManager() {
    std::call_once(init_flag, [this]() {
        thread_pool_manager_ = std::make_unique<ThreadPoolManager>();
    });
    return thread_pool_manager_.get();
}
```
or
```
std::mutex mtx;

ThreadPoolManager *getThreadPoolManager() {
    std::lock_guard<std::mutex> lock(mtx);
    if (!thread_pool_manager_) {
        thread_pool_manager_ = std::make_unique<ThreadPoolManager>();
    }
    return thread_pool_manager_.get();
}
```

```

File: nntrainer/engine.h [hunk 2]
```diff
@@ -95,6 +96,8 @@ class Engine : public Singleton<Engine> {
    */
   ~Engine() = default;
 
+  void release();
```
djeong20, line 103: Please add a description here.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -16,6 +16,7 @@
 #include <thread>
 #include <vector>
 
+#include <engine.h>
```
jijoongmoon, line 19: There might be a cyclic dependency. engine --> app_context --> layer and layer include the engine.h.
how about having the thread pool ownership by network itself. 
If the engine has the pool ownership, there will be only one thread pool for multiple network model as well. so.. let's make a network have one thread pool.

nntrainer/engine.cpp
```diff
@@ -62,6 +62,8 @@ void Engine::initialize() noexcept {
   }
 };
 
+void Engine::release() { thread_pool_manager_.reset(); }
```
myungjoo, line 65: Don't you need to proect with
```
  std::lock_guard<std::mutex> lock(thread_pool_manager_mutex_);
```

nntrainer/engine.cpp
```diff
@@ -170,4 +172,14 @@ int Engine::registerContext(const std::string &library_path,
...
 
+ThreadPoolManager *Engine::getThreadPoolManager() {
+  std::lock_guard<std::mutex> lock(thread_pool_manager_mutex_);
+
+  if (!thread_pool_manager_) {
+    thread_pool_manager_ = std::make_unique<ThreadPoolManager>();
+  }
+
+  return thread_pool_manager_.get();
+}
```
myungjoo, line 183: This appears to be called by many functions. If this is called frequently, consider using more lightweight mechanisms other than mutex: e.g., atomic-op, or std::call_once

PR #3565, opened 2025-11-16
[optimizer] Fix AdamW logic and apply decoupled weight decay

minseo25:
This patch corrects the AdamW optimizer implementation to match the paper and the [PyTorch reference implementation](https://github.com/pytorch/pytorch/blob/main/torch/optim/adam.py#L416).

## Dependency of the PR

None

<details><summary>{adamw: fix implementation and add decoupled weight decay}</summary><br />

Description:
The previous AdamW implementation did not follow the AdamW paper.

- Weight decay was applied via context.calcWeightDecayGradient() rather than as decoupled weight decay.
- The optimizer state buffers (m, v) were kept in the same dtype as the weights, instead of always using FP32 as in Adam.
- There was no configurable weight_decay property for AdamW.
- As a result, it diverged on simple setups where Adam converged (e.g., MNIST).

This patch:

- Implements true decoupled weight decay at the parameter level (w ← w - lr * wd * w), following the paper and the PyTorch AdamW/decoupled-weight-decay implementation.
- Adds a weight_decay property to AdamW so that users can configure the decay factor.
- Aligns AdamW’s optimizer state behavior with Adam by keeping the first and second moment buffers (m, v) in FP32 even under mixed precision.
- Reuses the same bias-corrected learning rate computation as Adam to keep the implementations consistent.

Self evaluation:

Signed-off-by: Minseo Kim <ms05251@naver.com>

</details>

### Summary

- Changed AdamW from an L2-style, gradient-based decay to true decoupled weight decay (w ← w - lr * wd * w) consistent with the AdamW paper and PyTorch’s decoupled_weight_decay=True path.
- Added an explicit weight_decay property to AdamW.
- Updated AdamW to keep its optimizer state buffers (m, v) in FP32.
- Shared the bias-corrected learning rate logic with Adam, making AdamW consistent with the existing Adam implementation in nntrainer.

### Results

Common configuration: MNIST dataset, 200 epochs, lr = 1e-3

image

---

**Numerical summary (after 200 epochs):**
```text
[MNIST] after_l2=114.490  delta_l2=73.9796  elems=3898  (adam)
[MNIST] after_l2=114.491  delta_l2=73.9791  elems=3898  (adamw-new, no wd)
[MNIST] after_l2=113.785  delta_l2=73.6573  elems=3898  (adamw-new, wd 0.01)
```

Signed-off-by: Minseo Kim <ms05251@naver.com>

### DIFF
--- a/nntrainer/optimizers/adamw.cpp
+++ b/nntrainer/optimizers/adamw.cpp
@@ -23,13 +23,16 @@
 
 namespace nntrainer {
 
-AdamW::AdamW() : adam_props(PropsB1(), PropsB2(), PropsEpsilon(), TorchRef()) {
+AdamW::AdamW() :
+  adam_props(PropsB1(), PropsB2(), PropsEpsilon(), TorchRef(),
+             PropsWeightDecayW()) {
   /** default properties */
-  auto &[b1, b2, eps, torch_ref] = adam_props;
+  auto &[b1, b2, eps, torch_ref, weight_decay] = adam_props;
   b1.set(0.9f);
   b2.set(0.999f);
-  eps.set(1.0e-7f);
+  eps.set(1.0e-8f);
   torch_ref.set(false);
+  weight_decay.set(0.0f);
 }
 
 AdamW::~AdamW() {}
@@ -37,7 +40,15 @@ AdamW::~AdamW() {}
 enum AdamParams { wm, wv };
 
 std::vector<TensorDim> AdamW::getOptimizerVariableDim(const TensorDim &dim) {
-  return {dim, dim};
+  /**
+   * @note We assume the optimizer parameters should be full precision to
+   * maintain the accuracy even in mixed precision training.
+   */
+  TensorDim wm_dim(dim);
+  TensorDim wv_dim(dim);
+  wm_dim.setDataType(ml::train::TensorDim::DataType::FP32);
+  wv_dim.setDataType(ml::train::TensorDim::DataType::FP32);
+  return {wm_dim, wv_dim};
 }
 
 void AdamW::exportTo(Exporter &exporter,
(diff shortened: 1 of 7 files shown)


songgot:
:+1:

haehun (approved):
Great work!

nntrainer/optimizers/adamw.cpp
```diff
@@ -12,32 +12,42 @@
  * @brief  This is the AdamW Optimizer.
  */
 
+#include <adamw.h>
```
djeong20: It’s not a big deal, but please revert this. The separation is intentional: first the standard C++ headers, then the external library includes, and finally this project’s headers.

nntrainer/optimizers/adamw.cpp
```diff
@@ -51,6 +61,20 @@ void AdamW::setProperty(const std::vector<std::string> &values) {
...
+static double getUpdatedLearningRate(
+  unsigned int iteration, double ll,
+  const std::tuple<PropsB1, PropsB2, PropsEpsilon, TorchRef, PropsWeightDecayW>
+    &adam_props) {
+  auto &beta1 = std::get<PropsB1>(adam_props).get();
+  auto &beta2 = std::get<PropsB2>(adam_props).get();
+  auto biasCorrection = [&](double f) { return 1.0 - pow(f, iteration + 1); };
+  ll *= sqrt(biasCorrection(beta2)) / biasCorrection(beta1);
+  return ll;
+}
```
djeong20: Is there a specific reason this function is declared static?

nntrainer/optimizers/adamw.h
```diff
@@ -15,15 +15,22 @@
 #define __ADAMW_H__
 #ifdef __cplusplus
 
-#include <tuple>
-
 #include <adam.h>
-
 #include <base_properties.h>
 #include <optimizer_devel.h>
+#include <tuple>
```
djeong20: same here

nntrainer/optimizers/adamw.cpp
```diff
@@ -51,6 +61,20 @@ void AdamW::setProperty(const std::vector<std::string> &values) {
   Optimizer::setProperty(left);
 }
 
+/**
+ * @brief bias-corrected learning rate (Adam)
+ */
+static double getUpdatedLearningRate(
+  unsigned int iteration, double ll,
```
songgot: Are there any other good names instead of ```double ll``` ?
'learning_rate' or 'lr'?

nntrainer/optimizers/adamw.cpp
minseo25: Reverted the change, thanks!

nntrainer/optimizers/adamw.h
minseo25: Also reverted this change

nntrainer/optimizers/adamw.cpp
minseo25: I initially followed Adam’s naming convention, but I’ve renamed both from ll to lr for clarity. Thanks!

nntrainer/optimizers/adamw.cpp
minseo25: No particular reason - I changed it to private to match adam.h. Thanks!

PR #3567, opened 2025-11-18
Add CUDA context support and build configuration

dkjung:
This PR adds CUDA context management files (cuda_context.h and cuda_context.cpp) that provide similar functionality to the existing OpenCL context. The changes include:

- Implementation of CudaContext class inheriting from Context and Singleton
- CUDA kernel management and execution interfaces
- Build system updates to support CUDA with enable-cuda option
- Conditional linking of CUDA runtime library for both Windows and Linux
- Addition of enable-cuda option in meson_options.txt
- Implementation of RMSNorm CUDA kernel and build configuration

### DIFF
--- a/nntrainer/engine.cpp
+++ b/nntrainer/engine.cpp
@@ -50,6 +50,12 @@ void Engine::add_default_object() {
 
   registerContext("gpu", &cl_context);
 #endif
+
+#ifdef ENABLE_CUDA
+  auto &cuda_context = nntrainer::CudaContext::Global();
+
+  registerContext("cuda", &cuda_context);
+#endif
 }
 
 void Engine::initialize() noexcept {
--- a/nntrainer/tensor/cuda_operations/meson.build
+++ b/nntrainer/tensor/cuda_operations/meson.build
@@ -1,0 +1,34 @@
+# Find CUDA compiler
+dep = dependency('cuda', version : '>=13', modules : ['cublas'])
+
+nvcc = find_program('nvcc', required: true)
+
+if nvcc.found()
+  cuda_sources = [
+    'rmsnorm_cuda.cu'
+  ]
+
+  cuda_headers = [
+    'rmsnorm_cuda.h'
+  ]
+
+  kernel_objects = []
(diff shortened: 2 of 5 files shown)


EunjuYang:
I think the commit body of bc7cec08321db753cf994e722d432942d164e6fb is missing. Please add some messages to the body instead of putting all messages in the title.

nntrainer/tensor/cuda_operations/meson.build
```diff
@@ -0,0 +1,34 @@
...
+
+  cuda_headers = [
+    'rmsnorm_cuda.h'
+  ]
+
+  kernel_objects = []
+  foreach kernel : cuda_sources
+    obj_name = kernel.replace('.cu', '.o')
+    obj = custom_target(obj_name,
+      command: [nvcc, '-c', '-Xcompiler', '/MD', '@INPUT@', '-o', '@OUTPUT@'],
```
myungjoo, line 19: /MD is for MSVC++. This expression should check if it's for Windows/MSVCC++.
For linux, add appropriate compiler option (-fPIC ? )

nntrainer/engine.cpp
```diff
@@ -50,6 +50,12 @@ void Engine::add_default_object() {
 
   registerContext("gpu", &cl_context);
 #endif
+
+#ifdef ENABLE_CUDA
+  auto &cuda_context = nntrainer::CudaContext::Global();
```
myungjoo, line 55: This may need to add ```#include "cuda_context.h"```.

nntrainer/cuda_context.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <nntrainer_log.h>
+#include <reshape_layer.h>
+
+#include <cuda.h>
+#include <cuda_runtime.h>
+
+namespace nntrainer {
+std::mutex cuda_factory_mutex;
+
+void CudaContext::initialize() noexcept {
```
myungjoo, line 28: Is it ok to let the caller of initialize() unaware of the init error?

PR #3569, opened 2025-11-18
INT4 GEMV OpenCL Kernel for Adreno GPU Compatibility

djeong20:
This PR updates the INT4 GEMV OpenCL kernel to ensure compatibility with the Adreno GPU while maintaining the Intel intrinsics.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/cl_operations/cl_kernels/int4_gemv.cl
+++ b/nntrainer/tensor/cl_operations/cl_kernels/int4_gemv.cl
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Modifications made by Donghyeon Jeong on September 13 2025:
 // - Limit its functionality exclusively to OS_IS_YX_OSV32_ISV2
+// - Portability updates (Adreno-compatible) while preserving Intel intrinsics:
 
 #if defined(cl_khr_fp16)
 #pragma OPENCL EXTENSION cl_khr_fp16 : enable
@@ -15,7 +16,7 @@
 #define CAT(x, y) __CAT(x, y)
 
 #define unroll_for __attribute__((opencl_unroll_hint)) for
-#define CEIL_DIV(a, b) (((a) + (b)-1) / (b))
+#define CEIL_DIV(a, b) (((a) + (b) - 1) / (b))
 #define ALIGN(a, b) (CEIL_DIV(a, b) * (b))
 #define MIN(a, b) ((a) < (b) ? (a) : (b))
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/cl_operations/cl_kernels/int4_gemv.cl
```diff
@@ -102,6 +116,54 @@
...
+// ==========================================================================
+// Block-read emulation (when Intel block-read intrinsics aren't present).
+// ==========================================================================
+#define BLOCK_READ_IMPL_1 ret = ptr[idx];
+
+#define BLOCK_READ_IMPL_2                                                      \
+  ret.s0 = ptr[idx];                                                           \
+  idx += get_max_sub_group_size();                                             \
+  ret.s1 = ptr[idx];                                                           \
+  idx += get_max_sub_group_size();
```
myungjoo, line 128: Be careful with such #define.

Someone may do...:
```
if (a == b)
  BLOCK_READ_IMPL_2;
else
  BLOCK_READ_IMPL_4;
```
which will break everything.

https://stackoverflow.com/questions/923822/whats-the-use-of-do-while0-when-we-define-a-macro

nntrainer/tensor/cl_operations/cl_kernels/int4_gemv.cl
```diff
@@ -160,6 +235,59 @@ DECLARE_BLOCK_READ_EMULATION(1, 16)
...
+// ==========================================================================
+#if !defined(cl_intel_subgroups)
+
+inline float sg_reduce_add_float(float v, __local float *buf_line) {
+  uint lid = get_sub_group_local_id(); // lane in X-dimension
+  buf_line[lid] = v;
+  barrier(CLK_LOCAL_MEM_FENCE);
+
+  uint sg_size = SUBGROUP_SIZE; // expected 16
+  for (uint stride = sg_size >> 1; stride > 0; stride >>= 1) {
```
piotrrak, line 252: I don't get it:

shouldn't
```
if (get_subgroup_local_id() != 0)
{
   for (uint i = 1; i <SUBGROUP_SIZE; ++i)
   {
       buf_line[0] = buf_line[0] + buf_line[i];
   }
}
```
as we would want to reduce all no just some.
Simply don't get it what's going on with this stride.

piotrrak: Oh think I get it, to make it work the way you want the reduction accumulation local buf_line[0] + to strided element should be atomic (local atomit fetch add) otherwise it races between subgroup threads. (race between is on buf_line[0] and modifiaction of buf_line[0])

Have you tested correctness of this reduction?

piotrrak:
According to this document qualcomm gpus support subgroup operations, as both `cl_khr_subgroup_*` as part of OpenCL 3.0 and `cl_qcom_subgroups_*`
Only part that should be reimplemented is `cl_intel_subgroup_local_block_io`

https://docs.qualcomm.com/bundle/publicresource/80-NB295-11_REV_C_Qualcomm_Snapdragon_Mobile_Platform_Opencl_General_Programming_and_Optimization.pdf

Please refer to chapter 8.9 and 9.2 in above document. I don't see the point in reimplementing those, as they are supported by non-intel  extensions. Reusing those should simplify debugging code.

In any case `subgroup_reduce_add` could be implemented as shuffle_down operation correctly.
image

PR #3338, opened 2025-07-18
[onnx] onnx interpreter for llama

baek2sm:
This PR contains multiple commits that get the LLM model to a compile‑ready state, including:
- Adding support for loading properties from ONNX operators.
- Adding support for loading properties from ONNX tensors.
- Implementing required layers such as Cast, Slice, and others.
- Adding documentation for ONNX.

The execution result of the example app is as follows (llama 1b):
```
================================================================================
          Layer name          Layer type    Output dimension         Input layer
================================================================================
         onnx__add_6               input             1:1:1:1
--------------------------------------------------------------------------------
 onnx__add_6/generat            multiout             1:1:1:1         onnx__add_6
--------------------------------------------------------------------------------
                 sin               input            1:1:1:64
--------------------------------------------------------------------------------
 sin/generated_out_0            multiout            1:1:1:64                 sin
--------------------------------------------------------------------------------

...

--------------------------------------------------------------------------------
   model_norm_cast_1                cast          1:1:1:2048      model_norm_mul
--------------------------------------------------------------------------------
    model_norm_mul_1            multiply          1:1:1:2048   model_norm_weight
                                                              model_norm_cast_1
--------------------------------------------------------------------------------
      lm_head_matmul              matmul         1:1:1:50304    model_norm_mul_1
                                                               onnx__matmul_3531
================================================================================
```

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [*]Passed [ ]Failed [ ]Skipped

Signed-off-by: Seungbaek Hong <sb92.hong@samsung.com>

PR #3570, opened 2025-11-19
[layer] fix duplicate indices issues for layer enum

baek2sm:
The current layer enum indices are managed separately in
`layer.h` and `nntrainer-api-common.h`.

For layers that are distributed through the ml-api,
`layer.h` maps to the enums defined in `nntrainer-api-common.h`.

But as the number of layers grew, I discovered that some unmapped layers
were receiving duplicate index values, leading to unintended behavior.

** Fix **
- Assign indices starting from 1000 to any layer that does not have a
  mapping, ensuring that unmapped layers never clash with existing
indices.

### DIFF
--- a/api/ccapi/include/layer.h
+++ b/api/ccapi/include/layer.h
@@ -111,20 +111,27 @@ enum LayerType {
     ML_TRAIN_LAYER_TYPE_LOSS_CROSS_ENTROPY_SOFTMAX, /**< Cross Entropy with
                                                        Softmax Loss Layer type
                                                      */
-  LAYER_TIME_DIST,       /**< Time Distributed Layer type */
-  LAYER_BACKBONE_TFLITE, /**< Backbone using TFLite */
-  LAYER_RESHAPE,         /**< Reshape Layer type */
-  LAYER_REDUCE_MEAN,     /**< Reduce mean Layer type */
+  LAYER_RMSNORM = ML_TRAIN_LAYER_TYPE_RMSNORM,      /**<RMS NORM Layer */
+  LAYER_TRANSPOSE = ML_TRAIN_LAYER_TYPE_TRANSPOSE,  /**< Transpose Layer type */
+  LAYER_CHANNEL_SHUFFLE =
+    ML_TRAIN_LAYER_TYPE_CHANNEL_SHUFFLE, /**< Channel Shuffle Layer type */
   LAYER_REDUCE_SUM =
     ML_TRAIN_LAYER_TYPE_REDUCE_SUM, /**< Reduce sum Layer type */
-  LAYER_LOSS_CONSTANT_DERIVATIVE,   /**< Synthetic loss layer to feed constant
-                                       derivative */
-  LAYER_UPSAMPLE2D,                 /**< Upsample 2D Layer type */
-  LAYER_RMSNORM = ML_TRAIN_LAYER_TYPE_RMSNORM,     /**<RMS NORM Layer */
-  LAYER_TRANSPOSE = ML_TRAIN_LAYER_TYPE_TRANSPOSE, /**< Transpose Layer type */
-  LAYER_CHANNEL_SHUFFLE =
-    ML_TRAIN_LAYER_TYPE_CHANNEL_SHUFFLE,      /**< Channel Shuffle Layer type */
-  LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN /**< Unknown */
+  LAYER_REDUCE_MEAN =
+    ML_TRAIN_LAYER_TYPE_REDUCE_MEAN,           /**< Reduce mean Layer type */
+  LAYER_RESHAPE = ML_TRAIN_LAYER_TYPE_RESHAPE, /**< Reshape Layer type */
+  LAYER_UPSAMPLE2D =
+    ML_TRAIN_LAYER_TYPE_UPSAMPLE2D, /**< Upsample 2D Layer type */
+
+  LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN, /**< Unknown */
+
+  LAYER_TIME_DIST =
+    ML_TRAIN_LAYER_EXPERIMENTAL, /**< Time Distributed Layer type */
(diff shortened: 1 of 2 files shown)


api/ccapi/include/layer.h
```diff
@@ -111,19 +111,22 @@ enum LayerType {
...
-  LAYER_TRANSPOSE = ML_TRAIN_LAYER_TYPE_TRANSPOSE, /**< Transpose Layer type */
-  LAYER_CHANNEL_SHUFFLE =
-    ML_TRAIN_LAYER_TYPE_CHANNEL_SHUFFLE,      /**< Channel Shuffle Layer type */
+
+  LAYER_TIME_DIST = 1000,       /**< Time Distributed Layer type */
+  LAYER_BACKBONE_TFLITE = 1001, /**< Backbone using TFLite */
+  LAYER_RESHAPE = 1002,         /**< Reshape Layer type */
+  LAYER_REDUCE_MEAN = 1003,     /**< Reduce mean Layer type */
+  LAYER_UPSAMPLE2D = 1004,      /**< Upsample 2D Layer type */
+  LAYER_LOSS_CONSTANT_DERIVATIVE = 1005, /**< Synthetic loss layer to feed
```
myungjoo: Is there any reason not to define ML_TRAIN_LAYER_* at nntrainer-api-common.h like others?

api/ccapi/include/layer.h
```diff
@@ -111,19 +111,22 @@ enum LayerType {
...
   LAYER_REDUCE_SUM =
     ML_TRAIN_LAYER_TYPE_REDUCE_SUM, /**< Reduce sum Layer type */
-  LAYER_LOSS_CONSTANT_DERIVATIVE,   /**< Synthetic loss layer to feed constant
-                                       derivative */
-  LAYER_UPSAMPLE2D,                 /**< Upsample 2D Layer type */
-  LAYER_RMSNORM = ML_TRAIN_LAYER_TYPE_RMSNORM,     /**<RMS NORM Layer */
-  LAYER_TRANSPOSE = ML_TRAIN_LAYER_TYPE_TRANSPOSE, /**< Transpose Layer type */
-  LAYER_CHANNEL_SHUFFLE =
-    ML_TRAIN_LAYER_TYPE_CHANNEL_SHUFFLE,      /**< Channel Shuffle Layer type */
+
```
songgot, line 125: It seems like we can use the ```LAYER_UNKNOWN``` value as a valid boundary value in an if statement.
bool
valid_layer (int layer) {
    if (layer < LAYER_UNKNOWN)
        return true;
    return false;
}
So, how about moving LAYER_UNKNOWN above LAYER_TIME_DIST?
```
  LAYER_UNKNOWN = ML_TRAIN_LAYER_TYPE_UNKNOWN, /**< Unknown Layer (999) - Boundary value for API layer validation */

  // ===== Internal Layers (1000+) - Safe range to avoid conflicts =====
  LAYER_TIME_DIST = 1000,       /**< Time Distributed Layer type */
  LAYER_BACKBONE_TFLITE = 1001, /**< Backbone using TFLite */
  LAYER_RESHAPE = 1002,         /**< Reshape Layer type */
  LAYER_REDUCE_MEAN = 1003,     /**< Reduce mean Layer type */
  LAYER_UPSAMPLE2D = 1004,      /**< Upsample 2D Layer type */
  LAYER_LOSS_CONSTANT_DERIVATIVE = 1005 /**< Synthetic loss layer to feed constant derivative */
};
```

api/ccapi/include/layer.h
baek2sm: @myungjoo I'll add `reduce_mean`, `upsample2d` and `reshape` layer to ML_TRAIN_LAYER_*, however, since the other layers don’t seem to be commonly used types, so I’m not sure whether they should be added to ML_TRAIN_LAYER.

api/ccapi/include/layer.h
baek2sm: This commit is just for preventing duplicate indices. I'm not sure what is the beneficial if we make `valid_layer` function.

api/nntrainer-api-common.h
```diff
@@ -83,6 +83,10 @@ typedef enum {
     45,                         /**< Channel Shuffle Layer type (Since 9.0)*/
   ML_TRAIN_LAYER_TYPE_NEG = 46, /**< Neg Layer type (Since 9.0)*/
   ML_TRAIN_LAYER_TYPE_REDUCE_SUM = 48, /**< ReduceSum Layer type (Since 9.0) */
+  ML_TRAIN_LAYER_TYPE_REDUCE_MEAN =
```
baek2sm, line 88: the reason why the index is start from 51 is that PR #3338 already uses indices to 50

api/ccapi/include/layer.h
songgot: I was thinking of C API. :), Thank you.

api/ccapi/include/layer.h
myungjoo: You may add a place holder macro for them, too:

E.g.,
```C
#define ML_TRAIN_LAYER_EXPERIMENTAL (10000)

...
LAYER_RESHAPE = ML_TRAIN_LAYER_EXPERIMENTAL + 1,
LAYER_REDUCE_MEAN = ML_TRAIN_LAYER_EXPERIMENTAL + 2,
...
```

baek2sm: Good idea. I added it. thanks!

api/ccapi/include/layer.h
```diff
@@ -17,6 +17,8 @@
 #ifndef __ML_TRAIN_LAYER_H__
 #define __ML_TRAIN_LAYER_H__
 
+#define ML_TRAIN_LAYER_EXPERIMENTAL (10000)
```
myungjoo: Move to api/nntrainer-api-common.h alont with other ML_TRAIN_LAYER_*.

songgot (approved):
LGTM :+1:

skykongkong8 (approved):

jijoongmoon (approved):
LGTM

PR #3576, opened 2025-11-20
Add ARM backend implementations for non-FP16 case

djeong20:
This PR implements functions in the ARM backend for when enable-fp16 is set to false.
Previously, compute_rotary_emb_value, compute_fp16vcache_fp32_transposed, and compute_kcaches would throw errors when used without enable-fp16 due to missing definitions.
The patch adds implementations for these functions specifically for the non-FP16 scenario, ensuring compatibility across configurations.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
@@ -1302,4 +1302,221 @@ void clamp(const float *input, float *output, size_t length, float lower_bound,
   }
 }
 
+#if defined(__aarch64__) || defined(_M_ARM64)
+static inline void load_fp16_4_to_chunk(const uint16_t *src, float *dst,
+                                        int chunk_size) {
+  int i = 0;
+  for (; i + 4 <= chunk_size; i += 4) {
+    uint16x4_t u16_vec = vld1_u16(src + i);
+    float16x4_t half = vreinterpret_f16_u16(u16_vec);
+    float32x4_t f32 = vcvt_f32_f16(half);
+    vst1q_f32(dst + i, f32);
+  }
+  for (; i < chunk_size; ++i) {
+    dst[i] = nntrainer::compute_fp16_to_fp32(src[i]);
+  }
+}
+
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -1302,4 +1302,219 @@ void clamp(const float *input, float *output, size_t length, float lower_bound,
...
+    float16x4_t half = vreinterpret_f16_u16(u16_vec);
+    float32x4_t f32 = vcvt_f32_f16(half);
+    vst1q_f32(dst + i, f32);
+  }
+  for (; i < chunk_size; ++i) {
+    dst[i] = nntrainer::compute_fp16_to_fp32(src[i]);
+  }
+}
+
+void compute_kcaches_uint16(const float *in, const uint16_t *kcache,
```
mnalewaj, line 1320: As I can see the method is a copy of:
template <>
void compute_kcaches(const float *in, const __fp16 *kcache, float *output,
                     int num_rows, int num_cache_head, int head_dim,
                     int gqa_size, int tile_size, size_t local_window_size) {
					 ...
					 }
From neon_impl_fp16.cpp
Only type is changed from __fp16 to uint16_t and other small corrections due to compilation errors.
Other methods also are copied from this file.
I would like to sure it is intended and this such solution is necessary due to compilation errors etc...

djeong20: Thank you for the review. I added this to ensure these kernels remain available when `enable-fp16` is false. If these kernels are not inherently tied to FP16, the `neon_impl_fp16` implementation could be removed. However, due to my limited familiarity with the kernel details and potential performance implications of minor changes, I chose to retain a copy for safety.

jijoongmoon: I agree with Mnalewaj. No need to be this PR, but we have to modify the code as mnalewaj said.

jijoongmoon (approved):
LGTM

PR #3582, opened 2025-11-25
std::tread and joint to remove threads to load 

jijoongmoon:
## Dependency of the PR 

This PR fixes load function using std::thread rather than futures.
For windows, the threads created by futures are not removed and the
total number of threads are increase.
We fixes this by using std::thread and join.

signed-off-by jijoong.moon<jijoong.moon@samsung.com>

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -737,13 +737,15 @@ void NeuralNetwork::load(const std::string &file_path,
         NNTR_THROW_IF((model_file_fd == -1), std::invalid_argument)
           << "Cannot open file : " << f_path;
       }
-      std::vector<std::future<void>> futures;
+      // std::vector<std::future<void>> futures;
+      std::vector<std::thread> threads;
+      threads.reserve(model_graph.size());
       for (auto iter = model_graph.cbegin(); iter != model_graph.cend();
            ++iter) {
         auto node = *iter;
         auto exec_order = std::get<0>((*iter)->getExecutionOrder());
 
-        futures.emplace_back(std::async(std::launch::async, [&, node] {
+        threads.emplace_back([&, node]() {
           if (!MMAP_READ) {
             auto local_model_file = checkedOpenStream<std::ifstream>(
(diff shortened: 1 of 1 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -777,39 +779,41 @@ void NeuralNetwork::load(const std::string &file_path,
...
+            ::munmap(mmap_ptr, f_size);
 #endif
           }
-        }));
+        });
       }
-
-      for (auto &f : futures)
-        f.get();
+      for (auto &t : threads)
```
DonghakPark:
```.cpp
    for (auto& t : threads) {
        if (t.joinable()) {
            t.join();
        }
    }
```
For avoid std::system_error, how about add check ```joinable()``` before join

DonghakPark (approved):
LGTM!

nntrainer/models/neuralnet.cpp
jijoongmoon: Thanks. fixed.

nntrainer/models/neuralnet.cpp
```diff
@@ -737,21 +737,23 @@ void NeuralNetwork::load(const std::string &file_path,
...
+        threads.emplace_back([&, node]() {
           if (!MMAP_READ) {
             auto local_model_file = checkedOpenStream<std::ifstream>(
               (v.size() == 2) ? v[1] : v[0], std::ios::in | std::ios::binary);
             node->read(local_model_file, false, exec_mode, fsu_mode,
                        std::numeric_limits<size_t>::max(), true, model_file_fd);
           } else {
 #if defined(_WIN32)
-            // Map per-task, then unmap immediately after: enables early release
+            // Map per-ask, then unmap immediately after: enables early release
```
songgot, line 756: Maybe typo? 'per-ask'  -> 'per-task'

songgot, line 756: Maybe typo? 'per-ask'  -> 'per-task'

PR #3589, opened 2025-11-27
Fixing tests

Anna-Szal:

<details><summary>l2norm test fix</summary><br />

**test**: blas_kernels --> l2norm
**problem**: the output was compared against CPU version of the algorithm which turned out to be less precise
**solution**: using fp64 version of the CPU algorithm as ground truth

**Self evaluation:**

</details>

<details><summary>rotary_emb test fix</summary><br />

**test**: attention_kernels --> rotary_emb
**problem**: InBufferC is used as third input buffer which 
- has "unused_buffer_bytes" size = 1 float
- is set as "read-only" but is being altered inside the kernel
**solution**: switch to outBufferB which is nor read-only and change its size to "buffer_size_bytes"

**Self evaluation:**

</details>

<details><summary>concatGPU test fix</summary><br />

**test**: layers--> ConcatGPU/LayerGoldenTest
**problem**: improper passing of buffers as arguments to SetKernelArguments()
**solution**: retrieve and store buffer pointer in a variable, then pass its address as the argument

**Self evaluation:**

</details>

### Summary

- bug fixes for several tests on GPU when built with OpenCL

### DIFF
--- a/test/unittest/unittest_blas_kernels_cl.cpp
+++ b/test/unittest/unittest_blas_kernels_cl.cpp
@@ -28,8 +28,6 @@
 #include <fp16.h>
 #include <layer_context.h>
 #include <tensor.h>
-#include <cblas.h>
-#include <cblas_interface.h>
 
 
 #define EXPECT_IN_RANGE(VAL, MIN, MAX)                                         \
@@ -1725,7 +1723,7 @@ TEST(blas_kernels, l2norm) {
                                   % MOD) *alpha);
   
   float gpu_result = nrm2Cl(A_fp32);
-  float cpu_result = cblas_dnrm2(size, B_fp32, 1);
+  float cpu_result = dnrm2(size, B_fp32, 1);
   delete[] B_fp32;
   
   EXPECT_FLOAT_EQ(gpu_result, cpu_result);
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -295,6 +295,14 @@ float snrm2(const unsigned int N, const float *X, const unsigned int incX) {
 #endif
 }
 
+double dnrm2(const unsigned int N, const double *X, const unsigned int incX) {
+#ifdef USE_BLAS
+  return __cblas_dnrm2(N, X, incX);
+#else
+  return __fallback_dnrm2(N, X, incX);
+#endif
+}
+
 void sgemm(const unsigned int TStorageOrder, bool TransA, bool TransB,
            const unsigned int M, const unsigned int N, const unsigned int K,
            const float alpha, const float *A, const unsigned int lda,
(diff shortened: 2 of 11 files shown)


nntrainer/layers/cl_layers/concat_cl.cpp
```diff
@@ -275,21 +275,22 @@ void ConcatLayerCl::concat_cl_axis3(const float *matAdata,
...
       break;
     }
+  
+    auto bufferInA = clbuffInstance.getInBufferA()->GetBuffer();
+    auto bufferInB = clbuffInstance.getInBufferB()->GetBuffer();
+    auto bufferOutA = clbuffInstance.getOutBufferA()->GetBuffer();
 
-    result = kernel_concat_ptr->SetKernelArguments(
-      0, clbuffInstance.getInBufferA()->GetBuffer(), sizeof(cl_mem));
+    result = kernel_concat_ptr->SetKernelArguments(0, &bufferInA, sizeof(cl_mem));
```
djeong20: Would this also work in the same way?
```suggestion
    result = kernel_concat_ptr->SetKernelArguments(0, &clbuffInstance.getInBufferA()->GetBuffer(), sizeof(cl_mem));
```

nntrainer/cl_buffer_manager.cpp
```diff
@@ -24,7 +24,7 @@ void ClBufferManager::initBuffers() {
   inBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   inBufferC = new opencl::Buffer(context_inst_, unused_buffer_bytes, true);
```
djeong20: Let's also revert inBufferC since it is used for the attention_kernel, which is also one of the failing tests
```suggestion
  inBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
```

test/unittest/unittest_blas_kernels_cl.cpp
```diff
@@ -28,6 +28,9 @@
 #include <fp16.h>
 #include <layer_context.h>
 #include <tensor.h>
+#include <cblas.h>
+#include <cblas_interface.h>
```
djeong20: I would not recommend directly including headers related to CBLAS. The reason is that there is an option called `enable-blas`, which only includes OpenBLAS when it is set to true. If you include the headers directly and `enable-blas` is set to false, it will result in an error. I personally suggest modifying the code accordingly.

https://github.com/nnstreamer/nntrainer/blob/b332f3f3f36f62b962db2027b779e37775f72ba1/nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp#L233-L239

nntrainer/layers/cl_layers/concat_cl.cpp
Anna-Szal: yes, this works. I will update all such instances accordingly

Anna-Szal:
thank you for your guidance! 
i will split this PR into separate 1-commit PRs and sign-off all my commits

PR #3594, opened 2025-12-01
[Optimizer] Optimizer Enhancements: Add Lion and Fix Core Gradient Logic

PJH6029:
## NNTrainer Optimizer Enhancements: Add Lion and Fix Core Gradient Logic

This PR introduces the **Lion optimizer**, adds a new MNIST-based application for testing and comparing optimizers, and fixes a critical latent bug in `Weight::applyGradient` that could cause future optimizers to fail.

## Dependency of the PR

None

<details><summary><b>Add Lion optimizer</b></summary><br />

This commit adds the Lion optimizer, based on the paper ["Symbolic Discovery of Optimization Algorithms"](https://arxiv.org/pdf/2302.06675).

Algorithm of Lion

This commit includes:
-   **Implementation:** `nntrainer/optimizers/lion.h` and `nntrainer/optimizers/lion.cpp` contain the core algorithm logic. The implementation follows the in-place modification pattern used by other optimizers like AdamW to ensure compatibility.
-   **Integration:** The optimizer has been wired into the build system and factory components, making it accessible via model configuration files.

**Self evaluation:**

Signed-off-by: Jeonghun Park <top231902@naver.com>

</details>

<details><summary><b>Add applications for testing optimizers</b></summary><br />

This commit adds a new application (`Optimizers`) designed to simplify the testing, validation, and comparison of different optimizers.

Key changes include:
-   **New Application:** `Applications/Optimizers` provides a standard environment for running a simple model on the MNIST or random dataset.
-   **Configuration:** Includes `.ini` files that are pre-configured to run with various optimizers (SGD, Adam, AdamW, and the new Lion), allowing for easy performance comparison.

The following figure validates the overall performance of the implemented Lion optimizer against existing optimizers, using the corrected `weight.cpp` logic.

---

**Figure 1: Lion Optimizer Performance Validation** - This chart displays the training loss, validation loss, and validation accuracy for SGD, Adam, AdamW, and the newly implemented Lion optimizer on the MNIST dataset. It serves to validate the proper functioning and competitive performance of the Lion optimizer.

Lion/Adam/AdamW/SGD performance

---

**Self evaluation:**

Signed-off-by: Jeonghun Park <top231902@naver.com>

</details>

<details><summary><b>Fix a logic of Weight::applyGradient in nntrainer/tensor/weight.cpp</b></summary><br />

This commit fixes a critical logic flaw in `Weight::applyGradient(double lr, Tensor &updated_grad)`.

### The Problem: A Latent Bug

The previous implementation had a bug in its non-mixed precision (FP32) path. It would fall back to calling `applyGradient(lr)`, which uses the weight's internal `grad` tensor and completely **ignores the `updated_grad` tensor** passed by the optimizer. This violates the function's contract, as the optimizer's calculated gradient is effectively discarded.

### Why Existing Optimizers Weren't Affected

Current optimizers like `AdamW` happen to work correctly due to a fortunate side-effect. They get a **reference** to the weight's internal gradient using `context.getGradient()` and modify it **in-place**. When the flawed `applyGradient` was called, it accidentally used this already-modified internal gradient, masking the bug.

FYI, the suggested `Lion` implementation also uses this trick to mitigate the bug. It uses the gradient''s reference to modify that in-place.

### Why The Fix Is Necessary

This is a potential bug. A future optimizer written in a different (but perfectly valid) style—for example, one that calculates the final update in a separate, temporary tensor—would fail. Such an optimizer would appear to behave like simple SGD because its carefully calculated update would be silently ignored.

### The Solution

This commit changes the FP32 path to explicitly use the gradient provided by the optimizer: `var->add_i(updated_grad, -lr);`. This makes the framework robust, logically correct, and future-proof, ensuring that any correctly implemented optimizer will work as expected, regardless of its internal implementation strategy.

The following figure specifically demonstrates the impact of this bug and the effectiveness of the fix.

---

**Figure 2: Demonstrating the Latent Bug and the Fix** - This figure specifically highlights the bug in the *previous* `weight.cpp` implementation. It compares the performance of a 'Non-In-Place' Lion optimizer (designed to be vulnerable to the bug by calculating updates in a separate tensor) against a standard SGD baseline. Two scenarios are shown: the 'Non-In-Place' Lion with the *buggy* `weight.cpp` (which should degrade to SGD-like behavior) and the same 'Non-In-Place' Lion with the *fixed* `weight.cpp` (which should exhibit correct Lion convergence). This clearly illustrates the impact of the latent bug and confirms the efficacy of the fix.

Non-In-Place Lion with buggy vs. fixed weight.cpp, and SGD baseline

> Note that the Lion implementation from commit 1 and 2 is different from the one used in the Figure 2. The committed one uses in-place operations.
---

**Self evaluation:**

Signed-off-by: Jeonghun Park <top231902@naver.com>

</details>

### Summary

- Adds the Lion optimizer, a new algorithm known for its memory efficiency and strong performance.
- Introduces a dedicated MNIST application to streamline the testing and comparison of different optimization algorithms.
- Corrects a latent bug in the core `Weight::applyGradient` method to make the optimizer framework robust and ensure future optimizers will work as intended, regardless of their implementation details.

Signed-off-by: Jeonghun Park <top231902@naver.com>

### DIFF
--- a/nntrainer/tensor/weight.cpp
+++ b/nntrainer/tensor/weight.cpp
@@ -137,7 +137,8 @@ void Weight::applyGradient(double lr, Tensor &updated_grad) {
     quantizeWeight();
     return;
   } else {
-    return applyGradient(lr);
+    /** FP32 (or matching dtype) path: apply the provided updated_grad directly */
+    var->add_i(updated_grad, -lr);
   }
 }
 


Applications/Optimizers/jni/main.cpp
```diff
@@ -0,0 +1,207 @@
...
+std::unique_ptr<ml::train::Model> create_model() {
+  std::vector<std::string> model_props = {"loss=mse",
+                                          "model_tensor_type=FP32-FP32"};
+  std::unique_ptr<ml::train::Model> model =
+    ml::train::createModel(ml::train::ModelType::NEURAL_NET, model_props);
+
+  model->addLayer(ml::train::createLayer("input", {"input_shape=1:1:10"}));
+
+  model->addLayer(ml::train::createLayer("fully_connected",
+                                         {"unit=5", "activation=softmax"}));
```
myungjoo, line 153: The output dimention is 5? Is it consistent with dataloader?

Applications/Optimizers/jni/main.cpp
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Copyright (C) 2024 Jeonghun Park <top231902@naver.com>
+ *
+ * @file   main.cpp
+ * @date   1 December 2025
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Jeonghun Park <top2313902@naver.com>
+ * @author Minseo Kim <ms05251@naver.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  This is the Lion Optimizer Header.
```
songgot: How about using ``` @brief  This is the main file for Optimizers MNIST application.``` ?

Applications/Optimizers/jni/main.cpp
```diff
@@ -0,0 +1,207 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Jeonghun Park <top231902@naver.com>
```
songgot, line 3: Which is correct, 'top231902' or 'top2313902' ?
These are being used in multiple files.

Applications/Optimizers/jni/main.cpp
PJH6029: Thanks for your clarification. I’ve updated the dataloader so that the output dimension is now 5 for consistency.
```cpp
      {{static_cast<unsigned int>(batch_size), 1u, 1u, 5u}},
      static_cast<unsigned int>(number_of_db)));
```

Applications/Optimizers/jni/main.cpp
PJH6029: Thanks. I’ve updated the comment in the `Optimizers/jni/main.cpp`, `Optimizers_MNIST/jni/main.cpp` and `lion.cpp` accordingly.

Applications/Optimizers/jni/main.cpp
PJH6029: top231902 is correct. I've also updated this too.

Applications/Optimizers/jni/main.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  // Measure deltas after training
+  auto ws_after = measure_weight_l2(*model);
+  auto delta = measure_delta_l2(*model, snap_before);
+  std::cout << "[DEBUG] weights_l2_after=" << ws_after.l2
+            << " delta_l2=" << delta.l2 << " elems=" << delta.num_elems
+            << std::endl;
+
+  return status;
+}
```
baek2sm: you need to add new line at the end of the script

PJH6029: Thanks. I've formatted all cpp/h files.

PR #3600, opened 2025-12-02
Enhancements in Thread Management, QINT4 Computations, and SVM Allocation Support

djeong20:
## Summary

- Added SVM allocation support with a secure setter pattern for MemoryData to improve memory management flexibility and security.

- Updated the FloatTensor class to support QINT4 computations using Q4_0 GEMM/GEMV, with plans to transition to standard INT4 computations in the future.

<details><summary>Add SVM allocation support with secure setter pattern for MemoryData</summary><br />

 This commit adds support for SVM allocation, using a secure setter pattern for MemoryData to enhance memory management flexibility and security.
This patch modifies memory_data.h to include SVM-specific members and methods, and updates memory_pool to handle SVM allocation with fallback.

**Self evaluation:**

</details>

<details><summary>QINT4 CPU Computations Using Q4_0 GEMM/GEMV</summary><br />

This commit updates the FloatTensor class to support computations using the QINT4 type by leveraging the Q4_0 GEMM/GEMV functionality.
In the future, this will be replaced with standard INT4 computations.
To effectively utilize this feature, the QINT4 data must be converted into the Q4_0 format.

**Self-evaluation:**

</details>

### DIFF
--- a/nntrainer/tensor/float_tensor.cpp
+++ b/nntrainer/tensor/float_tensor.cpp
@@ -17,6 +17,8 @@
 #include <cpu_backend.h>
 #include <float_tensor.h>
 #include <int4_tensor.h>
+#include <q4_0_utils.h>
+
 #include <tensor.h>
 #include <util_func.h>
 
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/float_tensor.cpp
```diff
@@ -722,67 +724,75 @@ void FloatTensor::dot(std::vector<Tensor *> input, std::vector<Tensor *> output,
...
+      }
       if (M == 1) {
-        gemm_q4_0(M, N, K, data, K, (void *)mdata, N, rdata, N);
+        gemv_int4_async_cl(mdatas, scales, data, rdatas, K, Ns,
+                           Int4QTensor::getGroupSize());
       } else {
-        Ns.push_back(N);
-        mdatas.push_back(mdata);
-        rdatas.push_back(rdata);
+        openvino_gemm_async_cl(data, mdatas, scales, rdatas, M, Ns, K,
```
jijoongmoon, line 768: I don't think this name is valid for ARM with SVM.

nntrainer/tensor/float_tensor.cpp
```diff
@@ -980,17 +990,24 @@ Tensor &FloatTensor::dotQInteger(Tensor const &input, Tensor &output,
...
-  if (M == 1) {
-    gemv_int4_cl(mdata, input.getScale<uint16_t>(), data, rdata, K, N,
-                 Int4QTensor::getGroupSize());
+  if (input.getMemoryData()->isSVM() && output.getMemoryData()->isSVM() &&
+      getMemoryData()->isSVM()) {
+    if (M == 1) {
+      gemv_int4_cl(mdata, input.getScale<uint16_t>(), data, rdata, K, N,
+                   Int4QTensor::getGroupSize());
+    } else {
+      openvino_sgemm_cl(data, mdata, input.getScale<uint16_t>(), rdata, M, N, K,
```
jijoongmoon, line 1005: Ditto.

nntrainer/tensor/float_tensor.cpp
djeong20: I agree. I think we should create a new PR for renaming

nntrainer/tensor/float_tensor.cpp
```diff
@@ -722,67 +724,75 @@ void FloatTensor::dot(std::vector<Tensor *> input, std::vector<Tensor *> output,
...
-                             Int4QTensor::getGroupSize());
+      gemm_q4_0(M, Ns[i], K, data, K, mdatas[i], Ns[i], rdatas[i], Ns[i]);
     }
-#endif
-  } else {
+  } else { // QINT4
+    /// @note It is essential to understand that this section of the code
+    /// requires the `input` data to be converted to Q4_0 type, not QINT4 type.
+    /// This should be replaced with standard CPU INT4 computation instead of
+    /// using Q4_0.
```
skykongkong8, line 789: I noticed this comment:
```cpp
/// @note It is essential to understand that this section of the code
/// requires the `input` data to be converted to Q4_0 type, not QINT4 type.
```
This is a big assumption! If a user accidentally passes unconverted QINT4 data, the results will be silently incorrect. Could we add a runtime check or assertion here? Something like:
```cpp
NNTR_THROW_IF(/* condition to detect unconverted data */, std::runtime_error)
<< "QINT4 data must be converted to Q4_0 format before CPU computation";
```

djeong20: Thank you for the detailed review. In my opinion, we need to implement a significant refactor for this. 

Specifically, the classes Q4_0_Tensor, Q4_K_Tensor, and Int4QTensor should be combined into a single class. Alternatively, we could create a base class from which all three classes inherit. 

Additionally, we should incorporate `QScheme` to manage this more effectively. Since these three classes have similar functionalities, when data is converted to Q4_0, the `QScheme` should be updated accordingly. This will help us determine the relevant configurations more efficiently.

PR #3602, opened 2025-12-04
CUDA int4 GEMM 

dkjung:
- Add CPU implementation of int4 GEMM with packed block optimization
- Add CUDA implementation of int4 GEMM operation
- Add unit tests for int4 GEMM operations
- Update quantize related files to support int4 operations
- Fix function name from int4 to int8 in quantize functions
- Add utility functions for matrix operations and vector generation

### DIFF
--- a/nntrainer/tensor/cuda_operations/gemm_int4_cpu.cpp
+++ b/nntrainer/tensor/cuda_operations/gemm_int4_cpu.cpp
@@ -1,0 +1,175 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2025 Samsung Electronics Co., Ltd. All Rights Reserved.
+ *
+ * @file   gemm_int4_cpu.cpp
+ * @brief  CPU implementation of int4 GEMM operation
+ * @author Samsung Electronics Co., Ltd.
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#include "gemm_int4_cpu.h"
+#include "fp16.h"
+#include "int4_utils.h"
+#include <cmath>
+#include <iostream>
+#include <vector>
+
+namespace nntrainer {
+
+void gemm_int4_cpu(const void *input, const void *weights, const void *scales,
+                   const void *input_scales, float *output, unsigned int M,
+                   unsigned int N, unsigned int K,
+                   unsigned int quantization_group_size) {
+
+  const int8_t *q_input = static_cast<const int8_t *>(input);
+  const uint8_t *q_weights = static_cast<const uint8_t *>(weights);
+  const uint16_t *w_scales = static_cast<const uint16_t *>(scales);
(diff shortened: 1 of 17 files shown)


nntrainer/tensor/cuda_operations/gemm_int4_cpu.cpp
```diff
@@ -0,0 +1,175 @@
...
+      // Block index = n_blk * (K / 2) + k_blk
+      // Each block is 32 bytes
+      const uint8_t *w_block = q_weights + (n_blk * (K / 2) + k_blk) * 32;
+
+      // Process for all M rows of input
+      for (unsigned int m = 0; m < M; ++m) {
+        // Load input values and scale
+        // k_start and k_start+1 are in the same group (assuming group_size >=
+        // 2)
+        unsigned int group_id_in_row = k_start / quantization_group_size;
```
myungjoo, line 122: This does not need to be calculated in this loop. This can be moved out of this loop block to save calculation.

nntrainer/tensor/cuda_operations/gemm_int4_cpu.cpp
```diff
@@ -0,0 +1,175 @@
...
+        unsigned int global_group_id = m * groups_in_row + group_id_in_row;
+
+        // Input scale
+        float i_scale = compute_fp16_to_fp32(in_scales[global_group_id * 2]);
+
+        // Input indices
+        // Note: input is quantized with padding, so we use block addressing
+        unsigned int offset_in_group = k_start % quantization_group_size;
```
myungjoo, line 130: This does not need to be calculated in this loop. This can be moved out of this loop block to save calculation.

myungjoo (commented):
Unit test should include the following cases:
- N % 32 > 0
- K % 2 > 0
- Small M (M=1, 5)
- CUDA Error Handling (calling CUDA-enabled binaries in non-CUDA environment. see if it handles the error nicely)

nntrainer/tensor/cuda_operations/gemm_int4_cpu.cpp
```diff
@@ -0,0 +1,175 @@
...
+  // Iterate over columns of output (N)
+  for (unsigned int n = 0; n < N; ++n) {
+    // 1. Dequantize the n-th row of weights (which is n-th col of B)
+    Int4Utils::dequantizePackedRow(
+      const_cast<uint8_t *>(q_weights), const_cast<uint16_t *>(w_scales), N, K,
+      quantization_group_size, n, dequantized_weight_row.data());
+
+    // 2. Compute dot product with each row of input (M)
+    for (unsigned int m = 0; m < M; ++m) {
+      float sum = 0.0f;
```
myungjoo, line 52: 1. Please check if having loop-m inside loop-k would be faster than loop-k inside loop-m or not.
2. Try to cache caculation as much as possible. (looks like the same calculation/memory-op is repeated in the loop)

PR #3603, opened 2025-12-04
Fix Windows Build Warnings

djeong20:
**Summary of Code Fixes:**

1. **Type Conversion Warnings (C4244, C4267, C4146, C4293):**
   - Fixed across **19 files** (layers, optimizers, dataset handlers, tensor operations).  
   - Added explicit type casts (e.g., `static_cast<unsigned int>`).  
   - Used literal suffixes for precision (e.g., `1.0f` instead of `1.0`).  

2. **MSVC Compiler Flag Warnings (D9024, D9025, D9027):**
   - Removed manual `/MT` and `/MD` flags from MSVC configuration.  
   - Resolved runtime library conflicts with subprojects (CLBlast, iniparser).  

3. **Hexadecimal Floating-Point Warnings (C4305, C4309):**
   - Fixed in `avx2_impl.cpp` by adding `f` suffix to literals (e.g., `0x1.c6af84b912394p-5f`).  
   - Ensured single-precision float type instead of double.


songgot (approved):
LGTM

PR #3608, opened 2025-12-05
Enhancing Performance of INT4 Data Transformation

djeong20:
This pull request speeds up the transformation of int4 data in the osv32_isv2 layout to block_q4_0x4 layers by utilizing ARM NEON and OpenMP. The previous version took about 7 to 8 milliseconds to transform a 3072x8192 matrix, while the current patch takes only 2 to 4 milliseconds.

**Self-evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -525,11 +525,17 @@ void create_q4_0_weights(const uint8_t *int4_weight, uint8_t *q4_0_weight) {
   __fallback_create_q4_0_weights(int4_weight, q4_0_weight);
 }
 
+/// @todo rename it to `transform_int4_osv32_isv2_to_q4_0`
 void transform_q4_0x_from_int4(size_t N, size_t K, const uint8_t *osv32_weights,
                                const uint16_t *osv32_scales,
                                size_t scale_group_size, void *dst_q4_0x) {
+#if defined(__aarch64__) || defined(_M_ARM64)
+  neon::transform_int4_osv32_isv2_to_q4_0x4(N, K, osv32_weights, osv32_scales,
+                                            scale_group_size, dst_q4_0x);
+#else
   Q4_0Utils::transformQ4_0x_FromInt4(N, K, osv32_weights, osv32_scales,
                                      scale_group_size, 4, dst_q4_0x);
+#endif
 }
 
 } /* namespace nntrainer */
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
@@ -27,6 +27,7 @@
 #ifdef ARMV7
 #include <armv7_neon.h>
 #endif
+#include "nntr_ggml_impl_common.h"
 #include <fallback_internal.h>
 #include <util_func.h>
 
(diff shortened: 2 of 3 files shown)


nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -1340,6 +1341,172 @@ void clamp(const float *input, float *output, size_t length, float lower_bound,
...
+                                         const uint8_t *osv32_weights,
+                                         const uint16_t *osv32_scales,
+                                         size_t scale_group_size,
+                                         void *dst_q4_0x4) {
+  NNTR_THROW_IF((!(scale_group_size == 32 || scale_group_size == 64 ||
+                   scale_group_size == 128)),
+                std::invalid_argument)
+    << "Scale group size must be 32/64/128";
+  NNTR_THROW_IF(K % QK4_0 != 0, std::invalid_argument)
+    << "K size must be divisable by QK4_0 (32)";
```
songgot: Is this a typo? 'divisable -> 'divisible'

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -1340,6 +1341,172 @@ void clamp(const float *input, float *output, size_t length, float lower_bound,
...
+  const size_t rows_count_pad = align(N, ROW_BLOCK_SIZE);
+  const size_t columns_count_pad = align(K, ROW_BLOCK_SIZE);
+  const size_t column_blocks_count = columns_count_pad / 2;
+  const size_t bytes_per_row_block_span = column_blocks_count * ROW_BLOCK_SIZE;
+  const size_t num_blocks_per_row = K / QK4_0;
+
+  block_q4_0x4 *dst_ptr = reinterpret_cast<block_q4_0x4 *>(dst_q4_0x4);
+
+#pragma omp parallel for schedule(static)
+  for (long long row_id = 0; row_id < (long long)N;
```
songgot: How about using code below?
```(long long)N```  ->  ```static_cast<long long>(N)```
```

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -528,8 +528,13 @@ void create_q4_0_weights(const uint8_t *int4_weight, uint8_t *q4_0_weight) {
 void transform_q4_0x_from_int4(size_t N, size_t K, const uint8_t *osv32_weights,
                                const uint16_t *osv32_scales,
                                size_t scale_group_size, void *dst_q4_0x) {
+#if defined(__aarch64__) || defined(_M_ARM64)
+  neon::transform_int4_osv32_isv2_to_q4_0x4(N, K, osv32_weights, osv32_scales,
```
jijoongmoon, line 533: We do need corret the fucntion name. It might be better to be transform_int4_to_q4_0. Also Q4_0Utils::transformQ4_0x_FromInt4 need to change.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
jijoongmoon: size_t might be better.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
djeong20: fixed! thanks

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
djeong20: Thank you for sharing your opinion! OpenMP 2.0, being an earlier version, does not support unsigned integral types. As a result, I utilized the `long long` type to ensure compatibility with OpenMP 2.0.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
djeong20: I agree. When we support AVX2 acceleration for the x86 arch, let's refactor the naming as well. For now, I left a comment that the naming should be changed

skykongkong8: In intrinsic-level transform, it is quite common to name function like `transform_A_B`, when this function transforms to A from B. This is why many casting/copying/transforming functions in `cpu_backend.h` is named like to-from manner, not like from-to manner. 😓
However, if many of team members believe this way is more intuitive, let's change to this naming convention from now on!
 
Just to make sure that everybody is sync-up with current rapid updates in Int4 Quantization formats, `osv32_isv2` , `q4_0x4`, and `qsi4cxp` are all subset of Int4 Quantization, and should be named differently!

PR #3609, opened 2025-12-05
[ backend ] Speed up transformation from int4 to Q4_0x (for arm and x86)

mnalewaj:
[ backend ] Speed up transformation from int4 to Q4_0x (for NEON and AVX2)
- adding multithreading and other improvements
For matrix K:3072 N:8192
NEON: ~3ms ---> ~1.35ms (Galaxy S23)
AVX2: ~17ms --> ~0.62ms (Intel(R) Core(TM) i7-12700)

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
@@ -1375,230 +1375,55 @@ void clamp(const float *input, float *output, size_t length, float lower_bound,
   }
 }
 
-/**
- * @brief Highly optimized version - processes 4 rows simultaneously
- * and writes directly to output block, eliminating intermediate copies.
- * Uses 128-bit NEON operations and prefetching for maximum throughput.
- */
-inline static void neon_transform_4rows_to_q4_0x4(
-  const uint8_t *__restrict row0_ptr, const uint8_t *__restrict row1_ptr,
-  const uint8_t *__restrict row2_ptr, const uint8_t *__restrict row3_ptr,
-  uint16_t scale0, uint16_t scale1, uint16_t scale2, uint16_t scale3,
-  block_q4_0x4 *__restrict out) {
-
-  // Prefetch next cache lines
-#ifndef _MSC_VER
-  __builtin_prefetch(row0_ptr + 64, 0, 3);
-  __builtin_prefetch(row1_ptr + 64, 0, 3);
-  __builtin_prefetch(row2_ptr + 64, 0, 3);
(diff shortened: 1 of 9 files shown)


djeong20:
> Hello, I had to find the cause https://github.com/nnstreamer/nntrainer/actions/runs/19970196417/job/57272706916?pr=3609#logs I found it and fixed it. Now I'll do what you need. Are you sure you don't want this NEON implementation? It is faster than current on the main - 2.1ms on single core (for matrix K:3072 N:8192). Multi core should be faster (probably about 2 times).

Hi @mnalewaj, could you please share the benchmark results for each device from your work and the current main branch?

djeong20:
Hi @mnalewaj, 

Thank you for your prompt response and quick follow-up. I conducted a new benchmark on the Galaxy Book 5 Edge using only a single thread, and here are the results.

| Resolution   | Previous | Your work |
|--------------|----------|-----------|
| 3072x256     | 0.081 ms | 0.078 ms  |
| 3072x3072    | 0.879 ms | 0.904 ms  |
| 3072x8192    | 2.912 ms | 3.594 ms  |
| 8192x3072    | 2.42 ms  | 3.048 ms  |

I believe we are very close to achieving our goal, and we should definitely consider adding multithreading to the latest solution.

djeong20:
Hi @mnalewaj, I tested your work on the Galaxy Book 5 Edge, and the performance is now quite similar to the previous version. You can go ahead and delete the `_old` implementation. Once you're finished, please mark this PR as ready for review and rebase to the latest upstream main branch. I appreciate your hard work!

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -1447,62 +1499,70 @@ void transform_int4_osv32_isv2_to_q4_0x4(size_t N, size_t K,
...
+    << "K size must be divisable by QK4_0 (32)";
+  NNTR_THROW_IF(N % 8 != 0, std::invalid_argument)
+    << "N size must be divisable by 8";
+
+  static constexpr const size_t NUM_Q4_0_BLOCKS = 4;
+  static constexpr const size_t ROW_BLOCK_SIZE = 32;
+  static constexpr const size_t COLUMN_BLOCK_SIZE = 2;
+  static constexpr const size_t ROW_BLOCK_BYTE_SIZE = 16;
 
+  // uint8_t dst_tmp[8 * ROW_BLOCK_BYTE_SIZE];
```
djeong20:
```suggestion
```

jijoongmoon (approved):
LGTM

PR #3611, opened 2025-12-08
[Application] move example applications into examples subdirectory

baek2sm:
the applications directory contains both example-purpose applications
and implementations of well-known models. To make the applications
folder easier to navigate, the example applications have been moved into
a separate `examples` subfolder.


nntrainer-windows-resource
djeong20, line 1: It seems you are using an outdated version of nntrainer-windows-resource, which has been included in this commit. Please remove the changes and update the submodule!

baek2sm: That file shouldn’t have been included in this commit. thank you !!

baek2sm:
Separating the example applications into its own directory is one possible approach, but I think merging all applications into a single application is the better direction, so I’ll close this PR.

PR #3615, opened 2025-12-10
[cpu_backend] Introduce version-selectable arm build & refactor `ggml_interface_impl.cpp`

skykongkong8:
### 1. Version-selectable android build
Previously, nntrainer android build supported hard-coded armv8.2-a build only.
By adding meson option `arm-arch` : `armv7l` / `armv8.2-a` / `armv9.2-a`, and json files at `tools/cross` directory, now we can build with specific hw-arch-version like:
```bash
./tools/package_android.sh --arm-arch=armv9.2-a
```
or
```bash
./tools/android_test.sh --arm-arch=armv8.2-a
```

### 2. Refactor `ggml_interface_impl.cpp`
As @myungjoo requested from #3543 
#### AS-IS
```
cpu_backend/ggml_interface/nntr_ggml_impl
ㄴ nntr_ggml_impl.cpp
```
#### TO-BE
```
cpu_backend/ggml_interface/nntr_ggml_impl
ㄴ nntr_ggml_impl_neon.cpp // NEON only (includes ISA extension preprocessors)
ㄴ nntr_ggml_impl_avx2.cpp // AVX2 only
ㄴ nntr_ggml_impl_sve.cpp // SVE and NEON (if SVE is unavailable)
...
```
And include one with meson at build time to minimize redundant preprocessors.

## Notes
- I intentionally removed unnecessary preprocessors like `#ifdef` from `ggml_interface_impl.cpp`-related files, but this  makes msvc compiler in Windows to catch inline assembly, making the **Windows arm build CI fail**. Are we going to get rid of msvc with clang compiler?

### DIFF
--- a/nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/meson.build
+++ b/nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/meson.build
@@ -14,7 +14,10 @@ ggml_impl_interface_sources = [
 arch = host_machine.cpu_family()
 arm_arch_opt = get_option('arm-arch')
 
-if arch == 'aarch64' or get_option('platform') == 'android'
+message('arm_arch_opt : ' + arm_arch_opt)
+
+# arm_arch_opt != 'none' case for cross-compilation scenario
+if arch == 'aarch64' or arm_arch_opt != 'none'
   if arm_arch_opt == 'armv9.2-a'
     # SVE/SVE2 implementation for armv9.2-a
     ggml_impl_interface_sources += 'nntr_ggml_impl_sve.cpp'
--- a/tools/package_android.sh
+++ b/tools/package_android.sh
@@ -33,6 +33,10 @@ for arg in "$@"; do
 done
 
 # If --arm-arch specified, read configuration from JSON file
+if [[ -z "$arm_arch" ]]; then
+    arm_arch="armv8.2-a"
+fi
+
 if [[ -n "$arm_arch" ]]; then
     # Convert dots to dashes for filename (e.g., armv8.2-a -> armv8-2-a)
     arch_filename=$(echo "$arm_arch" | sed 's/\./-/g')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -67,9 +67,9 @@ option('libiomp_root', type:'string', value: './libiomp_win')
 
 # ARM architecture options
 option('arm-arch', type: 'combo',
-       choices: ['auto', 'armv7l', 'armv8.2-a', 'armv9.2-a'],
-       value: 'auto',
-       description: 'Target ARM architecture. "auto" detects from host for native builds, defaults to armv8.2-a for cross-compilation.')
+       choices: ['none', 'armv7l', 'armv8.2-a', 'armv9.2-a'],
+       value: 'none',
+       description: 'Target ARM architecture. "none" disables arm related features.')
 option('arm-march', type: 'string',
        value: '',
        description: 'ARM march flag (e.g., "-march=armv8.2-a+fp16+dotprod+i8mm"). If empty, derived from arm-arch.')


tools/package_android.sh
```diff
@@ -41,8 +41,10 @@ if [[ -n "$arm_arch" ]]; then
         echo "Using ARM architecture config from: $json_file"
         # Read values from JSON using Python (portable, no jq dependency)
         enable_fp16=$(python3 -c "import json; print(json.load(open('$json_file'))['enable_fp16'])")
-        # Add arm-arch to meson args
+        arm_march=$(python3 -c "import json; print(json.load(open('$json_file'))['arm_march'])")
```
songgot: How about running the Python process just once and loading all the variables at once?
```
eval $(python3 -c "
import json
try:
    data = json.load(open('$json_file'))
    print(f'enable_fp16={data.get(\"enable_fp16\")}')
    print(f'arm_march=\"{data.get(\"arm_march\", \"\")}\"')
except Exception as e:
    exit(1)
")
```

nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/nntr_ggml_impl_avx.cpp
```diff
@@ -0,0 +1,9884 @@
...
+            rhs_mat_2367ABEF_13,
+            (_MM_PERM_ENUM)221); // B12(28-31) B13(28-31) B12(28-31) B13(28-31)
+                                 // B16(28-31) B17(28-31) B16(28-31) B17(28-31)
+                                 // B1A(28-31) B1B(28-31) B1A(28-31) B1B(28-31)
+                                 // B1E(28-31) B1F(28-31) B1E(28-31) B1F(28-31)
+
+          uint32_t utmp_00[4], utmp_01[4], utmp_10[4], utmp_11[4];
+
+          // Scales and Mins of corresponding sub blocks from different Q4_K
+          // structures are stored together The below block is for
```
songgot: I think it's a typo. 'multipled' -> 'multiplied'.

tools/package_android.sh
skykongkong8: That's a great suggestion! It would reduce process overhead and help avoiding any unlikely race condition if the file changes between calls during the build by accident (more secure). I will update this 👍

meson.build
```diff
@@ -176,10 +176,23 @@ if get_option('enable-fp16')
...
+       extra_defines += '-DARM_ARCH_ARMV9_2A=1'
+       extra_defines += '-DENABLE_SVE2=1'
+     elif arm_arch_opt == 'armv8.2-a'
+       add_project_arguments('-march=armv8.2-a+fp16+dotprod+i8mm', language: ['c', 'cpp'])
+       extra_defines += '-DARM_ARCH_ARMV8_2A=1'
+     else
+       # auto: default to armv8.2-a+fp16+dotprod+i8mm for native aarch64 builds
+       add_project_arguments('-march=armv8.2-a+fp16+dotprod+i8mm', language: ['c', 'cpp'])
+       extra_defines += '-DARM_ARCH_ARMV8_2A=1'
+     endif
```
EunjuYang, line 196: I think it would be helpful to emit a diagnostic message for the default case.
Currently, the fallback is silent, so it's hard for users to know whether they're getting the intended default (`arm-arch=auto`) or have endedup here due to a misconfiguration.
Therefore, I suggest adding a `message()` for this fallback.

skykongkong8: Sounds reasonable! I applied this change👍

skykongkong8:
#3656

jijoongmoon:
For better maintain, we do need add build tests according to the architecture.

tools/cross/android_armv9-2-a.json
```diff
@@ -0,0 +1,6 @@
+{
+    "arm_arch": "armv9.2-a",
+    "arm_march": "armv9.2-a+fp16+sve2+sme",
+    "enable_fp16": true,
+    "app_abi": "arm64-v8a"
+}
```
jijoongmoon, line 6: I don't think we do need this sepearte json files. Could you merge them? and add new line at the end of file.

.github/workflows/windows_arm.yml
```diff
@@ -1,34 +0,0 @@
...
-      - name: Prepare MSVC
-        uses: bus1/cabuild/action/msdevshell@v1
-        with:
-          architecture: arm64
-      - name: Prepare Build
-        run: meson setup --native-file windows-native.ini -Denable-blas=false builddir
-      - name: Run Build
-        run: meson compile -C builddir
-      - name: Run Test Suite
-        run: meson test -C builddir
```
jijoongmoon, line 34: I'm not sure deleting this file. It is needed to support windows arm build. @djeong20 , Could you confirm it?

nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/meson.build
```diff
@@ -4,16 +4,39 @@ ggml_impl_interface_headers = [
...
-  'nntr_ggml_impl.cpp',
   'nntr_ggml_impl_quant.cpp',
   'nntr_ggml_impl_fp16_fp32.cpp',
 ]
 
+# Build-time exclusive source selection (ARM first, then x86)
+arch = host_machine.cpu_family()
+arm_arch_opt = get_option('arm-arch')
+
+if arch == 'aarch64' or get_option('platform') == 'android'
```
myungjoo: This condition (android or aarch64) appears very strange and dangerous.

What if:
- armv7l + linux
- x64 + android
- ....

?



Plus, mixing up os and arch has made the code very unreadable. And this appears to be a bug.

.github/workflows/windows_arm.yml
skykongkong8: I thought we already discussed about removing Windows msvc build script from CI in offline.
@djeong20 also agreed.

tools/cross/android_armv9-2-a.json
skykongkong8: Got it.

nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/meson.build
skykongkong8: I see! Architecture selection should be strictly done with architecture only, not OS.. my mistake

skykongkong8: Uh, but in a second thought, I just recalled `get_option('platform') == 'android'` should also take place to check the cross-compilation scenario from linux PC. This option is produced from cross-compilation scenario only, not from the native android build.
In such scenario, I think we have no choice but have to provide explicit information like  `get_option('platform') == 'android'` or 
how about strictly limiting like:
```c++
is_aarch64 = host_machine.cpu_family() == 'aarch64'
# Cross-compilation fix for Android NDK without proper cross-file
if not is_aarch64 and get_option('platform') == 'android'
  if cxx.get_define('__aarch64__') != ''
    is_aarch64 = true
  endif
endif
```

Or maybe I can just fix `get_option('platform') == 'android'` to `get_option('platform') == 'android-crosscompile'` to minimize confusion.

skykongkong8: As we discussed offline, when cross-compiling in Linux PC, for aarch64 android target with ndk-build,
```
arch = host_machine.cpu_family()
target_arch = target_machine.cpu_family()
build_arch = build_machine.cpu_family()
...
message('host_arch: ' + arch)
message('target_arch: ' + target_arch)
message('build_arch: ' + build_arch )
```
produces:
```
Message: host_arch: x86_64
Message: target_arch: x86_64
Message: build_arch: x86_64
```

I will rather add an option `aarch64-crosscompile`. Thanks!!

skykongkong8: In a third thought, additional option is no longer needed since I already made one in this PR.
```
# arm_arch_opt != 'none' case for cross-compilation scenario
if arch == 'aarch64' or arm_arch_opt != 'none'
  if arm_arch_opt == 'armv9.2-a'
...
```
In native-build, specific architecture will be given from meson through `arch`
In cross-compilation, we always have `arm_arch_opt`.

PR #3625, opened 2025-12-11
rotary_emb_kernel fix

Anna-Szal:

<details><summary>rotary_emb_kernel fix</summary><br />

attention kernels attempted to write to a read-only buffer of size "unused_buffer_bytes" = sizeof(float).

Please note that the 2 tests affected by this bug (rotary_emb_kernel_FP32 and rotary_emb_kernel_FP32_case2) still randomly fail (from run to run), which means there's still more to debug. 

**Self evaluation:**

</details>

### Summary

- fixed a bug in CL Buffer Manager 

### DIFF
--- a/nntrainer/cl_buffer_manager.cpp
+++ b/nntrainer/cl_buffer_manager.cpp
@@ -22,9 +22,9 @@ namespace nntrainer {
 void ClBufferManager::initBuffers() {
   inBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   inBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  inBufferC = new opencl::Buffer(context_inst_, unused_buffer_bytes, true);
+  inBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   outBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
-  outBufferB = new opencl::Buffer(context_inst_, unused_buffer_bytes, false);
+  outBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
 
   data_input = context_inst_.createSVMRegion(buffer_size_bytes);
   for (unsigned int i = 0; i < max_qs; ++i) {
--- a/nntrainer/tensor/cl_operations/attention_kernels_templates.h
+++ b/nntrainer/tensor/cl_operations/attention_kernels_templates.h
@@ -86,14 +86,14 @@ inline static void rotary_emb_cl_internal(
       break;
     }
 
-    result = cl_buffer_manager.getInBufferC()->WriteDataRegion(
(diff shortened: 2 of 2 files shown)


nntrainer/cl_buffer_manager.cpp
```diff
@@ -22,9 +22,9 @@ namespace nntrainer {
 void ClBufferManager::initBuffers() {
   inBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   inBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  inBufferC = new opencl::Buffer(context_inst_, unused_buffer_bytes, true);
+  inBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   outBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
-  outBufferB = new opencl::Buffer(context_inst_, unused_buffer_bytes, false);
+  outBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, false);
```
myungjoo, line 27: This is additional 2 x 32MiB. @jijoongmoon Is it ok?

nntrainer/cl_buffer_manager.cpp
```diff
@@ -22,9 +22,9 @@ namespace nntrainer {
 void ClBufferManager::initBuffers() {
   inBufferA = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
   inBufferB = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
-  inBufferC = new opencl::Buffer(context_inst_, unused_buffer_bytes, true);
+  inBufferC = new opencl::Buffer(context_inst_, buffer_size_bytes, true);
```
myungjoo, line 25: ```attention_kernels_template.h``` is the only one using inBufferC. Isn't there any mechanism not to allocate this by default? (allocate only when someone first call getInBufferC();?)

This applies to other in/out Buffer A/B/C. You are by default allocating 32MiB x 5, which is A LOT for embedded devices.

PTAL: @jijoongmoon  (you now really need a memory-usage regression test)

djeong20: I will respond on behalf of Jijoong. I believe all buffers should be allocated upon their initial call. I will create a new PR that does not allocate a predefined buffer unless it is called.

djeong20: OpenCL buffers are now lazily initialized in #3643

Anna-Szal: since there is an incoming refactoring PR that will take care of this issue, do I just leave mine without changes?
(and changed the commit message only)

nntrainer/cl_buffer_manager.cpp
songgot: How about using '''unused_buffer_bytes'''?

PR #3626, opened 2025-12-11
l2norm test fix

Anna-Szal:

<details><summary>l2norm test fix</summary><br />

**test**: blas_kernels --> l2norm

**problem**: the output was compared against CPU version of the algorithm which turned out to be less precise

**solution**: using fp64 version of the CPU algorithm as ground truth.  To avoid including headers related to CBLAS directly, double-float versions of backend functions were implemented.

**Self evaluation:**

</details>

### Summary

- fix for failing blas_kernels.l2norm test

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
@@ -809,6 +809,12 @@ void sscal(const unsigned int N, const float alpha, float *X,
  * @param[in] X float * for Vector X
  */
 float snrm2(const unsigned int N, const float *X, const unsigned int incX);
+/**
+ * @brief     dnrm2 computation : Euclidean norm
+ * @param[in] N number of elements in X
+ * @param[in] X double * for Vector X
+ */
+double dnrm2(const unsigned int N, const double *X, const unsigned int incX);
 
 /**
  * @brief     copy function : Y = X
--- a/nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.cpp
+++ b/nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.cpp
@@ -73,6 +73,11 @@ float __cblas_snrm2(const unsigned int N, const float *X,
   return cblas_snrm2(N, X, incX);
 }
 
+double __cblas_dnrm2(const unsigned int N, const double *X,
+                    const unsigned int incX) {
+  return cblas_dnrm2(N, X, incX);
+}
+
 void __cblas_sgemm(const unsigned int TStorageOrder, bool TransA, bool TransB,
                    const unsigned int M, const unsigned int N,
                    const unsigned int K, const float alpha, const float *A,
--- a/nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.h
+++ b/nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.h
@@ -82,6 +82,13 @@ void __cblas_sscal(const unsigned int N, const float alpha, float *X,
 float __cblas_snrm2(const unsigned int N, const float *X,
                     const unsigned int incX);
 /**
+ * @brief     dnrm2 computation : Euclidean norm
+ * @param[in] N number of elements in X
+ * @param[in] X double * for Vector X
+ */
+double __cblas_dnrm2(const unsigned int N, const double *X,
+                    const unsigned int incX);
+/**
  * @brief     sgemm computation  : Y = alpha*op(A)*op(B) + beta*C,
  * where op(X) is one of X or X**T
  * @param[in] TStorageOrder Row major / Col major
--- a/nntrainer/tensor/cpu_backend/fallback/fallback.h
+++ b/nntrainer/tensor/cpu_backend/fallback/fallback.h
@@ -668,6 +668,13 @@ void sscal(const unsigned int N, const float alpha, float *X,
  */
 float snrm2(const unsigned int N, const float *X, const unsigned int incX);
 /**
+ * @brief     dnrm2 computation : Euclidean norm
+ * @param[in] N number of elements in X
+ * @param[in] X double * for Vector X
+ */
+double dnrm2(const unsigned int N, const double *X, const unsigned int incX);
+
+/**
  * @brief     copy function : Y = X
  * @param[in] N number of elements in X
  * @param[in] X float * for Vector X
--- a/nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
+++ b/nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
@@ -67,6 +67,19 @@ float __fallback_snrm2(const unsigned int N, const float *X,
   return sqrt(sum);
 }
 
+double __fallback_dnrm2(const unsigned int N, const double *X,
+                       const unsigned int incX) {
+  assert(incX > 0);
+  double sum = 0.0;
+  double tmp;
+
+  for (unsigned int i = 0; i < N; i++) {
+    tmp = X[i * incX];
+    sum += tmp * tmp;
+  }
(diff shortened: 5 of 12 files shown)


skykongkong8:
Please apply check c++ format checker

test/include/nntrainer_test_util.h
```diff
@@ -191,6 +191,20 @@ class ScopedIni {
...
   } while (0)
 
+#define GEN_TEST_INPUT_DOUBLE(input, equation_i_j_k_l)        \
+  do {                                                        \
+    for (int i = 0; i < batch; ++i) {                         \
+      for (int j = 0; j < channel; ++j) {                     \
+        for (int k = 0; k < height; ++k) {                    \
+          for (int l = 0; l < width; ++l) {                   \
+            float val = (equation_i_j_k_l);                   \
+            input[i*batch + j*channel + k*height + l] = val;  \
```
myungjoo: The index appears very strange...

```
((i*channel + j)*height + k)*width + l
```
?

test/include/nntrainer_test_util.h
```diff
@@ -191,6 +191,20 @@ class ScopedIni {
     }                                             \
   } while (0)
 
+#define GEN_TEST_INPUT_DOUBLE(input, equation_i_j_k_l)        \
```
myungjoo: Please take batch, channel, height, width as arguments. This damages readability and has potential errors for other developers later. 

This doesn't incur runtime overhead.

test/include/nntrainer_test_util.h
```diff
@@ -191,6 +191,20 @@ class ScopedIni {
...
+  do {                                                        \
+    for (int i = 0; i < batch; ++i) {                         \
+      for (int j = 0; j < channel; ++j) {                     \
+        for (int k = 0; k < height; ++k) {                    \
+          for (int l = 0; l < width; ++l) {                   \
+            float val = (equation_i_j_k_l);                   \
```
myungjoo: This is float. The caller in unittest_blas_kernel_cl.cpp is using double... is this intentional? (it's ok if it's intentional.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -721,6 +721,12 @@ void sscal(const unsigned int N, const float alpha, float *X,
  * @param[in] X float * for Vector X
  */
 float snrm2(const unsigned int N, const float *X, const unsigned int incX);
+/**
+ * @brief     snrm2 computation : Euclidean norm
```
myungjoo: snrm2? dnrm2!

nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.h
```diff
@@ -73,6 +73,13 @@ void __cblas_sscal(const unsigned int N, const float alpha, float *X,
  */
 float __cblas_snrm2(const unsigned int N, const float *X,
                     const unsigned int incX);
+/**
+ * @brief     snrm2 computation : Euclidean norm
```
myungjoo: snrm2? dnrm2!

nntrainer/tensor/cpu_backend/fallback/fallback.h
```diff
@@ -667,6 +667,13 @@ void sscal(const unsigned int N, const float alpha, float *X,
  * @param[in] X float * for Vector X
  */
 float snrm2(const unsigned int N, const float *X, const unsigned int incX);
+/**
+ * @brief     snrm2 computation : Euclidean norm
```
myungjoo: snrm2? dnrm2!

nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
```diff
@@ -67,6 +67,19 @@ float __fallback_snrm2(const unsigned int N, const float *X,
   return sqrt(sum);
 }
 
+double __fallback_dnrm2(const unsigned int N, const double *X,
+                       const unsigned int incX) {
+  assert(incX > 0);
+  double sum = 0.0f;
```
myungjoo: ```0.0f``` --> ```0.0```? (it's double. not float)

test/include/nntrainer_test_util.h
Anna-Szal: a brain glitch :D thanks

test/include/nntrainer_test_util.h
```diff
@@ -103,92 +103,107 @@ class ScopedIni {
...
-        for (int k = 0; k < width; ++k) {           \
-          for (int l = 0; l < channel; ++l) {       \
-            float val = (eqation_i_j_k_l);          \
-            input.setValue(i, l, j, k, val);        \
-          }                                         \
-        }                                           \
-      }                                             \
-    }                                               \
+#define GEN_TEST_INPUT_NHWC(input, eqation_i_j_k_l)                            \
+  do {                                                                         \
```
songgot, line 107: Was the space intentional?

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.h
```diff
@@ -570,7 +570,12 @@ void sscal(const unsigned int N, const float alpha, float *X,
  * @param[in] X float * for Vector X
  */
 float snrm2(const unsigned int N, const float *X, const unsigned int incX);
-
+/**
+ * @brief     snrm2 computation : Euclidean norm
+ * @param[in] N number of elements in X
+ * @param[in] X double * for Vector X
+ */
+double dnrm2(const unsigned int N, const double *X, const unsigned int incX);
```
songgot, line 578: There is no 'incX' in @param.

test/include/nntrainer_test_util.h
Anna-Szal: i used c++ format checker as was suggested. It does look better in editor I would say..

nntrainer/tensor/cpu_backend/x86/x86_compute_backend.h
Anna-Szal: i basically copied the comment from snrm2 which doesn't have description for incX either.. would you like me to write it for both functions? eg. how about "@param[in] incX stride within X"?

test/include/nntrainer_test_util.h
songgot: Good, Thank you.

nntrainer/tensor/cpu_backend/cblas_interface/cblas_interface.cpp
```diff
@@ -73,6 +73,14 @@ float __cblas_snrm2(const unsigned int N, const float *X,
   return cblas_snrm2(N, X, incX);
 }
 
+double __cblas_dnrm2(const unsigned int N, const double *X,
+                    const unsigned int incX) {
+#ifdef BLAS_NUM_THREADS
+  openblas_set_num_threads(BLAS_NUM_THREADS);
+#endif
```
myungjoo: Inconsistent with other functions in cblas_interface.cpp.

This has a side effect of setting blas-num-thread. Once this function set the num-thread, the effect remains and affects later calls of other functions.

For consistent behavior, other functions also need to call num-thread, or this function needs to return to "previous" num-thread.

What's the intention of calling num-thread here? (appears not related to the pull request topic)

Anna-Szal: this was a part of my battle with blas kernel tests timeout.. I added these lines and increased number of threads during building process, but now i am not at all sure that these lines specifically were crucial, also it is an ugly solution.. 
I currently have removed them and just build repo with extended test time limit, i don't know if this is a good solution either, since tests are intended to run automatically on github as i understand (unittest_blas_kernels_cl currently takes ~140 sec to run on my machine)

PR #3628, opened 2025-12-12
[cpu_backend/arm] Add KleidiAI QSI8D32P-QSI4C32P GEMM kernels with thread-backend support

skykongkong8:
This PR introduces optimized KleidiAI GEMM kernels for qsi8d32p_qsi4c32p quantization format on ARM, supporting both NEON/dotprod and SVE/i8mm instruction sets. Note that currently there is no official armv9.2-a build in the current nntrainer (planned to be introduced via #3615 in the near future), so SVE kernels won't work.

## Changes

- Add `kai_matmul_clamp_f32_qsi8d32p_qsi4c32p` ukernels with multiple tile configurations (4x4, 4x8, 8x4, 8x8) for NEON dotprod, NEON i8mm, and SVE i8mm
- Implement LHS quantization packing (`qsi8d32p_f32`) and RHS repacking (`qsi4c32p`) routines
- Add KleidiAI interface (`kleidiai_qsi8d32p.cpp`) with automatic kernel variant selection
- Extend thread-backend meson option to support OpenMP and std::thread parallelization for KleidiAI GEMM
- Add benchmark function `run_gemm_benchmark_comparison()` to compare latency of `qsi8d32p_qsi4c32p`, 
`qai8dxp_qsi4cxp`, and `gemm_q4_0<float>` implementations
- Update unit tests with new test cases for the `qsi8d32p_qsi4c32p` GEMM variant

## Impact
- This handles the same amount of information of quantization scheme along with the current GPU int4 kernels, allowing more possibilities for heterogenous hw-backend runtime.
- Since this is a group-quantized kernel, we can enjoy more precision-supplement  effect w.r.t. channelwise quantized kernels introduced from #3519


test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp
```diff
@@ -684,11 +684,490 @@ TEST(nntrainer_cpu_backend_standalone, qai8dxp_qsi4cxp_3072x512x512_CMP) {
...
+  // Benchmark: gemm_q4_0<float>
+  // ============================================================
+  nanoseconds total_time_q4_0 = nanoseconds(0);
+  for (uint32_t i = 0; i < test_iters; ++i) {
+    auto t1 = high_resolution_clock::now();
+    nntrainer::gemm_q4_0<float>(M, N, K, activation.data(), K,
+                                (void *)q4_0_repacked_qWeight.data(), N,
+                                dst_q4_0.data(), N);
+    auto t2 = high_resolution_clock::now();
+    total_time_q4_0 += duration_cast<nanoseconds>(t2 - t1);
```
myungjoo, line 1255: A test case that can never fail? (no assert/expect?)

skykongkong8: This code is `run_gemm_benchmark_comparison`, so this is only about latency comparison w.r.t. current 4bit GEMM kernels : `qsi8d32p_qsi4c32p_packed`, `qai8dxp_qsi4cxp_packed`, and `gemm_q4_0<float>`. More kernels with 4bit quantization scheme might be introduced in the near future, so I intentionally implemented like this.

test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp
```diff
@@ -680,11 +683,611 @@ TEST(nntrainer_cpu_backend_standalone, qai8dxp_qsi4cxp_3072x512x512_CMP) {
...
+    nntrainer::nntr_gemm_qsi8d32p_qsi4c32p_unpacked(
+      M, N, K, (void *)activations, (void *)rhs_native_mtx_qs4c32, nullptr,
+      dst.data(), transB); // scales are embedded in qs4c32 format
+  // #### MAIN TESTED METHOD ####
+  auto t2 = high_resolution_clock::now();
+  auto dt = duration_cast<nanoseconds>(t2 - t1);
+  if (print) {
+    std::cout << "[INFO] test_gemm_qsi8d32p_qsi4c32p_unpacked: " << dt.count()
+              << " ns " << dt.count() / 1'000 << " us "
+              << dt.count() / 1'000'000 << " ms " << std::endl;
```
haehun, line 724: Is there point in printing same thing in three different scales?

skykongkong8: From dimension to dimension, function latency might vary.
In case of small gemm problem for example, it might take a few nanoseconds, while it would take millions-scale nanosecond for big gemm problem.
> small :
> 20 ns 0 us 0 ms
> big :
> 123123123 ns 123123 us 123 ms

And often, different gemm functions are compared with each other, making readability of this print statement worse like:
> 123123123 ns 123123 us 123 ms
> 654654654 ns 654654 us 654 ms
> 123456789 ns 123456 us 123 ms

For easier comparison, I prefer to print all `ns`, `us`, and `ms` otherwise I should constantly change them for arbitrary matrix dimensions.

skykongkong8: From dimension to dimension, function latency might vary.
In case of small gemm problem for example, it might take a few nanoseconds, while it would take millions-scale nanosecond for big gemm problem.
> small :
> 20 ns 0 us 0 ms
> big :
> 123123123 ns 123123 us 123 ms

And often, different gemm functions are compared with each other, making readability of this print statement worse like:
> 123123123 ns 123123 us 123 ms
> 654654654 ns 654654 us 654 ms
> 123456789 ns 123456 us 123 ms

For easier comparison, I prefer to print all `ns`, `us`, and `ms` otherwise I should constantly change them for arbitrary matrix dimensions.

skykongkong8:
> LGTM!

@h0g1 
Discussed offline in-person :)
You can leave anything more if you have further questions👍

meson.build
```diff
@@ -172,7 +172,7 @@ if get_option('enable-fp16')
      # On ARM systems, you must enable this type explicitly with the -mfp16-format
      #   command-line option in order to use it.
      ## About FP16-SIMD in aarch64
-     # FP16-SIMD is supported since armv8.2. If you enable this forcibly, it won't be
+     # FP16-SIMD is supported since armv9.2. If you enable this forcibly, it won't be
```
djeong20:
```suggestion
     # FP16-SIMD is supported since armv8.2. If you enable this forcibly, it won't be
```

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -528,6 +528,92 @@ void nntr_gemm_qai8dxp_qsi4cxp_packed(
...
   bool transB = true, T lower_bound = std::numeric_limits<T>::lowest(),
   T upper_bound = std::numeric_limits<T>::max());
+
+/**
+ * @brief get size of memory to allocate for rhs weight packing of qsi8d32p to
+ * qsi4c32p
+ *
+ * @param n row length if not transposed
+ * @param k col length if not transposed
+ * @return size_t size of memory to allocate
```
djeong20, line 540: Please add a description for `idx_variant` and `transB`.

meson.build
skykongkong8: Seems like I missed this while changing everything all at once 😓 Thanks!

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
skykongkong8: Fixed!

PR #3632, opened 2025-12-12
[onnx] fix issue of weight layer

baek2sm:
It's rebased version PR of #3413 and fixed some unittest issues.

The input layer was identified during compilation based on the number of
input connections(or property). Howerver, weight layer also has no input
connection, so it causes some issues. this commit resolved these issues.


baek2sm:
I've uploaded other pr

PR #3643, opened 2025-12-15
Lazy Buffer Initialization and Refactor OpenCL Buffer Manager

djeong20:
## Dependency of the PR
None

<details><summary>Lazy Buffer Initialization and Refactor OpenCL Buffer Manager</summary><br />

This PR implemented lazy buffer initialization for OpenCL buffers and reorganized the buffer manager structure.

- opencl::Buffer objects are lazily initialized in their getters
- Moved nntrainer/cl_buffer_manager to nntrainer/opencl/opencl_buffer_manager

**Self-evaluation:**

Donghyeon Jeong <dhyeon.jeong@samsung.com>

</details>


nntrainer/opencl/opencl_buffer_manager.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include <cstring>
+
+#include <opencl_buffer_manager.h>
+#include <opencl_loader.h>
+
+namespace nntrainer {
+
+void ClBufferManager::initBuffers() {
+  data_input = context_inst_.createSVMRegion(buffer_size_bytes);
```
jijoongmoon, line 22: For now, I think it is OK. But one thing to check. What is the default value of buffer_size_bytes? We don't have to set it accordingly depending on the networks, do we?

djeong20: The default size is 32MB. I completely agree that the size should be allocated dynamically based on the network. The main goal of this pull request (PR) is to reduce the preallocated buffer size. The issue you raised should definitely be addressed, but it would be better to handle it in a separate PR.

jijoongmoon (approved):
LGTM

PR #3644, opened 2025-12-15
[do not merge][layers] update weight layer

baek2sm:
The current version of NNTrainer determines whether a layer is an input layer based on the p
resence of an input shape. If there is no input shape, the layer is treated as an input layer, w
hich causes a problem because weight layers also lack an input shape and are therefore mistakenl
y considered input layers.

 To address this, the commit introduces a weight_dim property that sets the dimension of weig
ht layers, allowing the system to distinguish between input layers and weight layers.

 However, this version is not the final set of changes. A separate PR incorporating the final
 modifications will be created in this week, and I will request a merge of that PR together with this one. The current version is just a working version commit before the SRIB commit.

** Things to be done:
 - Since a new property was added to layer_node, the current layer_impl cannot access it. I plan to improve this by parsing the property into a local variable during the finalize stage of layer_node so that it can be used at the layer_impl level.
 - Add unit tests for weight layer.


nntrainer/layers/layer_node.cpp
```diff
@@ -490,6 +502,10 @@ const std::vector<TensorDim> LayerNode::getOutputDimensions() const {
   return dims;
 }
 
+bool LayerNode::isWeightNode() const {
+  return layer->getType() == "weight" ? true : false;
+}
```
songgot: How about using code below?
```
return layer->getType() == "weight";
```

baek2sm: thank you that's my fault

songgot: Thank you.

Issue #3654, opened 2025-12-18
Create `Encoder` like `CausalLM`


EunjuYang:
We currently have an application called **CausalLM**. This app defines a CausalLM model and registers it with a factory, allowing users to easily run transformer models in a causal‑LM configuration.

In the same spirit, we also need to add support for **embedding** models. The embedding functionality should be implemented so that it can directly leverage the embedding layer defined in CausalLM. This capability is essential for future on‑device use‑cases, such as on‑device Retrieval‑Augmented Generation (RAG). By providing a straightforward way to build and use an on‑device embedder, we will enable these advanced features to be deployed efficiently.

EunjuYang:
What do you think about adding a new class called Transformer to create a unified interface for various transformer-based models?
By doing this, we can share the common architecture of transformer models across different use case models.
The following figure doesn't include the 'shared' structure, but we can further update Qwen3CausalLM by modifying it to inherit both Qwen3 (which inherits Transformer) and CausalLM. This approach will also allow us to easily implement Qwen3-Embedding.

<img width="300" height="200" alt="Image" src="https://github.com/user-attachments/assets/61066842-1fc2-4c66-b6a3-5aeeac934776" />

EunjuYang:
<img width="300" height="200" alt="qwen3_transformer" src="https://github.com/user-attachments/assets/065af845-0039-49a1-8df6-19d0a54639a3" />

PR #3670, opened 2025-12-31
[CausalLM] Add a new class `Transformer` and update `CausalLM` to inherit the class.

EunjuYang:
## Dependency of the PR
None.

<details><summary>[CauslLM] Add a Transformer class</summary><br />

- Add a Transformer class as a parent class of CausalLM
- This commit aims to update CausalLM inherits the Transformer class

**Self evaluation:**

</details>

<details><summary>[CausalLM] update CausalLM to inherit Transformer</summary><br />

- This commit updates CausalLM class to inherit Transformer class
- This commit updates factory to use Transformer as a basis class type
- This commit adds Transformer.cpp to build file

**Self evaluation:**

</details>

### Summary

- Please see #3654 
- Image
- What this PR aims is to implement a new common parent class for Transformer-based architectures.
- In this PR, I implemented `Transformer` class and updates `CausalLM` to inherit the `Transformer` class.
- In this PR, I updated factory to use `Transformer` type instead of `CausalLM` type.
- I verified this PR on both Qwen3-4b and Qwen3-30b-a3b model.

### DIFF
--- a/Applications/CausalLM/factory.h
+++ b/Applications/CausalLM/factory.h
@@ -14,7 +14,7 @@
 #ifndef __CAUSALLM_FACTORY_H__
 #define __CAUSALLM_FACTORY_H__
 
-#include <causal_lm.h>
+#include <transformer.h>
 #include <unordered_map>
 
 namespace causallm {
@@ -25,7 +25,7 @@ namespace causallm {
 class Factory {
 public:
   using Creator =
-    std::function<std::unique_ptr<CausalLM>(json &, json &, json &)>;
+    std::function<std::unique_ptr<Transformer>(json &, json &, json &)>;
 
   static Factory &Instance() {
     static Factory factory;
@@ -36,8 +36,9 @@ public:
     creators[key] = creator;
   }
 
-  std::unique_ptr<CausalLM> create(const std::string &key, json &cfg,
-                                   json &generation_cfg, json &nntr_cfg) const {
+  std::unique_ptr<Transformer> create(const std::string &key, json &cfg,
+                                      json &generation_cfg,
+                                      json &nntr_cfg) const {
     auto it = creators.find(key);
     if (it != creators.end()) {
       return (it->second)(cfg, generation_cfg, nntr_cfg);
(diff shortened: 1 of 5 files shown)


baek2sm (approved):
LGTM
It seems like a good structure to expand with applications such as SentenceTransformer

Applications/CausalLM/models/transformer.cpp
```diff
@@ -0,0 +1,393 @@
+// SPDX-License-Identifier: Apache-2.0
```
junbong-yu, line 1: Considering that Transformer is used not only for Causal model but also for Bidirectional model (Encoder), I think it is worth considering to transfer this code to another directory someday in the future.

Applications/CausalLM/models/transformer.h
```diff
@@ -0,0 +1,193 @@
...
+    file >> data;
+    return data;
+  } catch (const json::parse_error &e) {
+    throw std::runtime_error("JSON parse error in " + file_path +
+                             " | Details: " + e.what());
+  }
+}
+} // namespace causallm
+
+#endif
```
Seunghui98, line 193:
```suggestion
#endif

```
how about this?? 😄

Applications/CausalLM/models/transformer.cpp
```diff
@@ -0,0 +1,393 @@
...
+    throw std::runtime_error(
+      "Transformer model is not initialized. Please call "
+      "initialize() before run().");
+  }
+  ///@note This part should be filled in.
+  /// The run action can be defined by the precedent classes.
+}
+
+std::vector<LayerHandle>
+Transformer::createTransformerDecoderBlock(const int layer_id,
```
jijoongmoon, line 228: It is not the DecoderBlock anymore, is it?

jijoongmoon (approved):
LGTM except minor.

Applications/CausalLM/models/transformer.cpp
EunjuYang: Thank you for your suggestion! I agree with you. I believe we can rename this application to better support various types of transformers and reorganize the directory structure.

Applications/CausalLM/models/transformer.cpp
EunjuYang: I referred to this model: [modeling_qwen3.py](https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/models/qwen3/modeling_qwen3.py#L366). I referred to this model. However, as you pointed out, it seems much better to update its name. I'll rename it to `createTransformerBlock()` instead of `createTransformerDecoderBlock()`. Does that make sense? :)

Issue #3674, opened 2025-12-31
Implement `Encoder` class which inherits `Transformer`


EunjuYang:
Encoder (or Embedding; Please suggest a good name for this feature) aims to mimic the action of `sentence-transformer`.
Sentence-transformer's model usually consists with multiple modules which are defined in `module.json`.
We may need to define the interface to load the Encoder model. Let me firstly start with very simple version, which is not fullly generalized.

PR #3678, opened 2025-12-31
[Transformer] add `model_type` to `nntr_cfg` and implement feature to validate the model_type matching

### LINKED ISSUE #3675: Add model type (CausalLM/Model/Embedding/Unknown ...)

EunjuYang:

<details><summary>[Transformer] add model_type to nntr_cfg</summary><br />

- This commit adds model_type to nntr_cfg.
- This commit defines model types supported in the transformer
- This commit validates the model type matches with the user's model

**Self evaluation:**

</details>

### Summary
See also #3674 
This PR resolves #3675 

- This PR adds the model type to nntr_cfg.
- Since the type of configuration files required varies depending on the model_type, we need to add a field to detect the model type and ensure it matches the corresponding model class.

### DIFF
--- a/Applications/CausalLM/models/transformer.cpp
+++ b/Applications/CausalLM/models/transformer.cpp
@@ -43,7 +43,35 @@ std::string LoadBytesFromFile(const std::string &path) {
   return buffer;
 }
 
-Transformer::Transformer(json &cfg, json &generation_cfg, json &nntr_cfg) {
+static ModelType strToModelType(const std::string &model_type) {
+
+  static const std::unordered_map<std::string, ModelType> model_type_map = {
+    {"Model", ModelType::MODEL},
+    {"CausalLM", ModelType::CAUSAL_LM},
+    {"Embedding", ModelType::EMBEDDING}};
+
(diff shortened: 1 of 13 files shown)


h0g1 (approved):
LGTM!

Applications/CausalLM/models/transformer.cpp
```diff
@@ -43,7 +43,35 @@ std::string LoadBytesFromFile(const std::string &path) {
   return buffer;
 }
 
-Transformer::Transformer(json &cfg, json &generation_cfg, json &nntr_cfg) {
+ModelType strToModelType(std::string model_type) {
```
songgot: How about using code below? Just point to the original string without copying it. (Fast!)
```
static ModelType strToModelType(const std::string& model_type)
```

EunjuYang: Thank you for the suggestion. It seems like a great idea to make it static and take the variable as a reference. By defining it as static, we can enforce the function to have internal linkage. Thank you.👍

songgot: Thank you :+1:

PR #3687, opened 2026-01-05
[ONNX][WIP] Loading dynamic ONNX model  

sachin-nntrainer:
This PR contains commit 82c05981cd953b7e0605dee381bdbf2e1fda198d

It enables loading dynamic onnx model as a static model into nntrainer. The user has to define a mapping of symbolic name to dim value. While loading in onnx interpreter, the symbolic name is replaced by its corresponding static value. 

This will enable loading a model with dynamic input dimension for multiple static inputs.

For example, consider a model having input defined as 1\*seq_len\*256. This one model can be loaded multiple times for various seq_len values.

Advantages in terms of LLM scenarios: 

1. Single .onnx model for both prefill and decode stage
2. Prefill input dim set number of tokens present in prompt rather than max context length (eg: 2048)

Instructions to test:
Model input setting and loading: /nntrainer/Applications/ONNX/jni/main.cpp. Different static values can be set here for a dynamic symbolic name.
After building execute /nntrainer/build/Applications/ONNX/jni/nntrainer_onnx_example

**Self evaluation:**
4. Run test: []Passed [ ]Failed [X]Skipped

Signed-off-by: Sachin Singh <sachin.3@samsung.com}


baek2sm:
please add doxygen tags to "Applications/ONNX/jni/fcModel/create_bin.py" like other python scripts :)

PR #3688, opened 2026-01-06
[CI FIX #3644] [layers] update weight layer + TFLite exporter is updated

EunjuYang:

<details><summary> [layers] update weight layer</summary><br />

 The current version of NNTrainer determines whether a layer is an input layer based on the p
resence of an input shape. If there is no input shape, the layer is treated as an input layer, w
hich causes a problem because weight layers also lack an input shape and are therefore mistakenl
y considered input layers.

 To address this, the commit introduces a weight_dim property that sets the dimension of weig
ht layers, allowing the system to distinguish between input layers and weight layers.

 However, this version is not the final set of changes. A separate PR incorporating the final
 modifications will be created in this week, and I will request a merge of that PR together with this one. The current version is just a working version commit before the SRIB commit.

** Things to be done:
 - Since a new property was added to layer_node, the current layer_impl cannot access it. I plan to improve this by parsing the property into a local variable during the finalize stage of layer_node so that it can be used at the layer_impl level.
 - Add unit tests for weight layer.

</details>

<details><summary> [TFList] node exporter is updated</summary><br />
- Since the layernode's property is updated, tflite exporter should be
  updated.
- This commit adds props::WeightDimension to safeTfResult so that the
  layer node can be successfully exported

**Self evaluation:**

</details>

### Summary

- Since we cannot wait for #3644 anymore, I added a CI fix commit on top of #3644.
- This PR includes an additional commit to resolve the unit test failure in the TFLite exporter.

### DIFF
--- a/nntrainer/utils/node_exporter.cpp
+++ b/nntrainer/utils/node_exporter.cpp
@@ -88,11 +88,12 @@ void Exporter::saveTflResult(const std::tuple<> &props,
 
 template <>
 void Exporter::saveTflResult(
-  const std::tuple<
-    props::Name, props::Distribute, props::Trainable,
-    std::vector<props::InputConnection>, std::vector<props::InputShape>,
-    props::SharedFrom, props::ClipGradByGlobalNorm, props::Packed,
-    props::WeightDtype, props::LossScaleForMixed, props::ComputeEngine> &props,
+  const std::tuple<props::Name, props::Distribute, props::Trainable,
+                   std::vector<props::InputConnection>,
+                   std::vector<props::InputShape>, props::WeightDimension,
+                   props::SharedFrom, props::ClipGradByGlobalNorm,
+                   props::Packed, props::WeightDtype, props::LossScaleForMixed,
+                   props::ComputeEngine> &props,
   const LayerNode *self) {
   createIfNull(tf_node);
   tf_node->setLayerNode(*self);
--- a/nntrainer/utils/node_exporter.h
+++ b/nntrainer/utils/node_exporter.h
@@ -220,6 +220,7 @@ class Unit;
 class Flatten;
 class Distribute;
 class Trainable;
+class WeightDimension;
 class InputShape;
 class WeightRegularizer;
 class WeightRegularizerConstant;
@@ -249,11 +250,12 @@ class LayerNode;
  */
 template <>
 void Exporter::saveTflResult(
-  const std::tuple<
-    props::Name, props::Distribute, props::Trainable,
-    std::vector<props::InputConnection>, std::vector<props::InputShape>,
-    props::SharedFrom, props::ClipGradByGlobalNorm, props::Packed,
-    props::WeightDtype, props::LossScaleForMixed, props::ComputeEngine> &props,
+  const std::tuple<props::Name, props::Distribute, props::Trainable,
+                   std::vector<props::InputConnection>,
+                   std::vector<props::InputShape>, props::WeightDimension,
+                   props::SharedFrom, props::ClipGradByGlobalNorm,
+                   props::Packed, props::WeightDtype, props::LossScaleForMixed,
+                   props::ComputeEngine> &props,
   const LayerNode *self);
 
 class BatchNormalizationLayer;


junbong-yu (approved):
LGTM

nntrainer/layers/layer_node.h
```diff
@@ -1046,9 +1053,10 @@ properties in the context/graph unless intended. */
   using PropsType =
     std::tuple<props::Name, props::Distribute, props::Trainable,
                std::vector<props::InputConnection>,
-               std::vector<props::InputShape>, props::SharedFrom,
-               props::ClipGradByGlobalNorm, props::Packed, props::WeightDtype,
-               props::LossScaleForMixed, props::ComputeEngine>;
+               std::vector<props::InputShape>, props::WeightDimension,
```
haehun, line 1056: Is it impossible for a layer to have multiple weights (like, weight and bias for linear layer)?

EunjuYang: Let me clarify your question first. Are you asking whether a general layer can have multiple weights?
If so, my answer is yes; it is possible for a layer to have multiple weights.

Second, I understand that the property name raises some questions, as you mentioned.
However, it appears that this property is designed specifically for "weight" layers.  (Please see [a6baa46](https://github.com/nntrainer/nntrainer/pull/3688/commits/a6baa467086643bf3c8388de9b737b60b88d3de6))
Regarding its design, one weight layer might take only one WeightDimension property.
As for its design, we may need to discuss its intention with @baek2sm.

baek2sm: @haehun Unlike operation layers, high-level layers such as fully connected (FC) layers can have multiple weights, but users do not directly specify the weight dimensions of these weights. (For example, in the case of an FC layer, when the user specifies the unit size, the weight dimensions are determined during the finalize stage based on this value.) On the other hand, weight layer represents the minimum unit of network graphs, so they contain only one weight, and the weight dimension property is intended for this purpose.

nntrainer/layers/layer_node.cpp
```diff
@@ -113,6 +113,17 @@ class InputShape : public GenericShape {
   using prop_tag = dimension_prop_tag;              /**< property type */
 };
 
+/**
+ * @brief Weight Dimension property which saves a single tensor dim
+ *
+ */
+class WeightDimension : public GenericShape {
```
jijoongmoon, line 120: This could go to the common_properties.h.

nntrainer/layers/layer_node.cpp
```diff
@@ -201,8 +212,9 @@ LayerNode::LayerNode(std::unique_ptr<nntrainer::Layer> &&l) :
   run_context(nullptr),
   layer_node_props(new PropsType(
     props::Name(), props::Distribute(), props::Trainable(), {}, {},
-    props::SharedFrom(), props::ClipGradByGlobalNorm(), props::Packed(),
-    props::WeightDtype(), props::LossScaleForMixed(), props::ComputeEngine())),
+    props::WeightDimension(), props::SharedFrom(),
```
jijoongmoon, line 215: I'm not sure WeightDimension is the layer_node property. Isn't it more like Weight Layer Property?

nntrainer/layers/layer_node.cpp
EunjuYang: @baek2sm Could you answer to this?

nntrainer/layers/layer_node.cpp
EunjuYang: @baek2sm Could you answer to this?

nntrainer/layers/layer_node.cpp
baek2sm: @jijoongmoon @EunjuYang 
I didn't see this comment. Initially, I attempted to add it to common_property, but there's a slightly complex issue. The layer_node layer determines based on whether hasInputShapeProperty() returns true; when `inputShapeProperty` is absent, it retrieves the `input_shape` property to assign the layer's dimension. The problem arises because the weight layer also lacks `InputShapeProperty` and requires dimension configuration at this stage. Since the accessible property here is `layer_node_props`, I added the `weight_dimension` property directly to the layer node property, like `input shape`.

nntrainer/layers/layer_node.cpp
baek2sm: @jijoongmoon @EunjuYang 
I didn't see this comment. Initially, I attempted to add it to common_property, but there's a slightly complex issue. The layer_node layer determines based on whether hasInputShapeProperty() returns true; when `inputShapeProperty` is absent, it retrieves the `input_shape` property to assign the layer's dimension. The problem arises because the weight layer also lacks `InputShapeProperty` and requires dimension configuration at this stage. Since the accessible property here is `layer_node_props`, I added the `weight_dimension` property directly to the layer node property, like `input shape`.

test/unittest/layers/meson.build
```diff
@@ -68,7 +68,6 @@ test_target = [
   'unittest_layers.cpp',
   'unittest_layers_impl.cpp',
   'unittest_layers_input.cpp',
-  'unittest_layers_weight.cpp',
```
Seunghui98, line 71: it's just out of curiousity.
Why do you remove this unit test..? 🤔

baek2sm: That unit test is not suitable for a weight layer because it doesn't have any inputs. The unit test for weight layer should be updated, too.

PR #3691, opened 2026-01-06
Reopen weight loading for ONNX model

sumon-98:
This version supports weight loading for ONNX models and is compatible only with models where each layer has a separate weight file. Weight bins are generated using a Python script for each weight file.

**Model and Weight Bin Generation:** To generate models and weight bins, use the following directory: `Applications/ONNX/jni/python_scripts`

**Self evaluation:**
2. Run test: [X]Passed [ ]Failed []Skipped

### DIFF
--- a/Applications/ONNX/jni/main.cpp
+++ b/Applications/ONNX/jni/main.cpp
@@ -20,21 +20,31 @@
 int main() {
   auto model = ml::train::createModel();
 
+  std::cout << "--------------------------------------Create Model "
+               "Done--------------------------------------"
+            << std::endl;
   try {
-    std::string path = "../../../../Applications/ONNX/jni/llama_example.onnx";
+    std::string path =
+      "../../../../Applications/ONNX/python/simple_model/model.onnx";
     model->load(path, ml::train::ModelFormat::MODEL_FORMAT_ONNX);
   } catch (const std::exception &e) {
     std::cerr << "Error during load: " << e.what() << "\n";
     return 1;
   }
 
+  std::cout << "--------------------------------------Load Model "
+               "Done--------------------------------------"
+            << std::endl;
   try {
-    model->compile();
+    model->compile(ml::train::ExecutionMode::INFERENCE);
   } catch (const std::exception &e) {
     std::cerr << "Error during compile: " << e.what() << "\n";
     return 1;
   }
 
+  std::cout << "--------------------------------------Compile Model "
+               "Done--------------------------------------"
+            << std::endl;
   try {
     model->initialize();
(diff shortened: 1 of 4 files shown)


Applications/ONNX/jni/main.cpp
```diff
@@ -20,29 +20,69 @@
...
+    return 1;
+  }
+
+  float *input1 = new float[4];
+  for (int i = 0; i < 4; i++) {
+    input1[i] = (i + 1);
+  }
+  std::vector<float *> in;
+  in.push_back(input1);
+  auto ans = model->inference(1, in);
```
Seunghui98, line 79: Could you explain what these lines are...? 🤔

nntrainer/models/neuralnet.cpp
```diff
@@ -745,12 +748,43 @@ void NeuralNetwork::load(const std::string &file_path,
...
+        threads.emplace_back([=]() {
+          if (!MMAP_READ) {
+            if (hasWeights) {
+              std::string file_path;
+              if (!base1.empty()) {
+                file_path = base1 + layerName + ".bin";
+              } else {
+                file_path = base0 + layerName + ".bin";
+              }
+              // std::cout << "Appended Layer: " << file_path << std::endl;
```
baek2sm: for debugging? please remove this line

nntrainer/tensor/tensor.h
```diff
@@ -520,7 +520,12 @@ class Tensor {
   template <typename T = float>
   static Tensor Map(T *buf, unsigned int bytes, const TensorDim &d,
                     size_t offset = 0) {
+    // commented out as weight layers are treated as inputs, need to bypass it
```
baek2sm: It seems inappropriate to bypass the bug detection part, we should fix the bug instead.. If there are still areas where weights are treated as inputs, it would be good to check and correct those parts.

Applications/ONNX/jni/main.cpp
sumon-98: These are some dummy values used as input to test if inference is working fine or not.

nntrainer/models/neuralnet.cpp
sumon-98: done

nntrainer/tensor/tensor.h
sumon-98: I have removed the additional conditional. I was not able to reproduce the bug. I believe it was fixed in the upstream.

baek2sm (approved):
LGTM. 
and It will be updated using single bin file.

PR #3686, opened 2026-01-05
[Wait for #3685] [Transformers/layer] add embedding_normalize_layer 

EunjuYang:
## Commits to be reviewed in this PR

<details><summary> [Transformers/layer] add embedding_normalize_layer  </summary><br />

- This commit adds a new layer for embedding model.
- This commit adds normalize_layer which is implemented in sentence_transformers/models/Normalize.py

Signed-off-by: Eunju Yang <ej.yang@samsung.com>
**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

</details>

### Summary

- This PR adds a new layer for embedding model.
- This PR adds normalize_layer which is implemented in sentence_transformers/models/Normalize.py
- see #3673 

Signed-off-by: Eunju Yang <ej.yang@samsung.com>

PR #3701, opened 2026-01-08
[causallm] add Qwen3Embedding

baek2sm:
- added Qwen3Embedding
- enable special token option of tokenizer for embedding model
- bug fix for embedding pooling 

**output comparison for token "h"**
- sentence_transformers: 0.00480586,  0.00295271, -0.01529285 ...
- nntrainer: 0.00478721, 0.00295418, -0.0152909 ...

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/main.cpp
+++ b/Applications/CausalLM/main.cpp
@@ -104,6 +104,20 @@ void stop_and_print_peak() {
             << std::endl;
 }
 
+std::string resolve_architecture(std::string model_type,
+                                 const std::string &architecture) {
+  std::transform(model_type.begin(), model_type.end(), model_type.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+
+  if (model_type == "embedding") {
+    if (architecture == "Qwen3ForCausalLM") {
+      return "Qwen3Embedding";
+    }
+  }
+
+  return architecture;
+}
+
 int main(int argc, char *argv[]) {
 
   /** Register all runnable causallm models to factory */
@@ -210,14 +224,7 @@ int main(int argc, char *argv[]) {
 
     if (nntr_cfg.contains("model_type")) {
       std::string model_type = nntr_cfg["model_type"].get<std::string>();
-      std::transform(model_type.begin(), model_type.end(), model_type.begin(),
-                     [](unsigned char c) { return std::tolower(c); });
-
-      if (model_type == "embedding") {
-        if (architecture == "Qwen3ForCausalLM") {
-          architecture = "Qwen3Embedding";
-        }
-      }
+      architecture = resolve_architecture(model_type, architecture);
     }
 
     auto model = causallm::Factory::Instance().create(architecture, cfg,
--- a/Applications/CausalLM/models/embedding.cpp
+++ b/Applications/CausalLM/models/embedding.cpp
@@ -194,7 +194,7 @@ std::vector<float *> Embedding::encode(const WSTR prompt,
 #if defined(_WIN32)
   std::wstring prompt_ = system_prompt + prompt + tail_prompt;
   std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
-  auto _input = tokenizer->Encode(converter.to_bytes(prompt_));
+  auto _input = tokenizer->Encode(converter.to_bytes(prompt_), true);
 #else
   std::string prompt_ = system_prompt + prompt + tail_prompt;
   auto _input = tokenizer->Encode(prompt_, true);
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -1170,18 +1170,17 @@ std::vector<float *> NeuralNetwork::incremental_inference(
     } else {
       last_out_buf_data = new float[batch_size * out_t.width()];
 
+      /**
+       * If the output tensor's height (sequence dimension) is 1, it
+       * indicates a reduced/pooled output (e.g., Pooling,
+       * EmbeddingNormalize). In this case, the output for any 'step' is
+       * located at index 0. Otherwise, we access the specific step index
+       * corresponding to the last processed token.
+       */
+      unsigned int effective_step = (out_t.getDim().height() == 1) ? 0 : step;
       for (unsigned int batch = 0; batch < batch_size; ++batch) {
         if (out->getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
-          /**
-           * If the output tensor's height (sequence dimension) is 1, it
-           * indicates a reduced/pooled output (e.g., Pooling,
-           * EmbeddingNormalize). In this case, the output for any 'step' is
-           * located at index 0. Otherwise, we access the specific step index
-           * corresponding to the last processed token.
-           */
-          unsigned int effective_step =
-            (out_t.getDim().height() == 1) ? 0 : step;
 
           const _FP16 *out_t_batch_ptr =
             out_t.getData<_FP16>() + batch * out_t.getDim().getFeatureLen() +
@@ -1193,15 +1192,6 @@ std::vector<float *> NeuralNetwork::incremental_inference(
           throw std::invalid_argument("Error: enable-fp16 is not set");
 #endif
         } else if (out->getDataType() == ml::train::TensorDim::DataType::FP32) {
-          /**
-           * If the output tensor's height (sequence dimension) is 1, it
-           * indicates a reduced/pooled output (e.g., Pooling,
-           * EmbeddingNormalize). In this case, the output for any 'step' is
-           * located at index 0. Otherwise, we access the specific step index
-           * corresponding to the last processed token.
-           */
-          unsigned int effective_step =
-            (out_t.getDim().height() == 1) ? 0 : step;
 
           const float *out_t_batch_ptr =
             out_t.getData() + batch * out_t.getDim().getFeatureLen() +


Applications/CausalLM/main.cpp
```diff
@@ -199,21 +205,41 @@ int main(int argc, char *argv[]) {
...
+                                                      generation_cfg, nntr_cfg);
     model->initialize();
     model->load_weight(weight_file);
 
+    bool do_sample = false;
+    if (generation_cfg.contains("do_sample")) {
+      if (!generation_cfg["do_sample"].is_null()) {
+        do_sample = generation_cfg["do_sample"].get<bool>();
+      }
+    }
```
songgot: You can change it with the following code.
```
bool do_sample = generation_cfg.value("do_sample", false);
```

Applications/CausalLM/models/embedding.cpp
```diff
@@ -197,14 +197,13 @@ std::vector<float *> Embedding::encode(const WSTR prompt,
...
 
   std::vector<int64_t> init_input;
-  unsigned int _len = _input.size();
-  unsigned int input_len = _len;
+  unsigned int input_len = _input.size();
 
-  if (_len > MAX_SEQ_LEN) {
+  if (input_len > MAX_SEQ_LEN) {
     input_len = MAX_SEQ_LEN;
   }
```
songgot: The code can be changed as follows.
```
unsigned int input_len = std::min((unsigned int)_input.size(), (unsigned int)MAX_SEQ_LEN);
```

Applications/CausalLM/main.cpp
baek2sm: thank you, I applied it!

Applications/CausalLM/models/embedding.cpp
```diff
@@ -23,21 +27,17 @@ Embedding::Embedding(json &cfg, json &generation_cfg, json &nntr_cfg) :
...
+  {"Pooling", "embedding_pooling"}, {"Normalize", "embedding_normalize"}};
 
 void Embedding::setupParameters(json &cfg, json &generation_cfg,
                                 json &nntr_cfg) {
   Transformer::setupParameters(cfg, generation_cfg, nntr_cfg);
 
   std::string modules_config_path = "modules.json";
   if (nntr_cfg.contains("module_config_path")) {
-    modules_config_path = nntr_cfg["modules_config_path"].get<std::string>();
+    modules_config_path = nntr_cfg["module_config_path"].get<std::string>();
```
junbong-yu, line 38: +1

junbong-yu (approved):
LGTM!!

Applications/CausalLM/main.cpp
```diff
@@ -199,21 +205,36 @@ int main(int argc, char *argv[]) {
...
+
+    if (nntr_cfg.contains("model_type")) {
+      std::string model_type = nntr_cfg["model_type"].get<std::string>();
+      std::transform(model_type.begin(), model_type.end(), model_type.begin(),
+                     [](unsigned char c) { return std::tolower(c); });
+
+      if (model_type == "embedding") {
+        if (architecture == "Qwen3ForCausalLM") {
+          architecture = "Qwen3Embedding";
+        }
```
Seunghui98: (optional)
How about introducing converting function for this? 😄
Considering that we support new embedding models more, it would be better to extract function separately for code management.

Applications/CausalLM/models/embedding.cpp
Seunghui98: 👍

Applications/CausalLM/tokenizers_cpp.h
```diff
@@ -12,27 +14,36 @@
...
-   * \param texts The input texts.
-   * \returns The encoded token ids.
+  /**
+   * @brief Encode text into ids with special tokens option.
+   * @param text The input text.
+   * @param add_special_tokens Whether to add special tokens.
+   * @return The encoded token ids.
+   */
+  virtual std::vector<int32_t> Encode(const std::string &text,
+                                      bool add_special_tokens) = 0;
```
Seunghui98, line 41: May I ask a question?
why do you introduce `add_special_token` check??? 🤔

Applications/CausalLM/main.cpp
baek2sm: Good idea. It can be applied in a separate PR. Thanks!

Applications/CausalLM/tokenizers_cpp.h
baek2sm: That's because this model tokenizes special tokens (such as eos) together and uses it!

Seunghui98: Oh, I understood it!! Thanks for letting me know 👍

Applications/CausalLM/main.cpp
EunjuYang: How about defining a naming rule for the embedding model? For example, appending a prefix to the original architecture. By doing this, we can eliminate the need for complicated if-else clauses.

Applications/CausalLM/models/embedding.cpp
```diff
@@ -197,16 +197,12 @@ std::vector<float *> Embedding::encode(const WSTR prompt,
   auto _input = tokenizer->Encode(converter.to_bytes(prompt_));
```
EunjuYang: Shouldn't we pass `true` for Windows?
```suggestion
  auto _input = tokenizer->Encode(converter.to_bytes(prompt_, true));
```

nntrainer/models/neuralnet.cpp
```diff
@@ -1173,19 +1173,39 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
 #ifdef ENABLE_FP16
+          /**
+           * If the output tensor's height (sequence dimension) is 1, it
+           * indicates a reduced/pooled output (e.g., Pooling,
+           * EmbeddingNormalize). In this case, the output for any 'step' is
+           * located at index 0. Otherwise, we access the specific step index
+           * corresponding to the last processed token.
+           */
+          unsigned int effective_step =
+            (out_t.getDim().height() == 1) ? 0 : step;
```
EunjuYang: What about moving these lines of code outside the loop? (Before line 1173) This would help reduce code duplication.

Applications/CausalLM/main.cpp
```diff
@@ -199,21 +205,36 @@ int main(int argc, char *argv[]) {
...
-      generation_cfg, nntr_cfg);
+    std::string architecture =
+      cfg["architectures"].get<std::vector<std::string>>()[0];
+
+    if (nntr_cfg.contains("model_type")) {
+      std::string model_type = nntr_cfg["model_type"].get<std::string>();
+      std::transform(model_type.begin(), model_type.end(), model_type.begin(),
+                     [](unsigned char c) { return std::tolower(c); });
+
+      if (model_type == "embedding") {
```
EunjuYang: We can reuse `strToModelType()` in transformer.cpp by declaring the function in header.

Applications/CausalLM/main.cpp
baek2sm: @EunjuYang Architecture name is not defined in our nntr_config(it's from the config file from the Hugging Face), so, do you want to add name of architecture to nntr_config.json? If then, I think we don't need to add prefix, rather then, how about adding the full name(like QwenEmbedding) to nntr_config?

Applications/CausalLM/main.cpp
baek2sm: Since the `model type (model, causallm, embedding)` and `architecture(Qwen3, Gemma3, etc)` are separated in the config, it seems better to define a new function rather than reuse this one.

nntrainer/models/neuralnet.cpp
baek2sm: that's good idea :)

Applications/CausalLM/main.cpp
baek2sm: I've made `resolve_architecture` function for now, and we can refactor this function after discussion

Applications/CausalLM/models/embedding.cpp
baek2sm: thanks, i've added it for windows, too!

Applications/CausalLM/main.cpp
EunjuYang: Oh, what I intended was the name specified in config.json. Since the architecture name is already provided, we could define the Qwen3Embedding model's architecture name as the prefix plus the architecture name (in config.json). This was just one of my suggestions. :) Let's discuss this further with the team during our offline meeting.

baek2sm: In my opinion, directly modifying the config.json file does not seem like a good approach. if we do so, users would be required to customize both `nntr_config` and `config.json` to match our format. It would be better to keep the `config.json` file aligned with the original version distributed by HuggingFace, and place all customization-related information in `nntr_config` for simple usage of our framework.

Applications/CausalLM/models/embedding.cpp
```diff
@@ -194,19 +194,15 @@ std::vector<float *> Embedding::encode(const WSTR prompt,
 #if defined(_WIN32)
   std::wstring prompt_ = system_prompt + prompt + tail_prompt;
   std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
-  auto _input = tokenizer->Encode(converter.to_bytes(prompt_));
+  auto _input = tokenizer->Encode(converter.to_bytes(prompt_)), true;
```
jijoongmoon: ```Encode(converter.to_bytes(prompt_), true)??```

Applications/CausalLM/layers/meson.build
```diff
@@ -9,7 +9,7 @@ causallm_embedding_src_abs = [meson.current_source_dir() / 'embedding_layer.cpp'
 causallm_reshaped_rms_norm_src_abs = [meson.current_source_dir() / 'reshaped_rms_norm.cpp']
 causallm_qkv_layer_src_abs = [meson.current_source_dir() / 'qkv_layer.cpp']
 causallm_embedding_pooling_src_abs = [meson.current_source_dir() / 'embedding_pooling_layer.cpp']
-
+causallm_embedding_normalize_src_abs = [meson.current_source_dir() / 'embedding_normalize_layer.cpp']
```
jijoongmoon, line 12: missing embedding_normalize_layer.cpp?? I cannot find this file in this PR.

Applications/CausalLM/models/embedding.cpp
baek2sm: thanks, fixed it!

Applications/CausalLM/layers/meson.build
baek2sm: That file was added in PR #3686, but meson wasn't updated.

jijoongmoon (approved):
LGTM

PR #3720, opened 2026-01-16
Causallm/update lmhead out dim

EunjuYang:
This draft changes tensor allocate size from 505069568 → 146800640 for our 1B benchmark model

AS-IS
```
01-16 17:16:17.755 11853 11853 I nntrainer: MemoryPool::allocate size: 666395280
01-16 17:16:17.887 11853 11853 I nntrainer: MemoryPool::allocate size: 505069568
```

TO-BE
```
01-16 17:42:00.912 17583 17583 I nntrainer: MemoryPool::allocate size: 666395280
01-16 17:42:01.042 17583 17583 I nntrainer: MemoryPool::allocate size: 146800640
```

WTD)
- In current, lm_head layer is created with FC layer, which contains redundant computation as a lmhead layer.
- [x] create lm_head layer which follows the finalize and incremental forwarding operation of tie_word_embedding

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -13,8 +13,11 @@
  */
 #include <algorithm>
 #include <cmath>
+#include <mutex>
 #include <thread>
 #include <vector>
+
+static std::mutex rope_init_mtx;
 
 #include <engine.h>
 #include <fp16.h>
@@ -82,7 +85,7 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
     std::get<nntrainer::props::MaxTimestep>(mha_core_props).get();
 
   /** max position embeddings */
-  unsigned int max_position_embeddings =
+  max_position_embeddings =
     std::get<props::MaxPositionEmbeddings>(mha_core_props).get();
 
   /** local window size */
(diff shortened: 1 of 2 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -1159,7 +1159,9 @@ std::vector<float *> NeuralNetwork::incremental_inference(
   // auto end_increment = std::chrono::high_resolution_clock::now();
   std::vector<float *> output;
 
-  unsigned int step = ((to - from) == 0) ? 0 : (to - from) - 1;
+  ///@note Always we take the first position of output
+  // unsigned int step = ((to - from) == 0) ? 0 : (to - from) - 1;
+  unsigned int step = 0;
```
EunjuYang, line 1164: I think we have to make an option to decide whether returning sequence or not :)
It is not generalized so it cannot support embedding's action. This should be handled before  merging this code.

jijoongmoon: It shouldn't always start from 0. If it is, it is hard to support the multi-turn scenario.

EunjuYang: It seems that the "step" is only used to obtain the output value.
Therefore, what I intended was to always take the output value from the first place, regardless of the sequence length.

EunjuYang:
###  1) Update in tie-word-embedding layer and neuralnet.cpp
image

### 2) Update in lmhead layer and neuralnet.cpp
image

nntrainer/models/neuralnet.cpp
baek2sm: I agree with @EunjuYang. The layers can still support multi-turn interactions throught the `from` and `to` variables. `step` is just used for the position of output values.

EunjuYang:
Let me reopen this PR as a separate PR which does not depend on attention-related PRs.
This feature alone should be merged to main ASAP 😊

EunjuYang:
This draft can be replaced with #3737. I will close this to avoid confusion.

PR #3730, opened 2026-01-21
[ tensor ] add a new tensor class to support QINT4 on ARM CPU

h0g1:
This PR intorduces a new tensor class (temporarily named Kai4Tensor) to support QINT4 on ARM CPU.
There are several notable points in this PR like the following:

First, this PR reflects packing style of each micro-kernel from KleidiAI to allocate memory in Kai4Tensor.

Second, this PR adds a member variable called '_idx_variant' in Kai4Tensor to choose micro-kernel.

Third, this PR modifies a dot funciton of float tensor to call micro-kernels from KleidiAI interface on ARM CPU with fp16 enabled.

Fourth, this PR adds a unittest for tensor dot operation on Q40 and QINT4 type on CPU.

TODO:

- Extend the availability of Kai4Tensor to other micro-kernels (currently only support for block quantization with block size = 32)

- Add more test cases for ggml vs kai and more variant kernels (currently, only support for correctness of dot function and kernel index is fixed)

Signed-off-by: h0g1 <h0g1.hong@samsung.com>


test/unittest/unittest_kai4_comparison.cpp
jaemini-shin, line 1: Looks like an empty file

nntrainer/tensor/tensor.cpp
```diff
@@ -423,6 +451,12 @@ bool Tensor::operator==(const Tensor &rhs) const {
     } else if (getDataType() == Tdatatype::QINT8) {
       return itensorCompare<CharTensor>(itensor_.get(), rhs.itensor_.get());
     } else if (getDataType() == Tdatatype::QINT4) {
+#if defined(ENABLE_FP16) && defined(__aarch64__)
+      return itensorCompare<Kai4Tensor>(itensor_.get(), rhs.itensor_.get());
+#else
+      return itensorCompare<Int4QTensor>(itensor_.get(), rhs.itensor_.get());
+#endif
+    } else if (getDataType() == Tdatatype::BCQ) {      
       return itensorCompare<Int4QTensor>(itensor_.get(), rhs.itensor_.get());
```
jaemini-shin, line 462:
```suggestion
#else
      return itensorCompare<Int4QTensor>(itensor_.get(), rhs.itensor_.get());
#endif
```

PR #3732, opened 2026-01-21
[WIP][cpu_backend] Optimize TanhGeLU activation using NEON intrinsics

baek2sm:
This commit implements NEON optimization for the TanhGeLU activation function
to improve performance on ARM devices.

It is tested with gelu activation function of Gemma3(270m), and the result is as below
- before: 7.6 ms
- after(neon): 1.3 ms

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
@@ -965,6 +965,85 @@ void swiglu(const unsigned int N, float *X, float *Y, float *Z, float alpha) {
   }
 }
 
+void tanh_gelu(const unsigned int N, const float *X, float *Y) {
+  unsigned int i = 0;
+  float32x4_t half = vdupq_n_f32(0.5f);
+  float32x4_t one = vdupq_n_f32(1.0f);
+  float32x4_t const_0_044715 = vdupq_n_f32(0.044715f);
+  float32x4_t sqrt_2_over_pi = vdupq_n_f32(0.7978845608f);
+
+  // 4x loop unrolling: process 16 elements
+  for (; N - i >= 16; i += 16) {
+    float32x4_t x0 = vld1q_f32(&X[i]);
+    float32x4_t x1 = vld1q_f32(&X[i + 4]);
+    float32x4_t x2 = vld1q_f32(&X[i + 8]);
+    float32x4_t x3 = vld1q_f32(&X[i + 12]);
+
+    float32x4_t cubed0 = vmulq_f32(x0, vmulq_f32(x0, x0));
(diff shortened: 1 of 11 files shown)


nntrainer/tensor/cpu_backend/arm/neon_mathfun.hxx
```diff
@@ -332,10 +336,29 @@ v4sf sin_ps(v4sf x) {
...
+/**
+ * @brief tanh function with simd
+ *
+ * @param x input register variable
+ * @return v4sf
+ */
+inline v4sf tanh_ps(v4sf x) {
+  v4sf one = vdupq_n_f32(1.0f);
+  v4sf two = vdupq_n_f32(2.0f);
+  v4sf minus_one = vdupq_n_f32(-1.0f);
```
h0g1, line 354: What is the variable 'minus_one' for? Seems like it is defined but not used.

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
```diff
@@ -965,6 +965,85 @@ void swiglu(const unsigned int N, float *X, float *Y, float *Z, float alpha) {
...
+
+    inner = vmulq_f32(sqrt_2_over_pi, inner);
+    float32x4_t tanh_res = tanh_ps(inner);
+
+    float32x4_t res = vaddq_f32(one, tanh_res);
+    res = vmulq_f32(res, x);
+    res = vmulq_f32(res, half);
+
+    vst1q_f32(&Y[i], res);
+  }
```
h0g1, line 1037: IMO, GELU computation can be more efficient by subsituting tanh(x) with its explicit expression.

When $y = \frac{2}{\pi}(x + 0.044715x^3)$,
GELU(x) = $0.5x(1 + tanh(y))$ = $0.5x(1 + \frac{e^{y} - e^{-y}}{e^{y} + e^{-y}})$ = $0.5x\frac{2e^{y}}{e^{y} + e^{-y}}$ = $\frac{x}{1 + e^{-2y}}$ = $\frac{x}{1 + e^{-\frac{4}{\pi}(x + 0.044715x^3)}}$.

How about directly implementing GELU() explicitly without calling tanh?

nntrainer/tensor/cpu_backend/arm/neon_mathfun.hxx
songgot, line 354: ```minus_one``` is not used in actual calculation logic.

h0g1: We raised the same concern!

nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
baek2sm: Good idea, I'll try to apply this formula!

PR #3733, opened 2026-01-27
[util] add image loader

baek2sm:
- added image loading function for png image to llm_util.cpp
(I'll rename that file after CausalLM related works(SLM, API, etc..) are merged)
- It uses stb_image library (MIT license)

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/llm_util.cpp
+++ b/Applications/CausalLM/llm_util.cpp
@@ -11,6 +11,8 @@
  * @bug    No known bugs except for NYI items
  */
 
+#define STB_IMAGE_IMPLEMENTATION
+#include "stb_image.h"
 #include <llm_util.hpp>
 
 std::vector<unsigned int> generate_multi_tokens(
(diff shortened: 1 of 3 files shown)


Applications/CausalLM/llm_util.hpp
```diff
@@ -18,10 +18,15 @@
 #include <math.h>    // INFINITY
 #include <optional>
 
+#include "stb_image.h"
```
songgot: Any file that includes llm_util.hpp will be forced to include the stb_image library.
Since there are no stb-related types in the function declarations in llm_util.hpp, the include must be moved to the llm_util.cpp file.

Applications/CausalLM/llm_util.cpp
```diff
@@ -88,3 +89,131 @@ float applyTKP(float *logits, int len, float temperature, unsigned int top_k,
 
   return top_indices_and_logits[0].second;
 }
+
+/**
+ * @brief Resize image using bilinear interpolation
+ */
+std::vector<unsigned char> resizeImage(const unsigned char *src, int src_w,
```
songgot, line 97: How about using 'static'?

Applications/CausalLM/llm_util.hpp
baek2sm: I've updated it, thanks

PR #3736, opened 2026-01-29
[Draft: Multi-Input]: support for multi-input

junbong-yu:
- How to use:
  - 1. sentences should be std::vector<std::string> type
  - 2. input sentences are treated as batched input, so you need to fix batch_size value in nntr_config.json

{commit message}

**Self evaluation:**

</details>

<details><summary>{commit-2}</summary><br />

{commit message}

**Self evaluation:**

</details>

### Summary

- {Summary of PR 1}
- {Summary of PR 2}

### DIFF
--- a/Applications/CausalLM/models/embedding.cpp
+++ b/Applications/CausalLM/models/embedding.cpp
@@ -202,6 +202,32 @@ void Embedding::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   }
 }
 
+void Embedding::run(const std::vector<WSTR> &prompts, bool do_sample,
+                    const std::vector<WSTR> system_prompts,
+                    const std::vector<WSTR> tail_prompts) {
+  try {
+    std::vector<float *> results =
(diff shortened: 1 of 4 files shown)


Applications/CausalLM/models/embedding.cpp
```diff
@@ -202,6 +202,32 @@ void Embedding::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   }
 }
 
+void Embedding::run(const std::vector<WSTR> &prompts, bool do_sample,
+                    const std::vector<WSTR> system_prompts,
+                    const std::vector<WSTR> tail_prompts) {
```
baek2sm, line 207: Then, now there are two run function (the existing single-batch version and the new multi-batch version), so.. how about removing the batch-for loop in the existing single-batch run function(because there is batch-for loop but it doesn't work)?

junbong-yu: Good Idea, batch size of single-batched run function is always 1, which means batch-for loop isn't necessary.

PR #3737, opened 2026-01-30
[CausalLM] Update LMHead related reatures and add LMHead layer

EunjuYang:
## Dependency of the PR
This is a reopened PR (#3720) without dependency on other PRs.

<details><summary>[CausalLM] enforce lmhead's output height 1 </summary><br />

- This commit updates neuralnet.cpp to make it always return sequence
  length = 1 for incremenatal_forwarding
- This commit updates tie_word_embedding to finalize its output
  dimension for lmhead tensor (height = 1)

</details>

<details><summary>[CausalLM] lm_head layer is added</summary><br />

- In the previous code, we used fc_layer for lm_head, which may degrade
  the prefill speed by calculating all of the sequence even we only need
  the last element.
- This commit adds lm_head layer which computes lmhead action
  efficiently

</details>

<details><summary>[CausalLM] Optimize RoPE memory usage with thread-safe lazy allocation</summary><br />
This commit optimizes memory usage in MHACoreLayer by implementing lazy
allocation for RoPE frequency tables.

Key changes:
- Implemented Lazy Allocation: RoPE table (FP32/FP16) are now allocated
  on-demand during the first inference step instead of being
pre-allocated in finalize. This prevents unused tables from consuming
memory (saving ~50MB).
- Added Thread Safety: Utilized Doubled-Checked Locking with
  `std::mutex` to ensure safe one-time allocation in multi-thread
environments without performance overhead.
- Fixed Bug: Resolved an issue where `max_position_embeddings` was
  shadowed by a local variable in finalize, ensuring correct table
sizing.

</details>

<details><summary>[meson] add lm_head_dep to meson</summary><br />

- This commit adds lmhead_dep to meason (layer dependency)

</details>

### Summary

- In the existing code, during incremental forwarding, the output shape of the tie_word_embedding layer always required taking outputs from the 'step'. This led to a structure where outputs were fetched from the step.
- However, through analysis, we found that by modifying the process so that for all from-to cases in incremental forwarding, results are fetched starting from offset=0 of the output tensor, we can reduce the memory usage of the output tensor.
- Therefore, we modified the tie_word_embedding layer, which previously reserved a large-sized output tensor, and added an lm_head layer to ensure that outputs are always written starting from offset 0 under any circumstances.
- We also updated neuralnet.cpp to support unified behavior for the modified layer (all other layers already write their output values starting from offset 0).

####  1) Update in tie-word-embedding layer and neuralnet.cpp
image

#### 2) Update in lmhead layer and neuralnet.cpp
image

Signed-off-by: EunjuYang <ej.yang@samsung.com>

### DIFF
--- a/Applications/CausalLM/layers/lm_head.cpp
+++ b/Applications/CausalLM/layers/lm_head.cpp
@@ -66,7 +66,10 @@ void LmHeadLayer::finalize(nntrainer::InitLayerContext &context) {
   ///@note lm_head's output dimension (height is always 1 !)
   auto const &in_dim = context.getInputDimensions()[0];
   output_dims[0] = in_dim;
-  is_nchw ? output_dims[0].width(unit) : output_dims[0].channel(unit);
+  if (is_nchw)
+    output_dims[0].width(unit);
+  else
+    output_dims[0].channel(unit);
   output_dims[0].height(1);
 
   output_dims[0].setTensorType(
@@ -108,7 +111,10 @@ void LmHeadLayer::setProperty(const std::vector<std::string> &values) {
 }
 
 void LmHeadLayer::forwarding(nntrainer::RunLayerContext &context,
-                             bool training) {}
+                             bool training) {
+  throw nntrainer::exception::not_supported(
+    "Forwarding for LMHead layer is not supported");
+}
 
 void LmHeadLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
                                          unsigned int from, unsigned int to,
@@ -155,10 +161,13 @@ void LmHeadLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
 
 void LmHeadLayer::calcDerivative(nntrainer::RunLayerContext &context) {
   throw nntrainer::exception::not_supported(
-    "calcDerivative for Embedding layer is not supported");
+    "calcDerivative for LMHead layer is not supported");
 }
 
-void LmHeadLayer::calcGradient(nntrainer::RunLayerContext &context) {}
+void LmHeadLayer::calcGradient(nntrainer::RunLayerContext &context) {
+  throw nntrainer::exception::not_supported(
+    "calcGradient for LMHead layer is not supported");
+}
 
 void LmHeadLayer::exportTo(nntrainer::Exporter &exporter,
                            const ml::train::ExportMethods &method) const {
--- a/Applications/CausalLM/layers/lm_head.h
+++ b/Applications/CausalLM/layers/lm_head.h
@@ -2,7 +2,7 @@
 /**
  * Copyright (C) 2026 Eunju Yang <ej.yang@samsung.com>
  *
- * @file   lmhead_layer.h
+ * @file   lm_head.h
  * @date   16 Jan 2026
  * @brief  This is LM_Head Layer Class of Neural Network
  * @see    https://github.com/nntrainer/nntrainer
@@ -11,8 +11,8 @@
  *
  */
 
-#ifndef __LMHEAD_LAYER_H__
-#define __LMHEAD_LAYER_H__
+#ifndef __LM_HEAD_H__
+#define __LM_HEAD_H__
 #ifdef __cplusplus
 
 #pragma once
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -640,10 +640,11 @@ void MHACoreLayer::one_batch_incremental_forwarding(
 void MHACoreLayer::precompute_freqs(int head_dim, unsigned int seq_len,
                                     float theta, bool is_fp16) {
   // compute the freqs only when it is the first time to call this function
-  if (!is_fp16 && freqs_cos != nullptr && freqs_cos->size() == seq_len)
-    return;
 #ifdef ENABLE_FP16
-  if (is_fp16 && freqs_cos_fp16 != nullptr && freqs_cos_fp16->size() == seq_len)
+  if (freqs_cos_fp16 != nullptr && freqs_cos_fp16->size() == seq_len)
+    return;
+#else
+  if (freqs_cos != nullptr && freqs_cos->size() == seq_len)
     return;
 #endif
 


Applications/CausalLM/layers/lm_head.h
```diff
@@ -0,0 +1,129 @@
...
+  inline static const std::string type = "lm_head";
+
+private:
+  std::tuple<nntrainer::props::Unit> lmhead_props;
+  std::array<unsigned int, 2> weight_idx; /**< indices of the weights */
+};
+} // namespace causallm
+
+#endif
+#endif
```
Seunghui98: Could you add empty end line?? 🙂

Applications/CausalLM/layers/lm_head.h
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2026 Eunju Yang <ej.yang@samsung.com>
+ *
+ * @file   lmhead_layer.h
```
jaemini-shin:
```suggestion
 * @file   lm_head.h
```
Loos like a typo

Applications/CausalLM/layers/lm_head.h
```diff
@@ -0,0 +1,129 @@
...
+ * @date   16 Jan 2026
+ * @brief  This is LM_Head Layer Class of Neural Network
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author Eunju Yang <ej.yang@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ */
+
+#ifndef __LMHEAD_LAYER_H__
+#define __LMHEAD_LAYER_H__
```
jaemini-shin:
```suggestion
#ifndef __LM_HEAD_H__
#define __LM_HEAD_H__
```
These symbols should be changed, too

nntrainer/models/neuralnet.cpp
```diff
@@ -1159,7 +1159,9 @@ std::vector<float *> NeuralNetwork::incremental_inference(
   // auto end_increment = std::chrono::high_resolution_clock::now();
   std::vector<float *> output;
 
-  unsigned int step = ((to - from) == 0) ? 0 : (to - from) - 1;
+  ///@note Always we take the first position of output
+  // unsigned int step = ((to - from) == 0) ? 0 : (to - from) - 1;
+  unsigned int step = 0;
```
junbong-yu, line 1164: If step is 0, how about removing 'step * out_t.width()'?
or, Is there any purpose on preserving obsolete code?

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -639,62 +638,93 @@ void MHACoreLayer::one_batch_incremental_forwarding(
  * @note seq_len -> max_position_embeddings
  */
 void MHACoreLayer::precompute_freqs(int head_dim, unsigned int seq_len,
-                                    float theta) {
+                                    float theta, bool is_fp16) {
   // compute the freqs only when it is the first time to call this function
-  if (freqs_cos != nullptr && freqs_cos->size() == seq_len)
+  if (!is_fp16 && freqs_cos != nullptr && freqs_cos->size() == seq_len)
+    return;
+#ifdef ENABLE_FP16
```
Seunghui98, line 643: it's just out of curiousity. 🤔
isn't this line unnecessary due to the check of `is_fp16`?

nntrainer/models/neuralnet.cpp
```diff
@@ -1170,21 +1172,13 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
-       */
-      unsigned int effective_step = (out_t.getDim().height() == 1) ? 0 : step;
       for (unsigned int batch = 0; batch < batch_size; ++batch) {
         if (out->getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
 
           const _FP16 *out_t_batch_ptr =
             out_t.getData<_FP16>() + batch * out_t.getDim().getFeatureLen() +
-            effective_step * out_t.width();
+            step * out_t.width();
```
junbong-yu, line 1181: step * out_t.width() is always 0 if step is 0

nntrainer/models/neuralnet.cpp
```diff
@@ -1195,7 +1189,7 @@ std::vector<float *> NeuralNetwork::incremental_inference(
 
           const float *out_t_batch_ptr =
             out_t.getData() + batch * out_t.getDim().getFeatureLen() +
-            effective_step * out_t.width();
+            step * out_t.width();
```
junbong-yu, line 1192: step * out_t.width() is always 0 if step is 0

junbong-yu (approved):
Nice Idea, I look forward this patch to save memory!!

Applications/CausalLM/layers/lm_head.cpp
```diff
@@ -0,0 +1,201 @@
...
+  /// EffDimFlag shouldn't be fixed like this.
+  context.setEffDimFlagInputDimension(0, 0b1001);
+  context.setDynDimFlagInputDimension(0, 0b1000);
+  bool is_nchw = (context.getFormat() == nntrainer::Tformat::NCHW);
+
+  /** set output dimensions */
+  ///@note lm_head's output dimension (height is always 1 !)
+  auto const &in_dim = context.getInputDimensions()[0];
+  output_dims[0] = in_dim;
+  is_nchw ? output_dims[0].width(unit) : output_dims[0].channel(unit);
```
jaemini-shin:
```suggestion
  if(is_nchw)
    output_dims[0].width(unit);
  else
    output_dims[0].channel(unit);
```
How about this style? Using ternary operator without returning value might be confusing.

Applications/CausalLM/layers/lm_head.cpp
```diff
@@ -0,0 +1,201 @@
...
+  }
+}
+
+void LmHeadLayer::setProperty(const std::vector<std::string> &values) {
+  auto remain_props = loadProperties(values, lmhead_props);
+  LayerImpl::setProperty(remain_props);
+}
+
+void LmHeadLayer::forwarding(nntrainer::RunLayerContext &context,
+                             bool training) {}
```
jaemini-shin: It'd be better to throw exception if it's unsupported method.

Applications/CausalLM/layers/lm_head.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+}
+
+void LmHeadLayer::calcDerivative(nntrainer::RunLayerContext &context) {
+  throw nntrainer::exception::not_supported(
+    "calcDerivative for Embedding layer is not supported");
+}
+
+void LmHeadLayer::calcGradient(nntrainer::RunLayerContext &context) {}
```
jaemini-shin: It'd be better to throw exception if it's unsupported method.

nntrainer/models/neuralnet.cpp
EunjuYang: Yes, you are correct. We can simply remove the step, but I didn't remove it for future logging and discussion purposes.
We may need agreement to remove this 'step' feature from the code.

Applications/CausalLM/layers/lm_head.h
EunjuYang: Thank you for your comment :) I applied the comment.

Applications/CausalLM/layers/lm_head.h
EunjuYang: Thank you for your comment :) I applied the comment.

Applications/CausalLM/layers/mha_core.cpp
EunjuYang: I updated the code. Thank you for the comment :)

Applications/CausalLM/layers/lm_head.cpp
EunjuYang: Thank you for your comment :) I applied the comment.

Applications/CausalLM/layers/lm_head.cpp
EunjuYang: Thank you for your comment :) I updated the code.

Applications/CausalLM/layers/tie_word_embedding.cpp
```diff
@@ -274,7 +275,6 @@ void TieWordEmbedding::incremental_forwarding_lmhead(
   input_step_dim.batch(1);
   input_step_dim.height(1);
   hidden_step_dim.batch(1);
-  hidden_step_dim.height(1);
```
jaemini-shin, line 277: In my understanding, this line was removed because output height is always 1 set by finalize_lmhead.
But it's a bit hard to notice this with just looking at this method.
How about leaving a comment that height is always 1?

jaemini-shin (approved):
LGTM!

Seunghui98 (approved):
LGTM 👍

PR #3746, opened 2026-02-06
[CausalLM] build on Tizen10.0

lhs8928:
This PR is for building the nntrainer app CausalLM on the QRB4210 board(Armv8-a).

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -1267,25 +1267,25 @@ void MHACoreLayer::updateTensorsByInputDimensions(
     std::get<nntrainer::props::MaxTimestep>(mha_core_props).get();
   unsigned int &max_new_tokens =
     std::get<props::MaxNewTokens>(mha_core_props).get();
-  max_position_embeddings =
-    std::get<props::MaxPositionEmbeddings>(mha_core_props).get();
   max_timestep = height + max_new_tokens;
 
-  ml::train::TensorDim kv_dim = input_dimensions[0];
-  kv_dim.width(kv_dim.width() / (num_heads_Q / num_heads_KV));
+  ml::train::TensorDim q_dim = context.getInput(INOUT_INDEX::QUERY).getDim();
+  q_dim.height(input_dimensions[0].height());
 
-  ml::train::TensorDim kv_cache_dim = kv_dim;
-#ifdef ENABLE_FP16
-  kv_cache_dim.setDataType(ml::train::TensorDim::DataType::FP16);
-#else
-  kv_cache_dim.setDataType(ml::train::TensorDim::DataType::UINT16);
-#endif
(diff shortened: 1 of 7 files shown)


Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -1272,21 +1271,25 @@ void MHACoreLayer::updateTensorsByInputDimensions(
...
-  kv_cache_dim.setDataType(ml::train::TensorDim::DataType::UINT16);
-#endif
+  ml::train::TensorDim kv_dim = context.getInput(INOUT_INDEX::KEY).getDim();
+
+  ml::train::TensorDim kv_cache_dim =
+    context.getTensor(tensor_idx[AttentionParams::cache_key]).getDim();
   kv_cache_dim.height(max_timestep);
 
-  context.updateInput(INOUT_INDEX::QUERY, input_dimensions[0]);
+  precompute_freqs(head_dim, max_position_embeddings, theta);
```
jaemini-shin: Is this change necessary?

meson.build
```diff
@@ -188,6 +192,10 @@ if get_option('enable-fp16')
      elif arm_arch_opt == 'armv8.2-a'
        add_project_arguments('-march=armv8.2-a+fp16+dotprod+i8mm', language: ['c', 'cpp'])
        extra_defines += '-DARM_ARCH_ARMV8_2A=1'
+     elif arm_arch_opt == 'armv8-a'
+       add_project_arguments('-march=armv8-a+fp16', language: ['c', 'cpp'])
+       extra_defines += '-DARM_ARCH_ARMV8_A=1'
+       message('meson.build arm_arch_opt: armv8-a+fp16')
```
jaemini-shin, line 198: How about reordering in `elif armv8-a` and then `elif armv8.2-a`?

Applications/CausalLM/meson.build
```diff
@@ -70,4 +70,17 @@ e = executable('nntr_causallm',
...
+if get_option('platform') == 'tizen'
+    install_subdir('res' / 'Gauss0-2b',
+        install_dir: application_install_dir / 'res')
+
+    install_data('install_rpms.sh',
+        install_dir: application_install_dir)
+
+    install_subdir('res' / 'qwen3-0.6b',
+        install_dir: application_install_dir / 'res')
+endif
```
jaemini-shin: May I ask why these lines are needed?

Applications/CausalLM/layers/mha_core.cpp
lhs8928: Nice catch!
I forgot that precompute_freqs takes argument as max_position_embeddings instead of seq_len.

Applications/CausalLM/meson.build
lhs8928: Good point!
I removed it.

meson.build
lhs8928: For arm arch opt, since it was already sorted in descending order, I added it at the end.

baek2sm (approved):
LGTM

PR #3750, opened 2026-02-20
Multi-input Support

junbong-yu:
Multi-input Support: models like intfloat/multilingual-e5-large-instruct requires multiple input handling as batch.
This pr provides support for multiple-input sentences

How to handle Batched multi-input:
  - 1. input sentences for run() should be std::vector<std::string> type
  - 2. input sentences are treated as batched input, so you need to fix 'batch_size' value in nntr_config.json

**Self evaluation:**

Signed-off-by: Junbong Yu <junbong.yu@samsung.com>

### DIFF
--- a/Applications/CausalLM/models/embedding.cpp
+++ b/Applications/CausalLM/models/embedding.cpp
@@ -179,9 +179,29 @@ void Embedding::addModule(const std::string &type, int idx) {
 
 void Embedding::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
                     const WSTR tail_prompt) {
-
   try {
     std::vector<float *> results = encode(prompt, system_prompt, tail_prompt);
+
+    std::cout << "Embedding Result:" << std::endl;
+    std::cout << "[";
+    int print_dim = (DIM > 10) ? 10 : DIM;
+    for (int i = 0; i < print_dim; ++i) {
+      std::cout << results[0][i] << (i == print_dim - 1 ? "" : ", ");
+    }
+    if (DIM > 10)
+      std::cout << ", ...";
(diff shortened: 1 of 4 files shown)


Applications/CausalLM/models/embedding.cpp
```diff
@@ -254,6 +274,92 @@ std::vector<float *> Embedding::encode(const WSTR prompt,
...
+    throw std::runtime_error("Embedding model is not initialized. Please call "
+                             "initialize() before encode().");
+  }
+
+  // Check if all input vectors have the same size
+  if (prompts.size() != system_prompts.size() ||
+      prompts.size() != tail_prompts.size()) {
+    throw std::runtime_error("All input vectors (prompts, system_prompts, "
+                             "tail_prompts) must have the same size.");
+  }
```
jaemini-shin, line 290: This validation logic has problem when using default parameters for system and tail prompt.
In this case, their size is 1({""}), but prompts size can be longer.

Applications/CausalLM/models/transformer.h
```diff
@@ -100,6 +100,18 @@ WIN_EXPORT class Transformer {
...
+   * @brief run the Transformer model
+   */
+  virtual void run(const std::vector<WSTR> &prompts, bool do_sample,
+                   const std::vector<WSTR> system_prompts =
+                     {
+                       "",
+                     },
+                   const std::vector<WSTR> tail_prompts = {
+                     "",
+                   });
```
jaemini-shin, line 113:
```suggestion
                   const std::vector<WSTR> &system_prompts =
                     {
                       "",
                     },
                   const std::vector<WSTR> &tail_prompts = {
                     "",
                   });
```
This might improve performance.

PR #3751, opened 2026-02-20
[Wait for #3750] support multiple from/to for incremental_inference/forwarding

junbong-yu:
support multiple from/to for incremental_inference/forwarding
  - related issue: https://github.com/nntrainer/nntrainer/issues/3715
  - layers like mha_core requires segregated from/to's
  - fix incremental_inference/forwarding interface to handle multiple
     from/to's

**Self evaluation:**

Signed-off-by: Junbong Yu <junbong.yu@samsung.com>

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -186,28 +186,11 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
 
 /************************************************************** */
 
-/**
- * @note This forwarding function is used for training mode.
- *       This will be implemented ASAP.
- * @date 2024-09-02
- */
-void MHACoreLayer::forwarding(nntrainer::RunLayerContext &context,
-                              bool training) {}
-
-/**
- * @note This incremental_forwarding method is invoked for inference mode.
- *       Please note that Transformer Decoder's MHA takes only one sequence at a
- * step. Incremental forwarding function is used for this.
- */
-void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
-                                          unsigned int _from, unsigned int _to,
-                                          bool training) {
+void MHACoreLayer::check_max_timestep(const unsigned int &_from,
+                                      unsigned int &from, unsigned int &to) {
 
   unsigned int max_timestep =
     std::get<nntrainer::props::MaxTimestep>(mha_core_props).get();
-
-  unsigned int from = _from;
-  unsigned int to = _to;
 
   if (to >= max_timestep) {
     // initial forwarding
@@ -222,6 +205,58 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
       to = max_timestep;
     }
   }
+}
+
+/**
+ * @note This forwarding function is used for training mode.
+ *       This will be implemented ASAP.
+ * @date 2024-09-02
+ */
+void MHACoreLayer::forwarding(nntrainer::RunLayerContext &context,
+                              bool training) {}
+
+/**
+ * @note This incremental_forwarding method is invoked for inference mode.
+ *       Please note that Transformer Decoder's MHA takes only one sequence at a
+ * step. Incremental forwarding function is used for this.
+ */
+void MHACoreLayer::incremental_forwarding(
+  nntrainer::RunLayerContext &context, const std::vector<unsigned int> &_from,
+  const std::vector<unsigned int> &_to, bool training) {
+  return internal_incremental_forwarding(context, _from, _to, training);
+}
+
+/**
+ * @note This incremental_forwarding method is invoked for inference mode.
+ *       Please note that Transformer Decoder's MHA takes only one sequence at a
+ * step. Incremental forwarding function is used for this.
+ */
+void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
+                                          unsigned int _from, unsigned int _to,
+                                          bool training) {
+  return internal_incremental_forwarding(context, _from, _to, training);
+}
+
+template <typename T>
+void MHACoreLayer::internal_incremental_forwarding(
+  nntrainer::RunLayerContext &context, T T_from, T T_to, bool training) {
+  unsigned int _from;
+  unsigned int _to;
+
+  if constexpr (std::is_same_v<T, unsigned int>) {
+    _from = T_from;
+    _to = T_to;
+  } else if constexpr (std::is_same_v<T, std::vector<unsigned int>>) {
+    _from = *(std::min_element(T_from.begin(), T_from.end()));
+    _to = *(std::max_element(T_to.begin(), T_to.end()));
+  } else {
+    throw std::runtime_error("unsupported from, to type");
+  }
+
+  unsigned int from = _from;
+  unsigned int to = _to;
+
+  check_max_timestep(_from, from, to);
 
   // util fn to compute tensor dimension for one step.
   auto get_step_dim = [to, from](const ml::train::TensorDim &dim) {
@@ -313,11 +348,11 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
       V_step.copyData(value_step);
       if (use_sink) {
         one_batch_incremental_forwarding(
-          batch, _from, from, to, Q_step, K_step, V_step, O_step, cache_key,
+          batch, T_from, T_to, Q_step, K_step, V_step, O_step, cache_key,
           cache_value, cache_key_dim, cache_key_step_dim, cache_value_dim,
           cache_value_step_dim, sink);
       } else {
-        one_batch_incremental_forwarding(batch, _from, from, to, Q_step, K_step,
+        one_batch_incremental_forwarding(batch, from, to, Q_step, K_step,
                                          V_step, O_step, cache_key, cache_value,
                                          cache_key_dim, cache_key_step_dim,
                                          cache_value_dim, cache_value_step_dim);
@@ -326,19 +361,19 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
 #else
       if (use_sink) {
         one_batch_incremental_forwarding(
-          batch, _from, from, to, query_step, key_step, value_step, output_step,
+          batch, T_from, T_to, query_step, key_step, value_step, output_step,
           cache_key, cache_value, cache_key_dim, cache_key_step_dim,
           cache_value_dim, cache_value_step_dim, sink);
       } else {
         one_batch_incremental_forwarding(
-          batch, _from, from, to, query_step, key_step, value_step, output_step,
+          batch, T_from, T_to, query_step, key_step, value_step, output_step,
           cache_key, cache_value, cache_key_dim, cache_key_step_dim,
           cache_value_dim, cache_value_step_dim);
       }
 #endif
     } else {
       one_batch_incremental_forwarding(
-        batch, _from, from, to, query_step, key_step, value_step, output_step,
+        batch, T_from, T_to, query_step, key_step, value_step, output_step,
         cache_key, cache_value, cache_key_dim, cache_key_step_dim,
         cache_value_dim, cache_value_step_dim);
     }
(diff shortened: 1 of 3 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -495,6 +495,54 @@ NeuralNetwork::incremental_forwarding(unsigned int from, unsigned int to,
...
+     lookahead](std::shared_ptr<LayerNode> node, bool training) -> void {
+    PROFILE_MEM_ANNOTATE("Forwarding for layer: " + node->getName());
+
+    auto f = std::get<0>(node->getExecutionOrder());
+    if (exec_mode == ExecutionMode::TRAIN or
+        (exec_mode == ExecutionMode::INFERENCE and !fsu_mode)) {
+      model_graph.flushCacheExcept(f);
+
+      if (node->getType() == "mha_core")
+        node->incremental_forwarding(from, to, training);
```
junbong-yu, line 522: Requires review for this line, especially.

Explanation: 
for now, mha_core is the only layer that matters with multiple from/to problem. With thie PR, only mha_core handles multiple from/to's. other layers like input, fully_connected, will handle incremental_forwarding as it used to do before(single from, to)

api/ccapi/include/model.h
```diff
@@ -322,6 +322,25 @@ class Model {
...
+   * @param[in] from vector of start steps for each batch
+   * @param[in] to vector of end steps for each batch
+   * @param[in] output_hidden_state return last hidden state if true else return
+   * all hidden state
+   * @retval list of output as float *
+   * @note The output memory must not be freed by the caller
+   */
+  virtual std::vector<float *> incremental_inference(
+    unsigned int batch, const std::vector<float *> &input,
+    const std::vector<float *> &label, unsigned int init_seq_len,
```
h0g1, line 340: Is the argument 'init_seq_len' still scalar for multi-input support?

junbong-yu: Good point!!
For now, vectorized from/to's are for ranging masked area.
Calculation requires maxium sequence length among vectoriezed input, which is single scalar value.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -549,10 +606,39 @@ void MHACoreLayer::one_batch_incremental_forwarding(
       : ((to - from) * to),
     num_heads_Q, query_step.getTensorType());
 
+  out_.setValue((float)0);
+
+  nntrainer::Tensor mask(
```
jaemini-shin, line 611: Can you explain why this mask is necessary? Is there any problem using `from = T_from[batch]`, `to = T_to[batch]`?

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -535,7 +590,9 @@ void MHACoreLayer::one_batch_incremental_forwarding(
   ml::train::TensorDim cached_key_dim = cache_key_dim;
   ml::train::TensorDim cached_value_dim = cache_value_dim;
   cached_key_dim.height(to);
+  cached_key_dim.batch(1);
   cached_value_dim.height(to);
+  cached_value_dim.batch(1);
```
jaemini-shin, line 595: Nice catch! 👍

api/ccapi/include/model.h
h0g1: Then, sth like 'max_seq_len' would be better name!

nntrainer/layers/fc_layer.cpp
```diff
@@ -268,7 +268,7 @@ void FullyConnectedLayer::incremental_forwarding(RunLayerContext &context,
   // @todo make it parallelized with batch axis
   for (unsigned int b = 0; b < hidden_.batch(); ++b) {
     Tensor input_step = input_.getSharedDataTensor(
-      input_step_dim, b * hidden_dim.getFeatureLen(), true);
+      input_step_dim, b * input_dim.getFeatureLen(), true);
```
EunjuYang, line 271: Nice catch 👍

nntrainer/models/neuralnet.cpp
EunjuYang: This code branch might result in unexpected outcomes when a user attempts to perform true 'multi-batch' incremental inference for a causal model. What are your thoughts on this? Additionally, could you share the details of the model graph you are considering?

nntrainer/models/neuralnet.cpp
```diff
@@ -1172,6 +1280,14 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
       last_out_buf_data = new float[batch_size * out_t.width()];
 
+      /**
+       * If the output tensor's height (sequence dimension) is 1, it
+       * indicates a reduced/pooled output (e.g., Pooling,
+       * EmbeddingNormalize). In this case, the output for any 'step' is
+       * located at index 0. Otherwise, we access the specific step index
+       * corresponding to the last processed token.
+       */
+      unsigned int effective_step = (out_t.getDim().height() == 1) ? 0 : step;
```
EunjuYang, line 1290: Unused parameter.
```suggestion

```

nntrainer/models/neuralnet.cpp
```diff
@@ -1217,6 +1333,95 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
+      last_out_buf_data = new float[batch_size * out_t.width()];
+
+      std::vector<unsigned int> steps(batch_size);
+      for (unsigned int b = 0; b < batch_size; ++b) {
+        steps[b] = ((to[b] - from[b]) == 0) ? 0 : (to[b] - from[b]) - 1;
+      }
+
+      for (unsigned int b = 0; b < batch_size; ++b) {
+        unsigned int effective_step =
+          (out_t.getDim().height() == 1) ? 0 : steps[b];
```
EunjuYang, line 1397: ditto
```suggestion

```

nntrainer/models/neuralnet.cpp
```diff
@@ -1217,6 +1333,95 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
+
+    if (output_hidden_state) {
+      last_out_buf_data = out_t.getData();
+    } else {
+      last_out_buf_data = new float[batch_size * out_t.width()];
+
+      std::vector<unsigned int> steps(batch_size);
+      for (unsigned int b = 0; b < batch_size; ++b) {
+        steps[b] = ((to[b] - from[b]) == 0) ? 0 : (to[b] - from[b]) - 1;
+      }
```
EunjuYang, line 1393: I find these lines of code are no longer necessary, as the lm_head layer has been updated to always write its output to step 0. With this update, I believe there is no need for a layer to write its output with an offset based on its step :)
```suggestion
```

nntrainer/models/neuralnet.cpp
```diff
@@ -1217,6 +1333,95 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
+  } else {
+    output_tensors =
+      incremental_inference(input_tensors, init_seq_len, from, to);
+  }
+
+  std::vector<float *> output;
+
+  ///@note Always we take the first position of output
+  // unsigned int step = ((to - from) == 0) ? 0 : (to - from) - 1;
+  unsigned int step = 0;
```
EunjuYang, line 1378: To resolve the output step issue, it might be better to eliminate the step variable. :)
```suggestion
```

nntrainer/models/neuralnet.cpp
```diff
@@ -1217,6 +1333,95 @@ std::vector<float *> NeuralNetwork::incremental_inference(
...
+#else
+          throw std::invalid_argument("Error: enable-fp16 is not set");
+#endif
+        } else if (out->getDataType() == ml::train::TensorDim::DataType::FP32) {
+          const float *out_t_batch_ptr = out_t.getData() +
+                                         b * out_t.getDim().getFeatureLen() +
+                                         step * out_t.width();
+          scopy(out_t.width(), out_t_batch_ptr, 1,
+                last_out_buf_data + b * out_t.width(), 1);
+        }
```
EunjuYang, line 1415: Remove the `step` variables.
```suggestion
#ifdef ENABLE_FP16
          const _FP16 *out_t_batch_ptr = out_t.getData<_FP16>() +
                                         b * out_t.getDim().getFeatureLen() +
                                         out_t.width();
          scopy(out_t.width(), out_t_batch_ptr, 1,
                last_out_buf_data + b * out_t.width(), 1);
#else
          throw std::invalid_argument("Error: enable-fp16 is not set");
#endif
        } else if (out->getDataType() == ml::train::TensorDim::DataType::FP32) {
          const float *out_t_batch_ptr = out_t.getData() +
                                         b * out_t.getDim().getFeatureLen() +
                                         out_t.width();
          scopy(out_t.width(), out_t_batch_ptr, 1,
                last_out_buf_data + b * out_t.width(), 1);
        }
```

Applications/CausalLM/layers/mha_core.h
```diff
@@ -414,9 +431,9 @@ WIN_EXPORT class MHACoreLayer : public nntrainer::LayerImpl {
 
   void compute_kcaches(nntrainer::Tensor &in, nntrainer::Tensor &cache,
                        nntrainer::Tensor &out, unsigned int from,
-                       size_t sequence_len, unsigned int num_heads,
-                       unsigned int group_size, unsigned int head_dim,
-                       BS::thread_pool<> &pool);
+                       size_t sequence_len, size_t mask_size,
+                       unsigned int num_heads, unsigned int group_size,
+                       unsigned int head_dim, BS::thread_pool<> &pool);
```
EunjuYang, line 436: Could you add a docstring for this? Please include an explanation of the new parameter `mask_size`.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -381,12 +416,13 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
...
+  unsigned int num_head, unsigned int group_size, unsigned int head_dim,
+  BS::thread_pool<> &pool) {
 
   // Dispatch based on data type (FP32 or FP16)
   if (in.getDataType() == ml::train::TensorDim::DataType::FP32) {
-    if (sequence_len == 1) {
+    if (mask_size == 1) {
       // Single token processing (common during generation)
       // Parallelize over KV heads for decoding since Q direction is always 1
       int row_to_compute = is_causal ? from + 1 : from + sequence_len;
```
EunjuYang, line 428: Should it be changed as well?
```suggestion
      int row_to_compute = is_causal ? from + 1 : from + mask;
```
or 

```suggestion
      int row_to_compute = from + 1;
```

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -381,12 +416,13 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
  */
 void MHACoreLayer::compute_kcaches(
   nntrainer::Tensor &in, nntrainer::Tensor &cache, nntrainer::Tensor &out,
-  unsigned int from, size_t sequence_len, unsigned int num_head,
-  unsigned int group_size, unsigned int head_dim, BS::thread_pool<> &pool) {
+  unsigned int from, size_t sequence_len, size_t mask_size,
```
EunjuYang, line 419: Could you explain what `sequence_len` and `mask_size` are ?

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -434,6 +469,7 @@ void MHACoreLayer::compute_kcaches(
     }
   } else if (in.getDataType() == ml::train::TensorDim::DataType::FP16) {
 #ifdef ENABLE_FP16
+    int num_cache_head = num_head / group_size;
     if (sequence_len == 1) {
```
EunjuYang, line 473: Should it be fixed like below?

```suggestion
    if (mask_size == 1) {
```

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -549,10 +606,39 @@ void MHACoreLayer::one_batch_incremental_forwarding(
...
+    mask_to = T_to[batch];
+
+    for (i = from; i < to; i++) {
+      for (j = from; j < to; j++) {
+        if (j >= mask_from && j < mask_to)
+          for (k = 0; k < num_heads_Q; k++) {
+            mask.setValue(0, 0, (i * to + j), k, 0.0f);
+          }
+      }
+    }
```
EunjuYang, line 632: Do we always need to compute this? I'm concerned about whether this update might affect cases where the mask isn't necessary.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -549,10 +606,39 @@ void MHACoreLayer::one_batch_incremental_forwarding(
...
+  }
+
   unsigned int gqa_size = num_heads_Q / num_heads_KV;
 
-  compute_kcaches(query_step, b_cached_key, out_, _from, to - from, num_heads_Q,
-                  gqa_size, head_dim, pool);
+  compute_kcaches(query_step, b_cached_key, out_, _from, to - from, to - from,
+                  num_heads_Q, gqa_size, head_dim, pool);
```
EunjuYang, line 638: If the `sequence_len` and `mask_size` values are same, is there any reason we have to additionally define `mask_size`?

Applications/CausalLM/models/embedding.h
```diff
@@ -56,6 +60,22 @@ WIN_EXPORT class Embedding : virtual public Transformer {
   std::vector<float *> encode(const WSTR prompt, const WSTR system_prompt = "",
                               const WSTR tail_prompt = "");
 
+  /**
+   * @brief Encode the prompt and return the embedding
+   * @param prompt User prompt
+   * @param system_prompt System prompt
+   * @param tail_prompt Tail prompt
```
EunjuYang, line 67:
```suggestion
   * @param prompts User prompt
   * @param system_prompts System prompt
   * @param tail_prompts Tail prompt
```

EunjuYang (commented):
Nice work! :) Supporting multi-batch incremental inference seems quite challenging. Additionally, this PR isn't solely focused on enabling true multi-batch incremental forwarding; rather, it aims to leverage multi-batch incremental forwarding as a workaround to address certain structural issues in the embedding model.

We might need further discussions to ensure it covers general cases effectively.

PR #3762, opened 2026-03-02
[Draft] 1st iteration of mezo training pipeline of a small fc layer model on MNIST

sachin-nntrainer:
This is the first iteration of the Mezo training pipeline tested on the MNIST dataset using a fully connected (FC) layer.

The code is based on the paper: [Mezo Training Pipeline.](https://arxiv.org/abs/2305.17333)

Currently, the nntrainer implementation is not converging, possibly due to one of the following two reasons:

1. The seed is not set properly.

2. I have written a custom cross-entropy loss function, which is prone to errors. I could not use nntrainer::cross_entropy_softmax because the loss function with the model in inference mode was not allowed.

The Mezo binary can be run using the following command:

./build/Applications/mezo/jni/mezo_example /sachin/personalization/data/MNIST/raw/t10k-images-idx3-ubyte /sachin/personalization/data/MNIST/raw/t10k-labels-idx1-ubyte 100 128 0.00001 0.001

The PyTorch PoC in this PR demonstrates Mezo being performed on the MNIST dataset using an FC layer. In this case, convergence has been achieved.

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -1900,62 +1900,39 @@ RunStats NeuralNetwork::trainMeZO(const std::vector<unsigned int> &batch_sizes,
     }
   }
 
-  // Get parameter pointers for MeZO
-  auto param_ptrs = getParameterPointers();
-  size_t total_params = 0;
-  for (auto *p : param_ptrs) {
-    total_params += p->getDim().getDataLen();
-  }
-
-  // Initialize random number generator for MeZO perturbations
-  std::mt19937 gen(42);
-  std::normal_distribution<float> normal_dist(0.0f, 1.0f);
-
   RunStats stats;
   stats.max_epoch = epochs;
-
-  ml_logi("Starting MeZO training with %zu parameters, epsilon=%.6f",
-          total_params, mezo_epsilon);
 
   for (unsigned int epoch = 0; epoch < epochs; ++epoch) {
     stats.epoch_idx = epoch;
     float epoch_loss = 0.0f;
     unsigned int batch_count = 0;
+    int seed = 42 + epoch;
 
     for (size_t b = 0; b < inputs.size(); ++b) {
       unsigned int cur_batch_size = batch_sizes[b];
       std::vector<float *> input_batch = {inputs[b]};
       std::vector<float *> label_batch = {labels[b]};
 
-      // Generate random perturbation vector z
-      std::vector<float> z(total_params);
-      for (size_t i = 0; i < total_params; ++i) {
-        z[i] = normal_dist(gen);
-      }
-
       // MeZO algorithm: estimate gradient using finite differences
       // 1. Compute L(θ+ε·z) with positive perturbation
-      perturbParameters(mezo_epsilon, z, +1);
+      perturbParameters(mezo_epsilon, seed);
       float loss_plus =
         computeLossForwardOnly(cur_batch_size, input_batch, label_batch);
 
       // 2. Compute L(θ-ε·z) with negative perturbation
-      perturbParameters(mezo_epsilon, z, -2);
+      perturbParameters(-2 * mezo_epsilon, seed);
       float loss_minus =
         computeLossForwardOnly(cur_batch_size, input_batch, label_batch);
 
       // 3. Restore original parameters
-      perturbParameters(mezo_epsilon, z, +1);
+      perturbParameters(mezo_epsilon, seed);
 
       // 4. Estimate gradient: g ≈ (L(θ+) - L(θ-)) / (2ε) · z
-      float diff = (loss_plus - loss_minus) / (2.0f * mezo_epsilon);
-      std::vector<float> gradient_estimate(total_params);
-      for (size_t i = 0; i < total_params; ++i) {
-        gradient_estimate[i] = diff * z[i];
-      }
+      float projected_grad = (loss_plus - loss_minus) / (2.0f * mezo_epsilon);
 
       // 5. Update parameters: θ = θ - η · g
-      updateParametersMeZO(gradient_estimate, learning_rate);
+      updateParametersMeZO(projected_grad, learning_rate, seed);
 
       epoch_loss += (loss_plus + loss_minus) / 2.0f;
       batch_count++;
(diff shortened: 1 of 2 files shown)


Applications/mezo/jni/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+      std::min(batch_size, static_cast<unsigned int>(num_samples - offset));
+    float *in_ptr = images.data() + offset * feature_len;
+    float *lbl_ptr = labels.data() + offset * label_len;
+    input_batches.push_back(in_ptr);
+    label_batches.push_back(lbl_ptr);
+    batch_sizes.push_back(cur_batch_size);
+  }
+
+  // Convert ml::train::Model to NeuralNetwork for MeZO training
+  auto neural_net = static_cast<nntrainer::NeuralNetwork *>(model.get());
```
EunjuYang, line 164: `NeuralNetwork` should be hidden from the NNTrainer user.
It is more desirable to access the model using `ml::train::Model`

Applications/mezo/jni/main.cpp
```diff
@@ -0,0 +1,189 @@
...
+    "epochs=" + std::to_string(epochs),
+    "learning_rate=" + std::to_string(learning_rate),
+    "mezo_epsilon=" + std::to_string(mezo_epsilon)};
+
+  std::cout << "\nStarting MeZO training..." << std::endl;
+
+  try {
+    // Train using MeZO algorithm with clipped dataset
+    auto train_stats = neural_net->trainMeZO(batch_sizes, input_batches,
+                                             label_batches, train_properties);
```
EunjuYang, line 177: Similarly,
In this regard, `trainMeZO` cannot be called directly.
All interfaces that the NNTrainer user can utilize are defined in `model.h`.

nntrainer/models/neuralnet.cpp
```diff
@@ -1856,6 +1856,226 @@ void NeuralNetwork::forEachLayer(
...
+    for (size_t b = 0; b < inputs.size(); ++b) {
+      unsigned int cur_batch_size = batch_sizes[b];
+      std::vector<float *> input_batch = {inputs[b]};
+      std::vector<float *> label_batch = {labels[b]};
+
+      // Generate random perturbation vector z
+      std::vector<float> z(total_params);
+      for (size_t i = 0; i < total_params; ++i) {
+        z[i] = normal_dist(gen);
+      }
```
EunjuYang: Is it desirable to save the `z` for every parameters?

> At each step, we first sample a random seed s, and then for each of z’s four uses in Algorithm 1, we
reset the random number generator by s and resample the relevant entry of z. Using this in-place
implementation, MeZO has a memory footprint equivalent to the inference memory cost.

nntrainer/models/neuralnet.cpp
```diff
@@ -1856,6 +1856,226 @@ void NeuralNetwork::forEachLayer(
...
+  std::vector<nntrainer::Tensor *> params;
+  forEachLayer(
+    [&](ml::train::Layer &layer, RunLayerContext &rc, void *user_data) {
+      LayerNode &ln = static_cast<LayerNode &>(layer);
+      for (unsigned int i = 0; i < ln.getNumWeights(); ++i) {
+        params.push_back(&ln.getWeight(i));
+      }
+    },
+    nullptr);
+  return params;
```
EunjuYang, line 1970: There's no check for `trainable`, which may cause unexpected update for frozen parameters.

nntrainer/models/neuralnet.cpp
```diff
@@ -1856,6 +1856,226 @@ void NeuralNetwork::forEachLayer(
...
+      data[i] += direction * epsilon * z[idx + i];
+    }
+    idx += param_size;
+  }
+}
+
+float NeuralNetwork::computeLossForwardOnly(unsigned int batch_size,
+                                            std::vector<float *> &inputs,
+                                            std::vector<float *> &labels) {
+  auto output = inference(batch_size, inputs)[0];
```
EunjuYang, line 1997: I'm not sure it is desirable to plug-in this feature into nntrainer.

nntrainer/models/neuralnet.cpp
```diff
@@ -1856,6 +1856,226 @@ void NeuralNetwork::forEachLayer(
...
+
+      // 3. Restore original parameters
+      perturbParameters(mezo_epsilon, z, +1);
+
+      // 4. Estimate gradient: g ≈ (L(θ+) - L(θ-)) / (2ε) · z
+      float diff = (loss_plus - loss_minus) / (2.0f * mezo_epsilon);
+      std::vector<float> gradient_estimate(total_params);
+      for (size_t i = 0; i < total_params; ++i) {
+        gradient_estimate[i] = diff * z[i];
+      }
```
EunjuYang: It seems not desirable to implement MeZo in this way.
We cannot obtain some gains by MeZo (equivalent memory footprint with inference).

nntrainer/models/neuralnet.cpp
```diff
@@ -1856,6 +1856,226 @@ void NeuralNetwork::forEachLayer(
...
+
+float NeuralNetwork::computeLossForwardOnly(unsigned int batch_size,
+                                            std::vector<float *> &inputs,
+                                            std::vector<float *> &labels) {
+  auto output = inference(batch_size, inputs)[0];
+
+  // Assuming output is logits and we need to apply softmax
+  // Assuming single output tensor with shape [batch_size, num_classes]
+  unsigned int num_classes = 10;
+  std::vector<float> softmax_output(batch_size * num_classes);
```
EunjuYang, line 2002: Hard-coded line should be removed from `neuralnet.cpp`

EunjuYang (commented):
The core SPSA gradient estimation math is correct — the perturbation sequence, the projected gradient formula $(\ell + - \ell -) / (2 \epsilon) \cdot z$, and the parameter update are all faithful to MeZo algorthm in the paper. 
However, the current implementation bypasses NNTrainer's existing training infrastructure entirely, which introduces several architectural and correcness issues that should be addressed before merging.

--------------------

### 1. Architecture: MeZO should live inside `train()`, not beside it

**Prooblem** `trainMeZO()` reimplements the entire training loop from scratch — epoch iteration, batch management, loss logging, stats collection — duplicating what `train()` and `train_run()` already provides. This means MeZO cannot use DataBuffer, validation, checkpointing, callbacks, or any future improvements to the training pipeline.

**Suggestion** Implement MeZO as a proper `Optimizer` subclass and motify the existing `train_for_iteration` lambda to branch on optimizer type:

```
// Current flow
// forwarding -> backwarding(iter++) -> flushCache
// Proposed MeZO flow:
auto train_for_iteration = [this, ...](RunStats &stat, DataBuffer &buffer){
    if (opt->getType == MeZO::type){
       // new step 
       // forward pass 1
       // forward pass 2
       // compute loss
       // restore original parameters
        // update
    } else {
        forwarding(true, stop_cb, stop_user_data);
        backwarding(iter++, stop_cb, stop_user_data);
    }
    model_graph.flushCache();
};
```

### 2. Memory allocation: gradient and derivative tensors are unnecessary.
MeZO does not compute gradients or derivatives, yet how memory is allocated determines wheter those tensors exists. if `allocate(ExecutionMode::TRAIN)` is called, the tensor manager allocates gradient tensors for every weight and derivative tensors for every layer.
So, when the optimizer is MeZO, allocate at INFERENCE level, e.g.,:
```
// In NeuralNetwork::train(), before train_run():
if (opt->getType() == MeZO::type){
    status = allocate(ExecutionMode::INFERENCE);
} else {
    status = allocate(ExecutionMode::TRAIN);
}
```


### 3. in-place implementation?
As I understood, MeZO aims to reduce the memory by sampling z without saving the value. However, it seems that this code saves the value for the whole parameters.
`std::vector<float> z(total_params);`
Please avoid to save the full perturbation vector

PR #3769, opened 2026-03-03
[Quantizer] add ggml quantizer

### LINKED ISSUE #3755: Support quantize() for Q4_0
Currently, nntrainer supports the quantization and dequantization kernels for the ggml format, including Q4K, Q6K, and Q40. However, there is a lack of interfaces at the tensor level or application level for utilizing these functionalities. To address this, quantizer needs to support the ggml format quantization feature.

* Constraints: Only supports float tensor inputs.
* Supported types: Q4K, Q6K, Q40.
* Existing quantization types do not support repack operations, but ggml must support repack operations.

EunjuYang:

<details><summary>[Quantizer] add ggml quantizer</summary><br />

- This commit adds a ggml quantizer which supports Q4_K, Q4_0, Q6_K
- This commit adds unittest for ggml quantizer
- This commit resolves #3755

**Self evaluation:**

</details>

### Summary

- add ggml quantizer to support Q4_K/Q4_0/Q6_K quantization/dequtization support

### DIFF
--- a/nntrainer/tensor/quantizer.cpp
+++ b/nntrainer/tensor/quantizer.cpp
@@ -8,6 +8,7 @@
  * @bug		No known bugs except for NYI items
  */
 
+#include <cpu_backend.h>
 #include <math.h>
 #include <quantizer.h>
 #include <tensor.h>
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/quantizer.cpp
```diff
@@ -237,4 +238,160 @@ QScheme BinaryCodeBasedQuantizer::qscheme() const {
...
 }
 
+/**
+ * @brief GgmlQuantizer class
+ */
+std::unique_ptr<Quantizer> GgmlQuantizer::create() {
+  return std::make_unique<GgmlQuantizer>(scheme_);
+}
+
+Tensor GgmlQuantizer::quantize(const Tensor &input, Tdatatype qtype) {
```
jaemini-shin, line 248: `qtype` is unused. Shouldn't `out_dtype` be `qtype` and update `scheme_`?

EunjuYang: You are right. The current implementation is based solely on `scheme_`, which is undesirable.
I think it would be much better to update it to use qtype instead of `scheme_`.

To achieve this, my initial idea was to change the quantization type to `ggml`, moving away from specific types like `Q4_0`, `Q4Kx8`, or `Q6_0`.
However, to proceed this, I first needed to review the previous implementation.

Since I also need to examine how the GPU side uses QScheme, I initially implemented the quantizer following the earlier convention. However, I believe we should now update this quantizer to support the ggml format while performing tensor quantization based on qtype.
Doing so would allow us to reuse a single ggml quantizer across various qtype configurations. :)

nntrainer/tensor/quantizer.cpp
```diff
@@ -237,4 +238,160 @@ QScheme BinaryCodeBasedQuantizer::qscheme() const {
...
+}
+
+Tensor &GgmlQuantizer::quantize(const Tensor &input, Tensor &output,
+                                float *scales, unsigned int *zero_points) {
+  NNTR_THROW_IF(input.getDataType() != Tdatatype::FP32, std::invalid_argument)
+    << "[GgmlQuantizer::quantize] Input tensor must be FP32.";
+
+  NNTR_THROW_IF(output.empty(), std::invalid_argument)
+    << "[GgmlQuantizer::quantize] Output tensor is empty.";
+
```
jayden0701, line 320: I think adding extra verification for Output tensors would be good.

1) If output tensor matches input tensor's dimensions,
2) If output tensor's data type matches scheme_ (If this doesn't matter, maybe changing output tensor's quantized type to scheme_ might be needed?)

EunjuYang:
> I appreciate your work! By the way, isn't there quantizer already implemented in somewhere like arm_compute_backend.cpp?

Yes. There are kernel functions for quantization that are used in this quantizer.
What I did was a form of wrapper operating at the tensor level. I believe the CPU backend functions should be abstracted at the tensor level.

jijoongmoon (approved):
LGTM

PR #3773, opened 2026-03-04
[causallm] Support deberta_v2 embedding model

Seunghui98:
This commit supports deberta_v2 embedding model.
- add debert_attention layer
- add shared_fully_connected layer to reuse proj weight for debert attention
- add weight converter for deberta_v2

for: https://github.com/nntrainer/nntrainer/issues/3716
draft: https://github.com/nntrainer/nntrainer/pull/3714

**Self evaluation:**

### output comparison for sample input("DeBERTa v3 is a powerful encoder model.")
- transformers
<details>

<summary> used python file</summary>

```python
import torch
from transformers import AutoTokenizer, AutoModel

model_name = "microsoft/mdeberta-v3-base"

# tokenizer (SentencePiece 기반 → use_fast=False 권장)
tokenizer = AutoTokenizer.from_pretrained(
    model_name,
    torch_dtype=torch.float32,
    use_fast=False
)

# encoder model
model = AutoModel.from_pretrained(model_name, torch_dtype=torch.float32)
model.eval()

# sample input
text = "DeBERTa v3 is a powerful encoder model."

inputs = tokenizer(
    text,
    return_tensors="pt",
    padding=True,
    truncation=True
)

with torch.no_grad():
    outputs = model(**inputs)

# outputs
torch.set_printoptions(precision=7, sci_mode=False)
print(outputs.last_hidden_state[0, 0, :20])
```

</details>

```console
`torch_dtype` is deprecated! Use `dtype` instead!
Asking to truncate to max_length but no maximum length is provided and the model has no predefined maximum length. Default to no truncation.
tensor([-0.0164417, -0.1296964, -0.0693817,  0.0173562, -0.1183329, -0.1043039,
         0.2496491,  0.2931238, -0.0955804,  0.1029398, -0.0788064, -0.5542011,
         0.7671070,  0.0329785,  0.2213515,  0.0977408,  0.4115512, -0.1821359,
        -0.3463724, -0.0511756])
```
- nntrainer
```console
Embedding Result (1 batch(es)):
Batch 0: [-0.0164312, -0.129666, -0.0692946, 0.0176021, -0.118328, -0.104244, 0.249876, 0.29289, -0.0954586, 0.103098, ...] (Total DIM: 768)
[e2e time]: 918 ms 
Max Resident Set Size: 1327672 KB

```

### Rationale: Why I added deberta_attention_layer and shared_fully_connected_layer
- This PR adds DeBERTa v2/v3 encoder support to nntrainer. 
- The main gap was that DeBERTa’s attention mechanism isn't mapped for the existing Transformer attention layer, so we need to add dedicated attention layer. 

1) deberta_attention_layer: DeBERTa-style attention is structurally different from standard attention
  - The existing attention implementation in nntrainer is designed around standard scaled dot-product multi-head attention (Q·Kᵀ, softmax, then ·V). 
  - DeBERTa v2/v3 attention adds disentangled/relative attention components that change both the required inputs and the score computation.
  - Additional relative embedding inputs: depending on configuration, the layer requires extra inputs beyond (Q, K, V):K_rel for content-to-position (C2P), Q_rel for position-to-content (P2C)
  - Log-bucket relative position mapping: DeBERTa does not use a simple relative index lookup. Instead, it maps relative distances into buckets using a log-bucket function (compatible with HF), which is used when adding the C2P/P2C terms.

2) shared_fully_connected_layer: needed to reuse projection weights in DeBERTa attention
- shared_fully_connected_layer is needed because DeBERTa’s attention reuses the same Q/K projection weights to produce the C2P and P2C relative projections, so those weights must be shared (rather than duplicated) to match the intended DeBERTa computation and HF weight layout.
- In shared mode, the layer skips per-layer weight read/save, so the weight is expected to be loaded/saved once by the owning instance. 

- [[causallm] Support deberta_v2 embedding model](https://github.com/nntrainer/nntrainer/commit/d7aa234ce24882c02cc60ff979782533aa558502)

### DIFF
--- a/Applications/CausalLM/layers/deberta_attention_layer.cpp
+++ b/Applications/CausalLM/layers/deberta_attention_layer.cpp
@@ -19,22 +19,22 @@
 
 #include <deberta_attention_layer.h>
 
-#include <algorithm>   // std::min, std::max
-#include <cmath>       // std::log, std::ceil, std::sqrt, std::tanh, std::abs
-#include <future>      // std::future
-#include <limits>      // std::numeric_limits
-#include <mutex>       // std::lock_guard, std::mutex
-#include <vector>      // std::vector
+#include <algorithm> // std::min, std::max
+#include <cmath>     // std::log, std::ceil, std::sqrt, std::tanh, std::abs
+#include <future>    // std::future
+#include <limits>    // std::numeric_limits
+#include <mutex>     // std::lock_guard, std::mutex
+#include <vector>    // std::vector
 
-#include <omp.h>
 #include <engine.h>
 #include <fp16.h>
 #include <nntrainer_error.h>
+#include <omp.h>
 
 #include <node_exporter.h>
 
 #if !defined(__ANDROID__)
-  #include <cblas.h>
+#include <cblas.h>
 #endif
 
 namespace causallm {
@@ -103,8 +103,8 @@ static thread_local RelativeIndexKey tl_rel_index_key{};
 static thread_local RelativeIndexValue tl_rel_index_value{};
 
 /**
- * Scratch buffer for unpacking key cache to FP32 in FP32 relative-attention path.
- * Reused per-thread to avoid repeated allocation/resizing overhead.
+ * Scratch buffer for unpacking key cache to FP32 in FP32 relative-attention
+ * path. Reused per-thread to avoid repeated allocation/resizing overhead.
  */
 static thread_local std::vector<float> tl_key_cache_fp32_buf;
 
@@ -156,7 +156,8 @@ void DebertaAttentionLayer::finalize(nntrainer::InitLayerContext &context) {
       expected_inputs++;
   }
 
-  NNTR_THROW_IF(context.getNumInputs() != expected_inputs, std::invalid_argument)
+  NNTR_THROW_IF(context.getNumInputs() != expected_inputs,
+                std::invalid_argument)
     << "DebertaAttentionLayer expects " << expected_inputs
     << " inputs, but got " << context.getNumInputs();
 
@@ -196,8 +197,8 @@ void DebertaAttentionLayer::finalize(nntrainer::InitLayerContext &context) {
   is_causal = false;
 
   local_window_size = std::max<unsigned int>(
-    query_dim.height(), max_position_embeddings > 0 ? max_position_embeddings
-                                                    : query_dim.height());
+    query_dim.height(),
+    max_position_embeddings > 0 ? max_position_embeddings : query_dim.height());
 
   attn_logit_softcapping = 0.0f;
 
@@ -229,8 +230,8 @@ void DebertaAttentionLayer::finalize(nntrainer::InitLayerContext &context) {
   output_dims[0] = query_dim;
   context.setOutputDimensions(output_dims);
 
-  prepare_bucket_table(std::max<unsigned int>(query_dim.height(),
-                                              max_position_embeddings));
+  prepare_bucket_table(
+    std::max<unsigned int>(query_dim.height(), max_position_embeddings));
 
 #if !defined(__ANDROID__)
   // Keep BLAS single-threaded inside the layer to avoid oversubscription.
@@ -238,7 +239,8 @@ void DebertaAttentionLayer::finalize(nntrainer::InitLayerContext &context) {
 #endif
 }
 
-void DebertaAttentionLayer::setProperty(const std::vector<std::string> &values) {
+void DebertaAttentionLayer::setProperty(
+  const std::vector<std::string> &values) {
   auto remain_props = loadProperties(values, deberta_props);
   LayerImpl::setProperty(remain_props);
 }
@@ -259,9 +261,9 @@ void DebertaAttentionLayer::prepare_bucket_table(unsigned int max_seq_len) {
 
   for (int diff = -static_cast<int>(max_seq_len);
        diff <= static_cast<int>(max_seq_len); ++diff) {
-    bucket_table[diff + offset] = compute_bucket_pos_impl(
-      diff, static_cast<int>(position_buckets),
-      static_cast<int>(max_relative_positions));
+    bucket_table[diff + offset] =
+      compute_bucket_pos_impl(diff, static_cast<int>(position_buckets),
+                              static_cast<int>(max_relative_positions));
   }
 
   bucket_table_max_seq_len = max_seq_len;
@@ -273,8 +275,8 @@ int DebertaAttentionLayer::lookup_bucket(int relative_pos) const {
     return relative_pos;
 
   const int offset = static_cast<int>(bucket_table_max_seq_len);
-  const int idx = clampv(relative_pos + offset, 0,
-                         static_cast<int>(bucket_table.size()) - 1);
+  const int idx =
+    clampv(relative_pos + offset, 0, static_cast<int>(bucket_table.size()) - 1);
   return bucket_table[idx];
 }
 
@@ -330,8 +332,8 @@ void DebertaAttentionLayer::incremental_forwarding(
       query_step_dim,
       batch * query_dim.getFeatureLen() + _from * query_dim.width(), true);
     nntrainer::Tensor key_step = key.getSharedDataTensor(
-      key_step_dim,
-      batch * key_dim.getFeatureLen() + _from * key_dim.width(), true);
+      key_step_dim, batch * key_dim.getFeatureLen() + _from * key_dim.width(),
+      true);
     nntrainer::Tensor value_step = value.getSharedDataTensor(
       value_step_dim,
       batch * value_dim.getFeatureLen() + _from * value_dim.width(), true);
@@ -655,13 +657,11 @@ void DebertaAttentionLayer::add_relative_attn_score(
    * Shared relative index cache across attention layers
 
    */
-  const unsigned int rel_len_q =
-    p2c ? rel_query.height() : 0u;
-  const unsigned int rel_len_k =
-    c2p ? rel_key.height() : 0u;
+  const unsigned int rel_len_q = p2c ? rel_query.height() : 0u;
+  const unsigned int rel_len_k = c2p ? rel_key.height() : 0u;
 
-  const RelativeIndexKey cache_key{
-    from, to, att_span, rel_len_q, rel_len_k, c2p, p2c};
+  const RelativeIndexKey cache_key{from,      to,  att_span, rel_len_q,
+                                   rel_len_k, c2p, p2c};
 
   RelativeIndexValue &rel_idx_local = tl_rel_index_value;
 
@@ -676,27 +676,26 @@ void DebertaAttentionLayer::add_relative_attn_score(
       rel_idx_local.p2c_idx.resize(static_cast<size_t>(S_q) * S_k);
     }
 
-    #pragma omp parallel for schedule(static)
-      for (unsigned int q = 0; q < S_q; ++q) {
-        for (unsigned int k = 0; k < S_k; ++k) {
-          if (c2p) {
-            const int rel = static_cast<int>(q + from) - static_cast<int>(k);
-            const int bucketed = lookup_bucket(rel);
-            rel_idx_local.c2p_idx[static_cast<size_t>(q) * S_k + k] =
-              clampv(bucketed + static_cast<int>(att_span), 0,
-                    static_cast<int>(rel_len_k) - 1);
-          }
+#pragma omp parallel for schedule(static)
+    for (unsigned int q = 0; q < S_q; ++q) {
+      for (unsigned int k = 0; k < S_k; ++k) {
+        if (c2p) {
+          const int rel = static_cast<int>(q + from) - static_cast<int>(k);
+          const int bucketed = lookup_bucket(rel);
+          rel_idx_local.c2p_idx[static_cast<size_t>(q) * S_k + k] =
+            clampv(bucketed + static_cast<int>(att_span), 0,
+                   static_cast<int>(rel_len_k) - 1);
+        }
 
-          if (p2c) {
-            const int rel_rev =
-              static_cast<int>(k) - static_cast<int>(q + from);
-            const int bucketed_rev = lookup_bucket(rel_rev);
-            rel_idx_local.p2c_idx[static_cast<size_t>(q) * S_k + k] =
-              clampv(-bucketed_rev + static_cast<int>(att_span), 0,
-                    static_cast<int>(rel_len_q) - 1);
-          }
+        if (p2c) {
+          const int rel_rev = static_cast<int>(k) - static_cast<int>(q + from);
+          const int bucketed_rev = lookup_bucket(rel_rev);
+          rel_idx_local.p2c_idx[static_cast<size_t>(q) * S_k + k] =
+            clampv(-bucketed_rev + static_cast<int>(att_span), 0,
+                   static_cast<int>(rel_len_q) - 1);
         }
       }
+    }
 
     tl_rel_index_key = cache_key;
     tl_rel_index_ready = true;
@@ -705,24 +704,22 @@ void DebertaAttentionLayer::add_relative_attn_score(
   if (score.getDataType() == ml::train::TensorDim::DataType::FP32) {
     float *score_ptr = score.getData<float>();
     const float *q_ptr = query_step.getData<float>();
-    const float *rel_query_ptr =
-      p2c ? rel_query.getData<float>() : nullptr;
-    const float *rel_key_ptr =
-      c2p ? rel_key.getData<float>() : nullptr;
+    const float *rel_query_ptr = p2c ? rel_query.getData<float>() : nullptr;
+    const float *rel_key_ptr = c2p ? rel_key.getData<float>() : nullptr;
 
     const uint16_t *key_u16_ptr =
       key_cache.getDataType() == ml::train::TensorDim::DataType::UINT16
         ? key_cache.getData<uint16_t>()
         : nullptr;
 
-    #ifdef ENABLE_FP16
+#ifdef ENABLE_FP16
     const _FP16 *key_fp16_ptr =
       key_cache.getDataType() == ml::train::TensorDim::DataType::FP16
         ? key_cache.getData<_FP16>()
         : nullptr;
-    #else
+#else
     const void *key_fp16_ptr = nullptr;
-    #endif
+#endif
 
     std::vector<float> &key_cache_fp32_buf = tl_key_cache_fp32_buf;
     const float *key_unpacked_ptr = nullptr;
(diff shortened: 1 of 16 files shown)


Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,442 @@
...
+  // HF: log_pos = ceil(log(abs_pos/mid)/log((max_position-1)/mid) * (mid-1)) +
+  // mid
+  const double num = std::log((double)abs_pos / (double)mid);
+  const double den = std::log((double)(max_position - 1) / (double)mid);
+  int log_pos = (int)std::ceil(num / den * (mid - 1)) + mid;
+
+  // bucket_pos = where(abs_pos <= mid, relative_pos, log_pos * sign)
+  const int bucket_pos = (abs_pos <= mid) ? relative_pos : (log_pos * sign);
+  return bucket_pos;
+}
```
Seunghui98: ## Note for Reviewers

you can refer [transformer code].(https://github.com/huggingface/transformers/blob/5c1c72b/src/transformers/models/deberta_v2/modeling_deberta_v2.py).

```python
@torch.jit.script
def make_log_bucket_position(relative_pos, bucket_size: int, max_position: int):
    sign = torch.sign(relative_pos)
    mid = bucket_size // 2
    abs_pos = torch.where(
        (relative_pos < mid) & (relative_pos > -mid),
        torch.tensor(mid - 1).type_as(relative_pos),
        torch.abs(relative_pos),
    )
    log_pos = (
        torch.ceil(torch.log(abs_pos / mid) / torch.log(torch.tensor((max_position - 1) / mid)) * (mid - 1)) + mid
    )
    bucket_pos = torch.where(abs_pos <= mid, relative_pos.type_as(log_pos), log_pos * sign)
    return bucket_pos
```
https://github.com/huggingface/transformers/blob/5c1c72b/src/transformers/models/deberta_v2/modeling_deberta_v2.py#L56-L69

Applications/CausalLM/layers/shared_fully_connected_layer.cpp
```diff
@@ -0,0 +1,218 @@
...
+  bool is_nchw = (context.getFormat() == ml::train::TensorDim::Format::NCHW);
+  /** set output dimensions */
+  auto const &in_dim = context.getInputDimensions()[0];
+  output_dims[0] = in_dim;
+  is_nchw ? output_dims[0].width(unit) : output_dims[0].channel(unit);
+
+  output_dims[0].setTensorType(
+    {context.getFormat(), context.getActivationDataType()});
+
+  context.setOutputDimensions(output_dims);
```
Seunghui98, line 63: ### Note for Reviewers

Theses code are from fc_layer in nntrainer.

https://github.com/nntrainer/nntrainer/blob/052a1962a8ab7116aa0126cb6153f7975548d881/nntrainer/layers/fc_layer.cpp#L76-L90

Applications/CausalLM/models/transformer.cpp
```diff
@@ -129,9 +129,13 @@ void Transformer::setupParameters(json &cfg, json &generation_cfg,
...
-  ROPE_THETA = cfg["rope_theta"].get<unsigned int>();
-  TIE_WORD_EMBEDDINGS = cfg["tie_word_embeddings"].get<bool>();
-  NORM_EPS = cfg["rms_norm_eps"];
+  ROPE_THETA =
+    cfg.contains("rope_theta") ? cfg["rope_theta"].get<unsigned int>() : 10000;
+  TIE_WORD_EMBEDDINGS = cfg.contains("tie_word_embeddings")
+                          ? cfg["tie_word_embeddings"].get<bool>()
+                          : false;
+  NORM_EPS =
+    cfg.contains("rms_norm_eps") ? cfg["rms_norm_eps"].get<float>() : 1e-5;
```
Seunghui98: ### Note for Reivewers

- deberta_v2's model config hasn't `rope_theta` and `rms_norm_eps`.

nntrainer/graph/network_graph.cpp
```diff
@@ -785,9 +785,9 @@ NetworkGraph::finalizeContext(const std::shared_ptr<LayerNode> &lnode,
    */
   std::vector<std::string> input_names;
   input_names.reserve(prev_inputs.size());
-  std::transform(
-    prev_inputs.begin(), prev_inputs.end(), std::back_inserter(input_names),
-    [](auto const &vg) -> const auto & { return vg->getName(); });
+  std::transform(prev_inputs.begin(), prev_inputs.end(),
+                 std::back_inserter(input_names),
+                 [](auto const &vg) -> const auto & { return vg->getName(); });
```
Seunghui98: ### Note for Reviewers

- maybe.. these codes are formatted.

nntrainer/graph/network_graph.cpp
```diff
@@ -978,9 +978,9 @@ NetworkGraph::refinalizeContext(const std::shared_ptr<LayerNode> &lnode,
...
```
Seunghui98: ### Note for Reviewers

- maybe.. these codes are formatted.

Applications/CausalLM/main.cpp
```diff
@@ -241,8 +249,13 @@ int main(int argc, char *argv[]) {
...
     // Initialize and run model
-    std::string architecture =
-      cfg["architectures"].get<std::vector<std::string>>()[0];
+    std::string architecture;
+
+    if (cfg.contains("architectures")) {
+      architecture = cfg["architectures"].get<std::vector<std::string>>()[0];
+    } else if (cfg.contains("model_type")) {
+      architecture = cfg["model_type"].get<std::string>();
+    }
```
Seunghui98: ## Note for Reviewers

- mdeberta-v2/v3's config uses the key of `model_type` rather than `architectures` for model architecture information.
```
{
	"model_type": "deberta-v2",
	"attention_probs_dropout_prob": 0.1,
	"hidden_act": "gelu",
	"hidden_dropout_prob": 0.1,
	"hidden_size": 768,
	"initializer_range": 0.02,
	"intermediate_size": 3072,
	"max_position_embeddings": 512,
	"relative_attention": true,
	"position_buckets": 256,
	"norm_rel_ebd": "layer_norm",
	"share_att_key": true,
	"pos_att_type": "p2c|c2p",
	"layer_norm_eps": 1e-7,
	"max_relative_positions": -1,
	"position_biased_input": false,
	"num_attention_heads": 12,
	"num_hidden_layers": 12,
	"type_vocab_size": 0,
	"vocab_size": 251000
}
```

nntrainer/graph/network_graph.cpp
```diff
@@ -1191,7 +1191,7 @@ int NetworkGraph::initialize(ExecutionMode mode,
 
   /** check if the given config of node is of input node */
   auto is_input_node = [](const LayerNode *node) -> bool {
-    return node->getInputConnections().empty();
+    return node->getInputConnections().empty() && node->getType() != "weight";
```
Seunghui98: ### Note for Reviewers

- deberta_v2 model uses `weight_layer` for relative_embedding.
- so, In this case, `input_connection` doesn't exist in the weight_layer, so the true is always returned and the throw occurs below.

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1027 @@
...
+ * @author Seunghui Lee <shsh1004.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  DeBERTa Attention Layer based on mha_core optimized path.
+ *
+ *         Main execution flow:
+ *           1) compute_kcaches()               -> content-to-content score
+ *           2) rescale content score           -> match DeBERTa scaling
+ *           3) add_relative_attn_score()       -> add c2p / p2c score
+ *           4) softmax_triangle()              -> softmax on score
+ *           5) compute_fp16vcache_transposed() -> score * value
```
Seunghui98, line 17: ## Note for Reviewers

you can refer this execution flow.
- I took the implementation in `mha_core` for `compute_kcaches()`, `softmax_triangle()`  and `compute_fp16vcache_transposed()` and used these.
- `rotary_embedding` is applied to Q and K projections in `mha_core` for position embedding. But, in `deberta_attention`, after the relative position embedding is separately added to the `attention_score`, we can calculate softmax value.

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1027 @@
...
+  const int sign = (relative_pos >= 0) ? 1 : -1;
+  const int abs_rp = std::abs(relative_pos);
+  const int abs_pos = (abs_rp < mid) ? (mid - 1) : abs_rp;
+
+  const double num = std::log((double)abs_pos / (double)mid);
+  const double den = std::log((double)(max_position - 1) / (double)mid);
+  int log_pos = (int)std::ceil(num / den * (mid - 1)) + mid;
+
+  return (abs_pos <= mid) ? relative_pos : (log_pos * sign);
+}
```
Seunghui98, line 61: ## Note for Reviewers

This implementation is from `transformers/models/deberta_v2/modeling_deberta_v2.py`. -> [ref](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deberta_v2/modeling_deberta_v2.py#L56-L69)

```python
@torch.jit.script
def make_log_bucket_position(relative_pos, bucket_size: int, max_position: int):
    sign = torch.sign(relative_pos)
    mid = bucket_size // 2
    abs_pos = torch.where(
        (relative_pos < mid) & (relative_pos > -mid),
        torch.tensor(mid - 1).type_as(relative_pos),
        torch.abs(relative_pos),
    )
    log_pos = (
        torch.ceil(torch.log(abs_pos / mid) / torch.log(torch.tensor((max_position - 1) / mid)) * (mid - 1)) + mid
    )
    bucket_pos = torch.where(abs_pos <= mid, relative_pos.type_as(log_pos), log_pos * sign)
    return bucket_pos
```

nntrainer/models/neuralnet.cpp
```diff
@@ -688,12 +689,13 @@ void NeuralNetwork::load(const std::string &file_path,
 
   size_t start_from = 0;
   std::vector<std::pair<size_t, size_t>> file_offset;
+  std::unordered_set<std::string> visit_weights;
```
jaemini-shin: Can you please check #3810 ? This PR might resolve this logic.

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1024 @@
...
+      }
+
+      for (auto &fut : futures)
+        fut.get();
+    }
+#else
+    NNTR_THROW_IF(true, std::invalid_argument) << "enable-fp16 is not set!";
+#endif
+  }
+}
```
Seunghui98, line 501: ### Note for Reviewers

This function is same as `compute_kcaches` in `mha_core` layer.

https://github.com/nntrainer/nntrainer/blob/0e50be090a31c803a8c9f98f0859e206b85fff82/Applications/CausalLM/layers/mha_core.cpp#L382-L482

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1024 @@
...
+        }));
+      }
+      for (auto &fut : futures)
+        fut.get();
+    }
+#else
+    NNTR_THROW_IF(true, std::invalid_argument) << "enable-fp16 is not set!";
+#endif
+  }
+}
```
Seunghui98, line 565: ### Note for Reviewers

This function is same as `softmax_triangle` in `mha_core` layer considering only `is_causal=false`.

https://github.com/nntrainer/nntrainer/blob/0e50be090a31c803a8c9f98f0859e206b85fff82/Applications/CausalLM/layers/mha_core.cpp#L938-L1038

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1024 @@
...
+        nntrainer::compute_fp16vcache_transposed(
+          row_num, in_data, vcache_data, output_data, num_cache_head, gqa_size,
+          head_dim, local_window_size, head_kv, head_kv + 1);
+      }
+    }
+#else
+    NNTR_THROW_IF(true, std::invalid_argument) << "enable-fp16 is not set!";
+#endif
+  }
+}
```
Seunghui98, line 655: ### Note for Reviewers

This function is same as `compute_fp16vcache_transposed` in `mha_core` layer.

https://github.com/nntrainer/nntrainer/blob/0e50be090a31c803a8c9f98f0859e206b85fff82/Applications/CausalLM/layers/mha_core.cpp#L1143-L1248

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1023 @@
...
+/**
+ * @brief incremental forwarding
+ */
+void DebertaAttentionLayer::incremental_forwarding(
+  nntrainer::RunLayerContext &context, unsigned int _from, unsigned int _to,
+  bool training) {
+
+  auto get_step_dim = [_from, _to](const ml::train::TensorDim &dim) {
+    auto step_dim = dim;
+    step_dim.batch(1);
```
EunjuYang, line 351: The non-incremental `forwarding()` copies `q` directly to `output` without performing any attention computation. If this code path is reached, it would return incorrect results (raw query vectors instead of attention output).

If non-incremental forwarding is not yet implemented, it would be safer to throw an exception (similar to `calcDerivative`/`calcGradient`) so callers get a clear error rather than silently wrong output:
```cpp
void DebertaAttentionLayer::forwarding(nntrainer::RunLayerContext &context,
                                       bool training) {
  throw nntrainer::exception::not_supported(
    "DebertaAttentionLayer::forwarding (non-incremental) is not supported yet");
}
```

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1023 @@
...
+  if (p2c) {
+    rel_query = context.getInput(next_input_idx++);
+  }
+  if (c2p) {
+    rel_key = context.getInput(next_input_idx++);
+  }
+
+  const unsigned int S_q = to - from;
+  const unsigned int S_k = to;
+  const unsigned int hidden = static_cast<unsigned int>(num_heads_Q * head_dim);
```
EunjuYang, line 686: `key_unpacked_ptr` is only assigned when `p2c` is true (inside the `if (p2c && ...)` block above). When `c2p=true, p2c=false`, this pointer remains `nullptr`, and this `NNTR_THROW_IF` will always fire — even though the c2p path doesn't need the unpacked key cache at all.

This check should be moved inside a `if (p2c)` guard, or made conditional:
```cpp
NNTR_THROW_IF(p2c && key_unpacked_ptr == nullptr, std::invalid_argument)
  << "FP32 p2c path expected UINT16 or FP16 key cache";
```

Applications/CausalLM/layers/shared_fully_connected_layer.h
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * @copydoc Layer::calcGradient(RunLayerContext &context)
+   */
+  void calcGradient(nntrainer::RunLayerContext &context) override;
+
+  /**
+   * @copydoc Layer::exportTo(Exporter &exporter, const ExportMethods &method)
+   */
+  void exportTo(nntrainer::Exporter &exporter,
```
EunjuYang, line 116: `supportBackwarding()` returns `true`, but both `calcDerivative()` and `calcGradient()` throw `not_supported`. This is contradictory — the graph builder may attempt to allocate gradient tensors and call backprop based on `supportBackwarding() == true`, only to hit a runtime exception.

Since this layer is inference-only, this should return `false`.

nntrainer/graph/network_graph.cpp
```diff
@@ -225,7 +225,7 @@ int NetworkGraph::checkCompiledGraph() {
   /** Dimension of input layers must be known */
   for (auto iter = cbegin(); iter != cend(); iter++) {
     auto lnode = (*iter);
-    if (lnode->getNumInputConnections() == 0) {
+    if (lnode->getNumInputConnections() == 0 && lnode->getType() != "weight") {
```
EunjuYang: This hardcoded `lnode->getType() != "weight"` check appears in 3 places in this file. It works, but it creates a tight coupling between the graph initialization logic and a specific layer type name.

If another layer type has similar behavior in the future (no inbound connections, self-contained dimensions), each of these 3 locations would need updating. A more maintainable approach might be to add a virtual method to the Layer interface (e.g., `bool providesOwnInputDimension() const`) that the weight layer overrides.

Not a blocker, but worth considering for long-term maintainability.

Applications/CausalLM/layers/deberta_attention_layer.cpp
```diff
@@ -0,0 +1,1023 @@
...
+  for (int diff = -static_cast<int>(max_seq_len);
+       diff <= static_cast<int>(max_seq_len); ++diff) {
+    bucket_table[diff + offset] =
+      compute_bucket_pos_impl(diff, static_cast<int>(position_buckets),
+                              static_cast<int>(max_relative_positions));
+  }
+
+  bucket_table_max_seq_len = max_seq_len;
+  bucket_table_ready = true;
+}
```
EunjuYang, line 324: This sets a global OpenBLAS thread count that affects all BLAS operations across the entire process, not just this layer. A few questions:

1. Why specifically 4? Is this tuned for a particular target device?
2. Could this interfere with other layers that may benefit from different thread counts?

Consider making this configurable via a property, or at least adding a comment explaining the rationale for this specific value.

Applications/CausalLM/meson.build
```diff
@@ -26,6 +26,9 @@ causallm_layer_dependencies += [
     causallm_qkv_layer_dep,
     causallm_embedding_pooling_layer_dep,
     causallm_embedding_normalize_layer_dep,
+    causallm_qkv_layer_dep,
```
EunjuYang: `causallm_qkv_layer_dep` appears twice in this list (it was already present on line 25). This is harmless but should be cleaned up.

Applications/CausalLM/models/deberta_v2/deberta_v2.h
```diff
@@ -0,0 +1,73 @@
...
+   * @brief register CustomLayers
+   */
+  void registerCustomLayers() override;
+
+  int MAX_RELATIVE_POSITIONS;
+  bool c2p;
+  bool p2c;
+  bool SHARE_ATT_KEY;
+  bool RELATIVE_ATTENTION;
+  int POSITION_BUCKETS;
```
EunjuYang: Naming inconsistency: `MAX_RELATIVE_POSITIONS`, `POSITION_BUCKETS`, `SHARE_ATT_KEY`, `RELATIVE_ATTENTION` use all-caps (matching the parent class convention), while `c2p` and `p2c` use lowercase. Consider making these consistent — e.g., `C2P_ENABLED` and `P2C_ENABLED`, or at least documenting the distinction.

Applications/CausalLM/models/deberta_v2/deberta_v2.cpp
```diff
@@ -0,0 +1,331 @@
...
+  try {
+    app_context->registerFactory(
+      nntrainer::createLayer<causallm::DebertaAttentionLayer>);
+    app_context->registerFactory(
+      nntrainer::createLayer<causallm::SharedFullyConnectedLayer>);
+  } catch (std::invalid_argument &e) {
+    std::cerr << "failed to register factory, reason: " << e.what()
+              << std::endl;
+  }
+}
```
EunjuYang, line 364: Catching all exceptions and printing to stderr allows execution to continue with partially initialized parameters (e.g., `c2p`, `p2c`, `POSITION_BUCKETS` may hold default/stale values). This could cause subtle issues downstream.

Consider either re-throwing after logging, or at minimum setting a flag that prevents `constructModel()` from proceeding with incomplete configuration.

nntrainer/models/neuralnet.cpp
```diff
@@ -706,8 +708,22 @@ void NeuralNetwork::load(const std::string &file_path,
...
+        weight->getVariableRef().setFileOffset(start_from - size);
+        file_offset.emplace_back(std::make_pair(start_from - size, size));
+      } else {
+        /// Shared/tied weights may appear multiple times across layers.
+        /// Only the first occurrence should advance `start_from` and set the
+        /// file offset; otherwise the computed layout is corrupted.
+        ///
+        /// Subsequent instances share the same backing memory, and the first
+        /// `Tensor::read()` populates that shared buffer.
+        // weight->getVariableRef().setFileOffset(...); // Intentionally skipped
```
EunjuYang: Good fix. Shared/tied weights appearing multiple times would cause `start_from` to advance incorrectly, corrupting the file offset layout for all subsequent weights. Tracking visited weight names and skipping duplicates is the right approach.

EunjuYang (changes requested):
## Review Summary

This is a substantial and well-thought-out addition — DeBERTa v2/v3 encoder support with disentangled attention, shared weight projection, and the necessary core infrastructure changes. The numerical comparison with HuggingFace transformers in the PR description is appreciated and shows the implementation is accurate.

### What works well

- **Algorithmic correctness**: The disentangled attention (C2P, P2C, log-bucket relative positions) faithfully implements the DeBERTa paper and matches the HF reference output.
- **Performance-conscious design**: Thread-local caching for relative index tables, pre-computed bucket lookup, `#pragma omp simd` for dot products, and reuse of the existing mha_core optimized paths (compute_kcaches, softmax, vcache).
- **Shared weight file offset fix** (`neuralnet.cpp`): Correctly prevents shared/tied weights from double-advancing `start_from`, which would corrupt the file layout for all subsequent weights. This is an important fix.
- **Backward-compatible transformer.cpp changes**: Making `rope_theta`, `tie_word_embeddings`, and `rms_norm_eps` optional with sensible defaults is clean and doesn't break existing models.
- **Clear PR description**: Design rationale for the new layers, reference links, and output comparison make this easy to review.

### Items to address

1. **`forwarding()` silently returns wrong results** — The non-incremental path just copies Q to output without attention. This should throw `not_supported` if it's not implemented yet, rather than returning raw query vectors.

2. **`key_unpacked_ptr` null check bug** — When `c2p=true, p2c=false`, the NNTR_THROW_IF on `key_unpacked_ptr == nullptr` will always fire because that pointer is only set in the p2c code path. The check needs to be conditional on `p2c`.

3. **`SharedFullyConnectedLayer::supportBackwarding()` returns `true`** — But `calcDerivative()` and `calcGradient()` both throw. This should return `false`.

### Suggestions (non-blocking)

- The `lnode->getType() != "weight"` hardcoded check in 3 places in `network_graph.cpp` could be replaced with a virtual method for better extensibility.
- `openblas_set_num_threads(4)` sets global BLAS state — worth making configurable or documenting the rationale.
- Duplicate `causallm_qkv_layer_dep` in `meson.build`.
- Member variable naming (`c2p`/`p2c` vs `MAX_RELATIVE_POSITIONS`) could be more consistent.
- Exception handling in `setupParameters` swallows errors — partial initialization could cause subtle downstream issues.

Overall, this is solid work. The three items above (forwarding placeholder, null check bug, supportBackwarding) are the main things to fix before merge.

Applications/CausalLM/layers/shared_fully_connected_layer.h
Seunghui98: I follow other implementation for exportTo(). so, I think we should keep is as it is.

```cpp
void MHACoreLayer::exportTo(nntrainer::Exporter &exporter,
                            const ml::train::ExportMethods &method) const {
  LayerImpl::exportTo(exporter, method);
  exporter.saveResult(mha_core_props, method, this);
}
```

nntrainer/graph/network_graph.cpp
Seunghui98: > Not a blocker, but worth considering for long-term maintainability.

I agree with you. but, I think it's better to go as it is in this PR and address this in anther PR.

Applications/CausalLM/layers/deberta_attention_layer.cpp
Seunghui98: Good point!
The value 4 is an empirically chosen tuning value from local x86 environment.
I observed that this gave the best TPS performance in practice.

I agree that this is better handled as a global BLAS policy rather than inside a single layer, since OpenBLAS thread count is process-wide.
so, I will remove these!!!

nntrainer/models/neuralnet.cpp
Seunghui98: Thanks for letting me know!
I'll reflect #3810.

jijoongmoon (approved):
LGTM

baek2sm:
I've rebased it

PR #3774, opened 2026-03-04
[cpu_backend] Optimize softmax kernel

jaemini-shin:
### Summary of this PR:
* Rewrite softmax kernel for avx, neon
* Deprecate non-inplace softmax kernerl

### Optimizations
* Access array in row major order
Previous kernel accessed array in column major order, which increases cache miss rate.
* Vectorize code
Some logics didn't use simd even if it can be vectorized.
* Use reciprocal (FP32 only)
When calculating `exp(x-max)/sum`, direct dividing `sum` is slow. Instead, pre-calculate recipocal `1/sum` than multiplying it is much efficient.
* Deprecate non-inplace softmax kernel
Unlike its name, non-inplace kernel was designed to work same as inplace kernel. Its interface doesn't have destination parameter, in other words, it's inplace one. Furthermore, it calculates exp twice, which is the bottle neck in softmax.

### Future items
* Optimize exp logic
Since exp is the main bottle neck of softmax, there're still more room to optimize
  * reduce polynomial order
  * remove redundant logic
  For softmax, input is always non positive. Using this, we can remove overflow cutoff.
  * rewrite float16x8_t exp
  Currently, float16x8_t exp works as two float32x4_t exp logic sequentially. If we can make this logic with float16x8_t simd only, its latency and throughput would be highly increased.
  * apply loop unrolling
  
---
### Softmax latency Comparison
Qwen3-4b with 1K length input
| | Before this PR | After this PR | Ratio |
|:--:|:--:|:--:|:--:|
| avx prefill | 20989527.11 | 12423431.33 | 0.59 |
| avx decode | 471529.5056 | 422052.9592 | 0.90 |
| neon(fp16) prefill | 22316899.5 | 16427413.1388889 | 0.68 |
| neon(fp16) decode | 270753.7624 | 214167.623480903 | 0.79 |

---
### Update
Due to accuracy of FP16, use division instead of reciprocal approximation.
Following charts are the errors from the unittest.

* FP32
image

* FP16
image

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/neon_impl.cpp
@@ -1432,56 +1432,74 @@ void exp_i(const unsigned int N, float *X) {
 
 static void softmax_row_inplace(float *qk_out, size_t start_row, size_t end_row,
                                 size_t num_heads) {
-  size_t row_range = end_row - start_row;
-  const size_t full_blocks = (num_heads / 4) * 4;
-  // const size_t remainder = num_heads % 4;
+  const size_t vec_end = num_heads & ~((size_t)3); // floor(num_heads / 4) * 4
 
+  // 1. find max for each head
   float *max_vals = new float[num_heads];
+
+  // initialize max_vals with first row of qk_out
+  std::memcpy(max_vals, qk_out + start_row * num_heads,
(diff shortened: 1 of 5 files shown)


jijoongmoon (approved):
LGTM

jijoongmoon:
Could you check the accuracy ?

PR #3778, opened 2026-03-06
[avx2_impl]: fix out of boundary memory access

junbong-yu:
compute_fp16vcache_fp32_transposed should calculate j index of sequence row from 0 to num_row -1

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -1169,7 +1169,7 @@ void MHACoreLayer::compute_fp16vcache_transposed(
           float *out = output.getData<float>() +
                        i * (num_cache_head * gqa_size * head_dim);
 
-          int row_num = is_causal ? (to - seq + i) : to;
+          int row_num = is_causal ? (to - seq + i) : to - 1;
           nntrainer::compute_fp16vcache_fp32_transposed(
             row_num, input, vcache.getData<uint16_t>(), out, num_cache_head,
             gqa_size, head_dim, local_window_size);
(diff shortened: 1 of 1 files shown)


nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -1546,7 +1546,7 @@ void compute_fp16vcache_fp32_transposed(int row_num, const float *in,
 
     for (int j = row_num < local_window_size ? 0
                                              : row_num + 1 - local_window_size;
-         j <= row_num; ++j) {
+         j < row_num; ++j) {
```
jaemini-shin: `row_num` is the row index of vcache to calculate, so this logic is validate.
IMO, we should fix the caller side of this function.
https://github.com/nntrainer/nntrainer/blob/052a1962a8ab7116aa0126cb6153f7975548d881/Applications/CausalLM/layers/mha_core.cpp#L1172

junbong-yu: Thank you for your comment. I totally agree with your option, and fixed patches
now row_num for bidirectional QK*V is 'to - 1'

EunjuYang (approved):
Awesome! Thank you for your work 👍
Could you please check the CI so that we can merge this PR quickly?

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -1180,7 +1180,7 @@ void MHACoreLayer::compute_fp16vcache_transposed(
     } else {
       // Single token processing (common during generation)
       // Parallelize over KV heads for decoding since Q direction is always 1
-      int row_num = is_causal ? to - 1 : to;
+      int row_num = is_causal ? to - 1 : to - 1;
```
Jungwon-Lee, line 1183: Is this intended??
`int row_num = to - 1;`

junbong-yu: Thank you for your review, I didn't catch that. it looks better to remove redundant code

PR #3789, opened 2026-03-11
[compute_fp16vcache_transposed] fix for FP16 && ANDROID

junbong-yu:
apply row_num fix path for FP16 && ANDROID code block as well

Signed-off-by: junbong.yu <junbong.yu@samsung.com>

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -1180,7 +1180,7 @@ void MHACoreLayer::compute_fp16vcache_transposed(
     } else {
       // Single token processing (common during generation)
       // Parallelize over KV heads for decoding since Q direction is always 1
-      int row_num = is_causal ? to - 1 : to - 1;
+      int row_num = to - 1;
 
       // Use OpenMP for lower overhead parallelization during decoding
       const float *in_data = in.getData<float>();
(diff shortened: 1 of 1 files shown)


Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -1180,7 +1180,7 @@ void MHACoreLayer::compute_fp16vcache_transposed(
     } else {
       // Single token processing (common during generation)
       // Parallelize over KV heads for decoding since Q direction is always 1
-      int row_num = is_causal ? to - 1 : to - 1;
```
junbong-yu, line 1183: Removing redundant according to previous review comment
https://github.com/nntrainer/nntrainer/pull/3778#discussion_r2903629803

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -1216,7 +1216,7 @@ void MHACoreLayer::compute_fp16vcache_transposed(
             in.getData<_FP16>() + start_idx * num_cache_head * gqa_size;
           _FP16 *out = output.getData<_FP16>() +
                        i * (num_cache_head * gqa_size * head_dim);
-          int row_num = is_causal ? (to - seq + i) : to;
+          int row_num = is_causal ? (to - seq + i) : to - 1;
```
jaemini-shin, line 1219:
```suggestion
          int row_num = is_causal ? (from + i) : to - 1;
```
(optional) This change would be easier to understand this logic.

junbong-yu: Good point, but is this safe even if window size is smaller than (from- to)?
if window size is smaller than (to - from), then 'seq' is setup to window size, which means from value is not the starting point of calculation anymore.

jaemini-shin: Oh, I forgot about local window. We should leave the current logic! :+1:

PR #3797, opened 2026-03-16
nntrainer-visualizer-extension

prachi-t17:
nntrainer-visualizer 

Signed-off-by: prachi.t : prachi.t@samsung.com


tools/nntrainer-visualizer/README.md
```diff
@@ -0,0 +1,26 @@
+
+#NNTrainer Visualiser
+
+a vscode extension for visualizing NNTrainer architecture and weights.
+
+How to install
+
+1. Install Node.js
```
Seunghui98, line 8: ## Note

- `node.js`'s version seems to be important. 
- If we install a low version with 'sudo apt install npm' (maybe.. under 20?), it seems that the dependency problem occurs in the process of installing @vscode/vsce.

PR #3799, opened 2026-03-16
[layers] add skip prefill option

baek2sm:
- added skip prefill option to some layers

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -149,6 +149,10 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
   /** Is Causal */
   is_causal = std::get<props::IsCausal>(mha_core_props).get();
 
+  if (!std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).empty())
+    skip_prefill =
+      std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).get();
+
   /** Tensor for KV-Cache */
 #ifdef ENABLE_FP16
   ml::train::TensorDim cache_key_dim(
@@ -516,8 +520,6 @@ void MHACoreLayer::one_batch_incremental_forwarding(
     batch * cache_value_dim.getFeatureLen() + from * cache_value_dim.width(),
     true);
 
-  apply_rotary_emb_tensor_v2(query_step, query_step, head_dim, _from, false);
-
   apply_rotary_emb_tensor_v2(key_step, b_cache_key_step, head_dim, _from,
                              false);
 
@@ -531,6 +533,11 @@ void MHACoreLayer::one_batch_incremental_forwarding(
     NNTR_THROW_IF(true, std::invalid_argument) << "enable-fp16 is not set!";
 #endif
   }
+  bool is_prefill = !from;
+  if (skip_prefill && is_prefill)
+    return;
+
+  apply_rotary_emb_tensor_v2(query_step, query_step, head_dim, _from, false);
 
   ml::train::TensorDim cached_key_dim = cache_key_dim;
   ml::train::TensorDim cached_value_dim = cache_value_dim;
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -153,7 +153,7 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
     skip_prefill =
       std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).get();
 
-  /** Tensor for KV-Cache */
+    /** Tensor for KV-Cache */
 #ifdef ENABLE_FP16
   ml::train::TensorDim cache_key_dim(
     {batch_size, 1, max_timestep, num_heads_KV * head_dim},
--- a/Applications/CausalLM/layers/rms_norm.cpp
+++ b/Applications/CausalLM/layers/rms_norm.cpp
@@ -24,8 +24,8 @@ void RMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
   std::vector<nntrainer::TensorDim> dim = context.getInputDimensions();
   context.setOutputDimensions(dim);
 
-  if (!std::get<props::SkipPrefill>(rms_props).empty())
-    skip_prefill = std::get<props::SkipPrefill>(rms_props).get();
+  if (!std::get<nntrainer::props::SkipPrefill>(rms_props).empty())
+    skip_prefill = std::get<nntrainer::props::SkipPrefill>(rms_props).get();
 
   nntrainer::TensorDim gamma_dim(
     1, 1, 1, dim[0].width(),
--- a/Applications/CausalLM/layers/rms_norm.cpp
+++ b/Applications/CausalLM/layers/rms_norm.cpp
@@ -23,6 +23,10 @@ static constexpr size_t SINGLE_INOUT_IDX = 0;
 void RMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
   std::vector<nntrainer::TensorDim> dim = context.getInputDimensions();
   context.setOutputDimensions(dim);
+
+  if (!std::get<props::SkipPrefill>(rms_props).empty())
+    skip_prefill = std::get<props::SkipPrefill>(rms_props).get();
+
   nntrainer::TensorDim gamma_dim(
     1, 1, 1, dim[0].width(),
     nntrainer::TensorDim::TensorType(context.getFormat(),
@@ -50,7 +54,9 @@ void RMSNormLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   ml::train::TensorDim in_step_dim = in_dim;
   ml::train::TensorDim out_step_dim = out_dim;
 
-  unsigned int _from = from;
+  bool is_prefill = !from;
+  if (skip_prefill && is_prefill)
+    return;
 
   in_step_dim.batch(1);
   in_step_dim.height(to - from);
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -62,6 +62,10 @@ void CausalLM::setupParameters(json &cfg, json &generation_cfg,
                    ? nntr_cfg["lmhead_dtype"]
                    : nntr_cfg["embedding_dtype"];
 
+  SKIP_PREFILL = nntr_cfg.contains("skip_prefill")
+                   ? nntr_cfg["skip_prefill"].get<bool>()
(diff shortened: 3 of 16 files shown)


Applications/CausalLM/layers/rms_norm.cpp
```diff
@@ -50,7 +54,9 @@ void RMSNormLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   ml::train::TensorDim in_step_dim = in_dim;
   ml::train::TensorDim out_step_dim = out_dim;
 
-  unsigned int _from = from;
+  bool is_prefill = !from;
+  if (skip_prefill && is_prefill)
+    return;
```
EunjuYang, line 60: **`from == 0` does not reliably mean "prefill".** This is a convention in the current codebase, but not a guaranteed contract of the `incremental_forwarding` interface.

Consider these scenarios where `from == 0` but it is **not** a prefill:
- A fresh conversation turn that starts inference from position 0
- A model reset that re-runs from the beginning
- Unit tests or debugging tools that call `incremental_forwarding(context, 0, 1, false)` for a single-token forward pass

In all of these cases, `!from` evaluates to `true`, and the layer would incorrectly skip its computation, leaving the output tensor uninitialized.

A more robust approach would be to pass an explicit `is_prefill` flag (or a phase/mode enum) through the context or as a parameter, rather than inferring it from positional arguments. If that's too invasive for now, at minimum the assumption should be clearly documented at the `SkipPrefill` property definition and at every skip site, so future developers understand the constraint.

---

nntrainer/layers/fc_layer.cpp
```diff
@@ -245,6 +247,10 @@ void FullyConnectedLayer::incremental_forwarding(RunLayerContext &context,
   Tensor &hidden_ = context.getOutput(SINGLE_INOUT_IDX);
   Tensor loraA, loraB, hidden_tmp_lora, hidden_out_lora;
 
+  bool is_prefill = !from;
+  if (skip_prefill && is_prefill)
+    return;
+
```
EunjuYang, line 253: When `skip_prefill && is_prefill` is true, this function returns without writing anything to the output tensor. The output buffer retains whatever stale or uninitialized data was there.

This is only safe if **every** downstream layer that consumes this output also skips its prefill. If any single layer in the chain does not have `skip_prefill=true`, it will read uninitialized memory as input, producing silently wrong results.

There is currently no validation that ensures all layers in the path are configured consistently. Consider either:
1. Filling the output with zeros (or NaN in debug builds) so corruption is detectable, or
2. Adding a model-level check during construction/compile that verifies skip_prefill consistency across the layer graph.

---

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -149,7 +149,11 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
   /** Is Causal */
   is_causal = std::get<props::IsCausal>(mha_core_props).get();
 
-  /** Tensor for KV-Cache */
+  if (!std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).empty())
+    skip_prefill =
+      std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).get();
+
+    /** Tensor for KV-Cache */
```
EunjuYang, line 156: Indentation issue: the `/** Tensor for KV-Cache */` comment is indented as if it's inside the `if` block, but the `if` has no braces, so it's actually unconditional code. This is visually misleading.

```cpp
  if (!std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).empty())
    skip_prefill =
      std::get<nntrainer::props::SkipPrefill>(*layer_impl_props).get();

  /** Tensor for KV-Cache */   // ← should align here
```

---

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -531,6 +533,11 @@ void MHACoreLayer::one_batch_incremental_forwarding(
     NNTR_THROW_IF(true, std::invalid_argument) << "enable-fp16 is not set!";
 #endif
   }
+  bool is_prefill = !from;
+  if (skip_prefill && is_prefill)
+    return;
+
+  apply_rotary_emb_tensor_v2(query_step, query_step, head_dim, _from, false);
```
EunjuYang, line 540: The early return here has a specific (and correct) design: key and value are already written to the KV cache above, and only the query rotary embedding + attention computation are skipped. This preserves KV cache integrity while avoiding the expensive attention score computation.

However, the output tensor (`attention_output_step`) is left unwritten. It would be helpful to add a brief comment here explaining this is intentional and that all downstream layers are expected to also skip during prefill:

```cpp
  if (skip_prefill && is_prefill)
    // KV cache is populated above. Skip attention computation — downstream
    // layers must also have skip_prefill=true as output is not written.
    return;
```

---

nntrainer/layers/common_properties.h
```diff
@@ -125,6 +125,21 @@ class TensorDimension : public TensorDimProperty {
...
+class SkipPrefill : public nntrainer::Property<bool> {
+public:
+  /**
+   * @brief Construct a new SkipPrefill object
+   *
+   */
+  SkipPrefill() : nntrainer::Property<bool>() {}
+  static constexpr const char *key = "skip_prefill";
+  using prop_tag = bool_prop_tag;
+};
```
EunjuYang, line 141: The `SkipPrefill` property description is too brief. Since this feature has non-obvious safety requirements, the documentation should mention:
1. What "prefill" means in this context (currently inferred as `from == 0`)
2. That skipped layers leave their output tensor unwritten
3. That **all** layers in the forward path must have this property set consistently — a partial configuration leads to silent data corruption

This is important for anyone configuring models with this option in the future.

---

nntrainer/layers/layer_impl.cpp
```diff
@@ -27,8 +27,8 @@ LayerImpl::LayerImpl() :
     std::make_unique<
       std::tuple<props::WeightRegularizer, props::WeightRegularizerConstant,
                  props::WeightInitializer, props::WeightDecay, props::BiasDecay,
-                 props::BiasInitializer, props::DisableBias, props::Print>>()) {
-}
+                 props::BiasInitializer, props::DisableBias, props::Print,
+                 props::SkipPrefill>>()) {}
```
EunjuYang, line 31: Adding `SkipPrefill` to `LayerImpl`'s base property tuple means every `LayerImpl`-derived layer (embedding, loss, addition, concat, etc.) now silently accepts the `skip_prefill` property, even though only a handful of layers actually implement the skip logic.

This isn't a correctness issue (unused properties are just ignored), but it widens the API surface unnecessarily. If a user sets `skip_prefill=true` on a layer that doesn't implement it, nothing happens — which could be confusing to debug.

An alternative would be to only add `SkipPrefill` to the layers that actually use it (as was done for RMSNorm and SwiGLU which have their own props tuples). For FC and MHA that inherit from LayerImpl, it could be added to their layer-specific props instead.

---

EunjuYang (changes requested):
## Review Summary

The optimization idea is sound — during autoregressive prefill, if the only purpose is to populate the KV cache and the next token is already known, skipping the attention output computation, FFN, and norm layers can save significant compute. The MHA layer correctly preserves KV cache integrity by writing K/V before the skip point.

### Main concern: `from == 0` is not a reliable prefill indicator

The core assumption `bool is_prefill = !from;` equates "position 0" with "prefill phase", but these are not the same thing. `from == 0` can occur in non-prefill contexts:
- A fresh conversation turn starting inference from position 0
- Model reset scenarios
- Single-token forward passes starting at position 0 (e.g., in tests)

In all these cases, the layer would incorrectly skip computation. A more robust solution would be to pass an explicit prefill flag through the context or as a parameter, rather than inferring it from positional arguments. See inline comment on `rms_norm.cpp` for details.

### Uninitialized output tensors

When a layer skips, its output tensor is left unwritten. This is only safe if **every** layer in the forward chain also skips. There is no validation that ensures this consistency — a single misconfigured layer would silently read garbage from the previous layer's unwritten output buffer. Consider adding either:
- A model-level consistency check during construction
- Defensive fills (zeros or NaN) on skip for debuggability

### Other items
- **Indentation bug** in `mha_core.cpp:152-156` — comment appears to be inside the `if` block but is actually outside it
- **`SkipPrefill` added to `LayerImpl` base** — this exposes the property to all `LayerImpl`-derived layers, even those that don't implement skip logic. Users who set it on non-implementing layers get no error and no effect, which could be confusing.
- **`SkipPrefill` documentation** is too brief — should document the `from == 0` assumption, uninitialized output behavior, and the requirement for consistent configuration across all layers in the path.
- **No tests** for the skip_prefill feature.

### Positive notes
- The MHA skip point is well-placed: KV cache is populated before the early return, preserving cache integrity.
- The `causal_lm.cpp` N-1 token handling is clean and well-commented.
- Adding `SkipPrefill` as a property (rather than hardcoding) keeps the feature configurable per deployment.

---

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -437,16 +441,37 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   } else {
     SYS_PROMP_LEN = 0;
   }
-  output = model->incremental_inference(BATCH_SIZE, input, label, init_len,
-                                        SYS_PROMP_LEN,
-                                        SYS_PROMP_LEN + input_len, false);
+  std::vector<unsigned int> id_list;
+
+  if (SKIP_PREFILL && init_len > 1) {
```
jaemini-shin, line 446: Checking `init_len > 1` should not be the only condition for skip prefill option. How about checking it unconditionally?

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -437,16 +441,37 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
...
-  std::vector<unsigned int> id_list(generate_multi_tokens(
-    output[0], NUM_VOCAB, BATCH_SIZE, 1, ids_history, _len));
+    output = model->incremental_inference(BATCH_SIZE, input, label,
+                                          init_len - 1, SYS_PROMP_LEN,
+                                          SYS_PROMP_LEN + input_len - 1, false);
 
-  if (init_len < INIT_SEQ_LEN)
-    registerOutputs(tokenizer, id_list, init_len, eos_list);
+    for (unsigned int b = 0; b < BATCH_SIZE; ++b)
+      id_list.push_back(skipped_token);
```
jaemini-shin, line 457:
```suggestion
    for (unsigned int b = 0; b < BATCH_SIZE; ++b)
      id_list.push_back(init_input[b * init_len - 1]);
```
`skipped_token` only works for single batch.

jaemini-shin (commented):
Could you please check following comments?

PR #3812, opened 2026-03-24
[GPU Backend] OpenCL Kernel for Flash Attention

anupSnap21530358:
## Changes
- Adreno compatible Flash Attention OpenCL kernel
- Implementation added for both half and full precision data type

## Testing
- Unit testcase added for both fp16 and fp32


nntrainer/tensor/cl_operations/cl_kernels/flash_attention.cl
```diff
@@ -0,0 +1,134 @@
...
+      }
+      output[out_offset + d] = result;
+    }
+  } else {
+    // Slow path for larger sequences - compute without shared memory
+    float exp_sum = 0.0f;
+    float scores[512]; // Temporary array for scores
+    
+    // First pass: compute scores and find maximum, and store scores
+    for (int key_idx = 0; key_idx < seq_len && key_idx < 512; key_idx++) {
```
EunjuYang, line 93: When `seq_len > 512`, this loop silently stops at index 512:
```c
for (int key_idx = 0; key_idx < seq_len && key_idx < 512; key_idx++) {
```
Keys beyond position 512 are ignored, which means softmax is computed over a partial set of scores and the value-weighted sum is also incomplete. This will produce incorrect results without any error indication.

Could you either:
1. Add a runtime check that returns an error / logs a warning when `seq_len > 512`, or
2. Implement a true chunked/tiled approach so arbitrary sequence lengths are handled correctly?

nntrainer/tensor/cl_operations/cl_kernels/flash_attention.cl
```diff
@@ -0,0 +1,134 @@
...
+  }
+  
+  // Calculate base offsets for tensors
+  const int qkv_offset = (batch_idx * num_heads + head_idx) * seq_len * head_dim;
+  const int mask_offset = batch_idx * seq_len * seq_len;
+  const int out_offset = qkv_offset + query_idx * head_dim;
+  
+  // Load query vector for this position
+  float local_query[256]; // Increased to handle larger head_dim
+  for (int d = 0; d < head_dim; d++) {
```
EunjuYang, line 32: `local_query[256]` is allocated in private memory per work-item, but there's no bounds check that `head_dim <= 256`. If `head_dim > 256`, this will write out of bounds. The same applies to `shared_scores[256]` (assumes `seq_len <= 256`) and `scores[512]`.

It would be helpful to either:
- Add compile-time or runtime guards, or
- Pass these sizes as kernel compile-time defines (e.g., `-D HEAD_DIM=128`) so the arrays can be sized appropriately.

nntrainer/tensor/cl_operations/cl_kernels/flash_attention.cl
EunjuYang, line 1: A note on the algorithm: the current implementation computes the full attention score row per query position (Q·K^T → softmax → ×V) in a single work-item. This is functionally correct for bounded sequence lengths, but it differs from the Flash Attention algorithm described in Dao et al. (2022), which processes Q/K/V in tiles using shared (local) memory with an online softmax to avoid materializing the full N×N attention matrix.

The current approach is closer to a **standard (naive) attention** kernel. The key characteristics of Flash Attention that are missing here:
- **Tiled computation** over K/V blocks with `__local` memory
- **Online softmax** — incrementally maintaining running max and exp_sum across tiles
- **O(N) SRAM usage** instead of O(N²)

This is totally fine as a starting point — just wanted to flag the naming so expectations are aligned. You might consider naming this `attention_cl` / `standard_attention_cl` for now, and reserving the "flash attention" name for when the tiled algorithm is implemented.

nntrainer/tensor/cl_operations/flash_attention_kernel.cpp
```diff
@@ -0,0 +1,229 @@
...
+    if (!result) {
+      printf("Failed to set scale argument\n");
+      break;
+    }
+
+    // Calculate work group sizes
+    const int work_groups_count[3] = {batch_size, num_heads, seq_len};
+    const int work_group_size[3] = {1, 1, 1};
+    
+    result = cl_context->command_queue_inst_.DispatchCommand(
```
EunjuYang, line 155: With work group size `{1, 1, 1}`, each work-item independently processes one (batch, head, query_position) tuple entirely on its own. A few observations:

1. **No local memory sharing** — K/V values are re-read from global memory per work-item. Using `__local` memory to share K/V tiles across work-items in a work group could significantly reduce global memory traffic.
2. **Non-coalesced memory access** — The inner loop reads K and V with a stride of `head_dim`, which is unfriendly to GPU memory coalescing.
3. **High private memory per work-item** — `local_query[256]` + `scores[256 or 512]` = up to 3KB per work-item in registers, which may cause register spilling on some devices.

These are potential follow-up optimizations, but worth keeping in mind — especially for Adreno where local memory and coalesced access patterns matter for performance.

nntrainer/tensor/cl_operations/flash_attention_kernel.cpp
```diff
@@ -0,0 +1,229 @@
...
+    // Write data to buffers
+    result = cl_buffer_manager.getInBufferA()->WriteDataRegion(
+      cl_context->command_queue_inst_, query_size, query);
+    if (!result) {
+      printf("Failed to write query data\n");
+      break;
+    }
+
+    result = cl_buffer_manager.getInBufferB()->WriteDataRegion(
+      cl_context->command_queue_inst_, key_size, key);
```
EunjuYang, line 56: The project uses `ml_loge()` / `ml_logw()` for error logging elsewhere. Using `printf()` here is inconsistent and the output may not be visible in all deployment contexts. Consider switching to the project's logging macros.

Also, on failure, the function silently returns without propagating the error to the caller. It might be worth returning a status code or throwing an exception so the caller knows the operation failed.

nntrainer/tensor/cl_operations/flash_attention_kernel.cpp
```diff
@@ -0,0 +1,229 @@
...
+    static_cast<ClContext *>(Engine::Global().getRegisteredContext("gpu"));
+  auto &cl_buffer_manager = ClBufferManager::Global();
+
+  do {
+    // Calculate buffer sizes
+    size_t query_size = batch_size * num_heads * seq_len * head_dim * sizeof(T);
+    size_t key_size = batch_size * num_heads * seq_len * head_dim * sizeof(T);
+    size_t value_size = batch_size * num_heads * seq_len * head_dim * sizeof(T);
+    size_t output_size = batch_size * num_heads * seq_len * head_dim * sizeof(T);
+    size_t mask_size = attention_mask ?
```
EunjuYang, line 44: These intermediate multiplications (`batch_size * num_heads * seq_len * head_dim`) are all `int * int * int * int`, which could overflow for larger inputs before being assigned to `size_t`. Consider casting at least the first operand:
```cpp
size_t query_size = static_cast<size_t>(batch_size) * num_heads * seq_len * head_dim * sizeof(T);
```

nntrainer/tensor/cl_operations/cl_kernels/flash_attention_fp16.cl
EunjuYang, line 1: This FP16 kernel is nearly identical to the FP32 version — the only differences are the input/output types and the `(float)` / `(half)` casts. To reduce maintenance burden, you could consider:
- Using a single `.cl` file with preprocessor macros (`#ifdef USE_FP16`) to select the type, or
- Generating both variants from a shared template

This way, future bug fixes or algorithmic changes only need to be made in one place.

test/unittest/unittest_flash_attention_kernels_cl.cpp
```diff
@@ -0,0 +1,300 @@
...
+    value.setRandUniform(0, 1);
+    attention_mask.setRandUniform(0, 1);
+    
+    // Copy data for CPU computation
+    output_cpu.copy(output_gpu);
+    
+    // Time GPU execution
+    double gpu_time = time_execution([&]() {
+      flash_attention_cl(query, key, value, output_gpu, &attention_mask,
+                         batch, num_heads, seq_len, head_dim, scale);
```
EunjuYang, line 170: Nit: `alpha` and `MOD` are declared but never used in both FP32 and FP16 test cases. The compiler may emit warnings for these.

test/unittest/unittest_flash_attention_kernels_cl.cpp
```diff
@@ -0,0 +1,300 @@
...
+    int batch = test_case.batch_size;
+    int num_heads = test_case.num_heads;
+    int seq_len = test_case.seq_len;
+    int head_dim = test_case.head_dim;
+    float scale = test_case.scale;
+    
+    nntrainer::TensorDim::TensorType t_type_nchw_fp32 = {
+      nntrainer::Tformat::NCHW, nntrainer::Tdatatype::FP32};
+    
+    // Create tensors
```
EunjuYang, line 147: It would be valuable to add a test case with `seq_len > 512` to verify correctness (or surface the current limitation). Also, a test with `head_dim > 256` would help catch the array bounds issue in the kernel.

Consider also adding a test case without an attention mask (`nullptr`) to exercise that code path.

nntrainer/tensor/cl_operations/cl_kernels/flash_attention.cl
```diff
@@ -0,0 +1,134 @@
+#pragma OPENCL EXTENSION cl_khr_fp16 : enable
```
EunjuYang, line 1: Minor: The FP32 kernel doesn't use FP16 types, so this pragma is unnecessary here. It's only needed in the FP16 variant.

EunjuYang (changes requested):
## Review Summary

Thanks for adding an OpenCL attention kernel with both FP32 and FP16 support — having GPU-accelerated attention is a great direction for nntrainer. I've left some comments on specific areas. Here's a high-level summary:

### Correctness
- **Sequence length > 512 produces silent incorrect results** — the "slow path" loop caps at 512 iterations, so keys beyond that position are ignored without any error. This is the most important item to address.
- **head_dim > 256 causes out-of-bounds writes** — `local_query[256]` has no bounds check. Same pattern for the score arrays.

### Naming
- The current implementation computes full Q·K^T scores per query position, applies softmax, then multiplies by V — this is standard (naive) attention. Flash Attention (Dao et al., 2022) specifically refers to a tiled algorithm with online softmax and O(N) SRAM usage, which avoids materializing the full N×N matrix. The kernel is functionally valid, but the naming may set inaccurate expectations. Consider renaming to `attention_cl` or `standard_attention_cl` for clarity.

### Performance opportunities (for follow-up)
- Work group size is `{1, 1, 1}` with no `__local` memory usage — there's room for significant speedup via tiling and memory coalescing
- High private memory usage per work-item (up to ~3-4KB) may cause register spilling

### Minor items
- `printf()` should be replaced with project logging macros (`ml_loge`)
- Integer overflow risk in buffer size calculations
- FP32/FP16 kernel files are nearly identical — could be unified with preprocessor macros
- Unused variables (`alpha`, `MOD`) in tests
- Missing test cases for `seq_len > 512`, `head_dim > 256`, and `nullptr` attention mask
- Missing newline at end of several files

The kernel logic itself (score computation → softmax → weighted sum) is correct within the supported bounds. Addressing the silent truncation for large sequences and the array bounds issues would make this a solid foundation to build on.

anupSnap21530358:
> ## Review Summary
> Thanks for adding an OpenCL attention kernel with both FP32 and FP16 support — having GPU-accelerated attention is a great direction for nntrainer. I've left some comments on specific areas. Here's a high-level summary:
> 
> ### Correctness
> * **Sequence length > 512 produces silent incorrect results** — the "slow path" loop caps at 512 iterations, so keys beyond that position are ignored without any error. This is the most important item to address.
> * **head_dim > 256 causes out-of-bounds writes** — `local_query[256]` has no bounds check. Same pattern for the score arrays.
> 
> ### Naming
> * The current implementation computes full Q·K^T scores per query position, applies softmax, then multiplies by V — this is standard (naive) attention. Flash Attention (Dao et al., 2022) specifically refers to a tiled algorithm with online softmax and O(N) SRAM usage, which avoids materializing the full N×N matrix. The kernel is functionally valid, but the naming may set inaccurate expectations. Consider renaming to `attention_cl` or `standard_attention_cl` for clarity.
> 
> ### Performance opportunities (for follow-up)
> * Work group size is `{1, 1, 1}` with no `__local` memory usage — there's room for significant speedup via tiling and memory coalescing
> * High private memory usage per work-item (up to ~3-4KB) may cause register spilling
> 
> ### Minor items
> * `printf()` should be replaced with project logging macros (`ml_loge`)
> * Integer overflow risk in buffer size calculations
> * FP32/FP16 kernel files are nearly identical — could be unified with preprocessor macros
> * Unused variables (`alpha`, `MOD`) in tests
> * Missing test cases for `seq_len > 512`, `head_dim > 256`, and `nullptr` attention mask
> * Missing newline at end of several files
> 
> The kernel logic itself (score computation → softmax → weighted sum) is correct within the supported bounds. Addressing the silent truncation for large sequences and the array bounds issues would make this a solid foundation to build on.

Hello @EunjuYang Thanks for the review. Please have a look at this PR (https://github.com/nntrainer/nntrainer/pull/3817/) which contains the most efficient implementation for flash attention. Please review this PR and provide your feedback. Both fp32 and fp16 kernel for flash attention is added.

PR #3813, opened 2026-03-24
MeZO training pipeline with an example of a small fc layer model on MNIST

sachin-nntrainer:
This PR is an updated code of #3762. It contains the suggested changes. However, I faced a few issues with the code design. 

1. I'm using opt->updateWeightsMeZO(param_ptrs, seed, loss_plus, loss_minus) in 11c5d459dbd60034182c43370795c6dcbcfc401a to updates weights. I feel this does not go well with nntrainer's current logic for updating weights (like backprop). Maybe i should implement something similar to backwarding_op for MeZO and update weights through  RunLayerContext.
2. I faced an error (invalid arg) will setting cross_softmax as loss function. I am yet to debug this. For DEMO purpose i have set MSE as loss function. 

Requesting the reviewer's opinion on this withstanding issues and current changes made to the code. 

Thankyou!

### DIFF
--- a/Applications/MeZO/jni/main.cpp
+++ b/Applications/MeZO/jni/main.cpp
@@ -1,0 +1,291 @@
+#include <algorithm>
+#include <arpa/inet.h>
+#include <climits>
+#include <cmath>
+#include <dataset.h>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <memory>
+#include <model.h>
+#include <nntrainer_log.h>
+#include <numeric>
+#include <optimizer.h>
+#include <random>
+#include <util_func.h>
+#include <vector>
+
+using namespace nntrainer;
+
+using LayerHandle = std::shared_ptr<ml::train::Layer>;
+using ModelHandle = std::unique_ptr<ml::train::Model>;
+
+// User data structure for dataset callback
+struct UserData {
+  std::vector<float> data;
+  std::vector<float> labels;
+  unsigned int num_samples;
+  unsigned int current_index;
+  unsigned int feature_size;
+  unsigned int label_size;
+  std::vector<unsigned int> indices;
+  std::mt19937 rng;
+
+  UserData(const std::vector<float> &d, const std::vector<float> &l,
+           unsigned int fs, unsigned int ls) :
+    data(d),
+    labels(l),
+    num_samples(d.size() / fs),
+    current_index(0),
+    feature_size(fs),
+    label_size(ls),
+    indices(num_samples),
+    rng(0) {
+    std::iota(indices.begin(), indices.end(), 0);
+    std::shuffle(indices.begin(), indices.end(), rng);
+  }
+};
+
+// Dataset callback function
+int getSample(float **outVec, float **outLabel, bool *last, void *user_data) {
+  UserData *ud = static_cast<UserData *>(user_data);
+
+  if (ud->current_index >= ud->num_samples) {
+    // Reset and shuffle indices for next epoch
+    ud->current_index = 0;
+    std::shuffle(ud->indices.begin(), ud->indices.end(), ud->rng);
+  }
+
+  unsigned int idx = ud->indices[ud->current_index];
+
+  // Copy input data
+  std::copy(ud->data.begin() + idx * ud->feature_size,
+            ud->data.begin() + (idx + 1) * ud->feature_size, *outVec);
+
+  // Copy label data
+  std::copy(ud->labels.begin() + idx * ud->label_size,
+            ud->labels.begin() + (idx + 1) * ud->label_size, *outLabel);
+
+  ud->current_index++;
+
+  *last = (ud->current_index >= ud->num_samples);
+
+  return 0;
+}
+
+static uint32_t read_u32_be(std::ifstream &ifs) {
+  uint32_t val = 0;
+  ifs.read(reinterpret_cast<char *>(&val), sizeof(val));
+  return ntohl(val);
+}
+
+// Function to load raw MNIST data
+std::pair<std::vector<float>, std::vector<float>>
+loadMNISTData(const std::string &data_file, const std::string &label_file) {
+
+  std::vector<float> data;
+  std::vector<float> labels;
+  static constexpr float MNIST_MEAN = 0.1307f;
+  static constexpr float MNIST_STD = 0.3081f;
+
+  std::ifstream ifs_images(data_file, std::ios::binary);
+  if (!ifs_images.is_open()) {
+    ml_loge("Failed to open MNIST images file: %s", data_file.c_str());
+    return {data, labels};
+  }
+
+  uint32_t magic_images = read_u32_be(ifs_images);
+  uint32_t num_images = read_u32_be(ifs_images);
+  uint32_t rows = read_u32_be(ifs_images);
+  uint32_t cols = read_u32_be(ifs_images);
+
+  if (magic_images != 2051) {
+    ml_loge("Invalid MNIST images magic: %u", magic_images);
+    return {data, labels};
+  }
+
+  std::ifstream ifs_labels(label_file, std::ios::binary);
+  if (!ifs_labels.is_open()) {
+    ml_loge("Failed to open MNIST labels file: %s", label_file.c_str());
+    return {data, labels};
+  }
+
+  uint32_t magic_labels = read_u32_be(ifs_labels);
+  uint32_t num_labels = read_u32_be(ifs_labels);
+
+  if (magic_labels != 2049) {
+    ml_loge("Invalid MNIST labels magic: %u", magic_labels);
+    return {data, labels};
+  }
+
+  if (num_images != num_labels) {
+    ml_loge("Images/labels count mismatch: %u vs %u", num_images, num_labels);
+    return {data, labels};
+  }
+
+  size_t image_size = static_cast<size_t>(rows) * static_cast<size_t>(cols);
+  data.reserve((size_t)num_images * image_size);
+  labels.reserve((size_t)num_images * 10); // 10 classes for one-hot encoding
+
+  // Read images with normalization
+  for (uint32_t i = 0; i < num_images; ++i) {
+    std::vector<unsigned char> buf(image_size);
+    ifs_images.read(reinterpret_cast<char *>(buf.data()), image_size);
+    if (!ifs_images) {
+      ml_loge("Error reading image %u", i);
+      return {data, labels};
+    }
+    for (size_t p = 0; p < image_size; ++p) {
+      // Normalize: (pixel / 255.0 - mean) / std
+      float pixel = static_cast<float>(buf[p]) / 255.0f;
+      float normalized = (pixel - MNIST_MEAN) / MNIST_STD;
+      data.push_back(normalized);
+    }
+  }
+
+  // Read labels and convert to one-hot encoding
+  for (uint32_t i = 0; i < num_labels; ++i) {
+    unsigned char lab = 0;
+    ifs_labels.read(reinterpret_cast<char *>(&lab), 1);
+    if (!ifs_labels) {
+      ml_loge("Error reading label %u", i);
+      return {data, labels};
+    }
+
+    // Convert to one-hot encoding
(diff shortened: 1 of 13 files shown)


nntrainer/optimizers/MeZO.cpp
```diff
@@ -0,0 +1,59 @@
...
+  for (const auto *ptr : weights) {
+
+    float *data = (float *)(ptr->getData());
+    size_t param_size = ptr->getDim().getDataLen();
+
+    for (size_t i = 0; i < param_size; ++i) {
+
+      data[i] += lr * normal_dist(gen) * projected_grad;
+    }
+  }
```
EunjuYang, line 56: **Bug (Critical):** The update direction is inverted. According to the MeZO paper, the weight update rule is:

```
θ = θ - lr * projected_grad * z
```

But this line does `data[i] += lr * z * projected_grad`, which performs gradient **ascent** (increasing the loss) instead of gradient **descent**.

This should be:
```cpp
data[i] -= lr * normal_dist(gen) * projected_grad;
```

This likely explains why convergence may be poor or non-existent.

nntrainer/models/neuralnet.cpp
```diff
@@ -1370,10 +1375,35 @@ int NeuralNetwork::train_run(
...
+        perturbParameters(-2 * mezo_epsilon, seed);
+        forwarding(true, stop_cb, stop_user_data);
+        float loss_minus = getLoss();
+
+        perturbParameters(mezo_epsilon, seed);
+        float projected_grad = (loss_plus - loss_minus) / (2.0f * mezo_epsilon);
+
+        auto param_ptrs = getParameterPointers();
+        opt->updateWeightsMeZO(param_ptrs, seed, loss_plus, loss_minus);
+
```
EunjuYang, line 1401: `projected_grad` is computed here but never used — `updateWeightsMeZO()` recomputes it internally from `loss_plus` and `loss_minus`. This is dead code and should be removed.

nntrainer/models/neuralnet.cpp
```diff
@@ -1370,10 +1375,35 @@ int NeuralNetwork::train_run(
...
+      } else {
+        ml_logi("train for iteration");
+        forwarding(true, stop_cb, stop_user_data);
+        backwarding(iter++, stop_cb, stop_user_data);
+      }
       // To avoid unconsidered memory leak, we need to clear the cache
```
EunjuYang, line 1407: **Design concern:** Embedding the entire MeZO training loop inside `NeuralNetwork::train_run()` with a type-check branch tightly couples the model to a specific optimizer. Every new gradient-free optimizer (e.g., CMA-ES, evolutionary strategies) would require modifying `NeuralNetwork` again.

**Suggested approach:** Consider adding a `trainStep()` virtual method to the `Optimizer` interface:

```cpp
// In Optimizer base class
virtual bool requiresBackprop() const { return true; }

virtual void trainStep(
  std::function<void()> forward_fn,
  std::function<float()> get_loss_fn,
  std::vector<Tensor*> &params
) { /* default: no-op, backprop optimizers use existing path */ }
```

Then MeZO overrides this to encapsulate all its logic (perturbation, dual forward passes, weight update) inside the optimizer class. `NeuralNetwork::train_run()` simply checks `requiresBackprop()`:

```cpp
if (opt->requiresBackprop()) {
  forwarding(...);
  backwarding(iter++, ...);
} else {
  auto params = getParameterPointers();
  opt->trainStep([&]{ forwarding(...); }, [&]{ return getLoss(); }, params);
  iter++;
}
```

This way:
- `NeuralNetwork` doesn't need to know about MeZO internals
- `perturbParameters()` moves into the MeZO class where it belongs
- `OptimizerWrapped` doesn't need MeZO-specific methods (`getMeZOEpsilon`, `updateWeightsMeZO`)
- Future gradient-free optimizers can follow the same pattern

You mentioned this concern in the PR description — I think the direction of implementing something like `backwarding_op` for MeZO is exactly right.

nntrainer/models/neuralnet.cpp
```diff
@@ -1370,10 +1375,35 @@ int NeuralNetwork::train_run(
...
+        perturbParameters(mezo_epsilon, seed);
+        float projected_grad = (loss_plus - loss_minus) / (2.0f * mezo_epsilon);
+
+        auto param_ptrs = getParameterPointers();
+        opt->updateWeightsMeZO(param_ptrs, seed, loss_plus, loss_minus);
+
+      } else {
+        ml_logi("train for iteration");
+        forwarding(true, stop_cb, stop_user_data);
+        backwarding(iter++, stop_cb, stop_user_data);
```
EunjuYang, line 1405: `iter++` is missing in the MeZO branch. In the backprop path, `iter` is incremented via `backwarding(iter++, ...)`. Without this, iteration-dependent logic (logging, LR scheduling, etc.) won't work correctly.

Add `iter++;` somewhere in this branch.

nntrainer/models/neuralnet.cpp
```diff
@@ -1894,4 +1924,38 @@ void NeuralNetwork::exports(const ml::train::ExportMethods &method,
...
+    nullptr);
+  return params;
+}
+
+void NeuralNetwork::perturbParameters(float epsilon, int seed) {
+  auto param_ptrs = getParameterPointers();
+  std::mt19937 gen(seed);
+  std::normal_distribution<float> normal_dist(0.0f, 1.0f);
+
+  for (const auto *ptr : param_ptrs) {
```
EunjuYang, line 1949: `getData()` returns `const void*`, so this C-style cast `(float *)(ptr->getData())` silently casts away constness. This is technically undefined behavior if the underlying memory is const.

Consider using a non-const accessor if one exists (e.g., `getData<float>()` or a mutable variant), or at the very least use `const_cast` explicitly to make the intent clear. The same issue exists in `MeZO.cpp:52`.

nntrainer/models/neuralnet.cpp
```diff
@@ -1267,7 +1268,11 @@ int NeuralNetwork::train(const std::vector<std::string> &values,
   model_graph.setBatchSize(
     std::get<props::TrainingBatchSize>(model_flex_props));
 
-  status = allocate(ExecutionMode::TRAIN);
+  if (opt->getType() == MeZO::type) {
+    status = allocate(ExecutionMode::INFERENCE);
```
EunjuYang, line 1272: Using `ExecutionMode::INFERENCE` to avoid gradient allocation is a smart optimization, but hardcoding the optimizer type check here is fragile.

Consider letting the optimizer declare its required execution mode:
```cpp
// In Optimizer base class
virtual ExecutionMode getRequiredExecutionMode() const {
  return ExecutionMode::TRAIN;
}
```
Then this becomes simply:
```cpp
status = allocate(opt->getRequiredExecutionMode());
```

nntrainer/optimizers/optimizer_wrapped.cpp
```diff
@@ -140,4 +141,31 @@ nntrainer::LearningRateScheduler *OptimizerWrapped::getLearningRateScheduler() {
...
+    MeZO *mezo_optimizer = static_cast<MeZO *>(optimizer.get());
+    return mezo_optimizer->getEpsilon();
+  } else {
+    // For non-MeZO optimizers, return a default value
+    ml_logw("getEpsilon called on non-MeZO optimizer, returning 0.0f");
+    return 0.0f;
+  }
+}
+
 } // namespace nntrainer
```
EunjuYang, line 171: Adding MeZO-specific methods (`updateWeightsMeZO`, `getMeZOEpsilon`) to the generic `OptimizerWrapped` class breaks the abstraction. Every new optimizer with custom behavior would require adding new methods here.

If the `trainStep()` approach described in my other comment is adopted, these methods become unnecessary — all MeZO-specific logic stays inside the `MeZO` class, and `OptimizerWrapped` remains a clean, generic wrapper.

nntrainer/optimizers/MeZO.h
EunjuYang, line 1: **Naming convention:** Other optimizers follow lowercase naming: `sgd.h`, `adamw.h`, `lion.h`. For consistency, consider renaming to `mezo.h` / `mezo.cpp`.

Similarly, the enum value `ML_TRAIN_OPTIMIZER_TYPE_MeZO` should be `ML_TRAIN_OPTIMIZER_TYPE_MEZO` to follow C enum conventions (all caps).

nntrainer/app_context.cpp
```diff
@@ -31,6 +31,7 @@
 #include <adamw.h>
 #include <lion.h>
 #include <sgd.h>
+#include<MeZO.h>
```
EunjuYang, line 34: Nit: Missing space after `#include`.
```cpp
#include <MeZO.h>
```

Applications/MeZO/jni/main.cpp
```diff
@@ -0,0 +1,291 @@
...
+      model->addLayer(layer);
+    }
+
+    // Set the optimizer
+    auto optimizer = ml::train::createOptimizer(
+      "MeZO", {"MeZO_learning_rate=0.00069", " MeZO_epsilon=0.021"});
+    model->setOptimizer(std::move(optimizer));
+
+    // Set model properties
+    model->setProperty({"epochs=5", "batch_size=32"});
```
EunjuYang, line 264: There is a leading space in `" MeZO_epsilon=0.021"` which may cause a property parsing failure or unexpected behavior.
```cpp
auto optimizer = ml::train::createOptimizer(
  "MeZO", {"MeZO_learning_rate=0.00069", "MeZO_epsilon=0.021"});
```

nntrainer/app_context.cpp
```diff
@@ -108,8 +109,8 @@
 #include <nnstreamer_layer.h>
 #endif
 
-/// add #ifdef across platform
-static std::string solib_suffix = ".so";
+  /// add #ifdef across platform
+  static std::string solib_suffix = ".so";
```
EunjuYang, line 113: These indentation changes are unrelated to MeZO. Please revert to keep the diff clean.

EunjuYang (changes requested):
## Review Summary

Thanks for working on MeZO support — zeroth-order optimization is a great addition for on-device training where backpropagation memory is a constraint. The overall MeZO algorithm flow (perturb → forward → estimate gradient → update) is implemented correctly in spirit, but there are a few issues to address before this is ready.

### Critical

1. **Weight update sign is inverted** (`MeZO.cpp:56`): The code does `+=` but MeZO requires `θ = θ - lr * projected_grad * z`, so it should be `-=`. Currently performing gradient ascent.
2. **`projected_grad` computed but unused** in `neuralnet.cpp` — it's recomputed inside `updateWeightsMeZO()`.
3. **`iter++` missing** in the MeZO training branch.

### Architecture

The main design concern is that MeZO-specific logic (perturbation, dual forward passes, custom weight update) is spread across `NeuralNetwork`, `OptimizerWrapped`, and `MeZO` classes. I've suggested a `trainStep()` virtual method pattern on the `Optimizer` interface that would:
- Keep all MeZO logic inside the `MeZO` class
- Remove MeZO-specific methods from `OptimizerWrapped`
- Remove `perturbParameters()` / `getParameterPointers()` from `NeuralNetwork`
- Make it easy to add future gradient-free optimizers

See the inline comment on `neuralnet.cpp:1378-1407` for the detailed proposal.

### Other items
- **const violation**: C-style cast discards `const` from `getData()` — potential UB
- **Naming convention**: `MeZO.h/cpp` should follow project convention (`mezo.h/cpp`)
- **Leading space** in optimizer property string `" MeZO_epsilon=0.021"`
- **Unrelated whitespace changes** in `app_context.cpp`
- **No unit tests** for the new optimizer
- **Missing newline** at end of `main.cpp` and `meson.build`

### Positive notes
- The MNIST example is clean and well-structured
- Using `ExecutionMode::INFERENCE` to skip gradient allocation is a good insight
- `applyGradient()` returning empty and `getOptimizerVariableDim()` returning `{}` correctly reflects that MeZO doesn't use per-parameter optimizer state

Looking forward to the next iteration!

sachin-nntrainer:
Thankyou @EunjuYang for a Comprehensive review. I will work towards fixing them !

Issue #3779, opened 2026-03-06
Need to decouple `from`, `to` of Q and KV


jaemini-shin:
Currently, we assume that the length of query and KV cache is same, i.e. it's okay to share same `from`, `to` parameters.
But if we adopt new logic such as KV eviction, this guarantee becomes broken.
We need to have independent `from`, `to` for query and KV cache.

PR #3814, opened 2026-03-24
[CausalLM] Fix bug in multiturn for Qwen3 example

jaemini-shin:
### Related issue
* #3779

<details><summary>{[CausalLM] Decouple query and KV cache's from for Qwen3}</summary><br />

This commit resolves issue when cache size is not equal to `from` for Qwen3 model.
* remove `from`, `to` dependency between qurey and KV cache
* improve documentations

</details>

<details><summary>{[CausalLM] Fix tensor offset of lmhead layers}</summary><br />

This commit fixes the wrong tensor offset of lmhead layers.

</details>

### Summary
This PR fixes the bug occurs in the multi turn use case with Qwen3 model: KV cache with 0 offset is overwritten in every single turn.
Now, `mha_core` doesn't rely on `from` to get the entry point of KV cache, but uses `cache_index` which means the current size of the caches.
Also, lmhead layers was using the wrong index.

### How to check?

You can easily validate this with following example:
* First turn : short input
* Second turn : long input which is longer than (input length + token gen count)

In my example:
```
first input:
"<|im_start|>user\n1+1=<|im_end|>\n<|im_start|>assistant\nThe answer is"
second input:
"<|im_start|>user\nvariable x is the answer of previous question. What is the x? Also, calculate the value of 100 + x.<|im_end|>\n<|im_start|>assistant\nThe answer is"
```

### Before this PR
```
 100 + x = 100 + 10 = 110. So, x is 10 and 100 + x is 110. But wait, I need to check the previous question to make sure I'm not making a mistake. The user didn't provide the previous question, so I can't know what x is. I should ask for clarification. But the user said ...
```

### After this PR
```
 2.
</think>

The value of **x** is **2**, as determined from the previous question.

Now, calculate **100 + x**:

$$
100 + x = 100 + 2 = 102
$$

So, the final result is **102**.<|im_end|>
```

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/lm_head.cpp
+++ b/Applications/CausalLM/layers/lm_head.cpp
@@ -141,9 +141,7 @@ void LmHeadLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   for (unsigned int b = 0; b < b_size; ++b) {
     nntrainer::Tensor input_step = input_.getSharedDataTensor(
       input_step_dim,
-      b * input_dim.getFeatureLen() +
-        (to - from == 1 ? 0 : (to - 1) * input_.width()),
-      true);
+      b * input_dim.getFeatureLen() + (to - from - 1) * input_.width(), true);
     nntrainer::Tensor hidden_step = hidden_.getSharedDataTensor(
       hidden_step_dim, b * hidden_dim.getFeatureLen(), true);
 
--- a/Applications/CausalLM/layers/tie_word_embedding.cpp
+++ b/Applications/CausalLM/layers/tie_word_embedding.cpp
@@ -281,9 +281,7 @@ void TieWordEmbedding::incremental_forwarding_lmhead(
   for (unsigned int b = 0; b < b_size; ++b) {
     nntrainer::Tensor input_step = input_.getSharedDataTensor(
       input_step_dim,
-      b * input_dim.getFeatureLen() +
-        (to - from == 1 ? 0 : (to - 1) * input_.width()),
-      true);
+      b * input_dim.getFeatureLen() + (to - from - 1) * input_.width(), true);
     nntrainer::Tensor hidden_step = hidden_.getSharedDataTensor(
       hidden_step_dim, b * hidden_dim.getFeatureLen(), true);
 


Applications/CausalLM/layers/lm_head.cpp
```diff
@@ -141,9 +141,7 @@ void LmHeadLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   for (unsigned int b = 0; b < b_size; ++b) {
     nntrainer::Tensor input_step = input_.getSharedDataTensor(
       input_step_dim,
-      b * input_dim.getFeatureLen() +
-        (to - from == 1 ? 0 : (to - 1) * input_.width()),
-      true);
+      b * input_dim.getFeatureLen() + (to - from - 1) * input_.width(), true);
```
jaemini-shin, line 144: Valid input range is [0, to - from).
Previous range was valid in prefill with `from == 0`.
In multi-turn example, prefill with `from != 0` is available.

Applications/CausalLM/layers/tie_word_embedding.cpp
```diff
@@ -281,9 +281,7 @@ void TieWordEmbedding::incremental_forwarding_lmhead(
   for (unsigned int b = 0; b < b_size; ++b) {
     nntrainer::Tensor input_step = input_.getSharedDataTensor(
       input_step_dim,
-      b * input_dim.getFeatureLen() +
-        (to - from == 1 ? 0 : (to - 1) * input_.width()),
-      true);
+      b * input_dim.getFeatureLen() + (to - from - 1) * input_.width(), true);
```
jaemini-shin, line 284: Valid input range is [0, to - from).
Previous range was valid in prefill with `from == 0`.
In multi-turn example, prefill with `from != 0` is available.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -343,29 +343,6 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
         cache_value_dim, cache_value_step_dim);
     }
   }
-
-  if (!_from) {
```
jaemini-shin, line 347: This logic is related to line 281. We don't need it for correct `batch_size`.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -492,38 +469,41 @@ void MHACoreLayer::one_batch_incremental_forwarding(
   ml::train::TensorDim &cache_value_step_dim) {
```
jaemini-shin, line 472: I removed dependency to `from`, `to`. We can replace them with `cache_index` and the height of the input(same as `to - from`).
This is the first step of removing `incremental_forwarding`.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -262,23 +262,23 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
...
+    get_step_dim(output_dim); // (1, 1, step_size, n_heads_Q * head_dim)
   ml::train::TensorDim cache_key_step_dim =
-    get_step_dim(cache_key_dim); // (B, 1, from-to, n_heads_KV * head_dim)
+    get_step_dim(cache_key_dim); // (1, 1, step_size, n_heads_KV * head_dim)
 
   ml::train::TensorDim cache_value_step_dim =
-    get_step_dim(cache_value_dim); // (B, 1, from-to, n_heads_KV * head_dim)
+    get_step_dim(cache_value_dim); // (1, 1, step_size, n_heads_KV * head_dim)
 
-  unsigned int batch_size = (_from) ? 1 : query_dim.batch();
```
jaemini-shin, line 281: This line could matter for multi batch input.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -532,33 +512,40 @@ void MHACoreLayer::one_batch_incremental_forwarding(
...
-  softmax_triangle(out_, to - from, num_heads_Q, from, pool);
+  softmax_triangle(out_, step_size, num_heads_Q, cache_from, pool);
 
   compute_fp16vcache_transposed(out_, b_cached_value, attention_output_step,
-                                from, num_heads_KV, gqa_size, head_dim, to,
-                                pool);
+                                cache_from, num_heads_KV, gqa_size, head_dim,
+                                cache_to, pool);
+
+  cache_index += step_size;
```
jayden0701: Looks Great!
But keeping single cache_index and adding step size for all batches could cause problems if multi batch.
Just leaving comment in case for someone who might want multibatch!

jaemini-shin: You're right! It should be increased once, I'll reflect it.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -250,9 +253,6 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
     sink = context.getWeight(sink_idx);
   }
 
-  const unsigned int num_heads_Q =
```
Seunghui98, line 253: Why do you remove this??

jaemini-shin: Because it's unused one! Also, there's a member variable with same name. That's why I removed this line.

Applications/CausalLM/layers/mha_core.cpp
jaemini-shin: updated!

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -215,6 +217,7 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
       throw std::invalid_argument(
         "to shouldn't greater than max_timestep for initial forwarding");
     } else {
```
EunjuYang, line 219: The `throw std::runtime_error(...)` makes the lines below it (`cache_shift = true; ...`) unreachable dead code. Consider either:
1. Removing the dead code entirely, or
2. If cache shift is planned for the future, keeping the dead code but adding a clear comment like `// TODO: implement cache shift` and removing the unreachable assignments.

As-is, a compiler may warn about unreachable code.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -343,29 +343,9 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
...
-                                      batch * cache_value_dim.getFeatureLen() +
-                                        from * cache_value_dim.width(),
-                                      true);
-
-      cache_key_nth_step.copyData(cache_key_0_step);
-      cache_key_nth_step.copyData(cache_value_0_step);
-    }
-  }
+  
+  // increase cache size
```
EunjuYang, line 347: `cache_index` is initialized to 0 in the constructor and incremented here, but I don't see any mechanism to reset it (e.g., for a new conversation session or when the model is reused across independent conversations). How is `cache_index` reset between separate multi-turn sessions? If the layer object is reconstructed each time, this is fine — but if it's reused, a `reset()` or `clear_cache()` method would be needed to avoid the cache index growing unboundedly across sessions.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -343,29 +343,9 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
         cache_value_dim, cache_value_step_dim);
     }
   }
```
EunjuYang, line 345: Nit: There is trailing whitespace on this line (spaces after the `}`). Minor, but worth cleaning up if your project has a lint/style check.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -532,33 +515,38 @@ void MHACoreLayer::one_batch_incremental_forwarding(
 #endif
   }
 
+  /// @todo replace step_size into input height
```
EunjuYang, line 518: `step_size` is already computed at the top of `incremental_forwarding` and passed indirectly via `from`/`to` to this function. Here it's recomputed from `to - from`. Consider passing `step_size` as a parameter (or using it from a member) instead of recomputing it, to keep the logic consistent and reduce the chance of divergence between the two computations.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -570,6 +558,8 @@ void MHACoreLayer::one_batch_incremental_forwarding(
   ml::train::TensorDim &cache_key_step_dim,
   ml::train::TensorDim &cache_value_dim,
   ml::train::TensorDim &cache_value_step_dim, nntrainer::Tensor &sink_step) {
```
EunjuYang, line 560: Good that this is annotated with `@todo` and `@note`. Just want to confirm: the second overload of `one_batch_incremental_forwarding` (the sink variant used by gpt-oss) still uses `_from`/`to` directly instead of `cache_index`. This means gpt-oss and Qwen3 paths now have divergent cache management strategies. Is there a risk that the gpt-oss path has the same multi-turn bug? If so, it would be good to track this as a follow-up issue.

Applications/CausalLM/layers/lm_head.cpp
```diff
@@ -141,9 +141,7 @@ void LmHeadLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   for (unsigned int b = 0; b < b_size; ++b) {
     nntrainer::Tensor input_step = input_.getSharedDataTensor(
       input_step_dim,
```
EunjuYang, line 143: Good fix. The old code had a subtle bug: when `to - from > 1` (prefill), it used `(to - 1) * input_.width()` which was incorrect when `from > 0`, because it treated the offset as absolute within the full sequence rather than relative to the step tensor. The new expression `(to - from - 1) * input_.width()` correctly computes the offset of the last row within the step tensor regardless of `from`.

Applications/CausalLM/layers/mha_core.cpp
jaemini-shin: AFAIK, you should make another model object to make a new session. With this assumption, `cache_index` should not be cleared until model object is destructed.
If this use case is necessary, I'll make a new PR to update this feature.

PR #3820, opened 2026-03-28
[test] add benchmark framework for x86 CPU backend

YongHyeon02:
## Dependency of the PR

None

<details><summary>[test] add general-purpose benchmark framework for cpu        
  backend</summary><br />                                                       
                                                                                  
  Add benchmark infrastructure for measuring x86 CPU backend performance:         
  - benchmark_utils.h: general-purpose timing/throughput measurement utilities
  - benchmark_alloc_counter.h: memory allocation counter                          
  - benchmark_x86_backend.cpp: benchmark tests for CPU dispatch functions         
  
  **Self evaluation:**                                                            
                                                                                
  </details>                                                                      
                  
  <details><summary>[benchmark] refactor to general-purpose measurement           
  pattern</summary><br />
                                                                                  
  Refactor benchmark framework to be reusable beyond specific projects:
  - Remove fallback comparison pattern, each test measures current backend
  - Add CLI parameters: --bench_iters, --bench_warmup, --bench_sizes              
  - Report throughput/bandwidth/GFLOPS via bench::report()
  - Tune defaults: iters 1000, warmup 10                                          
                  
  **Self evaluation:**                                                            
   
  </details>                                                                      
   
  <details><summary>[benchmark] apply clang-format style</summary><br />          
                  
  Apply clang-format to benchmark source files

  **Self evaluation:**
   
  </details>

<details><summary>fix: add doxygen @brief tags and increase benchmark timeout</summary><br />
                                                                                                                                        
  Add missing @brief doxygen tags and increase benchmark test timeout.                                                                  
                                                                                                                                        
  **Self evaluation:**                                                                                                                  

  </details>                                                                                                                            
                             
  <details><summary>refactor: interleave FP32/FP16 benchmark results for easy comparison</summary><br />                                
   
  Interleave FP32 and FP16 results per operation (op_A FP32 → op_A FP16 → op_B FP32 → ...)                                              
  instead of separate FP32/FP16 test suites, for easier side-by-side comparison.
                                                                                                                                        
  **Self evaluation:**       
   
  </details>                                                                                                                            
                             
  <details><summary>build: remove benchmark from meson test registration</summary><br />

  Remove benchmark binary from meson test() registration so CI does not                                                                 
  run it. The benchmark is build-only and must be run manually.
                                                                                                                                        
  **Self evaluation:**       

  </details>

### Summary
- Add general-purpose benchmark framework for x86 CPU backend with timing, throughput, and memory allocation measurement
- Refactor to configurable, reusable measurement pattern with CLI parameters
- Interleave FP32/FP16 results per operation for easy comparison
- Remove benchmark from meson test registration (build-only, manual run)
- Apply clang-format code style


YongHyeon02:
> Appreciate your contribution! By the way how about displaying results in an interleaved pattern like op A (fp32) - op A(fp16) - op B (fp32) - op B (fp16) - ... for easy comparison between fp32 mode and fp16 mode of each operation?

I've updated the code!

baek2sm (approved):
LGTM

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,716 @@
...
+  float epsilon = 1e-6f;
+  std::string sz = "H=" + std::to_string(H) + ",W=" + std::to_string(W);
+
+  {
+    auto X = generate_random_vector<float>(N);
+    std::vector<float> Y(N);
+
+    auto stats = bench::measure(
+      [&]() {
+        nntrainer::rms_norm_wrt_width_fp16_intrinsic<float>(X.data(), Y.data(),
```
baek2sm: The FP32 RMSNorm benchmark calls rms_norm_wrt_width_fp16_intrinsic<float>(), but that x86 specialization forwards to an NYI fallback and throws at runtime. Please use rms_norm_wrt_width_fp32_intrinsic() here or skip this case until the fp16 intrinsic path is implemented.

baek2sm (changes requested):

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: Thanks for the catch! I've switched the FP32 case to call `rms_norm_wrt_width_fp32_intrinsic` directly so the benchmark exercises the actual AVX2 path.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,716 @@
...
+    [&]() {
+      nntrainer::sgemm(
+        0, false, false, M, N, K, 1.0f, (const _FP16 *)A_u16.data(), K,
+        (const _FP16 *)B_u16.data(), N, 0.0f, (_FP16 *)C_u16.data(), N);
+    },
+    g_bench_warmup, g_bench_iters);
+
+  std::string sz =
+    std::to_string(M) + "x" + std::to_string(N) + "x" + std::to_string(K);
+  bench::report("sgemm", "FP16", sz, stats, {.flop_count = 2.0 * M * N * K});
```
jaemini-shin: Designated Initializer has introduced since C++20, where nntrainer uses C++17.

jaemini-shin (changes requested):

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: I've replaced the designated initializers across all `bench::report()` call sites with explicit `bench::Metrics` construction.

test/unittest/meson.build
```diff
@@ -119,6 +119,19 @@ foreach target: test_target
   )
 endforeach
 
+# Benchmark binary — build only, not registered as a test.
+# Run manually: ./benchmark_x86_backend --bench_iters=100
+if get_option('enable-fp16')
```
jaemini-shin: This benchmark includes both fp32 and fp16 test. It'd be better to use other option, such as `enable-test`.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,752 @@
...
+// ============================================================================
+// Attention Kernel Benchmarks (Performance + Allocation Counting)
+// ============================================================================
+
+using AttnParams = std::tuple<int, int, int, size_t>;
+
+/**
+ * @brief Test fixture for attention kernel benchmarks with allocation counting
+ */
+class Bench_Attention : public ::testing::TestWithParam<AttnParams> {};
```
jaemini-shin: We always keep the KV cache as 16bit regardless of `enable-fp16` option.
* `enable-fp16=true`
  * Q: fp16
  * K: fp16
  * V: fp16
* `enable-fp16=false`
  * Q: fp32
  * K: fp16 in u16 type
  * V: fp16 in u16 type

You need to rewrite test code according to these conditions.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+static unsigned int g_bench_iters = 1000;
+static unsigned int g_bench_warmup = 10;
+static std::vector<unsigned int> g_bench_sizes = {256, 1024, 4096};
+
+// ============================================================================
+// Helpers
+// ============================================================================
+
+#if defined(ENABLE_FP16) || defined(__AVX2__)
```
jaemini-shin:
```suggestion
#if defined(ENABLE_FP16)
```
`nntrainer::compute_fp32_to_fp16` and `nntrainer::compute_fp16_to_fp32` doesn't require AVX2 support.

test/include/benchmark_alloc_counter.h
jaemini-shin, line 1: Is this allocation counter necessary?

It looks like this is for verifying zero heap allocation, but this feature might not be critical.

test/unittest/meson.build
YongHyeon02: I've switched the gate to `if get_option('enable-test')`!

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: I've split into two fixtures:
- Bench_Attention (always built): Q=float, K/V=uint16_t. Uses `compute_kcaches<uint16_t>` and `compute_fp16vcache_fp32_transposed`. Covers `enable-fp16=false`.
- Bench_Attention_FP16 (under `#ifdef ENABLE_FP16`): Q=_FP16, K/V=_FP16. Uses the _FP16-typed overloads. Covers `enable-fp16=true`.

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: Dropped the guard entirely so
`convert_f32_to_f16_u16` is unconditionally available.
Also removed the unused `convert_f16_u16_to_f32` helper that the old guard was hiding.

test/include/benchmark_alloc_counter.h
YongHyeon02: Agreed — Removed `benchmark_alloc_counter.h` along with all its uses,
the GCC pragma, and the `allocs=` / `RecordProperty("alloc_*", ...)` reporting.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+// ---- Set A: FP32 Q + uint16(FP16-bits) KV cache (available in all builds) ----
+
+/**
+ * @brief Test fixture for attention kernel benchmarks (FP32 Q, uint16 KV)
+ */
+class Bench_Attention : public ::testing::TestWithParam<AttnParams> {};
+
+TEST_P(Bench_Attention, compute_kcaches) {
+  auto [head_dim, num_cache_head, gqa_size, window_size] = GetParam();
```
jaemini-shin: With these parameters, sequence length of Q,K,V,O is always 1 and window_size is ignored.
What I recommend is using `[seq_len, num_cache_head, gqa_size, tile_size]`.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+  std::string sz = "hd=" + std::to_string(head_dim) +
+                   ",nch=" + std::to_string(num_cache_head) +
+                   ",gqa=" + std::to_string(gqa_size) +
+                   ",w=" + std::to_string(window_size);
+  bench::report("compute_kcaches", "FP32", sz, stats);
+
+  RecordProperty("latency_ns", stats.avg_ns);
+}
+
```
jaemini-shin, line 814: In inference, nntrainer doesn't use `softmax` but `softmax_row` or `softmax_row_inplace`. You should add them to measure attention correctly.
Both of them have same internal implementation, so it'd be enough to check only one item.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,752 @@
...
+  m.num_elements = N;
+  m.total_bytes = N * sizeof(uint16_t);
+  bench::report("sscal", "FP16", "N=" + std::to_string(N), stats, m);
+}
+
+GTEST_PARAMETER_TEST(Sizes, Bench_FP16_BLAS,
+                     ::testing::ValuesIn(g_bench_sizes));
+
+// ============================================================================
+// GEMM Benchmarks (Performance + Allocation Counting)
```
jaemini-shin:
```suggestion
// GEMM Benchmarks (Performance)
```
As of now, allocation counting is redundant.

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: Thanks — split the `compute_kcaches` cases into a new `Bench_KCache` / `Bench_KCache_FP16` fixture with the recommended `[seq_len, num_cache_head, gqa_size, tile_size]` layout. `head_dim` is fixed at 128 (typical for current LLMs) so the dimensions you suggested actually drive the workload, and `local_window_size` is set to `seq_len` so the full Q range is processed.

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: Thanks — added `Bench_SoftmaxRow` that exercises `softmax_row_inplace`, which is the production attention path. As you noted, `softmax_row` and `softmax_row_inplace` share the same internal implementation, so benchmarking the inplace variant covers both. The original `Bench_Activations::softmax` is kept for now as a baseline comparison.

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: Done — updated to `// GEMM Benchmarks (Performance)`.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+ */
+class Bench_KCache : public ::testing::TestWithParam<KCacheParams> {};
+
+TEST_P(Bench_KCache, compute_kcaches) {
+  auto [seq_len, num_cache_head, gqa_size, tile_size] = GetParam();
+  const int head_dim = kKCacheHeadDim;
+  const int total_heads = num_cache_head * gqa_size;
+  const size_t local_window_size = static_cast<size_t>(seq_len);
+
+  auto in_f32 = generate_random_vector<float>(total_heads * head_dim);
```
jaemini-shin: Input tensor should be in `[batch, seq_len, num_q_heads, head_dim]`.
So its size should be `seq_len * total_heads * head_dim`.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+  auto kcache_f32 =
+    generate_random_vector<float>(num_cache_head * seq_len * head_dim);
+  auto kcache_u16 = convert_f32_to_f16_u16(kcache_f32);
+  std::vector<float> output(total_heads * seq_len, 0.0f);
```
jaemini-shin: We use different output shape whether it's causal or not.
* Causal attention
  * `seq_len * (seq_len + 1) * num_q_heads / 2`
  * Instead of using mask, we don't calculate for `k_idx > q_idx`.
* Non-causal attention
  * `seq_len * seq_len * num_q_heads`

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+
+  auto stats = bench::measure(
+    [&]() {
+      nntrainer::compute_kcaches<uint16_t>(
```
jaemini-shin: Every attention related kernels(`compute_kcaches`, `softmax_in_row`, `compute_fp16vcache_fp32_transposed`, ...) should be called for every single q row.
```cpp
// ignoring batch
for(int i=0;i<seq_len;i++){
  float* in = in_f32.data() + i * num_q_heads * head_dim;
  uint16_t* k = kcache_u16.data();
  float* out = output.data() + i * (i+1) * num_q_heads / 2;
  nntrainer::compute_kcaches ...
}
```

In runtime, we also use threadpool on this loop to accelerate attention. You can find the actual code in `mha_core.cpp`, check the `one_batch_incremental_forwarding`.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+
+  RecordProperty("latency_ns", make_record_property_value(stats.avg_ns));
+}
+
+GTEST_PARAMETER_TEST(Configs, Bench_KCache, kKCacheConfigs);
+
+/**
+ * @brief Test fixture for attention kernel benchmarks (FP32 Q, uint16 KV)
+ */
+class Bench_Attention : public ::testing::TestWithParam<AttnParams> {};
```
jaemini-shin, line 899: It looks a bit awkward to using different test parameters. It'd be better to use same parameters along the attention tests.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+
+TEST_P(Bench_Attention, compute_fp16vcache) {
+  auto [head_dim, num_cache_head, gqa_size, window_size] = GetParam();
+  int total_heads = num_cache_head * gqa_size;
+  int row_num = static_cast<int>(window_size) - 1;
+  const int attention_rows = row_num + 1;
+
+  auto in_f32 = generate_random_vector<float>(total_heads * attention_rows);
```
jaemini-shin: It's input shape is same as the output of `compute_kcache()`.
* Causal attention
  * `seq_len * (seq_len + 1) * num_q_heads / 2`
* Non-causal attention
  * `seq_len * seq_len * num_q_heads`

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+  auto vcache_f32 =
+    generate_random_vector<float>(num_cache_head * (int)window_size * head_dim);
+  auto vcache_u16 = convert_f32_to_f16_u16(vcache_f32);
+  std::vector<float> output(total_heads * head_dim, 0.0f);
```
jaemini-shin: Output shape should be `[batch, seq_len, num_q_heads, head_dim]`.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+// Multihead Row-Softmax Benchmarks (FP32 vs FP16 interleaved)
+// Production attention path uses softmax_row / softmax_row_inplace; both share
+// the same internal implementation, so benchmarking the inplace variant covers
+// both.
+// ============================================================================
+
+/**
+ * @brief Test fixture for multihead row-softmax benchmarks
+ */
+class Bench_SoftmaxRow
```
jaemini-shin: It'd be better to move this test between kcache and vcache.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,998 @@
...
+  : public ::testing::TestWithParam<std::tuple<unsigned int, unsigned int>> {};
+
+TEST_P(Bench_SoftmaxRow, softmax_row_inplace) {
+  auto [num_rows, num_heads] = GetParam();
+  const unsigned int N = num_rows * num_heads;
+  std::string sz =
+    "rows=" + std::to_string(num_rows) + ",heads=" + std::to_string(num_heads);
+
+  {
+    auto X_orig = generate_random_vector<float>(N);
```
jaemini-shin: Its shape should be the same as output of `compute_kcache()`.
* Causal attention
  * `seq_len * (seq_len + 1) * num_q_heads / 2`
* Non-causal attention
  * `seq_len * seq_len * num_q_heads`

jaemini-shin (commented):
I left some comments on attention tests, PTAL.

baek2sm:
**Review verdict: 💬 Comment (mergeable; only non-blocking nits)**

### 1. Summary

Adds a reusable, self-contained benchmark framework (`bench::measure`, `bench::report`, `bench::compare` in `test/include/benchmark_utils.h`) plus a concrete `benchmark_x86_backend` binary exercising the x86/FP16 CPU backend (GEMM, elementwise, RMSNorm, softmax, attention KV-cache kernels). The binary is built only on x86/x86_64 when `enable-test=true` and is intentionally not registered as a meson `test()` target (timing measurements are non-deterministic).

Included:
- `bench::Stats` / `bench::Metrics` and `measure()` / `measure_with_setup()` templates with warmup, isolating setup cost.
- Human-readable table reporter with auto-scaled time units, throughput, bandwidth and GFLOPS, plus `compare()` for baseline-vs-optimized speedup rows.
- CLI flags `--bench_iters`, `--bench_warmup`, `--bench_sizes` parsed before `InitGoogleTest`.
- FP32 and FP16 (`#ifdef ENABLE_FP16`) cases for elementwise ops, activations, HGEMM/GEMV, RMSNorm, multihead softmax, and attention KV-cache.
- `RecordProperty` GFLOPS/latency hooks for the daily `ubuntu_benchmarks.yml`.

### 2. Notes (non-blocking)

- `test/unittest/benchmark_x86_backend.cpp` has no `Copyright (C)` line, while `benchmark_utils.h` does. Not CI-checked, but inconsistent — consider aligning. (P2)
- Commit `994cfee9` body mentions "RoPE" benchmarks, but no RoPE/rotary benchmark is present in the diff. Minor message inaccuracy. (P3)
- `Bench_Attention::compute_fp16vcache` computes `row_num = static_cast<int>(window_size) - 1`; current fixtures all have `window_size ≥ 128` so it is safe, but a future `window_size=0` config would underflow before the cast. Consider an `ASSERT_GE(window_size, 1u)` guard. (P3)

### 3. Stability for integration with latest main

The change is additive only (two new files + two guarded `meson.build` lines); it touches no library source or existing test. The x86-only meson guard (`host_machine.cpu_family() in ['x86','x86_64']`) excludes ARM/Tizen/Android cross-compile jobs, and the `ENABLE_FP16` guards mirror existing tests in the same directory. All called symbols are declared in `cpu_backend.h` on current upstream/main. Compilation is still covered by the Ubuntu/Windows x86_64 CI legs (where `enable-test=true` is default), so link/compile regressions are caught; the binary simply isn't auto-run. No new meson option, dependency, or exported symbol — blast radius is effectively zero.

Applies cleanly on top of current upstream/main (the CausalLM weight-converter commit in the branch tip is already upstream, so it doesn't appear in the PR diff). No rebase required.

### Overall

Sound and safe to merge. The only feedback is cosmetic (copyright line, commit-message wording) plus a defensive assert; none blocks merge.

baek2sm:
**Final Verdict: ✅ Approve**

---

Adds a header-only micro-benchmark framework for the x86 CPU backend plus a standalone benchmark binary that times the FP32/FP16 kernels against varying problem sizes.

**Included features**
- `bench::` timing/statistics/reporting utilities (`test/include/benchmark_utils.h`), dependency-free.
- `benchmark_x86_backend` GTest binary covering element-wise, activation, softmax, RMSNorm, BLAS, HGEMM/GEMV and attention-cache kernels, with CLI flags for iterations/warmup/sizes.
- Meson target gated to `x86`/`x86_64` host families (build-only; not registered as a runnable test).

---

### 2. Related Issues

- **P1 (fixed)** Linux GCC/clang portability: structured-binding capture in lambdas, `nullptr`→`T*` template-deduction, and FP16 `softmax_row_inplace` sink type-consistency were corrected; x86-unimplemented FP16 kernels (`softmax_row_inplace<_FP16>`, the kcache/vcache ops) are now ARM-guarded so the x86 link succeeds.
- **P1 (fixed)** 4 CodeQL high-severity `cpp/integer-multiplication-cast-to-long` alerts — benchmark size computations widened to `size_t` before multiplication.
- **P2 (fixed)** clang-format-14 formatting.

No remaining P0/P1.

---

The change is additive and test-only: no `nntrainer` core code is touched, and the benchmark binary is gated to x86/x86_64 hosts and built (not executed) in CI. FP16 ops that exist only on ARM are compiled out on x86, while x86-implemented FP16 ops remain exercised. The full CI matrix — including CodeQL, all `meson_test` ubuntu variants (±fp16), Windows MSVC/arm, Android, Tizen, cross-ARM64, static checks and cpp-linter — is green.

---

A self-contained x86 benchmark harness with all portability, security-scan, and formatting issues resolved and a fully green CI matrix. Approve.

YongHyeon02:
@jaemini-shin 
I'll keep the per-row loop serial. Defaulting to causal (`IsCausal=true`, matching decoder models) while still supporting the non-causal shape on a small subset. Memory is bounded by keeping `num_q_heads` realistic, so the `seq_len ∈ {128, 512, 1024, 2048}` sweep stays as-is. Let me know if you'd prefer otherwise.

test/unittest/benchmark_x86_backend.cpp
```diff
@@ -0,0 +1,1123 @@
...
+        }
+      },
+      g_bench_warmup, g_bench_iters);
+
+    bench::Metrics m;
+    m.num_elements = score_rows * total_heads;
+    bench::report("softmax_row_inplace", "FP32", sz, stats, m);
+  }
+
+#if defined(ENABLE_FP16) && (defined(__ARM_NEON) || defined(__ARM_NEON__))
```
jaemini-shin, line 864:
```suggestion
#if defined(ENABLE_FP16)
```
Now this file is built for x86 target only.

jaemini-shin (commented):
CI was not enabled in this PR... 😭

test/unittest/benchmark_x86_backend.cpp
YongHyeon02: Thanks for the review.
The guard exists because the kernels inside are FP16 (Set B) variants only implemented for ARM so far (`arm/arm_compute_backend_fp16.cpp`). Switching to `#if defined(ENABLE_FP16)` would cause link errors on x86 since there's no implementation there yet.
The x86 implementation is in progress in a separate PR and hasn't merged yet.
How should we proceed — should I just remove the guard for now and fix it once the x86 PR lands?

jaemini-shin: IMO, it looks fine to proceed as it is and fix it when fp16 attention for x86 is fully merged!

jaemini-shin (approved):
LGTM!

PR #3823, opened 2026-03-31
[CausalLM] Causal activation fp16

lhs8928:
Commit 1:     [Application] enable resetInputDimension in causallm
    
     - Enable resetInputDimension in Causal lm
    
Commit 2:     [causallm] applied fp16 input/output dtype on causallm
    
     - Added generate for fp16 input/output
    
Commit 3:     [dotQnK] enable dot operation of q4_0 and fp16
    
     - Enable dotQnK on half tensor
     - set thread number for gemm
     - Optimizing gemm of q4_0 and fp16 by removed temporary local variable(C32) and declared as global variable
    
Commit 4:     [embedding] Directly dequantize q4_0 to half tensor
    
     - Added dequantize_row_q4_0 for half tensor
    
Commit 5:     [swiglu] improve performance of swiglu for fp16 data type
    
     - Loop unrolling swiglu to improve computation speed for fp16 data type
    
Commit 6:     [normalization] set rms norm weight as fp32
    
     - Set weight dtype of rms norm as fp32 even though input/output dtype is not fp32
    
### DIFF
--- a/api/ccapi/include/tensor_dim.h
+++ b/api/ccapi/include/tensor_dim.h
@@ -17,9 +17,10 @@
 #ifdef __cplusplus
 
 #include <array>
-#include <iosfwd>
-
 #include <bitset>
+#include <cstdint>
+#include <iosfwd>
+#include <variant>
 #include <vector>
 
 #ifdef ENABLE_FP16
@@ -39,6 +40,14 @@ namespace train {
  */
 class TensorDim {
 public:
+#ifdef ENABLE_FP16
+  using IO_TensorType = std::variant<float *, uint32_t *, uint16_t *, uint8_t *,
+                                     int16_t *, int8_t *, _FP16 *>;
+#else
+  using IO_TensorType = std::variant<float *, uint32_t *, uint16_t *, uint8_t *,
+                                     int16_t *, int8_t *>;
+#endif
+
   static constexpr const size_t MAXDIM = 4;
 
   /**
--- a/Applications/CausalLM/layers/embedding_layer.cpp
+++ b/Applications/CausalLM/layers/embedding_layer.cpp
@@ -236,6 +236,21 @@ void EmbeddingLayer::save(std::ofstream &file,
   }
 }
 
+void EmbeddingLayer::updateTensorsByInputDimensions(
+  nntrainer::RunLayerContext &context,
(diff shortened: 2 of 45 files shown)


api/ccapi/include/tensor_dim.h
```diff
@@ -39,6 +40,14 @@ namespace train {
  */
 class TensorDim {
 public:
+#ifdef ENABLE_FP16
+  using IO_TensorType = std::variant<float *, uint32_t *, uint16_t *, uint8_t *,
```
jijoongmoon, line 44: We could only use the first IO_TensorType define without ENABLE_FP16.

baek2sm:
**Final Verdict: ✅ Approve**

---

Adds an end-to-end FP16-activation inference path for CausalLM while keeping normalization weights in FP32 for numerical stability.

**Included features**
- `IO_TensorType` typed inference I/O plus FP16 `inference` / `incremental_inference` overloads.
- `dequantize_row_q4_0` → FP16 through the full backend dispatch chain; `HalfTensor::dot` Q4_0 path wired to `dotQnK`.
- Unrolled FP16 NEON SwiGLU fast path; transformer/attention norm layers pinned to `weight_dtype=fp32`.
- Per-thread GEMM scratch reuse to avoid repeated allocation in the decode loop.

---

### 2. Related Issues

- **P3** Thread-safety hardened — the shared GEMM scratch (`C32`) is now `thread_local`, eliminating a data race under concurrent inference.
- **P3** Corrected the public `model.h` / `neuralnet.h` `incremental_inference(IO_TensorType)` doc; added a guard that throws on an unsupported output dtype instead of returning a null variant; removed dead commented-out swiglu code.
- **P3** `tensor_dim.h` keeps an `#ifdef ENABLE_FP16` split on the variant (unifying would break `std::get<_FP16*>` callers) — acceptable.

No P0/P1/P2 issues.

---

Clean merge onto the latest main; the redundant `SharedFullyConnectedLayer::save` commit auto-dropped (already upstream). FP16 paths are `#ifdef ENABLE_FP16`-guarded across arm/x86/fallback, so non-FP16 builds are unaffected. The full CI matrix including the FP16 unit-test lanes is green.

---

A complete FP16-activation pipeline with thread-safety, documentation, and UB-guard issues resolved, and a fully green CI matrix. Approve.

PR #3827, opened 2026-04-01
Causallm/make transformer base

EunjuYang:

<details><summary>[CausalLM] Extract TransformerBase abstract class from Transformer</summary><br />

Introduce TransformerBase as an abstract base class that defines the
common interface (initialize, load_weight, save_weight, run) and shared
state (model, tokenizer, basic dimension params) for all Transformer
variants. This enables QNNTransformer to coexist alongside the existing
NNTrainer-based Transformer without inheriting HuggingFace-specific
config parsing logic.

- Create transformer_base.h with pure virtual interface
- Refactor Transformer to inherit from TransformerBase
- Update Factory to use TransformerBase as the base type
- Move LoadJsonFile, ModelType enum, and type aliases to
  transformer_base.h
- Update api.cpp (Transformer to TransformerBase)

</details>

<details><summary>[CausalLM] Refactor run() - remove do_sample, add void* touput and </summary><br />

- Remove do_sample from TransformerBase::run() interface since only
  CausalLM uses it. CausalLM now read do_sample
  from generation_config.json during setupParameters
  and store as DO_SAMPLE member variable.
- Add void* output_buf parameter to run() for retrieving results:
  CausalLM fills std::vector<std:: string>*, SentenceTransformer fills
std::vector<float*>*. nullptr (default) keeps existing stdout-only
behavior.
- Add simple overload run(prompt, output_buf) that delegates to the full
  version with empty system/tail prompts.

</details>

### Summary

- create `TransformerBase`
- update `run` prototype for common TransformerBase

### DIFF
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -102,6 +102,9 @@ void CausalLM::setupParameters(json &cfg, json &generation_cfg,
   TEMPERATURE = generation_cfg.contains("temperature")
                   ? generation_cfg["temperature"].get<float>()
                   : 0.7;
+  DO_SAMPLE = generation_cfg.contains("do_sample")
+                ? generation_cfg["do_sample"].get<bool>()
+                : false;
   global_token_len = 0;
 }
 
@@ -330,8 +333,12 @@ void CausalLM::registerCustomLayers() {
   }
 }
 
-void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
-                   const WSTR tail_prompt, bool log_output) {
+void CausalLM::run(const WSTR prompt, void *output_buf, bool log_output) {
+  run(prompt, WSTR(), WSTR(), output_buf, log_output);
+}
+
+void CausalLM::run(const WSTR prompt, const WSTR system_prompt,
+                   const WSTR tail_prompt, void *output_buf, bool log_output) {
 
   auto start_total = std::chrono::high_resolution_clock::now();
   if (!is_initialized) {
@@ -523,7 +530,7 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
 
   // post process of model output
   std::vector<unsigned int> id_list(
-    generate(output[0], do_sample, 1, ids_history, init_len));
+    generate(output[0], DO_SAMPLE, 1, ids_history, init_len));
 
   if (init_len < INIT_SEQ_LEN)
     registerOutputs(tokenizer, id_list, init_len, eos_list, log_output);
@@ -559,7 +566,7 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
       model->incremental_inference(BATCH_SIZE, input, label, input_len,
                                    token_generation_idx - 1 + global_token_len,
                                    token_generation_idx + global_token_len);
-    std::vector<unsigned int> ids_list(generate(output_interval[0], do_sample));
+    std::vector<unsigned int> ids_list(generate(output_interval[0], DO_SAMPLE));
 
     // Feed the newly generated token back as the next input token.
     // token_generation_idx always starts at input_len + 1, so we are
@@ -603,6 +610,10 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   free(input_sample);
 
   global_token_len += (generation_cnt + init_len);
+
+  if (output_buf != nullptr) {
+    *static_cast<std::vector<std::string> *>(output_buf) = output_list;
+  }
 
   auto finish_generation = std::chrono::high_resolution_clock::now();
   auto generation_duration =
--- a/Applications/CausalLM/api/causal_lm_api.cpp
+++ b/Applications/CausalLM/api/causal_lm_api.cpp
@@ -45,7 +45,7 @@
 
 using json = nlohmann::json;
 
-static std::unique_ptr<causallm::Transformer> g_model;
+static std::unique_ptr<causallm::TransformerBase> g_model;
 static std::mutex g_mutex;
 static bool g_initialized = false;
 static std::string g_architecture = "";
--- a/Applications/CausalLM/api/causal_lm_api.cpp
+++ b/Applications/CausalLM/api/causal_lm_api.cpp
@@ -575,7 +575,7 @@ ErrorCode runModel(const char *inputTextPrompt, const char **outputText) {
     }
 
     // We assume single batch request for this API.
-    g_model->run(input, false, "", "", g_verbose);
+    g_model->run(input, "", "", nullptr, g_verbose);
 
     auto causal_lm_model = dynamic_cast<causallm::CausalLM *>(g_model.get());
     g_last_output = ""; // Reset last output
--- a/Applications/CausalLM/factory.h
+++ b/Applications/CausalLM/factory.h
@@ -15,7 +15,7 @@
 #define __CAUSALLM_FACTORY_H__
 
 #include <ostream>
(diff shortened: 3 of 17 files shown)


Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -285,8 +288,12 @@ void CausalLM::registerCustomLayers() {
   }
 }
 
-void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
-                   const WSTR tail_prompt, bool log_output) {
+void CausalLM::run(const WSTR prompt, void *output_buf, bool log_output) {
+  run(prompt, "", "", output_buf, log_output);
```
baek2sm: On Windows, WSTR expands to std::wstring, so the narrow "" arguments here are not convertible for Windows

jijoongmoon (approved):
LGTM

Seunghui98 (commented):
## Description

This PR extracts a `TransformerBase` abstract class from the existing `Transformer` and reshapes the `run()` interface across all variants (CausalLM, Bert, SentenceTransformer, TimmViT). The motivation — letting QNN-based variants coexist with the NNTrainer-based path without inheriting HF-config parsing logic — is reasonable, and the move of `do_sample` into `generation_config.json` plus the addition of an `output_buf` for programmatic result retrieval are both sensible directions. Overall the refactor is clean and the test updates / `causal_lm_api.cpp` migration look consistent.

That said, there are a few design and API-contract points worth a closer look before merging.

## Review Points

### 1. `void* output_buf` is type-unsafe (high priority)
`transformer_base.h:113-133` declares the output buffer as `void*` with the expected type encoded only in a docstring (`std::vector<std::string>*` for CausalLM, `std::vector<float*>*` for embedding/ViT). A wrong cast is silent UB. Consider:
- `std::variant<std::vector<std::string>*, std::vector<float*>*>`, or
- a small polymorphic `RunOutput` struct, or
- templating `run()` on the output type (with non-template virtual dispatch underneath).

This is the single biggest risk surface introduced by the PR — please double-check whether the type-erased approach is intentional or a placeholder.

### 2. Memory ownership of `float*` outputs
`bert_transformer.cpp:280-285`, `sentence_transformer.cpp:294-300`, `timm_vit_transformer.cpp:385-390` hand raw `float*` arrays to the caller and say "caller is responsible for deallocation," but the allocator (`new[]` vs `malloc`) is not documented. Please:
- Document the exact deallocation API (`delete[]`), or
- Wrap in `std::unique_ptr<float[]>` / return owning containers so the contract is enforced by the type system.

### 3. `TransformerBase` still holds NNTrainer-specific state
`transformer_base.h:154-167` puts `ModelHandle model`, `std::unique_ptr<tokenizers::Tokenizer> tokenizer`, `NUM_VOCAB`, `DIM`, `NUM_LAYERS`, `MAX_SEQ_LEN`, `BATCH_SIZE`, etc. as protected members of the base. If the goal is to let QNNTransformer inherit `TransformerBase` without dragging in NNTrainer, then `ModelHandle` (which is `std::unique_ptr<ml::train::Model>`) being in the base partially defeats the abstraction. Worth confirming whether QNN really needs all of these, or whether they belong in an intermediate `NNTrainerTransformerBase` layer.

### 4. `save_weight` overload throws by default
`transformer_base.h:91-103` provides a default implementation that throws `std::runtime_error`. This forces every subclass to override anyway, so making it pure virtual would be cleaner and gives a compile-time error instead of a runtime one. Alternatively, split the dtype-conversion variant into a separate optional interface.

### 5. `do_sample` is no longer runtime-controllable
`causal_lm.cpp:105-107` now reads `do_sample` once in `setupParameters` and stores it as `DO_SAMPLE`. Callers can no longer flip sampling on a per-call basis without re-initializing the model. Is this an intentional regression? If runtime control is still desired, add a setter (`setDoSample(bool)`) or restore it as an optional override parameter.

### 6. Two `run()` overloads vs. single one with defaults
Every variant now exposes both `run(prompt, output_buf, log_output)` and `run(prompt, system_prompt, tail_prompt, output_buf, log_output)`. Because the full overload already has defaults for `system_prompt` and `tail_prompt`, the simple overload is effectively a forwarder. This creates overload-resolution ambiguity risk (e.g., a future `WSTR`-convertible second arg) and doubles maintenance surface. Consider keeping only the full overload.

### 7. Doxygen tag inconsistency
`bert_transformer.h:106-109` and `transformer.h:103-107` use `@brief TransformerBase::run(...)` where it should be `@copydoc TransformerBase::run(...)` (the latter actually inherits the docstring). Minor, but worth fixing for consistency with the other `@copydoc` sites.

### 8. Potential memory-leak fix in `TimmViTTransformer::run`
`timm_vit_transformer.cpp:385-390` now deletes `output[]` when `output_buf == nullptr`. The previous code did not — was this an existing leak, or was the caller responsible somewhere else? If it's a leak fix, worth calling out in the commit message.

### 9. Confirm `output_list` is fully populated in `CausalLM::run`
`causal_lm.cpp:608-610` assigns `output_list` to `output_buf` near the end of `run()`. Please verify it is populated correctly in all early-termination paths (EOS reached, max-token reached) so callers passing a non-null buffer never observe truncated/empty results unexpectedly.

### 10. Public-API binary break
`causal_lm_api.cpp:48,560` changes the global `unique_ptr<Transformer>` to `unique_ptr<TransformerBase>` and the `run()` signature. If `causal_lm_api` is a published surface for downstream consumers (looks like it from the C-style `runModel`), this is an ABI break. Make sure the version bump / changelog reflects that, and consider whether a thin backwards-compat shim is desirable.

---

### Where to look next (priority order)
1. `transformer_base.h:113-133` — the `void*` contract (item 1).
2. `transformer_base.h:154-167` — what really belongs in the base vs. an NNTrainer-specific intermediate (item 3).
3. `causal_lm.cpp:105-107, 525, 562, 608-610` — `DO_SAMPLE` semantics and `output_list` completeness (items 5, 9).
4. `bert_transformer.cpp` / `sentence_transformer.cpp` / `timm_vit_transformer.cpp` — ownership/lifetime of returned `float*` (items 2, 8).
5. Header doxygen + double-overload cleanup (items 6, 7).

---

baek2sm:
**Final Verdict: ✅ Approve**

---

Extracts an abstract `TransformerBase` from `Transformer` so alternative model families share one interface, and streamlines the inference entry point (`run()` drops the per-call `do_sample`, gains a `void* output_buf` overload).

**Included features**
- New `TransformerBase` abstract base; `Transformer` and all concrete models (`CausalLM`, `SentenceTransformer`, `BertTransformer`, `TimmViTTransformer`) inherit it.
- `Factory` and `causal_lm_api` now operate on `TransformerBase`.
- `do_sample` is read once from `generation_config.json` and stored, removed from the `run()` signature.
- Fixes a `TimmViTTransformer` output memory leak when the caller does not take ownership.

---

### 2. Related Issues

- **P3** `void* output_buf` is type-erased; passing a wrong pointer type would be silent UB — an interface-design note for a follow-up.
- **P3** `causal_lm_api.cpp` hard-codes `do_sample=false` (pre-existing, unchanged).
- **P3** Minor comment typo ("dellocation") in `sentence_transformer.cpp`.

No P0/P1/P2 issues.

---

The single rebase conflict (`transformer.h`) was resolved by keeping upstream's `formatFromExtension` virtual while relocating the shared members into `TransformerBase`; the merged tree compiles (Windows MSVC 500/500). The change is confined to `Applications/CausalLM/` — `deberta_v2`, `timm_vit`, and `sentence_transformer` all inherit the new chain and build; no `nntrainer/` core code is touched. The full CI matrix is green.

---

Cleanly rebased onto the latest main with the only conflict resolved correctly, no regression to dependent models, and a fully green CI matrix. Approve.

PR #3830, opened 2026-04-01
Optimized Remaining Layernorm Kernels Using SIMD

b-saianirud:
Added SIMD optimizations for add_i, pow and used existing SIMD implementation for inverse sqrt. All other kernels that are used in Layernorm implementation are already using SIMD/BLAS optimizations.


jaemini-shin (changes requested):
`neon` is not the only backend, functions in `cpu_backend.h` should be architecture-agnostic. Could you please implement following cases?
* AVX for x86 backend
* fallback for non SIMD

Issue #3821, opened 2026-03-30
[Feature Proposal] Optimize x86 FP32/FP16 CPU Backend with AVX2 SIMD and Cache-Blocked FP16 GEMM


KWSMooBang:
## Current Status of the x86 CPU Backend

The x86 backend already has AVX2 implementations for several FP32 and FP32-oriented kernels, including `ele_mul`, `ele_add`, `transpose_matrix`, `swiglu`, `softmax_row`, `compute_kcaches`, `compute_fp16vcache_fp32_transposed`, `compute_rotary_emb_value`, and `rms_norm_wrt_width_fp32_intrinsic`.

However, there are still important gaps:

- Several FP32 kernels still rely on BLAS or fallback code paths, for example `ele_sub`, `ele_div`, `softmax`, `inv_sqrt_inplace`, `max_val`, and BLAS-backed `sgemv` / `sgemm` / `sdot` / `sscal` / `snrm2` / `isamax`.
- The FP16 backend in `x86_compute_backend_fp16.cpp` is still mostly fallback-based. Most FP16 arithmetic and utility functions such as `ele_*`, `sscal`, `snrm2`, `sdot`, `saxpy`, `softmax`, `swiglu`, `transpose_matrix`, and `compute_rotary_embedding_value` currently call `__fallback_*`.
- Within `x86_compute_backend_fp16.cpp`, SIMD coverage is limited mainly to contiguous FP16 <-> FP32 conversion (`vcvt_f16_f32`, `vcvt_f32_f16`) and `is_valid`.
- FP16 `sgemm` / `sgemv` currently allocate temporary FP32 buffers and perform full-buffer FP16 -> FP32 -> FP16 conversion around BLAS or fallback calls.

This results in:

- Unnecessary FP16 -> FP32 -> FP16 conversions
- Extra temporary allocations and memory traffic
- Limited SIMD utilization in the x86 FP16 path
- A wider performance and feature gap relative to the ARM FP16 backend

---

## Proposed Solution

Optimize the remaining x86 gaps mainly with AVX2-based SIMD kernels, using F16C conversion instructions where appropriate, while preserving the FP32 AVX2 kernels that already exist. The implementation plan is organized into the following three goals.

#### 1. Optimize remaining FP32 scalar / fallback operations with AVX2 SIMD
Replace high-impact FP32 fallback paths with AVX2 SIMD where it provides clear benefit
- [x] Target element-wise operations such as `ele_sub` and `ele_div`
[...]

PR #3834, opened 2026-04-02
[x86] feat: Add AVX2 implementation for x86 CPU Backend

YongHyeon02:
## Dependency of the PR

- Depends on: #3820 (benchmark tooling, not yet merged)
- Related to: #3821

<details><summary>feat: Add AVX2 implementation for x86 CPU Backend (f3b778e2)</summary><br />

- Add `avx2_impl.cpp` / `avx2_impl.h` with optimized FP32 AVX2 intrinsic implementations for:
  - `ele_sub`, `ele_div`, `softmax`, `tanh_gelu`, `tanh_gelu_mul`, `inv_sqrt_inplace`, `max_val`
- Integrate AVX2 dispatch into `x86_compute_backend.cpp`
- Add comprehensive unit tests verifying floating-point precision of all implemented routines

**Self evaluation:**

</details>

<details><summary>feat: Add SIMD implementation of sine, cosine function for x86 CPU Backend (32078f44)</summary><br />

- Add `avx2_mathfun.h` / `avx2_mathfun.hxx` providing vectorized sin/cos using AVX2 (ported from `neon_mathfun` approach)
- Refactor `tanh_gelu` AVX2 kernel to use SIMD trigonometric approximation instead of scalar fallback, further improving throughput

**Self evaluation:**

Signed-off-by: KWSMooBang <40kimwoosung@gmail.com>
</details>

<details><summary>feat: fix stride handling and add AVX2 paths for elementwise ops (811ae443)</summary><br />

- Fix incorrect stride handling in elementwise operations when `stridex != 1` or `stridey != 1`
- Add AVX2-accelerated paths for strided elementwise sub/div/mul/add
- Expand strided operation coverage from fallback-only to AVX2+fallback

**Self evaluation:**

Signed-off-by: SHPark524 <twinpsh1@gmail.com>
</details>

<details><summary>fix: correct X stride in elementwise ops fallback path (71b1cac1)</summary><br />

- Fix bug where `X` pointer was incorrectly advanced by `stridey` instead of `stridex` in fallback paths for `ele_sub` and `ele_div`

</details>

<details><summary>test: add missing tests for tanh_gelu_v2 and calc_trigonometric_vals_dup (c0b21501)</summary><br />

- Add unit tests for `calc_trigonometric_vals_dup` (SIMD sin/cos) and `tanh_gelu_v2` (AVX2 tanh_gelu using SIMD trig)
- Verify numerical correctness against scalar reference

</details>

<details><summary>fix: correct stride handling and zero-input behavior in AVX2 kernels (9e8906a9)</summary><br />

- Fix `ele_sub`/`ele_div` AVX2 fast path to only use contiguous vector loads when `i_stride == 1`; fall back to scalar for `i_stride > 1`
- Fix `inv_sqrt_inplace` AVX2 Newton-Raphson to produce `inf` (not `NaN`) for zero inputs, matching scalar path behavior
- Skip NYI `rms_norm_fp16_template_float` test with `GTEST_SKIP()` instead of letting it throw

**Self evaluation:**

</details>

<details><summary>fix: fix test loop indices and optimize inv_sqrt with FMA (f533d803)</summary><br />

- Fix `ele_sub`/`ele_div` test loop starting index from `i = -1` to `i = 0`
- Optimize `inv_sqrt_inplace` Newton-Raphson refinement using FMA (`_mm256_fnmadd_ps`): `y * (1.5 - 0.5*x*y²)` reduces dependency chain length from 3 to 2

**Self evaluation:**

</details>

## Performance Improvements

Benchmark environment: Intel Core i5-12400F (6C/12T, AVX2), Linux WSL2
Benchmark tool: `benchmark_x86_backend` (iters=1000, warmup=50)
Build: `enable-fp16=false`, GCC 11.4 (identical config for both baseline and AVX2)

### FP32 Activation & Element-wise Operations

| Function | Size | Baseline (Avg) | AVX2 (Avg) | Speedup | AVX2 Throughput |
|:---------|:-----|---------------:|-----------:|--------:|----------------:|
| ele_sub | N=256 | 226 ns | 71 ns | **3.2x** | 3.61 Gelem/s |
| ele_sub | N=1024 | 1.09 us | 129 ns | **8.4x** | 7.92 Gelem/s |
| ele_sub | N=4096 | 3.22 us | 395 ns | **8.2x** | 10.38 Gelem/s |
| ele_div | N=256 | 225 ns | 102 ns | **2.2x** | 2.50 Gelem/s |
| ele_div | N=1024 | 885 ns | 191 ns | **4.6x** | 5.36 Gelem/s |
| ele_div | N=4096 | 3.87 us | 689 ns | **5.6x** | 5.95 Gelem/s |
| softmax | N=256 | 2.39 us | 197 ns | **12.1x** | 1.30 Gelem/s |
| softmax | N=1024 | 9.59 us | 682 ns | **14.1x** | 1.50 Gelem/s |
| softmax | N=4096 | 45.31 us | 2.62 us | **17.3x** | 1.56 Gelem/s |
| tanh_gelu | N=256 | 3.54 us | 113 ns | **31.3x** | 2.27 Gelem/s |
| tanh_gelu | N=1024 | 14.87 us | 363 ns | **41.0x** | 2.82 Gelem/s |
| tanh_gelu | N=4096 | 54.19 us | 1.39 us | **39.0x** | 2.95 Gelem/s |
| tanh_gelu_mul | N=256 | 2.57 us | 124 ns | **20.7x** | 2.07 Gelem/s |
| tanh_gelu_mul | N=1024 | 10.72 us | 386 ns | **27.8x** | 2.65 Gelem/s |
| tanh_gelu_mul | N=4096 | 51.95 us | 1.42 us | **36.6x** | 2.88 Gelem/s |
| inv_sqrt | N=256 | 413 ns | 92 ns | **4.5x** | 2.79 Gelem/s |
| inv_sqrt | N=1024 | 1.61 us | 228 ns | **7.1x** | 4.49 Gelem/s |
| inv_sqrt | N=4096 | 6.32 us | 806 ns | **7.8x** | 5.08 Gelem/s |
| max_val | N=256 | 288 ns | 70 ns | **4.1x** | 3.67 Gelem/s |
| max_val | N=1024 | 1.12 us | 157 ns | **7.1x** | 6.53 Gelem/s |
| max_val | N=4096 | 4.45 us | 556 ns | **8.0x** | 7.36 Gelem/s |

### Highlights

- **tanh_gelu**: Up to **41.0x** speedup at N=1024 — biggest gain from replacing scalar math with SIMD sin/cos approximation
- **tanh_gelu_mul**: Up to **36.6x** speedup at N=4096
- **softmax**: Up to **17.3x** speedup at N=4096 (from 45.31 us to 2.62 us)
- **inv_sqrt_inplace**: Up to **7.8x** speedup at N=4096 (from 6.32 us to 806 ns) — uses FMA-based Newton-Raphson with shorter dependency chain
- **max_val**: Up to **8.0x** speedup at N=4096 (from 4.45 us to 556 ns)
- **ele_sub**: Up to **8.4x** at N=1024; **ele_div**: Up to **5.6x** at N=4096

### Known Limitations

- **RMS Norm**: `rms_norm_wrt_width_fp16_intrinsic` is NYI for x86 (throws exception). Planned for separate PR.
- **FP16 on x86**: Uses software emulation (convert FP16<->FP32 per element). Not accelerated by this PR.

## Summary

This PR introduces AVX2-optimized FP32 compute kernels for the x86 CPU backend, delivering **2x-41x speedups** across arithmetic, activation, and reduction operations. Key changes:

1. **AVX2 intrinsic implementations** (`avx2_impl.cpp/h`): `ele_sub`, `ele_div`, `softmax`, `tanh_gelu`, `tanh_gelu_mul`, `inv_sqrt_inplace`, `max_val`
2. **SIMD sin/cos approximation** (`avx2_mathfun.h/hxx`): Vectorized trigonometric functions enabling fast `tanh_gelu` computation
3. **Strided elementwise operations**: AVX2 paths with correct stride handling for non-contiguous memory layouts
4. **FMA-optimized Newton-Raphson**: `inv_sqrt_inplace` uses `_mm256_fnmadd_ps` to reduce dependency chain length
5. **Comprehensive unit tests**: Precision validation for all implemented routines

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -79,7 +79,7 @@
 #define _nnt_ATTR_ALWAYS_INLINE [[gnu::always_inline]]
 #endif
 
-#if __has_cpp_attribute(unikely)
+#if __has_cpp_attribute(unlikely)
 #define UNLIKELY [[unlikely]]
 #else
 #define UNLIKELY
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -328,6 +328,10 @@ avx2_approx_swiglu_alpha(__m256 x, __m256 s, __m256 alpha) noexcept -> __m256 {
 
 namespace nntrainer::avx2 {
 
+// Forward declarations for internal helpers used across the file
+static inline __m256 exp256_ps(__m256 x);
+static float hsum_avx(__m256 v);
+
 /**
  * @brief struct of q4_0x8 block
  */
@@ -701,16 +705,25 @@ static inline void convert_q4_0x8_noshuffle(const void *src,
 
 // ================== wrappers for your K,N combinations ==================
 // K = 3072 (UNIT = 768)
+/**
+ * @brief convert_q4_0x8_shuffle for K=3072 N=98304
+ */
 void convert_q4_0x8_shuffle_K3072_N98304(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = (N*8)/UNIT = 1024
   convert_q4_0x8_noshuffle<768, 1024>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=3072 N=36864
+ */
 void convert_q4_0x8_shuffle_K3072_N36864(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = 384
   convert_q4_0x8_noshuffle<768, 384>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=3072 N=3072
+ */
 void convert_q4_0x8_shuffle_K3072_N3072(const void *src, uint16_t *d_out,
                                         uint8_t *qs_out) {
   // groups = 32
@@ -718,16 +731,25 @@ void convert_q4_0x8_shuffle_K3072_N3072(const void *src, uint16_t *d_out,
 }
 
 // K = 8192 (UNIT = 2048)
+/**
+ * @brief convert_q4_0x8_shuffle for K=8192 N=98304
+ */
 void convert_q4_0x8_shuffle_K8192_N98304(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = 384
   convert_q4_0x8_noshuffle<2048, 384>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=8192 N=36864
+ */
 void convert_q4_0x8_shuffle_K8192_N36864(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = 144
   convert_q4_0x8_noshuffle<2048, 144>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=8192 N=3072
+ */
 void convert_q4_0x8_shuffle_K8192_N3072(const void *src, uint16_t *d_out,
                                         uint8_t *qs_out) {
   // groups = 12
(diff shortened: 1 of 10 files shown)


KWSMooBang:
## Add Features

feat: Add AVX2 SIMD implementation for sine and cosine functions in x86 CPU Backend
- Add new AVX2 math function headers for SIMD vectorization (`avx2_mathfun.h`, `avx2_mathfun.hxx`). Support AVX2 acceleration for math functions `sine` and `cosine` utilizing optimized intrinsic instructions (`avx2_impl.cpp`, `avx2_impl.h`).

## Performance Improvements

Benchmark evaluation shows significant speedups 'sine' and 'cosine' function

| Function | Size | Baseline (Avg) | After P1 (Avg) | Speedup | Baseline Throughput | After P1 Throughput |
|----------|------|----------------|----------------|---------|---------------------|---------------------|
| sine     | N=256 | 4.87 us       | 312 ns         | 15.6x   | 52.56 Melem/s       | 820.51 Melem/s      |
| sine     | N=1024| 10.64 us      | 1.41 us        | 7.5x    | 96.24 Melem/s       | 726.24 Melem/s      |
| sine     | N=4096| 28.33 us      | 5.42 us        | 5.2x    | 144.58 Melem/s      | 755.72 Melem/s      |
| cosine   | N=256 | 1.62 us       | 297 ns         | 5.4x    | 158.02 Melem/s      | 861.95 Melem/s      |
| cosine   | N=1024| 6.07 us       | 1.24 us        | 4.8x    | 168.70 Melem/s      | 825.81 Melem/s      |
| cosine   | N=4096| 18.70 us      | 3.69 us        | 5.0x    | 219.04 Melem/s      | 1.11 Gelem/s        |

SHpark524:
## Add features

feat: AVX2 elementwise ops
- Extend AVX2 support for mul/add/sub/div to general (alpha, beta) cases.
- Restrict AVX2 to contiguous/broadcast inputs (i_stride 0 or 1).
- Fix incorrect pointer increment in fallback path.

baek2sm (approved):
LGTM

test/unittest/unittest_nntrainer_cpu_backend.cpp
```diff
@@ -1439,6 +1460,293 @@ DECLARE_transform_int4_test_K_N(1024, 648, 64);
...
 DECLARE_transform_int4_test_K_N(3072, 8192, 32);
 
+// ============================================================================
+// P1: AVX2 replacement tests for formerly-fallback FP32 functions
+// ============================================================================
+
+static void run_ele_sub_test(const unsigned int N, float alpha, float beta,
+                             unsigned int i_stride, unsigned int o_stride) {
+  const int TEST_CNT = 20;
+  for (int i = -1; i < TEST_CNT; i++) {
```
jaemini-shin:
```suggestion
  for (int i = 0; i < TEST_CNT; i++) {
```

test/unittest/unittest_nntrainer_cpu_backend.cpp
```diff
@@ -1439,6 +1460,293 @@ DECLARE_transform_int4_test_K_N(1024, 648, 64);
...
+}
+
+TEST(nntrainer_cpu_backend_standalone, ele_sub_3072_alpha1_beta0_istr_2) {
+  run_ele_sub_test(3072, 1.f, 0.f, 2, 1);
+}
+
+static void run_ele_div_test(const unsigned int N, float alpha, float beta,
+                             unsigned int i_stride, unsigned int o_stride) {
+  const int TEST_CNT = 20;
+  for (int i = -1; i < TEST_CNT; i++) {
```
jaemini-shin:
```suggestion
  for (int i = 0; i < TEST_CNT; i++) {
```

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -327,6 +327,10 @@ avx2_approx_swiglu_alpha(__m256 x, __m256 s, __m256 alpha) noexcept -> __m256 {
 
 namespace nntrainer::avx2 {
 
+// Forward declarations for internal helpers used across the file
+static inline __m256 exp256_ps(__m256 x);
+static float hsum_avx(__m256 v);
+
```
jaemini-shin, line 333: These lines are unnecessary, all the new functions are added at the end.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -1145,6 +1167,222 @@ void gelu_v2(const unsigned int N, const float *X, float *Y) {
...
+                                     (y + 0.044715f * y * y * y)));
+    X[i] = gelu_y * Z[i];
+  }
+}
+
+void tanh_gelu_v2_mul(const unsigned int N, float *X, float *Y, float *Z) {
+  tanh_gelu_mul(N, X, Y, Z);
+}
+
+float max_val(const unsigned int N, float *X) {
```
jaemini-shin, line 1209:
```suggestion
static inline float max_val(const unsigned int N, float *X) {
```
The only usage is softmax, let's reduce function call overhead.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -1145,6 +1167,222 @@ void gelu_v2(const unsigned int N, const float *X, float *Y) {
...
+  const __m256 half = _mm256_set1_ps(0.5f);
+
+  for (; i + 8 <= N; i += 8) {
+    __m256 x = _mm256_loadu_ps(&X[i]);
+    __m256 is_zero = _mm256_cmp_ps(x, zero, _CMP_EQ_OQ);
+    __m256 est = _mm256_rsqrt_ps(x);
+    // Newton-Raphson: y = 0.5 * y * (3 - x * y * y)
+    __m256 xy2 = _mm256_mul_ps(x, _mm256_mul_ps(est, est));
+    __m256 refined = _mm256_mul_ps(
+      _mm256_mul_ps(half, est), _mm256_sub_ps(three, xy2));
```
jaemini-shin:
```suggestion
  const __m256 zero = _mm256_setzero_ps();
  const __m256 inf_val = _mm256_set1_ps(INFINITY);
  const __m256 three_half = _mm256_set1_ps(1.5f);
  const __m256 half = _mm256_set1_ps(0.5f);

  for (; i + 8 <= N; i += 8) {
    __m256 x = _mm256_loadu_ps(&X[i]);
    __m256 is_zero = _mm256_cmp_ps(x, zero, _CMP_EQ_OQ);
    __m256 est = _mm256_rsqrt_ps(x);
    // Newton-Raphson: y = y * (1.5 - 0.5 * x * y * y)
    __m256 half_x = _mm256_mul_ps(one_half, x);
    __m256 yy = _mm256_mul_ps(y, y);
    __m256 refined = _mm256_fnmadd_ps(
      half_x, yy, three_half);
```
This one might result better latency.
* reduce vector chain length: 3 to 2
* use FMA

jaemini-shin (commented):
Could you please check the followings?

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
YongHyeon02: Thanks for the suggestion, but `max_val` is declared in `avx2_impl.h` and called externally from `cpu_backend.cpp` (via `nntrainer::avx2::max_val`), so making it `static inline` would break linkage.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
YongHyeon02: These forward declarations are needed because `softmax` (line 1234) calls `exp256_ps` (line 1748) and `hsum_avx` (line 2363), which are defined later in the file. Removing them without moving the definitions would cause a compile error.

jaemini-shin: Sorry for lack of information. First of all, those I didn't catch that `softmax` is before `exp256_ps` and `hsum_avx`. But using forward declaration can make code hard to maintain.
So, here's my suggestion:
* Are you going to expose these helpers? If not, it recommend to use anonymous namespace. There's no reason to put them under `nntrainer::avx2` namespace!

YongHyeon02: Thanks for pointing it out! The helpers are planned to be reused in #3821, where we'll implement AVX2+F16C SIMD kernels for FP16 in `avx2_impl_fp16.cpp`. Rather than placing them in an anonymous namespace, How about moving them into a shared internal header `avx2_internal.h` under namespace `nntrainer::avx2::internal` ?

jaemini-shin: Since AVX2 doesn't have FP16 processing unit, you should reuse most of the FP32 AVX2 functions. In this point of view, making new file might become a [shot gun surgery](https://refactoring.guru/smells/shotgun-surgery).

How do you think about following process?
* move `exp256_ps` into `avx2_mathfun.hxx`
* reorder the functions into similar groups(e.b. softmax family, swiglu family, ...)
* remove static for reused functions
* implement wrappers for FP16 mode

YongHyeon02: Thanks for the feedback — the shotgun surgery concern is a fair point. Let me share some thoughts before making changes.

**On the wrapper approach for FP16:**
Pure wrappers have a performance cost for compute-heavy kernels. Consider `softmax`: a wrapper would need FP16→FP32 conversion of the full input, then `softmax` on FP32, then FP32→FP16 conversion of the output — that's 4 memory passes plus two temporary FP32 buffers. In contrast, inline F16C conversion inside the SIMD loop (FP16 load → `_mm256_cvtph_ps` → compute via `exp256_ps`/`hsum_avx` → `_mm256_cvtps_ph` → FP16 store) does the same work in 2 memory passes with zero extra allocations. 

**On `avx2_mathfun.hxx`:** 
That file is cephes-based sin/cos port under the zlib license. Placing our project-specific helpers (`exp256_ps`, `hsum_avx`, `rcp_ps`, GELU polynomials, swiglu helpers, etc.) alongside third-party code would blur the licensing boundary, and most of those aren't really "math functions".

**Proposal:** 
  Would it work to introduce a separate internal header that collects the shared SIMD helpers as I proposed before?

The higher-level compute functions (`softmax`, `ele_*`, `swiglu`, etc.) would still have type-specialized implementations for performance, but they'd share the low-level primitives. Reordering by functional group (softmax family, swiglu family, …) is orthogonal and I'm happy to apply that as well. Let me know if this direction works for you.

jaemini-shin: Thanks for reply.

> **On the wrapper approach for FP16:** 

You're right. What I intended was same as yours, but I missed some point. :sob: 

> **On `avx2_mathfun.hxx`:** That file is cephes-based sin/cos port under the zlib license. Placing our project-specific helpers (`exp256_ps`, `hsum_avx`, `rcp_ps`, GELU polynomials, swiglu helpers, etc.) alongside third-party code would blur the licensing boundary, and most of those aren't really "math functions".

Fortunately, it doesn't be a problem. We already use this approach for arm neon.

> **Proposal:** Would it work to introduce a separate internal header that collects the shared SIMD helpers as I proposed before?

As you supposed, introducing a shared helper won't be a big problem if we put only internal functions that doesn't need to expose to outside. If not, we should check several header files to find definitions.(`avx2_impl.h`, new helper header, ...)!

I'm looking forward for your next work. Please keep going on! :smile:

jaemini-shin (approved):
LGTM!

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
YongHyeon02: Thanks for your reply!

jaemini-shin:
It'd be better to proceed on another PR. For new commits, it requires another review!

YongHyeon02:
> It'd be better to proceed on another PR. For new commits, it requires another review!

Okay, I'll handle that commit in a separate PR!

jijoongmoon:
Could you check if you didn't add those two files, /usr/include/nntrainer/avx2_mathfun.h  and  /usr/include/nntrainer/avx2_mathfun.hxx for Tizen and Ubuntu packaging?

YongHyeon02:
> Could you check if you didn't add those two files, /usr/include/nntrainer/avx2_mathfun.h and /usr/include/nntrainer/avx2_mathfun.hxx for Tizen and Ubuntu packaging?

I've added both headers to `packaging/nntrainer.spec` and `debian/nntrainer-dev.install`.

baek2sm:
**Review verdict: ✅ Approve**

### 1. Summary

This PR adds 256-bit AVX2 SIMD kernels to the x86 CPU backend for several FP32 entry points that previously fell through to the scalar fallback, and fixes a packaging gap where two newly-installed math-helper headers were missing from the Tizen spec and Debian dev manifest.

Included:
- AVX2 implementations in `nntrainer::avx2` for `tanh_gelu`, `tanh_gelu_mul`, `tanh_gelu_v2_mul`, `max_val`, `softmax`, `inv_sqrt_inplace`, `sine`, `cosine`, `calc_trigonometric_vals_dup`, `ele_sub`, `ele_div`, plus AVX2 fast-paths for the stride-1/broadcast cases of `ele_mul`/`ele_add`.
- New `avx2_mathfun.h`/`.hxx` carrying a Cephes-based `sincos256_ps` (adapted from `neon_mathfun`, zlib-licensed, correctly attributed).
- Scalar `__fallback_calc_trigonometric_vals_dup<float>` replacing a prior hard `throw`, so non-NEON / non-AVX2 platforms can run rotary-embedding instead of crashing.
- Packaging fix declaring the two new headers in `nntrainer.spec` (`%ifarch %{ix86} x86_64`) and `debian/nntrainer-dev.install`.
- GTest unit tests comparing every new AVX2 kernel against the scalar reference (MSE ≤ 1e-5), including zero/boundary cases and broadcast (stride-0) variants.

### 2. Notes (non-blocking)

- `test/unittest/.../unittest_nntrainer_cpu_backend.cpp` — the `rms_norm_fp16_template_float` test is a permanent `GTEST_SKIP()` stub for out-of-scope functionality. Harmless (skip ≠ failure) but reduces test-suite signal; consider removing or tracking via an issue. (P3)
- The AVX2 fast-path for the elementwise ops only covers `o_stride == 1 && i_stride ∈ {0,1}`; the non-unit-stride tests therefore exercise the scalar path. Not a bug, but the test names are slightly misleading. (P3)

### 3. Stability for integration with latest main

- **Git-mergeability**: clean. Merge-base is `ffd354ef80d`; all upstream/main commits since branch-off touch unrelated modules (CausalLM, safetensors, ChatTemplate, doxygen tidy). No file-level or semantic conflicts with `avx2_impl.*`, `fallback_internal.*`, or the cpu-backend unit test.
- **Cross-platform**: the x86 subdir compiles only when `host_machine.cpu_family() ∈ {x86, x86_64}`, so ARM/Android(NDK)/Tizen-aarch64 builds never see the changed files. AVX2 macros are set unconditionally on x86 by both MSVC (`/arch:AVX2`) and GCC/Clang (`-mavx2 -mfma`). The new headers are only reachable from x86 TUs.
- **Packaging**: the second commit resolves a real Tizen GBS x86_64 / Ubuntu pdebuild "Installed but unpackaged files" failure that the first commit would otherwise introduce.

Checks executed: full diff + surrounding-code analysis, meson architecture gating, upstream/main divergence, packaging-script logic, doxygen/license headers, commit hygiene (signed-off, body length, author). The Windows MSVC build reasoning confirms portability; arm/clang legs reasoned about, not locally executed.

### Overall

Well-scoped, correctly gated so ARM/Android/Tizen builds are unaffected, adequately tested, and it fixes a packaging defect that would have broken Tizen/Debian CI. Ready to merge.

baek2sm:
**Final Verdict: ✅ Approve**

---

Adds 256-bit AVX2 SIMD kernels for the x86 CPU backend's FP32 entry points, with a scalar fallback fix for non-SIMD platforms.

**Included features**
- AVX2 kernels: `tanh_gelu` (+`_mul`/`_v2_mul`), `max_val`, `softmax`, `inv_sqrt_inplace`, `sine`/`cosine`/`calc_trigonometric_vals_dup`, `ele_sub`/`ele_div`, plus an AVX2 fast path for `ele_mul`/`ele_add`.
- Cephes-style `sincos256_ps` helper (`avx2_mathfun.h/.hxx`).
- Fixes `__fallback_calc_trigonometric_vals_dup<float>` to compute instead of throwing on non-NEON platforms.
- GTest coverage of every new kernel against the scalar reference; Tizen/Debian manifests updated.

---

### 2. Related Issues

- **P1 (fixed)** Added the missing `// SPDX-License-Identifier: Apache-2.0` tag to `avx2_mathfun.h` and `avx2_mathfun.hxx` (static-check requirement), retaining the zlib/Cephes attribution block.
- **P2 (fixed)** Corrected an `__has_cpp_attribute(unikely)` typo to `unlikely` so the branch-prediction hint applies.

No remaining P0/P1 issues.

---

Dispatch is wired through `nntrainer::avx2::*` in `x86_compute_backend.cpp` with no `#ifdef` at call sites and no change to the default-throw / `supports_*()` contract. The `-mavx2 -mfma` flags are already applied to the whole x86 build, and the x86 subdir is gated to `x86`/`x86_64` host families, so ARM/Android/Tizen-aarch64 are isolated. No dependency on #3820. The full CI matrix is green.

---

A clean, well-isolated AVX2 backend addition with correct dispatch wiring and scalar fallback; the SPDX and typo issues are fixed and the full CI matrix is green. Approve.

PR #3842, opened 2026-04-03
[Qnn] fix codes to support qnn

EunjuYang:
Please note that this PR contains cherry-pick from #3826

### DIFF
--- a/nntrainer/layers/input_layer.cpp
+++ b/nntrainer/layers/input_layer.cpp
@@ -71,16 +71,17 @@ void InputLayer::exportTo(Exporter &exporter,
 }
 
 void InputLayer::finalize(InitLayerContext &context) {
+  TensorDim::DataType input_dtype =
+    context.getInputDimensions()[0].getDataType();
+  TensorDim::DataType output_dtype = context.getActivationDataType();
 
   std::vector<TensorDim> output_dims = context.getInputDimensions();
   for (auto &d : output_dims) {
-    d.setDataType(context.getActivationDataType());
+    d.setDataType(output_dtype);
   }
-
   context.setOutputDimensions(output_dims);
-  is_inplace = true;
-  if (context.getActivationDataType() != ml::train::TensorDim::DataType::FP32)
-    is_inplace = false;
+
+  is_inplace = input_dtype == output_dtype ? true : false;
 }
 
 void InputLayer::updateTensorsByInputDimensions(
--- a/nntrainer/layers/input_layer.cpp
+++ b/nntrainer/layers/input_layer.cpp
@@ -74,13 +74,8 @@ void InputLayer::exportTo(Exporter &exporter,
 
 void InputLayer::finalize(InitLayerContext &context) {
   TensorDim::DataType input_dtype =
-    std::get<props::InputTensorDataType>(input_props).empty()
-      ? context.getInputDimensions()[0].getDataType()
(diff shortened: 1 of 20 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -1066,6 +1067,90 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
   return out;
 }
```
haehun, line 1069: I wonder if we could re-use code from normal `NeuralNetwork::inference` later

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -600,6 +620,9 @@ template <typename T> void UIntTensor<T>::copy(const void *buf) {
     const uint16_t *data = (const uint16_t *)buf;
     uint16_t *rdata = (uint16_t *)getData();
     copy_u16((const unsigned int)size(), data, rdata);
+#ifdef ENABLE_NPU
+    return;
+#endif
```
haehun, line 625: This is issue on NPU side. They should follow `uint_tensor` format like other contexts. I'll try fixing them later.

haehun (approved):
Good to go for now

nntrainer/layers/input_layer.cpp
```diff
@@ -33,7 +33,9 @@ namespace nntrainer {
 static constexpr size_t SINGLE_INOUT_IDX = 0;
 
 InputLayer::InputLayer() :
-  Layer(), input_props(props::Normalization(), props::Standardization()) {}
+  Layer(),
+  input_props(props::Normalization(), props::Standardization(),
+              props::InputTensorDataType(), props::TensorDataType()) {}
```
jijoongmoon, line 38: So from this PR, we can set the input tensor type then.?

nntrainer/models/neuralnet.cpp
```diff
@@ -1044,7 +1043,9 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
   if (!validateInput(X))
     throw std::invalid_argument("Input validation failed.");
 
+#ifndef ENABLE_NPU
   allocate(ExecutionMode::INFERENCE);
```
jijoongmoon, line 1047: I'm not sure if it is right or not. even it is, we do need allocate with rpcmem in memory pool even for the input and output.

I think it is not a good idea using #if def for ENABLE_NPU. Let's visit later.

nntrainer/layers/input_layer.cpp
EunjuYang: @lhs8928 @haehun , Could you please answer to this comment?

nntrainer/models/neuralnet.cpp
EunjuYang: @haehun , Could you please answer to this comment?

nntrainer/models/neuralnet.cpp
```diff
@@ -1066,6 +1067,90 @@ sharedConstTensors NeuralNetwork::inference(sharedConstTensors X,
   return out;
 }
 
+std::vector<IO_TensorType>
+NeuralNetwork::inference(unsigned int batch_size,
+                         const std::vector<IO_TensorType> &input,
+                         const std::vector<IO_TensorType> &label) {
```
jijoongmoon, line 1073: Wonder whether is we do need this. There are many inference functions like inference, incremental inference, and this IO_TensorType Inference. we need to restructuring this.

PR #3847, opened 2026-04-07
Implement SwiGLU layer with proper forward and backward pass

niket-agarwal:
- Add forwarding for training mode which internally calls incremental_forwarding
- Add calcDerivative implementation for backpropagation support
- Update unit tests to validate the implementation
- Move swiglu from Applications/CausalLM to nntrainer/layers

## Files Modified
-`Applications/CausalLM/layers/meson.build`
-`Applications/CausalLM/layers/swiglu.h`
-`Applications/CausalLM/meson.build`
-`Applications/CausalLM/models/transformer.cpp`
-`nntrainer/layers/meson.build`
-`nntrainer/layers/swiglu.cpp`
-`nntrainer/layers/swiglu.h`
-`test/unittest/layers/meson.build`
-`test/unittest/layers/unittest_layers_swiglu.cpp`
-`Applications/CausalLM/jni/Android.mk‎`
-`debian/nntrainer-dev.install`
-`packaging/nntrainer.spec`


Applications/CausalLM/jni/Android.mk
```diff
@@ -80,7 +80,7 @@ LOCAL_SRC_FILES := \
     ../models/qwen3_moe/qwen_moe_layer.cpp \
     ../layers/reshaped_rms_norm.cpp \
     ../layers/rms_norm.cpp \
-    ../layers/swiglu.cpp \
+    ../../../nntrainer/layers/swiglu.cpp \
```
EunjuYang: I think this line is not needed. Since you moved the layer to nntrainer's layer, if you register the layer to app_context, then you don't need to seperately add this code in here.

EunjuYang (changes requested):
1. **Missing factory registration** in `app_context.cpp`

Now that SwiGLULayer has been promoted from a CausalLM custom layer to a built-in nntrainer layer (added to layer_sources in nntrainer/layers/meson.build), it should also be registered with the default `AppContext`, just like the other built-in layers in the same directory (e.g. ChannelShuffle, Upsample2dLayer).

Note that `LAYER_SWIGLU` and the public factory helper `ml::train::layer::Swiglu(...)` already exist in `api/ccapi/include/layer.h`:
```
LAYER_SWIGLU = ML_TRAIN_LAYER_TYPE_SWIGLU, /**< Swiglu Layer type */
...
inline std::unique_ptr<Layer>
Swiglu(const std::vector<std::string> &properties = {}) {
  return createLayer(LayerType::LAYER_SWIGLU, properties);
}
```
However, without a corresponding registerFactory call in app_context.cpp, any caller going through the public API (or loading a model that references type=swiglu) will fail to resolve the factory — the layer is only reachable today because Applications/CausalLM/models/transformer.cpp::registerCustomLayers() still manually registers it.

Please add the registration alongside the other built-in layers in nntrainer/app_context.cpp:
```
#include <swiglu.h>
...
registerFactory(nntrainer::createLayer<SwiGLULayer>, SwiGLULayer::type,
                LayerType::LAYER_SWIGLU);
```

2. Once that is in place, the manual `registerFactory(nntrainer::createLayer<nntrainer::SwiGLULayer>)` in `Applications/CausalLM/models/transformer.cpp` becomes **redundant** and should be **removed** to avoid a double-registration attempt (and to finish the "move from CausalLM to nntrainer/layers" intent stated in the commit message).

Applications/CausalLM/models/transformer.cpp
```diff
@@ -435,7 +435,8 @@ void Transformer::registerCustomLayers() {
     static_cast<nntrainer::AppContext *>(ct_engine.getRegisteredContext("cpu"));
 
   try {
-    app_context->registerFactory(nntrainer::createLayer<causallm::SwiGLULayer>);
+    app_context->registerFactory(
+      nntrainer::createLayer<nntrainer::SwiGLULayer>);
```
EunjuYang: You don't need to register nntrainer's layer here once you add the layer in the app_context.

Applications/CausalLM/jni/Android.mk
niket-agarwal: Fixed in the latest commit. Thanks!

Applications/CausalLM/models/transformer.cpp
niket-agarwal: Right! Fixed, thanks :)

nntrainer/layers/swiglu.h
```diff
@@ -17,91 +17,76 @@
...
-    return SwiGLULayer::type;
-  };
+  const std::string getType() const override { return SwiGLULayer::type; };
 
   /**
    * @copydoc Layer::setProperty(const std::vector<std::string> &values)
    */
-  WIN_EXPORT void
-  setProperty(const std::vector<std::string> &values) override{};
+  void setProperty(const std::vector<std::string> &values) override {};
```
EunjuYang: I found that the CI failure is because of this part in `swiglu.h`:
```suggestion
  void exportTo(Exporter &exporter,
                const ml::train::ExportMethods &method) const override{};

  /**
   * @copydoc Layer::getType()
   */
  const std::string getType() const override { return SwiGLULayer::type; };

  /**
   * @copydoc Layer::setProperty(const std::vector<std::string> &values)
   */
  void setProperty(const std::vector<std::string> &values) override{};
```

niket-agarwal: Resolved. Thanks :)

EunjuYang (commented):
# PR 3847 Review — SwiGLU Layer Promotion with Forward/Backward Implementation

## Overview
- Moves `Applications/CausalLM/layers/swiglu.{cpp,h}` into `nntrainer/layers/`
- Changes namespace `causallm` → `nntrainer`, removes `PLUGGABLE` and `WIN_EXPORT` macros
- Fills in the previously empty `forwarding()` by delegating to `incremental_forwarding`
- Implements `calcDerivative()` for FP32 (throws for FP16)
- Registers `LAYER_SWIGLU` factory in `app_context.cpp`, adds one unit test
- Updates Android.mk / meson / deb / spec manifests

## Strengths
- **Mathematically correct gradient.** The expansion `Swish'(g) = σ(g) + g·σ(g)·(1−σ(g))` is accurate, and reusing `d_gate` as a scratch buffer keeps extra allocations down to a single `sig_gate` tensor.
- Follows the existing pattern (as with `LAYER_FC` / `LAYER_ADDITION`) of registering the same `LayerType` in both `app_context` (CPU) and `cl_context` (OpenCL) — no conflict with `SwiGLULayerCl`.
- `LayerType::LAYER_SWIGLU` and `ML_TRAIN_LAYER_TYPE_SWIGLU` are already defined in `api/ccapi/include/layer.h`, so no ABI additions are required.
- Removes `#pragma once` while keeping the existing `#ifndef __SWIGLU_LAYER_H__` guard — consistent with project convention.
- The golden generator in `gen_layer_tests.py` already emits `swiglu` for shape `(2,3,3,3)×(2,3,3,3)`, matching the unit test input.

## Main Issues

### 1. `forwarding()` looks batch-size-1 only (important)
```cpp
// For batch size = 1 training, call incremental_forwarding with the actual input length
auto to = in1.getDim().height();
incremental_forwarding(context, 0, to, training);
```
The comment itself says "batch size = 1." Since `to` is derived from `height()` only, and `incremental_forwarding` is sequence-length-slice based, this may misbehave or not cover the full batch when `batch > 1`. Now that the layer is promoted to core, it needs to work in general batched training. At minimum, turn the comment into an explicit TODO or add an assert/throw when that assumption is violated.

### 2. FP16 backward is unimplemented
`calcDerivative` throws `std::invalid_argument` on FP16, yet `supportBackwarding()` returns `true`, so the failure surfaces only at runtime. Document this in the header, and ideally check the data type in `finalize()` to fail earlier and more clearly.

### 3. Leftover SwiGLU duplicates
- `Applications/LLaMA/jni/swiglu.{cpp,h}` still exists — another SwiGLU copy.
- The PR description only mentions the CausalLM path.
- While consolidating CausalLM, it would be good to also migrate LLaMA to use `nntrainer::SwiGLULayer`, or to state explicitly why it is kept. Otherwise there are three copies (core, CL, LLaMA/jni).

### 4. Docs / comments
- The header brief still says `"A SwiGLU layer for llama."`. Since it is now a general-purpose layer, please update the wording.
- `~SwiGLULayer() {}` can be `= default;` for consistency with the modernized constructor.
- `updateTensorsByInputDimensions` never uses its `input_dimensions` parameter — if this is an interface constraint, mark it with `/* input_dimensions */` or add a comment explaining why.

### 5. Allocation / performance
- `nntrainer::Tensor sig_gate(gate.getDim(), true);` allocates on every step. The comment says "avoid allocations," but one allocation still happens. If there is an existing `context.getTensor` scratch-buffer pattern in the codebase, prefer that.
- `gate`/`d_gate` and `up`/`d_up` could share memory in frameworks that optimize in-place. If nntrainer guarantees separate buffers, fine — otherwise, add a defensive assert.

### 6. Test coverage
- `unittest_layers_swiglu.cpp` has only one golden test (FP32, NCHW, single shape).
- Please confirm that `LayerGoldenTestParamOptions::DEFAULT` also validates backward — since the backward path is the core contribution of this PR, a flag that explicitly verifies gradients should be enabled.
- There is no FP16 forward test (`swiglufp16` golden). Even a forward-only FP16 test would be valuable given that backward throws.

### 7. Minor style
- The blank line between `exportTo` and `getType` was removed, breaking the comment-block separation.
- Removing `#include <utility>` is fine — it was unused.
- `incoming_deriv` is bound as `const Tensor&`, but `getIncomingDerivative` returns a non-const reference. Implicit const binding works, but please confirm this matches project conventions.

## Risk Summary
- **High**: the `forwarding()` batch-size-1 assumption. Needs validation for the general training path.
- **Medium**: FP16 backward throws at runtime — it is safer to block this at configuration time.
- **Medium**: the duplicate SwiGLU in `Applications/LLaMA/jni` causes naming and maintenance confusion.
- **Low**: documentation/comment cleanup, scratch-buffer reuse.

Overall the direction is good (promotion to core + backward support) and the math is correct. Before merging, I recommend (1) verifying batch>1 training behavior and (2) clarifying the plan for the `Applications/LLaMA/jni` copy.

---

baek2sm (approved):
LGTM

nntrainer/layers/swiglu.h
```diff
@@ -4,7 +4,7 @@
  *
  * @file   swiglu.h
  * @date   14 July 2023
- * @brief  Implementation of SwiGLU activation function
+ * @brief  Implementation of custom SwiGLU activation function
```
jaemini-shin: Does it need to emphasize custom swiglu?

jaemini-shin (approved):
LGTM with minor comment!

nntrainer/layers/swiglu.h
niket-agarwal: No, it can be skipped. Thanks!

Jungwon-Lee (approved):
LGTM!!

jijoongmoon (approved):
LGTM except the issue addressed by enju-yang. It is great to move swiglu into nntrainer as a formal layer.

niket-agarwal:
> LGTM except the issue addressed by enju-yang. It is great to move swiglu into nntrainer as a formal layer.

All the reviews by @EunjuYang have been addressed. Please check below files for swiglu as a formal layer inside nntrainer.
nntrainer/layers/swiglu.cpp
nntrainer/layers/swiglu.h

Thanks for review!

baek2sm:
**Final Verdict: ✅ Approve**

---

Promotes the SwiGLU layer from two duplicated application-local copies (CausalLM and LLaMA) into a single registered core `nntrainer::SwiGLULayer`.

**Included features**
- Core `nntrainer/layers/swiglu.{h,cpp}` with forward and FP32 backward, registered in `AppContext`.
- Removes both app-local implementations and their build / factory-registration wiring.
- Golden-test coverage (FP32 forward + backward, FP16 forward).

---

### 2. Related Issues

- **P3** Fixed `incremental_forwarding` to honor the `from` offset (it previously always wrote rows `0..n`, wrong for `from > 0`); replaced the misleading `util_simd.h` include with `cpu_backend.h`; documented that FP16 backward is not yet implemented.
- **P3** `supportBackwarding()` returns true while FP16 backward throws — documented; completing FP16 backward is a reasonable follow-up.

No P0/P1/P2 issues.

---

Clean merge onto the latest main. The pre-existing OpenCL `LAYER_SWIGLU` registration in `ClContext` is untouched (separate context with its own factory map), and there are no dangling references to the removed app-local headers in CausalLM or LLaMA. The full CI matrix, including the SwiGLU golden test, is green.

---

A clean consolidation into a core layer with a genuine correctness fix (incremental `from`-offset) and a green CI matrix. Approve.

PR #3849, opened 2026-04-08
[Quick.AI] NNTrainer for QNNTransformer (2026-04-08-v2)

EunjuYang:


baek2sm:
While resolving conflicts with main, I've force-pushed and it was my fault. So I then restored the branch to its original state based on the SHA, and added a single merge commit on top.

PR #3853, opened 2026-04-09
[Fix] Fixed quantization to handle large weights

jayden0701:
## Summary

Changed index type from 'int' to 'int64_t' for reading layer weights in quantization.
Before, during quantization, it used int type index for reading weights.
Which made layer weights that contain more than 2^31 number of weights impossible to read.

This PR fixes that error

**Self evaluation:**

Signed-off-by: Joonseok Oh <jrock.oh@samsung.com>


jayden0701:
@codex review

PR #3866, opened 2026-04-13
[Fix] Fixed quant to handle large weights

jayden0701:
## Summary

Changed index type from 'int' to 'int64_t' for reading layer weights in quantization.
Before, during quantization, it used int type index for reading weights.
Which made layer weights that contain more than 2^31 number of weights impossible to read.

This PR fixes that error

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/nntr_ggml_impl_quant.cpp
+++ b/nntrainer/tensor/cpu_backend/ggml_interface/nntr_ggml_impl/nntr_ggml_impl_quant.cpp
@@ -385,13 +385,13 @@ void quantize_row_q4_0_ref(const float *__restrict x, block_q4_0 *__restrict y,
 
   assert(k % qk == 0);
 
-  const int nb = k / qk;
+  const int64_t nb = k / qk;
 
-  for (int i = 0; i < nb; i++) {
+  for (int64_t i = 0; i < nb; i++) {
     float amax = 0.0f; // absolute max
     float max = 0.0f;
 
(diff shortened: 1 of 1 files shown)


Jungwon-Lee (approved):
LGTM

jaemini-shin:
Just for curiosity, does this overflow really happen?

* `Q4_0` is 32, `k >= 2^31 * 32`.
* `float` is 4 bytes, so a single row is greater than `2^31 * 32 * 4` bytes.
  * `274,877,906,944 bytes ~= 274 GB`

jayden0701:
> Just for curiosity, does this overflow really happen?
> 
> * `Q4_0` is 32, `k >= 2^31 * 32`.
> * `float` is 4 bytes, so a single row is greater than `2^31 * 32 * 4` bytes.
>   
>   * `274,877,906,944 bytes ~= 274 GB`

Great question. The overflow doesn't actually happen in `int nb = k / qk`

it happens when
`  for (int i = 0; i < nb; i++) 
{
...
      const float v = x[i * qk + j];
}`

when we multiply `qk` onto `i` which is int, making it overflow.
(since `qk*i <= k` and, `k` is `int64_t` which can be bigger than integer)

So, while it is okay to just set the`i` in the for loop to `int64_t`,

I also changed `int nb` to `int64_t` for consistency,
Since there were other parts in the same code where `nb` was already set to `int64_t`.

baek2sm (approved):
Nice work, LGTM

jaemini-shin (approved):
LGTM!

jaemini-shin (approved):
LGTM!

PR #3869, opened 2026-04-14
Rms norm fallback fp16

b-saianirud:
13/14 testcases passed for rms norm fallback fp16 code
Experiment done:
All computations done in fp32, just the loading storing in fp16 (The input and output is fp16).
Conclusion: The loss of precision happens not in the fp16 computations, but the input is received with the loss of precision, and hence, no computational changes will resolve this issue


Jungwon-Lee:
The new RMSNorm test files are added to the repository, but they are not registered in `test_target`, so they will not be built or run by Meson/CI.

Please add the new tests to `test/unittest/meson.build` under the correct feature/architecture conditions. Otherwise this PR adds test files without actual CI coverage.

Jungwon-Lee:
`unittest_rms_norm_fp16.cpp` and `unittest_rms_norm_fp16_half.cpp` appear to be identical.
Please remove the duplicate.

baek2sm:
**Review verdict: ✅ Approve** (all CI checks green)

### Summary
Fills in the ARM NEON fp16 intrinsic implementation of `__fallback_rms_norm_wrt_width_fp16_intrinsic<_FP16>` for the fallback compute path, with unit tests.

### Fixes applied on top to make it portable and CI-clean
- Wrapped the NEON intrinsic implementation in `#if defined(__ARM_NEON)` with a non-NEON `NYI` throw — previously the bare NEON types would fail to compile on the x86_64 `enable-fp16` CI leg.
- Changed the `hsumv_f16`/`hsumvq_f16` reduction helpers to return `float` (avoiding an fp16-narrowing precision loss in the square-sum).
- Registered the new tests in `meson.build` (ARM/aarch64 + fp16 guarded), made them exercise the actual `_FP16` fallback path, and removed a byte-identical duplicate test file.
- clang-format-14 on changed lines; sign-off added for DCO.

### Stability for integration with latest `main`
Additive, cleanly mergeable. The non-NEON path is a runtime `NYI` matching the project's default-throw convention; the NEON path runs on ARM/Android CI. Windows MSVC build verified locally (478/478, 22/22). All required checks are green.

PR #3880, opened 2026-04-16
[CausalLM] Add `TurboQuant` feature (naive version)

EunjuYang:

<details><summary>[Feat] add naive-version TurboQuant & unittest</summary><br />

- This commit adds codes for TurboQuant.
- This commit only implements fallback version which is not optimized
  version.
- This commit adds a unittest to test turboquant-related kernels

**Self evaluation:**

</details>

<details><summary>[TurboQuant] apply turboquant to mha / model </summary><br />

- This commit applies turboquant kernels to mha_core
- This commit creates use_turboquant property and add it to Transformer
- This commit add turboquant property to Qwen3
- This commit add a unittest for mha_turboquant integration

**Self evaluation:**

</details>

<details><summary>[res] add use_turboquant field</summary><br />

- This commit add use_turboquant field to qwen3/qwen3-4b
  nntr_config.json
- If one wants to use turboquant, one can update its value from `false`
  to `true`.

**Self evaluation:**

</details>

### Summary

- This PR adds a TurboQuant to CausalLM.
- This PR contains kernels which are not optimized.
- This PR adds TurboQuant to Qwen3

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -325,6 +325,17 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   unsigned int batch_size = (_from) ? 1 : query_dim.batch();
 
   if (use_turboquant) {
+    // The naive TurboQuant kernels operate on FP32 data only. Reject any
+    // non-FP32 tensors up-front instead of reinterpreting their bytes via
+    // getData<float>() (which would corrupt KV cache for FP16 activations).
+    if (query.getDataType() != ml::train::TensorDim::DataType::FP32 ||
+        key.getDataType() != ml::train::TensorDim::DataType::FP32 ||
+        value.getDataType() != ml::train::TensorDim::DataType::FP32 ||
+        output.getDataType() != ml::train::TensorDim::DataType::FP32) {
+      throw std::invalid_argument(
+        "TurboQuant path currently supports FP32 tensors only");
+    }
+
     // TurboQuant 4-bit packed KV cache path
     nntrainer::Tensor &cache_key_scales =
       context.getTensor(tensor_idx[AttentionParams::cache_key_scales]);
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -28,6 +28,7 @@ static std::mutex rope_init_mtx;
 #include <node_exporter.h>
 
 #include <cstdint>
+#include <turboquant_utils.h>
 
 inline float convert_scalar(uint16_t h) {
   return nntrainer::compute_fp16_to_fp32(h);
@@ -52,7 +53,7 @@ MHACoreLayer::MHACoreLayer() :
     props::SlidingWindow(), props::MaxNewTokens(), props::RopeTheta(),
     props::MaxPositionEmbeddings(), props::UseSink(), props::RopeScalingType(),
     props::RopeScalingFactor(), props::RopeScalingMaxPositionEmbeddings(),
-    props::AttnLogitSoftcapping(), props::IsCausal()),
+    props::AttnLogitSoftcapping(), props::IsCausal(), props::UseTurboQuant()),
   sm(nntrainer::ActivationType::ACT_SOFTMAX),
   epsilon(1e-3),
   cache_index(0),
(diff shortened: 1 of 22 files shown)


Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -633,6 +770,189 @@ void MHACoreLayer::one_batch_incremental_forwarding(
...
+                                     : (seq_len * cache_to),
+                           num_heads_Q, query_step.getTensorType());
+
+    const uint8_t *kc_packed =
+      cache_key.getData<uint8_t>() + batch * cache_key_dim.getFeatureLen();
+    const float *kc_norms = cache_key_scales.getData<float>() +
+                            batch * cache_key_scales.getDim().getFeatureLen();
+    const float *signs = tq_rot_signs.data();
+
+    unsigned int seq =
```
DonghakPark, line 1110: This truncates the prefill loop to `local_window_size`, but `attention_output_step` still has `seq_len` rows and `softmax_triangle()` is called with `seq_len`. When `seq_len > local_window_size`, only the first `seq` query rows get `QK` scores, while the value-cache phase below reads rows based on `cache_to - seq + i` and writes only the first `seq` output rows. That leaves part of the prefill output uncomputed and can also read uninitialized attention scores for sliding-window models. The local window should limit the attended key rows inside the kernel, not the number of query tokens processed; please iterate all `seq_len` queries or adjust the output indexing/tests accordingly.

DonghakPark (dismissed):
Thanks for the TurboQuant baseline work. I found one correctness issue that should be fixed before merge. It affects prefill when TurboQuant is used with a sliding/local window shorter than the input sequence.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -149,29 +150,75 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
...
+  use_turboquant = std::get<props::UseTurboQuant>(mha_core_props).get();
+
+  if (use_turboquant) {
+    /**
+     * TurboQuant v2: norm + rotation + Lloyd-Max codebook
+     * - Packed KV cache: UINT8, width = num_heads_KV * head_dim / 2
+     * - Per-head L2 norms: FP32, width = num_heads_KV
+     * - Rotation signs: generated once at finalize
+     */
+    unsigned int packed_width = num_heads_KV * head_dim / 2;
```
jaemini-shin: How about adding a validation if num_heads_KV is even? Although most of the models use even number, but there is no guarantee for that.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -217,7 +264,6 @@ void MHACoreLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
       throw std::invalid_argument(
         "to shouldn't greater than max_timestep for initial forwarding");
     } else {
-      throw std::runtime_error("NYI: cache shift is not available");
```
jaemini-shin, line 424: It looks like some changes in commit bd8134b is reverted in this PR. Could you please check it?

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -633,6 +770,189 @@ void MHACoreLayer::one_batch_incremental_forwarding(
...
+  const unsigned int to, nntrainer::Tensor &query_step,
+  nntrainer::Tensor &key_step, nntrainer::Tensor &value_step,
+  nntrainer::Tensor &attention_output_step, nntrainer::Tensor &cache_key,
+  nntrainer::Tensor &cache_value, nntrainer::Tensor &cache_key_scales,
+  nntrainer::Tensor &cache_value_scales, ml::train::TensorDim &cache_key_dim,
+  ml::train::TensorDim &cache_key_step_dim,
+  ml::train::TensorDim &cache_value_dim,
+  ml::train::TensorDim &cache_value_step_dim) {
+
+  auto &pool =
```
jaemini-shin, line 999: This code would be in conflict with #3893. We should decide which one to merge first.

jaemini-shin (commented):
Could you please check these comments?

baek2sm:
**Review verdict: 💬 Comment** (rebased and locally green on x86, but the ARM/Android build gates are red — not approving)

### 1. Summary
Adds the naive-version `TurboQuant` feature: 4-bit quantized KV-cache kernels (`turboquant_utils.h`), a `use_turboquant` property on `MHACoreLayer`, a `UseTurboQuant` Qwen3 config field, and two unit-test files. The branch was rebased onto the latest `upstream/main` (all commits signed-off, original authorship preserved); the `use_turboquant` path was reconciled to nest under upstream's `!use_external_cache` external-KV-cache model.

### 2. CI status / blocking issue
The x86 build is healthy locally — **Windows MSVC compiled 504/504 and all 27 unit tests pass**, including the 9 `turboquant` kernel tests and the 11 `turboquant_mha_integration` accuracy tests (cosine > 0.99 vs FP32). However, **`Build CausalLM for Android (arm64-v8a)` and `Cross Build (x64 -> ARM64)` fail** — the `MHACoreLayer` TurboQuant integration path is `enable-transformer`-gated and compiles only on those targets, which we cannot build on Windows.

### 3. What needs consideration for verification (and our limitation)
The default Windows native build has `enable-transformer=false`, so the **`mha_core.cpp` TurboQuant path was not locally compiled** — only the standalone kernels (in the core library) were. If you verify, focus on:
- The Android/NDK and ARM cross-compile error in the `mha_core.cpp` `use_turboquant` branch (likely an ARM/NDK-specific type or API mismatch surfaced only when `enable-transformer` is on) — get the exact `ndk-build` error.
- The naive fallback kernels delegate identically on x86 and ARM (no NEON intrinsics added), so kernel numerics should match; the failure is most likely a **compile** issue in the gated integration code, not a math issue.
- Minor: a comment typo `caculcate` in `turboquant_utils.h` may trip the spelling static check.

The kernel math is well-tested on x86; the blocker is the ARM/Android compile of the gated MHA integration, which needs a target-toolchain build to resolve.

PR #3883, opened 2026-04-17
[Dependon #3880] [TurboQuant] add x86-avx2 kernel 

EunjuYang:
## Dependency of the PR
#3880 

<details><summary>[TurboQuant] add x86-avx2 kernel </summary><br />

- This commit add avx2_turboquant.{h/cpp}
- This commit add this avx2 kernel to x86_compute_backend.cpp

**Self evaluation:**

</details>

### Summary

- This commit add avx2_turboquant.{h/cpp}
- This commit add this avx2 kernel to x86_compute_backend.cpp

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
@@ -14,6 +14,7 @@
 #include <assert.h>
 
 #include <avx2_impl.h>
+#include <avx2_turboquant.h>
 #ifdef USE_BLAS
 #include <cblas_interface.h>
 #endif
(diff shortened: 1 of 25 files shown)


nntrainer/tensor/cpu_backend/x86/x86_compute_backend.cpp
```diff
@@ -503,4 +508,45 @@ void transform_int4_osv32_isv2_to_q4_0(size_t N, size_t K,
     N, K, osv32_weights, osv32_scales, scale_group_size, 8, dst_q4_0x);
 #endif
 }
+
+void quantize_kv_turboquant(const float *input, uint8_t *out_packed,
+                            float *out_norms, const float *rot_signs,
+                            int head_dim, int num_heads) {
+  if (head_dim == 64 || head_dim == 128) {
```
DonghakPark, line 515: For x86 builds these public backend entry points now return without doing anything when `head_dim` is not 64 or 128. That bypasses both the scalar fallback and the `get_codebook()` validation path, leaving `out_packed`, `out_norms`, or `output` unchanged/uninitialized while the caller continues as if the kernel succeeded. The AVX2 implementation already has defensive fallback branches, but this dispatcher prevents them from being reached. Please call the fallback implementation (or throw explicitly) in the non-64/128 path for all three TurboQuant entry points.

DonghakPark (dismissed):
I found one x86-specific correctness issue in the AVX2 dispatch layer. Please fix this before merge; otherwise unsupported head dimensions can silently produce invalid output on x86 builds.

PR #3893, opened 2026-04-23
Introduce new thread manager

jaemini-shin:
This PR introduces a new thread manager inspired by https://github.com/Maratyszcza/pthreadpool.
It replaces following components:
* bs threadpool
* openmp

Note that it hasn't replaced openblas yet.

### Features
* fix affinity
  * pinning thread to specific physical core
* work stealing
  * fast thread can steal job of slow one
* spin wait

### How to use
There are two ways to configure number of threads.
* (high priority) use environment variable `NNTR_NUM_THREADS`
* add meson compile flag `-Dnntr-num-threads`

Here comes the example code: just wrap the loop body
```cpp
// single thread
for(int i=0; i<1000; i++)
  a[i] += 1;

// thread manager
auto &tm = nntrainer::ThreadManager::Global();
tm.parallel_for(0, 1000, [=](size_t i) {
  a[i] += 1;
});
```

For work stealing, it is not desirable to divide whole loop with number of total threads.
Try non-tiled loop and then increase the tile size.

### Performance(qwen3-4b)
* Before this PR(main, omp thread 4)
```
=================[ LLM with NNTrainer ]===================
prefill: 512 tokens, 7540 ms, 67.9045 TPS
generation: 128 tokens, 6949 ms, 18.4199 TPS
total: 14490 ms
peak memory: 2634448 KB
==========================================================
[e2e time]: 15118 ms 
Max Resident Set Size: 2634448 KB
```
* After this PR(S25U, thread num 4)
```
=================[ LLM with NNTrainer ]===================
prefill: 512 tokens, 5356 ms, 95.5937 TPS
generation: 128 tokens, 6435 ms, 19.8912 TPS
total: 11793 ms
peak memory: 2686824 KB
==========================================================
[e2e time]: 12601 ms 
Max Resident Set Size: 2686824 KB
```
* After this PR(S25U, thread num 7)
```
=================[ LLM with NNTrainer ]===================
prefill: 512 tokens, 4199 ms, 121.934 TPS
generation: 128 tokens, 5906 ms, 21.6729 TPS
total: 10107 ms
peak memory: 2689288 KB
==========================================================
[e2e time]: 10910 ms 
Max Resident Set Size: 2689288 KB
```

### Notable Changes
* remove bs threadpool
* remove openmp
  * there might be some left openmp configurations, not affecting on operation
* optimize some kernels

---

## Commits to be reviewed
Sorry for big PR, but I'll explain in short for easy review.
### Introduce unified Thread Manager
* New thread manager is introduced with some unittests
* Not used in main code yet
### Replace bs thread pool into new thread pool
* remove bs threadpool
* use new thread manager instead
* API is almost same as before
### Replace openmp into new thread pool
* remove openmp
* use new thread manager instead
* API is a bit different, but scheme is same
* most of the changes are removing linking openmp library
### Optimize ggml kernel
* optimize some kernels to fit with new thread manager
* for better performance, each thread should hold several loops
### Refactor ThreadManager
* remove redundant APIs
* reduce call stack in thread manager

---

## Review analysis (Claude)

Full review was also left as a PR review comment. This section gives a per-commit breakdown plus an overall findings list.

#### 1. `2cf61ec` Introduce unified Thread Manager  (+1115 / −4, 10 files)

New files only; no call-site touched yet.
* `nntrainer/utils/thread_manager.h` (+354), `thread_manager.cpp` (+268)
* `nntrainer/utils/thread_manager_util.h` (+60), `thread_manager_util.cpp` (+208)
* `test/unittest/unittest_thread_manager.cpp` (+198) — this is the original, larger test file
* packaging / Android.mk / meson glue

Design: flat pool with main-thread participation, `futex` on Linux/Android + mutex/cv on Windows, `SPIN_COUNT=1,000,000`-step spin before blocking, work-stealing via `modulo_decrement` + `range_length` CAS + `range_end.fetch_sub`. Per-thread `thread_info` is `CACHELINE_ALIGNED` to prevent false sharing between owner (reads `range_start`, dec `range_length`) and stealers (dec `range_end`). Affinity enumerates physical cores via `/sys/devices/system/cpu/*/topology/thread_siblings_list` + `cpuinfo_max_freq`, handles Intel hybrid (`cpu_core` / `cpu_atom`).

Issues introduced here (all still present at tip):
* **Uninitialized `command_`**. `std::atomic<uint32_t> command_` has no default value. Pre-C++20, default-constructed atomics are not value-initialized. Workers launch and compare to `INIT(=0)`; if garbage ≠ INIT, they take the `RUN` branch and call `task_()` which is an empty `std::function` → crash. Set `command_.store(INIT, relaxed)` before spawning workers.
* **Nested `parallel_for` deadlocks.** `parallelize()` holds `execution_mutex_` for the whole dispatch and also participates as tid=0. A nested `parallel_for` inside the user lambda will block on the mutex; the outer call is stuck in `wait_worker_threads()` because that worker never reaches `checkin()`. Needs documentation, re-entrance detection, or a thread-local guard.
* **`is_x86` has no `#else`** in `thread_manager_util.cpp` → riscv64/ppc builds fail to compile.
* **`NNTR_NUM_THREADS=0` / malformed** passes `std::stoul` and underflows `num_workers = compute_threads - 1` to UINT32_MAX, then `reserve(huge)` throws. Also `std::stoul` can throw on bad input and there's no try/catch.
* **`core_map[0]` unchecked** when affinity is on and `getCoresByPerformance()` returns empty.
* **Malformed SPDX identifier**: `Apache-2.0, BSD 2-Clause "Simplified" License` → should be `Apache-2.0 AND BSD-2-Clause` for REUSE/FOSSology compliance.
* **`setConfig` semantics**: `config_` is read exclusively inside lazy `initialize()`. Any `setConfig` after first `Global()` is silently ignored. Either document "must be called before first use", make it throw when late, or plumb config through `Engine::initialize()`.

Tests at this point are reasonable baseline (empty range, single element, offset, accumulation, multi-call, thread count), but miss the new properties: work-stealing under imbalance, affinity, re-entrance, concurrent callers, env-vs-compile-flag precedence. They also regress in commit 5 (see below).

#### 2. `5bc4155` Replace bs thread pool into new thread pool  (+550 / −3749, 14 files)

Net deletion of ~3200 LOC. The star removals:
* `nntrainer/utils/bs_thread_pool.h` (−2850)
* `nntrainer/utils/bs_thread_pool_manager.hpp` (−56), `.cpp` (−50)
* `Engine::getThreadPoolManager()` + member + mutex removed from `engine.{h,cpp}`

Mechanical migration pattern: `submit_loop(0, n, [=](int i){ ... }); multi_future.wait();` → `tm.parallel_for(0, n, [=](size_t i){ ... });`. Touches the three `ggml_interface_{bs_threadpool,fp16,mixed}.cpp` variants, `Applications/CausalLM/layers/mha_core.{cpp,h}`, `qkv_layer.cpp`.

Issues:
* **Orphan Doxygen block in `engine.h`.** After removing `getThreadPoolManager()`, a `/** * */` header pair is left dangling above `parseComputeEngine`:
  ```cpp
    /**
     *
    /**
     *
     * @brief Parse compute Engine keywords ...
  ```
  Cosmetic but noisy.
* **Suboptimal loop shape retained.** Many sites still split `[0, thread_num)` exactly (1 tile per worker) — work stealing can't help with range_length==1. Strong perf win here is just "removed pool overhead + pinning", not actual stealing. The follow-up commit 4 does fix this for one file; the other two variants stay flat.
* **`[=]` captures of heavy vectors.** `__ggml_q4_0_4x8_q8_0_GEMM_GEMM` captures `QA` (`std::vector<char>`, potentially MB) by value into the `std::function<void(size_t)>` stored in `task_`. `parallelize()` blocks until workers finish, so `[&]` is safe and avoids the copy. Same pattern elsewhere.
* **Re-entrance from the `pb` remainder loop.** `for (pb = M4*4; pb < M; pb++) { tm.parallel_for(...); }` re-enters the pool 1–3 times per GEMM. Each re-entry re-pays dispatch cost. Consider hoisting the `pb` loop inside a single `parallel_for` dispatched over `pb * chunk` iterations.
* **MHA migration API simplification.** `compute_kcaches(..., BS::thread_pool<> &pool)` now takes no pool parameter — cleaner. In the prefill branch the old code used `pool.submit_task` which allowed heterogeneous per-task futures. Replacing with `parallel_for(0, seq, ...)` is static-work-distribution with stealing — equivalent or better for this pattern, but worth confirming no call-site actually needed submit_task's flexibility.

#### 3. `f8c4722` Replace openmp into new thread pool  (+982 / −1816, ~70 files)

Biggest breadth. Removes the OpenMP dependency end-to-end:
* meson: drops `enable-openmp`, `omp-num-threads`, `libiomp_root`, and the entire OpenMP detection block (−42 in `meson.build`).
* Windows: removes `libiomp_win/{arm64,x64}/*` binary artifacts and `include/omp.h` (−741 LOC header).
* Android: updates 22+ `Applications/**/jni/Android.mk` files (−LDFLAGS `-fopenmp`/ `-lomp` / `libomp`).
* Kernels: `#pragma omp parallel for` → `tm.parallel_for(...)` across `neon_impl.cpp`, `neon_impl_fp16.cpp`, `avx2_impl.cpp`, `kleidiai_interface_*.cpp`, `ggml_interface_omp.cpp`, `float_tensor.cpp`, `tensor.cpp`, `mha_core.cpp` (more), 7 CausalLM MoE layers.
* Removes `neon_setting.h` (−62).

Issues:
* **Stale `thread-backend` meson option** still sits in `meson_options.txt` with `value: 'omp'`. With OpenMP gone, this selects between the three `ggml_interface_*.cpp` variants that are now ~90% identical. Either drop the option or consolidate the files.
* **`topK` loses `collapse(3)`**. `FloatTensor::topK` converts `#pragma omp parallel for collapse(3)` to a single `parallel_for(0, batch*channel*height, ...)` with manual index decomposition. Semantically equivalent; loses the compiler's loop nest rescheduling but pick up work-stealing in return. Correctness intact.
* **`getBatchSlice` is now work-stealing** instead of `schedule(static)`. Likely a small improvement due to uneven memcpy cost.
* **`configurations/windows-cross-arm-ci.ini` touched** — good, keeps Windows cross-build in sync; worth verifying the CI actually goes green without libiomp.
* **`avx2_impl.cpp` (+80/−83)** — diff is substantial; worth a careful look to ensure no reduction loop silently became sequential and no false-sharing was introduced by the grain change.
* **`neon_setting.h` deletion** — confirm no downstream file relied on its constants (grep showed no survivors; double-check Android builds).

#### 4. `cbe03c1` Optimize ggml kernel  (+61 / −49, 1 file)

Single-file touch on `ggml_interface_omp.cpp`. The squashed commit message shows four original commits ("Optimize ggml kernel" × 3 + "Fix allocation bug"). Two substantive improvements:

1. **Chunked 2D dispatch** for GEMM/GEMV:
   ```cpp
   unsigned int row_chunk_size = 16;
   unsigned int col_chunk_size = 16;
   unsigned int row_loop = (M + row_chunk_size - 1) / row_chunk_size;
   unsigned int col_loop = (N + col_chunk_size - 1) / col_chunk_size;
   tm.parallel_for(0, col_loop * row_loop, [=](size_t i) {
     unsigned int r = i / col_loop, c = i % col_loop;
     ...
   });
   ```
   This is the loop shape that actually lets work-stealing pay off. The `M % 4 != 0` path was also rewritten to split the M4 block from the 1–3 remainder rows, with the GEMM tiled and remainder GEMV tiled.

2. **Allocation fix** for the `M % 4 != 0` path:
   ```diff
   - unsigned int qa_size = qa_4_rows_size * (((M >> 2) << 2) / 4 + 1);
   + unsigned int qa_size = qa_4_rows_size * M4 + qa_row_size * (M % 4);
   ```
   The old formula over-allocated in the blocked layout only; the new one correctly sizes the M4 blocks plus per-row remainder. This is a real bug fix.

Issues:
* **Squashed commit is hard to review** — four logical commits with the same title collapsed into one. Regression potential for an allocation fix bundled with perf work; please split in future.
* **Asymmetry with siblings.** Only `ggml_interface_omp.cpp` gets the chunked scheme. `ggml_interface_bs_threadpool.cpp` and `ggml_interface_mixed.cpp` still dispatch flat; they remain vulnerable to the same allocation bug if the `M % 4` path is live there too — please audit and mirror the fix, or consolidate the three variants.
* **No test for `M % 4 != 0`**. The fixed allocation path isn't exercised by any `unittest_nntrainer_tensor.cpp` or ggml-related test I saw. Add `M=7,9,11` cases.
* **Magic `chunk_size = 16`** — a one-line comment justifying the choice (cache line / SIMD tile) would help the next maintainer pick a sane value for a new quant format.

#### 5. `80df0c8` Refactor ThreadManager  (+260 / −403, 10 files)

Tidy-up pass. thread_manager.h: `+9/−40` removing redundant APIs. thread_manager.cpp: `+2/−2`. Callers updated: `neon_impl_fp16.cpp` (+110/−121), `kleidiai_interface_qsi8d32p_qsi4c32p_omp.cpp` (+63/−77), `kleidiai_interface_qai8dxp_qsi4cxp.cpp` (+30/−30), both ggml_interface_bs_threadpool and _mixed (+18/−18 each), ggml_interface_omp (+8/−8), ggml_interface_fp16 (+2/−2).

Issues:
* **Test-coverage regression**: `test/unittest/unittest_thread_manager.cpp` changed by `−87, +0`. The original file in commit 1 had ~198 lines; tip has 111. 87 lines of tests were **deleted without replacement** as part of a "refactor" commit. Given that commit 1's tests already missed work stealing / affinity / re-entrance / concurrency, removing further tests weakens coverage right as the API stabilizes. Please restore the deleted tests or justify why each was redundant.
* **Missed opportunity**: this was the natural place to fix the correctness items introduced in commit 1 (uninitialized `command_`, `is_x86` `#else`, SPDX, env-var validation). All still open at tip.
* **Kleidiai paths rewritten (~ 93+ changed lines).** `kleidiai_interface_qsi8d32p_qsi4c32p_omp.cpp` was a significant OpenMP accelerated path; please confirm micro-benchmarks on aarch64 S25U haven't regressed specifically for this backend (the PR's e2e numbers don't isolate it).

### Overall findings (consolidated)

Correctness / portability (recommended before merge):
1. `command_` atomic is not initialized before workers are spawned — pre-C++20 UB. *(introduced in 1, unfixed in 5)*
2. Nested / re-entrant `parallel_for` calls deadlock. *(introduced in 1)*
3. Orphan Doxygen block in `engine.h`. *(introduced in 2)*
4. `is_x86` constexpr has no `#else`. *(introduced in 1)*
5. `NNTR_NUM_THREADS=0` / malformed value causes underflow / uncaught throw. *(introduced in 1)*
6. `core_map[0]` unchecked. *(introduced in 1)*
7. **`unittest_thread_manager.cpp` lost 87 lines of tests in commit 5.** *(new finding)*

Design / maintenance:
8. `setConfig` ignored after lazy init. *(1)*
9. Stale `thread-backend` option + three near-identical `ggml_interface_*.cpp`. *(3)*
10. Default `compute_threads = hardware_concurrency()/2` undercounts non-SMT; use `getPhysicalCoreCount()`. *(1)*
11. Malformed SPDX identifier. *(1)*
12. Chunked GEMM/GEMV (`cbe03c1`) only applied to the `_omp` variant; same allocation fix may be needed in the two sibling files. *(4)*

Perf / minor:
13. Heavy `[=]` captures in `std::function` task; `[&]` is safe. *(2, 3)*
14. Many call-sites still dispatch `[0, thread_num)` flat — work stealing can't help until over-decomposed. *(2)*
15. MHA `pb` remainder loop re-enters the pool per iteration. *(2)*
16. Magic `chunk_size = 16` without rationale comment. *(4)*
17. Windows has no affinity, no SMT/hybrid detection. *(1)*
18. Future-dated `@file` headers (2026). *(1)*
19. Main thread's affinity never released (`Engine::release()` is empty). *(2)*

### Recommendation

Address items **1, 3, 4, 5, 7** (correctness + test regression) before merge. Items **2, 8, 9** are open design decisions that should get an answer now because they constrain how future call-sites should be written. Others are cleanup. Net-positive PR: ~3k LOC down, two external threading deps removed, real perf gains.


Applications/CausalLM/layers/embedding_layer.cpp
```diff
@@ -114,8 +115,8 @@ void EmbeddingLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
 
     int iter = to - from;
 
-#pragma omp parallel for
-    for (int i = 0; i < iter; ++i) {
+    auto &tm = nntrainer::ThreadManager::Global();
+    tm.parallel_for(0, static_cast<size_t>(iter), [&](size_t i) {
```
jaemini-shin, line 119: An example of using new thread manager API.
Just wrap with `tm.parallel_for`. You need start loop index, end loop index, and callback according to loop body.
Note that it loops the range of `[start, end)`.

nntrainer/tensor/cpu_backend/ggml_interface/ggml_interface_omp.cpp
```diff
@@ -30,27 +30,24 @@ void __ggml_q4_0_4x8_q8_0_GEMM(const unsigned int M, const unsigned int N,
...
-      M_step_start = (M_step_start % NB_COLS)
-                       ? M_step_start + NB_COLS - (M_step_start % NB_COLS)
-                       : M_step_start;
-      M_step_end = (M_step_end % NB_COLS)
-                     ? M_step_end + NB_COLS - (M_step_end % NB_COLS)
-                     : M_step_end;
+    // compute multithreaded GEMV
+    tm.parallel_for(0, loop, [=](size_t idx) {
+      unsigned int M_step_start = chunk_size * idx;
+      unsigned int M_step_end = std::min(chunk_size * (idx + 1), (size_t)N);
```
jaemini-shin, line 50: This is the good example of using new thread manager.
Using a fixed tile(chunk) is much better than doing loop for the number of threads.

.github/workflows/ubuntu_clean_meson_build_clang.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: install additional packages for features
       run: sudo apt-get install -y python3-dev python3-numpy python3
     - name: Install llvm
-      run: sudo apt install -y llvm libomp-dev
```
jijoongmoon, line 27: So, from now on, we do not use the openmp any more, then.

Applications/CausalLM/layers/embedding_layer.cpp
jijoongmoon: It might be better to provide the user guide for thread manager with markdown format

Seunghui98 (commented):
## Code Review — #3893 "Introduce new thread manager"

Nice consolidation: bs_thread_pool (~3k LOC header-only dep) + OpenMP are replaced by one in-tree primitive, with a real perf win on S25U (prefill 67.9 → 121.9 TPS at 7 threads). Design is clean (futex on Linux/Android, mutex+cv on Windows, spin threshold, work-stealing, big.LITTLE-aware pinning). Below are the issues I found while reading `thread_manager.{h,cpp}`, `thread_manager_util.*`, and representative call-sites.

### Blocking / correctness

1. **`command_` atomic is never initialized before workers read it.**
   `nntrainer/utils/thread_manager.h`:
   ```cpp
   CACHELINE_ALIGNED std::atomic<uint32_t> command_;
   ```
   `ThreadManager::ThreadManager() {}` is empty, and `initialize()` doesn't assign `command_` before `compute_workers_.emplace_back(...)`. Pre-C++20, default-constructed `std::atomic<T>` is not value-initialized — the storage is indeterminate. Each worker starts in `thread_main` with `last_command = INIT (=0)` and calls `wait_for_new_command(INIT)`. If the garbage happens to equal `INIT`, everything works (the common case, which is why it's passing tests); otherwise `command & COMMAND_MASK` can be `RUN` → `thread_parallelize()` runs against an unset `task_` (empty `std::function`) → crash. Please `store(INIT, relaxed)` before spawning workers.

2. **Nested / re-entrant `parallel_for` deadlocks.**
   `parallelize()` holds `std::lock_guard<std::mutex>(execution_mutex_)` for the whole dispatch and the main thread participates via `thread_parallelize(0)`. If the user callback (or a helper it calls) invokes `parallel_for` again, the inner call waits on `execution_mutex_`; meanwhile the outer call is blocked in `wait_worker_threads()` because the thread running the nested call never reaches `checkin()`. This is easy to trip in layer code that calls into helpers which themselves parallelize. Minimum: document the restriction; better: detect re-entrance and fall back to sequential.

3. **Dangling Doxygen comment in `nntrainer/engine.h`.**
   After removing `getThreadPoolManager()`, the `/** * */` block that used to precede it is now orphaned right above `parseComputeEngine`:
   ```cpp
     /**
      *
     /**
      *
      * @brief Parse compute Engine keywords ...
   ```
   Two `/**` in a row — looks like a merge artifact.

4. **`thread_manager_util.cpp::is_x86` has no fallback branch.**
   ```cpp
   constexpr static bool is_x86 =
   #if defined(__x86_64__) || ...
     true;
   #elif defined(__aarch64__) || ...
     false;
   #endif
   ```
   A non-x86/non-ARM host (riscv64, ppc, etc.) leaves `is_x86` without an initializer and fails to compile. The repo has riscv64 references elsewhere. Add an `#else` default.

5. **`NNTR_NUM_THREADS=0` / negative env value is unchecked.**
   `std::stoul` parses `"0"` happily → `compute_threads=0` → `num_workers = compute_threads - 1 = UINT32_MAX` (underflow) → `reserve` blows up. Also, `std::stoul` throws on non-numeric input, which will abort a process whose user typo'd the env var. Validate (`>= 1`) and wrap in try/catch with a warning.

6. **`core_map[0]` access after empty return.**
   If `getCoresByPerformance()` ever returns an empty vector on the x86/hybrid path (e.g., sysfs partially unreadable), the subsequent `pinSelfToCore(core_map[0])` is UB. A size check before indexing is cheap insurance.

### Design / maintenance

7. **`setConfig` is effectively ignored after first `Global()`.**
   `config_` is read exclusively inside `initialize()`, and the `Singleton` initializes lazily on first access. Any `setConfig` call after *any* caller has touched the singleton (very likely via some static initialization elsewhere) is silently dropped. Either document this as "must be called before first use", make it throw when late, or take the config as a ctor argument / init routine invoked from `Engine::initialize`.

8. **Stale `thread-backend` meson option + three nearly-identical ggml interfaces.**
   `meson_options.txt` still has `option('thread-backend', ..., value: 'omp')`, but `ggml_interface_omp.cpp`, `ggml_interface_bs_threadpool.cpp`, and `ggml_interface_mixed.cpp` now all go through `ThreadManager::parallel_for`. The three files are ~90% identical post-change. Strong candidate for consolidation into a single `ggml_interface.cpp` — otherwise every future kernel tweak needs three copies.

9. **Default `compute_threads` uses logical cores, but pinning uses physical.**
   ```cpp
   uint32_t hw = std::thread::hardware_concurrency();
   return hw > 0 ? hw / 2 : 1;
   ```
   On a non-SMT 4-core ARM laptop this returns 2 (half the cores) even though `getPhysicalCoreCount()` would return 4. Use `getPhysicalCoreCount()` for parity with the actual pinning logic.

10. **SPDX identifier is malformed.**
    `thread_manager.h`:
    ```
    // SPDX-License-Identifier: Apache-2.0, BSD 2-Clause "Simplified" License
    ```
    This isn't valid SPDX syntax; tooling (REUSE, FOSSology, etc.) won't parse it. Should be `Apache-2.0 AND BSD-2-Clause`.

11. **`@file` dates are in the future (20 March 2026 / 20 April 2026)** across the new files. Cosmetic but worth checking whether the project prefers authoring date or creation date.

### Perf / minor

12. **Heavy captures in `[=]` lambdas.**
    E.g. `ggml_interface_bs_threadpool.cpp::__ggml_q4_0_4x8_q8_0_GEMM_GEMM` captures `QA` (`std::vector<char>`, potentially MB) by value into the `std::function` stored in `task_`. Since `parallelize()` blocks until workers finish, `[&]` is safe and avoids the copy. Not a regression vs. the old bs_threadpool version, but an opportunity. The same applies in multiple `ggml_interface_*.cpp` sites.

13. **Work-stealing helps only when loops are over-decomposed.**
    The PR description says it, but many call-sites still split `[0, thread_num)` exactly (e.g., `ggml_interface_omp.cpp::__ggml_q4_K_8x8_q8_K_GEMM`). With range_length == 1 per worker, stealing can't help and the main contribution is removed OpenMP overhead + pinning. Where the PR does chunk by `chunk_size=16` (GEMV path), stealing can actually pay off. Maybe unify on chunked dispatch across the GEMM/GEMV paths.

14. **Test coverage misses the new properties.**
    `unittest_thread_manager.cpp` covers basic semantics well. But none of the features advertised in the PR are exercised:
    - Work stealing with mixed fast/slow iterations (e.g., sleep in odd indices).
    - Affinity (skip if sysfs unavailable, but at least check `pinSelfToCore` happy path).
    - Re-entrant / nested dispatch (related to issue 2 — would currently hang).
    - Concurrent `parallel_for` calls from different caller threads (`execution_mutex_` should serialize them cleanly).
    - Thread-count selection via env vs. compile flag vs. default.

15. **`unittest_thread_manager.cpp` header points to `nnstreamer/nntrainer`** instead of `nntrainer/nntrainer`.

16. **Windows path has no pinning and `getPhysicalCoreCount()` falls back to `hardware_concurrency()`.**
    This is acknowledged with `@todo`, but on hybrid x86 Windows (Alder Lake+) the default worker count will include E-core siblings and typically under-perform. Worth calling out in the docs.

### Nits

17. `thread_manager.cpp::thread_main`'s switch has an `INIT:` case that does nothing — intentional, but a `// NOP` comment would clarify.
18. `parseCpuList` mixes `int i` with `uint32_t` bounds; `std::stoi` also throws on `"0-"` / empty tokens without context-rich errors. Consider `std::from_chars` or at least wrapping exceptions.
19. `main thread's affinity is never restored.` `Engine::release()` is now empty; the main thread retains its pin until process exit. Harmless but asymmetric.
20. `thread_infos_` is `unique_ptr<thread_info[]>` — fine since `alignof(thread_info) == 64` with `CACHELINE_ALIGNED`. Worth an inline comment explaining the invariant.

### Recommendation

Approve once issues **1** (uninitialized `command_`), **3** (broken comment in `engine.h`), **4** (portability `#else`), and **5** (env-var validation) are addressed — those are correctness/portability. Issues 2, 7, 8 are followup candidates but worth a decision now because they constrain how call-sites should be written. The rest are cleanup / documentation.

Great work on the perf numbers and removing two external thread primitives in one pass.

---

nntrainer/utils/thread_manager.h
```diff
@@ -0,0 +1,334 @@
...
+ * @date   23 April 2026
+ * @brief  Unified thread manager for compute, inspired by pthreadpool
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author Jaemin Shin <jaemin2.shin@samsung.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+/*
+Copyright 2019 Google LLC
```
jijoongmoon, line 16: Any reaons for this?

jijoongmoon (approved):
LGTM except minor.

nntrainer/utils/thread_manager.h
jaemini-shin: This code includes the modification of [pthreadpool](https://github.com/Maratyszcza/pthreadpool/blob/master/LICENSE), and it requires the copyright and license notice.

nntrainer/utils/thread_manager.cpp
```diff
@@ -0,0 +1,272 @@
...
+  if (has_active_threads == 0)
+    return;
+#elif defined(_WIN32)
+  size_t active_threads = active_threads_.load(std::memory_order_acquire);
+
+  if (active_threads == 0)
+    return;
+#endif
+
+  for (uint32_t i = SPIN_COUNT; i != 0; i--) {
```
junbong-yu, line 183: Isn't there overhead on this code?

jaemini-shin: This spin wait loop is the main key for performance. In this loop, each thread keeps trying to fetch a command with extremely low latency. On the other hand, using futex/mutex would make the thread sleep which results the higher latency.

jaemini-shin: This structure is almost the same as `wait_for_new_command` right above!

junbong-yu (approved):
LGTM

junbong-yu:
If this thread-pool manager is designated for specific purposes such as 'model load' or 'computation' rather than 'general use', it would be more appropriate to clearly indicate a dedicated pool for these purposes.

baek2sm (approved):
Great work, LGTM

PR #3895, opened 2026-04-24
[graph] Skip input tensor check for weight-only layers

Seunghui98:
This commit skips input tensor check for weight-only layers.

## Overview
Update `nntrainer/graph/network_graph.cpp` so that graph construction does not fail for weight-only layers without input tensor.

## Description
- WeightLayer is a source node that does not consume any input tensor — it only outputs its stored weight. 
- However, the graph initialization logic in network_graph.cpp determines whether a node is a source node based solely on the presence of input connections (getInputConnections().empty()). 
- This change adds the required exception handling so valid weight-only layers can be processed correctly.

**Self evaluation:**

### DIFF
--- a/nntrainer/graph/network_graph.cpp
+++ b/nntrainer/graph/network_graph.cpp
@@ -225,7 +225,7 @@ int NetworkGraph::checkCompiledGraph() {
   /** Dimension of input layers must be known */
   for (auto iter = cbegin(); iter != cend(); iter++) {
     auto lnode = (*iter);
-    if (lnode->getNumInputConnections() == 0) {
+    if (lnode->getNumInputConnections() == 0 && lnode->getType() != "weight") {
       if (!lnode->hasInputShapeProperty()) {
         ml_loge("Layer with no inbound connection need input_shape property");
         return ML_ERROR_INVALID_PARAMETER;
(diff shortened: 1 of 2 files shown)


junbong-yu (approved):
LGTM

junbong-yu (approved):
Looks Better Now!

jaemini-shin (commented):
Looks like there are other checks for input node in line 1359, 1571. Is it safe not to update these lines?

Seunghui98:
@jaemini-shin 
> Looks like there are other checks for input node in line 1359, 1571. Is it safe not to update these lines?

yes. This PR skips checking **validation of the input tensor** when the weight layer is input_node, so it is not related to the lines!

nntrainer/graph/network_graph.cpp
```diff
@@ -225,7 +225,13 @@ int NetworkGraph::checkCompiledGraph() {
...
   for (auto iter = cbegin(); iter != cend(); iter++) {
     auto lnode = (*iter);
-    if (lnode->getNumInputConnections() == 0) {
+    /**
+     * Weight layers may have input connections in the graph, but they do not
+     * consume input tensors. They own their weights and expose them as output
+     * tensors instead. Therefore, treat weight-only layers as input-like nodes
+     * here to avoid looking up non-existent input tensors for them.
+     */
+    if (lnode->getNumInputConnections() == 0 && lnode->getType() != "weight") {
```
Seunghui98: @jaemini-shin 
After disussion, I added comments for this background.

jaemini-shin (approved):
LGTM!

PR #3898, opened 2026-04-24
Add safetensors format support to NeuralNetwork save/load

jijoongmoon:
Adds MODEL_FORMAT_SAFETENSORS to the C API (ml_train_model_format_e) and the C++ API (ml::train::ModelFormat)

 so callers can select the safetensors layout when saving or loading model weights.

NeuralNetwork::save() serializes weights as:
  [8B]  header_size (little-endian uint64)
  [header_size B] JSON header with per-tensor {dtype, shape,
                  data_offsets: [start, end]}
  [data] raw tensor bytes in graph order

NeuralNetwork::load() consumes the same layout: it reads the header size and JSON body, resolves a name -> (offset, size) map, and assigns each weight's file offset by name. This removes the ordering constraint that the legacy BIN format imposes and makes loading safe against graph reordering. The BIN path keeps the sequential offset assignment it always used.

In INFERENCE mode the load path spawns one std::thread per graph node, each mmap'ing the file with POSIX_MADV_RANDOM and invoking node->read() against its own view so large weight files stream concurrently. Training mode keeps sequential loading.

The dtype string mapping, JSON header builder, and minimal JSON header parser all live in nntrainer::safetensors (in nntrainer/utils/safetensors_util.{h,cpp}) so that the save/load code in neuralnet.cpp stays short and readable, and so that tools outside the model loader can reuse the same serialization.

convertBinToSafetensors(bin_path, st_path) loads the weights in BIN form and re-serializes them as safetensors for in-place upgrades.

Applications/CausalLM/models/transformer.cpp::load_weight / save_weight now pick the format from the file extension (*.safetensors -> MODEL_FORMAT_SAFETENSORS, otherwise BIN) so CausalLM pipelines pick up the new format with no call-site change.

test/unittest/unittest_nntrainer_safetensors.cpp adds a round-trip suite (save -> load, deterministic re-save byte equality, header layout check, missing-file negative) to pin the on-disk contract.

docs/weight-format-specification.md documents both the legacy BIN layout and the new safetensors layout (header, dtype codes, offsets, alignment).

### DIFF
--- a/Applications/CausalLM/models/transformer.cpp
+++ b/Applications/CausalLM/models/transformer.cpp
@@ -229,7 +229,16 @@ void Transformer::load_weight(const std::string &weight_path) {
   }
 
   try {
-    model->load(weight_path, ml::train::ModelFormat::MODEL_FORMAT_BIN);
+    // Auto-detect format from file extension: *.safetensors -> safetensors,
+    // anything else -> legacy BIN.
+    ml::train::ModelFormat format = ml::train::ModelFormat::MODEL_FORMAT_BIN;
+    const std::string ext = ".safetensors";
+    if (weight_path.size() >= ext.size() &&
+        weight_path.compare(weight_path.size() - ext.size(), ext.size(), ext) ==
+          0) {
+      format = ml::train::ModelFormat::MODEL_FORMAT_SAFETENSORS;
+    }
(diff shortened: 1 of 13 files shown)


jijoongmoon:
This PR is included in #3915

PR #3896, opened 2026-04-24
ml::train::Tensor API with Pimpl + symbolic graph compile

jijoongmoon:
### Redesign ml::train::Tensor API with Pimpl + symbolic graph compile

Reworks the public ml::train::Tensor class from a thin Var_Grad
wrapper into a standalone, Pimpl-backed API that supports:

- Symbolic tensors (constructed from TensorDim) that act as graph
  placeholders bound to internal storage at compile time.
- Eager tensors (fromData/zeros/ones) that carry data immediately.
- Lazy operation chaining (chain/add_i/subtract_i/multiply_i/
  divide_i/pow_i/inv_sqrt_i/eval) implemented with std::function
  queues over the existing internal nntrainer::Tensor ops -- no
  dependency on the LazyTensor class.
- Graph-based Model::compile(Tensor &input, Tensor &output, ...)
  overloads (single input, multi-output, multi-input/multi-output)
  that walk backwards from the output tensors to discover all
  layers and wire input_layers automatically, replacing manual
  addLayer() + string-based input wiring.

The implementation is split across four translation units to keep
each file focused:

  src/tensor_api_impl.h     -- shared Impl / SymbolicGraphNode types
  src/tensor_api.cpp        -- construction, state, data, factories
  src/tensor_api_ops.cpp    -- eager numeric ops + manipulation
  src/tensor_api_lazy.cpp   -- chain / add_i / ... / eval
  src/tensor_api_graph.cpp  -- LayerHandle, implicit ops, compile

api/ccapi/README.md documents the API: motivation, tensor states,
graph construction, compile overloads, lazy chaining, migration
from the legacy addLayer+input_layers pattern, and an end-to-end
example.

### Migrate sample apps and ccapi tests to symbolic tensor graph API
Rewrites 13 example applications (MNIST, Resnet, SimpleFC, LLaMA,
PicoGPT, MixedPrecision, Multi_input, YOLOv2, YOLOv3, ProductRatings,
Android/PicoGPTJNI) to build their models via the new
Model::compile(Tensor, Tensor, ...) graph API instead of the
addLayer + string input_layers pattern.

test/ccapi/unittest_ccapi_tensor.cpp is added with 86 tests
covering symbolic/eager tensor construction, lazy operation
chaining, graph-based compile (single, multi-input, multi-output),
leaf tensor wiring, and picogpt/causallm style integrations.

MNIST/Resnet/ProductRatings meson.build files are updated to drop
the now-unused .ini arguments from the app test invocation — the
[...]

PR #3899, opened 2026-04-25
[Wait 4 #3896] Add ComputeOps function-pointer dispatch table (CPU backends)

jijoongmoon:
Introduces a runtime dispatch table so that tensor operations can be
routed to a vendor's implementation at runtime instead of being
selected at preprocessor time by arch macros.

  nntrainer::ComputeOps (compute_ops.h) is a plain struct of function
  pointers grouped by op family: FP32 BLAS, element-wise, activation,
  transpose/copy, quantized GEMM (Q4_0/Q4_K/Q6_K), quantize/dequantize/
  repack, FP16 variants. Each pointer may be null when a backend does
  not support that op; the caller is expected to fall back to a
  portable implementation.

The dispatch table is exposed both globally (g_compute_ops, for direct
consumers) and through ContextData::setComputeOps / getComputeOps so
that every Context can advertise its own vendor's table in a uniform
way. ContextData is moved to its own header (nntrainer/context_data.h)
to avoid the context.h -> layer_devel.h cycle and to let vendor
ContextData subclasses (QNN, future CUDA/CL) keep their compute ops
private. A type-safe `as<T>()` template is added for vendor downcast.

Per-arch ops table files register wrappers around the existing
compute_backend functions:

  nntrainer/tensor/cpu_backend/arm/arm_ops_table.cpp   -> get_arm_ops()
  nntrainer/tensor/cpu_backend/x86/x86_ops_table.cpp   -> get_x86_ops()
  nntrainer/tensor/cpu_backend/fallback/
     fallback_ops_table.cpp                            -> get_fallback_ops()

init_backend() (per-arch compute_backend.cpp) now assigns g_compute_ops
to the correct table, and AppContext::initialize() publishes the table
to its ContextData so downstream callers can reach it via
context.getContextData()->getComputeOps().

This patch is intentionally additive: the existing direct-call paths
(e.g. nntrainer::sgemm) keep working unchanged. No tensor op is routed
through the table yet. Subsequent PRs will (1) remove #ifdef-based
vendor dispatch in float_tensor.cpp / tensor.cpp by routing through
the table, (2) extend the op set (util_simd, acti_func, OpenCL ops,
QNN plugin), and (3) shrink cpu_backend.h's compile-time forwarder
once all callers are switched.


jijoongmoon:
This PR is included in #3915

PR #3900, opened 2026-04-27
Junbong memory leak fix

junbong-yu:
Fix FSU/NPU memory leak

### DIFF
--- a/nntrainer/mem_allocator.cpp
+++ b/nntrainer/mem_allocator.cpp
@@ -14,9 +14,25 @@
 #include <cstdlib>
 #include <limits>
 #include <mem_allocator.h>
+#include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <numeric>
+#include <stdexcept>
 #include <vector>
+
+#if defined(_WIN32)
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#include <sysinfoapi.h>
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+#if defined(__ANDROID__) && ENABLE_NPU
+#include <dynamic_library_loader.h>
+#endif
 
 namespace nntrainer {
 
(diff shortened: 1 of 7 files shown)


jijoongmoon:
Please give more information in PR comment to understand better. 
What was the issues and problem and how you can resolve the issues. 
Also you should have to leave the signed-off in the PR comment as well.

jijoongmoon:
please fix the ci issues.

nntrainer/mem_allocator.cpp
```diff
@@ -28,4 +44,116 @@ void MemAllocator::alloc(void **ptr, size_t size, size_t alignment) {
...
+}
+
+RpcMemAllocator::~RpcMemAllocator() {
+  if (library_handle != nullptr) {
+    DynamicLibraryLoader::freeLibrary(library_handle);
+    library_handle = nullptr;
+  }
+}
+
+void RpcMemAllocator::alloc(void **ptr, size_t size,
```
jijoongmoon, line 128: RpcMemAllocator should be in qnn_context data. Not in here.

nntrainer/mem_allocator.h
```diff
@@ -30,6 +31,65 @@ class MemAllocator {
...
 };
+
+/**
+ * @brief Page-aligned host memory allocator.
+ *
+ * Uses std::aligned_alloc on POSIX and _aligned_malloc on Windows so the
+ * returned block satisfies the requested alignment. Replaces the ad-hoc
+ * ALIGNED_ALLOC/ALIGNED_FREE macros previously inlined in MemoryPool.
+ */
+class CpuMemAllocator : public MemAllocator {
```
jijoongmoon, line 42: If you want to define the CpuMemAllocator, then it could be in app_context data.

nntrainer/tensor/memory_pool.cpp
```diff
@@ -270,6 +262,20 @@ std::shared_ptr<MemoryData> MemoryPool::getMemory(unsigned int idx) {
  */
 void MemoryPool::deallocate() {
   if (mem_pool != nullptr) {
+    // When memory_ptrs entries are independently allocated blocks (FSU
+    // inference path or Android+NPU rpcmem path), free each unique pointer
+    // before releasing the sentinel mem_pool. Duplicate entries can exist
+    // because the allocation paths reuse a pointer across multiple logical
+    // indices via the offset_ptr map; dedup to avoid double free.
```
jijoongmoon, line 269: "Later, we should unify the CPU and GPU memory allocation strategy. By calculating the total memory required beforehand, we can allocate a single, large contiguous block and manage it using offsets."

nntrainer/tensor/tensor_pool.cpp
```diff
@@ -14,15 +14,40 @@
...
 #include <tensor_wrap_specs.h>
 #include <util_func.h>
 
 namespace nntrainer {
 
+void TensorPool::setAllocator(const std::string &name) {
+  std::shared_ptr<MemAllocator> alloc;
+
+  if (name == "npu") {
+    alloc = tryCreateRpcMemAllocator();
```
jijoongmoon, line 33: I think it should be from qnn_context.

jijoongmoon (changes requested):
Mostly, it looks good to me. But there are several points which are not align in the architectural point of view. I think each HW Context should be provides it's own proper MemoryAlloctor.

jijoongmoon:
please amend the title. You should use proper title. Not "**Junbong memory leak fix"**

PR #3901, opened 2026-04-27
[Wait 4 #3899] Feature/qnn integration

jijoongmoon:
This is the QNN (Qualcomm AI Engine Direct) backend slot, squashed
into a single coherent commit. It plugs the QNN context into the
ComputeOps + ContextData architecture established in the upstream
commits of this series and is gated behind enable-npu so default
builds are unaffected.

What's in
---------
* nntrainer/qnn/ — Qualcomm QNN SDK glue (BackendExtensions, IBackend,
  PAL/Logger, IOTensor wrapper, RPC memory manager). Brought in from
  pr/3826 and adapted. Compiled into a separate libqnn_context.so
  plugin so the Qualcomm code never lands in the main libnntrainer.
* nntrainer/qnn_context.{h,cpp} — QNNContext : Context, Singleton<>.
  initialize() mirrors AppContext / ClContext: bind CPU fallback
  ComputeOps via ensureComputeOps(), set them on this context's
  ContextData (so tensor ops outside the QNN graph still run on CPU),
  use ContextData::as<QNNBackendVar>() for type-safe vendor downcast,
  register the QNN-specific layers (QNNLinear, WeightLayer,
  TensorLayer, QNNGraph).
* nntrainer/engine.cpp — registerContext("qnn", &qnn_context) under
  #if ENABLE_NPU == 1.
* nntrainer/meson.build — enable-npu option pulls in the qnn subdir
  AFTER nntrainer_dep is declared so libqnn_context.so can link
  against libnntrainer.

Cleanups (would have been a separate fixup commit but folded in):
* QNNRpcManager destructor was iterating ptrToFdAndMemHandleMap_
  with a range-for and calling map.erase(mem.first) inside it —
  iterator-invalidation UB. Plus the entire map gets destroyed when
  the object dies, so the per-iteration erase was dead code too.
  Removed it.
* Five unguarded std::cout breadcrumbs in QNNContext (init called,
  before/after dynamicload, before create device, op-package name)
  → ml_logd.
* Three exit(1) calls in QNNContext::init() → return -1, matching
  the same function's existing return-on-failure convention. A
  library function calling exit() hard-kills the host process with
  no recovery path.

Verification
  - Default build (enable-npu=false): QNN code skipped. 1908 tests
    PASSED across the architectural side.
  - QNN-enabled build: cannot exercise here (no QNN SDK / libQnnHtp
    in this environment). Code is the same shape as the upstream
    pr/3826 with the cleanups applied.

Out of scope (tracked):
  - QNNRpcManager has no mutex around its maps. Single-threaded init
    is currently assumed; making that explicit (or adding a mutex)
    is a separate fix.
  - 46K LOC of Qualcomm SDK headers ("Confidential and Proprietary")
    sit in-tree; submodule-ization is its own PR.


baek2sm:
It has already been merged with #3915

PR #3902, opened 2026-04-27
[ Wait 4 #3901] CausalLM: migrate to ml::train::Tensor symbolic graph API

jijoongmoon:
Bring Applications/CausalLM in line with the rest of the migrated apps
(LLaMA / PicoGPT / MNIST / Resnet / SimpleFC / YOLOv2 / YOLOv3 /
ProductRatings / MixedPrecision / Multi_input / Android-PicoGPTJNI),
which already build their models with the new ml::train::Tensor +
LayerHandle::operator() pattern landed in feature/tensor-api.

This is a pure refactor — no behavioural change, no KV cache work.
KV cache externalization (PR-B series) lands on top of this.

What changes
------------
* transformer.h/.cpp — base class
  - typedefs: LayerHandle = ml::train::LayerHandle (was raw shared_ptr),
    + Tensor = ml::train::Tensor.
  - constructModel() returns std::pair<Tensor, Tensor> (input, output) and
    no longer owns model creation / addLayer(). Derived classes extend
    by taking the (input, output_after_norm) and feeding more layers.
  - initialize() now does: createModel → setProperty → constructModel →
    compile(x, y, INFERENCE). The (x, y) overload internally compiles +
    initializes + allocates, replacing the legacy 3-step sequence.
  - createTransformerDecoderBlock / createAttention / createMlp signatures
    all take Tensor inputs and return Tensor — the 'input_layers' string
    plumbing is gone. Multi-input ops use vector form: mha({q, k, v}),
    addition({a, b}), multiply({a, b}), swiglu({gate, up}).

* causal_lm.h/.cpp — extends Transformer's symbolic graph with the
  LM-head and returns the final {input, logits} pair. Inference code
  in run() / generate() / save_kvcache() / load_kvcache() is unchanged.

* Per-model files — virtual override signatures updated, internals
  rewritten to thread Tensor values through layer-call:
    - qwen2/qwen2_causallm                (createAttention)
    - qwen3/qwen3_causallm                (createAttention with reshaped
                                            rms-norm on Q and K)
    - qwen3_moe/qwen3_moe_causallm        (createMlp = qwen_moe layer)
    - qwen3_slim_moe                      (createMlp = moe_slim)
    - qwen3_cached_slim_moe               (createMlp = moe_cached_slim)
    - gemma3/gemma3_causallm              (createAttention with q/k norm,
                                            createMlp with gate-gelu *
                                            up = GeGLU,
                                            createTransformerDecoderBlock
                                            with extra post_attention_norm
                                            and post_ffn_norm)
    - gpt_oss/gptoss_causallm             (createAttention with sink +
                                            yarn rope-scaling,
                                            createMlp = gpt_oss_moe)
    - gpt_oss_cached_slim                 (same shape, gpt_oss_moe_slim_cached)

* sentence_transformer.h/.cpp — addModule(type, idx) is now
  addModule(type, idx, Tensor input) → Tensor output. constructModel
  threads the running tensor through the module loop and returns
  {input, final_output}. Modules with no layer mapping pass the input
  through unchanged so the chain is not broken. Throws if a non-
  Transformer module appears before any Transformer module sets up
  the input tensor.

What does NOT change
--------------------
* Custom layer implementations (mha_core, qkv_layer, swiglu, rms_norm,
  reshaped_rms_norm, embedding_*, lm_head, tie_word_embedding, all MoE
  layers) — internal code untouched, only how they're wired.
* main.cpp — no graph build, only model dispatch + tokenizer plumbing.
* tokenizer / quantize / api wrapper / huggingface_tokenizer.cpp — not
  relevant.
* Inference loop — incremental_inference() / save_kvcache /
  load_kvcache use the same KV cache path that lives inside mha_core.
  PR-B will externalize that.

Verification
------------
* meson setup build -Denable-transformer=true -Denable-tflite-backbone=
  false -Denable-tflite-interpreter=false ; ninja -C build
  → 658/658 targets compiled, including libcausallm.so, nntr_causallm,
    test_api, nntr_quantize, all per-model libs and all custom layer libs.
* clang-format clean.
* Cannot run end-to-end inference here (no model weights / tokenizer
  artifacts on disk in this environment). Architectural refactor, no
  numerical drift expected: the symbolic graph compile produces the
  same layer topology + property strings the old code produced
  through addLayer + input_layers.


baek2sm (approved):
This looks like good work overall. Since several other PRs also touch the CausalLM path, we may need to be careful about merge ordering here.

Would it make sense to migrate the non-CausalLM parts in this PR first, while leaving the CausalLM path unchanged for now? We could then consider migrating CausalLM later, possibly in the Quick.AI repo.

baek2sm:
It has already been merged with #3915

PR #3903, opened 2026-04-27
[ Wait 4 #3902 ] KVCacheManager: standalone host-side KV cache for CausalLM

jijoongmoon:
Add Applications/CausalLM/kv_cache_manager.{h,cpp} — a host-side
container that owns the KV cache memory for all attention layers in a
transformer model. This is the first piece of the KV-cache externali-
zation track (PR-B series). It does NOT yet replace mha_core's internal
cache — that wiring lands in PR-B2. This PR is intentionally standalone:
the class compiles and the unit tests run, but no production code path
calls into it yet.

API
---
  void allocate(num_layers, batch_size, max_seq_len,
                num_heads_kv, head_dim, dtype, format)

  unsigned int getPosition() / setPosition(pos) / advance(step) / reset()

  Tensor &getKeyCache(layer_idx)              // full underlying cache
  Tensor &getValueCache(layer_idx)
  Tensor  getKeyCacheWriteView(layer_idx, batch, step_size)   // [pos .. pos+step)
  Tensor  getValueCacheWriteView(layer_idx, batch, step_size)
  Tensor  getKeyCacheReadView(layer_idx, batch, read_len)     // [0 .. read_len)
  Tensor  getValueCacheReadView(layer_idx, batch, read_len)

  void save(path) / save(path, seq_len) / load(path, seq_len)

Storage layout per layer: (batch, 1, max_seq_len, num_heads_kv*head_dim).
Write/read views are zero-copy `getSharedDataTensor()` slices into the
owning buffer, so binding them to mha_core via setExternalTensors in
PR-B2 will not introduce any copies.

Out of scope
------------
* mha_core wiring (PR-B2: 5-input mode + setExternalTensors).
* KVCacheManager <-> CausalLM lifecycle integration (PR-B5).
* Cache eviction policies, compression, paged attention — design hooks
  only.

Verification
------------
* meson setup build -Denable-transformer=true -Denable-tflite-backbone=
  false -Denable-tflite-interpreter=false ; ninja -C build
  → libcausallm.so + nntr_causallm + test_api + nntr_quantize +
    unittest_kv_cache_manager all build clean.
* build/Applications/CausalLM/unittest_kv_cache_manager
  → 17 tests from KVCacheManagerTest, all PASSED:
      allocate_basic / allocate_invalid_params /
      cache_tensor_dimensions / position_management /
      position_bounds_check / invalid_layer_idx /
      write_view_dimensions / read_view_dimensions /
      write_view_points_to_correct_location /
[...]

PR #3904, opened 2026-04-27
[Wait 4 #3903] mha_core: 5-input external KV cache mode + bind from CausalLM 

jijoongmoon:
Wire the host-owned KVCacheManager (PR-B1) into the inference path by
giving mha_core an external KV cache mode and connecting it through the
symbolic graph.

Framework
---------
* api/ccapi/include/model.h: new public virtual
    Model::setExternalTensors(const std::vector<nntrainer::Tensor> &data,
                              const std::vector<std::string> &names)
  to bind externally-owned tensors into compiled input layers by name.
* nntrainer/models/neuralnet.h: NeuralNetwork override delegates to
    NetworkGraph::setExternalTensors.
* nntrainer/graph/network_graph.h: move setExternalTensors from private
  to public so the wrapper can call it.

mha_core
--------
* finalize() now accepts 3, 4, or 5 inputs. With 5 it sets
    use_external_cache = true
  and skips the internal requestTensor("cache_key/cache_value")
  allocation entirely.
* New public API:
    setCacheIndex(idx) / getCacheIndex()
  to drive the write position from the host (e.g. KVCacheManager).
* Implemented forwarding(context, training):
    - Reads cache from input slot 3 (cache_key) and slot 4 (cache_value)
      instead of internally-owned tensors.
    - step_size derived from input.height(); from = cache_index;
      to = cache_index + step_size.
    - Reuses the existing one_batch_incremental_forwarding() helper for
      both FP32 and FP16 / Android paths and the use_sink branch.
    - Advances cache_index by step_size on completion.
* incremental_forwarding(_from, _to, ...) routes to forwarding() when
  use_external_cache is true (sets cache_index = _from first), so the
  legacy NeuralNetwork::incremental_inference path keeps working without
  any change in callers. PR-B3 will then unify forwarding/incremental
  layer-wide and remove this trampoline.

Graph wiring
------------
* Transformer base gains createKVCachePlaceholders(layer_id, n_heads):
  produces a {cache_k, cache_v} pair of symbolic placeholder Tensors
  named "cache_k_l<i>" / "cache_v_l<i>". Cache shape is
    (BATCH_SIZE, 1, INIT_SEQ_LEN + NUM_TO_GENERATE,
     NUM_HEADS / GQA_SIZE * HEAD_DIM)
  with FP16 (or UINT16 on non-FP16 builds) — matching mha_core's
  pre-existing internal cache type so saved buffers stay binary-
  compatible.
* createAttention in Transformer base + all 5 in-tree overrides
  (qwen2, qwen3, gemma3, gpt_oss, gpt_oss_cached_slim) now request the
  placeholders via that helper and call mha({q, k, v, cache_k, cache_v}).
  MoE variants (qwen3_moe / qwen3_slim_moe / qwen3_cached_slim_moe)
  inherit Qwen3Transformer's createAttention and pick this up
  automatically.

CausalLM lifecycle
------------------
* Adds a KVCacheManager kv_cache member, plus
    allocateAndBindKVCache()  — first call: KVCacheManager::allocate
                                 + Model::setExternalTensors for every
                                 cache_k_l<i> / cache_v_l<i> name. Idempotent.
    setKVCachePosition(pos)   — KVCacheManager::setPosition + push to
                                 every mha_core's setCacheIndex.
    advanceKVCachePosition(s) — KVCacheManager::advance only (mha_core
                                 advances its own cache_index in
                                 forwarding()).
* run() now starts with allocateAndBindKVCache() + setKVCachePosition(0)
  before the prefill / generate loop, so every inference session sees a
  correctly-bound, zeroed cache without changing the user-facing API.
* save_kvcache / load_kvcache rewritten as thin wrappers around
  KVCacheManager::save / load. The forEachLayer + per-layer
  context.getTensor(0/1) plumbing is gone — there are no internal cache
  tensors to dump in external mode. load_kvcache also calls
  setKVCachePosition() so the next forwarding() resumes at the loaded
  offset.

Verification
------------
* meson setup build -Denable-transformer=true -Denable-tflite-backbone=
  false -Denable-tflite-interpreter=false ; ninja -C build
  → 204/204 targets clean (libcausallm.so, nntr_causallm, test_api,
    nntr_quantize, libmha_core_layer.so, all per-model libs,
    unittest_kv_cache_manager).
* build/Applications/CausalLM/unittest_kv_cache_manager
  → 17/17 PASSED (KVCacheManager unit tests carry over from PR-B1).
* clang-format clean.
* End-to-end inference parity: cannot be exercised here without model
  weights / tokenizer artifacts. The two paths run against equivalent
  cache memory layouts (same dtype, same dimension, same write offset
  formula `batch * featureLen + cache_pos * kv_width`), so byte-level
  bit-identity with the legacy 3-input path is the design intent.


jijoongmoon:
This PR is included in #3915

PR #3905, opened 2026-04-28
Enhance/causallm multiturn api

EunjuYang:
## Dependency of the PR
#3867

One can ignore commits from #3867.
Please only review the following commits.

<details><summary>[nntrainer] Expose LayerNode::getLayer() as public API</summary><br />

LayerNode wraps a concrete nntrainer::Layer and is the object handed to
NeuralNetwork::forEachLayer() callbacks (as ml::train::Layer&, via the
LayerNode's base-class inheritance). Reaching the wrapped implementation
from user code currently requires a second dynamic_cast that is silently
denied because getLayer() sits under a private access specifier.

This blocks a legitimate caller pattern: application-level code iterates
layers of a known type (e.g. MHACoreLayer) and needs to invoke methods
specific to that type. Promoting the two getLayer() accessors to public
enables that pattern without changing their behavior.

</details>

<details><summary>[CausalLM] Add MHACoreLayer::resetCache() for multi-turn support</summary><br />

MHACoreLayer maintains cache_index as the absolute write position into
cache_key / cache_value. That index also drives RoPE position encoding
on the next incremental forward, so it must be rewound to 0 before a
new conversation begins — otherwise the second conversation starts
writing at the tail of the previous one and reads back stale KV.

Add two overloads:

  - resetCache()                        : cheap inline, cache_index = 0.
    Sufficient on its own, because the attention read range is bounded
    by cache_index + step_size so garbage past the new write head is
    never dereferenced.

  - resetCache(RunLayerContext &)        : defense-in-depth variant used
    by CausalLM::resetConversation(). Also zero-fills cache_key /
    cache_value so stale entries cannot influence the next turn even
    under future code paths that might read beyond the freshly-written
    range. Guarded by isAllocated() because the tensors carry no backing
    memory until the first incremental_inference() call allocates them.

</details>

<details><summary>[CausalLM] Add MHACoreLayer::resetCache() for multi-turn support</summary><br />

MHACoreLayer maintains cache_index as the absolute write position into
cache_key / cache_value. That index also drives RoPE position encoding
on the next incremental forward, so it must be rewound to 0 before a
new conversation begins — otherwise the second conversation starts
writing at the tail of the previous one and reads back stale KV.

Add two overloads:

  - resetCache()                        : cheap inline, cache_index = 0.
    Sufficient on its own, because the attention read range is bounded
    by cache_index + step_size so garbage past the new write head is
    never dereferenced.

  - resetCache(RunLayerContext &)        : defense-in-depth variant used
    by CausalLM::resetConversation(). Also zero-fills cache_key /
    cache_value so stale entries cannot influence the next turn even
    under future code paths that might read beyond the freshly-written
    range. Guarded by isAllocated() because the tensors carry no backing
    memory until the first incremental_inference() call allocates them.

</details>

<details><summary>[CausalLM] Fix position indexing in run() for multi-turn conversations</summary><br />
The MHA layer keeps cache_index across run() calls, so KV cache state
naturally persists for multi-turn use. CausalLM::run() however was
mixing two different position conventions when stitching multi-turn:

  - Prefill called the model with (SYS_PROMP_LEN, SYS_PROMP_LEN+input_len),
    which ignored prior turns entirely and re-wrote the cache from the
    fixed system-prompt base.
  - Generation called the model with
    (token_generation_idx - 1 + global_token_len,
     token_generation_idx     + global_token_len),
    which DID account for prior turns.

The two phases therefore disagreed on the absolute position of the new
tokens in the cache, so RoPE encodings, the causal mask and the cache
write head all drifted apart on turn N>=2 and the model produced
nonsense.

Fix by snapshotting the absolute base position once per turn and
threading it consistently through prefill, generation and the
teacher-forcing replay index:

  base_pos = SYS_PROMP_LEN + global_token_len

Also:
  - Clear pending_ids_ at the top of run() so a partial UTF-8 / mid-
    punctuation tail from a prior turn does not corrupt this turn's
    decoded output.
  - Add a context-overflow guard before prefill: if base_pos + init_len
    + NUM_TO_GENERATE would exceed MAX_SEQ_LEN, free the input buffer,
    log a clear message, and throw. The C API already wraps run() in a
    try/catch and surfaces this as CAUSAL_LM_ERROR_INFERENCE_FAILED so
    the user is told to call resetConversation().

</details>

<details><summary>[CausalLM] Add CausalLM::resetConversation() to clear multi-turn state</summary><br />
Provide an explicit way to start a fresh conversation on an already-
loaded model without paying the model-load cost again. resetConversation()
returns the instance to the same state it had right after load:

  - global_token_len, SYS_PROMP_LEN, has_run_, output_list and
    pending_ids_ are all cleared.
  - Every MHACoreLayer's KV-cache write head (cache_index) is rewound
    to 0 and its cache_key / cache_value tensors are zero-filled when
    already allocated, so neither the position math nor the attention
    reads can pull in stale state from the previous conversation.

Reaching the wrapped MHACoreLayer through forEachLayer requires going
through LayerNode::getLayer() — the callback receives an
ml::train::Layer& that is in fact a LayerNode wrapper, so a direct
dynamic_cast to MHACoreLayer is silently denied. The lambda routes
through the now-public LayerNode accessor.

</details>

<details><summary>[CausalLM] Add Qwen3-1.7B to C API model enumeration</summary><br />
Register QWEN3-1.7B in the ModelType enum, the path map
(./models/qwen3-1.7b), the reverse name lookup, and the
test_api name->enum dispatch so users can load the larger
model for better multi-turn evaluation.

</details>

<details><summary>[CausalLM] Add resetConversation C API and automatic multi-turn promp…</summary><br />
…t splicing

Adds the public `resetConversation()` C entry point and a `g_turn_count`
state variable so the API layer can distinguish the first turn of a
conversation from continuations.

For continuation turns, the templated prompt is rewritten so that the
system block is dropped and the new user turn is spliced onto the
existing KV cache with a leading `\n` boundary. This matches the way the
chat template would have laid out the conversation if it had been
formatted in a single shot, which lets the cached system + prior turns
be reused verbatim.

</details>

<details><summary>[CausalLM] Add --verify-memory and --multi-turn modes to test_api </summary><br />

Refactors the inference path in test_api into a `run_turn()` helper and
adds two trailing flags to exercise the new multi-turn API end-to-end:

* `--verify-memory` runs a scripted two-turn regression test ("My name
  is Alice" -> "What is my name?"), asserts that the second response
  recalls "Alice" from the cached prior turn, then calls
  `resetConversation()` and confirms the model no longer claims to know
  the name. Returns non-zero on assertion failure.
* `--multi-turn` drops into an interactive REPL with `/reset` and
  `/quit` commands so the multi-turn behavior can be smoke-tested by
  hand.

Existing single-turn invocations are unaffected.

</details>

#### Test result 
```
./test_api QWEN3-0.6B "" 1 W4A32 0 --multi-turn
```
```
you> My name is Alice. Please remember my name.
📝 Input Prompt:
  My name is Alice. Please remember my name.

⚡ Running inference...

💬 Output:
  <think>
  Okay, so I need to remember my name, which is Alice. Let me think. First, I should make sure I'm not mixing up my name with someone else's. Maybe I should write it down in my mind or on a card. But how do I do that without looking at something? Maybe I can visualize it. Like, if I picture a white house with a big door, that's Alice. Or maybe a tree with a big leaf. That's Alice. Or a car with a big engine. That's Alice. So, maybe I can think of different images that represent Alice. But how do I choose the right one? Maybe I can pick a simple image that's easy to remember. Like a house with a big door. That's Alice. So, I can say "Alice, the house with a big door." That's a good way to remember it. Alternatively, maybe a tree with a big leaf. That's also a good memory. But which one is better? Well, maybe the house is more common. But I can choose either. Let me pick the house. So, the answer would be "Alice, the house with a big door." That's a good way to remember it.
  </think>
  
  To remember your name, "Alice" is a simple and memorable name. Here's a way to recall it:
  
  **"Alice, the house with a big door."**  
  
  This is a classic and effective way to remember the name. Let me know if you'd like to add more details!<|im_end|>

  📊 Prefill  tokens=17  111.00 ms  153.2 tok/s
  📊 Generation tokens=313  4610.00 ms  67.9 tok/s

you> What is my name?
📝 Input Prompt:
  What is my name?

⚡ Running inference...

💬 Output:
  <think>
  Okay, the user is asking for their name. But wait, the previous message was about remembering their name. Let me check. The user just said, "What is my name?" But maybe they forgot to mention their name in the previous message. Let me make sure. In the history, the user wrote, "My name is Alice. Please remember my name." Then the assistant responded with "Alice, the house with a big door." Now the user is asking, "What is my name?" So, the user is probably asking for their name again. But maybe they forgot to mention it. Let me respond appropriately. Maybe the user is testing if I can remember their name even if they didn't mention it. So, I should confirm their name and maybe add a note. Let me do that.
  </think>
  
  Your name is Alice. If you ever need to remember it again, you can say, "Alice, the house with a big door." That's a good way to recall it. Let me know if you need anything else!<|im_end|>

  📊 Prefill  tokens=13  94.00 ms  138.3 tok/s
  📊 Generation tokens=213  3483.00 ms  61.2 tok/s

you> /reset
✓ Conversation state reset.

you> What is my name?
📝 Input Prompt:
  What is my name?

⚡ Running inference...

💬 Output:
  <think>
  Okay, the user is asking, "What is my name?" I need to respond appropriately. Since I'm a language model, I don't have a personal name. But maybe I can be helpful. I should acknowledge that I don't have a name but still be useful. I can offer assistance with tasks or help with something else. Let me make sure my response is friendly and helpful.
  </think>
  
  I don't have a name, but I'm here to help! If you have any questions or need assistance, feel free to ask! ��<|im_end|>

  📊 Prefill  tokens=12  43.00 ms  279.1 tok/s
  📊 Generation tokens=113  1585.00 ms  71.3 tok/s

```

### DIFF
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -34,6 +34,7 @@
 
 #include <common.h>
 #include <layer_context.h>
+#include <layer_node.h>
 #include <lm_head.h>
 #include <mha_core.h>
 #include <nntrainer_error.h>
@@ -242,6 +243,42 @@ void CausalLM::registerOutputs(
       }
     }
   }
+}
+
+void CausalLM::resetConversation() {
+  global_token_len = 0;
+  SYS_PROMP_LEN = 0;
+  pending_ids_.clear();
+  output_list.clear();
+  has_run_ = false;
+
+  if (!model)
+    return;
+
+  // Reset cache_index in every MHACoreLayer and zero the underlying
+  // cache_key / cache_value tensors so no stale KV entries from the
+  // previous conversation can influence the next run.
+  //
+  // NeuralNetwork::forEachLayer hands us a reference to the LayerNode (which
+  // inherits from ml::train::Layer), not the concrete layer implementation.
+  // A direct dynamic_cast from ml::train::Layer& to MHACoreLayer* therefore
+  // always fails silently, which was the real root cause of cache_index
+  // leaking across turns (observed as cache_index(277) != _from(0) after a
+  // resetConversation()). Go through LayerNode::getLayer() to reach the
+  // wrapped implementation.
+  std::function<void(ml::train::Layer &, nntrainer::RunLayerContext &, void *)>
+    fn = [](ml::train::Layer &l, nntrainer::RunLayerContext &context,
+            void * /*user_data*/) {
+      if (l.getType() != causallm::MHACoreLayer::type)
+        return;
+      auto *node = dynamic_cast<nntrainer::LayerNode *>(&l);
+      if (!node)
+        return;
+      auto *mha = dynamic_cast<causallm::MHACoreLayer *>(node->getLayer());
+      if (mha)
+        mha->resetCache(context);
+    };
+  model->forEachLayer(fn, nullptr);
 }
 
 void CausalLM::save_kvcache(std::string path, int to_) {
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -610,10 +610,9 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
        ++token_generation_idx) {
 
     allocateAndBindKVCache();
-    auto output_interval =
-      model->incremental_inference(BATCH_SIZE, input, label, input_len,
-                                   token_generation_idx - 1,
-                                   token_generation_idx);
+    auto output_interval = model->incremental_inference(
+      BATCH_SIZE, input, label, input_len, token_generation_idx - 1,
+      token_generation_idx);
     std::vector<unsigned int> ids_list(generate(output_interval[0], do_sample));
 
     // Feed the newly generated token back as the next input token.
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -347,6 +347,11 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
 
   has_run_ = false;
 
+  // Multi-turn safety: drop any decoder state left over from a previous run.
+  // If a prior turn ended mid-punctuation or mid-UTF-8 token, pending_ids_
+  // would leak into this turn's decoding and corrupt output.
+  pending_ids_.clear();
+
   output_list.clear();
   for (unsigned int b = 0; b < BATCH_SIZE; ++b) {
     output_list.push_back("");
@@ -517,6 +522,17 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   allocateAndBindKVCache();
   const unsigned int prefill_from = SYS_PROMP_LEN + global_token_len;
   const unsigned int prefill_to = prefill_from + input_len;
+
+  if (prefill_from + init_len + NUM_TO_GENERATE > MAX_SEQ_LEN) {
+    free(input_sample);
+    std::cerr << "[CausalLM] context overflow: base_pos=" << prefill_from
+              << " input=" << init_len << " gen=" << NUM_TO_GENERATE
+              << " max_seq_len=" << MAX_SEQ_LEN
+              << ". Call resetConversation() to start a new conversation."
+              << std::endl;
+    throw std::runtime_error("CausalLM context overflow");
+  }
+
   setKVCachePosition(prefill_from);
   output = model->incremental_inference(BATCH_SIZE, input, label, init_len,
                                         prefill_from, prefill_to, false);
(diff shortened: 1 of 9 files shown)


Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -159,6 +160,42 @@ void CausalLM::registerOutputs(
...
+  // wrapped implementation.
+  std::function<void(ml::train::Layer &, nntrainer::RunLayerContext &, void *)>
+    fn = [](ml::train::Layer &l, nntrainer::RunLayerContext &context,
+            void * /*user_data*/) {
+      if (l.getType() != causallm::MHACoreLayer::type)
+        return;
+      auto *node = dynamic_cast<nntrainer::LayerNode *>(&l);
+      if (!node)
+        return;
+      auto *mha = dynamic_cast<causallm::MHACoreLayer *>(node->getLayer());
```
jaemini-shin, line 277: Isn't it possible to use static_cast? From line 187, it guarantees that `l` is `nntrainer::LayerNode` and node is `causallm::MHACoreLayer`.

Applications/CausalLM/models/causal_lm.cpp
jaemini-shin: Since `token_generation_idx` starts from `input_len + 1`, this if block is never used.

PR #3932, opened 2026-05-19
[causallm] Fix MHA_Core layer to use actual step size in forwarding

Seunghui98:
This commit fixes MHA_Core layer to use actual step size in external KV cache forwarding path
- Add incremental_step_size member to MHACore Layer to representation _to - _from for incremetal forwarding
- Replace query.height() with incremental_step_size in forwarding() so generation steps compute attention for 1 token, not MAX_SEQ_LEN

## Description

In `MHACoreLayer::incremental_forwarding()`, the external KV cache path passed `_from` to set `cache_index` but silently dropped `_to`. As a result, `forwarding()` fell back to `query.height()` (= MAX_SEQ_LEN = 1024)
as the step size for every call, including single-token generation steps.

This caused each generated token to run the full 1024-token attention computation — identical cost to a complete prefill — instead of attending over just 1 query token against the KV cache.

Fix by storing `_to - _from` in a new member `incremental_step_size` before delegating to `forwarding()`, and using it in place of `query.height()` when non-zero.

### Summary

- Fix external KV cache path ignoring `_to` in `incremental_forwarding()`, causing ~MAX_SEQ_LEN× slowdown during autoregressive generation

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -293,7 +293,8 @@ void MHACoreLayer::forwarding(nntrainer::RunLayerContext &context,
     sink = context.getWeight(sink_idx);
   }
 
-  unsigned int step_size = query.height();
+  unsigned int step_size =
+    (incremental_step_size > 0) ? incremental_step_size : (unsigned int)query.height();
   unsigned int from = cache_index;
   unsigned int to = cache_index + step_size;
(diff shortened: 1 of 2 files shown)


baek2sm (approved):
Thanks, LGTM

jijoongmoon (approved):
This was big bug. LGTM!

PR #3919, opened 2026-05-13
[ChatTemplate] Update chat template to support advanced format

Jungwon-Lee:
#### [Chat_Tempalte] Update Chat template to support jinja

- Update Chat template to support jinja

</details>

#### [ChatTemplate] Remove legacy chat template 

- Remove legacy chat template dependencies

</details>

Signed-off-by: Jungwon-Lee <jungone.lee@samsung.com>

PR #3933, opened 2026-05-19
[Wait for #3919] Add Gemma4 and optimizations

jayden0701:
__New Features:__

- Add Gemma4ForCausalLM model implementation with full architecture support

  - Support for sliding/full attention layers with different RoPE configurations
  - Implement KV-shared layers for memory efficiency
  - Add per-layer input embedding (PLE) support
  - Support double-wide MLP for KV-shared layers

__New Custom Layers:__

- Add PerLayerSliceLayer for selecting per-layer input from packed embedding tensor
- Add ScalarMultiplyLayer for scalar multiplication with optional weight loading
- Add LogitSoftCappingLayer for final logit softcapping (x/softcap -> activation -> *softcap)

__MHA Core Enhancements:__

- Add UseRope property to enable/disable RoPE embeddings
- Add RopePartialRotaryFactor for partial rotary position embeddings
- Add proportional RoPE scaling type support
- Refactor RoPE cache to use shared_ptr with cache key for better memory management

__RMS Norm Enhancements:__

- Add UseGamma property to optionally disable gamma scaling in RMSNorm
- Add ReverseRmsNorm

__Skip Prefill Optimization:__

- Add skip_prefill support to LmHead, TieWordEmbedding, ActivationLayer, AdditionLayer, and ReshapedRMSNorm layers
- Enable optimization for KV-shared layers during prefill phase

__Build System Updates:__

- Update Android.mk and meson.build for new layers and Gemma4 model
- Update install_android.sh for nntr_quantize installation

__Other Changes:__

- Add empty constructor to Transformer base class for child class flexibility
- Update RoPE parameter handling to support rope_parameters config structure
- Register Gemma4 model in main.cpp and quantize.cpp
- Cross platform quanization feature is also included

Signed-off-by: Oh Joonseok <jrock.oh@samsung.com>


Applications/CausalLM/layers/rms_norm.cpp
```diff
@@ -83,8 +83,6 @@ void RMSNormLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
       nntrainer::rms_norm_wrt_width_fp32_intrinsic(
         in_step.getData<float>(), out_step.getData<float>(), dim.height(),
         dim.width(), epsilon);
-
-      throw std::invalid_argument("Error: enable-fp16 is not set");
```
baek2sm: it is fp32 condition-path, so it should be removed(bug).

baek2sm:
```
<bos><|turn>userWrite one short friendly sentence.<turn|><|turn>model
Hope you're having a wonderful day!<turn|>
```

outputs are same with a python model

jijoongmoon (approved):
LGTM

nntrainer/tensor/cpu_backend/arm/neon_impl_fp16.cpp
```diff
@@ -2261,42 +2261,40 @@ void rms_norm_wrt_width_fp16_intrinsic(const float *__restrict X,
...
-      sum_h = sum_h + hx * hx;
+      sum_sq += rowX[i] * rowX[i];
     }
 
-    float mean_single = sum_h / W;
+    float mean_single = sum_sq / W;
     float scale_single = 1.F / std::sqrt(mean_single + eps_h);
-    float16x8_t scale_v = vdupq_n_f16(scale_single);
 
+    // Apply scaling (can still use FP16 for the element-wise multiply)
```
jaemini-shin, line 2290: Every backend implementation should be worked in the same manner, but this logic doesn't.

IMHO, it'd be better to check about current design: use `rms_norm_wrt_width_fp32_intrinsic` regardless of `ENABLE_FP16`.

* Rationale
FP16 option is used for enhance memory bandwidth by storing tensors in half precision. With this strategy, we get better latency, but worse accuracy for both precision and range.

But for this case, there's no advantage using FP16 casting: both input and output are in FP32 type.
* FP16 calculation latency might be shadowed due to the casting from and to FP32.
* Accuracy drops due to FP32 casting.

The previous implementation was meaningful since every internal calculation was in FP16. But for now, there's no reason for keeping it.

jayden0701: I think your point is valid.
We should use `rms_norm_wrt_width_fp32_intrinsic` regardless of `ENABLE_FP16` on normalization.

Than do you think `rms_norm_wrt_width_fp16_intrinsic` should be removed to avoid usage?

jaemini-shin: IMO, it'd be better to remove current `rms_norm_wrt_width_fp16_intrinsic` with float in/out.
If we truely need `rms_norm_wrt_width_fp16_intrinsic` for FP16 activation(probably not sure for accuracy issue), we should implement FP16 in/out interface which is different to this one!

jayden0701:
Checked valid output on x86 / ARM with q4_0 dtype FC / PLE and q6_k embedding(tie_word_embeddings)

baek2sm (approved):
LGTM

Jungwon-Lee:
LGTM

haehun (approved):
LGTM

jayden0701:
Made New PR https://github.com/nntrainer/nntrainer/pull/3954 containing this PR.
Will close

PR #3937, opened 2026-05-20
[CI] enable CausalLM model unittests

baek2sm:
- Enable CausalLM in the Ubuntu Meson CI path.
- Scope the transformer app gate so CI only builds CausalLM, not legacy transformer(llama and pico gpt) apps.

### DIFF
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -39,6 +39,11 @@ jobs:
     - name: install build systems
       run: sudo apt install meson ninja-build
     - run: |
+        causallm_option="-Denable-transformer=false"
+        if [ "${{ matrix.meson_options }}" = "-Denable-fp16=false" ]; then
+          causallm_option="-Denable-transformer=true"
+        fi
+
         meson \
           --buildtype=plain \
           --prefix=/usr \
@@ -47,6 +52,7 @@ jobs:
           --bindir=lib/nntrainer/bin \
           --includedir=include \
           -Dinstall-app=true \
+          ${causallm_option} \
           -Dreduce-tolerance=false \
           -Denable-debug=true \
           -Dml-api-support=enabled \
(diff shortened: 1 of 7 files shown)


Applications/meson.build
```diff
@@ -51,10 +51,6 @@ if host_machine.system() != 'windows'
 endif
 
 if get_option('enable-transformer')
-  subdir('LLaMA/jni')
-  if host_machine.system() != 'windows'
-  subdir('PicoGPT/jni')
-  endif
   if get_option('platform') != 'tizen'
     subdir('CausalLM')
   endif
```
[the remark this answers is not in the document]
baek2sm: It's intentional.

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -58,6 +64,27 @@ jobs:
           -Denable-capi=enabled \
           ${{ matrix.meson_options }} \
           build
+    - name: Check CausalLM model tests are enabled
+      if: ${{ matrix.meson_options == '-Denable-fp16=false' }}
```
jaemini-shin, line 68: Does it have to disable fp16?

baek2sm: It's not ready for enabling causallm CI with FP16 yet. we have to fix some issues and add fp16 unit-test for enabling it.

jaemini-shin (approved):
LGTM!

nntrainer/layers/layer_context.h
```diff
@@ -15,6 +15,7 @@
 #ifndef __LAYER_CONTEXT_H__
 #define __LAYER_CONTEXT_H__
 
+#include <map>
```
jijoongmoon, line 18: Is it intentional? If not, then please remove.

jijoongmoon (approved):
LGTM. Except minor.

nntrainer/layers/layer_context.h
baek2sm: Intentional. `layer_context.h` already uses `std::map` for the `tensor_map` member under `#ifdef DEBUG` (around line 1011), so it previously relied on a transitive include for `<map>`. Adding the explicit `#include <map>` makes the header self-contained, which avoids build breaks on toolchains (e.g. MSVC) that don't pull `<map>` in transitively. I'll keep it.

Applications/meson.build
```diff
@@ -51,10 +51,6 @@ if host_machine.system() != 'windows'
 endif
 
 if get_option('enable-transformer')
-  subdir('LLaMA/jni')
-  if host_machine.system() != 'windows'
-  subdir('PicoGPT/jni')
-  endif
```
Seunghui98, line 57: Why do you remove them?

.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -39,6 +39,11 @@ jobs:
     - name: install build systems
       run: sudo apt install meson ninja-build
     - run: |
+        causallm_option="-Denable-transformer=false"
+        if [ "${{ matrix.meson_options }}" = "-Denable-fp16=false" ]; then
+          causallm_option="-Denable-transformer=true"
+        fi
```
Seunghui98, line 45: Why are the options of `enable-fp16` and `enable-transformer` dependent on each other?

Seunghui98 (approved):
LGTM 👍 , but I have two questions( https://github.com/nntrainer/nntrainer/pull/3937/changes#r3411389959, https://github.com/nntrainer/nntrainer/pull/3937/changes#r3411381902).

PR #3939, opened 2026-05-20
[Optimization] rope caching for optimization

baek2sm:
## Summary
- Cache resolved RoPE cos/sin frequency table pointers per MHA layer
- Avoid repeated mutex, cache-key construction, and map lookup on warm decode calls

## Verification
- Observed a small decode TPS improvement(4-6%) in local Windows testing

### DIFF
--- a/Applications/CausalLM/layers/mha_core.h
+++ b/Applications/CausalLM/layers/mha_core.h
@@ -407,9 +407,13 @@ private:
   inline static std::vector<std::vector<float>> *freqs_cos = {};
   inline static std::vector<std::vector<float>> *freqs_sin = {};
   inline static std::vector<float> thetas;
+  std::vector<std::vector<float>> *cached_freqs_cos = {};
+  std::vector<std::vector<float>> *cached_freqs_sin = {};
 #ifdef ENABLE_FP16
   inline static std::vector<std::vector<_FP16>> *freqs_cos_fp16 = {};
   inline static std::vector<std::vector<_FP16>> *freqs_sin_fp16 = {};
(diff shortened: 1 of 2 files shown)


Applications/CausalLM/layers/mha_core.h
```diff
@@ -407,9 +407,13 @@ WIN_EXPORT class MHACoreLayer : public nntrainer::LayerImpl {
   inline static std::vector<std::vector<float>> *freqs_cos = {};
   inline static std::vector<std::vector<float>> *freqs_sin = {};
   inline static std::vector<float> thetas;
+  std::vector<std::vector<float>> *cached_freqs_cos = {};
+  std::vector<std::vector<float>> *cached_freqs_sin = {};
```
jaemini-shin, line 411: Is it necessary to declare these new variables? Is there any problem checking `freqs_cos == nullptr`?

baek2sm: because `freqs_cos` is static variable, It can cause value mismatch when if theta value is different (for some model, it has different theta value between sliding and global layer). so, We've handled it by look-up structure for getting freqs_cos every step, and now, this patch remove repeated look-up call using cached_freqs_cos for every instance.

jijoongmoon:
I don't understand why do we need this cached ? we already have the sin and cos which pre-computed in initialization. and it just assign them local variable.

jijoongmoon (changes requested):
I don't understand why do we need this cached ? we already have the sin and cos which pre-computed in initialization. and it just assign them local variable.

PR #3940, opened 2026-05-21
Fix model load crashes on Android

Seunghui98:

<details><summary>[api] tensor-api: preserve leaf tensor dtype for auto-created inputs</summary><br />

[[api] tensor-api: preserve leaf tensor dtype for auto-created inputs](https://github.com/nntrainer/nntrainer/commit/7a818e4c0ecf4e3470a05d1e369c7f4858881723) 

This commit preserve leaf tensor dtype for auto-created inputs.
- When Tensor API compile() finds additional leaf tensors, it creates input layers for tem automatically.
- However, the generated input layer only kept the tensor name and shape, dropping the dtype.
- This caused FP16 KV-cache placeholders created from symbolic tensors to be restored as the default dtype(FP32).
- On Android builds with ENABLE_FP16=1, the cache buffer was allocated as FP16 but the synthesized input placeholder became FP32, causing dtype mismatch.

**Self evaluation:**

</details>

<details><summary>[neuralnet] share a single mmap across model-load workers</summary><br />

[[neuralnet] share a single mmap across model-load workers](https://github.com/nntrainer/nntrainer/commit/0cab25b9cadf095f5cc4357d3e6730de4d0b1f71) 

This commit shares a single mmap across model-load workers.
- Parallel weight loading previously let each worker mmap the full weight file independently.
- For large models, this created many mappings of the same file and could exceed Android's virtual memory or map-count limits, causing mmap failures.
Create a single read-only mmap in the parent thread and shared it with all workers.
- Since each weight already has a fixed file offset, workers can safely read from the shared mapping without synchronization.
Run POSIX_MADV_DONTNEED only after all workers finish, avoiding races with other readers of the shared mapping.

**Self evaluation:**

</details>

### Summary

Fix two crashes that prevented Q4_0 models (e.g. Qwen3-0.6B) from loading on Android.

#### 1. Share a single mmap across model-load workers 
- Each worker thread was calling `mmap()` on the full `.bin` independently, consuming `graph_size × file_size` of virtual address space (~300 GB for a ~3 GB bin with 100 nodes)
- Fix: map the file once in the parent and share the read-only view across all workers; `munmap` after join

#### 2. Propagate leaf tensor dtype into auto-created input layers 
- With `ENABLE_FP16`, KV cache placeholders synthesized by `compile()` were defaulting to FP32, while `KVCacheManager` allocated FP16,causing `allocateAndBindKVCache: cache placeholder dtype mismatch`.
- Fix: carry `leaf_info.dim.getDataType()` as `input_dtype=` when non-FP32

**Self evaluation:**

### DIFF
--- a/api/ccapi/src/tensor_api_graph.cpp
+++ b/api/ccapi/src/tensor_api_graph.cpp
@@ -406,8 +406,37 @@ int Model::compile(std::vector<Tensor> &inputs, std::vector<Tensor> &outputs,
     std::string leaf_shape = std::to_string(leaf_info.dim.channel()) + ":" +
                              std::to_string(leaf_info.dim.height()) + ":" +
                              std::to_string(leaf_info.dim.width());
-    auto leaf_layer =
-      createLayer("input", {"name=" + leaf_name, "input_shape=" + leaf_shape});
+    auto dtype_to_str = [](nntrainer::TensorDim::DataType dt) -> const char * {
+      using DT = nntrainer::TensorDim::DataType;
+      switch (dt) {
+      case DT::FP32: return "FP32";
+      case DT::FP16: return "FP16";
+      case DT::QINT4: return "QINT4";
+      case DT::QINT8: return "QINT8";
+      case DT::QINT16: return "QINT16";
+      case DT::UINT4: return "UINT4";
+      case DT::UINT8: return "UINT8";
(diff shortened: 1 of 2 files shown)


haehun (approved):
Nice catch!

jijoongmoon (approved):
LGTM!

PR #3943, opened 2026-05-23
Causal Flash attention

jijoongmoon:
## Summary

  This PR generalises the existing non-causal flash-attention path in
  `MHACoreLayer` to cover **causal LLM prefill with GQA and sliding
  window**, validated end-to-end on **Qwen3-0.6B** on Galaxy S25 / S26
  Ultra and on x86. It also lands a self-tuned **AVX2 + F16C fused
  hsgemm kernel** so x86 can use the same flash path without falling
  back to OpenBLAS-sgemm-with-staging, an **ALL-FP16 ARM path** that
  matches the pre-V-JEPA `mha_core` precision exactly
  (FP16 storage + FP32 partial accumulation, no FP32 score buffer),
  a `use_flash_attention` `nntr_config.json` option, and two
  independent `nntrainer` bug fixes that surfaced while bringing the
  path up at `ENABLE_FP16=1`.

  V-JEPA encoder support that introduced the original non-causal flash
  path is **not part of this PR** — it lives on a separate branch.
  
  ## Causal flash (`MHACoreLayer::gemm_attention`)
  
  - Lifts the `!is_causal && gqa_size == 1` gate. Single
    `gemm_attention(query, K, V, out, N_kv, N_q, cache_from)` covers
    both encoder non-causal and causal-LLM prefill.
  - **Causal upper-bound block skip**: when the smallest `k_abs` in a
    key block is past the largest `q_abs` in the current query block,
    `break` the key loop — this and every later key block contribute
    zero FLOPs.
  - **Causal boundary mask**: blocks straddling the diagonal compute
    the full QK tile and then set
    `S[i, k] = -INFINITY` for `k > q_abs(i) - kb`. Bounded waste — at
    most one boundary block per query-block per head.
  - **GQA pointer math**: `h_kv = h_q / gqa_size`. Q is de-interleaved
    at `num_heads_Q` granularity; K / V at `num_heads_KV` granularity.
  - **Sliding-window** lower-bound block skip + in-block lower mask.
  - Min-prefill gate (`FLASH_MIN_PREFILL = 32`): single-token decode
    always falls through to the existing per-row dot kernels.

  ## x86 AVX2 + F16C fused hsgemm (`mha_hsgemm_avx2`)

  - Inline `FP32 × FP16-bits → FP32` GEMM:
    - QK path: 4-row register block, F16-load + `_mm256_cvtph_ps` to
      widen K, FMA into 4 parallel FP32 accumulators.
    - AV path: 8-wide N block, broadcast `A[m, k]`, contiguous F16
      load of B, FMA.
  - Removes a 14 MB Phase-1 FP32 staging buffer that the previous
    "convert-then-sgemm" path required and halves the K / V memory
    traffic by reading FP16 bits directly.
  - Used on x86 (`__x86_64__ || __i386__`) and `-march=native -mavx2
    -mfma` already enables the required F16C intrinsics on every
    modern x86_64 CPU.

  ## ALL-FP16 ARM path (pre-V-JEPA precision)

  When `forwarding()` passes FP16 query / output (the
  `ENABLE_FP16 && __ANDROID__` build), the flash dispatches to a
  fully FP16 path with FP32 partial accumulation only inside the GEMM
  kernels — no FP32 score buffer anywhere:
  - QK: `nntrainer::neon::custom_hgemm` (FP16 × FP16 → FP16).
  - Softmax: read FP16 from `Sp16`, compute exp in FP32 register,
    store FP16 back into `Sp16` (fused; no separate FP32→FP16
    conversion pass).
  - AV: `nntrainer::neon::custom_hgemm` consuming `Sp16` and the
    FP16 V tile.
  - Output scatter writes FP16 to `attention_output_step` when the
    caller's output buffer is FP16.
  
  Otherwise (FP32 Q from `Q4_0-FP32` activation) the path uses the
  existing `shgemm` + fused FP16 softmax `Sp16` + `custom_hgemm` AV
  sequence.
  
  ## Build / config wiring
  
  - New `use_flash_attention` field in `nntr_config.json` (default
    `true`). `Transformer::createAttention` and
    `Qwen3Transformer::createAttention` plumb the flag to `mha_core`'s
    `use_gemm_attention` property.
  - ARM build (`Applications/CausalLM/jni/Android.mk`) switched to
    `ENABLE_FP16=1` so the in-`mha_core` attention runs at the same
    FP16-storage + FP32-partial-accumulation precision the NPU target
    expects.
  
  ## Two `nntrainer` fixes (surfaced while enabling FP16 activation)

  Both correct in isolation, regardless of `model_tensor_type`:
  
  1. **`HalfTensor::dot` Q4_0 case** was an empty `break` — the FP16-
     activation FC layer with a Q4_0 weight left its output buffer
     uninitialised. Routed to `dotQnK`, which is already implemented
     and dispatches to `gemm_q4_0_fp16`.
  2. **`InputLayer::finalize`** silently promoted any FP32 input dim
     to the model's activation dtype, clobbering integer-valued
     inputs (Qwen3 vocab id 151643 saturates FP16 to `-inf` and
     breaks the embedding bounds check). Now preserves the declared
     input dtype.
  3. (Bonus) **Embedding's hard FP32 input check** removed in
     `embedding_layer.cpp`, `tie_word_embedding.cpp`, and
     `nntrainer/layers/embedding.cpp` so future FP16-activation
     models can instantiate.
  
  ## Benchmarks — Qwen3-0.6B Q4_0, 1 003-token prefill + 32-token greedy decode

  `NNTR_NUM_THREADS=8`, S25 Ultra, output token-identical to the
  reference path:
  
  | Build           | Path                            | Prefill        | Decode         | e2e        |
  |---|---|---|---|---|
  | `ENABLE_FP16=0` | reference (no flash)            | 2 967 ms (338) | 586 ms (54.6) | 3 580 ms   |
  | `ENABLE_FP16=0` | flash (V-JEPA-style `shgemm`)   | **1 458 ms (688)** | 583 ms (54.9) | **2 070 ms** |
  | `ENABLE_FP16=1` | reference (FP16 throughout)     | 1 752 ms (572) | 484 ms (66.1) | 2 259 ms   |
  | `ENABLE_FP16=1` | flash, Phase-1 Q FP16→FP32 + `shgemm` | 1 535 ms (653) | 586 ms (54.6) | 2 149 ms   |
  | `ENABLE_FP16=1` | flash, ALL-FP16 (`custom_hgemm`)| 1 656 ms (606) | 486 ms (65.8) | 2 166 ms   |

  On the ARM device the flash path is ~2 × faster than the reference
  attention at this sequence length, with peak RAM unchanged
  (~965 MB) — the model weights and KV cache dominate.

  Kernel profile (`std::chrono::steady_clock` summed across all 2 688
  per-inference QK calls, 8 workers, same workload):
  
  | QK kernel                       | Aggregate | Per call |
  |---|---|---|
  | `nntrainer::shgemm` (scopy + `__cblas_sgemm`) | 1 649 ms | 613 µs |
  | `nntrainer::neon::custom_hgemm`               | 2 295 ms | 854 µs |
  
  OpenBLAS sgemm out-tunes the in-tree FP16 GEMM by ~28 % on these
  shapes (M = 256, K = 128, N = 512 for QK; M = 256, K = 512, N = 128
  for AV) — tuning `custom_hgemm`'s register / cache blocking would
  flip the ranking and make the all-FP16 path strictly faster.
  
  Detailed precision matrix and benchmark sweep:
  `Applications/CausalLM/layers/FLASH_ATTENTION.md`.

  - QK: `nntrainer::neon::custom_hgemm` (FP16 × FP16 → FP16).
  - Softmax: read FP16 from `Sp16`, compute exp in FP32 register,
    store FP16 back into `Sp16` (fused; no separate FP32→FP16
    conversion pass).
  - AV: `nntrainer::neon::custom_hgemm` consuming `Sp16` and the
    FP16 V tile.
  - Output scatter writes FP16 to `attention_output_step` when the
    caller's output buffer is FP16.
  
  Otherwise (FP32 Q from `Q4_0-FP32` activation) the path uses the
  existing `shgemm` + fused FP16 softmax `Sp16` + `custom_hgemm` AV
  sequence.
  
  ## Build / config wiring
  
  - New `use_flash_attention` field in `nntr_config.json` (default
    `true`). `Transformer::createAttention` and
    `Qwen3Transformer::createAttention` plumb the flag to `mha_core`'s
    `use_gemm_attention` property.
  - ARM build (`Applications/CausalLM/jni/Android.mk`) switched to
    `ENABLE_FP16=1` so the in-`mha_core` attention runs at the same
    FP16-storage + FP32-partial-accumulation precision the NPU target
    expects.
  
  ## Two `nntrainer` fixes (surfaced while enabling FP16 activation)

  Both correct in isolation, regardless of `model_tensor_type`:
  
  1. **`HalfTensor::dot` Q4_0 case** was an empty `break` — the FP16-
     activation FC layer with a Q4_0 weight left its output buffer
     uninitialised. Routed to `dotQnK`, which is already implemented
     and dispatches to `gemm_q4_0_fp16`.
  2. **`InputLayer::finalize`** silently promoted any FP32 input dim
     to the model's activation dtype, clobbering integer-valued
     inputs (Qwen3 vocab id 151643 saturates FP16 to `-inf` and
     breaks the embedding bounds check). Now preserves the declared
     input dtype.
  3. (Bonus) **Embedding's hard FP32 input check** removed in
     `embedding_layer.cpp`, `tie_word_embedding.cpp`, and
     `nntrainer/layers/embedding.cpp` so future FP16-activation
     models can instantiate.
  
  ## Benchmarks — Qwen3-0.6B Q4_0, 1 003-token prefill + 32-token greedy decode

  `NNTR_NUM_THREADS=8`, S25 Ultra, output token-identical to the
  reference path:
  
  | Build           | Path                            | Prefill        | Decode         | e2e        |
  |---|---|---|---|---|
  | `ENABLE_FP16=0` | reference (no flash)            | 2 967 ms (338) | 586 ms (54.6) | 3 580 ms   |
  | `ENABLE_FP16=0` | flash (V-JEPA-style `shgemm`)   | **1 458 ms (688)** | 583 ms (54.9) | **2 070 ms** |
  | `ENABLE_FP16=1` | reference (FP16 throughout)     | 1 752 ms (572) | 484 ms (66.1) | 2 259 ms   |
  | `ENABLE_FP16=1` | flash, Phase-1 Q FP16→FP32 + `shgemm` | 1 535 ms (653) | 586 ms (54.6) | 2 149 ms   |
  | `ENABLE_FP16=1` | flash, ALL-FP16 (`custom_hgemm`)| 1 656 ms (606) | 486 ms (65.8) | 2 166 ms   |

  On the ARM device the flash path is ~2 × faster than the reference
  attention at this sequence length, with peak RAM unchanged
  (~965 MB) — the model weights and KV cache dominate.

  Kernel profile (`std::chrono::steady_clock` summed across all 2 688
  per-inference QK calls, 8 workers, same workload):
  
  | QK kernel                       | Aggregate | Per call |
  |---|---|---|
  | `nntrainer::shgemm` (scopy + `__cblas_sgemm`) | 1 649 ms | 613 µs |
  | `nntrainer::neon::custom_hgemm`               | 2 295 ms | 854 µs |
  
  OpenBLAS sgemm out-tunes the in-tree FP16 GEMM by ~28 % on these
  shapes (M = 256, K = 128, N = 512 for QK; M = 256, K = 512, N = 128
  for AV) — tuning `custom_hgemm`'s register / cache blocking would
  flip the ranking and make the all-FP16 path strictly faster.
  
  Detailed precision matrix and benchmark sweep:
  `Applications/CausalLM/layers/FLASH_ATTENTION.md`.

### DIFF
--- a/Applications/CausalLM/jni/Android.mk
+++ b/Applications/CausalLM/jni/Android.mk
@@ -27,6 +27,7 @@ CAUSALLM_COMMON_INCLUDES := \
     $(LOCAL_PATH)/../models/qwen3_cached_slim_moe \
     $(LOCAL_PATH)/../models/gemma3 \
     $(LOCAL_PATH)/../models/timm_vit \
+    $(LOCAL_PATH)/../models/vjepa2_vit \
     $(LOCAL_PATH)/../models/deberta_v2 \
     $(LOCAL_PATH)/../third_party/minja/include \
     $(LOCAL_PATH)/../third_party \
@@ -52,14 +53,14 @@ include $(PREBUILT_STATIC_LIBRARY)
 include $(CLEAR_VARS)
 
 LOCAL_ARM_NEON := true
-LOCAL_CFLAGS += -std=c++17 -Ofast -mcpu=cortex-a53 -Ilz4-nougat/lib -DENABLE_FP16=1 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1 -mtune=cortex-a76 -O3 -ffast-math
+LOCAL_CFLAGS += -std=c++17 -Ofast -mcpu=cortex-a53 -Ilz4-nougat/lib -DENABLE_FP16=0 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1 -mtune=cortex-a76 -O3 -ffast-math
 LOCAL_LDFLAGS += -Llz4-nougat/lib/obj/local/$(TARGET_ARCH_ABI)/
 LOCAL_CXXFLAGS += -std=c++17 -frtti
-LOCAL_CFLAGS += -pthread -fexceptions -DENABLE_FP16=1 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1 -mtune=cortex-a76 -O3 -ffast-math
-LOCAL_LDFLAGS += -fexceptions -DENABLE_FP16=1 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1 -mtune=cortex-a76 -O3 -ffast-math
+LOCAL_CFLAGS += -pthread -fexceptions -DENABLE_FP16=0 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1 -mtune=cortex-a76 -O3 -ffast-math
+LOCAL_LDFLAGS += -fexceptions -DENABLE_FP16=0 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1 -mtune=cortex-a76 -O3 -ffast-math
 LOCAL_ARM_MODE := arm
 LOCAL_MODULE := causallm_core
-LOCAL_LDLIBS := -llog -landroid -DENABLE_FP16=1 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1
+LOCAL_LDLIBS := -llog -landroid -DENABLE_FP16=0 -DUSE__FP16=1 -D__ARM_NEON__=1 -march=armv8.2-a+fp16+dotprod+i8mm -DUSE_NEON=1
 
 LOCAL_SRC_FILES := \
     ../chat_template.cpp \
@@ -97,6 +98,10 @@ LOCAL_SRC_FILES := \
     ../models/gemma3/embedding_gemma.cpp \
     ../models/gemma3/function.cpp \
     ../models/timm_vit/timm_vit_transformer.cpp \
+    ../models/vjepa2_vit/vjepa2_vit.cpp \
+    ../layers/vjepa_rope_layer.cpp \
+    ../layers/vjepa_gelu_layer.cpp \
+    ../layers/vjepa_layernorm_layer.cpp \
     ../models/deberta_v2/deberta_v2.cpp \
     ../layers/deberta_attention_layer.cpp \
     ../layers/shared_fully_connected_layer.cpp \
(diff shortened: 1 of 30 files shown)


baek2sm:
**Final Verdict: ✅ Approve**

---

Adds causal flash attention to `MHACoreLayer` for LLM prefill, generalizing the existing non-causal flash path.

**Included features**
- Causal + sliding-window masking with GQA pointer arithmetic; `use_flash_attention` config option.
- Fused AVX2+F16C x86 `hsgemm` kernel (reads FP16 KV cache directly) and a V-JEPA-style all-FP16 NEON ARM path.
- Ships the V-JEPA 2.1 ViT-B video encoder + custom layers and a RoPE frequency-table cache.
- Two core bug fixes: `HalfTensor::dot` Q4_0 path and removal of the implicit FP32→activation promotion in `InputLayer::finalize`.

---

### 2. Related Issues

- **P3** CI brought green: added doxygen `@author`/`@bug`/class `@brief`, applied clang-format-14, and corrected the mixed-precision golden-test harness so FP16-input models read the FP32-layout golden stream correctly (the implicit-promotion removal had misaligned the read).
- **P3** The x86 intrinsic guard would be more robust as `defined(__AVX2__) && defined(__F16C__)` rather than `__x86_64__` — non-blocking follow-up.

No P0/P1/P2 issues.

---

The widest-blast change is the deliberate removal of implicit FP32→activation promotion in `InputLayer::finalize`; the only dependent test (`unittest_mixed_models`) was reconciled via the harness read fix and now passes on the FP16 lanes. `UseGemmAttention` defaults to false so existing MHA callers are unaffected. The full CI matrix (Windows/Ubuntu ±fp16/Android/Tizen/cross-ARM64/static/linter) is green.

---

A substantial, well-contained feature; all CI gates root-caused and fixed (formatting, doxygen, and a genuine golden-harness alignment bug), with the full matrix green. Approve.

PR #3910, opened 2026-05-01
[Wait 4 #3834] [x86] feat: Add AVX2 FP16 implementation for x86 CPU Backend

KWSMooBang:
## Dependency of the PR
+ Depends on: #3820 , #3834
+ Related to: #3821 

## Commits to be reviewed in this PR

<details><summary>refactor: extract shared AVX2 helpers into avx2_internal.h (5699b4e5) </summary><br />

- Add `avx2_internal.h` for shared AVX2 utility functions.
- Move common helpers used by FP32/FP16 kernels:
  - `exp256_ps`
  - `rcp_ps`
  - `hsum_avx`
  - GELU/SwiGLU polynomial helper routines
  - AVX2 constants and compiler attribute helpers
- Reduce duplicated helper code before adding FP16 AVX2/F16C kernels.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: YongHyeon02 <dyddyd8574@snu.ac.kr>

</details>

<details><summary>feat: replace FP16 scalar fallbacks with AVX2+F16C SIMD implementations (192e3d43)</summary><br />

Add AVX2+F16C optimized FP16 backend implementations and connect them to the x86 FP16 dispatch layer.

Add `avx2_impl_fp16.cpp` implementations for:
- FP16 conversion:
  - `vcvt_f16_f32`
  - `vcvt_f32_f16`
- Element-wise ops:
  - `ele_mul`
  - `ele_add`
  - `ele_sub`
  - `ele_div`
- BLAS-like vector ops:
  - `saxpy`
  - `sdot`
  - `snrm2`
  - `sscal`
  - `custom_scopy`
- Reduction / activation / normalization:
  - `max_val`
  - `softmax`
  - `inv_sqrt_inplace`
  - `swiglu`
  - `rms_norm_wrt_width_fp16`
- LLM-related helper ops:
  - `compute_rotary_embedding_value`
  - `isamax`
  - `transpose_matrix`
  - `scopy_int4_to_float16`
  - `scopy_int8_to_float16`

Implementation details:
- Use F16C conversion instructions to convert FP16 data to FP32 SIMD vectors.
- Perform arithmetic in FP32 using AVX2 intrinsics.
- Convert results back to FP16 before storing.
- Preserve the existing NNTrainer CPU backend public interface.

Integrate AVX2 dispatch into `x86_compute_backend_fp16.cpp` for:
[...]

PR #3967, opened 2026-06-05
[Wait 4 #3910] [x86] feat: Add cache-blocked FP16 GEMM/GEMV for x86 CPU Backend

YongHyeon02:
## Dependency of the PR

- Depends on: #3910
- Benchmark tooling: #3820

## Summary

This PR replaces the x86 FP16 `sgemm` / `sgemv` (and the mixed-precision `shgemm` / `hsgemm` / `shgemv` / `hsgemv`) **triple-conversion** path — which allocated full FP32 copies of every operand, called CBLAS, and converted the result back — with a self-contained **cache-blocked GEMM** that converts FP16→FP32 *during packing* in L1/L2-sized blocks and computes in FP32 with AVX2+F16C micro-kernels.

Key properties:

1. **Zero per-call allocations.** A thread-local reusable workspace holds the
   FP32 accumulator panel and the packed A/B buffers, replacing the previous 3 `new float[]` allocations per call (up to ~63 MB/call for large GEMM).
2. **Conversion folded into packing.** No full-matrix FP16↔FP32 copies; the FP16→FP32 widening happens 8/16 lanes at a time inside the packing loop.
3. **Goto-style blocking** with a 6×16 primary micro-kernel plus an `mxn` edge kernel suite (1×8 … 6×16), zero-padded M/N edges, and small/skinny/`K`-small fast paths.
4. **Multi-threaded** via the unified `ThreadManager`, parallelized over output panels that are subdivided to feed all worker threads.
5. **Mixed precision reuses the same kernels** through a templated `<AType, BType, CType>` entry point; the FP32-output variants skip the final narrowing.

<details><summary>feat: add x86 AVX2+F16C FP16 GEMV for CPU backend (b46fe300)</summary><br />

- Add a custom AVX2+F16C matrix-vector path for FP16 `sgemv` and mixed-precision `shgemv` / `hsgemv` through a shared templated `gemv_impl`
- NoTrans stages the X vector once in FP32 and reduces along unit-stride A rows; TransA accumulates scaled rows into an FP32 scratch output
- Removes the per-call FP32 allocation and full-matrix conversion of the legacy triple-conversion sgemv

**Self evaluation:**

</details>

<details><summary>feat: add x86 cache-blocked FP16 GEMM for CPU backend (838fefdf)</summary><br />

- Add `nntrainer/tensor/cpu_backend/x86/hgemm/`: a cache-blocked FP16 GEMM that converts FP16→FP32 during packing and computes in FP32 with AVX2+F16C micro-kernels
- Goto-style blocking, a 6×16 primary kernel plus an `mxn` edge-kernel suite, zero-padded M/N edges, and small/skinny/`K`-small fast paths
- Thread-local zero-allocation workspace, per-microarchitecture block sizes, and `ThreadManager` panel parallelism subdivided to feed all worker threads
- Route FP16 `sgemm` and mixed-precision `shgemm` / `hsgemm` (row-major) to the new path
- Expose the public `hgemm` / `shgemm` / `hsgemm` entry points under `nntrainer::avx2` (matching the GEMV entry points); keep all blocking, packing, kernel, and workspace internals in `nntrainer::avx2::internal`
- Unit tests for the cache-blocked `sgemm`, mixed-precision `shgemm` / `hsgemm`, the FP16 `sgemv` / `shgemv` / `hsgemv` paths, and workspace reallocation statistics — aligned/unaligned shapes, all transpose combinations, alpha/beta scaling, zero-size and zero-K cases, strided vectors, and fast-path branches, registered under the `enable-fp16` gate

**Self evaluation:**

</details>

## Performance Improvements

Benchmark environment: Intel Core i5-12400F (6 P-cores / 12T, AVX2+F16C), Linux WSL2
Benchmark tool: `benchmark_x86_backend` (iters=30, warmup=8), GCC 12.3, `enable-fp16=true`
Reference: legacy FP16 triple-conversion path (`new[]`×3 + FP16↔FP32 convert + CBLAS), measured in the same binary.
`speedup = legacy / new`; ST = single thread (`NNTR_NUM_THREADS=1`, `OPENBLAS_NUM_THREADS=1`), MT = 6 threads (`6`/`6`).

### FP16 sgemm

| M×N×K | ST speedup | MT speedup |
|:------|-----------:|-----------:|
| 1024×1024×1024  | 1.19x | **1.87x** |
| 256×1024×1024   | 1.18x | 1.54x |
| 256×3072×3072   | 1.33x | **2.33x** |
| 1024×3072×3072  | 1.05x | 1.46x |
| 16×1024×1024    | 2.67x | **8.85x** |
| 1×4096×4096     | 5.20x | **14.30x** |
| 1×768×1024      | 12.16x | 7.34x |
| 256×256×256     | 1.73x | 1.74x |

### FP16 sgemv

| M×N | ST speedup | MT speedup |
|:----|-----------:|-----------:|
| 1×768      | 18.93x | 10.22x |
| 16×1024    | 4.22x  | 6.82x |
| 256×1024   | 3.74x  | 4.72x |
| 256×3072   | 4.41x  | 5.65x |
| 1024×3072  | 6.05x  | 4.57x |

### Highlights

- **Per-call allocations: 3 → 0** for every FP16 GEMM/GEMV call (thread-local workspace).
- Faster than the legacy convert+CBLAS path on every non-trivial shape, in both single- and multi-threaded runs.
- Memory-bound GEMV gains the most (4–19×); skinny/tall GEMM 5–14× under MT.
- Mixed-precision `shgemm`/`hsgemm`/`shgemv`/`hsgemv` reuse the same kernels with FP32 output (no narrowing).

## Known Limitations

- **Very small shapes** (e.g. 7×17×33) are slower than a direct CBLAS call; absolute cost is sub-microsecond and these are handled correctly via the scalar fast path.
- **Column-major (`TStorageOrder != 0`)** FP16 GEMM/GEMV inputs still fall back to the legacy/CBLAS path; the new kernels are row-major only and nntrainer dispatches row-major in practice.
- Block-size auto-tuning covers common Intel/AMD microarchitectures; unrecognized CPUs use conservative defaults.

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -19,13 +19,13 @@
 #include <cmath>
 #include <cstdint>
 #include <cstring>
+#include <fallback_internal.h>
 #include <fp16.h>
 #include <immintrin.h>
 #include <limits>
-#include <type_traits>
-#include <fallback_internal.h>
 #include <nntrainer_error.h>
 #include <thread_manager.h>
+#include <type_traits>
 #include <util_func.h>
 #include <vector>
 
@@ -731,8 +731,8 @@ void swiglu(const unsigned int N, float *X, const float *Y, const float *Z,
 }
 
 // poly_gelu_tanh_avx2 / poly_gelu_erf_avx2 are defined in avx2_internal.h
-using nntrainer::avx2::internal::poly_gelu_tanh_avx2;
 using nntrainer::avx2::internal::poly_gelu_erf_avx2;
+using nntrainer::avx2::internal::poly_gelu_tanh_avx2;
 
 // exp256_ps, hsum_avx, rcp_ps are also in avx2_internal.h
 using nntrainer::avx2::internal::exp256_ps;
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -328,6 +328,10 @@ avx2_approx_swiglu_alpha(__m256 x, __m256 s, __m256 alpha) noexcept -> __m256 {
 
 namespace nntrainer::avx2 {
 
+// Forward declarations for internal helpers used across the file
+static inline __m256 exp256_ps(__m256 x);
+static float hsum_avx(__m256 v);
+
 /**
  * @brief struct of q4_0x8 block
  */
@@ -701,16 +705,25 @@ static inline void convert_q4_0x8_noshuffle(const void *src,
 
 // ================== wrappers for your K,N combinations ==================
 // K = 3072 (UNIT = 768)
+/**
+ * @brief convert_q4_0x8_shuffle for K=3072 N=98304
+ */
 void convert_q4_0x8_shuffle_K3072_N98304(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = (N*8)/UNIT = 1024
   convert_q4_0x8_noshuffle<768, 1024>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=3072 N=36864
+ */
 void convert_q4_0x8_shuffle_K3072_N36864(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = 384
   convert_q4_0x8_noshuffle<768, 384>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=3072 N=3072
+ */
 void convert_q4_0x8_shuffle_K3072_N3072(const void *src, uint16_t *d_out,
                                         uint8_t *qs_out) {
   // groups = 32
@@ -718,16 +731,25 @@ void convert_q4_0x8_shuffle_K3072_N3072(const void *src, uint16_t *d_out,
 }
 
 // K = 8192 (UNIT = 2048)
+/**
+ * @brief convert_q4_0x8_shuffle for K=8192 N=98304
+ */
 void convert_q4_0x8_shuffle_K8192_N98304(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = 384
   convert_q4_0x8_noshuffle<2048, 384>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=8192 N=36864
+ */
 void convert_q4_0x8_shuffle_K8192_N36864(const void *src, uint16_t *d_out,
                                          uint8_t *qs_out) {
   // groups = 144
   convert_q4_0x8_noshuffle<2048, 144>(src, d_out, qs_out);
 }
+/**
+ * @brief convert_q4_0x8_shuffle for K=8192 N=3072
+ */
 void convert_q4_0x8_shuffle_K8192_N3072(const void *src, uint16_t *d_out,
                                         uint8_t *qs_out) {
   // groups = 12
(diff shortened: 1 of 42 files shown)


jijoongmoon (approved):
LGTM

test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp
```diff
@@ -174,16 +188,14 @@ void run_quant_test_fp16(const uint32_t M, const uint32_t K, const uint32_t N,
 
   // GROUND TRUTH TRANSB SGEMM for reference
   auto t1 = high_resolution_clock::now();
-  for (int tc = 0; tc < 20; ++tc) {
-    nntrainer::sgemm(0, false, true, M, N, K, 1.F, activation.data(), K,
-                     weight_fp16.data(), K, 0.F, ref_dst.data(), N);
-  }
```
jaemini-shin, line 180: Please do not try to modify existing test codes.

test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp
```diff
@@ -9,15 +9,29 @@
  */
 
 #include "int4_utils.h"
+#ifdef __ANDROID__
+// KleidiAI interface is only wired in via Android.mk; on Linux/meson the
+// header (and its qsi8d32p/qsi4c32p backend) is unavailable.
```
jaemini-shin: This one can be an candidate, but this approach would mess up the entire code:
* This file is being built for android only.

How about implementing separate test file such as `unittest_nntrainer_cpu_banckend_fp16_x86.cpp`?
You just need to add this new file into `meson.build`, not for `Android.mk`.

jaemini-shin (changes requested):
Could you please check following comments?
Also, try using `clang-format-14` to pass static format check.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm.h
```diff
@@ -0,0 +1,103 @@
...
+ * @bug    No known bugs except for NYI items
+ * @brief  Entry point for the x86 cache-blocked FP16 GEMM
+ */
+
+#ifndef __X86_HGEMM_H_
+#define __X86_HGEMM_H_
+
+#include <tensor_dim.h>
+
+namespace nntrainer::avx2 {
```
jaemini-shin: Could you please different namespace such as only `nntrainer`? This name space is for `avx2_impl.h`, using same namespace on different file makes it hard to maintain.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_util.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+const HgemmBlockSizes &get_hgemm_block_sizes() {
+  static const HgemmBlockSizes sizes = tune_block_sizes();
+  return sizes;
+}
+
+float *aligned_alloc_f32(std::size_t n_floats) {
+  void *p = nullptr;
+  std::size_t bytes = ((n_floats * sizeof(float) + 63u) / 64u) * 64u;
+  if (posix_memalign(&p, 64, bytes) != 0) {
```
jaemini-shin: x86 codes should be compatible to non posix systems such as windows.
How about using `std::aligned_alloc`, `std::free` for non windows system and `_algined_malloc`, `_aligned_free` for windows?

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_common.h
```diff
@@ -0,0 +1,54 @@
...
+/// Default C-panel tile sizes used inside the outer blocking panels. These are
+/// intentionally smaller than the outer M/N blocks so the FP32 accumulator
+/// panel stays cache-resident while K blocks are accumulated.
+#define X86_HGEMM_C_M_BLOCKING 192
+#define X86_HGEMM_C_N_BLOCKING 128
+
+/// Micro-kernel tile dimensions.
+/// 6 x 16 = 12 YMM accumulators + 2 YMM B + 1 YMM A broadcast = 15 of 16 YMMs.
+#define X86_HGEMM_MR 6
+#define X86_HGEMM_NR 16
```
jaemini-shin: Could you explain this number in detail?

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_workspace.cpp
```diff
@@ -0,0 +1,117 @@
...
+  float *next = aligned_alloc_f32(required);
+  aligned_free(buffer);
+  buffer = next;
+  capacity = required;
+  ++realloc_count;
+  return buffer;
+}
+
+HgemmWorkspace &get_hgemm_workspace() {
+  thread_local HgemmWorkspace workspace;
```
jaemini-shin, line 168: `thread_local` variable ends when the current thread terminates. Since we are using threadpool, once initialized workspace doesn't end. This could look like a memory leak from outside.
Please ensure releasing buffer after hgemm is finished for each thread.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_workspace.cpp
```diff
@@ -0,0 +1,117 @@
...
+                                     std::size_t required) {
+  if (required <= capacity) {
+    return buffer;
+  }
+
+  float *next = aligned_alloc_f32(required);
+  aligned_free(buffer);
+  buffer = next;
+  capacity = required;
+  ++realloc_count;
```
jaemini-shin: Are these counters necessary for non `ENABLE_TEST` case? It'd be better not to count for actual use

nntrainer/tensor/cpu_backend/x86/x86_compute_backend_fp16.cpp
```diff
@@ -150,68 +167,54 @@ void scopy(const unsigned int N, const _FP16 *X, const unsigned int incX,
...
-  delete[] A_;
-  delete[] B_;
-  delete[] C_;
-#else
-  __fallback_sgemm(TStorageOrder, TransA, TransB, M, N, K, alpha, A, lda, B,
-                   ldb, beta, C, ldc);
-#endif
+  if (TStorageOrder) {
+    __fallback_sgemm(TStorageOrder, TransA, TransB, M, N, K, alpha, A, lda, B,
+                     ldb, beta, C, ldc);
```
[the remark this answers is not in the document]
YongHyeon02: Restored the `FP32 conversion + __cblas_sgemm` path for column-major cases.

test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp
YongHyeon02: I've reverted `test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp` to its original state.

test/unittest/unittest_nntrainer_cpu_backend_fp16.cpp
YongHyeon02: I've moved all x86-specific tests added by this PR to the new `unittest_nntrainer_cpu_backend_fp16_x86.cpp` instead.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm.h
YongHyeon02: I've moved the `x86/hgemm/` tree from `nntrainer::avx2` to a dedicated `nntrainer::hgemm` namespace.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_util.cpp
YongHyeon02: I've branched allocation to use `_aligned_malloc` / `_aligned_free` on Windows and `std::aligned_alloc` / `std::free` elsewhere.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_workspace.cpp
YongHyeon02: I've registered workspaces in a process-wide registry; added `release_all_hgemm_workspaces()` and an `std::atexit` hook for deterministic cleanup at exit.

nntrainer/tensor/cpu_backend/x86/hgemm/hgemm_workspace.cpp
YongHyeon02: I've wrapped all realloc counter code in `#iifdef ENABLE_TEST` so it is compiled out in production build. Thank you for your kind review! :)

jaemini-shin:
Looks like you do not have a permission to run CI. Please tag me to run CI in next PR!

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -13,318 +13,37 @@
  */
 
 #include "avx2_impl.h"
+#include "avx2_internal.h"
 #include <array>
-#if __has_include(<bit>)
-#include <bit>
-#endif
```
jaemini-shin, line 19: I'm not sure removing these headers are safe or not...

YongHyeon02: The `<bit>` include moved to `avx2_internal.h` along with all the code that uses it (`compat_bit_cast`, `constexpr_popcount`, `Exp2Table` exp2 helper).

jaemini-shin (approved):
LGTM, nice work! 👍

baek2sm (approved):
Nice work, thank you. LGTM.

PR #3975, opened 2026-06-10
[meson] Update compile configurations

jaemini-shin:
This commit updates C/C++ compilation configurations.
* update C standard from gnu89 to gnu17
* trim warning options

**Self evaluation:**

---

nntrainer has used an old C standard `gnu89`. Let's update it to the default standard of ubuntu 22.04.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -8,7 +8,7 @@ title: Getting Started
 
 The following dependencies are needed to compile/build/run.
 
-* gcc/g++ >= 7 ( std=c++17 is used )
+* gcc/g++ >= 8 ( std=gnu17, std=c++17 is used )
    - (note)  >= 13 is recommended to enable fp16 support
 * meson >= 0.55.0
 * libopenblas-dev and base
@@ -29,6 +29,8 @@ sudo apt install nntrainer
 ```
 
 ## Clean build with pdebuild (Ubuntu 18.04)
+> [!WARNING]  
+> Ubuntu 18.04 is deprecated! Information below may be outdated.
 
 Use the NNStreamer PPA to resolve additional build-dependencies (Tensorflow/Tensorflow-Lite).
 
--- a/meson.build
+++ b/meson.build
@@ -5,7 +5,7 @@ project('nntrainer', 'c', 'cpp',
   default_options: [
     'werror=true',
     'warning_level=1',
-    'c_std=gnu89',
+    'c_std=gnu17',
     'cpp_std=c++17',
     'buildtype=release'
   ]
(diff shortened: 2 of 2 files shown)


meson.build
```diff
@@ -5,7 +5,7 @@ project('nntrainer', 'c', 'cpp',
   default_options: [
     'werror=true',
     'warning_level=1',
-    'c_std=gnu89',
+    'c_std=gnu17',
```
jaemini-shin, line 8: `gnu17` is the default standard of GCC 11, which is the default version in ubuntu 22.04.

docs/getting-started.md
```diff
@@ -8,7 +8,7 @@ title: Getting Started
 
 The following dependencies are needed to compile/build/run.
 
-* gcc/g++ >= 7 ( std=c++17 is used )
+* gcc/g++ >= 8 ( std=gnu17, std=c++17 is used )
```
jaemini-shin, line 11: `gnu17` is available from gcc 8+

meson.build
```diff
@@ -75,7 +75,6 @@ warning_flags = [
   '-Wno-error=varargs',
   '-Wno-error=ignored-attributes',
   '-Wno-error=sign-compare',
-  '-Wdefaulted-function-deleted',
```
jaemini-shin, line 78: This option is for C++ only.

meson.build
```diff
@@ -92,15 +91,16 @@ else
...
+warning_cxx_flags = [
+  '-Wdefaulted-function-deleted'
+]
+
 warning_c_flags = [
   '-Wmissing-declarations',
   '-Wmissing-include-dirs',
   '-Wmissing-prototypes',
   '-Wnested-externs',
-  '-Waggregate-return',
```
jaemini-shin, line 100: This option blocks RVO(return value optimization).

meson.build
```diff
@@ -92,15 +91,16 @@ else
...
-  '-Wold-style-definition',
-  '-Wdeclaration-after-statement',
```
jaemini-shin, line 102: This is the main reason for updating C standard. It has been off since `C99`.
```C
// before C99 not allowed
for(int i=0,...)
```

meson.build
```diff
@@ -92,15 +91,16 @@ else
...
-  '-Wno-error=varargs'
```
jaemini-shin, line 103: It's already included in above `warning_flags`.

haehun (approved):
Can we use lambdas or std::variants freely now?

Jungwon-Lee (approved):
LGTM

jaemini-shin:
@haehun
> Can we use lambdas or std::variants freely now?

It updates the C standard, not C++ standard. C++ standard is still C++17.
AFAIK, lambda and std::variants has already supported in C++17.

jayden0701 (approved):
LGTM!

PR #3977, opened 2026-06-11
[ Applications ] Enable FP16 Activation & Flash Attention

jijoongmoon:
## Enable FP 16 Activation and VJEPA Support

This PR enable FP16 Activation and VJEPA Support.
Also, Flash Attention is enabled.


dlwlzzero:
Are you still working on this? If not, it might be better to close it to keep the PR list clean.

PR #3978, opened 2026-06-11
[poc] quant and run kai on android device

jaemini-shin:
1. build and install android
2. prepare fp32 weight binary on device
3. quantize and run on device


dlwlzzero:
Will you check this one is also in progress?

PR #3979, opened 2026-06-15
[test] CausalLM tiny-model unit tests and HF differential reference fixtures

Jungwon-Lee:
## Summary

Adds unit-test coverage for all `Applications/CausalLM` model classes — both text-generation and embedding/encoder — using a **differential reference fixture** methodology: a tiny model is built in Python (HuggingFace `transformers`), golden outputs are saved to JSON, and the C++ model is validated against them at test time.

---

### Test framework: differential reference fixtures

Each fixture directory under `test/unittest/models/causallm_reference/` contains:

| File | Description |
|---|---|
| `config.json` | HuggingFace-compatible model config |
| `nntr_config.json` | nntrainer runtime config (dtype, seq-len, model_type, …) |
| `generation_config.json` / `modules.json` | Generation or SentenceTransformer module config |
| `tokenizer.json` | Minimal word-level tokenizer for the tiny vocab |
| `*.bin` | FP32 weights in nntrainer layout (excluded from git, shipped as `causallm_reference.tar.gz`) |
| `meta.json` | Per-fixture tolerances (`embedding_atol`, `cosine_min`, `*_q40` variants) and generator provenance |
| `reference_tokens.json` / `reference_embedding.json` | Golden outputs from HuggingFace |

Generator scripts live in `test/unittest/models/causallm_reference/generators/`. Each script builds a tiny model, runs HuggingFace, and writes the complete fixture including binary weights.

Tests skip automatically (`GTEST_SKIP`) when the fixture directory is absent, so CI without the binary tarballs stays green.

---

### Text-generation models

New differential tests comparing greedy-decode output and prefill logits against HuggingFace:

| Fixture | Class | Notes |
|---|---|---|
| `qwen2_tiny` | `Qwen2ForCausalLM` | |
| `qwen3_tiny` | `Qwen3ForCausalLM` | |
| `gemma3_tiny` | `Gemma3ForCausalLM` | |
| `gemma4_tiny` | `Gemma4ForCausalLM` | |
| `qwen3_moe_tiny` | `Qwen3MoeForCausalLM` | |
| `qwen3_slim_moe_tiny` / `qwen3_cached_slim_moe_tiny` | `Qwen3MoeForCausalLM` (cached) | |
| `gptoss_tiny` / `gptoss_cached_slim_tiny` | `GptOssForCausalLM` (cached) | |

Q4_0 quantized generation tests (`runQ40DifferentialChecks`) are gated on the `NNTR_QUANTIZE_BIN` environment variable and skip automatically when unset.

---

### Embedding/encoder models

New embedding differential framework added to `causallm_test_utils`: `EmbeddingTestAdapter<T>`, `runFp32EmbeddingDifferentialChecks`, `runQ40EmbeddingDifferentialChecks`, `expectEmbeddingNear` (per-element atol + cosine similarity). Tests drive each model through its own `encode(prompt)` path, absorbing per-model input-format differences (single input, 3-input BERT, etc.).

| Fixture | Class | HF architecture | Pooling |
|---|---|---|---|
| `qwen3_embedding_tiny` | `Qwen3Embedding` | `Qwen3Embedding` | last-token |
| `qwen2_embedding_tiny` | `Qwen2Embedding` | `Qwen2Model` | last-token |
| `kalm_embedding_tiny` | `Qwen2Embedding` | `Qwen2Model` (KaLM config) | mean |
| `embedding_gemma_tiny` | `EmbeddingGemma` | `Gemma3ForCausalLM` | mean + L2 |
| `tinybert_tiny` | `MultilingualTinyBert` | `BertForMaskedLM` | raw hidden state (no pooling), Linux-only |
| `deberta_v2_tiny` | `DebertaV2` | `DebertaV2ForMaskedLM` | mean + L2 |

Q4_0 quantized embedding tests added for all 6 models, also gated on `NNTR_QUANTIZE_BIN`.

---

### Bug fixes (from #3982)

Two encoder correctness fixes found while validating embedding models against HuggingFace, also tracked in [#3982](https://github.com/nntrainer/nntrainer/pull/3982):

- **BERT RoPE**: `BertTransformer::createAttention` defaulted `use_rope=true` with `rope_theta=0`, producing `nan` for all encoder outputs. Fixed by passing `use_rope=false`.
- **DeBERTa V2 p2c index**: the p2c gather index had the wrong sign (missing the HuggingFace post-gather transpose), corrupting every off-diagonal attention score. Corrected sign matches HuggingFace output (cosine > 0.9999).

### Android verification — bug fixes

Two bugs found while running the test suite on Android (arm64-v8a, FP16 enabled):

- **LayerNorm FP16 overflow** (`nntrainer/layers/layer_normalization_layer.cpp`): The `ENABLE_FP16` optimized path called `rms_norm_wrt_width_fp16_intrinsic(float*, ...)`, which accumulates squared values in FP16 and is intentionally deprecated on ARM (throws at runtime). DeBERTa-v2's LayerNorm inputs exceed FP16 range (~65504), causing overflow. Fixed by switching to `rms_norm_wrt_width_fp32_intrinsic` — the same fix already applied to the CausalLM `rms_norm.cpp` layers.
- **Tokenizer absolute path in `nntr_quantize`** (`Applications/CausalLM/quantize.cpp`): Fixture `nntr_config.json` files store `tokenizer_file` as an absolute host path. `nntr_quantize` runs as a subprocess and cannot reach that path on a different machine. Fixed by adding a fallback: if the absolute path does not exist, resolve by filename relative to the model directory.

---

### Gemma4 differential fixture fixes

The `gemma4_tiny` fixture previously diverged from HuggingFace (max prefill logit diff ~0.75, wrong argmax). Root-cause analysis (layer-by-layer activation bisection vs HF) showed the model code was correct; the problems were in the fixture generator and build plumbing. Now fixed — max FP32 logit diff **~2e-4**, FP32 + Q4_0 both pass:

- **Weight save order**: `generate_gemma4_reference.py` wrote weights in an order that did not match nntrainer's weight-load order. Reordered to emit `q_norm` right after `wq`, `k_norm` right after `wk`, and the global per-layer-input weights inside layer 0 (after `per_layer_input_gate`) — matching the production `res/gemma4/weight_converter.py`. (0.75 → 0.011)
- **Per-attention-type RoPE**: the fixture config omitted `rope_parameters`. Gemma4 uses `theta=10000` for sliding layers and `theta=1e6` + proportional `partial_rotary_factor=0.25` for full layers. Added it to the generated config. (0.011 → 2e-4)
- **Q4_0 argmax stability**: the tiny model's top-2 logits were nearly tied, so Q4_0 quantization noise reordered the argmax. Changed the generator default seed to 18 so the top logits are far enough apart.
- **Fixture extraction path**: `Applications/CausalLM/meson.build` extracted the fixture tarballs (entries prefixed `causallm_reference/`) into `models/causallm_reference/`, doubling the path so the bins never landed where the tests look. Because the bins are git-ignored and only shipped via the tarballs, **every** causallm differential test was silently skipping in CI. Now extracts into `models/`.

### `nntr_quantize` extension

- Registers `MultilingualTinyBert` and `DebertaV2` factory models so Q4_0 quantization works for encoder models.
- Adds BERT (`ffn_fc1`, `ffn_down`) and DeBERTa (`intermediate`, `output_dense`) layer names to the FC-layer dtype map.
- Resolves `module_config_path` and `tokenizer_file` relative paths to absolute before model creation.

---

## Test plan

```bash
# 1. Unpack fixtures
tar -xf causallm_reference.tar.gz -C test/unittest/models/causallm_reference/

# 2. Build
ninja -C build Applications/CausalLM/unittest_causallm_models

# 3. FP32 differential tests (no env vars needed)
./build/Applications/CausalLM/unittest_causallm_models

# 4. Q4_0 quantized tests
NNTR_QUANTIZE_BIN=./build/Applications/CausalLM/nntr_quantize \
  ./build/Applications/CausalLM/unittest_causallm_models

# 5. Android (arm64-v8a) — see Applications/CausalLM/TESTING.md
./Applications/CausalLM/run_unittest_android.sh          # full suite
./Applications/CausalLM/run_unittest_android.sh --cache  # skip rebuild
```

- Tests without fixtures → `GTEST_SKIP` (CI-safe)
- Tests without `NNTR_QUANTIZE_BIN` → Q4_0 tests skip automatically

---

### x86 (Ubuntu host, FP16 disabled)

Run with `NNTR_QUANTIZE_BIN` set (21 differential tests, fixtures present):

| Test suite | FP32 | Q4_0 |
|---|---|---|
| `Gemma3DifferentialTest` | ✅ PASS | ✅ PASS |
| `Gemma4DifferentialTest` | ✅ PASS | ✅ PASS |
| `Qwen3MoeDifferentialTest` | ✅ PASS | — |
| `Qwen2DifferentialTest` | ✅ PASS | ✅ PASS |
| `Qwen3DifferentialTest` | ✅ PASS | ✅ PASS |
| `Qwen3EmbeddingDifferentialTest` | ✅ PASS | ✅ PASS |
| `Qwen2EmbeddingDifferentialTest` | ✅ PASS | ✅ PASS |
| `KalmEmbeddingDifferentialTest` | ✅ PASS | ✅ PASS |
| `EmbeddingGemmaDifferentialTest` | ✅ PASS | ✅ PASS |
| `TinyBertDifferentialTest` | ✅ PASS | ✅ PASS |
| `DebertaV2DifferentialTest` | ✅ PASS | ✅ PASS |

**21 passed, 0 failed** — the previously-failing Gemma4 differential tests are fixed (see *Gemma4 differential fixture fixes* above).

### Android (arm64-v8a, FP16 enabled, NDK r26d, API 29)

| Result | Count |
|--------|-------|
| PASSED | 55 |
| SKIPPED | 0 |
| FAILED | 0 |

All 55 tests pass including FP32 differential, Q4_0 differential, and tiny-model deterministic tests.


haehun (approved):
So now we can test and compare logits with reference python models.

Seunghui98:
Could you rebase upstream/main ?

test/unittest/models/causallm_reference/deberta_v2_tiny/nntr_config.json
```diff
@@ -0,0 +1,15 @@
...
+  "fc_layer_dtype": "FP32",
+  "init_seq_len": 4,
+  "lmhead_dtype": "FP32",
+  "max_seq_len": 16,
+  "model_file_name": "nntr_deberta_v2_tiny_fp32.bin",
+  "model_tensor_type": "FP32-FP32",
+  "model_type": "Embedding",
+  "module_config_path": "modules.json",
+  "num_to_generate": 1,
+  "tokenizer_file": "/home/jwon/Desktop/workspace/testcode/nntrainer/test/unittest/models/causallm_reference/deberta_v2_tiny/tokenizer.json"
```
Seunghui98: I think you should change to a relative path, not your local path.

test/unittest/models/causallm_reference/embedding_gemma_tiny/nntr_config.json
```diff
@@ -0,0 +1,15 @@
...
+  "fc_layer_dtype": "FP32",
+  "init_seq_len": 4,
+  "lmhead_dtype": "FP32",
+  "max_seq_len": 8,
+  "model_file_name": "nntr_embedding_gemma_tiny_fp32.bin",
+  "model_tensor_type": "FP32-FP32",
+  "model_type": "Embedding",
+  "module_config_path": "modules.json",
+  "num_to_generate": 1,
+  "tokenizer_file": "/home/jwon/Desktop/workspace/testcode/nntrainer/test/unittest/models/causallm_reference/embedding_gemma_tiny/tokenizer.json"
```
Seunghui98: ditto

test/unittest/models/causallm_reference/deberta_v2_tiny/nntr_config.json
Jungwon-Lee: Thank you for your comment. 
I missed the rename the file path. i updated now

test/unittest/models/causallm_reference/generators/generate_deberta_v2_reference.py
```diff
@@ -0,0 +1,332 @@
...
+POOLING_CONFIG = {
+    "word_embedding_dimension": HIDDEN,
+    "pooling_mode_cls_token": False,
+    "pooling_mode_mean_tokens": True,
+    "pooling_mode_max_tokens": False,
+    "pooling_mode_mean_sqrt_len_tokens": False,
+    "pooling_mode_weightedmean_tokens": False,
+    "pooling_mode_lasttoken": False,
+    "include_prompt": True,
+}
```
Seunghui98, line 96: it's just out of curiousity.
I guess we don't need to have **config.json files for model because we use hardcorded constant values for config when we test this. 
so, Why don't we remove **config.json files under model_tiny path ?

Seunghui98: @Jungwon-Lee Could you check this comment?

Jungwon-Lee: `generate_*_reference.py` generates both binary weight files and JSON config files (`config.json, generation_config.json, nntr_config.json`) as pre-built fixtures. These fixtures are committed to the repo so tests can run without any Python dependencies — torch, numpy, and transformers (including model-specific packages like `transformers.models.gemma4)` are only needed when regenerating fixtures.

 The config values are hardcoded in the generator scripts, but the C++ test utilizes them from the JSON files at runtime. So even though the values come from hardcoded constants, they still need to be present on disk as JSON for the test binary to read.

jayden0701 (approved):
LGTM!

test/unittest/models/causallm_reference/generators/generate_deberta_v2_reference.py
Seunghui98: Thanks for letting me know
but... I don't know it, so I'm asking...
what is difference between the test `generate_*_reference.py` and the C++ test performed by using the config files of the model?

Seunghui98: Are the python tests in `generate_*_reference.py` for golden test?

Jungwon-Lee: > Are the python tests in `generate_*_reference.py` for golden test?

Yes,  `generate_*_reference.py` generates model fixture and logits.
The logits are used for golden test as a ground truth.

jaemini-shin (approved):
LGTM!

PR #3980, opened 2026-06-15
[neuralnet] enable network sharing

lhs8928:
[neuralnet] enable network sharing
    
     - Modified model compile and initialize to take a model reference pointer as a parameter for network sharing
     - Added test case for network sharing

    **Self evaluation:**
    
### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -124,8 +124,10 @@ NeuralNetwork::NeuralNetwork() :
   initialized(false),
   compiled(false),
   loadedFromConfig(false),
+  loadedWeight(false),
   exec_mode(ExecutionMode::TRAIN),
-  ct_engine(&Engine::Global()) {}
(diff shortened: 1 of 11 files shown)


nntrainer/models/neuralnet.cpp
```diff
@@ -187,7 +191,8 @@ void NeuralNetwork::setTrainConfig(const std::vector<std::string> &values) {
     << " of first element: " << left_props.front();
 }
 
-int NeuralNetwork::compile(ExecutionMode mode) {
+int NeuralNetwork::compile(ExecutionMode mode,
+                           std::shared_ptr<ml::train::Model> ref_model) {
```
jijoongmoon, line 198: Is there any reason to change this signature? I think there are missing code change here.

lhs8928: This is intentional because the Transformer is already implemented to invoke initialize() internally inside compile().

PR #3986, opened 2026-06-16
[kv_cache_manager] reduce peak memory

jaemini-shin:
This commit removes redundant initialization which results peak memory explosion.

**Self evaluation:**

---
With this PR we can reduce peak mem from 3.3 GB to 1.4GB for qwen3-0.6B model.

### DIFF
--- a/Applications/CausalLM/kv_cache_manager.cpp
+++ b/Applications/CausalLM/kv_cache_manager.cpp
@@ -68,8 +68,6 @@ void KVCacheManager::allocate(unsigned int num_layers, unsigned int batch_size,
                                    {format, dtype});
     layer_caches_[i].key_cache = nntrainer::Tensor(cache_dim, true);
     layer_caches_[i].value_cache = nntrainer::Tensor(cache_dim, true);
-    layer_caches_[i].key_cache.setZero();
-    layer_caches_[i].value_cache.setZero();
   }
 }
 


Applications/CausalLM/kv_cache_manager.cpp
```diff
@@ -68,8 +68,6 @@ void KVCacheManager::allocate(unsigned int num_layers, unsigned int batch_size,
                                    {format, dtype});
     layer_caches_[i].key_cache = nntrainer::Tensor(cache_dim, true);
     layer_caches_[i].value_cache = nntrainer::Tensor(cache_dim, true);
-    layer_caches_[i].key_cache.setZero();
-    layer_caches_[i].value_cache.setZero();
```
jaemini-shin, line 72: These lines are redundant because it will be overwritten by `k_proj`, `q_proj`.
With these lines, it results peak memory explosion of `MAX_SEQ_LEN` for each KV cache and layer.

jijoongmoon (approved):
LGTM

Jungwon-Lee (approved):
Great Work!!

Seunghui98:
### Review: reduce peak memory by dropping `setZero()`

**Summary:** Removing the two post-allocation `setZero()` calls cuts peak memory 3.3GB → 1.4GB on qwen3-0.6B. Correct and safe to merge.

**Why it works:** `Tensor(cache_dim, true)` reserves the buffer, but on Linux physical pages aren't committed until touched. `setZero()` writes every byte, forcing *all* layers' KV cache pages (full `max_seq_len`) to commit at init time — that's the peak spike. Dropping it lets pages commit lazily as generation writes them.

**Why it's safe (redundancy confirmed):** Two facts make the zero-init unnecessary:
- The used region `[0, cache_pos)` is always overwritten by the K/V projections (write views) **before it is read**.
- The unused tail `[cache_pos, max_seq_len)` is **never read** — `get{Key,Value}CacheReadView` bound the range to `read_len` (= `cache_pos + step`), and `save()` only writes `seq_len`.

So nothing ever reads uninitialized memory. ✔️

**Minor suggestion:** add a one-line comment so this isn't "fixed" back later:
```cpp
// Intentionally not zero-initialized: reads are bounded to [0, cache_pos)
```

Seunghui98 (approved):
LGTM 👍

PR #3988, opened 2026-06-17
[Wait 4 #3967] [x86] feat: Enable FP16 attention (kcache/vcache/softmax) on x86 CPU Backend

YongHyeon02:
## Dependency of the PR

- Depends on: #3967
- Benchmark tooling: #3820

## Summary

This PR (1) removes all per-call heap allocations from the existing **FP32-input** x86 attention kernels and (2) adds the missing **FP16-input** x86 attention kernels, so FP16-mode LLM (CausalLM) inference runs on x86 — it previously did not, because these kernels existed only on ARM (`neon_impl_fp16.cpp`).

Key properties:

1. **Zero per-call allocations** in the FP32-input kernels. `compute_kcaches`, `compute_fp16vcache_fp32_transposed`, and `softmax_row(_with_sink)_inplace` replaced their per-call `std::vector` / `new[]` / `new __m256[]` scratch with reusable `static thread_local` buffers (one allocation per worker thread, amortised to 0). Measured: **1 / 2 / 2 → 0** allocations per call (and a per-head `sumRem` site, **10 → 0** for `head_dim % 8 != 0`).
2. **Conversion fused into the dot product.** `compute_kcaches` widens the FP16 key row to FP32 8 lanes at a time inside the FMA loop (F16C), instead of staging the whole row in a `tmp_fp32` buffer first — ~1.6–1.8× faster.
3. **Hot-loop scratch pointers cached into locals.** The thread_local buffers' `.data()` pointers are resolved once per call so the inner FMA loops do plain pointer arithmetic (no per-access TLS/vector indirection) — required to keep the vcache kernel competitive (see commit 3 / Performance).
4. **New FP16-input AVX2+F16C kernels** under `nntrainer::avx2`: `compute_kcaches(const _FP16*)`, `compute_fp16vcache_transposed(const _FP16*)`, and `softmax_row_inplace(_FP16*)` (FP16 and FP32 sink overloads). x86 has no native FP16 arithmetic, so all math accumulates in **FP32** (F16C widen on load, narrow on store) — functionally matching the ARM kernels, more accurate than FP16 accumulation.
5. **Same 3-layer dispatch as ARM** (`x86_compute_backend.h` decls → thin wrappers in `x86_compute_backend_fp16.cpp` → `avx2::` impls), gated by plain `#ifdef ENABLE_FP16`; the shared `cpu_backend.h` is untouched.

<details><summary>feat: add x86 FP16-input attention kernels and remove kernel allocations (1afe91e0)</summary><br />

- Remove per-call allocations from the FP32-input kernels in `avx2_impl.cpp` (fold FP16→FP32 widening into `compute_kcaches`' dot loop; reusable `static thread_local` scratch for the vcache + softmax kernels; drop the non-RAII `new __m256[]`)
- Add FP16-input AVX2+F16C kernels in `avx2_impl_fp16.cpp` accumulating in FP32
- Public `nntrainer::` overloads in `x86_compute_backend.h`, delegating wrappers in `x86_compute_backend_fp16.cpp`, decls in `avx2_impl.h` (mirrors ARM layering)
- x86 attention unit tests vs an in-test FP32 software reference (5e-3 tol)

**Self evaluation:** 1. Build [X]Passed  2. Run [X]Passed

</details>

<details><summary>fix: harden x86 attention remainder and scale paths (8ab7f7a7)</summary><br />

- Guard the `rem == 0` remainder handling in the vcache kernels against empty scratch; precompute `1 / sqrt(head_dim)` once for the QK kernels
- Add `rem == 0` local-window / head-slice tests for the FP16 QK, FP16 vcache, and FP32-input vcache paths

**Self evaluation:** 1. Build [X]Passed  2. Run [X]Passed

Signed-off-by: KWSMooBang <40kimwoosung@gmail.com>
</details>

<details><summary>fix: cache x86 attention vcache scratch pointers to avoid TLS indirection (e21b39ed)</summary><br />

- The compiler did not hoist the `static thread_local std::vector` data pointers out of the vcache hot loop, so the inner FMA loop paid per-access TLS/vector indirection — regressing `compute_fp16vcache_fp32_transposed` by ~20% for `gqa=1` (MHA) while `gqa>1` improved
- Resolve each thread_local `.data()` pointer once into a local; hot loops then do plain pointer arithmetic (matching the pre-rewrite local-buffer access), 0 per-call allocations preserved. Applied to the FP16-input vcache kernel too

**Self evaluation:** 1. Build [X]Passed  2. Run [X]Passed

</details>

## Performance Improvements

Intel Core i5-12400F (6C/12T, AVX2+F16C), WSL2, GCC 12.3, `enable-fp16=true`, single-thread. One iteration = one full **causal** attention forward for a token block of length `seq_len` (`num_cache_head = 8`). Baseline = the pre-PR kernels on
#3967's head (`f26a5d3e`).

> Latency before/after is measured by **same-session interleaved A/B** (one bench binary, `LD_LIBRARY_PATH`-swapped `libnntrainer.so`, min-of-N): cross-session runs on this box drift up to ~2× and are not comparable.

### Allocation elimination (measured, per call)

| Kernel | before | after |
|:--|--:|--:|
| `compute_kcaches` (head_dim=128) | 1 | **0** |
| `compute_fp16vcache` (head_dim=128, rem=0) | 2 | **0** |
| `compute_fp16vcache` (head_dim=130, rem=2, 8 heads) | 10 | **0** |
| `softmax_row_inplace` (num_heads=64) | 2 | **0** |

### FP32-input `compute_kcaches` (Q·Kᵀ) — before → after

| seq_len, gqa, head_dim | speedup |
|:--|--:|
| 128, 1, 64 | **1.83×** |
| 512, 1, 64 | **1.84×** |
| 512, 4, 128 | **1.84×** |
| 1024, 1, 128 | **1.59×** |
| 1024, 4, 128 | **1.70×** |

`softmax_row_inplace` (FP32): parity (alloc-only change).

### FP32-input `compute_fp16vcache` (softmax·V) — before → after (final)

| seq_len, gqa, head_dim | speedup |
|:--|--:|
| 512, 1, 64 (MHA) | **1.19×** |
| 1024, 1, 128 (MHA) | ~parity |
| 512, 4, 128 (GQA) | **1.22×** |
| 1024, 4, 128 (GQA) | **1.16×** |

The initial commits regressed `gqa=1` to ~0.8×; commit `e21b39ed` (scratch-pointer caching) recovers it to parity-or-better and is 1.1–1.4× faster than the unfixed version — see the benchmark file for the full A/B.

### FP16-input vs FP32-input on x86 (both P4 kernels, same binary)

- `compute_fp16vcache_transposed` (softmax·V): **1.7–2.5× faster** than FP32-input at `gqa=1` (memory-bound; halved weight+output I/O), ~1.05× at `gqa=4`.
- `compute_kcaches`: ~1.5× faster at small `seq_len`, ~0.82–0.87× for `seq_len ≥ 512` (x86 must F16C-widen both Q and K). `softmax_row`: ~parity.

### Highlights

- **FP16-mode LLM inference on x86: non-functional → working** (the headline).
- **Per-call allocations eliminated** for every FP32-input attention kernel.
- **`compute_kcaches` 1.6–1.8× faster**; **`compute_fp16vcache` parity-to-1.2×** (after the TLS-caching fix); **FP16-input vcache 1.7–2.5× faster** than FP32-input.

## Known Limitations

- **`sum / sqrt(head_dim)` → `sum * (1/sqrt(head_dim))`** is not bit-exact vs the pre-PR FP32 output (within the 1e-4 test tolerance); deliberate, applied to both QK kernels.
- x86 FP16-input kernels accumulate in FP32, so they are numerically close to but not bit-identical to the ARM FP16-accumulate kernels (compared at 5e-3).
- `compute_kcaches` FP16-input is slower than FP32-input for `seq_len ≥ 512` (no native FP16 FMA on x86 → both operands need F16C widening); it is kept for functionality + model-level FP16 KV-cache memory savings.

### DIFF
--- a/debian/nntrainer-dev.install
+++ b/debian/nntrainer-dev.install
@@ -33,6 +33,8 @@
 /usr/include/nntrainer/weight.h
 /usr/include/nntrainer/quantizer.h
 /usr/include/nntrainer/avx2_impl.h
+/usr/include/nntrainer/avx2_mathfun.h
+/usr/include/nntrainer/avx2_mathfun.hxx
 # todo: update dataset headers
 /usr/include/nntrainer/databuffer.h
 /usr/include/nntrainer/databuffer_factory.h
--- a/nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
+++ b/nntrainer/tensor/cpu_backend/fallback/fallback_internal.cpp
@@ -409,9 +409,18 @@ void __fallback_calc_trigonometric_vals_dup(unsigned int N_half, float *angle,
                                             float *cos_, float *sin_,
                                             unsigned int from,
                                             float attention_scaling) {
-  throw std::runtime_error(
-    "Error: No implementation of rotary embedding layer incremental_forwarding "
-    "with SIMD acceleration except for NEON!");
+  const float from_f = static_cast<float>(from);
+  for (unsigned int i = 0; i < N_half; ++i) {
+    const float a = from_f * angle[i];
+    cos_[i] = std::cos(a) * attention_scaling;
+    sin_[i] = std::sin(a) * attention_scaling;
+  }
+  const unsigned int N = 2 * N_half;
+  for (unsigned int j = N_half, j_half = 0; j < N && j_half < N_half;
+       ++j, ++j_half) {
+    cos_[j] = cos_[j_half];
+    sin_[j] = sin_[j_half];
+  }
 }
 
 void __fallback_swiglu(const unsigned int N, float *X, float *Y, float *Z) {
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -1687,10 +1687,12 @@ void compute_fp16vcache_fp32_transposed(int row_num, const float *in,
             accPtr, _mm256_fmadd_ps(inVec, bVec, _mm256_loadu_ps(accPtr)));
         }
 
-        float *remPtr = &sumRem.data()[h * rem];
-        int base = num_blocks * 8;
-        for (int r = 0; r < rem; ++r) {
-          remPtr[r] += a_val * tmp_fp32[base + r];
+        if (rem > 0) {
+          float *remPtr = &sumRem[(size_t)h * rem];
+          int base = num_blocks * 8;
+          for (int r = 0; r < rem; ++r) {
+            remPtr[r] += a_val * tmp_fp32[base + r];
+          }
         }
       }
     }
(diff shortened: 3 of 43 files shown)


YongHyeon02:
@jaemini-shin Could you please run the CI?

PR #3992, opened 2026-06-17
[tensor] reduce peak memory

jaemini-shin:

[tensor] reduce peak memory

This commit removes default initialization list for fp16 and uint8 tensor.
This logic should have included in PR #3986, but it didn't.

</details>

### Summary
This PR adds missing logic from #3986.
Now the peak memory reduces correctly.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/uint_tensor.cpp
+++ b/nntrainer/tensor/uint_tensor.cpp
@@ -155,7 +155,7 @@ template <typename T> void UIntTensor<T>::allocate() {
 
     mem_data = new MemoryData(
       (void *)(new T[dim.getDataLen() + (sizeof(float) + sizeof(unsigned int)) /
-                                          sizeof(T) * scale_size()]{}));
+                                          sizeof(T) * scale_size()]));
     data = std::shared_ptr<MemoryData>(mem_data, [](auto *mem_data) {
       delete[] mem_data->template getAddr<T>();
       delete mem_data;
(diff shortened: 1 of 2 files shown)


nntrainer/tensor/uint_tensor.cpp
```diff
@@ -155,7 +155,7 @@ template <typename T> void UIntTensor<T>::allocate() {
 
     mem_data = new MemoryData(
       (void *)(new T[dim.getDataLen() + (sizeof(float) + sizeof(unsigned int)) /
-                                          sizeof(T) * scale_size()]{}));
+                                          sizeof(T) * scale_size()]));
```
jaemini-shin, line 158: `uint_tensor` used for KV cache has additional data: scale and zeropoint.
These parameters used to set as 0 at this point and not changed at `initialze()`.

nntrainer/tensor/uint_tensor.cpp
```diff
@@ -297,12 +297,22 @@ template <typename T> void UIntTensor<T>::initialize() {
 
   /// @note Sampling from the normal/uniform distribution is invalid
   switch (initializer) {
-  case Initializer::ZEROS:
+  case Initializer::ZEROS: {
     setZero();
+    float *scale = (float *)getScale();
+    std::fill(scale, scale + scale_size(), 1.0f);
+    unsigned int *zerop = getZeroPoint();
+    std::fill(zerop, zerop + scale_size(), 0u);
```
jaemini-shin, line 305: So we need to fill values in order to pass comparison unittest.

Seunghui98 (commented):
### Summary
This PR completes the peak-memory reduction work started in #3986.

The core idea: large tensor buffers were allocated with value-initialization (`new T[...]{}`). Zeroing the whole buffer at allocation time forces the OS to commit (fault in) every physical page immediately, which inflates peak memory. Dropping `{}` lets the allocation stay demand-paged, so pages are only committed when actually written.

Because `UIntTensor`'s buffer also holds the `scale` / `zero_point` regions at its tail — which were previously zeroed implicitly by that `{}` — `initialize()` now sets them explicitly (`scale = 1.0`, `zero_point = 0`) for the `ZEROS` and `ONES` initializers.

**Where to look:** `UIntTensor::initialize()` in `uint_tensor.cpp` is the heart of the change — verify the scale/zero_point defaults match what the previous zero-init produced. The `half_tensor.cpp` change is just the `{}` removal (no scale/zero_point to handle).

### Notes / questions
- **`NONE` / `default` cases:** scale and zero_point are now left uninitialized in those paths. Is that always safe — i.e., are they guaranteed to be overwritten by a subsequent weight load before use?
- **Minor:** the scale/zero_point fill block is duplicated between `ZEROS` and `ONES`. Could be hoisted before the `switch` (it's identical) or factored into a small helper.

Otherwise looks correct and on point for the stated goal.

Seunghui98 (approved):
LGTM 👍

jijoongmoon (approved):
LGTM

PR #3993, opened 2026-06-18
[DDTree] Parallelize and speed up buildTree per-depth preprocessing

Seunghui98:
## Dependency of the PR

None.

<details><summary>[DDTree] Parallelize and speed up buildTree per-depth preprocessing</summary><br />

The per-depth top-k/logsumexp loop in buildTree was the dominant cost:
for each of depthLimit rows it allocated an idx[vocab] array, ran
std::partial_sort over the full vocab, computed a separate full-row max,
and then logsumexp, all single-threaded. For a realistic config
(depth 15, vocab 150272, budget 31) this takes ~13 ms per call.

Each depth row is independent, so distribute the rows across the
nntrainer ThreadManager via parallel_for. Within a row, replace the
idx[vocab] + partial_sort top-k with a single sequential pass that keeps
a bounded min-heap of size topk (cache-friendly, no per-row vocab-sized
allocation) and computes the running max in the same pass. The
logsumexp(logZ) still accumulates double std::exp in token-index order
with the same max, so topLogProbs/topTokenIds and the resulting tree are
byte-identical to the previous sequential version (golden parity
preserved). With 4 threads this is ~3x faster on a host benchmark
(~13 ms -> ~4 ms); the gain scales with the number of worker threads.

**Self evaluation:**

</details>

<details><summary>[DDTree/Test] Add sequential-reference equivalence test for buildTree</summary><br />

The embedded Python golden vectors only exercise small trees, so they do
not cover the large-vocab path or the parallel top-k selection. Add a
self-contained pre-optimization reference (idx[vocab] + partial_sort,
single-threaded) and a randomized test that asserts the optimized
buildTree reproduces it byte-for-byte (nodeTokenIds, nodeDepths, parents,
visibility, nodeCount, currentLength) across many random inputs. Cases
include the realistic depth 15 / vocab 150272 / budget 31 config plus
tie-prone small rows and budget-edge cases; replicas of the big case
guard against nondeterminism from the parallel path.

**Self evaluation:**

</details>

### Summary

- Parallelize buildTree's per-depth top-k/logsumexp preprocessing over the
  nntrainer ThreadManager and replace the idx[vocab] + std::partial_sort top-k
  with a bounded-heap single-pass selection. Output is byte-identical to the
  previous sequential version (golden parity preserved); ~3x faster with 4
  threads on a host benchmark.
- Add a sequential-reference equivalence unit test that checks the optimized
  buildTree matches the pre-optimization algorithm byte-for-byte on large-vocab
  random inputs. Verified locally with a host g++ build of unittest_ddtree
  (22/22 tests passed, including MatchesPythonGolden and the new
  MatchesSequentialReferenceRandom).

### DIFF
--- a/nntrainer/utils/ddtree/ddtree.cpp
+++ b/nntrainer/utils/ddtree/ddtree.cpp
@@ -13,7 +13,10 @@
 
 #include <algorithm>
 #include <cmath>
+#include <limits>
 #include <queue>
+
+#include <thread_manager.h>
 
 namespace nntrainer {
 namespace ddtree {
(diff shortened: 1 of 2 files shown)


haehun (approved):
This is very good speedup, I tested it too

nntrainer/utils/ddtree/ddtree.cpp
```diff
@@ -37,25 +40,74 @@ DDTreeStructure buildTree(const float *draftLogits, int depthLimit, int vocab,
...
+    const float *row = draftLogits + d * static_cast<size_t>(vocab);
+
+    // Single sequential pass top-k via a threshold + small array: the hot path
+    // is one well-predicted `x < thresh` compare that rejects the ~99.98% of
+    // tokens outside the top-k, so it avoids the per-element priority_queue
+    // push/pop and Cand churn. `thresh` is the worst kept logit once `buf` is
+    // full; ties (x == thresh) fall through to the full `better` comparison so
+    // the lower token index still wins. The selected set/order is identical to
+    // a partial_sort (logit/token is a total order), preserving golden parity.
+    std::vector<Cand> buf;
```
jaemini-shin, line 75: (optional) Locating `buf` outside of thread loop might get better performance. Allocating buffers inside the thread loop usually hurts its performance.

nntrainer/utils/ddtree/ddtree.cpp
```diff
@@ -64,10 +116,10 @@ DDTreeStructure buildTree(const float *draftLogits, int depthLimit, int vocab,
     topLogProbs[d].resize(topk);
     topTokenIds[d].resize(topk);
```
jaemini-shin, line 117: These lines could move outside of the loop.

nntrainer/utils/ddtree/ddtree.cpp
```diff
@@ -64,10 +116,10 @@ DDTreeStructure buildTree(const float *draftLogits, int depthLimit, int vocab,
...
     for (int r = 0; r < topk; ++r) {
-      topTokenIds[d][r] = idx[r];
-      topLogProbs[d][r] = row[idx[r]] - logZ; // fp32 subtraction
+      topTokenIds[d][r] = buf[r].token;
+      topLogProbs[d][r] = buf[r].logit - logZ; // fp32 subtraction
```
jaemini-shin, line 120: These lines could result false sharing between threads. Is it possible to align with size of cache line 64?

jaemini-shin (approved):
LGTM!

dlwlzzero (approved):
LGTM

nntrainer/utils/ddtree/ddtree.cpp
Seunghui98: buf is a local variable inside the lambda — each parallel invocation owns its own independent instance.

nntrainer/utils/ddtree/ddtree.cpp
Seunghui98: The write at line 120 goes to each d's independently heap-allocated buffer, so no false sharing occurs there.

nntrainer/utils/ddtree/ddtree.cpp
Seunghui98: FYI from claude )  Moving it outside would introduce a data race unless replaced with a per-depth pre-allocated array, but reserve(31) across 15 depths is ~3.7 KB of allocations total, far below any measurable threshold.

PR #4004, opened 2026-06-24
fix(causallm): hold split multi-byte UTF-8 tokens during streaming

jayden0701:

<details><summary>7f91b7e — fix(causallm): hold split multi-byte UTF-8 tokens during streaming</summary><br />

fix(causallm): hold split multi-byte UTF-8 tokens during streaming

A multi-byte UTF-8 character (e.g. a Korean syllable such as '똻') can be
split across consecutive byte-fallback tokens during generation. The
streaming detokenizer in registerOutputs() flushed each token's decoded
text immediately, emitting a half-formed code point that rendered as
broken glyphs (U+FFFD).

The previous incomplete-token guard compared the tail against an empty
string literal and never fired. Replace it with a real check that holds
the delta while the decoded text ends in either an incomplete raw UTF-8
sequence (native BPE decode) or a trailing U+FFFD replacement character
(the HuggingFace/Rust tokenizer decodes partial byte-fallback bytes via
from_utf8_lossy). Holding is bounded by kMaxHeldByteTokens so genuinely
invalid output still drains instead of stalling the stream.

This feature was made into additional header file, so it can be used by external projects
(e.g. QuickAI)

Verified on device with gemma4-cpu: Korean syllables that previously
streamed as broken glyphs now render correctly.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

**Self evaluation:**

</details>

### Summary

- Fix broken glyph rendering when multi-byte UTF-8 characters (e.g. Korean syllables) are split across consecutive byte-fallback tokens during streaming generation. The previous guard (`decoded_str.compare(size-3, 3, "") == 0`) was a dead branch that never fired because it compared against an empty string literal. Replace it with proper UTF-8 incompleteness detection (`endsWithIncompleteUtf8` for native BPE decode, `endsWithReplacementChar` for HuggingFace/Rust tokenizer) and a bounded hold mechanism (`kMaxHeldByteTokens = 6`) that prevents stalling on genuinely invalid output.
- Extract the UTF-8 stream-hold logic into a standalone header (`utf8_stream_util.h` under `causallm::utf8stream` namespace) so it can be reused by external projects (e.g. QuickAI) and the QNN generation loops without depending on `causal_lm.cpp` internals.

### DIFF
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -41,6 +41,7 @@
 
 #include <causal_lm.h>
 #include <llm_util.hpp>
+#include <utf8_stream_util.h>
 
 namespace causallm {
 
@@ -233,9 +234,7 @@ void CausalLM::registerOutputs(
       if (std::find(puncts.begin(), puncts.end(), decoded_str.back()) !=
           puncts.end()) {
         // last symbol is a punctuation, hold on
-      } else if (decoded_str.size() >= 3 &&
-                 decoded_str.compare(decoded_str.size() - 3, 3, "") == 0) {
-        // ends with an incomplete token, hold on
+      } else if (utf8stream::shouldHold(decoded_str, pending_ids_.size())) {
       } else {
         if (log_output) {
           std::cout << decoded_str;
--- a/Applications/CausalLM/utf8_stream_util.h
+++ b/Applications/CausalLM/utf8_stream_util.h
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * @file   utf8_stream_util.h
+ * @brief  Helpers for holding incomplete multi-byte UTF-8 characters during
+ *         streaming detokenization.
+ * @date   24 June 2026
+ * @see    https://github.com/nnstreamer/nntrainer
+ * @author Joonseok Oh <jrock.oh@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ * A multi-byte UTF-8 character (e.g. a Korean syllable like '똻', 3 bytes) can
+ * be generated as several consecutive byte-fallback tokens. Decoding only a
+ * prefix of those tokens yields either incomplete raw bytes (native BPE) or a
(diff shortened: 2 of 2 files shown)


jijoongmoon (approved):
LGTM

dlwlzzero (approved):
LGTM

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -233,9 +234,7 @@ void CausalLM::registerOutputs(
       if (std::find(puncts.begin(), puncts.end(), decoded_str.back()) !=
           puncts.end()) {
         // last symbol is a punctuation, hold on
-      } else if (decoded_str.size() >= 3 &&
-                 decoded_str.compare(decoded_str.size() - 3, 3, "") == 0) {
-        // ends with an incomplete token, hold on
+      } else if (utf8stream::shouldHold(decoded_str, pending_ids_.size())) {
```
jaemini-shin, line 237:
```suggestion
      } else if (utf8stream::shouldHold(decoded_str, pending_ids_.size())) {
                // ends with an incomplete token, hold on
```
It'd be better to keep this comment.

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -233,9 +234,7 @@ void CausalLM::registerOutputs(
       if (std::find(puncts.begin(), puncts.end(), decoded_str.back()) !=
```
jaemini-shin, line 234: (optional) Current design seems not consistent, How about following design for a better maintainability?

* as is
```cpp
if (A) continue;
if (B) {}
else if (C) {}
else {
  // main logic
}
```

* to be
```cpp
if (A || B || C) continue;
// main logic
```

jaemini-shin (approved):
LGTM with some comments!

Applications/CausalLM/models/causal_lm.cpp
jayden0701: great. I will add follow up patch

baek2sm (approved):
LGTM

PR #4011, opened 2026-06-25
fix(layers): generalize incremental_forwarding for multi-token prefill

jayden0701:
## Summary

- Fix incremental_forwarding in BinaryOperationLayer, UnaryOperationLayer, EmbeddingLayer, and MultiOutLayer to support multi-token prefill steps when from > 0
- These layers unconditionally rejected any step size greater than 1 once from > 0, causing a crash on the second consecutive run() call in multi-turn inference

## Root cause

After the first run, global_token_len accumulates so the second run's prefill arrives with from > 0 and step = to - from > 1 simultaneously. All four layers threw here:

NNTR_THROW_IF(to - from != 1, std::invalid_argument)
  << "incremental step size is not 1";
from = 0;
to = 1;  // step forcibly clamped to 1

The fix normalizes the window to 0-based coordinates while preserving the step size — consistent with how AdditionLayer, FCLayer, ActivationLayer, and other layers already behave:

to = to - from;
from = 0;

### DIFF
--- a/nntrainer/layers/embedding.cpp
+++ b/nntrainer/layers/embedding.cpp
@@ -121,10 +121,9 @@ void EmbeddingLayer::incremental_forwarding(RunLayerContext &context,
   unsigned int out_dim = std::get<props::OutDim>(embedding_props);
 
   if (from) {
-    NNTR_THROW_IF(to - from != 1, std::invalid_argument)
-      << "incremental step size is not 1";
+    // Normalize to 0-based while preserving step size for multi-token prefill
+    to = to - from;
     from = 0;
-    to = 1;
   }
 
   Tensor &weight = context.getWeight(weight_idx);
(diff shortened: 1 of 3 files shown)


baek2sm (approved):
Nice Catch, Thanks. LGTM

jaemini-shin (approved):
LGTM!

nntrainer/layers/embedding.cpp
```diff
@@ -121,10 +121,9 @@ void EmbeddingLayer::incremental_forwarding(RunLayerContext &context,
   unsigned int out_dim = std::get<props::OutDim>(embedding_props);
 
   if (from) {
-    NNTR_THROW_IF(to - from != 1, std::invalid_argument)
-      << "incremental step size is not 1";
+    // Normalize to 0-based while preserving step size for multi-token prefill
+    to = to - from;
     from = 0;
```
jaemini-shin, line 126: (optional) Using new variable such as `step = from - to` seems better than overwriting `from` and `to`.
i.e. decouple `from` and `to` from actual `incremental_forwarding` logic.
It makes easy to migrate to `forwarding` and doesn't corrupt given parameters from caller which makes easy to debug.

jijoongmoon (approved):
LGTM

PR #4023, opened 2026-06-29
[CausalLM/layers] implement MHA Core backward pass for LoRA training

niket-agarwal:
## Dependency of the PR

None

## Commits to be reviewed

<details>
<summary>[CausalLM/layers] implement MHA Core backward pass for LoRA training</summary>

Enable gradient computation through the Multi-Head Attention core layer.
Adds training-mode forward that caches inv_k/inv_v scaling factors needed
for backpropagation, and the full calcGradient() implementation that
propagates gradients to attention weights and input tensors. Also adds
supportBackwarding() returning true to signal training support.

</details>

### Summary

- Add training-aware `forwarding()` that caches `inv_k` and `inv_v` scaling factors as intermediate tensors for use during backpropagation
- Implement `calcGradient()` to propagate gradients through the scaled dot-product attention mechanism to Q, K, V weight matrices
- Override `supportBackwarding()` to return `true`, enabling this layer in training graphs

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -7,6 +7,8 @@
  * @see    https://github.com/nntrainer/nntrainer
  *         https://arxiv.org/abs/1706.03762
  * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @author Anirudh <b.saianirud@samsung.com>
+ * @author Niket Agarwal <niket.a@samsung.com>
  * @bug    No known bugs except for NYI items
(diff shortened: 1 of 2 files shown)


baek2sm:
please rebase it to main !

Jungwon-Lee (changes requested):
Requesting changes because the new training path does not preserve the configured attention semantics: it always uses full-prefix causal attention, omits logit softcapping and attention-sink probability mass, and its backward pass applies inverse RoPE unconditionally even when `use_rope=false`. These differences make training optimize a graph that does not match inference. Please cover causal/non-causal, sliding-window, sink, softcapping, and no-RoPE cases with numerical gradient tests.

PR #4024, opened 2026-06-29
[CausalLM/layers] implement LM Head backward pass for LoRA training

sumon-98:
## Dependency of the PR

None

## Commits to be reviewed

<details>
<summary>[CausalLM/layers] implement LM Head backward pass for LoRA training</summary>

Enable gradient computation through the Language Model Head layer.
Adds calcGradient() that accumulates weight gradients and propagates
input gradients, along with g_lm_head_read_row() for reading specific
rows of the gradient matrix. Also fixes right-pad position tracking
so that gradient accumulation correctly skips padding tokens.

</details>

### Summary

- Add `calcGradient()` to accumulate weight gradients and propagate input gradients through the LM head projection matrix
- Implement `g_lm_head_read_row()` helper for reading specific rows of the gradient tensor during sparse gradient accumulation
- Fix right-pad position tracking so gradient accumulation skips over padding tokens when computing loss-masked updates

### DIFF
--- a/Applications/CausalLM/layers/lm_head.cpp
+++ b/Applications/CausalLM/layers/lm_head.cpp
@@ -7,6 +7,8 @@
  * @brief  This is lmhead layer
  * @see    https://github.com/nntrainer/nntrainer
  * @author Eunju Yang <ej.yang@samsung.com>
+ * @author Pranjal Thapliyal <p.thapliyal@samsung.com>
+ * @author Sumon Nath <sumon.nath@samsung.com>
  * @bug    No known bugs except for NYI items
  *
  */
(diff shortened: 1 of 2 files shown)


Jungwon-Lee (changes requested):
Requesting changes because one thread-local scalar is used as the LM-head read row for every sample in a batch. Mixed-length batches therefore read logits from, and propagate gradients to, the wrong token for most samples. In addition, `calcGradient()` silently does nothing although the LM-head weight is trainable by default. Please track positions per batch item (preferably from a mask) and either implement weight gradients or enforce a frozen LM head in the layer contract.

PR #4025, opened 2026-06-29
[CausalLM/layers] implement TieWordEmbedding backward pass for LoRA training

b-saianirud:
## Dependency of the PR

None

## Commits to be reviewed

<details>
<summary>[CausalLM/layers] implement TieWordEmbedding backward pass for LoRA training</summary>

Enable gradient computation through the Tied Word Embedding layer.
Adds calcGradient() that accumulates gradients into the shared embedding
weight, preserving the weight-sharing semantics between the embedding
matrix and the output projection. This is essential for correct
gradient flow during LoRA fine-tuning of the Qwen3 model.

</details>

### Summary

- Add `calcGradient()` to accumulate input gradients into the shared embedding weight matrix, correctly handling the tied weight relationship between the input embedding and the output LM head projection
- Override `supportBackwarding()` to return `true`, enabling gradient flow through this layer during LoRA training

### DIFF
--- a/Applications/CausalLM/layers/causallm_common_properties.h
+++ b/Applications/CausalLM/layers/causallm_common_properties.h
@@ -17,6 +17,8 @@
  * @brief	This defines a qwen3 causal language model.
  * @see		https://github.com/nnstreamer/
  * @author	Eunju Yang <ej.yang@samsung.com>
+ * @author	Pranjal Thapliyal <p.thapliyal@samsung.com>
+ * @author	Anirudh <b.saianirud@samsung.com>
  * @bug		No known bugs except for NYI items
  *
  */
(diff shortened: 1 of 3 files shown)


Jungwon-Lee (changes requested):
Requesting changes due to several correctness regressions: the last-token position is shared across the whole batch, the LM-head weight-gradient path uses the full batched `dy` against a single-sample input view, and the Q6_K save path reverses the embedding dimensions. Also, `read()` now drops `start_offset` and `read_from_offset`, changing existing shared/partial weight-loading semantics. Please preserve the offsets and add batch>1 tied-embedding forward/backward plus save/load round-trip tests.

PR #4026, opened 2026-06-29
[CausalLM/layers] implement RMSNorm backward pass for LoRA training

niket-agarwal:
## Dependency of the PR

None

## Commits to be reviewed

<details>
<summary>[CausalLM/layers] implement RMSNorm backward pass for LoRA training</summary>

Enable gradient computation through the Root Mean Square normalisation
layer. Adds an inv_rms intermediate tensor (ITERATION_LIFESPAN) cached
during the forward pass and used in calcGradient() to compute exact
gradients with respect to both the normalised input and the gamma scale
weight. Also overrides supportBackwarding() to return true.

</details>

### Summary

- Cache `inv_rms` (reciprocal RMS) as an `ITERATION_LIFESPAN` scratch tensor during `forwarding()` for use in backpropagation
- Implement `calcGradient()` to compute exact gradients with respect to the input and the learnable gamma scale weight using the cached inv_rms values
- Override `supportBackwarding()` to return `true`, enabling this layer in training computation graphs

### DIFF
--- a/Applications/CausalLM/layers/rms_norm.cpp
+++ b/Applications/CausalLM/layers/rms_norm.cpp
@@ -4,22 +4,34 @@
  *
  * @file   rms_norm.cpp
  * @date   19 July 2023
- * @brief  Implementation of custom RMS normalization function
+ * @brief  Implementation of RMS normalisation.
+ *
+ *         Forward:  out = x * inv_rms * gamma
+ *         where     inv_rms = 1 / sqrt(mean(x²) + epsilon)
+ *         Normalisation is over the width (last) dimension.
+ *
+ *         This is the special case of ReshapedRMSNorm where
+ *         feature_size == width — one chunk per row, no reshape needed.
+ *
  * @see    https://github.com/nntrainer/nntrainer
  * @author Seungbaek Hong <sb92.hong@samsung.com>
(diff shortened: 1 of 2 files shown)


Jungwon-Lee (changes requested):
Requesting changes. This PR uses `RMSParams::inv_rms` even though that enum entry is only introduced by another PR, so it does not compile against main despite declaring no dependency. Moreover, gamma and the `inv_rms` cache inherit `getWeightDataType()` but backward accesses both through raw `float *`; FP16 or block-quantized configurations can misinterpret storage or corrupt memory. Please make the dependency explicit/self-contained and allocate training caches in FP32 activation dtype or reject unsupported dtypes before raw access.

Jungwon-Lee (changes requested):
Requesting changes. This PR uses RMSParams::inv_rms even though that enum entry is only introduced by another PR, so it does not compile against main despite declaring no dependency. Moreover, gamma and the inv_rms cache inherit getWeightDataType() but backward accesses both through raw float pointers; FP16 or block-quantized configurations can misinterpret storage or corrupt memory. Please make the dependency explicit/self-contained and allocate training caches in FP32 activation dtype or reject unsupported dtypes before raw access.

PR #4027, opened 2026-06-29
[CausalLM/layers] implement ReshapedRMSNorm backward pass for LoRA training

b-saianirud:
## Dependency of the PR

None

## Commits to be reviewed

<details>
<summary>[CausalLM/layers] implement ReshapedRMSNorm backward pass for LoRA training</summary>

Enable gradient computation through the ReshapedRMSNorm layer, which
applies RMS normalisation over feature_size-wide chunks of the width
dimension (per-head or per-group normalisation). Adds calcGradient()
with the same chunked iteration pattern as incremental_forwarding, and
overrides supportBackwarding() to return true.

</details>

### Summary

- Implement `calcGradient()` using the same chunked iteration pattern as `incremental_forwarding()`, computing exact gradients for both the normalised input and the gamma scale weight per feature-size chunk
- Override `supportBackwarding()` to return `true`, enabling this layer in LoRA fine-tuning training graphs

### DIFF
--- a/Applications/CausalLM/layers/reshaped_rms_norm.cpp
+++ b/Applications/CausalLM/layers/reshaped_rms_norm.cpp
@@ -4,50 +4,176 @@
  *
  * @file   reshaped_rms_norm.cpp
  * @date   19 July 2023
- * @brief  Implementation of custom RMS normalization function
+ * @brief  Implementation of ReshapedRMSNorm — RMS normalisation over
+ *         feature_size-wide chunks of the width dimension.
+ *
+ *         This allows per-head or per-group normalisation without changing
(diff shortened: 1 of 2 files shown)


Jungwon-Lee (changes requested):
Requesting changes because the backward implementation reads quantized/FP16 weight-typed buffers through raw float pointers and reads the input after its planned lifetime may have ended. There are also two inference regressions: use_gamma=false is ignored and gamma is always requested/applied, while the FP32 branch calls rms_norm_wrt_width_fp16_intrinsic() whenever ENABLE_FP16 is defined. Please restore the no-gamma and FP32-kernel paths, use dtype-safe training buffers, and preserve/cache the input needed for derivative computation.

PR #4029, opened 2026-06-29
[CausalLM] add LoRA fine-tuning training pipeline for Qwen3

niket-agarwal:
## Summary

Adds end-to-end LoRA fine-tuning training pipeline for the Qwen3 causal language model on top of the training infrastructure introduced in the preceding PR.

- **`lora_train.cpp/.h`**: Introduces `TrainingDataGenerator`, which tokenizes a text file into sliding-window samples, shuffles each epoch with a fixed seed, and exposes an `ml::train` data-generator callback that right-pads short sequences and masks padding positions in the loss.
- **`qwen3_causallm.cpp`**: Overrides `hasLoRA` / `appendLoRAProps` to inject LoRA adapter rank/alpha into Q/K/V/O and MLP gate/up/down FC layers; non-LoRA layers are marked `trainable=false` so only adapters are updated.
- **`causal_lm.cpp`**: Sets the LM head `trainable=false` when `LORA_RANK > 0` to keep the token projection frozen during LoRA fine-tuning.
- **`train_qwen3_lora_master.cpp`**: CLI entry point with `--lr`, `--epochs`, `--patience` (early stopping), `--max_samples`, `--lora_qat`, `--lora_q4`, `--seed`, and `--output_path` flags; saves the LoRA adapter weights on completion.
- **`main.cpp`**: Adds `--lora_file_name` argument support for loading a saved LoRA adapter during inference.
- **`meson.build`**: Adds `lora_train` and `train_qwen3_lora_master` build targets.

## Test plan

- [ ] Build with `ninja -C build Applications/CausalLM/lora_train` and `ninja -C build Applications/CausalLM/train_qwen3_lora_master`
- [ ] Run `train_qwen3_lora_master` with a small text corpus and verify loss decreases over epochs
- [ ] Verify saved `.bin` adapter loads back correctly via `main --lora_file_name`
- [ ] Check that non-LoRA layers (attention, LM head, embedding) produce zero gradients
- [ ] Confirm padding-token positions are masked (loss should not decrease on PAD tokens)

### DIFF
--- a/Applications/CausalLM/lora_train.cpp
+++ b/Applications/CausalLM/lora_train.cpp
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2026 Eunju Yang <ej.yang@samsung.com>
+ *
+ * @file   lora_train.cpp
+ * @date   01 Apr 2026
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author Eunju Yang <ej.yang@samsung.com>
+ * @author Sumon Nath <sumon.nath@samsung.com>
+ * @author Niket Agarwal <niket.a@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @brief  LoRA training data pipeline implementation
+ */
+
+#include "lora_train.h"
+
+#include <algorithm>
+#include <fstream>
+#include <iostream>
+#include <numeric>
+#include <stdexcept>
+
+#include <lm_head.h>  // for g_lm_head_read_row
+
+namespace causallm {
+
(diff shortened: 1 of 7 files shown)


Jungwon-Lee (changes requested):
Requesting changes. This branch calls LoRA helpers that are not present on main, so it does not compile without the preceding infrastructure PR. The default target list includes FFN modules, but this branch only adds LoRA handling to Qwen3 attention; the FFN base weights remain trainable and the saved adapter cannot reproduce the trained model. The new executable also includes `sys/resource.h` and uses `getrusage`/`/proc` without Windows guards, breaking the supported Windows build. Please declare/include the stacked dependencies, apply LoRA/freeze handling to every advertised target, and use the existing cross-platform memory helper.

PR #4034, opened 2026-07-01
[nntrainer] Reduce peak memory

jaemini-shin:

[nntrainer] Reduce peak memory by removing redundant initialization

This commit removes redundant initialization in mem_allocator.

Self evaluation:

</details>

<details><summary>{commit-2}</summary><br />

[nntrainer] Reduce peak memory in weight loading

This commit reduces peak memory by using per-task mmap pointers.
As of now, nntrainer loads weight into tensor by using memcpy
which doubles peak memory: mmap pointer region and tensor region.
By using per-task mmap pointers, we can reduce the incremental amount for each task.

However, using a single shared mmap pointer was introduced to resolve runtime error
caused by too many concurrent creation of mmap pointers.
By limiting the upper bound of I/O threads, this issue can be resolved.

**Self evaluation:**

</details>

## Details

Let me explain how using per-task mmap pointers reduces peak memory.

Let's assume that we are loading a model:
* layer0: 100 bytes
* layer1: 200 bytes

### C1: using shared mmap
* make a mmap
* loading layer0
  * memcpy
    * read mmap (+100, peak: 100, current: 100)
    * write tensor (+100, peak: 200, current: 200)
* loading layer1
  * memcpy
    * read mmap (+200, peak: 400, current: 400)
    * write tensor (+200, peak: 600, current: 600)
* delete mmap (-300, peak: 600, current: 300)

As a result, at most 600 bytes is in physical memory in a moment.

### C2: using per-task mmap
* loading layer0
  * make a mmap
  * memcpy
    * read mmap (+100, peak: 100, current: 100)
    * write tensor (+100, peak: 200, current: 200)
  * delete mmap (-100, peak: 200, current: 100)
* loading layer1
  * make a mmap
  * memcpy
    * read mmap (+200, peak: 300, current: 300)
    * write tensor (+200, peak: 500, current: 500)
  * delete mmap (-200, peak: 500, current: 300)

As a result, at most 500 bytes is in physical memory in a moment.

## Summary
For q6k-q40 quantized qwen3-0.6b with 1K input:
* as is
  * peak memory: 1.4GB
* first commit
  * peak memory: 900MB
* first commit + second commit
  * peak memory: 800MB

## To do
* Remove memcpy for weights which do not need additional packing

### DIFF
--- a/nntrainer/mem_allocator.cpp
+++ b/nntrainer/mem_allocator.cpp
@@ -57,11 +57,6 @@ void MemAllocator::alloc(void **ptr, size_t size, size_t alignment) {
   NNTR_THROW_IF(*ptr == nullptr, std::runtime_error)
     << "MemAllocator::alloc: aligned_alloc(" << alignment << ", "
     << aligned_size << ") failed";
-
-  // MemoryPool callers historically expected zeroed buffers (calloc
-  // semantics). Preserve that — kernels that read uninitialised
-  // gradient slots would otherwise see garbage.
-  std::memset(*ptr, 0, aligned_size);
 }
 
 void MemAllocator::free(void *ptr) {
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -24,6 +24,8 @@
 #include "layer_context.h"
 #include "model.h"
 #include "model_common_properties.h"
+#include <algorithm>
+#include <atomic>
 #include <cmath>
 #include <compute_ops.h>
 #include <cstdint>
@@ -33,6 +35,7 @@
 #include <future>
 #include <iomanip>
 #include <sstream>
+#include <thread>
 
 #include <activation_realizer.h>
 #include <adamw.h>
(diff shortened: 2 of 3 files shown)


nntrainer/models/neuralnet.cpp
jaemini-shin, line 1: This change is almost the same as reverting https://github.com/nntrainer/nntrainer/pull/3940.

nntrainer/models/neuralnet.cpp
```diff
@@ -1006,40 +1009,23 @@ void NeuralNetwork::load(const std::string &file_path,
...
-                                 MAP_PRIVATE, model_file_fd, 0);
-        NNTR_THROW_IF((shared_mmap_ptr == MAP_FAILED), std::runtime_error)
-          << "mmap failed for " << f_path << " (" << shared_mmap_size
-          << " bytes)";
-        (void)::posix_madvise(shared_mmap_ptr, shared_mmap_size,
-                              POSIX_MADV_RANDOM);
-      }
-#endif
+      // Load weights with bounded thread number not to exceed mmap limits
+      constexpr size_t MAX_LOAD_THREADS = 8;
```
jaemini-shin, line 1013: To avoid a crash due to mmap explosion, limiting the upper bound is necessary.

nntrainer/mem_allocator.cpp
```diff
@@ -57,11 +57,6 @@ void MemAllocator::alloc(void **ptr, size_t size, size_t alignment) {
   NNTR_THROW_IF(*ptr == nullptr, std::runtime_error)
     << "MemAllocator::alloc: aligned_alloc(" << alignment << ", "
     << aligned_size << ") failed";
-
-  // MemoryPool callers historically expected zeroed buffers (calloc
-  // semantics). Preserve that — kernels that read uninitialised
-  // gradient slots would otherwise see garbage.
-  std::memset(*ptr, 0, aligned_size);
```
jaemini-shin, line 64: Since tensors are initialized when creating actual tensor base object, this logic is redundant.

jijoongmoon (approved):
LGTM!

jayden0701 (approved):
LGTM with question.

Q: Does this commit change load time?

jaemini-shin:
> LGTM with question.
> 
> Q: Does this commit change load time?

@jayden0701 This might change load time.

Load latency depends on the bandwidth of memory. Using more threads helps fully using bandwidth, thus limiting 8 threads might not enough to reach it compared to before. If load time is increased, adjusting the upper bound of thread would resolve this issue.

dlwlzzero (approved):
LGtM

nntrainer/models/neuralnet.cpp
Seunghui98: Could you explain what is `load_threads`...?
I dont' know it.. 🤔

nntrainer/mem_allocator.cpp
Seunghui98: good point!

Seunghui98 (approved):
LGTM 👍 
but  i have a question.
Do this approach introduce additional overhead cost to remove mmap per layer??

nntrainer/models/neuralnet.cpp
jaemini-shin: Before using shared mmap pointer, we had made threads with the size of model layers. Each thread tried to make a separate mmap pointer, which might over the maximum number of mmap that the device can provide.
To overcome this limit, I introduced the upper limit of load workers. i.e. only `MAX_LOAD_THREADS` workers can load weight in the same moment.

nntrainer/models/neuralnet.cpp
```diff
@@ -1006,40 +1009,23 @@ void NeuralNetwork::load(const std::string &file_path,
...
+      std::vector<std::shared_ptr<LayerNode>> load_nodes(model_graph.cbegin(),
+                                                         model_graph.cend());
+      const size_t num_load_nodes = load_nodes.size();
+      const size_t num_load_threads =
+        std::min<size_t>(num_load_nodes, MAX_LOAD_THREADS);
 
-      // std::vector<std::future<void>> futures;
-      std::vector<std::thread> threads;
-      threads.reserve(model_graph.size());
```
jaemini-shin, line 1036: @Seunghui98 check these lines!

PR #4035, opened 2026-07-02
fix(qnn): resolve QNN graph lookup case-insensitively

jayden0701:
### Problem
QNN graph names are case-sensitive, but `GraphCore::ensureName` lowercases every layer name. As a result, models whose QNN graph names contain uppercase letters (e.g. `gemma-4-e2b-qnn`'s `gemma_4_E2B_..._prefill`) fail at runtime with "cannot find graph" / "Unable to retrieve graph handle", because the lowercased lookup key never matches the binary's real graph name. All-lowercase models only work by coincidence.

### Solution
- `getGraphPtr` now falls back to a **case-insensitive match** when the exact lookup misses, while preserving the exact-match fast path for all-lowercase models.
- `graphRetrieve` passes the binary's **real-cased name** (`graphInfo->graphName`) to QNN's case-sensitive `graphRetrieve` instead of the lowercased key.

### Files Changed
- `nntrainer/qnn/jni/qnn_context_var.h` (+35 / -6)

### DIFF
--- a/nntrainer/qnn/jni/qnn_context_var.h
+++ b/nntrainer/qnn/jni/qnn_context_var.h
@@ -24,6 +24,7 @@
 #include "qnn_rpc_manager.h"
 
 #include <algorithm>
+#include <cctype>
 #include <cstdint>
 #include <fcntl.h>
 #include <functional>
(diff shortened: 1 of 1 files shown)


jijoongmoon (approved):
LGTM!

jaemini-shin (approved):
LGTM!

Seunghui98 (approved):
LGTM 👍

lhs8928 (approved):
LGTM

baek2sm (approved):
LGTM

PR #4040, opened 2026-07-03
[nntrainer] Fix is_prefill to recognize multi-token prefill steps

jayden0701:

<details><summary>[nntrainer] Fix is_prefill to recognize multi-token prefill steps</summary><br />

skip_prefill-capable layers computed is_prefill as simply !from, which
only recognized the very first prefill call (from == 0) as prefill.
During multi-turn inference, a later turn's prefill re-enters these
layers with from > 0 and a multi-token step (to - from > 1) -- e.g. the
KV-shared/sliding-window layers in gemma4 -- so is_prefill
evaluated false and the layer computed normally instead of skipping.

mha_core.cpp already treated any step > 1 as prefill regardless of
from; the affected layers now match that convention:

  bool is_prefill = !from || (to - from) > 1;

Without this, a skip_prefill layer downstream of one that already
skipped (e.g. mha_core) would compute over that layer's stale output
tensor, producing well-defined but incorrect hidden states that
propagate into later layers' KV cache writes.

Affected: fc_layer, addition_layer, activation_layer, multiply_layer
(nntrainer/layers), and rms_norm, reshaped_rms_norm, rms_reverse_norm,
swiglu, per_layer_slice, scalar_multiply, logit_softcapping
(Applications/CausalLM/layers). Also reorders rms_reverse_norm's check
to compute is_prefill before normalizing from/to, so the skip decision
sees the original step size.

lm_head / tie_word_embedding's lmhead path intentionally left
unchanged -- they always compute only the last row of the given range,
and the SKIP_PREFILL multi-token segment's output is discarded by the
caller (causal_lm.cpp), so not skipping there wastes compute but does
not affect correctness.

**Self evaluation:**

</details>

### Summary

- Fix `is_prefill` detection in 10 layers (`fc_layer`, `addition_layer`, `activation_layer`, `multiply_layer`, `rms_norm`, `reshaped_rms_norm`, `rms_reverse_norm`, `swiglu`, `per_layer_slice`, `scalar_multiply`, `logit_softcapping`) so multi-token, non-zero-offset prefill (`from > 0 && to - from > 1`) is correctly treated as prefill, matching `mha_core.cpp`'s existing convention.

### DIFF
--- a/Applications/CausalLM/layers/logit_softcapping.cpp
+++ b/Applications/CausalLM/layers/logit_softcapping.cpp
@@ -60,7 +60,7 @@ void LogitSoftCappingLayer::forwarding(nntrainer::RunLayerContext &context,
 void LogitSoftCappingLayer::incremental_forwarding(
   nntrainer::RunLayerContext &context, unsigned int from, unsigned int to,
   bool training) {
-  bool is_prefill = !from;
+  bool is_prefill = !from || (to - from) > 1;
   if (skip_prefill && is_prefill)
     return;
 
--- a/Applications/CausalLM/layers/per_layer_slice.cpp
+++ b/Applications/CausalLM/layers/per_layer_slice.cpp
@@ -40,7 +40,7 @@ void PerLayerSliceLayer::forwarding(nntrainer::RunLayerContext &context,
 void PerLayerSliceLayer::incremental_forwarding(
   nntrainer::RunLayerContext &context, unsigned int from, unsigned int to,
   bool training) {
-  bool is_prefill = !from;
+  bool is_prefill = !from || (to - from) > 1;
   if (skip_prefill && is_prefill)
     return;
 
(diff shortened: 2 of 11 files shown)


Jungwon-Lee (approved):
LGTM

Applications/CausalLM/layers/logit_softcapping.cpp
```diff
@@ -60,7 +60,7 @@ void LogitSoftCappingLayer::forwarding(nntrainer::RunLayerContext &context,
 void LogitSoftCappingLayer::incremental_forwarding(
   nntrainer::RunLayerContext &context, unsigned int from, unsigned int to,
   bool training) {
-  bool is_prefill = !from;
+  bool is_prefill = !from || (to - from) > 1;
```
jaemini-shin, line 63: IMO, `to - from == 1` is not a necessary condition for decode stage.
* In chunked prefill, a tail chunk could be the length of 1.

If we need to strictly judge prefill and decode, additional information should be given by the caller's side.

As of now, this change would cover most of the cases, but we should handle this edge case later.

jayden0701: I think that will be important when we get rid of `incremental_forwarding(and from/to)`.
I will keep this in mind. Thanks!

jaemini-shin (approved):
LGTM!

baek2sm (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #4042, opened 2026-07-06
[nntrainer] Allow leading underscore in layer/tensor names

dlwlzzero:
## Summary

`Name::isValid` rejected any layer/tensor name whose first character was
`_`, because the allowed pattern required the leading character to be
alphanumeric:

[a-zA-Z0-9][-_./a-zA-Z0-9]*

ONNX-exported QNN graphs commonly name tensors with a leading underscore.
Since nntrainer layer names
must match the graph tensor names exactly for input binding, loading such a
model threw `std::invalid_argument` during name validation.

## Changes

- **`nntrainer/layers/common_properties.cpp`** — widen the leading-character
  class in `Name::isValid` from `[a-zA-Z0-9]` to `[_a-zA-Z0-9]`, so names
  beginning with `_` are accepted. Interior characters are unchanged.
- **`nntrainer/layers/layer_node.cpp`** — comment only (no behavior change).
  Document that `createLayerNode` forwarding `properties` to
  `Engine::createLayerObject` is intentional and is *not* a double-dispatch
  onto the raw layer: `Engine` uses them only for `parseComputeEngine`
  (`engine=...` Context selection) and calls `Context::createLayerObject(type)`
  with no properties.

### DIFF
--- a/nntrainer/layers/common_properties.cpp
+++ b/nntrainer/layers/common_properties.cpp
@@ -42,7 +42,10 @@ void Name::set(const std::string &value) {
 }
 
 bool Name::isValid(const std::string &v) const {
-  static std::regex allowed("[a-zA-Z0-9][-_./a-zA-Z0-9]*");
+  // Leading '_' is allowed: ONNX-exported QNN graphs name tensors like
+  // "_model_embed_tokens_Gather_Gather_output_0" and layer names must match
+  // the graph tensor names exactly for input binding.
+  static std::regex allowed("[_a-zA-Z0-9][-_./a-zA-Z0-9]*");
   return !v.empty() && std::regex_match(v, allowed);
 }
 


nntrainer/layers/layer_node.cpp
```diff
@@ -165,6 +165,12 @@ std::unique_ptr<LayerNode>
 createLayerNode(const ml::train::LayerType &type,
                 const std::vector<std::string> &properties) {
   auto &eg = nntrainer::Engine::Global();
+  // properties must reach createLayerObject so Engine::parseComputeEngine can
+  // see an "engine=..." key and route to the right Context (e.g. "qnn").
+  // This does NOT double-dispatch properties onto the raw layer: Engine::
+  // createLayerObject only forwards properties to parseComputeEngine and
+  // calls ct->createLayerObject(type) with no properties, so the raw layer
+  // factory always gets {} regardless of what's passed here.
```
jaemini-shin: Looks like this comment is not related to this PR. Is it necessary to include this comment?

jijoongmoon (commented):
It might be better to set the rule as in below.
For the layer name, it is ok to have capital letter and others charactor as well.
But for the layer type, we do need lowering.

nntrainer/layers/layer_node.cpp
dlwlzzero: Thanks for checking, I'll remove it.

jaemini-shin:
Looks like signed off is not included in your commit message

jijoongmoon (commented):
I think we have to have different rult for name. for the layer name, I think it is ok to have capital letter as well. but for the layer type, it might be better to have lower letter.

dlwlzzero:
> It might be better to set the rule as in below. For the layer name, it is ok to have capital letter and others charactor as well. But for the layer type, we do need lowering.

GraphCore::ensureName() in graph_core.cpp also lowercases the name (on every addNode()), so it stores a lowercased name regardless of this fix. It's tangled with duplicate-name/auto-numbering logic though, so I'd rather handle it in a separate PR instead of bundling it here.

baek2sm (approved):
LGTM
I also agree with @dlwlzzero . This PR addresses an issue distinct from lower case and upper case letters.

jaemini-shin (approved):
LGTM!

PR #4044, opened 2026-07-06
causallm: add Q4_0-FP16 activation support

Jungwon-Lee:
## Summary
Split out of #4000 ("Add Flash Attention and FP16 Activation"), which bundled two independent
features. This PR carries the **FP16-activation** half only (17 commits, unmodified authorship
except one split commit noted below).

- Adds ggml Q4_0/Q6_K FP16-activation GEMM/GEMV kernels (with an x86 non-NEON fallback)
- Fixes FC bias dtype, embedding dequant, RMSNorm/ReshapedRMSNorm, tie_word_embedding,
  per_layer_slice, and scalar_multiply for FP16-activation correctness
- Adds Q4_0-FP16 activation test suites for Qwen2, Qwen3, Gemma3, Gemma4, GptOss (ENABLE_FP16 only)

Companion PR (Flash Attention half): #4045

Note: commit `causallm: fix x86 Q4_0-FP16 GEMM/GEMV fallback for non-NEON builds` is a split of
original commit e04b1f93, which touched both this feature and a flash-only AVX2 kernel; only the
FP16-activation hunk is included here (the flash hunk went to the companion PR).

## Test plan
- [x] x86 build (gcc-13, `-Denable-fp16=true -Denable-transformer=true`) - clean except two
      pre-existing baseline gaps in mha_core.cpp/deberta_attention_layer.cpp unrelated to this
      PR (flash-only x86 fp16 cache kernels, present on main before this PR)
- [x] unittest_nntrainer_cpu_backend - 51/51 pass
- [x] unittest_mixed_models - 4/4 pass
- [x] unittest_embedding_sidecar_lut - 10/10 pass
- [x] Commit-integrity check: this PR + the flash-attention PR reproduce #4000's tree exactly
      (verified via git diff - empty)
- [x] Android ARM FP16 device validation (left to reviewer per project convention)

PR #4000, opened 2026-06-19
Add Flash Attention and FP16 Activation

Jungwon-Lee:
## Summary

This PR adds two complementary features to the CausalLM inference path:

1. **GEMM-based flash attention** — online-softmax attention for long-context prefill, with GQA support and causal/sliding-window block-skip
2. **FP16 activation end-to-end** — Q4_0 and Q6_K weight × FP16 activation GEMM/GEMV kernels (NEON), plus the layer-level fixes needed to run the complete forward pass with FP16 activations

---

### Algorithm

Implements flash attention (Phase 2 tiling, online softmax) for the MHA forward pass in `mha_core`. The implementation:

- Activates for prefill only (`step_size >= 32`); single-token decode keeps the existing per-row dot path
- Supports GQA: Q-head / KV-head groups are mapped correctly without materialising extra KV copies
- Applies causal masking and sliding-window masking with block-skip to avoid computing fully-masked (Q, K) tiles
- Config: `use_flash_attention` (default `true`) in `nntr_config.json`, propagated through `Transformer` / `Qwen3Transformer` → `mha_core`

### Kernels

The QK and AV GEMM primitives are promoted to `cpu_backend` library functions:

| Kernel | ISA | Direction | File |
|--------|-----|-----------|------|
| `hgemm_f16xf16_f32_fmlal` | ARMv8.2-A FMLAL | FP16 × FP16 → FP32 (QK) | `cpu_backend/arm/hgemm/` |
| `hsgemm_fp16bits_avx2` | AVX2 + F16C + FMA | FP32 × FP16-bits → FP32 (QK, AV) | `cpu_backend/x86/avx2_impl.cpp` |

On x86, K and V are read as raw `uint16_t` FP16-bits directly from the KV cache — no FP32 staging buffer — saving ~14 MB per call and one memory pass.

### Performance (Qwen3-0.6B, Q4_0-FP32, 8 threads)

**Long prefill (1003 tokens) + 32-token decode:**

| Device | Flash OFF (prefill ms / TPS) | Flash ON (prefill ms / TPS) | Prefill speedup |
|--------|------------------------------|-----------------------------|-----------------|
| S25 Ultra | 2967 ms / 338 TPS | 1491 ms / 672 TPS | **2.0×** |
| S26 Ultra | 3047 ms / 329 TPS | 1768 ms / 567 TPS | **1.7×** |
| x86 Ryzen 7 | 5994 ms / 167 TPS | 6242 ms / 160 TPS | ~equal |

[...]

PR #4045, opened 2026-07-06
causallm: add GEMM-based flash attention

Jungwon-Lee:
## Summary
Split out of #4000 ("Add Flash Attention and FP16 Activation"), which bundled two independent
features. This PR carries the **Flash Attention** half only (10 original commits + 1 split-off
commit noted below, unmodified authorship).

- Adds GEMM-based flash attention to `mha_core` (GQA + online softmax), moved into the
  cpu_backend library with declared kernels
- 4x2 register-blocked QK fmlal ARM kernel (~1.8x faster, bit-identical), thread_local Phase 1
  buffers, lazy per-tile copy
- Enables flash attention for all CausalLM models (and Bert/TimmViT embedding models)
- Adds `docs/flash-attention.md` design doc

Companion PR (FP16 Activation half): #4044 (pr/fp16-activation)

Note: this PR includes a new commit `causallm: guard hsgemm_fp16bits_avx2 with explicit
AVX2/F16C/FMA target`, split from original commit e04b1f93 (which also touched FP16-activation
code that went to the companion PR instead). Only the flash-relevant AVX2 target-attribute hunk
is included here.

The GEMM/flash-attention path is built and validated on the ARM Android FP16 device build; on
non-FP16 x86 builds it compiles via the FP32-fallback path (unvalidated on x86, per existing
comment in `mha_core.cpp`).

## Test plan
- [x] x86 build (gcc-13, `-Denable-fp16=false -Denable-transformer=true`) — fully clean, 777/777
      targets
- [x] `unittest_causallm_models` — 69/69 pass (incl. `qwen3_flash_tiny` reference test with
      `use_flash_attention: true`)
- [x] `unittest_kv_cache_manager` — 17/17 pass
- [x] `unittest_cancel_api` — 7/7 pass
- [x] `unittest_embedding_sidecar_lut` — 10/10 pass
- [x] Commit-integrity check: this PR ∪ the FP16-activation PR reproduce #4000's tree exactly
      (verified via `git diff` — empty)
- [ ] Android ARM FP16 device validation (left to reviewer per project convention)


nntrainer/tensor/cpu_backend/arm/hgemm/hgemm.cpp
```diff
@@ -195,3 +195,136 @@ void hgemm_K1(const __fp16 *A, const __fp16 *B, __fp16 *C, unsigned int M,
...
+// the rows are contiguous).
+//
+// Requires FEAT_FHM (asimdfhm in /proc/cpuinfo). The target attribute pulls
+// the fp16fml ISA extension in only for this function so the rest of the TU
+// can stay on the build-wide -march flags.
+// Single-output FP16xFP16->FP32 dot, used for the M/N tails of the blocked
+// kernel below. The 4x2 block reproduces this exact accumulation order per
+// output, so blocked and tail results are bit-identical (no token drift).
+__attribute__((
+  target("arch=armv8.2-a+fp16+fp16fml+dotprod+i8mm"))) static inline float
```
jaemini-shin: nntrainer is using `armv8.2-a+fp16+dotprod+i8mm`, and `fp16fml` is optional extension for `armv8.2-a`. This might result runtime error on some devices. We need a kernel that doesn't rely on `fp16fml` and remove this attribute.

nntrainer/tensor/cpu_backend/cpu_backend.h
```diff
@@ -25,6 +25,26 @@
...
+// -march=...+fp16): hgemm.h and neon_impl.h declare bare __fp16 signatures,
+// which is not a usable type on ARM targets built without that extension
+// (e.g. Tizen armv7l/aarch64, which do not support fp16 NEON).
+#if !defined(__x86_64__) && !defined(__i386__) && defined(__ARM_NEON) &&       \
+  (defined(ENABLE_FP16) || defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC))
+#include <hgemm.h>     // hgemm_f16xf16_f32_fmlal (QK, FMLAL)
+#include <neon_impl.h> // nntrainer::neon::hgemm_f16xf16_f16 (AV) + exp_ps
+#elif defined(__x86_64__) || defined(__i386__)
+#include <avx2_impl.h> // nntrainer::avx2::hsgemm_fp16bits_avx2 (QK + AV)
+#endif
```
jaemini-shin, line 46: This kind of divergence makes hard to maintain the code. IMO, it'd be better to define shgemm for flash attention outside of hgemm, so that no additional dispatch occurs.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -35,6 +35,30 @@
...
+ *       targets -march=...+fp16), since __fp16 is not a usable type on ARM
+ *       targets built without that extension (e.g. Tizen armv7l/aarch64,
+ *       which do not support fp16 NEON).
+ */
+void shgemm(const unsigned int TStorageOrder, bool TransA, bool TransB,
+            const unsigned int M, const unsigned int N, const unsigned int K,
+            const float alpha, const float *A, const unsigned int lda,
+            const __fp16 *B, const unsigned int ldb, const float beta, float *C,
+            const unsigned int ldc);
+#endif
```
jaemini-shin, line 60: IMO, it'd be better to move `shgemm` outside of `ENABLE_FP16` block.

### RATIONALE
`shgemm` has two steps: `scopy`(convert f16 to f32), `sgemm`.
`scopy` uses `vcvt_f32_f16`, `vld1_f16` which can be used without `+fp16` flag.
So, by moving this function outside of FP16 block can easily resolve this issue.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -781,6 +795,365 @@ void MHACoreLayer::one_batch_incremental_forwarding(
                                 cache_to);
 }
 
+void MHACoreLayer::gemm_attention(nntrainer::Tensor &query_step,
```
jaemini-shin, line 798: This function should be delagated into separate `AVX2` and `neon` kernels.

nntrainer/tensor/cpu_backend/arm/meson.build
```diff
@@ -21,13 +21,20 @@ nntrainer_inc += include_directories('matrix_transpose_neon')
...
 
 
+# cpu_backend.h declares hgemm.h's flash-attention GEMM kernels (e.g.
+# hgemm_f16xf16_f32_fmlal) outside ENABLE_FP16 so ARM+NEON translation
+# units built with ENABLE_FP16=0 can still see them and resolve the
+# definitions at link time against a library built with enable-fp16=true.
+# Keep hgemm.h on the include path regardless of enable-fp16; only the
+# kernel sources themselves stay gated below.
+nntrainer_inc += include_directories('hgemm')
+nntrainer_inc_abs += meson.current_source_dir() / 'hgemm'
```
jaemini-shin, line 31: IMO, it is not a good candidate to support flash attention. We need to rearrange the location of kernel functions.

jaemini-shin (commented):
Here are some comments, PTAL!

nntrainer/tensor/cpu_backend/arm/hgemm/hgemm.cpp
Jungwon-Lee: Fixed the SIGILL risk with a runtime `HWCAP_ASIMDFHM` gate (falls back to `shgemm`). The FMLAL kernel gives ~1.8x on QK, so I'd keep it as an optional fast path and isolate it behind a single dispatch entry point in a follow-up: #4090

nntrainer/tensor/cpu_backend/arm/meson.build
Jungwon-Lee: Agreed — will move the flash kernels into a dedicated `flash_attention/` location as part of the same follow-up: #4090

Applications/CausalLM/layers/mha_core.cpp
Jungwon-Lee: Agreed. Will extract the arch-dependent GEMMs into `cpu_backend` entry points and drop the `#if` branches from `mha_core` in a follow-up (this PR is already large): #4090

Applications/CausalLM/models/transformer.cpp
```diff
@@ -144,6 +144,9 @@ void Transformer::setupParameters(json &cfg, json &generation_cfg,
   FC_LAYER_DTYPE = nntr_cfg["fc_layer_dtype"];
   EMBEDDING_FILE_NAME = nntr_cfg.value("embedding_file_name", std::string());
   PLE_FILE_NAME = nntr_cfg.value("ple_file_name", std::string());
+  USE_FLASH_ATTENTION = nntr_cfg.contains("use_flash_attention")
+                          ? nntr_cfg["use_flash_attention"].get<bool>()
+                          : true;
```
jaemini-shin:
```suggestion
  USE_FLASH_ATTENTION = nntr_cfg.contains("use_flash_attention")
                          ? nntr_cfg["use_flash_attention"].get<bool>()
                          : false;
```
It seems natural to use false as default.

jaemini-shin (approved):
LGTM, I'll work on the following work related to this PR!

PR #4051, opened 2026-07-07
[nntrainer] remove TensorLayer class

haehun:
## Dependency of the PR
None

<details><summary>remove TensorLayer class</summary><br />

TensorLayer was introduced in the early stages of QNN integration as a scratch-tensor holder via context.requestTensor(), but became unnecessary as the QNN integration matured and no persistent consumer remained. WeightLayer is kept as it is required by the ONNX importer (onnx_interpreter.cpp uses createLayerNode("weight") for every model initializer).

Removes tensor_layer.{h,cpp} from nntrainer/layers, drops the LAYER_TENSOR enum value from the CCAPI, and updates packaging manifests and ARCHITECTURE.md accordingly.

**Self evaluation:**

Signed-off-by: haehun.yang <haehun.yang@samsung.com>

</details>

### Summary

- Remove TensorLayer class

Signed-off-by: haehun.yang <haehun.yang@samsung.com>


Jungwon-Lee:
The C API keeps `ML_TRAIN_LAYER_TYPE_TENSOR = 39` as deprecated for ABI, but `LAYER_TENSOR` is removed from the C++ ccapi enum (`api/ccapi/include/layer.h`). For consistency, should we deprecate the C++ enum instead of removing it? (The C enum value can't be fully removed anyway due to ABI.)

haehun:
> The C API keeps `ML_TRAIN_LAYER_TYPE_TENSOR = 39` as deprecated for ABI, but `LAYER_TENSOR` is removed from the C++ ccapi enum (`api/ccapi/include/layer.h`). For consistency, should we deprecate the C++ enum instead of removing it? (The C enum value can't be fully removed anyway due to ABI.)

I think it's better to remove enums than deprecating them if consistency matters. From now on, they cannot be used (and there were no use of TensorLayer anyway)

jaemini-shin (approved):
LGTM!

jayden0701 (approved):
LGTM

PR #4052, opened 2026-07-07
[build] Export nntrainer ABI cflags to prebuilt consumers

jaemini-shin:
## Dependency of the PR

* #4050 

## Commits to be reviewed in this PR

<details><summary>[build] Export nntrainer ABI cflags to prebuilt consumers</summary><br />

Applications linking the prebuilt libnntrainer had to hardcode the
-march and FP16 defines the library was built with; a mismatch breaks
the ABI silently (-DENABLE_FP16/-DUSE__FP16 gate the _FP16 type in the
public tensor_dim.h, and USE__FP16 even changes name mangling).

Make the flags flow from the nntrainer build instead:

- Android-prebuilt.mk becomes a meson template. The generated
  android_build_result/Android.mk now carries the exact
  @MESON_ARM_MARCH@ / FP16 defines the build used and exports them
  via LOCAL_EXPORT_CFLAGS (plus the existing LOCAL_EXPORT_C_INCLUDES),
  so every module that links nntrainer/ccapi-nntrainer inherits them
  automatically.
- CausalLM drops its inline prebuilt definitions, NNTRAINER_INCLUDES
  and the hardcoded -march/-DENABLE_FP16/-DUSE__FP16 in favor of
  including the generated Android.mk. -DUSE_NEON stays app-side: it
  only selects the app's NEON call paths and is referenced by no
  installed nntrainer header.

Verified with build_android.sh: the .text sections of all four
artifacts are byte-identical to the previous hardcoded-flags build.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Signed-off-by: Jaemin Shin <jaemin980311@gmail.com>

</details>

### Summary

This PR makes `CausalLM` builds reuse `libnntrainer`'s compile flags.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Signed-off-by: Jaemin Shin <jaemin980311@gmail.com>

PR #4053, opened 2026-07-07
[CausalLM] Remove app-side USE_NEON flag

jaemini-shin:
## Dependency of the PR
* #4052 

<details><summary>[CausalLM] Remove app-side USE_NEON flag</summary><br />

mha_core.cpp kept two copies of the RoPE table setup: a NEON path
calling nntrainer::calc_trigonometric_vals_dup() under #ifdef USE_NEON
and an inline scalar #else copy, selected by an app-side -DUSE_NEON=1
in Android.mk. That dispatch belongs in nntrainer's cpu_backend, which
already resolves calc_trigonometric_vals_dup() to the NEON kernel on
ARM builds; the only blocker was that the fallback backend threw a
runtime_error instead of computing.

- fallback_internal.cpp: give __fallback_calc_trigonometric_vals_dup a
  real scalar implementation (cos/sin computed once per angle and
  duplicated into both halves in one pass), matching the removed #else
  code
- mha_core.cpp: drop both #ifdef USE_NEON blocks and call
  calc_trigonometric_vals_dup() unconditionally
- Android.mk: drop -DUSE_NEON=1 from CAUSALLM_COMMON_CFLAGS

Android builds keep the exact NEON path they had (same call, resolved
by the ARM compute backend); non-NEON builds now go through the new
fallback (also used by the x86 backend) instead of app-local scalar
loops.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

</details>

### Summary

This PR removes `USE_NEON` flag in `CausalLM` which is only used in calculating RoPE.
The previous logic moved to fallback implementation, so that we could remove branch with `USE_NEON` flag.

**Self evaluation:**

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


dlwlzzero (approved):
LGTM 👍

baek2sm (approved):
LGTM

jayden0701 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #4062, opened 2026-07-08
feat(embedding): mmap sidecar LUT instead of eager full read

haehun:

<details><summary>feat(embedding): mmap sidecar LUT instead of eager full read</summary><br />

Back QuantLut bytes with a read-only mmap of the LUT file (new MappedBytes RAII owner) instead of reading the whole file into a std::vector during EmbeddingLayer::finalize(). Embedding is a sparse row gather by token id, so only the rows the model actually uses fault in on demand during forwarding; MADV_RANDOM disables readahead so the entire LUT is never paged in. This removes the eager LUT read that dominated QNN model load time.

MappedBytes exposes data()/size()/empty(), so all decode consumers and the sidecar unit test are unchanged. Windows (no mmap) falls back to a full read.

**Self evaluation:**

Signed-off-by: haehun.yang <haehun.yang@samsung.com>

</details>

### Summary

- mmap sidecar LUT instead of eager full read

Signed-off-by: haehun.yang <haehun.yang@samsung.com>

### DIFF
--- a/Applications/CausalLM/layers/embedding_layer.cpp
+++ b/Applications/CausalLM/layers/embedding_layer.cpp
@@ -50,10 +50,18 @@ namespace {
 std::mutex quant_lut_cache_mutex;
 std::unordered_map<std::string, std::weak_ptr<QuantLut>> quant_lut_cache;
 
+/**
+ * @brief True if the given path has a ".json" extension (i.e. a LUT
+ * manifest rather than a raw UINT16 LUT file).
+ */
 bool hasJsonExtension(const std::string &path) {
   return std::filesystem::path(path).extension() == ".json";
 }
 
+/**
+ * @brief Resolve a LUT path referenced from a JSON manifest, treating it as
+ * relative to the manifest's directory unless it is already absolute.
+ */
 std::filesystem::path resolveLutPath(const std::string &manifest_path,
                                      const std::string &lut_path) {
   std::filesystem::path path(lut_path);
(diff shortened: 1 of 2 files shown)


Applications/CausalLM/layers/embedding_layer.cpp
```diff
@@ -355,6 +340,108 @@ void decodePacked4BitRowToFloatType(const QuantLut &lut, size_t token_idx,
...
+    map_addr_ = other.map_addr_;
+    owned_ = std::move(other.owned_);
+    other.ptr_ = nullptr;
+    other.size_ = 0;
+    other.map_addr_ = nullptr;
+  }
+  return *this;
+}
+
+MappedBytes MappedBytes::fromVector(std::vector<uint8_t> data) {
```
jaemini-shin:
```suggestion
MappedBytes MappedBytes::fromVector(std::vector<uint8_t> &&data) {
```
This can fully utilize move semantics

Applications/CausalLM/layers/embedding_layer.h
```diff
@@ -65,11 +65,51 @@ class OutputQuantOffset final : public nntrainer::Property<int> {
...
+ *
+ * Embedding is a sparse row gather by token id, so the model only ever touches
+ * a small fraction of a large LUT. Memory-mapping lets those rows fault in on
+ * demand during forwarding instead of eagerly reading the whole file during
+ * layer finalize() — which used to dominate QNN model load time.
+ *
+ * Exposes data()/size()/empty() so it is a drop-in for the std::vector<uint8_t>
+ * it replaces. Move-only (a mapping must have a single owner to munmap once).
+ */
+class WIN_EXPORT MappedBytes {
```
jaemini-shin, line 80: Just for curiosity, how do methods in this class pass doxygen tag check?

jaemini-shin (commented):
Here are some comments PTAL!

Applications/CausalLM/layers/embedding_layer.h
haehun: There might be a bug in doxygen tag check. I'll take a look at it.

haehun: Indeed a bug: https://github.com/nntrainer/nntrainer/pull/4066

jayden0701:
Does this improve the E2E time? I am wondering if this just moves load time from init to inference.

PR #4063, opened 2026-07-08
causallm: resize graph to actual prompt length via resetInputDimension

Jungwon-Lee:
https://github.com/nntrainer/nntrainer/pull/3881

## Summary
- Wire up `NetworkGraph::resetInputDimension()` so `CausalLM::run()` resizes the graph to the real prompt length during prefill and to a single token before the decode loop, instead of keeping a graph sized for a fixed `INIT_SEQ_LEN` for the whole session.
- Add/fix `updateTensorsByInputDimensions()` overrides across the layers used in CausalLM graphs (fc, embedding, mha_core, activation, rms_norm variants, scalar_multiply, tie_word_embedding, logit_softcapping, and all 5 MoE layer variants) so each correctly resizes its own tensors instead of falling back to the base class's unconditional throw, or naively overwriting tensor width with the broadcast hidden_size-based dimension when its true width differs (per-head tensors, GQA KV width, lm_head's fixed height-1 output, etc).
- Fix a pre-existing bug where prompts `>= INIT_SEQ_LEN` silently dropped the first generated token.
- `network_graph`: exclude `InputLayer` from the `resetInputDimension` broadcast, since its raw token-id tensor convention differs from every other layer's height-based one.

## Commits
Ordered so every commit up to "wire up resetInputDimension" only adds inert per-layer support (the call site isn't wired up yet), and the switch is flipped last, once every layer type used by a CausalLM graph can handle it:
1. `fc_layer`: add height-only `updateTensorsByInputDimensions`
2. `causallm/embedding_layer`: add height-only `updateTensorsByInputDimensions`
3. `causallm/mha_core`: stop resizing the KV cache in `updateTensorsByInputDimensions`
4. `network_graph`: exclude `InputLayer` from the broadcast, always allocate
5. `activation_layer`: add height-only `updateTensorsByInputDimensions`
6. `causallm/moe`: add `updateTensorsByInputDimensions` across all 5 MoE layer variants (router_logits/expert_mask resize)
7. `causallm`: wire up `resetInputDimension()` before prefill, force KV cache rebind — the actual switch
8. `test/causallm`: end-to-end `runPrompt()` regression coverage for Qwen2/Qwen3/Gemma3/Gemma4
9. `causallm`: stop dropping the first prefill token for long prompts
10. `causallm`: resize graph to a single token before the decode loop
11. `causallm`: preserve tensor width in `updateTensorsByInputDimensions` overrides (fc/activation/rms_norm variants/scalar_multiply/mha_core/tie_word_embedding/logit_softcapping)

## Test plan
- [x] `unittest_nntrainer_graph.cpp`: `InputLayer` exclusion from the broadcast
- [x] `unittest_layers_fully_connected.cpp` / `unittest_layers_activation.cpp`: height-only resize
- [x] `unittest_causallm_moe_resetinputdim.cpp`: end-to-end `runPrompt()` grow/shrink for Qwen3MoE
- [x] `unittest_causallm_run_prompt_resetinputdim.cpp`: end-to-end `runPrompt()` grow/shrink for Qwen2/Qwen3/Gemma3/Gemma4
- [x] Full `unittest_causallm_models` suite passes

Known follow-ups noted in commit messages (out of scope here):
- `GptOssMoELayer`: allowing `resetInputDimension` past the MoE layer exposes a pre-existing SIGSEGV in GptOss's sliding-window attention path; not covered by the new tests.
- `SlimMoELayer` / `CachedSlimGptOssMoELayer`: not covered by the new end-to-end test because their lazy mmap-based expert loading hangs in the in-process tiny-test harness; verified instead by code inspection against the identical `MoELayer` pattern.

### DIFF
--- a/Applications/CausalLM/layers/embedding_pooling_layer.cpp
+++ b/Applications/CausalLM/layers/embedding_pooling_layer.cpp
@@ -152,6 +152,19 @@ void EmbeddingPoolingLayer::incremental_forwarding(
   }
 }
 
+void EmbeddingPoolingLayer::updateTensorsByInputDimensions(
+  nntrainer::RunLayerContext &context,
+  std::vector<nntrainer::TensorDim> input_dimensions) {
+  // Input height is the current sequence length; output stays fixed at
+  // height 1 (pooled to a single vector per finalize()), so only the input
+  // needs to track the new step size.
+  unsigned int height = input_dimensions[0].height();
+
+  nntrainer::TensorDim input_dim = context.getInput(SINGLE_INOUT_IDX).getDim();
+  input_dim.height(height);
+  context.updateInput(SINGLE_INOUT_IDX, input_dim);
+}
+
 void EmbeddingPoolingLayer::calcDerivative(
   nntrainer::RunLayerContext &context) {
   throw nntrainer::exception::not_supported(
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -493,11 +493,13 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   };
   input = build_inference_inputs();
 
-  ///@note contains possible bug
-  // std::vector<ml::train::TensorDim> input_dims;
-  // ml::train::TensorDim input_dim(1, 1, input_len, DIM);
-  // input_dims.push_back(input_dim);
-  // model->resetInputDimension(input_dims);
+  std::vector<ml::train::TensorDim> input_dims;
+  input_dims.push_back(
+    ml::train::TensorDim(1, 1, input_len, static_cast<unsigned int>(DIM)));
+  model->resetInputDimension(input_dims);
+  // resetInputDimension() silently rebinds KV cache placeholders back to
+  // internal memory; force allocateAndBindKVCache() to redo external binding.
+  kv_cache_bound = false;
 
   auto start_prefill = std::chrono::high_resolution_clock::now();
 
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -575,9 +575,7 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
 
     // post process of model output
     id_list = generate(output[0], do_sample, 1, ids_history, init_len);
-
-    if (init_len < INIT_SEQ_LEN)
-      registerOutputs(tokenizer, id_list, init_len, eos_list, log_output);
+    registerOutputs(tokenizer, id_list, init_len, eos_list, log_output);
   }
   // output should be deallocated after use
   for (auto &out : output) {
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -599,6 +599,12 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
 
   auto start_generation = std::chrono::high_resolution_clock::now();
 
+  std::vector<ml::train::TensorDim> decode_input_dims;
+  decode_input_dims.push_back(
+    ml::train::TensorDim(1, 1, 1, static_cast<unsigned int>(DIM)));
+  model->resetInputDimension(decode_input_dims);
+  kv_cache_bound = false;
+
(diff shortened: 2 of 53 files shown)


Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -575,9 +575,10 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
 
     // post process of model output
     id_list = generate(output[0], do_sample, 1, ids_history, init_len);
-
-    if (init_len < INIT_SEQ_LEN)
-      registerOutputs(tokenizer, id_list, init_len, eos_list, log_output);
+    // The old init_len < INIT_SEQ_LEN guard silently dropped the first
+    // token for prompts >= INIT_SEQ_LEN; no longer needed with
+    // resetInputDimension().
+    registerOutputs(tokenizer, id_list, init_len, eos_list, log_output);
```
Jungwon-Lee, line 578: The old init_len < INIT_SEQ_LEN guard silently dropped the first token for prompts >= INIT_SEQ_LEN; no longer needed with resetInputDimension().

nntrainer/graph/network_graph.cpp
```diff
@@ -338,13 +338,14 @@ void NetworkGraph::resetInputDimension(std::vector<TensorDim> dims) {
...
+    /// @note InputLayer is excluded: its raw input tensor doesn't follow the
+    /// height-based convention `dims` uses for every other layer.
+    if ((*iter)->isFinalized() && (*iter)->getType() != InputLayer::type) {
       (*iter)->updateTensorsByInputDimensions(dims);
     }
   }
 
-  if (allocated)
-    allocateTensors(exec_mode);
+  allocateTensors(exec_mode);
```
jaemini-shin, line 348: Did you check the overhead of deallocation and allocation? If it takes long, we should introduce some additional optimizations.

jayden0701:
Does it cover scenarios in multiturn where |turn2 prefill length| > |initial prefill length| ? 
Because the old version of resetInputDimensions will fail in that case

nntrainer/graph/network_graph.cpp
Jungwon-Lee: It introduces minor reallocation overhead (~5ms), 
But, it can reduce the initialization time to allocate initial activation tensor pool, because INIT_SEQ_LEN become constant (we use 1).

Jungwon-Lee:
> Does it cover scenarios in multiturn where |turn2 prefill length| > |initial prefill length| ? Because the old version of resetInputDimensions will fail in that case

@jayden0701 It works fine even when the second turn's prefill is longer than the first one. 
No issues there!

nntrainer/graph/network_graph.cpp
jaemini-shin: > It introduces minor reallocation overhead (~5ms), But, it can reduce the initialization time to allocate initial activation tensor pool, because INIT_SEQ_LEN become constant (we use 1).

Does it mean initialization uses `INIT_SEQ_LEN == 1` and reallocation uses `INIT_SEQ_LEN` from config file? If so, it'd be better to initialize with real `INIT_SEQ_LEN` from config file.

Jungwon-Lee: Initialization uses `INIT_SEQ_LEN`, but we can set `INIT_SEQ_LEN=1` from the config file.
Then, the model can dynamically reallocate the prompt length.

Applications/CausalLM/layers/embedding_pooling_layer.cpp
```diff
@@ -152,6 +152,19 @@ void EmbeddingPoolingLayer::incremental_forwarding(
   }
 }
 
+void EmbeddingPoolingLayer::updateTensorsByInputDimensions(
```
jaemini-shin, line 155: Is this layer regarded as input layer? If not, since its input is output of precedent layer, we should not modify input dimension

jaemini-shin: Some other layers also update input dimension, too. Could you please check this?

Jungwon-Lee: > Is this layer regarded as input layer? If not, since its input is output of precedent layer, we should not modify input dimension

You're right, the input is logically the output of the preceding layer.

But, although they share the same memory buffer (zero-copy), they maintain independent separate Tensor objects with `dim`/`strides` metadata. Therefore, `updateInput()` must be called explicitly to keep the input Tensor's dimension in sync.

PR #4060, opened 2026-07-08
[test] enrich kleidiai benchmark

jaemini-shin:
## Commits to be reviewed in this PR

<details><summary>[test] enrich kleidiai benchmark</summary><br />

- Exclude the benchmark tests from the default run via the
  DISABLED_ prefix and add a --bench shortcut flag that composes
  with --gtest_filter
- Evict the packed RHS from the whole cache hierarchy (dc civac)
  before every iteration, so the benchmark measures production-like
  cold weights; the LHS stays warm on purpose since activations are
  produced right before the run
- Fix out-of-bounds writes to the reserved total_time_qai8dxp vector

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Signed-off-by: Jaemin Shin <jaemin980311@gmail.com>

</details>

### Summary

This PR updates kleidiai benchmark to get more accurate result to real use case.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Signed-off-by: Jaemin Shin <jaemin980311@gmail.com>

PR #4065, opened 2026-07-09
[nntrainer] optimize qai8dxp_qsi4cxp GEMM with cache-aware tiling

jaemini-shin:

<details><summary>[nntrainer] optimize qai8dxp_qsi4cxp GEMM with cache-aware tiling</summary><br />

Replace the m_step/n_step one-dimensional split heuristic in
__kai_gemm_qai8dxp_qsi4cxp with cache-aware 2D tiling, assuming a
fixed 1 MiB per-core L2:

- n_chunk: the micro-kernels iterate row-blocks outer / columns
  inner, so the RHS panel is re-read every m_step rows. Size it to
  an even split of n whose panel fits in 2/3 of L2.
- m_chunk: sized for ~16 tasks per thread so work stealing can
  balance heterogeneous cores, and capped so the LHS panel reused
  by consecutive tiles stays L2-resident beside the RHS panel,
  trading oversized panels for more tasks.
- Tiles are walked row-major; small m splits n finer to keep all
  cores fed, and m <= m_step keeps the previous GEMV n-split path.

Measured on 8 threads, M=1024 N=2560 K=4096, variant 8 (i8mm
8x8x32), interleaved A/B with 60s cooldowns: 10.99ms -> 10.21ms
median (-7.2%, 6/6 wins) with 3x lower run-to-run variance.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

</details>

## Summary

This PR improves heuristics for qai8dxp_qsi4cxp GEMM especially for the fastest kernel 8(matmul_clamp_f32_qai8dxp4x8_qsi4cxp8x8_8x8x32_neon_i8mm).

The following results are measured on Galaxy S26+.

### kernel performance
Since kernel level benchmark on main branch is not valid, I used the #4060 which resolved this issue.
* 1024x2560x4096 (ukernel 3~8)
 

| Kernel | optimized (ns) | PR 4060 (ns) |
|--------|-----------|-------------|
| ukernel[3] matmul_clamp_f32_qai8dxp4x4_qsi4cxp8x4_8x8x32_neon_dotprod | 17692640 | 17855062 |
| ukernel[4] matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x4_16x4x32_neon_dotprod | 18724320 | 19493789 |
| ukernel[5] matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_4x4x32_neon_i8mm | 14166047 | 15385062 |
| ukernel[6] matmul_clamp_f32_qai8dxp4x8_qsi4cxp4x8_8x4x32_neon_i8mm | 11728945 | 13745500 |
| ukernel[7] matmul_clamp_f32_qai8dxp4x8_qsi4cxp8x8_4x8x32_neon_i8mm | 12334203 | 12401266 |
| ukernel[8] matmul_clamp_f32_qai8dxp4x8_qsi4cxp8x8_8x8x32_neon_i8mm | 10184508 | 11624187 |

### model performance
* qwen3-0.6b, 1024 input, 512 output

|  | optimized (TPS) | main (TPS) |
|--------|-----------|-------------|
| Prefill | 581.158 | 570.474 |
| Decode | 53.8267 | 49.0234 |

**Self evaluation:**

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi4cxp.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi4cxp.cpp
@@ -32,6 +32,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+#include <algorithm>
+
 #include <kleidiai_interface.h>
 #include <thread_manager.h>
 
@@ -254,19 +256,11 @@ void __kai_gemm_qai8dxp_qsi4cxp_rhs_unpacked(
   size_t m, size_t n, size_t k, void *lhs_native_mtx_f32,
   void *rhs_native_mtx_qs4cx, void *rhs_scales_f32, float *dst_act_mtx_f32,
   size_t idx_variant, bool is_nxk, float lower_bound, float upper_bound) {
-  const size_t mr = ukernel_variants[idx_variant].ukernel.get_mr();
-  const size_t nr = ukernel_variants[idx_variant].ukernel.get_nr();
-  const size_t kr = ukernel_variants[idx_variant].ukernel.get_kr();
-  const size_t sr = ukernel_variants[idx_variant].ukernel.get_sr();
-
   // Get the size in bytes for the packed matrices
-  const size_t lhs_packed_size =
-    kai_get_lhs_packed_size_lhs_quant_pack_qai8dxp_f32(m, k, mr, kr, sr);
   const size_t rhs_packed_size =
     __kai_get_rhs_packed_size_qsi4cxp_qs4cxs1s0(n, k, idx_variant, is_nxk);
 
   // Allocate the matrices
-  uint8_t *lhs_packed_mtx_qa8dx = new uint8_t[lhs_packed_size];
   uint8_t *rhs_packed_mtx_qs4cx = new uint8_t[rhs_packed_size];
 
   // RHS packing
@@ -278,7 +272,6 @@ void __kai_gemm_qai8dxp_qsi4cxp_rhs_unpacked(
   __kai_gemm_qai8dxp_qsi4cxp(m, n, k, lhs_native_mtx_f32, rhs_packed_mtx_qs4cx,
                              dst_act_mtx_f32, idx_variant);
 
-  delete[] lhs_packed_mtx_qa8dx;
   delete[] rhs_packed_mtx_qs4cx;
 }
 
@@ -288,7 +281,6 @@ void __kai_gemm_qai8dxp_qsi4cxp(size_t m, size_t n, size_t k,
                                 float *dst_act_mtx_f32, size_t idx_variant,
                                 float lower_bound, float upper_bound) {
   const size_t mr = ukernel_variants[idx_variant].ukernel.get_mr();
-  const size_t nr = ukernel_variants[idx_variant].ukernel.get_nr();
   const size_t kr = ukernel_variants[idx_variant].ukernel.get_kr();
   const size_t sr = ukernel_variants[idx_variant].ukernel.get_sr();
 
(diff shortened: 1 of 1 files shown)


jayden0701 (approved):
LGTM

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi4cxp.cpp
```diff
@@ -288,7 +290,6 @@ void __kai_gemm_qai8dxp_qsi4cxp(size_t m, size_t n, size_t k,
                                 float *dst_act_mtx_f32, size_t idx_variant,
                                 float lower_bound, float upper_bound) {
   const size_t mr = ukernel_variants[idx_variant].ukernel.get_mr();
-  const size_t nr = ukernel_variants[idx_variant].ukernel.get_nr();
```
Seunghui98, line 291: Why do you remove `nr`?
Is it unused? 🤔

Seunghui98 (commented):
## Description

Scope is a single function: `__kai_gemm_qai8dxp_qsi4cxp` in `kleidiai_interface_qai8dxp_qsi4cxp.cpp`. The old `m_step > n_step` one-dimensional split heuristic is replaced by (a) an always-separate LHS quantize+pack pass, (b) an `m <= m_step` GEMV fast path that splits only over `n`, and (c) a cache-aware 2D tiling path for everything else, sized against a fixed 1 MiB per-core L2.

## Summary

- The tiling math looks sound. `m_chunk`/`n_chunk` are integer multiples of `m_step`/`n_step`, so every `get_lhs_packed_offset` / `get_rhs_packed_offset` / `get_dst_offset` call lands on a legal index for all 9 variants (checked: every variant has `m_step % mr == 0` and `n_step % nr == 0`). Tiles are disjoint in both dimensions, and `K` is never split, so results stay bit-identical to `main` — this is a scheduling-only change.
- No out-of-range tiles: `(m_tiles-1) * m_chunk < m` and `(n_tiles-1) * n_chunk < n` hold from the `ceil_div` derivations, so `m_to_process`/`n_to_process` are always positive.
- No division by zero: `getComputeThreadCount()` returns `workers + 1 >= 1`, and the `m_block_bytes` divisor is guarded by `rhs_panel_bytes < l2_cache_bytes` plus `std::max(..., 1)`.
- Nice side effect worth noting: for variant 4 (`16x4x32`, `m_step=16 > n_step=4`), decode (`m=1`) previously took the parallelize-over-m branch with `m_loop == 1`, i.e. it ran **single-threaded**. The new `m <= m_step` path makes it n-parallel. That likely explains a good part of the decode TPS gain, independent of the tiling itself.

## Bug

No functional bug found in the changed range. The concerns below are performance/portability, not correctness.

1. **Little cores vs. the hardcoded 1 MiB L2.** `l2_cache_bytes` is a compile-time constant, but `ThreadManager` pins workers across heterogeneous cores. On big.LITTLE parts the little-core L2 is often 128–512 KiB, so a tile sized for 1 MiB will thrash there while the big cores are fine. Since the tile size is uniform across threads, the little cores set the tail latency that the 16-tasks-per-thread work stealing is meant to absorb. Worth (a) at least naming it as a tunable/configurable constant rather than a bare `constexpr` in the function body, and (b) measuring on a device whose little-core L2 is clearly below 1 MiB.

2. **`n_chunk` can collapse to a single `n_step` and cause dst false sharing.** In the "increase n block to reach the task target for small m" branch, if `m_tiles` is small and `n_loop <= target_tasks`, you get `n_tiles == n_loop`, `n_blocks == 1`, so `n_chunk == n_step`. For the `n_step = 4` variants that is a 16-byte column slice, meaning up to 4 threads write into the same 64-byte cache line of every `dst` row. Concrete shape: `m = m_step + 1`, `n = 512`, variant 5/6 (`n_step = 4`), 8 threads. Correctness is unaffected, but the write-sharing may eat the gain. A floor on `n_chunk` at one cache line (`n_chunk * sizeof(float) >= 64`) would remove the risk cheaply.

3. **The new barrier between packing and matmul for large-M / small-N.** The old `m_step > n_step` branch fused packing and matmul in one `parallel_for`, so each thread consumed the rows it had just packed while they were still hot in its own L2. Now there is always a full barrier plus a round trip through the packed LHS buffer. The reported M=1024 N=2560 numbers cover the wide case; a tall-and-narrow shape (e.g. M=2048, N=512, K=4096) is where the fused path used to win, so it would be good to confirm no regression there.

4. **Small-K shapes.** With small `K`, `n_block_bytes` shrinks enough that `n_tiles` clamps to 1 and the whole RHS becomes one panel, pushing all parallelism onto `m`. Worth a quick sanity run at something like K=128 to confirm the task count and tile shape stay reasonable.

5. **Minor / optional:** `delete[] lhs_packed_mtx_qa8dx` is now duplicated across two return paths. A `std::unique_ptr<uint8_t[]>` would make the early return leak-proof against future edits.

Overall the direction and the measurements look convincing to me — items 1 and 2 are the ones I would want addressed or explicitly ruled out before merge.

---

Seunghui98 (approved):
LGTM

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi4cxp.cpp
jaemini-shin: Yes, `nr` is unused.

`nr` denotes the packed sized in `n` dimension of weight(rhs). Since weight has been packed before gemm, this variable is unused!

PR #4071, opened 2026-07-10
fix: prevent EOS/EOT token leakage in CPU generation path

jayden0701:
## Summary

Prevent EOS/EOT tokens such as `<turn|>` from leaking into decoded CPU-generation output, without changing token history or KV-cache/generation accounting.

## Root cause

`CausalLM::registerOutputs()` decoded every sampled token before the generation loop marked EOS. Moving the EOS-state update before `registerOutputs()` hid the token, but it also skipped `ids_history` registration and moved `generation_cnt` after the all-EOS break. The latter made `global_token_len` one cache step short because `incremental_inference()` had already consumed the previous input token in that iteration.

## Changes

- Keep EOS handling inside `registerOutputs()`'s decode/output path:
  - record EOS in `ids_history`;
  - do not add EOS itself to `pending_ids_` or decoded output;
  - flush any complete text that was pending immediately before EOS;
  - discard an incomplete UTF-8 fragment at end-of-stream.
- Update `eos_list` only after output/history registration.
- Increment `generation_cnt` before the all-EOS break, matching the cache step already completed by `incremental_inference()`.
- Add a Qwen2 regression test that forces one normal token followed by EOS and verifies:
  - decoded output contains only the normal token (`world`);
  - both the normal token and EOS remain in token history;
  - generation/cache-step accounting remains `1`.

### DIFF
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -245,32 +245,45 @@ void CausalLM::registerOutputs(
 
   static const std::vector<char> puncts{',', '!', ':', ';', '?'};
   for (size_t b = 0; b < ids.size(); ++b) {
-    if (!eos_list[b]) {
+    if (eos_list[b])
+      continue;
+
+    ids_history[b * MAX_SEQ_LEN + pos] = ids[b];
+
+    const bool is_eos = std::find(EOS_TOKEN_ID.begin(), EOS_TOKEN_ID.end(),
+                                  ids[b]) != EOS_TOKEN_ID.end();
+    if (!is_eos)
       pending_ids_.push_back(static_cast<int>(ids[b]));
-      ids_history[b * MAX_SEQ_LEN + pos] = ids[b];
-      std::string decoded_str = tokenizer->Decode(pending_ids_);
 
-      if (decoded_str.empty()) {
-        continue;
-      }
+    if (pending_ids_.empty())
+      continue;
 
-      if (std::find(puncts.begin(), puncts.end(), decoded_str.back()) !=
-          puncts.end()) {
-        // last symbol is a punctuation, hold on
-      } else if (utf8stream::shouldHold(decoded_str, pending_ids_.size())) {
-      } else {
-        if (log_output && streamer_ == nullptr) {
-          std::cout << decoded_str;
-          std::cout.flush();
-        }
-        output_list[b].append(decoded_str);
-        if (streamer_ != nullptr &&
-            streamer_put(streamer_, decoded_str.c_str()) != 0) {
-          requestStop();
-        }
+    const std::string decoded_str = tokenizer->Decode(pending_ids_);
+    const bool should_hold_utf8 =
+      decoded_str.empty() ||
+      utf8stream::shouldHold(decoded_str, pending_ids_.size());
+    if (should_hold_utf8) {
+      if (is_eos)
         pending_ids_.clear();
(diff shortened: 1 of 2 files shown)


Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -644,9 +653,16 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
...
     if (is_finish) {
+      // Do NOT increment generation_cnt for the EOS token: its K/V was never
+      // written to the cache (incremental_inference processes the *input*
+      // token, not the sampled output). Keeping generation_cnt in sync with
+      // actual cache contents ensures global_token_len stays correct for
+      // multi-turn conversations — no extra forward pass needed.
       break;
     }
 
+    ++generation_cnt;
```
jaemini-shin: Don't we have to count EOS tokens in generation_count? We just skip the printing EOS token, not exclude it from generated tokens.

Jungwon-Lee:
@jayden0701 The code works well, but from a code quality perspective, it is a bit difficult to understand.
Could you consider refactoring `registerOutput `to improve readability and maintainability?

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -576,6 +588,15 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
...
     if (init_len < INIT_SEQ_LEN)
       registerOutputs(tokenizer, id_list, init_len, eos_list, log_output);
+
+    // Mark EOS after output registration so the sampled token remains in
+    // history while registerOutputs() suppresses it from decoded output.
+    for (unsigned int j = 0; j < BATCH_SIZE; ++j) {
+      if (std::find(EOS_TOKEN_ID.begin(), EOS_TOKEN_ID.end(), id_list[j]) !=
+          EOS_TOKEN_ID.end()) {
+        eos_list[j] = true;
+      }
```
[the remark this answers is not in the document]
jaemini-shin: If expected token from prefill is EOS, we should skip the decode stage but it doesn't.

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -246,18 +246,30 @@ void CausalLM::registerOutputs(
   static const std::vector<char> puncts{',', '!', ':', ';', '?'};
   for (size_t b = 0; b < ids.size(); ++b) {
     if (!eos_list[b]) {
-      pending_ids_.push_back(static_cast<int>(ids[b]));
       ids_history[b * MAX_SEQ_LEN + pos] = ids[b];
+
+      const bool is_eos = std::find(EOS_TOKEN_ID.begin(), EOS_TOKEN_ID.end(),
```
jaemini-shin: Is it necessary to suppress printing EOS? We have already been printing special tokens such as `<im_start>`, but suppressing EOS tokens only seems strange.

jaemini-shin (commented):
Could you explain which use case does this PR try to resolve? In this PR, skip printing EOS tokens is the only change, and I couldn't figure out the purpose of this change.

Applications/CausalLM/models/causal_lm.cpp
jayden0701: It was a request from the users of QuickAI. 
And QNN runs also suppress EOS tokens. So I thought doing the same for cpu models is reasonable.

PR #4081, opened 2026-07-15
[HTP] Add Hexagon NPU (HexKL) backend: WH-baked prefill, NPU decode routing

dlwlzzero:
## Dependency of the PR

None — this is the full HexKL/HTP NPU backend integration, branched from main at da736573.

<details><summary>[HTP] Add backend skeleton and HtpContext/ComputeOps infrastructure</summary><br />

- Add meson options: enable-htp, hexkl-sdk-root, hexkl-lib-subdir
- Export HexKL SDK includes via hexkl_inc; wire HTP backend into Android
  ndk-build (jni/)
- Add ARM architecture flags for Oryon CPU (nosve+nosve2)
- Expose CpuComputeOps via cpu_ops_table.h for backend inheritance
- Make HtpComputeOps inherit from CpuComputeOps for CPU fallback
- Add HtpContext: engine-registered context with HTP ops; register under
  engine="htp" in Engine singleton
- Gate shgemm call site with supports_shgemm() predicate

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Implement shgemm_f32f16_f32 and prefill/decode routing</summary><br />

- Implement shgemm_f32f16_f32 via sdkl_npu_mm_f32f16_f32; reject beta!=0
  (NPU write-only, no accumulation)
- Fix TStorageOrder bug and enforce M%32/K%32/N%32 constraints; add sdkl
  probes
- Add HtpMemAllocator (sdkl_npu_alloc) and install it in the context; pad
  M to a multiple of 32 inside shgemm_f32f16_f32
- Cache WH-converted weights as a fallback when weights are not
  NPU-resident
- Route decode (M=1) GEMV to NPU shgemm when supported; route prefill to
  NPU and decode to CPU, bypassing WHCache for M>1
- Read compute_engine from nntr_config and guard FSU; inject
  compute_engine into Qwen3 attention and lm_head FCs
- Fix SVE detection for Oryon CPU (nosve+nosve2)
- Add unittest_nntrainer_htp_backend (accuracy, fallback, e
  kernel accuracy/perf tests, and a CPU-vs-HTP greedy token compare
  driver

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Add QINT8 u8i8 quantized path and Cbr />

- Add shgemm_u8i8 interface and hexkl_mm declaration; add getZpCorr
  accessor to CharTensor and extend memory layout
- Add QINT8 weight quantization and HTP dispatch; implement
  shgemm_u8i8_i32 with M-padding and WH layout support
- Fix per-channel scale indexing for QINT8 weights; wire shgemm_u8i8
  override in HtpComputeOps
- Add compute_engine support to ConfigModifier and a --comp
  sweep option to benchmark_android.py
- Route QINT8 dotQInteger to NPU when M%64==0; tighten u8i8
  kernel guards
- Add HtpU8i8Test and HtpDispatchTest; add QINT8 weight ser
  roundtrip coverage
- Update CausalLM models/layers for quantized inference; update Android
  benchmark and install scripts

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[nntrainer] Update build configuration and add HTP integration docs</summary><br />

- Update .gitignore for HTP backend artifacts; add base_pro
  extensions for compute_engine; update CPU backend build infra for
  quantization
- Add HTP backend integration guide (build, environment, se
  and backend docs (operator support, memory layout, caching)
- Add E2E performance results and unit test guide, GEMINI.md, and a
  u8i8 proof-of-concept design handoff doc
- Update CLAUDE.md with architecture and implementation details

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Add persistent WH residency cache wbr />

- Per-call RM->WH weight staging dominated TTFT on every pr
  add PrefillWHCache mapping each RM weight pointer to its NPU-resident
  WH buffer, populated once (at warmup) and reused for the
  lifetime
- Real prefill calls become lookup -> sdkl_npu_mm_f32f16_f32 ->
  copy-back; a transient fallback (identical to the old pat
  alloc failures
- Add prefillWHCacheSize()/prefillWHCacheClear() introspection hooks and
  a cache-reuse-across-calls test
- Add an opt-in warmup_prefill config that runs one throwaway prefill
  forward before the first timed prefill so the cache is populated ahead
  of the measured region

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Remove HtpMemAllocator; fix Android NDK r26d build</summary><br />

- HtpMemAllocator allocated all model tensors (weights, act
  NPU DMA memory, exhausting the DSP mapping pool and causing
  sdkl_npu_free Err=1 and crashes; remove it so only per-call transient
  buffers and the decode WHCache (64 MB cap) consume NPU DM
- Convert PrefillWHCache to a FIFO-eviction struct (cap 128 MB), kept
  disabled pending precise pool-budget characterization
- Add FP16 weight save support in layer_devel.h (clone FP32
  before serializing), matching the existing QINT8 path
- Fix Android cross-build for NDK r26d/API 26: replace
  std::aligned_alloc (unavailable at minSdkVersion 26) with
  posix_memalign; add -llog and -landroid link deps

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Replace evicting WH cache with pin-once residency</summary><br />

- Add an NPU DMA pool-size probe test (allocate 4 MB chunks
  failure) since the HexKL SDK exposes no pool-query API; measures the
  usable residency budget that gates the prefill weight pin cap
- The FIFO-eviction cache thrashed to ~0% hits (prefill tou
  distinct weights once each, evicting before reuse); replace it with
  pin-once: convert+pin each weight until a byte cap, then fall back to
  the transient path with no eviction/thrashing; return nullptr instead
  of throwing on exhaustion
- Add per-phase transient prefill timing test and a sustained-pin NPU
  pool probe test
- Update the HTP backend guide for pin-once residency: document the
  measured pool budget (B_transient=4000 MB, B_raw=448 MB,
  B_effective~48 MB) and the updated routing table; HtpKernelTest suite
  grew from 13 to 18 tests

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Add pre-baked WH registry and scratill</summary><br />

- Add a host-side registry (registerPrefillWH/lookupPrefillWH) holding
  already-WH-layout FP16 weight bytes keyed by the RM weight pointer; in
  the M>1 prefill branch a registry hit stages the WH bytes into NPU
  memory with a plain memcpy and skips rm_to_wh, a miss fal
  the existing pin-cache/transient conversion
- This is the runtime half of offline WH baking; host-resid
  sidesteps the ~48 MB NPU pin cap
- Grow X/A/W staging buffers on demand and reuse them acros
  prefill matmuls instead of alloc/free per call

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Offline WH bake: byte-identity test and trailer writer</summary><br />

- Add a byte-identity test confirming sdkl_cpu_rm_to_wh_f16_inplace is
  deterministic and produces identical WH bytes from host and NPU
  buffers for the three prefill FC shapes; this is the gate
  an offline-baked WH weight bit-for-bit substitutable at runtime
- Add a self-describing WH trailer (wh_trailer.{h,cpp}) written after
  the normal weight data when WH bake is requested; each 32
ed RM->WH via sdkl_cpu_rm_to_wh_f16_inplace
  and stored keyed by weight name, with RM bytes unchanged so non-HTP
  readers ignore the trailer
- Footer format is [u64 trailer_bytes][u32 count]["WHF1"] so a reader
  can locate and skip it from EOF; codec is unit-tested (round-trip +
  plain-data rejection)

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Load WH trailer at model load; fix bake-gate dtype resolution and exception safety</summary><br />

- After the BIN weight read, when HTP is active, parse the optional WH
  trailer and register each entry's WH bytes against the matching
  loaded weight (by name) via registerPrefillWH, so FP16 prefill hits
  the no-rm_to_wh fast path; best-effort — an absent/mismatched trailer
  leaves the transient path intact
- Fix NeuralNetwork::save()'s MODEL_FORMAT_BIN case to reso
  layer's dtype from layer_dtype_map (falling back to the global dtype)
  instead of hardcoding FP16 once the WH-bake branch fired, so
  --fc_dtype/--embd_dtype/--lmhead_dtype per-layer override
  respected; add a regression test covering a mixed-dtype 2-FC model
- Make the WH-bake collector reset exception-safe: wrap g_wh_collector
  in an RAII guard so a throwing layer_node->save() can't leave it
  dangling, and make the WHBakeGate test's own flag reset and file
  cleanup exception-safe (gtest ASSERT_* early-returns previously
  skipped both)

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Wire FP16_WH bake into nntr_quantize; fix INFERENCE-mode registration; record E2E speedup</summary><br />

- Enable ENABLE_HEXKL for the nntr_quantize ndk-build target:
  nntr_quantize's --fc_dtype FP16_WH flag calls
  nntrainer::setWHBakeRequested(), declared only under ENABLE_HEXKL, but
  the Android.mk path never defined this flag, so the guarded call
  compiled out silently; scope -Drestrict=__restrict, Hexag
  include paths, and a direct libsdkl.so link to the nntr_quantize
  target only
- Move the WH-trailer registration block in NeuralNetwork::load()'s BIN
  case to run unconditionally after the INFERENCE/non-INFERENCE branch —
  it previously lived only in the non-INFERENCE branch, but
  Applications/CausalLM always loads with ExecutionMode::INFERENCE, so
  registerPrefillWH() was never called in the one real call
  feature exists for; add a load()-level regression test
- nntr_quantize --fc_dtype FP16_WH now emits an integrated
  weights + WH trailer); on S25 Ultra, loading it through
  nntrainer_causallm (INFERENCE mode) registers all 196/196 pre-baked WH
  weights, and FP16 prefill drops from 3178 ms (7.24 TPS) t
  (239.58 TPS) on a 23-token prompt — a 33.1x speedup — with decode
  throughput unchanged; peak RSS grows by ~1.62 GiB and the integrated
  bin by ~880.8 MB versus the RM-only baseline
- Fix two ~2% GiB-rounding errors in the memory table and note the
  single-run-per-configuration measurement caveat
- Free each WH-trailer entry's bytes immediately after regi
  instead of keeping the whole ~880 MB trailer buffer alive for the
  entire registration loop, removing an avoidable duplicate-memory peak
  that directly inflated the feature's measured RSS cost

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Skip sdkl_npu_free after NPU finalize (teardown Err=1)</summary><br />

HtpBackend (a function-local static) is destroyed before the
namespace-scope NPU scratch/cache statics, so their destructors called
sdkl_npu_free on an already-finalized NPU, printing Err=1 x3 at exit.
Guard every teardown free with a plain npuAlive() flag that HtpBackend
lowers immediately before sdkl_npu_finalize; the finalize reclaims the
memory, so skipping the frees is safe, not a leak.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Fix consumer build breakage: ENABLE_FP16=0 and missing header install</summary><br />

- shgemm_u8i8_i32's cleanup() (compiled even without ENABLE
  npuFreeIfAlive, but the helper was defined inside the ENABLE_FP16
  block, breaking the ENABLE_HEXKL=1, ENABLE_FP16=0 int8 configuration;
  define it once at the top of hexkl_mm.cpp under ENABLE_HE
  ENABLE_FP16, and hoist <remote.h> alongside it since sdkl.h needs
  CDSP_DOMAIN_ID/CDSP1_DOMAIN_ID
- layer_devel.h is an installed public header that #includes
  <wh_trailer.h> under ENABLE_HEXKL, but wh_trailer.h was never added to
  install_headers, so consumers building against installed headers (e.g.
  Applications/CausalLM ndk-build) failed with a missing-file error; add
  it to the enable-htp header install block alongside htp_context.h

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Add pre-baked WH diagnostics: disable + runtime verify toggles</summary><br />

NNTR_HTP_DISABLE_PREBAKED_WH=1 forces lookupPrefillWH to miss so prefill
uses the known-good transient RM->WH path (escape hatch / bisection aid).
NNTR_HTP_VERIFY_PREBAKED_WH=1 recomputes the transient conv
pre-baked hit and memcmps it against the registered bytes, logging OK /
MISMATCH per weight to stderr.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Fix weight-layout mismatch in shgemm NPU path (garbage generation)</summary><br />

FC-family layers (fc_layer.cpp, shared_fully_connected_layer.cpp) store
weights [K,N] row-major and call dot(..., trans=false, trans_in=false),
but hexkl_mm.cpp's NPU shgemm path and layer_devel.h's offline WH-bake
both hardcoded the opposite [N,K] assumption without ever reading
TransB, silently mislabeling every FC/QKV/FFN weight's byte
before NPU tiling. CPU was always correct (it honors the layout via
ldb); NPU was always garbage; pre-baked and transient conve
produced identical garbage since both derived bytes from the same wrong
assumption. Existing unit tests never caught this because t
constructed weights the implementation's (wrong) way.

Fix: add a real transpose before WH-baking whenever TransB=
- hexkl_mm.cpp: new copyForWHBake() helper, wired into the transient and
  pin-cache RM->WH conversion paths.
- layer_devel.h: Tensor::transpose("0:2:1") before the offl
  reusing the pattern already used there for Q4_0/QS4CX.
- unittest_nntrainer_htp_kernels.cpp: add
  Accuracy_f32f16_f32_Prefill_TransBFalse_KNLayout, which r
  real production calling convention; add a prefillWHCacheClear() call
  there to avoid a heap-address-reuse false positive against the pin
  cache; add a permanent NNTR_HTP_VERIFY_PREFILL_MM diagnos
  diffs live NPU output against a CPU reference on real inputs.

Verified on-device (S25 Ultra): re-baked qwen3-0.6b-hmx-htp now
generates coherent output at full NPU prefill speed (323 TP
392 real prefill matmuls matching the CPU reference within FP16 noise
and zero mismatches.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[docs] Update HTP backend guide: build, l counts, devlog</summary><br />

- Record ENABLE_FP16=0/ENABLE_HEXKL=1 build support and wh_trailer.h
  header install (88779af0).
- Document npuAlive teardown guard explaining the exit-time Err=1 x3
  fix (62a11ef3).
- Fix stale hexkl_mm.cpp line reference for PREFILL_WH_PIN_
- Document TransB [K,N] vs [N,K] weight-layout fix and the three
  NNTR_HTP_* diagnostic env toggles.
- Add diagnostic-toggle run examples; update suite counts t
  kernels=22, backend=13, drop nonexistent HtpFallbackTest.
- Fix out-of-order section numbering in the unittest guide.
- Reconcile M/N/K alignment wording with hexkl_mm.cpp.
- Add 09_development_notes.md: per-component status dashboard and a
  commit-seeded devlog.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped
 Rewrite build guide to armv8 golden path; fix prompt truncation and source-dtype validation</summary><br />

Rewrite 02_build_and_env.md as a linear end-to-end pipeline (env ->
HTP lib build -> CausalLM app build -> device deploy -> on-device
FP16_WH bake -> run qwen3-0.6b -> verify), and switch its golden path
from armv9.2-a/armv9_android26 to armv8.2-a/armv8_android26 since the
two compile to functionally equivalent code today but armv9 causes a
SIGILL in libsdkl.so's init path on S25 Ultra (SM8750). Sync CLAUDE.md
and 09_development_notes.md to match.

While validating the rewritten guide end-to-end, fix two real bugs
found along the way:
- install_android.sh generated on-device wrapper scripts that passed
  "$@" unquoted, so a quoted multi-word prompt got re-split on
  whitespace and only the first word reached the model. Quote as "$@"
  in all three generated scripts.
- quantize.cpp only warned (not errored) when the source
  nntr_config.json declared a non-FP32 model_tensor_type, which could
  silently produce a structurally valid but numerically cor
  Make this a hard error; also force compute_engine=htp in the output
  config for FP16_WH bakes, and have main.cpp auto-prefer
  nntr_config_quantized.json when both exist.

Document both failure modes in a new Troubleshooting section, and add
an FP32-source precondition note to the bake pipeline descr

</details>

<details><summary>[docs] Translate and consolidate HTP backend guide to English</summary><br />

Restructure docs/backend_guide/htp_backend/ from 9 Korean docs down to
5 English ones, merging related content and replacing stale
measured-result dumps with fillable recording templates (introduction,
build/run pipeline, backend internals, unittest guide, e2e performance
results). Updated all cross-doc references (§ numbers, filenames, doc
guide links) to match the new structure.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Update backend docs and unittest suite for scratch-buffer reuse, ndk-build flow, e2e prompts</summary><br />

Docs (03/04/05) now describe the reusable NpuScratch staging buffers
(vs. per-call alloc/free), the ndk-build-based build/deploy flow for
HTP unittests, and record the latest on-device unittest and e2e run
results plus example prompt files for prefill-length sweeps.

Unittest changes drop kernel/dispatch tests for paths the backend no
longer exercises and add coverage for the new scratch-reuse and
offline-WH-bake paths. MNotAlignedThrows is renamed to
NNotAlignedThrows since M-misalignment is a supported padded path, not
a throw condition. Android.mk links unittest_nntrainer_htp_backend
against libccapi-nntrainer.so and drops the removed
sdkl_rm_to_wh_i8_probe standalone diagnostic target.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[HTP] Fix Android CI failures: doxygen/clang-format in unittests, sdkl gating in CausalLM Android.mk</summary><br />

Two unrelated CI failures fixed together:
1. doxygen-tag/cpp-linter: unittest_nntrainer_htp_backend.cpp and
   unittest_nntrainer_htp_kernels.cpp were missing required
   @author/@bug header tags and had inline comments that doxygen's
   advanced check misread as unterminated multi-line commen
   and reformatted with clang-format-14.
2. "Build CausalLM for Android" CI job: jni/Android.mk unconditionally
   declared a prebuilt sdkl module pointing at a hardcoded
   path, breaking the CPU-only CI target with no Hexagon SDK available.
   Add an ENABLE_HEXKL toggle (default 0) gating the sdkl prebuilt
   module/cflags/link, with HEXKL_ADDON_ROOT read from the
   instead of hardcoded. Verified with `ndk-build -n` that the default
   path emits no sdkl/ENABLE_HEXKL flags and ENABLE_HEXKL=1 threads the
   expected flags through.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

tale test expectations:
- CharTensor::getMemoryBytes()/read layout gained an int32_t
  zero-point-correction array alongside the float scale array.
- CharTensor::scale_size() for PER_CHANNEL_AFFINE returns height(), not
  width().

NeuralNetwork::load() pre-computes each weight's file offset before the
file is read, using scale_size() to size the QINT8 qparam region; at
that point the tensor's qscheme is still the constructor default
PER_TENSOR_AFFINE, so scale_size() returns 1 and the qparam bytes are
badly under-counted, drifting every subsequent weight's offset. Since
the QINT8 save path always transposes the weight to [N,K], the on-disk
per-channel scale/zp count equals the runtime tensor's width(); use
width() for the pre-calc size instead. Fixes
Qwen3Qint8WeightRoundTripTest.SaveLoadRestoresQuantMetadata.

Test alignments (no production behavior change): TensorPool
minMemoryRequirement accounts for the zp_corr array; Tensor_04_p
supplies height()-many scales; Qwen3 inspectWeight() sizes from
scale_size(); save_bin_with_fp16_dtype_throws_n gated behin
#ifdef ENABLE_FP16.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[docs] Untrack CLAUDE.md so it stays local, not merged into main</summary><br />

CLAUDE.md carries HTP-branch-specific Claude Code guidance and should
not leak into main via a future merge. Untrack it and ignore it going
forward; the file itself remains on disk for local use.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[CI] Fix doxygen tag check failures</summary><br />

Doxygen check flagged a false-positive multi-line-comment opener from an
embedded /* */ inside a // comment in neuralnet.cpp, plus several missing
@brief tags on structs/classes/enums in float_tensor.cpp an

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>

<details><summary>[nntrainer] Reuse pre-baked WH registry on HTP decode path</summary><br />

hexkl_mm's shgemm compute probed straight into g_wh_cache (
rm_to_wh per miss) for any M==1 call, so a decode path that later routes
through NPU would thrash on models larger than the cache instead of
reusing the WH registry already baked during prefill. Probe
lookupPrefillWH first and, on a hit, stage the pre-baked bytes into
scratch with a plain memcpy (no rm_to_wh); fall back to g_wh_cache on a
miss. No new kernel; per-token staging and Mp=32 padding re
dedicated GEMV kernel would remove those, out of scope).

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>
outing and CPU fallback on x86
via mocked ComputeOps.

Supporting changes to make the decision usable and verifiable:
- build_android.sh now auto-detects enable-htp from the meson builddir
  and forwards ENABLE_HEXKL/HEXKL_ADDON_ROOT to nntr_quantize, so
  `--fc_dtype FP16_WH` models are always built with WH-bake support
  instead of silently emitting a trailerless FP16 model.
- causal_lm.cpp dumps GENERATED_TEXT alongside TOKEN_IDS under
  HTP_E2E_DUMP_IDS for human-readable CPU-vs-HTP comparison.
- htp_e2e_compare.py compares CPU golden vs HTP generated t
  side and reads DEVICE/MODEL_DIR from env vars instead of hardcoded
  constants; benchmark_android.py gained a matching quantized-config
  probe.
- Docs (02/03/05) updated to describe the unconditional NPU routing and
  the WH-bake build/verification steps, without baking in
  date/commit-ID specifics.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [ ]Skipped

</details>
                                                                                                                                                    ### Summary

- Add a new HTP (Qualcomm Hexagon Tensor Processor) NPU backend via the HexKL SDK, registered as a new `Context`/`ComputeOps` vendor (`"htp"`)      alongside CPU/GPU.
- Implement FP16 (`shgemm_f32f16_f32`) and QINT8 (`shgemm_u8i8_i32`) NPU matmul kernels with the required M/N/K=32/64 alignment constraints and CPU fallback.
- Introduce offline WH (weight-layout) baking: `nntr_quantize --fc_dtype FP16_WH` writes a self-describing trailer of pre-converted WH-layout
weights, loaded and registered at model-load time so prefill RM->WH conversion (measured 33.1x prefill speedupon-device).
- Fix a real weight-layout bug (TransB not honored before Wced garbage NPU output for every FC/QKV/FFN weight; verified correct on-device with zero mismatches against a CPU reference.
- After fixing the WH-registry reuse gap on the decode path, route decode (M==1) to NPU unconditionally by default (same as prefill), since on-device A/B benchmarking showed parity-or-better throughp.
- Along the way: fix an unrelated QINT8 load-offset drift bug, multiple CI/build breakages (doxygen, Android NDK r26d, consumer header install, CI's CPU-only Android build), and rewrite/translate/consolidate the HTP backend documentation (Korean → English, 9 docs → 5).


Applications/CausalLM/benchmarks/README.md
```diff
@@ -38,6 +38,7 @@ python3 benchmark_android.py \
 - `-n, --n-gen`: Number of generation tokens (default: 0)
 - `-t, --n-threads`: Number of OMP threads (default: 4)
 - `-b, --batch-size`: Batch size (default: 1)
+- `--compute-engine`: compute_engine 값, 콤마 구분 (e.g., `htp,cpu`). 미지정 시 `nntr_config.json` 기존 값 사용
```
baek2sm, line 41: 오호. 한글?

Applications/CausalLM/models/causal_lm.cpp
```diff
@@ -233,6 +237,7 @@ std::pair<Tensor, Tensor> CausalLM::constructModel() {
     withKey("unit", NUM_VOCAB),
     withKey("disable_bias", "true"),
     withKey("weight_dtype", LMHEAD_DTYPE),
+    withKey("engine", "cpu"),
```
baek2sm, line 240: only cpu?

PR #4084, opened 2026-07-15
[tensor] Implement Q8_0 block quantization format and optimized GEMM kernels

baek2sm:
- **Q8_0 Weights Support**: Introduced core `Q8_0` block quantized datatype and representation (`Q8_0_Tensor`) with
safetensors de-serialization support.
- **Q8_0-based Conv Optimizations**: Implemented custom math kernels for optimized `indirect conv` and `depthwise
conv` utilizing vectorized hardware acceleration stubs.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: SeungBaek Hong <sb92.hong@samsung.com>

PR #4085, opened 2026-07-15
[Wait for #4084][nhwc] Support and optimize NHWC layout for Conv2D and related layers

baek2sm:
It contains 5 commits and wait for #4084 

- **NHWC Helper Layer Optimizations**: Added parallelized, vector-aligned helper operations for Pooling2D, Concat,
Slice, and Upsample2D under the NHWC layout.
- **Optimized Conv2D Math**: Implemented direct convolution, depthwise convolution, exact SiLU LUT (Lookup Table)
optimization, and vectorized bias addition under NHWC.

### Self evaluation:

Signed-off-by: SeungBaek Hong sb92.hong@samsung.com


nntrainer/layers/conv2d_layer.cpp
```diff
@@ -25,27 +32,461 @@
...
+        lut[i] = static_cast<_FP16>(silu);
+      }
+    }
+  });
+  return lut.data();
+}
+#endif
+
+/**
+ * @brief In-place SiLU / swish (x * sigmoid(x)) over a contiguous buffer.
```
jijoongmoon: Seems like implement this actvation again.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -25,27 +32,461 @@
...
+                                           int in_ch,
+                                           ::nntrainer::block_q8_0 *dst) {
+  const int nb = in_ch / 32;
+  auto &tm = ThreadManager::Global();
+  const unsigned int chunk = 512;
+  const size_t loops = ((size_t)n_spatial + chunk - 1) / chunk;
+  tm.parallel_for(0, loops, [=](size_t idx) {
+    unsigned r0 = (unsigned)idx * chunk;
+    unsigned r1 = std::min(r0 + chunk, (unsigned)n_spatial);
+    for (unsigned r = r0; r < r1; ++r) {
```
jijoongmoon: Seems like it only parallelized with cores. How about use intrinsics?

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -25,27 +32,461 @@
...
+ * plain block_q8_0, then Q8_0_Tensor::dot repacks to x4) into one, and lets the
+ * caller invoke Q8_0_Tensor::dot_prepacked_x4 (no per-conv repack, no per-conv
+ * QA malloc). Output bytes are identical to what the two-pass path produced.
+ * dst layout: M4=owoh/4 groups of block_q8_0x4 (136 B/blk) followed by (owoh %
+ * 4) remainder rows as plain block_q8_0 (34 B/blk) — exactly what
+ * dot_prepacked_x4 expects. dst must hold the same total as the block_q8_0
+ * buffer (136 B per 4 rows == 4 * 34 B). Q8_0 requires in_ch % 32 == 0.
+ */
+static inline void quantize_nhwc_q8_0x4_rows(const _FP16 *src, int in_ch,
+                                             int owoh, void *dst) {
```
jijoongmoon: I'm not sure this quantization locate here. How about move this codes to tensor implementation? conv2d_layer is more general.

jijoongmoon:
I'm not sure but it does not use conv op implementation in Q8_0 tensor. It uses direct implementation in conv2d_layer. Could you check it? Cause I think conv2d layer should be more general for other tensors as well.

baek2sm:
> I'm not sure but it does not use conv op implementation in Q8_0 tensor. It uses direct implementation in conv2d_layer. Could you check it? Cause I think conv2d layer should be more general for other tensors as well.

you are right. It still needs more refactoring for generalization. thank you

baek2sm:
Thanks for the review @jijoongmoon and Codex — all items addressed in 7b4d71ea (patched atop the existing commits; the Q8_0 dependency commits are left untouched):

**Codex P1 – Q8_0 save bias skip (layer_devel.h)**
The Q8_0 save path now skips bias-like tensors (K == 1) before the 32-divisibility check, mirroring the Q4_0 path. A biased FC layer no longer throws on K % 32 != 0.

**Codex P1 – conv0 fast path NCHW corruption (conv2d_layer.cpp)**
The conv0 FP16 fast path now requires both input and output to be NHWC, so the NHWC pointer-arithmetic kernel can no longer fire for an NCHW model of the same logical shape.

**Codex P1 – NHWC max-pool training indices (pooling2d_layer.cpp)**
The NHWC max forward now records the winning argmax index (NHWC linear offset) into pool_helper when training, so backward no longer consumes stale/uninitialized indices. NHWC backward itself is not yet implemented and now rejects NHWC explicitly in calcDerivative instead of silently taking the NCHW gradient path.

**Codex P2 – NHWC average pooling fall-through (pooling2d_layer.cpp)**
Added a dedicated NHWC average/global-average path so it no longer falls through to the legacy NCHW loop that mis-handles channel-last storage.

**@jijoongmoon – SiLU duplication (conv2d_layer.cpp)**
Removed the duplicated SiLU LUT epilogue in the conv0 path; both conv0 and the generic conv fused-activation epilogue now reuse the shared `convApplySwishInplace` helper.

**@jijoongmoon – quantize_nhwc_q8_0x4_rows location**
Already moved into Q8_0_Tensor in adef5e27; conv2d now calls `Q8_0_Tensor::quantize_nhwc_q8_0x4_rows`.

**@jijoongmoon – conv2d using Q8_0 conv op directly**
Partial: conv2d already dispatches to `Q8_0_Tensor::dot_prepacked_x4` / `convQ4_0Indirect`. The remaining in-layer Q8_0 scratch ownership and `NNTR_CONV_Q8ACT` dispatch are tagged `TODO(conv-generalization)` for a follow-up that moves the full NHWC Q8_0 conv into a Q8_0_Tensor conv op, as you noted further generalization is still needed.

On the parallel_for/intrinsics note: the NHWC pooling/conv paths use ParallelBatch for thread parallelism; SIMD intrinsics are used in the conv0 and Q8_0 GEMM kernels. Happy to vectorize the remaining scalar loops in a follow-up if that's the priority.

baek2sm:
Update on the generalization point @jijoongmoon raised ("conv2d layer should be more general for other tensors as well") — addressed without compromise in c3d04e81, on top of the earlier review-feedback patch (7b4d71ea). The Q8_0 dependency commits are untouched; both fixes are layered as patches atop them.

**SiLU/Swish moved out of conv2d_layer (acti_func.h)**
`get_silu_lut_fp16` + the in-place swish helper (FP16 64k LUT, NEON hard-swish approximation under NNTR_APPROX_SILU=1, scalar fallback) now live in `nntrainer::applySwishInplace` in acti_func.h. conv2d_layer.cpp no longer defines or duplicates any activation code — both the conv0 and the generic conv fused-activation epilogues call the shared helper. acti_func.h now includes arm_neon.h under `__ARM_NEON`.

**Q8_0 conv ops moved into Q8_0_Tensor**
Two high-level NHWC Q8_0-activation conv ops added to Q8_0_Tensor:
- `conv_nhwc_q8act_1x1` — fused `quantize_nhwc_q8_0x4_rows` + `dot_prepacked_x4`
- `conv_nhwc_q8act_indirect` — fused `quantize_nhwc_q8_0_rows` + `convQ4_0Indirect`

conv2d_layer's NHWC Q8_0-activation path now calls these instead of driving `quantize_nhwc_*` / `dot_prepacked_x4` / `convQ4_0Indirect` inline, so conv2d holds no `block_q8_0*` reinterpret and no activation-quantization logic — it only hands over the per-batch scratch slice and the weight/output tensors. The kernel call sequence is unchanged, so output is bit-identical; only the ownership/location moved. The earlier `TODO(conv-generalization)` marker is removed.

**Status of all review items**
- Codex P1 Q8_0 save bias skip — done (7b4d71ea)
- Codex P1 conv0 NHWC format guard — done (7b4d71ea)
- Codex P1 NHWC max-pool training indices — done (7b4d71ea); NHWC backward explicitly rejected until implemented
- Codex P2 NHWC average pooling path — done (7b4d71ea)
- jijoongmoon SiLU duplication — done (7b4d71ea dedup + c3d04e81 moved to acti_func.h)
- jijoongmoon quantize_nhwc location — already in Q8_0_Tensor (adef5e27)
- jijoongmoon conv2d using Q8_0 conv op / generalization — done (c3d04e81)

Build-verified locally with `meson -Denable-fp16=false` (634/634 targets linked). FP16/ARM NEON paths still need CI confirmation (local gcc 11.4 cannot build ENABLE_FP16).

PR #4086, opened 2026-07-19
[x86] refactor: Extract shared AVX2 helpers into avx2_internal.h

YongHyeon02:
## Dependency of the PR

- None (based on current `main`).
- First PR of the x86 AVX2/FP16 series re-split from #3834 / #3910 / #3967 / #3988 into small, independently reviewable units.

## Summary

Refactoring of `avx2_impl.cpp`: the file-local SIMD helper primitives (`exp256_ps`, `rcp_ps`, `hsum_avx`, `avx2_approx_swiglu(±alpha)`, `poly_gelu_tanh/erf_avx2`, `Exp2Table`, …) move into a new internal header `avx2_internal.h` under `namespace nntrainer::avx2::internal`. Missing doxygen briefs are added on the q4_0 shuffle helpers touched here.

This is the structure agreed with @jaemini-shin in the #3834 review thread: shared helpers live in one internal header (never in `avx2_impl.h`, never forward-declared per file), so the upcoming FP16 kernel PRs can reuse the same primitives without duplication.

- No behavior change; the public API header `avx2_impl.h` is untouched. Compiling `avx2_impl.cpp` on this commit and on `main` with identical flags yields the same symbol set and instruction stream (only assert() line-number/string operands shift).
- `avx2_internal.h` is internal and not installed — a note in the x86 `meson.build` records this.
- Unreferenced code is dropped rather than moved: the scalar `approx_exp_exp2_lookup` template, a commented-out low-precision `exp256_ps` variant, `compat_bit_cast`/`concept17_BinaryCastable` (left without users by that removal), and the unused `UNLIKELY`/`_nnt_ATTR_NOINLINE` macros.
- The header #undef's its compiler-shim macros at the end, so nothing leaks into including TUs.
- New swiglu / tanh_gelu_v2 / gelu_v2 accuracy tests against the scalar fallback references are added to `unittest_nntrainer_cpu_backend` (the moved primitives were previously exercised only by an Android-only test binary); tolerances are set from a measured dense-sweep of the approximation error, and the tests are gated to AVX2 builds since other architectures dispatch these entry points to different kernels.
- Every helper in `avx2_internal.h` carries function-level doxygen documentation, and the header suppresses `-Wattributes` around its attributed definitions itself so FP16 includers need no pragma of their own.

<details><summary>[cpu_backend] Extract shared AVX2 helpers into internal header (fccc2d4a)</summary><br />

Move the file-local AVX2 helper primitives of avx2_impl.cpp (exp256_ps,
rcp_ps, hsum_avx, swiglu/gelu polynomial approximations and the exp2
lookup table) into a new avx2_internal.h under namespace
nntrainer::avx2::internal, so that upcoming FP16 x86 kernels can reuse
the same primitives without forward declarations or code duplication.

The helpers are defined as plain inline functions with function-level
doxygen documentation, and the gelu/erf polynomial coefficients are
constexpr constants scoped to the namespace. The header suppresses
-Wattributes around its attributed definitions itself, so future FP16
includers do not need to carry the pragma.
The compiler shim macros (_nnt_CC_VECTORCALL, _nnt_ATTR_FLATTEN,
_nnt_ATTR_ALWAYS_INLINE and the __has_* fallbacks) are #undef'ed at the
end of the header, so it leaks no macros into including translation
units. avx2_impl.cpp reaches the primitives through explicit
using-declarations rather than a blanket using-directive, keeping its
dependencies visible.

Unreferenced code paths are dropped along the way: the scalar
approx_exp_exp2_lookup template, a commented-out low-precision variant
of exp256_ps, the compat_bit_cast/concept17_BinaryCastable helpers that
removal leaves without users, and the unused UNLIKELY and
_nnt_ATTR_NOINLINE macros.

The swiglu and gelu primitives were previously exercised only by an
Android-only test binary, so swiglu/tanh_gelu_v2/gelu_v2 accuracy tests
against the scalar fallback references are added to
unittest_nntrainer_cpu_backend to pin the moved code down on x86 CI. The
tests are gated to AVX2 builds because other architectures dispatch these
entry points to different kernels whose error bounds the x86-measured
tolerances do not describe, and the swiglu inputs are drawn from one RNG
stream so the fixed seed cannot correlate them. A
note in the x86 meson.build records that avx2_internal.h is intentionally
not installed with nntrainer_headers.

Also adds missing doxygen briefs on the q4_0 wrappers and rewrites a
plain multi-line comment in compute_fp16vcache_fp32_transposed to //
form, which the doxygen tag checker rejects.

Generated code is verified unchanged: compiling avx2_impl.cpp on this
commit and on main with identical flags yields the same symbol set and
instruction stream, differing only in the source line numbers and string
addresses that assert() embeds into its failure messages.

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -13,77 +13,23 @@
  */
 
 #include "avx2_impl.h"
+#include "avx2_internal.h"
 #include <array>
-#if __has_include(<bit>)
-#include <bit>
-#endif
 #include <cassert>
 #include <cmath>
 #include <cstdint>
 #include <cstring>
+#include <fallback_internal.h>
 #include <fp16.h>
 #include <immintrin.h>
 #include <limits>
-#if __has_include(<numbers>)
-#include <numbers>
-#endif
-#include <type_traits>
-#if __has_include(<version>)
-#include <version>
-#endif
-#include <fallback_internal.h>
 #include <nntrainer_error.h>
 #include <thread_manager.h>
+#include <type_traits>
 #include <util_func.h>
 #include <vector>
 
 #include "nntr_ggml_impl_common.h"
-
-#if !defined(__has_constexpr_builtin)
-#define __has_constexpr_builtin(x) (0)
-#endif
-
-#if !defined(__has_cpp_attribute)
-#define __has_cpp_attribute(x) (0)
-#endif
-
-// VECTORCALL calling-conv (default for x86_64-linux-gnu)
-#if _MSC_VER >= 1700
-#define _nnt_CC_VECTORCALL __vectorcall
-#else
-#define _nnt_CC_VECTORCALL
-#endif
-
-// Flatten attribute
-#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::flatten)
-#define _nnt_ATTR_FLATTEN [[msvc::flatten]]
-#elif __has_cpp_attribute(gnu::flatten)
-// clang, g++
-#define _nnt_ATTR_FLATTEN [[gnu::flatten]]
-#else
-#define _nnt_ATTR_FLATTEN
-#endif
-
-#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::noinline)
-#define _nnt_ATTR_NOINLINE [[msvc::noinline]]
-#elif __has_cpp_attribute(gnu::flatten)
-// clang, g++
-#define _nnt_ATTR_NOINLINE [[gnu::noinline]]
-#else
-#define _nnt_ATTR_NOINLINE
-#endif
-
-#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::forceinline)
-#define _nnt_ATTR_ALWAYS_INLINE [[msvc::forceinline]]
-#elif __has_cpp_attribute(gnu::always_inline)
-#define _nnt_ATTR_ALWAYS_INLINE [[gnu::always_inline]]
-#endif
-
-#if __has_cpp_attribute(unikely)
-#define UNLIKELY [[unlikely]]
-#else
-#define UNLIKELY
-#endif
 
 #if !defined(_MSC_VER) && !defined(__clang__)
 #pragma GCC diagnostic ignored "-Wattributes"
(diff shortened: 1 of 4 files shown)


nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -993,121 +730,14 @@ void swiglu(const unsigned int N, float *X, const float *Y, const float *Z,
...
-    _mm256_cmp_ps(x, _mm256_set1_ps(gelu_end_tanh), _CMP_GT_OQ);
-
-  y = _mm256_and_ps(y, gt_start);
-  y = _mm256_and_ps(y, le_end);
-  __m256 x_hi = _mm256_and_ps(x, gt_end);
-
-  return _mm256_add_ps(y, x_hi);
-}
+// poly_gelu_tanh_avx2 / poly_gelu_erf_avx2 are defined in avx2_internal.h
+using nntrainer::avx2::internal::poly_gelu_erf_avx2;
```
baek2sm: `using namespace nntrainer::avx2::internal;` is already declared at global scope (line 46), so I'm not sure why these individual using declarations are repeated here.

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
```diff
@@ -0,0 +1,455 @@
...
+#endif
+#endif
+
+#if !defined(_nnt_ATTR_ALWAYS_INLINE)
+#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::forceinline)
+#define _nnt_ATTR_ALWAYS_INLINE [[msvc::forceinline]]
+#elif __has_cpp_attribute(gnu::always_inline)
+#define _nnt_ATTR_ALWAYS_INLINE [[gnu::always_inline]]
+#endif
+#endif
```
Seunghui98: `_nnt_ATTR_ALWAYS_INLINE` has no `#else` fallback, so on a compiler that supports neither `msvc::forceinline` nor `gnu::always_inline` the macro stays undefined and every use site fails to compile. This was pre-existing in the .cpp, but now that it lives in a shared header it affects every future TU — please add an empty `#else` definition like the other attribute macros above.

---

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
```diff
@@ -0,0 +1,455 @@
...
+  __m128 shuf = _mm_movehdup_ps(vlow);
+  __m128 sums = _mm_add_ps(vlow, shuf);
+  shuf = _mm_movehl_ps(shuf, sums);
+  sums = _mm_add_ss(sums, shuf);
+  return _mm_cvtss_f32(sums);
+}
+
+// --- GELU polynomial approximation (tanh variant) ---
+
+#define gelu_start_tanh -4.38086284326899f
```
Seunghui98: These `#define`s (`gelu_start_tanh`, `tanh_c_gelu_p*`, and the `erf_*` block below) previously had file scope inside avx2_impl.cpp; in a header they now leak into every including TU and can silently collide with other code. Please either convert them to `constexpr float` constants inside `nntrainer::avx2::internal`, or `#undef` them at the end of the header.

---

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
```diff
@@ -0,0 +1,455 @@
...
+  auto neg_alpha_x = avx2_negate_ps(alpha_x);
+  auto inv_sigmoid = _mm256_add_ps(avx2_approx_exp_e2lookup<8>(neg_alpha_x),
+                                   _mm256_set1_ps(1.0f));
+  auto swiglu_nonscaled = _mm256_div_ps(x, inv_sigmoid);
+  return _mm256_mul_ps(swiglu_nonscaled, s);
+}
+
+// --- exp256_ps (used by softmax) ---
+
+static inline __m256 exp256_ps(__m256 x) {
```
Seunghui98: `exp256_ps`, `rcp_ps`, `hsum_avx`, and both `poly_gelu_*_avx2` are `static inline` in a header, which gives each TU its own internal-linkage copy and triggers `-Wunused-function` in TUs that include the header without using them (e.g. the upcoming FP16 TU). Plain `inline` is the conventional choice for header-defined functions and avoids both issues.

---

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -13,318 +13,37 @@
...
-avx2_approx_swiglu_alpha(__m256 x, __m256 s, __m256 alpha) noexcept -> __m256 {
-  auto alpha_x = _mm256_mul_ps(alpha, x);
-  auto neg_alpha_x = avx2_negate_ps(alpha_x);
-  auto inv_sigmoid = _mm256_add_ps(avx2_approx_exp_e2lookup<8>(neg_alpha_x),
-                                   _mm256_set1_ps(1.0f));
-  auto swiglu_nonscaled = _mm256_div_ps(x, inv_sigmoid);
-  return _mm256_mul_ps(swiglu_nonscaled, s);
-}
-} // namespace
+using namespace nntrainer::avx2::internal;
```
Seunghui98: This file-scope `using namespace nntrainer::avx2::internal;` already makes the per-symbol using-declarations at L734–740 redundant — please keep only one of the two. I'd prefer dropping this blanket directive and keeping the explicit using-declarations (or qualified calls), so it stays visible which internal helpers this TU actually depends on.

---

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
```diff
@@ -0,0 +1,455 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Sungsik Kong <ss.kong@samsung.com>
+ *
+ * @file   avx2_internal.h
+ * @date   10 Apr 2026
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author Sungsik Kong <ss.kong@samsung.com>
```
Seunghui98: Nit: the header credits Sungsik Kong (2024) with `@date 10 Apr 2026`, but this is code moved out of avx2_impl.cpp by this PR. Please align copyright/author/date with the actual authorship convention used in the repo (and fix the date — the PR was opened in July 2026).

---

Seunghui98 (commented):
## Description
Pure refactoring PR: file-local AVX2 helper primitives of `avx2_impl.cpp` (exp/swiglu/gelu approximations, `Exp2Table`, `compat_bit_cast`, `hsum_avx`, `rcp_ps`, attribute macros) move into a new internal header `avx2_internal.h` under `nntrainer::avx2::internal`, so the upcoming FP16 kernel PRs can reuse them without duplication. Public API (`avx2_impl.h`) is untouched; doxygen briefs are added on q4_0 wrappers and test helpers.

## Summary of review
The move is faithful — I diffed the relocated bodies against the originals and found no logic changes. Two silent improvements are included (typo fix `unikely`→`unlikely` for `UNLIKELY`, and `_nnt_ATTR_NOINLINE` now correctly probes `gnu::noinline` instead of `gnu::flatten`); both are correct. Tests only gain comments. Overall structure matches what was agreed in the #3834 thread.

## Points to double-check
- **Silently deleted code**: the scalar `approx_exp_exp2_lookup()` template was *removed*, not moved (it was unused dead code, so the build passes). Since the PR claims a pure move, this should either be mentioned in the description or the helper moved to the header if the FP16 PRs will need a scalar tail path.
- **Macro hygiene across future TUs**: `RESTRICT`, `UNLIKELY`, `_nnt_*` macros are now defined by a header shared with other TUs; the `#if !defined(...)` guards make include order matter if another file defines them differently.
- Behavior fixes noted above (`UNLIKELY`, `NOINLINE`) — confirm they are intended in a "no behavior change" PR (they are attribute-only, so codegen may change slightly but semantics don't).

## Requested changes (see inline comments)
1. Add an empty `#else` fallback for `_nnt_ATTR_ALWAYS_INLINE` (currently undefined on compilers lacking both attributes).
2. Don't leak the `gelu`/`erf` coefficient `#define`s from the header — use `constexpr` constants or `#undef` at the end.
3. Prefer `inline` over `static inline` for header-defined functions (avoids per-TU copies and `-Wunused-function`).
4. Remove the redundancy between file-scope `using namespace ...internal` and the per-symbol using-declarations in `avx2_impl.cpp`.
5. Nit: fix header copyright/author/date metadata.

None of these block the series — the refactor itself looks good.

---

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
jaemini-shin: I agree with this. It'd be better to remove this line.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -701,33 +420,51 @@ static inline void convert_q4_0x8_noshuffle(const void *src,
 
 // ================== wrappers for your K,N combinations ==================
 // K = 3072 (UNIT = 768)
+/**
```
jaemini-shin, line 428: Multi line comments for doxygen tag must be started with `/**`, please add a empty line.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -701,33 +420,51 @@ static inline void convert_q4_0x8_noshuffle(const void *src,
...
+ * @brief convert_q4_0x8_shuffle for K=3072 N=3072
+ */
 void convert_q4_0x8_shuffle_K3072_N3072(const void *src, uint16_t *d_out,
                                         uint8_t *qs_out) {
   // groups = 32
   convert_q4_0x8_noshuffle<768, 32>(src, d_out, qs_out);
 }
 
 // K = 8192 (UNIT = 2048)
+/**
```
jaemini-shin, line 456: Multi line comments for doxygen tag must be started with `/**`, please add a empty line.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
YongHyeon02: I've removed the blanket directive and consolidated the explicit `using` declarations at the top of `namespace nntrainer::avx2`.

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
YongHyeon02: I've converted all marcros to `constexpr inline float` inside `nntrainer::avx2::internal`.

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
YongHyeon02: I've added the empty `#else` fallback.

nntrainer/tensor/cpu_backend/x86/avx2_internal.h
YongHyeon02: I've changed all to plain `inline`.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
YongHyeon02: I've fixed it. The blanket directive dropped, explicit `using` declarations kept and moved around their first use.

YongHyeon02:
@jaemini-shin  Could you please run the CIs again? Thanks!

jaemini-shin (approved):
LGTM!

PR #4087, opened 2026-07-19
[Wait 4 #4086] [x86] perf: Remove allocations and TLS indirection from FP32 attention kernels

YongHyeon02:
## Dependency of the PR

- Depends on: #4086 (`split/x86-01-internal-header`)

## Summary

- `compute_kcaches` (FP32 in / FP16 kcache): converts the key row FP16→FP32
  inline in the dot loop via F16C instead of staging into a temporary buffer,
  and precomputes `1/sqrt(head_dim)` outside the loops.
- `compute_fp16vcache_fp32_transposed` and the `softmax_row(_with_sink)_inplace`
  helpers: thread-local scratch `.data()` pointers are resolved into locals once
  so the hot loops do plain pointer arithmetic (no repeated TLS/vector
  indirection), plus remainder/scale-path hardening.

**Why**: Same-session A/B benchmarking (i5-12400F) showed 1.4–2.0x on compute_kcaches
and 1.1–1.4x on the vcache path at gqa=1; zero per-call allocations.

**Notes for reviewers**: Covered by the existing `compute_kcaches`, `compute_fp16vcache_fp32_transposed`,
  and `softmax_row*` unit tests, plus new tests added here for the paths this PR introduces:
  `head_dim % 8 == 0` shapes, windowed/head-ranged invocations against scalar references, and
  same-thread thread_local scratch reuse (large shape then smaller shape). The reciprocal
  `1/sqrt(head_dim)` scale can differ from the former division in the last ULP; the scores feed
  a softmax, so the drift is not observable in practice.

<details><summary>[cpu_backend] Optimize x86 FP32 attention kernels without allocations (c448e1b3)</summary><br />

Rework compute_kcaches to convert the FP16 key cache to FP32 inline inside
the dot-product loop instead of staging rows in a temporary buffer, and
pre-compute the 1/sqrt(head_dim) scale outside the loops. Multiplying by
the reciprocal can differ from the former division in the last ULP; the
scores feed a softmax, so the drift is not observable in practice. Cache
the thread-local scratch data() pointers of the vcache and softmax_row
helpers in locals so the hot loops avoid repeated TLS and vector
indirections.

Add unit tests for the paths this change introduces: head_dim % 8 == 0
shapes (the new remainder-guard branches), windowed and head-ranged
invocations checked against scalar references, and same-thread scratch
reuse where each kernel runs a large shape and then a smaller one so the
grown thread_local buffers must still produce correct results.

**Self evaluation:**

</details>

### DIFF
--- a/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
+++ b/nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
@@ -13,77 +13,23 @@
  */
 
 #include "avx2_impl.h"
+#include "avx2_internal.h"
 #include <array>
-#if __has_include(<bit>)
-#include <bit>
-#endif
 #include <cassert>
 #include <cmath>
 #include <cstdint>
 #include <cstring>
+#include <fallback_internal.h>
 #include <fp16.h>
 #include <immintrin.h>
 #include <limits>
-#if __has_include(<numbers>)
-#include <numbers>
-#endif
-#include <type_traits>
-#if __has_include(<version>)
-#include <version>
-#endif
-#include <fallback_internal.h>
 #include <nntrainer_error.h>
 #include <thread_manager.h>
+#include <type_traits>
 #include <util_func.h>
 #include <vector>
 
 #include "nntr_ggml_impl_common.h"
-
-#if !defined(__has_constexpr_builtin)
-#define __has_constexpr_builtin(x) (0)
-#endif
-
-#if !defined(__has_cpp_attribute)
-#define __has_cpp_attribute(x) (0)
-#endif
-
-// VECTORCALL calling-conv (default for x86_64-linux-gnu)
-#if _MSC_VER >= 1700
-#define _nnt_CC_VECTORCALL __vectorcall
-#else
-#define _nnt_CC_VECTORCALL
-#endif
-
-// Flatten attribute
-#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::flatten)
-#define _nnt_ATTR_FLATTEN [[msvc::flatten]]
-#elif __has_cpp_attribute(gnu::flatten)
-// clang, g++
-#define _nnt_ATTR_FLATTEN [[gnu::flatten]]
-#else
-#define _nnt_ATTR_FLATTEN
-#endif
-
-#if _MSC_VER >= 1700 || __has_cpp_attribute(msvc::noinline)
-#define _nnt_ATTR_NOINLINE [[msvc::noinline]]
-#elif __has_cpp_attribute(gnu::flatten)
-// clang, g++
-#define _nnt_ATTR_NOINLINE [[gnu::noinline]]
-#else
-#define _nnt_ATTR_NOINLINE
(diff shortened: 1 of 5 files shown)


baek2sm:
LGTM for the standalone commit b1274d4c of this PR.
However, please check my comment on commit 1234815 from the dependency PR #4086.

nntrainer/tensor/cpu_backend/x86/avx2_impl.cpp
```diff
@@ -864,8 +864,12 @@ static void softmax_row_inplace(float *qk_out, size_t start_row, size_t end_row,
...
   const size_t vec_end = num_heads & ~((size_t)7); // floor(num_heads / 8) * 8
 
-  // 1. find max for each head
-  float *max_vals = new float[num_heads];
+  // 1. find max for each head (reusable thread-local scratch: 0 per-call alloc)
+  static thread_local std::vector<float> max_vals_buf;
+  static thread_local std::vector<float> sum_vals_buf;
+  max_vals_buf.resize(num_heads);
+  sum_vals_buf.resize(num_heads);
+  float *max_vals = max_vals_buf.data();
```
jaemini-shin, line 877: Could you measure the latency with or without this thread_local variable? There are some problems with this approach.
- We are using thread pool, i.e. these thread_local variables do not deleted after this function ends.
- This function is in dynamic library. In this case, accessing thread_local variable could be very slow.

YongHyeon02: | rows          | new[]   | thread_local |   delta |
|---------------|---------|--------------|--------|
| 1             | 66.7 ns |      49.6 ns | −25.7% |
| 4             | 133.6   |      119.9   | −10.2% |
| 16            | 400.8   |      401.4   |  +0.1% |
| 256           | 5597.4  |      5755.7  |  +2.8% |
| 1 (6 threads) | 70.0    |      51.0    | −27.2% |

@jaemini-shin  Yes, I also think it's the slow global-dynamic model, and the disassembly shows 6 `__tls_get_addr` calls per invocation. That said, all 6 are in the prologue and none are in the hot loops, so it's O(1) per call. The net trade is 2 malloc/free pairs → 6 TLS lookups, and the lookups come out cheaper. This weighs most in prefill, where each query row is split into a short call across the worker threads.

The +2.8% at large rows doesn't look like TLS. To check, I built a stack-buffer variant with zero TLS and zero alloc, and it shows the same +3.6% there — looks like code alignment from the function growing and relocating. SIMD instruction counts are identical across the variants.

jaemini-shin: Let's keep these thread_local variables. But could you please leave a todo comment about memory leakage? We need information in code when we need to reduce memory usage.

YongHyeon02: Okay, I left the comment about it! Thanks.

YongHyeon02:
@jaemini-shin Could you run the CIs?

PR #4089, opened 2026-07-20
[CausalLM] Let KVCacheManager own the KV cache write position

Jungwon-Lee:
## Summary

Make the host-owned `KVCacheManager` the single source of truth for the KV cache write position, instead of each `mha_core` layer tracking and self-advancing its own `cache_index`.

Previously, on every step the host pushed the position into every layer via a `cache_index=` property, and each layer independently advanced its own index inside `forwarding()`. This duplicated position state across all attention layers and the host-side `KVCacheManager`.

## Changes

- **Bind once**: `CausalLM::allocateAndBindKVCache()` binds every `mha_core` layer to the shared `KVCacheManager` via `MHACoreLayer::setKVCacheManager()` (pointer not owned).
- **Delegate**: When a manager is bound, `getCacheIndex()` / `forwarding()` read the write position from the manager and no longer self-advance. `cache_index` remains only as a fallback for the unbound / back-compat path.
- **Host advances once per step**: `advanceKVCachePosition()` is called after each `incremental_inference`; `setKVCachePosition()` no longer fans the index out to every layer.
- **`LayerNode::getLayer()`** moved to public so the host can downcast a layer node to `MHACoreLayer` for binding.
- `meson.build`: expose the parent include dir so `mha_core` can include `kv_cache_manager.h`.

## Notes

- Behavior is unchanged when no `KVCacheManager` is bound (fallback path preserved).


baek2sm (approved):
thanks, LGTM

dlwlzzero (approved):
Nice Work 👍

PR #4092, opened 2026-07-20
Move CausalLM app, C API, xgrammar, and QNN app infra into nntrainer (public, no private-model references)

jayden0701:
## Summary

Moves the on-device app layer that used to live only in the private QuickAI
project — the C API (`quick_dot_ai_api`), the Android app/AAR, xgrammar, and
the QNN model-runtime infra for public QNN models — into nntrainer's own
`Applications/CausalLM/`. After this change, **nntrainer alone builds and
runs the full public CausalLM app** (CPU models always, QNN models opt-in),
with no dependency on QuickAI's build system. QuickAI is reduced to an
optional proprietary model plugin that overlays on top of this public build;
this tree stays free of any private model references throughout.

139 files changed, +23629/-3189.

## Motivation

Pre-factoring for a 3-repo split that's coming next:
`QuickAI-internal (proprietary models)` → `QuickAI-external (app + API,
public)` → `nntrainer (engine, public)`. Staging the app/API inside
nntrainer now, instead of jumping straight to a new repo, let each step be
verified independently (x86 build, Android build, on-device run) before the
final split.

## What changed

**API / app / AAR relocation**
- `api/` (the `quick_dot_ai_api` C ABI, ~12 files) and `api-app/test_api.cpp`
  moved into `Applications/CausalLM/`, replacing the old `causal_lm_api`.
- `Android/` (the `QuickDotAI` AAR + `SampleTestAPP` demo) moved into
  `Applications/CausalLM/Android/`; AAR now references the canonical header
  instead of a bundled copy (removes header drift).
- The app optionally loads an external model-extension plugin at runtime
  (`System.loadLibrary`, best-effort, try/catch) — lets a private overlay
  add extra models to the catalog without the public app knowing anything
  about it.

**xgrammar + QNN folded into `libcausallm.so`**
- Added `mlc-ai/xgrammar` as a submodule under `Applications/CausalLM/xgrammar`
  (pinned). The wrapper + vendored library now compile into `libcausallm.so`
  exactly once (a `static_library` `link_whole`'d in), so the
  `XGrammarManager` Meyers singleton has a single definition that every
  consumer (`libquick_dot_ai_api.so`, an optional model plugin) borrows at
  runtime — CPU-only builds get it too, it isn't QNN-specific.
- The public QNN model-runtime infra (`qnn/*.cpp`) and the public
  `gemma4-e2b-qnn` model are now folded into `libcausallm.so` as well, gated
  by a single new `enable-qnn` option. One flag turns on the QNN infra + QNN
  models + the multimodal run path together — there's no separate
  `enable-qnn-models` or `enable-experimental-multimodal` gate anymore.

**New self-contained app build**
- New meson project at `Applications/CausalLM/app_build/`: links a
  **prebuilt** nntrainer engine (`find_library`) instead of requiring a
  from-source engine build in the same project, so this app build has zero
  dependency on QuickAI's meson. It builds `libcausallm.so`,
  `libquick_dot_ai_api.so`, and the CLIs below.
- Consolidated orchestration into a single script,
  `Applications/CausalLM/build_android.sh`: submodule/tokenizer checks →
  engine build (`tools/package_android.sh`, skippable) → app meson build →
  stage `prebuilt_libs/` → gradle install → push a working CLI to
  `/data/local/tmp/Quick.AI/`. Dropped 8 scripts that this replaces:
  `build.sh`, `apk-build-install.sh`, `apk_install_android.sh`,
  `install_android.sh`, `install_android_standalone.sh`, `build_api_lib.sh`,
  `build_test_app.sh`, `run_test.sh`.
- nntrainer's own desktop meson build (`meson setup … && ninja -C …`,
  building the engine + the `Applications/CausalLM` subdir target) is
  **unaffected** by any of this — it's a separate meson definition that
  still uses in-tree dependencies.

**Shared ABI configuration between legacy and standalone Android consumers**
- Both Android consumers — the legacy in-tree `jni/Android.mk` path and the
  new standalone `app_build/` meson project — now build against one shared
  ABI configuration emitted by the engine build, so a prebuilt
  `libnntrainer.so` and the app build can't drift on FP16 flags or `-march`.
  The engine build writes a generated `nntrainer-abi.ini`
  (`jni/nntrainer-abi.ini.template`) recording `nntrainer_abi_schema`,
  `nntrainer_abi_cflags` (incl. `-DENABLE_FP16=1`/`-DUSE__FP16=1`),
  `nntrainer_enable_fp16`, and `nntrainer_arm_march`
  (`-march=armv8.2-a+fp16+dotprod+i8mm`); the standalone meson project
  reads it back via `find_library` metadata so the app compiles with the
  same flags the engine was built with.
- `build_android.sh` now supports a CPU-only AAR packaging mode
  (`--skip-qnn --assemble-aar`) that excludes every QNN `.so`
  (`libqnn_context.so`, `libQnn*.so`) from the AAR's `jni/arm64-v8a/` while
  still shipping the CPU runtime libs. The legacy `jni/Android.mk` compat
  modules are installed alongside the prebuilt libs so old consumers keep
  linking.
- Generated prebuilt paths are now relocatable (no absolute build-tree paths
  baked in), and the selected Android NDK path is the single source of truth
  across the native meson cross-compile and the Gradle `externalNativeBuild`.
- Supported build modes are documented in
  `Applications/CausalLM/Android/Architecture.md` and the
  `Applications/CausalLM/README.md` / `api/README.md` rewrites.

**New path-based CLI**
- Added `nntr_causallm` (`Applications/CausalLM/main.cpp`, built by the new
  app project): takes a filesystem `<model_dir>` directly — reads
  `config.json`/`nntr_config.json`, resolves the architecture, loads and
  runs — no catalog/model-name lookup needed, unlike the C-API-driven
  `quick_dot_ai_test`. Pushed to device with a `run_causallm.sh` wrapper.

**Singleton identity anchored to the owner library (generalized)**
- The earlier ad-hoc out-of-line definitions for `Engine::Global()`,
  `causallm::Factory::Instance()`, and `ThreadManager` (each added
  one-off during the migration to stop per-`.so` duplication) are replaced
  by a single general mechanism: a new `NNTRAINER_SINGLETON_API` visibility
  macro in `utils/singleton.h` plus an explicit
  `template<> Singleton<T>::Global()` specialization declared in the header
  and defined in the owning library's single translation unit. Every
  built-in singleton — `Engine`, `AppContext`, `ClContext`, the three
  OpenCL managers, `Profiler`, `ThreadManager` — is now anchored this way, so
  any consumer `.so` (the app, a model plugin) that includes the header
  calls into the one definition exported from `libnntrainer.so` instead of
  instantiating its own hidden copy.
- Added a cross-DSO regression test (`test/unittest/singleton/`): two probe
  shared libraries (`singleton_probe_a`, `singleton_probe_b`) each return
  their `Profiler` / `ThreadManager` singleton address, and
  `unittest_singleton_dso.cpp` asserts both probes see the same instance —
  this fails the build if the visibility/specialization is ever regressed.
- The plugin-owned `QnnContext` lifecycle is kept explicit (a model plugin
  still owns its own QNN context; it is *not* folded into the engine
  singleton), so the public engine has no QNN link-time dependency.

**Bugs found and fixed during the migration**
- `Engine::Global()` and `causallm::Factory::Instance()` must each have
  exactly one definition shared by every consumer `.so`; a duplicated
  Factory (e.g. from a model plugin compiling its own copy) makes
  `Factory::create()` return `nullptr` even though the model is visible in
  the catalog. Both are now defined out-of-line in a single translation
  unit (via the `NNTRAINER_SINGLETON_API` mechanism above) and documented
  as a "rebuild every consumer lib" hazard.
- Shared `ThreadManager` across shared libraries (previously duplicated the
  same way; now covered by the same mechanism and the cross-DSO test).
- FP16 defines (`ENABLE_FP16`/`USE__FP16`) weren't propagated to the new
  app-build meson project, causing an FP32 fallback + a Tensor vtable ABI
  mismatch against the FP16-built engine (~30% slower, +12% memory on
  device) — now propagated via the shared `nntrainer-abi.ini` metadata so
  the app build always matches the engine build.
- QNN backend init on a device with `libqnn_context.so` present but no
  usable QNN backend fell through into an unguarded null function-pointer
  call (SIGSEGV, even on the CPU path) instead of returning an error; now
  guarded so CPU inference keeps working when QNN init fails.
- The app's speculative-decoding toggle used to hardcode a specific private
  model's catalog id; it's now driven by a `sd_variant_id` field + a
  capability bit in the model descriptor, so the public app makes no
  reference to any proprietary model name.

**Public tree scrub**
- Removed all proprietary model references from the public tree; a
  repo-wide scrub for the private model family's name now only turns up an
  unrelated `erf()` math comment in `tensor.h` (a false positive).

## Build entry points (after this PR)

`build_android.sh` is the single entry point for the Android native libraries,
the QuickDotAI AAR, and the sample app. With **no options** it builds the
CPU-only `jni/Android.mk` targets only — it does **not** invoke Gradle or
touch a device. App/AAR modes are opt-in via the flags below.

```bash
export ANDROID_NDK=/path/to/android-ndk
cd Applications/CausalLM

# --- Android (CPU by default; QNN opt-in) ---
./build_android.sh                              # CPU-only Android.mk targets;
                                                # artifacts in jni/libs/arm64-v8a/
./build_android.sh --assemble-aar               # + standalone app + QuickDotAI AAR/APK
                                                #   (no device install)
./build_android.sh --install                    # + install sample APK + push CLIs to
                                                #   /data/local/tmp/Quick.AI/
./build_android.sh --enable-qnn                 # QNN backend + QNN AAR libs
                                                #   (requires QNN_SDK_ROOT; implies --assemble-aar)
./build_android.sh --skip-qnn --assemble-aar    # explicit CPU-only AAR (no QNN libs)
./build_android.sh --native-only                # standalone app native libs, skip Gradle
./build_android.sh --cache                      # reuse a compatible engine build (build on miss)
./build_android.sh --skip-engine                # strictly reuse an existing engine build (fail if absent)
./build_android.sh --clean                      # recreate the standalone app build dir

# --- nntrainer desktop build (unchanged; builds engine + Applications/CausalLM) ---
meson build -Denable-transformer=true
ninja -C build
./build/Applications/CausalLM/nntr_causallm <model_dir>   # path-based CLI
```

The full option table is in `Applications/CausalLM/README.md` ("Android Build
& Test"); the supported build modes are documented in
`Applications/CausalLM/Android/Architecture.md`.

## Verification

- x86: `meson`/`ninja` build green (engine + `Applications/CausalLM` subdir +
  unit tests, including the new `unittest_singleton_dso` cross-DSO test).
- Android CI (`.github/workflows/causallm_android.yml`): `build_android.sh
  --skip-qnn --assemble-aar --legacy-ndk` runs clean; the workflow then
  verifies (a) the generated `nntrainer-abi.ini` carries
  `nntrainer_abi_schema = 1`, the `-march=armv8.2-a+fp16+dotprod+i8mm` flag,
  and `-DENABLE_FP16=1`/`-DUSE__FP16=1`; (b) the CPU-only AAR contains the
  CPU runtime libs (`libquickai_jni.so`, `libquick_dot_ai_api.so`,
  `libcausallm.so`, `libnntrainer.so`, `libccapi-nntrainer.so`,
  `libc++_shared.so`) and *no* QNN libs; (c) the standalone app
  `compile_commands.json` was compiled with the same `-march` + FP16 flags;
  (d) the public C header `quick_dot_ai_api.h` compiles cleanly as C under
  the NDK clang (`-fsyntax-only`); (e) `build_android.sh --cache
  --skip-qnn --native-only` reuses the engine cache.
- Android on-device: `build_android.sh` clean build → gradle install
  succeeds → app launches with no crash; model catalog shows public models
  only (no private ones); CPU model load/streaming/cancel confirmed
  on-device; the FP16-parity and QNN-init-guard fixes above were each
  isolated and confirmed on-device (non-QNN device with
  `libqnn_context.so` present no longer crashes; CPU throughput/memory back
  to baseline).
- QNN device: `gemma4-e2b-qnn` load/generate, and the capability-driven
  speculative-decoding toggle, confirmed working.
- `nntr_causallm` compiles and links cleanly as part of this app build
  (exercised indirectly via a downstream overlay build); a dedicated
  standalone on-device run of this specific binary is a good follow-up if
  reviewers want it called out separately.
- Guardrail checked throughout: no nntrainer submodule-pointer bumps (n/a
  here, this IS the submodule), and a repo-wide scrub for proprietary model
  references stays clean after every slice.

## Out of scope

- The proprietary models themselves stay out of this tree entirely — they
  live in the downstream QuickAI-internal project and are overlaid as an
  optional plugin `.so` at build time. Nothing in this PR references any of
  their names.


Applications/CausalLM/api/quick_dot_ai_api.cpp
```diff
@@ -0,0 +1,3294 @@
...
+  int kv_len = 0;
+};
+
+// Globals shared across all handles — options set via setOptions() apply
+// process-wide regardless of which handle is active.
+static std::mutex g_registry_mutex;
+static bool g_use_chat_template = true;
+static bool g_verbose = false;
+static std::string g_last_output = "";
+static std::optional<causallm::ChatTemplate> g_chat_template;
```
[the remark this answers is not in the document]
jayden0701: This is going to be handled in following PRs where I will fix api's

Applications/CausalLM/api/quick_dot_ai_api.cpp
```diff
@@ -0,0 +1,3294 @@
...
+    return CAUSAL_LM_ERROR_INVALID_PARAMETER;
+  }
+
+  causallm::XGrammar *grammar = nullptr;
+  // Step 1: Check if tool exists in XGrammarManager
+  if (causallm::XGrammarManager::Instance().hasTool(tool_name)) {
+    LOGD("[runModelWithToolHandle] Tool '%s' found in XGrammarManager, using "
+         "existing grammar",
+         tool_name);
+    grammar = causallm::XGrammarManager::Instance().getGrammar(tool_name);
```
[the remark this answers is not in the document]
jayden0701: This is going to be handled in following PRs where I will fix api's

Applications/CausalLM/api/quick_dot_ai_api.cpp
```diff
@@ -0,0 +1,3294 @@
...
+  }
+
+  // NOTE: We intentionally do NOT take the mutex here to avoid blocking
+  // when run() is holding the lock. The requestStop() method is thread-safe
+  // (uses atomic<bool>), and the models vector is not modified during run()
+  // (only during load/unload which do take the mutex). This allows immediate
+  // cancellation from any thread (e.g., UI cancel button handler).
+  LOGD("[DEBUG] cancelModelHandle: checking state without mutex, "
+       "initialized=%d, models.size=%zu",
+       handle->initialized, handle->models.size());
```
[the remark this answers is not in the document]
jayden0701: This is going to be handled in following PRs where I will fix api's

Applications/CausalLM/api/quick_dot_ai_api.cpp
```diff
@@ -0,0 +1,3294 @@
...
+  if (input_already_formatted) {
+    return input;
+  }
+
+  return apply_chat_template(architecture, input);
+}
+
+static std::string get_quantization_suffix(ModelQuantizationType type) {
+  return "";
+  switch (type) {
```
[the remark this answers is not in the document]
jayden0701: Quant Type will be disabled

Applications/CausalLM/api/callback_streamer.cpp
```diff
@@ -13,39 +15,44 @@
...
-  self->base.vtable = &callback_streamer_vtable;
-  self->callback = callback;
+  self->base.vtable = &kCallbackStreamerVTable;
+  self->callback = cb;
   self->user_data = user_data;
   self->cancelled = 0;
 }
 
-} // extern "C"
+} // extern "C"
```
jayden0701: I will add newline

Applications/CausalLM/api/callback_streamer.cpp
```diff
@@ -3,8 +3,10 @@
  * Copyright (C) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
  *
```
jayden0701, line 4: basically same file

Applications/CausalLM/api/callback_streamer.h
```diff
@@ -3,13 +3,21 @@
  * Copyright (C) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
```
jayden0701, line 3: just added comments

nntrainer/utils/profiler.cpp
```diff
@@ -24,6 +24,16 @@
 #include <profiler.h>
```
jayden0701, line 24: changed files from here are just Singleton changes

nntrainer/qnn/qnn_context.h
```diff
@@ -61,8 +61,6 @@
 #endif
 #endif
```
jayden0701, line 63: QNN does not inherit singleton anymore.
QNN is created with createfunc() and destoryed by destroyfunc(). Which makes it not actually a singleton pattern.

And QNNContext is defined in libqnn_context.so . And QNN is optional. Which makes it hard to compile and manage.

initializeOnce() is used to ensure only one init.

jayden0701:
static check for doxygen will pass when https://github.com/nntrainer/nntrainer/pull/4132 is merged 
(checked with local + online github CI)

dlwlzzero (approved):
LGTM 👍

dlwlzzero:
> static check for doxygen will pass when #4132 is merged (checked with local + online github CI)

I think #4132 didn't solve the CI failures, will you check?

jayden0701:
> > static check for doxygen will pass when #4132 is merged (checked with local + online github CI)
> 
> I think #4132 didn't solve the CI failures, will you check?

Running CI on this PR runs the old doxygen-tag.sh .   
The error message `.github/workflows/static.check.scripts/doxygen-tag.sh: line 76: local: can only be used in a function` proves this, since this error only appears on CI before  #4132

jaemini-shin:
Did you try rebasing it to main branch? It looks like CI uses the workflow file in this PR!

jayden0701:
> Did you try rebasing it to main branch? It looks like CI uses the workflow file in this PR!

Oh I didn't know rebase works that way! I will try :)

meson_options.txt
```diff
@@ -74,6 +74,21 @@ option('nnstreamer-subplugin-install-path', type: 'string', value: 'lib/nnstream
...
+# Compiles the public QNN model-runtime infra (Applications/CausalLM/qnn/*)
+# and the public gemma4-e2b-qnn model into libcausallm.so, and defines
+# ENABLE_QNN_MODELS for the api layer (the dynamic_cast<Quick_Dot_AI_QNN *>
+# / make_unique<Gemma4_E2B_QNN> paths in api/quick_dot_ai_api.cpp). Requires
+# enable-transformer + enable-npu: the QNN backend (libqnn_context.so) is
+# registered into nntrainer's Engine at runtime by qnn/quick_dot_ai_qnn.cpp
+# (Engine::Global().registerContext("libqnn_context.so", ...)), not linked
+# directly, so no QNN SDK headers are needed to compile this option's
+# sources. Off by default; proprietary QNN models are layered on top by a
+# separate downstream project, not in this public tree.
```
Seunghui98: I think Claude wrote it down verbosely.
Could you check this??

jayden0701: It looks too long...I will shorten it. Thank you!

Applications/CausalLM/build_android.sh
```diff
@@ -1,206 +1,644 @@
...
-        log_info "  Checked paths:"
-        log_info "    - $libs_path"
-        log_info "    - $obj_path"
-        return 1
+        echo "[3] Building the Android engine."
+    fi
+    (
+        cd "$NNTRAINER_ROOT"
+        ./tools/package_android.sh \
+            -Dmmap-read=false \
```
jaemini-shin:
```suggestion
```
This line should be removed.

Seunghui98 (commented):
## Review — #4092 (CausalLM app / C-API / xgrammar / QNN relocation into nntrainer)

### Description
This PR relocates the on-device app layer that previously lived only in the private QuickAI project into nntrainer's own `Applications/CausalLM/`: the `quick_dot_ai_api` C ABI, the `QuickDotAI` Android AAR + sample app, the `xgrammar` submodule + wrapper, and the public QNN model-runtime infra. After the change nntrainer alone builds and runs the full **public** CausalLM app (CPU always, QNN opt-in via a single flag), with QuickAI reduced to an optional proprietary model plugin overlaid at build time. It also generalizes the cross-`.so` singleton-identity mechanism (`NNTRAINER_SINGLETON_API` in `utils/singleton.h` + explicit `Singleton<T>::Global()` specializations anchored in the owner library) and adds a shared `nntrainer-abi.ini` so the standalone app build can't drift from the engine build on FP16 / `-march` flags.

### Summary
- **Scope:** 139 files, +23,644 / −3,195. The overwhelming majority is *moved* code (old `causal_lm_api` → new `quick_dot_ai_api`, Android tree, QNN tree) plus 8 build scripts consolidated into one `build_android.sh`. The genuinely *new logic* to scrutinize is small and concentrated: `utils/singleton.h`, the per-singleton `Global()` specializations, the QNN init null-guard, the ABI-ini propagation, and the new cross-DSO test.
- **Engine-side singleton change looks correct.** The primary `Singleton<T>::Global()` stays inline in the header, and each owner declares an explicit `template<> ... Global()` specialization (so consumers don't implicitly instantiate their own copy) with the definition in the owning library's single TU. `unittest_singleton_dso` is a good regression guard.
- **The QNN null-function-pointer guard is the right fix** — `init()` now returns `-1` when the backend/system function-pointer tables fail to load, so a device with `libqnn_context.so` present but no usable QNN backend keeps the CPU path alive instead of dereferencing a null `qnnInterface.*` pointer.

### Code quality / style
- Consistent with the codebase; doc comments and SPDX headers are in place on new files.
- `build_android.sh` doing submodule check → engine build → app meson → gradle → device push in one script is a big usability win over the 8 deleted scripts.
- `meson_options.txt` — the `enable-qnn-models` / `enable-npu` option descriptions are very verbose (already flagged by @Seunghui98 at line 90); worth trimming to 1–2 lines and moving the SDK-acquisition prose into `docs/`.
- Naming drift to double-check: the PR description talks about a single new `enable-qnn` flag, but the committed `meson_options.txt` still exposes `enable-api` + `enable-qnn-models` (no `enable-qnn`). Please confirm the description matches the actual option names so users aren't misled.

### Potential bugs
1. **`return false` from an `int`-returning function → error paths silently report success** (`nntrainer/qnn/qnn_context.cpp`, `int QNNContext::init()`). Several failure branches — `beforeBackendInitialize()` failure, the `m_backendConfig` `calloc` failure, and `afterBackendInitialize()` failure — do `return false;`. In an `int` function `false` is `0`, and the caller `QNNContext::initialize()` treats `init() != 0` as failure — so `0` means **success**. On any of those failures `init()` reports success and `initialize()` proceeds to `setMemAllocator()` / `registerFactory()` on a half-initialized backend — exactly the SIGSEGV class this PR set out to prevent with the new `return -1` guards a few lines above. These `return false` lines look pre-existing, but they sit squarely in the init path you're hardening; recommend converting them to `return -1;` while you're here. **(strongest finding)**
2. **Open P1 concurrency/lifetime issues in `api/quick_dot_ai_api.cpp` are real, not just style** — the bot-flagged threads (process-global `g_chat_template`, the shared mutable `XGrammar` matcher from `getGrammar()`, and the `cancelModelHandle()` vs `unloadModelHandle()` use-after-free) can corrupt output across independent handles or crash under concurrent/multi-model use. Deferring them to a follow-up API PR is reasonable, but please open a tracking issue so they aren't lost — the PR advertises "independent handle" behavior that the current code does not actually provide.
3. **Unreachable quant switch (P2)** — the unconditional early return in the quant-variant resolver makes W4A32/W16A16/W8A16/W32A32/UNKNOWN all map to the same directory. Author says quant type will be disabled; if so, remove/guard the dead switch and the public API/CLI options that expose those choices, otherwise callers silently load the wrong weights.
4. **Minor:** in `init()` the "logging initialized in the backend" branch logs `ml_logw("Logging not available in the backend.")` while the failure branch logs "Unable to initialize logging" — the success/failure log strings look swapped. Cosmetic only.

### What you should look at more closely
- **Test coverage vs. surface area.** +23k lines but the only *new* automated test is the cross-DSO singleton check, and this PR *deletes* `api/test_api.cpp` (−306) and `api/unittest_cancel_api.cpp` (−309). Please confirm the cancel/unload path is still covered somewhere — that's precisely the code with an open use-after-free P1.
- **Singleton coverage.** The probe test exercises only `Profiler` and `ThreadManager`. `Engine`, `AppContext`, `ClContext`, and the three OpenCL managers use the same mechanism but aren't in the probe — a regression on any of those wouldn't be caught. Also worth a one-line note in the header that consumers must include the *derived* header (not just `singleton.h`) so they see the explicit specialization and don't instantiate their own inline copy.
- **QNN `init()` error contract.** Given finding #1, audit every `return` in `init()` for the `int` (0 = success) convention, not just the two new guards.
- **Reviewability.** A 139-file / +23k PR mixing pure moves with real fixes and new build infra is very hard to review with confidence. The description does call out the fixes, but a short "pure-move vs. logic-change" file list (or splitting the singleton/ABI/QNN-guard fixes into their own small PRs) would materially raise reviewer confidence in the non-mechanical parts.

### Bottom line
The engine-side changes (singleton mechanism, QNN null-guard, ABI-ini propagation) are sound and well-motivated, and the migration itself appears faithful. The blocking concerns are the `return false`→success bug in `init()` (please fix in this PR) and making sure the deferred P1 data-race/use-after-free items in the API layer are tracked rather than dropped. Everything else is polish (option verbosity, doc/naming drift, test coverage for the moved API surface).

---

Seunghui98 (commented):
> ⚠️ This supersedes my earlier, longer review comment (kept for history; GitHub doesn't allow editing a submitted review). Please read this compact version instead.

## Review — #4092 (compact)

**What:** Relocates QuickAI's public app layer (C API, Android AAR, xgrammar, public QNN) into nntrainer `Applications/CausalLM/`. 139 files, +23.6k/−3.2k — mostly *code moves*; the real new logic is the cross-`.so` singleton mechanism, the QNN init null-guard, and ABI-ini propagation.

**Good:** Singleton specialization (`NNTRAINER_SINGLETON_API`) is correct; cross-DSO test is a nice guard; QNN null-function-pointer guard is the right fix.

**Bugs / must-fix:**
- 🔴 `qnn/qnn_context.cpp` — `int init()` has failure branches (`beforeBackendInitialize` / `calloc` / `afterBackendInitialize`) that `return false;`. In an `int` function that's `0` = **success**, and the caller treats `0` as success → proceeds on a half-initialized backend, re-introducing the exact crash this PR guards against. Change to `return -1;`.
- 🟠 API P1s (global `g_chat_template`, shared mutable `XGrammar` matcher, `cancel`/`unload` use-after-free) are real data-race/lifetime bugs. Deferring to a follow-up API PR is fine, but open a tracking issue.
- 🟡 Quant switch is unreachable (P2): every variant maps to the same dir. If quant is being disabled, remove the dead switch + the exposed options.

**Look closer:** this PR deletes ~615 lines of API tests but adds only the singleton test — confirm the `cancel` path stays covered. Also `enable-qnn` (desc) vs `enable-qnn-models` (code) naming, and the verbose `meson_options.txt` block already flagged by @Seunghui98.

**Bottom line:** engine-side changes are solid; blockers are the `init()` `return false` bug (fix here) and tracking the API P1s.

---

jayden0701:
> > ⚠️ This supersedes my earlier, longer review comment (kept for history; GitHub doesn't allow editing a submitted review). Please read this compact version instead.
> 
> ## Review — #4092 (compact)
> **What:** Relocates QuickAI's public app layer (C API, Android AAR, xgrammar, public QNN) into nntrainer `Applications/CausalLM/`. 139 files, +23.6k/−3.2k — mostly _code moves_; the real new logic is the cross-`.so` singleton mechanism, the QNN init null-guard, and ABI-ini propagation.
> 
> **Good:** Singleton specialization (`NNTRAINER_SINGLETON_API`) is correct; cross-DSO test is a nice guard; QNN null-function-pointer guard is the right fix.
> 
> **Bugs / must-fix:**
> 
> * 🔴 `qnn/qnn_context.cpp` — `int init()` has failure branches (`beforeBackendInitialize` / `calloc` / `afterBackendInitialize`) that `return false;`. In an `int` function that's `0` = **success**, and the caller treats `0` as success → proceeds on a half-initialized backend, re-introducing the exact crash this PR guards against. Change to `return -1;`.
> * 🟠 API P1s (global `g_chat_template`, shared mutable `XGrammar` matcher, `cancel`/`unload` use-after-free) are real data-race/lifetime bugs. Deferring to a follow-up API PR is fine, but open a tracking issue.
> * 🟡 Quant switch is unreachable (P2): every variant maps to the same dir. If quant is being disabled, remove the dead switch + the exposed options.
> 
> **Look closer:** this PR deletes ~615 lines of API tests but adds only the singleton test — confirm the `cancel` path stays covered. Also `enable-qnn` (desc) vs `enable-qnn-models` (code) naming, and the verbose `meson_options.txt` block already flagged by @Seunghui98.
> 
> **Bottom line:** engine-side changes are solid; blockers are the `init()` `return false` bug (fix here) and tracking the API P1s.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

The bugs that are associated with the QNN/APP side will be fixed in the following PRs.

Seunghui98 (approved):
The PR is too big😓, so I'm not sure if it's bug-free. 
But, if there will be a bug, I think it would be better to catch it then.
LGTM 👍

Applications/CausalLM/ANDROID_BUILD_UNIFICATION_PROGRESS.md
jaemini-shin, line 1: Please remove this unnecessary file

Applications/CausalLM/build_android.sh
```diff
@@ -40,70 +35,71 @@ SKIP_ENGINE=false
...
-        --legacy-ndk)
-            LEGACY_NDK_REQUESTED=true
+        --cache)
+            USE_BUILD_CACHE=true
             ;;
-        --assemble-aar|--aar)
-            APP_MODE_REQUESTED=true
-            ASSEMBLE_AAR=true
+        --skip-engine)
+            SKIP_ENGINE=true
```
jaemini-shin: Could you explain the difference between `--cache` and `--skip-engine`?

Applications/CausalLM/build_android.sh
```diff
@@ -40,70 +35,71 @@ SKIP_ENGINE=false
...
-        --skip-engine)
-            SKIP_ENGINE=true
-            ;;
-        --clean)
-            CLEAN=true
+        --install)
+            INSTALL=true
             ;;
-        --enable-qnn)
+        --qnn|--enable-qnn)
```
jaemini-shin: Did you intentionally keep deprecated options such as `--enable-qnn`?

Applications/CausalLM/build_android.sh
```diff
@@ -603,42 +655,69 @@ if [[ "$("${ADB[@]}" get-state 2>/dev/null)" != "device" ]]; then
...
+"${ADB[@]}" shell chmod 755 \
+    "$DEVICE_DIR/nntr_causallm" \
+    "$DEVICE_DIR/quick_dot_ai_test" \
+    "$DEVICE_DIR/nntr_quantize" \
+    "$DEVICE_DIR/nntr_safetensors_info"
 
-echo "=== Done (installed on $ANDROID_SERIAL) ==="
-echo "Device CLI: $DEVICE_DIR"
+if [[ "$BUILD_APP" == true ]]; then
+    echo "=== Done (native artifacts pushed and APK installed on $ANDROID_SERIAL) ==="
```
jaemini-shin, line 675: Is it necessary to push artifacts when app mode is on? If not, it'd be better to install app only.

dlwlzzero (commented):
It might be a minor point, why aren't the QNN models located inside the `Applications/CausalLM/models` folder?

Applications/CausalLM/build_android.sh
jayden0701: Removed it! Thanks

Applications/CausalLM/build_android.sh
jayden0701: unified it to --cache!

Applications/CausalLM/build_android.sh
jayden0701: Since app and cli use the same libs, same changes are applied to cli and app. 
So I thought updating the cli(and the native runtime) and app at the same time is resonable.

jayden0701:
> It might be a minor point, why aren't the QNN models located inside the `Applications/CausalLM/models` folder?

Since QNN build is optional, I thought all QNN-related codes(backend, models) should be in one folder to simplify the build process.

.github/workflows/static.check.scripts/doxygen-tag.sh
```diff
@@ -39,7 +39,7 @@ echo "::group::Doxygen tag check started"
 
 for file in `cat $files`; do
   if [[ `file $file | grep -E "(ASCII|Unicode) text" | wc -l` -gt 0 ]]; then
-      # In case of source code files: *.c|*.h|*.cpp|*.py|*.sh|*.php )
+      # Only C/C++ source files are checked.
```
jijoongmoon, line 42: I don't understand why you change the source code to be considered by doxygen. I means code documentation is necessary for every files including python, torch, and others as well.

jayden0701: For *.sh, *.php, there were actually no checks for them in the script. So I removed them in the comments.
For python, `doxygen-tag.sh` just checked if there is @package or @brief, and by looking closely, I found out all the current python files just wrote @brief with some C ctyle header. Which is not appropriate for python style doxygen.
And some python files (that were merged before this PR) didn't have any tags at all.

And by looking at the actual doxygen check by `doxygen-build.sh` , I found out it actually never outputs ERROR!
Which means the doxygen check never actually worked.

I could revive the python tag check, but I think remaking the CI related to doxygen might be better.
(e.g. make warn-as-error in doxygen-build, only on important parts like api)

Applications/CausalLM/build_android.sh
jaemini-shin, line 1: ### Comparison with previous `build_android.sh` + `install_android.sh`
* `/data/local/tmp/nntrainer/causallm` -> `/data/local/tmp/Quick.AI`
* No `$INSTALL_DIR/models`
* No `run_quantize.sh`
* No `run_safetensors_info.sh`
* Default NNTR_NUM_THREADS: 4 -> 7

---
Making `models` directory, missing scripts might be necessary.

jaemini-shin (approved):
LGTM!

.github/workflows/static.check.scripts/doxygen-tag.sh
jijoongmoon: If the doxygen format for python is not appropriate, then it is better way to make it better. not removing it. It is also better for reviewers and developers.

PR #4005, opened 2026-06-24
[CausalLM] Add V-JEPA 2.1 → LFM2 video-language model

Jungwon-Lee:
## Summary

Adds the **V-JEPA 2.1 + LFM2** vision-language model to the CausalLM application.

Pipeline: video → **VJEPA2ViT** (ViT-B/16 encoder, 3D axial RoPE) → **VjepaProjector** (pixel-unshuffle + MLP merger) → merge with text embeddings → **Lfm2CausalLM** (`run_with_embeddings`) → generate.

> **Note on scope / stacking:** this branch is built on top of the LFM2 work in #3998, so the diff here also contains the LFM2 commits. #3998 should land first (or be reviewed together); the V-JEPA-specific changes are the encoder/projector/VL-pipeline + flash-attention commits on top.

## Included

- **Causal flash attention** (`mha_core`): GQA + x86 AVX2 hsgemm, optional non-causal GEMM path, ALL-FP16 + ARM FP16 precision paths, `use_flash_attention` config option.
- **V-JEPA2 ViT encoder** + custom layers: `vjepa_rope` (3D axial RoPE), `vjepa_gelu`, `vjepa_layernorm` (thread-parallel, FP32/FP16).
- **VjepaProjector** (VoRA merger) and the **VjepaLfm2ForConditionalGeneration** pipeline (`run_video` / `run_video_bin`, `runVisionToLM`).
- Runtime wiring in `main.cpp` (`Lfm2VLVJepa21BModel` / `VJEPA2ViT`), factory + quantize registration, meson + **Android ndk-build** wiring.
- Q4_0-FP16 activation bring-up on the ViT, video/image utilities, sample video fixture, golden-test fixes.

## Excluded by design

The alternate SigLIP LFM2-VL path and ddtree speculative decoding from the source branch are **not** included (V-JEPA→LFM2 scope only).

## Notable base changes

Required for the FP16 activation path: removal of the forced-FP32 input-dtype promotion in core `embedding.cpp` / `input_layer.cpp` / `network_graph.cpp` (token-ID inputs are declared FP32 explicitly on the input layer instead). Base `Transformer` gains a `multimodal_pointer` typedef, an `initialize(const std::string&)` shim, and a public `getTokenizer()` accessor.

## Verification

- ✅ x86 build clean (`unittest_causallm_models` + `nntr_quantize`).
- ✅ CausalLM unit tests: **65/65 passing** (`run_unittest_x86.sh`).
[...]

PR #4094, opened 2026-07-20
[Activation] Integrate parallel FP16 GELU support

Jungwon-Lee:
## Summary
- parallelize exact-erf GELU over the global thread pool for large tensors
- fix the core GELU FP16 path by evaluating in FP32 and storing FP16 results
- add FP32 large-tensor and FP16 regression coverage

This folds the behavior introduced by `vjepa_gelu_layer` in #4005 into the existing core GELU activation, so V-JEPA can use the standard activation layer instead of maintaining a custom GELU implementation.

## Testing
- `git diff --check`
- compiled `unittest_nntrainer_activations.cpp` with `ENABLE_FP16=1` and `-Werror`
- attempted full macOS unit-test build; blocked by an existing unrelated error in `nntr_ggml_impl_fp16_fp32.cpp` (`vm_map_t` / `vm_address_t` undeclared)

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -18,6 +18,9 @@
 
 #include <common_properties.h>
 #include <cpu_backend.h>
+#include <thread_manager.h>
+
+#include <type_traits>
 
 #if defined(_WIN32)
 #define _USE_MATH_DEFINES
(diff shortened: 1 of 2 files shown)


Jungwon-Lee:
Review follow-up:

The FP32 chunking and FP16-in-FP32 implementation look correct, and the CI matrix is green. I still recommend moving the small-workload check before `ThreadManager::Global()`. At present, even `size < 4096` initializes the global pool and can create/pin worker threads before immediately selecting the serial path. This can regress first-call latency for small inference workloads.

Please also add before/after latency data around the threshold, at minimum for sizes 2048, 4095, 4096, and a representative V-JEPA tensor, with 1 and multiple compute threads. This is especially important because the threshold is otherwise not justified by measurements.

The existing unresolved thread describes the initialization issue in more detail: https://github.com/nntrainer/nntrainer/pull/4094#discussion_r3614633978

nntrainer/layers/acti_func.h
```diff
@@ -436,8 +439,40 @@ class ActiFunc {
...
+    if (thread_count <= 1 || size < 4096) {
+      gelu_chunk(0, size);
+      return t_out;
+    }
+
+    thread_manager.parallel_for(
+      0, static_cast<size_t>(thread_count), [=](size_t thread_index) {
+        gelu_chunk(size * thread_index / thread_count,
+                   size * (thread_index + 1) / thread_count);
+      });
```
jaemini-shin, line 475: Please do not bound the loop index to thread count. Please pass size with no adjustment.

PR #4095, opened 2026-07-20
[Activation] Parallelize SiLU and SwiGLU

Jungwon-Lee:
## Summary
- add a one-pass, thread-pool-parallel path for contiguous SiLU/Swish tensors
- evaluate FP16 SiLU in FP32 and cast only the final result
- parallelize CausalLM SwiGLU across tensor rows while preserving the existing NEON/AVX/FP16 kernels
- retain the existing strided-tensor fallback for Swish
- add large FP32 and FP16 SiLU regression coverage

The SwiGLU backend kernels are intentionally left single-call SIMD kernels because several MoE call sites already parallelize across tokens. Parallelism is added at the sequential `SwiGLULayer` call site to avoid nested `ThreadManager::parallel_for()` calls.

This PR is independent of #4094 and is based directly on `main`.

## Testing
- `git diff --check`
- compiled `unittest_nntrainer_activations.cpp` with `ENABLE_FP16=1` and `-Werror`
- compiled `Applications/CausalLM/layers/swiglu.cpp` with `ENABLE_FP16=1` and `-Werror`
- full macOS unit-test linking remains blocked by an existing unrelated error in `nntr_ggml_impl_fp16_fp32.cpp` (`vm_map_t` / `vm_address_t` undeclared)

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -18,6 +18,7 @@
 
 #include <common_properties.h>
 #include <cpu_backend.h>
+#include <thread_manager.h>
 
 #if defined(_WIN32)
 #define _USE_MATH_DEFINES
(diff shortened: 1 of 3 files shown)


Jungwon-Lee:
Blocking review findings:

1. The contiguous Swish fast path no longer preserves the existing empty-output behavior. A default `Tensor output;` is contiguous but has no allocation, so `t_out.getData<T>()` returns a null destination before the old `Tensor::apply()` path can allocate it. Please allocate the output or exclude empty outputs before entering the pointer loop. Existing thread: https://github.com/nntrainer/nntrainer/pull/4095#discussion_r3614667634

2. More importantly, this introduces a nested global-thread-pool deadlock in the existing Qwen3 MoE path. Qwen3 MoE already runs experts inside `ThreadManager::parallel_for()`, and each expert calls `acti_func.run_fn()` with Swish. Once an expert activation reaches 4096 elements, this new Swish path calls `parallel_for()` again. `ThreadManager` explicitly forbids nested `parallel_for()`, and its execution mutex makes this a deadlock rather than only oversubscription. Long prefill inputs can reach this condition.

Please choose one parallel level or add a reliable in-parallel serial fallback. The new SwiGLU row parallelism should also use a workload threshold; otherwise very small multi-row calls initialize and wake the pool without evidence that this is beneficial. A Qwen3 MoE long-prefill timeout regression test is needed before merge.

nntrainer/layers/acti_func.h
```diff
@@ -398,6 +399,33 @@ class ActiFunc {
...
+      auto &thread_manager = ThreadManager::Global();
+      const unsigned int thread_count = thread_manager.getComputeThreadCount();
+      if (thread_count <= 1 || size < parallel_threshold) {
+        swish_chunk(0, size);
+      } else {
+        thread_manager.parallel_for(
+          0, static_cast<size_t>(thread_count), [=](size_t thread_index) {
+            swish_chunk(size * thread_index / thread_count,
+                        size * (thread_index + 1) / thread_count);
+          });
```
jaemini-shin: Please do not bound the loop index to thread count.

Jungwon-Lee: Thanks for pointing this out. You are right—ThreadManager::parallel_for() already partitions the supplied range among the available workers, so manually limiting the range to thread_count and splitting the tensor is unnecessary. I have changed it to parallel_for(0, size, swish_at), while keeping the serial path for workloads below the threshold.

PR #4096, opened 2026-07-20
[LayerNorm] Add parallel width-axis fast path

Jungwon-Lee:
## Dependency of the PR

- Motivated by the V-JEPA LayerNorm implementation in #4005. This PR is based on `main` and has no code dependency on #4005.

<details><summary>[LayerNorm] add parallel width-axis fast path</summary><br />

[LayerNorm] add parallel width-axis fast path

- Add FP32 and FP16 width-axis LayerNorm backend entry points with scalar fallbacks and ARM NEON implementations.
- Keep FP16 input/output conversion at the boundary while accumulating mean, variance, and affine operations in FP32.
- Use the optimized path for contiguous inference tensors normalized over axis 3, with row-level ThreadManager parallelism.
- Preserve the generic implementation for training, non-contiguous tensors, other axes, and unsupported affine tensor types.

**Self evaluation:**

</details>

### Summary

- Integrate the V-JEPA width-axis LayerNorm behavior into the core LayerNormalization layer.
- Add portable FP32/FP16 fallbacks and ARM NEON kernels using the RMSNorm-style naming convention.
- Validate backend tail handling and FP32/FP16 LayerNormalization inference paths in Docker/Linux.

### DIFF
--- a/nntrainer/layers/layer_normalization_layer.cpp
+++ b/nntrainer/layers/layer_normalization_layer.cpp
@@ -22,6 +22,7 @@
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <node_exporter.h>
+#include <thread_manager.h>
 #include <util_func.h>
 
 namespace nntrainer {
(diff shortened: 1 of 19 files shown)


Jungwon-Lee:
Blocking correctness issue:

The fast-path predicate accepts any contiguous tensor, including NHWC. For NHWC with `channel() > 1`, logical width values are separated by a channel-sized stride, but `layerNormWrtWidthParallel()` treats `input.getData() + row * width` as a contiguous width row. It therefore mixes channel and width values when calculating mean/variance and applies gamma/beta to the wrong elements. Please restrict the path to NCHW or implement stride-aware processing. Existing thread: https://github.com/nntrainer/nntrainer/pull/4096#discussion_r3615071404

Please add an NHWC, channel>1, axis=3 inference test and an incremental-forwarding test. I also recommend replacing the `rows >= 4` parallel threshold with an element/byte-based threshold. The current test shape has only 48 elements but still wakes the pool, so small inference tensors can regress despite this being a performance PR.

nntrainer/layers/layer_normalization_layer.cpp
```diff
@@ -38,6 +39,52 @@ enum LNParams {
   temp_normalized_size,
 };
 
+static void layerNormWrtWidthParallel(const Tensor &input, Tensor &output,
+                                      const Tensor &gamma, const Tensor &beta,
+                                      size_t rows, size_t width,
+                                      float epsilon) {
+  auto &thread_manager = ThreadManager::Global();
+  const size_t thread_count = thread_manager.getComputeThreadCount();
+  const size_t tasks = rows >= 4 ? std::min(rows, thread_count) : 1;
```
jaemini-shin, line 48:
```suggestion
  const size_t tasks = rows;
```
Out threadpool gets good performance when tasks >> thread count. Please do not bound the tasks to thread count.

nntrainer/layers/layer_normalization_layer.cpp
```diff
@@ -38,6 +39,52 @@ enum LNParams {
...
+
+  auto run_rows = [&](size_t task) {
+    const size_t row_begin = rows * task / tasks;
+    const size_t row_end = rows * (task + 1) / tasks;
+    if (input.getDataType() == TensorDim::DataType::FP32) {
```
jaemini-shin, line 53: Please move this if clause out of the lambda. Conditional branch inside the hot path makes poor performance.
```cpp
    if (input.getDataType() == TensorDim::DataType::FP32) {
      tm.parallel_for(0, tasks, [&](size_t task){
        // layer_norm_wrt_width_fp32_intrinsic
      });
    }
    else {
     tm.parallel_for(0, tasks, [&](size_t task){
        // layer_norm_wrt_width_fp16_intrinsic
      });
    }
```

nntrainer/layers/layer_normalization_layer.cpp
```diff
@@ -147,6 +194,14 @@ void LayerNormalizationLayer::forwarding(RunLayerContext &context,
   Tensor &temp_full_size = output;
   Tensor &temp_norm_size = inv_std_dev;
 
+  if (!training && supportsLayerNormWrtWidthFastPath(input, output, gamma, beta,
+                                                     normalize_axes)) {
+    const size_t width = input.width();
+    const size_t rows = input.size() / width;
+    layerNormWrtWidthParallel(input, output, gamma, beta, rows, width, epsilon);
+    return;
+  }
```
jaemini-shin, line 203: We are assuming tensors are in NCHW format

nntrainer/layers/layer_normalization_layer.cpp
```diff
@@ -38,6 +39,52 @@ enum LNParams {
   temp_normalized_size,
 };
 
+static void layerNormWrtWidthParallel(const Tensor &input, Tensor &output,
+                                      const Tensor &gamma, const Tensor &beta,
```
jaemini-shin, line 43: Using pointer instead of Tensor might get better performance. We should create new Tensor object for every b, c loops.

jaemini-shin (commented):
Here are some points that could get better performance, PTAL!

Also, add doxygen brief tag for every new functions.

PR #4097, opened 2026-07-20
[Embedding] Optimize contiguous row lookup

Jungwon-Lee:
## Dependency of the PR

None. This PR is based on `main`.

<details><summary>[Embedding] optimize contiguous row lookup</summary><br />

Use direct row copies for contiguous FP32 and FP16 embedding tensors and parallelize sufficiently large prefill workloads across batch-token rows. Keep the tensor-view implementation as a fallback for unsupported layouts and data types.

**Self evaluation:**

</details>

### Summary

- Avoid per-token Tensor view construction for contiguous FP32/FP16 embedding lookup.
- Parallelize embedding row copies for prefill workloads of at least 64 KiB.
- Preserve the existing Tensor-based path for unsupported data types and layouts.
- Docker x86 FP16 build passed; all 15 `*Embedding*` layer tests passed.

### DIFF
--- a/nntrainer/layers/embedding.cpp
+++ b/nntrainer/layers/embedding.cpp
@@ -17,8 +17,10 @@
 #include <nntrainer_error.h>
 #include <nntrainer_log.h>
 #include <node_exporter.h>
+#include <thread_manager.h>
 #include <util_func.h>
 
+#include <cstring>
 #include <iostream>
 
 namespace nntrainer {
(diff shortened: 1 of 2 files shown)


Jungwon-Lee:
The direct-copy design is well scoped and should remove meaningful Tensor-view overhead, but this PR is not merge-ready while CodeQL is failing.

In the fallback loop, `get_index(b * token_count + i)` performs the multiplication as `unsigned int` before converting the result to `size_t`. Please use `get_index(static_cast<size_t>(b) * token_count + i)`. CodeQL report: https://github.com/nntrainer/nntrainer/runs/88404828409

The existing embedding tests also stay below the 64 KiB parallel threshold, so they validate direct copying but not the new parallel branch. Please add a workload above 64 KiB, including multiple batches and invalid-index validation, and rerun the cancelled/failed FP16 jobs. Before/after prefill latency would make the 64 KiB threshold defensible.

jaemini-shin (commented):
How much gain does this PR result? IMO, it looks hard to get a huge gain.

- For direct copy embedding, it's a memory bound operation. i.e. there's no huge gain with multi threading
- Threshold is too big: we usually expect width < 2K, which is much less than 64K
- Making `indices` buffer seems unnecessary process.

PR #4098, opened 2026-07-20
[Qwen3MoE] batch expert token projections

Jungwon-Lee:
## Dependency of the PR

None

<details><summary>[Qwen3MoE] batch expert token projections</summary><br />

Batch tokens assigned to the same expert so prefill uses matrix projections instead of per-token GEMMs. Keep the single-token path for decode, use compact expert outputs, and scatter results after parallel expert computation to avoid concurrent output accumulation.

**Self evaluation:**

</details>

### Summary

- Batch Qwen3 MoE expert projections for tokens routed to the same expert during prefill.
- Preserve the single-token decode path and scatter compact expert outputs after parallel computation.
- Avoid concurrent accumulation into the shared output tensor.

### DIFF
--- a/Applications/CausalLM/models/qwen3_moe/qwen_moe_layer.cpp
+++ b/Applications/CausalLM/models/qwen3_moe/qwen_moe_layer.cpp
@@ -25,6 +25,8 @@
 #include <acti_func.h>
 #include <algorithm>
 #include <cmath>
+#include <cstring>
+#include <functional>
 #include <node_exporter.h>
 #include <qwen_moe_layer.h>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


Jungwon-Lee:
Blocking correctness issue for quantized expert weights:

For multiple assigned tokens, `packed_input` is shaped as `[num_tokens, 1, 1, hidden_size]`. FP32 dot flattens the first three dimensions, so this works there. The Q4/Q6/QINT/QS4CX dot paths derive GEMM `M` from `getDim().height()`, which is 1 for this shape. As a result, only one token row is computed for a multi-token expert before the compact output is scattered. Existing thread: https://github.com/nntrainer/nntrainer/pull/4098#discussion_r3615377355

Please place tokens in the matrix-height dimension, for example `[1, 1, num_tokens, hidden_size]` for NCHW, or make all quantized dot paths use the same flattened-M calculation as FP32. The compact expert output shape should follow the same convention.

Test coverage is currently insufficient: the tiny Qwen3 MoE test zeros all expert projections, so incorrect batched expert math can still produce the expected zero residual, while the differential test skips when the FP32 weight fixture is absent. Please add a focused test with non-zero expert weights, at least two tokens routed to the same expert, and FP32 plus Q4_0 comparisons against the per-token reference.

The GEMV-to-GEMM change has strong prefill potential, but please report latency/tokens-per-second and peak RSS for sequence lengths 1, 8, 32, 128, and 512. Batched intermediates increase temporary memory even though compact expert outputs reduce the full-output allocation.

PR #4099, opened 2026-07-21
[test] Fix unittest_opencl_kernels_int4/qk_k to build against the cur…

jijoongmoon:
Both OpenCL int4/QK-K unit tests fail to compile on upstream/main with -Denable-opencl=true, for two unrelated reasons:

1. blas_kernels.h's INT4 GEMM entry points were renamed from openvino_gemm_cl/openvino_sgemm_cl/openvino_gemm_async_cl to gemm_int4_cl/sgemm_int4_cl/gemm_int4_async_cl ("Standardize Naming Conventions"), but unittest_opencl_kernels_int4.cpp was never updated and still calls the old (now nonexistent) symbols at 9 call sites, so the TU fails with "use of undeclared identifier".

2. Both unittest_opencl_kernels_int4.cpp and unittest_opencl_kernels_qk_k.cpp use a bare `Q4_0` as the Q4_0 block-size constant (`K % Q4_0`, `K * N / Q4_0`). No header defines a plain `Q4_0` in scope here: the only declaration reachable is the scoped enumerator `QScheme::Q4_0` in quantizer.h, which does not leak unqualified. The intended constant is `QK4_0` (= 32), a plain #define already reachable via the q4_0_utils.h include both files already carry.

Both fixes are straight identifier substitutions with no signature or behavior changes: 9 call sites renamed in unittest_opencl_kernels_int4.cpp (cross-checked against blas_kernels.h's current declarations), plus 3 occurrences (6 tokens) of Q4_0 -> QK4_0 in that same file and 1 occurrence in unittest_opencl_kernels_qk_k.cpp.

### DIFF
--- a/test/unittest/unittest_opencl_kernels_int4.cpp
+++ b/test/unittest/unittest_opencl_kernels_int4.cpp
@@ -37,8 +37,8 @@ static void run_dequantization_test_(const uint32_t K, const uint32_t N) {
     generate_random_vector<float>(N * K, -2.0f, 2.0f);
 
   // Dequantization Q4_0
-  if (K % Q4_0 == 0 && N % 8 == 0) {
-    size_t q4_data_size = K * N / Q4_0 * sizeof(block_q4_0);
+  if (K % QK4_0 == 0 && N % 8 == 0) {
+    size_t q4_data_size = K * N / QK4_0 * sizeof(block_q4_0);
     std::vector<uint8_t> q4_weight(q4_data_size);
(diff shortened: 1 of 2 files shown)


dlwlzzero (approved):
LGTM

jaemini-shin (approved):
LGTM!

Seunghui98 (approved):
LGTM 👍

Jungwon-Lee (approved):
LGTM

PR #4100, opened 2026-07-21
[causallm] Warn when a prompt is silently truncated to fit the prefil…

jijoongmoon:
CausalLM::run() clamps the tokenized prompt to num_allow_str (MAX_SEQ_LEN - NUM_TO_GENERATE) tokens with no diagnostic when the prompt is longer. Because the excess tokens are dropped from the END of the prompt, any trailing instruction (e.g. "... Now summarize the above.") can be cut entirely, and the model then continues the body instead of following the instruction — with nothing in the output or logs to explain why.

Warn to stderr whenever this clamp actually fires, printing the input token count, the allowed length, and the number of tail tokens dropped. This is a pure diagnostic: the clamp behavior and generated tokens are unchanged; a prompt within num_allow_str prints nothing extra.


jaemini-shin (approved):
LGTM!

dlwlzzero (approved):
LGTM

Seunghui98 (approved):
LGTM 👍

PR #4102, opened 2026-07-21
[build] Make DeepQ jsoncpp/libcurl dependencies optional

jijoongmoon:
Applications/ReinforcementLearning/DeepQ/jni/meson.build resolves jsoncpp and libcurl via dependency(...) with the default required: true. Since enable-app defaults to true and Applications/meson.build unconditionally subdirs into ReinforcementLearning/DeepQ/jni, any host missing libjsoncpp-dev or libcurl4-openssl-dev fails meson configure for the entire project, even when the user only wants libnntrainer, CausalLM, or nntr_quantize built.

Mark both dependencies required: false and skip the DeepQ executable target (via subdir_done()) with a warning naming the required apt packages when either is missing. No behavior change when both dependencies are present.

### DIFF
--- a/Applications/ReinforcementLearning/DeepQ/jni/meson.build
+++ b/Applications/ReinforcementLearning/DeepQ/jni/meson.build
@@ -6,8 +6,20 @@ run_command(['cp', '-lr', res_path, nntr_deepq_resdir], check: true)
 
 env_dir='../../Environment'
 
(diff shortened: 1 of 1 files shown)


jaemini-shin (approved):
LGTM!

dlwlzzero (approved):
LGTM

Jungwon-Lee (approved):
LGTM!

PR #4103, opened 2026-07-21
[CL] meson: install missing GPU headers + stop installing generated k…

jijoongmoon:
A from-scratch Android (ndk-build) app build failed because several OpenCL headers the app includes (via the installed <prefix>/include/nntrainer tree) were never added to the meson install lists:
  - nntrainer/opencl: opencl_buffer.h, opencl_loader.h, opencl_device_info.h
  - nntrainer/tensor/cl_operations: blas_kernels.h, attention_kernels.h These are pulled in transitively through cl_context.h and the blas/attention kernel interfaces, so the app could only build when stale copies happened to linger in the install dir (an x86-only-looking build in practice).

Also stop installing the generated per-kernel string headers (swiglu.h, geglu.h, ... and cl_kernels.h). They are INTERNAL to libnntrainer -- consumed by the generated *.cpp via the build-dir include path, never by consumers. Installing them shadows app-side layer headers of the same name (CausalLM's swiglu.h = SwiGLULayer), which broke the app build with a confusing collision.

Build-only change (install lists); no source/runtime behavior change.

### DIFF
--- a/nntrainer/tensor/cl_operations/cl_kernels/meson.build
+++ b/nntrainer/tensor/cl_operations/cl_kernels/meson.build
@@ -85,7 +85,12 @@ foreach k : cl_op_kernels
 
   # source.full_path() does not work: https://github.com/mesonbuild/meson/issues/5273#issuecomment-1851811417
   # can be used when meson is upgraded to 1.4.0
-  nntrainer_headers += meson.current_build_dir() / basename + '.h'
+  # NOTE: the generated per-kernel string headers (e.g. swiglu.h, geglu.h) are
+  # INTERNAL to libnntrainer (consumed by the *.cpp below via the build-dir
+  # include path). They are intentionally NOT added to nntrainer_headers, i.e.
+  # not installed: installing them into <prefix>/include/nntrainer shadows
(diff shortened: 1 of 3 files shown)


nntrainer/tensor/cl_operations/cl_kernels/meson.build
```diff
@@ -95,7 +100,6 @@ configure_file(
   output: 'cl_kernels.h',
   configuration: {'includes': all_includes},
 )
-nntrainer_headers += meson.current_build_dir() / 'cl_kernels.h'
 
 nntrainer_inc += include_directories('..')
 nntrainer_inc_abs += meson.current_build_dir() / '..'
```
jaemini-shin: Please add a new line to pass cpp linter.

jijoongmoon: Done. Thanks.

jaemini-shin (approved):
LGTM!

jayden0701 (approved):
LGTM!

Seunghui98 (approved):
LGTM 👍

Jungwon-Lee (approved):
LGTM!

baek2sm (approved):
LGTM

PR #4104, opened 2026-07-21
[nntrainer] Add the QS4CX-FP16 model tensor type (int4 weight, fp16 act)

jijoongmoon:
Upstream already supports the QS4CX weight type with an fp32 activation (WQS4CXA32 / "QS4CX-FP32"). This adds the fp16-activation sibling — WQS4CXA16 / "QS4CX-FP16" — to the ModelTensorDataTypeInfo enum, its EnumList, and EnumStr, so a model whose config declares a per-channel int4 weight with fp16 activations parses at load instead of failing with "No matching enum for value: QS4CX-FP16".

The enum is indexed (EnumList/EnumStr run in parallel with the enum values), so this is purely additive; no dtype-mapping switch changes (the same as the existing QS4CX-FP32 entry, which no code path switches on by name).

Repro:
  meson setup build -Denable-transformer=true -Denable-fp16=true \
    -Denable-opencl=true --buildtype=plain
  ninja -C build

### DIFF
--- a/nntrainer/models/model_common_properties.h
+++ b/nntrainer/models/model_common_properties.h
@@ -203,13 +203,15 @@ struct ModelTensorDataTypeInfo {
     WQ40A32,
     WQ40A16,
     WQS4CXA32,
+    WQS4CXA16,
   };
   static constexpr std::initializer_list<Enum> EnumList = {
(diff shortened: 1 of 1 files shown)


jaemini-shin (approved):
LGTM!

dlwlzzero (approved):
LGTM

PR #4105, opened 2026-07-21
[cpu_backend] Add whole-op ComputeOps dispatch surface + CPU baseline

jijoongmoon:
nntrainer already routes primitive kernels (sgemm, ele_add, softmax, ...) through the ComputeOps table so the arch/backend is picked at link time. This extends that pattern one level up, to whole tensor-level ops, so a single backend-neutral Layer can own structure/shape/orchestration while ComputeOps owns the whole kernel. Unlike the existing raw-pointer ops, these take Tensors, so a GPU ComputeOps (added in a later change) can introspect residency and bind device buffers directly; the CPU reference impls added here operate on the host buffers.

Interface (compute_ops.h / .cpp), each defaulting to a "not implemented" throw:
  geglu / swiglu / sigmoid_glu / sigmoid_add   activation-GLU whole-ops
  residual_op                                  one residual-add operand
  fc                                           fully-connected GEMM whole-op
  fc_prebuild_weight                           optional load-time weight xform
  apply_activation                             fused activation epilogue

CpuComputeOps gains reference implementations for all of them (cpu_ops_table.cpp). To let other backends inherit the CPU host paths (a CUDA UVM ComputeOps is host-coherent and reuses them), the CpuComputeOps class definition is moved out of cpu_ops_table.cpp into a new cpu_ops_table.h; the .cpp change is otherwise a mechanical extract (no behavior change to the existing raw-pointer overrides).

The one wired caller in this change is the fused-activation epilogue: a new `fused_activation` layer property (FusedActivation, distinct from the node-level `activation` realization property) lets FullyConnectedLayer and Conv2DLayer apply an activation in place after GEMM+bias via out.getOps()->apply_activation(...), instead of routing the output through a separate ActivationLayer node. This is value-identical to the standalone node and is opt-in: a layer that sets no fused_activation is byte-unchanged. Tensor gains a getOps() forwarder so the neutral layer reaches the right table without an #ifdef.

getComputeOps() is moved out-of-line (declaration in the header, definition in compute_ops.cpp): the previous inline body dereferenced the extern g_compute_ops data symbol from consumer modules, which does not export cleanly under a shared libnntrainer on Windows.

Behavior-preserving: the new whole-ops have no callers yet besides the opt-in fused-activation path, so default model graphs are unchanged. Verified with a gcc-13 -Werror plain build (enable-fp16=false, enable-test=true) and the regression suite: unittest_nntrainer_cpu_backend (53), unittest_nntrainer_tensor (267), unittest_nntrainer_tensor_advanced (26), unittest_layers (1020), unittest_nntrainer_models (125), unittest_models (156), unittest_safetensors_quantize (9) all pass.

Repro:
  ninja -C build
  meson test -C build unittest_nntrainer_cpu_backend unittest_layers

### DIFF
--- a/nntrainer/layers/common_properties.h
+++ b/nntrainer/layers/common_properties.h
@@ -1016,6 +1016,21 @@ public:
 };
 
 /**
+ * @brief FusedActivation — a layer-internal activation applied inline in the
+ *        compute layer's forward (after GEMM+bias), instead of as a separate
+ *        ActivationLayer node. Distinct key from `activation` (which is a
+ *        LayerNode *realization* property consumed by ActivationRealizer) so
+ * the FusionRealizer can move the activation onto the compute layer without it
+ *        being split back out into a node.
+ */
+class FusedActivation final
+  : public EnumProperty<nntrainer::props::ActivationTypeInfo> {
+public:
+  using prop_tag = enum_class_prop_tag;
+  static constexpr const char *key = "fused_activation";
+};
+
+/**
  * @brief HiddenStateActivation Enumeration Information
  *
  */
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -308,6 +308,19 @@ void Conv2DLayer::finalize(InitLayerContext &context) {
   NNTR_THROW_IF(context.getNumInputs() != 1, std::invalid_argument)
     << "Convolution layer takes only one input";
 
+  // The fusion is a forward-only epilogue: calcDerivative()/calcGradient() do
+  // not chain the activation derivative. Reject it wherever a backward pass
+  // can run instead of silently dropping that term, the same posture the
+  // tflite exporter takes at its own boundary (node_exporter.cpp).
+  if (auto &fused_act = std::get<props::FusedActivation>(conv_props);
+      !fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE) {
+    if (context.getExecutionMode() != ml::train::ExecutionMode::INFERENCE)
+      throw exception::not_supported(
+        "fused_activation on conv2d is inference-only: the backward pass does "
+        "not chain the activation derivative. Use a standalone activation "
+        "layer for training.");
+  }
+
   const TensorDim &in_dim = context.getInputDimensions()[0];
 
   auto &weight_regularizer =
--- a/nntrainer/layers/conv2d_layer.cpp
+++ b/nntrainer/layers/conv2d_layer.cpp
@@ -299,7 +299,8 @@ Conv2DLayer::Conv2DLayer(
   padding(padding_),
   conv_props(props::FilterSize(), std::array<props::KernelSize, CONV2D_DIM>(),
              std::array<props::Stride, CONV2D_DIM>(), props::Padding2D(),
-             std::array<props::Dilation, CONV2D_DIM>()) {
+             std::array<props::Dilation, CONV2D_DIM>(),
+             props::FusedActivation()) {
   wt_idx.fill(std::numeric_limits<unsigned>::max());
 }
 
@@ -475,6 +476,13 @@ void Conv2DLayer::forwarding(RunLayerContext &context, bool training) {
       throw std::invalid_argument("[Conv2D] adding bias failed");
     }
   }
+
+  // Fused activation epilogue dispatched through the op table (backend-
+  // neutral: CPU host ActiFunc / GPU kernel). Value-identical to the standalone
+  // ActivationLayer node it replaces.
+  auto &fused_act = std::get<props::FusedActivation>(conv_props);
+  if (!fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+    hidden_.getOps()->apply_activation(hidden_, (int)fused_act.get());
 }
 
 void Conv2DLayer::calcDerivative(RunLayerContext &context) {
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -50,6 +50,19 @@ FullyConnectedLayer::FullyConnectedLayer() :
 }
 
 void FullyConnectedLayer::finalize(InitLayerContext &context) {
+  // The fusion is a forward-only epilogue: calcDerivative()/calcGradient() do
+  // not chain the activation derivative. Reject it wherever a backward pass
+  // can run instead of silently dropping that term, the same posture the
+  // tflite exporter takes at its own boundary (node_exporter.cpp).
+  if (auto &fused_act = std::get<props::FusedActivation>(fc_props);
+      !fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE) {
+    if (context.getExecutionMode() != ml::train::ExecutionMode::INFERENCE)
+      throw exception::not_supported(
+        "fused_activation on fully_connected is inference-only: the backward "
+        "pass does not chain the activation derivative. Use a standalone "
+        "activation layer for training.");
+  }
+
   auto &weight_regularizer =
     std::get<props::WeightRegularizer>(*layer_impl_props);
   auto &weight_regularizer_constant =
@@ -338,6 +351,14 @@ void FullyConnectedLayer::incremental_forwarding(RunLayerContext &context,
         hidden_step.add_i(bias);
       }
     }
+
+    // Same fused activation epilogue as forwarding(), applied to this step's
+    // view only: hidden_ still holds the positions written by earlier
+    // incremental calls, so running it over the whole output would activate
+    // them a second time.
+    if (auto &fused_act = std::get<props::FusedActivation>(fc_props);
+        !fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+      hidden_step.getOps()->apply_activation(hidden_step, (int)fused_act.get());
   }
 }
 
--- a/nntrainer/layers/fc_layer.cpp
+++ b/nntrainer/layers/fc_layer.cpp
@@ -42,7 +42,8 @@ enum LORAParams { loraA, loraB, loraTmp, loraOut };
 FullyConnectedLayer::FullyConnectedLayer() :
   LayerImpl(),
   lora_scaling(1.0f),
-  fc_props(props::Unit(), props::LoraRank(), props::LoraAlpha()),
+  fc_props(props::Unit(), props::LoraRank(), props::LoraAlpha(),
+           props::FusedActivation()),
   quantizer(nullptr) {
   weight_idx.fill(std::numeric_limits<unsigned>::max());
   lora_idx.fill(std::numeric_limits<unsigned>::max());
@@ -249,6 +250,14 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
       hidden_.add_i(bias);
     }
   }
+
+  // Fused activation epilogue dispatched through the op table, so the
+  // fusion is backend-neutral: CpuComputeOps runs the host ActiFunc, a GPU
+  // ComputeOps can fuse it into the GEMM epilogue. Eliminates the separate
+  // ActivationLayer node; value-identical to it.
+  auto &fused_act = std::get<props::FusedActivation>(fc_props);
+  if (!fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+    hidden_.getOps()->apply_activation(hidden_, (int)fused_act.get());
 }
 
 void FullyConnectedLayer::incremental_forwarding(RunLayerContext &context,
--- a/nntrainer/layers/fc_layer.h
+++ b/nntrainer/layers/fc_layer.h
@@ -113,12 +113,15 @@ public:
 
 private:
   float lora_scaling;
-  std::tuple<props::Unit, props::LoraRank, props::LoraAlpha>
+  std::tuple<props::Unit, props::LoraRank, props::LoraAlpha,
+             props::FusedActivation>
     fc_props;                             /**< fc layer properties :
                                                 unit - number of output neurons,
                                                 lora_rank - rank of lora (optional)
                                                 lora_scaling - scaling factor of LoRA apply, i.e.,
-                                             lora_scaling = alpha / lora_rank */
+                                             lora_scaling = alpha / lora_rank
+                                                fused_activation - inline activation
+                                             applied after GEMM+bias (T10) */
   std::array<unsigned int, 2> weight_idx; /**< indices of the weights */
   std::array<unsigned int, 4> lora_idx;   /**< indices of the lora weights */
   std::unique_ptr<nntrainer::Quantizer> quantizer;
--- a/nntrainer/tensor/cpu_backend/compute_ops.h
+++ b/nntrainer/tensor/cpu_backend/compute_ops.h
@@ -40,6 +40,8 @@
 
 namespace nntrainer {
 
+class Tensor;
+
 /**
  * @class ComputeOps
  * @brief Abstract dispatch interface for tensor compute kernels.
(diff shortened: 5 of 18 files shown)


nntrainer/layers/common_properties.h
```diff
@@ -1015,6 +1015,21 @@ class Activation final
...
 };
 
+/**
+ * @brief FusedActivation — a layer-internal activation applied inline in the
+ *        compute layer's forward (after GEMM+bias), instead of as a separate
+ *        ActivationLayer node. Distinct key from `activation` (which is a
+ *        LayerNode *realization* property consumed by ActivationRealizer) so
+ * the FusionRealizer can move the activation onto the compute layer without it
+ *        being split back out into a node. [T10]
+ */
```
jaemini-shin, line 1025:
```suggestion
/**
 * @brief FusedActivation — a layer-internal activation applied inline in the
 *        compute layer's forward (after GEMM+bias),
 */
```
Most of the comments are too verbose and expose redundant tags such as `T10`.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -475,6 +476,13 @@ void Conv2DLayer::forwarding(RunLayerContext &context, bool training) {
       throw std::invalid_argument("[Conv2D] adding bias failed");
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table (backend-
+  // neutral: CPU host ActiFunc / GPU kernel). Value-identical to the standalone
+  // ActivationLayer node it replaces.
```
jaemini-shin:
```suggestion
```
ditto here

nntrainer/layers/fc_layer.cpp
```diff
@@ -249,6 +250,14 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
       hidden_.add_i(bias);
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table, so the
+  // fusion is backend-neutral: CpuComputeOps runs the host ActiFunc, a GPU
+  // ComputeOps can fuse it into the GEMM epilogue. Eliminates the separate
+  // ActivationLayer node; value-identical to it.
```
jaemini-shin:
```suggestion
```
ditto here

nntrainer/layers/fc_layer.cpp
```diff
@@ -249,6 +250,14 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
...
+  auto &fused_act = std::get<props::FusedActivation>(fc_props);
+  if (!fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+    hidden_.getOps()->apply_activation(hidden_, (int)fused_act.get());
```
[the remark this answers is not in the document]
jaemini-shin: Agree with this review. Shouldn't it be applied to `incremental_forwarding`, too?

nntrainer/layers/fc_layer.h
```diff
@@ -113,12 +113,15 @@ class FullyConnectedLayer : public LayerImpl {
...
+  std::tuple<props::Unit, props::LoraRank, props::LoraAlpha,
+             props::FusedActivation>
     fc_props;                             /**< fc layer properties :
                                                 unit - number of output neurons,
                                                 lora_rank - rank of lora (optional)
                                                 lora_scaling - scaling factor of LoRA apply, i.e.,
-                                             lora_scaling = alpha / lora_rank */
+                                             lora_scaling = alpha / lora_rank
+                                                fused_activation - inline activation
+                                             applied after GEMM+bias (T10) */
```
jaemini-shin, line 124:
```suggestion
                                             applied after GEMM+bias */
```
ditto here

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
                                               float *sin_);
 #endif // ENABLE_FP16
 
+  // ===========================================================================
+  // Whole-op (Tensor-level) ops — the §11 op_table pattern: a thin neutral
+  // Layer owns structure/shape/orchestration while ComputeOps owns the whole
+  // kernel. Unlike the raw-pointer ops above, these take Tensors so the backend
+  // impl can introspect residency (isClMem/getClMem/isSVM) and bind device
+  // buffers directly. Default throws; CPU/CL/CUDA subclasses override. [T7]
```
jaemini-shin:
```suggestion
  // Tensor-level ops
```
ditto here

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
...
+   *        per-row element count.
+   */
+  virtual void swiglu(const Tensor &in1, const Tensor &in2, Tensor &out,
+                      unsigned int active_rows, unsigned int row_offset);
+
+  /**
+   * @brief Sigmoid-GLU over the first `active_rows` rows starting at
+   *        `row_offset`: out = sigmoid(in1) * in2 ({gate, up} -> result).
+   *        in1/in2/out share shape; width() is the per-row element count.
+   *        gauss4 attention output gate.
```
jaemini-shin:
```suggestion
```

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
...
+   */
+  virtual void sigmoid_glu(const Tensor &in1, const Tensor &in2, Tensor &out,
+                           unsigned int active_rows, unsigned int row_offset);
+
+  /**
+   * @brief Sigmoid-add over the first `active_rows` rows starting at
+   *        `row_offset`: out = sigmoid(in1) + in2 ({gate, emb} -> result).
+   *        in1/in2/out share shape; width() is the per-row element count.
+   *        gauss4 PLE mix (method=1).
```
jaemini-shin:
```suggestion
```

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
...
+   */
+  virtual void sigmoid_add(const Tensor &in1, const Tensor &in2, Tensor &out,
+                           unsigned int active_rows, unsigned int row_offset);
+
+  /**
+   * @brief One residual-add operand: hidden = input (accumulate=false, the
+   *        first operand) or hidden += input (accumulate=true). The neutral
+   *        AdditionLayer calls this per input so the GPU backend can keep the
+   *        residual stream device-resident (cl_mem/SVM) while CPU/CUDA run the
+   *        host Tensor copy/add on the managed buffer.
```
jaemini-shin, line 428:
```suggestion
   * @brief One residual-add operand: hidden = input (accumulate=false, the
   *        first operand) or hidden += input (accumulate=true).
```
IMO, it doesn't not need to be exposed in virtual interface.

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
...
+  virtual void residual_op(Tensor &hidden, const Tensor &input,
+                           bool accumulate);
+
+  /**
+   * @brief Fully-connected GEMM: output = input * weight. The neutral
+   *        FullyConnectedLayer owns the weight/bias binding and calls this for
+   *        the matmul, so the quantized accelerator path (OpenCL v8c w4a8 /
+   *        CUDA cuda_fc_qint4) lives in the op table instead of a forked Layer.
+   *        input/weight may carry residency state the impl reads, hence
+   * non-const.
```
jaemini-shin, line 439:
```suggestion
   * @brief Fully-connected GEMM: output = input * weight.
```
ditto here

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
...
+   * @brief Fused activation epilogue: apply an element-wise activation in place
+   *        on a compute layer's output, after GEMM+bias. This is the op-table
+   *        half of the FusionRealizer (§10 T10): the neutral conv/fc layer
+   * calls out.getOps()->apply_activation(out, act) instead of routing the data
+   *        through a separate ActivationLayer node, so the fusion is
+   *        backend-neutral — CpuComputeOps runs the host ActiFunc, and a GPU
+   *        backend can override with a kernel that fuses it into the GEMM
+   *        epilogue. @p act_type is an nntrainer::ActivationType cast to int
+   * (kept as int here so compute_ops.h stays free of the layers headers);
+   * ACT_NONE is a no-op.
```
jaemini-shin, line 460:
```suggestion
   * @brief Fused activation epilogue: apply an element-wise activation in place
   *        on a compute layer's output, after GEMM+bias.
```
ditto here

nntrainer/tensor/cpu_backend/cpu_ops_table.cpp
```diff
@@ -16,350 +16,193 @@
...
+// OpenCL swiglu_cl kernel exactly (avoids the x*exp(x)/(1+exp(x)) overflow).
+inline float silu(float x) { return x / (1.0f + std::exp(-x)); }
+// sigmoid -- matches the OpenCL sigmoid_glu/sigmoid_add kernels and the CUDA
+// ELTWISE_SRC form (1/(1+exp(-x))) so the three backends agree token-for-token.
+inline float sigmoidf(float x) { return 1.0f / (1.0f + std::exp(-x)); }
+} // namespace
+
+// out = gelu_tanh(in1) * in2 over rows [row_offset, row_offset+active_rows).
+// row_offset is 0 on every current caller (the live token is at the buffer
+// base for the host/SVM/UVM paths); the offset is honored for generality.
```
jaemini-shin, line 51: These are duplicated comments with `compute_ops.h`. There is no need to increase the lines.

nntrainer/tensor/cpu_backend/cpu_ops_table.h
jaemini-shin, line 1: We need doxygen tags for each methods.

jaemini-shin (commented):
The overall logic looks fine, but comments are too verbose. Could you please share your opinion?

Seunghui98 (commented):
## Description

Extends the `ComputeOps` dispatch table from raw-pointer primitives to whole tensor-level ops (`geglu` / `swiglu` / `sigmoid_glu` / `sigmoid_add` / `residual_op` / `fc` / `fc_prebuild_weight` / `apply_activation`), adds CPU reference impls, extracts `CpuComputeOps` into a new `cpu_ops_table.h` so other backends can inherit the host paths, moves `getComputeOps()` out-of-line, adds `Tensor::getOps()`, and wires one caller: a new `fused_activation` layer property applied after GEMM+bias in `FullyConnectedLayer` and `Conv2DLayer`.

## Summary

- The op-table extraction and `getComputeOps()` out-of-lining look clean, and the header note explaining the Windows shared-build data-export problem is the kind of rationale that saves the next person from re-inlining it.
- `Tensor::getOps()` resolves against the existing `TensorBase::getOps()` (`tensor_base.h:355`), so the forwarder is well-founded.
- The tuple-arity change is handled properly on both sides: `node_exporter.h` and `.cpp` were updated together, and the tflite exporters throw `not_supported` when a fused activation is set rather than silently dropping it. That's easy to miss and it wasn't missed.
- `CpuComputeOps::apply_activation` builds the same `ActiFunc` the standalone node would and correctly branches on `supportInPlace()`, with a clear error when fp16 is requested without `enable-fp16`.

## Bug

**`FullyConnectedLayer::incremental_forwarding` never applies the fused activation.**

`forwarding()` gained the epilogue (`fc_layer.cpp:254-260`), but `incremental_forwarding()` ends at the bias add and returns — no `apply_activation` call anywhere in it. So one layer with one `fused_activation` value produces two different results depending on which entry point runs, and the incremental path returns pre-activation values.

Failure scenario: a `fully_connected` layer configured with `fused_activation=relu` inside a CausalLM graph. Batch inference through `forwarding()` applies ReLU; prefill/decode through `incremental_forwarding()` does not, so decode emits raw GEMM+bias output and disagrees with prefill on the same weights. It is latent today only because nothing sets the property yet — but the FusionRealizer described in the PR body is exactly the thing that will start setting it, and `incremental_forwarding` is the LLM path. `Conv2DLayer` has no incremental variant, so this is fc-only.

1. **The backward pass has the same gap.** `calcDerivative` and `calcGradient` are unchanged, so nothing chains the activation derivative when `fused_activation` is set. A standalone `ActivationLayer` node contributes that derivative itself; folding the activation into the fc/conv forward without the backward half makes gradients wrong for a training graph. The "value-identical to the standalone node" claim holds for forward inference but not for training — worth either implementing the backward half or rejecting `fused_activation` at `finalize()` when the layer is trainable, the way the tflite exporter already rejects it.

2. **The new whole-ops have no callers and no tests.** `geglu` / `swiglu` / `sigmoid_glu` / `sigmoid_add` / `residual_op` / `fc` ship with CPU reference impls that nothing in this PR exercises — no test file is touched, and the listed regression suites don't reach them. They'll sit unverified until the GPU PR lands and then be debugged through two layers at once. A small per-op smoke test against the existing equivalents (`swiglu` vs the current SwiGLU layer, `residual_op` vs `AdditionLayer`) would pin the contract now, while it's cheap.

3. **The new CPU impls trust their inputs.** They take dtype and stride from `in1` (`in1.width()`, `in1.getDataType()`) and then index `in2` and `out` with the same offsets, with no check that the three agree on shape or dtype. A caller passing an `out` of a different dtype gets a silent reinterpret rather than the `invalid_argument` the unsupported-dtype branch would give. Since these are the reference impls every backend will be checked against, a cheap size/dtype assert at the top seems worth it.

4. **Minor:** the PR body says the new interface methods each default to a "not implemented" throw, but `fc_prebuild_weight` defaults to an inline no-op (correctly so, given it's an optional transform). Only the description is off.

---

nntrainer/layers/fc_layer.cpp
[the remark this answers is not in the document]
Seunghui98: My claude's reivew is as follows. How about supporting fused activation for incremental_forwarding?

## Bug
**`FullyConnectedLayer::incremental_forwarding` never applies the fused activation.**

`forwarding()` gained the epilogue (`fc_layer.cpp:254-260`), but `incremental_forwarding()` ends at the bias add and returns — no `apply_activation` call anywhere in it. So one layer with one `fused_activation` value produces two different results depending on which entry point runs, and the incremental path returns pre-activation values.

Failure scenario: a `fully_connected` layer configured with `fused_activation=relu` inside a CausalLM graph. Batch inference through `forwarding()` applies ReLU; prefill/decode through `incremental_forwarding()` does not, so decode emits raw GEMM+bias output and disagrees with prefill on the same weights. It is latent today only because nothing sets the property yet — but the FusionRealizer described in the PR body is exactly the thing that will start setting it, and `incremental_forwarding` is the LLM path. `Conv2DLayer` has no incremental variant, so this is fc-only.

nntrainer/layers/fc_layer.cpp
jijoongmoon: done.

jijoongmoon:
Superseded by the re-cut series against current main: #4304. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

The review on this PR was harvested first, and every point is applied in #4304:

- **The reported bug is fixed.** `FullyConnectedLayer::incremental_forwarding` now applies the same fused-activation epilogue `forwarding()` does, so one layer with one `fused_activation` value no longer produces two different results depending on the entry point.
- **The backward gap is closed by construction.** The FusionRealizer is installed for inference compiles only. The fused forward is value-identical to the standalone node, but there is no fused backward, so a training graph keeps the separate `ActivationLayer` that contributes the activation derivative.
- **The tflite exporters follow the widened tuples.** `node_exporter`'s `fully_connected` and `conv2d` specializations were re-typed for the extra property and now refuse a node carrying a fused activation rather than exporting one that silently drops it.
- **The whole-ops have callers and tests.** `unittest_compute_ops_dispatch` covers `layer_norm`, `activation`, `supports_activation`, `residual_op`, `swiglu` and the unimplemented-op throw; `unittest_layers_layer_normalization` gains two inference-mode golden cases. `LayerNormalizationLayer`, `ActivationLayer`, `AdditionLayer`, `FullyConnectedLayerCl` and a new neutral `GeGLULayer` are the in-tree callers.
- **The CPU reference implementations validate their operands.** They now reject an `in2`/`out` that disagrees with `in1` on shape or dtype instead of reinterpreting it silently.
- **The description error is corrected.** `fc_prebuild_weight` is documented as defaulting to a no-op, not a throw.
- Comment volume was cut substantially in the re-cut; the remaining rationale is the part that explains a non-obvious constraint.

Branch preserved at `upstream-pr/trackE1-computeops-dispatch-shim`.

PR #4106, opened 2026-07-21
[engine] Open the backend registry + neutral device-capability probe

jijoongmoon:
Groundwork so a compute backend (OpenCL/CUDA/NPU) can register itself and
advertise its capabilities without the core needing a compile-time enum edit or
a static_cast to a concrete Context subtype. No behavior change: every new seam
is either a default that reproduces today's path or an inert tag nothing reads
yet.

Registry (T3):
- Context gains a non-template registerLayerFactory(PtrFactoryType<Layer>, key,
  int_key) virtual (base returns -1 "unsupported"); AppContext overrides it to
  forward to the existing per-class registerFactory<Layer>. Engine gains a
  registerLayerFactory(engine_name, ...) facade that resolves the target Context
  by name via getRegisteredContext() and dispatches to it — so a backend or the
  app can register a factory into a specific backend without a static_cast.
- The `engine=` layer property changes from a closed EnumProperty
  (ComputeEngineTypeInfo) to a registry-open Property<std::string>:
  LayerNode::compute_engine is now the context NAME, normalized against the live
  registry in Engine::parseComputeEngine (unknown/unavailable -> "cpu" fallback,
  matching what createLayerObject already resolved). getComputeEngineType()
  still returns the name string (unchanged return type), so existing callers
  (network_graph, neuralnet) are untouched. A CUDA row is added to the enum
  table for the compat setter, and toLayerComputeEngine() maps a registered name
  back to the residency-plane enum via the new Context::residencyEngine()
  (base=CPU) instead of the central name->enum string table.

Device capabilities (T1):
- A backend-neutral DeviceCaps POD (backend/arch/vendor/unified_memory/subgroups/
  compute_units/max_alloc/... — no cl_mem/CUDA types) with a one-line toString(),
  exposed via a Context::caps() virtual (base returns a default CPU snapshot).
  AppContext::initialize() logs it once. Log-only: no decision site reads it.

Tag threading (inert foundation):
- TensorSpecV2 gains engine (producing layer's LayerComputeEngine, default CPU)
  and a TensorRole hint (default GENERIC); InitLayerContext stamps the layer's
  engine onto its output specs at the single outSpec() chokepoint;
  RunLayerContext carries a run_engine. Nothing consumes these yet (the residency
  resolver that will is a later change), so output/allocation is byte-identical.

The per-backend concrete overrides (ClContext/CudaContext caps()/residencyEngine
probing), the ExecPlan resolver that will consume DeviceCaps, and the
TensorPool residency-class derivation that will consume the engine tag are each
deferred to their own follow-up changes; this one only lands the neutral base.

Depends-on: the ComputeOps whole-op dispatch change ("[cpu_backend] Add whole-op
ComputeOps dispatch surface + CPU baseline") — must merge first.

Verified with a gcc-13 -Werror plain build (enable-fp16=false, enable-test=true)
and the regression suite: unittest_nntrainer_appcontext (12),
unittest_nntrainer_models (125), unittest_models (156), unittest_layers (1020),
unittest_nntrainer_cpu_backend (53), unittest_nntrainer_tensor (267),
unittest_compiler (38), unittest_nntrainer_graph (12) all pass.

Repro:
  ninja -C build
  meson test -C build unittest_nntrainer_appcontext unittest_nntrainer_graph

### DIFF
--- a/api/ccapi/include/common.h
+++ b/api/ccapi/include/common.h
@@ -47,10 +47,11 @@ enum class ExecutionMode {
  * @brief     Enumeration of layer compute engine
  */
 enum LayerComputeEngine {
-  CPU, /**< CPU as the compute engine */
-  GPU, /**< GPU as the compute engine */
-  QNN, /**< QNN as the compute engine */
-  HTP, /**< HTP (Hexagon/HMX NPU) as the compute engine */
+  CPU,  /**< CPU as the compute engine */
+  GPU,  /**< GPU (OpenCL) as the compute engine */
+  QNN,  /**< QNN as the compute engine */
+  HTP,  /**< HTP (Hexagon/HMX NPU) as the compute engine */
+  CUDA, /**< CUDA (NVIDIA GPU) as the compute engine */
 };
 
 /**
--- a/api/ccapi/include/common.h
+++ b/api/ccapi/include/common.h
@@ -47,9 +47,10 @@ enum class ExecutionMode {
  * @brief     Enumeration of layer compute engine
  */
 enum LayerComputeEngine {
-  CPU, /**< CPU as the compute engine */
-  GPU, /**< GPU as the compute engine */
-  QNN, /**< QNN as the compute engine */
+  CPU,  /**< CPU as the compute engine */
+  GPU,  /**< GPU (OpenCL) as the compute engine */
+  QNN,  /**< QNN as the compute engine */
+  CUDA, /**< CUDA (NVIDIA GPU) as the compute engine */
 };
 
 /**
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -259,6 +259,10 @@ void AppContext::initialize() noexcept {
 
     add_default_object();
     add_extension_object();
+
+    // Log device capabilities once (log-only, docs/ARCHITECTURE_REFACTOR.md
+    // §10 T1). AppContext inherits the base CPU snapshot (host-coherent).
+    ml_logi("[AppContext] %s", caps().toString().c_str());
   } catch (std::exception &e) {
     ml_loge("registering layers failed!!, reason: %s", e.what());
   } catch (...) {
@@ -703,6 +707,15 @@ template const int AppContext::registerFactory<nntrainer::Optimizer>(
 template const int AppContext::registerFactory<nntrainer::Layer>(
   const FactoryType<nntrainer::Layer> factory, const std::string &key,
   const int int_key);
+
+// Non-template seam (Context::registerLayerFactory override): forwards to the
+// per-class registerFactory<Layer> here in the same TU so the explicit
+// instantiation is used and no template crosses the .so boundary.
+int AppContext::registerLayerFactory(PtrFactoryType<nntrainer::Layer> factory,
+                                     const std::string &key,
+                                     const int int_key) {
+  return registerFactory<nntrainer::Layer>(factory, key, int_key);
+}
 
 /**
  * @copydoc const int AppContext::registerFactory
--- a/nntrainer/context.h
+++ b/nntrainer/context.h
@@ -15,6 +15,7 @@
 #define __CONTEXT_H__
 
 #include <algorithm>
+#include <cstdint>
 #include <functional>
 #include <memory>
 #include <mutex>
@@ -39,6 +40,67 @@ namespace nntrainer {
 
 // ContextData lives in its own header so that layer_context.h / layer_node.h
 // can pull it in without triggering the context.h → layer_devel.h cycle.
+
+/**
+ * @struct DeviceCaps
+ * @brief Read-only snapshot of device capabilities, probed ONCE per backend at
+ *        Context init from real device queries (clGetDeviceInfo /
+ *        cudaGetDeviceProperties via the per-backend ContextManagers) rather
+ *        than from NNTR_* env flags. Currently LOG-ONLY — no decision site
+ *        reads it yet; it is the input the ExecPlan resolver will consume (see
+ *        docs/ARCHITECTURE_REFACTOR.md §10 T1/T4). Fields describe attributes
+ *        (what the device can do), never identity (who it is); unknown values
+ *        stay at the defaults below.
+ */
+struct DeviceCaps {
+  std::string backend = "cpu";  /**< "cpu" / "gpu" (OpenCL) / "cuda" */
+  std::string device_name = ""; /**< human-readable device name */
+  std::string arch = "";        /**< backend arch tag, e.g. "compute_120" */
+  uint32_t vendor_id = 0;       /**< OpenCL CL_DEVICE_VENDOR_ID; 0 = n/a */
+  bool integrated = true;       /**< host+device share one physical pool
+                                     (host-coherent); CPU = true */
+  bool unified_memory = false;  /**< single-pointer SVM/UVM available */
+  bool subgroups = false;       /**< OpenCL cl_intel_subgroups (XMX/DPAS) */
+  uint32_t compute_units = 0;   /**< OpenCL CL_DEVICE_MAX_COMPUTE_UNITS */
+  uint64_t max_alloc_bytes = 0; /**< per-alloc cap (CL MAX_MEM_ALLOC_SIZE);
+                                     0 = unknown/unbounded */
+  bool image_v8c = true;        /**< device uses the image2d v8c path (FC GEMM +
+                                     KV attention) rather than the cl_mem buffer
+                                     path. Both report CL_DEVICE_IMAGE_SUPPORT, so
+                                     this is not a clean query — it is set from
+                                     vendor_id at init (Intel NEO's compiler
+                                     rejects the integer-coord read_imageui v8c
+                                     kernel ⇒ buffer; Adreno/unknown ⇒ image). The
+                                     V8C_BUF cell of the resolver. */
+  bool dpas = false; /**< OpenCL cl_intel_subgroup_matrix_multiply_accumulate
+                          — the actual systolic-array/DPAS matrix engine
+                          (Xe2/Xe3 "Arc"/"Battlemage" and later). NOT the
+                          same as `subgroups`: cl_intel_subgroups is
+                          advertised by every Intel GPU since Gen9
+                          (Meteor-Lake Xe-LPG included) and has no matrix
+                          unit, so gating XMX on it silently ropes
+                          non-DPAS Intel iGPUs into the DPAS kernel
+                          (IGC emulates it — catastrophic slowdown). This
+                          is the real XMX-capability gate. Declared LAST
+                          so appending it leaves the other field offsets
+                          unmoved (ABI-safe for an app built against the
+                          old DeviceCaps). */
+
+  /**
+   * @brief One-line human-readable dump for the init-time log.
+   */
+  std::string toString() const {
+    std::ostringstream os;
+    os << "DeviceCaps{backend=" << backend << ", device=\"" << device_name
+       << "\", arch=" << (arch.empty() ? "-" : arch) << ", vendor_id=0x"
+       << std::hex << vendor_id << std::dec << ", integrated=" << integrated
+       << ", unified_memory=" << unified_memory << ", subgroups=" << subgroups
+       << ", image_v8c=" << image_v8c << ", dpas=" << dpas
+       << ", compute_units=" << compute_units
+       << ", max_alloc_bytes=" << max_alloc_bytes << "}";
+    return os.str();
+  }
+};
 
 /**
  * @class Context contains user-dependent configuration for  support
@@ -196,6 +258,65 @@ public:
    * @return return 0 for success
    */
   virtual int load(const std::string &file_path) { return 0; };
+
+  /**
+   * @brief Read-only device capability snapshot for this backend, probed once
+   *        at init. The base returns CPU caps (host-coherent, no accelerator);
+   *        ClContext / CudaContext override with a probed snapshot. LOG-ONLY
+   * for now (docs/ARCHITECTURE_REFACTOR.md §10 T1) — no decision site reads it
+   * yet, so adding/overriding it is byte-identical.
+   *
+   * @return const DeviceCaps& capabilities of the device backing this context
+   */
+  virtual const DeviceCaps &caps() const {
+    static const DeviceCaps
+      cpu_caps; // backend="cpu", integrated=true, defaults
+    return cpu_caps;
+  }
+
+  /**
+   * @brief Register a layer factory under this backend. The base default is
+   *        "unsupported" (returns -1); each concrete Context overrides it to
+   *        forward to its own registerFactory<Layer>. This is the non-template
+   *        seam that lets callers register a layer on any backend through
+   *        Engine::registerLayerFactory(engine, creator) WITHOUT a
+   *        static_cast to a concrete ClContext/CudaContext (whose
+   *        registerFactory is a per-class explicit template instantiation, an
+   *        ABI hazard across the .so boundary). [docs/ARCHITECTURE_REFACTOR.md
+   *        §10 T3 / §11 S1]
+   *
+   * @param factory layer creator (createLayer<T> result)
+   * @param key string key (empty ⇒ derived from the layer's getType())
+   * @param int_key integer key (-1 ⇒ auto-assigned)
+   * @return registered integer key, or -1 if the backend cannot register
+   */
+  virtual int registerLayerFactory(PtrFactoryType<nntrainer::Layer>,
+                                   const std::string & = "", const int = -1) {
+    ml_logw("[Context] this backend does not support layer registration");
+    return -1;
+  }
+
+  /**
+   * @brief Which residency plane (LayerComputeEngine) this backend's tensors
+   *        live on. This is the single authority the layer graph consults to
+   *        map a registered engine NAME onto the tensor/residency-plane enum
+   *        (toLayerComputeEngine, layer_node.cpp) — retiring the central
+   *        name→enum string table (ComputeEngineTypeInfo::EnumStr) in favour of
+   *        a per-context declaration. The base is CPU (host residency);
+   *        ClContext→GPU, CudaContext→CUDA, QNNContext→QNN override it. A new
+   *        aliased/added backend just declares its plane here, with no central
+   *        table to edit (add-only). [docs/ARCHITECTURE_REFACTOR.md §10 T3]
+   * @note  NEW vtable tail: appended after every pre-existing slot so a rebuilt
+   *        libnntrainer.so stays ABI-compatible with an app/ccapi built against
+   *        the old vtable. The QNN context lives in the libqnn_context.so
+   *        plugin, which subclasses Context, so that plugin must be rebuilt
+   *        alongside libnntrainer.so (an old plugin lacks this slot entirely —
+   *        calling residencyEngine() on it is UB).
+   * @return residency-plane enum backing this context's tensors
+   */
+  virtual ml::train::LayerComputeEngine residencyEngine() const {
+    return ml::train::LayerComputeEngine::CPU;
+  }
 
 private:
   /**
--- a/nntrainer/engine.cpp
+++ b/nntrainer/engine.cpp
@@ -98,6 +98,10 @@ Engine::parseComputeEngine(const std::vector<std::string> &props) const {
       // self-registers a Context (e.g. "npu", "exynos") then resolves with no
       // enum edit; an unknown/unavailable engine falls back to "cpu" instead of
       // resolving to a name that getRegisteredContext would later reject.
+      // An engine tag is therefore a registered Context NAME, not a device
+      // taxonomy: names are flat keys bound one-per-Context, so "gpu" and
+      // "cuda" are disjoint by construction and no tag is a superset of
+      // another. Nothing downstream should read a hierarchy into them.
       // [docs/ARCHITECTURE_REFACTOR.md §10 T3]
       std::string name = value;
       std::transform(name.begin(), name.end(), name.begin(),
--- a/nntrainer/engine.cpp
+++ b/nntrainer/engine.cpp
@@ -93,15 +93,17 @@ Engine::parseComputeEngine(const std::vector<std::string> &props) const {
     std::string key, value;
     int status = nntrainer::getKeyValue(prop, key, value);
     if (nntrainer::istrequal(key, "engine")) {
-      constexpr const auto data =
-        std::data(props::ComputeEngineTypeInfo::EnumList);
-      for (unsigned int i = 0;
-           i < props::ComputeEngineTypeInfo::EnumList.size(); ++i) {
-        if (nntrainer::istrequal(value.c_str(),
-                                 props::ComputeEngineTypeInfo::EnumStr[i])) {
-          return props::ComputeEngineTypeInfo::EnumStr[i];
-        }
-      }
+      // Validate against the LIVE registered-context name set, not the closed
+      // LayerComputeEngine enum + string list. A vendor backend that
+      // self-registers a Context (e.g. "npu", "exynos") then resolves with no
+      // enum edit; an unknown/unavailable engine falls back to "cpu" instead of
+      // resolving to a name that getRegisteredContext would later reject.
+      // [docs/ARCHITECTURE_REFACTOR.md §10 T3]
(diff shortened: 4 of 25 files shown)


nntrainer/layers/fc_layer.cpp
```diff
@@ -249,6 +250,14 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
...
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table, so the
+  // fusion is backend-neutral: CpuComputeOps runs the host ActiFunc, a GPU
+  // ComputeOps can fuse it into the GEMM epilogue. Eliminates the separate
+  // ActivationLayer node; value-identical to it.
+  auto &fused_act = std::get<props::FusedActivation>(fc_props);
+  if (!fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+    hidden_.getOps()->apply_activation(hidden_, (int)fused_act.get());
```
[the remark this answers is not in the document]
jayden0701: This problem is from #4105. If fixed in #4105, it will be okay

jayden0701 (commented):
https://github.com/jijoongmoon/nntrainer/blob/c701444798b17381121e2ad7296aedaa19f36fbb/nntrahttps://github.com/jijoongmoon/nntrainer/blob/c701444798b17381121e2ad7296aedaa19f36fbb/nntrainer/layers/layer_node.cpp#L808


I think `refinalize()` needs to pass engine tag too

just like in `finalize()` where it passed


```
  auto context = InitLayerContext(
    actual_input_dims, out_info, getInPlaceType() != InPlaceType::NONE,
    getName(), scope, max_norm, tensor_type, loss_scale, mode,
    toLayerComputeEngine(compute_engine));
```iner/layers/layer_node.cpp#L808

File: api/ccapi/include/common.h [hunk 2]
```diff
@@ -47,9 +47,10 @@ enum class ExecutionMode {
  * @brief     Enumeration of layer compute engine
  */
 enum LayerComputeEngine {
-  CPU, /**< CPU as the compute engine */
-  GPU, /**< GPU as the compute engine */
-  QNN, /**< QNN as the compute engine */
+  CPU,  /**< CPU as the compute engine */
+  GPU,  /**< GPU (OpenCL) as the compute engine */
+  QNN,  /**< QNN as the compute engine */
+  CUDA, /**< CUDA (NVIDIA GPU) as the compute engine */
```
Seunghui98, line 54:
```suggestion
  GPU_OPENCL,  /**< GPU (OpenCL) as the compute engine */
  QNN,  /**< QNN as the compute engine */
  GPU_CUDA, /**< CUDA (NVIDIA GPU) as the compute engine */
```

How about this? This seems to be more clearly distinguished.

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -475,6 +476,13 @@ void Conv2DLayer::forwarding(RunLayerContext &context, bool training) {
       throw std::invalid_argument("[Conv2D] adding bias failed");
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table (backend-
```
Seunghui98:
```suggestion
  // fused activation epilogue dispatched through the op table (backend-
```
I guess this is from claude's word.

nntrainer/layers/fc_layer.cpp
```diff
@@ -249,6 +250,14 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
       hidden_.add_i(bias);
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table, so the
```
Seunghui98:
```suggestion
  // fused activation epilogue dispatched through the op table, so the
```
ditto?

nntrainer/layers/fc_layer.h
```diff
@@ -113,12 +113,15 @@ class FullyConnectedLayer : public LayerImpl {
...
+  std::tuple<props::Unit, props::LoraRank, props::LoraAlpha,
+             props::FusedActivation>
     fc_props;                             /**< fc layer properties :
                                                 unit - number of output neurons,
                                                 lora_rank - rank of lora (optional)
                                                 lora_scaling - scaling factor of LoRA apply, i.e.,
-                                             lora_scaling = alpha / lora_rank */
+                                             lora_scaling = alpha / lora_rank
+                                                fused_activation - inline activation
+                                             applied after GEMM+bias (T10) */
```
Seunghui98, line 124:
```suggestion
                                             applied after GEMM+bias */
```
ditto

Seunghui98 (commented):
## Description

Opens the backend registry so a compute backend can register itself without a compile-time enum edit: `Context` gains non-template `registerLayerFactory`, `residencyEngine()` and a `caps()` device-capability probe; `Engine` gains a name-dispatched `registerLayerFactory` facade; the `engine=` layer property becomes a registry-open string validated against the live context set; and `TensorSpecV2` / `RunLayerContext` gain inert engine/role tags. Stacked on #4105, whose diff it carries.

## Summary

- The registry-open direction is sound, and the ABI reasoning is unusually careful for this kind of change — appending `CUDA` at the tail of `LayerComputeEngine`, appending `residencyEngine()` at the vtable tail, and routing registration through a non-template virtual to keep the per-class `registerFactory<Layer>` explicit instantiation from crossing the `.so` boundary are all the right calls.
- I checked for exhaustive `switch` statements over `LayerComputeEngine` that appending `CUDA` would break under `-Werror`; there are none in-tree.
- The `compute_engine` member is still cached from the property in `finalize()` (`layer_node.cpp:686-688`), so the string-typed member and `getComputeEngineType()`'s unchanged return type line up with the existing callers.

## Bug

**`Engine::registerLayerFactory`'s documented `-1` return is unreachable for an unregistered engine — it throws instead.**

```cpp
  // engine.h — "@return registered integer key, or -1 if the backend cannot register"
  int registerLayerFactory(const std::string &engine, ...) const {
    return getRegisteredContext(engine)->registerLayerFactory(creator, key, int_key);
  }
```

`getRegisteredContext` throws `std::invalid_argument` when the name is absent from `engines` (`engine.h`), so the `-1` path only ever fires when the context *is* registered but its base override declines. Failure scenario: the add-only vendor backend this facade exists for — a plugin calling `Engine::Global().registerLayerFactory("npu", createLayer<MyLayer>)` in a build or load order where the NPU context has not registered yet gets an uncaught `std::invalid_argument` at registration time, rather than the `-1` the contract promises it can branch on. Either document the throw or give the facade a non-throwing lookup and return `-1`.

1. **A stale QNN plugin against the new `residencyEngine()` vtable slot.** The doc comment states plainly that an old `libqnn_context.so` "lacks this slot entirely — calling `residencyEngine()` on it is UB". That plugin is loaded dynamically, so a version-skewed `.so` on a device is a realistic deployment, and `toLayerComputeEngine()` calls `residencyEngine()` on whatever `getRegisteredContext()` returns for *every layer* at `finalize()` and `configureRunContext()`. A hazard this specific deserves an actual guard at plugin load (an ABI/version symbol) rather than a comment — otherwise the failure mode is a jump through a garbage vtable entry with no diagnostic.

2. **The unknown-engine fallback is now silent.** `props::ComputeEngine` moves from `EnumProperty` (which rejected an unknown value outright) to `Property<std::string>`, and `parseComputeEngine` returns `"cpu"` for anything unregistered without logging. Turning `engine=gpu` on a CPU-only build into a graceful downgrade is a clear improvement over the old late throw from `getRegisteredContext` — but the same path now also swallows `engine=gpuu`, which used to be a hard error. One `ml_logw` when a non-empty requested name doesn't resolve would keep the graceful behavior and stop a typo from silently costing the accelerator.

3. **The `DeviceCaps` ABI claim is stronger than what holds.** The `dpas` comment says being declared last makes it "ABI-safe for an app built against the old `DeviceCaps`". That's true for reading pre-existing fields through the `const DeviceCaps&` that `caps()` returns, but the struct holds `std::string` members and appending a field changes `sizeof`, so any copy or any backend that constructs one across a version boundary is not safe. Worth narrowing the comment to the read-through-reference case so nobody relies on the broader reading.

4. **`toLayerComputeEngine()` uses `catch (...)` as normal control flow.** "Name not registered in this build" is an expected condition on the cpu-only path, not an exceptional one, and it's evaluated per layer at finalize. Beyond the cost, the bare `catch (...)` also swallows genuinely unexpected failures from `getRegisteredContext` and silently reports CPU residency. A non-throwing `tryGetRegisteredContext` would serve both this and item 1.

5. **None of the new seams has a caller or a test.** `isComputeEngineGPU` / `isComputeEngineCUDA` / `isComputeEngineCPU` and `Engine::registerLayerFactory` have zero in-tree callers, and no test file is touched — so the listed regression suites confirm "nothing broke" but not that any of this works. That matches the deliberate inert-foundation framing, though a small `unittest_nntrainer_appcontext` case registering a factory by engine name and asserting the resolved key would pin the facade's contract (including item 1) before the follow-ups build on it.

6. **Carried from #4105:** this PR's diff includes the fused-activation epilogue, and with it the `FullyConnectedLayer::incremental_forwarding` gap I raised on that PR — the activation is applied in `forwarding()` but not in the incremental path. Flagging it here only because it rides along in this diff; the discussion belongs on #4105.

---

nntrainer/app_context.cpp
```diff
@@ -259,6 +260,10 @@ void AppContext::initialize() noexcept {
 
     add_default_object();
     add_extension_object();
+
+    // Log device capabilities once (log-only, docs/ARCHITECTURE_REFACTOR.md
+    // §10 T1). AppContext inherits the base CPU snapshot (host-coherent).
```
Seunghui98, line 264:
```suggestion
    // AppContext inherits the base CPU snapshot (host-coherent).
```

I guess this is from claude's work.

nntrainer/context.h
```diff
@@ -197,6 +259,65 @@ class Context {
...
+  /**
+   * @brief Register a layer factory under this backend. The base default is
+   *        "unsupported" (returns -1); each concrete Context overrides it to
+   *        forward to its own registerFactory<Layer>. This is the non-template
+   *        seam that lets callers register a layer on any backend through
+   *        Engine::registerLayerFactory(engine, creator) WITHOUT a
+   *        static_cast to a concrete ClContext/CudaContext (whose
+   *        registerFactory is a per-class explicit template instantiation, an
+   *        ABI hazard across the .so boundary). [docs/ARCHITECTURE_REFACTOR.md
+   *        §10 T3 / §11 S1]
```
Seunghui98, line 286:
```suggestion
   *        ABI hazard across the .so boundary).
```

ditto

nntrainer/context.h
```diff
@@ -197,6 +259,65 @@ class Context {
...
+   *
+   * @param factory layer creator (createLayer<T> result)
+   * @param key string key (empty ⇒ derived from the layer's getType())
+   * @param int_key integer key (-1 ⇒ auto-assigned)
+   * @return registered integer key, or -1 if the backend cannot register
```
Seunghui98, line 291:
```suggestion
   * @return registered integer key, or throws if the backend cannot register
```
How about this?

`getRegisteredContext` throws `std::invalid_argument` when the name is absent from `engines` (`engine.h`), so the `-1` path only ever fires when the context *is* registered but its base override declines. 

```cpp
  // engine.h — "@return registered integer key, or -1 if the backend cannot register"
  int registerLayerFactory(const std::string &engine, ...) const {
    return getRegisteredContext(engine)->registerLayerFactory(creator, key, int_key);
  }

File: nntrainer/context.h [hunk 9]
```diff
@@ -197,6 +259,65 @@ class Context {
...
+  /**
+   * @brief Which residency plane (LayerComputeEngine) this backend's tensors
+   *        live on. This is the single authority the layer graph consults to
+   *        map a registered engine NAME onto the tensor/residency-plane enum
+   *        (toLayerComputeEngine, layer_node.cpp) — retiring the central
+   *        name→enum string table (ComputeEngineTypeInfo::EnumStr) in favour of
+   *        a per-context declaration. The base is CPU (host residency);
+   *        ClContext→GPU, CudaContext→CUDA, QNNContext→QNN override it. A new
+   *        aliased/added backend just declares its plane here, with no central
+   *        table to edit (add-only). [docs/ARCHITECTURE_REFACTOR.md §10 T3]
```
Seunghui98, line 308:
```suggestion
   *        table to edit (add-only).
```
I guess this is from claude's work.

nntrainer/engine.cpp
```diff
@@ -93,15 +93,17 @@ Engine::parseComputeEngine(const std::vector<std::string> &props) const {
...
-                                 props::ComputeEngineTypeInfo::EnumStr[i])) {
-          return props::ComputeEngineTypeInfo::EnumStr[i];
-        }
-      }
+      // Validate against the LIVE registered-context name set, not the closed
+      // LayerComputeEngine enum + string list. A vendor backend that
+      // self-registers a Context (e.g. "npu", "exynos") then resolves with no
+      // enum edit; an unknown/unavailable engine falls back to "cpu" instead of
+      // resolving to a name that getRegisteredContext would later reject.
+      // [docs/ARCHITECTURE_REFACTOR.md §10 T3]
```
Seunghui98, line 113:
```suggestion
```
ditto

nntrainer/engine.h
```diff
@@ -174,6 +174,29 @@ class Engine : public Singleton<Engine> {
...
   std::string parseComputeEngine(const std::vector<std::string> &props) const;
 
+  /**
+   * @brief Register a layer factory on a backend by engine name, without a
+   *        static_cast to a concrete Context. Resolves the engine to its
+   *        Context and dispatches through the Context::registerLayerFactory
+   *        virtual (each backend forwards to its own registerFactory<Layer>).
+   *        This is the registration facade for vendor add-only backends and
+   *        the Application layer. [docs/ARCHITECTURE_REFACTOR.md §10 T3 / §11
+   * S1]
```
Seunghui98, line 184:
```suggestion
   *        the Application layer.
```
ditto

nntrainer/app_context.cpp
```diff
@@ -704,6 +709,15 @@ template const int AppContext::registerFactory<nntrainer::Layer>(
   const FactoryType<nntrainer::Layer> factory, const std::string &key,
   const int int_key);
 
+// Non-template seam (Context::registerLayerFactory override): forwards to the
+// per-class registerFactory<Layer> here in the same TU so the explicit
+// instantiation is used and no template crosses the .so boundary. [T3]
```
Seunghui98:
```suggestion
// instantiation is used and no template crosses the .so boundary.
```
ditto

nntrainer/utils/base_properties.h
```diff
@@ -811,19 +811,23 @@ class TensorType final : public EnumProperty<nntrainer::TensorTypeInfo> {
...
+  static constexpr std::initializer_list<Enum> EnumList = {
+    Enum::CPU, Enum::GPU, Enum::QNN, Enum::CUDA};
+  static constexpr const char *EnumStr[] = {"cpu", "gpu", "qnn", "cuda"};
 };
 
 /**
- * @brief ComputeEngine Enumeration Information
- *
+ * @brief ComputeEngine name property — the node's backend as a registered
+ *        Context NAME (registry-open [ARCHITECTURE_REFACTOR.md §10 T3]: any
```
Seunghui98, line 821:
```suggestion
 *        Context NAME (registry-open: any
```
ditto

nntrainer/tensor/tensor_wrap_specs.h
```diff
@@ -96,6 +96,25 @@ typedef std::tuple<TensorDim, Initializer, bool, const std::string,
...
   VarGradSpec;
 
+/**
+ * @brief Semantic role of a tensor — a HINT for residency / placement
+ *        decisions (e.g. an attention KV cache wants image2d backing on Adreno;
+ *        weights stay device-resident). `GENERIC` by default; role-bearing
+ *        layers tag their tensors. Currently a FOUNDATION (Mem M3): carried on
+ *        the spec but NOT yet consumed — the residency resolver (M4) will read
+ *        it to replace the env/model-name heuristics. Inert today ⇒
+ *        byte-identical. [docs/ARCHITECTURE_REFACTOR.md §10 T5 / Mem M3]
```
Seunghui98, line 106:
```suggestion
 *        byte-identical.
```
I guess this is from claude's work.

nntrainer/tensor/tensor.h
```diff
@@ -2044,6 +2044,16 @@ class Tensor {
     return itensor_->getContextData();
   }
 
+  /**
+   * @brief Get the ComputeOps this tensor dispatches through.
+   *
+   * Priority: attached ContextData (per-vendor ops, e.g. ClComputeOps /
+   * CudaComputeOps) > the global CPU table. This is how a neutral Layer
+   * reaches the right backend's whole-op kernel without an #ifdef: e.g.
+   * `in1.getOps()->geglu(...)`. [T7]
```
Seunghui98:
```suggestion
   * `in1.getOps()->geglu(...)`.
```
ditto

nntrainer/tensor/cpu_backend/compute_ops.h
```diff
@@ -375,6 +377,90 @@ class ComputeOps {
                                               float *sin_);
 #endif // ENABLE_FP16
 
+  // ===========================================================================
+  // Whole-op (Tensor-level) ops — the §11 op_table pattern: a thin neutral
+  // Layer owns structure/shape/orchestration while ComputeOps owns the whole
+  // kernel. Unlike the raw-pointer ops above, these take Tensors so the backend
+  // impl can introspect residency (isClMem/getClMem/isSVM) and bind device
+  // buffers directly. Default throws; CPU/CL/CUDA subclasses override. [T7]
```
Seunghui98:
```suggestion
  // buffers directly. Default throws; CPU/CL/CUDA subclasses override.
```

nntrainer/context.h
```diff
@@ -40,6 +41,67 @@ namespace nntrainer {
...
 // can pull it in without triggering the context.h → layer_devel.h cycle.
 
+/**
+ * @struct DeviceCaps
+ * @brief Read-only snapshot of device capabilities, probed ONCE per backend at
+ *        Context init from real device queries (clGetDeviceInfo /
+ *        cudaGetDeviceProperties via the per-backend ContextManagers) rather
+ *        than from NNTR_* env flags. Currently LOG-ONLY — no decision site
+ *        reads it yet; it is the input the ExecPlan resolver will consume (see
+ *        docs/ARCHITECTURE_REFACTOR.md §10 T1/T4). Fields describe attributes
```
Seunghui98, line 51:
```suggestion
 *        reads it yet; it is the input the ExecPlan resolver will consume
 *        Fields describe attributes
```

nntrainer/context.h
```diff
@@ -40,6 +41,67 @@ namespace nntrainer {
...
+  uint64_t max_alloc_bytes = 0; /**< per-alloc cap (CL MAX_MEM_ALLOC_SIZE);
+                                     0 = unknown/unbounded */
+  bool image_v8c = true;        /**< device uses the image2d v8c path (FC GEMM +
+                                     KV attention) rather than the cl_mem buffer
+                                     path. Both report CL_DEVICE_IMAGE_SUPPORT, so
+                                     this is not a clean query — it is set from
+                                     vendor_id at init (Intel NEO's compiler
+                                     rejects the integer-coord read_imageui v8c
+                                     kernel ⇒ buffer; Adreno/unknown ⇒ image). The
+                                     V8C_BUF cell of the resolver. [T8] */
```
Seunghui98:
```suggestion
                                     V8C_BUF cell of the resolver. */
```

nntrainer/context.h
```diff
@@ -197,6 +259,65 @@ class Context {
    */
   virtual int load(const std::string &file_path) { return 0; };
 
+  /**
+   * @brief Read-only device capability snapshot for this backend, probed once
+   *        at init. The base returns CPU caps (host-coherent, no accelerator);
+   *        ClContext / CudaContext override with a probed snapshot. LOG-ONLY
+   * for now (docs/ARCHITECTURE_REFACTOR.md §10 T1) — no decision site reads it
```
Seunghui98, line 266:
```suggestion
   * for now  — no decision site reads it
```

nntrainer/engine.h
```diff
@@ -174,6 +174,29 @@ class Engine : public Singleton<Engine> {
...
+   *
+   * @param engine registered context name ("cpu"/"gpu"/"cuda"/...)
+   * @param creator layer creator (createLayer<T> result)
+   * @param key string key (empty ⇒ derived from getType())
+   * @param int_key integer key (-1 ⇒ auto-assigned)
+   * @return registered integer key, or -1 if the backend cannot register
```
Seunghui98, line 190:
```suggestion
   * @return registered integer key, or throws if the backend cannot register
```
ditto https://github.com/nntrainer/nntrainer/pull/4106/changes#r3663162584

nntrainer/layers/layer_node.h
```diff
@@ -991,16 +991,41 @@ class LayerNode final : public ml::train::Layer, public GraphNode {
...
+  bool isComputeEngineGPU() const;
+
+  /**
+   * @brief Query whether this node requests the CUDA compute engine via its
+   *        `engine` property (the additive NVIDIA backend). Same direct-read
+   *        rationale as isComputeEngineGPU(); used to route the graph's memory
+   *        allocator to CUDA Unified Memory.
+   * @return true iff the `engine` property is set and equals CUDA
+   */
+  bool isComputeEngineCUDA() const;
```
jaemini-shin, line 1040: These names are ambiguous. `cuda` is a subset of `gpu` but it treated as disjoint in this scope

nntrainer/layers/layer_node.cpp
```diff
@@ -132,29 +132,38 @@ class SharedFrom : public Name {
...
+static ml::train::LayerComputeEngine
+toLayerComputeEngine(const std::string &name) {
+  // A registered context DECLARES its own residency plane via
+  // Context::residencyEngine() -- the single authority, retiring the central
+  // name->enum string table below. Registry ALIASES collapse onto their backend
+  // for free: "npu" is registered as an alias of the "qnn" context
+  // (engine.cpp), so getRegisteredContext("npu") returns the QNN context and
+  // residencyEngine() reports QNN -- no per-alias special case here, and a new
+  // aliased/added backend just declares its plane in its Context override
+  // (add-only, no edit here). [docs/ARCHITECTURE_REFACTOR.md §10 T3]
```
jaemini-shin, line 151: Similar comments appeared again.

jaemini-shin (commented):
IMO, `gpu` and `cuda` should be renamed so that they could be entirely disjoint.

jijoongmoon:
Superseded by the re-cut series against current main: #4306. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

Review from this PR, applied in #4306:

- **`ClContext` / `QNNContext` never override `registerLayerFactory`, so the facade silently returns -1 for every non-CPU backend** (codex, P2 — also raised on #4108). Fixed: `ClContext` now overrides it exactly as `AppContext` does, forwarding to its own `registerFactory<Layer>` in the same translation unit. Verified with an OpenCL-enabled build.
- **Plan-specific tags in shipped comments** (`(T11)`, `§10 T*`, the `V8C_BUF` cell note) — every one removed; where the comment pointed at a real design rule it now cites the section of `docs/backend_guide/ARCHITECTURE_REFACTOR.md` that states it.
- **Comment alignment / verbosity** — the comments were re-flowed while re-authoring, and `clang-format-diff` over the whole range is clean.
- **`@return registered integer key, or -1 if the backend cannot register`** — corrected on both the `Context` virtual and the `Engine` facade: an unregistered engine name throws `std::invalid_argument` from `getRegisteredContext()`, and -1 only ever means a registered backend declined.
- **`GPU` → `GPU_OPENCL` / `CUDA` → `GPU_CUDA` in `LayerComputeEngine`** — not taken. Renaming existing members of a public ccapi enum breaks every downstream user of `LayerComputeEngine::GPU`, and the enum is now only the residency plane; the dispatch name is a string. Happy to revisit as a standalone API change if the maintainers want it.
- The `fc_layer.cpp` fused-activation findings belong to the op-table change and were fixed there (#4304): the epilogue now runs on the incremental path, and the fusion realizer is installed for inference compiles only, so a training graph keeps the standalone activation node with its derivative.

Branch preserved at `upstream-pr/trackE2-engine-registry-devicecaps`.

PR #4108, opened 2026-07-21
[context] Add the ExecPlan resolver over DeviceCaps + ModelFeatures (shadow)

jijoongmoon:
The resolver follow-up deferred from the engine-registry change: the single
place where device attributes (and the model's declared features) decide which
kernel families run, replacing scattered NNTR_* env flags and device-name
branches. SHADOW for now: it is resolved and logged once at init, but no
decision site reads it, so behavior is byte-identical.

Resolver (T4):
- GemmPath enum (CPU / DP4A / XMX / CUBLAS): which quantized-FC GEMM family a
  device should run, derived from DeviceCaps attributes only — never from an
  env flag or a device-name string match.
- ExecPlan POD (gemm_path, host_coherent, decode_gpu) with a one-line
  toString() for the shadow log.
- resolveExecPlan(const DeviceCaps &): pure function — cuda -> CUBLAS,
  gpu -> (dpas ? XMX : DP4A), else CPU; host_coherent mirrors caps.integrated.
- Only cleanly caps-derivable cells are resolved. Cells that are NOT a pure
  function of caps (kv backing image2d-vs-buffer — a driver compile quirk with
  no probe; queue-sync coherence overrides — wrong => garbage) deliberately
  stay conservative env overrides and are not shadowed.

Model features (T11):
- ModelFeatures POD: what the model IS, declared by the model itself (qk/v
  norms, MLP kind, norm placement, sliding window, KV-share/skip-prefill, dual
  head_dim, per-layer embedding, softcaps, LM-head scheme, decode-GPU
  eligibility, head_dim) — replaces is_<model> name proxies, so a new model is
  "set the flags" with no backend edit.
- resolveExecPlan(const DeviceCaps &, const ModelFeatures &) matcher overload:
  caps-derived cells come from the caps-only resolver; decode_gpu is granted
  only when the model asks for it AND the backend is a real GPU ("gpu"/"cuda").

AppContext::initialize() logs the caps-only resolution once next to the
existing DeviceCaps line (log-only). The per-backend concrete caps() probes
and the model-side getModelFeatures() declarations that feed the matcher land
in their own follow-up changes; unit tests (ExecPlanResolverTest, 14 cases)
cover both overloads and the toString dumps here.

Depends-on: the engine-registry change ("[engine] Open the backend registry +
neutral device-capability probe") — must merge first.

Verified with a gcc-13 -Werror plain build (enable-fp16=false,
enable-test=true) and the full meson suite: 43/43 test targets pass, including
unittest_nntrainer_appcontext (26 — 14 new ExecPlanResolverTest cases),
unittest_compiler (38), unittest_nntrainer_graph (12), unittest_layers (1020),
unittest_nntrainer_cpu_backend (53), unittest_nntrainer_tensor (267),
unittest_nntrainer_models (125), unittest_models (156).

Repro:
  ninja -C build
  meson test -C build unittest_nntrainer_appcontext

### DIFF
--- a/nntrainer/context.h
+++ b/nntrainer/context.h
@@ -31,6 +31,7 @@
 #include <layer.h>
 #include <layer_devel.h>
 #include <mem_allocator.h>
+#include <model_features.h>
 #include <optimizer.h>
 #include <optimizer_devel.h>
 
@@ -142,14 +143,24 @@ inline const char *toString(GemmPath p) {
  *        decision site reads it, so it is byte-identical.
  *
  *        Only cleanly caps-derivable cells are resolved here. Cells that are
- * NOT a pure function of caps stay env overrides for now and are NOT shadowed:
- *        - kv_backing (Adreno image2d vs Intel cl_mem buffer) — both advertise
- *          image2d; the split is the NEO "cannot compile read_imageui" quirk,
- *          which has no probe. (NNTR_V8C_BUF / NNTR_KV_IMG_ATTN)
- *        - queue sync (NNTR_XE3_SYNC) — a new-ISA coherence regression; wrong ⇒
- *          garbage, so it stays a conservative override.
+ *        NOT a pure function of caps stay env overrides for now and are NOT
+ *        shadowed:
+ *        - kv_backing (image2d vs cl_mem buffer) — both kinds of device
+ *          advertise image2d; the split is a compiler quirk (one vendor's
+ *          driver rejects the integer-coordinate read_imageui kernel), which
+ *          has no direct probe and is derived from the vendor id instead.
+ *        - the SVM coherence drain — a coherence regression whose failure mode
+ *          is wrong output, so it stays a conservative decision. It does have
+ *          a probe (the device's fine-grain-SVM capability), and is the first
+ *          cell scheduled to become resolver-authoritative.
  *        Model-dependent cells (head_dim attention path, KV-share/skip-prefill)
- *        arrive with ModelFeatures (T11).
+ *        arrive with ModelFeatures.
+ *
+ *        Plan ownership: once the resolver is authoritative, the resolved
+ *        ExecPlan belongs to the compiled MODEL (one plan per model instance,
+ *        resolved at compile from that model's ModelFeatures); a Context owns
+ *        only its DeviceCaps. Two models in one process then get two plans
+ *        over one caps.
  */
 struct ExecPlan {
   GemmPath gemm_path = GemmPath::CPU;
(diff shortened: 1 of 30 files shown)


nntrainer/layers/fc_layer.cpp
```diff
@@ -249,6 +250,14 @@ void FullyConnectedLayer::forwarding(RunLayerContext &context, bool training) {
...
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table, so the
+  // fusion is backend-neutral: CpuComputeOps runs the host ActiFunc, a GPU
+  // ComputeOps can fuse it into the GEMM epilogue. Eliminates the separate
+  // ActivationLayer node; value-identical to it.
+  auto &fused_act = std::get<props::FusedActivation>(fc_props);
+  if (!fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+    hidden_.getOps()->apply_activation(hidden_, (int)fused_act.get());
```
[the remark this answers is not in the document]
jayden0701: from #4105

nntrainer/layers/conv2d_layer.cpp
```diff
@@ -475,6 +476,13 @@ void Conv2DLayer::forwarding(RunLayerContext &context, bool training) {
       throw std::invalid_argument("[Conv2D] adding bias failed");
     }
   }
+
+  // [T10] fused activation epilogue dispatched through the op table (backend-
+  // neutral: CPU host ActiFunc / GPU kernel). Value-identical to the standalone
+  // ActivationLayer node it replaces.
+  auto &fused_act = std::get<props::FusedActivation>(conv_props);
+  if (!fused_act.empty() && fused_act.get() != ActivationType::ACT_NONE)
+    hidden_.getOps()->apply_activation(hidden_, (int)fused_act.get());
```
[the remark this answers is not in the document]
jayden0701: from #4105

nntrainer/context.h
```diff
@@ -102,6 +102,175 @@ struct DeviceCaps {
...
+ * @brief The backend's resolved execution decisions, derived from DeviceCaps.
+ *        This is the output of the ExecPlan resolver — the single place where
+ *        device attributes (and later ModelFeatures) decide which kernels run,
+ *        replacing scattered NNTR_* env flags. Currently a SHADOW
+ *        (docs/ARCHITECTURE_REFACTOR.md §10 T4): resolved + logged + asserted
+ *        ==  the current env-driven choice, but NOT yet authoritative — no
+ *        decision site reads it, so it is byte-identical.
+ *
+ *        Only cleanly caps-derivable cells are resolved here. Cells that are
+ * NOT a pure function of caps stay env overrides for now and are NOT shadowed:
```
jayden0701: Comment alignment needed

nntrainer/context.h
```diff
@@ -102,6 +102,175 @@ struct DeviceCaps {
...
+ *        - kv_backing (Adreno image2d vs Intel cl_mem buffer) — both advertise
+ *          image2d; the split is the NEO "cannot compile read_imageui" quirk,
+ *          which has no probe. (NNTR_V8C_BUF / NNTR_KV_IMG_ATTN)
+ *        - queue sync (NNTR_XE3_SYNC) — a new-ISA coherence regression; wrong ⇒
+ *          garbage, so it stays a conservative override.
+ *        Model-dependent cells (head_dim attention path, KV-share/skip-prefill)
+ *        arrive with ModelFeatures (T11).
```
jayden0701: I think plan-specific comments like (T11) should be removed.
Maybe using AI for making comments less verbose throughout the PRs might help

nntrainer/context.h
```diff
@@ -102,6 +102,175 @@ struct DeviceCaps {
...
+}
+
+/**
+ * @struct ModelFeatures
+ * @brief What the model IS, declared by the model itself (NOT inferred from a
+ *        model-name/`is_gemma2` proxy). This is the other half of the resolver
+ *        input: `ModelFeatures × DeviceCaps → ExecPlan`. Fields are attributes
+ *        (independent feature combos), so a new model is "set the flags" with
+ * no backend edit. Currently consumed only by the SHADOW matcher overload below
+ * (log-only, byte-identical). docs/ARCHITECTURE_REFACTOR.md §10 T11.
```
jayden0701: ditto

Seunghui98 (commented):
## Description

Third in the stack (#4105 → #4106 → this), whose diff it carries. The incremental part is confined to three files: `context.h` adds `GemmPath` / `ExecPlan` / `ModelFeatures` and the two `resolveExecPlan` overloads, `AppContext::initialize()` logs the caps-only resolution once, and `unittest_nntrainer_appcontext.cpp` gains 14 `ExecPlanResolverTest` cases.

## Summary

- `resolveExecPlan` is genuinely a pure function of its inputs — no env reads, no device-name substring matching. That was the stated goal of the change and it holds.
- Being explicit about which cells are deliberately *not* resolved (`kv_backing`, the Xe3 queue-sync override) and why is the right call, and documenting "no probe exists for this" beats quietly resolving it wrong.
- This is the first PR in the stack to ship tests with its new surface. 14 cases with `_p`/`_n` naming covering both overloads, the caps/matcher agreement invariant, and the `toString` dumps — a real improvement over #4105 and #4106, which added seams with no coverage at all.

## Bug

None. It's shadow-only, nothing reads the plan, and the resolver does what it says.

1. **The resolver dispatches on a free-form string, with a silent CPU default — and the PR directly beneath it added an enum authority for exactly this.** `resolveExecPlan` branches on `c.backend == "cuda"` / `== "gpu"`, where `DeviceCaps::backend` is an unvalidated `std::string` each backend's `caps()` fills in by hand. A vendor writing `"GPU"`, `"opencl"`, or a stray trailing space lands in the `else` and gets `GemmPath::CPU` with no diagnostic. Because this is shadow-only, that typo stays invisible until the resolver becomes authoritative — and then a real GPU silently runs CPU kernels, which is the hardest class of bug to notice from the outside. #4106 introduced `Context::residencyEngine()` as "the single authority" specifically to retire name-string→enum mapping; keying the resolver off that enum, or at least normalizing/validating `backend`, would stop the two mechanisms from drifting apart.

2. **There's no cell for the QNN/HTP backend.** `GemmPath` is CPU/DP4A/XMX/CUBLAS, but the tree already registers an HTP context (`htp_context.h`, wired in `engine.cpp`) and #4106's `residencyEngine()` names QNN as a residency plane. Today a QNN device falls through the `else` to `GemmPath::CPU`. That may well be intended for now, but it's worth stating in the enum comment — otherwise the first NPU integration reads the fallthrough as a defect rather than a deferral.

3. **`ModelFeatures::decode_rope_gpu` has no route to an `ExecPlan` cell.** It's documented as a strict subset of `decode_gpu` that drives `gpu_decode_rope` (gemma2 wants attn=GPU but rope=HOST), but the matcher sets only `p.decode_gpu`, and `ExecPlan` has no corresponding field. As it stands the distinction the comment describes cannot be expressed in the output. Either add the cell or mark it deferred, so the POD isn't read as a complete contract.

4. **The tests can't tell "resolved to CPU" from "fell through to CPU".** All 14 cases use exactly `"cpu"`, `"gpu"`, `"cuda"`, so the `else` branch is only ever reached by the input that's *supposed* to reach it. One case with an unrecognized or odd-cased backend — asserting whatever behavior you settle on in item 1 — would pin the single input that silently misroutes.

5. **Minor:** `ModelFeatures` lands 13 fields in a widely-included header with exactly one of them consumed by the matcher. That matches the declared foundation framing, but the struct's shape is being committed before its readers exist, so some churn is likely once the model-side `getModelFeatures()` declarations land.

---

nntrainer/context.h
```diff
@@ -40,6 +41,236 @@ namespace nntrainer {
...
 // can pull it in without triggering the context.h → layer_devel.h cycle.
 
+/**
+ * @struct DeviceCaps
+ * @brief Read-only snapshot of device capabilities, probed ONCE per backend at
+ *        Context init from real device queries (clGetDeviceInfo /
+ *        cudaGetDeviceProperties via the per-backend ContextManagers) rather
+ *        than from NNTR_* env flags. Currently LOG-ONLY — no decision site
+ *        reads it yet; it is the input the ExecPlan resolver will consume (see
+ *        docs/ARCHITECTURE_REFACTOR.md §10 T1/T4). Fields describe attributes
```
Seunghui98, line 52:
```suggestion
 *        reads it yet; it is the input the ExecPlan resolver will consume
 *        Fields describe attributes
```

I guess this is from claude's work.

nntrainer/context.h
```diff
@@ -40,6 +41,236 @@ namespace nntrainer {
...
+  uint64_t max_alloc_bytes = 0; /**< per-alloc cap (CL MAX_MEM_ALLOC_SIZE);
+                                     0 = unknown/unbounded */
+  bool image_v8c = true;        /**< device uses the image2d v8c path (FC GEMM +
+                                     KV attention) rather than the cl_mem buffer
+                                     path. Both report CL_DEVICE_IMAGE_SUPPORT, so
+                                     this is not a clean query — it is set from
+                                     vendor_id at init (Intel NEO's compiler
+                                     rejects the integer-coord read_imageui v8c
+                                     kernel ⇒ buffer; Adreno/unknown ⇒ image). The
+                                     V8C_BUF cell of the resolver. [T8] */
```
Seunghui98:
```suggestion
                                     V8C_BUF cell of the resolver. */
```

jijoongmoon:
Superseded by the re-cut series against current main: #4306. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

**The red on this PR was root-caused before the successor opened, and it was not this code.** The failing job was `meson_test (ubuntu-24.04)` (run 31096349777), and the failing case was `nntrainer_fallback_kleidiai.quant_nxk_qs8cx_f32_n32_k128` in `unittest_nntrainer_fallback` — a file this PR does not touch:

```
unittest_nntrainer_fallback.cpp:1169: Failure
The difference between q_f and ref is 0.0038874149322509766, which exceeds
scale / 2, where ... scale / 2 evaluates to 0.0038874098099768162.
```

The overshoot is 5.1e-9, about one ULP: an exact half-LSB rounding tie, from a generator seeded with `std::random_device`, so it is nondeterministic and hits the clang lane more often than the gcc lanes. main fixed the test itself in `578a8fba4` ("[test] Add epsilon to kleidiai quant test"), which changes exactly that line to `scale / 2 + TOLERANCE` — and it landed thirteen minutes after this job finished, one commit past this PR's base. Sibling rungs that ran after it were green on the same leg. That fix is in the base of #4306 and the leg is green locally; no change to the resolver was needed.

Review from this PR, applied in #4306: the `registerLayerFactory` override missing on the non-CPU contexts is fixed (`ClContext` now overrides it); the plan-specific comment tags are gone; the comment alignment and verbosity were addressed while re-authoring, and `clang-format-diff` over the range is clean. `ModelFeatures` keeps its own dependency-free header. The `fc_layer` / `conv2d_layer` fused-activation findings belong to the op-table change and were fixed there (#4304).

Branch preserved at `upstream-pr/trackE3-execplan-resolver`.

PR #4111, opened 2026-07-21
[opencl] Add the XMX/DPAS int8xint4 prefill GEMM (Intel Xe2/Xe3)

jijoongmoon:
Drop-in fast path for the M>4 buffer-path v8c FC GEMM on Intel GPUs with a
real matrix engine: raw-nibble weights + the i8_u8 DPAS matrix-MAD builtin
accumulate acc_raw, then the SAME bias-correction + scale epilogue as the
dp4a kernel — byte-identical output, ~1.8x faster prefill on DPAS parts.

Kernel (int8_int8_gemm_xmx.cl):
- intel_sub_group_i8_u8_matrix_mad_k32 over the unmodified v8c weight
  packing; weights via transposed 2D block reads, activations via 8x32 2D
  block reads. Shape-adaptive tile (MT/NT/SG_M compile options; NT=2/SG_M=4
  measured uniformly best in-model across the validated LLMs, NNTR_XMX_NT /
  NNTR_XMX_SGM override for tuning). M is padded by the grid; an M_valid
  store guard keeps padded rows from landing.
- All three required extensions are declared explicitly
  (subgroup_matrix_multiply_accumulate, subgroup_2d_block_io, subgroups) so
  a compiler without the matrix engine or the block-io builtins rejects the
  build cleanly instead of software-emulating it. The header comment now
  names the actual file.

Dispatch (blas_kernels.cpp, inside gemm_int8_v8c_cl):
- Engages only when NNTR_FC_XMX is on AND the device actually has DPAS
  (DeviceCaps::dpas, i.e. cl_intel_subgroup_matrix_multiply_accumulate) AND
  the shape fits (M>4, buffer path, N%64==0, K%64==0); anything else falls
  through to the dp4a kernel. The capability gate deliberately does NOT use
  cl_intel_subgroups (advertised by every Intel GPU since Gen9, including
  matrix-engine-less Xe-LPG parts, where the emulated DPAS builtin ran at a
  fraction of dp4a's speed). NNTR_FC_XMX=0 force-disables;
  NNTR_FC_XMX_FORCE=1 (value-checked) bypasses the capability check for
  debugging. A one-shot warning names the device when XMX was
  requested/defaulted but no DPAS engine exists, and a one-shot gate
  diagnostic reports why XMX was or was not selected (quiet by default).
- ClContext's Intel env-default block turns NNTR_FC_XMX on (overwrite=0)
  only when caps().dpas is present.

suite 43/43 green; enable-opencl suite identical to the parent branch
(43 Ok / 2 fail / 2 timeout, unittest_layers failing set unchanged at the 3
pre-existing cases). The kernel source is the field-validated one from the
downstream tree; runtime engagement is exercised by model-level runs (an
int4-weight LLM prefill), not by this unit suite.

Repro:


nntrainer/tensor/cl_operations/cl_kernels/int8_int8_gemm_xmx.cl
```diff
@@ -0,0 +1,207 @@
...
+    // first; IGC overlaps them with the independent MMA chain below.
+    uint8 wraw[NT];
+#pragma unroll
+    for (int j = 0; j < NT; ++j) {
+      uint draw[8];
+      intel_sub_group_2d_block_read_transpose_32b_16r8x1c(
+          (__global void *)wgt, K / 2, N, K / 2,
+          (int2)(k0 >> 3, n0 + j * 16), draw);
+      wraw[j] = *(uint8 *)draw;
+    }
```
jayden0701, line 152: By the gate `(M>4, buffer path, N%64==0, K%64==0)` , `K==64` can come here. And by `K / 2`, it will be sent as `width == 32`

But by looking at Restrictions in https://www.intel.com/content/www/us/en/developer/articles/technical/opencl-20-shared-virtual-memory-overview.html , it restricts `width >= 64bytes` . Adding `K >= 128` to the gate might be needed.

jijoongmoon:

Detail, so nothing is lost. This branch is a seven-rung stack, and the lower rungs are already open separately as #4304, #4306, #4308, #4309 and #4310; #4311 carries only this branch's own top two commits, the XMX/DPAS kernel and the `K >= 128` gate, folded into one commit because the second only corrects the first.

Two things changed on the way across. The `ClContext` `setenv` block that defaulted `NNTR_FC_XMX` on is gone: the default is derived from the probed capability at the one site that reads it, which is behaviour-identical and keeps a `setenv` shim out of the series. And the profiling label on the dispatch is stripped — that machinery belongs to the infrastructure PR.

Three of the review's findings were live on the code and are fixed in #4311's last commit:

- **(1) unvalidated `atoi` on the tuning env vars.** Correct and reproduced: `NNTR_XMX_SGM=0` divides by zero. Both overrides are now `strtol`-parsed, range-checked, and an out-of-range or non-numeric value is logged and ignored. Verified on an Xe DPAS device: `NNTR_XMX_SGM=0` and `NNTR_XMX_NT=abc` now fall back to the default tile with byte-identical output instead of crashing.
- **(2) the kernel ignores the weight-pitch argument.** Correct, and closer to live than the review says: `buf_kernel` is `use_buf || fc_buf_probe`, so the probe override can select the buffer kernel while the row stride is the padded one. The dispatch gate now requires `v8c_wrow_bytes(K) == K/2` rather than assuming it.
- **(4) "shape-adaptive tile" is not adaptive.** Correct; the comment was the wrong half. It now says the tile is one measured default and why the per-shape sweep results were not adopted.

Left open, and said so on #4311: **(3)** there is still no automated A/B of the DPAS kernel against dp4a. What exists is a hand-run differential on an Xe iGPU that does advertise `cl_intel_subgroup_matrix_multiply_accumulate` — the program compiles, the gate opens, and the output is byte-identical to dp4a at 32x512x512, 32x1536x6144, 64x2048x2048 and 8x4096x1024, with K=64 correctly declining. Items 5 (type punning), 6 (non-atomic one-shot log flags) and 8 (the GRF flag) were read and judged not worth changing a field-validated kernel for; 7 is fixed — `NNTR_FC_XMX_FORCE=1` now logs every time it is honoured.

The branch is preserved at `upstream-pr/trackG3-xmx-dpas`, so this is reversible.

PR #4112, opened 2026-07-21
[opencl] Add the GPU attention/RoPE/KV kernels + GLU whole-ops, fix rmsnorm tails

jijoongmoon:
The attention-side GPU compute paths on the OpenCL infrastructure + v8c FC
PRs: fp16 attention kernels (two-conv staged and flash), the in-place RoPE +
KV scatter/gather machinery they feed on, GeGLU/SwiGLU whole-op dispatches,
and standalone rmsnorm/swiglu kernel bugfixes.

Attention kernels:
- two_conv_attention.cl: staged QK -> softmax -> SV fp16 attention in image
  and buffer (-DTCA_BUFFER_ONLY) families, with kv-int8, OHWI-layout, and
  image-view variants for the KV-mirror path; fused_row single-row variant.
- flash_attention.cl: one-pass online-softmax prefill (LDS tree or
  sub_group_reduce_add reduction, buffer-path default = SG) and a split-KV
  two-pass flash decode (per-(head,chunk) partials + online-softmax reduce)
  for the M=1 decode-parallelism cliff.
- attention_kernels.cpp host dispatchers: in-place fp16 RoPE over a
  device-cached cos/sin LUT; K/V scatter into OHWI-layout KV mirrors and the
  matching gathers; device KV-mirror creation + image views; GPU fp16 copy
  helpers. cl_mem-resident operands arrive via defaulted parameters (the
  residency planner wires them in a later change); SVM/host binding works
  today. The base (non-OHWI) two-conv prefill wrapper ships default-off
  behind NNTR_MHA_GPU_FORCE_BROKEN: its per-layer drift amplifies on small
  models, and the honest default is the CPU fallback until it is re-validated
  (the gate names the env to re-test).
- Programs prewarm at context init (softmax/quant/rope/norm/lm_head) so
  their one-time builds stay out of the first timed prefill.

GLU whole-ops:
- geglu_cl_op / swiglu_cl_op: Tensor-level dispatches (gelu_tanh/silu gate x
  up) bound SVM-direct on the SVM pool, host-bounce otherwise, with
  active_rows/row_offset for decode live-row processing. ClComputeOps gains
  geglu/swiglu overrides forwarding to them — inert until the neutral GLU
  layers land (model-track change), same forward-infrastructure pattern as
  the earlier PRs.

Kernel fixes (standalone, affect existing callers):
- rmsnorm.cl: the float4 reduction silently DROPPED the tail elements when
  width % 4 != 0 (and produced nothing for width < 4); add scalar tails for
  the sum and the output store. FIXES the pre-existing RMSNorm-CL golden
  failure in unittest_layers (its 2:3:3:3 shape has width 3).
- rmsnorm_fp16.cl: fp16 coop-variant counterpart of the same contract.
- swiglu.cl / swiglu_fp16.cl: numerically stable SiLU — the
  x*exp(x)/(1+exp(x)) form overflows to NaN for x > ~88; use
  x/(1+exp(-x)).
- addition_cl work-group: lws was the placeholder {1,1,1} (one work-item per
  work-group, ~1/64 of the SIMD wave); use 64 with a guarded rounded-up
  global size. addition_cl gains an SVM-direct in-place accumulate branch
  (use_svm), used by add_i_cl when both tensors are SVM-resident.
- rmsnorm_layer_cl: route through the fp16 whole-op with skip_prefill
  support and shape-tail correctness on the CL layer path.

Env defaults at ClContext init (setenv overwrite=0, explicit env always
wins): NNTR_MHA_GPU on for Intel and Adreno (no host NEON on the Intel
hosts; measured GPU-favored on both), NNTR_KV_IMG_ATTN on for Adreno
(image2d KV/attention family).

Validated on the Intel iGPU host (gcc-13 -Werror, both configs):
- Non-OpenCL build: full meson suite 43/43 green.
- enable-opencl build: -Werror clean; suite topline unchanged
  (43 Ok / 2 fail / 2 timeout); unittest_attention_kernels_cl passes;
  unittest_layers' failing set SHRINKS again, 3 -> 2: the RMSNorm-CL golden
  that fails on every ancestor branch PASSES with the width-tail fix. The
  remaining two are pre-existing (one enable-clblast=false stub throw, one
  SwiGLULayerCl golden numeric that predates this series and whose layer is
  replaced by the neutral-GLU model-track change).
- The OpenCL unittest compile-fix change ("[Test] Fix the OpenCL int4/qk_k
  unittests") was applied for validation only (not part of this change).

Repro:


nntrainer/tensor/cl_operations/cl_kernels/two_conv_attention.cl
```diff
@@ -0,0 +1,1149 @@
...
+  const int head_q = get_global_id(2);
+  if (m >= M || x0 >= d) return;
+  const int head_kv = head_q / gqa;
+
+  const sampler_t smp = CLK_NORMALIZED_COORDS_FALSE | CLK_ADDRESS_CLAMP |
+                        CLK_FILTER_NEAREST;
+  const int v_row0 = head_kv * d + x0;
+  const long score_base =
+      (long)head_q * (long)M * (long)N_kv + (long)m * (long)N_kv;
+  // Causal: scores[m][n]=0 for n > the query's ABSOLUTE position, so only the
```
jayden0701, line 900: I think this function is hardcoded for causal mode only, can there be cases where this function can be used in non-causal mode? It seems some functions are able to select causal/non-causal, whereas this function doesn't

jijoongmoon:
Superseded by the re-cut series against current main: #4312. Closing to keep the queue reviewable — please review the new series instead.

This PR bundles three independent pieces, and they land in three places, so the mapping is worth stating explicitly rather than leaving it to be discovered:

1. **The GLU whole-ops** (`geglu_cl_op` / `swiglu_cl_op`, their kernels, and the `ClComputeOps` overrides that forward to them) shipped in #4308, the OpenCL whole-op table PR.
2. **The GPU attention, RoPE and KV kernels** — `two_conv_attention.cl`, `flash_attention.cl` and the `attention_kernels.cpp` host dispatchers, including the in-place RoPE over a device-cached table and the OHWI KV mirrors — ship in #4312, which is what this comment is closing against. The sliding-window mask and the explicit causal flag on the OHWI `sv_matmul` kernels are there too.
3. **The standalone kernel fixes** in this PR are not in either: the `rmsnorm.cl` / `rmsnorm_fp16.cl` width-tail correction (the float4 reduction silently drops the tail when the width is not a multiple of four, which is the live cause of the `RMSNormGPU` golden failure in `unittest_layers`) has a different consumer from the attention kernels and rides with the PR that carries those files. Replacement bundle to follow; the branch is preserved at `upstream-pr/trackG4-attention-kernels` so nothing is lost in the meantime.

The attention half was deliberately held back until its successor existed rather than closed against #4308 alone, since that would have left the larger half of this PR pointing at a PR that does not contain it.

PR #4113, opened 2026-07-21
[opencl] Parallelize the v8c weight upload (chunk crew + submit-and-go)

jijoongmoon:
Model-load performance for the v8c FC weight build. The per-weight nibble
repack + upload in make_v8c_weight_backing_from_qs4cx was fully serial with a
blocking write per chunk; on a multi-GB weight set the load phase paid one
submit->wait round trip per 16MB chunk on one core.

- Chunk crew: for large weights (>= 64MB, in practice the untied lm_head)
  a small worker crew (<= 4 threads) packs and uploads independent chunks
  concurrently. Staging stays bounded at workers x 16MB; blocking writes to
  disjoint offsets are safe on the shared queue and the per-channel row-sum
  rows are disjoint across chunks.
- Submit-and-go for the serial path: non-blocking writes hand their staging
  buffer to a pending-upload registry, freed once the write's event
  completes; correctness needs no barrier (the in-order queue sequences all
  later GEMMs after the writes), only the staging lifetime. Bounded at
  256MB: pushing past the cap drains everything queued so far.
  NNTR_V8C_UPLOAD_ASYNC=0 restores the blocking writes (A/B lever).
- v8c_flush_pending_uploads(): memory-hygiene reclaim of any still-pending
  staging, called at the first FC dispatch (relaxed-atomic no-op afterward).

The weight-cache build already runs OUTSIDE the cache lock (the parent
change), so concurrent weight-load workers now repack AND upload in
parallel; with the build inside a serial upload, a multi-second model init
was measured with ~97% of its time in this section.

suite 43/43 green; enable-opencl suite identical to the parent branch
(43 Ok / 2 fail / 2 timeout, unittest_layers failing set unchanged). The
upload paths are exercised by model-level int4-weight loads (not by this
unit suite). The OpenCL unittest compile-fix change was applied for
validation only.

Repro:


jijoongmoon:

Detail. This branch is a stack whose lower rungs are already open as #4304, #4306, #4308, #4309 and #4310; #4311 carries only this branch's own top commit, the chunk crew and the submit-and-go registry. Every raw `cl*` call in it was qualified through the ICD loader — unqualified calls link today only because CLBlast leaks the real symbols, and #4302 removes that — and `clReleaseEvent`, which the registry needs, was added to the loader's export list.

The review's findings, checked against the tree rather than taken on their word:

- **(1) the async upload is unordered whenever the queue is out-of-order.** The premise does not hold in this series: `CommandQueueManager::CreateCommandQueue()` passes properties `0` unconditionally after #4308, so there is no `NNTR_GPU_SVM_POOL` branch that can restore an out-of-order queue and no configuration in which the writes are unordered against the GEMM. The reasoning is right and would be a real defect the moment an out-of-order queue returns; #4311's commit message now states the in-order queue as the invariant the design rests on, so a future change that breaks it has something to break.
- **(2) the staging bound in the description is not the real bound.** Correct — the builder is itself called concurrently by the loader's weight workers, so the crew was per call and the peak was a multiple of what the message claimed. #4311 draws the worker count from a global budget, degrading to a serial pack when it is spent, and the commit message states the per-call structure instead of claiming a global bound.
- **(3) the crew's blocking writes serialize on the shared queue.** Correct. The commit message now says outright that what this parallelizes is the CPU permute, not the upload, and that the permute is the pole.
- **(4) no test coverage.** Still true and stated as an open item on #4311. What has been added is a device A/B on an Xe iGPU showing the quantized GEMM's output is byte-identical across the paths this touches; a host-side byte-for-byte comparison of the two upload modes remains unwritten.
- **(5) scope.** Answered by the re-cut itself: the infrastructure is #4302/#4304/#4306/#4308/#4309/#4310, and #4311's own diff is 12 files.
- **(6) mixed atomic/mutex discipline** and **(7) the unexplained constants** were read; the constants now carry their reason in the comment. **(8)** the `=false` / `=off` truthiness footgun is real but the idiom is tree-wide, so fixing it in one site would make the surface less consistent, not more — worth one shared helper in a separate change. **(9)** the pitch comment was corrected in #4310.

The branch is preserved at `upstream-pr/trackG6-load-perf`, so this is reversible.

PR #4114, opened 2026-07-21
[opencl] Planner-decided static tensor residency (cl_mem plane)

jijoongmoon:
The residency layer over the OpenCL stack: each tensor's memory plane
(HOST / SVM / device cl_mem) is decided STATICALLY by the memory planner at
allocation, and every producer/consumer binds by that class — replacing
per-edge runtime flips and the host map/unmap round-trips they forced.

Substrate:
- ResidencyClass (HOST/SVM/GPU_CLMEM + reserved IMAGE2D/RPCMEM) on
  MemoryData, with a static residency_ field distinct from the runtime
  device_valid overlay. MemAllocator::supportsResidency() capability map.
  Tensor::isClMem()/getClMem() so layers ask the tensor how to bind, plus an
  opt-in non-owning tv::TensorBacking pointer (propagated across shared-data
  views — a view consumer must see its source's backing).
- ClContext::residencyEngine() = GPU; the cl_mem sub-pool defaults on for
  Intel/Adreno (NNTR_GPU_CLMEM_POOL, overwrite=0 env default).
- Queue-manager map/unmap gain a force parameter and an NNTR_SVM_RESIDENT
  mode that skips the defensive per-op host maps on an all-GPU chain
  (genuine host boundaries pass force=true).
- ClSVMAllocator: the weight pool keeps the plain SVM MemoryPool (no weight
  ever binds its per-offset cl_mem; the quantized FCs read their own packed
  backing) — NNTR_CLMEM_WEIGHT_PLANE overrides both ways.

Planner (residency_planner.h / residency_policy.h):
- The classification pass is a declarative, testable unit: a pure function
  of (engine, all-consumers-GPU, dtype, role, name, fan-out) plus per-pool
  config. The app-specific boundary names (input RAISE / output LOWER /
  engine-neutral layer types) move OUT of core into an app-populated
  ResidencyPolicy — core no longer hardcodes application layer names.
- TensorPool threads the requesting layer's engine + TensorRole through
  request/view/requestOrExtend (defaulted params — non-GPU callers are
  unchanged), accumulates all_consumers_gpu across view consumers, classifies
  at allocate(), stamps MemoryData, and zero-fills exactly the per-offset
  cl_mem buffers a GPU_CLMEM tensor binds (ClBufferPool::ensureZeroFilled).
- NetworkGraph stamps each consumer's engine onto its producers' outputs
  and resolves producer->consumer resident edges; MultiOut (an in-place
  identity fan-out) and app-declared engine-neutral types do not veto the
  shared source's GPU residency.

Consumers (blas_kernel_interface.cpp):
- clmem_raise_cl / clmem_lower_cl: the two genuine host<->device boundaries
  as explicit copies (host-written input raised to the device plane; the
  final output lowered for the host lm_head) — device kernels reading
  host-mapped coarse SVM intermittently see stale data on the measured
  drivers, so boundaries must not rely on the map protocol.
- clmem_residual_op_cl + ClComputeOps::residual_op/scopy family: the
  residual chain runs on-device (cl_mem copy/add kernels), and the former
  AdditionLayerCL fork is DELETED — the core AdditionLayer is registered on
  the gpu context and dispatches through the op table.
- dotCl_v8c gains the residency paths: cl_mem-resident inputs quantize
  IN PLACE (NNTR_FC_QUANT_DIRECT, skipping the staging copy), GPU_CLMEM
  outputs are stored DIRECTLY by the GEMM's M_valid store guard
  (NNTR_V8C_DIRECT_OUT) or written by the device copy kernel, and directly
  chained FC->FC edges rebind the consumer's act-quant source to the
  producer's actual store target. The GLU whole-ops bind cl_mem operands
  with a kernel-side row offset for decode live-row processing.
- Segment-A GPU RMSNorm producers (rmsnorm_resident_fp16/fp32) with
  TensorBacking outputs + the publish/readback bridge, env-gated off by
  default (NNTR_RESIDENT_ACT/NNTR_DEVRES overlays; the static-class path is
  the production route).

suite 43/43 green (the substrate + planner compile into every build; the
classification is HOST for non-GPU allocators, byte-identical). CL suite
topline unchanged with the cl_mem pool DEFAULT-ON (43 Ok / 2 fail /
2 timeout; unittest_layers failing set identical to the parent branch's two
pre-existing cases, and the AdditionGPU goldens pass through the new
neutral-layer route). Lever smokes: NNTR_GPU_CLMEM_POOL=0 (legacy SVM) and
NNTR_SVM_RESIDENT=1 both hold the same failing set. The OpenCL unittest
compile-fix change was applied for validation only.

Depends-on: the attention/GLU change ("[opencl] Add the GPU
attention/RoPE/KV kernels + GLU whole-ops, fix rmsnorm tails") — must merge
first.

Repro:


jijoongmoon:
Superseded by the re-cut series against current main: #4309. The work in this PR is included there,
rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review
the new series instead. The branch is preserved on the fork, so this is reversible.

**Row re-verified before closing, and there is a scope split worth stating.** The residency content
this PR is named for — `residency_planner.h`, `residency_policy.h`, the `ResidencyClass` on
`MemoryData`, the `TensorPool` classify pass, `manager.cpp`, and the `cl_mem` pool — is all in #4309.
The rest of this branch is its parent rung's: the GPU attention/RoPE/KV kernels and the GLU whole-ops
from `0f199a4fb` (#4112) sit under it, and those go to #4308 (the GLU whole-ops) and to the GPU
attention PR later in the series. Neither half is lost; they are simply reviewed where they belong.

**What changed in the re-cut, and why.** The classifier is the same decision, stated without the
levers: the six `NNTR_CLMEM_*` environment variables that steered it are gone, and what core cannot
derive — a raised input boundary, a lowered output boundary, an excluded tensor — is declared once by
the application through `ResidencyPolicy`. #4309 adds **zero** new `getenv`/`setenv` calls. The
hard-coded `"cache_"` name test in core became `ResidencyPolicy::exclude_patterns` for the same
reason: core should carry no model-specific tensor names.

The allocator side is also stronger than it was here. `MemAllocator` now answers
`isHostAddressable` / `isDeviceVisible` / `isSVM` / `supportsDevicePool` / `supportsResidency` and
builds its own pool through `makePool()`, so the choice of pool kind follows from what the allocator
produces rather than from a `getName() == "gpu-svm"` comparison.

`TensorRole` was deliberately **not** carried over. Measured on this branch, its only use is a
`switch` in which every case falls through to the same result — a surface with no caller, which is
what #4105's review asked us to avoid. It belongs with the first layer that actually tags a tensor
with a role.

Verified on an Intel Xe integrated GPU: `unittest_cl_residency` 9/9, covering device placement, the
host-consumer and host-producer and FP32 downgrades, the declared raise and exclude boundaries, the
host-allocator case, and two tensors sharing a planner offset sharing one `cl_mem` handle.

PR #4115, opened 2026-07-21
[layers] Promote the CausalLM LLM layers to core + extract the model registry

jijoongmoon:
The parameter-general LLM layers that every CausalLM model shares move from
Applications/CausalLM/layers/ into core (nntrainer/layers/llm/), registered
by AppContext itself — an application (or SDK embedder) no longer has to
duplicate per-layer registration to run a transformer, and the backend
contexts can route the same layer types to their accelerated dispatches.

Promoted (compiled straight into libnntrainer, the loss/ subdir pattern):
- swiglu / geglu: backend-neutral GLU layers dispatching through the op
  table (in1.getOps()->swiglu/geglu) — CpuComputeOps on cpu, ClComputeOps
  on gpu. The former app-side causallm::SwiGLULayer and the OpenCL
  SwiGLULayerCl fork are both DELETED; the gpu context registers the
  neutral class gated on its kernel registration
  (registerSwiGLU/GeGLUClKernels).
- lm_head (untied host lm_head FC), tie_word_embedding (tied lm_head; its
  GPU Q6_K/Q4_0 GEMV is #if ENABLE_OPENCL internally, so it compiles
  host-only too), logit_softcapping, scalar_multiply (+ the OpenCL-resident
  scalar_multiply_gpu variant, compiled and registered only with
  enable-opencl), qkv_layer (pure-host QKV projection; no in-tree model
  wires it yet — inert/additive), and the env-gated layer_prof.h per-layer
  profiler header.
- App side: the duplicate layer sources/plugins are removed from
  Applications/CausalLM (meson plugin blocks + dep references dropped), and
  the app registration sites now rely on the core registrations. Include
  sites resolve to the core headers (swiglu.h -> swiglu_layer.h).
- The SwiGLULayerCl unit tests move to the neutral SwiGLULayer + whole-op
  entry points; the OpenCL kernel suites' swiglu references switch to
  swiglu_cl_op.h.

Model registry extraction:
- causallm::registerAllModels() (models/model_registry.{h,cpp}) — the full
  model-factory registration list moves verbatim out of main.cpp
  (idempotent via call_once), so alternative entry points (tests, SDK
  wrappers) register the same model set without duplicating the list.
  main.cpp shrinks to a single call.

FIXES a pre-existing failure: the SwiGLU-CL layer golden in unittest_layers
(numeric mismatch on every ancestor branch) PASSES through the neutral
layer + whole-op route.

suite 43/43 green (including the CausalLM model unit tests — the promoted
layers drive the real models); enable-opencl suite topline unchanged
(43 Ok / 2 fail / 2 timeout) with unittest_layers' failing set down to ONE
(the enable-clblast=false stub throw); one transient
unittest_attention_kernels_cl failure appeared under full parallel suite
load and passed 3/3 re-runs single-target and a full-suite re-run
(GPU-contention flake, pre-existing class). The OpenCL unittest compile-fix
change was applied for validation only.

Depends-on: the residency change ("[opencl] Planner-decided static tensor
residency (cl_mem plane)") — must merge first.

Repro:


jijoongmoon:
This PR is being replaced by the re-cut series against current `main`, and its two halves go to two different PRs, so it is not being closed against just one of them.

**The LLM layer promotion is now open as #4314.** That PR moves `qkv_layer`, `lm_head`, `logit_softcapping`, `scalar_multiply` and `tie_word_embedding` into `nntrainer/layers/llm/` as genuine renames, folds the application SwiGLU fork into the neutral `SwiGLULayer`, registers the promoted classes on the OpenCL and CUDA contexts, and adds the `CudaContext::registerLayerFactory` override so a layer can be registered on any engine.

**The model registry half is not in #4314.** `Applications/CausalLM/models/model_registry.{cpp,h}` and its consumers belong to the CausalLM PR that follows it in the series.

Leaving this open until that second half has a link, rather than pointing it at a PR that contains only part of it. The branch is preserved at `upstream-pr/trackA1-model-registry`, so nothing is lost either way.

jijoongmoon:
Superseded by the re-cut series against current main: the layer promotion half is #4314 and the model registry half is #4316, which adds `Applications/CausalLM/models/model_registry.{cpp,h}` and routes the command-line app, the C API and the quantizer through it. Both halves of this PR now have a successor, so closing it here. Please review the new series instead.

PR #4116, opened 2026-07-21
[Application/CausalLM] transformer: skip QS4CX repack on non-CPU engines

jijoongmoon:
Transformer::repack_weight() runs the QS4CX weight repack (KAI rhs-pack),
which is only consumed by the ARM CPU (KAI) inference path -- the "gpu"
(OpenCL) and "cuda" engines consume the plain QS4CX blob directly. On a
non-CPU run the whole per-weight loop is redundant, single-threaded CPU work;
on a large model it can pin one core (on Adreno the GPU sits idle at 0% while
one CPU core climbs toward a thermal shutdown, with 96%+ of CPU time in
kai_run_rhs_pack). Skip it unless the compute engine is CPU.

Adds a small causallm_engine() helper to llm_util.hpp (reads NNTR_ENGINE:
explicit cpu/cuda, else "gpu" when OpenCL is built, else "cpu"). It is
[[maybe_unused]] since not every translation unit that includes the header
calls it (so an -Werror=unused-function build stays clean).

Validated on x86 (gcc-13 -Werror, enable-cuda + enable-opencl + enable-fp16):
the CausalLM translation units compile clean with the guard in place.

Depends-on: trackA1-model-registry

---

Scope note on the two commits of this PR
----------------------------------------

The first commit skips the QS4CX repack on the coarse build-flag/engine
signal; the second narrows it to a per-layer decision keyed on each layer's
`engine=` property. The per-layer form is the intended shape, but it is
INERT on this branch and on every branch between it and the engine-stamping
change: nothing stamps `engine=` on the layer graph yet, so every layer
reports the default engine and the per-layer gate resolves exactly like the
coarse one.

Measured at the current tips: `withKey("engine"` occurs 0 times under
Applications/ on trackA1-model-registry, trackA2-repack-guard and
trackA3-gpu-layer-variants, and 69 times on trackDEF-gemma4-gpu-prereq,
which is where the stamping lands. Until that change merges, this PR is a
no-op refactor of the gate rather than a behaviour change, and the thermal
/ throughput benefit the first commit describes is what is actually in
effect.

One consequence a reviewer should know about, inherited from the first
commit rather than introduced by the second: once stamping exists, a layer
stamped `engine=gpu` whose OpenCL FC path falls back to the host KAI route
at runtime (the documented Adreno image-cap fallback) will reach
`getPackedData()` with nothing packed, because there is no lazy pack. The
per-layer gate is strictly narrower than the coarse gate it replaces, so
this is not a new regression, but the repack-skip idea as a whole is
unproven against that fallback and cannot be settled without a run on that
device.


jijoongmoon:
Superseded by the re-cut series against current main: #4316. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

PR #4117, opened 2026-07-21
[cuda] Add the CUDA backend context, memory, and stream infrastructure

jijoongmoon:
The bring-up layer of the additive NVIDIA CUDA backend: a "cuda" Context
registered alongside (not replacing) the OpenCL "gpu" context, selected at
runtime via engine=cuda. Kernels and layer compute paths follow in stacked
changes; this lands the infrastructure they run on.

Infrastructure (nntrainer/cuda/):
- ContextManager: device selection + primary-context retain, device probe
  (name/arch/integrated/concurrentManagedAccess), Windows delay-load image
  preloading.
- StreamManager: the in-order compute stream + drain policy helpers.
- Module/Kernel: NVRTC runtime compilation with an on-disk PTX cache keyed
  on (source, options) hash, cuModule loading, and kernel-launch wrappers —
  the CUDA analogue of the OpenCL registerClKernel machinery, with the same
  in-memory module cache shape.
- Buffer/BufferManager: device buffer RAII + a lazily-initialized staging
  set.

Context (cuda_context.{h,cpp}):
- CudaContext with the DeviceCaps probe (backend="cuda", unified_memory —
  UVM is the default pool), registerCudaKernel (hot path: single key +
  lookup, no source copy; owning module cache), registerLayerFactory via the
  engine facade, and the layer registrations for the backend-neutral classes
  (fully_connected/addition/geglu/swiglu/logit_softcapping/scalar_multiply/
  qkv/lm_head/tie_word_embedding — all correct host-on-UVM until the CUDA
  compute-ops land).
- Deferred bring-up: on a dual-backend build this context would otherwise
  initialize on the FIRST Engine touch of ANY run, and cuInit wakes a
  runtime-PM-suspended dGPU over PCIe (measured: the D3cold wake alone costs
  seconds and taxed every non-CUDA run). The context registers and
  initializes only when NNTR_ENGINE=cuda (NNTR_CUDA_EAGER_CTX/EAGER_INIT
  restore the eager behavior for A/B).
- CudaMemAllocator: cudaMallocManaged (UVM) pool — host-addressable AND
  device-accessible, the SVM analogue — with a device-only mode, calloc
  semantics via device-side memset, and a host fallback. isSVM() reports
  FALSE by contract: OpenCL SVM fast paths must never bind a CUDA-engine
  tensor on a unified build (a silently-failing CL call leaves the layer
  output unwritten — field-diagnosed as whole-model garbage).

The decode-step seam:
- Context::runDecode(nn, from, to, input, label): one decode/prefill
  forward step, dispatched through the resolved Context. The base is the
  plain graph walk (incremental_forwarding) — CPU/OpenCL byte-identical; a
  backend with its own decode engine (the CUDA-graph capture/replay in a
  later change) overrides it. NeuralNetwork::getDecodeContext() resolves the
  context from the authoritative per-layer engine property (NEVER a
  hardcoded backend order: on a unified build a blind "prefer cuda" would
  capture an engine=gpu run's decode onto CUDA and the KV/hidden-state
  handoff would cross backends incoherently).

Build: -Denable-cuda (+cuda-path), independent of and non-exclusive with
enable-opencl; links cudart/nvrtc/cublas/cublasLt + the driver API (with the
lib64/stubs fallback), Windows /DELAYLOAD for the CUDA DLLs + nvcuda.dll (a
unified binary must load on machines with no NVIDIA driver; the cudart-only
auto-probe must be callable before nvcuda is touched).

The promoted LLM layers' CUDA fast-path gate (NNTR_LLM_CUDA_FAST_PATH,
introduced in the layer-promotion change) stays off here; the CUDA
elementwise/FC changes flip it back to ENABLE_CUDA when their kernels land.

- Non-CUDA configs byte-identical: non-OpenCL suite 43/43 green; OpenCL
  suite topline unchanged (43 Ok / 2 fail / 2 timeout, both pre-existing).
  full suite topline identical to the OpenCL-only build (the CUDA context
  stays dormant without NNTR_ENGINE=cuda, per the gate).
- Runtime bring-up smoke with NNTR_ENGINE=cuda + eager flags: device probe
  (sm_120, discrete, 7.6 GiB), DeviceCaps log, and all layer factories
  registered on the live device; the appcontext suite passes under it.
- The OpenCL unittest compile-fix change was applied for validation only.

Depends-on: the layer-promotion change ("[layers] Promote the CausalLM LLM
layers to core + extract the model registry") — must merge first.

Repro:


jijoongmoon:

PR #4118, opened 2026-07-21
[cuda] Add the CUDA element-wise decode kernels + enable the LLM-layer fast paths

jijoongmoon:
The first CUDA compute kernels of the additive NVIDIA backend: the small
per-op element-wise transforms that otherwise break the device-resident
decode chain (each would drop the fp16 activation to a host loop between the
surrounding GEMMs). NVRTC-compiled, fp16 I/O with FP32 math, one kernel per
op, dispatched through the CUDA stream and module cache from the backend
infrastructure change.

nntrainer/cuda/cuda_elementwise.{h,cpp} (three kernels + host wrappers):
- swiglu_fp16   out = silu(gate) * up          (qwen3/llama FFN gate)
- scalar_mul_fp16  out = in * scalar
- softcap_fp16  out = cap * tanh(in / cap)      (final logit softcapping)

These are exactly the kernels the three promoted neutral LLM layers call.
The layer-promotion change ("[layers] Promote the CausalLM LLM layers …")
staged their CUDA fast paths behind NNTR_LLM_CUDA_FAST_PATH because the
kernels did not exist yet; this change lands the kernels and flips those
three layers' gate to ENABLE_CUDA:
- SwiGLULayer          -> cuda_swiglu_fp16
- ScalarMultiplyLayer  -> cuda_scalar_mul_fp16 (opt-in NNTR_CUDA_ELTWISE)
- LogitSoftCappingLayer -> cuda_softcap_fp16

Each fast path is additionally guarded on ENABLE_FP16 and on the input being
device-accessible; it falls through to the existing host path otherwise
(including on the host-coherent UVM pool, where the CpuComputeOps loop is
correct — the kernel is the acceleration, not the correctness path). The
TieWordEmbedding fast path keeps the NNTR_LLM_CUDA_FAST_PATH gate: it needs
the int4 FC kernel, which lands in a later change that flips it in turn.

- Non-CUDA configs byte-identical: the flipped gate evaluates false without
  ENABLE_CUDA, so the plain and OpenCL builds preprocess the layers exactly
  as before (plain suite 43/43 green; OpenCL suite topline unchanged).
- Bring-up smoke with NNTR_ENGINE=cuda: the CUDA context registers the three
  layers on the live device and the appcontext suite passes under it.

Depends-on: the CUDA backend infrastructure change ("[cuda] Add the CUDA
backend context, memory, and stream infrastructure").

Repro:


jijoongmoon:

PR #4119, opened 2026-07-21
[cuda] Add the QS4CX int4 FC dequant-GEMM + install the CUDA ComputeOps

jijoongmoon:
The int4 FullyConnected path for the CUDA backend: a QS4CX (per-channel int4)
fused dequant-GEMM, dispatched through a CudaComputeOps that the cuda context
installs. This is the correctness floor -- the plain-decode GEMM (one thread
per output element, float accumulation); the w4a8 dp4a / int8-Tensor-Core
fast paths and the weight-residency levers are later changes.

Why the CUDA backend needs its own FC GEMM: the host Tensor::dot for a QS4CX
weight is NYI on x86 (the packed-int4 host kernel is KAI/ARM-only). So on the
cuda context the QS4CX FC MUST take a device path; without it the FC throws.

- nntrainer/cuda/cuda_fc_qint4.{h,cpp}: the QS4CX dequant-GEMM (NVRTC).
  Reads the PLAIN QS4CX payload (row-major [N][(K+1)/2] nibbles, uint4 =
  int4+8) directly -- no host/UVM copy of the weight -- dequantizing inline and
  scaling by an N-entry per-channel fp16 scale built once per weight
  (cuda_fc_qs4cx_scales_to_uvm_fp16). fp32-activation
  (cuda_fc_qs4cx_gemm_fp32) and fp16-activation (cuda_fc_qs4cx_gemm_fp16_naive,
  staged through fp32) variants.
- nntrainer/cuda/cuda_compute_ops.cpp: CudaComputeOps, inheriting CpuComputeOps
  (engine=cuda tensors are host-coherent UVM, so every op the backend does not
  accelerate is correct via the CPU body). Overrides only fc(): QS4CX -> the
  device dequant-GEMM when the operands are device-accessible, else the host
  dot fallback. get_cuda_ops() returns the singleton, mirroring get_cl_ops().
- nntrainer/cuda_context.cpp: install it with
  getContextData()->setComputeOps(get_cuda_ops()) (the analogue of the OpenCL
  context's get_cl_ops install). FullyConnectedLayerCl (registered on the cuda
  context in the infrastructure change) dispatches its GEMM through this.
- compute_ops.h: the get_cuda_ops() declaration, guarded by ENABLE_CUDA next
  to get_cl_ops().

- Non-CUDA configs untouched: every new source is behind -Denable-cuda and the
  get_cuda_ops() declaration is ENABLE_CUDA-guarded, so the plain and OpenCL
  builds are unchanged (plain suite 43/43 green; OpenCL suite topline
  unchanged).
- Bring-up smoke with NNTR_ENGINE=cuda: get_cuda_ops installs on the live
  device (the FC now resolves to the CUDA path) and the appcontext suite
  passes under it.

The int4-FC numerics are exercised by a full model decode on the device (no
unit fixture, since the QS4CX host reference is ARM-only); the kernel bodies
are carried unchanged from the validated CUDA backend.

Depends-on: the CUDA RMSNorm change ("[cuda] Add the CUDA RMSNorm layer +
kernel …").

Repro:


jijoongmoon:

PR #4120, opened 2026-07-21
[cuda] QS4CX int4 FC fast paths: w4a8 dp4a + cuBLAS int8-IMMA prefill

jijoongmoon:
Two accelerated int4 FC paths on top of the naive QS4CX dequant-GEMM floor,
dispatched in CudaComputeOps::fc() (cuBLAS-i8 first at M>=32, then dp4a, then
the naive floor -- each falls to the next). The int32 accumulate is exact, so
all three are bit-identical.

- w4a8 dp4a (cuda_fc_qint4): FC_QINT4_DP4A_SRC + a lazy pointer-keyed derived
  weight cache (ensure_dp4a_cache_locked / dp4a_repack_and_gemm) built on first
  FC. int8-activation x int4-weight via __dp4a on the int ALU; the real decode
  path.
- cuBLAS int8-IMMA prefill (cuda_blas_manager): cublas_i8_gemm_fp16 + the
  int8-epilogue kernels + a persistent per-weight int8 cache; the same
  int8xint4->int32 GEMM on the IMMA Tensor Cores (~10x the dp4a GEMM throughput
  at prefill M). BlasManager wraps the cuBLAS handle + tuning.

Validated on an RTX 5060 (gcc-13 -Werror): enable-cuda+opencl unified build
clean; plain suite 43/43 (gate-off byte-identical); a device-probe compiles the
NVRTC sources and launches the full dp4a + cuBLAS-i8 chains with correct output;
a 1K-prefill run produces byte-identical generated text across dp4a and cuBLAS.

Depends-on: trackC4-cuda-int4-fc


jijoongmoon:

PR #4121, opened 2026-07-21
[cuda] Add device-only-memory copy ops to CudaComputeOps (DEV_ACT pool)

jijoongmoon:
Completes the device-only activation pool. The CudaMemAllocator already
supports a device_only mode (real cudaMalloc, not host-addressable) selected
by Manager::activationAllocator under NNTR_CUDA_DEV_ACT -- that pool keeps the
CPU off the activation pages (no page-fault thrash) while weights stay UVM.
But a Tensor::copy() on such an activation would hit the inherited
CpuComputeOps host loop and fault on the non-host-addressable pointer.

This adds the copy-op overrides so a copy involving a device-only endpoint is
correct:
- scopy_fp32 / scopy_fp16: route a contiguous same-type device-only copy
  through a stream-ordered cudaMemcpyAsync (ordered against the producing
  kernels on the backend stream); a copy the host reads next (D2H) drains
  first. Host / host-coherent UVM keep the CPU path unchanged.
- scopy_fp32_to_fp16 / scopy_fp16_to_fp32: converting copies with a
  device-only endpoint stage through a host temporary (copy_any D2H/H2D).
- A strided copy on device-only memory is unsupported (does not occur in the
  forward path) and fails loudly rather than faulting.

With the UVM default (NNTR_CUDA_DEV_ACT off) dev_only() is false for every
pointer, so device_copy() returns false immediately and all ops fall through
to the unchanged CpuComputeOps behavior -- no change to the default path.

- Non-CUDA configs untouched: the change is in a cuda-gated source (plain
  suite 43/43; OpenCL suite topline unchanged).
- Bring-up smoke passes with NNTR_ENGINE=cuda both at the UVM default and with
  NNTR_CUDA_DEV_ACT=1 (device-only activation pool active). The copy-op bodies
  are carried unchanged from the device-validated CUDA backend.

Depends-on: the cuBLAS int8-IMMA FC change ("[cuda] Add the cuBLAS int8-IMMA
int4 FC prefill path").

Repro:


jijoongmoon:

PR #4122, opened 2026-07-21
[cuda] Prefetch QS4CX weights to the device during load (wprefetch)

jijoongmoon:
A load-time memory lever for the CUDA backend: migrate each QS4CX weight's
managed plain payload (+ its fp32 scale tail) to the device with
cudaMemPrefetchAsync as it is read, so the FC weight bytes leave host RSS and
the GEMM reads them from VRAM instead of faulting managed pages in on first
use.

- cuda_fc_qint4.{h,cpp}: cuda_fc_qs4cx_prefetch_weight(). Discrete GPU only
  (returns false on integrated, where managed pages don't migrate and the
  device pool gives no bandwidth benefit); a no-op on a non-managed pointer.
- models/neuralnet.cpp: in the parallel weight-load worker, after a node's
  weights are read, prefetch its QS4CX weights. Gated by NNTR_CUDA_WPREFETCH>=2
  AND at least one layer on the cuda engine -- doing it inline (as-read) keeps
  the FC bytes from accumulating in host RSS during load, which is the load-time
  RSS peak. The engine gate keeps the CUDA call off OpenCL/CPU runs of a
  dual-enabled binary (a stray CUDA call would otherwise create the CUDA
  context).

Default off (NNTR_CUDA_WPREFETCH unset -> 0): the load path is unchanged, and
the whole block is #if ENABLE_CUDA so non-CUDA builds never see it. This is
the prefetch mechanism; the derived-cache prewarm and the host-page decommit
(drop_plain) that build on it are app-orchestrated and land separately.

- Non-CUDA configs untouched: the additions are #if ENABLE_CUDA guarded, so
  the plain and OpenCL builds are unchanged (plain suite 43/43; OpenCL suite
  topline unchanged).
- Device probe: cuda_fc_qs4cx_prefetch_weight on a cudaMallocManaged buffer
  issues the prefetch on the discrete RTX and returns success (the pointer
  stays managed/device-accessible). The function body is carried unchanged
  from the device-validated CUDA backend.

Depends-on: the CUDA device-only copy-ops change ("[cuda] Add
device-only-memory copy ops to CudaComputeOps (DEV_ACT pool)").

Repro:


jijoongmoon:
Row re-verified against #4313 before closing, and it does **not** match, so this stays open.

The map routed this PR to the CUDA-backend rung. Measured on #4313's own range: its `CudaComputeOps` overrides are the copy family, `geglu`, `layer_norm`, `activation` and `fc` — there is **no `fc_prebuild_weight` override**, which is the seam this PR's tip commit reseats the weight-prefetch hook onto. So this PR's own content (the two `wprefetch` commits) is not in #4313; only the device-only copy ops it inherits from #4121 are, and those were closed against #4313 on their own row.

Leaving this open until the rung that carries the load-time weight prefetch is cut. Branch preserved at `upstream-pr/trackC12-cuda-weight-prefetch` (or whichever head it is on the fork).

jijoongmoon:

As noted above, the change here is NOT carried by that series: the CUDA backend's ComputeOps has no `fc_prebuild_weight` override, which is the seam this PR's weight-prefetch hook depends on.

PR #4123, opened 2026-07-21
[cuda] QS4CX FC: optional transient JIT int8 weight unpack (cuBLAS-i8 prefill)

jijoongmoon:
Add an opt-in (NNTR_CUDA_I8_JIT) transient JIT variant of the int4->int8
weight unpack that feeds the cuBLAS INT8-IMMA prefill FC. Instead of the
persistent per-weight int8 cache, the resident dp4a packed-int4 weight is
transpose-unpacked to int8 on the GPU per prefill into a reusable scratch
(nothing stays resident; the per-channel rowsum is shared with the dp4a
cache -- same values). This trades a small per-call unpack cost for the int8
cache's VRAM: a TTFT-vs-VRAM knob, default off. cuda_fc_qs4cx_cublas_i8_gemm_fp16
keeps the persistent ensure_i8_cache path unless NNTR_CUDA_I8_JIT=1.

The unpack ships two NVRTC kernels: a byte-granular i8_jit_unpack and a
vectorized i8_jit_unpack_v4 (K%8==0 && N%4==0, which covers every FC in the models measured here)
that runs the ~1.8GB/prefill unpack traffic at near-memcpy bandwidth via
4-byte coalesced loads/stores through a shared tile.

The int32 IMMA accumulate is exact, so the JIT path is bit-identical to both
the persistent-cache cuBLAS path and dp4a; only the weight-unpack staging
differs.

- enable-cuda + enable-opencl + enable-fp16 build compiles clean; the branch
  is fully preprocessed out of non-CUDA builds (default off, no behavior
  change when NNTR_CUDA_I8_JIT is unset).
- 1K-prefill run with NNTR_CUDA_I8_JIT=1 (forcing the JIT unpack of every FC
  weight) produces byte-identical generated text to the persistent-cache
  cuBLAS-i8 path -- the NVRTC unpack kernels compile and launch correctly and
  reproduce the same int8 weights.

Depends-on: trackC5b-cuda-wprefetch


jijoongmoon:

PR #4124, opened 2026-07-21
[cuda] Q6_K GPU lm_head + enable the tie_word_embedding CUDA fast path

jijoongmoon:
Add lmhead_gemv_q6_k_cuda: a Q6_K lm_head GEMV that reads the FP16 hidden and
the (managed) Q6_K weight on the device and writes FP16 logits to the device
output -- no host bounce, so the Q6_K lm_head (gemma2/qwen3) runs under a
device-only activation pool (NNTR_CUDA_DEV_ACT) where a host GEMV would fault.
Returns false (caller falls back to host) when hidden % 256 != 0 or dispatch
fails.

With the kernel in place, flip tie_word_embedding.cpp's CUDA fast path from the
placeholder NNTR_LLM_CUDA_FAST_PATH gate (defined nowhere -> dead) to
ENABLE_CUDA. This activates the layer's device path: the Q6_K lm_head GEMV above
plus the device-only embedding staging (the latter is gated on the per-tensor
emb_dev_only flag, so it is inert on the default managed/UVM pool and only
engages under NNTR_CUDA_DEV_ACT). NNTR_LLM_CUDA_FAST_PATH was a temporary gate
used while the CUDA kernels landed across this series; this is its last user, so
the macro is now gone from the tree.

- enable-cuda + enable-opencl + enable-fp16 build compiles clean with the
  tie_word CUDA path now live; non-CUDA builds are unchanged (the whole block is
  ENABLE_CUDA-guarded, so the plain/OpenCL preprocessing is unaffected).
- The Q6_K GPU lm_head produces correct output on the Q6_K-lm_head models
  (qwen3, gemma2) at 1K input; the emb_dev_only staging stays inert on the
  default managed pool, matching the host-path output.

Depends-on: trackC4d-cuda-i8-jit


jijoongmoon:

PR #4125, opened 2026-07-21
[opencl][cuda] NNTR_DETERMINISTIC: reproducible execution, on by default

jijoongmoon:
Make nntrainer's GPU execution reproducible by default, with an opt-out for
callers that want the last bit of throughput. The default (env unset) now runs
the deterministic contract; `NNTR_DETERMINISTIC=0` restores the prior
non-deterministic behavior.

The contract:
- OpenCL kernel ring-rotation (cl_context.cpp): hand out K=8 rotating clones of
  each kernel instead of a process-global singleton. This fixes a *measured
  token-altering* hazard -- a dispatcher re-binding args on the singleton can
  hit an object whose previous enqueue the driver has not locked in yet. The
  clones are cheap (program cache) with zero per-call overhead, so this is a
  free correctness fix, hence default-on.
- CUDA per-op drains kept (cuda_context.cpp pins NNTR_CUDA_ASYNC=0 unless the
  caller set it) so decode is reproducible.
- cuBLAS FP32 pedantic math mode pinned (cuda_blas_manager.cpp).
- OpenCL attention reduction: the fixed LDS tree is available behind the
  explicit `NNTR_DETERMINISTIC=1` opt-in; the sub-group reduction stays the
  default (attention_kernels.cpp) -- see the measured reason below.

Finer opt-in levers for bisecting a nondeterminism source stay default-off:
NNTR_CL_KERNEL_RING=K, NNTR_CL_LOCKSTEP (clFinish per enqueue, brutal cost),
NNTR_CL_SYNC_IO, NNTR_POISON_FILL, NNTR_CUBLAS_PEDANTIC.

## Past-window prefill regression, found and recovered — the tip commit

This claim belongs to one commit and is stated separately so it is not read as a
property of the contract as a whole: **`[opencl] Keep the sub-group flash
reduction default; LDS tree behind strict opt-in`**.

The first cut of the contract flipped the Block-Q flash kernel's d-dot reduction
from `sub_group_reduce_add` to the fixed LDS tree unless `NNTR_DETERMINISTIC=0`.
Block-Q is the kernel every **past-window** prefill call lands on -- the DPAS/XMX
route requires `win == 0` -- so the flip taxed exactly the `M > W` regime while
`M <= W` prefill calls kept their speed. The flip's own validation ran CUDA and a
mobile GPU, neither of which reaches this Intel-only (`cl_intel_subgroups`)
buffer path, which is how it could measure "no material cost" on lanes where the
new default never engages.

Serialized per-kernel timing (`NNTR_FLASH_TIMER`) attributes the whole gap to the
sliding bucket alone: **39.5 ms vs 13.6 ms** over the identical 16 windowed calls
of a 1K prefill on a W=1024 sliding-window model, with the full-attention buckets
equal (64.5 vs 69.4 ms) and identical routing on both sides.

Cells on Intel Xe3 (fp16 activations, QS4CX weights; best-of-2, one discarded
warm-up, 45 s cooldowns), LDS-tree default -> this commit, against the
pre-contract reference bar:

| model (window) | cell | before | after | bar |
|---|---|---|---|---|
| W=1024 sliding | 1K M=1073 XMX prefill | 2495.4 (**-12.1 %** vs bar) | **2869.0** | 2838.6 |
| W=1024 sliding | 29K XMX prefill (ring on) | 872.4 (**-19.6 %** vs bar) | **1177.8** | 1084.5 |
| W=1024 sliding | 29K dp4a prefill (ring on) | 606.7 | **755.6** | 741.5 |
| gemma4 (W=512) | 1K M=999 XMX prefill | 3092.9 | **4094.3** | 3948.6* |

(* single non-protocol spot reference.) Decode moved toward the bar everywhere
(19.21 -> 20.73 vs bar 21.19 on the W=1024 model's 1K cell); no decode cell
regressed. The 29K runs kept the KV window ring engaged (logged every run).

Determinism is not weakened by keeping the sub-group default: the same default
lane already runs `sub_group_reduce_add` / `_max` inside the XMX flash kernel for
every full-attention prefill call, two consecutive runs of that lane are
byte-identical (fixed dispatch on the in-order queue), token output is unchanged
for four models on short and long goldens, and `NNTR_FLASH_SG=0/1` still forces
either path explicitly.

## Validation

Validated on an RTX 5060 and a Galaxy S26 Ultra (Adreno):
- No measurable cost: gemma4 CUDA 1K prefill 5581 (off) vs 5775 (on) TPS;
  a 35-layer sliding-window QS4CX-FP16 model on Adreno, 1K prefill 1258 vs 1244 TPS -- both within run-to-run noise.
- Output unchanged: the generated tokens are byte-identical with the contract
  on vs off (the only textual diff is opt-in [CAP-AUDIT] debug logging).
- Reproducible: two `NNTR_DETERMINISTIC` runs are byte-identical.
- The determinism files compile clean (enable-cuda+opencl+fp16); every block is
  env-gated so `NNTR_DETERMINISTIC=0` reproduces the exact pre-change path.

**Merge button: "Rebase and merge", never squash.** The past-window recovery
above is attributed to a single commit; squashing collapses it into the contract
commit and destroys that revert handle.

Depends-on: #4124


jijoongmoon:
Superseded by the re-cut series against current main: #4312. Closing to keep the queue reviewable — please review the new series instead.

The OpenCL half of the determinism contract is in #4312, in two pieces:

- **The kernel ring-rotation**, in `ClContext::registerClKernel`. It is carried as its own commit and, as this PR's own follow-up argued, it is **unconditional** there: the singleton's argument re-bind reaching an object whose previous enqueue the driver has not locked in is a correctness bug, not a reproducibility preference, so it does not sit under the `NNTR_DETERMINISTIC` opt-out. `NNTR_CL_KERNEL_RING=K` remains the diagnostic override and is now parsed with `strtol` and range-checked, so a typo cannot silently select the legacy singleton.
- **The attention reduction shape.** `sub_group_reduce_add` is the one vendor-opaque reduction on the flash path, so `NNTR_DETERMINISTIC=1` selects the fixed LDS tree instead. It is an explicit opt-in rather than the default, for the reason this PR's last commit records: the tree costs the measured sub-group win on the buffer path and buys no reproducibility the default lane does not already have, since the same lane runs sub-group reductions inside the matrix-engine kernel anyway and a fixed dispatch on an in-order queue is already run-to-run reproducible.

The CUDA half (the per-op drain policy in `cuda_context.cpp` and the math-mode pinning in the cuBLAS manager) is not in this PR's successor, because the CUDA backend has not been cut yet. Replacement bundle to follow; the branch is preserved at `upstream-pr/trackD1-determinism`.

For the record, the mapping table in the series plan routed this PR to the execution-plan PR (#4306), which was wrong — that PR ships no determinism content at all. It was left open there rather than closed against a PR that does not contain it.

PR #4126, opened 2026-07-21
[cuda] M2-B CUDA-graph decode: capture+replay the per-token forward (opt-in)

jijoongmoon:
A per-token decode issues ~1000 tiny GPU kernels; the CPU launch/dispatch
latency between them is the decode bottleneck (GPU only ~30-47% utilized).
CudaContext::runDecode captures the per-token forward into a CUDA graph once
and replays it, collapsing that launch overhead. This is a DECODE-speed path
only (it does not touch prefill correctness or throughput).

Gated by NNTR_CUDA_M2B (default OFF); with the graph flags unset runDecode is a
plain eager walk == the base Context::runDecode, so engine=cuda without the
flags stays byte-identical.

Part A ships:
  - the CudaContext::runDecode override (the CUDA-graph capture/replay state
    machine: M1 decode graph / M2-B single-capture coherent replay / prefill
    graph), relocated verbatim from the old neuralnet.cpp ENABLE_CUDA block and
    appended at the vtable tail;
  - the neuralnet M2-B accessors the runDecode seam drives through
    (setM2BSkipAll, isPrefillCaptureDisabled/setPrefillCaptureDisabled,
    getLayerNode, feedInputsLabels) plus their state.

Full M2-B decode CORRECTNESS additionally needs the app-side decode kernels
(per-layer-slice / argmax / scalar-mul slot writes) that refresh the embedding
staging and read back the sampled token between replays. Those kernels have no
core consumer here (they live in the app-side mha_core / per_layer_slice /
causal_lm), so they are intentionally not part of this core change: the
runDecode override ships default-off and is inert until NNTR_CUDA_M2B is set.

Depends-on: trackC5e-cuda-q6k-lmhead-tieword


jijoongmoon:

PR #4127, opened 2026-07-21
[cuda] Auto-tune CUDA env defaults by GPU class (discrete vs integrated)

jijoongmoon:
The CudaContext ctor now fills a tuned NNTR_CUDA_* default set once at bring-up,
so a bare `NNTR_ENGINE=cuda` run gets full-GPU residency without the caller
exporting ~15 flags. Every setenv uses overwrite=0, so any explicit user env
ALWAYS wins (including =0 to disable a default).

The common set (both HW classes) moves rope/attn/geglu/eltwise/qknorm/FC onto
the GPU (NNTR_CUDA_ROPE/ATTN/KV_UVM/GEGLU/ELTWISE/QKNORM, FLASH_DECODE=64,
BLOCKQ, FC_CUDA_CUBLAS, PREWARM). Discrete GPUs (!integrated &&
concurrentManagedAccess) additionally get the device-resident/decode-graph
profile: DEV_ACT, VCOPY_PREFILL, RMSNORM_CUDA_OFF=all (raises the CUDA RMSNorm
row cap), M2B, and ASYNC (kept off under NNTR_DETERMINISTIC). Integrated
(Tegra/Orin) and WDDM (cMA==0) keep the safe base profile — managed pools +
per-op drains — because those add-ons let a host op touch managed/device memory
around in-flight kernels, which is only legal under cMA=1.

This ships last in the CUDA stack because several of these defaults turn ON
features that gate app-side (rope/attn) or only just landed (M2B, in the
preceding track): they are inert until their consumer/app-side kernels exist,
which is expected and harmless (an unread setenv costs nothing).

Depends-on: trackC5d-cuda-graph-m2b


jijoongmoon:

PR #4129, opened 2026-07-21
[build] fp16-impl option: uint16 Half wrapper fallback for compilers …

jijoongmoon:
MSVC cl.exe has no native `_Float16`/`__fp16` scalar, so an x86_64 fp16 build cannot use the native half type there. Add a uint16-backed `nntrainer::Half` (IEEE754 binary16) that computes every op in float and rounds back, and a meson `fp16-impl` combo option (auto/native/wrapper) to select the backing type:

  - auto (default): the meson probe compiles a tiny `_Float16` arithmetic snippet with the actual compiler; native `_Float16` is used when it succeeds (GCC/clang), otherwise the build falls back to the wrapper and emits -DUSE_HALF_WRAPPER=1. Selection is capability-based, not compiler-name/version based.
  - native: force `_Float16`.
  - wrapper: force the Half wrapper even on GCC/clang, so the MSVC path is exercisable as a Linux test harness.

`_FP16` resolves to `::nntrainer::Half` under USE_HALF_WRAPPER (in tensor_dim.h and nntr_ggml_impl.h) and to `_Float16` otherwise. The wrapper is a self-contained leaf header (std lib + immintrin.h only, no nntrainer include) and is layout-identical to native `_Float16` (2-byte standard layout, static_assert-checked) so model files, checkpoints and GPU kernel buffers are interchangeable between native and wrapper builds. The x86 AVX2 fp16 SIMD interface (avx2_impl.h / avx2_impl_fp16.cpp) is retyped from `_Float16*` to `_FP16*` so it accepts the Half type; the kernels already reinterpret through `__m128i*`, so this is a no-op on the native path (`_FP16` == `_Float16` there) and only makes the same files compilable on MSVC, where `_Float16` does not exist.

auto/native remain unchanged on GCC/clang (native `_Float16`, byte identical). Validated on Linux x86_64 (GCC 13.3):
  - `meson -Dfp16-impl=wrapper ... && ninja` compiles + links the full library and apps with -DUSE_HALF_WRAPPER=1 (was the whole point: build the Half wrapper where the native path is bypassed).
  - `meson -Dfp16-impl=native` builds clean and unchanged; both emit -DENABLE_FP16=1, only the wrapper build adds -DUSE_HALF_WRAPPER=1.
  - a same-binary bit-identity check (wrapper Half vs native _Float16): all 65536 f16->f32 patterns, an f32->f16 value grid (subnormals / overflow / signs) and +,-,*,/ arithmetic agree bit-for-bit (65747 checks, 0 mismatch).

### DIFF
--- a/api/ccapi/include/half_fp16.h
+++ b/api/ccapi/include/half_fp16.h
@@ -1,0 +1,311 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2026 Jijoong Moon <jijoong.moon@samsung.com>
+ */
+/**
+ * @file   half_fp16.h
+ * @date   04 Jul 2026
+ * @brief  uint16-backed IEEE754 binary16 `Half` wrapper for toolchains that
+ *         lack a native `_Float16` / `__fp16` scalar type (notably MSVC
+ * cl.exe).
+ * @see    https://github.com/nntrainer/nntrainer
+ * @see    docs/design/msvc_fp16_half_wrapper.md
+ * @author Jijoong Moon <jijoong.moon@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ * @note   This is a SELF-CONTAINED LEAF header. It must include only the C++
+ *         standard library (<cstdint>/<cstring>/<cmath>) plus <immintrin.h>
+ *         under MSVC. It must NOT include any nntrainer header (in particular
+ *         tensor_dim.h), so that tensor_dim.h can pull it in without an include
+ *         cycle.
+ *
+ * @note   Selected only when the build defines USE_HALF_WRAPPER (emitted by the
+ *         meson capability probe when native _Float16 arithmetic is
+ *         unavailable, e.g. MSVC). On GCC/clang/ARM the `_FP16` macro keeps
+ *         resolving to the native `_Float16`/`__fp16` and this header is not
+ *         used — those builds stay byte-identical.
+ *
+ * @note   Layout is IEEE754 binary16 (`uint16_t bits_`), bit-for-bit identical
+ *         to native `_Float16`, so model files, checkpoints and GPU kernel
+ *         buffers are interchangeable between a native-fp16 build and a
+ *         wrapper build. Every arithmetic operator computes in `float` and
+ *         rounds the result back to half, matching how GCC/clang lower scalar
+ *         `_Float16` arithmetic (promote to float, operate, round back).
+ */
+
+#ifndef __HALF_FP16_H__
+#define __HALF_FP16_H__
+#ifdef __cplusplus
+
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#include <type_traits>
+
+#if defined(_MSC_VER)
+#include <immintrin.h>
+#endif
+
+namespace nntrainer {
+
+/**
+ * @brief Internal bit-level fp16<->fp32 conversion helpers.
+ * @note  The software path is a byte-for-byte copy of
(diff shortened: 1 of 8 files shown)


jijoongmoon:
Superseded by the re-cut series against current main: #4307. The work in this PR is included there,
rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please
review the new series instead. The branch is preserved at `upstream-pr/...` on the fork.

The 08-26/08-28 review on this PR was harvested before the re-cut was pushed, and three of its
findings were real. They are fixed in #4307, not carried forward:

| finding | disposition in #4307 |
| --- | --- |
| `half_fp16.h` is appended to `ccapi_headers` but listed in neither `packaging/nntrainer.spec` nor any `debian/*.install` | **Fixed.** Added to both. Verified against the meson install plan on one host with identical flags: plain `main` = 115 installed headers with 6 spec-unclaimed / 15 debian-unclaimed; the new branch = 116 with the **same** 6 / 15, i.e. zero new unpackaged headers. Without this the Tizen GBS `check-files` stage and `dh_missing --fail-missing` would both have failed, and no meson/Android/Windows job can see it because none of them runs `meson install`. |
| the compound-assignment operators only take `Half`, so `half op= float` narrows the operand first and rounds twice | **Fixed, and it was reachable.** Live call sites: `nntrainer/tensor/cpu_backend/fallback/fallback_internal_fp16.cpp` `hsgemm_loop`, `C[m * ldc + n] += (beta) * c_old` with `C` an `_FP16 *` and both operands `float`. Native `_Float16` widens, computes in float and rounds once; narrowing first also turns an operand above the half range into `inf` *before* the addition rather than after. SFINAE mixed overloads added, computing in `std::common_type<float, T>`. Pinned by a test: 65504 − 70000 is −4496 through the mixed overload and +inf through the homogeneous one. |
| `_mm_cvtps_ph` / `_mm_cvtph_ps` are reached on bare `_MSC_VER` | **Fixed.** MSVC's x64 baseline is SSE2 and it has no per-function target attribute, so that is an illegal instruction on pre-Ivy-Bridge hardware, not a slow path. Narrowed to `defined(_MSC_VER) && defined(__AVX2__)`. GCC/clang keep the software converter unconditionally, so no non-MSVC behaviour changes. |
| no test covers the wrapper | **Fixed.** `test/unittest/unittest_half_fp16.cpp`, 8 cases, built wherever a native half exists to compare against — which is every GCC/clang build including `-Dfp16-impl=wrapper`. All 65536 half patterns widen bit-identically; a >65 000-entry probe grid spanning subnormals, overflow and the rounding edges narrows bit-identically; single add/sub/mul/div and all six comparisons agree over >500 000 operand pairs; plus mixed compound assignment, the layout/storage-interchange contract, and `++`/`--`. Multi-operation expressions are deliberately not compared — excess precision makes them differ by design, and that caveat is stated in the PR description. |
| NaN payloads may differ between the software and intrinsic paths | Accepted as-is and handled in the test, which treats all NaNs as equal rather than asserting payload equality. |

Verification in #4307: three Linux configurations (`fp16=false`, `fp16=true` → native `_Float16`,
`fp16=true -Dfp16-impl=wrapper`) build with tests on, and the native and wrapper arms run the same
suites to identical counts — `unittest_half_fp16` 8/8, `unittest_nntrainer_tensor` 268,
`_tensor_fp16` 130, `_tensor_pool_fp16` 96, `cpu_backend` 53, `fallback` 78, `activations` 20, zero
failures on either side. No MSVC build was performed and none is claimed; the added CI job is the
mechanism that builds this configuration on `cl.exe`.

PR #4130, opened 2026-07-21
[utils] ThreadManager::Global(): out-of-line singleton (per-DLL insta…

jijoongmoon:
Move the ThreadManager::Global() accessor out of line: declare it in the header and define the function-local static in thread_manager.cpp instead of inheriting the inline Singleton<T>::Global() template body.

On MSVC, an inline/header Global() is instantiated once per DLL, so a shared-linked build ends up with a separate worker pool per module. Defining Global() out of line pins it to a single translation unit and yields one process-wide instance. This mirrors the same fix already shipped for the OpenCL context/buffer/command-queue managers and compute_ops getComputeOps.

Behavior is identical on Linux/Android (still one function-local static instance); this only changes where the singleton is emitted.

Depends-on: (none -- independent, off b3a5face2)

### DIFF
--- a/nntrainer/utils/thread_manager.cpp
+++ b/nntrainer/utils/thread_manager.cpp
@@ -40,6 +40,13 @@ ThreadManagerConfig ThreadManager::config_ = {};
 
 ThreadManager::ThreadManager() {}
 
+ThreadManager &ThreadManager::Global() {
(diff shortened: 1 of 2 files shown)


jaemini-shin (approved):
LGTM!

Jungwon-Lee (approved):
LGTM!

PR #4131, opened 2026-07-21
[build/ci][windows] GPU machine file + MSVC OpenCL+CUDA+fp16 verify CI

jijoongmoon:
Windows GPU build support, two pieces:

- configurations/windows-native-gpu.ini: a meson native-file for a Windows GPU
  build (OpenCL + CUDA + fp16). Pure option toggles -- enable-opencl/cuda/fp16,
  enable-blas=false (no shipped Windows profile routes GEMM through OpenBLAS,
  and its per-instance pools cost several GB of private commit under static
  linking), enable-test=false, platform=windows, vsenv. It deliberately does
  NOT pin default_library: SHARED is the intended default flavor (single
  nntrainer.dll, faster prefill, far smaller private commit; validated both
  flavors), and meson's own default is already 'shared', so a machine-file pin
  would silently override a command-line -Ddefault_library on every ninja regen.

- .github/workflows/msvc_gpu_verify.yml: a windows-2022 MSVC x64 compile+link
  job -- install CUDA Toolkit + vcpkg OpenCL, meson setup --native-file the ini
  above, ninja -k 0. Compile/link verification of the OpenCL + CUDA +
  fp16-Half-wrapper build (GH runners have no GPU, so no run). The base already
  ships windows.yml / windows_cross_compile.yml / windows_arm_clang.yml; this
  adds the GPU-config MSVC job.

No product-model content. Validated on a Windows MSVC box: the ini
configures the GPU build; both shared and static flavors golden.

Depends-on: trackC5e-cuda-q6k-lmhead-tieword, trackW1-msvc-fp16-half


jijoongmoon:
Superseded, **in part**, by the re-cut series against current main: #4307. Closing to keep the queue
reviewable — please review the new series instead. The branch is preserved on the fork.

**What #4307 carries:** the Windows MSVC verification job, re-written fp16-only —
`.github/workflows/windows_msvc_fp16.yml` plus `configurations/windows-native-fp16.ini`. It
configures `-Denable-fp16=true` and lets the `fp16-impl=auto` probe select the `Half` wrapper, so the
code path that exists for `cl.exe` is compiled and linked on `cl.exe` on every PR.

**What it deliberately does not carry, and why:** the GPU variants. The workflow as written here
passes `enable-cuda`, `cuda-path` and `enable-clblast` — **none of which is an option on `main`** — so
it would die at `meson setup` on its first run rather than fail a build. Those variants belong with
the backend PRs that create the options; the machine file for them is not shipped either, for the
same reason.

Two further defects in this PR were structurally resolved by the rewrite rather than patched:

- the trigger was `push: branches: [ 'ci/msvc-*' ]`, so the job could never run on a pull request,
  which is the only place it is useful. The new one is `pull_request: branches: [ main ]` with a
  `concurrency` group.
- the job is compile + link only, with `ninja -k 0` so one run surfaces every remaining error. The
  unit-test suite still has test-only MSVC portability gaps, and the fp32 job in `windows.yml` keeps
  running `meson test`. That is stated in the workflow header rather than left implicit.

Every option the new workflow and its machine file pass was checked mechanically against
`meson_options.txt` on the current base, which is exactly the check this PR's version failed.

PR #4132, opened 2026-07-21
fix(ci): stop misfiring doxygen-tag.sh checks

jayden0701:
## Summary
- Checks @brief in only file head, not for function/struct
- Allows inline comments like `func(/*do_sample=*/false)`

## Detail
`doxygen-tag.sh` (the "/Checker/ Doxygen tag check" step in
`static.check.yml`) had four independent bugs that made it produce
false failures, silently skip real ones, or crash with a shell error —
none related to actual documentation quality:

- **Dead per-function/struct check, printing a bash error on every
  file.** `local function_check_flag=...` was used outside a function,
  which bash rejects (`local: can only be used within a function`).
  Because that assignment silently never took effect, the ctags-based
  "this function/struct needs `@brief`" check it fed has been a
  complete no-op since this script was imported from TAOS-CI in 2024.
  Wiring it up properly instead of removing it would newly require a
  `@brief` above every function/struct in every file this repo has ever
  touched — measured against `main`, that's ~7,000 functions and ~166
  structs across 582 files. Dropped the dead branch instead.
- **False positives on ordinary inline comments.** The "comment should
  begin with `/**`" rule flagged any line containing `/*` that didn't
  *end* with `*/`, which misfires on common inline comments like
  `/*thiz*/` or `/*do_sample=*/false` that close on the same line.
  Anchored the check to lines that themselves open a comment and don't
  close it on that same line.
- **File-header tag check wasn't actually scoped to the header.** The
  "`@file`/`@brief`/`@author`/`@bug` tag is required at the top of
  file" rule grepped the *whole file*, so any of those words appearing
  in a per-function comment anywhere below silently satisfied the
  file-header requirement. Restricted the search to the file's actual
  leading comment region.
- **A single bad Python file could hide every later violation.** The
  `.py` branch did `break` on failure instead of `continue`, which
  exits the outer `for file` loop entirely — every file listed *after*
  a tag-less `.py` file (regardless of language) was silently never
  checked. Changed to `continue`.

Also restored `report_path=${report_path:-/tmp/doxygen-tag-report.log}`,
which this branch's copy of the script had dropped, reintroducing an
"ambiguous redirect" error on every debug log line (previously fixed
once already).

Scoping the file-header check correctly surfaced 21 existing files on
`main` whose header relies on `@note`/`@details` instead of an actual
`@brief` line (`Applications/CausalLM/models/**`, `Applications/LLaMA/jni/main.cpp`,
`nntrainer/layers/centroid_knn.h`) — they only "passed" before because
some unrelated `@brief` elsewhere in the file satisfied the old,
unscoped whole-file grep. Added a proper `@brief` line to each so a
future PR touching them doesn't hit a new failure from the fixed check.

## Test plan
- [x] `bash -n` on the script
- [x] Reproduced each bug with a minimal test case and confirmed the
      fix resolves it (mis-ordered `.py`/`.cpp` file lists, a header-less
      file with only per-function tags, inline-comment false positives)
- [x] Re-ran the check against a broader sample of real repo files
      (including the 21 backfilled ones) to confirm no new false
      positives from the header-scoping change

### DIFF
--- a/.github/workflows/static.check.scripts/doxygen-tag.sh
+++ b/.github/workflows/static.check.scripts/doxygen-tag.sh
@@ -28,6 +28,7 @@ fi
 
 files=$1
 failed=0
+report_path=${report_path:-/tmp/doxygen-tag-report.log}
 
 if [ ! -f $files ]; then
   echo "::error The file $files does not exists."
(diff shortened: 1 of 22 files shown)


jaemini-shin (approved):
LGTM!

dlwlzzero (approved):
LGTM 👍

PR #4133, opened 2026-07-21
[kleidiai] Add `matmul_clamp_f16_qai8dxp_qsi4cxp`

jaemini-shin:

<details><summary>[kleidiai] Add matmul_clamp_f16_qai8dxp_qsi4cxp</summary><br />

- Add ukernels for matmul_clamp_f16_qai8dxp_qsi4cxp
- Add pack kernel for matmul_clamp_f16_qai8dxp_qsi4cxp
- Add unittest for gemm / gemv

</details>

### Summary
This PR updates the following kleidiai kernel
- lhs: FP16 -> channel wise int 8
- rhs: channel wise int 4
- out: FP16

Note that this kernel is effectively available for android only.
- other platforms doesn't use aarch64 + FP16 as of now

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/kai_interface/kai/pack/meson.build
+++ b/nntrainer/tensor/cpu_backend/arm/kai_interface/kai/pack/meson.build
@@ -20,6 +20,16 @@ pack_headers = [
     'kai_rhs_pack_nxk_qsi4cxp_qs4cxs1s0_neon.h',
 ]
 
+if get_option('enable-fp16')
+pack_sources += [
+  'kai_lhs_quant_pack_qai8dxp_f16_neon.c',
+]
+
+pack_headers += [
+  'kai_lhs_quant_pack_qai8dxp_f16_neon.h',
+]
+endif
+
 foreach s : pack_sources
   nntrainer_sources += meson.current_source_dir() / s
 endforeach
--- a/nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_f16_qai8dxp_qsi4cxp.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_f16_qai8dxp_qsi4cxp.cpp
@@ -1,0 +1,330 @@
+/**
+ * Copyright (C) 2024 Arm Limited and/or its affiliates
+ *
+ * @file   kleidiai_interface_f16_qai8dxp_qsi4cxp.cpp
+ * @date   21 July 2026
+ * @see    https://github.com/ARM-software/kleidiai
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author Jaemin Shin <jaemin980311@gmail.com>
+ *
+ * @brief  Modified computational backend components of
+ * kleidiai. Portions of this file are derived from Arm
+ * Limited code licensed under the Apache License, Version 2.0, with
+ * modifications
+ *
+ * @note   Licensed under the Apache License, Version 2.0 (the "License");
+ *         you may not use this file except in compliance with the License.
+ *         You may obtain a copy of the License at
+ *             http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * @modifications
+ *   - [2026-07-21] Integrated and adapted Arm-provided code into
+ *     nntrainer CPU backend for FP16 support
+ *
+ * @bug    No known bugs except for NYI items
+ */
+//
+// SPDX-FileCopyrightText: Copyright 2025 Arm Limited and/or its affiliates
+// <open-source-office@arm.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <kleidiai_interface.h>
+#include <thread_manager.h>
+
+#include "kai/matmul_clamp_f16_qai8dxp_qsi4cxp/kai_matmul_clamp_f16_qai8dxp_qsi4cxp_interface.h"
+
+// Include micro-kernel variants
+#include "kai/matmul_clamp_f16_qai8dxp_qsi4cxp/kai_matmul_clamp_f16_qai8dxp1x4_qsi4cxp4x4_1x4_neon_dotprod.h"
+#include "kai/matmul_clamp_f16_qai8dxp_qsi4cxp/kai_matmul_clamp_f16_qai8dxp1x8_qsi4cxp4x8_1x4_neon_dotprod.h"
+#include "kai/matmul_clamp_f16_qai8dxp_qsi4cxp/kai_matmul_clamp_f16_qai8dxp4x4_qsi4cxp4x4_16x4_neon_dotprod.h"
+
+#if defined(__ARM_FEATURE_MATMUL_INT8)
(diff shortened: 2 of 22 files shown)


test/unittest/unittest_nntrainer_kleidiai.cpp
```diff
@@ -209,6 +209,105 @@ INSTANTIATE_TEST_SUITE_P(
...
+    for (size_t i = 0; i < M * N; i++) {
+      dst[i] = static_cast<float>(dst_f16[i]);
+    }
+
+    // Compute cosine similarity
+    double cosine_sim = cosine_similarity(ref_dst.data(), dst.data(), M * N);
+
+    // Expect high cosine similarity (directional alignment)
+    EXPECT_GE(cosine_sim, cosine_threshold)
+      << "Low cosine similarity at F16 variant " << idx_variant;
```
jaemini-shin, line 294: Pass when cosine similarity between fp32 matmul and f16 kernel is greater than 0.99

Seunghui98 (commented):
## Description

Adds the `matmul_clamp_f16_qai8dxp_qsi4cxp` path (FP16 LHS → per-channel int8, int4 RHS, FP16 out): 4 vendored ukernels + asm, a new `kai_lhs_quant_pack_qai8dxp_f16_neon` pack kernel, an `nntrainer` interface layer, meson wiring behind `enable-fp16`, and two cosine-similarity unit tests.

## Summary

Structure mirrors the existing f32 `qai8dxp_qsi4cxp` path closely, which makes it easy to follow. Build gating looks right: `kai_interface` is only entered for aarch64/android, and `jni/meson.build` supplies `-march=...+fp16+dotprod+i8mm`, so the new pack kernel's `FEAT_FP16` requirement is satisfied on the target platform. One improvement over the f32 version: `_rhs_unpacked` no longer allocates an unused LHS buffer.

## Bug

**`kai/pack/meson.build` — `pack_headers` is assigned instead of appended.**

```meson
pack_sources += [ 'kai_lhs_quant_pack_qai8dxp_f16_neon.c' ]

pack_headers = [                                   # <-- should be +=
  'kai_lhs_quant_pack_qai8dxp_f16_neon.h',
]
```

`pack_sources` uses `+=` but `pack_headers` uses `=`, so with `enable-fp16=true` the 7 existing pack headers are wiped and only the f16 one reaches `nntrainer_headers`. Result: the f32/qsi8d32p pack headers stop being installed in the fp16 build. Almost certainly a typo.

1. **Scheduler duplication vs. #4065.** `__kai_gemm_f16_qai8dxp_qsi4cxp` copies the pre-#4065 `m_step > n_step` heuristic verbatim, so f16 won't inherit the tiling work landing in that PR — and it carries the same weakness: both f16 GEMM variants are `m_step=16 > n_step=4`, so `M=1` takes the parallelize-over-m branch with `m_loop == 1` and runs **single-threaded**. Worth either sharing one scheduler between the f32/f16 interfaces or rebasing on #4065 once it merges.

2. **Test shapes are all step-aligned.** `(1, 512, 3072)` and `(768, 768, 768)` are divisible by every `m_step`/`n_step` in play (4, 16), so the ragged tail — where `get_*_offset` math actually goes wrong — is never exercised. Adding one shape like `(17, 130, 100)` would cover it. Also untested: the `is_nxk=false` (kxn) pack path and non-default `lower_bound`/`upper_bound`.

3. **`enable-fp16` now implies `+fp16` in `-march`.** The new pack kernel hard-`#error`s without `__ARM_FEATURE_FP16_{SCALAR,VECTOR}_ARITHMETIC`. The default `arm-march` covers this, but a config that sets `enable-fp16=true` with a march lacking `+fp16` turns into a hard build failure rather than a soft fallback — worth confirming no existing build config does that.

4. **Minor:** `mr`/`nr`/`kr`/`sr` in `__kai_gemm_f16_qai8dxp_qsi4cxp_rhs_unpacked` and `nr` in `__kai_gemm_f16_qai8dxp_qsi4cxp` are now unused (`-Wunused-variable`).

The meson `+=` fix is the only blocking item for me; the rest are follow-ups.

---

nntrainer/tensor/cpu_backend/arm/kai_interface/kai/pack/meson.build
```diff
@@ -20,6 +20,16 @@ pack_headers = [
...
 ]
 
+if get_option('enable-fp16')
+pack_sources += [
+  'kai_lhs_quant_pack_qai8dxp_f16_neon.c',
+]
+
+pack_headers = [
+  'kai_lhs_quant_pack_qai8dxp_f16_neon.h',
+]
```
Seunghui98:
```suggestion
pack_headers += [
  'kai_lhs_quant_pack_qai8dxp_f16_neon.h',
]
```
maybe i guess this is a bug. Could you check it?

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_f16_qai8dxp_qsi4cxp.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+void __kai_gemm_f16_qai8dxp_qsi4cxp_rhs_unpacked(
+  size_t m, size_t n, size_t k, void *lhs_native_mtx_f16,
+  void *rhs_native_mtx_qs4cx, void *rhs_scales_f32, void *dst_act_mtx_f16,
+  size_t idx_variant, bool is_nxk, float lower_bound, float upper_bound) {
+  const size_t mr = ukernel_variants[idx_variant].ukernel.get_mr();
+  const size_t nr = ukernel_variants[idx_variant].ukernel.get_nr();
+  const size_t kr = ukernel_variants[idx_variant].ukernel.get_kr();
+  const size_t sr = ukernel_variants[idx_variant].ukernel.get_sr();
```
Seunghui98: I guess they are unused. so, How about removing them?

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_f16_qai8dxp_qsi4cxp.cpp
```diff
@@ -0,0 +1,336 @@
...
+  delete[] rhs_packed_mtx_qs4cx;
+}
+
+void __kai_gemm_f16_qai8dxp_qsi4cxp(size_t m, size_t n, size_t k,
+                                    void *lhs_native_mtx_f16,
+                                    void *rhs_packed_mtx_qs4cx,
+                                    void *dst_act_mtx_f16, size_t idx_variant,
+                                    float lower_bound, float upper_bound) {
+  const size_t mr = ukernel_variants[idx_variant].ukernel.get_mr();
+  const size_t nr = ukernel_variants[idx_variant].ukernel.get_nr();
```
Seunghui98: I guess this is unused.
How about removing it?

Seunghui98 (approved):
LGTM Thanks 👍

jayden0701 (approved):
LGTM!

haehun (approved):
LGTM

PR #4134, opened 2026-07-21
[ThreadManager] Prevent nested parallel deadlock

Jungwon-Lee:
## Dependency of the PR

None

<details><summary>[ThreadManager] Prevent nested parallel deadlock</summary><br />

[ThreadManager] Prevent nested parallel deadlock

Track callback execution per thread and run directly nested parallel_for calls sequentially. This avoids re-entering the shared execution mutex while preserving serialization for independent callers.

The fallback applies only when parallel_for is invoked directly from a callback on the same thread; execution context is not propagated to child threads.

Add regression coverage that verifies nested callbacks execute sequentially on the calling thread and every iteration executes exactly once.

**Self evaluation:**

- clang-format 14.0.6 check passed.
- C++ syntax check passed.
- An isolated Linux harness verified direct nesting and independent callers.
- The full Meson test target was skipped because the available container lost its Docker backend connection.

</details>

### Summary

- Prevent direct nested parallel_for calls from re-entering the shared execution mutex.
- Execute directly nested work sequentially while preserving serialization for independent callers.
- Document the cross-thread nesting limitation and add regression coverage.

### DIFF
--- a/nntrainer/utils/thread_manager.cpp
+++ b/nntrainer/utils/thread_manager.cpp
@@ -32,11 +32,27 @@ inline size_t modulo_decrement(size_t i, size_t n) {
   return i - 1;
 }
 
+class ParallelRegionGuard {
+public:
+  explicit ParallelRegionGuard(bool &in_parallel_region) :
+    in_parallel_region_(in_parallel_region),
+    previous_value_(in_parallel_region) {
+    in_parallel_region_ = true;
+  }
+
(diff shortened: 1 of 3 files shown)


dlwlzzero (approved):
LGTM overall. The doxygen tag check is failing in CI, take a look and fix it!

jaemini-shin (changes requested):
Since this thread pool is not designed to use nested loop for performance, it'd be better to prohibit rather than allowing it.

PR #4136, opened 2026-07-22
[kleidiai] Add dequantization for QS4CX

jaemini-shin:

<details><summary>[kleidiai] Add dequantization for QS4CX</summary><br />

- Add dequant_qs4cx_f32 for arm, x86, fallback
- Add nxk, kxn fallback implementation
- Add unittest

</details>

### Summary
This PR adds fallback for QS4CX dequantization.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -639,9 +639,15 @@ void dequant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_qs4cx,
 
 void dequantize_row_qs4cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs4cx,
                           void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+#ifdef __ARM_NEON
+  nntrainer::neon::dequantize_row_qs4cx_neon(
+    n_idx, k, (const uint8_t *)rhs_native_mtx_qs4cx,
+    (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+#else
   __fallback_dequantize_row_qs4cx_f32(
     n_idx, k, (const uint8_t *)rhs_native_mtx_qs4cx,
     (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+#endif
 }
 
 size_t get_rhs_packed_size_qsi4cxp_qs4cxs1s0(size_t n, size_t k,
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -623,6 +623,27 @@ void quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
   }
 }
 
+void dequant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_qs4cx,
+                       void *rhs_scales_f32, void *rhs_native_mtx_f32,
+                       bool is_nxk) {
+  if (is_nxk) {
+    __fallback_dequant_nxk_qs4cx_f32(
+      n, k, (const uint8_t *)rhs_native_mtx_qs4cx,
+      (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+  } else {
+    __fallback_dequant_kxn_qs4cx_f32(
+      n, k, (const uint8_t *)rhs_native_mtx_qs4cx,
+      (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+  }
+}
+
+void dequantize_row_qs4cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs4cx,
+                          void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+  __fallback_dequantize_row_qs4cx_f32(
+    n_idx, k, (const uint8_t *)rhs_native_mtx_qs4cx,
+    (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+}
+
 size_t get_rhs_packed_size_qsi4cxp_qs4cxs1s0(size_t n, size_t k,
                                              size_t idx_variant, bool is_nxk) {
 #ifndef ARMV7
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
@@ -1483,6 +1483,46 @@ void quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
                      bool is_nxk);
 
 /**
+ * @brief qs4cx dequantization of rhs matrix. Inverse of quant_qs4cx_f32.
+ * Converts quantized int4 back to float32 using per-channel scales.
+ * qs4cx refers to quantized symmetric 4-bit channel-wise quantization.
+ *
+ * @warning You should allocate memory for output before use:
+ *  - rhs_native_mtx_f32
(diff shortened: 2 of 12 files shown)


nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -1482,6 +1482,46 @@ void quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
...
+ *    n * k * sizeof(float)
+ *
+ * @param[in] n N for (M, K) * (K, N) = (M, N) in noTrans GEMM
+ * @param[in] k K for (M, K) * (K, N) = (M, N) in noTrans GEMM
+ * @param[in] rhs_native_mtx_qs4cx quantized matrix data
+ * @param[in] rhs_scales_f32 per-channel quantization scales
+ * @param[out] rhs_native_mtx_f32 dequantized matrix data in float32
+ * @param[in] is_nxk true if the quantized matrix is stored in nxk format
+ */
+void dequant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_qs4cx,
```
jaemini-shin, line 1501: This is a dequant for tensor.

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
```diff
@@ -1482,6 +1482,46 @@ void quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
...
+ *  - rhs_native_mtx_f32
+ *    k * sizeof(float)
+ *
+ * @param[in] n_idx channel (row) index to dequantize
+ * @param[in] k K length of the row (channel)
+ * @param[in] rhs_native_mtx_qs4cx full nxk quantized matrix data
+ * @param[in] rhs_scales_f32 per-channel quantization scales
+ * @param[out] rhs_native_mtx_f32 dequantized row data in float32 (k floats)
+ */
+void dequantize_row_qs4cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs4cx,
```
jaemini-shin, line 1522: This is a dequant for row

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -623,6 +623,33 @@ void quant_qs4cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
...
+    __fallback_dequant_kxn_qs4cx_f32(
+      n, k, (const uint8_t *)rhs_native_mtx_qs4cx,
+      (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+  }
+}
+
+void dequantize_row_qs4cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs4cx,
+                          void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+#ifdef __ARM_NEON
+  nntrainer::neon::dequantize_row_qs4cx_neon(
```
jaemini-shin, line 643: As of now, only arm has optimized kernel.

Seunghui98 (commented):
## Description

Adds QS4CX dequantization: `__fallback_dequant_{nxk,kxn}_qs4cx_f32` and `__fallback_dequantize_row_qs4cx_f32` in the fallback layer, a NEON row kernel `dequantize_row_qs4cx_neon`, public `dequant_qs4cx_f32` / `dequantize_row_qs4cx` entry points on the arm/x86/fallback backends, and unit tests for both the fallback and NEON paths.

## Summary

- Round-trip semantics check out against the existing quantizers. Both `__fallback_quant_{nxk,kxn}_qs4cx_f32` store `recip_scale0 = 1/scale0` in `rhs_scales_f32[n_idx]` and both take their f32 input in **nxk** layout, so `code * scale` and the nxk-shaped output of `__fallback_dequant_kxn_qs4cx_f32` are the correct inverses. The `@note output matrix will be the (n,k) shape` on the kxn variant is accurate, not a slip.
- The NEON kernel is correct as far as I can tell. `vzipq_u8(low_nibbles, high_nibbles)` yields `val[0] = e0..e15`, `val[1] = e16..e31` (zip1/zip2 on A64, VZIP on A32 — same result), matching the comment; `vsubq_s8(..., 8)` is safe since nibbles are `0..15`; and the widening chain preserves element order. Bounds are fine at every step: the 32-element loop needs `i/2 + 16 <= k/2`, the 16-element block needs `i/2 + 8 <= k/2`, and both follow from their guards, with the `< 16` remainder left to scalar code.
- Test shapes cover the paths that matter: `k=255` exercises the 32-loop, the 16-block, and a 15-element odd tail in one go, and `k=7`/`k=255` cover the unused high nibble of the last byte.

## Bug

None found. The items below are API consistency and coverage.

1. **`dequant_qs4cx_f32` on ARM never uses NEON.** `arm_compute_backend.cpp` dispatches straight to the fallback for both layouts, even though the `is_nxk` case is exactly `n` iterations of the NEON row kernel you just added. Right now the per-row API is accelerated and the whole-matrix API — the more likely hot path — is not. A loop over `dequantize_row_qs4cx_neon` would close that with a few lines.

2. **`dequantize_row_qs4cx` is nxk-only but doesn't say so in its signature.** Unlike `dequant_qs4cx_f32`, it has no `is_nxk` parameter and no guard, so a kxn buffer passed in silently produces garbage. Either add the flag or rename to `dequantize_row_nxk_qs4cx` so the constraint is visible at the call site.

3. **Stride is spelled two different ways.** The NEON kernel uses `(k + 1) / 2`, the fallback uses `roundup(k, 2) / 2`. They're equal for all `k`, but since the layout contract lives in that one expression, having a single form (or a shared helper) makes future changes safer.

4. **`dequantize_row_qs4cx_perf_vs_fallback` is a benchmark inside the unit test target.** It asserts byte-identity, which is good, but the timing part has no threshold and just prints, so it adds runtime to every CI run without being able to fail on a regression. Worth either moving it to the benchmark target or adding an actual bound.

5. **Minor:** `dequant_qs4cx_f32` / `dequantize_row_qs4cx` are byte-identical in `fallback.cpp` and `x86_compute_backend.cpp`, and near-identical in `arm_compute_backend.cpp`. That matches the existing pattern in this directory, so no objection — just noting it in case item 1 makes the arm copy actually diverge.

---

test/unittest/unittest_nntrainer_kleidiai.cpp
```diff
@@ -488,6 +461,124 @@ INSTANTIATE_TEST_SUITE_P(
...
+  const double iters = static_cast<double>(iterations);
+  std::cout << "[INFO] dequantize_row_qs4cx n=" << n << " k=" << k
+            << " (avg of " << iterations << " iters)" << std::endl;
+  std::cout << "[INFO] fallback: " << fallback_ns / iters / 1000.0 << " us, "
+            << std::endl;
+  std::cout << "[INFO] neon    : " << neon_ns / iters / 1000.0 << " us, "
+            << std::endl;
+  std::cout << "[INFO] speedup : "
+            << static_cast<double>(fallback_ns) / static_cast<double>(neon_ns)
+            << "x" << std::endl;
```
Seunghui98, line 607: it's just out of curiousity. 🤔
I think it is unnatural to print the test time to the unit test, is this necessary?

jaemini-shin: This PR was uploaded before #4060. I'll make this test disabled with `--bench`!

Seunghui98 (approved):
LGTM 👍

jayden0701 (approved):
LGTM!

jijoongmoon (approved):
Just for the confirmation, is this for the dynamic quantization & dequantization for KAI? 
LGTM

jaemini-shin:
> Just for the confirmation, is this for the dynamic quantization & dequantization for KAI? LGTM

@jijoongmoon It could be used for Q-DQ, but what I intended was for row-wise DQ for embedding layer.

PR #4141, opened 2026-07-24
[tensor] NNTR_MEM_PLANNER selector: choose basic|v1|v2|v3 for the activation pool (default v1, unchanged)

jijoongmoon:
Manager::finalizeTensorPool hardcodes OptimizedV1Planner even though OptimizedV2Planner/OptimizedV3Planner already ship in-tree, and carries a `//@todo change V3 and validate` right at the call site. This adds an NNTR_MEM_PLANNER env selector (basic|v1|v2|v3) so planners can be A/B-ed without a rebuild. Unset or unknown values keep the exact previous behavior (v1; BasicPlanner still used whenever optimizations are off).

Answering the @todo with data: on a CausalLM inference workload (uniform transformer activations, identical-size tensors recurring per layer), V3's byte-granular gap-fill splits vacant intervals and then extends the pool top — classic best-fit fragmentation — measurably growing the host pool versus V1's same-size slot reuse, while outputs stay byte-identical across planners and no layout throws occur. V1's same-size reuse is effectively optimal for this shape of workload, so it stays the default; the selector makes that conclusion reproducible on any model.

Repro:
  meson setup build -Denable-transformer=true -Denable-fp16=true --buildtype=plain
  ninja -C build && meson test -C build unittest_memory_pool unittest_nntrainer_tensor_pool
  NNTR_MEM_PLANNER=v3 <any CausalLM run>   # output byte-identical, pool size differs

### DIFF
--- a/nntrainer/tensor/manager.cpp
+++ b/nntrainer/tensor/manager.cpp
@@ -908,16 +908,28 @@ void Manager::flushCacheExcept(unsigned int order) {
 
 void Manager::finalizeTensorPool(TensorPool &pool, unsigned int start,
                                  unsigned int end) {
-  if (enable_optimizations) {
-    if (exec_mode == ExecutionMode::INFERENCE && enable_fsu) {
-      //@todo change V3 and validate
-      pool.finalize(OptimizedV1Planner(), start, end);
-    } else {
-      pool.finalize(OptimizedV1Planner(), start, end);
-    }
-  } else {
+  // NNTR_MEM_PLANNER=basic|v1|v2|v3 selects the layout planner (default v1,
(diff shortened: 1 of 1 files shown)


jijoongmoon:
Closing: **no content in this scope.**

The re-cut series against current main (#4302 … #4320) does not carry this change, so there is no
successor PR to point at. `NNTR_MEM_PLANNER` has zero occurrences anywhere in the range the series
covers, which was verified before writing this — the selector cannot be cut from it and nothing in
a new PR body would substantiate it.

This is therefore not a supersession. The branch is preserved at
`upstream-pr/track0-mem-planner-selector` and the PR can be reopened as-is if the memory-planner
selector is wanted on its own footing.

PR #4075, opened 2026-07-14
[Build] Fix clang invalid-feature-combination error from -march=native on AVX10.1 runners

Jungwon-Lee:
## Summary
- `meson_test` CI (ubuntu-22.04/24.04, clang) has been failing on `main`'s scheduled builds for weeks, unrelated to any specific PR.
- GitHub Actions runners now expose AVX10.1-capable CPUs. `meson.build` unconditionally adds `-march=native -mavx2 -mfma`; clang detects the conflicting feature combination and (under `-Werror`) hard-fails with `invalid feature combination: +avx10.1-256; will be promoted to avx10.1-512`.
- Add `-Wno-error=invalid-feature-combination` to the clang warning flags, gated by meson's `has_argument` check (same pattern as the other `-Wno-error=*` flags nearby), so older clang without this diagnostic is unaffected.

## Test plan
- [ ] CI `meson_test (ubuntu-22.04)` and `meson_test (ubuntu-24.04)` pass on this PR
- [ ] `main` scheduled build passes after merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4142, opened 2026-07-24
[causallm] Report peak private commit alongside peak working set

jijoongmoon:
On Windows the peak working set alone under-describes a run's memory: WDDM folds GPU allocations into the process WS, and paged-out private pages leave it entirely. The commit charge (peak private commit) is the stable host-private footprint, so the CausalLM perf summary now reports both.

On POSIX the new getPeakCommitKb() returns 0 and the print is suppressed — Linux output is unchanged apart from the existing peak line gaining a "(working set)" label. No behavior change anywhere else; metrics only.

Repro:
  meson setup build -Denable-transformer=true --buildtype=plain
  ninja -C build && meson test -C build unittest_causallm_models

---
Note: this branch also carries the one-line `-Wno-error=invalid-feature-combination` clang guard from #4075 — the clang meson_test cells fail on any AVX10.1-capable runner (`-march=native` + explicit `-mavx2/-mfma` collision, a runner-pool regression also hitting main since 2026-07-14) and this PR happened to land on such a runner. Identical content to #4075's fix; whichever lands second rebases clean.

### DIFF
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -673,6 +673,7 @@ void CausalLM::run(const WSTR prompt, bool do_sample, const WSTR system_prompt,
   auto total_duration = std::chrono::duration_cast<std::chrono::milliseconds>(
     finish_total - start_total);
   size_t peak_memory = getPeakMemoryKb();
+  size_t peak_commit = getPeakCommitKb();
 
   if (log_output) {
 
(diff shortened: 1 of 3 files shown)


jijoongmoon:
Closing: **no content in this scope.**

The re-cut series against current main (#4302 … #4320) does not carry this change, so there is no
successor PR to point at. `peak_commit` has zero occurrences anywhere in the range the series
covers, which was verified before writing this — the metric cannot be cut from it and nothing in a
new PR body would substantiate it.

This is therefore not a supersession. The branch is preserved at
`upstream-pr/track0-causallm-peak-commit-metric` and the PR can be reopened as-is if the
peak-private-commit metric is wanted on its own footing.

PR #4143, opened 2026-07-24
[cuda] Add the CUDA RMSNorm layer (FP32-safe accumulation) + the cuda_layers meson element

jijoongmoon:
The host fp16 RMSNorm path accumulates the square-sum in fp16, which overflows on large-magnitude residuals (the row collapses to 0 and the model emits garbage). This adds CudaRMSNormLayer — FP32 accumulation on both paths — bound by the cuda context for the core "rms_norm" type, exactly as the gpu context binds RMSNormLayerCl. Decode-size fp16 rows run a device NVRTC kernel; wide prefill stays on the multithreaded host path (the kernel syncs per call, so rows<=32 gates the device route).

New meson element nntrainer/layers/cuda_layers/ (enable-cuda, symmetric to cl_layers) carries the layer; registration is promoted into cuda_context.add_default_object() rather than app-side per-engine factories.

Repro:
  meson setup build -Denable-cuda=true -Denable-fp16=true --buildtype=plain
  ninja -C build   # RTX smoke: rms_norm registers on the cuda context (12/12 appcontext smoke)

Note: this rung was built and validated in the same pass as #4117/#4118 but its PR was never opened — filling the gap so the C-stack (#4119+) has a complete review chain.

Depends-on: #4118


jijoongmoon:

PR #4144, opened 2026-07-24
[causallm] Clamp the runtime context length to max_position_embeddings

jijoongmoon:
setupParameters accepted MAX_SEQ_LEN/INIT_SEQ_LEN beyond the model's max_position_embeddings, but the RoPE tables are only built up to that cap — a longer run walks compute_rotary_emb_value off the end of the table (SIGSEGV, gdb-confirmed on a 32768-cap model asked for 33,536). Clamp both knobs to max_position_embeddings at setup and warn once, so an oversized request degrades to the model's real limit instead of crashing.

Repro (pre-fix): request a context beyond the model's max_position_embeddings and prefill past the cap — first out-of-table RoPE read faults. Post-fix: clamped + one-shot warning, run completes.


Applications/CausalLM/models/transformer.cpp
```diff
@@ -179,6 +180,26 @@ void Transformer::setupParameters(json &cfg, json &generation_cfg,
...
+                 "[causallm] WARNING: requested window (max_seq_len=%u, "
+                 "init_seq_len=%u) exceeds the model's "
+                 "max_position_embeddings=%u; clamping (RoPE tables only "
+                 "cover the model range).\n",
+                 MAX_SEQ_LEN, INIT_SEQ_LEN, MAX_POSITION_EMBEDDINGS);
+    if (MAX_SEQ_LEN > MAX_POSITION_EMBEDDINGS)
+      MAX_SEQ_LEN = MAX_POSITION_EMBEDDINGS;
+    if (INIT_SEQ_LEN > MAX_SEQ_LEN)
+      INIT_SEQ_LEN = MAX_SEQ_LEN;
+  }
```
jayden0701, line 208: The _[P2] Clamp the generation budget after shrinking max length_ claim by chatgpt-codex-connector is correct.
Need to add below


```suggestion
  }
  // NUM_TO_GENERATE is parsed independently of MAX_SEQ_LEN above, so the
  // clamp just above (or a plain misconfiguration) can leave
  // NUM_TO_GENERATE >= MAX_SEQ_LEN. CausalLM::run() computes prefill budgets
  // as MAX_SEQ_LEN - NUM_TO_GENERATE in unsigned arithmetic; left unchecked,
  // that wraps around to a huge value instead of rejecting the request,
  // silently disabling the truncation/bounds checks it exists to enforce.
  if (NUM_TO_GENERATE < 0 ||
      static_cast<unsigned int>(NUM_TO_GENERATE) >= MAX_SEQ_LEN) {
    const int clamped =
      (MAX_SEQ_LEN > 0) ? static_cast<int>(MAX_SEQ_LEN) - 1 : 0;
    std::fprintf(stderr,
                 "[causallm] WARNING: num_to_generate=%d does not leave room "
                 "for a prompt within max_seq_len=%u; clamping "
                 "num_to_generate to %d.\n",
                 NUM_TO_GENERATE, MAX_SEQ_LEN, clamped);
    NUM_TO_GENERATE = clamped;
  }
```

jijoongmoon:
Re-checked against #4316 in the re-cut series: that PR clamps the generation window against max_seq_len and the history-buffer stride, but it does not clamp the runtime context length to max_position_embeddings, which is what this PR does. Leaving this open rather than closing it against a PR that does not carry its content.

jijoongmoon:

As noted above, the change here is NOT carried by that series: the shipped generation-window clamp bounds the output buffer but does not clamp the runtime context length to max_position_embeddings.

PR #4145, opened 2026-07-24
[gpu/cl] ClBufferPool: skip the cl_mem plane for SVM-only offsets

jijoongmoon:
ClBufferPool allocates both an SVM arena and a cl_mem plane for every pooled offset range. Offsets whose consumers are all SVM-classified never touch the cl_mem plane — allocating it anyway doubles the long-lived GPU memory for those ranges (a MAX_LIFESPAN SVM-resident tensor, e.g. an int8 KV cache at long context, wastes hundreds of MB this way). This skips the cl_mem plane for SVM-only offset ranges and asserts if such a range ever sees a real cl_mem bind; ranges shared with a GPU_CLMEM consumer keep both planes, so real cl_mem bindings stay byte-identical.

Repro:
  meson setup build -Denable-opencl=true -Denable-fp16=true --buildtype=plain
  ninja -C build   # libnntrainer.so links; NNTR_CLMEM_SVM_SKIP=0 opts out for A/B

Depends-on: #4114


jijoongmoon:
Superseded by the re-cut series against current main: #4309, though not by transplant — the defect
this PR fixes cannot occur there. Closing to keep the queue reviewable; the branch is preserved on the
fork, so this is reversible.

**Row re-verified before closing.** This PR's content is a patch on `ClBufferPool::allocate()`: it
eagerly created one `cl_mem` sub-buffer for every planner offset, so an offset used only by
SVM-classified tokens — a long-lived cache the host also touches, say — committed an equally large
device buffer that nothing would ever bind. The fix collected the SVM-only tokens, skipped their
offsets, and put a `NNTR_CLMEM_SVM_SKIP=0` kill-switch behind it in case the skip was ever wrong.

In #4309 the pool is lazy rather than eager. `ClBufferPool::deviceMemory()` creates the buffer for a
planner offset on the first request and returns the same handle for every later token at that offset;
`allocate()` records the offset map and allocates nothing. A tensor the planner classified to the
shared plane never asks, so no buffer is created for its offset — which is exactly the outcome this
PR computed, arrived at by construction instead of by a second pass over the classification. A pool
whose tensors all classify to the shared plane creates no device buffer at all.

That also retires the kill-switch. There is no always-allocate behaviour left to restore, and the
"skip" is no longer a decision that could be wrong: nothing is skipped, nothing is eager.
`NNTR_CLMEM_SVM_SKIP` does not exist in #4309, which adds zero new `getenv`/`setenv` calls in total.

The reuse property this PR was careful about is preserved and now pinned by a test: tensors the
planner placed at one offset have disjoint lifetimes and must bind the SAME `cl_mem`, because
distinct handles over one region are not coherent through a per-handle device cache.
`unittest_cl_residency` asserts it directly (9/9 on an Intel Xe integrated GPU).

PR #4148, opened 2026-07-24
[GPU/OpenCL] flash attention prefill: fp16 XMX/DPAS tile kernel (default-on on DPAS devices)

jijoongmoon:
Adds flash_attention_prefill_f16_xmx — an fp16 DPAS/XMX tile kernel for the flash-attention prefill path — and routes flash_attention_prefill_f16_cl to it on DPAS-capable devices (ClContext caps().dpas; NNTR_FLASH_XMX=0 opts out, falling back to the existing blockq kernel through the untouched pre-existing dispatch ternary). The kernel: 8-row×16-key subgroup tiles via f16_f16_matrix_mad_k16, dual-subgroup d-split workgroups (FXA_NSG, default 4 at d=512), TM=16 dual-fragment row tiles sharing K/V loads, batched cross-subgroup exchange, V-tile SLM staging. Lever guards clamp XB against the per-WG SLM budget and force NSG=1 on non-divisible shapes. Two opt-in diagnostics ship with it: NNTR_FLASH_XMX_CHECK=N (dual-run host-reference compare on the first N calls) and NNTR_FLASH_TIMER.

Why: at d>=256 the blockq prefill path is scalar-ALU bound; measured on open models this kernel takes gemma4 (d=512) 32K full-attention prefill ~3.4x, with gemma2 (d=256) and qwen3 (d=128) covered by the same source (per-d compile variants).

Validation: -Werror clean builds (opencl+fp16 on/off); Intel Xe3 hardware CHECK-mode runs across all three head_dims — gemma4 d=256/512, gemma2 d=256, qwen3 d=128 — maxdiff 1e-4..8e-3, over-tolerance 0 of ~28k checked elements per call, zero crashes; NNTR_FLASH_XMX=0 fallback verified structurally (dispatch ternary unchanged vs base) and empirically (three models, clean runs). Note for reviewers: at this rung the in-repo CausalLM app cannot yet drive this entry point (flash_attention_prefill_f16_cl has no in-tree caller until the upcoming CausalLM GPU-dispatch PRs), so the hardware evidence above comes from the development tree that this kernel was extracted from verbatim — a semantic diff of the extracted files against that tree's fp16 path was verified empty.

Depends-on: #4112


jijoongmoon:
Superseded by the re-cut series against current main: #4312. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

Scope note. This branch's head commit (the fp16 XMX/DPAS flash prefill tile kernel, default-on where the device advertises `cl_intel_subgroup_matrix_multiply_accumulate`) is carried in the new PR as its own commit, on top of the attention kernel family the same branch depends on. Two deliberate differences from what is here:

- The KV ring-capacity kernel argument that this branch adds on top of the tile kernel is **not** carried. It is only meaningful with a bounded sliding-window KV cache, and both the ring modulo and the host side that computes a capacity ship together in the long-context PR later in the series. A parameter that can only ever be zero at this rung would be dead code upstream.
- Every raw OpenCL entry point in the dispatch file is called through the ICD loader (`opencl::cl*`) rather than linked directly, and `clEnqueueBarrierWithWaitList` was added to the loader's export list for it.

Evidence for the kernel is stated the way the arithmetic requires: it is not bit-identical to the scalar block-Q kernel, so md5 equality is not claimed. `NNTR_FLASH_XMX_CHECK` runs both kernels on the same buffers and reports 0 of 8192 elements over tolerance with a maximum absolute difference of 0.000244 on an Intel Xe iGPU, and `NNTR_FLASH_XMX=0` reproduces the scalar path exactly.

PR #4149, opened 2026-07-24
[GPU/OpenCL] Fused RMSNorm+quant resident kernel, per-layer-slice fp16 body, clmem probe capture

jijoongmoon:
Three core CL pieces consumed by the CausalLM GPU layer-variant PR that follows: (1) fused_rmsnorm_quant_resident_fp32 — a fused RMSNorm→per-row-int8-quant kernel writing a device-resident output (new fused_rmsnorm_quant.cl), with its self-check host reference; (2) the out-of-line body for per_layer_slice_cl_fp16 (its declaration already ships in blas_kernels.h — this makes the symbol real); (3) clmem_probe_capture, a small probe hook for cl_mem residency diagnostics. All verbatim extractions; no consumer in-tree until the next PR, dormant-but-compiled (same precedent as the attention-kernel PRs).

Validation: -Werror clean at {opencl+fp16, opencl+fp16-off, opencl-off} × clblast=false; the only test failures are the two pre-existing Q4_0-macro files already addressed by #4099.

Depends-on: #4116


jijoongmoon:
Note before this one is closed, so the record is accurate: **#4311 carries none of this PR's content**, and this PR should stay open until the rung that does.

The re-cut plan maps this PR to two new PRs. The one that just opened, #4311, is the XMX/DPAS prefill GEMM and the weight-upload path; when its contents were cut it took nothing from `afee1a02d`. Reading that commit is what settled it: all three symbols it adds — `per_layer_slice_cl_fp16`, `fused_rmsnorm_quant_resident_fp32` and `clmem_probe_capture` — exist to be linked by the GPU layer-variant slice that arrives with the model layers, and the commit says so itself. They are env-gated or simply unused until that caller is registered. Shipping 745 lines of host dispatch plus a debug value probe one rung early would put them in a PR whose reviewer has no way to see them exercised, which is the opposite of what the split is for.

So this PR's real home is the later rung, whole, next to its consumers — not split across two. Its current review (dated 2026-08-27, at this branch's head) has not been harvested yet and should be read when that rung is cut, not now.

The branch is preserved at `upstream-pr/trackG8-cl-fused-rmsnorm-slice-kernels`.

jijoongmoon:
Leaving this open deliberately, and recording the split so it is not lost.

#4318 carries one of this PR's three parts: the **per-layer-slice FP16 body**.
That is `per_layer_slice_cl_fp16` with its kernel and host entry point, the
`PerLayerSliceLayerGPU` consumer, and the host `PerLayerSliceLayer` dtype fix
(the FP32 branch was being taken for an FP16 tensor, striding four bytes over
two-byte cells).

The other two parts are **not** in it:

- the fused RMSNorm+quant resident kernel, and
- the clmem probe capture,

both of which depend on residency-plane machinery that is not in this cut's
ancestry. Closing this issue against that PR would therefore be false, so it
stays open until those two find a home.

jijoongmoon:

As noted above, only one of this PR's three parts is carried by the series: the per-layer-slice FP16 body, in #4318. The fused RMSNorm+quant resident kernel and the clmem probe capture are NOT carried anywhere in the series.

The branch is preserved, and the remaining two parts remain a candidate for a follow-up PR after the series lands.

PR #4150, opened 2026-07-24
[CausalLM] GPU-resident RMSNorm + per-layer-slice layer variants (additive registration)

jijoongmoon:
Adds rms_norm_gpu and per_layer_slice_gpu — CL-resident variants of the CausalLM RMSNorm and per-layer-slice layers — registered additively under the "gpu" engine factory (try/catch, in Transformer::registerCustomLayers). Nothing selects them unless a model config asks for the gpu engine, so default behavior is unchanged; this is the groundwork rung for the CausalLM GPU attention-dispatch PRs.

One deliberate deviation from verbatim extraction: the NNTR_NORM_DBG debug block in rms_norm_gpu.cpp used _FP16 without an ENABLE_FP16 guard (latent gap in the source tree, never exercised there) — wrapped in #ifdef ENABLE_FP16 so fp16-off builds compile.

Validation: full libcausallm.so builds clean (-Werror) at fp16-off and opencl-off; at opencl+fp16 the NEW libraries and the touched transformer.cpp compile and link cleanly, while two pre-existing sibling files (mha_core.cpp/deberta_attention_layer.cpp) fail at this rung with x86 fp16 backend signature mismatches that reproduce identically WITHOUT this PR (present on the base; flagged separately) — not introduced or touched here.

Depends-on: #4149 (trackG8), #4116


jijoongmoon:
Superseded by #4318, which lands this content on the re-cut series.

That PR adds `RMSNormLayerGPU` and `PerLayerSliceLayerGPU` and registers them on
the gpu context under the same type strings the host classes use, which is the
additive registration this PR proposed: each context owns its own factory map, so
a node asking for `engine=gpu` resolves to the GPU class and every other engine
keeps resolving to the host one.

It also carries a correctness fix this PR did not have. The FP32 `rmsnorm_cl`
kernel bounds its loops with `i < W / 4`, and that division truncates, so a width
that is not a multiple of four loses its last `W % 4` elements from both the sum
of squares and the output. At the golden test's `2:3:3:3` the width is 3, so both
loops ran zero iterations.

Closing here; the discussion is preserved and the branch is untouched.

PR #4151, opened 2026-07-24
[GPU/OpenCL] flash attention prefill: distributed psum-exchange reduction (FXA_XRED, Intel Xe2/Xe3 DPAS)

jijoongmoon:
The XMX/DPAS flash-attention prefill kernel's cross-subgroup psum exchange was all-to-all: every one of NSG subgroups read all NSG partials for all TM rows (NSG²·TM·KT SLM reads/tile, NSG-fold redundant). At d=512/NSG=4 this exchange is ~56% of the kernel and is pure SLM read/write volume (the workgroup barriers themselves cost ~0). FXA_XRED restructures it: each subgroup reduces only the rows it owns once, publishes the full score to a shared `ssum` buffer, then all read back (~2·NSG·TM·KT reads). Per-g sum order is preserved, so the output is bit-identical. Default-on for NSG>1 (compiles out entirely at NSG=1, i.e. d≤256); `NNTR_FLASH_XMX_XRED=0` forces the old path.

Intel Xe2/Xe3 only: the whole path is behind the `use_xmx` gate (caps().dpas = the Intel `cl_intel_subgroup_matrix_multiply_accumulate` extension); Adreno never advertises it, so this is a no-op there.

Validation (Intel Xe3, open models): gemma4 d512 full-attention bucket −31.5% (35.5→24.3 ms / 3 calls); end-to-end prefill gemma4 xmx **32K +40.3%** (718→1008 TPS) / dp4a +26.8%, since the d512 full-attention is 76.5% of the 32K prefill wall; 1K isolated-XRED A/B +4.9%; `NNTR_FLASH_XMX_CHECK` over-tolerance 0 / bit-identical on every call; gemma2 (d256, NSG=1) unaffected.

Repro:
  meson setup build -Denable-opencl=true -Denable-fp16=true --buildtype=plain
  ninja -C build   # NNTR_FLASH_XMX_CHECK=6 over-tol 0; NNTR_FLASH_XMX_XRED=0 for the A/B

Depends-on: #4148


jijoongmoon:
Superseded by the re-cut series against current main: #4312. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

Scope note. The distributed psum-exchange reduction is carried as its own commit in the new PR, directly on top of the tile kernel it optimizes, with the commit message rewritten to state the mechanism (each sub-group's partial score is exchanged between sub-groups, batched over several key tiles, instead of going through a shared-local round trip and a work-group barrier per tile) rather than the campaign shorthand. The batch depth stays adjustable through the same environment override for A/B work.

Measured on an Intel Xe iGPU, the tile kernel and this reduction together are worth a 32K-context prefill of 1165.3 TPS against 678.0 without them, and a 1K prefill of 4087.4 against 3141.8.

PR #4152, opened 2026-07-24
[cuda] Fused decode GEMV (dp4a act-quant folded in): opt-in paper-3.7 reference, default off

jijoongmoon:
dp4a_gemv_fused_h folds the per-row asymmetric int8 activation quantization into the decode GEMV kernel — the ML Drift paper's §3.7 decode form (quantize inside the operational kernel). fmin/fmax partition-independence plus integer dp4a make the output bit-identical to the existing act_quant_i8 + dp4a_gemv pair (verified byte-identical generated text, 3× A/B on gemma2 1K).

Shipped default-OFF behind `NNTR_CUDA_FC_FUSED_DECQ=1`: on RTX 5060 it measured a stable ~31% decode LOSS on both the SAFE and M2-B-graph profiles (gemma2 1K: 63-64 → 43.9 TPS) — the per-block quantize barriers serialize the weight streaming that the split kernels pipeline, and a discrete GPU's launch + q8-round-trip cost is smaller than that stall. It ships as (a) the paper-conformance reference and (b) an A/B lever for launch/sync-tax-bound environments (WDDM without graphs, mobile-class GPUs) where the paper's premise holds — measure before enabling.

Repro:
  meson setup build -Denable-cuda=true -Denable-fp16=true --buildtype=plain
  ninja -C build   # NNTR_CUDA_FC_FUSED_DECQ=1 to engage; default off = existing split-kernel path

Depends-on: #4120


jijoongmoon:
Row re-verified against #4313 before closing, and it does **not** match, so this stays open.

Two of the three commits on this branch — the w4a8 dp4a + cuBLAS int8-IMMA fast paths and the `dp4a_gemv` K % 4 != 0 fix — are in #4313 and were closed against it on #4120's row. This PR's own tip, the opt-in fused decode GEMV with the activation quant folded into the GEMV kernel, is **not**: #4313's `cuda_fc_qs4cx.cpp` has `dp4a_gemv`, `dp4a_gemm`, `dp4a_gemm_reg`, `fpact_gemv_i4_h` and `q6k_gemv`, and none of them folds the quant in. The fused *norm + act-quant* kernel it does have (`rmsnorm_quant_i8_h`, `NNTR_CUDA_FUSED_NORMQ`) is a different change.

Leaving this open rather than closing it against a PR that does not contain it. Branch preserved on the fork.

jijoongmoon:

As noted above, the change here is NOT carried by that series: the fused decode GEMV with the activation quant folded into the GEMV kernel itself is not among the shipped kernels (the series ships separate dp4a/fpact GEMV kernels and a separate norm+quant fusion, but not this specific fusion).

PR #4153, opened 2026-07-27
[meson] Fix CI failure on x86 configuration

jaemini-shin:

<details><summary>[meson] Fix CI failure on x86 configuration</summary><br />

This commit removes redundant -march=native flag which confilcts with
following -mavx2 -mfma at CI runs.
Since -march=native enables every available extensions, let's remove
this line and keep necessary flags only.

</details>

### Summary

This PR fixes the CI failure caused by `-march=native` which enables every available CPU extensions.

**Self evaluation:**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -131,10 +131,10 @@ elif arch == 'x86' or arch == 'x86_64'
   if cxx_compiler_id == 'msvc'
     add_project_arguments(['/arch:AVX2'], language: ['c','cpp'])
     # Note: C runtime library (/MT or /MD) is controlled by b_vscrt option
+    message('/arch:AVX2 added for AVX hardware acceleration.')
   else
-    add_project_arguments(['-march=native'], language: ['c','cpp'])
-    add_project_arguments(['-mavx2', '-mfma'], language: ['c','cpp'])
-    message('-march=native added for AVX hardware acceleration.')
+    add_project_arguments(['-mavx2', '-mfma', '-mf16c'], language: ['c','cpp'])
+    message('-mavx2, -mfma, -mf16c added for AVX hardware acceleration.')
   endif
 endif
 


haehun (approved):
LGTM!

Seunghui98 (commented):
## Description

Single hunk in the top-level `meson.build` x86 branch: drops `-march=native`, adds `-mf16c` alongside the existing `-mavx2 -mfma`, and adds a `message()` to the MSVC branch.

## Summary

- The `-mf16c` addition is load-bearing, not cosmetic. `x86/avx2_impl.cpp` calls `_mm256_cvtph_ps` / `_mm256_cvtps_ph` with only a `__TIZEN__ && !__F16C__` guard around them, so on any non-Tizen x86 build F16C must be enabled at compile time. It used to arrive implicitly via `-march=native`; without this line the build would break. Worth reflecting in the PR title, which reads as a pure removal.
- No build break from losing the wider ISA: every AVX512 path in `ggml_interface/nntr_ggml_impl/nntr_ggml_impl_utils.h` is behind `#if defined(__AVX512F__)` / `__AVX512VNNI__`, so they simply compile out.
- The runtime baseline doesn't effectively tighten either — F16C shipped with Haswell, the same generation as AVX2, so `-mf16c` excludes nothing that `-mavx2` didn't already exclude.

## Bug

Nothing wrong in the diff itself. One behavioral consequence worth verifying before merge (item 1).

1. **The Tizen F16C fallback becomes dead code.** `avx2_impl.cpp` carries a scalar path guarded by `__TIZEN__ && !__F16C__` (around lines 1540–1578), which implies some Tizen x86 target is built *without* F16C today. `packaging/nntrainer.spec` builds `%{ix86}` and `x86_64` through this same meson branch, so adding `-mf16c` unconditionally now defines `__F16C__` there and that fallback never compiles. If it exists because a Tizen emulator target can't execute F16C, this trades a CI failure for a runtime SIGILL. If the fallback is simply stale, it'd be good to delete it in the same PR so the intent is unambiguous.

2. **The commit message's stated cause doesn't match how the compiler works.** `-march=native` does not "conflict with" a following `-mavx2 -mfma` — GCC/Clang `-m<feature>` flags are additive and the later ones only ever turn features on. The real failure is almost certainly `-march=native` enabling ISA above the baseline of the machine that *runs* the tests (build host ≠ test host, or a container reporting different CPUID). Worth correcting the message, otherwise the next person hits a perf question and re-adds the flag.

3. **AVX512/VNNI paths are now unreachable on every x86 build.** The `__AVX512VNNI__` branches in `nntr_ggml_impl_utils.h` cover the Q4/Q8 dot products, i.e. the ggml GEMM hot path. Dropping to plain AVX2 everywhere is the right default for portability, but if local x86 ggml throughput matters to anyone, an opt-in meson option (`-Dx86-arch=native`) would preserve it without putting CI back where it was.

4. **Minor:** `-mtune=native` goes away with the same flag, so scheduling is now tuned for the generic model. Negligible next to the ISA question, just noting it's part of the change.

---

Seunghui98 (approved):
LGTM 👍

jayden0701 (approved):
LGTM

PR #4154, opened 2026-07-27
[CausalLM] Fix parallel_for deadlock in MoE expert loops

Jungwon-Lee:
## Dependency of the PR

N/A

<details><summary>[CausalLM] Fix parallel_for deadlock in MoE expert loops</summary><br />

[CausalLM] Fix parallel_for deadlock in MoE expert loops

ThreadManager::parallelize() uses a non-recursive execution_mutex_,
so nesting parallel_for (outer expert loop) inside dot() (which also
calls parallel_for) causes deadlock. Replace the outer parallel_for
with a serial for loop across all three MoE layer variants
(qwen3_moe, qwen3_slim_moe, qwen3_cached_slim_moe). The internal
GEMV/GEMM parallelism in dot() is preserved.

Also remove unused variables (active_experts, total_work, use_parallel)
in forwarding() that would trigger -Wunused-but-set-variable, and fix
-Wsign-compare in incremental_forwarding().

**Self evaluation:**

</details>

### Summary

- Replace the outer `parallel_for` with a serial `for` loop in all three MoE layer variants (qwen3_moe, qwen3_slim_moe, qwen3_cached_slim_moe) to resolve a deadlock caused by `ThreadManager::parallelize()` using a non-recursive `execution_mutex_` when nested inside `dot()`.
- Remove unused variables (`active_experts`, `total_work`, `use_parallel`) in `forwarding()` that would trigger `-Wunused-but-set-variable` build failure.
- Fix `-Wsign-compare` in `incremental_forwarding()` by casting `num_experts` to `int` for loop comparison.
- Internal GEMV/GEMM parallelism in `dot()` is preserved, minimizing performance impact.

### DIFF
--- a/Applications/CausalLM/models/gpt_oss/gpt_oss_moe_layer.cpp
+++ b/Applications/CausalLM/models/gpt_oss/gpt_oss_moe_layer.cpp
@@ -237,51 +237,22 @@ void GptOssMoELayer::forwarding(nntrainer::RunLayerContext &context,
     }
   }
 
-  // Adaptive optimization based on workload
-  const int active_experts =
-    std::count_if(expert_assignments.begin(), expert_assignments.end(),
-                  [](const auto &assignments) { return !assignments.empty(); });
+  // Serial outer loop: the expert GEMV/GEMM parallelizes internally via
+  // ThreadManager (dot() calls parallel_for), and nesting parallel_for
+  // deadlocks because ThreadManager::parallelize() uses a non-recursive
+  // execution_mutex_.
+  for (int expert_idx = 0; expert_idx < static_cast<int>(num_experts);
+       ++expert_idx) {
+    const auto &assignments = expert_assignments[expert_idx];
+    if (assignments.empty())
+      continue;
 
-  // Calculate total work (sum of token assignments across all experts)
-  int total_work = 0;
-  for (const auto &assignments : expert_assignments) {
-    total_work += assignments.size();
-  }
-
-  // Use parallel processing only when it's beneficial
-  const bool use_parallel = (total_work > 4) && (active_experts > 1);
-
-  if (use_parallel) {
-    // Parallel processing for larger workloads
-    auto &tm = nntrainer::ThreadManager::Global();
-    tm.parallel_for(
-      0, static_cast<size_t>(num_experts), [&](size_t expert_idx) {
-        const auto &assignments = expert_assignments[expert_idx];
-        if (assignments.empty())
-          return;
-
-        // Use optimized expert forward computation without memory copies
-        compute_expert_forward(
-          input, output, assignments,
-          context.getWeight(expert_gate_proj_indices[expert_idx]),
-          context.getWeight(expert_up_proj_indices[expert_idx]),
-          context.getWeight(expert_down_proj_indices[expert_idx]), hidden_size);
-      });
-  } else {
-    // Sequential processing for smaller workloads
-    for (int expert_idx = 0; expert_idx < static_cast<int>(num_experts);
-         ++expert_idx) {
-      const auto &assignments = expert_assignments[expert_idx];
-      if (assignments.empty())
-        continue;
-
-      // Use optimized expert forward computation without memory copies
-      compute_expert_forward(
-        input, output, assignments,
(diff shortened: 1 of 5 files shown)


jaemini-shin (approved):
LGTM!

Seunghui98 (commented):
## Description

Replaces the outer per-expert `parallel_for` with a serial `for` in three MoE layers (`qwen3_moe`, `qwen3_slim_moe`, `qwen3_cached_slim_moe`), drops the now-dead `active_experts` / `total_work` / `use_parallel` heuristic in `MoELayer::forwarding`, and cleans up the loop index types.

## Summary

The diagnosis holds up. `ThreadManager::parallelize()` takes a plain `std::lock_guard` on a non-recursive `execution_mutex_`, and it stores the callback in a single shared `task_` member — so nesting is unsupported at two independent levels, not just the lock. `compute_expert_forward` calls `Tensor::dot()` three times per expert (gate/up/down), each of which re-enters `parallel_for`, so the outer loop was violating the `@warning nested parallel_for is forbidden` contract already documented on `parallel_for`. Serializing the outer loop is a correct fix, and the conversions are faithful — no behavior beyond the loop structure changed.

## Bug

None in the changed code.

1. **The same bug is still live in the gpt_oss MoE layers.** `Applications/CausalLM/models/gpt_oss/gpt_oss_moe_layer.cpp:253-268` is a verbatim copy of the pre-fix `qwen3_moe` block — same `use_parallel = (total_work > 4) && (active_experts > 1)` gate, same `parallel_for` wrapping `compute_expert_forward`. `gpt_oss_cached_slim/gpt_oss_moe_layer_cached.cpp:324` (calling `compute_expert_forward` at 350 and 372) has it too. If the PR title says "MoE expert loops", these two are part of the same set and will deadlock identically.

2. **Consider fixing it in `ThreadManager` instead of at each call site.** A `thread_local` "already inside a parallel region" flag that makes `parallel_for` run the body inline on re-entry would cover all five MoE variants at once, protect future call sites, and downgrade a re-introduction from a hang to a perf question. Given the constraint is already documented but unenforceable by the caller, enforcing it centrally seems worth more than five separate serial loops.

3. **"minimizing performance impact" deserves a number for the FSU variants.** For `qwen3_slim_moe` and `qwen3_cached_slim_moe` the outer `parallel_for` wasn't only adding FLOPs — it overlapped `activate()` (mmap + storage read) for one expert with compute on another. Serializing removes that overlap entirely, which is a different cost profile from the pure-compute `qwen3_moe` case. A decode TPS before/after on the FSU path would confirm the claim.

4. **Worth noting in the commit message why this looked intermittent.** `parallel_for` short-circuits to inline execution when `end - begin == 1` or there are no workers, and the removed `use_parallel` gate additionally required `total_work > 4 && active_experts > 1`. So the hang only reproduced on multi-token steps that routed to 2+ experts — otherwise it reads as a flaky failure.

5. **Minor:** in `CachedSlimMoELayer::incremental_forwarding`, the `cache_mutex` `lock_guard`s around `hit_count` / `miss_count` / `loaded_expert_deque` are now uncontended overhead in a serial loop, and the `#ifdef DEBUG` hit/miss timings now measure serialized sections, so numbers collected before and after this change aren't comparable.

---

Seunghui98 (approved):
LGTM 👍

Seunghui98:
@Jungwon-Lee 

My Claude review is as follows. so, I have to fix it !!

> The same bug is still live in the gpt_oss MoE layers. Applications/CausalLM/models/gpt_oss/gpt_oss_moe_layer.cpp:253-268 is a verbatim copy of the pre-fix qwen3_moe block

Jungwon-Lee:
> @Jungwon-Lee
> 
> My Claude review is as follows. so, I have to fix it !!
> 
> > The same bug is still live in the gpt_oss MoE layers. Applications/CausalLM/models/gpt_oss/gpt_oss_moe_layer.cpp:253-268 is a verbatim copy of the pre-fix qwen3_moe block

I fixed it too. Thank you!!

jijoongmoon:
I think we need to support the recursive optimization of paralle_for. Because we have to parallelize not only computation but also weight loading as well. Please keep that in mind. And also it might be better to measure the performance as well.

PR #4159, opened 2026-07-28
[poc] cpp-linter test

jaemini-shin:

PR #4160, opened 2026-07-28
[ci] Fix uncatched whitespace errors

jaemini-shin:

<details><summary>[ci] Fix uncatched whitespace errors</summary><br />

- Add an explicit git diff --check for whitespace

cpp-linter runs with lines-changed-only, so it only reports clang-format
replacements whose start line falls inside the PR diff. When a changed
line is a blank line carrying trailing whitespace, clang-format emits a
single replacement that begins at the newline terminating the previous
line (the offset lands on the preceding, unchanged line). cpp-linter
attributes that fix to the previous line, which is not in the changed-line
set, and drops it -- so trailing whitespace introduced on such lines slips
through the format check.

</details>

### Summary
`clang-format-14` with `lines-changed-only` option which used in nntrainer's CI
can't detect some unformatted white spaces.
This PR introduces a explicit `git diff --check` which checks white space errors.

You can find the example at https://github.com/nntrainer/nntrainer/pull/4159.

### To do
* Introduce pre-commit hook
  * Remove white space check if matters

**Self evaluation:**

### DIFF
--- a/.github/workflows/cpp_linter.yml
+++ b/.github/workflows/cpp_linter.yml
@@ -11,6 +11,13 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: check whitespace errors on changed lines
+        run: |
+          echo "Checking whitespace errors on changed lines"
+          git diff --check ${{ github.event.pull_request.base.sha }}...HEAD \
+            -- . ':(exclude).github' ':(exclude)Applications/CausalLM/third_party'
       - name: install dev packages
         run: |
           echo "Install mandatory dev packages to avoid false-positive reports from cpp-linter"


Seunghui98 (commented):
## Description

Adds an explicit `git diff --check` step to `.github/workflows/cpp_linter.yml` (plus `fetch-depth: 0` on the checkout) to catch trailing-whitespace introduced on changed lines, which cpp-linter's `lines-changed-only` attribution drops when clang-format's replacement offset lands on the preceding unchanged line.

## Summary

The mechanism is right and the diagnosis in the commit message matches what the tools actually do:

- `base.sha...HEAD` is the correct spelling — three-dot in `git diff` means `merge-base(base.sha, HEAD)..HEAD`, and since `actions/checkout` puts you on the merge ref (which has `base.sha` as a parent) the merge-base *is* `base.sha`, so this diffs exactly the PR's changes.
- `git diff --check` reports only on lines the diff **adds**, so it sidesteps the attribution problem entirely rather than reproducing it — that's the part that makes this a real fix and not a second copy of the same bug.
- The repo has no `.gitattributes` and no `core.whitespace` override, so the defaults apply: `blank-at-eol`, `space-before-tab`, `blank-at-eof`. `blank-at-eol` is exactly the case #4159 slipped through on. Confirmed the check covers the reported failure.
- `Applications/CausalLM/third_party` exists at that path, so the exclude pathspec is live and not a silent no-op.
- `git diff --check` exits non-zero on findings, so the step fails the job on its own without needing a `failing fast` companion.

## Bug

None. The step does what it claims.

1. **The exclusion set diverges from the cpp-linter step in the same job.** The `cpp-linter-action` below ignores `.github|Applications/CausalLM/third_party`; the new step excludes only `third_party`. So `.github` is now whitespace-checked by one half of this workflow and skipped by the other, with a comment right underneath explaining the `third_party` skip and nothing explaining the divergence. Worth making them match — or, if covering `.github` is deliberate, saying so, since this PR edits a `.github` file itself and is the first thing the new rule applies to.

2. **The check is repo-wide but lives in a job named "C++ Format Checker".** `git diff --check` doesn't filter by language, so a trailing space added to a `.md`, `.py` or `.sh` file now fails a job whose name and every other step are about C++ formatting — a confusing place to go looking. That isn't hypothetical: 20 `.md` files and 23 `.py`/`.sh`/`.yml` files in the tree already carry trailing whitespace today, so the first person to edit one of those lines gets a C++ format failure. Markdown makes it worse, since two trailing spaces are the standard hard-line-break idiom and would be flagged as an error. Two ways out — scope the pathspec to C/C++ sources (`-- '*.cpp' '*.h' '*.hxx' ...`), which matches the job's purpose, or move the step to `static.check.yml` where a repo-wide rule belongs.

3. **`fetch-depth: 0` clones the full history on every PR.** It's the safe choice and I wouldn't block on it, but it's not free on a repo this size, and it's paid by every PR just to resolve one merge-base. If CI time starts to matter, fetching only what's needed for `base.sha` is the cheaper equivalent.

---

haehun (approved):
LGTM!

jayden0701 (approved):
LGTM

jijoongmoon (approved):
LGTM

PR #4161, opened 2026-07-29
[CausalLM] Fix tensor dimension of QS4CX embedding

jaemini-shin:

<details><summary>[CausalLM] Fix tensor dimension of QS4CX embedding</summary><br />

Since nntrainer uses width axis as quantized axis, we should transpose
the dimension of weight to allocate proper size of memory.

Self evaluation:

</details>

### Summary

Let's fix the dimension of QS4CX embedding in order to allocate proper size of memory.

As of now, quantization axis of per-channel quantized tensor should be the width axis. But in embedding layer, it is quantized along the height axis.

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/embedding_layer.cpp
+++ b/Applications/CausalLM/layers/embedding_layer.cpp
@@ -513,9 +513,27 @@ void EmbeddingLayer::finalize(nntrainer::InitLayerContext &context) {
 
   dim.setTensorType({context.getFormat(), context.getWeightDataType()});
 
-  dim.height(in_dim);
-  dim.width(out_dim);
   dim.batch(1);
+
+  /**
+   * @note nntrainer's per-channel quantized tensor is in following shape
+   * - quantized data: (H, W)
+   * - scales: (W)
+   *
+   * For embedding, there are in_dim elements in scale.
+   * So we should use (out_dim, in_dim) dimension although actual tensor
+   * shape is (in_dim, out_dim)
+   *
+   * @todo Add other types that needs to be transposed
+   * @todo Allow other axis can be a quantization axis
+   */
+  if (context.getWeightDataType() == nntrainer::TensorDim::DataType::QS4CX) {
+    dim.height(out_dim);
+    dim.width(in_dim);
+  } else {
+    dim.height(in_dim);
+    dim.width(out_dim);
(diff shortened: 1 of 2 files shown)


jayden0701 (commented):
It looks good, but one thing comes on my mind.
Like in the commit comment, if QS4CX is possible on embedding,
than i think QS4CX format lm_head might be possible. 

In that case I think `is_transposed` flag should be on lm_head?

jaemini-shin:
> It looks good, but one thing comes on my mind. Like in the commit comment, if QS4CX is possible on embedding, than i think QS4CX format lm_head might be possible.
> 
> In that case I think `is_transposed` flag should be on lm_head?

@jayden0701 You're right, the main purpose on this PR is QS4CX embedding and lm_head. Since QS4CX weight is only available for fully connected layer, I'm going to set `is_transposed` as true when supporting QS4CX lm_head!

jayden0701 (approved):
LGTM!

Seunghui98 (approved):
LGTM

nntrainer/layers/fc_layer.cpp
```diff
@@ -120,9 +120,17 @@ void FullyConnectedLayer::finalize(InitLayerContext &context) {
...
 
-  weight_idx[FCParams::weight] = context.requestWeight(
-    weight_dim, weight_initializer, weight_regularizer,
-    weight_regularizer_constant, weight_decay, "weight", true);
+  /**
+   * @note For QS4CX, weights are stored in (N,K) format while the tensor
+   * dimension is (K,N). To clarify this it, let's explicitly mark this feature.
+   */
+  bool is_transposed =
+    (context.getWeightDataType() == TensorDim::DataType::QS4CX);
```
jijoongmoon: I think this sholuldn't be set internally. If we set this is_transposed, then it only used for the QS4CX datatype. For now, it is ok but We do need to fix later. Please write TODO

jaemini-shin: @jijoongmoon  Do you mean this should be handled at model side not at each layer? Or is it enough to leave TODO message on other types?

jijoongmoon: How about fc_layer property?

PR #4162, opened 2026-07-29
[HTP] Add the Hexagon (HexKL) backend skeleton and its ComputeOps table

Seunghui98:
## Dependency of the PR

None.

<details><summary>[HTP] Add the Hexagon (HexKL) backend skeleton and its ComputeOps table</summary><br />

Registers a third compute vendor alongside CPU and GPU: an HtpContext
under engine="htp", backed by HtpBackend, which owns the FastRPC session
to the Hexagon DSP through the HexKL Macro API (sdkl.h / libsdkl.so).

No kernel is wired to it yet, so get_htp_ops() hands back the CPU table
as-is: every op falls through to the CPU implementation and an
engine="htp" model runs correctly, just without acceleration. The
accelerated overrides land with the kernels they dispatch to, and so does
the HtpComputeOps subclass that carries them -- with nothing overridden,
a subclass here would only duplicate get_cpu_ops(), so this commit does
not touch the CPU backend at all.

That fallback is the point of the design rather than a placeholder: the
NPU can fail to come up (no device, skel absent, pool exhausted) and the
context has to keep working when it does. HtpContext catches during
bring-up and installs the CPU table instead, so a model built for
engine="htp" never fails for lack of an NPU.

Build wiring: three meson options (enable-htp, hexkl-sdk-root,
hexkl-lib-subdir), off by default. The SDK is proprietary and not
vendored, so the caller passes the extracted addon root. hexkl-lib-subdir
defaults to armv8_android26 deliberately -- the armv9 prebuilt uses
instructions the Android app process faults on for current SM8750/V79
devices.

Every change to meson.build and jni/meson.build sits inside
`if get_option('enable-htp')`. With the option off -- which is the
default, and what CI builds -- both files behave exactly as they do
today; no existing target's compiler flags or source list changes.

Tests
-----
unittest_compute_ops_dispatch gains a smoke check that the HTP ops table
resolves and that an op on it still computes a correct result. Asserting
the result rather than a non-null pointer is what catches a table whose
entries throw, which is what the base ComputeOps defaults do -- so an
HTP context that stopped falling through to the CPU fails here. The check
needs no FP16, so it runs in every configuration that enables HTP.

**Self evaluation:**

</details>

### Summary

- Registers a third compute vendor (engine="htp") backed by the HexKL Macro API. No kernel is wired to it yet, so `get_htp_ops()` returns the CPU table directly and this PR does not touch the CPU backend -- the only change outside `htp_backend/` is a 5-line `get_htp_ops()` declaration in `compute_ops.h`.
- Three meson options, all off by default; the proprietary SDK is not vendored, and with `enable-htp` off no existing target's flags or sources change.
- Full suite 31/31 on an x86 host.


nntrainer/tensor/cpu_backend/cpu_ops_table.cpp
```diff
@@ -16,347 +16,10 @@
  * dispatch picks the right symbol at link time.
  */
 
-#include <compute_ops.h>
-#include <cpu_backend.h>
+#include <cpu_ops_table.h>
 
 namespace nntrainer {
 
-class CpuComputeOps : public ComputeOps {
```
dlwlzzero, line 24: This PR is related to HTP, but I don't understand why there are changes in the CPU area. Could you explain it to me?

dlwlzzero (commented):
Moreover, where are the docs I made about HTP?

nntrainer/tensor/cpu_backend/cpu_ops_table.cpp
Seunghui98: It seems to be extracted as `cpu_ops_table.h` for management.
This code is probably from your draft, so I'll erase it because it doesn't seem necessary here.

dlwlzzero: I see, I'll check why this change happened.

Seunghui98:
> Moreover, where are the docs I made about HTP?

@dlwlzzero Docs will be included in other prs.

meson_options.txt
```diff
@@ -33,6 +33,18 @@ option('enable-npu', type: 'boolean', value: false)
 # Alternatively: export QNN_SDK_ROOT=/opt/qcom/aistack/qnn-2.47.0
 # See docs/backend_guide/QNN_BUILD.md for details.
 option('qnn-sdk-root', type: 'string', value: '')
+# HTP (Hexagon Tensor Processor / HMX) backend via the Qualcomm HexKL
+# CPU Macro API (sdkl.h / libsdkl.so). Proprietary SDK is NOT vendored;
+# pass the extracted hexkl_addon root, e.g.:
+#   meson setup build -Denable-htp=true \
+#     -Dhexkl-sdk-root=/path/to/Hexagon_SDK/.../addons/hexkl_addon
+# ARM/Android only: there is no x86 libsdkl.so.
```
jaemini-shin: Only Qualcomm device can use it, not for every arm

meson.build
```diff
@@ -271,6 +271,39 @@ if get_option('opencl-kernel-path') != ''
...
+  hexkl_dep = declare_dependency(
+    dependencies: libsdkl,
+    include_directories: hexkl_inc)
+  extra_defines += '-DENABLE_HEXKL=1'
+  # sdkl.h uses the C99 `restrict` keyword in C++ function signatures; the NDK
+  # aarch64 C++ compiler rejects it.  Map restrict → __restrict to preserve the
+  # optimization hint while satisfying the C++ front-end.  This flag is
+  # module-wide under the guard; acceptable because HTP is the only sdkl.h
+  # consumer in nntrainer and the macro only affects an optimization hint.
+  extra_defines += '-Drestrict=__restrict'
```
jaemini-shin: This will affect through whole project. Is it possible to remove this flag? kleidiai is using `restrict` keyword but it works well.

Seunghui98: Good point! I'll update it!!

nntrainer/htp_context.cpp
```diff
@@ -0,0 +1,70 @@
...
+std::unique_ptr<nntrainer::Layer>
+HtpContext::createLayerObject(const int int_key,
+                              const std::vector<std::string> &properties) {
+  return AppContext::Global().createLayerObject(int_key, properties);
+}
+
+void HtpContext::initialize() noexcept {
+  // Ensure the CPU backend ops table exists — it is both the fallback
+  // and the base class of HtpComputeOps.
+  init_backend();
```
jayden0701: I think we should use `ensureComputeOps()` instead of `init_backend()`
since `ensureComputeOps()` is the canocical entry point

```cpp
void ensureComputeOps() {
  std::call_once(g_compute_ops_init_flag, []() { init_backend(); });
}
```

Seunghui98: Good point! I'll update it!!

Seunghui98:
@dlwlzzero @jaemini-shin @jayden0701 Could you reivew it again?

nntrainer/tensor/cpu_backend/cpu_ops_table.cpp
Seunghui98: I confirmed that it is necessary to add hexkl next PR.
so, I upload a new PR for this. https://github.com/nntrainer/nntrainer/pull/4168.
Could you review this?

jayden0701 (approved):
LGTM!

dlwlzzero (approved):
LGTM

jaemini-shin (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #4163, opened 2026-07-29
[HTP] Add the FP16 NPU matmul and route prefill and decode to it

Seunghui98:
## Dependency of the PR

The backend skeleton PR (#4162).

<details><summary>[HTP] Add the FP16 NPU matmul and route prefill and decode to it</summary><br />

Implements shgemm_f32f16_f32 on top of sdkl_npu_mm_f32f16_f32: FP32
activations against FP16 weights on the Hexagon HMX array. This is the
first kernel wired to the HTP ComputeOps table, so supports_shgemm() now
tracks whether the NPU actually came up and FloatTensor gates on it.

Both matmul paths route through it. Prefill (M>1) was the obvious target;
decode (M==1) goes to the same kernel rather than the CPU hsgemv, since
it reuses the resident weight and measured at parity or better on device.
Either path falls back to the CPU when the NPU is down or N is not
32-aligned -- the HMX tile width is a hard constraint, and checking it at
the call site keeps the CPU path reachable instead of throwing.

Two constraints are worth calling out because they are not obvious:

  - beta != 0 is rejected. The NPU write path does not accumulate into
    the destination, so honouring beta would mean a read-modify-write the
    kernel cannot do.
  - M is padded up to the SDKL tile internally and only the real rows are
    returned, so a caller never has to align M itself.

The weight side is where the cost is. sdkl wants WH (weight-Hexagon)
tiling, and converting per call dominated everything else, so converted
weights stay resident in NPU memory keyed by their host pointer. Prefill
touches each weight once, which makes an evicting cache thrash to ~0%
hits, so residency is pin-once up to a byte cap and then falls back to
the transient path rather than evicting. Activation and accumulator
staging buffers are grown on demand and reused across calls.

The CPU backend is touched here for one reason: HtpComputeOps has to
inherit CpuComputeOps so every op the NPU does not accelerate keeps
working, and CpuComputeOps was defined inside cpu_ops_table.cpp where
nothing else could reach it. It moves to a cpu_ops_table.h declaration --
a pure move; no method body changes. Deriving from ComputeOps instead
would make each non-overridden op throw, which is what the base defaults
do. The skeleton commit deliberately did none of this: with nothing to
override it returned get_cpu_ops() directly and left the CPU backend
alone.

FC-family layers store weights [K, N] and call dot(trans=false,
trans_in=false); the WH bake needs [N, K]. copyForWHBake does the real
transpose instead of relabelling the dims, which is what makes the NPU
output match the CPU rather than being silently scrambled.

Tests
-----
test/unittest/jni_htp/ is the ndk-build project for the on-device suites,
and it lands here rather than with the backend skeleton because this is
the commit that gives it something to test. The meson build does not
produce Android test binaries, so the two gtest suites below are built by
this Android.mk and run over adb.

unittest_nntrainer_htp_kernels (7 tests) and unittest_nntrainer_htp_backend
(8 tests) cover accuracy against a CPU reference for prefill and for the
[K, N] calling convention, M-padding, the beta and N-alignment guards,
residency reuse and pin-once behaviour, and scratch reuse across mixed
shapes.

ZeroWeightGivesZero is worth a note: a zero weight is degenerate on the
CPU but not here, because the accumulator is staged in a scratch buffer
that outlives the call, so a kernel that forgets to clear it returns the
previous call's C and still matches a CPU reference everywhere the two
happen to agree. The test primes the output with a non-zero sentinel and
runs after the accuracy test has filled that scratch, which is what gives
it teeth.

unittest_compute_ops_dispatch gains gating tests proving a tensor whose
ops advertise shgemm reaches the override, and one whose ops do not falls
back.

The device suites are compiled only with enable-htp and are skipped at
runtime when no NPU is present. On an x86 host the library and
unittest_compute_ops_dispatch build and pass.

**Self evaluation:**

</details>

### Summary

- shgemm_f32f16_f32 on HMX, with prefill and decode both routed to it and a CPU fallback when the NPU is down or N is not 32-aligned.
- WH-converted weights stay resident in NPU memory, pin-once rather than evicting, because prefill touches each weight once.
- Carries the `CpuComputeOps` -> `cpu_ops_table.h` move (a pure move, no body changes): `HtpComputeOps` must inherit it so the ops the NPU does not accelerate keep working. This is the first commit with something to override, which is why the move is here and not in the skeleton PR.
- Adds the jni_htp ndk-build project and two on-device gtest suites (kernels 7, backend 8); they build only with enable-htp and self-skip without an NPU. Host suite 31/31.


Seunghui98:
we don't need this PR, so I closed it

PR #4168, opened 2026-07-30
[tensor] Extract CpuComputeOps into its own header

Seunghui98:
## Dependency of the PR

None.

<details><summary>[tensor] Extract CpuComputeOps into its own header</summary><br />

CpuComputeOps was defined inside cpu_ops_table.cpp, so nothing outside
that translation unit could name it -- including an accelerator backend
that wants to inherit it. An accelerator typically overrides a handful
of ops and needs the other eighty to keep working; the base ComputeOps
defaults throw, so inheriting CpuComputeOps is what makes that possible.
Today ClComputeOps derives from ComputeOps and accepts the throw for
everything it does not implement, which is a fine choice for the GPU but
not the only one a backend should be able to make.

This is a pure move: the class declaration goes to cpu_ops_table.h and
cpu_ops_table.cpp keeps only get_cpu_ops(). No method body is edited, no
signature changes, and nothing new is compiled -- every one of the 338
moved lines is byte-identical to what it replaced. The header joins the
installed set (meson, debian, spec) because a downstream backend needs it
at build time.

**Self evaluation:**

</details>

### Summary

Moves the `CpuComputeOps` declaration from `cpu_ops_table.cpp` into a new
`cpu_ops_table.h`. Nothing else changes: no method body, no signature, no
new code compiled.

Why it is worth doing, in numbers: `ComputeOps` declares 88 virtual ops and
every base default throws (`NI(op)` -> `throwNotImplemented`).
`CpuComputeOps` implements 74 of them. An accelerator backend that
accelerates a handful of ops -- the HTP backend accelerates three -- has to
choose between inheriting that CPU table or having the other eighty-five
throw the moment a model touches them. `ClComputeOps` takes the second
option today, which suits the GPU; a backend that wants to stay usable when
its device is absent needs the first, and cannot take it while the class is
locked inside a .cpp.

- Pure move: all 338 relocated lines are byte-identical to what they replaced.
- Header added to the installed set (meson / debian / spec) so a downstream
  backend can include it.
- Prerequisite for the HTP backend PRs; contains no HTP code and no behaviour
  change for any existing backend.

Signary-off-by placeholder removed -- see trailer below.


nntrainer/tensor/cpu_backend/cpu_ops_table.cpp
```diff
@@ -16,347 +16,10 @@
  * dispatch picks the right symbol at link time.
  */
 
-#include <compute_ops.h>
-#include <cpu_backend.h>
+#include <cpu_ops_table.h>
```
Seunghui98, line 19: ## Note for Reviewers
#### Why do you extract `cpu_ops_table`?
`ComputeOps` has 88 virtual ops, and every base implementation throws. 
`CpuComputeOps` fills in 74 of them. So a backend that accelerates only a few ops -- HTP accelerates three -- either inherits that CPU table or lets the remaining 85 throw as soon as a model reaches them. 
The GPU accepts the throws; HTP cannot, because it has to keep working with no NPU present.

jayden0701 (approved):
LGTM!

haehun (approved):
LGTM, it looks better to move them into headers

jijoongmoon (approved):
LGTM

PR #4171, opened 2026-07-30
[tensor] Support runtime page size for virtual tensor mmap

Jungwon-Lee:
## Dependency of the PR

None

<details><summary>[tensor] Support runtime page size for virtual tensor mmap</summary><br />

[tensor] Support runtime page size for virtual tensor mmap

Virtual tensors aligned model-file offsets to a hard-coded 4 KiB boundary before calling mmap. This fails with EINVAL on systems using larger pages, such as 16 KiB Android configurations, because mmap requires its file offset to be aligned to the runtime page size.

Query _SC_PAGESIZE during activation and deactivation and use it to calculate the aligned mapping offset and length. Keep the mmap result local until the mapping succeeds so a failed mapping does not leave MAP_FAILED in tensor state, and request MADV_WILLNEED on Linux as well as Android.

Add a virtual tensor lifecycle test with an unaligned file offset to verify that activation reads the expected values and deactivation releases the mapping.

**Self evaluation:**

</details>

### Summary

- Align virtual tensor mmap offsets to the runtime system page size instead of assuming 4 KiB pages.
- Keep failed mmap results out of tensor state and cover unaligned virtual tensor activation and deactivation.

### DIFF
--- a/nntrainer/tensor/tensor.cpp
+++ b/nntrainer/tensor/tensor.cpp
@@ -1663,7 +1663,11 @@ void Tensor::activate() {
 #else
 
   auto file_offset = getFileOffset();
-  size_t off = (file_offset / 4096) * 4096;
+  const long system_page_size = sysconf(_SC_PAGESIZE);
+  NNTR_THROW_IF(system_page_size <= 0, std::runtime_error)
+    << "[activate] failed to query the system page size";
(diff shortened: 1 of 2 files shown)


dlwlzzero (approved):
LGTM

baek2sm (approved):
LGTM

PR #4174, opened 2026-07-30
[layers] Fix 2x output-buffer overflow in ActiFunc::gelu/tanhGelu under FP16 activations

jijoongmoon:
`ActiFunc::gelu` and `ActiFunc::tanhGelu` are templates on the activation element type `T`, but their
bodies hardcode `getData<float>()` on an **element** count. `Tensor::getData<T>()` is an unchecked
cast that does not consult the tensor's real data type, so with `T = _FP16` the call hands a buffer of
N 2-byte elements to a routine that reads N 4-byte floats and writes N 4-byte floats: it reads 2x past
the end of the input and **overflows the output buffer by 2x**. Heap corruption, not merely wrong
numerics.

This is reachable. `ActivationLayer::finalize` really does select the `_FP16` instantiation when
`getActivationDataType() == FP16`, and ten of the `ModelTensorDataTypeInfo` strings are `*-FP16`.
In-tree consumers of the two affected activation nodes: `gelu` in the BERT, TIMM-ViT and DeBERTa-v2
CausalLM models, `tanh_gelu` in Gemma3. Any of those loaded with an `*-FP16` `model_tensor_type`
corrupts the heap in its MLP activation. It went unnoticed because every `model_tensor_type` committed
to this tree is `FP32-FP32` or `Q4_0-FP32`, so CI only ever instantiates `T = float`, where the cast
happens to be correct.

Fix keeps the accelerated fp32 path (`gelu_v2` / `tanh_gelu` are AVX2- and NEON-vectorized;
de-SIMDing fp32 would be a real regression) behind an `if constexpr (std::is_same_v<T, float>)` and
routes every other `T` through `Tensor::apply<T>()`, which walks the tensor with the correct element
type and additionally throws if the output dtype disagrees — exactly what the neighbouring
`sigmoidGelu` already does. The lambdas accumulate in float and cast back to `T`, reproducing the
existing `__fallback_gelu_v2` / `__fallback_tanh_gelu` fp16 references bit for bit.

**New in this rung:** `[layers] Fix 2x output-buffer overflow in ActiFunc::gelu/tanhGelu under FP16
activations`.

**Not included:** no new activation kernels, no dispatch changes.

**Validated:** CPU build (`-Denable-opencl=false -Denable-cuda=false -Denable-fp16=false`) and an
fp16-enabled build compile clean; the fp32 path is unchanged by construction.
**Honest empty cell:** there is **no in-tree `*-FP16` activation fixture**, so this cannot be closed
by a committed regression test today. Closing it needs either a committed FP16-activation reference
config or an ASAN build over one of the four affected models.

---

### DIFF
--- a/nntrainer/layers/acti_func.h
+++ b/nntrainer/layers/acti_func.h
@@ -19,10 +19,8 @@
 #include <common_properties.h>
 #include <cpu_backend.h>
 
-#if defined(_WIN32)
-#define _USE_MATH_DEFINES
-#include <math.h>
-#endif
+#include <cmath>
+#include <type_traits>
 
 namespace nntrainer {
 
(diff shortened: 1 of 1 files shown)


jijoongmoon:
Consolidated into #4194.

PR #4175, opened 2026-07-30
[init-latency] Stop double zero-filling QS4CX weight payloads

jijoongmoon:
`QS4CX_Tensor::allocate()` zeroed its payload **twice** — `new uint8_t[size()]{}` and then
`initialize() -> setZero()` — and on the only path that reaches it in an inference process both
passes are dead stores: the weight's self-owned heap payload is overwritten in full by
`NeuralNetwork::load` (`TensorBase::read` writes `bytes()`, which for QS4CX equals `size()`;
`copy_qs4cx` scopy's `size()`). The zeroing was pure init latency.

This allocates uninitialized and skips `initialize()` on that path (`putData()` is still called, so
the `MemoryData` invalidate hook fires exactly as before). Page faults are **not** saved — they move
to the `read()` that fills the buffer — but the two full-arena writes are.

The argument is only as strong as the claim "every reader writes all `size()` bytes", so that was
audited on this tree and the audit is in the commit message: `TensorBase::read(std::ifstream)` /
`read(ReadSource)` write `bytes() == size()`; `copy_qs4cx` covers `size()`; `readFSU` is a no-op and
the path is gated `!enable_fsu`. There is no other reader — `QS4CX_Tensor` has no `read()` override
here. A tripwire is recorded in the `qs4cxAllocUninitialized()` doc comment for the one known partial
writer that does **not** exist in this tree (a legacy on-disk QINT4 → QS4CX transcode leaves an
N/2-byte hole between the nibble and scale regions for even K), so whoever ports that read path in
cannot miss the requirement.

Escape hatch: `NNTR_QS4CX_ALLOC_ZERO=1` restores the old double zero-fill — which is also the bisect
handle for any init-latency claim built on this rung.

Measured on a prior device session (35-layer sliding-window QS4CX-FP16 model, Intel Xe3, OpenCL,
warm, 3 runs each): `model_graph.initialize` **731.8 ms → 8.3 ms**, whole init **1593/1572/1514 ms →
893/979/947 ms**, peak RSS ~3.15 GB → ~1.7–2.0 GB, loader-join time +81 ms (the faults that moved),
generated text byte-identical to baseline.

**New in this rung:** `[init-latency] L1: stop double zero-filling QS4CX weight payloads`.

**Not included:** the Windows `VirtualAlloc` bypass branch (this tree's `allocate()` has only the CRT
heap branch), and the legacy QINT4 transcode guard (nothing to guard here — recorded as a tripwire
instead of a stub).

**Not verified on this tree:** golden-text identity with the lever on, and the hatch control
(`NNTR_QS4CX_ALLOC_ZERO=1`) reproducing the pre-port `init_ms`. The numbers above are from the prior
device session, not from this branch.

---

### DIFF
--- a/nntrainer/tensor/qs4cx_tensor.cpp
+++ b/nntrainer/tensor/qs4cx_tensor.cpp
@@ -9,10 +9,60 @@
  */
 
 #include <cpu_backend.h>
+#include <cstdlib>
 #include <qs4cx_tensor.h>
 #include <tensor.h>
 
 namespace nntrainer {
+
+namespace {
+/**
+ * @brief VALUE-checked env truthiness: set, non-empty, and not starting with
+ *   '0', so that NNTR_QS4CX_ALLOC_ZERO=0 really is "off" rather than "present,
+ *   therefore on". Kept file-local on purpose: this translation unit must
+ *   compile against a tree that ships no shared env helper, and a header that
+ *   only this file would consume does not belong in the public include set.
+ */
+bool envOn(const char *name) {
+  const char *e = std::getenv(name);
+  return e != nullptr && e[0] != '\0' && e[0] != '0';
+}
+
+/**
+ * @brief [init-latency L1] Is this process's QS4CX payload allocation
+ *   load-destined, i.e. may allocate() hand back UNINITIALIZED memory?
+ *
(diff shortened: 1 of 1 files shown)


jijoongmoon:
Consolidated into #4212.

PR #4177, opened 2026-07-30
[tensor] x86 host QS4CX GEMM for HalfTensor::dot

jijoongmoon:
`origin/main`'s `half_tensor.cpp` has **zero** QS4CX references, so `HalfTensor::dot` has no QS4CX case
at all: an fp16-activation QS4CX model has no host GEMM to fall back to on x86. This adds one.

Fix by delegation rather than a fourth hand-rolled decoder: the new case calls
`gemm_qai8dxp_qsi4cxp_rhs_unpacked(..., is_nxk=true)` — the exact x86 entry point
`FloatTensor::dotQs4cx` already uses — with the fp16 activation widened to fp32 per chunk and the
fp32 result narrowed back, instead of decoding the packed nibble layout a third time. Staging is
`chunk` floats per worker, not `M*N` (an untied lm_head is `N = 262144`).

**Why this is one squashed commit, not two.** The internal precursor of this port hand-decoded the
nibble payload and got the layout wrong three independent ways (wrong axis order, wrong nibble half,
signed instead of unsigned offset — the full derivation against the two in-tree layout authorities,
the GPU v8c weight builder and the host reference GEMM, is in the commit message). Publishing that
precursor even for one commit would introduce a known-wrong decoder that never existed upstream, so
only the corrected version is here.

Also carries four small x86 fp16-activation gaps that the same work unmasked, unrelated to the QS4CX
decoder: `rms_norm` / `reshaped_rms_norm` read gamma at `context.getWeightDataType()` instead of a
hardcoded FP32 (matching `*-FP16` packages that store norm gammas at the model weight dtype),
`fallback_internal_fp16`'s `rms_norm_wrt_width_fp16` scalar reference (was an NYI throw), and
`Transformer::repack_weight` skipping on NYI per weight so x86 (no KleidiAI rhs-pack) loads the plain
QS4CX blob instead of aborting.

**New in this rung:** `[tensor] x86 host QS4CX GEMM for HalfTensor::dot (squash, correct decoder only)`.

**Not included:** the ARM case in the same `#if/#elif` block
(`nntr_gemm_qai8dxp_qsi4cxp_packed<_FP16>`) — it depends on the KleidiAI fp16 packed-weight
infrastructure, which is a separate PR in this series, and does not exist on `main` (0 grep hits in
`half_tensor.cpp`). The x86 `#elif` therefore becomes a standalone `#if`; the case body, guard
condition and parallel-chunk logic are otherwise unchanged.

**Not verified here (git-only work; measurement host was in use):** x86 + `ENABLE_FP16` + OpenCL
build; `NNTR_ENGINE=cpu` on qwen3/gemma4/gemma2; `Gemma4DifferentialTest.Q40CloseToFP32Reference`;
`unittest_causallm_models`. The switch-case adaptation was verified by inspection against
`float_tensor.cpp`'s identical `dotQs4cx` pattern only.

**Three-axis note:** structure ++ (one decoder per backend, matching the existing x86/ARM/GPU QS4CX
authorities instead of adding a fourth); performance ~ (adds an `M*K` fp32 staging copy plus int8
requantization per call, on the host QS4CX-FP16 fallback path only); memory 0 (chunk-sized staging).

---

### DIFF
--- a/Applications/CausalLM/layers/reshaped_rms_norm.cpp
+++ b/Applications/CausalLM/layers/reshaped_rms_norm.cpp
@@ -35,14 +35,22 @@ void ReshapedRMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
     << "feature size must be a divisor of width";
 
   if (use_gamma) {
-    // gamma follows the model weight dtype (packed=false clamps it to the
-    // activation dtype), matching what the quantizer wrote to the bin.
-    // Hardcoding FP32 misloads/overruns *-FP16 bins whose gammas are FP16;
-    // the multiply site's dtype-cast guard covers either stored dtype.
+    // gamma is never quantized, but a bin stores it at the model's FLOAT
+    // dtype, so an FP16-activation package holds 2 bytes per element. Ask for
+    // the context's weight dtype when that dtype is itself a float type, and
+    // FP32 otherwise -- the same rule as rms_norm.cpp and the core
+    // nntrainer/layers/llm/rms_norm_layer.cpp, which carries the full note.
+    // These q/k/v norms are counted in the same packed weight stream, so a
+    // constant FP32 here shifts every later offset on an FP16 package.
+    const auto norm_w_dtype = context.getWeightDataType();
+    const auto norm_dtype =
+      (norm_w_dtype == nntrainer::TensorDim::DataType::FP32 ||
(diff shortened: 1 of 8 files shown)


jijoongmoon:
Consolidated into #4194.

PR #4179, opened 2026-07-30
[context] Make factory int_key assignment position-independent and fail loud

jijoongmoon:
Two coupled defects in the factory-registration path, both reproducible on `origin/main`.

1. **The auto `int_key` is derived from a count.**
   `int assigned_int_key = int_key == -1 ? str_map.size() + 1 : int_key;` (`app_context.cpp:682`).
   `str_map.size()` is a count, not a high-water mark, and the collision guard only runs for
   `int_key != -1`, so `int_map[k] = type` silently **rebinds** an existing key whenever the string
   map's size collides with an explicitly-assigned key. A registration mistake then resolves to the
   wrong layer type with nothing logged.

2. **`add_default_object()` runs before `setMemAllocator()`** (`cl_context.cpp:85` / `:89`). Any
   registration throw during default-object setup therefore leaves a fully "initialised" context with
   a **null `MemAllocator`**, and every model built on it dies later in the `TensorPool` constructor
   with a symptom that names nothing.

Fix: make the auto key position-independent (derive it from the actual maximum in use, not from a
count), run the collision guard on every assignment, and fail loud on a collision. Order
`setMemAllocator()` before `add_default_object()`. `AppContext::initialize` already orders them the
safe way (`app_context.cpp:251` vs `:260`), so this propagates an in-tree precedent rather than
inventing one. The idiom exists in **four** contexts; this change covers the core three plus the QNN
context.

**New in this rung:** `[context] Make factory int_key assignment position-independent and fail loud`.

**Not included:** the `cuda_context.cpp` copy of the same idiom — that file does not exist on `main`;
it is covered by the CUDA-context PR in this series.

**Honest note:** the canary for defect 1 was run first and did **not** fire. The mitigation already in
the tree (a comment asking contributors to append last, plus explicit keys 9001/9002) held. The
mechanism is still there, and defect 2's null-allocator window is unconditional.

**Validated:** CPU and OpenCL builds compile clean; registration order/keys/error paths for existing
registrations are unchanged.

---

### DIFF
--- a/nntrainer/app_context.cpp
+++ b/nntrainer/app_context.cpp
@@ -679,7 +679,13 @@ const int AppContext::registerFactory(const FactoryType<T> factory,
     return int_key;
   }
 
-  int assigned_int_key = int_key == -1 ? str_map.size() + 1 : int_key;
+  // Auto key from max(existing) + 1, not from the registration COUNT: a
+  // count-derived key shifts when a registration is inserted mid-list and can
+  // walk onto an explicitly-requested LayerType value, which the assignment
+  // below would then silently rebind. AppContext keeps its pr/3963 "duplicate
+  // explicit key is a no-op, reuse it" policy above; this only removes the
+  // position dependence of the auto path.
+  int assigned_int_key = int_key == -1 ? nextAutoIntKey(int_map) : int_key;
 
   str_map[assigned_key] = factory;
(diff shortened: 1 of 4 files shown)


jijoongmoon:
Consolidated into #4194.

PR #4180, opened 2026-07-30
[build/android] Make the CLBlast-free OpenCL build configurable and reachable

jijoongmoon:
Three independent Android build defects, none of which any CI job can see (no CI job builds the
CausalLM app or runs `ndk-build`). Together they are why the Android GPU build was unreachable from
the documented entry point, and they are the fix for the Android arm64 "Build CausalLM" job that is
red on the app chain of this series.

1. **`-Denable-clblast=false` cannot configure on `platform=android`.**
   `meson -Dplatform=android -Denable-opencl=true -Denable-clblast=false` dies at configure time with
   `jni/meson.build:77: ERROR: Unknown variable "clblast_root"`. The CLBlast template substitution was
   gated on `clblast_dep.found()`, but on the `enable-clblast=false` path `clblast_dep` is a bare
   `declare_dependency()` — whose `.found()` is **true**. The guard never fires and the body reads a
   variable only assigned in the `true` branch. The same guard also owned `MESON_CL_ROOT`, so even
   without the crash the staged `libOpenCL.so` / CL headers would drop out of the generated
   `Android.mk`. Fix: gate on `get_option('enable-clblast')` — the same authority that decides whether
   `clblast_root` gets assigned — hoist `MESON_CL_ROOT` out of the CLBlast branch, and export a
   `MESON_HAS_CLBLAST` marker so `Android.mk.in` can drop the CLBlast static-library module instead of
   compiling ~100 CLBlast sources from an empty root prefix. The `clblast_dep.found()` idiom was
   vacuously correct before the option existed; adding the option turned a vacuous test into a live one
   on a path nothing exercises.

2. **`build_android.sh` cannot forward meson options.** It accepts only `--cache` and `exit 1`s on
   anything else, and invokes `tools/package_android.sh` with no arguments — and `enable-opencl`
   defaults to false. So the documented way to build the Android app could only ever produce a CPU-only
   nntrainer. `package_android.sh` already filters `-D*` / `--arm-arch=*` out of its argv and forwards
   them to meson, so the only missing piece was passing them through:
   `./build_android.sh -Denable-opencl=true -Denable-clblast=false` now works. Also warns when meson
   args are combined with `--cache`, since that path skips the reconfigure and the cached options win.

3. **A missing jni source ⇒ undefined vtable.** `jni/Android.mk` never listed
   `models/bert/multilingual_tinybert_16mb.cpp`, although meson does. `MultilingualTinyBert` declares
   all virtuals inline except `encode()`, which is an implicit override and therefore the class's key
   function — so the vtable is emitted only in that TU, and the two Android modules that instantiate
   the class (`causallm_core` via the model registry, `nntr_quantize` via `quantize.cpp`) link with an
   undefined `vtable for causallm::MultilingualTinyBert`. The jni source lists are hand-maintained and
   completely independent of the meson lists, so adding a model to the app fails only on a real NDK
   build.

**New in this rung:** the three commits above.

**Not included:** the `ENABLE_OPENCL` export to prebuilt consumers — that one has an app-chain
dependency and rides a separate PR in this series.

**Not verified here:** NDK build + on-device deploy. The three defects are configure-time /
link-time and were each reproduced from the error text quoted above.

### The `enable-clblast` declaration, and how it relates to #4191

This PR is the declaring PR of record for `enable-clblast`. Its first commit introduces the
`get_option('enable-clblast')` reads in `nntrainer/meson.build` and `jni/meson.build` **and** declares
the option in `meson_options.txt`, in that same commit. Meson resolves option names at configure time,
so a tree that reads an undeclared option aborts before any target is built; keeping the declaration
with its first reader means no commit in this PR leaves a tree that cannot run
`meson setup -Denable-opencl=true`.

#4191 carries a byte-identical declaration, because it reads the option as well. The two declarations
are the same text at the same position, immediately after `option('enable-opencl', ...)`, so the two
PRs merge cleanly in either order and whichever lands second is a no-op for `meson_options.txt`.

The reach of the knob differs between the two trees, which is worth stating precisely. On this PR's
tree alone, `-Denable-clblast=false` controls the Android path — the `MESON_HAS_CLBLAST` marker that
drops the CLBlast static-library module from the generated `Android.mk` — and the
`subprojects/CLBlast/include` entry on the compile line. It does not by itself stop a host build from
configuring and compiling the CLBlast cmake subproject. #4191 adds the gate on that subproject and
`-DENABLE_CLBLAST=1`, which is what makes `-Denable-clblast=false` actually skip the multi-minute
CLBlast compile, together with the matching `#ifdef ENABLE_CLBLAST` in
`nntrainer/tensor/cl_operations/clblast_interface.cpp`. That guard is what lets the translation unit
compile once the CLBlast headers are gone, so an `ndk-build` of the CLBlast-free configuration wants
both changes: this PR for the configure step, #4191 for the source-level guard. Merging the two yields
a single option declaration, the include-directory gate from this PR, and the `#ifdef` from #4191. The
comment above that gate in `nntrainer/meson.build` records the same dependency, so a reader of the
build file does not have to infer it.

---

### DIFF
--- a/Applications/CausalLM/build_android.sh
+++ b/Applications/CausalLM/build_android.sh
@@ -6,16 +6,29 @@ set -e
 
 # Parse options
 USE_BUILD_CACHE=0
+MESON_ARGS=()
 while [[ $# -gt 0 ]]; do
     case $1 in
         --cache)
             USE_BUILD_CACHE=1
             shift
             ;;
+        -D*|--arm-arch=*)
+            # Forwarded verbatim to tools/package_android.sh, which already
+            # filters -D*/--arm-arch=* out of its own argv and hands them to
+            # meson. Without this, options the nntrainer build needs -- most
+            # importantly -Denable-opencl=true for the GPU build -- are
+            # unreachable from this script.
+            MESON_ARGS+=("$1")
+            shift
+            ;;
         *)
             echo "Unknown option: $1"
-            echo "Usage: $0 [--cache]"
-            echo "  --cache  Reuse existing nntrainer builddir if available"
+            echo "Usage: $0 [--cache] [-D<meson-option>=<value>]... [--arm-arch=<version>]"
+            echo "  --cache           Reuse existing nntrainer builddir if available"
+            echo "  -D<opt>=<val>     Forwarded to meson, e.g."
+            echo "                    -Denable-opencl=true -Denable-clblast=false"
+            echo "  --arm-arch=<ver>  Forwarded to meson, e.g. --arm-arch=armv9.2-a"
             exit 1
             ;;
     esac
@@ -103,12 +116,16 @@ log_info "NNTRAINER_ROOT: $NNTRAINER_ROOT"
 log_info "Build cache: $([ "$USE_BUILD_CACHE" -eq 1 ] && echo 'enabled' || echo 'disabled (default)')"
 log_info "ANDROID_NDK: $ANDROID_NDK"
 log_info "Working directory: $(pwd)"
+log_info "meson args: ${MESON_ARGS[*]:-(none)}"
 
 # Step 1: Build nntrainer for Android if not already built
 log_step "1/4" "Build nntrainer for Android"
 
 if [ "$USE_BUILD_CACHE" -eq 1 ] && [ -f "$NNTRAINER_ROOT/builddir/android_build_result/lib/arm64-v8a/libnntrainer.so" ]; then
     log_info "Build cache enabled: reusing existing nntrainer builddir (skipping)"
+    if [ ${#MESON_ARGS[@]} -gt 0 ]; then
+        log_warning "meson args are ignored with --cache; the cached builddir keeps its own options."
+    fi
 else
     log_info "Building nntrainer for Android..."
     cd "$NNTRAINER_ROOT"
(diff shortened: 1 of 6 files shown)


jijoongmoon:
Consolidated into #4191.

#4191 is now the task-level Android build PR: "the Android GPU build was silently a CPU build." It carries all three commits from this PR (script option forwarding, the `enable-clblast` false-path configure crash, the missing `multilingual_tinybert_16mb.cpp` jni source) plus the `ENABLE_OPENCL` export to prebuilt consumers and the loud failure on `engine=gpu`/`engine=cuda` in a build without the backend. Same failure mode, overlapping files (`Applications/CausalLM/jni/Android.mk`, `jni/meson.build`, `nntrainer/meson.build`, `meson_options.txt`, `llm_util.hpp`), one review.

Containment verified mechanically: this PR's entire delta (`git diff <merge-base>..head`) **reverse-applies cleanly** against the new #4191 tree, i.e. every line this PR adds is present there and every line it removes is absent. (`git cherry` flags one commit only because the byte-identical `enable-clblast` declaration in `meson_options.txt` was already on #4191's branch — the duplicate that both bodies documented is now resolved to a single declaration.)

Both halves of the `enable-clblast` knob — the Android `MESON_HAS_CLBLAST` marker and include gate from here, the host-side cmake-subproject skip and the `#ifdef ENABLE_CLBLAST` guard from #4191 — now sit in one PR, which is what a CLBlast-free `ndk-build` actually needs.

#4191 lands with "Rebase and merge" (never squash), so each of these three defects keeps its own revert handle.

PR #4181, opened 2026-07-30
[cpu_backend] Add QS8CX dequantization

jaemini-shin:

<details><summary>[cpu_backend] Add QS8CX dequantization</summary><br />

- Add dequant_qs8cx_f32 and dequantize_row_qs8cx for arm, x86, fallback
  - only nxk format is implemented
- Add unittest for fallback correctness

</details>

### Summary

This PR updates dequantization for `QS8CX` type matrix and row.

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
@@ -657,6 +657,23 @@ void quant_qs8cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
                                  (float *)rhs_scales_f32);
 }
 
+void dequant_qs8cx_f32(size_t n, size_t k, void *rhs_native_mtx_qs8cx,
+                       void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+  __fallback_dequant_nxk_qs8cx_f32(n, k, (const int8_t *)rhs_native_mtx_qs8cx,
+                                   (const float *)rhs_scales_f32,
+                                   (float *)rhs_native_mtx_f32);
+}
+
+void dequantize_row_qs8cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs8cx,
+                          void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+  // qs8cx dequantization is a memory-bound int8->f32 widening. The compiler
+  // auto-vectorizes the scalar fallback into NEON at -O2/-O3, so a hand-written
+  // NEON kernel offered no speedup over the fallback and was dropped.
+  __fallback_dequantize_row_qs8cx_f32(
+    n_idx, k, (const int8_t *)rhs_native_mtx_qs8cx,
+    (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+}
+
 size_t get_rhs_packed_size_qsi4cxp_qs4cxs1s0(size_t n, size_t k,
                                              size_t idx_variant, bool is_nxk) {
 #ifndef ARMV7
--- a/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
+++ b/nntrainer/tensor/cpu_backend/arm/arm_compute_backend.h
@@ -1544,6 +1544,44 @@ void quant_qs8cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
                      void *rhs_native_mtx_qs8cx, void *rhs_scales_f32);
 
 /**
+ * @brief qs8cx dequantization of rhs matrix in nxk format. Inverse of
+ * quant_qs8cx_f32. Converts quantized int8 back to float32 using per-channel
+ * scales. qs8cx refers to quantized symmetric 8-bit channel-wise quantization.
+ *
+ * @warning You should allocate memory for output before use:
+ *  - rhs_native_mtx_f32
+ *    n * k * sizeof(float)
+ *
+ * @param[in] n N for (M, K) * (K, N) = (M, N) in noTrans GEMM
+ * @param[in] k K for (M, K) * (K, N) = (M, N) in noTrans GEMM
+ * @param[in] rhs_native_mtx_qs8cx quantized matrix data
+ * @param[in] rhs_scales_f32 per-channel quantization scales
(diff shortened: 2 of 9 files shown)


nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -657,6 +657,23 @@ void quant_qs8cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
...
+
+void dequantize_row_qs8cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs8cx,
+                          void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+  // qs8cx dequantization is a memory-bound int8->f32 widening. The compiler
+  // auto-vectorizes the scalar fallback into NEON at -O2/-O3, so a hand-written
+  // NEON kernel offered no speedup over the fallback and was dropped.
+  __fallback_dequantize_row_qs8cx_f32(
+    n_idx, k, (const int8_t *)rhs_native_mtx_qs8cx,
+    (const float *)rhs_scales_f32, (float *)rhs_native_mtx_f32);
+}
```
jaemini-shin, line 675: Unlike `QS4CX`, row dequantization is more likely memory bound and well auto-vectorized.

Seunghui98 (commented):
## Description

QS8CX counterpart to the QS4CX dequantization in #4136: adds `__fallback_dequant_nxk_qs8cx_f32` and `__fallback_dequantize_row_qs8cx_f32`, wires public `dequant_qs8cx_f32` / `dequantize_row_qs8cx` entry points on the arm/x86/fallback backends, and adds four fallback unit tests. nxk only.

## Summary

Round-trip semantics check out against `__fallback_quant_nxk_qs8cx_f32`:

- The quantizer stores `recip_scale0 = 1/scale0` in `rhs_scales_f32[n_idx]`, and dequant multiplies by it — correct inverse.
- Layout matches on both sides: `rhs + n_idx * k`, one byte per value, no packing. The row helper's stride is plain `k`, which is right for int8 (the `(k + 1) / 2` of the QS4CX version doesn't apply here).
- Codes are clamped to `[INT8_MIN, INT8_MAX]`, and dequant reads through an `int8_t*`, so `-128` sign-extends correctly. No off-by-one at the range ends.

The comment on the ARM `dequantize_row_qs8cx` — that the scalar loop auto-vectorizes at -O2/-O3 and a hand-written NEON kernel gave no speedup, so it was dropped — is worth having in the tree. A recorded negative result stops the next person from re-doing the experiment.

## Bug

None found.

1. **The `_odd_k` tests don't test anything here.** For QS4CX, `k=7` was meaningful: nibble packing made the last byte half-used, so odd `k` exercised a real tail path. QS8CX has one byte per value and a stride of exactly `k`, so `k=7` and `k=8` run identical code. The edge cases that would actually be informative for int8 are a **constant-valued row** (`rmin0 == rmax0` takes the `scale0 = 1.f` branch, so the stored scale is `1.0f` and dequant returns the raw codes) and a row that saturates at `INT8_MIN`/`INT8_MAX`. Swapping one `_odd_k` case for a constant row would cover a branch nothing currently reaches.

2. **The assertions restate the implementation.** Both tests compute `expected = (float)code * rhs_scales_f32[n_idx]`, which is exactly what the dequant does, so they can catch an indexing or stride mistake but not a wrong formula — if the scale were stored non-reciprocal on both sides, every test would still pass. That's still worth having for the stride coverage, but one extra assertion comparing against the *original* f32 within a quantization tolerance would close the formula end of it.

3. **`dequant_qs8cx_f32` has no `is_nxk` parameter, unlike its QS4CX sibling.** That matches `quant_qs8cx_f32`, which is also nxk-only, so the pair is self-consistent and I don't think it needs the flag. But the name doesn't carry the constraint either, so a kxn buffer passed by a caller who generalized from `dequant_qs4cx_f32(..., is_nxk)` gets silent garbage. A one-line `@note nxk only` on the public declarations (or `dequant_nxk_qs8cx_f32` as the name) would make it unmissable.

4. **The measurement behind the "NEON gave no speedup" comment isn't recorded.** Since that claim is the justification for shipping without a NEON path, the shape and device it was measured on would be useful in the commit message — otherwise the next person has only the conclusion and no way to know whether it still holds on different hardware.

5. **Minor:** the dispatch bodies in `fallback.cpp`, `x86_compute_backend.cpp` and `arm_compute_backend.cpp` are byte-identical apart from the ARM comment. That's the established pattern in this directory, so no objection — noting it only because the ARM copy is the one carrying the rationale, and the full-matrix `dequant_qs8cx_f32` there has no equivalent note even though it takes the same fallback route for the same reason.

---

nntrainer/tensor/cpu_backend/arm/arm_compute_backend.cpp
```diff
@@ -657,6 +657,23 @@ void quant_qs8cx_f32(size_t n, size_t k, void *rhs_native_mtx_f32,
...
+  __fallback_dequant_nxk_qs8cx_f32(n, k, (const int8_t *)rhs_native_mtx_qs8cx,
+                                   (const float *)rhs_scales_f32,
+                                   (float *)rhs_native_mtx_f32);
+}
+
+void dequantize_row_qs8cx(size_t n_idx, size_t k, void *rhs_native_mtx_qs8cx,
+                          void *rhs_scales_f32, void *rhs_native_mtx_f32) {
+  // qs8cx dequantization is a memory-bound int8->f32 widening. The compiler
+  // auto-vectorizes the scalar fallback into NEON at -O2/-O3, so a hand-written
+  // NEON kernel offered no speedup over the fallback and was dropped.
```
Seunghui98, line 671: it's just out of curiousity.

> NEON kernel offered no speedup over the fallback and was dropped.

May I ask a question for reason?

Seunghui98: Ah, I found it!!

> Unlike QS4CX, row dequantization is more likely memory bound and well auto-vectorized.

Seunghui98 (approved):
LGTM 👍

jayden0701 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #4182, opened 2026-07-30
[causallm] Consolidate attn_logit_softcapping parse into base setupParameters (dead override)

jijoongmoon:
The Gemma-family attention soft-cap parse lived in a **dead override**.

`setupParameters()` is only ever dispatched from base-class **constructor** bodies
(`Transformer::Transformer`, `CausalLM::CausalLM`). C++ resolves a virtual call made inside a base
constructor to the *base* version, never a derived override, because the derived object does not exist
yet. A model whose own constructor body adds no `setupParameters()` call of its own therefore never
runs its override, and every key parsed only there keeps its default. `Gemma3Transformer` is exactly
that shape, so the `attn_logit_softcapping` parse it held never ran, `ATTN_LOGIT_SOFTCAPPING` kept its
`0.0f` default, and — since every consumer is gated `if (attn_logit_softcapping > 0.0f)` — the
`c*tanh(s/c)` soft-cap was skipped on every layer, on every backend.

Fix: move the parse into `Transformer::setupParameters` — the one function the base constructor
provably reaches — and delete the per-model copies. The parse stays gated on
`contains(...) && !is_null(...)`, so any model whose config lacks the key keeps the `0.0f` default and
is byte-preserving.

Affected vs unaffected, audited per model:
- **gemma3 / EmbeddingGemma** — plumbing cured, **numerics unchanged**: the override was unreachable,
  but the in-tree gemma3 configs carry `"attn_logit_softcapping": null`, so the gate leaves `0.0f`.
- **gemma4** — unchanged. Its override is live (`Gemma4Transformer`'s ctor body explicitly calls
  `setupParameters()`, which calls the base first). The deleted copy re-read the same key from the same
  cfg into the same member, so the base parse cannot double-apply. `final_logit_softcapping` stays in
  gemma4 (its own member).
- **qwen2 / qwen3 / qwen3_moe / gpt_oss / lfm2 / bert / deberta / xlm_roberta / timm_vit** — no such
  key in their configs; `0.0f` default preserved.
- The real numerics victim is **gemma2**, whose shipped configs all carry
  `"attn_logit_softcapping": 50.0`. Gemma2 is not on this tree yet — it arrives in the next PR of this
  series, and this fix is its prerequisite: without it the new model would silently run with its
  soft-cap off.

**New in this rung:** `[causallm] Consolidate attn_logit_softcapping parse into base setupParameters`,
plus `Gemma3ConfigPlumbingTest` and a shared `makeTinyCtorOnlyNntrainerConfig()` helper in
`causallm_test_utils`.

The test is ctor-level parameter *visibility*: a probe subclass exposes the protected member and
constructs the model with no tokenizer, no weights and no graph compile, so each assertion observes
exactly what the constructor chain parsed. That is precisely the layer where the bug lived — a derived
override can be silently unreachable while everything still builds and links. Three cases: key present
(the regression guard, which **fails on the unfixed tree**), key absent, key explicitly null.

**Merge-order note:** this PR ships `causallm_test_utils.{h,cpp}`, which the Gemma2 PR's unit test
consumes. Please land it **before** the Gemma2 model PR.

**Validated:** CPU build clean; the three new test cases pass on the fixed tree and the
"key present" case fails on the unfixed tree.

---

### DIFF
--- a/Applications/CausalLM/models/gemma3/gemma3_causallm.cpp
+++ b/Applications/CausalLM/models/gemma3/gemma3_causallm.cpp
@@ -70,10 +70,7 @@ void Gemma3Transformer::setupParameters(json &cfg, json &generation_cfg,
   if (cfg.contains("layer_types")) {
     layer_types = cfg["layer_types"].get<std::vector<std::string>>();
   }
-  if (cfg.contains("attn_logit_softcapping") &&
-      !cfg["attn_logit_softcapping"].is_null()) {
-    ATTN_LOGIT_SOFTCAPPING = cfg["attn_logit_softcapping"].get<float>();
-  }
+  // attn_logit_softcapping now parsed in Transformer::setupParameters (base).
 }
 
 Tensor Gemma3Transformer::createTransformerDecoderBlock(const int layer_id,
--- a/Applications/CausalLM/models/gemma4/gemma4_causallm.cpp
+++ b/Applications/CausalLM/models/gemma4/gemma4_causallm.cpp
@@ -115,10 +115,8 @@ void Gemma4Transformer::setupParameters(json &cfg, json &generation_cfg,
     layer_types = cfg["layer_types"].get<std::vector<std::string>>();
   }
 
-  if (cfg.contains("attn_logit_softcapping") &&
-      !cfg["attn_logit_softcapping"].is_null()) {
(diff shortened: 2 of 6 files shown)


jijoongmoon:
Consolidated into #4194.

PR #4183, opened 2026-07-30
[causallm] Honor nntr_config lmhead_untie: untied FC lm_head in the generic and Gemma4 graphs

jijoongmoon:
The app only read `config.json`'s `tie_word_embeddings` when choosing the lm_head layer type;
`nntr_config.json`'s `lmhead_untie` key was ignored entirely. Untied-serialized packages (tied HF
config plus a separate transposed `[hidden, vocab]` head record appended to the bin) therefore built a
**tied** graph against an **untied** bin. That used to misload silently; with the loader tripwire in
this series it now fails fast with a model-layout mismatch — either way the package is unusable.

- `Transformer::setupParameters` parses `lmhead_untie` into a new `LMHEAD_UNTIE` member. It lives on
  `Transformer`, not `CausalLM`, because `embedding0`'s layer-type choice happens in
  `<model>Transformer::constructModel` scope, which does not see `CausalLM` members in the diamond
  hierarchy.
- New `CausalLM::buildLmHeadOutput(h, add_skip_prefill)`: untied ⇒ an independent
  `fully_connected` output with its own weight (`weight_dtype = LMHEAD_DTYPE`, optional
  `skip_prefill` so the head stays decode-only); tied ⇒ the existing
  `tie_word_embeddings` / `lm_head` choice with `shared_from`, unchanged. Shared by the generic path
  and by Gemma4, each keeping the exact skip-prefill gate its inline copy had.
- `Gemma4Transformer::constructModel` gates `embedding0` on
  `TIE_WORD_EMBEDDINGS && !LMHEAD_UNTIE`: with an untied head there is nothing to share, so
  `embedding0` becomes a plain `embedding_layer` (lookup-identical, same weight record), which also
  makes the `embedding_file_name` sidecar legal for it.

Byte-preserving for every package without the key: `LMHEAD_UNTIE` defaults false, the emitted property
strings and their order are unchanged, and no `skip_prefill` is added on paths that did not add it
before. The generic base `constructModel` intentionally keeps a `tie_word_embeddings`-typed
`embedding0` even when untied, because a standalone `TieWordEmbedding` is a plain lookup.

**New in this rung:** `[causallm] Honor nntr_config lmhead_untie: untied FC lm_head in the generic and
Gemma4 graphs`.

**Not included:** no change to the tied path's numerics, no new kernels, no loader changes (the
tripwire that makes a mismatch diagnosable is a separate PR in this series).

**Not verified here:** tied goldens (qwen3 / gemma2 / gemma4 packagings) byte-identical, and an
untied-serialized gemma4 package passing the load tripwire and producing coherent output. Both are
pending build-host time; the change is byte-preserving by construction for every package that does not
set the key.

---

### DIFF
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -220,26 +220,43 @@ void CausalLM::advanceKVCachePosition(unsigned int step_size) {
   kv_cache.advance(step_size);
 }
 
-std::pair<Tensor, Tensor> CausalLM::constructModel() {
-
-  // base transformer (input, output_norm)
-  auto [x, h] = Transformer::constructModel();
-
+/**
+ * [lmhead-untie] When nntr_config.json sets lmhead_untie, build
+ * output_of_causallm as an independent fully_connected layer with its own
+ * weight even for a tied-embedding model, so the lm_head can carry a
+ * different dtype than the embedding (untied-serialized packages such as
+ * gemma4_qs4cx_fp16 ship a separate transposed [hidden, vocab] head record
+ * that a tied graph cannot load). Untie is the config flag, NOT derived from
+ * LMHEAD_DTYPE: a quantizer constructs this same untied graph from the FP32
+ * source and quantizes output_of_causallm via the dtype map on save.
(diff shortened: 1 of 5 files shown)


jijoongmoon:
Consolidated into #4194.

PR #4184, opened 2026-07-30
[Wait for #4105,4106][Applications] Register custom layers through the Engine facade, not a downcast

jijoongmoon:
> **Depends on #4106** — please merge that first. The registration facade this PR calls
> (`Engine::registerLayerFactory`, its per-backend seam `Context::registerLayerFactory`, and the
> `AppContext` override) is **not** on `main`; it is introduced by #4106. An earlier revision of this
> PR stated the facade was already on `main` — it is not, and on plain `main` all 16 translation units
> here fail to compile with `error: 'class nntrainer::Engine' has no member named
> 'registerLayerFactory'`, taking out `libcausallm` plus the `Custom/LayerClient`, `LLaMA`,
> `SimpleShot`, `YOLOv2` and `YOLOv3` targets and therefore every Ubuntu/Tizen/pdebuild job. The
> branch is now stacked on #4106 (which is itself stacked on #4105), so the call sites and the facade
> can only land together.

19 app files registered their custom layers by downcasting the context:

```
auto app_context = static_cast<nntrainer::AppContext *>(ct_engine.getRegisteredContext("cpu"));
app_context->registerFactory(nntrainer::createLayer<X>);
```

This converts them to `ct_engine.registerLayerFactory("cpu", nntrainer::createLayer<X>)`. Semantics
are **identical, not merely similar**: `AppContext::registerLayerFactory` forwards straight to
`registerFactory<nntrainer::Layer>`, which is the exact overload the downcast resolved to — every
converted site passed `nntrainer::createLayer<X>`, whose type is `PtrFactoryType<Layer>`, so `T` was
already deduced as `Layer` everywhere. `Engine::getRegisteredContext` throws `std::invalid_argument`,
the same exception type the surrounding `try`/`catch` blocks already handle. No registration order,
key, `int_key` or error-path change.

**Why this is the 16-file half.** The source change touches 19 files; 16 of them are entirely public
and need nothing beyond the facade — `deberta_v2`, `gemma3`, `gemma4`, `gpt_oss`,
`gpt_oss_cached_slim`, `lfm2`, `qwen3`, `qwen3_cached_slim_moe`, `qwen3_moe`, `qwen3_slim_moe`,
`sentence_transformer`, `Custom/LayerClient`, `LLaMA`, `SimpleShot`, `YOLOv2`, `YOLOv3`. This PR is
exactly those 16, each hunk byte-identical to the source. The remaining files
(`causal_lm.cpp`, `transformer.cpp`) depend on the model-registry PR of this series and ride a
separate tail PR.

Verification of the *split*, not just the port: `git grep -nE '[-]>registerFactory[[:space:]]*\('
-- Applications test` is 29 hits / 18 files on `main` and 6 hits / 2 files here — exactly the
two files this PR deliberately excludes. The 16 files it does touch carry 0 remaining downcast calls.

A note for whoever maintains the contribution checklist: the prescribed single-line detection grep
(`static_cast<.*Context *\*>.*registerFactory`) never matches, because every real site splits the cast
and the call across lines (39 lines apart in one case). Check the **call** instead — it is single-line
by construction. Do not use a bare downcast grep: `static_cast<.*Context *\*>` alone matches 155 lines
tree-wide, almost all of them `static_cast<ClContext*>(getRegisteredContext("gpu"))` inside kernel
wrappers fetching a kernel, which is a different pattern.

**New in this PR:** `[Applications] Register custom layers through the Engine facade, not a
downcast`.

**Known remaining, deliberately not converted:** three SimpleShot unit tests call
`AppContext::Global().registerFactory(...)` and then `createObject<Layer>` on the same object — direct
singleton use, not a context downcast, and the facade has no `createObject` counterpart.

**Verified:** `g++ -fsyntax-only` with the flags an x86 OpenCL build actually uses, over the 14 of
these 16 files that build configures — they fail on the pre-#4106 base with the missing-member error
above and compile clean with #4106 underneath (14/14). `clang-format` 14.0.6 clean on the changed
lines.

**Not verified here:** a link of the touched Applications and one model run per app family to
confirm the custom layer types still resolve at runtime. Startup-path-only change; one extra virtual
call per custom layer type, nothing on any inference path.

---


jijoongmoon:
Consolidated into #4214.

Both rungs are about how a process acquires a backend: #4214 stops the engine layer from constructing every context unconditionally and brings up only the one the process asked for (symmetric ClContext / CudaContext gate), and this PR stops the Applications from reaching past the Engine facade with a downcast. **The downcast is exactly what breaks once contexts are conditional** — a `static_cast<AppContext *>` asserts which concrete type `getRegisteredContext(name)` returned, and once bring-up is conditional "which contexts exist" becomes a function of `NNTR_ENGINE`. The facade path throws a named, catchable error instead, which is what makes #4214's `[Engine] gpu Context is not registered` line possible rather than a crash. Reviewed apart, that coupling is invisible.

Containment verified: this PR's rung commit **reverse-applies cleanly** against the new #4214 tree, i.e. every line it adds is present and every line it removes is absent. The post-change downcast sweep on the survivor is `3 hits / 1 file` (`transformer.cpp`, the file this rung deliberately excludes), down from 29 hits / 18 files on `main`.

The `[Wait for #4105,4106]` prefix has been dropped from the commit subject and the PR title — it was a merge-order marker, and #4214's base already carries both.

#4214 lands with "Rebase and merge" (never squash), so both rungs keep their own revert handles.

PR #4185, opened 2026-07-30
[init-latency] Overlap the chat-template renderer warm-up with model load

jijoongmoon:
Two commits that take the ~170–190 ms `minja` chat-template renderer construction off the path to the
first token.

1. **Warm the renderer off-thread, race-safe.** `ChatTemplate::Load()` starts a side thread that
   builds the renderer; the first `apply()` joins it. On a caller that does
   create → initialize → load_weight → repack between `Load()` and `apply()`, the overlap is total.
   The join is guarded (the naive `if (warm_thread.joinable()) warm_thread.join();` outside a lock lets
   two concurrent `apply()` callers both see `joinable()` and both join — the fix for that is folded
   into this commit, not left as a follow-up).

2. **Make the CLI actually overlap.** On the app's own CLI path the first commit bought nothing:
   `Load()` and `apply()` were nine lines apart, before the model object even existed, so the warm
   thread was joined almost immediately and the construction was still paid serially — and *ahead* of
   `initialize()`, squarely on the path to the first token. This splits the decision from the render:
   *which* input to use is still resolved in place, only the render moves to just after
   `repack_weight()`, handing the warm thread the same window a create/initialize/load/repack caller
   already gets.

Error timing is deliberately preserved: the deferred branch is taken only when a chat input exists
**and** a template loaded, and both are decided at the original point — as are the `sample_input`
lookups that throw when the key is missing. A malformed config still fails before model load.

**New in this rung:** the two commits above.

**Not included:** no change to template semantics, rendering output, or the tokenizer path.

**Validated:** CPU build clean; the double-join race is closed by construction (single guarded join
under the lock). Generated text is unaffected — this moves *when* the renderer is built, not what it
produces.

---

### DIFF
--- a/Applications/CausalLM/chat_template.cpp
+++ b/Applications/CausalLM/chat_template.cpp
@@ -18,6 +18,7 @@
 #include <cstdlib>
 #include <fstream>
 #include <mutex>
+#include <optional>
 #include <sstream>
 #include <stdexcept>
 #include <thread>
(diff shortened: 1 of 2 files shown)


jijoongmoon:
Consolidated into #4212.

PR #4186, opened 2026-07-30
[causallm] Embedding sidecar LUT: mmap the payload read-only (lazy page-in)

jijoongmoon:
Makes the embedding **sidecar LUT** payload lazily paged instead of heap-resident, and gives the loader
an actionable error plus GGML row-payload support. The sidecar infrastructure itself (the
`quantized_lut_path` manifest keys, the config wiring and the unit test) is already on `main`; this is
the *consumption* half.

1. **mmap the payload (the memory lever).** The sidecar table (raw UINT16, or the `ufixed8`/`sfixed4`
   JSON-manifest payload) was read whole into a heap vector at finalize, so a multi-hundred-MB embedding
   table became resident before the first token. Now attached with
   `::mmap(PROT_READ, MAP_PRIVATE)` + `MADV_RANDOM`. `QuantLut::bytes` is intentionally left empty on
   the successful mmap path; `QuantLut::payload_size()` is the documented size accessor regardless of
   which backing store is active.
2. **Actionable unsupported-datatype error.** The loader only accepted `ufixed8`/`sfixed4` manifests
   (plus raw UINT16 by non-`.json` path); anything else hit a terse throw. It now spells out what is
   supported and the two ways forward. Same dtypes rejected as before — only the message changes.
3. **GGML `q4_0`/`q6_k` row payloads (see "separable" below).** Adds a manifest datatype whose payload
   is the byte-identical Q4_0/Q6_K row table an in-bin embedding weight would hold. Decode **reuses the
   existing** `incremental_forwarding` dequant loop — `dequantize_row_q{4_0,6_K}`, the FP32-temp cast
   for non-FP32 activations and the scale multiply are untouched; only the row base pointer is chosen
   between the weight tensor and the mmap'd sidecar, and the row stride is hoisted (same value the old
   per-branch `num_blocks_per_row` math produced, since the weight width is `out_dim`). Sidecar output
   is therefore bit-exact vs the monolithic package by construction. Because `MADV_RANDOM` disables
   readahead, a ~1K-token prefill would otherwise pay ~1K serialized major faults, so the exact rows of
   each incoming batch are batch-`MADV_WILLNEED`'d up front (POSIX only; the Windows mapping pages on
   demand via `MapViewOfFile`).
4. **Unit-test fix.** Two pre-existing assertions read `bytes.size()` directly and started failing once
   mmap began succeeding on a real filesystem; they now read `payload_size()`.

**Separable commit, stated up front:** commit 3 (`[OPTIONAL]`, ~123 lines) can be dropped to keep GGML
LUT support deferred — commit 2's actionable error then stands on its own. It is included because it is
pure reuse of a dequant this file already consumes (zero new dependencies) and because every sidecar
package currently in use declares a GGML LUT manifest, so without it sidecar consumption has no
package to exercise it. Owner call at review time.

**New in this rung:** the five commits above (mmap, actionable error, GGML routing, test fix, error-text
restore).

**Not included:** no change to the monolithic (in-bin) embedding path for packages without a
`quantized_lut_path` — byte-identical by construction (no `quant_lut`, `ggml_lut` false, same weight
rows through the same branches; the only removed work is constructing an unused `Tensor` view).

**Base note:** this rung targets `main` directly. Its consuming files are
`Applications/CausalLM/layers/embedding_layer.{cpp,h}` plus the sidecar unit test, and the sidecar
infrastructure and that test are already upstream — so no part of the app chain of this series is
needed.

**Expected effect, pending re-measurement on this tree:** on a prior device session the lazy page-in
measured RSS −1.27 GB, init −41 %, load residual −44 %, and the `MADV_WILLNEED` batch prefetch removed
a ~100–160 ms serialized-fault stall on NVMe. Cited as expected values; the golden requirement
(sidecar output bit-exact vs monolithic) held on those runs and is re-stated here as the gate.

---

### DIFF
--- a/Applications/CausalLM/layers/embedding_layer.cpp
+++ b/Applications/CausalLM/layers/embedding_layer.cpp
@@ -276,6 +276,49 @@ std::shared_ptr<QuantLut> loadSfixed4Manifest(const std::string &manifest_path,
   return lut;
 }
 
+/**
+ * @brief GGML row-block sidecar: the payload is the byte-identical Q4_0/Q6_K
+ *        row table an in-bin embedding weight would hold, so decode reuses
+ *        dequantize_row_q{4_0,6_K} and the outputs match the in-bin path
+ *        bit-exactly. Manifest:
+ *          {"datatype": "q4_0"|"q6_k", "size": <out_dim>,
+ *           "rows": <in_dim, optional>, "lut-path": "<payload>"}
+ */
+std::shared_ptr<QuantLut> loadGgmlManifest(const std::string &manifest_path,
+                                           const nlohmann::json &json,
+                                           nntrainer::TensorDim::DataType dt) {
+  const auto lut_path = requireJsonStringField(json, "lut-path", manifest_path);
+
+  auto lut = std::make_shared<QuantLut>();
+  lut->out_dim = requireJsonSizeField(json, "size", manifest_path);
+  lut->ggml_dtype = dt;
+
+  const size_t block = (dt == nntrainer::TensorDim::DataType::Q6_K) ? 256 : 32;
+  const size_t block_bytes =
+    (dt == nntrainer::TensorDim::DataType::Q6_K) ? 210 : 18;
+  NNTR_THROW_IF(lut->out_dim % block != 0, std::invalid_argument)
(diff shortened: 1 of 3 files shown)


jijoongmoon:
Consolidated into #4212.

PR #4187, opened 2026-07-30
[layers] Backend-neutral SigmoidGLU / SigmoidAdd layers + their OpenCL whole-ops

jijoongmoon:
Promotes the fused sigmoid gating pair as model-neutral core. The dispatch shim already declares
`ComputeOps::sigmoid_glu` / `::sigmoid_add` and ships a `CpuComputeOps` implementation, but the tree
carries **no Layer that reaches them and no CL override** — two unreachable virtuals. This arms them.

- `nntrainer/layers/llm/sigmoid_{glu,add}_layer.{h,cpp}`: one thin backend-neutral Layer each (type
  `"sigmoid_glu"` / `"sigmoid_add"`), owning structure and shape only and delegating the kernel through
  `in.getOps()->sigmoid_glu / sigmoid_add` — the `SwiGLULayer` shape exactly. `sigmoid_glu` is the
  sigmoid-gated attention output gate (`out = sigmoid(gate) * up`); `sigmoid_add` is the sigmoid-gated
  per-layer-embedding mix (`out = sigmoid(gate) + emb`). Both honour the `SkipPrefill` property the
  other GLU layers already take.
- `nntrainer/tensor/cl_operations/sigmoid_{glu,add}_cl_op.{h,cpp}` plus four `.cl` kernels (fp32 +
  fp16): the whole-op CL bodies in `swiglu_cl_op`'s shape. The fp16 variants upcast to fp32 for the
  sigmoid and combine, so wide gating intermediates do not overflow half.
- `ClComputeOps` overrides, so an `engine=gpu` graph containing these nodes dispatches instead of
  hitting the base `NI()` throw.
- Registration on cpu (`AppContext`) and gpu (`ClContext`), appended **last** in
  `add_default_object` with **explicit** `int_key`s 9001/9002. The auto `int_key` is
  `str_map.size() + 1` (a count), so a mid-list insertion shifts every later auto key and can collide
  with an explicit one; appending last with explicit keys is the documented collision discipline. When
  the collision-proof auto-key mechanism lands (separate PR in this series), these fold into it.

Conformance: one class, one type string, no per-backend Layer fork; the Layer body calls only the
`ComputeOps` virtual and the raw CL wrapper stays inside `cl_operations/` and out of any installed
header.

Depends-on: #4115

**New in this rung:** `[layers] Backend-neutral SigmoidGLU / SigmoidAdd layers + OpenCL whole-ops`.

**Two consumers, in two different lineages** — which is why this is its own PR rather than folded into
either of them: the long-context layer rung (#4198) edits both of these layers and carries a patch twin of
this commit, and the CUDA lane registers these two layer types on the cuda context. Landing this PR early
makes both inherited copies evaporate on rebase, so it is the cheapest merge in the series.

**The deliberately deferred backend, named so the gap reads as tracked rather than forgotten:** the
cuda-context factory for these two layers ships in **#4209**, alongside the matching `CudaComputeOps`
overrides — registered last with explicit `int_key`s 9001/9002, mirroring the CL registration here. Without
that pairing an `engine=cuda` graph stamping a fused gate aborts with `Key is not found for the object. Key:
sigmoid_glu`, so the two must not be separated; they are not separated.

**Not included:** no CUDA override or cuda registration (both in #4209), and no model uses these node types
in this PR — it closes the two unreachable virtuals the dispatch shim already shipped and becomes the
prerequisite for the long-context layer rung, which edits both layers.

**Validated:** CPU-only and fp16+OpenCL builds compile clean; with no graph emitting these node types
the default execution path is byte-identical by construction.

---

**Merge button: "Rebase and merge" — please do not squash.** This PR is a single commit today, but
the never-squash convention holds for the whole series: patch identity is what lets the inherited copies of
this commit in #4198 and the CUDA lane drop out cleanly when this one lands.

Depends-on: #4115


jijoongmoon:
Consolidated into #4201.

PR #4188, opened 2026-07-30
[causallm] Add the Gemma2 model class (Gemma2ForCausalLM)

jijoongmoon:
Adds `Gemma2ForCausalLM` — a public Gemma2 model class built on the model registry — so the
Gemma2 packagings become loadable and testable in-tree.

Gemma2's shipped configs all carry `"attn_logit_softcapping": 50.0`, which makes the soft-cap parse
fix its hard prerequisite: on a tree where that parse sits in an unreachable derived override, the new
model would silently run with its attention soft-cap **off**. That fix is a separate main-direct PR in
this series (see `Depends-on`), and it also ships the `causallm_test_utils` helpers the Gemma2 unit test
consumes.

Depends-on: #4115, #4182

**New in this rung:** `[causallm] Add the Gemma2 model class (Gemma2ForCausalLM)`. The
soft-cap-consolidation commit also appears in this branch's diff — it is the same commit under review
in the main-direct PR named above and lands once, whichever merges first.

**Not included:** engine stamping for the new graph (the node-level `engine=` property that routes it
to the GPU FC/norm variants) — that is the next rung; this PR is CPU-correct on its own.

**Validated:** CPU build clean; `unittest_causallm_gemma2` passes; the model's generated text
reproduced a pre-existing reference continuation byte-for-byte once the BOS and soft-cap prerequisites
were in place (the BOS fix is the graph-correctness PR of this series).

---

Depends-on: #4115, #4182


jijoongmoon:
**Superseded by #4193**, which contains these commits and reviews them together.

Both commits of this branch are present in #4193's diff as **byte-identical patches** (verified with
`git patch-id --stable`: `92fdfc686…` for `[causallm] Add the Gemma2 model class (Gemma2ForCausalLM)`,
`fc9d1d699…` for the soft-cap parse consolidation), so no content is lost. #4193's body now carries this
rung's documentation in full: the Gemma2 model class and the model-registry shape it uses, why the
`attn_logit_softcapping` parse consolidation (#4182) is a hard prerequisite rather than a nicety, and the
`unittest_causallm_gemma2` gate.

Why consolidate: every branch of this series is a stack, and a cross-repo PR must use `base: main`, so
each rung's diff replays the same inherited ~20k lines. Folding the three engine-selection rungs into one
PR means a reviewer confronts that diff once instead of three times — and the `engine=` stamping commit,
which is the *only* consumer of the OpenCL layer-variant chain, is then reviewed together with the commits
that give its nodes a GPU home.

Closing in favour of #4193.

PR #4189, opened 2026-07-30
[CausalLM] MHACore GPU prefill dispatch (OpenCL flash/two-conv attention)

jijoongmoon:
Restructures the `MHACore` layer to dispatch **prefill** attention onto the OpenCL flash / two-conv /
OHWI-image kernels the earlier GPU rungs of this series ship, with the host x86/NEON per-row dot and
GEMM-flash fallbacks preserved for the non-GPU path. The prefill dispatch is env-gated
(`NNTR_MHA_GPU`, `NNTR_MHA_GPU_IMG`, `NNTR_MIN_PREFILL`, `NNTR_V8C_BUF`); `softmax_triangle`,
`apply_rotary_emb_tensor_v2` (with the process-wide flat RoPE LUT for the GPU concat-RoPE path) and the
`gemm_attention` path come along. Every kernel it calls resolves against the base of this rung.

**Prerequisite fix — this is the point of the rung.** `opencl + fp16 + transformer` did **not build**
at the earlier app rungs: `mha_core.cpp` and `deberta_attention_layer.cpp` call the all-FP16
`nntrainer::compute_kcaches` / `compute_fp16vcache_transposed` / `compute_rotary_emb_value` (and the
`softmax_row[_inplace]<_FP16>` instantiations), which the cpu_backend only builds for ARM/NEON. This
`mha_core` carries its own x86-only FP16 host implementations of those (scalar, FP32-accumulate), and
the deberta hunk drops `ENABLE_FP16` on x86 (that model is ARM-encoder-only), so the fp16-ON x86 app
build is green from here on.

Depends-on: #4150

**New in this rung:** `[CausalLM] MHACore GPU prefill dispatch (OpenCL flash/two-conv attention)` plus
a comment-only cleanup commit.

**Explicitly NOT included** — each is a later rung, stripped cleanly, and the file compiles with it
absent:
- CUDA (`#if ENABLE_CUDA` regions);
- int8 KV cache (the internal int8 cache with per-(token, head) fp16 scales, its quant/dequant
  helpers, int8 read/write, the CL int8 kernel dispatch, and the `cache_key_scale` /
  `cache_value_scale` inputs);
- the KV window ring (`kv_ring_cap` / `cacheRow`; pre-ring linear cache semantics are kept here);
- GPU **decode** (flash-decode, GPU-RoPE-decode and the per-layer decode properties — this rung is
  prefill only);
- the `ROPE_EXTEND` opt-in.

This rung supports only the external FP16 5-input cache and the plain 3/4-input internal FP16 cache.

**Validated:** fp16 + OpenCL x86 build green (it was red before this rung — that is the build gate);
GPU dispatch is env-gated, so the default path is unchanged.

---

Depends-on: #4150


jijoongmoon:
**Superseded by #4192**, which contains these commits and reviews them together.

`upstream-pr/trackA4-mha-core-gpu-prefill` is a **git ancestor** of #4192's head — verified with
`git rev-list <this head> ^<#4192 head>` = **0 commits**, i.e. every commit of this PR is literally in that
PR's history, same SHAs. Nothing is lost. #4192's body now carries this rung's documentation in full: the
prefill dispatch decision table (which shape / dtype / residency combination selects flash vs OHWI image vs
two-conv vs host, with every fallback condition named), the `opencl + fp16 + transformer` build gate that is
the point of the rung, the explicit not-included list (CUDA regions, int8 KV, KV ring, GPU decode,
`ROPE_EXTEND`), the `env_compat.h` packaging justification, and the honest 28-raw-`*_cl`-call-site
deviation with its named paydown shape.

Why consolidate: #4192's trace hooks sit in the `forwarding()` / `incremental_forwarding()` exits that this
dispatch creates, so it cannot merge before this rung — bundling delays nothing. And "host fallbacks are
NAMED, not hidden" is the acceptance criterion for a 2,637-line dispatch, so the 194 lines of default-off
diagnostics belong in the same review rather than in a second full pass over the same file.

Closing in favour of #4192.

PR #4190, opened 2026-07-30
[causallm] Stamp engine= on the layer graph so the GPU FC/norm variants get used

jijoongmoon:
Stamps the `engine=` property node-for-node on the layer graph so that the GPU FC / norm layer
variants registered by the earlier rungs of this series are actually selected. Without it a graph built
on an `engine=gpu` context still instantiates the host layer classes, and every GPU layer variant in
the tree is dead code reachable only by hand-editing a model.

The stamp is a graph-construction property, not a backend test inside forwarding logic: the resolved
engine comes from the context, the layer classes stay backend-neutral, and no `#if` fence or
`getName() == "gpu"` predicate is introduced.

Depends-on: #4116, #4188, #4182

**New in this rung:** `[causallm] Stamp engine= on the layer graph so the GPU FC/norm variants get
used`.

The other two commits in this branch's diff (the soft-cap consolidation and the Gemma2 model class) are
the same commits under review in their own PRs named above; they land once.

**Not included:** the Gemma4-specific stamping and its host-node removals (a later rung — Gemma4 still
has host-only nodes at this point and stamping it here would be premature), and the `lm_head` stamp.

**Validated:** CPU and fp16+OpenCL builds clean; 4-model short-prompt text byte-identical to the
pre-change baseline on the CPU engine (the stamp is inert without a GPU context).

---

Depends-on: #4116, #4188, #4182


jijoongmoon:
**Superseded by #4193**, which contains these commits and reviews them together.

All three commits of this branch are present in #4193's diff as **byte-identical patches** (verified with
`git patch-id --stable`: `6906877de…` for `[causallm] Stamp engine= on the layer graph so the GPU FC/norm
variants get used`, plus `92fdfc686…` Gemma2 and `fc9d1d699…` soft-cap), so no content is lost. #4193's
body now carries this rung's documentation in full: the `residencyEngine()` authority, the
`causallm_engine()` contract, the stamped-site inventory, the per-site reasons for every node deliberately
left on the host, and the measured claim that belongs to this commit (~2.3 / 1.8 TPS host-FC-bound on
qwen3 with an OpenCL build, identical to `NNTR_ENGINE=cpu`, because zero `withKey("engine", ...)` call
sites existed in `Applications/`).

Why consolidate: this stamp is the only consumer of the whole OpenCL layer-variant chain, and a node
cannot be stamped before it has a GPU home — which is exactly what the Gemma4 prerequisite commits in
#4193 supply. Reviewed apart, this rung's stamps land on nodes whose GPU home arrives in another PR, and
both PRs replay the same inherited `base: main` diff.

Closing in favour of #4193.

PR #4191, opened 2026-07-30
[build/android] The Android GPU build was silently a CPU build: make it reachable (script/CLBlast/jni source) and export ENABLE_OPENCL, failing loudly without it

jijoongmoon:
**Task-level PR: "the Android GPU build was silently a CPU build."** Four independent build defects, one
failure mode. Absorbs #4180 (now closed and pointed here).

`Applications/CausalLM` on Android compiled with **no `-DENABLE_OPENCL`**, so every
`#if defined(ENABLE_OPENCL)` block in the app (`mha_core.cpp`, `rms_norm_gpu.cpp`,
`per_layer_slice_gpu.cpp`, `llm_util.hpp`) was dropped and `causallm_engine()` returned `"cpu"`. The app
then linked a GPU-capable `libnntrainer.so` and ran the host fallbacks — roughly **10x slower, with no
error and no log line saying GPU was not used**. This exact class has bitten before: a prior tree was
missing the same define and a model measured **191 TPS instead of 2400**.

And underneath that, the GPU build was not even *reachable*: the documented entry point could not
forward the meson option, the CLBlast-free configuration crashed at configure time, and a model source
was missing from the hand-maintained jni list. No CI job can see any of this — no CI job builds the
CausalLM app or runs `ndk-build`.

**Merge button: "Rebase and merge" — never squash.** The build-export hunks and the runtime throw live
in one commit (disclosed below); squashing would bury that distinction entirely.

**Base legality:** #4180's three commits are ladder-free and cherry-picked **clean** onto this branch.
The reverse was tested and rejected: moving this branch's two commits down onto main-direct #4180
conflicts in `llm_util.hpp`, because the `engine=` resolution those commits guard is introduced by the
app lineage this branch already sits on — rebasing across bases is not something this consolidation
does.

Depends-on: #4193

---

## 1. `-Denable-clblast=false` cannot configure on `platform=android`

`meson -Dplatform=android -Denable-opencl=true -Denable-clblast=false` dies at configure time with
`jni/meson.build:77: ERROR: Unknown variable "clblast_root"`. The CLBlast template substitution was
gated on `clblast_dep.found()`, but on the `enable-clblast=false` path `clblast_dep` is a bare
`declare_dependency()` — whose `.found()` is **true**. The guard never fires and the body reads a
variable only assigned in the `true` branch. The same guard also owned `MESON_CL_ROOT`, so even without
the crash the staged `libOpenCL.so` / CL headers would drop out of the generated `Android.mk`.

Fix: gate on `get_option('enable-clblast')` — the same authority that decides whether `clblast_root`
gets assigned — hoist `MESON_CL_ROOT` out of the CLBlast branch, and export a `MESON_HAS_CLBLAST` marker
so `Android.mk.in` can drop the CLBlast static-library module instead of compiling ~100 CLBlast sources
from an empty root prefix. The `clblast_dep.found()` idiom was vacuously correct before the option
existed; adding the option turned a vacuous test into a live one on a path nothing exercises.

## 2. `build_android.sh` cannot forward meson options — the GPU build was unreachable

It accepted only `--cache` and `exit 1`'d on anything else, and invoked `tools/package_android.sh` with
no arguments — and `enable-opencl` defaults to false. **So the documented way to build the Android app
could only ever produce a CPU-only nntrainer.** `package_android.sh` already filters `-D*` /
`--arm-arch=*` out of its argv and forwards them to meson, so the only missing piece was passing them
through: `./build_android.sh -Denable-opencl=true -Denable-clblast=false` now works. Also warns when
meson args are combined with `--cache`, since that path skips the reconfigure and the cached options
win.

## 3. A missing jni source ⇒ undefined vtable

`jni/Android.mk` never listed `models/bert/multilingual_tinybert_16mb.cpp`, although meson does.
`MultilingualTinyBert` declares all virtuals inline except `encode()`, which is an implicit override and
therefore the class's **key function** — so the vtable is emitted only in that TU, and the two Android
modules that instantiate the class (`causallm_core` via the model registry, `nntr_quantize` via
`quantize.cpp`) link with an undefined `vtable for causallm::MultilingualTinyBert`. The jni source lists
are hand-maintained and completely independent of the meson lists, so adding a model to the app fails
only on a real NDK build.

## 4. Export `ENABLE_OPENCL` to prebuilt consumers, and fail loudly without it

How the flag is supposed to reach the app: meson `enable-opencl` → `extra_defines
'-DENABLE_OPENCL=1'` → for the library, `jni/meson.build` `MESON_CFLAGS`; for prebuilt consumers, the
`NNTRAINER_ABI_DEFINES` filter → `android_build_result/Android.mk`'s `NNTRAINER_EXPORT_CFLAGS` →
`LOCAL_EXPORT_CFLAGS` on the prebuilt modules → the app's `LOCAL_CFLAGS`. The last hop already works
(the `-march` and FP16 defines travel it today); the filter simply did not select `ENABLE_OPENCL`, so it
stopped at the library.

Fix: widen the existing export filter rather than hardcoding the flag in the app makefile — which is
what the export mechanism exists to prevent, since a hardcoded value can disagree with the library and
would then be an ABI break instead of a missing feature. Plus a loud failure when a graph asks for
`engine=gpu` (or `engine=cuda`) in a build that has no such backend, so the silent 10x fallback becomes
an error.

On x86 the question never arises: meson applies `extra_defines` project-wide via
`add_project_arguments`, so `Applications/` inherits the define automatically. Only the
prebuilt/`ndk-build` path needs the explicit export, and no CI job builds an `ndk-build` app against the
generated prebuilt `Android.mk`.

**Residual disclosed honestly: the export commit mixes two kinds of change.** A reviewer may want to
accept them separately, so here is which hunk is which:

- *mechanical build-export widening* — `jni/meson.build` (the `NNTRAINER_ABI_DEFINES` filter),
  `jni/Android-prebuilt.mk.in`, and `Applications/CausalLM/jni/Android.mk`;
- *runtime behaviour change* — `Applications/CausalLM/llm_util.hpp`, which throws `std::invalid_argument`
  when a graph asks for `engine=gpu` in a build without OpenCL (and the matching `engine=cuda` case).

Only a commit split would fully separate them, and that is out of scope for this consolidation pass; the
split is offered if a reviewer asks.

---

## `enable-clblast`: one declaration, two halves, now in one PR

This is the main reason the two PRs belonged together. `enable-clblast` was declared **twice**, once in
each PR, byte-identically and at the same position immediately after `option('enable-opencl', ...)`,
because each branch read the option and Meson resolves option names at configure time — a tree that
reads an undeclared option aborts before any target is built. Consolidated, there is exactly one
declaration, and the two halves of the knob finally sit together:

| half | what it does | came from |
|---|---|---|
| Android | the `MESON_HAS_CLBLAST` marker that drops the CLBlast static-library module from the generated `Android.mk`; gating the `subprojects/CLBlast/include` entry on the compile line | #4180 |
| host | skipping the CLBlast cmake subproject entirely (this is what makes `-Denable-clblast=false` actually avoid the multi-minute CLBlast compile) and `-DENABLE_CLBLAST=1`; plus the `#ifdef ENABLE_CLBLAST` guard in `nntrainer/tensor/cl_operations/clblast_interface.cpp` that lets that TU compile without the CLBlast headers | this branch |

A CLBlast-free `ndk-build` wants both. The comment above the gate in `nntrainer/meson.build` records the
dependency, so a reader of the build file does not have to infer it.

## Overlapping files, which is why this is one review

`Applications/CausalLM/jni/Android.mk`, `jni/meson.build`, `nntrainer/meson.build`,
`meson_options.txt` and `llm_util.hpp` are all touched by both halves.

---

**Not verified here:** NDK build + on-device deploy. The export gap was traced hop by hop through the
generated `Android.mk`, and the app-side consumers are the four TUs named above; the three build defects
are configure-time / link-time and were each reproduced from the error text quoted above.

**Interaction note:** the ndk include-list surface here overlaps the installed-header correction in
#4103 — the two touch the same generated makefile.

---

fork branch as its base, so this PR targets `main` and its diff therefore also contains the parent rungs
named in `Depends-on:`. Please review the commits described above; the remainder are under review in
their own PRs and will disappear from this diff as the parents land.

**Merge order:** the `attn_logit_softcapping` and `lmhead_untie` rungs inherited from this branch's base
lineage also land in the main-direct inference-correctness PR (#4194). Land **#4194 first** so those
copies arrive as no-ops under rebase-and-merge rather than as conflicts.

Depends-on: #4193


jijoongmoon:
Superseded, **in part**, by the re-cut series against current main: #4307. Closing to keep the queue
reviewable. The branch is preserved on the fork, and this comment states plainly what did and did not
travel, so the residue is not lost.

**What #4307 carries — the export half, and only that.** `jni/meson.build` selects which of
`extra_defines` the generated `Android-prebuilt.mk` exports to prebuilt consumers, and the filter
matched only the two FP16 defines. A library built with `-Denable-opencl=true` therefore handed
consumers cflags with no trace of OpenCL, while `engine.h` — an installed header — pulls
`cl_context.h` in under `#if defined(ENABLE_OPENCL)`. An ndk app then declared a different `Engine`
than it linked and silently dropped its own OpenCL blocks. #4307 widens the filter to
`ENABLE_OPENCL` and `CL_TARGET_OPENCL_VERSION`, and states the rule the filter follows: a define
belongs there when it changes what an *installed* header declares.

Two review points from this PR's own thread were applied while re-cutting:

- `CL_TARGET_OPENCL_VERSION` is matched **by prefix**, not exact-matched against `=200`. Pinning the
  value would silently drop the define the day the build option changes, which is the same class of
  bug as never exporting it.
- `USE_HALF_WRAPPER` was added to the same filter. It selects the `_FP16` type in `tensor_dim.h`, so
  by the filter's own criterion it is as ABI-affecting as `ENABLE_FP16`, even though no android
  toolchain selects it today.

The other ask on this thread — a third `cl_options` matrix cell so the regression is gated — is
**already satisfied on `main`**: `.github/workflows/android.yml` runs
`cl_options: [ "-Denable-opencl=false", "-Denable-opencl=true" ]`.

**What #4307 does NOT carry.** The rest of this PR — the `NNTR_ENGINE` fail-loudly handling and the
residency-policy work — is not in it and is not superseded by it. That material belongs to the
context/engine line (#4306) and the residency-plane PR that follows it, and should be reviewed there.
Reopen this PR if that line does not pick it up.

Verified with NDK r27 (27.2.12479018), `arm64-v8a`, traced end to end: the generated
`NNTRAINER_EXPORT_CFLAGS` is
`-march=armv8.2-a+fp16+dotprod+i8mm -DENABLE_FP16=1 -DUSE__FP16=1 -DENABLE_OPENCL=1 -DCL_TARGET_OPENCL_VERSION=200`,
and `ndk-build -n -B` shows `-DENABLE_OPENCL=1` and `-DCL_TARGET_OPENCL_VERSION=200` on the app's own
`llm_util.cpp` compile line. On `main` that line has neither.

PR #4192, opened 2026-07-30
[CausalLM] MHACore GPU prefill dispatch (OpenCL flash / two-conv attention) + name every host bounce it can take

jijoongmoon:
Restructures `MHACore` to dispatch **prefill** attention onto the OpenCL flash / two-conv / OHWI-image
kernels the earlier GPU rungs of this series ship — and, in the same PR, makes every host bounce that
dispatch can take **say so**. The two halves are one unit: "host fallbacks are NAMED, not hidden" is the
acceptance criterion for a 2,637-line dispatch, and the trace is what turns a host bounce from an argument
reconstructed by reading ~600 lines of gates into a 30-second measurement. The capture hooks also sit in
the `forwarding()` / `incremental_forwarding()` exits that this dispatch creates, so they could not have
landed earlier anyway.

## 1. The prefill dispatch (and the build gate that is the point of the rung)

The prefill dispatch is env-gated (`NNTR_MHA_GPU`, `NNTR_MHA_GPU_IMG`, `NNTR_MIN_PREFILL`, `NNTR_V8C_BUF`);
`softmax_triangle`, `apply_rotary_emb_tensor_v2` (with the process-wide flat RoPE LUT for the GPU
concat-RoPE path) and the `gemm_attention` path come along. Every kernel it calls resolves against the base
of this rung.

**Prerequisite fix.** `opencl + fp16 + transformer` did **not build** at the earlier app rungs:
`mha_core.cpp` and `deberta_attention_layer.cpp` call the all-FP16 `nntrainer::compute_kcaches` /
`compute_fp16vcache_transposed` / `compute_rotary_emb_value` (and the `softmax_row[_inplace]<_FP16>`
instantiations), which the cpu_backend only builds for ARM/NEON. This `mha_core` carries its own x86-only
FP16 host implementations of those (scalar, FP32-accumulate), and the deberta hunk drops `ENABLE_FP16` on
x86 (that model is ARM-encoder-only), so the fp16-ON x86 app build is green from here on. **That build
gate is the review gate for this rung.**

### Dispatch decision table (prefill)

| # | condition | route |
|---|---|---|
| 0 | `use_gemm_attention == false` | host per-row dot (`compute_kcaches` + `softmax_row`) — the property is unstamped until the flash-plumbing rung, so this is today's default |
| 1 | `use_gemm_attention && step_size < FLASH_MIN_PREFILL` (`min_prefill_thr(head_dim)`, env `NNTR_MIN_PREFILL`) and not `NNTR_CUDA_ATTN` | host `gemm_attention` — blocking + softmax bookkeeping does not pay below the threshold |
| 2 | above threshold, `NNTR_MHA_GPU=1`, **Q fp16 and out fp16**, `head_dim > 0`, `num_heads_KV > 0`, `num_heads_Q % num_heads_KV == 0` | GPU candidate; otherwise host `gemm_attention` |
| 3 | GPU candidate **and** `svm_ok` = all four of Q / cached-K / cached-V / out have `MemoryData` **and** `isSVM()` | device path; if `svm_ok` is false the non-SVM wrapper (host-tensor cache) is used, i.e. a host bounce |
| 4 | device path, `NNTR_KV_IMG_ATTN=1`, `head_dim % 8 == 0`, and the layer is **not past its sliding window** | Adreno OHWI **image** attention (`qk_matmul_f16_ohwi_img`) — preempts flash on success |
| 5 | device path, otherwise | **flash** kernels (SVM buffer), which take `local_window`; the OHWI two-conv prefill path is selected instead when `NNTR_KV_OHWI_GPU_FORCE`/`_ohwi_gpu_on` holds |
| 6 | any kernel returns false (shape, alignment, scratch, mirror init) | falls back to the host path for that call |

Row 4's window exclusion is a defect fixed at the source and recorded at the call site:
`qk_matmul_f16_ohwi_img` has only the causal upper-bound mask and **no window lower bound**, so a
sliding-window layer past its window would silently compute full causal attention over evicted keys
(measured: a W=512 model's 999-token Adreno prefill degenerates into word salad, severity growing with
`cache_to - W`). Those calls are routed to the flash kernels, which do take `local_window`.

## 2. Name every FC that leaves the GPU path (`NNTR_FC_DIVERT_TRACE`)

`ClComputeOps::fc` had **no** way to report that an FC diverted to a host path: `dotCl_v8c` returns a bare
bool and the tree had no trace hook. A host bounce is not a harmless slow path — it reads the input and
writes the output through host pointers, which is exactly why the fallback has to bracket itself with
`clmem_lower_cl` / `clmem_raise_cl`.

- `dotCl_v8c`: all 16 `return false` sites now go through a `V8C_REJECT("<why>")` macro recording a static
  reason string in a thread-local slot (thread-local because the eager weight prebuild runs on the loader
  pool). No behaviour change — same control flow, same returns. As a side effect the reject taxonomy
  becomes readable at a glance: env off, weight dtype, K mismatch, `N%8` / `K%32`, activation dtype,
  weight-backing build, `huge_n`, the five scratch/copy failures, `y_fp16`, the host-bounce readback, and
  the catch-all. `v8c_last_reject_reason()` exposes it.
- `ClComputeOps::fc` prints **one** line per divert under `NNTR_FC_DIVERT_TRACE=1` (default off,
  value-checked so `=0` disables), on a stamped graph:

```
[fc-divert] layer0_ffn_gate:out0 <- layer0_ffn_gate:weight
            in[1x1x413x1024 FP16] w[1024x3072 QS4CX] out[FP16]
            out_clmem=1 reason=huge_n weight and not the M==1 lm-head GEMV case
```

i.e. the tensor, the shapes, the dtypes, the output residency class and the condition — enough to map the
event onto the gate that caused it without a rebuild. Durable value beyond this series: the one
configuration where a host bounce is *expected* (an untied QS4CX lm_head at M>1, `huge_n`) and the ones
where it is a *regression* (any per-layer FC) become distinguishable in a log, on any device — including
the Adreno and Windows lanes where attaching a debugger is impractical.

**Every diagnostic in this PR is default-off and value-checked** (`=0` really disables):
`NNTR_FC_DIVERT_TRACE`, `NNTR_CLMEM_PROBE` (+ `NNTR_CLMEM_PROBE_DRAIN`), `NNTR_CLMEM_DUALOUT`,
`NNTR_CLMEM_OUTCHECK`, `NNTR_CLMEM_OUTBAR`. Two differential GPU-path probes are restored with their
enabling guard, extended to cover `attention_out` and `mha_core`; that guard is what made
`layer0_wq:out0:y`, `layer0_ffn_gate/up:out0:y`, `layer0_attention_out:out0:y` and
`layer0_attention:mha_out` real comparisons for the first time — and that capture is what exposed the
positional gate/up load-order defect fixed in #4194.

## Disclosures

**Known architecture deviation, disclosed rather than hidden.** `mha_core.cpp` still contains **28 raw
`nntrainer::*_cl(...)` dispatch call sites** against the rule that an app layer should reach the backend
through one private helper called once from `forwarding()`. It is tracked in the GPU contribution
checklist. The paydown shape is already written and reviewable: #4195 converts the reverse-RMSNorm raw call
into a `ComputeOps` whole-op virtual with named rejects, and an **attention whole-op virtual** is the
declared follow-up for this file. Nothing here regresses the rule; the count is the honest starting point.

**Packaging additions justified.** `debian/nntrainer-dev.install`, `packaging/nntrainer.spec` and
`nntrainer/utils/meson.build` each gain exactly one entry: `env_compat.h`. `mha_core.cpp` is an app TU
compiled against the *installed* headers and it includes `<env_compat.h>` for `nntr_env_on()` — the
value-checking env helper that makes `=0` mean off (a presence check cannot be turned off). Without the
install entry the app build fails on a packaged nntrainer.

Depends-on: #4150

**New in this rung:** 6 commits — the MHACore GPU prefill dispatch, a comment-only cleanup, the restored
differential probes, the probe-guard extension covering `attention_out` / `mha_core`, the divert trace, and
a `-Wformat` fix in the trace (`%u` → `%zu` for `size_t` dims, which `-Werror=format` rejects at compile
time in every `-Denable-opencl=true` build).

**Supersedes #4189,** closed in favour of this PR: `trackA4-mha-core-gpu-prefill` is a **git ancestor** of
this branch (`git rev-list #4189-head ^this-head` = 0 commits), so its entire content is contained here and
its documentation is folded into the sections above.

**Explicitly NOT included** — each is a later rung, stripped cleanly, and the file compiles with it absent:
CUDA (`#if ENABLE_CUDA` regions, #4205); the int8 KV cache (internal int8 cache with per-(token, head) fp16
scales, its quant/dequant helpers, int8 read/write, the CL int8 kernel dispatch, and the
`cache_key_scale` / `cache_value_scale` inputs); the KV window ring (`kv_ring_cap` / `cacheRow` — pre-ring
linear cache semantics are kept here, #4200); GPU **decode** (flash-decode, GPU-RoPE-decode and the
per-layer decode properties — this rung is prefill only); and the `ROPE_EXTEND` opt-in. This rung supports
only the external FP16 5-input cache and the plain 3/4-input internal FP16 cache.

**Validated:** fp16 + OpenCL x86 build green (it was red before this rung — that is the build gate);
fp16-off configuration also builds, since the tripwire is dtype-agnostic. GPU dispatch is env-gated, so the
default path is unchanged and generated text is byte-identical with all gates off. `NNTR_FC_DIVERT_TRACE=1`
emits **zero** lines on the tied qwen3 / gemma2 packages, which is what the static audit predicted (every
per-layer FC passes all gates) — a line naming a per-layer FC would falsify that audit, so the trace is
also the standing witness for it.

---

**Merge button: "Rebase and merge" — please do not squash.** The dispatch, the probes and the trace are
separate revert handles, and the `-Wformat` fix must stay separable from the trace it fixes.

Depends-on: #4150


jijoongmoon:
Consolidated into #4201.

PR #4193, opened 2026-07-30
[causallm] Resolve engine= for real: give every graph node a GPU home, then stamp gemma2 / gemma4 / qwen3 and the lm_head

jijoongmoon:
One story: **a graph node declares where it runs, and every node that cannot yet run there is NAMED.**

This rung consolidates what were three separate PRs (#4188 Gemma2 model class, #4190 `engine=` stamping,
and the Gemma4 prerequisites) into the single unit they form, because they cannot be reviewed apart: the
stamping commit is the *only* consumer of the whole OpenCL layer-variant chain (#4109-#4114, #4150) —
before it, **zero `withKey("engine", ...)` call sites existed anywhere in `Applications/`**, so that entire
chain was dead code — and a node cannot be stamped until it *has* a GPU home, which is what the Gemma4
prerequisite commits supply. Reviewed apart, the middle rungs are code that is rewritten before it merges.

## What the rung does, in dependency order

**1. Give every graph node a GPU home.**

- **Collapses Gemma4's `tanh_gelu` + `multiply` pairs into the fused `geglu` node** — the neutral GeGLU
  layer and its `ClComputeOps::geglu` override come from earlier rungs of this series; this is the
  consumer.
- **SVM-aware `ReshapedRMSNormLayer` + gpu-context registration** as ONE class on two contexts, so the
  per-head q/k norms stop being a host segment in the middle of a device graph.
- **The per-layer-embedding lookup table must carry `embedding_dtype`, not `fc_layer_dtype`** — a real
  dtype-plumbing defect: the table is an embedding, and reading it at the FC dtype mis-sizes it for any
  package where the two differ.
- **Declares `mha_core` engine-neutral** — `ResidencyPolicy` had no populator for it anywhere in the tree,
  so the residency planner could not see the attention node at all.
- Adds an `NNTR_VNORM_HOST` kill switch and an `NNTR_RMSN_TRACE` hook on `ReshapedRMSNormLayer` (default
  off) so the norm can be bisected to the host without a rebuild.

**2. Add the Gemma2 model class (`Gemma2ForCausalLM`)**, built on the model registry, so the Gemma2
packagings become loadable and testable in-tree. Gemma2's shipped configs all carry
`"attn_logit_softcapping": 50.0`, which makes the soft-cap parse consolidation its hard prerequisite: on a
tree where that parse sits in an unreachable derived override, the new model would silently run with its
attention soft-cap **off**. `unittest_causallm_gemma2` covers it (it consumes the `causallm_test_utils`
helpers that ship with the soft-cap PR).

**3. Stamp `engine=`.** The engine tag's authority is `Context::residencyEngine()` (see
`nntrainer/layers/layer_node.cpp` — one authority, no per-alias special case, no `getName() == "gpu"`
predicate and no `#if` fence in any layer body). The app side resolves it through `causallm_engine()`,
whose contract is: unset ⇒ `"gpu"` under `ENABLE_OPENCL`, `"cpu"` in a non-OpenCL build,
`NNTR_ENGINE=cpu` opts out, `NNTR_ENGINE=cuda` selects cuda. Stamped: 13 sites in the base builders, 4 in
qwen3, 14 in gemma2, then the Gemma4 graph once its host-only nodes are gone, and finally the `lm_head`.

Every stamped type was checked to have a real gpu-context factory, because `createLayer` on a type absent
from the resolved context **throws** `exception::not_supported` — an unregistered type is *not* a silent
cpu fallback: `fully_connected` → `FullyConnectedLayerCl`, `addition` → `AdditionLayer`
(`ClComputeOps::residual_op`), `swiglu` → `SwiGLULayer`, `geglu` → `GeGLULayer`, `rms_norm` →
`causallm::RMSNormLayerGPU`, `tie_word_embeddings` → `TieWordEmbedding`.

**Deliberately NOT stamped — a host bounce is a defect to be NAMED, and each one carries its reason at
the call site** (`Applications/CausalLM/models/transformer.cpp`):

| node | line | why it stays on the host |
|---|---|---|
| `embedding0` | :269 | `"embedding_layer"` has no gpu-context registration and `createLayer` on an unregistered type throws, so an untied model would fail at graph build. The lookup is a gather, not a GEMM. |
| `cache_k_l*` / `cache_v_l*` | :496 | `"input"` has no gpu-context registration, and their storage is host-owned by `KVCacheManager` and bound via `setExternalTensors`, so the graph allocator never backs them. |
| `mha_core` | :549 | cpu-context only; it selects its own GPU kernels internally per path rather than from the node's engine. |
| `reshaped_rms_norm` (qwen3 q/k-norm) | in the stamping commit's message | cpu-context only *at that commit*; the SVM-aware registration in this same rung is what closes it. |

## Two measured claims, each on its own commit

**`[causallm] Stamp engine= on the layer graph so the GPU FC/norm variants get used`** — before it every
model ran host-FC-bound: **~2.3 / 1.8 TPS on qwen3 with an OpenCL build, identical to `NNTR_ENGINE=cpu`**,
because the FCs went to the CPU `FullyConnectedLayer` instead of `FullyConnectedLayerCl` and its v8c GEMM.
Two further consequences it fixes: `neuralnet.cpp` only routes the tensor pool to the gpu context's SVM
allocator when the graph contains a GPU-engine node, so nothing was device-visible; and
`Transformer::repack_weight`'s `causallm_engine() != "cpu"` skip was already assuming a GPU graph that
never existed.

**`[causallm] Stamp engine= on the lm_head (the last node cannot stay on the host)`** — the head is the
LAST node and consumes `output_norm`'s activation. With the block chain stamped, that activation lives on
the gpu context's `cl_mem`/SVM plane, so a host head reads a stale host shadow. Measured on the untied
QS4CX-head gemma4 package, OpenCL engine:

| | output | prefill | decode |
|---|---|---|---|
| unstamped | `<pad>`-class garbage | 1.76 | **0.23** |
| stamped | "The capital of South Korea is **Seoul**." | 313.7 | **20.6** |

So this commit is what makes that model *work*, not merely what makes it fast. `"lm_head"` itself (reached
only when `tie_word_embeddings=false` **and** `nntr_config` has no `lmhead_untie`) has no gpu registration
and is excluded by the condition, keeping today's host path.

Depends-on: #4150, #4182, #4183

## New in this rung / inherited

**New in this rung:** 9 of the 11 commits in the diff — (1) the Gemma2 model class, (2) the layer-graph
`engine=` stamp, (3) the Gemma4-graph stamp, (4) the `lm_head` stamp, (5) the fused geglu collapse, (6) the
SVM-aware `ReshapedRMSNormLayer` + gpu-context registration, (7) the PLE `embedding_dtype` fix, (8) the
`mha_core` engine-neutral / `ResidencyPolicy` declaration, (9) the norm kill-switch + trace hook.

**Inherited, NOT claimed as new — and these are patch TWINS, not git ancestors:** `[causallm] Consolidate
attn_logit_softcapping parse into base setupParameters` and `[causallm] Honor nntr_config lmhead_untie` are
byte-identical patches of the commits under review as **#4182** and **#4183** (verified by
`git patch-id --stable`). They were cut on this branch independently rather than merged in, so they will
**vanish from this diff on rebase** once those two PRs land — please **merge #4182 and #4183 first**.

**Supersedes #4188 and #4190,** which are closed in favour of this PR: #4188's two commits and #4190's
three commits are patch-identical to commits contained here (`92fdfc686…` Gemma2, `fc9d1d699…` soft-cap,
`6906877de…` stamping), so nothing of either is lost and a reviewer confronts the inherited 20k-line
`base: main` diff once instead of three times.

**Not included:** the graph-correctness defect fixes (gate/up load order, BOS, staging zero — #4194) and
the SVM KV plane (#4196), both of which build on this rung; and no CUDA path.

**One expected merge interaction with #4194.** That PR is a `main`-anchored defect fix that reorders
`createMlp` to create `ffn_gate` before `ffn_up`; this PR's stamping commit adds `withKey("engine",
causallm_engine())` at those same two sites (the stamp lines had to be dropped from #4194 because
`causallm_engine()` does not exist on `main`). So whichever of the two lands second shows **one small
textual conflict in `createMlp`**, whose intended resolution is "**gate-first and stamped**" — the exact
combination every downstream branch of this series already carries, so no golden moves from resolving it
that way.

**Validated:** fp16 + OpenCL build clean; CPU build clean and `unittest_causallm_gemma2` passes.
4-model short-prompt text byte-identical to the pre-change baseline over 3 consecutive runs on the cpu
lane (the stamp is inert without a GPU context), and per-model on the GPU lane: gemma4 golden (garbage →
golden via the lm_head stamp), qwen3 golden with the FCs on the device, gemma2 golden once the soft-cap
and BOS prerequisites are in place. The assembled series measures within run-to-run noise of the
pre-series reference bars on Intel XMX and dp4a at a 1K prompt (gemma4 +0.8 % / −0.6 % prefill, qwen3
+3.3 % / +2.6 %, gemma2 −1.2 % / −1.1 %; decode within −0.9…−2.8 %).

---

**Merge button: "Rebase and merge" — please do not squash.** Each commit carries its own rationale, its
own measured claim where it has one, and its own revert handle; squashing also destroys the patch identity
that lets the inherited twin commits above drop out cleanly when #4182 and #4183 land.

Depends-on: #4150, #4182, #4183


jijoongmoon:
Consolidated into #4201.

PR #4194, opened 2026-07-30
[causallm] Inference correctness: FFN gate/up load order, prompt BOS, FP16 activation overflow, loader fail-loud, q4_0 shape gate, factory int_key, softcap, lmhead_untie

jijoongmoon:
**Task-level PR.** Eight previously separate rungs, consolidated into one review: every commit here is a
defect that makes an inference path **return a wrong answer or fall back silently instead of failing
loud**. One review question throughout: *does this path fail loud, or does it lie?* None of them needs a
GPU backend to review, and none has a feature dependency.

Absorbs #4174, #4176, #4177, #4178, #4179, #4182, #4183 (all main-direct singletons, now closed and
pointed here). 14 commits, all replayed clean onto `main` — see the per-rung sections below.

**Merge button: "Rebase and merge" — never squash.** Each rung must keep its own revert handle.

**Base is legal for every member:** all eight branches were ladder-free (their entire delta versus `main`
is their own rungs — no stacked-series content), so all 14 commits replay onto `upstream/main`
(`2b9140ba`) with zero conflicts.

---

## 1. `ActiFunc::gelu` / `tanhGelu` overflow the output buffer by 2x under FP16 activations

`gelu` and `tanhGelu` are templates on the activation element type `T`, but their bodies hardcode
`getData<float>()` on an **element** count. `Tensor::getData<T>()` is an unchecked cast that does not
consult the tensor's real dtype, so with `T = _FP16` the call hands a buffer of N 2-byte elements to a
routine that reads N 4-byte floats and writes N 4-byte floats: it reads 2x past the end of the input and
**overflows the output buffer by 2x**. Heap corruption, not merely wrong numerics.

Reachable: `ActivationLayer::finalize` really does select the `_FP16` instantiation when
`getActivationDataType() == FP16`, and ten `ModelTensorDataTypeInfo` strings are `*-FP16`. In-tree
consumers: `gelu` in BERT, TIMM-ViT and DeBERTa-v2; `tanh_gelu` in Gemma3. It went unnoticed because
every `model_tensor_type` committed to this tree is `FP32-FP32` or `Q4_0-FP32`, so CI only ever
instantiates `T = float`, where the cast happens to be correct.

Fix keeps the accelerated fp32 path (`gelu_v2` / `tanh_gelu` are AVX2- and NEON-vectorized; de-SIMDing
fp32 would be a real regression) behind `if constexpr (std::is_same_v<T, float>)` and routes every other
`T` through `Tensor::apply<T>()`, which walks with the correct element type and throws if the output
dtype disagrees — exactly what the neighbouring `sigmoidGelu` already does. The lambdas accumulate in
float and cast back to `T`, reproducing the existing `__fallback_gelu_v2` / `__fallback_tanh_gelu` fp16
references bit for bit.

**Honest empty cell:** there is no in-tree `*-FP16` activation fixture, so this cannot be closed by a
committed regression test today. Closing it needs either a committed FP16-activation reference config or
an ASAN build over one of the four affected models.

### Automated-review findings carried over from the absorbed PR

The singleton attracted two bot findings. Neither is lost:

- **`acti_func.h` / MSVC `M_PI` (P1).** Including `<cmath>` before `_USE_MATH_DEFINES` is defined means
  the include-guarded CRT math header is already in, so a later `_USE_MATH_DEFINES` cannot expose
  `M_PI` — and a header cannot guarantee definition order for its consumers, so `geluPrime()` would
  break any MSVC TU that includes it. **Already fixed on this branch**: the header now carries
  `constexpr static inline double PI = 3.14159265358979323846;` as a plain class constant (same value as
  the CRT/glibc `M_PI`, so results are unchanged) with the ordering rationale in the doc comment.
- **CodeQL alert 448, `acti_func.h:491` — "multiplication result may overflow float before conversion to
  double"** in `geluPrime`'s `pow(x * tmp, 2)`. Assessed as a false positive and **deliberately not
  changed here**: `tmp` is `1/sqrt(2) < 1`, so `x * tmp` strictly shrinks the magnitude and cannot
  overflow for any representable `x` in either `float` or `_FP16`. Every candidate silencing edit
  (widening the intermediate to `double`) would change the intermediate precision of a gradient path
  that has no committed reference, so it is recorded here as a known, understood alert rather than
  silently re-numbered.

## 2. Loader: fail loud on a weight-name miss and on a model-layout mismatch

Two silent-wrong-data / late-crash paths, both hardening, no behaviour change on a well-formed package.

- **Name-miss reads the container header as weight data.** `NeuralNetwork::load`'s safetensors path
  assigns each weight its file offset **by name**; on a miss it just `continue`d, leaving `file_offset`
  at its default `0`. The weight is then read from byte 0: the 8-byte header length followed by the
  ASCII JSON header. Those byte pairs decode as fp16 5.3e4…6.1e4 — large-but-finite values
  indistinguishable from a plausible weight to everything downstream, and the EOF tripwire cannot catch
  it because offset 0 never overruns. Now logs the tensor name and the file. Warning, not a throw: a
  genuinely optional record is a legitimate miss, and hard-erroring needs a per-model audit this change
  is not making.
- **Layout mismatch SIGSEGVs on a loader thread.** Per-tensor offsets are trusted blindly by the
  parallel mmap reader, so any graph-vs-file size disagreement surfaces as a `memcpy` past the end of the
  mapping — a crash on a worker thread, far from the actual bug. Now compares the computed layout total
  against the real file size up front and throws with both numbers. Demonstrated on a QINT4-FP16 qwen3
  package where the graph sizes QINT4 records by `Int4QTensor::getMemoryBytes()` while the container on
  disk stores the plain PR#3978 layout (**graph expects 375,483,784 vs 348,646,792 on disk**): that
  package now errors with an actionable message instead of crashing.

**Validated:** silent on all four target packages (a 35-layer sliding-window QS4CX-FP16 model, gemma4,
qwen3, gemma2 — **0 warnings each**; they load through the positional `.bin` path, not this one), so
neither message is the cause of any current defect.

## 3. x86 host QS4CX GEMM for `HalfTensor::dot`

`main`'s `half_tensor.cpp` has **zero** QS4CX references, so `HalfTensor::dot` has no QS4CX case at all:
an fp16-activation QS4CX model has no host GEMM to fall back to on x86. This adds one.

Fix by delegation rather than a fourth hand-rolled decoder: the new case calls
`gemm_qai8dxp_qsi4cxp_rhs_unpacked(..., is_nxk=true)` — the exact x86 entry point
`FloatTensor::dotQs4cx` already uses — with the fp16 activation widened to fp32 per chunk and the fp32
result narrowed back. Staging is `chunk` floats per worker, **not `M*N`** (an untied lm_head is
`N = 262144`).

**Why one squashed commit:** the internal precursor hand-decoded the nibble payload and got the layout
wrong three independent ways (wrong axis order, wrong nibble half, signed instead of unsigned offset —
full derivation against the two in-tree layout authorities is in the commit message). Publishing that
precursor even for one commit would introduce a known-wrong decoder that never existed upstream.

Also carries four small x86 fp16-activation gaps the same work unmasked: `rms_norm` /
`reshaped_rms_norm` read gamma at `context.getWeightDataType()` instead of a hardcoded FP32 (matching
`*-FP16` packages that store norm gammas at the model weight dtype); `fallback_internal_fp16`'s
`rms_norm_wrt_width_fp16` scalar reference (was an NYI throw); `Transformer::repack_weight` skipping on
NYI per weight so x86 (no KleidiAI rhs-pack) loads the plain QS4CX blob instead of aborting; and a
rejection of `model_tensor_type=QS4CX-FP16` where no host kernel exists.

**Not included:** the ARM case in the same `#if/#elif` block
(`nntr_gemm_qai8dxp_qsi4cxp_packed<_FP16>`) — it depends on KleidiAI fp16 packed-weight infrastructure
that does not exist on `main`.

**Three-axis note:** structure ++ (one decoder per backend, matching the existing x86/ARM/GPU QS4CX
authorities instead of adding a fourth); performance ~ (adds an `M*K` fp32 staging copy plus int8
requantization per call, on the host QS4CX-FP16 fallback path only); memory 0 (chunk-sized staging).

## 4. Gate the CL q4_0 accel path on prepack shape eligibility — **7 of 10 crashing suites go green**

`ClComputeOps::supports_gemm_q4_0_{batch,accel}_fp32()` return `true` **unconditionally**, so
`float_tensor.cpp`'s Q4_0 dispatch sends every `M > 1` GEMM into the CL prepack. Both CL entry points
prepack the Q4_0x8 weight on the host with `unpack_q4_0x8_transpose16`, whose x86 implementation is an
AVX2 256-wide K unroll that asserts `(K % 256) == 0` and `(N % 8) == 0`. Any Q4_0 weight with
`K % 256 != 0` therefore **aborts the process** on `engine=gpu`: the tiny CausalLM fixtures
(hidden_size = 64) crash **7 of the 10 `Q40CloseToFP32Reference` suites**, while the same suites pass on
`NNTR_ENGINE=cpu`. Reproduces on `main` — a pre-existing upstream defect.

Fix at the CL seam, not at the neutral dispatch site: the wrappers that claimed support now check shape
eligibility (`K % 256 == 0 && N % 8 == 0`) and route ineligible shapes to the host `gemm_q4_0` — the
exact function the cpu engine uses, on the same host/SVM pointers. **That host bounce is NAMED in a code
comment rather than hidden.** Eligible shapes take the identical CL calls as before: zero behaviour
change.

The same seam also needed the generic branch: the dispatch's `else` path (`M == 1`, i.e. every decode
step) calls `gemm_q4_0_fp32`, which `ClComputeOps` never overrode, so the base class threw "not
implemented by this backend" on the first generated token — a latent defect the prefill crash masked.

**Measured:** the 10 `Q40CloseToFP32Reference` suites on an OpenCL + fp32 build — **7 CRASH → PASS, 3
PASS-still**. Highest value-per-line change in this bundle.

**Known NOT fixed:** the tiny-BERT Q4_0-on-gpu path completes after this change but yields NaN
embeddings from a **distinct** pre-existing gpu-engine Q4_0 weight-load defect that the crash used to
hide.

## 5. Factory `int_key` assignment position-independent + fail loud

- `int assigned_int_key = int_key == -1 ? str_map.size() + 1 : int_key;` — `str_map.size()` is a count,
  not a high-water mark, and the collision guard only ran for `int_key != -1`, so `int_map[k] = type`
  silently **rebinds** an existing key whenever the string map's size collides with an
  explicitly-assigned key. A registration mistake then resolves to the wrong layer type with nothing
  logged.
- `add_default_object()` ran **before** `setMemAllocator()` in `cl_context.cpp`. Any registration throw
  during default-object setup left a fully "initialised" context with a **null `MemAllocator`**, and
  every model built on it died later in the `TensorPool` constructor with a symptom that names nothing.

Fix: derive the auto key from the actual maximum in use, run the collision guard on every assignment,
fail loud on collision; order `setMemAllocator()` before `add_default_object()`. `AppContext::initialize`
already orders them the safe way, so this propagates an in-tree precedent. Covers the core three
contexts plus QNN.

**Honest note:** the canary for the first defect was run and did **not** fire — the in-tree mitigation (a
comment asking contributors to append last, plus explicit keys 9001/9002) held. The mechanism is still
there; the second defect's null-allocator window is unconditional.

## 6. Three inference-path defects: FFN gate/up load order, prompt BOS, token-id staging

All three found by **differencing a layer-0 device capture against a reference**, not by reading code.

**(a) The generic FFN loads `gate_proj` into `ffn_up` — a *loader-order* defect.** The
`swiglu({up, gate}, {1, 0})` wiring on `main` is already correct (`tensor_api.h:641-655` documents that
the slot map inverts). The defect is upstream of it: **the weight loader is positional.** File offsets
are assigned in graph-creation order, not by name, and converters write `gate_proj` → `up_proj` →
`down_proj`. `Transformer::createMlp` created `ffn_up` **first**, so `ffn_up` consumed `gate_proj`'s
bytes. Net computation: **`silu(up_proj) * gate_proj`** — coherent-looking activations, wrong function.
The capture showed a *swap*, not a diff:

```
layer0_ffn_up:out0:y     this tree b2d89b8a30493f9b   reference a62dba9a0458afda
layer0_ffn_gate:out0:y   this tree a62dba9a0458afda   reference b2d89b8a30493f9b
```

Fix: create `ffn_gate` first, `ffn_up` second, `swiglu({gate, up})` with no slot map. This also makes the
layer **names** semantically true, which matters for every name-keyed consumer (differential probes,
divert trace, residency dump). Result: **qwen3 is golden on both lanes for the first time.**

Scope is exactly the models that do not override `createMlp` — on `main`, qwen3 and any future generic
model. `gemma3`, `gemma4`, `bert`, `gpt_oss` (+ `gpt_oss_cached_slim`), `lfm2`, `qwen3_moe`
(+ `qwen3_slim_moe`, `qwen3_cached_slim_moe`) and `timm_vit` all override `createMlp` and were already
gate-first, so they are **byte-identical**. Why no test caught it: `unittest_causallm_models` fixtures
load weights **name-keyed** and the defect is **positional** — 86/86 before and after.

**(b) Missing BOS.** `CausalLM::run()` used the 1-argument `tokenizer->Encode(prompt_)`; for a tokenizer
with `TemplateProcessing` / `add_bos_token=true` the leading BOS is **silently dropped**. The capture
diverges at the **first tensor, on a size**: `layer0_attention_norm:in` `bytes=23040` (5x2304x2) against
the reference's `27648` (6x2304x2), and the reference's row-0 first fp16 is `0x8000` (-0.0) = the BOS
embedding. Fix is `Encode(prompt_, /*add_special_tokens=*/true)` — **not** a hard-coded BOS insert, which
would double-prepend for templates that already emit one.

**(c) Unzeroed token-id staging buffer.** `BATCH_SIZE * MAX_SEQ_LEN` floats were `malloc`'d and only
`input_len` written, so the graph read uninitialised memory past the prompt: `input0` entered at
`min=-5.92e+35 max=5.62e+21` against the reference's `min=0 max=2.37e+05`. Fix is `calloc`.

**(d)** A fourth commit makes (b) multi-turn-safe: the special tokens are added **only at sequence
position 0**, and the cached rows are counted the same way, so a continued conversation does not
re-prepend BOS on every turn.

**Validated:** 4-model short-prompt text byte-identical to the reference over 3 consecutive runs, and the
per-layer hash stream matches line for line on the captured model. Two of the three defects each moved a
model from wrong-but-plausible output to golden (qwen3 via the loader order, the Gemma2 packaging via the
BOS); gemma4 is unchanged and serves as the regression control. The two goldens that change here change
**by design** — they were wrong before.

## 7. `attn_logit_softcapping` lived in a dead override

`setupParameters()` is only ever dispatched from base-class **constructor** bodies. C++ resolves a
virtual call made inside a base constructor to the *base* version, never a derived override. A model
whose own constructor body adds no `setupParameters()` call therefore never runs its override, and every
key parsed only there keeps its default. `Gemma3Transformer` is exactly that shape, so the
`attn_logit_softcapping` parse never ran, `ATTN_LOGIT_SOFTCAPPING` kept its `0.0f` default, and — since
every consumer is gated `if (attn_logit_softcapping > 0.0f)` — the `c*tanh(s/c)` soft-cap was skipped on
every layer, on every backend.

Fix: move the parse into `Transformer::setupParameters` — the one function the base constructor provably
reaches — and delete the per-model copies. Gated on `contains(...) && !is_null(...)`, so any model whose
config lacks the key keeps `0.0f` and is byte-preserving.

Audited per model: **gemma3 / EmbeddingGemma** plumbing cured, **numerics unchanged** (in-tree configs
carry `"attn_logit_softcapping": null`); **gemma4** unchanged (its override is live and re-read the same
key into the same member); **qwen2 / qwen3 / qwen3_moe / gpt_oss / lfm2 / bert / deberta / xlm_roberta /
timm_vit** have no such key. The real numerics victim is **gemma2**, whose shipped configs all carry
`"attn_logit_softcapping": 50.0` — this fix is its prerequisite.

Ships `Gemma3ConfigPlumbingTest` and a shared `makeTinyCtorOnlyNntrainerConfig()` helper in
`causallm_test_utils`. The test is ctor-level parameter *visibility*: a probe subclass exposes the
protected member and constructs the model with no tokenizer, no weights and no graph compile. Three
cases: key present (the regression guard, which **fails on the unfixed tree**), key absent, key
explicitly null.

**Merge-order note:** this bundle ships `causallm_test_utils.{h,cpp}`, which the Gemma2 model PR's unit
test consumes. Please land this **before** the Gemma2 model PR.

## 8. Honor `nntr_config`'s `lmhead_untie`

The app only read `config.json`'s `tie_word_embeddings` when choosing the lm_head layer type;
`nntr_config.json`'s `lmhead_untie` was ignored entirely. Untied-serialized packages (tied HF config plus
a separate transposed `[hidden, vocab]` head record appended to the bin) therefore built a **tied** graph
against an **untied** bin. That used to misload silently; with rung 2's loader tripwire it now fails fast
with a model-layout mismatch — either way the package is unusable.

- `Transformer::setupParameters` parses `lmhead_untie` into a new `LMHEAD_UNTIE` member. It lives on
  `Transformer`, not `CausalLM`, because `embedding0`'s layer-type choice happens in
  `<model>Transformer::constructModel` scope, which does not see `CausalLM` members in the diamond
  hierarchy.
- New `CausalLM::buildLmHeadOutput(h, add_skip_prefill)`: untied ⇒ an independent `fully_connected`
  output with its own weight (`weight_dtype = LMHEAD_DTYPE`, optional `skip_prefill` so the head stays
  decode-only); tied ⇒ the existing `tie_word_embeddings` / `lm_head` choice with `shared_from`,
  unchanged. Shared by the generic path and by Gemma4.
- `Gemma4Transformer::constructModel` gates `embedding0` on `TIE_WORD_EMBEDDINGS && !LMHEAD_UNTIE`: with
  an untied head there is nothing to share, so `embedding0` becomes a plain `embedding_layer`
  (lookup-identical, same weight record), which also makes the `embedding_file_name` sidecar legal.

Byte-preserving for every package without the key: `LMHEAD_UNTIE` defaults false, emitted property
strings and their order are unchanged, and no `skip_prefill` is added on paths that did not add it
before.

**Not verified here:** tied goldens (qwen3 / gemma2 / gemma4 packagings) byte-identical, and an
untied-serialized gemma4 package passing the load tripwire and producing coherent output. Both pending
build-host time; the change is byte-preserving by construction for every package that does not set the
key.

---

## Merge order

Rungs 7 and 8 (`attn_logit_softcapping`, `lmhead_untie`) are also re-included inside two of the stacked
consolidation branches in this series. That is the stacked series' pre-existing shape, not something this
bundle creates, and it dictates order: **land this PR before those stacked PRs**, so the cascade copies
arrive as no-ops under rebase-and-merge rather than as conflicts.

Rung 6(a) reorders the same `ffn_up` / `swiglu` lines that the `engine=` selection rung stamps with
`withKey("engine", causallm_engine())`. Whichever lands second shows one small textual conflict in
`createMlp`; the intended resolution is "**gate-first *and* stamped**", which is exactly the combination
the downstream feature branches already carry, so no goldens move as a result.

---

### DIFF
--- a/Applications/CausalLM/layers/reshaped_rms_norm.cpp
+++ b/Applications/CausalLM/layers/reshaped_rms_norm.cpp
@@ -35,14 +35,22 @@ void ReshapedRMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
     << "feature size must be a divisor of width";
 
   if (use_gamma) {
-    // gamma follows the model weight dtype (packed=false clamps it to the
-    // activation dtype), matching what the quantizer wrote to the bin.
-    // Hardcoding FP32 misloads/overruns *-FP16 bins whose gammas are FP16;
-    // the multiply site's dtype-cast guard covers either stored dtype.
+    // gamma is never quantized, but a bin stores it at the model's FLOAT
+    // dtype, so an FP16-activation package holds 2 bytes per element. Ask for
+    // the context's weight dtype when that dtype is itself a float type, and
+    // FP32 otherwise -- the same rule as rms_norm.cpp and the core
+    // nntrainer/layers/llm/rms_norm_layer.cpp, which carries the full note.
+    // These q/k/v norms are counted in the same packed weight stream, so a
+    // constant FP32 here shifts every later offset on an FP16 package.
+    const auto norm_w_dtype = context.getWeightDataType();
+    const auto norm_dtype =
+      (norm_w_dtype == nntrainer::TensorDim::DataType::FP32 ||
+       norm_w_dtype == nntrainer::TensorDim::DataType::FP16)
+        ? norm_w_dtype
+        : nntrainer::TensorDim::DataType::FP32;
     nntrainer::TensorDim gamma_dim(
       1, 1, 1, feature_size,
-      nntrainer::TensorDim::TensorType(context.getFormat(),
-                                       context.getWeightDataType()));
+      nntrainer::TensorDim::TensorType(context.getFormat(), norm_dtype));
     wt_idx[RMSParams::gamma] = context.requestWeight(
       gamma_dim, nntrainer::props::InitializerInfo::Enum::NONE,
       nntrainer::WeightRegularizer::NONE, 1.0f, 0.0f, "gamma", true);
--- a/Applications/CausalLM/layers/reshaped_rms_norm.cpp
+++ b/Applications/CausalLM/layers/reshaped_rms_norm.cpp
@@ -35,13 +35,14 @@ void ReshapedRMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
     << "feature size must be a divisor of width";
 
   if (use_gamma) {
-    // gamma is unquantized FP32 on disk; request FP32 regardless of activation
-    // dtype (FP16 would reinterpret the FP32 bytes and corrupt gamma). The FP16
-    // path casts gamma down at the multiply site.
+    // gamma follows the model weight dtype (packed=false clamps it to the
+    // activation dtype), matching what the quantizer wrote to the bin.
+    // Hardcoding FP32 misloads/overruns *-FP16 bins whose gammas are FP16;
+    // the multiply site's dtype-cast guard covers either stored dtype.
     nntrainer::TensorDim gamma_dim(
       1, 1, 1, feature_size,
       nntrainer::TensorDim::TensorType(context.getFormat(),
-                                       nntrainer::TensorDim::DataType::FP32));
+                                       context.getWeightDataType()));
     wt_idx[RMSParams::gamma] = context.requestWeight(
       gamma_dim, nntrainer::props::InitializerInfo::Enum::NONE,
       nntrainer::WeightRegularizer::NONE, 1.0f, 0.0f, "gamma", true);
--- a/Applications/CausalLM/layers/rms_norm.cpp
+++ b/Applications/CausalLM/layers/rms_norm.cpp
@@ -28,17 +28,29 @@ void RMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
   if (!std::get<nntrainer::props::SkipPrefill>(rms_props).empty())
     skip_prefill = std::get<nntrainer::props::SkipPrefill>(rms_props).get();
 
-  // gamma follows the model weight dtype (packed=false clamps it to the
-  // activation dtype). Quantized *-FP16 bins store the norm weights as FP16;
-  // hardcoding FP32 here positionally misloads every weight after the first
-  // gamma AND overruns the mmap'd file by 2 bytes/element (SIGSEGV in the
-  // parallel load worker). The multiply site carries a dtype-cast guard, so
-  // FP32-gamma bins (upstream's own layout) still load when the model weight
-  // dtype is FP32.
+  // gamma is never quantized, but it is not unconditionally FP32 either: a bin
+  // stores it at the model's FLOAT dtype, so an FP16-activation package holds
+  // 2 bytes per element. Ask for the context's weight dtype when that dtype is
+  // itself a float type, and FP32 otherwise -- see the note in the core
+  // nntrainer/layers/llm/rms_norm_layer.cpp, which this layer mirrors:
+  //   - a hard FP32 over-requests on an FP16 package and, because weights are
+  //     a packed stream in graph order, shifts every later offset until the
+  //     loader rejects the layout;
+  //   - a bare getWeightDataType() is only safe while every caller passes
+  //     packed=false (LayerNode::finalize() then sets tensor_type[1] =
+  //     tensor_type[2] before InitLayerContext exists); the float-type test
+  //     keeps a caller that forgets it from requesting a block-quantized
+  //     gamma.
+  // The FP16 forward path casts gamma down at the multiply site either way.
+  const auto norm_w_dtype = context.getWeightDataType();
+  const auto norm_dtype =
+    (norm_w_dtype == nntrainer::TensorDim::DataType::FP32 ||
+     norm_w_dtype == nntrainer::TensorDim::DataType::FP16)
+      ? norm_w_dtype
+      : nntrainer::TensorDim::DataType::FP32;
   nntrainer::TensorDim gamma_dim(
     1, 1, 1, dim[0].width(),
-    nntrainer::TensorDim::TensorType(context.getFormat(),
-                                     context.getWeightDataType()));
+    nntrainer::TensorDim::TensorType(context.getFormat(), norm_dtype));
   wt_idx[RMSParams::gamma] = context.requestWeight(
     gamma_dim, nntrainer::props::InitializerInfo::Enum::NONE,
     nntrainer::WeightRegularizer::NONE, 1.0f, 0.0f, "gamma", true);
--- a/Applications/CausalLM/layers/rms_norm.cpp
+++ b/Applications/CausalLM/layers/rms_norm.cpp
@@ -28,14 +28,17 @@ void RMSNormLayer::finalize(nntrainer::InitLayerContext &context) {
   if (!std::get<nntrainer::props::SkipPrefill>(rms_props).empty())
     skip_prefill = std::get<nntrainer::props::SkipPrefill>(rms_props).get();
 
-  // gamma is unquantized and stored as FP32 in the bin. Request it as FP32
-  // regardless of the activation dtype; declaring it FP16 reinterprets the
-  // on-disk FP32 bytes as FP16 and corrupts gamma (≈FP16-max garbage). The
-  // FP16 forward path casts gamma down to FP16 at the multiply site.
+  // gamma follows the model weight dtype (packed=false clamps it to the
+  // activation dtype). Quantized *-FP16 bins store the norm weights as FP16;
+  // hardcoding FP32 here positionally misloads every weight after the first
+  // gamma AND overruns the mmap'd file by 2 bytes/element (SIGSEGV in the
+  // parallel load worker). The multiply site carries a dtype-cast guard, so
+  // FP32-gamma bins (upstream's own layout) still load when the model weight
+  // dtype is FP32.
   nntrainer::TensorDim gamma_dim(
     1, 1, 1, dim[0].width(),
     nntrainer::TensorDim::TensorType(context.getFormat(),
-                                     nntrainer::TensorDim::DataType::FP32));
+                                     context.getWeightDataType()));
   wt_idx[RMSParams::gamma] = context.requestWeight(
     gamma_dim, nntrainer::props::InitializerInfo::Enum::NONE,
     nntrainer::WeightRegularizer::NONE, 1.0f, 0.0f, "gamma", true);
--- a/Applications/CausalLM/meson.build
+++ b/Applications/CausalLM/meson.build
@@ -284,6 +284,28 @@ if get_option('enable-test')
     suite: 'unittests'
   )
 
+  # Prompt special-token policy: no model, no graph, no external assets -- the
+  # test builds its own BOS-prepending tokenizer.json blob, so it only needs the
+  # tokenizer backend.
+  unittest_prompt_special_tokens = executable(
+    'unittest_prompt_special_tokens',
+    [
+      meson.current_source_dir() / 'models' / 'unittest_prompt_special_tokens.cpp',
+      meson.current_source_dir() / 'huggingface_tokenizer.cpp',
+    ],
+    dependencies: [gtest_main_dep],
+    include_directories: causallm_inc,
+    link_args: tokenizer_link_args,
+    build_by_default: true,
+    install: false,
+  )
+
+  test('unittest_prompt_special_tokens', unittest_prompt_special_tokens,
+    args: '--gtest_output=xml:@0@/@1@.xml'.format(meson.build_root(), 'unittest_prompt_special_tokens'),
+    timeout: 60,
+    suite: 'unittests'
+  )
+
   unittest_cancel_api = executable(
     'unittest_cancel_api',
     [
--- a/Applications/CausalLM/models/causal_lm.cpp
+++ b/Applications/CausalLM/models/causal_lm.cpp
@@ -220,26 +220,43 @@ void CausalLM::advanceKVCachePosition(unsigned int step_size) {
   kv_cache.advance(step_size);
 }
 
-std::pair<Tensor, Tensor> CausalLM::constructModel() {
-
-  // base transformer (input, output_norm)
-  auto [x, h] = Transformer::constructModel();
-
+/**
+ * [lmhead-untie] When nntr_config.json sets lmhead_untie, build
+ * output_of_causallm as an independent fully_connected layer with its own
+ * weight even for a tied-embedding model, so the lm_head can carry a
+ * different dtype than the embedding (untied-serialized packages such as
+ * gemma4_qs4cx_fp16 ship a separate transposed [hidden, vocab] head record
+ * that a tied graph cannot load). Untie is the config flag, NOT derived from
+ * LMHEAD_DTYPE: a quantizer constructs this same untied graph from the FP32
+ * source and quantizes output_of_causallm via the dtype map on save.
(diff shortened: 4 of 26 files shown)


jijoongmoon:
## `Transformer::createMlp` gate/up operand order: this branch carries the swapped form

This branch's `Applications/CausalLM/models/transformer.cpp::createMlp` creates `ffn_gate`
before `ffn_up` and calls `swiglu({gate, up})` with no input-slot map. That is the **wrong**
direction of the FFN gate/up fix and should be reverted to up-first before this merges.

**Why it's wrong:** the `.bin` loader assigns each FFN weight's file offset positionally, by
graph-creation order, not by layer name. The in-tree converters write the FFN weights in
`up_proj, gate_proj, down_proj` order. Creating `ffn_gate` first means the `ffn_gate` layer
actually consumes the file's `up_proj` bytes, and `ffn_up` consumes `gate_proj`'s bytes — so
`swiglu({gate, up})` computes `silu(up_proj) * gate_proj` instead of `silu(gate_proj) *
up_proj`. The output is coherent-looking (finite, plausible activations) rather than obviously
broken, which is why this has round-tripped more than once.

**Measured cost, on an isolated fix (up-first + explicit `swiglu({up, gate}, {1, 0})` slot
map):**

| suite | before (gate-first) | after (up-first + slot map) |
|---|---|---|
| LFM2-MoE differential | 1.06e-2 | 2.98e-5 |
| LFM2 (dense) differential | 9.75e-3 | 5.11e-5 |
| Qwen3 differential | 5.99e-3 | 6.24e-5 |
| Qwen2-tiny differential | 5.60e-3 | 3.43e-5 |

**Fix shape:** create `ffn_up` first, `ffn_gate` second (so the layers land in file order), and
wire `swiglu({up, gate}, {1, 0})` — the explicit slot map keeps swiglu's input 0 bound to the
gate projection's *output tensor* while the two layers themselves stay in file-write order.
Models that override `createMlp` (Gemma2/3/4, the MoE variants, BERT, TIMM-ViT) are unaffected —
they already create their gate/up equivalents in the order their own converters expect.

**Note on how this regressed:** an earlier pass at this exact bug went the other way — creating
gate first and dropping the slot map — based on a differential probe compared against a
reference that itself carried the same gate-first layout, so the probe matched two wrong
implementations to each other and read as a fix. The table above is from an independent
re-derivation (checking the converters' actual on-disk write order directly, plus the
differential suites) and points the other way. Recommend treating "up-first, explicit slot map"
as the settled direction and not re-flipping it based on a differential probe alone unless the
reference side is also independently confirmed.

---

Affected branches: this PR's `dd6e6449d` is the origin, and the same commit was replayed into the branches behind #4198, #4201, #4211, #4222, #4227, #4228 — those need the same reversal before merging. Related fixture issue: #4266.

jijoongmoon:
Partly superseded by #4316 in the re-cut series against current main: the FFN gate/up load order, the prompt BOS via add_special_tokens, the attention-logit soft-cap parsing and lmhead_untie all land there. The remaining items in this PR — the FP16 activation overflow, the loader fail-loud behaviour and the q4_0 shape gate — belong to the later application-side rungs of the same series, so this stays open until those are posted rather than being closed against a successor that carries only part of it.

jijoongmoon:
Further superseded by #4317, which is the next application-side rung of the same re-cut series. It carries the FP16 activation path for the attention core and the DebertaV2 attention layer — the two translation units whose missing all-FP16 CPU kernels made `-Denable-fp16=true` fail to build for this application at all — together with the GPU attention core and the KV-cache manager.

Of the items listed here, that leaves the loader fail-loud behaviour and the q4_0 shape gate, which belong to the remaining application-side rungs. This PR stays open until those are posted rather than being closed against successors that carry only part of it; replacement bundle to follow, branch preserved at `upstream-pr/graph-correctness`.

jijoongmoon:

As noted above, four of this PR's eight fixes are carried, in #4316/#4317: the FFN gate/up load order, the prompt BOS via add_special_tokens, the attention-logit soft-cap parsing, and lmhead_untie.

The remaining four are NOT carried anywhere in the series, and remain open candidates for a follow-up PR after the series lands:
- the FP16 activation overflow fix in gelu/tanhGelu for non-float tensor types
- the loader fail-loud behavior on a model-layout mismatch
- the q4_0 CL-accel shape eligibility gate
- the factory int_key auto-assignment fix

The branch is preserved.

PR #4195, opened 2026-07-30
[compute_ops] Reverse-RMSNorm as a whole-op virtual; retire the Layer's raw CL call

jijoongmoon:
Refactors the OpenCL reverse-RMSNorm GPU path into the conforming **whole-op** shape, and retires the
Layer's raw kernel call.

The port this replaces violated the op-table contract three ways: the Layer body called the raw kernel
wrapper `rms_reverse_norm_cl_fp16` directly from inside `incremental_forwarding`; the call sat behind a
**silent** four-way `isSVM()` fall-through (if any predicate was false, control fell through with no log
and no throw into a host FP32-temp block — a hidden host bounce installed on the very path whose hidden
host bounce was the original defect); and it grew the installed `blas_kernels.h` ABI by one raw symbol.

**Motion, not modification** — old code moves to a new home, point by point:
- the `+81`-line wrapper moves **token-identical** (whitespace-normalised diff empty) out of
  `blas_kernels.cpp` into a new `nntrainer/tensor/cl_operations/rms_reverse_norm_cl_op.cpp`, anonymous
  namespace: same kernel, same argument binding, same launch geometry. `blas_kernels.{cpp,h}` return to
  their pre-port blobs **byte-identical**, so the raw symbol leaves the installed ABI and the new header
  is deliberately not in the installed CL header list.
- the Layer's four-way SVM predicate becomes a **named reject** per leg in the whole-op, in the same
  short-circuit order — so the silent fall-through becomes a reportable reason.
- a `ComputeOps::rms_reverse_norm` virtual with an `NI()` default; `CpuComputeOps` overrides it with the
  existing host body moved verbatim; the Layer body collapses to one `getOps()->rms_reverse_norm(...)`
  call. Because the CL ops class derives from the base (not from the CPU class), a tensor that misses
  the SVM contract now **throws** instead of silently bouncing — the hidden-host-bounce defect
  dissolves structurally rather than being patched.

Depends-on: #4193, #4103

**Merge-order note:** land **after** the CL install-list correction (#4103), which is what makes the
`blas_kernels.h` ABI restoration meaningful — the two changes move that header in the same direction and
sequencing them the other way re-adds a raw symbol that #4103 removes.

## Its own measured claim, kept separate from the parent rung's

The defect this pays down was not cosmetic. The per-layer-embedding model's blank output was a **host island
between two GPU ops reading an undrained SVM shadow**: the capture showed a uniform `-0` plane at
`layer0_ple_post_norm` where the reference had `-0.881..1.26`. With the op on the device path the OpenCL lane
matches the reference **token for token**, at **444 prefill / 20.9 decode TPS against the reference's
500 / 20.8** — i.e. the correctness fix costs nothing measurable on the decode side and is within the
prefill band of a different implementation of the same kernel.

**Three-axis note.** Structure: **+++** (a raw symbol leaves the installed ABI, a silent fall-through becomes
a named reject, the Layer body collapses to one call). Performance: **0 by construction** — same kernel, same
launch geometry, one virtual call per whole op. Memory: **0**. The GPU-contribution-checklist grep for raw
`*_cl` calls in app/layer bodies moves 50 lines / 10 files → **49 / 9**.

**Cross-backend gap, named:** #4209 adds a genuine **CUDA device override** for this same
`ComputeOps::rms_reverse_norm` virtual. Until then `CudaComputeOps` inherits the identical host body from
`CpuComputeOps` — which is why this refactor needs no CUDA edit at all, and why no reviewer should read the
missing CUDA override as an oversight.

**Diagnostic-only residual, disclosed:** with `NNTR_GPU_SVM_POOL=0` **and** the full GPU flag set, the
whole-op emits its named reject (as designed) and the run then dies with `CL_OUT_OF_RESOURCES` — a
combination no normal path uses (the SVM pool is on by default on the gpu engine). It is a diagnostic
configuration, not a supported one; the named reject is what makes it visible at all rather than a silent
degradation.

**New in this rung:** the whole-op refactor plus the two commits it rests on in this branch.

**Not included:** the CUDA override for this op (later rung), and the private model that consumes this
layer — the *layer* is upstream, only the model is not.

**Validated:** fp16 + OpenCL build clean; the moved wrapper is token-identical, so the GPU numeric path
is unchanged by construction, and the host path is the same body relocated. The behavioural change is
strictly that a missing SVM contract now throws instead of silently degrading.

---

**Merge button: "Rebase and merge" — please do not squash.** The wrapper move, the
`blas_kernels.{cpp,h}` ABI restoration and the virtual's introduction are separate revert handles, and the
token-identical move is only verifiable as a move while it stays its own commit.

Depends-on: #4193, #4103


jijoongmoon:
Consolidated into #4201.

PR #4196, opened 2026-07-30
[causallm] Allocate the KV cache from the SVM pool and make it survive the per-call placeholder fill

jijoongmoon:
Puts the KV cache on the OpenCL SVM plane **and makes the flag survive an inference call** — the second
half is the one that actually changes behaviour.

The first commit allocates the per-layer K/V from the `gpu-svm` `MemoryPool`, so the cache tensors
themselves report `isSVM() == true`. On its own it changed **no output**, because the flag did not
survive a single call:

`NeuralNetwork::incremental_inference` wraps every input buffer in a fresh `Tensor::Map`, which builds a
brand-new `MemoryData` with `svm_allocation = false`. `TensorPool::fillPlaceholder` then installs exactly
that `MemoryData` on the placeholder and `syncDependents()` propagates it to every dependent view —
including `mha_core`'s `cache_k` / `cache_v` views. So the SVM-ness of the pool allocation was
overwritten on **every** call, one call before `mha_core` ever looked at it.

`mha_core` gates its whole GPU/flash attention chain on `svm_ok = q && k && v && o all isSVM()`. With the
flag clobbered on `k` and `v` that predicate is unsatisfiable, so attention fell through to the host
GEMM — which reads the K/V plane through host pointers while the K/V producers wrote it on the device.
That is both a silent performance cliff and a coherence hazard.

The second commit makes the placeholder fill preserve the pool's SVM attribute instead of replacing it,
so `svm(q/k/v/o)` reads `1111` on every layer and the GPU attention chain is reachable.

**Residency is selected by a predicate, not by a backend name.** `KVCacheManager` selects on
`it->second->isSVM()` — the same `MemAllocator` predicate core itself uses — and the choice is deliberate,
with the caveat recorded in-comment right above it: `CudaMemAllocator` **overrides `isSVM()` to false**
even though its memory is host-coherent UVM, so a cuda run takes the host branch here by design. This rung
also folds in the fix for a backend-naming violation that the original port carried (a
`getName() == "gpu-svm"` / `getenv("NNTR_ENGINE")` selector), so no public commit in this series ever
contains that selector, and there is no `getenv("NNTR_ENGINE")` anywhere in the final code.

**The first commit is inert by design** — allocating the K/V from the `gpu-svm` pool changed **no output**,
because the flag did not survive one call. That is why the pair is the minimum reviewable unit: commit 1 is
the allocation, commit 2 is what makes it observable (`svm(q/k/v/o)` 1001 → **1111** on all 28 attention
layers of the measured model, which is what makes `mha_core`'s `svm_ok` satisfiable at all).

Depends-on: #4193, #4194

**New in this rung:** the two commits above (SVM-pool KV allocation; surviving the placeholder fill).

**Inherited, NOT claimed as new:** the three graph-correctness commits of #4194 and the Gemma4-prerequisite
lineage of #4193. One nuance a reviewer should know rather than discover: #4194 has been **re-anchored onto
`main`**, which required dropping two `withKey("engine", causallm_engine())` lines from its generic-FFN
commit (that function does not exist on `main`; #4193 re-adds them). This branch still carries the
pre-re-anchor form of that commit, so on rebase it collapses to nothing only once **both** #4194 and #4193
have landed — rebase this branch after both, not after #4194 alone.

**Not included:** chunked prefill and the KV window ring — both build on this plane and are separate
rungs; and the default-on flash flip, which is measurably wrong before this rung lands (forcing it on
without the surviving SVM plane regresses gemma4 to an immediate `<eos>`).

**Validated:** fp16 + OpenCL build clean; `svm(q/k/v/o)=1111` confirmed on all 28 attention layers of
the measured model; 4-model short-prompt text byte-identical over 3 consecutive runs.

---

**Merge button: "Rebase and merge" — please do not squash.** The inert allocation commit and the
placeholder-fill fix are separate revert handles: the second one is the behaviour change, and keeping it
separable is what makes "the flag was clobbered one call before anyone looked at it" checkable.

Depends-on: #4193, #4194


jijoongmoon:
Consolidated into #4201.

PR #4197, opened 2026-07-30
[CausalLM] Chunked forward prefill: cap the activation plane at long context

jijoongmoon:
`NNTR_PREFILL_CHUNK=C` drives the prefill **forward** in C-token chunks instead of one M-token block.
Each chunk feeds its tokens at input row 0 and writes KV at the absolute range `[from, from+clen)`; the
next chunk attends over the accumulated cache. This is the exact token-feed pattern the decode call
shape already uses (tokens at row 0, absolute KV position via `from`), generalized from `chunk=1` to
`chunk=C` — not a new execution model.

**What it buys.** The prefill *activation plane*, not the KV cache, is the dominant long-context memory
term. It is sized to `INIT_SEQ_LEN` because the shared activation tensors are built at
`{1,1,1,INIT_SEQ_LEN}` and `resetInputDimension` is disabled, so an unchunked whole-document prefill
needs `INIT_SEQ_LEN >= the document` and the plane grows linearly with context. Chunking decouples the
two: `INIT_SEQ_LEN` becomes the activation/chunk height (small) while `MAX_SEQ_LEN` stays the KV capacity
(large).

|---|---|---|
| `chunk=1024` | **4,272 MB** | **1085 TPS** (+15 % over the full plane) |

So chunking is not a memory-for-speed trade here — at long context it is faster *and* ~8 GB smaller,
because the full plane no longer fits comfortably.

`NNTR_PREFILL_CHUNK=0` is the sole documented opt-out for this rung, and the default is unchunked, so
existing behaviour is byte-identical unless the knob is set.

Depends-on: #4196

**New in this rung:** `[CausalLM] Chunked forward prefill: cap the activation plane at long context`.

**Not included / IMPORTANT:** the shared activation layers still assume "`from > 0` implies a
single-token decode step". Chunk 2 of a chunked prefill therefore aborts (`incremental step size is not
1`) until the **layer** rung lands — that is the very next PR in this series and this rung is not usable
without it. They are split because the layer change touches six layers and is reviewable on its own.

**Validated:** the 32K cell above; the default (unchunked) path is untouched.

---

Depends-on: #4196


jijoongmoon:
**Superseded by #4198**, which contains these commits and reviews them together.

`upstream-pr/trackCtx-chunked-prefill` is a **git ancestor** of #4198's head — verified with
`git rev-list <this head> ^<#4198 head>` = **0 commits**, so this rung's commit is literally in that PR's
history. #4198's body now carries this rung's documentation in full: the forward-chunk mechanism and why it
is the decode call shape generalized rather than a new execution model, the activation-plane measurement
(32K: 12,814 → 4,272 MB with +15 % prefill, and the note that ~8 GB of the 32K footprint is activation
plane and **not** KV, kept separate from the ring's KV number), the chunk-size sweep behind the 4096
default, the ARM/Adreno 0 default with its measured OHWI V-mirror restride reason (−34 % on a 1073-token
prompt), and the `NNTR_PREFILL_CHUNK=0` opt-out contract.

Why consolidate: chunking is **default-ON at 4096** on x86/x64, and without the layer half every chunk-2
call aborts (`incremental step size is not 1` / `Creating shared tensor of size bigger than tensor memory`)
for any prompt past the chunk size. Two PRs create a default-on broken state that a bisect can land inside;
one PR removes that bisect point. The capability *is* the `from > 0` block-call contract, so every layer
participating in it belongs in the same merge.

Closing in favour of #4198.

PR #4198, opened 2026-07-30
[CausalLM] Chunked forward prefill: cap the activation plane at long context (model + layer block-call contract)

jijoongmoon:
`NNTR_PREFILL_CHUNK=C` drives the prefill **forward** in C-token chunks instead of one M-token block. Each
chunk feeds its tokens at input row 0 and writes KV at the absolute range `[from, from+clen)`; the next
chunk attends over the accumulated cache. This is the exact token-feed pattern the decode call shape
already uses (tokens at row 0, absolute KV position via `from`), generalized from `chunk=1` to `chunk=C` —
not a new execution model.

## Why the model half and the layer half must merge as one PR

The capability **is** the `from > 0` block-call contract, so every participant has to arrive together.
`effectivePrefillChunk()` defaults to **4096 on x86/x64** (0 on ARM/Adreno — see below), i.e. this rung is
**default-ON**, and the shared activation layers still encoded the old "`from > 0` implies a single-token
decode step" assumption three different ways. Every one of them fires on **chunk 2** of a chunked prefill
(`from = C`, `to - from = C`):

```
[!] FATAL ERROR: incremental step size is not 1
[!] FATAL ERROR: Creating shared tensor of size bigger than tensor memory
```

Shipping the two halves as two PRs would create a **default-on broken state that a bisect can land inside**
— any prompt past the chunk size aborts. Consolidating them removes that bisect point instead of creating
one. The three assumptions and their fixes:

1. `fc_layer_cl` normalized `from` by writing `from = 0; to = 1;` — collapsing a C-token chunk to a single
   row (behind an assert that threw first). Now `to = to - from; from = 0;` — 0-based but step-preserving.
2. `fc_layer_cl`'s skip-prefill gate was `skip_prefill && from == 0`, so on chunk 2 a KV-shared projection
   did **not** skip and asked for a 1024-row view of an activation only sized for decode.
3. The four GLU layers (SwiGLU, GeGLU, SigmoidGLU, SigmoidAdd) carried the same single-row assumption in
   their incremental path.

## What it buys — the activation plane, not the KV cache

The prefill *activation plane* is the dominant long-context memory term. It is sized to `INIT_SEQ_LEN`
because the shared activation tensors are built at `{1,1,1,INIT_SEQ_LEN}` and `resetInputDimension` is
disabled, so an unchunked whole-document prefill needs `INIT_SEQ_LEN >= the document` and the plane grows
linearly with context. Chunking decouples the two: `INIT_SEQ_LEN` becomes the activation/chunk height
(small) while `MAX_SEQ_LEN` stays the KV capacity (large).

|---|---|---|
| `chunk=1024` | **4,272 MB** | **1085 TPS** (+15 % over the full plane) |

So at long context this is not a memory-for-speed trade: it is faster **and** ~8 GB smaller, because the
full plane no longer fits comfortably. **This number is the activation plane and nothing else** — roughly
8 GB of the 32K footprint is activation plane, not KV. The KV-side saving is a different rung with its own
number (#4200, the window ring: 514.5 MB of KV rows at 32K); please do not read the two as one claim.

**Chunk-size default.** The equal-thermal ring-on sweep on Xe3 at 32K is monotone in throughput but not in
memory (1024 → 1117.1, 2048 → 1125.4, 4096 → 1136.0, 8192 → 1146.3 TPS at peak WS
2696 / 2950 / 3419 / 4511 MB), so 4096 buys +1.7 % prefill for +723 MB while 8192's further +0.9 pp costs
another 1.1 GB; CUDA independently wants 4096 for tensor-core-efficient prefill GEMMs. Hence 4096.

**ARM/Adreno defaults to 0 (no auto-chunk), for a measured reason:** the per-layer OHWI V-mirror keeps a
texture-pitch-aligned image sized to the **first** call's `cache_to`, and a second chunk whose `cache_to`
crosses that stride's rounding boundary **re-scatters every already-cached KV row of every layer**.
Measured on a packed model at a 1.2K prompt: auto-chunking a 1073-token prompt into 1024+49 gave
**845 TPS against the single-block path's 1277 TPS (−34 %)**. An explicit `NNTR_PREFILL_CHUNK` is still
always honoured there.

**One knob, one rung.** `NNTR_PREFILL_CHUNK=0` is the sole opt-out and reproduces the legacy single-block
prefill **byte-identically**. It also disables the KV ring transitively (`kvRingCap` needs `C > 0` to bound
the live span); the ring's own opt-out `NNTR_KV_WINDOW_RING=0` does **not** alter chunking. Each knob owns
exactly its own rung.

Depends-on: #4196, #4187

**New in this rung:** 2 commits — `[CausalLM] Chunked forward prefill: cap the activation plane at long
context` (the model half, which also carries the default and the single measured claim above) and
`[layers] Chunked prefill: accept a from>0 block call in FC + the four GLU layers` (the layer half, a pure
precondition with no number of its own).

**Inherited, NOT claimed as new:** commit `0f9418ae5` `[layers] Backend-neutral SigmoidGLU / SigmoidAdd
layers + OpenCL whole-ops` is the **same patch** as the commit under review in **#4187** (this rung edits
both of those layers, hence the dependency). It will drop out of this diff on rebase once #4187 lands —
please merge #4187 first; it is additive and the cheapest merge in the series.

**Supersedes #4197,** closed in favour of this PR: `trackCtx-chunked-prefill` is a **git ancestor** of this
branch (`git rev-list` = 0 commits not contained), so its content is fully here and its documentation is
folded into the sections above.

**Not included:** the KV window ring (#4200) and the flash-attention default flip (#4211).

**Validated:** with both halves the 32K cell completes and summarizes a ~29,190-token document coherently
with the ring engaged at `cap=3072`; without the layer half, chunk 2 aborts with the two errors quoted
above — that failure/success pair is the gate. `NNTR_PREFILL_CHUNK=0` reproduces the pre-chunking output
byte-identically.

---

**Merge button: "Rebase and merge" — please do not squash.** The model half and the layer half must stay
separate commits (they are separate revert handles), and squashing would also destroy the patch identity
that lets the inherited #4187 commit drop out on rebase.

Depends-on: #4196, #4187


jijoongmoon:
### Merge-order note (dependency, not a change request on this diff)

This is the branch where a `from > 0` forward call first becomes reachable:
`[CausalLM] Chunked forward prefill: cap the activation plane at long context`
splits the prefill into blocks, and `[layers] Chunked prefill: accept a from>0
block call in FC + the four GLU layers` lets the layers take those blocks.
Before this branch every prefill call arrived with `from == 0`, so a
`skip_prefill` guard spelled `from == 0` was indistinguishable from the general
predicate. From here on it is not.

Two spellings coexist at this head:

* general (correct) — `nntrainer/layers/addition_layer.cpp:52` and
  `Applications/CausalLM/layers/per_layer_slice.cpp:43`:
  `bool is_prefill = !from || (to - from) > 1;`
* `from`-only (lagging) —
  `Applications/CausalLM/layers/rms_norm_gpu.cpp:124` (`skip_prefill && from == 0`),
  `Applications/CausalLM/layers/per_layer_slice_gpu.cpp:44` (same),
  `nntrainer/layers/llm/scalar_multiply.cpp:115` (`is_prefill = !from`),
  `nntrainer/layers/llm/scalar_multiply_gpu.cpp:64` (`from == 0`).

With chunked prefill enabled, prefill blocks 2..N arrive with `from > 0` and
`to - from > 1`. The `from`-only guards answer "not prefill" for those blocks,
so a layer that `skip_prefill` contracts to be decode-only executes on a
multi-row prefill block instead — the case the guard exists to exclude. It
fails silently (wrong numbers on a long prompt), not loudly.

So: please land the guard alignment (the `#4115` / `#4150` layer cluster)
before this PR, or fold the four one-line predicates in here. Merging this
first is what makes those four guards exploitable.

jijoongmoon:
Consolidation status — **left open deliberately; not closed.**

This PR carries human-authored review work (the merge-order note above), so it is being kept rather than folded away, pending your sign-off.

**Content status:** every commit of this PR is now contained in #4201, verified with `git range-diff` (each commit has a `=`/`!` counterpart on the #4201 tip; none unmatched). #4201 is the task-level long-context PR — chunked prefill, KV window ring, SVM KV plane, GPU prefill dispatch, `engine=` resolution, RoPE LUT cap, and the flash-attention hatch — and this PR's activation-plane numbers (32K: 12,814 MB → 4,272 MB, prefill 1085 TPS, +15%), the 4096 chunk-size sweep, and the ARM/Adreno `chunk=0` default with its 845-vs-1277 TPS measurement have all been transcribed into #4201's body.

**Your merge-order note is RESOLVED there, not lost.** It asked that the guard alignment land before chunked prefill, because prefill blocks 2..N arrive with `from > 0` **and** `to - from > 1`, so the four `from`-only guards answer "not prefill" and a `skip_prefill` decode-only layer executes on a multi-row prefill block — silently. #4201 already contains `[layers] Chunked prefill: treat every multi-row step as prefill in the lagging skip_prefill guards`, ordered **before** the chunking rungs in the same branch, and all four sites now read the general predicate:

| site | now |
|---|---|
| `Applications/CausalLM/layers/rms_norm_gpu.cpp:128` | `bool is_prefill = !from \|\| (to - from) > 1;` |
| `Applications/CausalLM/layers/per_layer_slice_gpu.cpp:48` | same |
| `nntrainer/layers/llm/scalar_multiply.cpp:134` | same |
| `nntrainer/layers/llm/scalar_multiply_gpu.cpp:80` | same |

Since #4201 lands as one rebase-and-merge, the ordering hazard the note describes cannot occur inside it.

**Your call:** close this PR as redundant (the analysis is preserved above and in #4201's body), or keep it open as the standalone chunked-prefill review. Nothing was closed here without you.

jijoongmoon:
Superseded by https://github.com/nntrainer/nntrainer/pull/4319, which carries this
content in three commits: the chunked forward prefill driver behind
`NNTR_PREFILL_CHUNK` (default off), the commit that decouples the activation plane
from the KV span so the prompt is bounded by the KV budget rather than by
`init_seq_len`, and the clamped-chunk commit that makes the default admit a long
prompt instead of truncating it to `init_seq_len`.

Verified present in that PR's diff before writing this: `CausalLM::run()`'s chunk
loop, `Transformer::prefillChunk()` as the single clamped answer every consumer
uses, and the `init_seq_len` layer property that keeps the model-side and
layer-side chunk equal to the row.

Closing in favour of #4319. If anything here is not covered, please reopen.

PR #4199, opened 2026-07-30
[CausalLM] Hoist the per-layer sliding window into a getLayerSlidingWindow() hook

jijoongmoon:
Hoists the per-layer sliding-window value into a single virtual hook. **Pure refactor, no behaviour
change** — and it is the prerequisite that makes the KV-ring rung safe.

Before: the window was computed at the property-stamping site with the `(layer_id + 1) %
SLIDING_WINDOW_PATTERN` formula in the base `Transformer`, and re-derived ad hoc from `layer_types`
inside each Gemma `createAttention()`. Any future consumer that needs "is layer *i* sliding, and how
wide" — KV-capacity sizing being the immediate one — would have to duplicate one of those formulas. A
prior tree did exactly that, duplicating the pattern formula at each consumer site, which is **wrong for
gemma2/gemma3/gemma4**: those derive sliding-ness from the config `layer_types` array, not from
`SLIDING_WINDOW_PATTERN`. A capacity computed by the pattern formula while the layer believes it is
full-attention writes absolute rows into a bounded buffer — silent corruption waiting to happen.

After, one hook: `Transformer::getLayerSlidingWindow(layer_id)` carries the expression copied verbatim
from `createAttention()` (unsigned modulo, as before), so the base graph is semantics-preserving by
construction; Gemma2/3/4 override it with their `layer_types` derivation — the exact code moved out of
their `createAttention()` bodies. Every site that needs the window reads this hook, so they cannot
disagree.

Depends-on: #4189, #4193, #4192, #4194, #4196, #4197, #4198, #4187

**New in this rung:** `[CausalLM] Hoist the per-layer sliding window into a getLayerSlidingWindow()
hook` — **one commit**.

**Please read only that commit.** This branch was cut on the attention-prefill lineage, so its diff
against `main` also replays the rungs already under review as #4193 (Gemma4 prerequisites), #4192
(GPU-path observability), #4194 (graph correctness), #4196 (SVM KV plane), #4197 (chunked prefill),
#4198 (chunked-prefill layers) and #4187 (neutral sigmoid layers). Those commits are the same work,
reviewed there; they disappear from this diff as those PRs land.

**Not included:** the ring itself — KV storage sized to the window follows as its own PR and is the
reason this hook exists. Splitting them is deliberate: the hook must land first, or the ring computes
capacities from a formula that disagrees with what the layer believes.

**Validated:** 4-model short-prompt goldens byte-identical — this commit only moves the window
derivation, it does not change any derived value.

---

Depends-on: #4189, #4193, #4192, #4194, #4196, #4197, #4198, #4187


jijoongmoon:
**Superseded by #4200**, which contains this commit and reviews it with its only consumer.

`upstream-pr/trackCtx-sliding-window-hook` is a **git ancestor** of #4200's head — verified with
`git rev-list <this head> ^<#4200 head>` = **0 commits**, so this rung's commit is literally in that PR's
history. #4200's body now carries this rung's documentation in full as "Half 1": what the hook replaces (the
`(layer_id + 1) % SLIDING_WINDOW_PATTERN` formula in the base plus the ad-hoc `layer_types` re-derivation
inside each Gemma `createAttention()`), the verbatim-move argument for why the base graph is
semantics-preserving, and the 4-model byte-identical golden result for the refactor on its own.

Why consolidate: the failure mode of splitting these two is recorded, not hypothetical. Shipped alone, the
ring sizes itself from a duplicated per-site window formula that is **wrong for gemma2/gemma3/gemma4** (they
derive sliding-ness from the config `layer_types` array, not from the pattern), and a ring computed that way
while the layer still believes it is full-attention writes **absolute rows into a `Wcap` buffer** — silent
memory corruption. The hook is the single place W is decided and the ring is its only consumer, so keeping
them in one merge makes the hook-before-ring ordering unbreakable. Attribution is unaffected: the hook is a
pure refactor with no measured number, and #4200 attributes the KV-row number to the ring commits.

Closing in favour of #4200.

PR #4200, opened 2026-07-30
[CausalLM] W-bounded attention gets W-bounded storage: one getLayerSlidingWindow() authority, then the KV ring

jijoongmoon:
A sliding-window attention layer only ever attends to its last **W** keys, and chunked prefill bounds one
launch's live key span to **W + C**. So its KV does not need `max_seq` rows — it needs a ring of `Wcap`
rows, with every absolute cache position mapping to physical row `pos % Wcap`. Full-attention layers and
full-attention models are untouched.

This PR carries **both halves of that capability**, and they must not be split — see the ordering law
below.

## Half 1: one `getLayerSlidingWindow()` authority (pure refactor, no behaviour change)

Before: the window was computed at the property-stamping site with the `(layer_id + 1) %
SLIDING_WINDOW_PATTERN` formula in the base `Transformer`, and re-derived ad hoc from `layer_types` inside
each Gemma `createAttention()`. Any consumer needing "is layer *i* sliding, and how wide" — KV-capacity
sizing being the immediate one — had to duplicate one of those formulas.

After, one hook: `Transformer::getLayerSlidingWindow(layer_id)` carries the expression copied verbatim from
`createAttention()` (unsigned modulo, as before), so the base graph is semantics-preserving by
construction; Gemma2/3/4 override it with their `layer_types` derivation — the exact code moved out of their
`createAttention()` bodies. Every site that needs the window reads this hook, so they cannot disagree.

**The ordering law, and why this is one PR.** A prior tree duplicated the pattern formula at each consumer
site, which is **wrong for gemma2/gemma3/gemma4**: those derive sliding-ness from the config `layer_types`
array, not from `SLIDING_WINDOW_PATTERN`. A ring sized from the pattern formula while the layer still
believes it is full-attention writes **absolute rows into a `Wcap` buffer** — silent memory corruption. The
hook is the single place W is decided and the ring is its only consumer, so shipping them as two PRs makes
the hook-before-ring ordering breakable by a merge-queue reorder. Shipping them together makes it
unbreakable. Attribution stays clean: the hook is a pure refactor with **no** number, and the measured
claim below belongs to the two ring commits.

## Half 2: the ring

`Wcap = (W/C + 2) * C`, i.e. a multiple of the chunk and `>= W + C`:
- a multiple of C means a C-aligned chunk write never straddles the wrap seam, so one step's write stays a
  single contiguous slice;
- `>= W + C` means the live window `[pos-W+1, pos+C)` never self-collides mod `Wcap`.

`kvRingCap()` is the single source of truth; every site — graph placeholder shape, `KVCacheManager`
allocation, `mha_core`'s write offset, the OpenCL kernels' read row — derives `Wcap` from it. Kill switch:
`NNTR_KV_WINDOW_RING=0` (it does not alter chunking; each knob owns exactly its own rung).

**This rung's own measured claim — KV rows, and nothing else.** Witness line, 28 of 35 layers sliding:

```
[kv-ring] chunk=1024 max_seq=32768 layers=35 (28 sliding W=1024 / 7 full): ENGAGED cap=3072 rows on 28/35
```

| | 16K | 32K |
|---|---|---|
| KV rows | 600,320 → 206,080 | 1,146,880 → 315,392 |
| private commit | 4051 → 3667 MB | 4895 → **3815 MB (−1080)** |

At an **equal chunk size** the ring's own throughput cost is noise: CUDA −0.1 %, Intel −1.9 % prefill. (An
earlier "the ring costs 7–11 %" reading was a chunk-size confound, not the ring.) Keep this separate from
#4198's number: that one is the **activation plane** (32K 12,814 → 4,272 MB), this one is **KV storage**.
A `NOT ENGAGED` line in the log **invalidates** any ring measurement — that is the review gate for this
rung.

The second ring commit is the SVM-plane half, and it is a real merge-semantics defect worth calling out:
`KVCacheManager::allocate` grew an SVM-pool branch that sized every layer at `max_seq_len` and returned
before reaching the ring-aware code below it. On the default GPU lane that **aborts at bind time**, because
the graph placeholders were already shaped at `Wcap`:

```
[!] FATAL ERROR: allocateAndBindKVCache: shape mismatch for layer 0
```

The SVM branch now derives both the requested byte size and the tensor height from `getLayerCap(i)`,
exactly as the host branch does, and the one-time witness line is emitted from **both** branches tagged
with which one ran.

Depends-on: #4198, #4196, #4193, #4192, #4194, #4187

## ★ Assembly fact a reviewer must know: this branch is a RE-CUT, not a descendant

This branch was cut fresh on the attention-prefill lineage (#4192's parent). Measured over that base it has
**25 own commits, and 23 of them are patch twins of work under review in other PRs of this series** — it is
**not** a git descendant of any of them. So until those PRs land and this branch is rebased, its `base:
main` diff replays them. **Please review only the three commits marked NEW below.**

| owning PR | inherited twin commits in this diff |
|---|---|
| #4193 | soft-cap consolidation, Gemma2 model class, layer-graph `engine=` stamp, `lmhead_untie`, fused geglu collapse, SVM-aware `ReshapedRMSNorm`, Gemma4-graph stamp, PLE `embedding_dtype`, `lm_head` stamp, norm kill-switch/trace, `mha_core` engine-neutral (**11**) |
| #4192 | differential probes, probe-guard extension, `NNTR_FC_DIVERT_TRACE` (**3**) |
| #4194 | prompt BOS, token-id staging zero (**2**) |
| #4196 | SVM-pool KV allocation, surviving the placeholder fill, **plus** the generic-FFN gate/up commit in its pre-re-anchor form (**3**) |
| #4198 | chunked forward prefill, the `from>0` block-call layer half (**2**) |
| #4187 | neutral SigmoidGLU / SigmoidAdd layers (**1**) |
| — | **NEW in this rung (3):** the `getLayerSlidingWindow()` hook, the KV ring, the ring-aware SVM-pool sizing |

**Merge order is therefore load-bearing.** Land the app-chain PRs first — #4194, #4187, #4193, #4192,
#4196, #4198 — then rebase this branch. If this merges first, those PRs go empty or conflicting.

Two consequences of that ordering, stated so nobody is surprised:

1. **The inherited gate/up twin is no longer patch-identical to #4194.** #4194 was re-anchored onto `main`,
   which required dropping the two `withKey("engine", causallm_engine())` lines from that commit (the
   function does not exist on `main`); #4193 re-adds them. This branch's copy still carries them. So that
   twin collapses to nothing on rebase only once **both** #4194 and #4193 have landed — and whichever of
   those two lands second will show one small textual conflict at `createMlp`, whose intended resolution is
   "gate-first **and** stamped", the combination this branch already has.
2. **The `-Wformat` correction is not in this lineage.** #4192 ships `%u` → `%zu` for the six `size_t`
   dimension fields of the divert-trace format string; this branch carries the pre-fix copy, which
   `-Werror=format` rejects wherever it is enabled (the `TensorDim` accessors return `size_t`). Land #4192
   before running an OpenCL CI job on this branch, or rebase first.

**Supersedes #4199,** closed in favour of this PR: `trackCtx-sliding-window-hook` is a **git ancestor** of
this branch (`git rev-list <hook head> ^<this head>` = 0 commits), so its content is fully contained and its
documentation is folded into Half 1 above.

**Downstream gate:** the CUDA lane's dispatch rung (#4205) is a **union merge onto exactly this tip**
(verified: this branch is a git ancestor of it), so the entire CUDA chain — #4205, #4207, #4209, #4210 — is
gated behind this PR. That is inherent to the code (the CUDA attention dispatch needs the ring-aware cache),
not to the packaging.

**Not included:** the flash-attention hatch (#4201) and its default flip (#4211, which must merge after
this one), and int8 KV.

**Validated:** ring on/off byte-identical at a fixed chunk on a prompt that **wraps** the ring, with the
`ENGAGED` witness present; 32K completes coherently at `cap=3072`; the memory table above; and 4-model
short-prompt goldens byte-identical for the hook commit alone (it only moves the window derivation).

---

**Merge button: "Rebase and merge" — please do not squash.** Squashing this PR would collapse the
no-number refactor into the measured ring commits and, worse, destroy the patch identity that lets the 23
inherited twin commits above drop out cleanly when their owning PRs land.

rungs named in `Depends-on:`. Please review the commits listed as **NEW in this rung**; the

Depends-on: #4198, #4196, #4193, #4192, #4194, #4187


jijoongmoon:
Consolidated into #4201.

PR #4201, opened 2026-07-30
[causallm] Long context: chunked prefill, KV window ring, SVM KV plane, GPU prefill dispatch, engine= resolution, RoPE LUT cap, flash attention opt-in

jijoongmoon:
**Task-level PR: everything that makes a long prompt work and fit.** Chunked prefill blocks the activation
plane, the KV ring bounds storage to the sliding window, the shared layers learn to accept a `from > 0`
block, the RoPE LUT stops being sized to `max_position_embeddings`, the per-layer-embedding reverse-norm
leaves the host, and flash attention becomes selectable from config.

They are also **mutually load-bearing**: chunking is what first makes a `from > 0` call reachable, which is
exactly what the lagging `skip_prefill` guards used to get wrong. Split apart, these rungs produce a
default-on broken state that a bisect can land inside.

Absorbs #4187, #4192, #4193, #4196, #4200 and #4195, plus a rung that never had a PR of its own (the host
FP32 RoPE LUT cap). **#4198 stays open by request** — see "Chunked prefill" below; its content is fully
contained here and its merge-order note is **resolved** here, but it carries review work and is not closed
unilaterally.

**Merge button: "Rebase and merge" — never squash.** Every rung is a separate revert handle, and several of
them (the inert SVM allocation vs the placeholder-fill fix; the window authority vs the ring) are only
checkable because they are separable.

**Merge position:** this PR must land **before #4211** (the flash default-on flip), which remains
deliberately standalone and merges strictly last of everything.

---

## 1. Chunked forward prefill — the activation plane at long context

`NNTR_PREFILL_CHUNK=C` drives the prefill **forward** in C-token chunks instead of one M-token block. Each
chunk feeds its tokens at input row 0 and writes KV at the absolute range `[from, from+clen)`; the next
chunk attends over the accumulated cache. This is the exact token-feed pattern the decode call shape
already uses (tokens at row 0, absolute KV position via `from`), generalized from `chunk=1` to `chunk=C` —
not a new execution model.

The prefill *activation plane* is the dominant long-context memory term. It is sized to `INIT_SEQ_LEN`
because the shared activation tensors are built at `{1,1,1,INIT_SEQ_LEN}` and `resetInputDimension` is
disabled, so an unchunked whole-document prefill needs `INIT_SEQ_LEN >= the document` and the plane grows
linearly with context. Chunking decouples the two: `INIT_SEQ_LEN` becomes the activation/chunk height
(small) while `MAX_SEQ_LEN` stays the KV capacity (large).

|---|--:|--:|
| `chunk=1024` | **4,272 MB** | **1085 TPS** (+15% over the full plane) |

At long context this is not a memory-for-speed trade: it is faster **and** ~8 GB smaller, because the full
plane no longer fits comfortably. **This number is the activation plane and nothing else** — the KV-side
saving is rung 2, with its own number.

**Chunk-size default 4096 on x86/x64.** The equal-thermal ring-on sweep on Xe3 at 32K is monotone in
throughput but not in memory (1024 → 1117.1, 2048 → 1125.4, 4096 → 1136.0, 8192 → 1146.3 TPS at peak WS
2696 / 2950 / 3419 / 4511 MB), so 4096 buys +1.7% prefill for +723 MB while 8192's further +0.9 pp costs
another 1.1 GB; CUDA independently wants 4096 for tensor-core-efficient prefill GEMMs.

**ARM/Adreno defaults to 0 (no auto-chunk), for a measured reason:** the per-layer OHWI V-mirror keeps a
texture-pitch-aligned image sized to the **first** call's `cache_to`, and a second chunk whose `cache_to`
crosses that stride's rounding boundary **re-scatters every already-cached KV row of every layer**.
Measured on a packed model at a 1.2K prompt: auto-chunking a 1073-token prompt into 1024+49 gave
**845 TPS against the single-block path's 1277 TPS (−34%)**. An explicit `NNTR_PREFILL_CHUNK` is still
always honoured there.

**The layer half, which is why it cannot ship separately.** The shared activation layers encoded the old
"`from > 0` implies a single-token decode step" assumption three different ways, and every one fires on
**chunk 2** (`from = C`, `to - from = C`):

```
[!] FATAL ERROR: incremental step size is not 1
[!] FATAL ERROR: Creating shared tensor of size bigger than tensor memory
```

1. `fc_layer_cl` normalized `from` by writing `from = 0; to = 1;` — collapsing a C-token chunk to a single
   row (behind an assert that threw first). Now `to = to - from; from = 0;` — 0-based but step-preserving.
2. `fc_layer_cl`'s skip-prefill gate was `skip_prefill && from == 0`, so on chunk 2 a KV-shared projection
   did **not** skip and asked for a 1024-row view of an activation only sized for decode.
3. The four GLU layers (SwiGLU, GeGLU, SigmoidGLU, SigmoidAdd) carried the same single-row assumption on
   their incremental path.

### The lagging `skip_prefill` guards — the merge-order concern raised on #4198, resolved here

#4198's review note identified four guards still spelled `from`-only, which become exploitable exactly when
chunked prefill makes `from > 0` reachable:

| site | old spelling |
|---|---|
| `Applications/CausalLM/layers/rms_norm_gpu.cpp` | `skip_prefill && from == 0` |
| `Applications/CausalLM/layers/per_layer_slice_gpu.cpp` | `skip_prefill && from == 0` |
| `nntrainer/layers/llm/scalar_multiply.cpp` | `is_prefill = !from` |
| `nntrainer/layers/llm/scalar_multiply_gpu.cpp` | `from == 0` |

With chunked prefill enabled, prefill blocks 2..N arrive with `from > 0` **and** `to - from > 1`. The
`from`-only guards answer "not prefill" for those blocks, so a layer that `skip_prefill` contracts to be
decode-only executes on a multi-row prefill block instead — the case the guard exists to exclude. It fails
silently (wrong numbers on a long prompt), not loudly.

The note asked that the guard alignment land **first**. **In this consolidated PR it does**: the commit
`[layers] Chunked prefill: treat every multi-row step as prefill in the lagging skip_prefill guards`
precedes the chunking rungs in this branch, and all four sites now read the general predicate
`bool is_prefill = !from || (to - from) > 1;` — the same spelling `addition_layer.cpp` and
`per_layer_slice.cpp` already used. The ordering hazard the note describes cannot occur inside a single
rebase-and-merge of this PR.

**One knob, one rung.** `NNTR_PREFILL_CHUNK=0` is the sole opt-out and reproduces the legacy single-block
prefill **byte-identically**. It also disables the KV ring transitively (`kvRingCap` needs `C > 0` to bound
the live span); the ring's own opt-out `NNTR_KV_WINDOW_RING=0` does **not** alter chunking.

## 2. KV window ring — W-bounded attention gets W-bounded storage

A sliding-window attention layer only ever attends to its last **W** keys, and chunked prefill bounds one
launch's live key span to **W + C**. So its KV does not need `max_seq` rows — it needs a ring of `Wcap`
rows, with every absolute cache position mapping to physical row `pos % Wcap`. Full-attention layers and
full-attention models are untouched.

**One `getLayerSlidingWindow()` authority first (pure refactor, no behaviour change).** Before, the window
was computed at the property-stamping site with the `(layer_id + 1) % SLIDING_WINDOW_PATTERN` formula in
the base `Transformer`, and re-derived ad hoc from `layer_types` inside each Gemma `createAttention()`. Any
consumer needing "is layer *i* sliding, and how wide" had to duplicate one of those formulas.
`Transformer::getLayerSlidingWindow(layer_id)` now carries the expression copied verbatim from
`createAttention()`; Gemma2/3/4 override it with their `layer_types` derivation — the exact code moved out
of their `createAttention()` bodies.

**This ordering is load-bearing, not stylistic.** A prior tree duplicated the pattern formula at each
consumer site, which is **wrong for gemma2/gemma3/gemma4**: those derive sliding-ness from the config
`layer_types` array, not from `SLIDING_WINDOW_PATTERN`. A ring sized from the pattern formula while the
layer still believes it is full-attention writes **absolute rows into a `Wcap` buffer** — silent memory
corruption.

`Wcap = (W/C + 2) * C`, i.e. a multiple of the chunk and `>= W + C`: a multiple of C means a C-aligned
chunk write never straddles the wrap seam, so one step's write stays a single contiguous slice; `>= W + C`
means the live window `[pos-W+1, pos+C)` never self-collides mod `Wcap`. `kvRingCap()` is the single source
of truth; every site — graph placeholder shape, `KVCacheManager` allocation, `mha_core`'s write offset, the
OpenCL kernels' read row — derives `Wcap` from it.

Witness line, 28 of 35 layers sliding:

```
[kv-ring] chunk=1024 max_seq=32768 layers=35 (28 sliding W=1024 / 7 full): ENGAGED cap=3072 rows on 28/35
```

| | 16K | 32K |
|---|--:|--:|
| KV rows | 600,320 → 206,080 | 1,146,880 → 315,392 |
| private commit | 4051 → 3667 MB | 4895 → **3815 MB (−1080)** |

At an **equal chunk size** the ring's own throughput cost is noise: CUDA −0.1%, Intel −1.9% prefill. (An
earlier "the ring costs 7–11%" reading was a chunk-size confound, not the ring.) A `NOT ENGAGED` line in
the log **invalidates** any ring measurement — that is the review gate for this rung.

The SVM-plane half of the ring is a real merge-semantics defect worth calling out: `KVCacheManager::allocate`
grew an SVM-pool branch that sized every layer at `max_seq_len` and returned before reaching the ring-aware
code below it. On the default GPU lane that **aborts at bind time**, because the graph placeholders were
already shaped at `Wcap` (`[!] FATAL ERROR: allocateAndBindKVCache: shape mismatch for layer 0`). The SVM
branch now derives both the requested byte size and the tensor height from `getLayerCap(i)`, exactly as the
host branch does, and the one-time witness line is emitted from **both** branches tagged with which one ran.

**KV element width comes from `TensorDim::getDataTypeSize()`, never a local dtype test.** A hand-rolled
`dtype == FP16 ? 2 : 4` charged the 2-byte **UINT16** cache dtype — the one a build without `ENABLE_FP16`
uses — four bytes, reserving exactly **2x** the device memory the SVM KV pool needs. The fix derives the
width from a `TensorDim`, and the regression test that pins it (UINT16 → 2, FP32 → 4, FP16 → 2 when the
build has it, asserting both `getDim().getDataTypeSize()` and the resulting `bytes()` of every K and V
plane, host-only, no device required) is restored as its own commit in this branch.

## 3. The KV cache on the SVM plane, and making the flag survive a call

The first commit allocates the per-layer K/V from the `gpu-svm` `MemoryPool`, so the cache tensors
themselves report `isSVM() == true`. On its own it changed **no output**, because the flag did not survive
a single call: `NeuralNetwork::incremental_inference` wraps every input buffer in a fresh `Tensor::Map`,
which builds a brand-new `MemoryData` with `svm_allocation = false`. `TensorPool::fillPlaceholder` then
installs exactly that `MemoryData` on the placeholder and `syncDependents()` propagates it to every
dependent view — including `mha_core`'s `cache_k` / `cache_v` views. **So the SVM-ness of the pool
allocation was overwritten on every call, one call before `mha_core` ever looked at it.**

`mha_core` gates its whole GPU/flash attention chain on `svm_ok = q && k && v && o all isSVM()`. With the
flag clobbered on `k` and `v` that predicate is unsatisfiable, so attention fell through to the host GEMM —
which reads the K/V plane through host pointers while the K/V producers wrote it on the device. Both a
silent performance cliff and a coherence hazard.

The second commit makes the placeholder fill preserve the pool's SVM attribute instead of replacing it:
`svm(q/k/v/o)` reads **1001 → 1111 on all 28 attention layers** of the measured model, which is what makes
`mha_core`'s `svm_ok` satisfiable at all. That is why the pair is the minimum reviewable unit — commit 1 is
the allocation, commit 2 is what makes it observable.

**Residency is selected by a predicate, not by a backend name.** `KVCacheManager` selects on
`it->second->isSVM()` — the same `MemAllocator` predicate core itself uses — with the caveat recorded
in-comment right above it: `CudaMemAllocator` **overrides `isSVM()` to false** even though its memory is
host-coherent UVM, so a cuda run takes the host branch here by design. This also folds in the fix for a
backend-naming violation the original port carried (a `getName() == "gpu-svm"` / `getenv("NNTR_ENGINE")`
selector), so no public commit in this series contains that selector, and there is no
`getenv("NNTR_ENGINE")` anywhere in the final code.

## 4. `engine=` resolved for real: every graph node gets a GPU home

**A graph node declares where it runs, and every node that cannot yet run there is NAMED.**

- **Collapses Gemma4's `tanh_gelu` + `multiply` pairs into the fused `geglu` node** — the neutral GeGLU
  layer and its `ClComputeOps::geglu` override come from earlier rungs; this is the consumer.
- **SVM-aware `ReshapedRMSNormLayer` + gpu-context registration** as ONE class on two contexts, so the
  per-head q/k norms stop being a host segment in the middle of a device graph.
- **The per-layer-embedding lookup table must carry `embedding_dtype`, not `fc_layer_dtype`** — a real
  dtype-plumbing defect: the table is an embedding, and reading it at the FC dtype mis-sizes it for any
  package where the two differ.
- **Declares `mha_core` engine-neutral** — `ResidencyPolicy` had no populator for it anywhere in the tree,
  so the residency planner could not see the attention node at all.
- Adds an `NNTR_VNORM_HOST` kill switch and an `NNTR_RMSN_TRACE` hook (default off) so the norm can be
  bisected to the host without a rebuild.
- **Adds the Gemma2 model class (`Gemma2ForCausalLM`)** on the model registry. Gemma2's shipped configs all
  carry `"attn_logit_softcapping": 50.0`, which makes the soft-cap parse consolidation (in #4194) its hard
  prerequisite: on a tree where that parse sits in an unreachable derived override, the new model silently
  runs with its attention soft-cap **off**.
- **Stamps `engine=`.** The tag's authority is `Context::residencyEngine()` — one authority, no per-alias
  special case, no `getName() == "gpu"` predicate, no `#if` fence in any layer body. The app side resolves
  it through `causallm_engine()`: unset ⇒ `"gpu"` under `ENABLE_OPENCL`, `"cpu"` in a non-OpenCL build,
  `NNTR_ENGINE=cpu` opts out, `NNTR_ENGINE=cuda` selects cuda. Stamped at 13 sites in the base builders,
  4 in qwen3, 14 in gemma2, then the Gemma4 graph once its host-only nodes are gone, and finally the
  `lm_head`.

Every stamped type was checked to have a real gpu-context factory, because `createLayer` on a type absent
from the resolved context **throws** `exception::not_supported` — an unregistered type is *not* a silent
cpu fallback.

**Before the stamping commit, zero `withKey("engine", ...)` call sites existed anywhere in
`Applications/`** — the entire OpenCL layer-variant chain was dead code with no consumer.

## 5. MHACore GPU prefill dispatch, and naming every host bounce it can take

The two halves are one unit: "host fallbacks are NAMED, not hidden" is the acceptance criterion for a
2,637-line dispatch, and the trace is what turns a host bounce from an argument reconstructed by reading
~600 lines of gates into a 30-second measurement.

**Prerequisite fix, and the review gate for this rung.** `opencl + fp16 + transformer` did **not build** at
the earlier app rungs: `mha_core.cpp` and `deberta_attention_layer.cpp` call the all-FP16
`nntrainer::compute_kcaches` / `compute_fp16vcache_transposed` / `compute_rotary_emb_value` (and the
`softmax_row[_inplace]<_FP16>` instantiations), which the cpu_backend only builds for ARM/NEON. This
`mha_core` carries its own x86-only FP16 host implementations of those (scalar, FP32-accumulate), and the
deberta hunk drops `ENABLE_FP16` on x86 (that model is ARM-encoder-only), so the fp16-ON x86 app build is
green from here on.

### Prefill dispatch decision table

| # | condition | route |
|---|---|---|
| 0 | `use_gemm_attention == false` | host per-row dot (`compute_kcaches` + `softmax_row`) — the default until the flash plumbing below stamps it |
| 1 | `use_gemm_attention && step_size < FLASH_MIN_PREFILL` (`min_prefill_thr(head_dim)`, env `NNTR_MIN_PREFILL`), not `NNTR_CUDA_ATTN` | host `gemm_attention` — blocking + softmax bookkeeping does not pay below the threshold |
| 2 | above threshold, `NNTR_MHA_GPU=1`, Q fp16 **and** out fp16, `head_dim > 0`, `num_heads_KV > 0`, `num_heads_Q % num_heads_KV == 0` | GPU candidate; otherwise host `gemm_attention` |
| 3 | GPU candidate **and** `svm_ok` = all four of Q / cached-K / cached-V / out have `MemoryData` **and** `isSVM()` | device path; if `svm_ok` is false the non-SVM wrapper is used, i.e. a host bounce |
| 4 | device path, `NNTR_KV_IMG_ATTN=1`, `head_dim % 8 == 0`, and the layer is **not past its sliding window** | Adreno OHWI **image** attention (`qk_matmul_f16_ohwi_img`) — preempts flash on success |
| 5 | device path, otherwise | **flash** kernels (SVM buffer), which take `local_window` |
| 6 | any kernel returns false (shape, alignment, scratch, mirror init) | falls back to the host path for that call |

Row 4's window exclusion is a **defect fixed at the source** and recorded at the call site:
`qk_matmul_f16_ohwi_img` has only the causal upper-bound mask and **no window lower bound**, so a
sliding-window layer past its window would silently compute full causal attention over evicted keys —
measured, a W=512 model's 999-token Adreno prefill degenerates into word salad, severity growing with
`cache_to - W`. Those calls are routed to the flash kernels, which do take `local_window`.

**`NNTR_FC_DIVERT_TRACE`.** `ClComputeOps::fc` had **no** way to report that an FC diverted to a host path:
`dotCl_v8c` returns a bare bool and the tree had no trace hook. A host bounce is not a harmless slow path.
Every FC that leaves the GPU path now names itself. The differential probes that ship alongside are all
default-off.

## 6. SigmoidGLU / SigmoidAdd as backend-neutral layers + their OpenCL whole-ops

The dispatch shim already declared `ComputeOps::sigmoid_glu` / `::sigmoid_add` and shipped a
`CpuComputeOps` implementation, but the tree carried **no Layer that reaches them and no CL override** —
two unreachable virtuals. This arms them.

One thin backend-neutral Layer each (type `"sigmoid_glu"` / `"sigmoid_add"`), owning structure and shape
only and delegating through `in.getOps()->sigmoid_glu / sigmoid_add` — the `SwiGLULayer` shape exactly.
`sigmoid_glu` is the sigmoid-gated attention output gate (`out = sigmoid(gate) * up`); `sigmoid_add` is the
sigmoid-gated per-layer-embedding mix (`out = sigmoid(gate) + emb`). Both honour `SkipPrefill`. Four `.cl`
kernels (fp32 + fp16); the fp16 variants upcast to fp32 for the sigmoid and combine, so wide gating
intermediates do not overflow half.

Registration on cpu (`AppContext`) and gpu (`ClContext`), appended **last** in `add_default_object` with
**explicit** `int_key`s 9001/9002 — the auto `int_key` is `str_map.size() + 1` (a count), so a mid-list
insertion shifts every later auto key and can collide with an explicit one. (The collision-proof auto-key
mechanism lands in #4194; these fold into it afterwards.)

**Named cross-backend gap:** the cuda-context factory for these two layers ships in the CUDA lane PR
(#4222), alongside the matching `CudaComputeOps` overrides, registered last with the same explicit keys.
Without that pairing an `engine=cuda` graph stamping a fused gate aborts with
`Key is not found for the object. Key: sigmoid_glu`.

## 7. Per-layer-embedding reverse-RMSNorm as a whole-op virtual

The port this replaces violated the op-table contract three ways: the Layer body called the raw kernel
wrapper `rms_reverse_norm_cl_fp16` directly from inside `incremental_forwarding`; the call sat behind a
**silent** four-way `isSVM()` fall-through (control fell through with no log and no throw into a host
FP32-temp block — a hidden host bounce installed on the very path whose hidden host bounce was the original
defect); and it grew the installed `blas_kernels.h` ABI by one raw symbol.

**Motion, not modification.** The `+81`-line wrapper moves **token-identical** (whitespace-normalised diff
empty) out of `blas_kernels.cpp` into a new `nntrainer/tensor/cl_operations/rms_reverse_norm_cl_op.cpp`,
anonymous namespace: same kernel, same argument binding, same launch geometry. `blas_kernels.{cpp,h}`
return to their pre-port blobs **byte-identical**, so the raw symbol leaves the installed ABI. The Layer's
four-way SVM predicate becomes a **named reject** per leg, in the same short-circuit order. The Layer body
collapses to one `getOps()->rms_reverse_norm(...)` call.

**Its own measured claim.** The defect this pays down was not cosmetic: the per-layer-embedding model's
blank output was a host island between two GPU ops reading an undrained SVM shadow — the capture showed a
uniform `-0` plane at `layer0_ple_post_norm` where the reference had `-0.881..1.26`. With the op on the
device path the OpenCL lane matches the reference **token for token**, at **444 prefill / 20.9 decode TPS
against the reference's 500 / 20.8** — the correctness fix costs nothing measurable on decode and is within
the prefill band of a different implementation of the same kernel.

Verified on the integration tree: cpu/fp16/cl builds clean; `unittest_compute_ops_dispatch` 11/11 including
the new reverse-norm dispatch test; `unittest_layers` 1020/1020; `unittest_nntrainer_models` 125/125;
4-model golden x3 byte-identical; layer-hash 1478/1478 identical to the reference tree. The raw-`*_cl`-call
grep over app/layer bodies moves 50 lines / 10 files → **49 / 9**.

**Merge-order note:** the `blas_kernels.h` ABI restoration is only meaningful once the CL install-list
correction (#4103) has landed — the two changes move that header in the same direction, and sequencing them
the other way re-adds a raw symbol #4103 removes.

## 8. Cap the host FP32 RoPE table at the live sequence length

`apply_rotary_emb_tensor_v2()`'s FP32 branch built its cos/sin tables at the model's full
`max_position_embeddings`, while the FP16 branch sixty lines below already capped them at
`rope_lut_positions() = min(MaxTimestep, max_position_embeddings)`. The FP32 branch was not simply
overlooked — `rope_lut_positions()` was declared and defined inside `#ifdef ENABLE_FP16`, so it was **not
visible from the FP32 path at all**. The helper has no FP16 dependency, so it moves out of the guard and is
used on both paths.

Cost of the table is `2 * positions * head_dim * sizeof(float)`, and unlike the KV cache it is fully
written at build time, so every byte is resident:

| shape | uncapped | capped |
|---|--:|--:|
| positions=40960, head_dim=128 | 40.0 MB | **1.0 MB** (max_seq_len=1024) |
| positions=131072, head_dim=256 | 256.0 MB | **16.0 MB** (max_seq_len=8192) |

Measured on a CPU FP32 lane (`enable-fp16=false`), RssAnon delta across the first forward — where the table
is built lazily: QS4CX-FP32 0.6B (`max_position_embeddings=40960`, `max_seq_len=1024`) **+33.6 / +31.5 MB
→ +3.5 / +3.5 MB**; Q4_0-FP32 0.6B **+32.5 → +6.5 MB**. RssAnon understates the true table size because the
allocator reuses ~10 MB of already-resident free chunks; a sweep of `NNTR_ROPE_LUT_CAP` over
1024/2048/10240/20480/40960 reproduces the predicted linear growth once that constant is accounted for.
Generated tokens are unchanged at every cap value including the uncapped one, and an FP16-lane gemma4
package is byte-identical before and after.

## 9. Flash attention becomes selectable (still default OFF)

`mha_core`'s `UseGemmAttention` property defaults to **false**, and nothing on this tree stamped it. The
base `Transformer` had no `USE_FLASH_ATTENTION` member at all, so

```
if (use_gemm_attention && (step_size >= FLASH_MIN_PREFILL || _cuda_attn_on))
```

was **unreachable** for gemma4 / gemma2 / qwen3 and every model built by the generic
`Transformer::createAttention` — the whole GPU/flash/gemm attention subsystem **plus** its CPU
`gemm_attention` prefill was dead code, and attention always took the per-row host path. Measurable:
gemma4 prefill sits at **235 TPS @ M=999** on the pre-change tree.

This rung ships the plumbing **opt-in**: `USE_FLASH_ATTENTION` on `Transformer` (read from the
`use_flash_attention` config key, default **false**), stamped as `use_gemm_attention` at the `mha_core`
sites. Default-off is deliberate and measured — turning it on **before** the SVM KV plane survives a call
is a regression (gemma4 golden → immediate `<eos>`, because the block then declines every GPU candidate on
`svm(q/k/v/o)=1001` and lands on `gemm_attention` reading a stale K/V plane). With the plane fix now in the
same PR, the hatch is A/B-able from `nntr_config.json` **without a rebuild**.

**The flip itself is #4211 and stays separate** so that "merge strictly last of everything" survives as a
real control; keeping them apart is also what makes the flip a self-contained `git revert`.

---

## Validation

- fp16 + OpenCL build clean; the fp16-ON **x86** app build is green from the MHACore rung on (it did not
  build before).
- `svm(q/k/v/o)=1111` on all 28 attention layers of the measured model.
- 4-model short-prompt text byte-identical over 3 consecutive runs; per-layer hash stream 1478/1478
  identical to the reference tree.
- With `use_flash_attention` at its default `false`, all four measured packages are byte-identical on both
  lanes — this PR changes no default-path behaviour on a short prompt.
- Ring gate: a `NOT ENGAGED` witness line invalidates any ring measurement.

## Merge order

- Land **#4194** (main-direct inference correctness) and **#4212** (cold start) first: the
  `attn_logit_softcapping` and `lmhead_untie` rungs inherited by this lineage also land there, and arriving
  second under rebase-and-merge makes them no-ops rather than conflicts.
- Land **#4103** (CL install-list correction) before this, per rung 7's ABI note.
- **#4211 merges strictly last of everything**, after this PR.

---

branch as its base, so this PR targets `main` and its diff therefore also contains the parent rungs of the
ladder (#4105 / #4106 / #4109 / #4115 / #4116 / #4149 / #4150), which are under review in their own PRs and
will disappear from this diff as they land.


jijoongmoon:
Superseded by https://github.com/nntrainer/nntrainer/pull/4319 for the long-context
core: chunked prefill, the sliding-window KV window ring (capacity sizing, the
cache-write modulo, and ring indexing in both the OpenCL and CUDA attention
kernels), and the CUDA split-KV path that reads it.

Two of the other items listed here landed on earlier rungs of the same stack
rather than in #4319, so this is closed against the stack and not against one PR:
the SVM/residency KV plane in #4309, and the GPU prefill dispatch and flash
attention in #4312. One item is deliberately not carried anywhere in the series:
the `use_flash_attention` opt-in, whose `USE_FLASH_ATTENTION` machinery originates
in a commit that is out of scope for this upstreaming, so it has no successor and
is not claimed to.

Closing in favour of #4319. If the long-context half is incomplete there, please
reopen.

PR #4202, opened 2026-07-30
[cuda] RoPE + attention device kernels (flash core, split-KV decode, int8-KV)

jijoongmoon:
Kernel-library half of the CUDA attention path. **No consumer is wired in this rung, so no lane changes
behaviour yet.**

- `cuda_rope.{h,cpp}` — split-half RoPE rotation on device fp16 rows, matching the host
  `apply_rotary_emb_tensor_v2` layout, plus a `_dpos` variant that reads the position from the device
  `pos` buffer so a captured graph replays with an 8-byte rewrite instead of a re-record.
- `cuda_attention.{h,cpp}` — fp32 attention core; interleaved fp16 flash-style attention (causal +
  sliding-window mask, GQA, optional softcap, ring-buffer aware via `ring_cap`); split-KV **decode**
  kernels with a prewarm hook for graph-capture-safe workspace allocation; int8-KV quantize and i8
  attention kernels (compiled, **never dispatched** — unused until an int8 KV plane exists).

All entry points are plain functions taking raw device/UVM pointers and ride only the existing seams
(the NVRTC kernel cache, the stream manager's dispatch/capture, the cuBLAS handle, the
integrated-device probe) — the same pattern as the in-tree int4 FC kernels.

**Two auto-default retractions are part of this rung, and they matter for the chain's story.** The
CUDA auto-default block previously armed `NNTR_CUDA_DEV_ACT` (device-only activation pool) and
`NNTR_CUDA_M2B` (graph decode) on a discrete GPU. Both are **disarmed here**, because their correctness
dependencies are not present at this point in the chain: with host layer segments still in the graph,
`DEV_ACT=1` is a guaranteed host-loop SIGSEGV, and `M2B=1` with no device-resident feed decodes garbage.
Later rungs complete those dependencies and re-arm the levers — see the note on the device-activation
rung, which carries the matching `Revert`. Disarm-then-re-arm is the honest sequence for a
default-changing lever whose deps land in between; it is not thrash.

Depends-on: #4126

**New in this rung:** the RoPE/attention kernel library, the GPU-class auto-tuning of the CUDA env
defaults, the two auto-default retractions, and a comment cleanup.

**Not included:** every consumer (dispatch into `mha_core`, KV residency, argmax, prewarm) — later rungs;
and the element-wise/RMSNorm kernel growth, which is the immediately following PR.

**Validated:** unified CUDA + OpenCL + fp16 build compiles clean (kernels are NVRTC-compiled and
device-probed); no dispatch site exists in this rung, so every lane is byte-identical to its parent.

---

Depends-on: #4126


jijoongmoon:
**Superseded by #4205**, which contains these commits and reviews them with their first consumer.

`upstream-pr/trackC6-cuda-rope-attention-kernels` is a **git ancestor** of #4205's head — verified with
`git rev-list <this head> ^<#4205 head>` = **0 commits**, so every commit here is literally in that PR's
history. #4205's body now carries this rung's documentation in full: the `cuda_rope` / `cuda_attention`
kernel inventory including the `_dpos` graph-replay variant and the int8-KV kernels that are compiled but
never dispatched, the "rides only existing seams" argument, the GPU-class auto-tuning of the CUDA env
defaults, and the full explanation of the two auto-default retractions (`DEV_ACT`, `M2B`) together with the
PR that re-arms `DEV_ACT` (#4209).

Why consolidate: these ~1,898 lines of device code have **exactly one consumer in the entire series** — the
`mha_core` CUDA dispatch in #4205 (`cuda_rope_fp16`, `cuda_rope_fp16_dpos`,
`cuda_attention_interleaved_fp16`, `cuda_set_pos`). Reviewed alone they are unreachable device code, which is
the anti-pattern the GPU contribution checklist exists to prevent. Bisect resolution is unaffected: upstream
lands PRs with **rebase-and-merge** (main has zero merge commits in its last 300), so each of these commits
keeps its own SHA and its own revert handle on `main`.

Closing in favour of #4205.

PR #4203, opened 2026-07-30
[cuda] Grow the element-wise and RMSNorm device kernel sets

jijoongmoon:
Companion to the RoPE/attention kernel rung: the remaining small device ops the later dispatch rungs call,
ported verbatim from the device-validated implementation (a single shared NVRTC source per file, so the
compiled module strings stay byte-identical to what was numerically verified).

- **`cuda_elementwise`** — adds `geglu`, the fused sigmoid gates (`sigmoid_glu` / `sigmoid_add`),
  residual add, the graph-decode slot V-copy (writes the KV cache at the device-read position with
  per-row ring mapping: `ring_cap > 0` maps each absolute row `% ring_cap`, so multi-row chunk writes stay
  in bounds on a ring-sized sliding-layer cache), per-layer slice copy, and a two-pass on-GPU greedy
  argmax (fp32/fp16; ties resolve to the lowest index, matching `std::max_element`; capture-safe scratch).
- **`cuda_rmsnorm`** — adds the fp32 kernel and the fp16 **reverse** norm (weight applied *before* the
  norm, `out_scale` folded), plus `rms_maybe_finish`: the post-op drain becomes skippable for device-only
  destinations under `NNTR_CUDA_DRAINSKIP_RMS=1`, default **off** (unset keeps the exact previous
  behaviour).

Existing entry points keep their signatures, so the current `swiglu` / `scalar_mul` / `softcap` /
`rmsnorm_fp16` consumers are unaffected, and everything new has zero consumers in this rung.

Depends-on: #4202

**New in this rung:** `[cuda] Grow the element-wise and RMSNorm device kernel sets` — one commit.

**Not included:** all consumers; and any default change (the one new lever is default-off).

**Validated:** unified CUDA + OpenCL + fp16 build clean; kernels device-probed via NVRTC. With no
consumer wired, every lane is byte-identical to its parent.

---

Depends-on: #4202


jijoongmoon:
**Superseded by #4205**, which contains this commit and reviews it with its first consumer.

`upstream-pr/trackC6b-cuda-elementwise-rmsnorm-growth` is a **git ancestor** of #4205's head — verified with
`git rev-list <this head> ^<#4205 head>` = **0 commits**. #4205's body now carries this rung's documentation
in full: the `cuda_elementwise` additions (geglu, the fused sigmoid gates, residual add, the graph-decode
slot V-copy with its per-row ring mapping, per-layer slice copy, the two-pass on-GPU argmax with
`std::max_element` tie semantics), the `cuda_rmsnorm` fp32 kernel and fp16 reverse norm with the default-off
`NNTR_CUDA_DRAINSKIP_RMS` drain-skip, and the statement that existing entry points keep their signatures so
current consumers are unaffected.

Because several of these kernels are first consumed two and three rungs later, #4205 now also lists them
explicitly with the PR that consumes each (`cuda_argmax_*` → #4207; `cuda_geglu_fp16`,
`cuda_sigmoid_glu_fp16` / `cuda_sigmoid_add_fp16`, `cuda_add_fp16`, `cuda_slice_copy_fp16`,
`cuda_rmsnorm_fp32`, `cuda_rms_reverse_norm_fp16` → #4209), so no reviewer discovers apparently dead code.

Closing in favour of #4205.

PR #4204, opened 2026-07-30
[cuda] Register the fused sigmoid gate layers on the cuda context

jijoongmoon:
`NNTR_ENGINE=cuda` **aborts at graph construction** for any model that stamps `engine=` on the fused
sigmoid gates:

```
[!] FATAL ERROR: Key is not found for the object. Key: sigmoid_glu
```

`sigmoid_glu` (attention output gate) and `sigmoid_add` (per-layer-embedding mix) are backend-neutral
layers registered on the cpu and gpu contexts but **never on the cuda context**, so a node carrying
`engine=` for them has no factory there.

This registers both in `CudaContext::add_default_object`. The layers dispatch through `getOps()`;
`CudaComputeOps` inherits the `CpuComputeOps` bodies, and `engine=cuda` tensors are host-coherent UVM, so
the host loops are numerically correct on the managed buffers — the same host-class-on-the-cuda-context
precedent that `addition` / `scalar_multiply` already set. Registered **last** with **explicit** high
`int_key`s (9001/9002, mirroring the CL context), because the auto `int_key` is `str_map.size() + 1`, so a
mid-list insertion shifts later auto keys onto explicit ones; explicit keys are collision-checked at
registration.

Depends-on: #4202, #4187

**New in this rung:** `[cuda] Register the fused sigmoid gate layers on the cuda context` — one commit.
The neutral-layer commit and a merge commit also appear in this diff: this rung is the union of the CUDA
kernel chain and the neutral sigmoid layers PR named above, because it registers layers that PR
introduces.

**Not included:** CUDA *kernel* overrides for these two ops — this rung only makes the layer types
constructible on the cuda context. A native device implementation lands with the device-activation rung.

**Validated:** the abort above no longer reproduces; a cuda-engine graph containing these node types
constructs and runs, and the host-loop path on UVM is numerically identical to the cpu engine.

---

Depends-on: #4202, #4187


jijoongmoon:
**Superseded by #4209**, which now carries this commit and reviews it next to the device overrides it exists
to reach.

Containment was **created deliberately, then verified**: this branch's unique content — the 18-line
`CudaContext::add_default_object` registration of `SigmoidGluLayer` / `SigmoidAddLayer` with explicit
`int_key`s 9001/9002 — was **cherry-picked onto #4209's branch**, and the result is patch-identical to the
commit here (`git patch-id --stable` = `4649092d3…` on both). Verified after the pick: `sigmoid_glu` and
`sigmoid_add` are registered at #4209's tip, where before the pick that file had **zero** hits for either
type. The neutral-layer commit this branch also carried is present in #4209's diff as a patch twin and stays
owned by **#4187**, which reviews it.

Why this mattered: this branch was a **dead-end leaf** — not an ancestor of #4205, #4207, #4209 or #4210 — so
if the CUDA lane had merged as previously stacked, the `CudaComputeOps::sigmoid_glu` / `::sigmoid_add` device
overrides defined in `cuda_compute_ops.cpp` would have been unreachable forever, and any `engine=cuda` graph
that stamps a fused gate would abort at construction with `Key is not found for the object. Key: sigmoid_glu`.
One neutral layer + its whole-op + the registration that makes it selectable belongs in one PR, and #4209's
body now names that abort string and the repair explicitly.

Closing in favour of #4209.

PR #4205, opened 2026-07-30
[cuda] Device RoPE / attention / element-wise / RMSNorm kernels and their first consumer: mha_core CUDA dispatch with UVM-routed pools

jijoongmoon:
The CUDA attention path as one reviewable unit: the device kernel library **and its first consumer**.

Shipped apart, the kernel halves are ~2,460 lines of device code with **zero call sites anywhere in the
tree** — the exact "kernels without a consumer" anti-pattern the GPU contribution checklist is written
against. Verified by call-site grep: the RoPE/attention kernels have exactly one consumer in the whole
series (this rung's `mha_core` dispatch: `cuda_rope_fp16`, `cuda_rope_fp16_dpos`,
`cuda_attention_interleaved_fp16`, `cuda_set_pos`), and the element-wise/RMSNorm growth is first consumed
here too. So they are reviewed together, with the residency precondition that makes the lane reachable at
all.

## 1. Kernel library

- **`cuda_rope.{h,cpp}`** — split-half RoPE rotation on device fp16 rows, matching the host
  `apply_rotary_emb_tensor_v2` layout, plus a `_dpos` variant that reads the position from the device `pos`
  buffer so a captured graph replays with an 8-byte rewrite instead of a re-record.
- **`cuda_attention.{h,cpp}`** — fp32 attention core; interleaved fp16 flash-style attention (causal +
  sliding-window mask, GQA, optional softcap, ring-buffer aware via `ring_cap`); split-KV **decode** kernels
  with a prewarm hook for graph-capture-safe workspace allocation; int8-KV quantize and i8 attention
  kernels which are **compiled but never dispatched** (unused until an int8 KV plane exists — named here so
  nobody hunts for the consumer).
- **`cuda_elementwise`** — `geglu`, the fused sigmoid gates (`sigmoid_glu` / `sigmoid_add`), residual add,
  the graph-decode slot V-copy (writes the KV cache at the device-read position with per-row ring mapping:
  `ring_cap > 0` maps each absolute row `% ring_cap`, so multi-row chunk writes stay in bounds on a
  ring-sized sliding-layer cache), per-layer slice copy, and a two-pass on-GPU greedy argmax (fp32/fp16;
  ties resolve to the lowest index, matching `std::max_element`; capture-safe scratch).
- **`cuda_rmsnorm`** — the fp32 kernel and the fp16 **reverse** norm (weight applied *before* the norm,
  `out_scale` folded), plus `rms_maybe_finish`: the post-op drain becomes skippable for device-only
  destinations under `NNTR_CUDA_DRAINSKIP_RMS=1`, default **off**.

All entry points are plain functions taking raw device/UVM pointers and ride only the existing seams (the
NVRTC kernel cache, the stream manager's dispatch/capture, the cuBLAS handle, the integrated-device probe) —
the same pattern as the in-tree int4 FC kernels. Existing entry points keep their signatures, so the current
`swiglu` / `scalar_mul` / `softcap` / `rmsnorm_fp16` consumers are unaffected.

**Kernels added here whose consumers arrive in later rungs — listed so no reviewer discovers dead code:**
`cuda_argmax_fp32` / `cuda_argmax_fp16` → consumed by #4207 (on-GPU greedy argmax); `cuda_geglu_fp16`,
`cuda_sigmoid_glu_fp16` / `cuda_sigmoid_add_fp16`, `cuda_add_fp16`, `cuda_slice_copy_fp16`,
`cuda_rmsnorm_fp32`, `cuda_rms_reverse_norm_fp16` → consumed by #4209 (the device-activation chain).

## 2. First consumer: `mha_core` fp16 GPU dispatch + KV-cache CUDA residency

Wires the attention/RoPE dispatch into `mha_core` and gives the KV cache a device-accessible residency
option: a RoPE LUT device-upload helper (a flat `[positions * half]` table cached by table identity — the
per-call host LUT row was **74 % of API time** under async submission in the profile this port carries
over), decode Q-RoPE on device, K-RoPE written straight into the UVM/device cache slice, and the GPU V-copy.

Gating note: the cuda context's existing safe auto-default block already defaults `NNTR_CUDA_ROPE` / `_ATTN`
/ `_KV_UVM` to 1 and its comment explicitly marked them *"inert until those consumers exist"* — this rung
**is** those consumers, so on a cuda run these paths engage by default and an explicit `=0` opts out per
lever. Non-cuda lanes are untouched (everything is under `ENABLE_CUDA`, and the `setenv`s only run on a
cuda-context init).

## 3. ★ The routing defect that made everything above it inert

`neuralnet.cpp` had **no CUDA pool-routing branch**, so every tensor came from the plain CPU allocator,
`dev_accessible` was false, and *every* CUDA fast path self-disabled. This is a **residency precondition,
not a perf extra** — without it the lane is host-bound and the kernels above are unreachable.

**This is the one throughput number in this PR and it belongs to commit `3848a1bf2` `[models] Route the
tensor pools to CUDA UVM for engine=cuda graphs` alone:** measured host-FC fallback count **55 → 0**, and
short-prompt prefill **1.07 → 35.7**, **2.08 → 40.7**, **0.45 → 13.95** and **2.99 → 42.4 TPS** on the four
measured models (qwen3 decode 1.9 → 15.6). A 20–30x step that is a *routing* fix, not a kernel one.

**Cross-backend behaviour change, disclosed.** On a unified build, the OpenCL lm_head GEMV is now gated to
layers actually stamped on the CL plane (commit `778df3f3a`). Previously an `NNTR_ENGINE=cuda` run on a
machine that also has an Intel iGPU would have its lm_head hijacked by the CL path — cross-backend
contamination that silently invalidates any CUDA measurement. This is the only hunk in the PR that changes
behaviour outside the cuda lane.

## 4. The two auto-default retractions, and where they are undone

The CUDA auto-default block previously armed `NNTR_CUDA_DEV_ACT` (device-only activation pool) and
`NNTR_CUDA_M2B` (graph decode) on a discrete GPU. Both are **disarmed** in this PR's kernel commits, because
their correctness dependencies are not present at this point in the chain: with host layer segments still in
the graph, `DEV_ACT=1` is a guaranteed host-loop SIGSEGV, and `M2B=1` with no device-resident feed decodes
garbage. These are **inertness-preserving retractions, not user-visible flips**. #4209 completes the
device-activation chain and carries the matching `Revert` that re-arms `DEV_ACT`; `M2B` deliberately stays
opt-in there, on a measured eager-vs-replay result. Disarm-then-re-arm is the honest sequence for a
default-changing lever whose dependencies land in between.

Depends-on: #4200, #4126

## Narrative scope, stated deliberately

This rung claims **correctness and residency, not performance parity**. The four measured models load and
generate under `NNTR_ENGINE=cuda`; gemma4 is byte-identical cpu↔cuda; gemma2 matches a pre-existing CUDA
reference by md5 (its previous "does not work" status was *never attempted*, not a defect); qwen3 is golden.
Hatches `ROPE` / `KV_UVM` / `PREWARM` are byte-identical, and the full kill-switch set returns the
pre-dispatch golden byte-for-byte; only `ATTN=0` is numerically involved, and it lands exactly on the host
path. The performance levers that close the remaining gap are #4207, #4209 and #4210.

## New in this rung / inherited

**New in this rung (9 commits):** the RoPE/attention kernel library, the element-wise + RMSNorm kernel
growth, the GPU-class auto-tuning of the CUDA env defaults, the two auto-default retractions, a comment
cleanup, the `mha_core` CUDA dispatch, the tensor-pool UVM routing fix, and the CL lm_head engine gate.

**Supersedes #4202 and #4203,** closed in favour of this PR: both branches are **git ancestors** of this one
(`git rev-list <each head> ^<this head>` = 0 commits), so their content is contained here verbatim and their
documentation is folded into sections 1 and 4 above.

**Inherited from the OpenCL app chain, NOT claimed as new:** this branch is a **union merge** of the CUDA
kernel chain with the long-context chain at the KV-ring tip, because the attention dispatch needs both the
device kernels and the ring-aware cache. Everything under #4200 (and, through it, #4193, #4192, #4194,
#4196, #4198, #4187) is replayed in this diff and reviewed there. **Rebase this branch after that chain
lands** — otherwise its `base: main` diff replays ~49 app commits on top of the CUDA chain. Note also that
#4192's `%u` → `%zu` `-Wformat` correction is not in this lineage, so land #4192 before running an OpenCL CI
job here.

**Not included:** argmax / load-time prewarm (#4207), the device-activation chain and its default re-arm
(#4209), plain-page drop (#4209), split-KV prefill (#4210) — in that order.

**Validated:** unified CUDA + OpenCL + fp16 build clean; kernels NVRTC-compiled and device-probed; 4-model
goldens as described above.

---

**Merge button: "Rebase and merge" — please do not squash.** The single throughput number in this PR is
attributed to one commit (`3848a1bf2`), the two auto-default retractions must stay individually revertible
because #4209 reverts one of them by SHA-level intent, and squashing would destroy the patch identity that
lets the inherited app-chain commits drop out on rebase.

Depends-on: #4200, #4126


jijoongmoon:
Consolidated into #4222.

PR #4206, opened 2026-07-30
[causallm/cuda] On-GPU greedy argmax, coherent logits readback, and load-time cache prewarm

jijoongmoon:
Two per-token host costs at the lm_head boundary on the cuda lane, plus the load-time prewarm that moves
the cold-start cache builds off the first timed prefill, plus a fallback tracer.

1. **On-GPU greedy argmax** (`NNTR_CUDA_ARGMAX`, opt-in, presence-gated). `incrementalInference` stashes
   the device-accessible logits pointer before any host copy; `generate()` reduces it to a token id with
   the two-pass on-GPU argmax and reads back **4 bytes**, skipping the full-vocab D2H pass and the host
   `std::max_element`. Gated to pure greedy: sampling, repetition penalty, bad-words and the logits
   processor all mutate or consume logits on the host, and batch must be 1. On any gate or kernel failure
   it falls through to the untouched host path (the host buffer is still produced and valid).
2. **Coherent logits readback** (unconditional on cuda runs). The host fp16→fp32 convert / fp32 memcpy now
   (a) drains first under async submission so a UVM read cannot be torn, and (b) stages D2H through a host
   temp when the activation pool is device-only — a raw host read of a `cudaMalloc` pointer faults. All
   cudart touches sit behind an engine check, so a non-cuda run of a unified binary never boots the CUDA
   runtime inside this timed path.
3. **Load-time prewarm of the derived weight caches + a teardown seam.** The cold first prefill paid for
   every FC's one-time derived-cache build (plain QS4CX → packed int4 + rowsum on the GPU) plus the
   first-call scratch `cudaMalloc`s. Three lifecycle entry points over the **existing** cache machinery
   (no new build path): a pointer-keyed idempotent prewarm wrapper (the build is a GPU repack straight
   from the plain payload, so prewarming costs no transient host RSS); a scratch pre-grow to decode bounds
   so the `M=1` FC never `cudaMalloc`s inside a graph capture; and a release entry point for the
   pointer-keyed caches on model reload (correctness never depends on it — caches are keyed by weight
   pointer and lazily rebuild).
4. **`NNTR_CUDA_FC_DBG`** traces silent host-dot fallbacks in `CudaComputeOps::fc`. Same rule as the
   OpenCL divert trace: a host bounce is named, not hidden.

Depends-on: #4205

**New in this rung:** the three commits above (argmax + coherent readback; prewarm + teardown; FC
fallback trace). The rest of the diff is the union lineage under review in #4205 and its ancestors.

**Not included:** the device-activation chain and its default flip, plain-page drop, split-KV prefill.

**Validated:** `NNTR_CUDA_ARGMAX` byte-identical to the host argmax path on the measured models (it is a
tie-break-compatible reduction); prewarm is behaviour-neutral by construction (same build, earlier); the
FC trace is default-off and value-checked.

---

Depends-on: #4205


jijoongmoon:
**Superseded by #4207**, which contains these commits and reviews the whole load-and-lm_head lifecycle
together.

`upstream-pr/trackC8-cuda-argmax-prewarm` is a **git ancestor** of #4207's head — verified with
`git rev-list <this head> ^<#4207 head>` = **0 commits**, so every commit here is literally in that PR's
history. #4207's body now carries this rung's documentation in full, and — as the consolidation's own
condition — it keeps **the two measured numbers separately attributed to their commits**:

- the on-GPU greedy argmax + coherent device-logits readback (this rung) is presented as the **decode**-path
  mechanism, with the profile decode cells named and the honest note that they were measured at the top of
  the chain, so #4209's `DEV_ACT` re-arm is a co-requisite and neither PR claims them alone;
- the cuBLAS-i8 load-time prewarm (the next rung) is presented as **the 1K prefill lever** with its own
  table (118–124 % of every same-package bar);
- `NNTR_CUDA_FC_DBG` keeps its role as this PR's named-fallback disclosure (most `CudaComputeOps::fc` ops
  still run host loops over UVM).

Why consolidate: both rungs mutate the same seam — `cuda_fc_qint4.{cpp,h}` plus the `causal_lm.cpp`
prewarm/teardown call — and split, the tree has an intermediate state where the prewarm covers only half the
derived cache families. Bisect resolution is unaffected: upstream rebase-and-merges, so each commit keeps its
own SHA and revert handle on `main`.

Closing in favour of #4207.

PR #4207, opened 2026-07-30
[cuda] Build both derived weight-cache families at load (CPU-side) + on-GPU argmax, with the silent host-dot fallback named

jijoongmoon:
One authority for one lifecycle: **both** derived weight-cache families are built at load instead of one,
and the two per-token host costs at the lm_head boundary are removed. Both halves edit the same seam —
`cuda_fc_qint4.{cpp,h}` plus the `causal_lm.cpp` prewarm/teardown call — so split they leave an intermediate
state in the tree where the prewarm covers only half the derived caches.

## What is in this PR, with the numbers attributed per commit

**A. `[causallm/cuda] Load-time prewarm of the derived weight caches + teardown seam`** — three lifecycle
entry points over the **existing** cache machinery (no new build path): a pointer-keyed idempotent prewarm
wrapper; a scratch pre-grow to decode bounds so the `M=1` FC never `cudaMalloc`s inside a graph capture; and
a release entry point for the pointer-keyed caches on model reload (correctness never depends on it — caches
are keyed by weight pointer and rebuild lazily). Behaviour-neutral by construction: same build, earlier.

**B. `[CUDA] Prewarm the cuBLAS int8 weight cache at load too (CPU-side, chunked)` — this is THE 1K prefill
lever, and the number below belongs to this commit alone.** It moves **272 lazy per-weight cuBLAS-i8 `[K,N]`
cache builds** — each with device-syncing `cudaMalloc`s that serialize the GPU pipeline — out of the **first
timed prefill** into load, CPU-side and 64 MB-chunked. Profiler-attributed as the dominant prefill
wall-time gap (prefill *kernel* time was already at parity). Side effect: because the plain UVM payload is
only ever host-read at load, its pages stop being migrated wholesale to VRAM. `skip_prefill` towers and an
untied lm_head are exempted from the eager i8 build (their cache would be dead VRAM); the lazy build remains
as the self-healing fallback.

**Measured** (RTX 5060 8 GB, 1K prompts, M=1073/999/1075/1107, one discarded warm-up + best-of-2 with 45 s
cooldowns, idle-witnessed, golden-gated per cell):

| model | baseline | defaults | profile | pre-series bar | % of bar |
|---|---|---|---|---|---|
| an internal 35-layer sliding-window model | 220.9 | 5474.5 | **6062.2** | 5014.0 | **121 %** |
| gemma4 (qs4cx_fp16)¹ | 203.0 | 9000.0 | **9990.0** | 8121.9 | **123 %** |
| gemma2 | 191.1 | 3445.5 | **3732.6** | 3161.8 | **118 %** |
| qwen3 | 1058.3 | 5562.8 | **6116.0** | 5125.0 | **119 %** |

GPU memory (whole-GPU peak delta) is at or under every bar (−0.1…−0.7 %; qwen3 −0.2 %, previously +10 %) —
the CPU-side prewarm is what keeps the plain pages host-resident. 32K (chunk 1024, KV ring, 29,190-token
prefill): prefill **2496.4** vs bar 2474.1–2488.2 (100.6 %), decode 32.45 vs 32.1–32.3 (met).

**E. `[cuda] Gate the load-time cuBLAS-i8 prewarm on the actual prompt length` — this fixes a defect
in B, measured.** B builds the cuBLAS-i8 `[K,N]` cache for **every** prefill-capable QS4CX FC at
load. Its only consumer is the dispatcher's `M >= 32` branch — and a normal turn prefills far below
that (the golden 20-token turn prefills at M=20/21), so on every short turn that build is **provably
dead work**, paid before the first token where the user is waiting. The prompt is already tokenized
when the prewarm starts, so the largest M the turn can ever present is known for free
(`min(prefill_chunk, INIT_SEQ_LEN)` when the prompt exceeds a chunk, else the prompt length); the
**eager** i8 build is gated on that maximum. `M >= 32` also stops being a literal in two places:
`cuda_fc_qint4.h` owns `CUDA_FC_I8_PREFILL_MIN_M` and both the dispatcher and the prewarm read it, so
they cannot drift into "build a cache nothing reads".

**Measured** on the same box (a 3.5B-class untied-lm_head QS4CX package, 20-token golden turn, quiet
host, 5 gated runs vs 4 control runs with `NNTR_CUDA_PREWARM_I8=1` = pre-gate behaviour, same
binary):

| metric | pre-gate | gated | delta |
|---|--:|--:|--:|
| prewarm lap | 1503 ms | **332 ms** | −1171 |
| full-span TTFT, card active | 3312 ms | **2138 ms** | **−1174 (−35 %)** |
| full-span TTFT, card suspended | 5510 ms | **4350 ms** | −1160 |
| e2e median (active / suspended) | 3946 / 6165 ms | **2778 / 4974 ms** | −1168 / −1191 |
| VRAM in use at the first token | — | — | **−2.6 GB** of dead i8 caches |
| peak RSS | 4431.7–4432.1 MB | unchanged | the caches are VRAM, not RSS |

**Long-generation runs keep today's behaviour** — that is why only the i8 half is gated:

- With a 1.2K prompt the gate reports `eager i8 ON` and the prewarm lap is **1497–1505 ms**,
  byte-for-byte the pre-gate behaviour, at prefill 5457 TPS and decode 42.5–42.7 TPS (3 runs). The
  case where the prewarm pays is untouched.
- Turning the **whole** prewarm off instead costs **45.0 → 31.4 TPS** of decode and only breaks even
  at **≈136 generated tokens**. This gate does not pay that price: it keeps the dp4a pack, the fp16
  scale buffers and the split-KV / decode scratch — which are what buy decode throughput — so the
  decode band is unchanged and there is no break-even to argue.

Safety: the gate skips only an **eager** build. The lazy in-path build still runs on first use (a
device-side repack, no host transient), so a later long-prompt turn self-heals at that one prefill.
`NNTR_CUDA_PREWARM_I8=1` restores the pre-gate eager build (the A/B control arm), `=0` disables it
entirely, unset = length-gated. Golden byte-identical on every run of both arms.

**C. `[causallm/cuda] On-GPU greedy argmax + coherent device-logits readback` — a different number, decode
path only.** `incrementalInference` stashes the device-accessible logits pointer before any host copy;
`generate()` reduces it to a token id with the two-pass on-GPU argmax and reads back **4 bytes**, skipping
the full-vocab D2H pass and the host `std::max_element`. Gated to pure greedy (`NNTR_CUDA_ARGMAX`, opt-in,
presence-gated): sampling, repetition penalty, bad-words and the logits processor all mutate or consume
logits on the host, and batch must be 1; on any gate or kernel failure it falls through to the untouched
host path. The **coherent logits readback** half is unconditional on cuda runs: the host fp16→fp32 convert /
fp32 memcpy now (a) drains first under async submission so a UVM read cannot be torn, and (b) stages D2H
through a host temp when the activation pool is device-only — a raw host read of a `cudaMalloc` pointer
faults. All cudart touches sit behind an engine check, so a non-cuda run of a unified binary never boots the
CUDA runtime inside this timed path.

This is the lever behind the **decode** column of the profile configuration
(**64.00 / 58.11 / 65.31 / 120.30 TPS**, model order as in the table above, = 103 % / 87 %¹ / 108 % / 114 %
of the pre-series bars; defaults column 41.56 / 37.15 / 51.45 / 81.01). Stated honestly: those cells were
measured at the top of the CUDA chain, so the device-activation chain (#4209) is a **co-requisite** for them
— both PRs name the same decode numbers and **neither claims them alone**. What this commit owns is the
mechanism: per token, one 4-byte readback instead of a full-vocab D2H plus a host reduction.

**D. `[cuda] NNTR_CUDA_FC_DBG: trace silent host-dot fallbacks in CudaComputeOps::fc`** — the named-fallback
instrument, and this PR's own architecture disclosure: **most ops in `CudaComputeOps::fc` still run host
loops over UVM**, and this is what makes that measurable rather than argued. Same rule as the OpenCL divert
trace: a host bounce is named, not hidden. Default-off and value-checked.

## Two things a reviewer should not have to discover

1. **The "profile" column is not the shipped default.** `NNTR_CUDA_ASYNC` stays **off** by default per this
   series' determinism contract (#4125); "profile" adds `NNTR_CUDA_ASYNC=1 NNTR_CUDA_ARGMAX=1`. Every
   profile cell is byte-identical and golden-clean, so it is a **reference-equivalent configuration** — read
   it as a ceiling, not as what a user gets out of the box. The "defaults" column is the shipped state.
2. **Commit B moves work onto the host CPU at load.** The prewarm changes from a GPU repack into CPU-side
   `ThreadManager` work with 64 MB chunked staging. On this box that is a net win at every cell, but it adds
   **load-time CPU cost on WDDM and Tegra, where it has never been measured**. If either matters to a
   reviewer, commit B is the single revert handle.

¹ The bar row is a different package (`lmint4`); measured here on the heavier `qs4cx_fp16` because `lmint4`
predates this lineage's weight-layout contract and fails to load — that is a packaging/loader rung, **not**
included here. Its prefill nonetheless exceeds even the `lmint4` bar.

Depends-on: #4205

**New in this rung:** the argmax + coherent readback (C), the load-time prewarm + teardown seam (A),
the `NNTR_CUDA_FC_DBG` trace (D), a `-Wunused-but-set-variable` fix for non-CUDA builds, the
cuBLAS-i8 load-time prewarm (B), and the i8 prewarm length gate (E) that fixes B's short-turn dead
work — plus the small follow-up fixes since landed on this branch, each carrying its own DCO-signed
rationale.

**Supersedes #4206,** closed in favour of this PR: `trackC8-cuda-argmax-prewarm` is a **git ancestor** of
this branch (`git rev-list <#4206 head> ^<this head>` = 0 commits), so its content is contained here verbatim
and its documentation is folded into sections A, C and D above.

**Inherited, NOT claimed as new:** the whole CUDA + long-context union lineage under #4205 (and, through it,
#4200 and the OpenCL app chain). Rebase after those land.

**Not included:** the device-activation chain and its `DEV_ACT` re-arm (#4209), plain-page drop (#4209),
split-KV prefill (#4210); the loader/packaging rung for the alternate gemma4 package; the per-layer-KV memory
folds; the heap-bypass host-weight staging chain.

**Validated:** `NNTR_CUDA_ARGMAX` is byte-identical to the host argmax path on the measured models (it is a
tie-break-compatible reduction); prewarm is behaviour-neutral by construction; the FC trace is default-off
and value-checked; every cell above is golden-gated.

---

**Merge button: "Rebase and merge" — please do not squash.** This PR carries **two distinct measured
levers** (the 1K prefill number on commit B, the decode mechanism on commit C) plus the load-time CPU-cost
trade that reviewers may want to revert independently. Squashing destroys exactly that attribution.

Depends-on: #4205


jijoongmoon:
Consolidated into #4222.

PR #4208, opened 2026-07-30
[cuda] Device-activation chain completion + re-arm the NNTR_CUDA_DEV_ACT auto-default

jijoongmoon:
Completes the device-resident forward chain on the CUDA context and **then** re-arms
`NNTR_CUDA_DEV_ACT` (the device-only activation pool) on the discrete-GPU profile.

4. An `nntr_env_on` value-check sweep for the new gates, so `=0` actually disables them, plus the
   async-drain seam.
5. An `NNTR_RMSNORM_CUDA_OFF` row-cap lever (`=all` forces all rows). The auto-profile had been setting
   this lever expecting it to exist; without it a 1K prefill under `DEV_ACT` is a **guaranteed host-loop
   SIGSEGV**.
   default).
7. **`Revert "[cuda] Do not auto-default NNTR_CUDA_DEV_ACT: host layer segments remain"`.**

### On that revert — please read before flagging it as thrash

The reverted commit is in this same chain, in the RoPE/attention-kernels rung (#4202). The sequence is
deliberate and each step is the honest state of the tree at that point:

- Early in the chain the auto-default block armed `DEV_ACT` on a discrete GPU **while host layer segments
  still existed in the graph** (`reshaped_rms_norm`, `per_layer_slice`, the sigmoid gates). In that state
  `DEV_ACT=1` is not a performance lever, it is a crash: a host loop dereferences a `cudaMalloc` pointer.
  #4202 therefore **disarms** it, and its message says exactly why.
- Items 1–6 of *this* rung are the missing dependencies: every remaining host segment gets a device
  dispatch, and item 5 supplies the row-cap lever the profile assumed.
- With the dependencies present, the disarm's stated rationale no longer holds, so it is reverted rather
  than silently re-edited — the revert keeps the reason for both decisions in the history, which a fresh
  edit would erase.

Measured effect: `DEV_ACT=1` went from a **guaranteed SIGSEGV** (qwen3 faults in the first q-norm host
loop) to **byte-identical across 4 models**. It is the decode-chain enabler for the numbers in #4207:
decode 41.56 / 37.15 / 51.45 / 81.01 at defaults and 64.00 / 58.11 / 65.31 / 120.30 at profile =
103 % / 87 %¹ / 108 % / 114 % of the pre-series bars.

The graph-decode lever (`M2B`) deliberately **stays opt-in**: measured on this lineage the eager
device-resident walk outruns the graph replay (replay-only ceiling 122.1 ≈ eager+async 121.2 vs
M2B+host-feed 61.4 TPS on qwen3 decode). That decision is gated on the eager-vs-replay measurement, not on
an assumed launch-bound dispatch; `M2B=1` is byte-identical ×4 as an opt-in.

Depends-on: #4207

**New in this rung:** the seven commits above. The rest of the diff is the union lineage under review in
#4207 and its ancestors.

**Not included:** an `M2B` auto-default; the alternate-package loader rung; the per-layer-KV folds; the
heap-bypass chain; the OpenCL reverse-norm GPU path (#4195); a device redirect for the sidecar-LUT
embedding path (owned by #4186).

¹ cross-package only — see the package footnote in #4207.

---

Depends-on: #4207


jijoongmoon:
**Superseded by #4209**, which contains these commits and — importantly — pairs them with the registration
they need.

`upstream-pr/trackC10-cuda-device-activation-chain` is a **git ancestor** of #4209's head — verified with
`git rev-list <this head> ^<#4209 head>` = **0 commits**, so every commit here is literally in that PR's
history. #4209's body now carries this rung's documentation in full: the seven device-dispatch/lever items,
the host segments they close, the `NNTR_RMSNORM_CUDA_OFF` row-cap lever the auto-profile had assumed, the
full "please read before flagging it as thrash" explanation of the `DEV_ACT` disarm→re-arm sequence with the
SIGSEGV→byte-identical-×4 measurement, and the eager-vs-replay result behind keeping `M2B` opt-in.

The consolidation also fixed a real defect that this stacking would have shipped: at this branch's tip
`cuda_compute_ops.cpp` defines the `sigmoid_glu` / `sigmoid_add` device overrides while `cuda_context.cpp`
registers **neither layer**, and the only carrier of the 18-line registration (#4204) was a dead-end leaf that
is not an ancestor of C7…C12. Merged as previously stacked, those overrides were unreachable forever and any
`engine=cuda` graph stamping a fused gate aborted with `Key is not found for the object. Key: sigmoid_glu`.
That registration has been cherry-picked onto #4209's branch, so the capability is complete in one PR.

Closing in favour of #4209.

PR #4209, opened 2026-07-30
[cuda] Complete the device-activation chain: every layer gets a device op, the fused gates get a cuda factory, and DEV_ACT is re-armed

jijoongmoon:
Completes the device-resident forward chain on the CUDA context, **repairs a half-wired capability that
would otherwise ship broken**, then re-arms `NNTR_CUDA_DEV_ACT` (the device-only activation pool) on the
discrete-GPU profile — and adds the opt-in plain-page drop that belongs to the same residency story.

## ★ The half-wiring this PR repairs — read this first

At the tips of the previous CUDA rungs, `nntrainer/cuda/cuda_compute_ops.cpp` defines
`CudaComputeOps::sigmoid_glu` and `::sigmoid_add` device overrides, while `nntrainer/cuda_context.cpp`
registers **neither layer** (grep for `Sigmoid` there: zero hits). The only carrier of the 18-line
registration was a **dead-end leaf branch that was not an ancestor of any later CUDA rung**. Merged as the
lane was previously stacked, the two device overrides would have been **unreachable forever**, and any
`engine=cuda` graph that stamps a fused gate aborts at construction with:

```
[!] FATAL ERROR: Key is not found for the object. Key: sigmoid_glu
```

So this PR now carries that registration: `SigmoidGluLayer` / `SigmoidAddLayer` in
`CudaContext::add_default_object`, appended **last** with **explicit** `int_key`s 9001/9002 mirroring the CL
context — because the auto `int_key` is `str_map.size() + 1`, so a mid-list insertion shifts later auto keys
onto explicit ones; explicit keys are collision-checked at registration. One neutral layer + its whole-op +
the registration that makes it selectable is **one** reviewable unit.

`sigmoid_glu` is the sigmoid-gated attention output gate; `sigmoid_add` is the sigmoid-gated
per-layer-embedding mix. The layer classes themselves are under review in #4187.

**Minor ordering note, disclosed:** the registration commit lands *after* the `CudaComputeOps` overrides in
commit order. It is purely additive (nothing between the two commits constructs one of these node types on
the cuda context), so the intermediate state is inert rather than broken; reordering would mean rewriting the
two rungs beneath it.

## The chain itself

4. An `nntr_env_on` value-check sweep for the new gates, so `=0` actually disables them, plus the async-drain
   seam.
5. An `NNTR_RMSNORM_CUDA_OFF` row-cap lever (`=all` forces all rows). The auto-profile had been setting this
   lever expecting it to exist; without it a 1K prefill under `DEV_ACT` is a **guaranteed host-loop
   SIGSEGV**.
   default).
7. The `SigmoidGlu` / `SigmoidAdd` cuda-context registration (above).
8. `Revert "[cuda] Do not auto-default NNTR_CUDA_DEV_ACT: host layer segments remain"`.
9. `NNTR_CUDA_DROP_PLAIN` (opt-in) — see the honest scope note below.

**Host segments closed by this chain** — the point of items 1-3 and 6 is that the graph stops bouncing
between device and host per layer: residual add, per-layer slice, q/k/v-norm, the fused sigmoid/geglu
whole-ops, the reverse norm, and the embedding H2D staging all gain a device path (or a pinned-staging path)
on the cuda context. What remains host-side is named by `NNTR_CUDA_FC_DBG` from #4207.

## On the revert — please read before flagging it as thrash

The reverted commit is in this same chain, in the RoPE/attention-kernels rung now under review as #4205. The
sequence is deliberate and each step is the honest state of the tree at that point:

- Early in the chain the auto-default block armed `DEV_ACT` on a discrete GPU **while host layer segments
  still existed in the graph**. In that state `DEV_ACT=1` is not a performance lever, it is a crash: a host
  loop dereferences a `cudaMalloc` pointer. #4205 therefore **disarms** it, and its message says exactly why.
- Items 1-6 of *this* rung are the missing dependencies, and item 5 supplies the row-cap lever the profile
  assumed.
- With the dependencies present, the disarm's stated rationale no longer holds, so it is **reverted** rather
  than silently re-edited — the revert keeps the reason for both decisions in the history, which a fresh edit
  would erase.

Measured effect: `DEV_ACT=1` went from a **guaranteed SIGSEGV** (qwen3 faults in the first q-norm host loop
of the fp16 `rms_norm` fallback) to **byte-identical across 4 models**. It is the decode-chain co-requisite
for the decode column reported in #4207 (41.56 / 37.15 / 51.45 / 81.01 at defaults and
64.00 / 58.11 / 65.31 / 120.30 at profile = 103 % / 87 %¹ / 108 % / 114 % of the pre-series bars) — that
column is named in both PRs and claimed by neither alone.

The graph-decode lever (`M2B`) deliberately **stays opt-in**: measured on this lineage the eager
device-resident walk outruns the graph replay (replay-only ceiling 122.1 ≈ eager+async 121.2 vs
M2B+host-feed 61.4 TPS on qwen3 decode). That decision rests on the eager-vs-replay measurement, not on an
assumed launch-bound dispatch; `M2B=1` is byte-identical ×4 as an opt-in.

## `NNTR_CUDA_DROP_PLAIN`: opt-in, and **inert on this lane today**

After the load-time prewarm, every derived cache (dp4a packed, optionally cuBLAS-i8, plus the fp16 scales)
exists — so the QS4CX **plain** payload is a pure pointer key. This adds `cuda_fc_qs4cx_drop_plain_pages` +
`cuda_fc_qs4cx_has_cache` and an opt-in `NNTR_CUDA_DROP_PLAIN=1` that drops its pages in place (`madvise` /
`DiscardVirtualMemory`), mirroring the OpenCL int4 path's post-build drop.

**Nobody should merge this believing it saves memory today.** On this lane the drop is only effective for
heap-backed payloads; on managed (UVM) pool pages `madvise` returns `EINVAL` and the call is a **no-op**. The
missing enabler is the key-only-dispatch + host-weight staging chain, which is **not ported** to this lane
and is a named follow-up. That is precisely why the lever is opt-in and why this disclosure is in the body
rather than only in the commit message.

For the 32K memory goal the currently-preferred configuration is instead the in-chain `NNTR_CUDA_I8_JIT`
lever (#4123): measured 32K GPU delta **4914 → 3684 MiB (−1230)** at a ~1.5 % prefill cost
(2496.4 → 2458.1 TPS), byte-identical output. The bar delta is 2780, so +904 MiB of residual remains; the
next named source for it is the per-layer-KV folds, **not** included here.

¹ cross-package only — see the package footnote in #4207.

Depends-on: #4207, #4187

**New in this rung:** 9 commits — the six device-dispatch/lever commits, the cuda-context registration of the
fused sigmoid gates, the `DEV_ACT` disarm revert, and the `DROP_PLAIN` plain-page drop.

**Supersedes #4208 and #4204,** closed in favour of this PR:
- `trackC10-cuda-device-activation-chain` is a **git ancestor** of this branch (`git rev-list` = 0 commits
  not contained), so #4208's content is here verbatim.
- #4204's unique content — the 18-line cuda-context registration — was **cherry-picked onto this branch**
  and is patch-identical to the original (`git patch-id --stable` `4649092d3…` on both), and the neutral
  layer commit it also carried is present here as a patch twin (owned by #4187). Verified after the pick:
  `sigmoid_glu` / `sigmoid_add` are registered at this tip.

**Inherited, NOT claimed as new:** the whole CUDA + long-context union lineage under #4207 → #4205 → #4200
and the OpenCL app chain. Rebase after those land.

**Not included:** an `M2B` auto-default; the heap-bypass host-weight staging chain (the named `DROP_PLAIN`
enabler); the per-layer-KV folds; the alternate-package loader rung; split-KV prefill (#4210); the OpenCL
reverse-norm GPU path (#4195); a device redirect for the sidecar-LUT embedding path (owned by #4186).

**Validated:** the `Key is not found for the object. Key: sigmoid_glu` abort no longer reproduces and a
cuda-engine graph containing those node types constructs and runs; `DEV_ACT=1` byte-identical ×4 (was a
SIGSEGV); `M2B=1` byte-identical ×4 as an opt-in; `DROP_PLAIN` default-off, and with the lever on the output
is unchanged and the drop is a no-op on UVM pages — which is the scope limitation stated above, not a
regression.

---

**Merge button: "Rebase and merge" — please do not squash.** The `DEV_ACT` revert must stay a revert, the
registration must stay separately revertible from the overrides it reaches, and `DROP_PLAIN` must remain a
one-commit removal for anyone who would rather not carry an inert lever.

Depends-on: #4207, #4187

**Cascade note.** The i8 prewarm length gate reviewed in #4207 (`[cuda] Gate the load-time cuBLAS-i8 prewarm on the actual prompt length`, −1174 ms of dead load-time work on a short turn) has been carried forward onto this branch so the stack stays consistent; it is **inherited from #4207, not claimed here**, and it will collapse out of this diff when that rung lands.


jijoongmoon:
Consolidated into #4222.

PR #4210, opened 2026-07-30
[cuda/attn] Split-KV chunked prefill for global-attention layers — default ON (32K prefill +61.9%)

jijoongmoon:
> **MERGE LAST OF THE CUDA CHAIN.** The tip commit is a **default-ON policy flip** resting on an owner
> decision plus a golden **re-baseline** (details below). Everything under it is default-inert. Land the
> rest of the CUDA chain first; merge this one last so the re-baseline is the final state, not an
> intermediate one.

Touches exactly one file (`nntrainer/cuda/cuda_attention.cpp`). Models are expressed as attention classes
(d512-global / d128-uniform), not by name.

**Mechanism.** The serial block-q prefill kernel launches one fixed 32-lane warp per (head, 4-row query
tile) that walks the **entire** causal K range. Once the growing KV exceeds L2 (context ≈ 9K) every warp
streams K/V from DRAM independently and the measured per-key cost plateaus ~1.96x higher — **86.3 % of the
d512-global 32K prefill wall lives in that kernel**. This rung extends the existing decode-only split-KV
pattern to chunked prefill: `attn_blockq_split_d{128,256,512}` walk a **fixed uniform** K partition
(`n_splits = ceil(N_kv/split_len)`, cap 32, deterministically re-derived) with the identical register
online-softmax body and write `(m, l, acc[d])` partials; `attn_blockq_split_reduce` flash-merges per row in
**fixed split order**, serial loops, **no atomics** — byte-stable run to run. Tiles are slabbed under a
64 MiB scratch budget (`NNTR_CUDA_SPLITKV_PREFILL_MB`); slab boundaries never touch numerics, and any
failure (including a misaligned K/V cache base) falls back to the serial kernel.

**Why the second commit exists (falsification narrative).** The first-cut split kernel recovered **0 %** —
it exactly reproduced the serial time. Standalone-probe attribution showed the walk is **latency**-bound
(~8 warps/SM from the 128-float q/acc register state; a 4x DRAM-traffic cut via shared-memory K/V staging
measured completely **flat**, ruling bandwidth out), dominated by 16 scalar u16 loads per K/V row slice.
Vectorized row-slice loads (uint2/uint4-pair; FP arithmetic order untouched, partials bit-identical by
probe) plus `__launch_bounds__(32, {24,16,12})` occupancy floors recovered it: probe d512 **320 → 169
ms/call (1.89x)**. Rejected by measurement: shared-KV staging (flat), a 2-key software pipeline, dual/quad
concurrent chains, warps-per-block packing.

**32K results** (RTX 5060, chunk 1024, ring on, async+argmax profile, idle-witnessed, warm-up + best-of-2,
serial controls same session):

| cell | serial | split-KV ON | delta |
|---|---|---|---|
| d512-global 32K prefill | 1760.3 | **2837.8–2850.3 TPS** | **+61.9 %** (162 % of the pre-series bar) |
| d512-global 32K ms/1K tok | 568.1 | 350.8 | −38 % |
| d128-uniform 32K prefill | 2489.6 | 2678.7 | +7.6 % |
| 32K decode (both) | 29.4 / 32.2–32.4 | unchanged | — |
| GPU peak | — | +64 / +42 MiB | the scratch budget |

1K cells are **byte-inert**: below the 4096-key engage threshold the serial kernel runs verbatim
(lever-on vs lever-off byte-identical), and decode (`N_q == 1`) never enters the path.

**Numerics, and the default-ON statement.** Device probe (random fp16, `N_kv=9216`, `d=128/256/512`,
softcap on/off): split vs serial differs on **0.67 % of fp16 outputs by at most 1 ulp**; against an fp64
reference both kernels are **equally accurate** (rms 6.58e-6 vs 6.58e-6). The d512-global 32K stream is
byte-identical to serial end to end. The single d128-uniform 32K divergence is **one near-tie argmax flip**
in the generated continuation (both continuations coherent), **re-baselined as the canonical golden**
following the established re-baseline precedent — with the serial stream retained as the bisect handle:
`NNTR_CUDA_SPLITKV_PREFILL=0` reproduces it byte-for-byte (`=N` sets a custom split length). The reduce is
a deterministic fixed-split serial merge with no atomics; run-to-run byte-stability verified ×3 on both 32K
classes.

**Commit structure, stated as a call.** Kernel and vectorization are kept separate (the falsification story
is the reviewable value), and the opt-in → default-ON pair is also kept separate: the opt-in commit is the
token-identity-contract record (it documents the d128 near-tie and the default-off contract decision), and
the flip commit is the golden re-baseline **decision point** — and the exact bisect handle for the one
token-stream change this rung introduces.

Depends-on: #4209

**Branch note:** this branch has been **rebased onto the current #4209 tip** so it inherits the fused-sigmoid
cuda-context registration that #4209 now carries (without it, an `engine=cuda` graph stamping a fused gate
aborts at construction). The four commits below are unchanged in content and order, and the tip is still the
default-ON flip, so a reviewer who wants this opt-in drops exactly one commit.

**New in this rung:** the four commits above. The rest of the diff is the union lineage under review in
#4209 and its ancestors — rebase again after that chain lands, and note that #4192's `%u` → `%zu`
`-Wformat` correction is not in this lineage yet, so land #4192 before running an OpenCL CI job here.

**Not included:** any change outside `cuda_attention.cpp`; no decode-path change; no memory rung.

---

**Merge button: "Rebase and merge" — please do not squash.** Four commits, four distinct decisions:
the kernel, the vectorization that is the falsification story, the opt-in token-identity record, and the
default-ON flip that is the only token-stream change here. Squashing destroys the one-commit rollback of the
flip and the one-commit removal of the re-baseline decision.

Depends-on: #4209

**Cascade note.** The i8 prewarm length gate reviewed in #4207 (`[cuda] Gate the load-time cuBLAS-i8 prewarm on the actual prompt length`, −1174 ms of dead load-time work on a short turn) has been carried forward onto this branch so the stack stays consistent; it is **inherited from #4207, not claimed here**, and it will collapse out of this diff when that rung lands.


jijoongmoon:
Consolidated into #4222.

**Merge-order note preserved:** "merge last of the CUDA chain" was an inter-PR ordering rule. It is now an intra-PR ordering inside #4222 and it holds — the split-KV enable commit `9489c35a5` sits after all four split-KV implementation rungs, and everything following it is fixes and additive opt-in levers. The whole chain lands atomically, so the golden re-baseline the flip rests on is the final state, not an intermediate one. If you would rather have split-KV land genuinely last as its own merge, say so and this PR is reopened as a separate rung.

PR #4211, opened 2026-07-30
[MERGE LAST] [causallm] Flip use_flash_attention on by default; keep the config hatch live for every model

jijoongmoon:
> **MERGE LAST.** This PR must be the **final merge of the entire series**. Landing it before the SVM KV
> plane (#4196) and the KV window ring (#4200) is a **measured regression**: forcing the flip on early
> makes gemma4 emit an immediate `<eos>`, because the attention block is then entered, declines every GPU
> candidate on `svm(q/k/v/o)=1001`, and lands on `gemm_attention` reading a stale K/V plane. The two rungs
> must land together, in this order.

Flips `use_flash_attention` on by default and stamps it node-for-node, then closes the config hatch the
flip broke.

**Why now.** The opt-in rung shipped the plumbing default-**off** precisely because the SVM KV plane did
not survive an inference call. The KV rung closed that: the plane now reports `svm(q/k/v/o)=1111` on all
28 layers. So this does what the opt-in rung's own doc comment said to do — flip the default, together
with the rungs it depends on. What it unlocks: on x86 `min_prefill_thr()` returns 1, so the
`use_gemm_attention` block now runs for every prefill step, and the whole GPU / flash / image / two-conv
attention subsystem **and** its CPU `gemm_attention` prefill stop being dead code.

**Second commit — the hatch fix (audit finding).** After the flip, the config key
`"use_flash_attention": false` stopped reaching 2 of the 4 models: the flip stamped `use_gemm_attention`
unconditionally `"true"` at gemma2's attention site and gemma4's two graph sites, while qwen3 and the
generic graph kept the `USE_FLASH_ATTENTION ? "true" : "false"` routing. So setting the key false
disabled the block for qwen3 and generic models but gemma2 and gemma4 **silently kept it on** — and a
config hatch that answers for half the models is worse than none, because a measured A/B with it lies for
the other half. This routes all three sites through `USE_FLASH_ATTENTION`. Behaviour at the default is
byte-identical (the member defaults true since the flip); the only behavioural change is that
`"use_flash_attention": false` now actually disables the block for every model — which is also the bisect
handle for this rung.

Depends-on: #4201, #4200, #4196

**New in this rung:** the flip commit and the hatch-restoration commit. The rest of this diff is the
lineage under review in #4193, #4192, #4194, #4196, #4197, #4198, #4187, #4199, #4200 and the opt-in PR
named above.

**Not included:** no kernel changes — this rung only decides which attention implementation the graph
selects by default.

**This is the only rung in the remaining set whose merge order is load-bearing.** Everything else in this
series is either default-inert or ordered by a code dependency; here the ordering *is* the correctness
argument, and the config hatch (`"use_flash_attention": false`, restored for every model by the second
commit) is the kill switch if a package regresses after the flip.

**Validated:** with the ring and SVM plane in place, 4-model short-prompt goldens hold and the flash path
is the one exercised; `"use_flash_attention": false` reproduces the pre-flip path for every model.

---

**Merge button: "Rebase and merge" — please do not squash.** The flip and the hatch restoration must
stay two commits: that pair is the entire rollback for the most regression-implicated change in the series
(`git revert` twice returns the pre-flip default and the pre-flip routing).

Depends-on: #4201, #4200, #4196


jijoongmoon:

The change here (flipping use_flash_attention on by default) is NOT carried by that series. The default flip was deliberately deferred, to be revisited after the series lands.

The branch is preserved.

PR #4212, opened 2026-07-30
[init-latency] Cold start: OpenCL kernel cache + pooled build, loader page advice + giants-first, tokenizer snapshot + async, chat-template overlap, QS4CX single zero-fill, sidecar mmap

jijoongmoon:
**Task-level PR.** One measured number: **time from process start to the first generated token.** Every
commit here removes one serialized cold-start cost. Reviewed together they read as one budget; reviewed
apart each looks like an unmotivated micro-optimization.

Absorbs #4213, #4215, #4218, #4185, #4175, #4186, plus two rungs that never had a PR of their own
(pooled eager CL program build; loader giants-first scheduling). **16 commits, all replayed clean onto
`main`.**

> **Branch-name caveat:** GitHub cannot repoint an open PR's head, so the ref is still
> `upstream-pr/track0-kernel-cache-default-on` even though the scope is now the whole cold-start
> budget. The title and this body carry the real story.

**Merge button: "Rebase and merge" — never squash.** Each lever must keep its own revert handle and its
own escape hatch.

**Base is legal for every member:** `trackInit-cl-parallel-build`, `trackInit-loader-giants-first` and
the tokenizer branch already had merge-base == `upstream/main` HEAD; the remaining three (#4185, #4175,
#4186) are ladder-free. Three absorptions are additionally proven by merge-noop (merging into this
branch yields this branch's exact tree): #4212 ⊂ cl-parallel-build, #4213 ⊂ giants-first, #4215 ⊂ #4218.

---

## Where the time goes, and which rung takes it

| phase | lever | commits |
|---|---|--:|
| OpenCL program compile | kernel binary cache default-on + pooled parallel eager build | 4 |
| weight page-in | loader `WILLNEED` advice + working `MADV_DONTNEED` + giants-first scheduling | 2 |
| tokenizer | persistent snapshot cache + async build on a side thread | 2 |
| chat template | renderer warm-up moved off-thread and actually overlapped | 2 |
| weight allocation | stop double zero-filling QS4CX payloads | 1 |
| embedding table | sidecar LUT eager read → lazy mmap page-in | 5 |

---

## 1. OpenCL: make the kernel binary cache safe, enable it, then build the rest on a pool

`enable-kernel-caching` defaulted to `false`, so every process start recompiled the whole OpenCL program
set from source and never wrote a binary.

| lane | metric | before | after |
|---|---|--:|--:|
| Adreno 840, 35-layer LLM | OpenCL program set-up | 3,961 ms | **2,493 ms (−37%)** |
| Intel Xe3 | OpenCL context bring-up | 0.48 s | **0.05 s (9.6x)** |

Memory moves the right way too: on the Adreno device peak RSS is **~781 MB lower** with caching on
(4,697 MB vs 5,478 MB) — compiling 50 programs has a larger working set than loading them. Goldens
byte-identical in both arms.

Three problems had to be fixed before the default could flip:

- **There was no writable cache location.** The only path was the CWD-relative
  `Program::DEFAULT_KERNEL_PATH`; an application whose working directory is not writable (the normal
  case for a packaged Android app) missed every read and failed every write, silently and forever.
  Resolution order is now: explicit environment override → an application-private/temporary writable
  location → the legacy CWD path, which is still *read* so existing caches are not orphaned.
  `tools/package_android.sh` passes the option through. Verified on device: 50 binaries land in the
  resolved directory and the next launch takes 50 hits.
- **Cache-directory creation used the throwing `create_directories` overload.** When the directory was
  not writable the exception escaped into `initialize()`'s `catch`, skipping the rest of the
  registration block: **72 registered layer factories instead of 76** — `concat`, `reshape`, `rmsnorm`
  and `swiglu` silently disappeared from the GPU context. Directory creation, binary write failure and
  binary rejection are all best-effort now, and a rejected binary falls back to compiling from source.
  All 76 factories present in every failure mode, including the unwritable-directory case.
- **The cache key hashed the kernel source only,** so two builds differing only in compile options (an
  fp16 `-D`, say) could be served each other's binary. The key now includes `compile_options`.

**Then the cold path itself.** Context initialization registered every eager OpenCL kernel serially on
the calling thread. With a cold binary cache *and* a cold driver cache on an Intel Xe3 iGPU, the 52
programs took **6.41–6.64 s of `clBuildProgram` (mean 6.50 s) = 72.8% of process-start-to-first-token**;
~84% of that (5.44 s) is the eager bring-up. `initBlasClKernels()`, `initAttentionClKernels()` and
`add_default_object()` become thin wrappers over three collectors, and the collected task list runs on a
pool of half the hardware threads. Layer factories are still registered afterwards, serially, in their
original order — a factory registered without an explicit integer key takes one from the current map
size, so reordering would renumber the keys.

| gate | before | after |
|---|--:|--:|
| eager build wall time (16 cores, pool of 8, 3 cold samples) | 5,444 ms | **1,234 / 1,325 / 1,239 ms (−77%)** |
| process start → first token | 7,219–8,223 ms (mean 7,640) | **3,003 / 3,021 / 3,004 ms (−61%)** |

Summed per-build time *rises* to ~10.2 s under the pool because the driver's compiler threads contend —
wall time is what the user waits for, and that is what is quoted. Warm runs unchanged to slightly
better; generated text byte-identical. `NNTR_CL_BUILD_THREADS=1` restores the serial path;
`-Denable-kernel-caching=false` restores the previous cache behaviour.

Concurrent builds make a latent cache hazard reachable, so `writeBinaryFile` now writes a temporary and
renames it into place — two builders of one program (two threads here, or simply two processes started
together, which was already possible) would otherwise interleave writes into the same cache file and a
reader could pick up the torn result.

**Logger side-fix, in the same failure class.** `Logger`'s constructor used the throwing
`create_directories` overload and then threw `std::runtime_error` if the log file could not be opened —
from inside `ml_log*()`, which every caller treats as non-throwing and several callers are `noexcept`.
On a working directory the process cannot write to, the **first log line terminated the process**. Found
while exercising a cold kernel build from a read-only working directory, where the run died before
reaching `main`'s first statement. Now uses the `error_code` overloads and degrades to a stderr notice.
Nothing diagnostic is lost — the `runtime_error` it replaces reported an empty path (its message was
built from a zero-initialized buffer nothing ever wrote to).

**Known and unchanged:** a few kernels fail to build on some Adreno drivers and never produce a cache
entry. Pre-existing; they now fall back to source compilation instead of aborting registration. Their
repeated build attempts cost ~76 ms per launch; eliminating that needs a per-process negative cache and
is left to a separate change.

## 2. Loader: page advice, a working `DONTNEED`, and giants-first scheduling

**The staging maps advised the kernel wrongly.** Loader workers each read a sequential sub-range of the
weight file, but the map was advised `POSIX_MADV_RANDOM`, which suppresses read-ahead — so every 4 KB
page faulted individually. Advising `WILLNEED` instead lets the kernel read ahead across the range each
worker actually walks. This is the headline change and it is free on both axes:

| lane | cold weight load | |
|---|--:|--:|
| Adreno 840, 35-layer model | 10,223 ms | **2,566 ms (−75%)** |
| Intel Xe3, 3.2 GB model | 17,225 ms | **4,051 ms (−76%)** |

**`posix_madvise(POSIX_MADV_DONTNEED)` was dead code.** Both loader sites called it with a comment
describing it as a peak-RSS optimization, but glibc documents `POSIX_MADV_DONTNEED` as a **no-op** — so
the tree contained a memory optimization a reader would reasonably believe works and that in fact did
nothing on Linux. Switched to `madvise(MADV_DONTNEED)`. **Be precise about what that buys: on its own,
nothing measurable** (2310 → 2307 MB peak RSS on Intel, i.e. zero) because the call sits immediately
before `munmap`, which already drops those pages. It is worth fixing because it is dead code in a place
that claims otherwise, and because it is the prerequisite for the lever below.

**Residency-gated reaper, opt-in (`NNTR_LOAD_REAP_MB=<MB>`).** With `MADV_DONTNEED` now effective,
staged pages can be dropped once consumed:

| arm | peak RSS | cold load |
|---|--:|--:|
| default (reaper off) | 2307 MB | 4051 ms |
| `NNTR_LOAD_REAP_MB=256` | **656 MB (−1651 MB)** | 4100 ms |

It ships **off** because its cost is device-specific: **+49…+109 ms on Intel Xe3** (negligible) but
**+4.1 s on Adreno 840** (prohibitive). The reason is structural — every worker maps the whole file, so
one large record trips any threshold and dropping its pages discards read-ahead other workers are still
using. On a warm page cache the result is also erratic (1234 / 2146 / 1806 MB) because the copy can
finish inside the reaper's tick. The durable fix is to stage through a single shared mapping instead of
one per worker, at which point the reaper stops fighting read-ahead and could default on; deliberately
out of scope here.

**Giants first.** The parallel loader hands nodes to workers off a shared cursor walking graph order, so
a node large enough to dominate the load starts whenever the graph happens to mention it, and the worker
that draws it runs alone for the rest of the load while every other worker waits at the join: **one
node's copy span becomes the join wall.** On a 35-layer decoder whose packed per-layer embedding is
~1.3 GB, that single node took 300–340 ms on one worker, against a barrier idle sum of 1.9–3.6 s out of
a 3.4–4.4 s busy sum and a worker-finish spread of 130–250 ms.

Serve the few nodes big enough to set the tail on their own first, largest first (threshold 128 MB of
weight payload — selects 3–4 nodes on every model measured: the packed embedding, an untied lm_head, the
input embedding); everything else keeps graph order. Sorting the *whole* queue largest-first balances
~25 ms better but starts every fat payload at once and grew the transient peak by 0.5–1.1 GB because the
staging maps then overlap — kept as `NNTR_LOAD_LPT=2` for measurement rather than made default.

Scheduling only: each node is still read exactly once by the same worker body, so bytes written and file
offsets read are unchanged, and `stable_partition` + `stable_sort` keep the order deterministic for a
given graph. `NNTR_LOAD_LPT=0` restores legacy graph order.

| entry point (warm, 6 runs, first discarded, text byte-identical every run) | before | after |
|---|--:|--:|
| command-line runner, start → first token | 809 ms | **774 ms** (join lap ~348 → ~336) |
| embedding API entry point | 792 ms | **694 ms** (join lap ~466–498 → ~390) |

Peak RSS stays inside the pre-change envelope on the CLI runner; the API entry point shows a ~200 MB
mid-load transient because the giants now overlap its tokenizer load window, steady state unchanged.

## 3. Tokenizer: snapshot cache, then build it on a side thread

`Transformer`'s constructor parsed the ~32 MB BPE `tokenizer.json` **synchronously**, before the model
factory and before any GPU bring-up — **626 ms median, 19% of a warm 3.3 s time-to-first-token** on the
reference desktop package.

The cost is not JSON text parsing, so no parser swap removes it: it is dominated by the `tokenizers`
crate's untagged+flatten deserialize buffering plus BPE construction (a full msgpack round-trip measured
~505 ms vs ~525 ms JSON — no win). So this caches the **result**, not the input.

**Safety contract — the reason this is acceptable on a correctness path:**

- **Byte-identical tokenization on a HIT.** The snapshot reconstruction mirrors the crate's own
  deserialize order (model, then components, then added tokens).
- **Silent fallback on key mismatch or corruption.** Source `(size, mtime)` mismatch, header version
  bump, short read, trailing bytes, FNV-1a payload checksum mismatch, or a loader rejection all fall
  back to the untouched `FromBlobJSON` path and schedule a rewrite. A failure is a debug log, never an
  error, and **never a different answer**.
- **Opt-out:** `NNTR_TOKENIZER_CACHE=0` disables it entirely — no read, no write.
- The (re)write runs on a background worker **after** the parsed tokenizer has been handed to the
  caller, so a MISS never adds to the critical path. The worker is a function-static `std::async` future
  rather than a detached thread, so its destructor joins at exit and a short-lived CLI process still
  persists the cache.

Cache location: `<tokenizer_file>.ntkc` next to the model, falling back to
`$XDG_CACHE_HOME/nntrainer/tokenizer/<hash>.ntkc`.

**Measured** (RTX 5060 Laptop host, 3.5B-class QS4CX package, quiet host, 4–5 runs per arm):

| arm | n | full-span TTFT | tokenizer lap | peak RSS |
|---|--:|--:|--:|--:|
| `NNTR_TOKENIZER_CACHE=0` (parse) | 4 | 2126 [2024–2147] | 597 [592–599] | 4366.2 MB |
| snapshot HIT (default) | 3 | **1713** [1709–1781] | **234** [226–240] | **4177.6 MB** |

**−413 ms full span, −363 ms on the lap, −189 MB peak RSS.** Gates, each its own run, all with
byte-identical golden output: MISS → 591 ms parse + background rewrite, and the rewritten file is
**md5-identical** to the original; post-rewrite HIT 227–241; bit-flip inside the checksummed window →
silent fallback to parse, correct output; truncation to 1/3 → same; stale key (`touch tokenizer.json`)
→ 591 then 230. A snapshot written on one host loaded unmodified on another — the point of keying it by
content rather than by build.

**Then move it off the critical path.** The snapshot cut the build to ~234 ms but left it fully
**exposed**: it ran in the constructor, before graph compile / pool commit / weight load. The build
depends on nothing those phases produce, so the constructor now starts
`std::async(std::launch::async, LoadTokenizerCached)` and `ensureTokenizer()` joins it under a mutex,
exactly once, immediately before the first tokenizer touch. `getTokenizer()` joins internally, so
external callers cannot observe an unjoined tokenizer. It **composes** with the snapshot cache rather
than duplicating it — the worker calls the same `LoadTokenizerCached()` entry point, so there is exactly
one parse and one cache decision either way.

Interleaved A/B (5 async/synchronous pairs **alternated** so GPU-clock and page-cache drift cannot land
on one arm only; goldens byte-identical on every arm):

| arm (same binary) | n | TOTAL median | delta |
|---|--:|--:|--:|
| `NNTR_TOKENIZER_ASYNC=0` | 5 | 1263 [1242–1268] | — |
| async (new default) | 5 | **1078** [964–1100] | **−185** (per-pair median −187) |

| entry | before | after |
|---|--:|--:|
| tokenizer exposed lap | 238.5 ms | **8.2 ms** (join BLOCKED **0.0 ms**) |
| CUDA context bring-up | 272.1 ms | 318.6 ms |
| peak RSS | 2200.5 MB | 2225.4 MB (**+24.8**) |
| first token / decode | 68.1 ms / 45.8–46.2 TPS | band unchanged |

**Why −185 and not −230:** the exposed lap falls by 230, but context bring-up **grows by 46 ms** — the
worker competes with `cuInit` for the same cores. Overlap credits are `max()`, not `sum()`, and that
rule bites *inside a single lever* here. The credit is the measured span (1262 → 1072), not the lap
delta.

**Memory is a real cost here: +24.8 MB peak RSS, causal, not noise.** With the build moved beside the
graph/pool/load phase, its transient peak now *overlaps* the weight-arena commit instead of preceding
it, so the process-wide maximum rises. Good trade for 190 ms on a 2.2 GB process, but it is a trade.

Exception safety: a worker failure rides the future and is re-raised at the join. `std::future::get()`
invalidates the future even when it throws, so the exception is latched in a **sticky**
`std::exception_ptr` and re-raised on **every** later `ensureTokenizer()` — otherwise a second caller
would sail past with `tokenizer == nullptr` and fault far from the cause. Join coverage audited over
every direct use of the member (`causal_lm`, `sentence_transformer`, `deberta_v2`,
`multilingual_tinybert_16mb`, `xlm_roberta`, and **both** lfm2 entry points including
`run_with_embeddings()`, which never `Encode`s but still `Decode`s through `registerOutputs()`).
`NNTR_TOKENIZER_ASYNC=0` restores the synchronous build verbatim.

**Reviewer note: `Applications/CausalLM/lib/libtokenizers_c.a` is regenerated** (36.1 → 40.2 MB) because
the snapshot payload is produced/consumed inside the Rust wrapper — three new C entry points
(`tokenizers_snapshot_from_json` / `tokenizers_snapshot_free` / `tokenizers_new_from_snapshot`). The
Windows build compiles the same crate from source (`tokenizers_c_win`), which is why `src/lib.rs` carries
the matching implementation. As with the prebuilt it replaces, the archive embeds cargo-registry paths
from the machine that built it in Rust panic-location strings; removing those needs a rebuild with
`--remap-path-prefix`. The snapshot deliberately bypasses serde for the heavy sections: BPE's custom
`Serialize` declares 8 struct fields but writes 10, which truncates vocab/merges in any length-prefixed
format, so only an explicit layout is safe.

## 4. Chat-template renderer warm-up, actually overlapped

The `minja` renderer construction costs ~170–190 ms on the path to the first token.
`ChatTemplate::Load()` now starts a side thread that builds the renderer and the first `apply()` joins
it. The join is guarded — the naive `if (warm_thread.joinable()) warm_thread.join();` outside a lock
lets two concurrent `apply()` callers both see `joinable()` and both join; that fix is folded into the
same commit, not left as a follow-up.

On the app's own CLI path the first commit alone bought **nothing**: `Load()` and `apply()` were nine
lines apart, before the model object even existed, so the warm thread was joined almost immediately and
the construction was still paid serially — and *ahead* of `initialize()`, squarely on the path to the
first token. The second commit splits the decision from the render: *which* input to use is still
resolved in place, only the render moves to just after `repack_weight()`, handing the warm thread the
same window a create/initialize/load/repack caller already gets.

Error timing is deliberately preserved: the deferred branch is taken only when a chat input exists
**and** a template loaded, and both are decided at the original point — as are the `sample_input`
lookups that throw when the key is missing. A malformed config still fails before model load. This moves
*when* the renderer is built, not what it produces.

## 5. Stop double zero-filling QS4CX weight payloads

`QS4CX_Tensor::allocate()` zeroed its payload **twice** — `new uint8_t[size()]{}` and then
`initialize() -> setZero()` — and on the only path that reaches it in an inference process both passes
are dead stores: the weight's self-owned heap payload is overwritten in full by `NeuralNetwork::load`.
The zeroing was pure init latency. This allocates uninitialized and skips `initialize()` on that path
(`putData()` is still called, so the `MemoryData` invalidate hook fires exactly as before). Page faults
are **not** saved — they move to the `read()` that fills the buffer — but the two full-arena writes are.

The argument is only as strong as the claim "every reader writes all `size()` bytes", so that was
audited on this tree and the audit is in the commit message: `TensorBase::read(std::ifstream)` /
`read(ReadSource)` write `bytes() == size()`; `copy_qs4cx` covers `size()`; `readFSU` is a no-op and the
path is gated `!enable_fsu`. A tripwire is recorded in the `qs4cxAllocUninitialized()` doc comment for
the one known partial writer that does **not** exist in this tree (a legacy on-disk QINT4 → QS4CX
transcode leaves an N/2-byte hole between the nibble and scale regions for even K), so whoever ports
that read path in cannot miss the requirement.

Measured (35-layer sliding-window QS4CX-FP16 model, Intel Xe3, OpenCL, warm, 3 runs each):
`model_graph.initialize` **731.8 ms → 8.3 ms**, whole init **1593/1572/1514 ms → 893/979/947 ms**, peak
RSS **~3.15 GB → ~1.7–2.0 GB**, loader-join time +81 ms (the faults that moved), generated text
byte-identical. Escape hatch `NNTR_QS4CX_ALLOC_ZERO=1` restores the old double zero-fill — which is also
the bisect handle for any init-latency claim built on this rung.

## 6. Embedding sidecar LUT: eager heap read → lazy mmap page-in

The sidecar infrastructure (the `quantized_lut_path` manifest keys, config wiring, unit test) is already
on `main`; this is the *consumption* half.

- **mmap the payload.** The sidecar table (raw UINT16, or the `ufixed8`/`sfixed4` JSON-manifest payload)
  was read whole into a heap vector at finalize, so a multi-hundred-MB embedding table became resident
  before the first token. Now attached with `::mmap(PROT_READ, MAP_PRIVATE)` + `MADV_RANDOM`.
  `QuantLut::bytes` is intentionally left empty on the successful mmap path; `QuantLut::payload_size()`
  is the documented size accessor regardless of backing store.
- **Actionable unsupported-datatype error.** Same dtypes rejected as before — only the message changes,
  and it now spells out what is supported and the two ways forward.
- **GGML `q4_0`/`q6_k` row payloads.** Adds a manifest datatype whose payload is the byte-identical
  Q4_0/Q6_K row table an in-bin embedding weight would hold. Decode **reuses the existing**
  `incremental_forwarding` dequant loop — `dequantize_row_q{4_0,6_K}`, the FP32-temp cast for non-FP32
  activations and the scale multiply are untouched; only the row base pointer is chosen between the
  weight tensor and the mmap'd sidecar, and the row stride is hoisted. Sidecar output is bit-exact vs
  the monolithic package by construction. Because `MADV_RANDOM` disables read-ahead, a ~1K-token prefill
  would otherwise pay ~1K serialized major faults, so the exact rows of each incoming batch are
  batch-`MADV_WILLNEED`'d up front (POSIX only; the Windows mapping pages on demand via
  `MapViewOfFile`).
- **Unit-test fix.** Two pre-existing assertions read `bytes.size()` directly and started failing once
  mmap began succeeding on a real filesystem; they now read `payload_size()`.

**Separable commit, stated up front:** the GGML row-payload commit (~123 lines) can be dropped to keep
GGML LUT support deferred; the actionable error then stands on its own. It is included because it is
pure reuse of a dequant this file already consumes (zero new dependencies) and because every sidecar
package currently in use declares a GGML LUT manifest, so without it sidecar consumption has no package
to exercise it. Owner call at review time.

Byte-identical by construction for packages without a `quantized_lut_path`: no `quant_lut`, `ggml_lut`
false, same weight rows through the same branches; the only removed work is constructing an unused
`Tensor` view.

**Expected effect, pending re-measurement on this tree:** on a prior device session the lazy page-in
measured **RSS −1.27 GB, init −41%, load residual −44%**, and the `MADV_WILLNEED` batch prefetch removed
a ~100–160 ms serialized-fault stall on NVMe. Cited as expected values; the golden requirement (sidecar
output bit-exact vs monolithic) held on those runs and is re-stated here as the gate.

---

## What the whole budget buys, attributed honestly

On the CUDA-lane reference box the *code* rungs of this campaign take a warm short turn from
**3326 → 1777 ms (−47%)** and peak RSS from 4432 → 4080 MB. Reaching the campaign's headline
**1504 ms / 2200 MB (−55% / −50%)** additionally needs a **packaging** change with no code in it: moving
the PLE and embedding tables into mmap'd sidecar payloads (`ple_file_name` / `embedding_file_name` —
runtime keys already consumed), which measured **−273 ms and −1978 MB peak RSS**, golden identical to
the canonical package. That is a recommendation for model packaging, **not part of this or any other PR
here**. Goldens are byte-identical across all 72 short-turn runs of the campaign, which share one md5.

## Merge order

The QS4CX single-zero-fill rung is also re-included inside the legacy-QINT4 compatibility branch of this
series. That is the stacked series' pre-existing shape; **land this PR first** so the copy arrives as a
no-op under rebase-and-merge rather than as a conflict.

---


jijoongmoon:
Superseded by the re-cut series against current main, for the OpenCL kernel-cache part: #4308. Closing to keep the queue reviewable — please review the new series instead.

This PR bundles eight independent cold-start items, and only the first is in #4308, so here is where each one goes rather than a bare supersession:

- **OpenCL kernel cache + pooled build → #4308, landed.** Compiled kernels are cached by default, keyed to source, options and device signature, so a binary built with different options or on a different driver is never reused. The `create_directories` call that threw inside a `noexcept initialize()` — and thereby skipped every layer registration on a read-only working directory — is now best-effort with a warning, and the allocator and ops table are installed before the registrations so a throw in that path can no longer leave a context with a null `MemAllocator`.
- Loader page advice + giants-first, tokenizer snapshot + async, chat-template overlap, sidecar mmap → the application-side PRs later in the series. Replacement bundle to follow.
- QS4CX single zero-fill → the QS4CX data-type PR, #4305.

The branch is preserved at `upstream-pr/track0-kernel-cache-default-on`, so this is reversible and nothing is lost.

On the `CHANGES_REQUESTED` review here: its items are against the application-side content, which #4308 does not carry, so they travel with those parts rather than being answered now.

PR #4213, opened 2026-07-30
[loader] Advise the staging maps WILLNEED, and make the DONTNEED hint actually work

jijoongmoon:
Two problems in the weight-loading path, plus an opt-in memory lever whose cost turns out to be
device-specific.

**1. The staging maps advise the kernel wrongly.** Loader workers each read a sequential sub-range of
the weight file, but the map is advised `POSIX_MADV_RANDOM`, which suppresses read-ahead — so every
4 KB page faults individually. Advising `WILLNEED` instead lets the kernel read ahead across the
range each worker actually walks. This is the headline change and it is free on both axes:

| lane | cold weight load | |
|---|--:|--:|
| Adreno 840, 35-layer model | 10,223 ms | **2,566 ms (−75%)** |
| Intel Xe3, 3.2 GB model | 17,225 ms | **4,051 ms (−76%)** |

**2. `posix_madvise(POSIX_MADV_DONTNEED)` is dead code.** Both loader sites call it with a comment
describing it as a peak-RSS optimization, but glibc documents `POSIX_MADV_DONTNEED` as a **no-op**, so
the tree currently contains a memory optimization a reader would reasonably believe works and that in
fact does nothing on Linux. Switched to `madvise(MADV_DONTNEED)`.

**Be precise about what that switch buys: on its own, nothing measurable.** The call sits immediately
before `munmap`, which already drops those pages — measured 2310 → 2307 MB peak RSS on Intel, i.e.
zero. It is worth fixing because it is dead code in a place that claims otherwise, and because it is
the prerequisite for the lever below, but this PR should not be read as a memory fix by itself.

**3. The residency-gated reaper (opt-in, `NNTR_LOAD_REAP_MB=<MB>`).** With `MADV_DONTNEED` now
effective, staged pages can be dropped once consumed. Peak RSS on the same 3.2 GB model:

| arm | peak RSS | cold load |
|---|--:|--:|
| default (reaper off) | 2307 MB | 4051 ms |
| `NNTR_LOAD_REAP_MB=256` | **656 MB (−1651 MB)** | 4100 ms |

Its cost is **device-specific**, which is why it ships off by default rather than on:

| lane | load-time cost of the reaper |
|---|---|
| Intel Xe3 | **+49…+109 ms** — negligible |
| Adreno 840 | **+4.1 s** — prohibitive |

The reason for the difference is structural: every worker maps the whole file, so one large record
trips any threshold and dropping its pages discards read-ahead other workers are still using. That
hurts far more on the slower device. On a warm page cache the result is also erratic (1234 / 2146 /
1806 MB) because the copy can finish inside the reaper's tick. Enable it where peak memory matters
more than load time.

The durable fix is to stage through a single shared mapping instead of one per worker, at which point
the reaper stops fighting read-ahead and could default on. That is a larger change and is left out of
this one deliberately.

### DIFF
--- a/nntrainer/models/neuralnet.cpp
+++ b/nntrainer/models/neuralnet.cpp
@@ -26,6 +26,7 @@
 #include "model_common_properties.h"
 #include <algorithm>
 #include <atomic>
+#include <chrono>
 #include <cmath>
 #include <compute_ops.h>
 #include <cstdint>
@@ -34,8 +35,14 @@
 #include <fstream>
 #include <future>
 #include <iomanip>
+#include <mutex>
 #include <sstream>
 #include <thread>
+
+#if !defined(_WIN32)
+#include <sys/mman.h>
+#include <unistd.h>
+#endif
 
 #include <activation_realizer.h>
 #include <adamw.h>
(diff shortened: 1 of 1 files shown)


jijoongmoon:
Consolidated into #4212.

PR #4214, opened 2026-07-30
[engine] How a process acquires a backend: conditional context bring-up + register custom layers through the Engine facade

jijoongmoon:
**Task-level PR: how a process acquires a backend.** Two rungs, one coupling — the engine layer stops
constructing every context unconditionally and brings up only the one the process asked for, and the
Applications stop reaching past the `Engine` facade with a downcast to register their custom layers. **The
downcast is exactly what breaks once contexts are conditional**, so reviewing them apart hides the
coupling.

Absorbs #4184 (now closed and pointed here).

**Merge button: "Rebase and merge" — never squash.** Two independent revert handles.

**Base legality:** #4184's base was the facade rung plus its parent; this branch's base is a strict superset
of that (it adds the CL infrastructure rung), so this is an up-lineage move inside one ladder, not a base
crossing. The commit cherry-picked clean. The `[Wait for #4105,4106]` prefix has been dropped from both the
commit subject and this PR's title — it was a merge-order marker, and this branch's base already carries
both.

**Depends-on: #4117** — the CUDA half of the bring-up predicate replaces an inline `getenv` that PR
introduces, so this cannot be main-direct.

---

## 1. Bring up only the compute engine the process actually asked for

A process runs **one** compute engine. The consumer layer has resolved exactly one from `NNTR_ENGINE` for a
long time (`causallm_engine()`: `cpu | cuda | gpu`, defaulting to `gpu` on an OpenCL build).
`Engine::add_default_object()` was the one place that did not agree: it constructed the OpenCL `ClContext`
**unconditionally** and the `CudaContext` behind an ad-hoc inline `getenv`. So each lane paid the other's
device bring-up — and a cpu-only run paid both.

Both are replaced with one predicate, `bringUpWanted(name, eager_env, on_by_default)`, applied
symmetrically:

| `NNTR_ENGINE` | ClContext | CudaContext |
|---|---|---|
| `cuda` | **skipped** (was brought up + ~50 dead program builds) | brought up |
| `gpu` | brought up | skipped (already true) |
| `cpu` | **skipped** (was brought up + discrete-GPU wake) | skipped |
| unset | brought up (the consumer's own default engine) | skipped |

`NNTR_CL_EAGER_CTX=1` / `NNTR_CUDA_EAGER_CTX=1` restore the unconditional bring-up (the A/B control, and the
escape hatch for a genuinely mixed process).

### Two of the three findings are defects, not just latency

1. **A cpu-engine run was paying the entire OpenCL bring-up.** One leaked-by-design `cl_context` + command
   queue + ~50 compiled CL programs, alive for the whole process, for work a cpu run can never use — plus
   the enumeration time itself (measured 2396 ms cold on the box below).
2. **The discrete GPU wakes on `gpu` and `cpu` runs, and it is not CUDA's fault.** The suspicion that an
   OpenCL run in a CUDA-enabled build pays `cuInit` is **measurably false** in this tree (no `CudaContext`
   construction, no CUDA context lap on `NNTR_ENGINE=gpu`). The ~2.4 s wake happens inside OpenCL platform
   enumeration: `clGetPlatformIDs` dlopens every vendor ICD and `/etc/OpenCL/vendors` holds the NVIDIA one.
   `NNTR_CL_DEVICE_FILTER` cannot prevent it (applied *after* enumeration); only `OCL_ICD_VENDORS` can.
   **This PR removes that wake from `cuda` and `cpu` runs by not enumerating at all; for an OpenCL run on a
   box with a discrete GPU it remains, and that is a deployment matter, stated here rather than claimed as
   fixed.**

### Measured

RTX 5060 Laptop + Intel Xe3, a 3.5B-class untied-lm_head QS4CX package, quiet host, goldens byte-identical
on every run of every arm.

| arm | before | after | note |
|---|--:|--:|---|
| CUDA run, card **active** | 2138 | **2119** | −19 ms, at the noise floor: warm, the ~200 ms of CL enumeration was mostly NVIDIA driver load that `cuInit` pays anyway, and it moved there (CUDA context lap 110 → 289 ms) |
| CUDA run, **OpenCL compiler cache cold** | **10,637** | **2007** | −8630 ms. The CL kernel line is not "0 cached" any more, it is **absent**: no CL builds at all |
| CUDA run, cold cache, card suspended | 12,476 | **4297** | −8179 ms |
| **cpu-engine run** | **19,027** | **16,917** | −2110 ms, and the discrete GPU is **never woken** (sysfs stayed/returned `suspended`; before this it flipped `suspended`→`active`) |
| OpenCL run, card suspended | 4305 / 4493 | 4258 / 4362 | unchanged — correct, this rung does not touch that lane |
| 1.2K prompt | prefill 5403–5457 TPS, decode 41.5–42.7 | prefill 5403–5430, decode 41.8 | unchanged: the init-path-only claim holds |

Acceptance was 5/5, including "mixed/fallback still correct": the consumer's own
`registerLayerFactory("gpu", …)` now reports `[Engine] gpu Context is not registered` on stderr and the
graph resolves those types on the cuda/cpu contexts, with the golden identical — so no node silently
changed engine.

### Why not lazy first-use bring-up

Two concrete things block it, and both are load-bearing:

1. **Layer factories live inside a Context.** `Engine::registerLayerFactory(engine, …)` resolves the
   Context first, and consumers register GPU-routed and CUDA-routed custom layers at model-construction
   time. With a lazy registry entry that registration is itself a "first use" and would construct — hence
   bring up — every backend the consumer registers layers for, on every run. **This is precisely the seam
   rung 2 is about.**
2. **The `MemAllocator` identity is fetched at graph-construction time**, not at first allocation:
   `NetworkGraph`'s constructor does `getRegisteredContext(engine_name)->getMemAllocator()`. A lazy entry
   would have to construct there too, collapsing the laziness for exactly the runs that matter.

The right shape is a lazy **device** inside an eagerly constructed Context (an `ensureDevice()` before the
first allocation/dispatch, with the factory table available without it) — which `CudaContext::initialize()`
already approximates by returning early and `ClContext` lacks, since its `cl_context`, queue and ~50
`clBuildProgram` calls all happen in `initialize()`. That is a larger core refactor and is not attempted
here.

### Landmines respected

Registry semantics untouched (an unregistered engine name still resolves to `cpu` in
`parseComputeEngine`); no new `ClContext::Global()` call from inside any bring-up — the gate is a `getenv`
**before** the singleton touch, so the known self-deadlock path is not introduced; not one `registerFactory`
added or reordered, so explicit-key ordering is untouched; the predicate sits behind the same
`ENABLE_OPENCL`/`ENABLE_CUDA` guards as its two call sites, so a neither-backend build has no unused
function.

## 2. Register custom layers through the Engine facade, not a downcast

19 app files registered their custom layers by downcasting the context:

```cpp
auto app_context = static_cast<nntrainer::AppContext *>(ct_engine.getRegisteredContext("cpu"));
app_context->registerFactory(nntrainer::createLayer<X>);
```

This converts them to `ct_engine.registerLayerFactory("cpu", nntrainer::createLayer<X>)`. Semantics are
**identical, not merely similar**: `AppContext::registerLayerFactory` forwards straight to
`registerFactory<nntrainer::Layer>`, which is the exact overload the downcast resolved to — every converted
site passed `nntrainer::createLayer<X>`, whose type is `PtrFactoryType<Layer>`, so `T` was already deduced
as `Layer` everywhere. `Engine::getRegisteredContext` throws `std::invalid_argument`, the same exception
type the surrounding `try`/`catch` blocks already handle. **No registration order, key, `int_key` or
error-path change.**

**Why the downcast is the thing rung 1 breaks.** A downcast to `AppContext *` asserts that
`getRegisteredContext(name)` returned that concrete type. Once bring-up is conditional, "which contexts
exist" becomes a function of `NNTR_ENGINE`, and a call site that reaches past the facade has no way to
report a missing one — it either resolves to a different concrete type or dereferences a failed cast. The
facade path throws a named, catchable error instead, which is what makes rung 1's
`[Engine] gpu Context is not registered` line possible rather than a crash.

**Scope:** the 16 files that need nothing beyond the facade — `deberta_v2`, `gemma3`, `gemma4`, `gpt_oss`,
`gpt_oss_cached_slim`, `lfm2`, `qwen3`, `qwen3_cached_slim_moe`, `qwen3_moe`, `qwen3_slim_moe`,
`sentence_transformer`, `Custom/LayerClient`, `LLaMA`, `SimpleShot`, `YOLOv2`, `YOLOv3`. The remaining files
(`causal_lm.cpp`, `transformer.cpp`) depend on the model-registry rung of this series and ride a separate
tail PR.

Verification of the *split*, not just the port: `git grep -nE '[-]>registerFactory[[:space:]]*\('
-- Applications test` is 29 hits / 18 files on `main` and **3 hits / 1 file** on this branch — exactly the
file this rung deliberately excludes. The 16 files it does touch carry 0 remaining downcast calls.

**A note for whoever maintains the contribution checklist:** the prescribed single-line detection grep
(`static_cast<.*Context *\*>.*registerFactory`) never matches, because every real site splits the cast and
the call across lines (39 lines apart in one case). Check the **call** instead — it is single-line by
construction. Do not use a bare downcast grep: `static_cast<.*Context *\*>` alone matches 155 lines
tree-wide, almost all of them `static_cast<ClContext*>(getRegisteredContext("gpu"))` inside kernel wrappers
fetching a kernel, which is a different pattern.

**Known remaining, deliberately not converted:** three SimpleShot unit tests call
`AppContext::Global().registerFactory(...)` and then `createObject<Layer>` on the same object — direct
singleton use, not a context downcast, and the facade has no `createObject` counterpart.

**Verified:** `g++ -fsyntax-only` with the flags an x86 OpenCL build actually uses, over the 14 of these 16
files that build configures — they fail on a pre-facade base with
`error: 'class nntrainer::Engine' has no member named 'registerLayerFactory'` and compile clean with the
facade underneath (14/14). `clang-format` 14.0.6 clean on the changed lines. The rung also completes
`lfm2_causallm.cpp`'s file header with `@bug No known bugs except for NYI items`, because converting its
registration block pulls it into the changed-file set that the advanced doxygen rule applies to.

**Not verified here:** a link of the touched Applications and one model run per app family to confirm the
custom layer types still resolve at runtime. Startup-path-only change; one extra virtual call per custom
layer type, nothing on any inference path.

---

branch as its base, so this PR targets `main` and its diff therefore also contains the parent rungs named in
`Depends-on:`. The two commits to review are `[engine] Bring up only the compute engine the process actually
asked for` and `[Applications] Register custom layers through the Engine facade, not a downcast`; the

Depends-on: #4117


jijoongmoon:
Superseded by the re-cut series against current main: #4306, which is this PR's primary successor. Closing to keep the queue reviewable — please review the new series instead.

Scope split: #4306 carries the `bringUpWanted()` bring-up gate and the facade conversions for the five non-CausalLM example applications (LayerClient, LLaMA, SimpleShot, YOLOv2, YOLOv3). The CausalLM call sites ride the CausalLM change later in the series, and the CUDA half of the gate rides the CUDA backend change.

The `CHANGES_REQUESTED` review here was harvested before the successor opened, and most of it is applied:

- **The comment cites code that does not exist on this branch** (`causallm_engine()` in `llm_util.hpp`) — correct, and the same class of defect as a doc anchor pointing at nothing. Rewritten to state the contract it actually depends on: `NNTR_ENGINE` is latched on first use, so it must be exported before the first `Engine::Global()`.
- **Lowercasing `NNTR_ENGINE` here but nowhere else** — the inconsistency cannot arise in the successor, because `engine.cpp` is the *only* reader of `NNTR_ENGINE` in that tree (the `cuda_context` and `cl_context` readers arrive with their own backends). To keep it that way, `requestedEngine()` now carries a note telling the next backend to route through it rather than comparing `getenv()` exactly, which is precisely the `NNTR_ENGINE=CUDA` failure you traced.
- **The measured win is conditional on the operator exporting the variable** — agreed, and said so in the successor's body.
- **Raise the "compiled in but not brought up" message from `ml_logi` to `ml_logw`** — done. The second half (warn again at `compile()` when a graph's `engine=gpu` nodes were normalized to cpu) is not in the successor: `NeuralNetwork::compile()` is not in its scope. Worth doing where that file next moves.
- **No test pins the new policy down** — six cases added to `unittest_nntrainer_appcontext`: name resolution against the live registry, the `cpu` fallback for an unregistered name, registration through the facade end to end, the throw on an unknown engine, the context-declared residency plane, and the CPU capability snapshot feeding the resolver.
- **Latch asymmetry inside `bringUpWanted`** — kept, and now documented as deliberate: the engine name is latched, the per-backend eager override is not, because it is read once per backend.
- **Two missed downcast sites (`transformer.cpp`, `causal_lm.cpp`)** — both are CausalLM files and belong to the later change in the series; noted there.
- **`NNTR_ENGINE` is load-bearing but undocumented** — the environment-flag reference is #4303, which is not merged yet; the row belongs there.

Thank you for the trace through `parseComputeEngine` → `isComputeEngineGPU()` → `NetworkGraph(..., "gpu")`; that confirmation is why the successor keeps the fallback rather than throwing.

Branch preserved at `upstream-pr/trackC1b-engine-conditional-bringup`.

PR #4215, opened 2026-07-30
[causallm] Persistent tokenizer snapshot cache: rebuild the BPE instead of re-parsing 32 MB of JSON

jijoongmoon:
`Transformer`'s constructor parses the ~32 MB BPE `tokenizer.json` **synchronously**, before the
model factory and before any GPU bring-up, so all of it sits on the critical path — measured
**626 ms median, 19 % of a warm 3.3 s time-to-first-token** on the reference desktop package.

The cost is not JSON text parsing, so no parser swap removes it: it is dominated by the `tokenizers`
crate's untagged+flatten deserialize buffering plus BPE construction (a full msgpack round-trip
measured ~505 ms vs ~525 ms JSON — no win). So this caches the **result**, not the input: the parsed
vocab/merges tables plus the small component sections are written to a snapshot file, and later runs
rebuild the tokenizer from it.

## Safety contract — the reason this is acceptable on a correctness path

- **Byte-identical tokenization on a HIT.** The snapshot reconstruction mirrors the crate's own
  deserialize order (model, then components, then added tokens), so encode/decode of any prompt
  equals the parse path.
- **Silent fallback on key mismatch or corruption.** Source `(size, mtime)` mismatch, header version
  bump, short read, trailing bytes, FNV-1a payload checksum mismatch, or a loader rejection all fall
  back to the untouched `FromBlobJSON` path and schedule a rewrite. A failure is a debug log, never
  an error, and **never a different answer**.
- **Opt-out:** `NNTR_TOKENIZER_CACHE=0` disables it entirely — no read, no write.
- The (re)write runs on a background worker **after** the parsed tokenizer has been handed to the
  caller, so a MISS never adds to the critical path. The worker is a function-static `std::async`
  future rather than a detached thread, so its destructor joins at exit and a short-lived CLI process
  still persists the cache (a detached writer's rename was verified not to land when `main()`
  returned first).

Cache location: `<tokenizer_file>.ntkc` next to the model, falling back to
`$XDG_CACHE_HOME/nntrainer/tokenizer/<hash>.ntkc` when that directory is not writable.

## Measured

RTX 5060 Laptop desktop host, a 3.5B-class QS4CX package, quiet host, 4–5 runs per arm, snapshot HIT
vs `NNTR_TOKENIZER_CACHE=0`:

| arm | n | full-span TTFT | tokenizer lap | peak RSS |
|---|--:|--:|--:|--:|
| `NNTR_TOKENIZER_CACHE=0` (parse) | 4 | 2126 [2024–2147] | 597 [592–599] | 4366.2 MB |
| snapshot HIT (default) | 3 | **1713** [1709–1781] | **234** [226–240] | **4177.6 MB** |

**−413 ms full span, −363 ms on the lap, −189 MB peak RSS** — rebuilding from the snapshot's
vocab/merges records never materializes the 32 MB JSON document plus serde's transcode buffers.

Gates — each its own run, all with byte-identical golden output:

| gate | tokenizer lap | outcome |
|---|--:|---|
| HIT | 226–240 | as measured above |
| MISS (snapshot parked) | 591 | parse + background rewrite; the rewritten file is **md5-identical** to the original |
| post-rewrite HIT | 227–241 | the exit-joined writer landed |
| bit-flip inside the checksummed window | 625 | silent fallback to parse, correct output |
| truncation to 1/3 | 609 | silent fallback to parse, correct output |
| stale key (`touch tokenizer.json`) | 591 → 230 | invalidate, re-derive + rewrite, then HIT |
| `NNTR_TOKENIZER_CACHE=0` | 597 | pure parse path, no cache file touched |

A snapshot written on one host loaded unmodified on another — the point of keying it by content
(source size/mtime/version + payload checksum) rather than by build.

## What a reviewer should know before approving

- **`Applications/CausalLM/lib/libtokenizers_c.a` is regenerated** (36.1 → 40.2 MB) because the
  snapshot payload is produced/consumed inside the Rust wrapper: three new C entry points
  (`tokenizers_snapshot_from_json` / `tokenizers_snapshot_free` /
  `tokenizers_new_from_snapshot`). The Windows build compiles the same crate from source
  (`tokenizers_c_win`), which is why `src/lib.rs` carries the matching implementation. As with the
  prebuilt it replaces, the archive embeds cargo-registry paths from the machine that built it in
  Rust panic-location strings; removing those needs a rebuild with `--remap-path-prefix`.
- **The snapshot deliberately bypasses serde for the heavy sections.** BPE's custom `Serialize`
  declares 8 struct fields but writes 10, which truncates vocab/merges in any length-prefixed
  format, so only an explicit layout is safe. The payload carries its own magic + version.
- **One call site changes.** `FromBlobJSON(LoadBytesFromFile(...))` becomes
  `LoadTokenizerCached(...)`; the parse path itself is untouched and remains the fallback for every
  failure mode.

**Where this sits in the whole span, attributed honestly.** On this box the three *code* rungs of
this campaign — the i8 prewarm length gate (#4207), the symmetric engine-conditional bring-up
(#4214) and the tokenizer snapshot cache (#4215) — take a warm short turn from **3326 → 1777 ms
(−47 %)** and peak RSS from 4432 → 4080 MB. Reaching the campaign's headline **1504 ms / 2200 MB
(−55 % / −50 %)** additionally needs a **packaging** change with no code in it: moving the PLE and
embedding tables into mmap'd sidecar payloads (`ple_file_name` / `embedding_file_name` — runtime keys
already consumed) measured **−273 ms and −1978 MB peak RSS** (UVM plane 2696 → 1373 MiB), golden
identical to the canonical package. That is a recommendation for model packaging, **not part of this
or any other PR here**. Goldens are byte-identical across all 72 short-turn runs of the campaign,
which share one md5.

Applies to `main` directly — no dependency on the backend series.

### DIFF
--- a/Applications/CausalLM/huggingface_tokenizer.cpp
+++ b/Applications/CausalLM/huggingface_tokenizer.cpp
@@ -129,4 +129,25 @@ Tokenizer::FromBlobByteLevelBPE(const std::string &vocab,
     vocab.data(), vocab.length(), merges.data(), merges.length(),
     added_tokens.data(), added_tokens.length()));
 }
+
+std::unique_ptr<Tokenizer> Tokenizer::FromSnapshot(const std::string &blob) {
+  TokenizerHandle handle =
+    tokenizers_new_from_snapshot(blob.data(), blob.length());
+  if (handle == nullptr) {
+    return nullptr; // caller falls back to the JSON parse path
+  }
+  return std::make_unique<HFTokenizer>(handle);
+}
+
+std::string Tokenizer::SnapshotFromJSON(const std::string &json) {
+  char *data = nullptr;
+  size_t len = 0;
+  tokenizers_snapshot_from_json(json.data(), json.length(), &data, &len);
+  if (data == nullptr || len == 0) {
+    return {};
+  }
+  std::string blob(data, len);
+  tokenizers_snapshot_free(data);
+  return blob;
+}
 } // namespace tokenizers
--- a/Applications/CausalLM/jni/Android.mk
+++ b/Applications/CausalLM/jni/Android.mk
@@ -78,6 +78,7 @@ LOCAL_SRC_FILES := \
     ../models/gpt_oss/gptoss_causallm.cpp \
     ../models/gpt_oss_cached_slim/gptoss_cached_slim_causallm.cpp \
     ../huggingface_tokenizer.cpp \
+    ../tokenizer_cache.cpp \
     ../llm_util.cpp \
     ../layers/embedding_layer.cpp \
     ../layers/embedding_pooling_layer.cpp \
--- a/Applications/CausalLM/meson.build
+++ b/Applications/CausalLM/meson.build
@@ -1,6 +1,7 @@
 causallm_src = [
     meson.current_source_dir() / 'chat_template.cpp',
     meson.current_source_dir() / 'huggingface_tokenizer.cpp',
+    meson.current_source_dir() / 'tokenizer_cache.cpp',
     meson.current_source_dir() / 'llm_util.cpp',
     meson.current_source_dir() / 'api' / 'callback_streamer.cpp',
     meson.current_source_dir() / 'api' / 'causal_lm_api.cpp',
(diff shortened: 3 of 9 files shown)


jijoongmoon:
Consolidated into #4212.

PR #4216, opened 2026-07-30
[cuda] Derive-once disk cache for the QS4CX weight packs (opt-in, default off)

jijoongmoon:
**This ships default-OFF, and the default is the measurement, not caution.** Read the trade table
before the mechanism: on this lane one of the two derived weight families is worth caching and the
other is not, and the disk axis disagrees with the latency axis. `NNTR_CUDA_PACK_CACHE=1` opts in.

## The trade, in full

Every launch re-derives the same derived weight bytes from the same plain nibbles: the dp4a
packed-int4 + row sums, and — when a prefill can reach the cuBLAS-i8 gate — the cuBLAS-i8 `[K,N]`
transpose-unpack. This persists them once and uploads from the mapped pack instead.

Measured on RTX 5060 Laptop, a 3.5B-class untied-lm_head QS4CX package, quiet host, goldens
byte-identical on every arm:

| cell (card active) | n | full-span TTFT | prewarm lap | disk |
|---|--:|--:|--:|--:|
| short turn, no cache | 5 | 2138 | 332 | — |
| short turn, pack HIT (**page cache warm**) | 4 | **2034** [2020–2042] | **232** | 1373 MB |
| short turn, pack HIT (**page cache COLD**) | 1 | **2854** | **1150** | 1373 MB |
| short turn, pack MISS (writes it) | 1 | 2445 | 672 | → 1373 MB |
| long prompt (1.2K), no cache | 3 | 3424 [3316–3448] | 1497–1505 | — |
| long prompt, pack HIT | 3 | **2378** [2262–2406] | **425–435** | 2.6 GB |
| long prompt, pack MISS (writes it) | 1 | 4020 | 2150 | → 2.6 GB (695 records) |

So: **−104 ms on a warm short turn, +716 ms on a cold page cache, −1046 ms (−31 %) on a long
prompt**, at a cost of **+1.4 GB / +2.6 GB of disk** per model.

Why the split, and why it can only be decided by measurement — this rung adds permanent cost
accounting (`cuda_fc_qs4cx_prewarm_stats`, reported as a `split` log line when the cache is enabled)
because the question is not "how long is the prewarm" but "how much of it is a **derive** a cache
could remove":

| workload | host derive | H2D | pack tee (MISS) | pack HIT | payload |
|---|--:|--:|--:|--:|--:|
| short turn (i8 gated off) | **150.9** | 83.8 | 214.7 | **107.2** | 1366 MB |
| 1.2K prompt (i8 engaged) | **1114.8** | 193.0 | 408.1 | **206.3** | 2570 MB |

The dp4a derive is bandwidth-bound and 16-way parallel (151 ms for 1366 MB), so caching it trades CPU
for I/O and only pays warm. The i8 `[K,N]` derive is the opposite — a column-strided transpose-unpack,
1115 ms for 2570 MB — and caching that wins a full second. Latency, memory and disk do not agree, so
the default is off with a recommendation: **enable it for prefill-heavy deployments** (long prompts,
where the i8 records exist and the page cache stays hot) that can afford 2.6 GB per model.

Other axes: **peak RSS is unchanged in every arm** (4431.6–4432.4 MB) — the hit path uploads from the
pack mmap in 64 MB chunks and `MADV_DONTNEED`s each consumed chunk, so transient residency is one
chunk, not the pack. Prefill TPS (5457 → 5457–5485) and decode (42.5 → 42.4–42.6) are unchanged,
i.e. the cached bytes **are** the derived bytes.

## Mechanism, and the one hard rule

`cuda_pack_cache.{h,cpp}` is a structural port of the OpenCL lane's v8c pack cache: same on-disk
layout (64 B header / blob regions / **index appended last**, so a temp file that never finalized
cannot validate), same identity rules, same failure semantics, same background-finalizer-joined-at-exit
write path. The two lanes now read the same way.

**Identity is by FILE + NAME, never by pointer.** This is carried from that work because this
codebase's pointer-keyed derived-weight caches have mis-hit under SVM/UVM before. The pack is bound
to the weight file's `(size, mtime_ns, format version)`; each record to `(record-name FNV-1a, N, K,
row_bytes, payload_len)` plus a sampled payload FNV (both 64 KB ends + 16 interior pages, ~192 KB per
record, so validating never costs what the cache saves) plus a full FNV over the row-sum block. The
record name is the graph-stable `<layer name>.<weight slot>#<kind>`, not an address. The in-memory
cache stays pointer-keyed and process-local; only the disk one is content-addressed.

Fail-safe: a missing, stale, truncated or bit-flipped pack simply derives as before. Verified per
gate — cache OFF, a MISS that writes the pack, a HIT, a bit-flip inside a checksummed window, a
truncation, and a stale mtime — all with identical output.

## Context: what this rung is *not*

The same campaign found a bigger, zero-code memory lever that reviewers should know about because it
changes what "init" costs on this lane: moving the PLE and embedding tables out of the model bin into
mmap'd sidecar payloads (runtime keys `ple_file_name` / `embedding_file_name`, already consumed)
measured **−273 ms and −1978 MB peak RSS** (UVM plane 2696 → 1373 MiB), golden identical to the
canonical package. **That is a packaging recommendation, not a patch, and no commit here implements
it** — it is stated so this PR's +2.6 GB of disk is weighed against the alternative that costs
nothing and saves more.

Depends-on: #4207 — this consumes the load-time prewarm and its length gate from that rung.

---

ARM KleidiAI) into nntrainer. New levers are gated so existing builds and the default execution path
stay byte-identical unless a feature is explicitly enabled — which for this rung means **nothing
changes unless `NNTR_CUDA_PACK_CACHE=1` is set**.

rungs named in `Depends-on:`. The commit to review here is the single
`[cuda] Derive-once disk cache for the QS4CX weight packs (opt-in, default off)`.


jijoongmoon:
Consolidated into #4222.

The one conflict this absorption required is disclosed in #4222's body: this rung was cut before the plain-page drop existed, so its prewarm call and the `NNTR_CUDA_DROP_PLAIN` teardown met at the same line in `causal_lm.cpp`. Resolved by keeping the drop seam and threading `pack_name.c_str()` through the prewarm, with the drop running after it so a pack rewrite still sees the plain payload it derives from. Containment verified line-level: **all 841 non-blank added lines of this rung are present on the new #4222 tip, 0 absent**.

PR #4217, opened 2026-07-30
[Wait for #4125][tensor] Carve the inference weight arena out of the allocator zero-fill

jijoongmoon:
**Depends-on: #4125** (`upstream-pr/trackD1-determinism`), which itself carries #4117 (the CUDA context/memory/stream infrastructure). This branch is based on #4125's head, not on `main` — see "Why this is not a main-direct change" below before reading the diff.

## Why this is not a main-direct change, and why it must not be read as "removing a safety fill"

I traced the provenance before choosing a base, because the shape of this change ("stop zero-filling the weight arena") looks like an optimization against `main` and is not one:

| file | on `main`? | who put the unconditional fill there |
|---|---|---|
| `nntrainer/mem_allocator.cpp` | present, but `alloc()` has **NO memset** | added by `a6c1dc91a` (route MemoryPool through per-vendor MemAllocator), **removed** by `5c061673e` ("Reduce peak memory by removing redundant initialization"), **re-added** by `b14130174` (the OpenCL infrastructure rework, in this stack) |
| `nntrainer/cl_svm_allocator.cpp` | **absent** | introduced with the file; poison hook added by the determinism change |
| `nntrainer/cuda_mem_allocator.cpp` | **absent** | present at the file's creation commit `4f5c35e5f` as `std::memset(hp,0,size)` + `cudaMemset(dptr,0,size)`, explicitly to "honor MemAllocator's calloc contract (per-process nondeterminism class)"; the determinism change `9cc2a7245` only re-parameterized `0` -> `cuda_fill_byte()` |

So on `main` this change would be a behavioural no-op: there is no fill to carve out. It is emphatically **not** a no-op on this base — the fill exists, it costs 142-198 ms of pool commit plus 92-124 ms of loader time, and skipping it for one plane is a real, measured, golden-preserving win.

That is the correct framing: **this is the weight-arena carve-out on the calloc/determinism fill those commits introduce**, not the removal of a safety fill. Every `zero == true` caller — activations, KV, gradients, every derived device cache, every training plane — keeps the fill and its comments verbatim. Worth noting on the way past: the base-class hop is the one place in this chain that was already *deliberately removed once* upstream for peak-memory reasons (`5c061673e`) and later restored.

## The mechanism is not "skip a memset" — it pays twice

`cudaMemset` over a never-touched **managed** arena makes the driver POPULATE and migrate the whole plane to the device; the host loader then faults every one of those pages back to write the weights into them. Dropping the fill therefore removes an arena-wide device memset **and** a plane-wide device->host migration inside the loader. The page faults are not saved — they move to the load that actually writes the bytes:

| lap (RTX 5060 Laptop, card active, medians) | before | after |
|---|--:|--:|
| arena commit, sidecar package (1 439 672 998 B) | 141.9 | **0.1 ms** |
| `load`, sidecar | 290.1 | **198.4 ms** |
| arena commit, canonical package (2 827 MB) | 197.9 | **0.1 ms** |
| `load`, canonical | 469.9 | **345.7 ms** |

Full span, process start to first token, short golden turn:

| cell | before | after |
|---|--:|--:|
| sidecar package, card active | 1490 | **1281 ms** (-209, -14.0%) |
| canonical package, card active | 1719 | **1356 ms** (-363, -21.1%) |
| sidecar, GPU in a suspended power state | 3736 | 3443 ms (-293) |
| second model (smaller QS4CX package) | 716 | 664 ms (arena lap 41.3 -> 0.05) |

Peak RSS unchanged (2200.6 vs 2200.5 MB) — the fill was a write into already-reserved address space, not an allocation. Decode 44.6-46.3 TPS and prefill 67-69 ms are inside the pre-change band.

## Scope is structural, not a heuristic

`init == false` reaches `TensorPool::allocate` from exactly one call site: `Manager::allocateWeights(_, exec_mode != INFERENCE)`, i.e. the weight pool in inference. The chain already existed for three hops and stopped one short of the allocator; this adds the last hop (`MemoryPool::allocate(bool zero)` -> `MemAllocator::alloc(..., bool zero)`) and honors it in `MemAllocator`, `ClSVMAllocator`, and in all three planes `CudaMemAllocator` can return (pinned host-mapped, managed/device, host fallback).

Both signatures default to `zero = true`, so no other caller changes behavior, and the `override` keyword on ClBufferPool / CachePool / ClSVMAllocator / CudaMemAllocator / QNNRpcManager makes the widened signature fail loudly instead of silently de-virtualizing. `QNNRpcManager` and `CachePool` accept `zero` and **document why they cannot act on it** (neither path has an arena-wide fill) rather than pretending to honor it.

**No plane-size gate hides the patched path on the CUDA lane**: `CudaMemAllocator` does not override `makePool()`, so the weight pool is a plain `MemoryPool` and takes the single-contiguous `allocate()` route at any size — verified live at both package sizes above. The OpenCL lane differs and this is stated so nobody reads a null result there as a failed port: `ClSVMAllocator` *does* override `makePool()`, and its `ClBufferPool` diverts any plane above `CL_DEVICE_MAX_MEM_ALLOC_SIZE` to `allocateFSU()`, which has no `zero` parameter at all.

## Gates

Goldens byte-identical on every arm of every gate (one md5 across both packages).

| gate | evidence |
|---|---|
| **poison discriminator** | `NNTR_WEIGHT_ARENA_ZERO=1 NNTR_POISON_FILL=1` fills the arena with **0x55 instead of 0x00**; generated text **byte-identical 3/3**. This is the proof that no consumer reads an arena byte the loader did not write — if one did, poison would change the output |
| **escape hatch** | `NNTR_WEIGHT_ARENA_ZERO=1` restores the old numbers exactly: sidecar arena lap back to 117-123 ms and full span to 1454 ms (inside the pre-change band), canonical back to 197.9 / 1719 |
| **both package sizes** | 1.44 GB and 2.83 GB weight planes both report the skip and both improve |
| **second model** | its own golden, identical under hatch and default |

The underlying invariant is pre-existing, not new: a weight the model file does not cover already throws in `NeuralNetwork::load` today, so "every byte of the weight plane is written before any read" is already enforced.

## The one reader that does not write every byte — held, with the reason

The legacy QINT4 -> QS4CX transcode is the single path that leaves a hole: it fills the `N*((K+1)/2)` nibble region and writes all `N` fp32 scales explicitly, but for **even K** the QS4CX scale offset is `N*(K+1)/2` evaluated left-to-right (= `N*K/2 + N/2`) while the nibble region is only `N*(K/2)` bytes, so an **`N/2`-byte gap** sits between them untouched — previously zero only by virtue of the arena fill.

**That transcode does not exist on this base**, so the carve-out for it is not in this diff; where it is present, `QS4CX_Tensor::read()` needs an unconditional `setZero()` before transcoding on both legacy overloads, and that belongs with the transcode's own change. Unconditional on purpose: that branch already does a full unpack of the record, so one memset is noise, and gating it on a pool decision three layers away would make the guarantee fragile. What it protects is **byte-preservation** of the plane (and any content-keyed identity computed over the payload), not output correctness — nothing reads the gap, which is what the poison discriminator independently establishes.

On the tree that does have the transcode, the path could **not** be exercised end to end: all three legacy-QINT4 packages available there fail to load with a pre-existing `Model layout mismatch`, identically with and without this change. Recorded as **unavailable**, never as passed.

## Validation status

Measured on Linux + NVIDIA (RTX 5060 Laptop). **Windows/WDDM and ARM/Adreno are PENDING.** The recycled-page hazard the fill exists for is most acute on WDDM; the argument above says the weight plane has no window for it, but that has not been run there — which is also why the escape hatch is a documented env var rather than a build flag.

`test/unittest/memory`: `CountingAllocator`'s `alloc` override is widened and forwards `zero`, so the test still routes through the allocator it counts.


jijoongmoon:

The change here is NOT carried by that series: the allocator zero-fill carve-out for the inference weight arena is not present in #4305 or anywhere else in the series.

PR #4218, opened 2026-07-30
[Wait for #4215][causallm] Build the tokenizer on a side thread and join it at first use

jijoongmoon:
**Depends-on: #4215** (`upstream-pr/track0-tokenizer-snapshot-cache`). This branch is based on that PR's head — it is the second half of the same lever and is meaningless without it.

## What #4215 left on the table

#4215 cut the tokenizer build from ~597 ms to ~234 ms, but left it fully **EXPOSED**: the build ran in `Transformer`'s constructor, **before** graph compile / pool commit / weight load, so all 234 ms were still serial time-to-first-token. The build depends on nothing those phases produce, so it can run beside them.

- the constructor starts `std::async(std::launch::async, LoadTokenizerCached)` and keeps the future;
- `ensureTokenizer()` joins it under a mutex, exactly once, and is called immediately before the first tokenizer touch of a run;
- `getTokenizer()` calls `ensureTokenizer()` first, so external callers cannot observe an unjoined tokenizer.

It **composes** with the snapshot cache rather than duplicating it: the worker calls the same `LoadTokenizerCached()` entry point, so there is exactly one parse and one cache decision either way — only the thread that pays for it changes. No second code path, no double-parse on a HIT.

App-side only (`Applications/CausalLM`); no core change.

## Measured — interleaved A/B, not two separate campaigns

3.5B-class untied QS4CX package, RTX 5060 Laptop, card active, short golden turn, goldens byte-identical on every arm. 5 async/synchronous pairs **alternated** so GPU-clock and page-cache drift cannot land on one arm only:

| arm (same binary) | n | TOTAL median | delta |
|---|--:|--:|--:|
| `NNTR_TOKENIZER_ASYNC=0` | 5 | 1263 [1242-1268] | — |
| async (new default) | 5 | **1078** [964-1100] | **-185** (per-pair median **-187**) |

Pooled over all runs of the cell: **1262 -> 1072 ms (-190)**.

| entry | before | after |
|---|--:|--:|
| tokenizer exposed lap | 238.5 | **8.2 ms** (join BLOCKED **0.0 ms**) |
| CUDA context bring-up | 272.1 | 318.6 ms |
| peak RSS | 2200.5 | 2225.4 MB (**+24.8**) |
| first token / decode | 68.1 ms / 45.8-46.2 TPS | band unchanged |
| canonical package | 1356 | 1154 ms |
| GPU in a suspended power state | 3500 | 3264 ms |

The overlap window is the GPU driver bring-up and it swallows the whole build: a representative trace requests the build at 1.9 ms, the worker finishes at 293.1 ms — entirely inside `cuInit` + primary-context retain (230.1 -> 332.8) — and the join at 584.9 ms blocks **0.0 ms**.

### Why -185 and not -230

The exposed lap falls by 230, but context bring-up **grows by 46 ms**: the worker competes with `cuInit` for the same cores. Overlap credits are `max()`, not `sum()`, and that rule bites **inside a single lever** here. The credit is the measured span (1262 -> 1072), not the lap delta — quoting -230 would be quoting a lap, not a latency.

## Exception safety

A worker failure is not swallowed. It rides the future and is re-raised at the join. `std::future::get()` invalidates the future even when it throws, so the exception is latched in a **sticky** `std::exception_ptr` (`tokenizer_error_`) and re-raised on **every** later `ensureTokenizer()` — otherwise a second caller would sail past with `tokenizer == nullptr` and fault far from the cause.

## Join coverage

Every direct use of the `tokenizer` member is dominated by a join: `causal_lm.cpp` (one join before both `Encode` calls; the generation loop's `Decode`/`registerOutputs` are downstream of it in the same run), `sentence_transformer`, `deberta_v2`, `multilingual_tinybert_16mb`, `xlm_roberta`, and **both** lfm2 entry points — `run()` before `Encode`, and `run_with_embeddings()`, which takes pre-computed embeddings and never calls `Encode` but still `Decode`s through `registerOutputs()`. The only other reader is `getTokenizer()`, which joins internally.

## Gates

Goldens byte-identical on every one.

| gate | result |
|---|---|
| snapshot HIT (default) | join BLOCKED 0.0 ms; exactly **one** cache decision in the trace, i.e. no double-parse |
| snapshot **MISS** | worker parses 853.3 ms, join BLOCKED **365.9 ms** — only the residue past the overlap window is exposed; one parse, not two |
| background rewrite after the MISS | written from the worker thread in 363.2 ms, **md5 identical** to the pre-existing snapshot |
| post-rewrite HIT | join BLOCKED 0.0 ms |
| worker **throws** (`tokenizer.json` absent) | `tokenizer async build FAILED -- rethrowing at join`, then the **same** fatal message and the same rc as the synchronous path. No crash, nothing swallowed |
| corrupt `tokenizer.json` | async and synchronous arms fail **identically** (the C tokenizer wrapper accepts a bad document and yields 0 tokens -> a `TensorDim` throw at prefill). Pre-existing wrapper weakness, unchanged here; recorded because it is what a corrupt file actually does |

## Memory is a real cost here

**+24.8 MB peak RSS, and it is causal, not noise.** With the build moved beside the graph/pool/load phase, its transient peak now *overlaps* the weight-arena commit instead of preceding it, so the process-wide maximum rises. 24.8 MB for 190 ms on a 2.2 GB process is a good trade, but it is a trade and it belongs in the record.

Escape hatch: `NNTR_TOKENIZER_ASYNC=0` restores the synchronous build — the pre-change code path verbatim.

### DIFF
--- a/Applications/CausalLM/huggingface_tokenizer.cpp
+++ b/Applications/CausalLM/huggingface_tokenizer.cpp
@@ -129,4 +129,25 @@ Tokenizer::FromBlobByteLevelBPE(const std::string &vocab,
     vocab.data(), vocab.length(), merges.data(), merges.length(),
     added_tokens.data(), added_tokens.length()));
 }
+
+std::unique_ptr<Tokenizer> Tokenizer::FromSnapshot(const std::string &blob) {
+  TokenizerHandle handle =
+    tokenizers_new_from_snapshot(blob.data(), blob.length());
+  if (handle == nullptr) {
+    return nullptr; // caller falls back to the JSON parse path
+  }
+  return std::make_unique<HFTokenizer>(handle);
+}
+
+std::string Tokenizer::SnapshotFromJSON(const std::string &json) {
+  char *data = nullptr;
+  size_t len = 0;
+  tokenizers_snapshot_from_json(json.data(), json.length(), &data, &len);
+  if (data == nullptr || len == 0) {
+    return {};
+  }
+  std::string blob(data, len);
+  tokenizers_snapshot_free(data);
+  return blob;
+}
 } // namespace tokenizers
--- a/Applications/CausalLM/jni/Android.mk
+++ b/Applications/CausalLM/jni/Android.mk
@@ -78,6 +78,7 @@ LOCAL_SRC_FILES := \
     ../models/gpt_oss/gptoss_causallm.cpp \
     ../models/gpt_oss_cached_slim/gptoss_cached_slim_causallm.cpp \
     ../huggingface_tokenizer.cpp \
+    ../tokenizer_cache.cpp \
     ../llm_util.cpp \
     ../layers/embedding_layer.cpp \
     ../layers/embedding_pooling_layer.cpp \
--- a/Applications/CausalLM/meson.build
+++ b/Applications/CausalLM/meson.build
@@ -1,6 +1,7 @@
 causallm_src = [
     meson.current_source_dir() / 'chat_template.cpp',
     meson.current_source_dir() / 'huggingface_tokenizer.cpp',
+    meson.current_source_dir() / 'tokenizer_cache.cpp',
     meson.current_source_dir() / 'llm_util.cpp',
     meson.current_source_dir() / 'api' / 'callback_streamer.cpp',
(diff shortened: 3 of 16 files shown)


jijoongmoon:
Consolidated into #4212.

PR #4219, opened 2026-07-30
[causallm] One prompt-preparation seam for the chat template

jijoongmoon:
Both front ends of `Applications/CausalLM` have to turn a user prompt into model input, and each was about to grow its own way of doing it. This adds one prompt-preparation seam next to the existing renderer, points the CLI at it, and pins its contract with a unit test. **No behaviour change** — the CLI's one existing render call site becomes a `buildPrompt()` call, and the literal `argv[2]` prompt path is left byte-for-byte alone.

The consumer that makes this worth having is the SDK path (`Applications/CausalLM/api`), which today cannot template a prompt the way the runner does. That is a separate PR so this one can be reviewed on its own terms; see the bottom of this description.

## The seam

`Applications/CausalLM/chat_template.h` gains four free functions beside `causallm::ChatTemplate` — no new renderer, no second template source:

| symbol | role |
|---|---|
| `enum class PromptTemplateMode { Auto, Never }` | apply-when-available vs feed-raw |
| `chatTemplateContext(nntr_cfg)` | the package's optional `chat_template_context` object from `nntr_config.json` (empty object when absent) |
| `makeUserRequest(user_text, render_context)` | wrap one user turn into a request |
| `buildPrompt(tmpl, request, mode, render_context)` | the single application point: merges the package context *under* the request (request keys win), then calls `ChatTemplate::apply` |
| `buildUserPrompt(tmpl, user_text, mode, render_context)` | single-turn convenience; returns `user_text` unchanged when nothing is applied |

Three design points, each of which cost something to learn:

1. **One renderer, not two.** The seam consumes the in-tree minja machinery (`third_party/minja`) that `ChatTemplate` already wraps. A chat template is a property of the **model package** (`chat_template.jinja`, else `tokenizer_config.chat_template`), never of a table in the code keyed on an architecture string — two packages of the same architecture legitimately carry different templates.
2. **Render failures propagate.** A template that cannot render is reported, never silently downgraded to a raw prompt. An instruction model fed an unmarked prompt drifts, and that failure reads as a model bug rather than a configuration one; it has cost real debugging time.
3. **`buildUserPrompt` is not an overload of `buildPrompt`.** A string literal converts to `nlohmann::json` as readily as to `std::string`, so an overload pair made every `buildPrompt(t, "text")` call site ambiguous. The new unit test caught that before any caller could.

### Why a package-level render context

Some templates branch on a *sibling* render key rather than on the messages. `ChatTemplate::apply` already forwards any extra top-level request key into the renderer, so the plumbing exists; what was missing was an owner for the value. The model package is that owner, via an optional `"chat_template_context"` object in `nntr_config.json`.

Concretely: one in-tree model's template emits `…<|im_start|>assistant\n` and **stops** unless the request carries `enable_thinking: false`, in which case it emits the `<think>\n\n</think>\n\n` form its validated prompt uses (85 B vs 104 B rendered). This is not a renderer defect — the template branches on `enable_thinking` and the upstream default is `True`, which here would open a think block nothing closes. Proven end-to-end that a package-level `"chat_template_context": {"enable_thinking": false}` produces **byte-identical** output to passing the key per call, so it is fixable in package data with no caller code. Flagged for the model-package owners below rather than worked around in code.

## CLI: refactor only

`main.cpp`'s single `chat_template->apply(nntr_cfg["chat_input"])` becomes a `causallm::buildPrompt(...)` call carrying the package context. Everything else about prompt selection is untouched:

- **`argv[2]` still bypasses the template entirely.** Callers that template their own input rely on those bytes arriving verbatim; that path is not touched by this PR.
- the `chat_input` → render → clear system head/tail sequence keeps its exact order and semantics;
- packages with no template keep falling through to `sample_input` raw.

## `unittest_chat_template` — new, 8/8

No model, no tokenizer, no backend: the shared `chat_template.cpp` plus gtest, so it runs anywhere the app builds. It pins package-context plumbing, opt-out byte-identity, request-keys-win precedence, render determinism, and raise-instead-of-raw.

```
[----------] 8 tests from ChatTemplateSeam
[  PASSED  ] 8 tests.
```

## Verification

Beyond the unit test, the seam was exercised through both front ends on four model packages, CPU lane, fp16 — reported in full on the SDK PR because that is where the equivalence claim lives. The two results that belong here:

| gate | result |
|---|---|
| four CLI goldens, baseline binary vs post-change binary, same command | **IDENTICAL 4/4** (160 / 49 / 161 / 145 generated bytes) |
| rendered prompt, cold renderer construction (18 KB jinja) / warm | 180–195 ms / **0.09–0.19 ms** |

The cold construction is a property of the renderer this PR does not change; it is measured here only to justify keeping the render where a model-load window can hide it, which is what the CLI already does and what the SDK PR preserves.

**Lane note:** the gates ran on the CPU lane (fp16). The seam is upstream of every backend — it produces prompt *bytes* before any engine is touched — so a GPU-lane repeat is confirmation, not discovery. This PR does not claim GPU-lane coverage.

## For the model-package owners (data, not code)

Two findings surfaced while rendering the in-tree packages. Both are packaging decisions, deliberately **not** patched in code here:

1. One public package's `chat_template.jinja` is a byte-identical copy of another (internal) package's template. Rendering it emits a thought-channel opening that the package's own validated `sample_input` does not contain, so the model answers in a "thought" style instead of the direct answer its golden records. Template and golden prompt disagree *inside the package*. Both front ends now do the same thing with it — equivalence holds either way — but the package needs either its real template or a `chat_template_context`.
2. The `enable_thinking` case described above.

## Depends-on / relation to other PRs

- No dependency. Applies directly to `main`.
- **Follow-up:** `[causallm/api] Apply the model package's chat template in the SDK path` consumes this seam, deletes the API's hardcoded per-architecture template table, and carries the SDK ≡ CLI byte-identity gate. It also depends on #4115 for the shared model registry.

### DIFF
--- a/Applications/CausalLM/chat_template.cpp
+++ b/Applications/CausalLM/chat_template.cpp
@@ -498,4 +498,57 @@ const std::string &ChatTemplate::sourcePath() const {
   return impl_->source_path;
 }
 
+nlohmann::json chatTemplateContext(const nlohmann::json &nntr_cfg) {
+  if (nntr_cfg.is_object() && nntr_cfg.contains("chat_template_context") &&
+      nntr_cfg["chat_template_context"].is_object())
+    return nntr_cfg["chat_template_context"];
+
+  return nlohmann::json::object();
+}
+
+nlohmann::json makeUserRequest(const std::string &user_text,
+                               const nlohmann::json &render_context) {
+  nlohmann::json request = nlohmann::json::object();
+  if (render_context.is_object())
+    request = render_context;
+
+  // The message list is ours to define; a render context may only supply
+  // sibling render keys (a thinking-channel flag, for instance), never the
+  // conversation itself.
+  request["messages"] =
+    nlohmann::json::array({{{"role", "user"}, {"content", user_text}}});
+  return request;
+}
+
+std::string buildPrompt(const ChatTemplate *tmpl, const nlohmann::json &request,
+                        PromptTemplateMode mode,
+                        const nlohmann::json &render_context) {
+  if (tmpl == nullptr || mode == PromptTemplateMode::Never)
+    return std::string();
+
+  // Package-supplied defaults fill only the keys the request left undefined:
+  // an explicit request is authoritative about its own render context.
+  if (render_context.is_object() && !render_context.empty() &&
+      request.is_object()) {
+    nlohmann::json merged = request;
+    for (auto it = render_context.begin(); it != render_context.end(); ++it) {
+      if (!merged.contains(it.key()))
+        merged[it.key()] = it.value();
+    }
+    return tmpl->apply(merged);
+  }
+
+  return tmpl->apply(request);
+}
+
+std::string buildUserPrompt(const ChatTemplate *tmpl,
+                            const std::string &user_text,
+                            PromptTemplateMode mode,
+                            const nlohmann::json &render_context) {
+  if (tmpl == nullptr || mode == PromptTemplateMode::Never)
+    return user_text;
+
+  return buildPrompt(tmpl, makeUserRequest(user_text, render_context), mode);
+}
+
 } // namespace causallm
--- a/Applications/CausalLM/chat_template.h
+++ b/Applications/CausalLM/chat_template.h
@@ -70,6 +70,81 @@ private:
   std::unique_ptr<Impl> impl_;
 };
 
+/**
+ * @brief Whether a caller's prompt should be run through the model's template.
+ */
+enum class PromptTemplateMode {
+  Auto,  /**< apply the model package's template when it has one */
+  Never, /**< feed the prompt exactly as given (already-templated input) */
+};
+
+/**
+ * @brief The model package's default chat render context.
+ *
+ * Read from the optional "chat_template_context" object in nntr_config.json.
+ * Some templates branch on a sibling render key rather than on the messages --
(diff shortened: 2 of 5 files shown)


jijoongmoon:
Consolidated into #4220.

#4220 is now the task-level SDK PR: this seam, the SDK path that applies the package's chat template through it, and a third rung that makes the SDK's load idempotent. The seam exists so the SDK can use it — landing it without its only consumer leaves dead abstraction, and landing the consumer without the seam is impossible.

Containment verified line-level: **all 273 non-blank added lines of this PR are present on the #4220 tip, 0 absent.** #4220's copy of the seam commit is a strict superset — the identical patch plus 4 comment lines in `Applications/CausalLM/meson.build` explaining why the test block is appended rather than inserted. Every measured number and design note from this body has been transcribed into #4220's description.

**⚠ One trade-off flagged for your call.** This PR was **main-direct** — no dependency, applies straight to `main`. #4220 sits on #4115 (the model registry, needed by the SDK consumer), so consolidating moves the seam onto that ladder and it now lands **with A1** rather than immediately. If that delay is not acceptable, reopen this PR and keep it standalone; the two copies are byte-compatible, so whichever lands second is a no-op.

#4220 lands with "Rebase and merge" (never squash), so each of the three rungs keeps its own revert handle.

PR #4220, opened 2026-07-30
[causallm/api] SDK path: one chat-template seam, apply the package template through it, and make load idempotent

jijoongmoon:
**Task-level PR: the SDK path, from prompt seam to idempotent load.** Three rungs, one story — the seam
exists so the SDK can use it, and the SDK is only usable if loading it twice is not a full reload.

Absorbs #4219 (now closed and pointed here) and a third rung that never had a PR of its own (idempotent
load). The branch head has been advanced to carry all three commits.

> **Branch-name caveat:** GitHub cannot repoint an open PR's head, so the ref is still
> `upstream-pr/trackA6-sdk-chat-template` even though it now also carries the load-idempotency rung.

**Merge button: "Rebase and merge" — never squash.** Three independent contracts, three revert handles.

**Depends-on: #4115** (`upstream-pr/trackA1-model-registry`) for `causallm::registerAllModels()`.

> ### ⚠ Trade-off that needs a call before merge
> #4219 was **main-direct** — it applied to `main` with no dependency. Consolidating it here moves it onto
> the #4115 ladder, so the seam now lands **with A1** rather than immediately. That is the price of one
> review instead of two. **If that delay is unacceptable, say so and #4219 is reopened as a standalone
> main-direct PR**; the seam commit in this branch is a byte-superset of it (identical patch plus 4 comment
> lines in `Applications/CausalLM/meson.build`), so either order works and whichever lands second is a
> no-op.

---

## 1. One prompt-preparation seam for the chat template

Both front ends of `Applications/CausalLM` have to turn a user prompt into model input, and each was about
to grow its own way of doing it. This adds one prompt-preparation seam next to the existing renderer,
points the CLI at it, and pins its contract with a unit test. **No behaviour change on the CLI** — its one
existing render call site becomes a `buildPrompt()` call, and the literal `argv[2]` prompt path is left
byte-for-byte alone.

`Applications/CausalLM/chat_template.h` gains four free functions beside `causallm::ChatTemplate` — no new
renderer, no second template source:

| symbol | role |
|---|---|
| `enum class PromptTemplateMode { Auto, Never }` | apply-when-available vs feed-raw |
| `chatTemplateContext(nntr_cfg)` | the package's optional `chat_template_context` object from `nntr_config.json` (empty object when absent) |
| `makeUserRequest(user_text, render_context)` | wrap one user turn into a request |
| `buildPrompt(tmpl, request, mode, render_context)` | the single application point: merges the package context *under* the request (request keys win), then calls `ChatTemplate::apply` |
| `buildUserPrompt(tmpl, user_text, mode, render_context)` | single-turn convenience; returns `user_text` unchanged when nothing is applied |

Three design points, each of which cost something to learn:

1. **One renderer, not two.** The seam consumes the in-tree minja machinery (`third_party/minja`) that
   `ChatTemplate` already wraps. A chat template is a property of the **model package**
   (`chat_template.jinja`, else `tokenizer_config.chat_template`), never of a table in the code keyed on an
   architecture string — two packages of the same architecture legitimately carry different templates.
2. **Render failures propagate.** A template that cannot render is reported, never silently downgraded to a
   raw prompt. An instruction model fed an unmarked prompt drifts, and that failure reads as a model bug
   rather than a configuration one; it has cost real debugging time.
3. **`buildUserPrompt` is not an overload of `buildPrompt`.** A string literal converts to
   `nlohmann::json` as readily as to `std::string`, so an overload pair made every
   `buildPrompt(t, "text")` call site ambiguous. The unit test caught that before any caller could.

**Why a package-level render context.** Some templates branch on a *sibling* render key rather than on the
messages. `ChatTemplate::apply` already forwards any extra top-level request key into the renderer; what
was missing was an owner for the value. The model package is that owner, via an optional
`"chat_template_context"` object in `nntr_config.json`. Concretely: one in-tree model's template emits
`…<|im_start|>assistant\n` and **stops** unless the request carries `enable_thinking: false`, in which case
it emits the `<think>\n\n</think>\n\n` form its validated prompt uses (85 B vs 104 B rendered). A
package-level `"chat_template_context": {"enable_thinking": false}` produces **byte-identical** output to
passing the key per call.

**`unittest_chat_template` — new, 8/8.** No model, no tokenizer, no backend: the shared `chat_template.cpp`
plus gtest, so it runs anywhere the app builds. It pins package-context plumbing, opt-out byte-identity,
request-keys-win precedence, render determinism, and raise-instead-of-raw.

## 2. Apply the model package's chat template in the SDK path

The SDK path could not answer a chat prompt the way the runner does. This makes it do so, through the
*same* renderer — and the equivalence is asserted **byte-for-byte** rather than by inspection.

### Headline gate — the SDK and the CLI produce the same bytes

Same prompt, same model package, four packages. The CLI reference is fed the exact prompt the SDK renders;
the SDK gets the plain user question and renders it itself. Generated text compared byte for byte.

| package (by template source) | CLI reference prompt | SDK input | generated | result |
|---|---|---|--:|---|
| release package — `chat_template.jinja`, 18 KB, untied int4 | package `sample_input` (= the render, 101 B) | "What is the capital of South Korea?" | 158 B | **IDENTICAL** |
| in-tree package — the same 18 KB `chat_template.jinja` | the render (101 B) | same question | 113 B | **IDENTICAL** |
| in-tree package — `tokenizer_config.json:chat_template`, 4 KB string | the render (104 B) | same question + `{"enable_thinking":false}` | 138 B | **IDENTICAL** |
| in-tree base package — **no** template | package `sample_input` (no template → SDK feeds raw) | "The capital of France is" | 143 B | **IDENTICAL** |

**PASS 4/4.** The CLI's streaming display brackets the generated text with one `\n` on each side (a print
artifact of `run()`; the SDK returns `getOutput(0)`, which is the text alone), so equality is asserted with
that bracket removed — both raw forms were kept.

| supporting gate | result |
|---|---|
| four CLI goldens, baseline binary vs post-change binary, same command | **IDENTICAL 4/4** — 160 / 49 / 161 / 145 generated bytes |
| the release package renders + generates its golden **through the SDK** | its `chat_template.jinja` renders to **exactly** the 101 bytes of the package's golden pre-rendered prompt, and the run produces the package's golden answer — reached through the SDK for the first time without a caller pre-templating anything |
| opt-out is byte-identical to today's raw behaviour | **IDENTICAL 2/2** (158 B, 113 B) — and for these packages the opt-out output *equals* the template-ON output, which is the point: the seam renders exactly what a hand-templating caller was passing in |

**Lane note:** these ran on the CPU lane (fp16, greedy where the package asks for it). The seam is upstream
of every backend — it produces prompt *bytes* before any engine is touched — so a GPU-lane repeat is
confirmation, not discovery. This PR does not claim GPU-lane coverage.

### API shape: default-ON with an explicit per-call opt-out

```c
typedef struct {
  bool apply_chat_template;      /* default when options == NULL: true      */
  const char *chat_context_json; /* merged OVER the package's own context   */
  int do_sample;                 /* <0 follow package, 0 greedy, >0 sample  */
} GenerationOptions;

WIN_EXPORT ErrorCode loadModelFromPath(BackendType compute, const char *model_dir);
WIN_EXPORT ErrorCode runModelWithOptions(const char *inputTextPrompt,
                                         const GenerationOptions *options,
                                         const char **outputText);
```

`runModelWithOptions()` returns `CAUSAL_LM_ERROR_INFERENCE_FAILED` when a template is present but cannot
render the prompt: the prompt is never quietly downgraded to raw text behind the caller's back.

Default-ON is not a preference, it is what the existing external contract already says: the shipped SDK
surface external consumers compile against today exposes `apply_chat_template = true`, and its own consumer
example documents `false` as the raw escape hatch. Inverting that default would break the one external
contract that exists. Nothing in-tree pre-templates *through the SDK*; the pre-templating that does happen
is on the **CLI `argv[2]`** path, which is left byte-for-byte alone. There is no double-wrap risk: the
opt-out is explicit and **per call**, not a global mode a caller has to discover, and the seam never
re-wraps text that came out of the seam.

**Compat mechanics.** Additive only: new symbols, no existing signature changed, no struct field reordered,
no field *meaning* changed. `Config`'s existing `bool use_chat_template` keeps its exact sense — including
zero-init = raw — so a stale caller behaves identically, and `runModel()` / `runModelStreaming()` keep both
their template rule and their greedy decoding. Text I/O stays UTF-8 `const char *` on every platform.

### One renderer, not two

`apply_chat_template()` used to fall back to a **hardcoded per-architecture template table**
(`[INST] … [/INST]`, ChatML, `<start_of_turn>`) whenever the package's renderer was missing *or threw*.
Deleted. It was a second template implementation keyed on an architecture string, and it could only ever
guess: two packages of the same architecture legitimately carry different templates, so it silently
mis-templated them and hid a broken template behind plausible-looking markers. A package with no template
now gets its prompt fed as given, with a warning — never re-templated from an architecture guess.

### Three divergences this fixed as a side effect

Each found by the equivalence gate, not by reading the code:

1. **The API kept its own drifted model-registration list.** `register_models()` carried a hand-maintained
   list of nine architectures that had already drifted from the runner's: **3 of the 4 gate packages would
   not load through the SDK at all**, failing with `Unknown architecture`, purely because nobody had added
   them in two places. It now delegates to `causallm::registerAllModels()` (#4115) — Windows exclusions
   included.
2. **The API ignored the package's `do_sample`.** `run()` was called with `do_sample=false` hard-coded, so
   a package whose `generation_config.json` asks for sampling decoded **differently in the SDK than in the
   runner** — reproducibly: same prompt, same 20 prefill tokens, two different answers, measured on a 0.6B
   package that sets `do_sample=true`. The options path now follows the package (tri-state override); the
   legacy `runModel()` keeps its greedy behaviour for compat.
3. **The API load path never called `repack_weight()`**, which the runner has always called after
   `load_weight()`. The two front ends were running different weight layouts for the same package. Both
   load paths now end in a shared `finish_load()` that calls it, so they cannot drift again.

`loadModelFromPath()` exists because there was previously **no way to load a model package at all**:
`ModelType` is a closed enum of one value with a `./models/<name>` path convention, so an SDK consumer
could not be pointed at a package the way the runner is — the equivalence claim above was not even
expressible against the old surface.

### Render cost, and where it sits

| measurement | value |
|---|--:|
| cold renderer construction (18 KB jinja) | **180–195 ms** |
| warm render (renderer cached) | **0.09–0.19 ms** |
| a 4 KB template from `tokenizer_config.json`, cold | 22 ms |
| SDK `init_ms` (create + initialize + load + repack), 0.6B package, CPU lane | 432–444 ms |

It is **not added to the critical path**: the template load happens inside `finish_load()`, i.e. inside the
model-load window, and the trace ordering shows it — `[init-trace] chat-template renderer warm done` prints
**before** the first graph-init trace line, so the first `apply()` joins a completed renderer and pays the
warm cost, not the cold one.

**`test_api` gained the harness:** `test_api <model_dir> <prompt> [raw] [chat_context_json]` prints only the
generated text between markers, so an SDK run can be diffed byte for byte against a runner run on the same
prompt and package — that is how the table at the top was produced. The mode is selected by probing for the
package's `config.json` rather than by looking for a path separator, because a package can be named by a
relative path with no separator in it. The built-in `ModelType` flow is untouched.

## 3. Reuse the loaded model instead of failing or dropping it

`loadModel()` had **no already-loaded guard at all.** A second call for the model already in memory re-ran
the whole config resolution and re-read the weights from disk, so a caller that loads before every request
paid a full load every time. Worse, the load path cleared `g_initialized` and replaced `g_model` **before
the replacement existed**, so a load that failed anywhere after that point — a missing weight file is
enough — left the process holding **neither the old model nor a new one**.

Loading is now idempotent. A request that matches the model in memory returns success and keeps that model:
no teardown, no re-read, no change to `g_model`, `g_initialized` or the chat template. The identity
compared is everything that selects a model — the entry point, the backend, the `ModelType` and the
`ModelQuantizationType`, and the package directory they resolved to, weakly canonicalised. The directory is
part of it because resolution is relative to the working directory, so the enums alone do not pin down
which weights were read. `loadModelFromPath()` gets the same guard keyed on its directory.

A request that **differs** still reloads — handing back the model in memory for a different request would
be a correctness trap — but it now builds the replacement to completion first and then swaps the model, its
chat template, its render context, its sampling default and its identity in one step under `g_mutex`. A
reload that fails leaves the previously loaded model loaded and runnable and returns the error. The cost is
that both models are resident for the length of the swap, which is the price of not being left with
neither.

A reused load leaves `initialization_duration_ms` alone rather than reporting its own ~0 ms: that number is
what the model being run cost to build, and zeroing it would tell a caller polling after a repeat load that
the model came up for free.

Measured on one model package on the CPU backend, host loadavg 16–18:

| case | before | after |
|---|--:|--:|
| repeat load of the same package | 736.01 ms, weights re-read | **0.01 ms, 0 bytes read** |
| failed load after a good one | `CAUSAL_LM_ERROR_NOT_INITIALIZED` on the next generate | generation still succeeds, **byte-identical output**, and the error is returned |

`unittest_load_api` covers both contracts and the cases that must **not** reuse: a different quantization,
a different backend, a different directory, and a model whose identity is unknown.

---

## For the model-package owners (data, not code)

Two package-data findings surfaced by the equivalence gate, deliberately **not** patched in code:

1. **One public package's `chat_template.jinja` is a byte-identical copy of another package's** (`diff`
   empty). Rendering it emits a thought-channel opening, and the model then answers in *thought* style
   ("The user is asking for the capital of … I need to recall …") instead of the direct answer the
   package's own golden records. The package's `sample_input` has **no** such channel, so template and
   golden prompt disagree *inside the package*. Equivalence still holds — both front ends do the same thing
   with it — but the package needs either its real template or a `chat_template_context`.
2. **The 0.6B package needs `enable_thinking:false`** to reach the ChatML + empty-think form its golden
   uses (85 B rendered without it, 104 B with it). Not a renderer defect: the template branches on
   `enable_thinking` and the upstream default is `True`, which here would open a think block nothing
   closes. Fixable in package data, no caller code.

The release package's `sample_input` matches its jinja render byte-for-byte; it is the clean reference of
the four.

---

branch as its base, so this PR targets `main` and its diff therefore also contains #4115's rungs, which are
under review in their own PR and will disappear from this diff as it lands.

Depends-on: #4115


jijoongmoon:
Re-checked against #4316 in the re-cut series: that PR routes the C API through the shared model registry and teaches its chat-template helper the Gemma turn markers, but it does NOT contain the single chat-template seam, the package-template application through it, or the idempotent load that this PR is about. Leaving this open rather than closing it against a PR that does not carry its content.

jijoongmoon:

As noted above, the change here is NOT carried by that series: the shipped SDK path routes through the shared model registry but does not add the single chat-template seam, the package-template application through it, or the idempotent load.

PR #4221, opened 2026-07-30
[CausalLM] Avoid weight shallow copy

jaemini-shin:

<details><summary>[CausalLM] Avoid weight shallow copy</summary><br />

getWeight() returns a reference, but it was bound to a value, shallow copying
the weight tensor on every incremental forward. Bind it
to a reference instead to avoid the copy.

</details>

### Summary
Let's remove redundant copy of weight tensor.

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/lm_head.cpp
+++ b/Applications/CausalLM/layers/lm_head.cpp
@@ -126,7 +126,7 @@ void LmHeadLayer::incremental_forwarding(nntrainer::RunLayerContext &context,
   if (skip_prefill && is_prefill)
     return;
 
-  nntrainer::Tensor weight =
+  nntrainer::Tensor &weight =
     context.getWeight(weight_idx[LmHeadParams::weight]);
 
   nntrainer::Tensor &input_ = context.getInput(SINGLE_INOUT_IDX);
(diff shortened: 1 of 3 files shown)


Jungwon-Lee (approved):
LGTM!
This looks faster now that the extra memory copying is removed. 
How much does this improve peak memory or speed?

Jungwon-Lee:
I found two more straightforward cases of the same pattern:

- `TieWordEmbedding::incremental_forwarding_lmhead()` in `tie_word_embedding.cpp`
- `gate_bias` in `gpt_oss_moe_layer.cpp`

Both receive the `Tensor&` returned by `getWeight()` as a value, invoking the `Tensor` copy constructor. Since they only read the weights and do not require separate lifetime management, could we update them to use references in this PR as well?

haehun (approved):
This is huge reduction!

jaemini-shin:
> LGTM! This looks faster now that the extra memory copying is removed. How much does this improve peak memory or speed?

@Jungwon-Lee
There's no significant change on both peak memory and speed because `Tensor` uses shallow copy for buffer `TensorBase`.
Despite of this, it's trivial that using reference is better than shallow copy!

jayden0701 (approved):
LGTM!

jijoongmoon (approved):
LGTM

PR #4222, opened 2026-07-30
[cuda] Device-resident CUDA lane: RoPE/attention/element-wise/RMSNorm kernels, load-time weight-cache prewarm, device-activation chain, split-KV 32K prefill, pack disk cache, and Engine-wide layer registration

jijoongmoon:
**Task-level PR: get the CUDA lane device-resident and keep it there.** Kernels land, the derived weight
caches are built once instead of per launch (in memory, then optionally on disk), activations stop
round-tripping to the host, split-KV lets a 32K prefill fit, and the custom layers get registered on every
backend the `Engine` actually brought up.

Absorbs #4205, #4207, #4209, #4210 and #4216 — all now closed and pointed here. Splitting this lane apart
is what produced five PRs that each re-included the previous four; this branch already carried **103 of the
104 commits involved** before consolidation, and now carries all of them.

**Merge button: "Rebase and merge" — never squash.** Every rung is individually revertable, and several
(the split-KV opt-in commit vs the default-ON flip; the kernel vs its vectorization) exist *as separate
commits* precisely so the falsification narrative and the policy decision stay independently reviewable.

**Base:** the same ladder lineage the whole C-chain already shares. Containment of #4205, #4207 and #4210
is proven by merge-noop (merging each into this branch yields this branch's exact tree), and #4210's tip
tree occurs verbatim inside this history at `6f9375301`. #4209's one residual hunk and #4216's rung were
cherry-picked (the latter with one seam resolution, described under rung 6).

### Merge-order control that used to be an inter-PR rule

"**#4210 / split-KV merges last of the CUDA chain**" was an ordering constraint between PRs. Absorbed, it
becomes an **intra-PR ordering**, and it holds: inside this branch the split-KV enable commit `9489c35a5`
sits **after** all four split-KV implementation rungs, and everything following it is fixes and additive
opt-in levers. The whole chain now lands atomically, which satisfies the intent — the golden re-baseline
the flip rests on is the final state of this PR, not an intermediate one. **If you want split-KV genuinely
last as its own merge, say so and #4210 goes back to being a separate PR.**

---

## 1. Device kernel library + its first consumer (`mha_core` CUDA dispatch)

Shipped apart, the kernel halves are ~2,460 lines of device code with **zero call sites anywhere in the
tree** — the exact "kernels without a consumer" anti-pattern the GPU contribution checklist is written
against.

- **`cuda_rope`** — split-half RoPE rotation on device fp16 rows, matching the host
  `apply_rotary_emb_tensor_v2` layout, plus a `_dpos` variant that reads the position from the device `pos`
  buffer so a captured graph replays with an 8-byte rewrite instead of a re-record.
- **`cuda_attention`** — fp32 attention core; interleaved fp16 flash-style attention (causal +
  sliding-window mask, GQA, optional softcap, ring-buffer aware via `ring_cap`); split-KV **decode** kernels
  with a prewarm hook for graph-capture-safe workspace allocation; int8-KV quantize and i8 attention kernels
  which are **compiled but never dispatched** (unused until an int8 KV plane exists — named here so nobody
  hunts for the consumer).
- **`cuda_elementwise`** — `geglu`, the fused sigmoid gates, residual add, the graph-decode slot V-copy
  (writes the KV cache at the device-read position with per-row ring mapping: `ring_cap > 0` maps each
  absolute row `% ring_cap`, so multi-row chunk writes stay in bounds on a ring-sized sliding-layer cache),
  per-layer slice copy, and a two-pass on-GPU greedy argmax (fp32/fp16; ties resolve to the lowest index,
  matching `std::max_element`; capture-safe scratch).
- **`cuda_rmsnorm`** — the fp32 kernel and the fp16 **reverse** norm (weight applied *before* the norm,
  `out_scale` folded), plus `rms_maybe_finish`: the post-op drain becomes skippable for device-only
  destinations under `NNTR_CUDA_DRAINSKIP_RMS=1`, default **off**.

All entry points are plain functions taking raw device/UVM pointers and ride only the existing seams (the
NVRTC kernel cache, the stream manager's dispatch/capture, the cuBLAS handle, the integrated-device probe).
Existing entry points keep their signatures.

**First consumer:** the `mha_core` fp16 GPU dispatch plus a device-accessible KV residency option — a RoPE
LUT device-upload helper (a flat `[positions * half]` table cached by table identity; the per-call host LUT
row was **74% of API time** under async submission), decode Q-RoPE on device, K-RoPE written straight into
the UVM/device cache slice, and the GPU V-copy. The cuda context's safe auto-default block already defaulted
`NNTR_CUDA_ROPE` / `_ATTN` / `_KV_UVM` to 1 with a comment marking them *"inert until those consumers
exist"* — this rung **is** those consumers, so they engage by default on a cuda run and an explicit `=0`
opts out per lever. Non-cuda lanes are untouched.

## 2. Build both derived weight-cache families at load, and get argmax off the host

One authority for one lifecycle. Three lifecycle entry points over the **existing** cache machinery (no new
build path): a pointer-keyed idempotent prewarm wrapper; a scratch pre-grow to decode bounds so the `M=1`
FC never `cudaMalloc`s inside a graph capture; and a release entry point for the pointer-keyed caches on
model reload (correctness never depends on it — caches are keyed by weight pointer and rebuild lazily).
Behaviour-neutral by construction: same build, earlier.

**The 1K prefill lever** is prewarming the **cuBLAS int8** weight cache too. It moves **272 lazy per-weight
cuBLAS-i8 `[K,N]` cache builds** — each with device-syncing `cudaMalloc`s that serialize the GPU pipeline —
out of the **first timed prefill** into load, CPU-side and 64 MB-chunked. Profiler-attributed as the
dominant prefill wall-time gap (prefill *kernel* time was already at parity). Side effect: because the
plain UVM payload is only ever host-read at load, its pages stop being migrated wholesale to VRAM.
`skip_prefill` towers and an untied lm_head are exempted from the eager i8 build (their cache would be dead
VRAM); the lazy build remains as the self-healing fallback.

Measured (RTX 5060 8 GB, 1K prompts, M=1073/999/1075/1107, one discarded warm-up + best-of-2 with 45 s
cooldowns, idle-witnessed, golden-gated per cell):

| model | baseline | defaults | profile | pre-series bar | % of bar |
|---|--:|--:|--:|--:|--:|
| 35-layer sliding-window | 220.9 | 5474.5 | **6062.2** | 5014.0 | **121%** |
| gemma4 (qs4cx_fp16) | 203.0 | 9000.0 | **9990.0** | 8121.9 | **123%** |
| gemma2 | 191.1 | 3445.5 | **3732.6** | 3161.8 | **118%** |
| qwen3 | 1058.3 | 5562.8 | **6116.0** | 5125.0 | **119%** |

GPU memory (whole-GPU peak delta) is at or under every bar (−0.1…−0.7%; qwen3 −0.2%, previously +10%) — the
CPU-side prewarm is what keeps the plain pages host-resident. 32K (chunk 1024, KV ring, 29,190-token
prefill): prefill **2496.4** vs bar 2474.1–2488.2 (100.6%), decode 32.45 vs 32.1–32.3 (met).

**A defect in that lever, fixed in the same rung.** The eager build covered **every** prefill-capable QS4CX
FC. Its only consumer is the dispatcher's `M >= 32` branch — and a normal turn prefills far below that (the
golden 20-token turn prefills at M=20/21), so on every short turn that build was **provably dead work**,
paid before the first token where the user is waiting. The prompt is already tokenized when the prewarm
starts, so the largest M the turn can present is known for free (`min(prefill_chunk, INIT_SEQ_LEN)` when the
prompt exceeds a chunk, else the prompt length); the eager i8 build is gated on that maximum. `M >= 32`
also stops being a literal in two places: `cuda_fc_qint4.h` owns `CUDA_FC_I8_PREFILL_MIN_M` and both the
dispatcher and the prewarm read it, so they cannot drift into "build a cache nothing reads".

## 3. Complete the device-activation chain; re-arm `DEV_ACT`

**A half-wired capability this repairs — read this first.** At the tips of the earlier CUDA rungs,
`nntrainer/cuda/cuda_compute_ops.cpp` defined `CudaComputeOps::sigmoid_glu` and `::sigmoid_add` device
overrides while `nntrainer/cuda_context.cpp` registered **neither layer** (grep for `Sigmoid` there: zero
hits). The only carrier of the 18-line registration was a dead-end leaf branch that was not an ancestor of
any later CUDA rung. Merged as the lane was previously stacked, the two device overrides would have been
**unreachable forever**, and any `engine=cuda` graph stamping a fused gate aborts at construction with
`[!] FATAL ERROR: Key is not found for the object. Key: sigmoid_glu`.

The registration is now here: `SigmoidGluLayer` / `SigmoidAddLayer` in `CudaContext::add_default_object`,
appended **last** with **explicit** `int_key`s 9001/9002 mirroring the CL context — because the auto
`int_key` is `str_map.size() + 1`, so a mid-list insertion shifts later auto keys onto explicit ones. (The
layer classes themselves land in #4201.)

The chain:

4. An `nntr_env_on` value-check sweep for the new gates, so `=0` actually disables them, plus the
   async-drain seam.
5. An `NNTR_RMSNORM_CUDA_OFF` row-cap lever (`=all` forces all rows). The auto-profile had been setting
   this lever expecting it to exist; without it a 1K prefill under `DEV_ACT` is a **guaranteed host-loop
   SIGSEGV**.
   default), including reading the embedding row dtype off `cur_weight` rather than the whole-weight handle
   in the device-only cast.
7. The `SigmoidGlu` / `SigmoidAdd` cuda-context registration.
8. Revert of "do not auto-default `NNTR_CUDA_DEV_ACT`" — the device-only activation pool is re-armed on the
   discrete-GPU profile.
9. `NNTR_CUDA_DROP_PLAIN` (opt-in): once every derived cache for a weight exists (dp4a packed [+ cuBLAS
   int8] + fp16 scales), the plain pages are droppable in place, the same way the OpenCL v8c path drops
   them after its backing build.

The point of items 1–3 and 6 is that the graph stops bouncing between device and host per layer: residual
add, per-layer slice, q/k/v-norm and the fused sigmoid/geglu segments all close.

## 4. Split-KV chunked prefill for global-attention layers — **32K prefill +61.9%**

Touches exactly one file (`nntrainer/cuda/cuda_attention.cpp`). Models are expressed as attention classes
(d512-global / d128-uniform), not by name.

**Mechanism.** The serial block-q prefill kernel launches one fixed 32-lane warp per (head, 4-row query
tile) that walks the **entire** causal K range. Once the growing KV exceeds L2 (context ≈ 9K) every warp
streams K/V from DRAM independently and the measured per-key cost plateaus ~1.96x higher — **86.3% of the
d512-global 32K prefill wall lives in that kernel**. This extends the existing decode-only split-KV pattern
to chunked prefill: `attn_blockq_split_d{128,256,512}` walk a **fixed uniform** K partition
(`n_splits = ceil(N_kv/split_len)`, cap 32, deterministically re-derived) with the identical register
online-softmax body and write `(m, l, acc[d])` partials; `attn_blockq_split_reduce` flash-merges per row in
**fixed split order**, serial loops, **no atomics** — byte-stable run to run. Tiles are slabbed under a
64 MiB scratch budget (`NNTR_CUDA_SPLITKV_PREFILL_MB`); slab boundaries never touch numerics, and any
failure (including a misaligned K/V cache base) falls back to the serial kernel.

**Why the vectorization is a separate commit (falsification narrative).** The first-cut split kernel
recovered **0%** — it exactly reproduced the serial time. Standalone-probe attribution showed the walk is
**latency**-bound (~8 warps/SM from the 128-float q/acc register state; a 4x DRAM-traffic cut via
shared-memory K/V staging measured completely **flat**, ruling bandwidth out), dominated by 16 scalar u16
loads per K/V row slice. Vectorized row-slice loads (uint2/uint4-pair; FP arithmetic order untouched,
partials bit-identical by probe) plus `__launch_bounds__(32, {24,16,12})` occupancy floors recovered it:
probe d512 **320 → 169 ms/call (1.89x)**. Rejected by measurement: shared-KV staging (flat), a 2-key
software pipeline, dual/quad concurrent chains, warps-per-block packing.

**32K results** (RTX 5060, chunk 1024, ring on, async+argmax profile, idle-witnessed, warm-up + best-of-2,
serial controls same session):

| cell | serial | split-KV ON | delta |
|---|--:|--:|--:|
| d512-global 32K prefill | 1760.3 | **2837.8–2850.3 TPS** | **+61.9%** (162% of the pre-series bar) |
| d512-global 32K ms/1K tok | 568.1 | 350.8 | −38% |
| d128-uniform 32K prefill | 2489.6 | 2678.7 | +7.6% |
| 32K decode (both) | 29.4 / 32.2–32.4 | unchanged | — |
| GPU peak | — | +64 / +42 MiB | the scratch budget |

1K cells are **byte-inert**: below the 4096-key engage threshold the serial kernel runs verbatim (lever-on
vs lever-off byte-identical), and decode (`N_q == 1`) never enters the path.

**Numerics, and the default-ON statement.** Device probe (random fp16, `N_kv=9216`, `d=128/256/512`, softcap
on/off): split vs serial differs on **0.67% of fp16 outputs by at most 1 ulp**; against an fp64 reference
both kernels are **equally accurate** (rms 6.58e-6 vs 6.58e-6). The d512-global 32K stream is byte-identical
to serial end to end. The single d128-uniform 32K divergence is **one near-tie argmax flip** in the
generated continuation (both continuations coherent), **re-baselined as the canonical golden** — with the
serial stream retained as the bisect handle: `NNTR_CUDA_SPLITKV_PREFILL=0` reproduces it byte-for-byte
(`=N` sets a custom split length). Run-to-run byte-stability verified x3 on both 32K classes.

## 5. Register the device-routed custom layers on every backend the Engine brought up

A model graph stamps `engine=` on the nodes it wants off the host, and the registry decides which backend
that name resolves to. So **every type a graph can stamp must be constructible on every backend the process
brought up**, or graph construction dies with `Key is not found: <type>`. That invariant was maintained by
hand in five places, and every place had drifted:

| type | cpu | gpu (OpenCL) | cuda |
|---|---|---|---|
| `reshaped_rms_norm` | 4 per-model registrations | `transformer.cpp` | **never** |
| `per_layer_slice` | 1 (one model) | `transformer.cpp` | **never** |
| `rms_reverse_norm` | **never** | **never** | `causal_lm.cpp`, behind an `#if` |

Every model stamping a per-head q/k/v norm therefore aborted before its first token; only the one model
whose graph happens to stamp types the core `CudaContext` registers itself could run. Three more defects
hid behind that one:

- `logit_softcapping` is a promoted core layer that `AppContext` and `CudaContext` register and `ClContext`
  alone did not, so one model's final soft-cap had to stay unstamped — and an unstamped node re-binds the
  **cpu** op table onto a **device** graph's logits (the activation pool is chosen graph-wide, the
  ComputeOps table per node). On `engine=cuda` that is a SIGSEGV in `CpuComputeOps::softcap` on a
  `cudaMalloc` pointer.
- Both device-routed app layers ended their device path with a **host fallback**, which is a SIGSEGV on the
  device-only activation pool. `per_layer_slice` is *repaired* (its fallback is a pure copy, so it routes
  through `cudaMemcpy(cudaMemcpyDefault)`; the FP16 branch probed only the input and the output is the half
  that faults, and the FP32 branch had no probe at all). `reshaped_rms_norm` is *guarded* — host norm math
  cannot run on memory the CPU cannot address, so it reports the contract violation by name instead of
  faulting.
- **`libcausallm.so` never linked `librms_reverse_norm_layer.so`** — `layers/meson.build` builds it,
  `causallm_layer_dependencies` never listed it, and `causal_lm.cpp` already referenced the class inside its
  CUDA block. **A plain `-Denable-cuda=true` app build has therefore never linked.** Verified by reverting
  this change set on the parent and reproducing the identical link error. If CI is green on the parent, CI
  is not building the CausalLM app with CUDA enabled.

**The shape of the fix:** one central table in `Transformer::registerCustomLayers`, walked once against a
new `Engine::getRegisteredContextNames()` — coverage becomes exhaustive **by construction** instead of by
review. Six hand-maintained sites collapse into it; no model file names a backend any more. The single
`try` that wrapped three `registerLayerFactory("gpu", …)` calls is also split, so one throw can no longer
skip the other two (the source of the misleading `gpu Context is not registered` line printed on every
non-OpenCL run).

### What the lane measures once it is constructible

RTX 5060, `pgrep` idle witness + loadavg per row, warm run discarded, best of 2. `bar` is the pre-ladder
reference matrix. GPU column is whole-GPU `memory.used` peak delta.

**CUDA 1K — three of these four cells did not exist before (rc=1, 12/12 runs):**

| model | prefill TPS | decode TPS | init ms | GPU peak MiB |
|---|--:|--:|--:|--:|
| 35-layer sliding-window | 5014.0 → **6166.7** | 62.26 → **80.60** | 916–920 → **568** | 5924 → **3526** |
| gemma4 | 8121.9 → **9990.0** | 67.23 → **71.64** | 1259–1280 → **775** | 5620 → **2682** |
| gemma2 | 3161.8 → **3706.9** | 60.49 → **74.77** | 598–616 → **409** | 5504 → **4732** |
| qwen3 | 5125.0 → **6254.2** | 105.96 → **124.51** | 362–364 → **1692** | — |

gemma2 is the non-regression control: it ran before, and reproduces the pre-change build to the digit
(3694.2 → 3706.9, 74.94 → 74.77, GPU peak 4732 → 4732). The bars also hold on **shipped defaults** with no
profile env exported (6096.6 / 9891.1 / 3668.9 / 6219.1 prefill).

**CUDA 32K** (`NNTR_PREFILL_CHUNK=1024`, KV window ring, M=29,190) — both cells also did not exist: 35-layer
long-context **2779.2** TPS (bar 2474–2488) / decode **36.78** (bar 32.1–32.3); gemma4 **2923.7** (bar
1755.8). Split-KV engagement proven by its own kill-switch (−8.0% / −63.4% with
`NNTR_CUDA_SPLITKV_PREFILL=0`), not by a log string.

**CUDA init**, short golden turn, warm caches: full-span **≈644 ms** sidecar / **≈768 ms** canonical.
Caveat: `init` is blind to the NVRTC tax — with the cubin and JIT caches cold it does not move while the
first prefill absorbs the compile.

**Intel/OpenCL non-regression:** two cells re-measured back-to-back against the unmodified parent binary in
the same session; both land at or above it.

**Goldens:** cpu-vs-cuda is byte-identical for gemma4; two models differ at a single token where the cuda
output *agrees with the OpenCL lane*, i.e. host fp16 is the outlier. All twelve previously-working lane
comparisons (cpu x4, OpenCL x4, and the pre-change cuda model) are byte-identical.

## 6. Derive-once disk cache for the QS4CX weight packs (opt-in, **default off**)

**The default is the measurement, not caution.** On this lane one of the two derived weight families is
worth caching and the other is not, and the disk axis disagrees with the latency axis.
`NNTR_CUDA_PACK_CACHE=1` opts in.

Every launch re-derives the same derived weight bytes from the same plain nibbles: the dp4a packed-int4 +
row sums, and — when a prefill can reach the cuBLAS-i8 gate — the cuBLAS-i8 `[K,N]` transpose-unpack. This
persists them once and uploads from the mapped pack instead.

| cell (card active) | n | full-span TTFT | prewarm lap | disk |
|---|--:|--:|--:|--:|
| short turn, no cache | 5 | 2138 | 332 | — |
| short turn, pack HIT (**page cache warm**) | 4 | **2034** [2020–2042] | **232** | 1373 MB |
| short turn, pack HIT (**page cache COLD**) | 1 | **2854** | **1150** | 1373 MB |
| short turn, pack MISS (writes it) | 1 | 2445 | 672 | → 1373 MB |
| long prompt (1.2K), no cache | 3 | 3424 [3316–3448] | 1497–1505 | — |
| long prompt, pack HIT | 3 | **2378** [2262–2406] | **425–435** | 2.6 GB |
| long prompt, pack MISS (writes it) | 1 | 4020 | 2150 | → 2.6 GB (695 records) |

So: **−104 ms on a warm short turn, +716 ms on a cold page cache, −1046 ms (−31%) on a long prompt**, at a
cost of **+1.4 GB / +2.6 GB of disk** per model.

The rung also adds permanent cost accounting (`cuda_fc_qs4cx_prewarm_stats`, reported as a `split` log line
when the cache is enabled), because the question is not "how long is the prewarm" but "how much of it is a
**derive** a cache could remove":

| workload | host derive | H2D | pack tee (MISS) | pack HIT | payload |
|---|--:|--:|--:|--:|--:|
| short turn (i8 gated off) | **150.9** | 83.8 | 214.7 | **107.2** | 1366 MB |
| 1.2K prompt (i8 engaged) | **1114.8** | 193.0 | 408.1 | **206.3** | 2570 MB |

The dp4a derive is bandwidth-bound and 16-way parallel (151 ms for 1366 MB), so caching it trades CPU for
I/O and only pays warm. The i8 `[K,N]` derive is the opposite — a column-strided transpose-unpack, 1115 ms
for 2570 MB — and caching that wins a full second. Latency, memory and disk do not agree, so the default is
off with a recommendation: **enable it for prefill-heavy deployments** (long prompts, where the i8 records
exist and the page cache stays hot) that can afford 2.6 GB per model.

**Peak RSS is unchanged in every arm** (4431.6–4432.4 MB) — the hit path uploads from the pack mmap in
64 MB chunks and `MADV_DONTNEED`s each consumed chunk, so transient residency is one chunk, not the pack.
Prefill TPS (5457 → 5457–5485) and decode (42.5 → 42.4–42.6) are unchanged, i.e. the cached bytes **are**
the derived bytes.

`cuda_pack_cache.{h,cpp}` is a structural port of the OpenCL lane's v8c pack cache: same on-disk layout
(64 B header / blob regions / **index appended last**, so a temp file that never finalized cannot
validate), same identity rules, same failure semantics, same background-finalizer-joined-at-exit write
path. The two lanes now read the same way. **Identity is by FILE + NAME, never by pointer.**

**Consolidation seam, disclosed.** This rung was cut on the prewarm branch, before the plain-page drop
existed, so its prewarm call and the `NNTR_CUDA_DROP_PLAIN` teardown met at the same line in
`causal_lm.cpp`. Resolved by keeping the drop seam and threading the pack-cache name through it: the
prewarm now takes `pack_name.c_str()` and the opt-in drop runs **after** it, so a pack rewrite still sees
the plain payload it derives from. That ordering is recorded in a comment at the call site.

---

## Still missing for full CUDA parity — not fixed here

The device-activation chain is not complete: `reshaped_rms_norm`'s guard reports a contract violation rather
than computing on device, so a device-only activation pool still cannot run every graph. That needs its own
change set. This PR's claim is narrower and exact: **the lane is constructible and correct for all four
models, and its numbers clear every bar.**

## Merge order

Land **#4194** (main-direct inference correctness) and **#4212** (cold start) first, so the rungs this
lineage re-includes arrive as no-ops under rebase-and-merge rather than as conflicts. This PR touches the
same `causal_lm.cpp` / `transformer.cpp` seams as the app-side edits under review elsewhere in the series;
merging it before those keeps the resolution to context only.

---


jijoongmoon:
Superseded in part by the re-cut series against current main: #4313, which carries this PR's CUDA-library half — the RoPE, attention, element-wise and RMSNorm device kernels, the device-activation chain and the Engine-side registration of the CUDA context — rebased onto main.

The rest of this PR is not in #4313 and is not being dropped: the split-KV **chunked prefill** belongs with the long-context work (#4313 ships the split-KV decode kernels only), and the app-side layer and KV-cache changes belong with the CausalLM rungs. For those, a replacement bundle is to follow; the branch is preserved at `upstream-pr/trackC13-cuda-lane-registration`.

The review's first structural point is also settled: this PR and #4228 were the same 224-file tree pushed to two numbers, and neither number is being used for that tree again — the series cuts one subsystem per PR, which is the review's second point.

Closing to keep the queue reviewable.

PR #4128, opened 2026-07-21
[Tensor/cpu_backend/arm] KAI f16 int4 (qai8dxp/qsi4cxp) matmul + disp…

jijoongmoon:
Add an ARM KleidiAI f16-in / f16-out int4 matmul path for the qai8dxp / qsi4cxp (per-channel int4 = QS4CX) family, so QS4CX-FP16 FCs and the untied QS4CX lm_head run without an fp16<->fp32 cast pass over the LHS/DST tensors. The QS4CX ARM base (qs4cx_tensor, KAI rhs packing, the f32 gemm_qai8dxp_qsi4cxp path, getPackedData) is already upstream; this adds the fp16 variant on top.

Kernel + pack:
- kai/matmul_clamp_f16_qai8dxp_qsi4cxp/...4x4x32_neon_i8mm.{c,h}: forked 4x4x32 NEON i8mm micro-kernel that does the fcvtn to fp16 just before the store (i8mm-gated, __ARM_FEATURE_MATMUL_INT8).
- kai/pack/kai_lhs_quant_pack_qai8dxp_f16.{c,h}: forked LHS quant-packer that reads __fp16 directly (pure NEON, no i8mm dependency).
- kleidiai_interface{.h,_qai8dxp_qsi4cxp.cpp}: the additive public entry nntr_kai_gemm_qai8dxp_qsi4cxp_olp_f16() (packs the LHS from fp16, parallelizes the matmul over N via ThreadManager). i8mm-gated to match the f32 entries.
- meson: register the new f16 matmul subdir + f16 pack source.
- tests: q4_kai_bytecompat / q4_roundtrip_kai / q4_kai_matmul_device.

Dispatch:
- arm_compute_backend_fp16.cpp: the nntr_gemm_qai8dxp_qsi4cxp_packed<_FP16>
  facade that packs the LHS and calls the kernel above.
- arm_compute_backend.h: the paired packed/_unpacked template decls.
- half_tensor.cpp: a QS4CX case that routes a QS4CX-FP16 FC through the facade, taking the pre-packed KAI weight via the existing getPackedData.
- avx2_impl.{h,cpp}: retype the x86 AVX2 fp16 signatures _Float16* -> _FP16* (a no-op on native, needed so those files compile under the MSVC Half wrapper).

[...]

PR #4223, opened 2026-07-31
[tensor] Load legacy per-channel int4 (QINT4) weight records as QS4CX

jijoongmoon:
**Depends-on: #4128** (base — the KAI `qsi4cxp` Section-A packing helpers this extends arrive there).
**Land #4212 first** if both are in flight: this branch re-includes the QS4CX single-zero-fill rung, which then arrives as a no-op.
**Rebase and merge only, never squash.**

Every model package whose `model_tensor_type` is `QINT4-*` fails to load on this tree. The loader
over-counts each int4 weight, the running file offset drifts, and the EOF tripwire fires. Reproduced
on three packages (a 35-layer model with an untied QINT4 lm_head, and two smaller ones):

```
FATAL ERROR: Model layout mismatch: graph expects 3,539,836,016 bytes of weights
             but file '...' has only 3,399,793,776 bytes
```

## Root cause

The on-disk int4 record and the in-memory int4 tensor are two different formats, and only one of them
is implemented.

**On disk**, an int4 record carries **one scale per output channel** — either
`[u16 0x06][KAI qsi4cxp Section-A nibbles][N fp16 scales]` or the plain container
`[u16 0x01][padded nibbles][N fp32 scales]`.

**In memory**, `Int4QTensor::scale_size()` returns the upstream group-of-32 count `H*W/32`, so
`getMemoryBytes()` — which is what advances `start_from` — claims `H*W/16` scale bytes instead of
`2*N`. For the package above that is **142,467,072 expected vs 2,424,832 on disk**: exactly the
140,042,240-byte delta in the error.

Nothing in this tree can *write* a group-32 int4 record, so there is no such package to preserve.

## The fix

A QINT4 weight materialises as QS4CX, and the legacy record is transcoded at read time — both
containers. The nibbles are re-laid out and the fp16→fp32 scale widening is exact, so the transcode
is **lossless**: a legacy package and its regenerated QS4CX twin emit byte-identical continuations.

## Verification

- The three previously-unloadable packages load and generate; the 35-layer one is **token-identical
  to a reference binary** on the same package.
- Canonical (non-legacy) packages are untouched — the flag is off, so `TensorBase::read` is unchanged.
- Verified on both the CPU engine and the OpenCL lane.

### DIFF
--- a/nntrainer/layers/layer_context.h
+++ b/nntrainer/layers/layer_context.h
@@ -82,8 +82,21 @@ public:
    * @return Tensor DataType of the the Weight
    */
   TensorDim::DataType getWeightDataType() {
-    return str_converter<enum_class_prop_tag, nntrainer::TensorDataTypeInfo>::
-      from_string(tensor_type[1]);
+    auto dt =
+      str_converter<enum_class_prop_tag,
+                    nntrainer::TensorDataTypeInfo>::from_string(tensor_type[1]);
+    // A QINT4 weight dtype -- whether inherited from a "QINT4-*"
+    // model_tensor_type graph default or set per layer -- is materialised as
+    // the canonical QS4CX class: both are the same KleidiAI qsi4cxp
+    // per-output-channel int4 format, and QS4CX is the one every compute path
+    // in this tree consumes. The on-disk bytes stay the legacy QINT4 record and
+    // are transcoded at read time (NeuralNetwork::load flags the tensor,
+    // QS4CX_Tensor::read does the re-layout), so existing QINT4 packages keep
+    // loading unchanged. This is the single point that makes the mapping
+    // universal across every model.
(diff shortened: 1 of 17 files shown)


jijoongmoon:
Superseded by the re-cut series against current main: #4305. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

Specifically: `Int4Utils::readLegacyQint4RecordToQs4cx` plus the `QS4CX_Tensor::read()` transcode carry this content, and `unittest_nntrainer_cpu_backend.qint4_legacy_reader_to_qs4cx` pins the reader against the in-place transcode. The branch is preserved at `upstream-pr/trackLegacy-qint4-record-compat`.

PR #4224, opened 2026-07-31
[opencl] Cut OpenCL cold start: prewarm v8c programs, cache the derived weight pack

jijoongmoon:
**Depends-on: #4113** (`trackG6-load-perf`, the branch this is cut on — both rungs edit the v8c region
`nntrainer/tensor/cl_operations/blas_kernels.cpp` that #4110 introduces and #4113 parallelizes).
**Rebase and merge only, never squash.**
Cross-reference: **#4212** is the main-direct half of the same cold-start story; **#4222**'s
`cuda_pack_cache` is the sibling implementation of rung (b) on the CUDA lane.

The OpenCL half of the cold-start budget. Two rungs, one number.

## (a) Prewarm the v8c programs the first prefill used to compile

Eager registration leaves a tail. The v8c GEMM / activation-quantization program is first built
*inside* the first prefill, and the int4 lm_head GEMV program inside the first decode — time the user
waits for the first token, and **invisible to any measurement of initialization**, which has already
returned.

Building them during bring-up has to happen from the TU that owns their sources *and their compile
options*: registering a program under different options produces one the hot path never looks up, so
the compile is paid twice and nothing leaves the critical path. The collector must also not call
`ClContext::Global()` — directly, or indirectly via `v8c_use_buffer_path()` — because it runs inside
that context's own one-time initialization and would wait on itself. Both traps are written down at
the declaration and the definition.

Measured on an Intel Xe3 iGPU with the binary cache **and** the driver cache empty, 3 samples each,
every sample passing an all-compiled / none-cached gate:

| | before | after |
|---|--:|--:|
| process start → first token | 3003–3021 ms | **2238–2489 ms** |
| mean | 3009 ms | **2354 ms (−22%)** |
| first-token lap | ~1.26 s | **346–362 ms** |

Warm runs are unchanged and the generated text is byte-identical. Flash-attention programs are
deliberately excluded: their compile options bake in head dimension and tunables that are not known
until the model is constructed.

## (b) Derive the v8c weight pack once and keep it on disk

Building a v8c GEMM weight permutes plain QS4CX nibbles into the GEMM byte layout and folds
per-channel row sums — a deterministic pure function, redone every launch. For an untied lm_head it is
**the longest single-node span of the model load: 250–270 ms of the loader's join.**

The pack is persisted next to the model (`<model>.v8cpack`, falling back to
`$XDG_CACHE_HOME/nntrainer/v8cpack`). A later launch maps it, skips permute and row-sum entirely and
uploads straight from the mapped pages: **37–46 ms for the 336 MB lm_head pack.** Goldens are
byte-identical. Memory-neutral.

**Identity is by name, never by pointer.** This tree has already had a derived-weight cache keyed by
data pointer serve the wrong entry once pool and SVM addresses recycled, so: the pack is bound to the
source file by size/mtime/format-version, each record by the weight **name**'s hash plus
N/K/row-stride/payload-length, guarded by a sampled payload checksum (both 64 KB ends plus 16 interior
pages — bounded at ~192 KB/record so validating can never cost what the cache saves) plus a full
checksum over the row-sum block. Any mismatch silently falls back to deriving.

Writing costs the load nothing: the miss path tees each packed chunk to a temp file with `pwrite` at
disjoint offsets, and index/header/fsync/atomic-rename happen on a background thread joined at exit,
so a partial rewrite never clobbers a valid pack.

Fault injection: a bit flipped inside the sampled window derives instead; 60% truncation and a header
identity mismatch both invalidate and rewrite; the rewritten pack reproduces the original byte for
byte. `NNTR_V8C_PACK_CACHE=0` leaves no cache code on the path.


jijoongmoon:

Detail. This branch is a stack whose lower rungs are already open as #4304, #4306, #4308, #4309, #4310 and (its immediate parent) the upload work now in #4311; #4311 carries this branch's own top two commits, the program prewarm and the derive-once pack cache. `v8c_pack_cache.h` is listed in the source list only, never in the installed header set, so it adds no unpackaged header to the Tizen and pdebuild legs.

The review's blocking findings were all real and are fixed in #4311's last commit:

- **(1) integer overflow in the on-disk header validation.** Correct, and correctly identified as a trust-boundary problem rather than a robustness nicety: the pack lives beside the model or under a directory an environment variable names. Every bound is now checked in subtraction form (`off > len - len_field`, with `len_field > len` rejected first), for the index region and for both blob regions of every record. The index count is additionally capped before it drives a `resize`, so a corrupted count cannot allocate gigabytes ahead of the check. A file shorter than the header was already rejected.
- **(2) the identity cannot see a same-size, same-mtime replacement.** Correct. Each index record now also carries a sampled fingerprint of the source nibbles it was derived from, and a hit must match it; the format magic is bumped to `NTV8PC02` so a pack written by the previous layout is rejected rather than reinterpreted with a shifted record size. The fingerprint is sampled, so it costs a fixed amount whether the lookup hits or misses.
- **(3) no tests for a new on-disk format.** Still true, and stated as an open item on #4311 with the reviewer's own argument for why it should be closed: the round-trip, the truncated file, the identity mismatch and the overflow case are all pure file I/O and need no OpenCL device. It belongs with whichever change next touches this file.

Non-blocking: **(4)** the unchecked `mkdir` results are unchanged and worth a follow-up; **(5)** the merge-order constraint is moot now that the dependency is a normal stacked PR; **(6)** the file is still POSIX-only and #4311's body says so; **(7)** is answered by the re-cut — #4311's own diff is 12 files.

The measured cold-start result carried over unchanged and is quoted on #4311.

The branch is preserved at `upstream-pr/trackInit-v8c-cold-start`, so this is reversible.

PR #4225, opened 2026-07-31
[causallm] Embedding sidecar LUT: mmap the payload read-only (lazy page-in)

jijoongmoon:
The embedding sidecar lookup table declared by `quantized_lut_path` was read whole into a heap
`std::vector<uint8_t>` at `finalize()`, so a multi-hundred-MB table became fully resident before the
first token was produced. This series maps the payload read-only instead, so rows page in on demand.

Base: this branch applies on top of `main` (base commit `308bdd4f3`, "[kleidiai] Add
`matmul_clamp_f16_qai8dxp_qsi4cxp`"). It has no dependency on any other pull request. `main` has moved
four commits ahead of that base since; none of those four touch
`Applications/CausalLM/layers/embedding_layer.{cpp,h}` or the sidecar unit test, so it still merges
cleanly. Rebasing onto current `main` is fine if preferred.

## Commits

1. `[causallm] Embedding sidecar LUT: mmap the payload read-only (lazy page-in)`
2. `[causallm] Sidecar LUT: make the unsupported-datatype error actionable`
3. `[causallm][OPTIONAL] Sidecar LUT: accept GGML q4_0/q6_k row payloads`
4. `[causallm][test] Sidecar LUT unit test: check payload_size(), not bytes.size()`
5. `[causallm] Sidecar LUT: restore the two-ways-forward text after GGML support landed`

Commit 3 is separable. Dropping it keeps GGML sidecar support deferred and leaves the actionable error
from commit 2 in place; commits 4 and 5 still apply. It is included because it is routing over the
dequant loop this file already runs (`dequantize_row_q4_0` / `dequantize_row_q6_K` were already consumed
here, so it adds no dependency), and because it is what makes the lazy page-in reachable for the
q4_0/q6_k sidecar packages that exist in practice.

## What changes

`QuantLut` gains `mmap_ptr`/`mmap_len` and `data()`/`payload_size()` accessors. `attachPayload()` opens
the file and calls `::mmap(PROT_READ, MAP_PRIVATE)` plus `madvise(MADV_RANDOM)` — token-id lookups are
random access, so readahead would defeat the point. If the map fails for any reason, it falls back to the
previous full read into `QuantLut::bytes`; every consumer now reads through the accessors, so decode
behaves identically on either backing store. `QuantLut` becomes non-copyable and owns the mapping,
unmapping in its destructor. On Windows the same `::mmap` call resolves to `MapViewOfFile` through the
existing `utils/mman_windows.h` shim, which also faults pages on demand. The process-wide `weak_ptr`
dedup in `get_or_load_quant_lut()` is unchanged and now dedups the mapping itself.

Commit 3 adds a `"q4_0"`/`"q6_k"` manifest datatype whose payload is the same row table an in-bin
embedding weight would hold. Decode reuses the existing `incremental_forwarding` dequant loop unchanged;
only the row base pointer is selected between the weight tensor and the mapping, and the row stride is
hoisted to the value the previous per-branch block math already produced. Sidecar output is therefore
bit-identical to the monolithic package by construction, not by coincidence.

## Effect, and what is actually measured

The honest claim is narrow, so let me separate what follows from the code from what rests on numbers.

**Provable by construction:** on the mmap path the heap vector is never allocated, so the payload is no
longer counted as anonymous resident memory at load. That is a property of the diff, not of a benchmark.

**Not established by this PR:** the magnitude. The commit messages carry figures from an earlier tree —
RSS reduction of roughly 1.27 GB, model init about 41% faster, load residual about 44% lower. Those were
single-run observations on a different tree and have **not** been reproduced on this branch. I am citing
them as the reason the change was made, not as a result this PR demonstrates. Treat them as an
expectation to be re-measured, and note that memory numbers on the devices in question have shown
run-to-run spreads wide enough to swallow differences of this order, so a single run per arm is not
sufficient evidence. I would rather understate this than repeat an earlier overclaim.

**A tradeoff worth naming explicitly:** commit 3 adds an `MADV_WILLNEED` prefetch to cover cold-start
prefill, because `MADV_RANDOM` disables readahead and a long prefill would otherwise serialize one major
fault per token (about 100–160 ms on NVMe on the earlier tree — again single-run, not re-measured).
`MADV_WILLNEED` pulls pages in, which pushes RSS **up**, so this is latency bought with memory, not a
free win. Two things bound the cost: it is gated on `iter > 1`, so it never fires during single-token
decode, and it advises only the specific rows of the incoming batch — token count times row stride, not
the table. The residency cost of that bounded set has not been measured. Reviewers who care more about
peak memory than prefill latency can drop commit 3 and keep the page-in behaviour.

## Compatibility

Packages without a `quantized_lut_path` are byte-identical by construction: no `QuantLut` is created,
`ggml_lut` is false, the loop reads the same weight rows through the same branches, and the only work
removed is the construction of a `Tensor` view that was never used on the quant branches.

## Verification status

Not run for this PR: no build, no model execution, no re-measurement. The claim that sidecar output is
bit-exact against the monolithic package follows from the shared decode path rather than from a run on
this branch. Please treat CI as the first compile signal. Two areas deserve reviewer attention in
particular: the Windows `_wopen` / `mman_windows.h` path, which no build here exercised, and the unit
test, which only compiles when tests are enabled — commit 4 fixes two assertions that read
`QuantLut::bytes` directly and would fail once the mmap path starts succeeding on a real filesystem.

Rebase and merge only, never squash.

### DIFF
--- a/Applications/CausalLM/layers/embedding_layer.cpp
+++ b/Applications/CausalLM/layers/embedding_layer.cpp
@@ -276,6 +276,49 @@ std::shared_ptr<QuantLut> loadSfixed4Manifest(const std::string &manifest_path,
   return lut;
 }
 
+/**
+ * @brief GGML row-block sidecar: the payload is the byte-identical Q4_0/Q6_K
+ *        row table an in-bin embedding weight would hold, so decode reuses
+ *        dequantize_row_q{4_0,6_K} and the outputs match the in-bin path
+ *        bit-exactly. Manifest:
+ *          {"datatype": "q4_0"|"q6_k", "size": <out_dim>,
+ *           "rows": <in_dim, optional>, "lut-path": "<payload>"}
+ */
+std::shared_ptr<QuantLut> loadGgmlManifest(const std::string &manifest_path,
+                                           const nlohmann::json &json,
+                                           nntrainer::TensorDim::DataType dt) {
+  const auto lut_path = requireJsonStringField(json, "lut-path", manifest_path);
+
+  auto lut = std::make_shared<QuantLut>();
+  lut->out_dim = requireJsonSizeField(json, "size", manifest_path);
+  lut->ggml_dtype = dt;
+
+  const size_t block = (dt == nntrainer::TensorDim::DataType::Q6_K) ? 256 : 32;
+  const size_t block_bytes =
+    (dt == nntrainer::TensorDim::DataType::Q6_K) ? 210 : 18;
+  NNTR_THROW_IF(lut->out_dim % block != 0, std::invalid_argument)
+    << "Malformed LUT manifest " << manifest_path << ": size " << lut->out_dim
+    << " must be a multiple of the " << block << "-wide quant block";
+  lut->row_bytes = block_bytes * (lut->out_dim / block);
+
+  attachPayload(*lut, resolveLutPath(manifest_path, lut_path));
+  NNTR_THROW_IF(lut->payload_size() == 0 ||
+                  lut->payload_size() % lut->row_bytes != 0,
+                std::runtime_error)
+    << "LUT binary size " << lut->payload_size()
+    << " is not consistent with row stride " << lut->row_bytes << " for "
+    << manifest_path;
+  lut->in_dim = lut->payload_size() / lut->row_bytes;
+
+  if (json.contains("rows")) {
+    const size_t rows = requireJsonSizeField(json, "rows", manifest_path);
+    NNTR_THROW_IF(rows != lut->in_dim, std::invalid_argument)
+      << "LUT manifest " << manifest_path << " declares rows=" << rows
+      << " but payload holds " << lut->in_dim;
+  }
+  return lut;
+}
+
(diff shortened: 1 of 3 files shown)


jijoongmoon:
Superseded by #4318, which lands the read-only mmap of the sidecar payload.

The payload is opened `O_RDONLY` and mapped `PROT_READ` / `MAP_PRIVATE`, with
`MADV_RANDOM` advised and the mapping released in the lookup table's destructor;
reading the file into a vector stays as the fallback when the mapping cannot be
made. On Windows the same is done through `MapViewOfFile` via the mman shim.

Two notes beyond this PR's scope. Because `MADV_RANDOM` disables readahead, a
pre-pass advises `MADV_WILLNEED` over the page ranges of exactly the row ids each
batch will touch, so a long prefill does not pay one synchronous major fault per
token. And the 4-bit per-block LUT half of the sidecar work lands separately, in
the following rung, so this PR's mmap point is the part that closes here.

Closing; the branch is untouched.

PR #4226, opened 2026-08-03
[htp] HVX kernel device bring-up on CDSP

dlwlzzero:
## Dependency of the PR

None. This is a standalone device-side test that links neither libnntrainer nor test_util; it only uses FastRPC and gtest.

<details><summary>8755f959 [htp] Add the FastRPC interface and DSP skel build for HVX bring-up</summary><br />

[htp] Add the FastRPC interface and DSP skel build for HVX bring-up

Defines a one-method IDL and the build script that turns it into a
loadable CDSP skel. qaic generates the stub and skel, hexagon-clang
builds them against v79 with HVX enabled; nothing here is hand-written
plumbing.

add_f32 returns AEE_EUNSUPPORTED for now, on purpose. The next commit's
test starts red against that return, which separates "the session
opened and the call reached the DSP" from "the kernel computes the
right values" -- two failures that are otherwise easy to confuse.

The build stays a manual step rather than a meson or ndk-build hook.
One kernel does not justify wiring a cross-compiler into the main build,
and a hook would have to be guarded for every environment without the
Hexagon SDK installed.

Co-authored-by: Sisyphus <noreply@opencode.ai>

**Self evaluation:**

</details>

<details><summary>960075e3 [htp] Verify an HVX kernel end to end on device</summary><br />

[htp] Verify an HVX kernel end to end on device

An nntrainer gtest opens an unsigned-PD CDSP session, calls an HVX
elementwise add across FastRPC, and compares the result bit for bit
against the CPU. 100 floats crosses three full 128-byte vectors and a
four-element tail, so one case covers both paths.

A failed session open is a FAIL, not a SKIP. Whether the DSP comes up on
the device is exactly what this test measures, and a skip would report
success for the measurement being absent.

Nothing locks HVX. The SDK's own profiling example uses HVX intrinsics
inside a skel with no lock -- the RPC thread already holds it -- so the
kernel only checks that a 128-byte context exists at all. Loads go
through HVX_UVector because FastRPC makes no alignment promise about
the buffers it hands the skel.

The module builds only under ifdef HEXAGON_SDK_ROOT and links neither
libnntrainer nor test_util; it needs FastRPC and gtest and nothing else.
Environments without the Hexagon SDK build exactly as before.

Deviation from plan (two changes, both required to build):

1. The test source is named unittest_hvx_kernels.cpp, not the plan's
   unittest_hvx_add.cpp. The LOCAL_MODULE (binary name) stays
   unittest_hvx_add so the existing adb push / run commands are
   unchanged.

2. The test defines its own int main(int argc, char**). The plan
   expected googletest_main to supply main(), but this repo's
   googletest_main module (Android.mk:82) compiles only gtest-all.cc;
   gtest_main.cc is absent. Every other unittest in this tree defines
   main at the bottom of its .cpp (e.g. unittest_common_properties.cpp:268). The
   new test follows that pattern rather than re-architecting the shared
   gtest module, which would touch every other test's build.

Co-authored-by: Sisyphus <noreply@opencode.ai>

**Self evaluation:**

</details>

<details><summary>8652ecf4 [htp] Cover the below-one-vector input for HVX add</summary><br />

[htp] Cover the below-one-vector input for HVX add

n=100 never exercises n_vec == 0, so the scalar tail runs unguarded for
any input shorter than a single 128-byte vector. This case pins that
boundary.

It likely passes without touching the kernel, which is the point: it is
a regression guard against a future rewrite that assumes at least one
full vector.

Co-authored-by: Sisyphus <noreply@opencode.ai>

**Self evaluation:**

</details>

### Summary

- Define a one-method FastRPC IDL (`nntr_hvx`) and a manual `build.sh` that runs `qaic` + `hexagon-clang` (v79, `-mhvx`) to produce `libnntr_hvx_skel.so` for the CDSP unsigned PD.
- Add a device-only gtest that opens an unsigned-PD CDSP session, calls an HVX elementwise add (`Q6_Vsf_vadd_VsfVsf`) across FastRPC, and compares the result bit-for-bit against the CPU. The TDD flow is preserved in history: commit 2 ships the stub returning `AEE_EUNSUPPORTED` first (RED proves the FastRPC path is up, separated from kernel correctness), commit 3 implements the HVX kernel (GREEN), commit 4 adds a below-one-vector regression guard.
- The test module builds only under `ifdef HEXAGON_SDK_ROOT` and links neither `libnntrainer` nor `test_util`; environments without the Hexagon SDK build exactly as before.
- Verified on Galaxy S25 Ultra (adb `R3CY205ZMND`, CDSP, v79): `[  PASSED  ] 2 tests.`

### DIFF
--- a/test/htp/.gitignore
+++ b/test/htp/.gitignore
@@ -1,0 +1,2 @@
+generated/
+build/
--- a/test/unittest/unittest_hvx_kernels.cpp
+++ b/test/unittest/unittest_hvx_kernels.cpp
@@ -1,7 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 /**
+ * Copyright (C) 2026 dlwlzzero <dlwlzzero@gmail.com>
+ *
  * @file   unittest_hvx_kernels.cpp
- * @brief  Device test: an HVX kernel on the CDSP returns what the CPU does.
+ * @date   03 Aug 2026
+ * @brief  Device test: an HVX kernel on the CDSP returns what the CPU does
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author dlwlzzero <dlwlzzero@gmail.com>
+ * @bug    No known bugs except for NYI items
  *
  * Runs on an Android device only. Requires libnntr_hvx_skel.so on
  * ADSP_LIBRARY_PATH. See test/htp/build.sh.
@@ -43,8 +49,7 @@ protected:
     remote_rpc_control_unsigned_module unsigned_pd = {CDSP_DOMAIN_ID, 1};
     int err = remote_session_control(DSPRPC_CONTROL_UNSIGNED_MODULE,
                                      &unsigned_pd, sizeof(unsigned_pd));
-    ASSERT_EQ(err, AEE_SUCCESS)
-      << "enabling unsigned PD failed: " << hex(err);
+    ASSERT_EQ(err, AEE_SUCCESS) << "enabling unsigned PD failed: " << hex(err);
 
     const std::string uri = std::string(nntr_hvx_URI) + "&_dom=cdsp";
     err = nntr_hvx_open(uri.c_str(), &handle_);
@@ -75,6 +80,18 @@ TEST_F(HvxAdd, MatchesCpuForFullVectorsPlusRemainder) {
   int err = nntr_hvx_add_f32(handle_, a.data(), n, b.data(), n, c.data(), n);
   ASSERT_EQ(err, AEE_SUCCESS) << "add_f32 failed: " << hex(err);
   EXPECT_EQ(c, expected);
+}
+
+TEST_F(HvxAdd, MatchesCpuBelowOneVector) {
+  // Fewer floats than a single vector holds: n_vec is 0 and only the
+  // scalar tail runs.
+  const std::vector<float> a = {1.5f};
+  const std::vector<float> b = {2.25f};
+  std::vector<float> c = {0.0f};
+
+  int err = nntr_hvx_add_f32(handle_, a.data(), 1, b.data(), 1, c.data(), 1);
+  ASSERT_EQ(err, AEE_SUCCESS) << "add_f32 failed: " << hex(err);
+  EXPECT_EQ(c[0], 3.75f);
 }
 
 } // namespace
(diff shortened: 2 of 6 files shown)


test/htp/.gitignore
```diff
@@ -0,0 +1,2 @@
+generated/
+build/
```
Seunghui98, line 2: we don't need to add gitignore in this PR.

Jungwon-Lee (commented):
The last commit `50c5f00` only contains formatting and Doxygen fixes.
Could you please squash it into the commits that originally introduces these files to keep the commit history clean and each change self-contained.

Jungwon-Lee:
I think this is a standalone HVX bring-up test rather than an nntrainer test.
So, I have some concerns about its current location is appropriate.

test/unittest/unittest_hvx_kernels.cpp
```diff
@@ -0,0 +1,119 @@
...
+/** @brief Render a FastRPC error as hex so the code is searchable. */
+std::string hex(int err) {
+  std::ostringstream os;
+  os << "0x" << std::hex << std::setw(8) << std::setfill('0')
+     << static_cast<unsigned>(err);
+  return os.str();
+}
+
+/**
+ * @brief Opens one unsigned-PD CDSP session for the whole test case.
```
Jungwon-Lee: I think the current comment may be inaccurate.
`Setup` runs for each `TEST_F`, so it would be better to clarify that a new FastRPC session is opend for each test.
```
Opens an unsigned-PD CDSP session for the each test.
```

dlwlzzero:
> The last commit `50c5f00` only contains formatting and Doxygen fixes. Could you please squash it into the commits that originally introduces these files to keep the commit history clean and each change self-contained.

Got it, I'll clean it up.

test/htp/.gitignore
dlwlzzero: But if `.gitignore` file is deleted, more build related files will be added in this PR. Why don't we just remain it?

dlwlzzero:
> I think this is a standalone HVX bring-up test rather than an nntrainer test. So, I have some concerns about its current location is appropriate.

As you mentioned, the current structure is standalone, but I plan to keep the existing unittest c codes and only change the env to the nntrainer's unittest env. Although there might be some wasted steps like adding or deleting files for now, I thought it would be better to do this work after merging, so I submitted a PR.

Jungwon-Lee (approved):
LGTM!
It would be helpful to integrate HVX into nntrainer

Seunghui98 (approved):
LGTM 👍

jaemini-shin (approved):
LGTM!

PR #4227, opened 2026-08-03
[CausalLM][gpu] Long context on Adreno: keep the KV ring and the image path alive

jijoongmoon:
At `max_seq_len >= 16K` on an Adreno the model produced garbage — coherent for a sentence or two, then an endless repetition loop that never emits EOS — at ~1/8 the prefill throughput. Two independent defects compounded.

**1. The KV ring was off on ARM, so local attention kept `max_seq` rows.**

`kvRingCap()` returns 0 unless `effectivePrefillChunk() > 0`, and that returned 0 on ARM/Adreno. The stated reason was throughput (a 1073-token prompt measured 845 TPS split vs 1277 single-block), but the split threshold is `min(C, INIT_SEQ_LEN)` — that measurement was `INIT_SEQ_LEN=1024` cutting a 1073-token prompt, not a property of `C` — and the -34% itself came from the V-mirror restride fixed in (2). A performance knob was silently deciding a storage question: a `W=1024` sliding layer held 32768 rows of KV. The non-ARM value was already 4096, so nothing changes there.

**2. The OHWI K mirror was sized from `max_timestep`, one image row per (head, seq).**

Its height was `num_heads_KV * max_seq`. `CL_DEVICE_IMAGE2D_MAX_HEIGHT` is 16384 on Adreno 840, which that clears only up to an 8K window. Past it `clCreateImage` fails and — image attention being all-or-nothing per process — *every* layer drops to the non-image path, which is not merely slower, it is wrong. Nothing was logged.

Fixes: size the mirror from the ring cap (the other mirror-init site already did, so the two now agree, and a constant size means the V stride never grows — the restride behind (1) cannot recur); and fold `1<<KIMG_GSH` consecutive sequence rows into one image row. Those rows are already contiguous in the buffer, so it is purely a view change (width `*= G`, height `/= G`) — no buffer layout change, no extra traffic. `kimg_gsh_for()` derives the shift from the device's own limits and publishes it for `tca_copts()`, so mirrors and the kernels' compile-time `KIMG_GSH` always agree; it is 0 — today's addressing, instruction for instruction — whenever the unpacked view already fits.

`tca_copts()` had to stop caching: the prewarm pass registers these kernels before any layer finalizes, and a value frozen at 0 there would contradict the mirrors built afterwards.

### Measured — 1737-token prompt, SM-S948U (Adreno 840)

A 35-layer sliding-window decoder (W=1024, full attention every 5th layer, H_kv=2, d=128).

| `max_seq_len` | before | after |
|---|---|---|
| 8192 | 1204 TPS, correct | **1204.58 TPS, correct** (GSH=0) |
| 32768 | 158 TPS, repetition loop | **925.41 TPS, correct** |

KV rows at 32K: 1,146,880 → 458,752 (448 MB).

The 8192 cell is the A/B for the packing cost: the adaptive shift leaves every window that works today on the unpacked path, so its output and throughput are unchanged. At 32K the packed view costs ~23% against 8192 (a 16x65536 → 64x16384 aspect change is a texture-cache lever on this part) — the price of the path existing at all; the alternative there is the 158 TPS fallback with wrong output.

### Verification

Desktop (x86 + OpenCL + CUDA) builds clean; unit suite unchanged at 61 pass / 2 fail, the two being the OpenCL kernel tests that fail on `main` here as well. Intel-GPU and CUDA end-to-end runs were **not** possible on this host (no `cl_khr_integer_dot_product` on the iGPU; CUDA 13 dropped sm_61 and needs r580+), so those backends are covered by compilation and the unit suite only — both changed values are no-ops there by construction.

### Not fixed here

The non-image attention path is still wrong past ~1300 tokens (reproducible at any window with `NNTR_KV_IMG_ATTN=0`). This change stops Adreno from landing on it; it does not repair it.

---
**Stacked PR.** Based on #4201, so the diff shows that PR's commits too — **only the last commit belongs to this one**. Merge #4201 first.


jijoongmoon:
Superseded by https://github.com/nntrainer/nntrainer/pull/4319 for the KV-ring
half, which is what this PR is mostly about. Verified present in that diff: the
OpenCL Block-Q prefill and split-KV decode kernels take a `ring_cap` and
modulo-map their K/V row reads, `mha_core` maps every cache-row write through
`cacheRow()` (with a fail-loud throw if a step write would straddle the wrap
seam), and `KVCacheManager::setLayerCaps()` sizes a sliding layer's allocation to
`Wcap` instead of `max_seq_len`. RoPE positions and the causal/window masks stay
in absolute space; only physical storage wraps.

The Android/Adreno build half landed earlier in the same stack, in #4307.

Closing in favour of #4319. If the image-path half is not covered, please reopen.

PR #4228, opened 2026-08-03
[build][android] Build rms_reverse_norm.cpp: transformer.cpp names it unconditionally

jijoongmoon:
`Transformer::registerCustomLayers`' `device_routed` table registers `causallm::RMSReverseNormLayer` with no build guard, but `jni/Android.mk` never compiled the translation unit that defines it. `libcausallm_core.so` links with an undefined vtable and every executable over it fails:

```
ld.lld: error: undefined symbol: vtable for causallm::RMSReverseNormLayer
```

Found building for arm64 with NDK r28b and running on an Adreno 840 (SM-S948U). Added to both ndk-build modules, next to the sibling `reshaped_rms_norm.cpp`.

---
**Stacked PR.** Based on #4222, so the diff shows that PR's commits too — **only the last commit belongs to this one**. Merge #4222 first.


jaemini-shin:
It looks like this PR could be consisted with the last commit only.

jijoongmoon:
Superseded by the re-cut series against current main: #4316. The work in this PR is included there, rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please review the new series instead.

PR #4229, opened 2026-08-03
[build][android] nntr_quantize needs the tokenizer TUs it now pulls in

jijoongmoon:
#4212 makes `Transformer`'s constructor build a tokenizer (the async snapshot path), so every module that compiles `models/transformer.cpp` now depends on `huggingface_tokenizer.cpp` and `tokenizer_cache.cpp`. The `causallm_core` module already lists both; `nntr_quantize` compiles `transformer.cpp` but lists neither, so the arm64 build fails at link:

```
ld.lld: error: undefined symbol: causallm::LoadTokenizerCached(...)
>>> referenced by transformer.cpp:174
```

It links `tokenizers_c` (the Rust C ABI) but not the C++ wrapper that defines `tokenizers::Tokenizer`, so the C archive alone cannot satisfy it.

Found building for arm64 with NDK r28b.

---

Unrelated to this diff but found on the same branch: the checked-in `Applications/CausalLM/lib/libtokenizers_c.a` was rebuilt on Ubuntu 24.04 (GCC 13.3) and references `std::ios_base_library_init()` (GLIBCXX_3.4.32), so it **cannot link on Ubuntu 22.04** (libstdc++ 6.0.30). Rebuilding the archive on 22.04 fixes it.

---
**Stacked PR.** Based on #4212, so the diff shows that PR's commits too — **only the last commit belongs to this one**. Merge #4212 first.


jijoongmoon:
Superseded by the re-cut series against current main: #4307. The work in this PR is included there,
rebased onto main with the CI failures resolved. Closing to keep the queue reviewable — please
review the new series instead. The branch is preserved on the fork.

Two things are better in the new cut than a straight re-land of this PR would have been:

- **It is not stacked.** #4307 is cut from `main`, so the blocker this PR carried — sitting on top of
  another in-flight branch — is gone.
- **The archive is fixed as well as the source list.** Adding `huggingface_tokenizer.cpp` to the
  `nntr_quantize` ndk module is only half of it. `build_tokenizer_android.sh` cloned
  `mlc-ai/tokenizers-cpp` at build time, unpinned, and that checkout also built the C++ wrapper — so
  the prebuilt archive carried a `huggingface_tokenizer.cc.o` member the tracked x86 archive does
  not, and `nntr_quantize` was resolving `tokenizers::Tokenizer::FromBlobJSON` out of it. One app,
  two builds of the same class depending on the module. #4307 builds the in-tree
  `tokenizers_c_win` crate with the NDK cross compiler instead, so the android archive is crate-only
  and matches the x86 one member for member: `llvm-ar t` gives 0 `hugg` members, `llvm-nm` gives 0
  `FromBlobJSON`, and all ten entry points `tokenizers_c.h` declares are defined. That makes the
  missing source-list entry a hard link error rather than a silent substitution, which is why the two
  commits ship together and in that order.

Verified with NDK r27 (27.2.12479018), `arm64-v8a`: `ndk-build` of `causallm_core`, `causallm_api`,
`nntrainer_causallm`, `test_api`, `nntr_quantize` and `nntr_safetensors_info` returns 0, and
`llvm-nm` finds `FromBlobJSON` defined once in `nntr_quantize`.

PR #4231, opened 2026-08-03
[kleidiai] Add matmul_clamp_f32_qai8dxp_qsi8cxp

jaemini-shin:

<details><summary>[kleidiai] Add matmul_clamp_f32_qai8dxp_qsi8cxp</summary><br />

- Add matmul_clamp_f32_qai8dxp_qsi8cxp ukernels
- Add rhs pack kernel
- Add interface

</details>

### Summary

Let's add kleidiai kernel for
- activation: fp32 -> channelwise int8
- weight: channelwise int8

**Self evaluation:**

### DIFF
--- a/nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi8cxp.cpp
+++ b/nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi8cxp.cpp
@@ -1,0 +1,343 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2024 Arm Limited and/or its affiliates
+ *
+ * @file   kleidiai_interface_qai8dxp_qsi8cxp.cpp
+ * @date   23 July 2026
+ * @see    https://github.com/ARM-software/kleidiai
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author Jaemin Shin <jaemin980311@gmail.com>
+ *
+ * @brief  Modified computational backend components of
+ * kleidiai. Portions of this file are derived from Arm
+ * Limited code licensed under the Apache License, Version 2.0, with
+ * modifications
+ *
+ * @note   Licensed under the Apache License, Version 2.0 (the "License");
+ *         you may not use this file except in compliance with the License.
+ *         You may obtain a copy of the License at
+ *             http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * @bug    No known bugs except for NYI items
+ */
+//
+// SPDX-FileCopyrightText: Copyright 2025 Arm Limited and/or its affiliates
+// <open-source-office@arm.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <algorithm>
+
+#include <kleidiai_interface.h>
+#include <thread_manager.h>
+
+#include "kai/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp_qsi8cxp_interface.h"
+
+// Include micro-kernel variants
+#include "kai/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp1x4_qsi8cxp4x4_1x4_neon_dotprod.h"
+#include "kai/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp1x8_qsi8cxp4x8_1x4_neon_dotprod.h"
+#include "kai/matmul_clamp_f32_qai8dxp_qsi8cxp/kai_matmul_clamp_f32_qai8dxp4x4_qsi8cxp4x4_16x4_neon_dotprod.h"
+
+#if defined(__ARM_FEATURE_MATMUL_INT8)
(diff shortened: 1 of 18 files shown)


Seunghui98 (commented):
## Description

Adds the `matmul_clamp_f32_qai8dxp_qsi8cxp` path (FP32 activation → per-channel int8, per-channel int8 weight): 4 vendored ARM ukernels, the `kai_rhs_pack_nxk_qsi8cxp_qsi8cx_neon` pack kernel, an nntrainer interface layer, and meson wiring.

## Summary

- The meson wiring is correct this time — `pack_sources` and `pack_headers` both get their new entry added to the existing lists, and the `enable-fp16` block below now uses `pack_headers +=`, so the header-list clobber from #4133 is gone.
- `__kai_gemm_qai8dxp_qsi8cxp_rhs_unpacked` does not repeat the spurious LHS allocation that the qsi4cxp version carried — it allocates only the RHS buffer it actually uses.
- The RHS pack path is nxk-only by construction (one pack kernel, no `is_nxk` flag), which is self-consistent rather than an omission.

## Bug

No correctness defect found. The item below is a performance regression, and I'd still treat it as the thing to fix before merge.

1. **This reintroduces the scheduler that #4065 already removed from the sibling file on this very base.** I checked the PR's base commit (`2b9140b`): `kleidiai_interface_qai8dxp_qsi4cxp.cpp` there contains the "Cache-aware 2D tiling" block and **zero** occurrences of `m_step > n_step`. The new `kleidiai_interface_qai8dxp_qsi8cxp.cpp` contains `m_step > n_step` and none of the tiling — so the two files sitting next to each other in the same directory now disagree about how to schedule the same shape of work, with the newer one using the older design.

   The consequence is concrete, not stylistic. Both GEMM variants added here are `m_step = 16`, `n_step = 4`:

   | variant | m_step | n_step |
   |---|---|---|
   | `qai8dxp1x4_qsi8cxp4x4_1x4_neon_dotprod` | 1 | 4 |
   | `qai8dxp1x8_qsi8cxp4x8_1x4_neon_dotprod` | 1 | 4 |
   | `qai8dxp4x4_qsi8cxp4x4_16x4_neon_dotprod` | **16** | 4 |
   | `qai8dxp4x8_qsi8cxp4x8_16x4_neon_i8mm` | **16** | 4 |

   For the bottom two, `m_step > n_step` is always true, so every call takes the parallelize-over-m branch. At decode (`m = 1`) that gives `m_loop == 1`, and `parallel_for(0, 1, ...)` runs the body inline — the whole GEMM is **single-threaded**. Restoring the single-threaded decode path is one of the things #4065 was written to fix.

   The cleanest resolution is to share one scheduler between the two interfaces rather than fork it a third time (#4133 made the same copy for the f16 variant). Failing that, porting the tiling across before merge would at least stop the regression from landing.

2. **17 files and +3458 lines with no test.** #4133, the closest precedent, at least shipped `gemv`/`gemm` cosine-similarity cases. Here there is nothing exercising the new interface, so a wrong `nr`/`kr`/`sr` in the variant table or a bad offset would compile and link cleanly and only surface in a model. A single cosine-similarity test mirroring the qsi4cxp one, including a shape whose `m`/`n` are not multiples of the step sizes, would cover the parts most likely to be wrong.

3. **Minor:** `nr` is fetched in `__kai_gemm_qai8dxp_qsi8cxp` but never used — only `mr`/`kr`/`sr` feed the LHS pack. `-Wunused-variable` will flag it in a `-Werror` build.

---

Seunghui98 (commented):
you need to check CI fails.

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi8cxp.cpp
```diff
@@ -0,0 +1,304 @@
...
+ * modifications
+ *
+ * @note   Licensed under the Apache License, Version 2.0 (the "License");
+ *         you may not use this file except in compliance with the License.
+ *         You may obtain a copy of the License at
+ *             http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * @modifications
+ *   - [2026-07-23] Integrated and adapted Arm-provided code into
+ *     nntrainer CPU backend
```
Seunghui98: What is this tag? 🤔

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi8cxp.cpp
```diff
@@ -0,0 +1,304 @@
...
+  delete[] rhs_packed_mtx_qs8cx;
+}
+
+void __kai_gemm_qai8dxp_qsi8cxp(size_t m, size_t n, size_t k,
+                                void *lhs_native_mtx_f32,
+                                void *rhs_packed_mtx_qs8cx,
+                                float *dst_act_mtx_f32, size_t idx_variant,
+                                float lower_bound, float upper_bound) {
+  const size_t mr = ukernel_variants[idx_variant].ukernel.get_mr();
+  const size_t nr = ukernel_variants[idx_variant].ukernel.get_nr();
```
Seunghui98:
```suggestion
```
I guess this is unused.

nntrainer/tensor/cpu_backend/arm/kai_interface/kleidiai_interface_qai8dxp_qsi8cxp.cpp
jaemini-shin: This is generated by claude, I'll fix it.

Seunghui98 (approved):
LGTM!

jayden0701 (approved):
LGTM!

PR #3949, opened 2026-05-27
[tensor_api] Decouple multi-input layer restriction

jaemini-shin:
This PR decouples restriction between layer node order and input index.
Since previous update, legacy nntrainer binaries were in conflict in swiglu layer.
* swiglu input[0] should be gate
* swiglu input[1] should be up
* weights are in up, gate, swiglu order(fixed in binary)
* layer nodes are in gate, up, swiglu order(by swiglu({gate, up}))

In this condition, we always get swiglu input[0] = up and swiglu input[1] = gate
regardless of changing the order like swiglu({up, gate}).
With this commit, we can set each orders in separate which keeps backward compatiblity.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jaemin Shin <jaemin980311@gmail.com>

PR #4233, opened 2026-08-03
Fix KV cache inputs being bound to the wrong compiled input slots

jayden0701:
## Summary

CausalLM bound KV cache pointers to compiled input slots by lexical name
sort, but since the caches became real `"input"` layers the compiled order
is graph-construction order (interleaved K/V per layer), not lexical. From
two layers on, every cache after the first was silently bound to the wrong
slot — and where per-layer widths differ (e.g. Gemma4 sliding vs global
head_dim) the mismatch is an out-of-bounds heap access, not just a
permutation.

This PR fixes the binding to look up the compiled input order by name
(`Model::getInputNames()`) instead of predicting it, and rejects undersized
buffers at bind time so a future mismatch fails with a diagnostic instead of
corrupting memory. It also makes Gemma4 FP16 KV caches explicit `"input"`
layers on all platforms, and hardens the raw-pointer inference paths to
reject caller vectors whose count does not match the compiled model.

## Problem

CausalLM sorted its KV cache pointers by name and handed the result to
`incremental_inference()` positionally. That matched the compiled graph only
while the caches were plain leaf `Tensor`s, which the ccapi graph builder
registers through a `std::map` and therefore in lexical order. Since the caches
became real `"input"` layers they are registered in graph construction order
instead, which interleaves them:

```
compiled:  input0, cache_k_l0, cache_v_l0, cache_k_l1, cache_v_l1
sorted:    input0, cache_k_l0, cache_k_l1, cache_v_l0, cache_v_l1
```

From two layers on, every cache after the first was bound to the wrong slot.

## Why this is not just a permutation

`TensorPool::fillPlaceholder()` rebinds by raw pointer and reinterprets whatever
it is given with the slot's dimension, so the mismatch is silent. Where the
per-layer widths differ it is also **out of bounds**. Measured on device with the
tiny Gemma4 fixture (sliding width 32, global width 64):

```
slot 1 'cache_k_l0' needs 256 elems <- bound buffer 'cache_k_l0' has 256 elems
slot 2 'cache_v_l0' needs 256 elems <- bound buffer 'cache_k_l1' has 512 elems   (wrong buffer)
slot 3 'cache_k_l1' needs 512 elems <- bound buffer 'cache_v_l0' has 256 elems   *** OVERFLOW ***
slot 4 'cache_v_l1' needs 512 elems <- bound buffer 'cache_v_l1' has 512 elems
```

Real Gemma4 has 35 layers and a 256/512 `head_dim`/`global_head_dim` split, where
lexical order additionally sorts `l10` before `l2`.

`save_kvcache()`/`load_kvcache()` address `KVCacheManager` by logical layer index,
independently of the graph binding, so a permuted binding also sends saved caches
to the wrong layers.

## Why the existing tests did not catch it

Between equally sized, distinct buffers a stable permutation is self-consistent
inside the graph — layer 0's V simply lives in the buffer named "K of layer 1",
and a single prefill writes and reads it consistently. The reference tests use
1- or 2-layer models with uniform widths, so the logits never moved.

## Fix

Look the order up instead of predicting it. Callers hand over host buffers keyed
by compiled input layer name, including the primary input, and
`buildOrderedInferenceInputs()` places each at the index the compiled graph
reports. Every compiled input must be matched by name; there is no positional or
name-pattern fallback to get wrong.

Each buffer also carries its size and is rejected unless it covers the slot it
binds to, so a future mismatch fails at bind time with a diagnostic instead of
corrupting the heap.

## Commits

1. `[ccapi] Expose compiled model input names in inference order` — adds
   `Model::getInputNames()`; non-pure with an empty default so out-of-tree
   `Model` implementations keep compiling.
2. `[CausalLM] Bind KV cache inputs by compiled input name` — the fix above,
   plus regressions. Also fixes LFM2 passing `NUM_LAYERS*2` pointers into a graph
   that only declares inputs for its attention layers.
3. `[CausalLM] Make Gemma4 FP16 KV caches explicit input layers` — removes the
   last plain-`Tensor` cache path so one binding order is correct on all
   platforms.
4. `[ccapi] Reject raw-pointer inference calls with mismatched counts` — the
   inference loops were sized by model dimensions but indexed the caller's
   vectors. **Must land after (2)**, which is what stopped CausalLM
   over-supplying pointers.

## Notes for reviewers

- (4) is a behaviour change: callers that relied on surplus pointers being
  ignored now get an error naming both counts.
- The new meson FP16 test target only materialises for a native aarch64 build —
  the android cross build skips `Applications` entirely and covers the test
  through `Applications/CausalLM/jni/Android.mk`.
- The tiny Gemma4 fixture now pins `global_head_dim` so the sliding and global
  caches keep different widths. With equal widths the regression cannot fail.

## Test plan

- x86_64 FP32: `unittest_causallm_models` 69 passed, `unittest_ccapi` graph suite
  12 passed.
- x86_64 ASAN: 69 passed, no sanitizer reports.
- Galaxy S26U via
  `Applications/CausalLM/jni` ndk-build: 86 passed, 0 failed (10 Q4_0
  differential tests skipped, `NNTR_QUANTIZE_BIN` unset). Verified the caches
  really compile as FP16 and that the compiled order is interleaved on ARM.
- Each of the four commits builds and passes on x86_64 (bisect-safe).
- Mutation check: forcing lexical order back into
  `buildOrderedInferenceInputs()` is now rejected —
  `"Host buffer for model input 'input0' is too small: 32 bytes offered, 1024
  bytes required by the compiled slot"`.
- Verified end-to-end: full-size Gemma4/LFM2 generation


jaemini-shin:
If this issue is occurred by the reorder of input tensors, https://github.com/nntrainer/nntrainer/pull/3949 could be a solution.

jayden0701:
> If this issue is occurred by the reorder of input tensors, #3949 could be a solution.

This issue is at a different level. #3949 changes input orders for a specific layer(tensor),
but this problem occurs because the order of KV caches given to incremental_forwarding()
doesn't match the sorted order of all inputs for attention layers. 

And I think matching cahces to inputs by actual names is needed for robustness in many situations.

api/ccapi/include/model.h
```diff
@@ -503,6 +503,13 @@ class Model {
    */
   virtual void exports(const ExportMethods &method,
                        const std::string file_path) = 0;
+
+  /**
+   * @brief Get input layer names in positional inference order.
+   *
+   * @note Override this for models that support named positional inputs.
+   */
+  virtual std::vector<std::string> getInputNames() { return {}; }
```
jaemini-shin:
```suggestion
  virtual std::vector<std::string> getInputNames() = 0;
```
How about using pure virtual method if possible?

Applications/CausalLM/models/lfm2/lfm2_causallm.cpp
```diff
@@ -557,7 +531,10 @@ void Lfm2CausalLM::run_with_embeddings(const void *inputs_embeds,
     input_sample = const_cast<float *>(embeds);
   }
 
-  auto input = build_inference_inputs(input_sample);
+  const size_t input_sample_bytes =
+    sizeof(float) * static_cast<size_t>(BATCH_SIZE) *
+    (needs_padding ? static_cast<unsigned int>(INIT_SEQ_LEN) : input_len) * DIM;
```
jaemini-shin:
``` cpp
  const size_t input_sample_bytes =
  sizeof(float) * static_cast<size_t>(BATCH_SIZE) *
  static_cast<size_t>(INIT_SEQ_LEN) *
  static_cast<size_t>(DIM);
```
At this point, input has been already padded, no need to branch.

Applications/CausalLM/models/lfm2/lfm2_causallm.cpp
```diff
@@ -505,33 +506,6 @@ void Lfm2CausalLM::run_with_embeddings(const void *inputs_embeds,
...
-      cache_inputs.emplace_back(
-        "cache_k_l" + std::to_string(i),
-        reinterpret_cast<float *>(kv_cache.getKeyCache(i).getData()));
-      cache_inputs.emplace_back(
-        "cache_v_l" + std::to_string(i),
-        reinterpret_cast<float *>(kv_cache.getValueCache(i).getData()));
-    }
-    std::sort(
-      cache_inputs.begin(), cache_inputs.end(),
-      [](const auto &lhs, const auto &rhs) { return lhs.first < rhs.first; });
```
jaemini-shin, line 525: Just for curiosity, is it possible to handle this issue with removing this sort?

Seunghui98 (commented):
## Description

Fixes KV cache buffers being bound to the wrong compiled input slots. CausalLM predicted the compiled input order by lexical name sort, which stopped matching once the caches became real `"input"` layers registered in graph-construction order. Replaces the prediction with a lookup via a new `Model::getInputNames()`, adds a size guard at bind time, makes Gemma4 FP16 caches explicit input layers, and rejects count-mismatched raw-pointer inference calls.

## Summary

The diagnosis and the fix both hold up, and I checked the two places this kind of change usually goes wrong:

- **`reinitialize()` stays in sync.** `input_dims_` is cleared at the top of `reinitialize()` and rebuilt, while `input_names_` persists from `initialize()`. The new per-element name/tensor assertions plus the closing `model_input_idx != input_names_.size()` check mean a desync fails loudly rather than leaving `getInputNames()` throwing "not synchronized" later. Given the new invariant is "these three vectors must agree", pinning it at the point of mutation is the right place.
- **The declared buffer size is honest at the main call site.** `causal_lm.cpp` passes `sizeof(float) * BATCH_SIZE * MAX_SEQ_LEN`, which is exactly the `malloc` two lines above it — the guard isn't being fed a number that just happens to be large enough.

Two things worth calling out as good practice: replacing prediction with lookup and leaving *no* positional or name-pattern fallback is what makes this fix durable rather than another guess that happens to be right today. And `RuntimeBindingsMatchCompiledInputs` asserting that the sliding and global cache widths actually differ — with a message saying the test is meaningless otherwise — is a test that defends its own premise. That's the thing that was missing when the original bug slipped through 1- and 2-layer uniform-width fixtures.

## Bug

None found in the change.

1. **The size guard validates against the compiled batch, not the batch the call actually runs with.** `buildOrderedInferenceInputs()` computes `slot_bytes` from `model->getInputDimension()`, which carries the batch captured at initialize. But `incremental_inference()` and `inference()` then do `in_dim[idx].batch(batch_size)` with the caller's value before constructing the tensors. A call with `batch_size` larger than the compiled batch passes the guard and then reads past the buffer — precisely the failure mode this PR exists to close, just through a different door. Every in-tree caller passes `BATCH_SIZE` consistently so nothing is broken today, but since the guard's stated purpose is "a future mismatch fails with a diagnostic instead of corrupting memory", it would be worth either scaling `slot_bytes` by the batch the caller will pass, or asserting in `inference`/`incremental_inference` that `batch_size` matches the compiled one.

2. **`Model::getInputNames()`'s empty default produces a misleading error rather than a clear one.** The rationale for non-pure-with-`{}` (out-of-tree `Model` implementations keep compiling) makes sense. But an implementation that doesn't override it reaches `buildOrderedInferenceInputs()` with an empty name list and a non-empty dimension list, and fails on `"Compiled model input names and dimensions are not synchronized"` — which points the reader at a graph-internal invariant rather than at "this model type doesn't support named inputs". A dedicated check for the empty-names case would save someone a debugging session.

3. **`SentenceTransformer::encode()` drops unmatched entries silently.** `CausalLM::buildInferenceInputs()` passes `&dropped` and `ml_logd`s each one; the SentenceTransformer call site omits the argument entirely, so a runtime buffer that matches no compiled input vanishes without a trace. Compiled inputs are still all required by name, so this can't cause a wrong binding — it just means a stale cache name there is invisible. Worth mirroring for consistency.

4. **Commit (4) is a public ccapi behaviour change and deserves more than a PR-body note.** Callers that relied on surplus pointers being ignored now get a throw. The ordering after (2) is correct and the error message naming both counts is the right call, but out-of-tree users of `inference()`/`incremental_inference()` will hit this at runtime with no compile-time signal. A line in the release notes seems warranted.

---

Applications/CausalLM/models/lfm2/lfm2_causallm.cpp
jayden0701: It would work...(Just like Gemma4)
But I think matching them by names would be more robust.

jaemini-shin: IMO, if this sorting is the reason, removing it makes no need to save every input name and compare it.
How does the overhead for this new logic? If this overhead is huge, it'd be better to remove this sort only.

jayden0701: Time complexity would be 
before : O(NlogN) (by std::sort)  
after : O(NlogN) (by std::map) + copy names and lookup O(N) + some O(N) operations

N is just 2*NUM_layers, and it happens only once before inference. Which won't make much difference.

When tested on device, no big differences were found (E2E)

+) I measured the differences in only binding time, and it was 3.41us to 5.92us . Which is not that big. (x86)

jaemini-shin: Then let's proceed as it is! But it'd be better to keep in mind that we might skip this check if we need extreme optimization for latency and memory.

jaemini-shin (approved):
LGTM!

Applications/CausalLM/models/lfm2/lfm2_causallm.cpp
jayden0701: Yup! For extreme cases I agree that this check can be skipped.

Seunghui98 (approved):
LGTM!

PR #4236, opened 2026-08-04
[htp] complete A8W4 DSP quantization pipeline with HMX matmul and activation quant

dlwlzzero:
## Dependency of the PR

None

<details><summary>66d2b1aa [htp] Link HexKL micro and bake u8i4 weights once</summary><br />

[htp] Link HexKL micro and bake u8i4 weights once

Add mm_u8i4_from_u8 and mm_u8i4_from_f32 to the FastRPC interface and a
DSP-side entry point that brings HMX up through hexkl_micro_hw_init and
round-trips the HMX lock. from_f32 returns AEE_EUNSUPPORTED until the
quantization kernels land.

build.sh now links libhexkl_micro.a. This requires Hexagon SDK 6.1.1.0 or
newer because HexKL ships a v79 static library only from lib/6.1.1.0 up;
6.4.0.1 is the combination HexKL's own examples were built against.

The harness logs the VTCM size so the budget in the design document can be
checked against hardware for the first time. Measured ~8 MB on S25 Ultra
(V79), ~13x the design's 598 KB peak footprint.

Add the VTCM partitioning, a one-shot weight bake and the HMX matmul loop
for the A8W4 path. HexKL's example converts each weight tile to WH layout
from inside the innermost loop; baking every tile up front lets the matmul
read them in place with no copy per tile.

The test quantizes weights on the host with the QS4CX algorithm, feeds a
fixed uint8 activation and requires the int32 accumulator to match a plain
C integer matmul bit for bit. Integer arithmetic is deterministic, so any
layout or wiring error shows up here rather than being absorbed into a
tolerance later.

The test also dumps the baked WH tile, which is the only record we have of
that layout -- HexKL does not document it.

**Self evaluation:**

Co-Authored-By: Sisyphus <noreply@opencode.ai>

</details>

<details><summary>23fe947f [htp] Quantize and dequantize activations on the DSP</summary><br />

[htp] Quantize and dequantize activations on the DSP

Add per-row asymmetric uint8 dynamic quantization and write the result
straight into the HMX activation layout. Eight-bit activation tiles are
flat row-major inside a tile, so no separate layout pass is needed and the
kernel can target VTCM directly instead of going through HexKL's
copy_submatrix helper, which its own header marks as debug-only.

Also add hvx_convert.h. HVX has no numeric conversion between int32 and
f32 -- Q6_Vw_equals_Vsf and friends only reinterpret bits, and the vcvt
family does not cover sf<->w -- so both directions go through a
magic-number identity that is exact for |x| <= 2^22. Our accumulators are
bounded by 255 * 8 * K, leaving about 2x margin at K=1024.

The quantization implementation is scalar initially; vectorization comes
next and is gated on this same test staying green.

Rounding is nearest-even on both sides. The host reference uses nearbyint
rather than round because HVX's float add rounds that way and the two
disagree at exact .5. Weight quantization keeps round, since it runs only
on the host.

Scalar K3 (int32 acc -> f32 dequant) completes the A8W4 accuracy
pipeline: S1/S2/S3 all pass. Vectorize K1 (row min/max reduction),
K2 (per-row uint8 quantization), and K3 (i32 -> f32 dequant) on HVX.
K3 uses Q6_Vsf_equals_Vw for the int32 -> f32 lane conversion, matching
the production pattern in llama.cpp's ggml-hexagon backend; the
magic-number trick that K2 needed (bit reinterpret, not numeric
conversion) does not apply in the K3 direction.

**Self evaluation:**

Co-Authored-By: Sisyphus <noreply@opencode.ai>

</details>

<details><summary>08db922d [test] Trim u8i4 harness and drop dead FastRPC endpoint</summary><br />

[test] Trim u8i4 harness and drop dead FastRPC endpoint

The per-task S1-S3 isolation tests and the HmxBringsUp smoke test
already proved out the individual pipeline stages; CheckShape's S1-S4
checks now cover the same ground end to end for every real workload
shape. Drop the now-redundant tests and the want_wh_dump debug dump
path so the suite only carries the four shapes that matter.

Drop the dead mm_u8i4_from_u8 FastRPC endpoint. Its only callers were
the unittest scaffolding removed above; mm_u8i4_from_f32 already exercises
the same weight-bake/HMX matmul path end to end, so the from_u8 entry point
and its WH-tile debug dump argument are unused now.

**Self evaluation:**

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

</details>

### Summary

- **Link HexKL micro and bake u8i4 weights once**: Add FastRPC interface with HMX initialization, implement one-shot weight baking to WH layout, validate VTCM budget on hardware (S25 Ultra: ~8MB measured)
- **Quantize and dequantize activations on the DSP**: Implement per-row asymmetric uint8 dynamic quantization with scalar and HVX kernels, complete A8W4 accuracy pipeline (S1/S2/S3 pass)
- **Trim u8i4 harness and drop dead FastRPC endpoint**: Consolidate test suite to shape-level coverage (S1-S4), remove redundant isolation tests and unused endpoints

---

### DIFF
--- a/nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.c
+++ b/nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.c
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2026 dlwlzzero <dlwlzzero@gmail.com>
+ *
+ * @file   hexkl_mm_u8i4.c
+ * @date   03 Aug 2026
+ * @brief  VTCM layout and HMX orchestration for the A8W4 matmul
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author dlwlzzero <dlwlzzero@gmail.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#include <AEEStdErr.h>
+
+#include "hexkl_micro.h"
+
+#include "hexkl_mm_u8i4.h"
+
+#define ROUND_UP_U32(v, a) ((((v) + ((a)-1)) / (a)) * (a))
+
+/** @brief Bytes in one packed i4 weight tile (32x32 values). */
+#define WEIGHT_TILE_BYTES 512u
+/** @brief Bytes in one int32 accumulator tile (64x32 values). */
+#define ACC_TILE_BYTES 8192u
+
+int hexkl_mm_u8i4_plan(uint8_t *vtcm_base, uint32_t vtcm_size, uint32_t m_pad,
+                       uint32_t k, uint32_t n, hexkl_mm_u8i4_layout *out) {
+  if (!vtcm_base || !out || k == 0 || n == 0 || m_pad == 0) {
+    return AEE_EBADPARM;
+  }
+  if ((k % HEXKL_HMX_INT8_BLOCK_N_INNER) != 0 ||
+      (n % HEXKL_HMX_INT8_BLOCK_N_COL) != 0 ||
+      (m_pad % HEXKL_HMX_INT8_BLOCK_N_ROW) != 0) {
+    return AEE_EBADPARM;
+  }
+
+  const uint32_t n_ktiles = k / HEXKL_HMX_INT8_BLOCK_N_INNER;
+  const uint32_t n_ntiles = n / HEXKL_HMX_INT8_BLOCK_N_COL;
+  const uint32_t n_rblocks = m_pad / HEXKL_HMX_INT8_BLOCK_N_ROW;
+
+  out->vtcm_base = vtcm_base;
+  out->vtcm_size = vtcm_size;
+
+  out->w_base = 0u;
+  out->w_size = n_ktiles * n_ntiles * WEIGHT_TILE_BYTES;
+
+  // Activation tiles must sit on HEXKL_HMX_ACTIVATION_ALIGNMENT, and each
+  // tile is exactly that many bytes, so aligning the base is enough.
+  out->act_base = ROUND_UP_U32(out->w_size, HEXKL_HMX_ACTIVATION_ALIGNMENT);
+  out->act_size = n_rblocks * n_ktiles * HEXKL_HMX_ACTIVATION_ALIGNMENT;
+
+  out->result_off = out->act_base + out->act_size;
+
+  // Put the config region at the top of the arena, rounded down to its
+  // alignment so the rounding can never eat into the result region.
+  if (vtcm_size < hexkl_micro_hmx_config_size()) {
+    return AEE_ENOMEMORY;
+  }
+  out->config_off = (vtcm_size - hexkl_micro_hmx_config_size()) &
+                    ~(HEXKL_HMX_CONFIG_ALIGNMENT - 1u);
+
+  if (out->result_off + ACC_TILE_BYTES > out->config_off) {
+    return AEE_ENOMEMORY;
+  }
+  return AEE_SUCCESS;
+}
+
+int hexkl_mm_u8i4_bake_weights(const hexkl_mm_u8i4_layout *L,
+                               const int8_t *w_i4_rm, uint32_t k, uint32_t n) {
+  if (!L || !w_i4_rm) {
+    return AEE_EBADPARM;
+  }
+  const uint32_t n_ktiles = k / HEXKL_HMX_INT8_BLOCK_N_INNER;
+  const uint32_t n_ntiles = n / HEXKL_HMX_INT8_BLOCK_N_COL;
+
+  for (uint32_t kt = 0; kt < n_ktiles; ++kt) {
+    for (uint32_t nt = 0; nt < n_ntiles; ++nt) {
+      const uint32_t off = L->w_base + (kt * n_ntiles + nt) * WEIGHT_TILE_BYTES;
+      int res =
+        hexkl_micro_hmx_rm_to_wh_i4(L->vtcm_base, off, w_i4_rm, kt, nt, n);
+      if (res != AEE_SUCCESS) {
+        return res;
+      }
+    }
+  }
+  return AEE_SUCCESS;
+}
+
+int hexkl_mm_u8i4_run(const hexkl_mm_u8i4_layout *L, uint32_t m_pad, uint32_t k,
+                      uint32_t n, int32_t *acc_i32) {
+  if (!L || !acc_i32) {
+    return AEE_EBADPARM;
+  }
+  const uint32_t n_ktiles = k / HEXKL_HMX_INT8_BLOCK_N_INNER;
+  const uint32_t n_ntiles = n / HEXKL_HMX_INT8_BLOCK_N_COL;
+  const uint32_t n_rblocks = m_pad / HEXKL_HMX_INT8_BLOCK_N_ROW;
+
+  for (uint32_t rb = 0; rb < n_rblocks; ++rb) {
+    for (uint32_t nt = 0; nt < n_ntiles; ++nt) {
+      hexkl_micro_hmx_acc_clear_int32();
+
+      for (uint32_t kt = 0; kt < n_ktiles; ++kt) {
+        const uint32_t act_off =
+          L->act_base + (rb * n_ktiles + kt) * HEXKL_HMX_ACTIVATION_ALIGNMENT;
+        const uint32_t w_off =
+          L->w_base + (kt * n_ntiles + nt) * WEIGHT_TILE_BYTES;
+        int res = hexkl_micro_hmx_mm_u8i4(L->vtcm_base, act_off, w_off);
+        if (res != AEE_SUCCESS) {
+          return res;
+        }
+      }
+
+      int res = hexkl_micro_hmx_acc_read_int32(L->vtcm_base, L->config_off,
+                                               L->result_off);
+      if (res != AEE_SUCCESS) {
+        return res;
+      }
+      // Going through copy_32b_to_submatrix keeps us off the accumulator's
+      // shuffled layout, which HexKL does not document.
+      res = hexkl_micro_hmx_copy_32b_to_submatrix(L->vtcm_base, L->result_off,
+                                                  acc_i32, rb, nt, m_pad, n);
+      if (res != AEE_SUCCESS) {
+        return res;
+      }
+    }
+  }
+  return AEE_SUCCESS;
+}
--- a/nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.h
+++ b/nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.h
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2026 dlwlzzero <dlwlzzero@gmail.com>
+ *
+ * @file   hexkl_mm_u8i4.h
+ * @date   03 Aug 2026
+ * @brief  VTCM layout and HMX orchestration for the A8W4 matmul
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author dlwlzzero <dlwlzzero@gmail.com>
+ * @bug    No known bugs except for NYI items
+ */
+
+#ifndef __NNTRAINER_HEXKL_MM_U8I4_H__
+#define __NNTRAINER_HEXKL_MM_U8I4_H__
+
+#include <stdint.h>
+
+/**
+ * @brief Where each region lives inside the VTCM arena.
+ *
+ * Byte offsets from @a vtcm_base, not pointers, because that is what the
+ * HexKL micro API takes.
+ */
+typedef struct {
+  uint8_t *vtcm_base;  /**< base of the VTCM arena from hexkl_micro_hw_init */
+  uint32_t vtcm_size;  /**< total arena size in bytes */
+  uint32_t w_base;     /**< all baked weight tiles, (K/32)*(N/32)*512 bytes */
+  uint32_t w_size;     /**< byte span of the weight region */
+  uint32_t act_base;   /**< activation AH tiles, m_pad*K bytes, 2048-aligned */
+  uint32_t act_size;   /**< byte span of the activation region */
+  uint32_t result_off; /**< 8192-byte accumulator readout, 2048-aligned */
+  uint32_t config_off; /**< HMX config region, 256-aligned */
+} hexkl_mm_u8i4_layout;
+
+/**
+ * @brief Computes the VTCM partitioning and checks it against the arena.
+ *
+ * Pure arithmetic apart from reading hexkl_micro_hmx_config_size(), so it
+ * is the one piece of this file that could be unit tested on the host.
+ *
+ * @return AEE_SUCCESS, AEE_EBADPARM on a shape violation, or
+ *         AEE_ENOMEMORY when the partitioning does not fit.
+ */
+int hexkl_mm_u8i4_plan(uint8_t *vtcm_base, uint32_t vtcm_size, uint32_t m_pad,
+                       uint32_t k, uint32_t n, hexkl_mm_u8i4_layout *out);
+
+/**
+ * @brief Converts every weight tile to WH layout once, into VTCM.
+ *
+ * HexKL's example calls rm_to_wh_i4 from the innermost matmul loop, which
+ * serializes the layout conversion against the multiply. Hoisting it here
+ * means the matmul reads the tiles in place with no copy at all.
+ *
+ * @param[in] w_i4_rm weights as int4 values in int8 containers, range
+ *                    [-8, 7], K rows by N columns, row-major
+ */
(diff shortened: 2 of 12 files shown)


test/unittest/unittest_hvx_mm_u8i4.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+  /**
+   * @brief Runs S1 through S4 for one shape.
+   *
+   * S1 and S2 are bit-exact: integer arithmetic is deterministic, so any
+   * layout or wiring error shows up as an exact mismatch rather than
+   * being absorbed into a tolerance. S3 allows f32 ordering slack. S4 is
+   * reported, not gated.
+   */
+  void CheckShape(uint32_t M, uint32_t K, uint32_t N, bool want_wh_dump) {
```
Jungwon-Lee: `want_wh_dump` is never used.
Could we remove this parameter?

test/unittest/unittest_hvx_mm_u8i4.cpp
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * @param[in]  w_f32 weight matrix, K rows by N columns, row-major
+ * @param[out] q_w   quantized values in [-8, 7], K by N, row-major
+ * @param[out] d     per-channel dequantization multiplier, N entries
+ * @param[out] colsum per-channel sum of q_w over K, N entries
+ */
+void quantize_weights_qs4cx(const std::vector<float> &w_f32, uint32_t K,
+                            uint32_t N, std::vector<int8_t> &q_w,
+                            std::vector<float> &d,
+                            std::vector<int32_t> &colsum) {
```
Jungwon-Lee, line 102: Could we use nntrainer's existing `quant_qs4cs_f32` here?

test/unittest/unittest_hvx_mm_u8i4.cpp
```diff
@@ -0,0 +1,476 @@
...
+  std::ostringstream os;
+  os << "0x" << std::hex << std::setw(8) << std::setfill('0')
+     << static_cast<unsigned>(err);
+  return os.str();
+}
+
+/** @brief Rounds @a v up to a multiple of @a a. */
+constexpr uint32_t round_up(uint32_t v, uint32_t a) {
+  return ((v + a - 1) / a) * a;
+}
```
Jungwon-Lee, line 46: Could we use a single checked round-up helper here?
The PR currently has three implementations.

test/unittest/unittest_hvx_mm_u8i4.cpp
dlwlzzero: Since I don't have background of CPU, are there any differences between two operations?

test/unittest/unittest_hvx_mm_u8i4.cpp
dlwlzzero: I think it might be better to do this job, after I merged HTP unittest env to nntrainer's. For now, I unified 3 separate round-up code for the following changes, is it okay?

Seunghui98:
@dlwlzzero Could you rebase this PR to upstream/main?

Seunghui98:
### 1. Description
This PR establishes the foundational HTP/DSP quantization pipeline for the A8W4 (`u8i4`) path, introducing Hexagon HMX matmul execution alongside HVX-bound per-row asymmetric activation quantization and dequantization.

### 2. Summary
* **FastRPC Stub/Skel Infrastructure:** Added a custom FastRPC interface (`test/htp/nntr_hvx.idl`) and build pipeline to produce `libnntr_hvx_skel.so` linked with `libhexkl_micro.a`.
* **HMX MatMul Orchestration:** Implemented VTCM tiling planners (`hexkl_mm_u8i4_plan`), one-shot weight baking to the proprietary WH layout, and the core HMX execution loops (`hexkl_mm_u8i4_run`).
* **HVX Activation Processing:** Created vectorized HVX kernels for per-row asymmetric `uint8` dynamic quantization and `int32` accumulator-to-`f32` dequantization with zero-point correction.

### 3. Bug / Risks
* **Strict 2^22 Magic-Number Limit:** The `hvx_convert.h` conversion trick is mathematically exact only for `|x| <= 2^22` (~4.19M). While accumulators are safely bounded at $K=1024$ (2.08M, a $2\\times$ margin), any long-context inference where $K \\ge 2048$ will exceed this bound and cause silent numeric corruption.
* **Severe Performance Bottleneck (Harness-only Design):** Running `hmx_bringup()` (hw_init & lock) and weight-baking synchronously on every single invocation introduces a massive latency penalty, making this PR purely an accuracy harness.

### 4. What to Look Into Further
* **Lifecycle Hoisting:** To transition from an accuracy harness to a production-ready path, look into moving `hmx_bringup` and `bake_weights` out of the per-call loop into the global session `open` and `close` lifecycles.
* **Doxygen Comment Compliance:** Some comments in `hexkl_mm_u8i4.c` and test files still use the standard C `/*` block comments. To prevent the repository's strict `doxygen-tag.sh` static checker from failing, ensure all multiline comments are converted to begin with `/**`.

nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.h
```diff
@@ -0,0 +1,72 @@
...
+ * @return AEE_SUCCESS, AEE_EBADPARM on a shape violation, or
+ *         AEE_ENOMEMORY when the partitioning does not fit.
+ */
+int hexkl_mm_u8i4_plan(uint8_t *vtcm_base, uint32_t vtcm_size, uint32_t m_pad,
+                       uint32_t k, uint32_t n, hexkl_mm_u8i4_layout *out);
+
+/**
+ * @brief Converts every weight tile to WH layout once, into VTCM.
+ *
+ * HexKL's example calls rm_to_wh_i4 from the innermost matmul loop, which
```
Seunghui98, line 50: ## Note for Reviewers

### What is `rm_to_wh`?
cpu(host) uses `rm`(row major) data type, but
npu uses `wh` data type for weight.

test/htp/build.sh
```diff
@@ -4,15 +4,32 @@
...
 # Override the target with: HEX_ARCH=v75 ./build.sh
+# Override HexKL with:      HEXKL_ROOT=/path/to/hexkl_addon ./build.sh
 
 set -eu
 
 : "${HEXAGON_SDK_ROOT:?source setup_sdk_env.source first}"
 : "${DEFAULT_HEXAGON_TOOLS_ROOT:?source setup_sdk_env.source first}"
 
 HEX_ARCH="${HEX_ARCH:-v79}"
+HEXKL_ROOT="${HEXKL_ROOT:-$HOME/Downloads/hexkl_addon}"
```
Seunghui98: I guess this is hardcorded path.
```
HEXKL_ROOT="${HEXKL_ROOT:-$HOME/Downloads/hexkl_addon}" << ~/Download/
```
Could you check this?

Seunghui98 (commented):
## Description

Brings up the A8W4 DSP path: links HexKL micro, bakes u8i4 weights to WH layout once instead of per tile, adds the HMX matmul loop, and adds HVX kernels for per-row asymmetric uint8 activation quantization (K1/K2) and int32→f32 dequantization (K3), plus a FastRPC endpoint and a shape-level test harness.

## Summary

The testing strategy here is the right one for this kind of work. Requiring the int32 accumulator to match a plain C integer matmul **bit for bit** means a layout or wiring mistake fails loudly instead of being absorbed into a float tolerance later — for a layout that HexKL does not document, that is the only way to know the bake is right. Baking every tile up front rather than converting inside the innermost loop is also a clear win over what HexKL's own example does.

The numeric care is visible too: the note that the host reference must use `nearbyint` rather than `round` because HVX's float add rounds to nearest-even, and that weight quantization keeps `round` since it only runs on the host, is the kind of detail that silently costs days when it is missing.

## Bug

No correctness defect found. The three items below are dead code, misplaced documentation, and an unstated precondition.

1. **`hvx_w_to_sf()` has no callers, and the header's opening rationale contradicts both the code and itself.** `hvx_convert.h` opens with "HVX has no numeric conversion between int32 and f32. `Q6_Vw_equals_Vsf` and `Q6_Vsf_equals_Vw` are bit reinterpretations… The magic-number identity below is the way across." Twenty lines later the same file says the opposite, citing the manual: "`Q6_Vsf_equals_Vw` / `Q6_Vw_equals_Vsf` look like reinterpret but are in fact NUMERIC conversions (V79 HVX Programmer's Reference Manual)". The code follows the second reading — `hvx_dequant_i32.c:47-48` converts the accumulator with `Q6_Vsf_equals_Vw` directly — which leaves `hvx_w_to_sf` defined and never used. Worth deleting the helper and the first paragraph, so the file stops arguing with itself about the one fact the whole approach rests on.

2. **The `2^22` bound is documented on the function it does not constrain.** The header attaches "Our accumulators are bounded by `255 * 8 * K`… Nothing here is safe for larger K without revisiting this bound" to the magic-number helpers. But the accumulator never passes through them. The only live user is `hvx_quant_u8.c:118`, which converts `input * inv_scale` — a uint8 code in [0, 255], five orders of magnitude short of 2^22. Meanwhile the path that actually scales with K carries no documented bound at all: `Q6_Vsf_equals_Vw` in the dequant is exact only while the value fits float32's 24-bit mantissa, i.e. `255 * 8 * K ≤ 2^24`, or K ≈ 8200. Past that the vector path (which computes `af - vzf*csf` in float) and the scalar tail at `hvx_dequant_i32.c:58-61` (which computes `arow[j] - z * colsum_w[j]` in int32) start to disagree. A future reader scaling K will follow the warning to the unused helper and miss the real constraint.

3. **`hvx_quant_pack_u8_ah` silently drops columns when `k % 32 != 0`.** `n_ktiles = k / TILE_INNER` truncates and there is no scalar tail, so the `memset` leaves the remaining columns zero and the quantized activation is wrong with no signal. The dequant kernel does handle its remainder (`n_vec * LANES` tail loop), so the two kernels treat a non-multiple dimension differently. The header documents `m_pad` as "rows after padding up to a multiple of 64" but says nothing about `k`. If `k % 32 == 0` is guaranteed by the HMX tiling, saying so in the doc comment — or asserting it — would keep the guarantee from quietly depending on every caller knowing it.

---

nntrainer/tensor/htp_backend/hvx/hvx_convert.h
```diff
@@ -0,0 +1,84 @@
...
+ * Q6_Vw_equals_Vsf look like reinterpret but are in fact NUMERIC
+ * conversions (V79 HVX Programmer's Reference Manual: "Convert IEEE
+ * floating point to integer... round to zero"), which would destroy the
+ * magic-number identity.
+ */
+static inline HVX_Vector hvx_w_to_sf(HVX_Vector v) {
+  const HVX_Vector magic = Q6_V_vsplat_R(HVX_CONVERT_MAGIC_BITS);
+  const HVX_Vector bits = Q6_Vw_vadd_VwVw(v, magic);
+  return Q6_Vsf_vsub_VsfVsf(bits, magic);
+}
```
Seunghui98: I guess this isn't used.
so, how about removing it?

dlwlzzero: I removed it. Thanks for checking :)

test/htp/build.sh
dlwlzzero: Yes, you're right. So I changed the hardcoded path to warning message.

test/htp/build.sh
```diff
@@ -4,15 +4,32 @@
...
+# Set HexKL path:           HEXKL_ROOT=/path/to/hexkl_addon ./build.sh
 
 set -eu
 
 : "${HEXAGON_SDK_ROOT:?source setup_sdk_env.source first}"
 : "${DEFAULT_HEXAGON_TOOLS_ROOT:?source setup_sdk_env.source first}"
 
 HEX_ARCH="${HEX_ARCH:-v79}"
+HEXKL_ROOT="${HEXKL_ROOT:?set HEXKL_ROOT to your hexkl_addon path}"
+HEXKL_SDK_VER="${HEXKL_SDK_VER:-6.4.0.1}"
```
Seunghui98:
```suggestion
HEXKL_SDK_VER="${HEXKL_SDK_VER:-6.4.0.2}"
```
Ah, Should we use `6.4.0.1` version??
I know that we use 6.4.0.2 for micro api.

dlwlzzero: I'm currently using `6.4.0.1` version, I think that version is not important unless `>6.4.0.x`.
So I'm considering making `HEXKL_SDK_VER` an environment variable so that users can specify it themselves.

test/htp/build.sh
```diff
@@ -4,15 +4,32 @@
...
+HEXKL_TOOLS_VARIANT="${HEXKL_TOOLS_VARIANT:-toolv19}"
+HEXKL_LIB="$HEXKL_ROOT/lib/$HEXKL_SDK_VER/hexagon_${HEXKL_TOOLS_VARIANT}_${HEX_ARCH}/libhexkl_micro.a"
+
+if [ ! -f "$HEXKL_LIB" ]; then
+    echo "Error: HexKL static library not found:" >&2
+    echo "  $HEXKL_LIB" >&2
+    echo "HexKL provides v79 only for lib/6.1.1.0 and newer." >&2
+    exit 1
+fi
```
Seunghui98, line 29:
```suggestion
if [ ! -f "$HEXKL_LIB" ]; then
    echo "Error: HexKL static library not found:" >&2
    echo "  $HEXKL_LIB" >&2
    echo "Available under $HEXKL_ROOT/lib:" >&2
    ls "$HEXKL_ROOT/lib" 2>/dev/null >&2 || echo "  (none)" >&2
    exit 1
fi
```

test/htp/build.sh
dlwlzzero: Btw, how about changing the name of `HEXKL_SDK_VER` to `HEXAGON_SDK_VER`?

Seunghui98 (approved):
LGTM 👍

test/unittest/unittest_hvx_mm_u8i4.cpp
Jungwon-Lee: Since the function has no differences, is it better to reuse the existing implementation rather instead of rewriting?

Jungwon-Lee: ah, this code is for the HTP unit test environment. you will integrate it with nntrainer soon, right?

jaemini-shin: This quantization logic is different from `quant_qs4cx_f32`. It stores one int4 nibble in int8 while `quant_qs4cx_f32` packs two nibbles in one int8.
Also, `quant_qs4cx_f32` adds 8 when packing nibbles, but it doesn't.

dlwlzzero: @Jungwon-Lee, as @jaemini-shin mentioned, there is a difference between two and due to the HMX format, I can't use `quant_as4cs_f32` instead of `quantize_weights_qs4cx`. Will you check?

Jungwon-Lee: ok, i got it
it's good

Jungwon-Lee (approved):
LGTM!

nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.c
```diff
@@ -0,0 +1,128 @@
...
+
+int hexkl_mm_u8i4_bake_weights(const hexkl_mm_u8i4_layout *L,
+                               const int8_t *w_i4_rm, uint32_t k, uint32_t n) {
+  if (!L || !w_i4_rm) {
+    return AEE_EBADPARM;
+  }
+  const uint32_t n_ktiles = k / HEXKL_HMX_INT8_BLOCK_N_INNER;
+  const uint32_t n_ntiles = n / HEXKL_HMX_INT8_BLOCK_N_COL;
+
+  for (uint32_t kt = 0; kt < n_ktiles; ++kt) {
```
jijoongmoon, line 76: I think this can be parallelized.

nntrainer/tensor/htp_backend/hmx/hexkl_mm_u8i4.c
```diff
@@ -0,0 +1,128 @@
...
+int hexkl_mm_u8i4_run(const hexkl_mm_u8i4_layout *L, uint32_t m_pad, uint32_t k,
+                      uint32_t n, int32_t *acc_i32) {
+  if (!L || !acc_i32) {
+    return AEE_EBADPARM;
+  }
+  const uint32_t n_ktiles = k / HEXKL_HMX_INT8_BLOCK_N_INNER;
+  const uint32_t n_ntiles = n / HEXKL_HMX_INT8_BLOCK_N_COL;
+  const uint32_t n_rblocks = m_pad / HEXKL_HMX_INT8_BLOCK_N_ROW;
+
+  for (uint32_t rb = 0; rb < n_rblocks; ++rb) {
```
jijoongmoon, line 98: how about parallelize this loop as well?

PR #4243, opened 2026-08-06
[HTP] Optimize u8i4 performance with HexKL micro-kernel, user-DMA, and cross-layer weight prefetching

Seunghui98:
# Dependency of the PR

#4236

<details><summary>cee4f06d [HTP] Add a dmlink user-DMA ring for weight staging</summary><br />

[HTP] Add a dmlink user-DMA ring for weight staging

Queues 2D transfers onto the single DMA engine by chaining descriptors
with dmlink, so a weight stream and the output write-back share one engine
and drain during compute instead of blocking on dmstart each.

Ported unchanged in behaviour from the doc13 measurement bench, which took
it from llama.cpp ggml-hexagon's dma-queue. Kept dtype-agnostic so the
u8i8 path can share it.

**Self evaluation:**

</details>

<details><summary>f4d02f53 [HTP] Bake u8i4 weights once and prefetch across a layer</summary><br />

[HTP] Bake u8i4 weights once and prefetch across a layer

hexkl_mm_u8i4 bakes the weight to WH layout on every matmul, ~72 us per
call by doc15 §3, and runs one matmul per call so there is nothing to
overlap the DMA with. This adds a registry that bakes once and keeps the
WH bytes resident, plus a run that takes several weights sharing one
activation (Q/K/V, gate/up) and streams the next weight into VTCM while
the current one computes.

Weights stay in DSP heap, not VTCM: only the two being double-buffered
need to be resident, so total weight bytes are not capped by the arena.

**Self evaluation:**

</details>

<details><summary>ad8a54bb [HTP] Hold HMX for the session and add the u8i4 layer endpoint</summary><br />

[HTP] Hold HMX for the session and add the u8i4 layer endpoint

nntr_hvx_open now does hw_init, takes the HMX lock and runs
setup_acc_read_int32 once for the session; close releases it and any
weights still registered. mm_u8i4_from_f32 reaches all three through the
handle instead of repeating the preamble per invoke.

mm_u8i4_layer is the performance entry point: handles in, f32 out, none
of the four rout buffers the accuracy harness returns. That harness keeps
its endpoint unchanged so unittest_hvx_mm_u8i4 still checks every stage.

**Self evaluation:**

</details>

<details><summary>ff47e4c3 [test] Cover the u8i4 layer endpoint</summary><br />

[test] Cover the u8i4 layer endpoint

Reuses HmxMmU8I4's quantize/dequant helpers as the reference rather than
duplicating them: the layer endpoint's job is to match what calling the
accuracy endpoint once per weight would produce, so that is what these
compare against.

Covers: several weights against one shared activation matching the
per-weight reference, a registered weight surviving repeated calls
bit-for-bit, a released handle rejected and its slot reused, a
mismatched-K handle rejected, and a printed (not asserted) per-call
cost comparison against the accuracy harness -- thermal state moves the
numbers between runs, so a threshold here would encode lab conditions
rather than a property of the code.

**Self evaluation:**

</details>

<details><summary>0b1ee7fc [test] Add an end-to-end device runner for the u8i4 layer endpoint</summary><br />

[test] Add an end-to-end device runner for the u8i4 layer endpoint

Builds the DSP skel, libnntrainer.so, and the ARM gtest, pushes all three
plus libc++_shared to the device, and runs unittest_hvx_mm_u8i4.

Verified on R3CY10WM83Y: all 9 tests pass, including the 4 existing
accuracy-harness shapes (the session-scoping refactor did not change their
result) and the 5 new layer-endpoint tests. Reported (not asserted)
speedup_vs_harness came out at 3.31x, above doc13 §3a's 1.7-2x -- that
number was DSP-only with no FastRPC in the timed region, this one is
end-to-end, so the session-scoped HMX and dropped debug buffers add on
top of the DMA/prefetch win doc13 measured in isolation.

**Self evaluation:**

</details>

### Summary

- **Add a dmlink user-DMA ring for weight staging**: chain descriptors onto one DMA engine instead of blocking dmstart per transfer, dtype-agnostic so u8i8 can reuse it
- **Bake u8i4 weights once and prefetch across a layer**: keep WH-baked weights resident and prefetch the next one while the current matmul runs, instead of re-baking (~72us) every call
- **Hold HMX for the session and add the u8i4 layer endpoint**: move hw_init/HMX lock/acc_read setup into open/close, add mm_u8i4_layer as the lean performance entry point next to the unchanged accuracy harness
- **Cover the u8i4 layer endpoint**: 5 new tests against the existing per-weight reference (multi-weight correctness, repeated-call determinism, handle release/reuse, mismatched-K rejection, per-call cost)
- **Add an end-to-end device runner for the u8i4 layer endpoint**: build+push+run script; verified on Galaxy S25 Ultra (V79): 9/9 tests pass, speedup_vs_harness = 3.31x (doc13 §3a's DMA/prefetch-only number was 1.7-2x)

---

### DIFF
--- a/nntrainer/tensor/htp_backend/hmx/hexkl_dma_ring.c
+++ b/nntrainer/tensor/htp_backend/hmx/hexkl_dma_ring.c
@@ -1,0 +1,131 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Copyright (C) 2026 SeungHui Lee <shsh1004.lee@samsung.com>
+ *
+ * @file   hexkl_dma_ring.c
+ * @date   06 Aug 2026
+ * @brief  Minimal dmlink user-DMA ring for cross-matmul weight prefetch
+ * @see    https://github.com/nntrainer/nntrainer
+ * @author SeungHui Lee <shsh1004.lee@samsung.com>
+ * @bug    No known bugs except for NYI items
+ *
+ * Ported from the doc13 §3a measurement bench's ring_push2d/ring_drain,
+ * itself lifted from llama.cpp ggml-hexagon's dma-queue.{c,h}. Behaviour is
+ * unchanged; only the names are de-abbreviated and the file stands alone so
+ * both the u8i4 and (later) u8i8 matmul modules can share one ring.
+ */
+
+#include "hexkl_dma_ring.h"
+
+#include <hexagon_protos.h>
+#include <hexagon_types.h>
+#include <hmx_hexagon_protos.h>
+#include <string.h>
+
+/** @brief One 2D DMA descriptor. Hardware-defined layout; do not reorder. */
+typedef struct __attribute__((aligned(128))) hexkl_dma_desc2d_s {
+  void *next;
+  uint32_t dst_stride : 24;
+  uint32_t desc_size : 2;
+  uint32_t dst_comp : 1;
+  uint32_t src_comp : 1;
+  uint32_t dst_bypass : 1;
+  uint32_t src_bypass : 1;
+  uint32_t order : 1;
+  uint32_t done : 1;
+  void *src;
+  void *dst;
+  uint32_t desc_type : 8;
+  uint32_t reserved0 : 24;
+  uint32_t row_size : 24;
+  uint32_t nrows_lo : 8;
+  uint32_t nrows_hi : 8;
+  uint32_t src_stride : 24;
+  uint32_t offset : 24;
+  uint32_t reserved1 : 8;
+} hexkl_dma_desc2d;
+
+static inline void hexkl_dma_link(void *cur, void *next) {
+  asm volatile(" release(%0):at" : : "r"(next));
+  asm volatile(" dmlink(%0, %1)" : : "r"(cur), "r"(next));
+}
+
+static inline void hexkl_dma_start(void *p) {
+  asm volatile(" release(%0):at" : : "r"(p));
+  asm volatile(" dmstart(%0)" : : "r"(p));
+}
+
+/** @brief Power of two, comfortably above the pushes any one call issues,
+ *         so a call's transfers never wrap into ones still in flight. */
+#define HEXKL_DMA_RING_N 256u
+
+static hexkl_dma_desc2d g_ring[HEXKL_DMA_RING_N] __attribute__((aligned(128)));
+static uint32_t g_push, g_pop;
+static hexkl_dma_desc2d *g_tail;
+static int g_started;
+
+static void hexkl_dma_ring_wait_idx(uint32_t i) {
+  long guard = 0;
+  while (!g_ring[i].done && guard++ < 50000000L) {
+    unsigned r = 0;
+    asm volatile(" %0 = dmpoll" : "=r"(r) : : "memory");
+    (void)r;
+  }
+}
+
+void hexkl_dma_ring_reset(void) {
+  memset(g_ring, 0, sizeof(g_ring));
+  g_push = 0;
+  g_pop = 0;
+  g_tail = &g_ring[HEXKL_DMA_RING_N - 1];
+  g_started = 0;
+}
+
+void hexkl_dma_ring_push2d(void *dst, const void *src, uint32_t dst_stride,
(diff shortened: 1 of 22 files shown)


Seunghui98:
Rebased onto #4236's current head (`2f22a22`) and today's `main` (`9a3b5b8`). That drops the 4 already-merged #4226 bring-up commits this branch was carrying and picks up #4236's two review fixes, so the diff no longer shows either as new work.

**Only the last 6 commits belong to this PR** — the 7 before them are #4236, still in review:

| commit | subject |
|---|---|
| `c4c7ce9` | [HTP] Add a dmlink user-DMA ring for weight staging |
| `28a8182` | [HTP] Bake u8i4 weights once and prefetch across a layer |
| `8f2bb81` | [HTP] Hold HMX for the session and add the u8i4 layer endpoint |
| `fc33d45` | [test] Cover the u8i4 layer endpoint |
| `4ed9363` | [test] Add an end-to-end device runner for the u8i4 layer endpoint |
| `bb019f6` | style: apply clang-format to modified htp/u8i4 files |

### Cleanups added on top

Five more commits, all of which came out of preparing this for review:

- **`23bea06`** — `[HTP] Derive the HMX tile sizes from HexKL instead of restating them`. Three files carried their own copies of the tile geometry: `hexkl_mm_u8i4.c` and `hexkl_mm_u8i4_dma.c` (512/8192, the latter with a comment saying they are "kept in sync by inspection") and `hvx_quant_u8.c` (64/32/2048, "mirrored from hexkl_micro.h" but never including it). The quantizer writes activation tiles at `ACT_TILE_BYTES` and the matmul reads them at `HEXKL_HMX_ACTIVATION_ALIGNMENT` — the same required number spelled two ways, agreeing only because both are 2048 today. `hexkl_hmx_geom.h` derives them from `HEXKL_HMX_*` and static-asserts the results, so a future SDK breaks the build instead of the data.

- **`7747b34`** — `[HTP] Validate the layer run's shape before deriving offsets from it`. The `K % 32` check ran after `k_tiles` was already computed from `K`; `config_off` arrives from the session and was only ever checked implicitly. Both now run before any offset is derived. Also fixes a comment that claimed the handle loop collects the total output width when it collects the maximum.

  This also points `nntr_hvx_mm_u8i4.c`'s `ROUND_UP` at the shared helper. @Jungwon-Lee's [earlier note](https://github.com/nntrainer/nntrainer/pull/4236#discussion_r3717848135) asked for one round-up implementation instead of three — `85ecd64` made them agree in form, this makes them one.

- **`5ed3af0`** — `[test] Stop baking one machine's install paths into the device runner`. `run_u8i4_layer_on_device.sh` defaulted `HEXAGON_SDK_ROOT`, `HEXKL_ROOT` and `ANDROID_NDK` to absolute paths under `$HOME/workspace`, and pinned the Hexagon toolchain to `19.0.04`. It also still exported `HEXKL_SDK_VER`, which stopped doing anything when `build.sh` began deriving the version from `HEXAGON_SDK_ROOT`'s directory name — a variable nothing reads, that still looked like it pinned the version. Plus: fixed `/tmp` log path → `mktemp`, hardcoded `linux-x86_64` NDK prebuilt dir → globbed, and a `[ -f x ] && adb push` that exits the whole script under `set -e` when the file is absent.

- **`917b391`** — `[htp] Replace the internal design-doc citations with what they say`. Sixteen comments pointed at `doc13 §3a`, `doc15 §4/§8 item 3`, `HTP_PR_PLAN.md PR①` and `PR②/③`. Those documents are not in the tree and are not headed upstream, so a reader who follows the pointer finds nothing. Each was a citation for a fact the comment already stated, so only the citation goes.

- **`a1cf6c2`** — clang-format over the two files the above touched.

### Not verified on device

The functional commits are unchanged from the revision that ran 9/9 on the S25 Ultra. The five cleanup commits are **not** compiled or run — no Hexagon SDK here — so `test/htp/run_u8i4_layer_on_device.sh` should be run once before merge. The one to watch is the `_Static_assert` block in `hexkl_hmx_geom.h`: it assumes `HEXKL_HMX_ACTIVATION_ALIGNMENT` and friends are integer constant expressions. If any of them is not, the build fails there loudly rather than silently.

---

PR #4252, opened 2026-08-06
[CausalLM] Store quantized models in self-contained output directories

Jungwon-Lee:
## Dependency of the PR

None.

#### Behavior change

| Invocation | Before | After |
| --- | --- | --- |
| `nntr_quantize <model_path>` without `--output` | The output path defaulted to the FP32 model directory, so quantized weights and configuration were created alongside the source files. | A dtype- and ISA-specific subdirectory is created under the FP32 model directory, and the quantized model is stored there. |
| `nntr_quantize <model_path> --output <path>` | The quantized model and auxiliary inference files were stored in the specified directory. | Unchanged. |

The generated default directory follows:

`quantized_fc_<fc>_embd_<embedding>_lmhead_<lmhead>_isa_<isa>`

- Reuse the existing auxiliary-file copy flow for the new default output directory.
- Add `tokenizer.model`, `spiece.model`, and `sentencepiece.bpe.model` to the copied tokenizer assets.
- Generate `nntr_config.json` from the original configuration so its relative paths are preserved.
- Explicit `--output` and intentional in-place output behavior remain unchanged.

**Self evaluation:**

### Summary

Keep automatically generated quantized models separate from the source FP32 model while preserving the existing explicit-output workflow.

### DIFF
--- a/Applications/CausalLM/quantize.cpp
+++ b/Applications/CausalLM/quantize.cpp
@@ -31,7 +31,8 @@
  *                           nntr_config.json, and the .bin weight file.
  *
  *   Options:
- *     --output, -o <path> Output directory (default: <model_path>)
+ *     --output, -o <path> Output directory (default: generated from the
+ *                         dtype configuration and target ISA)
  *     --fc_dtype <type>   Target dtype for FC layers (default: Q4_0)
  *     --embd_dtype <type> Target dtype for embedding layer (default: FP32)
  *     --lmhead_dtype <type> Target dtype for LM head layer (default: FP32)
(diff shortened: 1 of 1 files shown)


Applications/CausalLM/quantize.cpp
```diff
@@ -867,6 +886,9 @@ int main(int argc, char *argv[]) {
                                  "special_tokens_map.json",
                                  "vocab.json",
                                  "merges.txt",
+                                 "tokenizer.model",
+                                 "spiece.model",
+                                 "sentencepiece.bpe.model",
```
Seunghui98, line 891: Why do we need to add these lines?

Issue #4169, opened 2026-07-30
Todo items for weight repacking


jaemini-shin:
- [x] transformer should not be layer dependant (#4157 )
  - deciding whether repack or not should be handled at derived model not transformer

PR #4254, opened 2026-08-07
[CausalLM] delegate repack to each layers

jaemini-shin:
## Related
* #4169

<details><summary>[CausalLM] delegate repack to each layers</summary><br />

This commit delegates whether and how to repack weights to each layer
not from the transformer object. With this commit, we don't have to
change transformer logic to add new rule to repack weights.

</details>

### Summary

Let's delegate repack logic to each layer.

**Self evaluation:**

### DIFF
--- a/Applications/CausalLM/layers/shared_fully_connected_layer.cpp
+++ b/Applications/CausalLM/layers/shared_fully_connected_layer.cpp
@@ -184,6 +184,14 @@ void SharedFullyConnectedLayer::exportTo(
   exporter.saveResult(fc_props, method, this);
 }
 
+void SharedFullyConnectedLayer::pack(nntrainer::RunLayerContext &context) {
+  for (auto &w : context.getWeights()) {
+    nntrainer::Tensor &var = w->getVariableRef();
+    if (var.getDataType() == nntrainer::TensorDim::DataType::QS4CX)
+      var.pack();
+  }
+}
+
 void SharedFullyConnectedLayer::read(
   std::ifstream &file, nntrainer::RunLayerContext &context, bool opt_var,
   ml::train::ExecutionMode mode, bool trainable,
--- a/Applications/CausalLM/models/transformer.cpp
+++ b/Applications/CausalLM/models/transformer.cpp
@@ -361,22 +361,12 @@ void Transformer::repack_weight() {
   }
   std::function<void(ml::train::Layer &, nntrainer::RunLayerContext &, void *)>
     fn = [](ml::train::Layer &l, nntrainer::RunLayerContext &context, void *) {
-      // repack FC layer only
-      if (l.getType() != "fully_connected" &&
-          l.getType() != "shared_fully_connected")
-        return;
-
-      auto weights = context.getWeights();
-      for (auto &w : weights) {
-        if (w->getVariableRef().getDataType() ==
-            ml::train::TensorDim::DataType::QS4CX) {
-          w->getVariableRef().pack();
-        }
-      }
+      // Each layer determines whether it repacks its own weights or not
+      static_cast<nntrainer::LayerNode &>(l).pack(context);
     };
   try {
     model->forEachLayer(fn, nullptr);
-    ml_logd("QS4CX weights repacked successfully");
+    ml_logd("weights repacked successfully");
   } catch (const std::exception &e) {
     throw std::runtime_error("Failed to repack weights: " +
                              std::string(e.what()));
--- a/nntrainer/layers/layer_devel.h
+++ b/nntrainer/layers/layer_devel.h
@@ -246,6 +246,15 @@ public:
   virtual void calcGradient(RunLayerContext &context) {}
 
   /**
+   * @brief     Pack the layer's weights into a backend-optimized layout
+   * @details   Invoked after weights are loaded.
+   * - The default implementation is a no-op
+   * - Each derived layer overrides this if weight repack needed
+   * @param     context Context of the layer, providing access to its weights
+   */
+  virtual void pack(RunLayerContext &context) {}
+
+  /**
    * @brief     set Property of layer
    * @param     values values of property
    * @throw std::invalid_argument invalid parameter.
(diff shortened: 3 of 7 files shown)


Applications/CausalLM/models/transformer.cpp
```diff
@@ -361,22 +361,12 @@ void Transformer::repack_weight() {
...
-
-      auto weights = context.getWeights();
-      for (auto &w : weights) {
-        if (w->getVariableRef().getDataType() ==
-            ml::train::TensorDim::DataType::QS4CX) {
-          w->getVariableRef().pack();
-        }
-      }
+      // Each layer determines whether it repacks its own weights or not
+      static_cast<nntrainer::LayerNode &>(l).pack(context);
```
jaemini-shin, line 365: With this PR, `transformer` don't have to care about `pack` implementation.

nntrainer/layers/layer_devel.h
```diff
@@ -245,6 +245,15 @@ class Layer {
...
   virtual void calcGradient(RunLayerContext &context) {}
 
+  /**
+   * @brief     Pack the layer's weights into a backend-optimized layout
+   * @details   Invoked after weights are loaded.
+   * - The default implementation is a no-op
+   * - Each derived layer overrides this if weight repack needed
+   * @param     context Context of the layer, providing access to its weights
+   */
+  virtual void pack(RunLayerContext &context) {}
```
jaemini-shin, line 255: Default is no-op.

Applications/CausalLM/layers/shared_fully_connected_layer.cpp
```diff
@@ -184,6 +184,14 @@ void SharedFullyConnectedLayer::exportTo(
   exporter.saveResult(fc_props, method, this);
 }
 
+void SharedFullyConnectedLayer::pack(nntrainer::RunLayerContext &context) {
+  for (auto &w : context.getWeights()) {
+    nntrainer::Tensor &var = w->getVariableRef();
+    if (var.getDataType() == nntrainer::TensorDim::DataType::QS4CX)
+      var.pack();
+  }
+}
```
jaemini-shin, line 193: Each layer now implements what to do.

jayden0701 (approved):
LGTM!

Seunghui98 (commented):
## Description

Moves the "which weights get repacked" decision out of `Transformer::repack_weight()` and into the layers: adds a virtual `Layer::pack(RunLayerContext&)` defaulting to a no-op, a `LayerNode::pack()` forwarder, and overrides on `FullyConnectedLayer` and `SharedFullyConnectedLayer`. `repack_weight()` now just calls `pack()` on every node.

## Summary

The direction is right — a type-string whitelist in the model object is exactly the kind of thing that rots as layers are added, and pushing the decision to the layer that owns the weights is the correct place for it. The `static_cast<LayerNode &>(l)` is also safe: `NeuralNetwork::forEachLayer` (`neuralnet.cpp:2271-2272`) does `static_pointer_cast<LayerNode>(*iter)` and passes `*ln`, so the dynamic type really is `LayerNode`.

## Bug

**`FullyConnectedLayerCl` silently loses its repack.**

The old whitelist matched on the type string:

```cpp
if (l.getType() != "fully_connected" && l.getType() != "shared_fully_connected")
  return;
```

Two classes answer `"fully_connected"`:

| class | header | base |
|---|---|---|
| `FullyConnectedLayer` | `nntrainer/layers/fc_layer.h:117` | `LayerImpl` |
| `FullyConnectedLayerCl` | `nntrainer/layers/cl_layers/fc_layer_cl.h:108` | `LayerImplCl` |

The old code repacked both, because it never looked at the C++ type. The new code only repacks classes that override `pack()`, and `FullyConnectedLayerCl` does not — it is not derived from `FullyConnectedLayer`, so it inherits the no-op `Layer::pack()`.

Result: in an OpenCL-engine model with QS4CX weights, the FC weights are no longer packed. Nothing throws and nothing logs — `repack_weight()` still reports `"weights repacked successfully"` — so the first symptom is whatever the unpacked layout does at inference time. Adding the same override to `FullyConnectedLayerCl` restores parity.

1. **This case is worth treating as evidence about the design, not just a missed file.** The PR's stated benefit is that adding a new repacking layer no longer requires touching `transformer.cpp` — but the cost is that *forgetting* to override `pack()` is now invisible, where previously forgetting to add a type string at least left the old whitelist visible in one place. `FullyConnectedLayerCl` is that failure happening on the first change. If the set of layers that need packing is meant to stay small and explicit, a `ml_logd` in `repack_weight()` naming the layers that actually packed something would make an omission visible in a log instead of in output quality.

2. **The two `pack()` bodies are identical.** `fc_layer.cpp:217-223` and `shared_fully_connected_layer.cpp:187-193` are the same loop, and a third copy is needed for the CL layer. Both classes derive from `LayerImpl`, but hoisting it there would change behaviour for every other `LayerImpl` subclass with QS4CX weights (embedding layers, for instance), which the old whitelist deliberately excluded — so a small shared helper called by each override looks safer than a base-class implementation.

3. **Minor:** `repack_weight()` no longer distinguishes "packed nothing" from "packed everything", and the log line lost its QS4CX qualifier. With the decision now spread across layers, that message is the only remaining signal that the pass ran at all.

---

Applications/CausalLM/models/transformer.cpp
```diff
@@ -361,22 +361,12 @@ void Transformer::repack_weight() {
   }
   std::function<void(ml::train::Layer &, nntrainer::RunLayerContext &, void *)>
     fn = [](ml::train::Layer &l, nntrainer::RunLayerContext &context, void *) {
-      // repack FC layer only
-      if (l.getType() != "fully_connected" &&
-          l.getType() != "shared_fully_connected")
-        return;
```
Seunghui98, line 367: I guess this condition is removed.
Coud you explain why we can remove it?

jaemini-shin: Before this PR, `transformer::repack_weight()` controlled conditions.

From this PR, `transformer::repack_weight()` calls each layer's `pack()`.
i.e. `fc_layer::pack()`, `shared_fully_connected::pack()` controls repacking, not from the `transformer::repack_weight()`.

Seunghui98: Ah, I got this! Thanks for letting me know 👍

Seunghui98 (approved):
LGTM

PR #4288, opened 2026-08-24
[CI] Fix GBS RPM cache size growing bug

jaemini-shin:

<details><summary>[CI] Fix GBS RPM cache size growing bug</summary><br />

RPM cache includes packages which are necessary for gbs build.
When these packages are newly updated, actions/cache/save saves both old
and new packages which causes RPM cache size explosion.
Let's use pruning script to save newest packages only.

</details>

### Summary
As of now, gbs RPM cache consumes most of the cache capacity due to bug.
Let's fix it and make free spaces for this repository's cache.

* Before this PR
a

* After this PR
b

* 5.88GB -> 1.05GB

### DIFF
--- a/.github/workflows/gbs-cache-prune.sh
+++ b/.github/workflows/gbs-cache-prune.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+##
+# @file  gbs-cache-prune.sh
+# @brief Keep only the newest version of each RPM in the GBS RPM cache.
+#
+# GBS never purges superseded packages from ~/GBS-ROOT/local/cache.
+# This keeps, per package (name+arch), only the most-recently-downloaded .rpm
+# and deletes older versions, bounding the cache to one snapshot.
+
+# A wrongly-pruned package only costs a re-download on the next run, never a
+# build failure, so erring toward pruning is safe.
+set -u
+
+CACHE="${1:-$HOME/GBS-ROOT/local/cache}"
+
+if [ ! -d "$CACHE" ]; then
+  echo "gbs-cache-prune: no cache dir at $CACHE, nothing to do"
+  exit 0
+fi
+
+echo "== gbs package cache before prune =="
+du -sh "$CACHE" 2>/dev/null || true
+before=$(find "$CACHE" -name '*.rpm' 2>/dev/null | wc -l)
+echo "rpm files: $before"
+
+# newest first (by mtime); keep first occurrence of each package key, list the rest
+prune_list=$(
+  find "$CACHE" -type f -name '*.rpm' -printf '%T@\t%p\n' 2>/dev/null \
+    | sort -rn \
+    | awk -F'\t' '
+        {
+          path = $2
+          n = split(path, seg, "/"); base = seg[n]
+          key = base
+          # strip trailing -<version>-<release>.<arch>.rpm to group all versions
+          sub(/-[^-]+-[^-]+\.[^.]+\.rpm$/, "", key)
+          if (key in seen) { print path }   # older duplicate -> prune
+          else             { seen[key] = 1 }
+        }'
+)
+
+# surface how aggressive the prune is, so a runaway regex is visible in the log
+pruned=$(printf '%s' "$prune_list" | grep -c . || true)
+echo "pruning $pruned of $before rpm(s)"
+if [ -n "$prune_list" ]; then
+  printf '%s\n' "$prune_list" | while IFS= read -r f; do [ -n "$f" ] && rm -f "$f"; done
+fi
+
+echo "== gbs package cache after prune =="
+du -sh "$CACHE" 2>/dev/null || true
+after=$(find "$CACHE" -name '*.rpm' 2>/dev/null | wc -l)
+echo "rpm files: $after"
+
+# a correct prune always keeps the newest of each key, so it can never empty a
+# non-empty cache; if it did, the key logic is broken -- flag it loudly
+if [ "$before" -gt 0 ] && [ "$after" -eq 0 ]; then
+  echo "gbs-cache-prune: ERROR pruned all $before rpm(s); key logic broken" >&2
+fi
--- a/.github/workflows/daily_build_tizen_arm.yml
+++ b/.github/workflows/daily_build_tizen_arm.yml
@@ -34,7 +34,7 @@ jobs:
       run: |
         echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
 
-    - name: restore gbs cache from main branch
+    - name: restore gbs RPM cache from main branch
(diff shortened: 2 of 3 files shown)


.github/workflows/gbs-cache-prune.sh
```diff
@@ -0,0 +1,58 @@
...
+  find "$CACHE" -type f -name '*.rpm' -printf '%T@\t%p\n' 2>/dev/null \
+    | sort -rn \
+    | awk -F'\t' '
+        {
+          path = $2
+          n = split(path, seg, "/"); base = seg[n]
+          key = base
+          # strip trailing -<version>-<release>.<arch>.rpm to group all versions
+          sub(/-[^-]+-[^-]+\.[^.]+\.rpm$/, "", key)
+          if (key in seen) { print path }   # older duplicate -> prune
```
dlwlzzero: The header says packages are grouped by name+arch, but this `sub()` strips `.<arch>` along with version and release, so the key is only the package name. The directory isn't part of the key either.

So if the cache has `foo-1.0-1.armv7l.rpm` and `foo-1.0-1.noarch.rpm`, or the same package in two repo subfolders, only the newest one is kept. The other is still needed, so GBS downloads it again on every run and this step deletes it again before the save. The build doesn't break, but that file never stays in the cache.

A key that matches the comment, with the directory added:

```awk
arch = base; sub(/\.rpm$/, "", arch); sub(/.*\./, "", arch)
name = base; sub(/-[^-]+-[^-]+\.[^.]+\.rpm$/, "", name)
key = substr(path, 1, length(path) - length(base)) name "." arch
```

I checked it locally on a small input: `foo` armv7l/noarch and `foo` in two folders are all kept, and only the older armv7l version is pruned. If name-only grouping is what you meant, updating the header comment is enough.

jaemini-shin: Thanks @dlwlzzero — adopted both points and verified against the real cache.

**Key** → now `dir + name + arch`. Keeping `arch` in the key also fixes the `noarch` vs arch-specific collapse (the earlier `key = path` idea kept the directory but still stripped arch, so it didn't). Header comment updated to match.

**Selection** → switched from newest-`mtime` to **highest rpm version** (`sort -V` on `version-release`), so a cache restore that scrambles mtimes can no longer keep the wrong file. The `@brief "newest version"` is now literally what the code does.

**Verified on the actual cache** — restored `gbs-cache-x86_64-2026-09-21` (2064 rpm / 2258 MB) read-only (never re-saved) and read ground-truth `NAME/ARCH/EPOCH:VERSION-RELEASE` via `rpm -qp`:

- parse mismatches: **0** — the filename `(dir, name, arch)` key equals the real rpm `NAME/ARCH` for every file, so no distinct packages are collapsed.
- 287 unique packages; pruning superseded versions takes **2258 MB → ~203 MB**.
- On this x86_64 snapshot no package name spans more than one arch or more than one repo subtree, so the collision doesn't currently occur — but the dir+arch key closes it regardless, and the version-based selection is the robust fix independent of mtime.

Commit updated (force-pushed).

dlwlzzero: Thanks for the quick update and the check against the real cache. The dir + name + arch key looks right to me.

I ran `be35d0f1` on a small synthetic cache. Every case gives the expected result except one:

| cache contents | kept |
|---|---|
| `foo` 1.9 / 1.10 | 1.10 ✅ |
| `foo` x86_64 / noarch | both ✅ |
| `foo` in `base/` and `unified/` | both ✅ |
| `bar` 2.0~rc1 / 2.0 | 2.0 ✅ |
| `qux` 1.2-3.1 / 1.2.1-1.1 | 1.2.1 ✅ |
| `baz` 3.1 (older mtime) / 3.0 (just downloaded) | **3.1 kept, 3.0 deleted** ❌ |

The last row is a version going **down**: the snapshot rolls back, so today's build downloads `baz-3.0` while `baz-3.1` is still in the cache. With version-based selection the prune deletes 3.0, which is the file the build actually used, and keeps the unused 3.1. The next run downloads 3.0 again and it gets deleted again. This repeats until a version above 3.1 shows up. The earlier mtime-based selection would have kept 3.0.

About the reason for the switch: I don't think a cache restore scrambles mtimes. `actions/cache` saves and restores with tar, and GNU tar restores the original mtimes on extract by default. I haven't checked this on a runner. A file downloaded by the current build always has the newest mtime, so mtime is a good signal for "what this build used".

Suggestion: go back to newest-mtime selection and keep the new key. The key was the real bug, and it's fixed now. Not a blocker; this only matters when a rollback happens.

<details><summary>repro</summary>

```bash
mkdir -p c/base c/unified
mk(){ touch -d "$2" "c/$1"; }
mk base/foo-1.9-1.1.x86_64.rpm  2026-09-01; mk base/foo-1.10-1.1.x86_64.rpm 2026-09-02
mk base/foo-1.10-1.1.noarch.rpm 2026-09-02; mk unified/foo-1.9-1.1.x86_64.rpm 2026-09-01
mk base/bar-2.0~rc1-1.1.x86_64.rpm 2026-09-01; mk base/bar-2.0-1.1.x86_64.rpm 2026-09-02
mk base/baz-3.1-1.1.x86_64.rpm 2026-09-01; mk base/baz-3.0-1.1.x86_64.rpm 2026-09-20
mk base/qux-1.2-3.1.x86_64.rpm 2026-09-01; mk base/qux-1.2.1-1.1.x86_64.rpm 2026-09-02
bash .github/workflows/gbs-cache-prune.sh c >/dev/null; find c -name '*.rpm' | sort
```
</details>

dlwlzzero (approved):
LGTM. The key fix resolves the collision. The mtime-vs-version note in the thread is non-blocking.

.github/workflows/gbs-cache-prune.sh
jaemini-shin: Thanks for catching the rollback case and running the repro. You're right — the version-based selection keeps the unused file while deleting the one the build just used, causing re-downloads until a higher version appears.

**Reverted to mtime selection with the new key in place.** mtime answers the right question ("what did this build download"), whereas version answers a different one. And `actions/cache` preserves mtimes on extract, so mtime is stable.

**Verified on the actual cache** — restored `gbs-cache-x86_64-2026-09-21` (2064 rpm / 2258 MB), prune with mtime+key, ran one `gbs build`:
- **prune-caused re-downloads: 0** 
- 1777 files pruned, 121 appeared during build, all 121 were brand-new upstream versions (acl-2.3.2, cmake-3.31.2, gcc-14.2.0, glibc-2.40, etc.) — none from the 1777 pruned set.

So mtime+key keeps what the build needs and avoids the rollback defect. Commit updated (force-pushed).

PR #4292, opened 2026-08-24
[draft] Feat/ced w8a8 int8 resident

baek2sm:
ced hugging face model ( timm vit )


dongkyu-kim:
**[Review — CED W8A8 int8-resident PR]**

Thanks for the huge amount of work here — the CED W8A8 int8-resident path is impressive, and the documentation (`W8A8_DESIGN.md`, `OPTIMIZATION_PLAYBOOK.md`, `VERIFICATION_PYTORCH.md` with 67/67 detections matching PyTorch FP32, 81/87 keypoints vs ONNX Runtime, 12.1 ms / 114 MB on Android arm64) sets a high bar for verification rigor. The `rms_norm` `forwarding()` fix (delegating to `incremental_forwarding`) is also a genuine bug fix.

That said, in its current form I don't think this is mergeable. The main issues below, roughly in priority order.

## 1. Scope: this PR bundles at least four independent concerns

The title says "CED W8A8 int8 resident", but the diff (~528 files, +53k/−2.2k) also contains:

- The wholesale `Applications/CausalLM` → `Applications/quick_ai` rename (namespace, meson vars, docs, unittests) — several hundred files by itself
- New YOLOv7 / YOLOv7-tiny detector apps
- YOLOv7Pose + the RTMCC GAU/SimCC head layers (`rtmcc_gau`, `rtmcc_head`) — an entirely separate model family
- FastViTKeyword + its custom attention layer
- New `xi_pooling` layer
- CI workflow rewrites and unrelated build-option flips (see §4)

Please split this: (a) the rename alone as one PR (purely mechanical, easy to review fast), (b) CED + W8A8 core engine support, (c) each new model app separately. Mixed-precision `_fp8` dot symbols in `moe_layer.cpp` / `moe_layer_cached.cpp` also appear to leak in from the separate FP8 effort (#4175) — those hunks should be dropped here.

## 2. Build regression in `Applications/meson.build` (blocking)

The `if get_option('enable-transformer')` / `platform != 'tizen'` guard was removed and `subdir('quick_ai')` is now unconditional. This breaks `enable-transformer=false` builds and Tizen builds — @jijoongmoon already confirmed the link failure (`libtokenizers_c.a` missing; worked around with `-Dquick_ai-tokenizer=false` on `jijoongmoon:tizen/ced-armv7l`). Please restore the guard (renamed to the quick_ai option) so minimal/Tizen builds keep working without the tokenizer stack.

Related: the PR is in `dirty` merge state and has zero CI check runs, so please rebase onto current `main` and get CI green. Also the title still says `[draft]` while the PR isn't marked as a draft — please pick one.

## 3. Quantization path divergence: `quantizer.cpp` vs `tensor.cpp`

`Quantizer::unknownquantize()` for `Q8_0_PERCH_DEQUANT` copies the full tensor into `weights_perch` to do taps-first→taps-last permutation — and then `Tensor::quantize()` → `quantize_with_Q8_0(..., true, true)` permutes/copies again. Beyond the double copy, the two paths are not equivalent: the `tensor.cpp` path reverses rows when `RT == 1`, while the `quantizer.cpp` path never does. A layer quantized through one path vs the other will produce different weight layouts for `RT == 1` tensors. Please consolidate into a single implementation (or assert one path is unreachable) and add a test that would catch layout divergence. The `getData<uint8_t>()` reinterpretation of an int8 tensor buffer in the quantizer path is also fragile — use a typed accessor or a comment justifying the aliasing.

## 4. Unrelated build/CI changes that should be split or reverted

- `meson_options.txt`: `werror` flipped from `disabled` to `enabled` — a global build-behavior change for all users, hidden in a feature PR.
- `jni/Android.mk` / `Application.mk`: `OPENCL=false` default flip, `NEON=true`, NNTrainer/TFLite version bumps — unrelated to CED.
- `.github/workflows/cpp_linter.yml`: moving off `cpp-linter-action` to a manual `clang-format --dry-run --Werror` loop loses PR annotations, fetches `origin/main` redundantly after `fetch-depth: 0`, and uses a two-dot `git diff origin/main HEAD` which flags files changed on main rather than in the PR (should be `origin/main...HEAD`).
- `.github/workflows/static.check.yml`: exempting `libtokenizers_c.a` / `libtokenizers_cpp.a` / `libsentencepiece.a` from the file-size check — please justify checking in prebuilt binaries at all, or generate/fetch them at build time.
- `Applications/quick_ai/json.hpp`: a ~24.6k-line vendored nlohmann header newly checked in (previously fetched by `prepare_encoder.sh`). Please keep fetching it at build time, or use a subproject/submodule — this is permanent repo bloat.
- `test/jni/gclayer.py` (debug helper with runtime `input()`) shouldn't ship.

## 5. Stale references left behind by the rename

- `Applications/quick_ai/models/FastViTKeyword/jni/main.cpp`: default `RES_DIR` still points at `Applications/CausalLM/models/FastViTKeyword/res` (pre-rename path).
- `.gitignore`: tail still has `Applications/CausalLM/models/YOLOv11/jni/` and `yolo11*.pt` entries.
- `Applications/quick_ai/meson.build`: the `error()` message still points to `Applications\CausalLM\build_tokenizer_windows.ps1`.
- `README.md` / `TESTING.md`: many `CausalLM` / `Applications\CausalLM` references remain.
- `Applications/quick_ai/models/qwen3_tq/.../qwen3_tq_embedding.h`: `#include "../qwen3/causallm_qwen3_embedding.h"` — please verify this path still exists after the rename; if the qwen3 headers were renamed this is a build break.
- `model_config.cpp` now reports `model_type = "quick_ai"` instead of `"CausalLM"` — is that an intentional API/behavior change for downstream consumers? If yes, call it out in the PR description.

## 6. YOLOv7 app bugs

`Applications/quick_ai/models/YOLOv7/jni/main.cpp`:
- Default `RES_DIR` is hardcoded to a personal path: `/home/seungbaek/Downloads/video/detector/res`. Please make it a required argument or a repo-relative default.
- In `loadImageLetterbox`: `for (int dx = 0; dy < nw; ++dx)` — the condition uses `dy` instead of `dx`. For wide images this is an infinite loop / out-of-bounds write. Needs a fix and ideally a test.

## 7. CED transformer concerns

- `runAudioFile` computes the front-end for **all** windows up front into one big buffer, so peak memory scales with clip length (the code comment admits this). For an on-device framework this should stream per window; at minimum, document the bound and gate behind a chunk size.
- Window 0 is inferred through a separate special-cased path and then windows 1..N in a loop — the duplicated code paths will drift; please unify.
- The manual NaN detection via bit-testing (because `-ffast-math` breaks `std::isnan`) is fragile — consider compiling this TU without fast-math, or centralizing the check.
- There's a redundant copy from the per-window front-end buffer into a contiguous chunk before inference.

## 8. New layers (`xi_pooling`, `rtmcc_gau`, `rtmcc_head`)

- `xi_pooling::forwarding` unconditionally delegates to `incremental_forwarding(..., false)` and ignores `from`/`to`; input is read via `getData<float>()` with no dtype guard — if the input ever arrives as QINT8-resident (the whole point of this PR), this silently misreads memory. Please assert the dtype or implement the quantized path. Same concern applies to the RTMCC layers' assumptions about the appended scale layout.
- These are inference-only FP32 islands — fine, but please state that explicitly in each header doc.

## 9. Backend correctness findings from @jijoongmoon

Please make sure these are addressed or tracked before merge (they affect more than this PR):
- `gelu_v2` runs twice (NEON result overwritten by scalar) in `arm_compute_backend.cpp` — missing `#else`, affects aarch64 too.
- `copy_fp32_u16` numeric-cast bug on the ARM fallback path → NaN in the ViT KV cache.
- `gemmPerChannelA8` accumulator folding per 32 values.
- Non-monotonic thread scaling.

## 10. Minor

- New files carry 2024/2025 copyright years — please use the current year.
- Stale `CausalLM` mentions in comments (e.g. `rtmcc_gau.cpp`), and the pre-existing `casuallm_` typos in meson variable names are being propagated rather than fixed — a rename PR is the right time to fix them.
- `timm_vit.h`: `attention_mask` allocated as `Float` though logically boolean — doubles memory; consider bool/uint8.
- `qwen3_tq` comment references `models/qwen3_tq/ji/` (typo for `jni/`).

## What's great

- The verification story (PyTorch parity numbers, ONNX Runtime keypoint parity, on-device latency/memory measurements) is exactly what a quantization PR should ship with.
- The `rms_norm` forwarding fix.
- The staged S0–S5 plan with accuracy gates in `W8A8_DESIGN.md` — please keep that discipline as the split-out PRs land.

Happy to re-review quickly once this is split, rebased, and the build guard is restored.

PR #4293, opened 2026-08-25
[CausalLM] Bound RoPE LUTs to runtime context

jayden0701:

<details><summary>d96ab492 [CausalLM] Bound RoPE LUTs to runtime context</summary><br />

[CausalLM] Bound RoPE LUTs to runtime context

Use the runtime KV-cache capacity instead of the model context length when building RoPE lookup tables.

Co-authored-by: Codex <noreply@openai.com>

**Self evaluation:**

</details>

### Summary

- [CausalLM] Bound RoPE LUTs to runtime context: RoPE lookup tables (cos/sin
  freqs) were previously built using `max_position_embeddings` from the model
  config, which can be far larger than the actual runtime KV-cache capacity
  (`max_timestep`). This caused unnecessary memory allocation for LUT entries
  that are never accessed at runtime. This commit changes the LUT size to be
  bounded by the runtime `max_timestep`, and removes the now-unused
  `max_position_embeddings` member from `MHACoreLayer`. The double-checked
  locking in `apply_rotary_emb_tensor_v2` is also hardened to re-check the LUT
  size so that a smaller table is rebuilt when the runtime capacity grows. The
  gemma4 unit-test fixture is updated to set `max_position_embeddings` to a
  deliberately large value (131072) so that the runtime-bounded RoPE LUT path
  is covered.

#### Theoretical memory savings

For Gemma4 (`max_position_embeddings = 131,072`, `head_dim = 256`, runtime
`max_seq_len = 2,048`), the RoPE LUT stores cos/sin tables of size
`seq_len × head_dim` per type (FP32 / FP16):

| | Before (max_position_embeddings) | After (max_timestep) | Savings |
|---|---|---|---|
| FP32 LUT | 2 × 131,072 × 256 × 4 B = **256 MB** | 2 × 2,048 × 256 × 4 B = **4 MB** | **~252 MB** |
| FP16 LUT | 2 × 131,072 × 256 × 2 B = **128 MB** | 2 × 2,048 × 256 × 2 B = **2 MB** | **~126 MB** |

#### Measured performance (Gemma4, S26U, 8 threads. prefill 1027 tokens + generation 296 tokens)

| Metric | Before | After | Change |
|---|---|---|---|
| Peak memory | 4,316,444 KB | 3,355,232 KB | **−961,212 KB (≈ 938 MB, −22.3%)** |
| Prefill TPS | 492.094 | 525.858 | **+6.9%** |
| Prefill time | 2,087 ms | 1,953 ms | **−134 ms (−6.4%)** |
| Generation TPS | 27.5246 | 27.8562 | **+1.2%** |
| Generation time | 10,754 ms | 10,626 ms | **−128 ms (−1.2%)** |

The measured memory reduction (~938 MB) exceeds the theoretical LUT savings
(~252 MB FP32) because the `std::vector<std::vector<float>>` structure incurs
per-vector allocation overhead (metadata + malloc headers) for each of the
131,072 rows, which is eliminated when the table is sized to the runtime
capacity.

### DIFF
--- a/Applications/CausalLM/layers/mha_core.cpp
+++ b/Applications/CausalLM/layers/mha_core.cpp
@@ -154,15 +154,13 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
   const unsigned int max_timestep =
     std::get<nntrainer::props::MaxTimestep>(mha_core_props).get();
 
-  /** max position embeddings */
-  max_position_embeddings =
-    std::get<props::MaxPositionEmbeddings>(mha_core_props).get();
-
   /** local window size */
   local_window_size = std::get<props::SlidingWindow>(mha_core_props).get();
 
   /** use rope */
   use_rope = std::get<props::UseRope>(mha_core_props).get();
+  if (use_rope)
+    checkMaxTimestepBound(max_timestep);
 
   /** attention scaling computation */
   rope_scaling_type = std::get<props::RopeScalingType>(mha_core_props).get();
@@ -864,9 +862,19 @@ void MHACoreLayer::one_batch_incremental_forwarding(
 
 /************************************************************** */
 
+void MHACoreLayer::checkMaxTimestepBound(unsigned int max_timestep) {
+  const unsigned int max_position_embeddings =
+    std::get<props::MaxPositionEmbeddings>(mha_core_props).get();
+  NNTR_THROW_IF(max_timestep > max_position_embeddings, std::invalid_argument)
+    << "max_timestep (" << max_timestep << ") exceeds max_position_embeddings ("
+    << max_position_embeddings
+    << "); RoPE positions would extrapolate beyond the model's trained "
+       "context length";
+}
+
 /**
  * @brief rotary embedding-related member function
- * @note seq_len -> max_position_embeddings
+ * @note seq_len is the runtime KV-cache capacity (max_timestep)
  */
 void MHACoreLayer::precompute_freqs(int head_dim, unsigned int seq_len,
                                     float theta, bool is_fp16) {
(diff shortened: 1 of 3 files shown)


Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -154,10 +154,6 @@ void MHACoreLayer::finalize(nntrainer::InitLayerContext &context) {
   const unsigned int max_timestep =
     std::get<nntrainer::props::MaxTimestep>(mha_core_props).get();
 
-  /** max position embeddings */
-  max_position_embeddings =
-    std::get<props::MaxPositionEmbeddings>(mha_core_props).get();
-
```
jaemini-shin, line 160: We still need this variable for checking `max_time <= max_position_embeddings` which is missing in this PR.

Applications/CausalLM/layers/mha_core.cpp
```diff
@@ -1086,11 +1082,15 @@ void MHACoreLayer::apply_rotary_emb_tensor_v2(nntrainer::Tensor &in,
   unsigned int max_timestep =
     std::get<nntrainer::props::MaxTimestep>(mha_core_props).get();
 
+  // RoPE positions are indexed by the runtime KV-cache position. The model's
+  // max_position_embeddings can be much larger than the configured execution
+  // length, so using it here would allocate unused LUT entries.
+
   if (in.getDataType() == ml::train::TensorDim::DataType::FP32) {
-    if (freqs_fp32 == nullptr) {
+    if (freqs_fp32 == nullptr || freqs_fp32->cos.size() < max_timestep) {
```
jaemini-shin, line 1102: Is it necessary to double check the condition?
It looks like it's impossible to have less size than `max_timestep` for non null `freqs_fp32`, since this is the only path to generate positional embedding,

Applications/CausalLM/layers/mha_core.cpp
jayden0701: So you mean we should add `max_time <= max_position_embeddings` as a check?

Applications/CausalLM/layers/mha_core.cpp
jayden0701: For now, it is correct, but `updateTensorsByInputDimensions()` can change `max_timestep` on runtime.
So I added this for future scenarios where `max_timestep `might change.

Applications/CausalLM/layers/mha_core.cpp
jaemini-shin: > So you mean we should add `max_time <= max_position_embeddings` as a check?

IMO, at least it'd be better to leave a warning message. `max_position_embeddings` means the maximum length that this model ensures. Of course it is not likely to have such an extreme case, but not keeping any guard looks inappropriate to me.

Applications/CausalLM/layers/mha_core.cpp
jaemini-shin: Oh I understood. Is there any additional plan for computing rope cache? If not, let me work on optimizing rope computation. It looks like there are many points to be optimized!

Applications/CausalLM/layers/mha_core.cpp
jayden0701: Good opinion! I'll update it accordingly

Applications/CausalLM/layers/mha_core.cpp
jayden0701: There is no plan for now. I'll look forward to ur optimizations :)

jaemini-shin (commented):
Could you please check formatting?

Applications/CausalLM/layers/mha_core.cpp
jayden0701: @jaemini-shin I applied max_time <= max_position_embeddings. could you check?

PR #4294, opened 2026-08-25
[build] Fix armv9.2-a build failure

jaemini-shin:

<details><summary>[build] Fix armv9.2-a build failure</summary><br />

The sve kernel files in nntrainer were for Vector Length(VL) 256, where
android arm cores that support sve are 128. This mismatch results both
wrong result and also build failure when figuring meson options.

Let's remove these unused files and fix build failure.

</details>

### Summary
- sve kernel files cannot be used in android devices.
  - ggml, kleidiai kernel files assumes vector length 256
    - The original codes have `VL == 256` guard, but they are not in nntrainer's code.
  - android arm cores use vector length 128
- sve kernels in kleidiai causes armv9.2-a build failure
- Let's drop these inappropriate sve files and fix build issue


dlwlzzero (approved):
LGTM. I checked that nothing in the tree still references the removed SVE files or symbols, and that `nntr_ggml_impl_neon.cpp` exports every function the removed `nntr_ggml_impl_sve.cpp` did, so armv9.2-a builds will link fine.

Two small optional follow-ups:
1. `meson.build` still adds `-DENABLE_SVE2=1` for armv9.2-a, but nothing uses `ENABLE_SVE2` after this PR. Could be dropped.
2. `kai_interface/kai/kai_common_sve_asm.S` is not built anywhere and has no callers (predates this PR). Could be removed as part of this cleanup.

